PR #1951, opened 2016-07-16
Remove allocations caused by string.Format in GetFullName method

dVakulen:
String format in SafeTimer's method was causing memory allocations on each call, replaced with raw strings. 

timeline64_2016-07-16_18-28-33

### DIFF
--- a/src/Orleans/Timers/SafeTimerBase.cs
+++ b/src/Orleans/Timers/SafeTimerBase.cs
@@ -118,15 +118,12 @@ namespace Orleans.Runtime
         private string GetFullName()
         {
             // the type information is really useless and just too long. 
-            string name;
             if (syncCallbackFunc != null)
-                name = "sync";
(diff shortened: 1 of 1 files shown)


gabikliot:
Good catch, although I think it was printing it only in verbose logging, which is not by default on in prob.

dVakulen:
It was called on every timer creation at this line: https://github.com/dotnet/orleans/blob/master/src/Orleans/Timers/SafeTimerBase.cs#L70

gabikliot:
Ohh, in that case it's worth optimizing and not even allocating the timer unless if it's is verbose. Or having it static. No need for timer per msg.

gabikliot:
Yes, better have it static and inline the sync async string in the printing statement.

gabikliot:
Could be in general a good idea to find out what exactly we allocate per msg everywhere and see if we can simply avoid allocating it.

PR #1946, opened 2016-07-15
[WIP] Messages pooling

dVakulen:
Related to https://github.com/dotnet/orleans/issues/1910

PR #1084, opened 2015-11-27
Reduce number of allocations during grain method call

dVakulen:
Currently on each grain method call string with debug context is being [allocated](https://github.com/dotnet/orleans/blob/master/src/Orleans/Runtime/GrainReference.cs#L354). 
Sample run of 28k grain methods calls.
Before fix:
image
After:

PR #1952, opened 2016-07-17
Remove debug context string allocations

dVakulen:
Debug context string allocations is one of the major memory traffic creators:

timeline64_2016-07-17_15-44-24

Adding of interning in https://github.com/dotnet/orleans/pull/1084 only partially mitigated the issue. 

Caching of this string will not only reduce memory pressure, but also reduce CPU time spent in this method (according to benchmarking of composing string every time vs getting it from <code>ConcurrentDictionary</code> - the latter shows 3x better performance) 

Profile results after change:

timeline64_2016-07-17_16-01-46

(The newly appeared dictionary initialize allocations on second result is due to first profile snapshot was created on build with https://github.com/dotnet/orleans/pull/1946, and the second one - without)

### DIFF
--- a/src/Orleans/Runtime/GrainReference.cs
+++ b/src/Orleans/Runtime/GrainReference.cs
@@ -330,14 +330,7 @@ namespace Orleans.Runtime
                 }
                 else
                 {
-                    int hash = 17;
-                    unchecked
-                    {
-                        hash = hash * 23 + InterfaceName.GetHashCode();
-                        hash = hash * 23 + InterfaceId;
-                        hash = hash * 23 + request.MethodId;
-                    }
-
+                    var hash = InterfaceId ^ request.MethodId;
                     if (!debugContexts.TryGetValue(hash, out debugContext))
                     {
                         debugContext = GetDebugContext(this.InterfaceName, GetMethodName(this.InterfaceId, request.MethodId), request.Arguments);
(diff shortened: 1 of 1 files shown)


src/Orleans/Runtime/GrainReference.cs
```diff
@@ -320,7 +324,26 @@ private Task<object> InvokeMethod_Impl(InvokeMethodRequest request, string debug
         {
             if (debugContext == null && USE_DEBUG_CONTEXT)
             {
-                debugContext = string.Intern(GetDebugContext(this.InterfaceName, GetMethodName(this.InterfaceId, request.MethodId), request.Arguments));
+                if (USE_DEBUG_CONTEXT_PARAMS)
```
dVakulen: This variable is turned off almost always.

gabikliot:
For hash key, you can just XOR methmet I'd with interface id. Both are already pseudo random hashes of names.

src/Orleans/Runtime/GrainReference.cs
gabikliot: Right. But why do u need a special handling for it? Just do the same as for regular debug context.

dVakulen: Because parameters can vary a lot, so there's no sense in caching resulting strings

gabikliot:
Lgtm.

PR #1953, opened 2016-07-17
Remove timer's logger allocations

dVakulen:
https://github.com/dotnet/orleans/pull/1951#issuecomment-233183132


gabikliot:
I think you can just keep the name logger (instead of currentLogger) to minimize churn, but otherwise looks good.

PR #1957, opened 2016-07-18
Add note about WeakReference

ReubenBond:
`CreateObjectReference` uses a `WeakReference` for observers, so users should ensure they hold their own reference to each client object.

Unless I'm missing something.


gabikliot:
Yes, you are correct.

PR #1958, opened 2016-07-18
Remove ActivationState enum boxing

dVakulen:
Due to virtual method calls on value type <code>ActivationState</code> instance is being boxed. 

dotmemory64_2016-07-18_18-28-49

(numbers per 10k grain method calls)


sergeybykov:
Good catch. Thank you, @dVakulen!

PR #1959, opened 2016-07-18
Fix Azure Web Sample page that was out-of-date

benjaminpetit:


sergeybykov:
Welcome to the club, @benjaminpetit!

Issue #1083, opened 2015-11-26
Dnx Console Host


elmortega:
Orleans version 1.1.0-beta1

I'm trying to host the silo in a Dnx Console App this way:    

```
    AppDomain hostDomain = AppDomain.CreateDomain("OrleansHost", null,
                 new AppDomainSetup()
                 {
                     AppDomainInitializer = InitSilo
                 });
```

But I get a System.IO.FileNotFoundException.

If a I run it in a normal console app works fine

sergeybykov:
I suspect you need to add pre-released `Microsoft.Framework.DependencyInjection` and `Microsoft.Framework.DependencyInjection.Abstractions` nugets to the project. We fixed that for -beta2. Which file does the `FileNotFoundException` complain about?

elmortega:
'OMS.ConsoleHost' is the Console project name

No se puede cargar el archivo o ensamblado 'OMS.ConsoleHost, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' ni una de sus dependencias. El sistema no puede encontrar el archivo especificado.

KodrAus:
I had this issue running in DNX too. You won't be able to run Orleans in DNX just yet (unless it's changed in the last few days), it's an active work in progress.

Have a look at #368 or #912

elmortega:
Just installed the Orleans version 1.1.0-beta2, doesn't work either

jdom:
yeah, as @KodrAus mentioned, Orleans is not supported on DNX just yet. BTW, are you targeting CoreCLR or full .NET?

elmortega:
@jdom full .NET

elmortega:
So that means that I can´t use orleans with Asp.Net 5 ?

sergeybykov:
@elmortega You should be able to use Orleans so long as you target full .NET I think. Try the final v1.1.0 we just published.

elmortega:
in my project.json:

"frameworks": {
    "dnx451": {
      "dependencies": {
        "Microsoft.Orleans.OrleansRuntime": "1.1.0"
      }
    }

elmortega:
I´m trying to initialize the GrainClient in a Asp.Net 5 app

            var clientconfig = new Orleans.Runtime.Configuration.ClientConfiguration();
            clientconfig.Gateways.Add(new IPEndPoint(IPAddress.Loopback, 30000));
            GrainClient.Initialize(clientconfig);

but I get...

[...]

PR #1961, opened 2016-07-18
Add Xproj and project.json based version of Adventure example

TsengSR:
As requested in https://github.com/dotnet/orleans/issues/1083 here's a dotnet-cli based version of the Adventure sample. 

It's mainly unchanged except for adding project.json and xproj files as well as changing the `System.Web.Script.Serialization.JavaScriptSerializer` to `Json.NET` which is also used in ASP.NET Core as `System.Web.Script.Serialization.JavaScriptSerializer` wasn't ported to .NET Core.

I had to create a new folder for it. Attempting to add the json/xproj files side-by-side in the existing Adventure folder resulted in being unable to restore nuget packages when opening the old cproj based solution.


sergeybykov:
This change puts a copy of the source files from Samples/Adventure to Samples/AdventureDotNetCli, which is obvious code duplication. Can we instead have the two sets of solution/project files in separate folders, but have them point to a single set of .cs and other files?

TsengSR:
One could maybe put the *.csproj files outside of their respective folders and put the *.xproj/project.json inside the project folders. Doesn't work other way around as *.xproj/project.json don't have the option to add/remove a single file from solution. 

One can exclude them for compilation and publishing via `project.json`, Gonna try that when I am back at home

TsengSR:
I've removed the *.cs files from the Adventure folder and referenced them from the AdventureDotNet folder. 

Still feels a bit "dirty", but there don't seem to be any other way. There are some workarounds to prevent VS restoring packages from `packages.xml` and only use `project.json`, but this would likely break support for VS 2013 where the .NET Core tools aren't available.

Samples/AdventureDotNetCli/AdventureGrainInterfaces/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Reflection;
```
shayhatsor: There's no need for AssemblyInfo.cs files in .net core projects.

TsengSR:
Removed the AssemblyInfo.cs files from th DotNetCli project

sergeybykov:
Thank you @TsengSR!

PR #1963, opened 2016-07-19
Eliminate closure allocations

dVakulen:
On each <code>IsOrleansShallowCopyable</code> call closure was allocated because of delegate that captured <code>Type</code> variable. 

dotmemory64_2016-07-19_20-20-44

### DIFF
--- a/src/Orleans/Serialization/TypeUtilities.cs
+++ b/src/Orleans/Serialization/TypeUtilities.cs
@@ -82,7 +82,7 @@ namespace Orleans.Serialization
 
         private static bool IsTypeFieldsShallowCopyable(TypeInfo typeInfo)
         {
-            return typeInfo.GetFields().All(f => !(f.FieldType == typeInfo) && IsOrleansShallowCopyable(f.FieldType));
+            return typeInfo.GetFields().All(f => !(f.FieldType.IsEquivalentTo(typeInfo)) && IsOrleansShallowCopyable(f.FieldType));
         }
 
         internal static bool IsSpecializationOf(this Type t, Type match)
--- a/src/Orleans/Serialization/TypeUtilities.cs
+++ b/src/Orleans/Serialization/TypeUtilities.cs
@@ -71,13 +71,18 @@ namespace Orleans.Serialization
 
             if (typeInfo.IsValueType && !typeInfo.IsGenericType && !typeInfo.IsGenericTypeDefinition)
             {
-                result = typeInfo.GetFields().All(f => !(f.FieldType == t) && IsOrleansShallowCopyable(f.FieldType));
+                result = IsTypeFieldsShallowCopyable(typeInfo);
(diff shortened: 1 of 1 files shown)


src/Orleans/Serialization/TypeUtilities.cs
```diff
@@ -80,6 +80,11 @@ internal static bool IsOrleansShallowCopyable(this Type t)
             return false;
         }
 
+        private static bool IsTypeFieldsShallowCopyable(TypeInfo typeInfo)
+        {
+            return typeInfo.GetFields().All(f => !(f.FieldType == typeInfo) && IsOrleansShallowCopyable(f.FieldType));
```
gabikliot: I think it is a wrong comparison now.
t used to be Type, now you compare f.FieldType to TypeInfo.

dVakulen: According to tests https://github.com/dotnet/corefx/blob/master/src/System.Reflection/tests/TypeInfo/TypeInfo_IsEquivalentTo.cs#L17 this is legit comparison, but I can change if it results iт confusion.

sergeybykov: I think `IsEquivalentTo()` would be a bit better readable.

dVakulen: Agreed.

sergeybykov:
Thank you, @dVakulen!

src/Orleans/Serialization/TypeUtilities.cs
```diff
@@ -80,6 +80,11 @@ internal static bool IsOrleansShallowCopyable(this Type t)
             return false;
         }
 
+        private static bool IsTypeFieldsShallowCopyable(TypeInfo typeInfo)
+        {
+            return typeInfo.GetFields().All(f => !(f.FieldType.IsEquivalentTo(typeInfo)) && IsOrleansShallowCopyable(f.FieldType));
```
jdom: this is not compatible with .NET core, since TypeInfo does not inherit from Type nor there is an extension method. We can use typeInfo.AsType() in that case

jdom: Sorry, to be completely accurate, it must be changed to:
```
typeInfo.GetFields().All(f => !f.FieldType.GetTypeInfo().IsEquivalentTo(typeInfo.AsType()) && IsOrleansShallowCopyable(f.FieldType))
```
which might default the purpose of this PR. The point is that it's currently not portable and we'll have to change it

dVakulen: According to tests in the corefx repository: https://github.com/dotnet/corefx/blob/master/src/System.Reflection/tests/TypeInfo/TypeInfo_IsEquivalentTo.cs#L17 
this is compatible with .NET core.

Issue #1863, opened 2016-06-20
Parallelize test run (extending xUnit)


jdom:
Hi, there's something I wanted to do for a long time, but never find the actual time to do so. Maybe somebody already has the expertise and would like to help out.
Today all of our tests run serially and take a long time. Ideally, once we start an in-memory test cluster, we should be able to run all of the tests that use that test cluster in parallel. Nevertheless, we cannot have 2 test clusters running tests at the same time, mainly because GrainClient relies on statics, so the tests would conflict with each other.

xUnit by default provides the ability to [run tests in different test collections in parallel](https://xunit.github.io/docs/running-tests-in-parallel.html) (effectively run all tests within 1 collection serially, but parallelize with running multiple collections concurrently). This is the exact opposite of what we need, as each of our collections are the ones that set up and tear down the cluster and client. So we need to run each test collection serially, but parallelize within it.

The work to support this test execution is not completely trivial, although there are some samples in xUnit's github repo that show how to make changes to it. I have not made an assestment of how much work that is, but here are the hints I got on where to look at (from a Slack chat with them):

> jdom [2:03 PM] is there a way to invert the condition for how to run tests in parallel? I would like to have all the tests in a certain fixture (being that a collection or class fixture) to run in parallel between each other, and once that entire fixture ran, then move to the next collection or class fixture
> bradwilson [3:25 PM] Sorry, there's no simple way to do this today (you could define your own unit test framework, but that's not trivial).
> jdom [3:30 PM] @bradwilson: ok. By that should I assume that the current extensibility points in xunit are not enough to accomplish this parallelization? I haven't tried yet, but if it is possible then we might give it a shot
> bradwilson [3:37 PM] @jdom No, the extensibility points are sufficient, but you would be extending an awful lot of infrastructure to make it happy.
> For example, take a look at what this does: https://github.com/xunit/samples.xunit/tree/master/ObservationExample

Another option is that if we could force each collection to be ran in a separate AppDomain, then at that point we could leverage xUnit's built in parallelization, given that each client will not be conflicting with other collections' usages of GrainClient (and its statics).

amccool:
Is this the true issue re: #467 ?

> mainly because GrainClient relies on statics, so the tests would conflict with each other.

but does this imply that pure grain code should execute cleanly in unit tests without a limit on parallelization?

jdom:
[...]

PR #1966, opened 2016-07-20
Add dynamic adding / removing stream providers functionality.

JingqiaoFu:
This change will add the functionality of dynamically adding / removing stream providers on silo at runtime, so that each silo doesn't have to be restarted to pick up newly added stream providers in the config file.

this feature can programmatically manipulate configuration objects in the silo instance. We do not address the issue where this new functionality may cause differences between the configuration object and the config files. Any dynamic change made at runtime will NOT be recorded back into config files. Therefore, if a silo is restarted, the changes made last time will be lost.  

User experience:  

1. User creates a stream provider configuration object, which contains a list of specifications of all the stream providers that needs to be loaded. 

2. User calls member function UpdateStreamProviders(streamProviderConfig) of management grain or silo itself, passes in the configuration object as the parameter. The stream providers in target silos will be updated based on the differences between their loaded stream providers and the list of providers in the object: If a provider in the configuration object already exists in a silo, it will be kept as is; if a provider in the configuration object does not exist in a silo, it will be loaded and started; if a provider that exists in a silo but is not in the configuration object, it will be stopped and removed from the silo.


jdom:
@dotnet-bot test this please
reason: removed dependency on MSTest since the PR was created

src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProvider.cs
```diff
@@ -8,12 +8,14 @@ namespace Orleans.Providers.Streams.SimpleMessageStream
     public class SimpleMessageStreamProvider : IInternalStreamProvider
     {
         public string                       Name { get; private set; }
+        private ProviderState               State { get; set; }
```
jason-bragg: Why track this separately, why not just ask the stateManager for the state?

src/Orleans/Providers/IOrleansProvider.cs
```diff
@@ -160,4 +161,23 @@ public static TimeSpan GetTimeSpanProperty(this IProviderConfiguration config, s
             return config.Properties.TryGetValue(key, out s) ? TimeSpan.Parse(s) : settingDefault;
         }
     }
+
+    /// <summary>
+    /// Exception thrown whenever a provider has failed to be initialized.
+    /// </summary>
+    [Serializable]
+    public class ProviderInitializationException : OrleansException
```
gabikliot: please put all the new exception in one file, better in the same file where all the other  OrleansExceptions are now.

jason-bragg: We added it here because it's an addition to the public surface of the provider model.  On initialization failure, providers are expected to throw this exception.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -487,13 +491,13 @@ private void DoStart()
             if (logger.IsVerbose) { logger.Verbose("Storage provider manager created successfully."); }
 
             // Load and init stream providers before silo becomes active
-            var siloStreamProviderManager = (StreamProviderManager) grainRuntime.StreamProviderManager;
+            var streamProviderManager = (StreamProviderManager)grainRuntime.StreamProviderManager;
```
gabikliot: why do you need to rename that? Lets keep unrelated churn low. If you want to refactor, unrelated to the change, please submit it is a separate PR.

test/TesterInternal/StreamingTests/DynamicStreamProviderConfigurationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+                DefaultStreamProviderSettings.Add(PersistentStreamProviderConfig.STREAM_PUBSUB_TYPE, StreamPubSubType.ImplicitOnly.ToString());
+
+                return new TestCluster(options);
+            }
+        }
+
+        public async void Dispose()
+        {
+            await RemoveAllProviders();
+            fixture.DefaultStreamProviderSettings.Remove(FailureInjectionStreamProvider.FailureInjectionModeString);
```
jdom: This is odd. Please do no update the fixture itself, as it is mutable state across tests and would break as soon as we add some test parallelization (#1863)

test/TesterInternal/StreamingTests/DynamicStreamProviderConfigurationTests.cs
```diff
@@ -0,0 +1,327 @@
...
+        [Fact, TestCategory("Functional"), TestCategory("Streaming"), TestCategory("Providers")]
+        public async Task DynamicStreamProviderConfigurationTests_AddProvidersAndThrowExceptionOnInitialization()
+        {
+            //Making sure the initial provider list is empty.
+            IEnumerable<String> providerNames = fixture.HostedCluster.Primary.Silo.TestHook.GetStreamProviderNames();
+            Assert.Equal(providerNames.Count(), 0);
+
+            bool exceptionThrown = false;
+            fixture.DefaultStreamProviderSettings.Remove(FailureInjectionStreamProvider.FailureInjectionModeString);
+            fixture.DefaultStreamProviderSettings.Add(FailureInjectionStreamProvider.FailureInjectionModeString,
```
jdom: Same here, please never update the shared fixture on a test by test basis

src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProvider.cs
JingqiaoFu: Good catch. This is obsolete code that I missed to delete.

test/TesterInternal/StreamingTests/DynamicStreamProviderConfigurationTests.cs
```diff
@@ -0,0 +1,328 @@
...
+                DefaultStreamProviderSettings.Add(PersistentStreamProviderConfig.STREAM_PUBSUB_TYPE, StreamPubSubType.ImplicitOnly.ToString());
+
+                return new TestCluster(options);
+            }
+        }
+
+        public async void Dispose()
+        {
+            await RemoveAllProviders();
+            //fixture.DefaultStreamProviderSettings.Remove(FailureInjectionStreamProvider.FailureInjectionModeString);
```
jdom: latest tests look good, just remove this comment

jason-bragg:
Looks good.
Please rebase and squash, and I'll merge it.

Also did you sign the contributors license agreement?  PR is still marked cla-required.

test/Tester/TestStreamProviders/FailureInjectionStreamProvider.cs
```diff
@@ -0,0 +1,69 @@
+
+using System;
+using System.Collections.Generic;
+using System.Management.Automation.Language;
```
jdom: be careful with including namespaces that are not really used. This is currently harmless in .NET Framework, but it's breaking when we multi target to .NET Core (which we are doing in other branches). This is why we recently removed all superfluous using statements in #1985. We should be careful in new code not to add more.

Issue #1912, opened 2016-07-07
Issue using latest azure storage (>= 7.1.2 and even >= 7.0.0)


onionhammer:
Orleans seems incompatible with WindowsAzure.Storage 7.1.2 

```
at Orleans.Runtime.Host.AzureSilo.Start(ClusterConfiguration config, String deploymentId, String connectionString)
Exc level 1: Orleans.Runtime.OrleansException: Exception trying to create or connect to the Azure table: The remote server returned an error: (415) Unsupported Media Type.
at Orleans.AzureUtils.OrleansSiloInstanceManager.<GetManager>d__14.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\ClassLibraryCommon\Table\CloudTable.cs:line 1183
```

jdom:
Without having investigated, I wonder if it's related to this:
> Changes in 7.1.2 :
- All: Support for 2015-12-11 REST version. Please see our REST API documentation and blogs for information about the related added features. If you are using the Storage Emulator, please update to Emulator version 4.4.

From their [changelog](https://github.com/Azure/azure-storage-net/blob/master/changelog.txt)

jdom:
Ping @centur who also ran into the same issue

jdom:
Relevant information, from [Versioning for the Azure Storage Services](https://msdn.microsoft.com/en-us/library/azure/dd894041.aspx)
> Beginning with version 2015-12-11, the Atom feed is no longer supported as a payload format for Table service operations. Version 2015-12-11 and later versions support only JSON for the payload format.

It is related to they dropping support for that media type. We need to figure out how to address this

jkonecki:
I can confirm the issue. Workaround is to downgrade Storage nuget package to 7.0.0

benjaminpetit:
I submitted #1968 to support WindowsAzure.Storage 7.0+

Azure made a breaking change in the version 7.0 of their library by [removing an enum value from TablePayloadFormat](https://github.com/Azure/azure-storage-net/commit/b4699d38f6583a2cb4136586e687f04c7784fae8#diff-f2dcdff30895201b510b4347e27c46ed). As they don’t enforce enum order, they basically shifted all of them.

[...]

PR #1968, opened 2016-07-20
Update to WindowsAzure.Storage 7.0.0

benjaminpetit:
#1912


jdom:
LGTM.
Just to contextualize other reviewers, we now have a legit reason to update our dependency to 7.0.0, so no point on trying to hack around it to be forward compat (for example we could get the value of the enum through reflection, but we decided it's best to avoid it). Unless someone states why we shouldn't force people to upgrade to 7.0.0, I think it's fine.
Also, once we move closer to CoreCLR, we will have to bump this version even higher anyway (7.1.13-preview is the first one with support for .NETStandard)

jdom:
BTW, something went wrong with the PR... if you look at Jenkins it kind of looks like it's including a ton of commits that are already in master. Maybe a rebase that went wrong.

benjaminpetit:
I think it was due to the conflict, now it seems ok

PR #1970, opened 2016-07-21
Reworked documentation to use DocFX

richorama:
Please browse here to see the working site: http://richorama.github.io/orleans/

The `readme.md` contains instructions on how to build the docs.


sergeybykov:
Looks totally awesome! The search box is what we were missing all alone.

Do I understand the editing process correctly, that it involves three steps now: 1) make edits to .md files; 2) run `docfx src\docfx.json` to generate .html files; 3) commit updated .md and .html files?

jdom:
I love that this would also enable to generate documentation of our APIs from source code, and you can jump right into the source code in github from it.
For our current markdown docs, they look awesome! we'd need to figure out how to automate this though, as we can't expect docs PRs to include both the md and the html files from a local run of the docfx tool.

shayhatsor:
That's just great! Finally !!! @richorama, are there any pitfalls or gotchas you've encountered ?

richorama:
@sergeybykov yes, that's the correct process. It's unfortunate that we have to check in the generated HTML, but GitHub doesn't support DocFX, so GH-Pages just sees it as a static site.

@jdom could we install DocFX (xcopy the binaries) onto the build server, and then have it compile the docs? It would then need to do a commit if there is a diff.

To automate the API documentation we would probably need to co-locate the doc source along with the source code (probably placing the image files elsewhere) in the same branch. You can then push  the compiled HTML to the gh-pages branch using [`git subtree`](https://gist.github.com/cobyism/4730490). You would want to control this so you push on an orleans release.

@shayhatsor Yes, there are a few quirks:

1. No search box on the main page (this looks like a 'feature' of DocFX - I'll raise it with them).
1. The directory structure leads to a curious way the menu works, i.e. when you open 'advanced concepts', you then lose where you are in the context of the documentation. This would seem to work better if you put all your files in a single directory. Obviously, that makes it hard to organise. I'll raise this as well.
1. The documentation for DocFX is well, lacking (quite ironic).
1. There is some replication in the menu items between `toc.yml` files, which leads to more to maintain. I'll raise this too.

sergeybykov:
The process doesn't scare me. It's less clear to me how an auto-commit from a build server would work.

richorama:
...and another thing, DocFX seems to expect you to split your documentation into separate sections (i.e. tutorials, API reference, documentation etc...) which our docs didn't previously have, so I created this as a top level folder strcuture and organised our content accordingly. 

This means that many of the pages have now moved, so anyone deep linking into a page from an external website will probably get a 404. I'm not sure how to get around that...

richorama:
@sergeybykov yes, you might get an endless loop...

shayhatsor:
> This means that many of the pages have now moved, so anyone deep linking into a page from an external website will probably get a 404. I'm not sure how to get around that...

IMHO we can live with that.

sergeybykov:
I don't expect many links pointing to specific doc pages to be out there. Most just point to the main page. So 404s should be fine I think.

jdom:
As an idea (not sure about implementation details, but other experts might piggyback on) is that we can have a repo with all the markdown files, and when one of them is updated, that can kick off a Jenkins build that generates the html using docfx and pushes the resulting html to the gh-pages of that repo (or to the Orleans main repo if we still want it there).
Then if we want to start publishing Orleans API documentation also, we can have a git submodule in that repo that points to the Orleans source code. For this we probably don't want to generate HTML for every single PR that gets into master, probably just manually trigger an update of the submodule link for every nuget release.

richorama:
Yes, I think that would work.

richorama:
@jdom instead of a separate repo for the md files, you could just have a branch, I guess you don't want to end up with multiple repos to keep track of (i.e. issues and PRs appearing in the wrong places etc...)

shayhatsor:
@jdom said:
> For this we probably don't want to generate HTML for every single PR that gets into master, probably just manually trigger an update of the submodule link for every nuget release.

This reminds me that I think we should version the documentation, one branch for each release and one for master.

sergeybykov:
Looks like one has to run docfx.exe from a clone of the main repo. Otherwise, the links point to the fork.

I ran it in my fork, and I see changes in every html file:
`content="docfx 2.0.1.0"` -> `content="docfx 2.2.2.0"` which is understandable
and
`href="https://github.com/richorama/orleans/..."` -> `href="https://github.com/sergeybykov/orleans/..."`

Unless there is a way to force it to point to `https://github.com/dotnet/orleans/`.

richorama:
I have asked the question: https://github.com/dotnet/docfx/issues/531

galvesribeiro:
Wow! Didn't saw that! Awesome! Great work @richorama! :shipit:

shayhatsor:
@sergeybykov, @jdom. Is there something that blocks us from merging this PR? The current documentation is being changed all the time, which already caused merge conflicts. Even if we'll need to locally run the DocFX tool for now, or manually edit some files before pushing - I believe it's worth it.

richorama:
There is an update to docfx which addresses the problem with running against a fork. I'll set this up in the configuration in a day or so.

jdom:
We can't run this manually for every contribution, or ask people to run it, otherwise it'll cause more friction, not less, which is one of the goals of revamping documentation.
We should explore automating this, or the other options suggested, but let's not rush this in until we find consensus

veikkoeeva:
In addition to latency, should there be a mention Orleans remains stable even on high load and retains high throughput? I could imagine "low latency" means easily to some readers as being at odds with throughput, though here they are looked at different angles (e.g. Orleans has state in memory reducing latency and then semi-cooperative scheduling for throughput).

richorama:
The 'improve this doc' links now point to the orleans repo despite being generated from a fork. You'll need [version 2.3](https://github.com/dotnet/docfx/releases) or greater of docfx.

@jdom could we explore the idea of splitting the 'compiled' docs (html) and source (markdown) into separate branches, and using the CI server to maintain the `gh-pages` branch.

As a side note, I see that github now supports [using any branch as the docs branch](https://github.com/blog/2228-simpler-github-pages-publishing), so we're not limited to `gh-pages`.

galvesribeiro:
With the latest updates from GH, we now can have docs on master and push doc changes along with code. We should even put a "rule" to push new features only with docs.

dVakulen:
Would it be hard to create snapshot every release in order to support docs per version like http://akka.io/docs/ ?

richorama:
@dVakulen I don't think so, you just run DocFX against the repo at the given commit, and push the compiled site to gh-pages

richorama:
My concern about putting documentation in `master` is the binary files (i.e. images and pptx files etc...). The documenation source is currently 200MB. 

I'm sure these files could be hosted elsewhere, the question is where...

galvesribeiro:
+1 to have a diff then... In fact, have an Organization for Orleans would save us this problem and many others...

sergeybykov:
@jdom @richorama In my opinion lack of automation shouldn't block this PR because building docs will naturally pick up all changes to doc sources made after a previous build. If the fork issue is resolved, I think we can move forward and start by manually running docs build. It's just a single command to execute.

jdom:
Ok, I talked to Sergey. I think it would be nice to split the docs to its own repo (`dotnet/orleans-docs` or something like that).
Then we can have a branch for the source markdown (ie: master), and a gh-pages branch with the built html files. For now we can start running docFX manually and pushing to that other branch, and in the future we could automate it so whenever the source branch changes, a CI build can build and immediately push to gh-pages.
Sounds reasonable and doable @richorama?

richorama:
@jdom it does indeed sound doable. However, it will mean the URL for the docs will change to http://dotnet.github.io/orleans-docs  - so assuming we retire the existing docs, anybody following a link to the old site will get a 404.

It would be possible to set a redirect in JavaScript on the home page, but I don't think an HTTP redirect is possible.

galvesribeiro:
@richorama if the URL will change anyway, there is way to use custom domains on gh-pages?

richorama:
@galvesribeiro I believe you can add a `CNAME` to a gh pages site, [according to the docs](https://help.github.com/articles/using-a-custom-domain-with-github-pages/).

Just to clarify, just switching from jekyll to docfx won't break links to the docs home page, but it will break deeper links in the site.

galvesribeiro:
So basically, can we (/me looking to @sergeybykov  and .Net Foundation) get a domain and put on top of it? These changes shouldn't break the whole ranking on searches anymore

richorama:
@galvesribeiro yes, you could buy something like orleans.io and point it at the docs.

galvesribeiro:
Yes I can buy it no problem, but I mean, it would be good if the team and/or the foundation keep its administration. (just in case I decide to start a rock band and not work with tech anymore :guitar: :smile: )

note: orleans.io is unnavailable :(

sergeybykov:
I checked. Looks like the issue with links pointing to forks is resolved now.

Let's merge this as a first step, so that we immediately benefit from the better usable docs. We can move to a separate repo as a separate follow-up effort.

@richorama Could you rebase this please. There's a minor conflict with the recent updates to docs.

richorama:
@sergeybykov synced!

richorama:
merged again!

Advanced-Concepts/Two-Way-Client-Observers-old.md
```diff
@@ -1,57 +0,0 @@
----
-layout: page
-title: Two Way Client Observers
----
-{% include JB/setup %}
-
-**THE BELOW IS WRONG!!!!!!**
```
jdom: was this file removed on purpose or a merge conflict? Because I see the .md file is gone, but there is html documentation generated for this.

jdom:
The diff does not open entirely in github, so I'll add a few loose comments here (as I'm only seeing them when I review the branch locally)

jdom:
In the combined log, I see several added links to `https://github.com/richorama/orleans/blob/gh-pages`. Could they have been generated before you updated DocFX to point to dotnet/orleans? Or maybe the links in the source where just incorrect before and now the html is generated incorrectly too?

jdom:
`manifest.json` contains a local path called `C:/Users/Richard/`

jdom:
Oh, wow! I just finished reviewing locally, and I'm really impressed by the level of care you took to fix links, headers, etc. This is a tremendous contribution not just by the result, but by the effort required.
Thanks a lot.
Other than those minor comments, the rest of the diff looks outstanding.

jdom:
I tried building with the latest docfx (2.4.0.0), and I no longer see the *improve this doc* button. Do you know why that is?

richorama:
@jdom 

* The 'improve this doc' link has moved to the top of the menu on the right hand side (I didn't see it there either, until I looked at the [commit on the docfx repo](https://github.com/dotnet/docfx/commit/78db9a712a6ae02a3d117d19082f2e6f8c6f34c5#diff-aa8cbb85472327adfee31e90af121c5e)). We customise the template to put a more obvious button in, but I'm wary of doing too much customisation, as I think we'd rather be on the defaults.
* The `manifest.json` is generated by docfx. I'm not sure of it's purpose, but it doesn't seem to be used. I wonder if we should exclude it from the repo?
* There were a few orphaned HTML pages which I removed in f30684e, these contained 'improve this doc' links still pointing to my fork.

jdom:
- I built locally, and still can't find the *Improve this doc* button (or an alternative for it). Nevertheless, I think we can fix this after I merge, as I don't want you to have to resolve merge conflicts once again.
- Not sure about the manifest.json, it looks like it might be safe to exclude. I did not see it in the gitignore file for core-docs repo, nor I see it committed either, maybe I'm looking at the wrong repo?

jdom:
Wiiii, merged! The documentation is now live at http://dotnet.github.io/orleans/ !!!!
Thanks a lot @richorama, this looks so much nicer

sergeybykov:
Thank you so much, @richorama! The docs look much more professional now.

As someone said, the most important contribution one can make to an OSS project is by improving its documentation. :-)

shayhatsor:
WOW! @sergeybykov, I can't believe it, I was actually just about to write that this contribution is one of the most important ones 😄

galvesribeiro:
Yay! Great work @richorama! :dancer:

gabikliot:
I think some of the presentations from https://github.com/dotnet/orleans/tree/gh-pages/Presentations got lost during the migration. I now only see one file there, but there used to be more. For example, it had http://dotnet.github.io/orleans/Presentations/Balancing Techniques in Orleans.pptx

sergeybykov:
I opened https://github.com/OrleansContrib/meetups/issues/13 last week for the same issue.

richorama:
@sergeybykov @gabikliot I'll get to it tomorrow :¬)

PR #1986, opened 2016-07-26
Replace [MethodImpl(MethodImplOptions.Synchronized)] with lock(this)

jdom:
Since that API with the attribute is not available in .NET Core, I replaced it with the exact same logic as what the attribute currently does.


jason-bragg:
Looks good.
Will merge after update

PR #1987, opened 2016-07-26
Add a function that checks the connection string for use during initialization

malavika1:


src/OrleansAzureUtils/Hosting/AzureSilo.cs
```diff
@@ -61,13 +71,43 @@ public AzureSilo()
...
+
+        internal static ClusterConfiguration DefaultConfiguration(IServiceRuntimeWrapper serviceRuntimeWrapper)
+        {
             var config = new ClusterConfiguration();
 
             config.Globals.LivenessType = GlobalConfiguration.LivenessProviderType.AzureTable;
-            config.Globals.DeploymentId = AzureClient.GetDeploymentId();
-            config.Globals.DataConnectionString = AzureClient.GetDataConnectionString();
+            config.Globals.DeploymentId = serviceRuntimeWrapper.DeploymentId;
+            config.Globals.DataConnectionString = serviceRuntimeWrapper.GetConfigurationSettingValue(AzureConstants.DataConnectionConfigurationSettingName);
```
jason-bragg: plz make sure GetConfigurationSettingValue

src/OrleansAzureUtils/Hosting/AzureSilo.cs
```diff
@@ -61,13 +71,43 @@ public AzureSilo()
             logger = LogManager.GetLogger("OrleansAzureSilo", LoggerType.Runtime);
         }
 
+        public async Task<bool> ValidateConfiguration(ClusterConfiguration config)
+        {
+            if (config.Globals.LivenessType == GlobalConfiguration.LivenessProviderType.AzureTable)
+            {
+                string deploymentId = serviceRuntimeWrapper.DeploymentId;
+                string connectionString = serviceRuntimeWrapper.GetConfigurationSettingValue(DataConnectionConfigurationSettingName);
```
jason-bragg: Only get values from the runtime if these fields are not set in the configuration

src/OrleansAzureUtils/Hosting/AzureSilo.cs
```diff
@@ -61,13 +71,43 @@ public AzureSilo()
...
+
+                try
+                {
+                    var manager = await OrleansSiloInstanceManager.GetManager(deploymentId, connectionString);
```
jason-bragg: only get manager if siloInstanceManager is not set, otherwise use siloInstanceManager.

src/OrleansAzureUtils/Hosting/AzureSilo.cs
```diff
@@ -61,13 +71,43 @@ public AzureSilo()
...
+                    var instances = await manager.DumpSiloInstanceTable();
```
jason-bragg: Since we're reading the instances, we may as well log them under verbose.

src/OrleansAzureUtils/Hosting/AzureSilo.cs
```diff
@@ -61,13 +71,43 @@ public AzureSilo()
...
+                }
+                catch (Exception exc)
+                {
+                    var error = String.Format("The connection string is incorrect. Connecting to the storage table has failed with {0}",
```
jason-bragg: We don't know "The connection string is incorrect."  We just know the operation failed.

src/OrleansAzureUtils/OrleansAzureUtils.csproj
```diff
@@ -43,6 +43,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
+    <Reference Include="Microsoft.WindowsAzure.ServiceRuntime, Version=2.7.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL" />
```
jason-bragg: We don't want a hard dependency on the runtime.  What necessitated this change?

src/OrleansAzureUtils/Hosting/AzureSilo.cs
```diff
@@ -110,9 +110,16 @@ internal static ClusterConfiguration DefaultConfiguration(IServiceRuntimeWrapper
             {
                 config.Globals.DataConnectionString = serviceRuntimeWrapper.GetConfigurationSettingValue(AzureConstants.DataConnectionConfigurationSettingName);
             }
-            catch (RoleEnvironmentException)
+            catch (Exception exc)
             {
-                config.Globals.DataConnectionString = null;
+                if (exc.ToString().Contains("RoleEnvironmentException"))
```
jason-bragg: This is probably safe, but what we really want it to ensure the type, and ToString has more than that in it.
Suggets exc.GetType().Name.Contains(...)

jason-bragg:
Looks good.
Please rebase and squash, and I'll merge it.

PR #1990, opened 2016-07-27
PowerShell Client Module

galvesribeiro:
This initial implementation is part of a series of improvements on tooling.

Many people were asking on Gitter for a way to interact with grains from a console.

This initial push enables user to start and stop `GrainClient` and call any grain from Powershell.

Currently GrainObservers, Streams, interceptors are not implemented. This first step allow only direct calls to grain methods as in any other client that uses `GrainClient` does.

A page will be added to docs site on a separated PR explaining how to use it and more work will come to support other client scenarios.

### DIFF
--- a/src/OrleansPSUtils/Orleans.psd1
+++ b/src/OrleansPSUtils/Orleans.psd1
@@ -1,0 +1,24 @@
+﻿@{
+	RootModule = 'OrleansPSUtils.dll'
+	ModuleVersion = '$version$'
+	GUID = 'afbcf490-ca4b-4039-9ffe-c6fec8a4f71b'
+	Author = 'Microsoft Orleans Team'
+	CompanyName = 'Microsoft'
+	Copyright = 'Copyright Microsoft 2016'
+	Description = 'Orleans Client Powershell module'
+	PowerShellVersion = '3.0'
+	CLRVersion = '4.0'
+	FunctionsToExport = '*'
+	VariablesToExport = '*'
+	AliasesToExport = '*'
+
+  PrivateData = @{
+    PSData = @{
+        Tags = @("Orleans", "Cloud-Computing", "Actor-Model", "Actors", "Distributed-Systems", "C#", ".NET")
+        LicenseUri = 'https://github.com/dotnet/Orleans#license'
+        ProjectUri = 'https://github.com/dotnet/orleans'
+        IconUri = 'https://raw.githubusercontent.com/dotnet/orleans/gh-pages/assets/logo_128.png'
+        ReleaseNotes = 'Initial Release $version$'
+    } 
+  }
+}
--- a/src/OrleansPSUtils/OrleansPSUtils.csproj
+++ b/src/OrleansPSUtils/OrleansPSUtils.csproj
@@ -1,0 +1,112 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <UsingTask TaskName="UpdateVersion" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v4.0.dll">
+    <ParameterGroup>
+      <InputFilename ParameterType="System.String" Required="true" />
+      <OutputFilename ParameterType="System.String" Required="true" />
+    </ParameterGroup>
+    <Task>
+      <Reference Include="System.Core" />
+      <Using Namespace="System" />
+      <Using Namespace="System.IO" />
+      <Code Type="Fragment" Language="cs">
+                  <![CDATA[
+                var version = File.ReadAllText(InputFilename)                                    
+                                    .Replace("\r\n", "")
+                                    .Replace("\n", "")
+                                    .Replace("\r", "");
+                var inputText = File.ReadAllText(OutputFilename);
+                File.WriteAllText(
+                    OutputFilename,
+                    inputText.Replace("$version$", version)
+                    );
+              ]]></Code>
+    </Task>
+  </UsingTask>
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{6AD37425-7CB4-4D23-80C3-A9D143329A66}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>OrleansPSUtils</RootNamespace>
+    <AssemblyName>OrleansPSUtils</AssemblyName>
+    <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <TargetFrameworkProfile />
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\Orleans\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\Orleans\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(OutDir)' != ''">
+    <OutDir>$(OutDir)\Orleans</OutDir>
+    <OutputPath>$(OutDir)</OutputPath>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="..\Build\GlobalAssemblyInfo.cs">
+      <Link>Properties\GlobalAssemblyInfo.cs</Link>
+    </Compile>
+    <Compile Include="GetGrain.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="StartGrainClient.cs" />
+    <Compile Include="StopGrainClient.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\OrleansCodeGenerator\OrleansCodeGenerator.csproj">
+      <Project>{8d937706-f6da-4d33-b0a9-24a260bd3080}</Project>
+      <Name>OrleansCodeGenerator</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\OrleansProviders\OrleansProviders.csproj">
+      <Project>{0054db14-2a92-4cc0-959e-a2c51f5e65d4}</Project>
+      <Name>OrleansProviders</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\Orleans\Orleans.csproj">
+      <Project>{bc1bd60c-e7d8-4452-a21c-290aec8e2e74}</Project>
+      <Name>Orleans</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="Orleans.psd1">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="project.json" />
+  </ItemGroup>   
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
+       Other similar extension points exist, see Microsoft.Common.targets.
+  <Target Name="BeforeBuild">
+  </Target>
+  <Target Name="AfterBuild">
+  </Target>
+  -->
+  <Target Name="AfterBuild">
+        <UpdateVersion InputFilename="$(SolutionDir)Build\Version.txt" OutputFilename="$(OutputPath)Orleans.psd1" />
+  </Target>
+</Project>
--- a/src/OrleansPSUtils/StartGrainClient.cs
+++ b/src/OrleansPSUtils/StartGrainClient.cs
@@ -1,0 +1,92 @@
+﻿using Orleans;
+using Orleans.Runtime.Configuration;
+using System;
+using System.IO;
+using System.Management.Automation;
+using System.Net;
+
+namespace OrleansPSUtils
+{
+    [Cmdlet(VerbsLifecycle.Start, "GrainClient", DefaultParameterSetName = DefaultSet)]
+    public class StartGrainClient : PSCmdlet
+    {
+        private const string DefaultSet = "Default";
+        private const string FilePathSet = "FilePath";
+        private const string FileSet = "File";
+        private const string ConfigSet = "Config";
+        private const string EndpointSet = "Endpoint";
+
+        [Parameter(Position = 1, Mandatory = true, ValueFromPipeline = true, ParameterSetName = FilePathSet)]
+        public string ConfigFilePath { get; set; }
+
+        [Parameter(Position = 2, Mandatory = true, ValueFromPipeline = true, ParameterSetName = FileSet)]
+        public FileInfo ConfigFile { get; set; }
+
+        [Parameter(Position = 3, Mandatory = true, ValueFromPipeline = true, ParameterSetName = ConfigSet)]
+        public ClientConfiguration Config { get; set; }
+
+        [Parameter(Position = 4, Mandatory = true, ValueFromPipeline = true, ParameterSetName = EndpointSet)]
+        public IPEndPoint GatewayAddress { get; set; }
+
+        [Parameter(Position = 5, ValueFromPipeline = true, ParameterSetName = EndpointSet)]
+        public bool OverrideConfig { get; set; } = true;
+
+        [Parameter(Position = 6, ValueFromPipeline = true, ParameterSetName = FilePathSet)]
+        [Parameter(Position = 6, ValueFromPipeline = true, ParameterSetName = FileSet)]
+        [Parameter(Position = 6, ValueFromPipeline = true, ParameterSetName = ConfigSet)]
+        [Parameter(Position = 6, ValueFromPipeline = true, ParameterSetName = EndpointSet)]
+        public TimeSpan Timeout { get; set; } = TimeSpan.Zero;
+
+        protected override void ProcessRecord()
+        {
+            try
+            {
+                WriteVerbose($"[{DateTime.UtcNow}] Initializing Orleans Grain Client");
+
+                switch (ParameterSetName)
+                {
+                    case FilePathSet:
+                        WriteVerbose($"[{DateTime.UtcNow}] Using config file at '{ConfigFilePath}'...");
+                        if (string.IsNullOrWhiteSpace(ConfigFilePath))
+                            throw new ArgumentNullException(nameof(ConfigFilePath));
+                        GrainClient.Initialize(ConfigFilePath);
+                        break;
+                    case FileSet:
+                        WriteVerbose($"[{DateTime.UtcNow}] Using provided config file...");
+                        if (ConfigFile == null)
+                            throw new ArgumentNullException(nameof(ConfigFile));
+                        GrainClient.Initialize(ConfigFile);
+                        break;
+                    case ConfigSet:
+                        WriteVerbose($"[{DateTime.UtcNow}] Using provided 'ClientConfiguration' object...");
+                        if (Config == null)
+                            throw new ArgumentNullException(nameof(Config));
+                        GrainClient.Initialize(Config);
+                        break;
+                    case EndpointSet:
+                        WriteVerbose($"[{DateTime.UtcNow}] Using default Orleans Grain Client initializer");
+                        if (GatewayAddress == null)
+                            throw new ArgumentNullException(nameof(GatewayAddress));
+                        GrainClient.Initialize(GatewayAddress, OverrideConfig);
+                        break;
+                    default:
+                        WriteVerbose($"[{DateTime.UtcNow}] Using default Orleans Grain Client initializer");
+                        GrainClient.Initialize();
+                        break;
+                }
+
+                if (Timeout != TimeSpan.Zero)
+                    GrainClient.SetResponseTimeout(Timeout);
+            }
+            catch (Exception ex)
+            {
+                WriteError(new ErrorRecord(ex, ex.GetType().Name, ErrorCategory.InvalidOperation, this));
+            }
+        }
+
+        protected override void StopProcessing()
+        {
+            GrainClient.Uninitialize();
+        }
+    }
+}
--- a/src/OrleansPSUtils/project.json
+++ b/src/OrleansPSUtils/project.json
@@ -1,0 +1,18 @@
+{
+  "dependencies": {
+    "Microsoft.CodeAnalysis.Analyzers": "1.0.0",
+    "Newtonsoft.Json": "7.0.1",
+    "Microsoft.CodeAnalysis.Common": "1.0.0",
+    "Microsoft.CodeAnalysis.CSharp": "1.0.0",
+    "System.Collections.Immutable": "1.1.36",
+    "System.Management.Automation.dll": "10.0.10586",
+    "System.Reflection.Metadata": "1.0.21"
+  },
+  "frameworks": {
+    "net451": {}
+  },
+  "runtimes": {
+    "win-anycpu": {},
+    "win": {}
+  }
+}
--- a/test/Tester/PSClientTests/PSClientTests.cs
+++ b/test/Tester/PSClientTests/PSClientTests.cs
@@ -1,0 +1,184 @@
+﻿using Orleans;
+using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using OrleansPSUtils;
+using System.Collections.Generic;
+using System.IO;
+using System.Management.Automation;
+using System.Management.Automation.Runspaces;
+using UnitTests.Tester;
+using Xunit;
+using Xunit.Abstractions;
+using Orleans.TestingHost;
+using UnitTests.GrainInterfaces;
+using System;
+
+namespace Tester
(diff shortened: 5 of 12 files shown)


jthelin:
Nice!

src/OrleansPSUtils/OrleansPSUtils.csproj
```diff
@@ -0,0 +1,108 @@
...
+    <Reference Include="System" />
+    <Reference Include="System.Collections.Immutable, Version=1.1.36.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\packages\System.Collections.Immutable.1.1.36\lib\portable-net45+win8+wp8+wpa81\System.Collections.Immutable.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="System.Core" />
+    <Reference Include="System.Management.Automation, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\System.Management.Automation.dll.10.0.10586.0\lib\net40\System.Management.Automation.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
```
jdom: In `master` we transitioned to using `project.json` for the nuget packages

src/OrleansPSUtils/StartGrainClient.cs
```diff
@@ -0,0 +1,92 @@
...
+        public IPEndPoint GatewayAddress { get; set; }
+
+        [Parameter(Position = 5, ValueFromPipeline = true, ParameterSetName = EndpointSet)]
+        public bool OverrideConfig { get; set; } = true;
+
+        [Parameter(Position = 6, ValueFromPipeline = true, ParameterSetName = FilePathSet)]
+        [Parameter(Position = 6, ValueFromPipeline = true, ParameterSetName = FileSet)]
+        [Parameter(Position = 6, ValueFromPipeline = true, ParameterSetName = ConfigSet)]
+        [Parameter(Position = 6, ValueFromPipeline = true, ParameterSetName = EndpointSet)]
+        public TimeSpan Timeout { get; set; } = TimeSpan.Zero;
```
jdom: why is this special and not rely on the client configuration?

src/OrleansPSUtils/StartGrainClient.cs
```diff
@@ -0,0 +1,92 @@
...
+        public string ConfigFilePath { get; set; }
+
+        [Parameter(Position = 2, Mandatory = true, ValueFromPipeline = true, ParameterSetName = FileSet)]
+        public FileInfo ConfigFile { get; set; }
+
+        [Parameter(Position = 3, Mandatory = true, ValueFromPipeline = true, ParameterSetName = ConfigSet)]
+        public ClientConfiguration Config { get; set; }
+
+        [Parameter(Position = 4, Mandatory = true, ValueFromPipeline = true, ParameterSetName = EndpointSet)]
+        public IPEndPoint GatewayAddress { get; set; }
```
jdom: There's a lot of "positional" arguments, but from what I think I understand, they are mutually exclusive, so you would never pass all the arguments by position, correct?

src/OrleansPSUtils/StartGrainClient.cs
```diff
@@ -0,0 +1,92 @@
...
+                        break;
+                    default:
+                        WriteVerbose($"[{DateTime.UtcNow}] Using default Orleans Grain Client initializer");
+                        GrainClient.Initialize();
+                        break;
+                }
+
+                if (Timeout != TimeSpan.Zero)
+                    GrainClient.SetResponseTimeout(Timeout);
+            }
```
jdom: since starting the client consumes resources on the silos, should we have some idle time where we disconnect the client? Not sure what's the best approach, but it's a little bit scary to leave tons of grain clients started unknowingly even after you are no longer running any scripts. Or maybe this already happens automatically by the grain client who only keeps connections open when it needs them? /cc @jason-bragg

test/Tester/PSClientTests.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        [Fact, TestCategory("BVT"), TestCategory("Tooling")]
+        public void GetGrain()
+        {
+            var getGrainCommand = new Command("Get-Grain");
+            getGrainCommand.Parameters.Add("GrainType", typeof(IManagementGrain));
+            getGrainCommand.Parameters.Add("LongKey", RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID);
+            _ps.Commands.AddCommand(getGrainCommand);
+
+            var results = _ps.Invoke<IManagementGrain>();
```
jdom: I'm confused, how is this working if you didn't call Start-GrainClient? Could it be picking up the grain client initialized by the DefaultClusterFixture outside of Powershell?

src/OrleansPSUtils/OrleansPSUtils.csproj
galvesribeiro: I remember you said that the nuget consumers like samples for example would still be on the packages.config. Ok will change that.

src/OrleansPSUtils/StartGrainClient.cs
galvesribeiro: What you mean? the `Timeout`? I just want to avoid to create another cmdlet just to have `GrainClient.SetResponseTimeout(timeout)` and since it requires the client initialized, just made a "powershell-way" to make this work.

src/OrleansPSUtils/StartGrainClient.cs
galvesribeiro: The position behaves differently when you have `ParameterSet`s. Basically it follow the order inside the own ParameterSet so yes, you will not pass more than 1 parameter for the same set. They are made by powershell runtime and you can see it on `Do-Something -?` and it will show you the possible sets acceptable.

src/OrleansPSUtils/StartGrainClient.cs
galvesribeiro: It is the same principle of `DbConnection`. You can call `Close()`, you can use `using()` and you can call `Dispose()`. All of these are ways for you to close a connection. You still need do it.

Can we do a timeout? Yes, we can by using PSSession. But that would require a lot more of complexity. Also, when the powershell Session is over, the objects are cleaned up. This is a behaviour of the powershell runtime and its session.

src/OrleansPSUtils/OrleansPSUtils.csproj
jdom: but this is not a nuget consumer, or did I get it wrong? I thought you were adding this to the main Orleans.sln

src/OrleansPSUtils/StartGrainClient.cs
jdom: ah, I see. Still, shouldn't you restart from position 1 for each of the sets? (completely naive question, I'm not familiar with it)

test/Tester/PSClientTests.cs
galvesribeiro: Yes, this is the case. Since many other tests cover the GrainClient itself, I just made a simple initial test to make sure the powershell is executing the correct things. Can add test for the others and use an exclusive silo for that.

src/OrleansPSUtils/OrleansPSUtils.csproj
galvesribeiro: Yeah you are right. Will change that.

test/Tester/PSClientTests.cs
jdom: but I don't understand... I get that you are just starting a normal silo cluster, but I don't understand how you are connecting using powershell if you don't call the Start-GrainClient command. By the fact that you created a new fixture, you are already exclusively starting up a new silo and tearing it down after this fixture completes

src/OrleansPSUtils/StartGrainClient.cs
galvesribeiro: No we can't use same numbers on diff properties :( That is why we have multiple attributes with the same number on diff sets. But that will behave as you expect.

For example:

```
PS C:\Users\xxxx\Documents\Visual Studio 2015\Projects\CodegenTests\src\TargetLibrary> Get-FileHash -?

NAME
    Get-FileHash

SYNTAX
    Get-FileHash [-Path] <string[]> [-Algorithm {SHA1 | SHA256 | SHA384 | SHA512 | MACTripleDES | MD5 | RIPEMD160}]
    [<CommonParameters>]

    Get-FileHash -LiteralPath <string[]> [-Algorithm {SHA1 | SHA256 | SHA384 | SHA512 | MACTripleDES | MD5 |
    RIPEMD160}]  [<CommonParameters>]

    Get-FileHash -InputStream <Stream> [-Algorithm {SHA1 | SHA256 | SHA384 | SHA512 | MACTripleDES | MD5 | RIPEMD160}]
     [<CommonParameters>]

```
Each one of those groups are a `ParameterSet`. In fact, you basicaly use the parameter by explicitly use the `-ParameterXXX value` sintax on most of the cases

test/Tester/PSClientTests.cs
galvesribeiro: GrainClient.IsInitialized is true. So if I call initialize again, it doesn't do anything... Check here: https://github.com/dotnet/orleans/pull/1990/files#diff-8e2208b94b1322807645ed26c86bc2adR51

src/OrleansPSUtils/project.json
```diff
@@ -0,0 +1,18 @@
+{
+  "dependencies": {
+    "Microsoft.CodeAnalysis.Analyzers": "1.0.0",
+    "Newtonsoft.Json": "7.0.1",
+    "Microsoft.CodeAnalysis.Common": "1.0.0",
+    "Microsoft.CodeAnalysis.CSharp": "1.0.0",
+    "System.Collections.Immutable": "1.1.36",
+    "System.Management.Automation.dll": "10.0.10586",
+    "System.Reflection.Metadata": "1.0.21"
+  },
```
jdom: remember that we are now using transitive restore. That means that you need to reference only the packages that you care about, and not its dependencies. For example, you probably can just reference `Microsoft.CodeAnalysis.CSharp` and avoid `Microsoft.CodeAnalysis.Analyzers, Microsoft.CodeAnalysis.Common, System.Collections.Immutable and System.Reflection.Metadata`.
BTW, do you even require `Microsoft.CodeAnalysis.CSharp` or is this a mistake?

jdom:
This looks good to me. 
@sergeybykov this will mean we'll probably have to start publishing this module in the Powershell gallery. Do you see any issues with merging this? An alternative is to use OrleansContrib (at least temporarily) although in the long term we probably want to have strong built-in PowerShell support (for cluster management mainly).

shayhatsor:
LGTM. @galvesribeiro please squash before we can merge. @jdom, I don't see any issues with merging this, only benefits.

galvesribeiro:
@shayhatsor done. Ah! Remember that while merging PRs GH can squash for you :)

test/Tester/project.json
```diff
@@ -10,7 +10,8 @@
     "WindowsAzure.Storage": "7.0.0",
     "xunit": "2.1.0",
     "xunit.runner.visualstudio": "2.1.0",
-    "Xunit.SkippableFact": "1.2.14"
+    "Xunit.SkippableFact": "1.2.14",
+    "System.Management.Automation.dll": "10.0.10586"
```
jdom: Since this one doesn't require anything from Tester, should we start creating a new test project for the Powershell DLL? I think that's going to be how we do things in the future, and it would help us in the short term since this is not portable to .NET Core.

src/OrleansPSUtils/project.json
jdom: If this is for using runtime codegen, since we are using transitive restore, I think you might not need to reference it here, you just reference the codegen project, and it should get the dependencies automatically.

test/Tester/project.json
galvesribeiro: It requires a cluster to run. Should we duplicated that test cluster stuff in another project? If yes I can move it right away and update the PR

jdom:
We'll then have to define the steps required to publish to the gallery. I assume these can be automated?
Not necessarily need to be addressed in this PR (in case it needs a script), but we'll need that automation in the future. if it's extremely straightforward, we'll just run that command or script when we release new versions of Orleans

test/Tester/PSClientTests/PSClientTests.cs
```diff
@@ -0,0 +1,145 @@
...
+using System.IO;
+using System.Management.Automation;
+using System.Management.Automation.Runspaces;
+using UnitTests.Tester;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Tester
+{
+    public class PowershellHostFixture : DefaultClusterFixture
```
jdom: you probably don't want to inherit from DefaultTestCluster fixture, and instead use it's base class and start the Silo cluster but not necessarily the grain client (I believe there's an overload to TestCluster's constructor where you just use the ClusterConfiguration but just pass a null ClientConfiguration). You can take the ClusterConfiguration object from the TestClusterOptions.ClusterConfiguration property.

galvesribeiro:
Yes, the process is the same as publish a nuget. You just call a command line that pass the build output path and the publish key just as nuget. Lets talk offline and I can make another PR for it once this is merged.

test/Tester/PSClientTests/PSClientTests.cs
```diff
@@ -0,0 +1,145 @@
...
+                    pair.Value);
+            }
+
+            // Client must be not initialized
+            Assert.NotNull(results[4]);
+            Assert.True((bool)results[4].BaseObject == GrainClient.IsInitialized);            
+        }
+
+        [Fact, TestCategory("BVT"), TestCategory("Tooling")]
+        public void GetGrainTest()
```
benjaminpetit: >GetGrainTest [](start = 20, length = 12)

It could be nice to test all possible values for key type

galvesribeiro: @benjaminpetit  you mean test each type of key supported?

benjaminpetit: Yes, you could try to get a `IGuidGrain`, `IStringGrain`, ect. Just to cover the switch statement in `GetGrain.ProcessRecord`

galvesribeiro:
@jdom and @benjaminpetit added the suggested changes.

test/Tester/project.json
jdom: Either we can move those utilities to another project, or if you don't want to take this work in this PR, you can do the same thing as TesterInternal, and have this new test project reference Tester in order to use `BaseTestClusterFixture`.
We can refactor `BaseTestClusterFixture` and some other extra stuff into their own project later. I believe that would be more appropriate instead of conflating that move in this PR.

src/OrleansPSUtils/Orleans.psd1
```diff
@@ -0,0 +1,24 @@
+﻿@{
+	RootModule = 'OrleansPSUtils.dll'
+	ModuleVersion = '1.2.3'
```
jdom: is there a way to have this populated by the version of the OrleansPSUtils (effectively the same as the rest of Orleans)?

benjaminpetit: Sorry I missed that earlier.

It could be nice to have this value synced with what is inside `Build\GlobalAssemblyInfo.cs`. I believe it could be done by writing a custom target in the csproj.

Same for the version in ReleaseNotes

benjaminpetit: At least update to 1.3.0, I think it will be enough for now

galvesribeiro: I tried to do this:

```
  <UsingTask TaskName="UpdateVersion" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v4.0.dll">
    <ParameterGroup>
      <InputFilename ParameterType="System.String" Required="true" />
      <OutputFilename ParameterType="System.String" Required="true" />
    </ParameterGroup>
    <Task>
      <Reference Include="System.Core" />
      <Using Namespace="System" />
      <Using Namespace="System.IO" />
      <Using Namespace="System.Text.RegularExpressions" />
      <Code Type="Fragment" Language="cs"><![CDATA[
				var version = File.ReadAllText(InputFilename);
				var inputText = File.ReadAllTest(OutputFilename);
				File.WriteAllText(
					OutputFilename,
					inputText.Replace("$version$", version)
					);
				
			  ]]></Code>
    </Task>
  </UsingTask>
```

and in the end of the .csproj this: 

```
<Target Name="AfterBuild">
    <ReplaceFileText InputFilename="$(SolutionDir)Build\Version.txt" OutputFilename="$(OutputDir)Orleans.psd1" />
  </Target>
```

I'm not good at MSBuild so if you can put something together for it I can quickly add it here.

Thanks

benjaminpetit: This works:
```
<UsingTask TaskName="UpdateVersion" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v4.0.dll">
    <ParameterGroup>
      <InputFilename ParameterType="System.String" Required="true" />
      <OutputFilename ParameterType="System.String" Required="true" />
    </ParameterGroup>
    <Task>
      <Reference Include="System.Core" />
      <Using Namespace="System" />
      <Using Namespace="System.IO" />
      <Code Type="Fragment" Language="cs">
                  <![CDATA[
                var version = File.ReadAllText(InputFilename)                                    
									.Replace("\r\n", "")
                                    .Replace("\n", "")
                                    .Replace("\r", "");
                var inputText = File.ReadAllText(OutputFilename);
                File.WriteAllText(
                    OutputFilename,
                    inputText.Replace("$version$", version)
                    );
              ]]></Code>
    </Task>
  </UsingTask>
```

and
````
  <Target Name="AfterBuild">
        <UpdateVersion InputFilename="$(SolutionDir)Build\Version.txt" OutputFilename="$(OutputPath)Orleans.psd1" />
  </Target>
````

File: src/OrleansPSUtils/OrleansPSUtils.csproj [hunk 11]
```diff
@@ -0,0 +1,82 @@
...
+    <AssemblyName>OrleansPSUtils</AssemblyName>
+    <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <TargetFrameworkProfile />
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\Orleans\</OutputPath>
```
benjaminpetit: Any reasons there is a trailing "Orleans" in the OuputPath? Same question for target Release

galvesribeiro: @BenjaminGibbs Yes. The folder that you call the publish command is the folder that will be packed and end up on the end-user machine. If I left debug/release, we would need an extra process to rename it.

src/OrleansPSUtils/Orleans.psd1
galvesribeiro: Thanks! My mistake was the AfterBuild that was in the wrong place and ofc the endline replaces you add. Now it worked like a charm :)

galvesribeiro:
- Added the fix for the version update by using the MSBuild code.
- Rebased on latest master.

galvesribeiro:
Build failure looks like the MSBuild task is breaking it on the CI mahine... @benjaminpetit / @jdom any idea?

benjaminpetit:
It seems that when you build with MSBuild the `OutputPath` variable is ignored, and you have to use `OutDir` instead. But of course, if you use `OutDir` in your csproj, it will not build in VS anymore. It should be an easy fix, I will post one.

But we have to find a way to copy all output from this project in a dedicated subfolder, right? Because right now with `Build.cmd`, everything goes to `Orleans\Binaries\{Debug|Release}`

galvesribeiro:
Yes, We need have a folder that will end up on the package. So when we call the pack command, it will build the package with this folder so later on, while installing, it will be `C:\local_ps_modules\Orleans` for instance. That is why I made the Output Dir to be `\{config}\Orleans\*`

benjaminpetit:
Adding this in the csproj:
````
  <PropertyGroup Condition="'$(OutDir)' != ''">
    <OutDir>$(OutDir)\Orleans</OutDir>
	<OutputPath>$(OutDir)</OutputPath>
  </PropertyGroup>
````
seems to do exactly what we want!

**Comment by galvesribeiro** [comment]
@benjaminpetit any particular position?

**Comment by benjaminpetit** [comment]
Just before the first `ItemGroup` node should be good

**Comment by galvesribeiro** [comment]
@benjaminpetit done! Thanks for the help!

:shipit:

File: test/Tester/PSClientTests/PSClientTests.cs (hunk 8 above)
Comment (jdom) [inline test/Tester/PSClientTests/PSClientTests.cs]: was this a regression? I thought you had an iteration where you were no longer inheriting from Default

File: test/Tester/project.json (hunk 7 above)
Comment (jdom) [inline test/Tester/project.json]: maybe the change was lost when rebasing? I thought you were creating a new test project

File: test/Tester/PSClientTests/PSClientTests.cs (hunk 8 above)
Comment (galvesribeiro) [inline test/Tester/PSClientTests/PSClientTests.cs]: This is outdated @jdom. The current code in the PR already has the change you suggested and inherit from `BaseTestClusterFixture`

Comment (galvesribeiro) [inline test/Tester/PSClientTests/PSClientTests.cs]: Check here https://github.com/dotnet/orleans/pull/1990/commits/06fc7f102a54823c04faa22b66aa7114a6631d97#diff-67c19d50200ddd6504cda9466e833332R18

**Comment by galvesribeiro** [comment]
Rebased

File: test/Tester/project.json (hunk 7 above)
Comment (galvesribeiro) [inline test/Tester/project.json]: Not for now. Will keep it there. We need revisit all the tests and make a test kit otherwise we will still keeping the tests duplicated but this time, spread across multiple projects. 

I think we should design a new testkit I'm just not sure this is the right time since we have tons of other things to do for the .net core move...

**Comment by shayhatsor** [comment]
LGTM

**Comment by shayhatsor** [comment]
Thanks @galvesribeiro !!

File: test/Tester/project.json (hunk 7 above)
Comment (jdom) [inline test/Tester/project.json]: I think this does not suffer from the same as the other provider tests and does not require any test kit. It would be trivial to just move to a new test project in the way I suggested.

**Comment by jdom** [comment]
Thanks @galvesribeiro, great contribution. LGTM, I'll merge right after you move this to it's own test project

**Comment by jdom** [comment]
Doh, didn't realized @shayhatsor had merged already. In the future, @shayhatsor, please give time for other reviewers to validate their comments after the submitter amends the PR. In this case there were still unaddressed comments. Nothing major in this case, but now it puts the pressure on us to fix it right away.

**Comment by galvesribeiro** [comment]
@jdom it was already merged. Also, related to the tests, until we have fully extracted the testkit, I see no point on have separated projects just to duplicate test code.

**Comment by jdom** [comment]
@galvesribeiro, in the future please when you address comments with new changes, please do not rebase every time. In some cases it caused regressions, but regardless of that, it's hard to know what are the new changes since the last time I reviewed it, and had to review all over again. We can use the squash and merge button anyway. It would only require rebasing if there's conflicts that do not allow the merge

**Comment by shayhatsor** [comment]
@jdom, I was under the impression that @galvesribeiro had already addressed all issues. I apologize for jumping the gun.

**Comment by jdom** [comment]
not sure I understand why you would need a test kit in this case. It's not a provider, you don't need to reuse any of the tests (which is what the test kit would address)

**Comment by jdom** [comment]
No problem, btw, I'm just mentioning this for future practice. I'll put out a PR on Monday to address the remaining comment, otherwise it provides more friction to the .net core migration.

**Comment by veikkoeeva** [comment]
@jdom Maybe put something to contributor guidelines? It might be useful to have something like:
1. While in your master `git pull upstream master`
2. `git checkout -b some_branch`
3. While working in the topic branch and new code gets pushed `git pull --rebase upstream master`
4. `git push --force-with-lease origin +some_branch`

Perhaps with even more detailed instructions (and having `rerere` on helps). It looks a common tripping point is bleeding commits between branches and/or updating the working branch (some_branch) gets foiled.

**Comment by jdom** [comment]
Hmm, unfortunately this merge broke most of our builds (functionals, load tests, reliability), since some of them still use VSO's way of building that puts everything in a single folder, and I guess the OutDir variable is not being set.

```
C:\Builds\5879\Orleans\CI-BVT\src\src\Orleans.sln - 1 error(s), 329 warning(s), View Log File
 C:\Builds\5879\Orleans\CI-BVT\src\src\OrleansPSUtils\OrleansPSUtils.csproj (110): The "UpdateVersion" task failed unexpectedly.
System.IO.FileNotFoundException: Could not find file 'C:\Builds\5879\Orleans\CI-BVT\src\src\OrleansPSUtils\$(OutDir)Orleans.psd1'.
File name: 'C:\Builds\5879\Orleans\CI-BVT\src\src\OrleansPSUtils\$(OutDir)Orleans.psd1'
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.StreamReader..ctor(String path, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize, Boolean checkHost)
   at System.IO.File.InternalReadAllText(String path, Encoding encoding, Boolean checkHost)
   at InlineCode.UpdateVersion.Execute() in c:\Users\xcgbuild\AppData\Local\Temp\tvxtw5sa.0.cs:line 63
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()
```

This is why we were going to run this PR by our build system before merging... Unfortunately it is very common that when we add build tasks, some issues arise. We are trying to get rid of this way of building in some of our tests, but we are not there yet.

Could someone please lend a hand to fix this? Otherwise we are blocked on running tests for other PRs and we'll either have to drop everything and do this emergency fix or revert this for now. I'd prefer not to revert if someone is willing to help asap.

galvesribeiro:
Without understand and have access to VSO I can't say what variables VSO is using that is making that don't work. 

BTW, lets just be clear. The merge doesn't break anything by itself. The common consensus in the past was that we had ALL the build processes working seamless between both Jenkins and VSO and we spent weeks on that migration. If there is something old there, we would never guess since we don't have access to it :(

jdom:
Let me clarify @galvesribeiro: you did nothing wrong. Nothing at all.
It's just that we take our time to merge, especially when there are dependencies like this, for a reason.
Just as an FYI, what we spent a lot of time making sure it works is for running xUnit tests with both vstest.exe and xunit.console.exe, and the reason is that the build system is still different (we did not get rid of it last time, we made it work across both, remember?). We definitely need to fix this, and we already started porting our battery of tests over in December, and there is a PR coming very soon with some more (@benjaminpetit has been working on it last week).

shayhatsor:
If this merge is causing any problem with internal builds, I see no harm in reverting for now.
@jdom, what should I watch out for? is it new projects ? changes in the build script(s) ?

jdom:
I'm troubleshooting now to avoid the revert, I think I'm close, just running the tests.

Yeah, both new projects and new build dependencies (whether it's build scripts, msbuild tasks, test scripts). Also external environmental dependencies.

jdom:
Nevermind, I wasn't close. @benjaminpetit is taking a look at it now

galvesribeiro:
I can't actively help on that since I cant simulate the problem here... If there is a way for me to do that let me know

jdom:
@galvesribeiro Do you know what's required to publish to the gallery? Signed build? We have one. Do we also need to sign any powershell script?

galvesribeiro:
@jdom sorry for the delay.

Just create an account on powershellgallery.com which is pretty similar to the nuget.org one and grab an APIKey. Invoke the [Publish-Module](https://technet.microsoft.com/en-us/library/dn807163.aspx) cmdlet and fill all the parameters and you should all be set. 

Regarding the signature, I would suggest you to sign the assembly. This project output a .Net Assembly just like any other class library, so the sign process should be the same. There is no real .ps1 script.

Let me know if you have any problems.

PR #1991, opened 2016-07-27
Replace buffer pool underlying collection

dVakulen:
Currently used <code>BlockingCollection</code> takes quite a portion of serialization time:

(results from mini benchmark of roundtrip serialization of POCO)
dottraceview64_2016-07-27_09-16-55

After replacing with <code>ConcurrentBag</code>:

dottraceview64_2016-07-27_09-21-46
(note, that <code>Release</code> method also disappeared from hot spots list)

### DIFF
--- a/src/Orleans/Messaging/BufferPool.cs
+++ b/src/Orleans/Messaging/BufferPool.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Threading;
 using Orleans.Runtime.Configuration;
 
 namespace Orleans.Runtime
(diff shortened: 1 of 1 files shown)


src/Orleans/Messaging/BufferPool.cs
```diff
@@ -82,6 +89,11 @@ public byte[] GetBuffer()
                 buffer = new byte[byteBufferSize];
                 allocatedBufferCounter.Increment();
             }
+            else if (limitBuffersCount)
+            {
+                Interlocked.Decrement(ref currentAllocatedBuffers);
```
dVakulen: Used separate counter with Interlocked as it twice more faster that accessing of ConcurrentBag Count property. But the difference in global perspective isn't much noticeable, so in readability purposes can change to .Count.

src/Orleans/Messaging/BufferPool.cs
```diff
@@ -1,20 +1,25 @@
...
+        private readonly int maxBuffersCount;
+        private readonly bool limitBuffersCount;
+        private readonly ConcurrentBag<byte[]> buffers;
         private readonly CounterStatistic allocatedBufferCounter;
         private readonly CounterStatistic checkedOutBufferCounter;
         private readonly CounterStatistic checkedInBufferCounter;
         private readonly CounterStatistic droppedBufferCounter;
         private readonly CounterStatistic droppedTooLargeBufferCounter;
 
+        private int currentAllocatedBuffers;
```
jason-bragg: currentAllocatedBuffers, to me seems an unclear name.  Since this is related to maxBufferCount and limitBufferCount, maybe currentBufferCount is a more clear name?

jason-bragg:
Race conditions can allow some minor impersision in the pool size management at max pool capacity, but this should be harmless, and, imo, isn't worth addressing.

One minor comment regarding naming, but otherwise looks good.

src/Orleans/Messaging/BufferPool.cs
dVakulen: Agreed.

PR #1993, opened 2016-07-27
Updated all nuspec to use new icon.

galvesribeiro:
Just update the nuspecs to use the new icons from #1989


jdom:
Awesome! Could it be an issue that the new image is several times bigger than an icon? Did you check if NuGet's UI displays it correctly?

galvesribeiro:
Just found out that that size don't work on nuget. @ReubenBond  already fixed that on #2002 so when it gets merged I'll update this PR.

jdom:
Reuben's PR is merged

galvesribeiro:
image

Now the icon is correct with the adjusted icon. :)

galvesribeiro:
Thanks!

PR #1996, opened 2016-07-28
Avoid dictionary allocation on creation of grain instance

dVakulen:
On each grain activation dictionary was allocated, presumably for the sake of immutability,
 but its only usage is at https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/Catalog/Catalog.cs#L634.

dottraceview64_2016-07-29_00-29-00

### DIFF
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -630,8 +630,7 @@ namespace Orleans.Runtime
         private void CreateGrainInstance(string grainTypeName, ActivationData data, string genericArguments)
         {
             string grainClassName;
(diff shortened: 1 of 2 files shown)


benjaminpetit:
Why not use the `TryGetPrimaryImplementation` from `GrainInterfaceData` in a proxy method in `GrainTypeManager` then use it here https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/Catalog/Catalog.cs#L634 instead?

It seems less risky to me for the long term.

gabikliot:
Can you change the function to return idictionary as well? To be robust to future usages.

gabikliot:
Or Benjamin's suggestion which is even better actually.

dVakulen:
Done.

PR #1997, opened 2016-07-28
Use GetTypeInfo() on code changed by #1963

jdom:
Use GetTypeInfo() in preparation for the .NET Core port (broken during #1963).
Minor cleanup: Replace DNXCORE50 deprecated constant with NETSTANDARD1_6

/cc @dVakulen


src/Orleans/Serialization/TypeUtilities.cs
```diff
@@ -82,7 +82,7 @@ internal static bool IsOrleansShallowCopyable(this Type t)
 
         private static bool IsTypeFieldsShallowCopyable(TypeInfo typeInfo)
         {
-            return typeInfo.GetFields().All(f => !(f.FieldType.IsEquivalentTo(typeInfo)) && IsOrleansShallowCopyable(f.FieldType));
+            return typeInfo.GetFields().All(f => !f.FieldType.GetTypeInfo().IsEquivalentTo(typeInfo.AsType()) && IsOrleansShallowCopyable(f.FieldType));
```
dVakulen: According to tests in the corefx repository: https://github.com/dotnet/corefx/blob/master/src/System.Reflection/tests/TypeInfo/TypeInfo_IsEquivalentTo.cs#L17 
this was compatible with .NET core.

dVakulen: And either <code>GetTypeInfo()</code> or <code>AsType()</code> call seems unnecessary.

jdom: But that example you are linking is actually doing Getty peinó and then passing a type (not a typeinfo)  to the IsEquivalentTo method, which is exactly what I'm doing too. BTW, I did this change in my coreclr branch because it did not compile for me, I'm not speculating. This works in full dotNET because TypeInfo inherits from Type, but that's not the case in netcore, and it's why we use the GetTypeInfo extension method and the AsType

dVakulen: Then this equality check isn't symmetric, but I believed it is..

jdom: yup, it's not. That's probably why they aren't just using IEquatable in that test.
With this new insight, do you recommend reverting your original PR or keep it this way? or use a different comparison?

dVakulen: There's no need to revert original PR, as it purpose was in extracting allocation causing code into separate method. So it's OK as it is.

jdom: but what about doing this at least:
`return typeInfo.GetFields().All(f => f.FieldType != typeInfo.AsType()) && IsOrleansShallowCopyable(f.FieldType));`
This way we don't get the TypeInfo unecessarily in the first place

jdom: Updated with that approach and added some test coverage

PR #2001, opened 2016-07-30
Optimize CheckLength method

dVakulen:
BinaryTokenStreamReader's <code>CheckLength</code> method takes ~21% of deserialization time (tested on simple POCO) due to excessive usage of <code>ArraySegment</code>'s properties: <code>Offset</code>, <code>Array</code> and <code>Count</code>.

dottraceview64_2016-07-30_00-11-32

During rewrite was removed repeatable accessing of the properties and unnecessary construction of the <code>ArraySegment</code> on each call. 

Results after:

dottraceview64_2016-07-30_00-06-32

### DIFF
--- a/src/Orleans/Serialization/BinaryTokenStreamReader.cs
+++ b/src/Orleans/Serialization/BinaryTokenStreamReader.cs
@@ -19,14 +19,18 @@ namespace Orleans.Serialization
     public class BinaryTokenStreamReader
     {
         private readonly IList<ArraySegment<byte>> buffers;
+        private readonly int buffersCount;
         private int currentSegmentIndex;
         private ArraySegment<byte> currentSegment;
         private byte[] currentBuffer;
         private int currentOffset;
+        private int currentSegmentOffset;
+        private int currentSegmentCount;
         private int totalProcessedBytes;
         private readonly int totalLength;
 
         private static readonly ArraySegment<byte> emptySegment = new ArraySegment<byte>(new byte[0]);
+        private static readonly byte[] emptyByteArray = new byte[0];
 
         /// <summary>
         /// Create a new BinaryTokenStreamReader to read from the specified input byte array.
@@ -49,7 +53,10 @@ namespace Orleans.Serialization
             currentSegment = buffs[0];
             currentBuffer = currentSegment.Array;
             currentOffset = currentSegment.Offset;
+            currentSegmentOffset = currentOffset;
+            currentSegmentCount = currentSegment.Count;
             totalLength = buffs.Sum(b => b.Count);
+            buffersCount = buffs.Count;
             Trace("Starting new stream reader");
         }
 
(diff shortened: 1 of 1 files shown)


src/Orleans/Serialization/BinaryTokenStreamReader.cs
```diff
@@ -115,52 +126,52 @@ private ArraySegment<byte> CheckLength(int n, out bool safeToUse)
...
                 StartNextSegment();
             }
 
-            if (currentOffset + n <= currentSegment.Offset + currentSegment.Count)
+            var nextOffset = currentOffset + n;
+            if (nextOffset <= currentSegmentOffset + currentSegmentCount)
             {
-                var result = new ArraySegment<byte>(currentBuffer, currentOffset, n);
-                currentOffset += n;
-                if (currentOffset >= currentSegment.Offset + currentSegment.Count)
```
dVakulen: This check was removed as it's equal to the one at https://github.com/dotnet/orleans/pull/2001/files#diff-4084318ed4204a1de3f94a50a63573feR135.

gabikliot:
@dvakulen, do you think you

dVakulen:
Results from https://github.com/dotnet/orleans/pull/2003:
Before: 

                Method | Repeats |      Median |    StdDev |
---------------------- |-------- |------------ |---------- |
   SerializerBenchmark |  300000 | 581.5529 ms | 2.2370 ms |
 DeserializerBenchmark |  300000 | 466.5629 ms | 1.4196 ms |

After:

                Method | Repeats |      Median |    StdDev |
---------------------- |-------- |------------ |---------- |
   SerializerBenchmark |  300000 | 578.9851 ms | 3.9789 ms |
 DeserializerBenchmark |  300000 | 403.7003 ms | 1.2128 ms |

src/Orleans/Serialization/BinaryTokenStreamReader.cs
jdom: are you sure? Without going into too much depth (you might have already checked and can guide me), they don't look equal. This was checking if `currentOffset` (which is now `+n` actually) is *greater* or equal.

dVakulen: because check at line https://github.com/dotnet/orleans/blob/master/src/Orleans/Serialization/BinaryTokenStreamReader.cs#L129 ensures that in can be only less or equal.

jdom: But doesn't that still leave the "equal" part as a possibility? I'm just saying this based on your original comment that it is the same check as line 124, but this is not necessarily true after the `currentOffset += n;`

gabikliot:
So almost no improvement fro serialization and about 15% for deserialization?

src/Orleans/Serialization/BinaryTokenStreamReader.cs
gabikliot: StartNextSegment at line 126 can change those values, so I agree with @jdom.

dVakulen: Changing values by StartNextSegment at line 126 doesn't matter because of return statement at line 137, so the new segment will be started anyway on the next CheckLength call.

dVakulen:
Yes.

gabikliot:
LGTM.

PR #2003, opened 2016-07-31
Add serialization benchmarks

dVakulen:
As requested in https://github.com/dotnet/orleans/pull/2001#issuecomment-236368141


src/OrleansCodeGenerator/project.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "Microsoft.CodeAnalysis.CSharp": "1.0.0"
+    "Microsoft.CodeAnalysis.CSharp": "1.3.2"
```
dVakulen: Had to update it in order to run with latest <code>BenchmarkDotNet</code> package. cc @ReubenBond

dVakulen:
It seems that newly added packages has to be restored manually on the build server.

src/OrleansCodeGenerator/project.json
jdom: can't it be solved with binding redirects on the benchmark project? Seems like a pretty big user facing update that we should be able to avoid.
Otherwise we'd have to update it everywhere, including the nuspec files

gabikliot:
perhaps this can just be added as a XUnit test, not under BVT so it does not run every time, but can be run on demand. That way you will save the extra new project.

dVakulen:
Other actor frameworks have separate projects for benchmarks - https://github.com/akkadotnet/akka.net/tree/dev/src/benchmark, https://github.com/akka/akka/tree/master/akka-bench-jmh; for now, while public benchmarks set contains only few tests it's ok to put it into project with tests.

gabikliot:
LGTM.

dVakulen:
I'll add more benchmarks which will cover basic usages of Orleans, so that developers could see perfomance impact of their changes without waiting for internal full test suite run or creating their own benchmarks. This will also justify creating of extra project.

jdom:
Yeah, I also prefer they are separate. We are already talking about splitting this conglomerate of tests in a a single (or 2 actually) project.
There's a few bad side effects of having them all together, such as parallelization of tests, but the biggest upcoming issue is with the port to coreclr, which means that in order to port the tests, we need to port the entire codebase first (as it is referenced in its entirety by the test projects) or deal with a huge amount of conditional compilation.

jdom:
BTW, please avoid merging until we resolve the issue with the CodeAnalysis upgrade

src/OrleansCodeGenerator/project.json
dVakulen: Fixed.

dVakulen:
Newly added benchmark shows stable numbers; this PR is ready for review, more benchmarks will be added in future, after fixing of perfomance issues found by profiling this one.

src/OrleansBenchmarks/OrleansBenchmarks.csproj
```diff
@@ -0,0 +1,111 @@
...
+  <ItemGroup>
+    <Content Include="BenchClientConfiguration.xml">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+    <Content Include="BenchServerConfiguration.xml">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <Import Project="..\packages\Microsoft.Diagnostics.Tracing.TraceEvent.1.0.41\build\Microsoft.Diagnostics.Tracing.TraceEvent.targets" Condition="Exists('..\packages\Microsoft.Diagnostics.Tracing.TraceEvent.1.0.41\build\Microsoft.Diagnostics.Tracing.TraceEvent.targets')" />
```
jdom: these are probably a remnant of before migrating to project.json, and I believe it's the reason why it's failing the build. Search for `..\packages\` and you'll see several instances

dVakulen: Thanks, fixed.

jdom:
Please move everything under a `test/benchmarks` folder. Please do the same with the solution folders.

src/BenchmarkGrainInterfaces/MapReduce/DataflowGrainsInterfaces.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Orleans;
+
+namespace GrainInterfaces
+{
+    public interface IDataflowGrain : IGrainWithGuidKey
```
jdom: It is interesting that this is imitating the DataFlow library. Nevertheless, is this valuable for the kind work you are doing or is it mainly a cool sample?
I'm not pushing back, just want to understand why we are adding it.

src/OrleansBenchmarks/BenchClientConfiguration.xml
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<!-- 
+	This is a sample client configuration file. 
+	For a detailed reference, see "Orleans Configuration Reference.html".
+-->
+<ClientConfiguration xmlns="urn:orleans">
+  <GatewayProvider ProviderType="Config"/>
+  <Gateway Address="localhost" Port="30000"/>
+  <Statistics MetricsTableWriteInterval="30s" PerfCounterWriteInterval="30s" LogWriteInterval="300s" WriteLogStatisticsToTable="true"/>
+</ClientConfiguration>
```
jdom: Let's use programmatic config instead.

src/OrleansBenchmarks/BenchServerConfiguration.xml
```diff
@@ -0,0 +1,31 @@
...
+  <Globals>
+    <StorageProviders>
+      <Provider Type="Orleans.Storage.MemoryStorage" Name="MemoryStore" />
+     </StorageProviders>
+    <SeedNode Address="localhost" Port="11111" />
+  </Globals>
+  <Defaults>
+    <Networking Address="localhost" Port="11111" />
+    <ProxyingGateway Address="localhost" Port="30000" />
+    <Tracing DefaultTraceLevel="Warning" TraceToConsole="true" TraceToFile="{0}-{1}.log">
```
jdom: we probably want to turn off console logging, as a quick run of this filled up the console with unrelated info (still let's do it in programmatic config)

src/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
+using GrainInterfaces;
+using Orleans;
+using Orleans.Concurrency;
+using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using Orleans.Runtime.Host;
+using Orleans.Serialization;
```
jdom: Please here and every other file, try to remove unnecessary using statements. It's usually harmless, but I found out it made the migration to .net core a lot harder, as some of these are not available.
It would be nice to be able to run these in .net core and compare in the future

src/OrleansBenchmarks/project.json
```diff
@@ -0,0 +1,25 @@
...
+    "Microsoft.Diagnostics.Tracing.TraceEvent": "1.0.41",
+    "System.Collections": "4.0.0",
+    "System.Diagnostics.Debug": "4.0.0",
+    "System.Globalization": "4.0.0",
+    "System.Linq": "4.0.0",
+    "System.Resources.ResourceManager": "4.0.0",
+    "System.Runtime": "4.0.0",
+    "System.Runtime.Extensions": "4.0.0",
+    "System.Threading": "4.0.0",
+    "System.Threading.Tasks": "4.0.0"
```
jdom: we are now using nuget transitive restore, so please just reference the nuget packages you depend on, but not the dependencies of your direct dependencies. This ways upgrades will be easier for us (and also migration to .net core)

src/OrleansBenchmarks/Program.cs
```diff
@@ -0,0 +1,52 @@
...
+        static void Main(string[] args)
+        {
+            if (args.Length == 0 || !_benchmarks.ContainsKey(args[0]))
+            {
+                Console.WriteLine("Running full benchmarks suite");
+                _benchmarks.Select(pair => pair.Value).ToList().ForEach(action => action());
+            }
+            else
+            {
+                _benchmarks[args[0]]();
```
jdom: please either add a readme or some easy way of figuring out how to run this. Potentially also a link to BenchmarkDotNet's doc for for advanced scenarios when running.

src/OrleansBenchmarks/Program.cs
```diff
@@ -0,0 +1,52 @@
...
+            },
+            ["Serialization"] = () =>
+            {
+                var summary = BenchmarkRunner.Run<SerializationBenchmarks>();
+            }
+        };
+
+        static void Main(string[] args)
+        {
+            if (args.Length == 0 || !_benchmarks.ContainsKey(args[0]))
```
jdom: Since the benchmarks take SO long to run, I'd make the default to just return the names of the available benchmarks. If the user wants to run all, make him opt in by pass `all` or something like that

src/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs
```diff
@@ -0,0 +1,175 @@
...
+        private int _currentRepeat = 0;
+
+        [Setup]
+        public void BenchmarkSetup()
+        {
+            AppDomain hostDomain = AppDomain.CreateDomain("OrleansHost", null, new AppDomainSetup
+            {
+                AppDomainInitializer = InitSilo,
+                AppDomainInitializerArguments = new string[] { "Primary", "BenchServerConfiguration.xml" },
+            });
```
jdom: Should we be using `TestCluster` instead to manage this for us? It will be simpler for when we need more than 1 silo (or maybe that will never be an option for these kinds of benchmarks?).
Also it will probably be easier if we migrate this to .net core, since once we update `TestCluster` to run silos without new AppDomains (as it's not supported), this will just work, otherwise this benchmark might require keeping track of started processes to run the silos.

src/BenchmarkGrainInterfaces/MapReduce/DataflowGrainsInterfaces.cs
dVakulen: Almost all of the code is needed for this benchmark in one form or another. But writing it in such way leaves the ability for it to evolve into something bigger than just a sample (i wouldn't add epithet 'cool' yet).

src/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs
dVakulen: Makes sense.

src/OrleansBenchmarks/Program.cs
dVakulen: Fixed

src/OrleansBenchmarks/Program.cs
dVakulen: Added comment

src/OrleansBenchmarks/project.json
dVakulen: Fixed

src/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs
dVakulen: Fixed

src/OrleansBenchmarks/BenchServerConfiguration.xml
dVakulen: Fixed

src/OrleansBenchmarks/BenchClientConfiguration.xml
dVakulen: Done

test/Benchmarks/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs.rej
```diff
@@ -0,0 +1,23 @@
+diff a/src/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs b/src/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs	(rejected hunks)
```
jdom: merge conflict

test/Benchmarks/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs
```diff
@@ -0,0 +1,138 @@
...
+            clusterConfig.Defaults.TraceToConsole = false;
+            _siloHandle = TestCluster.StartOrleansSilo(
+                new TestCluster(),
+                Silo.SiloType.Primary,
+                clusterConfig,
+                nodeConfig);
+
+            var clientConfig = ClientConfiguration.LocalhostSilo();
+            clientConfig.TraceToConsole = false;
+            GrainClient.Initialize(clientConfig);
```
jdom: you should be able to replace this whole set up with the equivalent higher level set up:
```
            TestClusterOptions.DefaultTraceToConsole = false;
            var options = new TestClusterOptions(1);
            options.ClusterConfiguration.ApplyToAllNodes(c => c.DefaultTraceLevel = Severity.Warning);
            _host = new TestCluster(options);
            _host.Deploy();
```
This has a few additional benefits (such as allowing to start more than 1 silo and it being configured automatically, using random ports for networking, and a little more future proof if we have to use some external process manager when we move to .net core).

Also, if there is a concept of tear down in the benchmark framework, please consider calling `_host.StopAllSilos();` to avoid this cluster leaking into the next benchmark being run.

test/Benchmarks/OrleansBenchmarks/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+                Console.WriteLine("Please, select benchmark, list of available:");
+                _benchmarks
+                    .Select(pair => pair.Key)
+                    .ToList()
+                    .ForEach(Console.WriteLine);
+                Console.WriteLine("All");
+                return;
+            }
+
+            if (args[0].Equals("all", StringComparison.InvariantCultureIgnoreCase))
```
jason-bragg: Will it ever make it here?  "all" is not in _benchmarks so first conditional will fire.
Suggest adding "all" as first in list with the below logic as it's action.

test/Benchmarks/OrleansBenchmarks/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            ["MapReduce"] = () =>
+            {
+                Console.WriteLine("Running MapReduce benchmark");
+                var mapReduceBenchmark = new MapReduceBenchmark();
+                mapReduceBenchmark.BenchmarkSetup();
+                var stopWatch = Stopwatch.StartNew();
+                mapReduceBenchmark.Bench().Wait();
+                Console.WriteLine($"Elapsed milliseconds: {stopWatch.ElapsedMilliseconds}");
+                Console.ReadLine();
```
jason-bragg: Later maintainers will likely follow whatever pattern initially put in place.   More maintainable initial pattern will require less refactoring later.  

Suggest action logic be moved into a function for each action.  That will keep this initialization from growing into spaghetti hell over time.

test/Benchmarks/OrleansBenchmarkGrainInterfaces/MapReduce/DataflowGrainsInterfaces.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Orleans;
+
+namespace OrleansGrainInterfaces.MapReduce
+{
+    public interface IDataflowGrain : IGrainWithGuidKey
```
jason-bragg: Suggest inheriting from IGrain.  Grain key can be defined in specializations.

test/Benchmarks/OrleansBenchmarkGrainInterfaces/MapReduce/DataflowGrainsInterfaces.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        Task Complete();
+
+        Task Fault();
+
+        Task Completion();
+    }
+
+    public interface ITargetGrain<in TInput> : IDataflowGrain
```
jason-bragg: Interface name is very generic.  Suspect this will lead to naming conflicts in future.  Especially if more dataflow related test scenarios are added.  Similar concern for other concrete grain types in this test harness.

test/Benchmarks/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs
```diff
@@ -0,0 +1,127 @@
...
+        private readonly int _intermediateStagesCount = 15;
+        private readonly int _pipelineParallelization = 4;
+        private readonly int _repeats = 50000;
+        private int _currentRepeat = 0;
+
+        [Setup]
+        public void BenchmarkSetup()
+        {
+            TestClusterOptions.DefaultTraceToConsole = false;
+            var options = new TestClusterOptions(1);
```
jason-bragg: Single silo will mean all grain calls are local, bypassing much of the cost.  Not sure this will result in a meaningful benchmark.

jason-bragg:
Very excited to see public performance tests!  Nice!

test/Benchmarks/OrleansBenchmarks/MapReduce/MapReduceBenchmark.cs
dVakulen: I've ran it with multiple silo, and the serialization + networking cost was the only overhead. And for those matters separate benchmarks can be created.

test/Benchmarks/OrleansBenchmarkGrainInterfaces/MapReduce/DataflowGrainsInterfaces.cs
dVakulen: It's meant to be evolved with time, and renaming won't be a problem for a while.

test/Benchmarks/OrleansBenchmarks/project.json
```diff
@@ -0,0 +1,16 @@
+﻿{
+  "dependencies": {
+    "BenchmarkDotNet": "0.9.0",
+    "System.Collections": "4.0.0",
+    "System.Linq": "4.0.0",
+    "System.Threading": "4.0.0",
+    "System.Threading.Tasks": "4.0.0"
```
jdom: I believe that when using net451 as the framework, you shouldn't need to reference these that are automatically included in the framework (and conflict with the versions from the framework).
I'm not entirely familiar with this yet, but take a look at the log: http://dotnet-ci.cloudapp.net/job/dotnet_orleans/job/innerloop_prtest/1785/consoleFull

And search for "Detected package downgrade". You'll see a ton of these:
```
14:29:55      1>EXEC : warning : Detected package downgrade: System.Collections from 4.0.11 to 4.0.0 [D:\j\workspace\innerloop_prtest933994bb\src\Orleans.sln]
14:29:55           OrleansBenchmarks (>= 1.0.0) -> Tester (>= 1.0.0) -> Microsoft.Extensions.DependencyInjection (>= 1.0.0) -> System.Collections (>= 4.0.11) 
14:29:55           OrleansBenchmarks (>= 1.0.0) -> System.Collections (>= 4.0.0)
14:29:55      1>EXEC : warning : Detected package downgrade: System.Threading from 4.0.11 to 4.0.0 [D:\j\workspace\innerloop_prtest933994bb\src\Orleans.sln]
14:29:55           OrleansBenchmarks (>= 1.0.0) -> Tester (>= 1.0.0) -> Microsoft.Extensions.DependencyInjection (>= 1.0.0) -> System.Threading (>= 4.0.11) 
14:29:55           OrleansBenchmarks (>= 1.0.0) -> System.Threading (>= 4.0.0)
<snip for brevity/>
```

test/Benchmarks/OrleansBenchmarkGrainInterfaces/MapReduce/DataflowGrainsInterfaces.cs
dVakulen: Plus, it's purpose in being generic - new dataflow related tests won't need another <code>ITargetGrain</code>

test/Benchmarks/OrleansBenchmarks/Program.cs
dVakulen: Fixed.

test/Benchmarks/OrleansBenchmarks/Program.cs
dVakulen: Fixed

test/Benchmarks/OrleansBenchmarkGrainInterfaces/MapReduce/DataflowGrainsInterfaces.cs
dVakulen: Fixed

test/Benchmarks/OrleansBenchmarks/project.json
dVakulen: Fixed

jdom:
LGTM. BTW, I tried running them locally, and I see nice and clear results for the Serialization tests, although not sure if the MapReduce are running to conclusion or what, since the output is just an elapsed time that looks handmade. Is that the expectation?

dVakulen:
Yes, BenchmarkDotNet doesn't appears to be well suited for this kind of tests, I was considering https://github.com/petabridge/NBench, but decided not to add additional dependency for now..

jdom:
@dotnet-bot test this please

jdom:
Thanks @dVakulen! It's great that we start adding these to github. We have some perf tests that we run on our side that would be a lot of baggage to just put in the public without a ton of rework. This is a nice fresh start to having publicly usable benchmarks.

Issue #2006, opened 2016-08-02
Amazon AWS integration


galvesribeiro:
This issue is part of #2005 by implementing a minimal set of providers to make Orleans to run well by integrating Amazon AWS services.
- [x] DynamoDB Grain State persistence provider (#2007)
- [x] DynamoDB Membership Provider (#2008) 
- [x] DynamoDB Reminder Provider (#2045) 
- [x] SQS Stream Provider (#2053) 
- [ ] DynamoDB Metrics and Statistics
- [ ] Kinesis Stream Provider

gabikliot:
Why do you think DynamoDB  is the right storage AWS layer to implement Membership  and Grain State provider? Why not SimpleDB? Or is it not popular any more?

gabikliot:
A couple of other questions:

1) What about the hosting model? Would you run a silo in a vanilla VM, started as a startup service or by some cron job?
2) Windows on AWS initially and then maybe Linux with Core CLR? In both cases, the hosting model can be identical for AWS and GCP.
3) What about telemetry/logging/monitoring?

galvesribeiro:
> Why do you think DynamoDB is the right storage AWS layer to implement Membership and Grain State provider? Why not SimpleDB? Or is it not popular any more?

We can always add SimpleDB provider to the list as well and implement it. SimpleDB is Amazon offer that compete with Azure Table Storage yes however, I've see people using DynamoDB more for NoSQL/DocDB today. It was not a preference choice. I just made it to help someone else but yes, I can help add that as well if you have interest in it.

> 1) What about the hosting model? Would you run a silo in a vanilla VM, started as a startup service or by some cron job?

Initially I'm going to lay down the foundation of a very basic providers. Mostly port what is on Azure Table Store providers to use DynamoDB but yes, run on VMs. After that, will look at extensibility points on Elastic Beanstalk hosting model and make something like what we have for Azure Cloud Services.

> 2) Windows on AWS initially and then maybe Linux with Core CLR? In both cases, the hosting model can be identical for AWS and GCP.
3) What about telemetry/logging/monitoring?

Given that AWS/GCP provide some PaaS hosting model for either Windows or Linux, I would explore that to make easy deployment as we have with Azure Cloud Services later on. .Net Core is something that will be pending until we finish the port.

galvesribeiro:
[...]

Issue #2005, opened 2016-08-02
Orleans and Cloud Service Providers


galvesribeiro:
Although Orleans can run perfectly on-premises and on any other Cloud Service Provider (CSP), I saw people asking about having a deeper/transparent integration with those CSPs by leverage their service offers.

This issue will track a series of issues that will implement minimal providers like the ones already created for Azure but instead, targeting other major CSPs:
- [ ] Amazon Web Services (AWS) - #2006 
- [ ] Google Cloud Platform (GCP) - #2057 

Feel free to make suggestions of other major CSPs to be included on this list or if you wish to contribute on those. Discussions about an specific CSP's provider implementation please keep inside its linked issue.

Lets put Orleans to run everywhere! :dancer:

gabikliot:
This is great! 
It is of course not new, we have been talking about it a lot in the past (e.g., #473). 
Also, some work have already been done in the GCP space (#1174 - Google Protobuf, part of OrleansGoogleUtils).
This line of work will take the framework to the next level and unblock a great potential of application mobility across cloud providers.

galvesribeiro:
@gabikliot Yep, I already saw the google part, I just started with AWS cause I have someone that want to use Orleans here but they are 100% on AWS so I just want to get them started asap.

When the time comes, I will get to google as well :)

galvesribeiro:
Ok... As much as I want that providers to reach the main branch and support the major CSP player, I'm dropping this issue and all its related PRs and closing it.

The reason is that after a long discussion with the core team (as much as I don't agree with the decision), the decision was to move it to OrleansContrib. By doing that, it will slow down the development of those components and I have no time/resource to invest on it. The reason it will slow down is that in order to properly test the whole providers we will need the internal tests from the main repo and we can't do that in a manageable way by having it on OrleansContrib. 

I feel sad about that and I still believe Orleans should reach other CSPs easily. Thanks to everyone who contributed with either comments or reviews on the related issues/PRs, but I'll drop this engagement since I have another higher priority and time consuming engagements on Orleans like .Net Core support.

I hope eveyone understand. Thanks.

gabikliot:
Really disappointed and frustrated by that decision as well. 
You listed all the exact reasons why I believe it should be in the main repo. 
We already have ZK, and Consul and SQL in the main repo. Why AWS/GCP are different?
Big mistake IMHO. 

[...]

PR #1682, opened 2016-04-13
Preliminary relational persistence queries

veikkoeeva:
An update storage provider for existing relational backends

* A storage provider for MySQL added and for SQL Server (hopefully) improved by easier setup, supported for any ADO.NET supported DB (initial scripts for SQL Server and MySQL).
* Includes support for changing serializers and deserializers at any level of granularity (meaning grain type, id, state data etc., [see tests](https://github.com/dotnet/orleans/pull/1682/commits/eac23b59f5e85fc37fed36d09fb9ce2eaefd0784#diff-e580375a3d638cc0620ee1a22d1d6fd5R108)). This means one can use any serializer or deserializer one wants and even change from one format to another.
* Supports evolving (or versions) or wholesale changing of data types.
* Allows downloading data as streams at any granularity (e.g. avoids LOH allocations and conserves resources otherwise too).
* Allows for sharding using the Orleans built-in sharding, hence supporting it for any ADO.NET backend (having scripts) and even for heterogenous vendor setup (even heterogenous storage setup, should be possible to slice both horizontally and vertically with some additional work).
* Supports multiple simultaneous deployments.
* In-storage special processing supported (e.g. XML and JSON for SQL Server).
* Jenkins hash function exposed and used to hash the index, wrapped to an interface so it can be changed and doesn't create backwards compability issue for the current implementation. This allows also the hashing functionality changed like serializers and deserializers.

Notes to follow...

Fixes #1176.

PR #2007, opened 2016-08-02
AWS DynamoDB Grain State Provider

galvesribeiro:

### DIFF
--- a/src/Orleans.sln
+++ b/src/Orleans.sln
@@ -64,6 +64,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Nuget", "Nuget", "{014CD19F
 		NuGet\Microsoft.Orleans.Core.nuspec = NuGet\Microsoft.Orleans.Core.nuspec
 		NuGet\Microsoft.Orleans.CounterControl.nuspec = NuGet\Microsoft.Orleans.CounterControl.nuspec
 		NuGet\Microsoft.Orleans.EventSourcing.nuspec = NuGet\Microsoft.Orleans.EventSourcing.nuspec
+		NuGet\Microsoft.Orleans.OrleansAWSUtils.nuspec = NuGet\Microsoft.Orleans.OrleansAWSUtils.nuspec
 		NuGet\Microsoft.Orleans.OrleansAzureUtils.nuspec = NuGet\Microsoft.Orleans.OrleansAzureUtils.nuspec
 		NuGet\Microsoft.Orleans.OrleansCodeGenerator.Build.nuspec = NuGet\Microsoft.Orleans.OrleansCodeGenerator.Build.nuspec
 		NuGet\Microsoft.Orleans.OrleansCodeGenerator.nuspec = NuGet\Microsoft.Orleans.OrleansCodeGenerator.nuspec
@@ -182,6 +183,8 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Extensions", "Extensions", "{F3C3FA92-FC69-4B94-8914-3B70E624B5B5}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OrleansPSUtils", "OrleansPSUtils\OrleansPSUtils.csproj", "{6AD37425-7CB4-4D23-80C3-A9D143329A66}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OrleansAWSUtils", "OrleansAWSUtils\OrleansAWSUtils.csproj", "{67738E6C-F292-46A2-994D-5B52E745205B}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -309,6 +312,10 @@ Global
 		{B99C744A-7F62-430C-9255-E64875D39486}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{B99C744A-7F62-430C-9255-E64875D39486}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{B99C744A-7F62-430C-9255-E64875D39486}.Release|Any CPU.Build.0 = Release|Any CPU
+		{67738E6C-F292-46A2-994D-5B52E745205B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{67738E6C-F292-46A2-994D-5B52E745205B}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{67738E6C-F292-46A2-994D-5B52E745205B}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{67738E6C-F292-46A2-994D-5B52E745205B}.Release|Any CPU.Build.0 = Release|Any CPU
 		{6AD37425-7CB4-4D23-80C3-A9D143329A66}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{6AD37425-7CB4-4D23-80C3-A9D143329A66}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{6AD37425-7CB4-4D23-80C3-A9D143329A66}.Release|Any CPU.ActiveCfg = Release|Any CPU
@@ -351,5 +358,6 @@ Global
 		{B99C744A-7F62-430C-9255-E64875D39486} = {01F3CC7E-F996-411E-AFD6-72673A826549}
 		{F3C3FA92-FC69-4B94-8914-3B70E624B5B5} = {4CD3AA9E-D937-48CA-BB6C-158E12257D23}
 		{6AD37425-7CB4-4D23-80C3-A9D143329A66} = {F3C3FA92-FC69-4B94-8914-3B70E624B5B5}
+		{67738E6C-F292-46A2-994D-5B52E745205B} = {F3C3FA92-FC69-4B94-8914-3B70E624B5B5}
 	EndGlobalSection
 EndGlobal
--- a/src/Orleans/Properties/AssemblyInfo.cs
+++ b/src/Orleans/Properties/AssemblyInfo.cs
@@ -24,6 +24,7 @@ using System.Runtime.InteropServices;
 [assembly: InternalsVisibleTo("OrleansRuntime")]
 [assembly: InternalsVisibleTo("OrleansHost")]
 [assembly: InternalsVisibleTo("OrleansAzureUtils")]
+[assembly: InternalsVisibleTo("OrleansAWSUtils")]
 [assembly: InternalsVisibleTo("OrleansManager")]
 [assembly: InternalsVisibleTo("LoadTestGrains")]
 [assembly: InternalsVisibleTo("UnitTests")]
@@ -32,4 +33,3 @@ using System.Runtime.InteropServices;
 [assembly: InternalsVisibleTo("TesterInternal")]
 [assembly: InternalsVisibleTo("TestInternalGrainInterfaces")]
 [assembly: InternalsVisibleTo("TestInternalGrains")]
-
--- a/src/OrleansAWSUtils/AWSUtils.cs
+++ b/src/OrleansAWSUtils/AWSUtils.cs
@@ -1,0 +1,66 @@
+﻿using Amazon;
+using System;
+
+namespace OrleansAWSUtils
+{
+    /// <summary>
+    /// Some basic utilities methods for AWS SDK
+    /// </summary>
+    internal static class AWSUtils
+    {
+        internal static RegionEndpoint GetRegionEndpoint(string zone = "")
+        {
+            switch (zone)
+            {
+                case "us-east-1":
+                    return RegionEndpoint.USEast1;
+                case "us-west-1":
+                    return RegionEndpoint.USWest1;
+                case "ap-south-1":
+                    return RegionEndpoint.APSouth1;
+                case "ap-northeast-2":
+                    return RegionEndpoint.APNortheast2;
+                case "ap-southeast-1":
+                    return RegionEndpoint.APSoutheast1;
+                case "ap-southeast-2":
+                    return RegionEndpoint.APSoutheast2;
+                case "ap-northeast-1":
+                    return RegionEndpoint.APNortheast1;
+                case "eu-central-1":
+                    return RegionEndpoint.EUCentral1;
+                case "eu-west-1":
+                    return RegionEndpoint.EUWest1;
+                case "sa-east-1":
+                    return RegionEndpoint.SAEast1;
+                default:
+                    return RegionEndpoint.USWest2;
+            }
+        }
+
+        /// <summary>
+        /// Validate DynamoDB PartitionKey.
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns></returns>
+        public static string ValidateDynamoDBPartitionKey(string key)
+        {
+            if (key.Length >= 2048)
+                throw new ArgumentException(string.Format("Key length {0} is too long to be an DynamoDB partition key. Key={1}", key.Length, key));
+
+            return key;
+        }
+
+        /// <summary>
+        /// Validate DynamoDB RowKey.
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns></returns>
+        public static string ValidateDynamoDBRowKey(string key)
+        {
+            if (key.Length >= 1024)
+                throw new ArgumentException(string.Format("Key length {0} is too long to be an DynamoDB row key. Key={1}", key.Length, key));
+
+            return key;
+        }
+    }
+}
--- a/src/OrleansAWSUtils/Properties/AssemblyInfo.cs
+++ b/src/OrleansAWSUtils/Properties/AssemblyInfo.cs
@@ -1,0 +1,25 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Orleans.CodeGeneration;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("OrleansAWSUtils")]
+[assembly: AssemblyDescription("Orleans - Windows AWS Helper Classes")]
+[assembly: AssemblyConfiguration("")]
+
+// Setting ComVisible to false makes the types in this assembly not visible 
+// to COM components.  If you need to access a type in this assembly from 
+// COM, set the ComVisible attribute to true on that type.
+[assembly: ComVisible(false)]
+
+// The following GUID is for the ID of the typelib if this project is exposed to COM
+[assembly: Guid("67738e6c-f292-46a2-994d-5b52e745205b")]
+
+[assembly: InternalsVisibleTo("UnitTests")]
+[assembly: InternalsVisibleTo("TesterInternal")]
+[assembly: InternalsVisibleTo("UnitTestGrains")]
+[assembly: SkipCodeGeneration]
+
--- a/src/OrleansAWSUtils/Storage/DynamoDBStorage.cs
+++ b/src/OrleansAWSUtils/Storage/DynamoDBStorage.cs
@@ -1,0 +1,548 @@
+﻿using Amazon.DynamoDBv2;
+using Amazon.DynamoDBv2.Model;
+using Amazon.Runtime;
+using Orleans;
+using Orleans.Runtime;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OrleansAWSUtils.Storage
+{
+    /// <summary>
+    /// Wrapper around AWS DynamoDB SDK.
+    /// </summary>
+    internal class DynamoDBStorage
+    {
+        private const string AccessKeyPropertyName = "AccessKey";
+        private const string SecretKeyPropertyName = "SecretKey";
+        private const string ServicePropertyName = "Service";
+        private const string ReadCapacityUnitsPropertyName = "ReadCapacityUnits";
+        private const string WriteCapacityUnitsPropertyName = "WriteCapacityUnits";
+
+        private string accessKey;
+        protected string secretKey;
+        private string service;
+        private int readCapacityUnits = 10;
+        private int writeCapacityUnits = 5;
+        private AmazonDynamoDBClient ddbClient;
+        private Logger Logger;
+
+        /// <summary>
+        /// Create a DynamoDBStorage instance
+        /// </summary>
+        /// <param name="dataConnectionString">The connection string to be parsed for DynamoDB connection settings</param>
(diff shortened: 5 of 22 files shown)


gabikliot:
I think there is duplication with the other pr.

galvesribeiro:
@gabikliot actually it is not duplicated... I made a mistake... I should wait this PR get merged before create a new branch for the other one. What I did was create a branch locally from this one so I can use the DynamoDB code into the Membership provider :(

If you can merge this one I can fix the other one.

Thanks

gabikliot:
I can review the membership provider. Someone else can review the storage. For me to review the membership provider, please make its pr clean and only include MBR provider code. I already posted some comments there.

galvesribeiro:
@gabikliot yea I'll make it clear but like I said, this PR need to be merged first since I need `DynamoDBDataManager<T>` to use it inside MBR on `OrleansSiloInstanceManager`.

Lets wait someone merge it here so I can work on the MBR one. Thanks

veikkoeeva:
In case work continues, some short notes I believe are applicable to this too: https://gitter.im/dotnet/orleans?at=57a35c77836d2d021163d9a6.

src/OrleansAWSUtils/Storage/DynamoDBDataManager.cs
```diff
@@ -0,0 +1,248 @@
...
+
+        public Task UpsertItemsAsync(IEnumerable<T> entries)
+        {
+            var config = new DynamoDBOperationConfig { ConsistentRead = true, SkipVersionCheck = true, OverrideTableName = TableName };
+            var batch = ddbContext.CreateBatchWrite<T>(config);
+            batch.AddPutItems(entries);
+            return batch.ExecuteAsync();
+        }
+
+        public Task UpsertItemAsync(T entry)
```
gabikliot: We need a separate Update which always relates to the previous version and updates the new version if previous was not changed
AND
Insert,, which always inserts if there is no prev version or fails otherwise.
Then any higher level retries should be done on top of it.

We can still have upsert for some situations when we don't care, but 95% of the time we do care.
In general, I always say that upsert is very harmfull and most of the time people use it wrongly instead of properly: try insert, if not read and try update. Otherwise, we run into a danger of a blink write.

Specifically, in both storage provider and MBR provider you will need Update, tryInsert and you will not need upsert.

src/OrleansAWSUtils/Storage/DynamoDBDataManager.cs
```diff
@@ -0,0 +1,248 @@
...
+
+        /// <summary> URL of the local DynamoDB instance or the AWS region. </summary>
+        protected string Service { get; set; }
+
+        private int readCapacityUnits;
+        private int writeCapacityUnits;
+
+        private DynamoDBContext ddbContext;
+        private AmazonDynamoDBClient ddbClient;
+        private DynamoDBOperationConfig operationConfig;
```
gabikliot: looks like this could be static, right?

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,334 @@
...
+            }
+
+            public Task<GrainStateRecord> Read(string partitionKey, string rowKey)
+            {
+                return tableManager.ReadSingleTableEntryAsync(partitionKey, rowKey);
+            }
+
+            public Task Write(GrainStateRecord record)
+            {
+                return tableManager.UpsertItemAsync(record);
```
gabikliot: I think this is wrong. You need to call Update here.

gabikliot: Look at https://github.com/dotnet/orleans/blob/master/src/OrleansAzureUtils/Providers/Storage/AzureTableStorage.cs#L496.
There is a good reason it does create OR update, but not upsert.
We want to fail if we try to do create but it already exists. Otherwise, we can overwrite by mistake.

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,334 @@
...
+                {
+                    record.StringData = string.Empty;
+                    record.Data = null;
+                    await tableDataManager.Write(record).ConfigureAwait(false);
+                    grainState.ETag = record.Version.HasValue ? record.Version.ToString() : string.Empty; // Update in-memory data to the new ETag
+                }
+            }
+            catch (Exception)
+            {
+                throw;
```
gabikliot: ??

src/OrleansAWSUtils/Storage/DynamoDBDataManager.cs
shayhatsor: Adding to @gabikliot note. When developing a storage provider it's important to think of what happens on "duplicate activation". If two duplicate grains attempt a write to the db at the same time, one of them must fail. This is easily achieved by making sure all write operations send the expected version with the data. I think of it as a conditional upsert operation.

src/Orleans.sln
```diff
@@ -64,6 +64,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Nuget", "Nuget", "{014CD19F
 		NuGet\Microsoft.Orleans.Core.nuspec = NuGet\Microsoft.Orleans.Core.nuspec
 		NuGet\Microsoft.Orleans.CounterControl.nuspec = NuGet\Microsoft.Orleans.CounterControl.nuspec
 		NuGet\Microsoft.Orleans.EventSourcing.nuspec = NuGet\Microsoft.Orleans.EventSourcing.nuspec
+		Microsoft.Orleans.OrleansAWSUtils.nuspec = Microsoft.Orleans.OrleansAWSUtils.nuspec
```
shayhatsor: all others are under `Nuget` folder

src/OrleansAWSUtils/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Orleans.CodeGeneration;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("OrleansAWSUtils")]
+[assembly: AssemblyDescription("Orleans - Windows Azure Helper Classes")]
```
shayhatsor: fix here

galvesribeiro:
Guys, dont review the current code. I had changed it and will let you know whenever I get it reopened probably later today. Thanks

galvesribeiro:
image

Ok, @gabikliot @shayhatsor (and anyone who can) now it is ready for review. Remember that this issue is just the Grain State storage provider and the core DataManager that will be used on other components like MBR, Reminder, etc.

As soon as I got feedback I'll update the PR with the changes. 

I'll wait this PR get completed before I start the refactory on the MBR one.

Thanks!

galvesribeiro:
BTW, if anyone want to run the tests, just follow the instructions [here](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html) that explain how to quickly download and run DynamoDB. No install required. Just unzip and call the Java commandline as explined.

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,415 @@
...
+using System.Threading.Tasks;
+
+namespace Orleans.Storage
+{
+    /// <summary>
+    /// AWS DynamoDB basic Storage provider
+    /// </summary>
+    public class DynamoDBStorage : IStorageProvider
+    {
+        internal const string TableNameDefaultValue = "OrleansGrainState";
```
shayhatsor: a suggestion, move all the configuration parsing to `DynamoDBStorageConfig` in order to separate concerns

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,415 @@
...
+        {
+            tableDataManager = null;
+            return TaskDone.Done;
+        }
+
+        /// <summary> Read state data function for this storage provider. </summary>
+        /// <see cref="IStorageProvider.ReadStateAsync"/>
+        public async Task ReadStateAsync(string grainType, GrainReference grainReference, IGrainState grainState)
+        {
+            if (tableDataManager == null) throw new ArgumentException("GrainState-Table property not initialized");
```
shayhatsor: I'm not sure this is needed, here and in other places. even then, it's not an `ArgumentException`

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,415 @@
...
+            if ((pkSize + rkSize + versionSize + dataSize) > MAX_DATA_SIZE)
+            {
+                var msg = string.Format("Data too large to write to DynamoDB table. Size={0} MaxSize={1}", dataSize, MAX_DATA_SIZE);
+                throw new ArgumentOutOfRangeException("GrainState.Size", msg);
+            }
+        }
+
+        private string GetKeyString(GrainReference grainReference)
+        {
+            var key = string.Format("{0}_{1}", serviceId, grainReference.ToKeyString());
```
shayhatsor: The implementation here and in similar places is the same as the Azure one. My knowledge of Azure Table or DynamoDB is close to nothing, so I don't know how similar they are. I think that before merging this, we must convince ourselves that following the Azure oriented implementation gives the best throughput in DynamoDB too.

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,415 @@
...
+                Name, serviceId, tableName, isDeleteStateOnClear);
+
+            if (config.Properties.ContainsKey(UseJsonFormatPropertyName))
+                useJsonFormat = "true".Equals(config.Properties[UseJsonFormatPropertyName], StringComparison.OrdinalIgnoreCase);
+
+            this.jsonSettings = SerializationManager.UpdateSerializerSettings(SerializationManager.GetDefaultJsonSerializerSettings(), config);
+
+            initMsg = String.Format("{0} UseJsonFormat={1}", initMsg, useJsonFormat);
+
+            Log.Info(ErrorCode.StorageProviderBase, "AWS DynamoDB Provider: {0}", initMsg);
```
shayhatsor: I suggest you either use the Azure error codes (like in the previous version of this PR) or copy paste and rename them like we did in the ado.net storage provider PR. If you feel like fixing this the right way, the best would be to merge all the storage provider errors and put them in the Orleans providers project. see https://github.com/dotnet/orleans/pull/1682#issuecomment-236450895. the important thing is that you shouldn't use the base error because it'd make it harder later to refactor.

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,415 @@
...
+                    throw;
+                }
+            }
+
+            public async Task Write(GrainStateRecord record)
+            {
+                if (logger.IsVerbose3) logger.Verbose3(ErrorCode.StorageProviderBase, "Writing: PartitionKey={0} RowKey={1} to Table={2} with ETag={3}", record.PartitionKey, record.RowKey, TableName, record.ETag);
+                string eTag = record.ETag == 0 ?
+                    await tableManager.CreateTableEntryAsync(record).ConfigureAwait(false) :
+                    await tableManager.UpdateTableEntryAsync(record, record.ETag.ToString()).ConfigureAwait(false);
```
shayhatsor: You have to wrap this with a try and catch and rethrow an `InconsistentStateException` when the write failed because the row already exists or there was a version conflict. Looking at the Azure code, this isn't implemented there, which is a bug.

a note for @veikkoeeva : having the insert and update separate will simplify the db scripts greatly (no need to lock anything). I don't know how I didn't think of it before.

shayhatsor:
@galvesribeiro, about the tests you've added. How are they different than the tests already found in Orleans? Do they test specific DynamoDB issues ? I suggest making base classes for these tests. In general, the similarity between the Azure Table implementation and the DynamoDB seems like a good candidate for refactoring to base classes. What do you think?

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
galvesribeiro: OK

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
galvesribeiro: OK

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
galvesribeiro: DDB is pretty much same thing then Azure tables, just in AWS.  Many of the rules on Azure apply to there and many parts of the code will remain the same which looks to me ok.

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
galvesribeiro: That ErrorCode enum is something very weird to me. It contains error codes from Azure so I put this instead.

I think unless the errors are REALLY uniform across all providers (which should contain a set of common error codes) each provider should have its own error codes. 

I'll create a separated PR after all the providers for AWS are done and before we start creating Nugets, only enhancing the logging across all AWS providers.

galvesribeiro:
@shayhatsor The tests are most of them the same. Some I added, some I changed.

I already talked with @jdom while doing the xunit migration that we should have the black box tests migrated to theories so we test against the abstraction/interface surface with multiple provider implementation. But again, this is indeed a refactory we can do, but I think it is not the focus of this PR.

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
jdom: I agree you shouldn't use the base number

veikkoeeva: In ADO.NET the base is moved higher than what it is in ATS or Blob storage (for reference [here](https://github.com/dotnet/orleans/blob/master/src/OrleansSQLUtils/Storage/Provider/AdoNetStorageProvider.cs#L28)). I tried to make it so they match the Azure ones, just with higher base.

jdom:
We will not use theories, as each provider should be able to be developed independent from the rest of the solution, and even if it was, each of them requires a different fixture for set up. We will be moving to a test kit as explained in #2011. The more we add to the same Tester project, the harder it is to port to coreclr by partitioning the work. If this PR doesn't address the test kit refactoring, we can do that later anyway, as a prereq for the migration to coreclr

src/Orleans/Properties/AssemblyInfo.cs
```diff
@@ -24,12 +24,12 @@
 [assembly: InternalsVisibleTo("OrleansRuntime")]
 [assembly: InternalsVisibleTo("OrleansHost")]
 [assembly: InternalsVisibleTo("OrleansAzureUtils")]
+[assembly: InternalsVisibleTo("OrleansAWSUtils")]
```
jdom: Why is this needed? In newer providers (such as the event hub stream provider) we started doing some work so that it doesn't require any internals, and any external provider could also be developed in the same way as built in ones. Is it possible to avoid this, and if not, let's try to find a way to make it better. 
The event hub provider did require some refactoring of the main Orleans assemblies, but in the end it improved them for all stream providers.

galvesribeiro:
Ok. Reset on master and made some fixes. The other enhancements suggested here comes with other PR before we start to generate Nugets to the whole OrleansAWSUtils.

src/Orleans/Properties/AssemblyInfo.cs
galvesribeiro: Right now I just had to add that since I'm using ErrorCode enum. I'll create a new enum with all the AWS specific error codes once we finish the review of all the PRs so before we release the code thiss will be out.

src/OrleansAWSUtils/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,430 @@
...
+        {
+            var parameters = dataConnectionString.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+
+            var serviceConfig = parameters.Where(p => p.Contains(ServicePropertyName)).FirstOrDefault();
+            if (!string.IsNullOrWhiteSpace(serviceConfig))
+            {
+                var value = serviceConfig.Split(new[] { '=' }, StringSplitOptions.RemoveEmptyEntries);
+                if (value.Length == 2 && !string.IsNullOrWhiteSpace(value[1]))
+                    service = value[1];
+            }
```
shayhatsor: this block, including the line above it should be refactored as a method. now it's just duplicated 4 times

src/OrleansAWSUtils/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,430 @@
...
+                if (description.Table != null)
+                    return description.Table;
+            }
+            catch (ResourceNotFoundException)
+            {
+                return null;
+            }
+            catch (Exception)
+            {
+                throw;
```
shayhatsor: ?

test/TestGrainInterfaces/IPersistenceTestGrains.cs
```diff
@@ -19,7 +19,7 @@ public interface IPersistenceTestGrain : IGrainWithGuidKey
     }
 
 
-    public interface IPersistenceTestGenericGrain<T> : IPersistenceTestGrain // IGrainWithGuidKey
+    public interface IPersistenceTestGenericGrain<T> : IPersistenceTestGrain // IGrainWithGuidKeyaws
     { }
```
shayhatsor: ?

src/OrleansAWSUtils/Storage/Provider/DynamoDBStorageProvider.cs
```diff
@@ -0,0 +1,340 @@
...
+using Orleans.Serialization;
+using Amazon.DynamoDBv2.Model;
+using Amazon.DynamoDBv2;
+using OrleansAWSUtils;
+using System.IO;
+using OrleansAWSUtils.Storage;
+
+namespace Orleans.Storage
+{
+    public class DynamoDBStorageProvider : IStorageProvider
```
shayhatsor: all public entities must have at least summary. Also, please add some design rationale, why DynamoDB, implementation details and specifics, why you chose certain defaults and what do they mean, etc.

src/Orleans/Properties/AssemblyInfo.cs
```diff
@@ -24,12 +24,12 @@
...
 [assembly: InternalsVisibleTo("OrleansManager")]
 [assembly: InternalsVisibleTo("LoadTestGrains")]
 [assembly: InternalsVisibleTo("UnitTests")]
 [assembly: InternalsVisibleTo("UnitTestGrainInterfaces")]
 [assembly: InternalsVisibleTo("UnitTestGrains")]
 [assembly: InternalsVisibleTo("TesterInternal")]
 [assembly: InternalsVisibleTo("TestInternalGrainInterfaces")]
-[assembly: InternalsVisibleTo("TestInternalGrains")]
-
+[assembly: InternalsVisibleTo("TestInternalGrains")]
```
shayhatsor: ?

src/OrleansAWSUtils/AWSUtils.cs
```diff
@@ -0,0 +1,66 @@
+﻿using Amazon;
+using System;
+
+namespace OrleansAWSUtils
+{
+    /// <summary>
+    /// Some basic utilities methods for AWS SDK
+    /// </summary>
+    public static class AWSUtils
```
shayhatsor: can this be internal?

src/OrleansAWSUtils/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,430 @@
...
+            {
+                var response = await ddbClient.CreateTableAsync(request);
+                TableDescription description = null;
+                do
+                {
+                    description = await GetTableDescription(tableName);
+
+                    await Task.Delay(2000);
+
+                } while (description.TableStatus != TableStatus.ACTIVE);
```
shayhatsor: can this go on forever ?

src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
galvesribeiro: Done. The parse is only on DynamoDBStorage for all fields of the connection strings and not spread everywhere.

test/TestGrainInterfaces/IPersistenceTestGrains.cs
galvesribeiro: Ouch, let me revert that.

src/OrleansAWSUtils/Storage/Provider/DynamoDBStorageProvider.cs
galvesribeiro: All provider implementations are never accessed directly so no-one will ever read comments. The only reason it is public is so it can be accessed by Orleans internals. I think is more useful a better documentation on the docs page.

src/Orleans/Properties/AssemblyInfo.cs
galvesribeiro: When added `[assembly: InternalsVisibleTo("OrleansAWSUtils")]` just removed an empty line. Not that it worth...

src/OrleansAWSUtils/AWSUtils.cs
galvesribeiro: Done!

src/OrleansAWSUtils/Storage/DynamoDBStorage.cs
galvesribeiro: Nop, it throw an exception and in 99.99% of the cases the loop never execute twice but, AWS docs say that we have to wait for active. I just changed it a bit to make the loop only run while `TableStatus.CREATING`.

src/OrleansAWSUtils/Storage/Provider/DynamoDBStorageProvider.cs
shayhatsor: When someone other than you tries to figure out your thought process when implementing this, they won't go to the Orleans documentation for details. Please add a short summary like I mentioned, I agree that a more inclusive and exhaustive documentation should be added to Orleans documentation. Also, having public entities without a summary raises compilation warnings in Orleans.

shayhatsor:
@galvesribeiro, thanks for this great contribution !

galvesribeiro:
@shayhatsor Thanks for having time to review and for the feedback :)

PR #2008, opened 2016-08-02
AWS DynamoDB Membership Provider

galvesribeiro:


src/OrleansAWSUtils/Providers/Storage/DynamoDBStorage.cs
```diff
@@ -0,0 +1,334 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Orleans.Storage
+{
+    /// <summary>
+    /// AWS DynamoDB basic Storage provider
+    /// </summary>
+    public class DynamoDBStorage : IStorageProvider
```
gabikliot: Why storage provider is part of membership table pull request? I thought you had a separate pr for it.

src/OrleansAWSUtils/Providers/Membership/OrleansSiloInstanceManager.cs
```diff
@@ -0,0 +1,278 @@
...
+        /// <summary>
+        /// Conditionally update the row for this entry, but only if the version matches with the current record in data store
+        /// </summary>
+        /// <param name="siloEntry">Silo Entry to be written</param>
+        /// <param name="tableVersionEntry">Version row to update</param>
+        /// <returns></returns>
+        internal async Task<bool> UpdateSiloEntryConditionally(SiloInstanceTableEntry siloEntry, SiloInstanceTableEntry tableVersionEntry)
+        {
+            await storage.UpsertItemAsync(tableVersionEntry);
+            await storage.UpsertItemAsync(siloEntry);
```
gabikliot: This should not be upsert. It should be update conditional conditioned on stage/version. Also the two updates should be applied atomically, in the same batch transaction.

gabikliot:
@galvesribeiro , my recommendation (based on https://github.com/dotnet/orleans/issues/2005#issuecomment-237136674 ) is to reopen this PR, remove all the code unrelated to MBR provider (remove storage provider) and I will review it thoroughly.

galvesribeiro:
Ok, this PR has the final membership code. All the changes on it are on `OrleansAWSUtils\Membership` and `OrleansAWSUtils\Statistics`. 

This PR still need be rebased on master once #2007 is merged since it depends on a common class (`Storage\DynamoDBStorage.cs`). All the tests for either the MembershipTable and Liveness were included and are passing.

galvesribeiro:
@gabikliot rebased on latest master which where the Storage Provider is already merged. Can you review it please?

Btw, kudos to @shayhatsor for the tips on the membership protocol :)

gabikliot:
Please separate and remove the statistics/metrics from this PR to a sepaarte PR.

galvesribeiro:
What is the problem with that? The PR was intent to be with it. It is 3 no-brain classes...

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+                var keys = new Dictionary<string, AttributeValue> { { $":{SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME}", new AttributeValue(deploymentId) } };
+                var records = await storage.QueryAsync(TABLE_NAME_DEFAULT_VALUE, keys, $"{SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME} = :{SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME}", item => new SiloInstanceRecord(item));
+
+                var toDelete = new List<Dictionary<string, AttributeValue>>();
+                foreach (var record in records)
+                {
+                    toDelete.Add(record.GetKeys());
+                }
+
+                if (records.Count <= 25)
```
gabikliot: I don't think you need a special case here, the general loop below is good. Also, 25? at least lets use constant.

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        public async Task<bool> UpdateRow(MembershipEntry entry, string etag, TableVersion tableVersion)
+        {
+            try
+            {
+                if (logger.IsVerbose) logger.Verbose("UpdateRow entry = {0}, etag = {1}", entry.ToFullString(), etag);
+                var siloEntry = Convert(entry);
+                int currentEtag = 0;
+                int.TryParse(etag, out currentEtag);
```
gabikliot: and what if TryParse failed?

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+                siloEntry.ETag = currentEtag;
+                siloEntry.ETag++;
+
+                bool result;
+
+                try
+                {
+                    var conditionalValues = new Dictionary<string, AttributeValue> { { CURRENT_ETAG_ALIAS, new AttributeValue { N = etag } } };
+                    var expression = $"{SiloInstanceRecord.ETAG_PROPERTY_NAME} = {CURRENT_ETAG_ALIAS}";
+                    await storage.UpsertEntryAsync(TABLE_NAME_DEFAULT_VALUE, siloEntry.GetKeys(),
```
gabikliot: I wrote before in another PR or in the prev. version of this PR about UpsertEntryAsync. I think your current code is probably correct, but my recommendation is not to use Upsert. Instead, use Insert (try insert which will failed if the row is already there) and Update which always updates and never inserts.

Upsert by itself CAN be used correctly, in theory, but I think most of the people will use it wrongly, and thus I also suggest to remove it from the AWS storage API and use TryInsert and TryUpdate instead. Upsert if one of those convenience shortcut methods, which are very dangerous IMHO.

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+
+        public async Task UpdateIAmAlive(MembershipEntry entry)
+        {
+            try
+            {
+                if (logger.IsVerbose) logger.Verbose("Merge entry = {0}", entry.ToFullString());
+                var siloEntry = ConvertPartial(entry);
+                siloEntry.ETag++;
+                var expression = $"attribute_exists({SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME}) AND attribute_exists({SiloInstanceRecord.SILO_IDENTITY_PROPERTY_NAME})";
+                await storage.UpsertEntryAsync(TABLE_NAME_DEFAULT_VALUE, siloEntry.GetKeys(), siloEntry.GetFields(), expression);
```
gabikliot: would that Upsert  NEVER touch any of the other fields except for 3 that are returned from ConvertPartial? Will it just overwrite the IAmAklive?
If yes, then why do you do:  siloEntry.ETag++;?
if it updated the etag/version, it means it also needs to check the previous version?

src/OrleansAWSUtils/Membership/SiloInstanceRecord.cs
```diff
@@ -0,0 +1,210 @@
...
+        public string StartTime { get; set; }
+        public string IAmAliveTime { get; set; }
+        public int ETag { get; set; }        
+
+        internal static SiloAddress UnpackRowKey(string rowKey)
+        {
+            var debugInfo = "UnpackRowKey";
+            try
+            {
+#if DEBUG
```
gabikliot: we are trying not to use compile time DEBUG.  Not a good C# practice in general I think.

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
```diff
@@ -102,43 +103,56 @@ protected virtual string GetAdoInvariant()
...
         }
 
-        protected async Task MembershipTable_ReadAll_EmptyTable()
+        protected async Task MembershipTable_ReadAll_EmptyTable(bool extendedProtocol = true)
         {
             var data = await membershipTable.ReadAll();
             Assert.NotNull(data);
 
-            logger.Info("Membership.ReadAll returned VableVersion={0} Data={1}", data.Version, data);
+            if (extendedProtocol)
```
gabikliot: I understand that your did not implement the extendedProtocol . That is fine, but it does not mean you don't want those checks. You do want most of those. Like that particular one `Assert.NotNull(data.Version.VersionEtag);` - yes, you do want this check. There should still be a valid non null Version. I actually think that you will need most of the checks that you now disabled.

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+                if (logger.IsVerbose) logger.Verbose("InsertRow entry = {0}", entry.ToFullString());
+                var tableEntry = Convert(entry);
+
+                bool result;
+
+                try
+                {
+                    var expression = $"attribute_not_exists({SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME}) AND attribute_not_exists({SiloInstanceRecord.SILO_IDENTITY_PROPERTY_NAME})";
+
+                    await storage.PutEntryAsync(TABLE_NAME_DEFAULT_VALUE, tableEntry.GetFields(true), expression);
```
gabikliot: would that fail if row already exists?

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+
+        public async Task<bool> UpdateRow(MembershipEntry entry, string etag, TableVersion tableVersion)
+        {
+            try
+            {
+                if (logger.IsVerbose) logger.Verbose("UpdateRow entry = {0}, etag = {1}", entry.ToFullString(), etag);
+                var siloEntry = Convert(entry);
+                int currentEtag = 0;
+                int.TryParse(etag, out currentEtag);
+                siloEntry.ETag = currentEtag;
```
gabikliot: siloEntry.ETag = currentEtag + 1; and comment - set the NEXT etag.

gabikliot:
It is no brainier for you. For me, as a reviewer that takes responsibility on what he reviews, that makes MY life much harder. Specifically, I want to review the MBR now with 100% attention. The extra unrelated code confuses me.
Plus, I may not even review the metrics and stats providers, so I want it to be very clear what I did review and what not.
Plus, it is just a good practice to keep unrelated features in unrelated PRs. Each provider is definitely unrelated from each other.

gabikliot:
In your AWS Storage Put method:
https://github.com/dotnet/orleans/blob/master/src/OrleansAWSUtils/Storage/DynamoDBStorage.cs#L208

What happens if you call this method without conditionExpression and without conditionValues  - that is: you THINK there is no previous version of the row, but when the call arrives to AWS the row is there. Would the call overwrite the prev data or would it fail?
If it does not fail, it meansyou have a blind write method, which is ALWAYS WRONG.

gabikliot:
I think you have to go back and add detailed explanation on the AWS storage class about concurrency semantics of each operation.
https://github.com/dotnet/orleans/blob/master/src/OrleansAWSUtils/Storage/DynamoDBStorage.cs#L201 "Create or Replace an entry in a DynamoDB Table" but conditional is optional? I don t think it should be optional.
If you want a blind write method, add another method and call it this way, but in 99% of cases we will be using CreateIfNotExist with valid condition and Update with valid condition.

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+                    await storage.PutEntryAsync(TABLE_NAME_DEFAULT_VALUE, tableEntry.GetFields(true), expression);
+                    result = true;
+                }
+                catch (ConditionalCheckFailedException)
+                {
+                    result = false;
+                }
+
+                if (result == false)
+                    logger.Warn(ErrorCode.MembershipBase,
```
veikkoeeva: Could this logging be moved inside the preceding catch clause?

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+                int.TryParse(etag, out currentEtag);
+                siloEntry.ETag = currentEtag;
+                siloEntry.ETag++;
+
+                bool result;
+
+                try
+                {
+                    var conditionalValues = new Dictionary<string, AttributeValue> { { CURRENT_ETAG_ALIAS, new AttributeValue { N = etag } } };
+                    var expression = $"{SiloInstanceRecord.ETAG_PROPERTY_NAME} = {CURRENT_ETAG_ALIAS}";
```
veikkoeeva: This generic `expression` could have a more descriptive name.

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,350 @@
...
+                        siloEntry.GetFields(), expression, conditionalValues);
+
+                    result = true;
+                }
+                catch (ConditionalCheckFailedException)
+                {
+                    result = false;
+                }
+
+                if (result == false)
```
veikkoeeva: Could this logging be moved inside the preceding catch?

src/OrleansAWSUtils/Membership/SiloInstanceRecord.cs
veikkoeeva: Could these be logged on `Debug` level. Maybe the idea here is that this class doesn't take logger as a parameter, but maybe the code using this could log relevant pieces.

src/OrleansAWSUtils/Membership/SiloInstanceRecord.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (fields.ContainsKey(I_AM_ALIVE_TIME_PROPERTY_NAME))
+                IAmAliveTime = fields[I_AM_ALIVE_TIME_PROPERTY_NAME].S;
+
+            int etag;
+            if (fields.ContainsKey(ETAG_PROPERTY_NAME) &&
+                int.TryParse(fields[ETAG_PROPERTY_NAME].N, out etag))
+                ETag = etag;
+        }
+        
+        public string DeploymentId { get; set; }
```
veikkoeeva: I don't know if there's any specific rule in the project about member variables, but I personally believe not placing them in the beginning of the class declaration and instead in the body makes it harder to grasp the size and shape of the class.

src/OrleansAWSUtils/Statistics/DynamoDBStatisticsPublisher.cs
```diff
@@ -0,0 +1,138 @@
...
+        private const string STATVALUE_PROPERTY_NAME = "StatValue";
+        private const string ISDELTA_PROPERTY_NAME = "IsDelta";
+        private const string DATE_TIME_FORMAT = "yyyy-MM-dd-" + "HH:mm:ss.fff 'GMT'";
+
+        private string deploymentId;
+        private string address;
+        private string name;
+        private bool isSilo;
+        private long clientEpoch;
+        private int counter;
```
veikkoeeva: What does this `counter` count?

src/OrleansAWSUtils/Statistics/DynamoDBStatisticsPublisher.cs
```diff
@@ -0,0 +1,138 @@
...
+                    new AttributeDefinition { AttributeName = ROW_KEY_PROPERTY_NAME, AttributeType = ScalarAttributeType.S }
+               });
+        }
+
+        public Task ReportStats(List<ICounter> statsCounters)
+        {
+            try
+            {
+                var toWrite = new List<Dictionary<string, AttributeValue>>();
+                foreach (var counter in statsCounters)
```
veikkoeeva: I don't have a compiler here now, but doesn't this trigger a compiler warning due to a similarly named member variable? Maybe `foreach(var statsCounter in statsCounters)`?

src/OrleansAWSUtils/Statistics/DynamoDBStatisticsPublisher.cs
```diff
@@ -0,0 +1,138 @@
...
+                {
+                    var fields = ParseCounter(counter);
+                    if (fields == null)
```
veikkoeeva: If this condition were inverted and `toWrite.Add(fields)` put in, you wouldn't need the `continue`.

src/OrleansAWSUtils/Statistics/DynamoDBStatisticsPublisher.cs
```diff
@@ -0,0 +1,138 @@
...
+                        continue;
+
+                    toWrite.Add(fields);
+                }
+
+                if (toWrite.Count <= 25)
```
veikkoeeva: What is the thing with `25`? Maybe something like `const int TheThingWith25 = 25`?

test/TesterInternal/MembershipTests/DynamoDBMembershipTableTest.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Threading.Tasks;
+using UnitTests.StorageTests.AWSUtils;
+using Xunit;
+
+namespace UnitTests.MembershipTests
+{
+    /// <summary>
+    /// Tests for operation of Orleans Membership Table using AWS DynamoDB - Requires access to external DynamoDB storage
+    /// </summary>
+    public class DynamoDBMembershipTableTest : MembershipTableTestsBase, IClassFixture<DynamoDBStorageTestsFixture>
```
veikkoeeva: Would these tests show as failed or skipped with a reason no AWS facilities found when I run the tests if I don't have AWS systems on my machine? If I don't have the libraries, how long would it take for either result to materialize (i.e. would the test be failed or skipped quickly)?

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
galvesribeiro: DynamoDB has a MAX_BATCH_SIZE of 25. Ok, createt a constant and removed the IF.

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
galvesribeiro: I agree with you that it can be dangerous but, if I do the way you are asking, it will add a bunch of boilerplate non-reusable code just as it was in Azure Table Storage provider. I'll explain why in another comment here.

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
galvesribeiro: Yes you are right. I should don't bump the ETag in that case and only update the IAmAlive field. Just changed. More details on the upsert later on.

src/OrleansAWSUtils/Membership/SiloInstanceRecord.cs
galvesribeiro: Ok, removing that. Was just a remaining copy from the Azure core. I wish I hadn't did that...

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
galvesribeiro: Yes, will explain more about it...

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
galvesribeiro: Done

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
galvesribeiro: Done

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
galvesribeiro: Done

src/OrleansAWSUtils/Membership/SiloInstanceRecord.cs
galvesribeiro: Done

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
galvesribeiro: You are right. While rerunning the tests, only the ones that increment and validates the increment of the tableVersion row fail. I'm re-enabling those tests and leaving the flag only on those who fail.

test/TesterInternal/MembershipTests/DynamoDBMembershipTableTest.cs
galvesribeiro: It should fail quickly. However, it is not so easy as in Azure or SQL to detect if DynamoDB local instance is "installed" since there is nothing to install. So basically all the tests will quickly fail as they can't connect to DynamoDB. Hard to tell  it to skip and that is why it is (as any other provider) not a BVT test and has its own category so people going to test it make sure to have the correct infrastructure before start.

galvesribeiro:
Ok... All the comments addressed. 

@veikkoeeva thanks for get back on the statistics but, I'm removing that from this PR as @gabikliot requested and will put another one with those and I appreciate if you can have some time to review that.

I noticed that probably there is a lack of knowledge of how DynamoDB works which is ok and I should had it detailed on the PR before submit it so let me explain a bit here...

DynamoDB has only 3 Write operations:

1. PUT ==  Create/Insert or Replace
2. UPDATE == Update or Create/Insert
3. DELETE

All of them doesn't deal with ETag natively (unlike in Azure Table Storage or auto-increment fields in SQL which has native ways to deal with it) and any kind of conditional operation isn't implicitly implied. So we have to deal with it by using Conditional Expressions that you can see on those string around the code.

Basically, if you want to PUT something and make sure there is no other row with the same ID (as we do while Creating a new record) we need to use an expression like that:

``` csharp
var expression = $"attribute_not_exists({SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME}) AND attribute_not_exists({SiloInstanceRecord.SILO_IDENTITY_PROPERTY_NAME})";
await storage.PutEntryAsync(TABLE_NAME_DEFAULT_VALUE, tableEntry.GetFields(true), expression);
```
So if there is a row with the same keys as the ones in the expression, it will throw `ConditionalCheckFailedException`.

Otherwise, if you want to replace it (blind write), just call Put without any conditional expression.

The same happens to the case where we need to check the ETag.

```csharp
var etagConditionalExpression = $"{SiloInstanceRecord.ETAG_PROPERTY_NAME} = {CURRENT_ETAG_ALIAS}";
await storage.UpsertEntryAsync(TABLE_NAME_DEFAULT_VALUE, siloEntry.GetKeys(),
                        siloEntry.GetFields(), etagConditionalExpression, conditionalValues);
```

Expressions are optional and that is the way AWS works. I understand you concern about the dangerous of the blind write. But, think that "just because USA allow people to carry guns, it doesn't mean they should shoot each other". The same should apply here. All the classes are internal, and the provider is only accessible by Orleans Runtime and never directly by the users. The `DynamoDBStorage` makes no assumptions of whether the user want or not have an ETag, neither if it should create or update. It is just a thin easy wrapper on top of AWS DynamoDB SDK which is reused by all providers that want to. If the consumer want to enforce ETag conditional, it can do so by passing the parameters as expected and the expression will be validated by DynamoDB runtime and throw the appropriate exception in case the condition is not met.

IMHO blind write is not ALWAYS wrong. It is a matter of decision. If you need a condition to be met? Just express it. Remember that SQL can delete all the records of a single table if you don't specify a `WHERE` condition in it. It doesn't mean people can't want to delete a whole table without a condition. Its a matter of judgement for your case what are the conditions (if any) and express it. In our case, all the conditions are expressed using the provided APIs from AWS and after consulting them and with @shayhatsor which basically saw all the steps while I was building this code, I came up with the code.

Please let me know if you need anything else to change or more clarification on how DynamoDB works.

test/TesterInternal/MembershipTests/DynamoDBMembershipTableTest.cs
veikkoeeva: I believe one way of doing this would be to try to reach *DynamoDb* with the default connection and if it fails, skip the tests using `SkippableFact`. The explanation could be the database was unreachable, likely not present.

However, this is not my call. It is only my concern that I would like to see yellow for tests with unmet environment setup instead of red and without *being in the know about it*. Maybe this could be put also so that I have a creeping nagging feeling we're accumulating test code that's mostly red to most of the people and has duplication in code and in structure.

Then optinally it would be nice to have a way to override environment setup per test-site (i.e. if I have a differing setup on my rig). Not impossible, but not something we have at the moment either.

veikkoeeva:
@galvesribeiro Sure, no problem for the other PR. :) About my comments, it would help to have more comments for people who'd like to see how things are implemented. Then variables named according to their specific purpose and why some default numbers have been chosen maybe with just some appropriately named constant. However, this again is only how I feel.

galvesribeiro:
@veikkoeeva  sure! I addressed all the comments you made. :)

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
```diff
@@ -0,0 +1,347 @@
...
+                }
+
+                siloEntry.ETag = currentEtag + 1;
+
+                bool result;
+
+                try
+                {
+                    var conditionalValues = new Dictionary<string, AttributeValue> { { CURRENT_ETAG_ALIAS, new AttributeValue { N = etag } } };
+                    var etagConditionalExpression = $"{SiloInstanceRecord.ETAG_PROPERTY_NAME} = {CURRENT_ETAG_ALIAS}";
```
gabikliot: should not conditional expression here include also, in adddition to etag, $"attribute_exists({SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME}) AND attribute_exists({SiloInstanceRecord.SILO_IDENTITY_PROPERTY_NAME})";
?

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
```diff
@@ -163,16 +176,20 @@ protected virtual string GetAdoInvariant()
...
             logger.Info("Membership.ReadRow returned VableVersion={0} Data={1}", data.Version, data);
 
             Assert.Equal(1, data.Members.Count);
 
-            Assert.NotNull(data.Version.VersionEtag);
-            Assert.NotEqual(newTableVersion.VersionEtag, data.Version.VersionEtag);
-            Assert.Equal(newTableVersion.Version, data.Version.Version);
+            if (extendedProtocol)
+            {
+                Assert.NotNull(data.Version.VersionEtag);
```
gabikliot: I think this assert should hold even if extendedProtocol is false.

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
```diff
@@ -211,12 +232,15 @@ protected virtual string GetAdoInvariant()
...
+        protected async Task MembershipTable_UpdateRow(bool extendedProtocol = true)
         {
             var tableData = await membershipTable.ReadAll();
-
             Assert.NotNull(tableData.Version);
-            Assert.Equal(0, tableData.Version.Version);
+
+            if (extendedProtocol)
+            {
+                Assert.Equal(0, tableData.Version.Version);
```
gabikliot: that also holds even if extendedProtocol is false, right?

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
```diff
@@ -243,52 +268,49 @@ protected virtual string GetAdoInvariant()
                 Assert.NotNull(etagBefore);
 
                 logger.Info("Calling UpdateRow with Entry = {0} correct eTag = {1} old version={2}", siloEntry,
-                    etagBefore, tableVersion);
-
-                ok = await membershipTable.UpdateRow(siloEntry, etagBefore, tableVersion);
-
-                Assert.False(ok, $"row update should have failed - Table Data = {tableData}");
+                    etagBefore, tableVersion != null ? tableVersion.ToString() : "null");
```
gabikliot: why tableVersion would be ever null?

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
```diff
@@ -243,52 +268,49 @@ protected virtual string GetAdoInvariant()
...
 
+                if (extendedProtocol)
+                {
+                    ok = await membershipTable.UpdateRow(siloEntry, etagBefore, tableVersion);
```
gabikliot: this definitely should hold unrelated of extendedProtocol. It uses etagBefore.

gabikliot:
@galvesribeiro , thanks for the explanation! It makes much more sense now.
It would help a lot if you heavily documented DynamoDBStorage manager, and included all the explanation you gave here.
Agree that "blind write is not ALWAYS wrong", but usually they are. I just think that concurrency control is hard in general and distributed concurrency control even more so, thus being explicit in the method signatures forces people to think harder about what they are doing. It puts the choices of  concurrency control in their face: what should I use? TryInsert and TryUpdate are much more explicit than Write("bunch of concurrency options"). But I agree its a matter of personal taste, so I won't insist on changing that. You cab keep it as is now.

Overall, looks good now, except for a couple of comments I made. 
I think you were too eager in the tests to exclude too much code if !extendedProtocol. I think the vast majority of the checks will stay, even if the version is not persisted.

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
```diff
@@ -114,23 +114,28 @@ protected virtual string GetAdoInvariant()
...
+        protected async Task MembershipTable_InsertRow(bool extendedProtocol = true)
         {
             var membershipEntry = CreateMembershipEntryForTest();
 
             var data = await membershipTable.ReadAll();
             Assert.NotNull(data);
             Assert.Equal(0, data.Members.Count);
 
-            bool ok = await membershipTable.InsertRow(membershipEntry, data.Version.Next());
+            TableVersion nextTableVersion = data.Version.Next();
```
shayhatsor: I think this line was added by mistake. Please run the tests on Azure to make sure you haven't mistakenly broke anything related to the extended protocol.

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
shayhatsor: I think so too. @galvesribeiro, please make another pass at the changes you made to the tests.

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
shayhatsor: Currently it wouldn't matter, as Orleans doesn't use the number returned in the version, other than to increment. It can be good as an indicator of sanity.

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
shayhatsor: it must never be null.

shayhatsor:
@galvesribeiro, as most of this work was done while consulting me - it looks good to me :smile: 
But, the changes to the tests make me a bit :worried:. As I mentioned earlier, please make another pass and make sure you only move the least amount of checks to the extended block. Then run all tests with Azure.

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
galvesribeiro: I'll change that to be more explicit but it isn't required. A row is never saved with ETAG as null, the initial value is 0, so basically this would remove the need for the double check on the key. But I agree with you that it would make more clear for the reader the intentions with the expression even if it is not required by the DynamoDB expression in this case.

galvesribeiro: Rethinking on that, no... I think we shouldn't change the filter and leave only the etag. Since the ETag is never null (initial value is 0), and the row will never be inserted without DEPLOYMENT_ID and SILO_IDENTITY since they are the keys, there is no point in a row to has ETag but no keys :)

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
galvesribeiro: Done

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
galvesribeiro: Done

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
galvesribeiro: Done

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
galvesribeiro: Done

src/OrleansAWSUtils/Membership/DynamoDBMembershipTable.cs
gabikliot: Unless there is a bug?

test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
galvesribeiro: Ok, on this test I actually added more things on the IF because on extended protocol, the tableVersion is taken in consideration while performing the update. If I remove the IF, the update will happen since the etag is correct and I ignore the tableVersion by not implement the extended protocol and all the following assertion will fail.

galvesribeiro:
Ok guys, thanks for the feedback @shayhatsor and @gabikliot. I made small changes on the tests to make sure we get more Asserts as you guys pointed and also run the Azure tests to make sure nothing were affected and as expected (the default for `extendedProtocol` is `true`) everything is passing just fine.

galvesribeiro:
> It would help a lot if you heavily documented DynamoDBStorage manager, and included all the explanation you gave here.

Just for the sake of notes here, I'll be adding that explanation in a final PR on all the AWS lib which will make some cosmetic changes like error enum, messages, and comments. Rest assured of it.

shayhatsor:
@galvesribeiro, thanks for the hard work and dedication !

galvesribeiro:
Thanks to everyone for the feedback! :smile:

PR #2009, opened 2016-08-02
Optionally configure Dependency Injection container using delegate.

ReubenBond:
This PR allows Dependency Injection to be configured using a delegate instead of only the `StartupTypeName` method used today. The existing method is still available. They are mutually exclusive and an exception will be thrown is a user tries to set both.

Delegate-based configuration allows for much more flexibility over the existing system, which requires the use of static variables to provide runtime configuration. Eg, how do I provide services from an existing container (maybe shared with ASP.NET) using the current system? The only answer I can come up with is to use global static variables - which clearly wont suffice.

### DIFF
--- a/src/Orleans/project.json
+++ b/src/Orleans/project.json
@@ -1,6 +1,7 @@
 {
   "dependencies": {
-    "Newtonsoft.Json": "7.0.1"
+    "Newtonsoft.Json": "7.0.1",
+    "Microsoft.Extensions.DependencyInjection": "1.0.0"
   },
   "frameworks": {
     "net451": {}
--- a/src/OrleansRuntime/Silo/Silo.cs
+++ b/src/OrleansRuntime/Silo/Silo.cs
@@ -32,6 +32,10 @@ using Orleans.Timers;
 
 namespace Orleans.Runtime
 {
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.DependencyInjection.Extensions;
+
+    using Orleans.Runtime.ReminderService;
 
     /// <summary>
     /// Orleans silo.
@@ -175,6 +179,28 @@ namespace Orleans.Runtime
             globalConfig = config.Globals;
             config.OnConfigChange("Defaults", () => nodeConfig = config.GetOrCreateNodeConfigurationForSilo(name));
 
+            // Configure DI using Startup type
+            bool usingCustomServiceProvider;
+            Action<IServiceCollection> configureServices = serviceCollection =>
+            {
+                serviceCollection.AddSingleton(this);
+                serviceCollection.AddTransient<GrainBasedMembershipTable>();
+                serviceCollection.AddTransient<GrainBasedReminderTable>();
+
+                serviceCollection.AddSingleton(config);
+                serviceCollection.AddTransient(_ => this.globalConfig);
+                serviceCollection.AddTransient(_ => this.nodeConfig);
+                serviceCollection.TryAddSingleton(_ => membershipFactory.GetMembershipTable(GlobalConfig.LivenessType, GlobalConfig.MembershipTableAssembly));
+                serviceCollection.AddSingleton(svc =>
+                {
+                    var silo = svc.GetRequiredService<Silo>();
+                    var membershipTable = svc.GetRequiredService<IMembershipTable>();
+                    return membershipFactory.CreateMembershipOracle(silo, membershipTable);
+                });
+            };
+
+            Services = StartupBuilder.ConfigureStartup(nodeConfig, configureServices, out usingCustomServiceProvider);
+
             if (!LogManager.IsInitialized)
                 LogManager.Initialize(nodeConfig);
 
@@ -219,9 +245,6 @@ namespace Orleans.Runtime
                 LocalDataStoreInstance.LocalDataStore = keyStore;
             }
 
-            // Configure DI using Startup type
-            bool usingCustomServiceProvider;
-            Services = StartupBuilder.ConfigureStartup(nodeConfig.StartupTypeName, out usingCustomServiceProvider);
 
             healthCheckParticipants = new List<IHealthCheckParticipant>();
             allSiloProviders = new List<IProvider>();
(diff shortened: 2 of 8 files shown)


src/Orleans/project.json
```diff
@@ -1,6 +1,7 @@
 {
   "dependencies": {
-    "Newtonsoft.Json": "7.0.1"
+    "Newtonsoft.Json": "7.0.1",
+    "Microsoft.Extensions.DependencyInjection": "1.0.0"
```
ReubenBond: What are your thoughts on this? It seems that the .Abstractions package isn't sufficient (eg, doesn't have `GetRequiredService` or `IServiceCollection`)

test/Tester/DependencyInjectionGrainTests.cs
```diff
@@ -39,7 +41,49 @@ protected override TestCluster CreateTestCluster()
...
+        public async Task CannotGetExplictlyRegisteredGrain()
+        {
+            ISimpleDIGrain grain = GrainFactory.GetGrain<ISimpleDIGrain>(GetRandomGrainId(), grainClassNamePrefix: "UnitTests.Grains.ExplicitlyRegistered");
+            var exception = await Assert.ThrowsAsync<OrleansException>(() => grain.GetTicksFromService());
+            Assert.Contains("Error creating activation for", exception.Message);
+            Assert.Contains(nameof(ExplicitlyRegisteredSimpleDIGrain), exception.Message);
+        }
+    }
+
+    public class DelegateDependencyInjectionGrainTests : OrleansTestingBase, IClassFixture<DelegateDependencyInjectionGrainTests.Fixture>
```
ReubenBond: Same tests, different configuration method.

n-sidorov:
I can relate to this. When we were integrating Orleans into our production system we faced this issue. 

Though we ended up with static variables I was actually thinking about something like this PR but a little bit different.

Maybe it's better to add ability to provide instance of "startup" class instead of delegate (especially when this delegate is basically "startup" class)? It also more similar to current implementation where you provide name of "startup" class.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -175,6 +179,28 @@ internal Silo(string name, SiloType siloType, ClusterConfiguration config, ILoca
             globalConfig = config.Globals;
             config.OnConfigChange("Defaults", () => nodeConfig = config.GetOrCreateNodeConfigurationForSilo(name));
 
+            // Configure DI using Startup type
+            bool usingCustomServiceProvider;
+            Action<IServiceCollection> configureServices = serviceCollection =>
+            {
+                serviceCollection.AddSingleton(this);
```
jdom: careful that you are passing `this` (and resolving it afterwards) before this object finished constructing. This can lead to very hard to troubleshoot errors, since people might resolve it and try to use it in different ways

src/Orleans/project.json
jdom: Hmm, not sure if it's avoidable (considering that we want to enhance how to configure Orleans by using DI, similar to AspNet MVC), but since this is currently the only binary that EVERY project needs to reference (including grain interfaces), I'm at least initially reluctant to add this dependency without considering alternative approaches.

test/Tester/DependencyInjectionGrainTests.cs
```diff
@@ -39,7 +41,49 @@ protected override TestCluster CreateTestCluster()
...
+    {
+        private class Fixture : BaseTestClusterFixture
+        {
+            protected override TestCluster CreateTestCluster()
+            {
+                return new TestCluster(new TestClusterOptions(1))
+                {
+                    AppDomainSiloCreatorType = typeof(SiloCreator)
```
jdom: I understand this change, just a heads up that this will probably be another thing we'll have to work around if we want a TestCluster that works in CoreCLR (where AppDomains are not available, and we'll probably have to resort to starting processes and IPC, unless we solve the issue with statics, which is very unlikely).
Not necessarily an action item for this PR, but just a heads up in case you thought of other approaches that are less reliant on AppDomains (or maybe just removing AppDomain suffix from the property name, as alternate implementations might use that creator type differently).

src/OrleansRuntime/Startup/StartupBuilder.cs
```diff
@@ -30,30 +44,32 @@ internal static IServiceProvider ConfigureStartup(string startupTypeName, out bo
...
+            configureServices(serviceCollection);
 
-            serviceCollection.AddTransient<GrainBasedMembershipTable>();
-            serviceCollection.AddTransient<GrainBasedReminderTable>();
-        }
+            usingCustomServiceProvider = serviceProviderBuilder != null;
+            if (usingCustomServiceProvider)
+            {
+                return serviceProviderBuilder(serviceCollection);
+            }
```
jdom: This might be a nice first step for now.
Nevertheless. we should probably start splitting the registrations into helper methods, so that we can eventually do something like in AspNet MVC, where the user decides what from the framework to put into the container. Most of the times that requires the user to add an explicit 1 line config call such as `services.AddMvc()` (which adds every default implementation in MVC) but allows users to pick and choose different implementations of some of the internal services if they want. For more info of what I mean: https://docs.asp.net/en/latest/fundamentals/dependency-injection.html

With the current approach, if Orleans' defaults are always registered, there is not an easy way for opting for a different implementation unless we rely on the nasty approach of specifying the concrete type name that will be resolved, by having explicit configuration knobs (type names) for each internal dependency we might want replaceable.

test/Tester/DependencyInjectionGrainTests.cs
ReubenBond: If we get rid of stateful statics and can just use the container instead, then we don't need IPC or AppDomains

src/OrleansRuntime/Startup/StartupBuilder.cs
ReubenBond: The user could create their own container (from their library of choice), override the default Orleans ones, and return that. Given the method signature (IServiceCollection -> IServiceProvider) I don't see another way yet. I'm not sure if I should be configuring those default providers in a different way to make that easier.

src/OrleansRuntime/Silo/Silo.cs
ReubenBond: Point taken. The container should never be used before it's been constructed and it's never actually accessed (only referenced) in the constructor, the lines below this are just declaring that the `IMembershipOracle` implementation requires a `Silo`.

I plan to make just minimal changes to use the container in this PR and intend to move more into the container in future PRs.
It would be best if the container itself was built outside of the `Silo` and the silo is constructed by it.

ReubenBond:
Closing this PR since we intend to align with ASP.NET via #2052 & related issues/PRs

Issue #1984, opened 2016-07-25
Build.cmd fails to build pre-release nuget packages in diff OS locales


jdom:
The reason is that [CreatePackages.bat line 47](https://github.com/dotnet/orleans/blob/81776b5a5a0d4a6306810671e71a753325adad88/src/NuGet/CreateOrleansPackages.bat#L47) uses the MS-DOS `date` and `time` commands to generate a dev string based on the current date and time (such as Microsoft.Orleans.Client.1.3.0-Dev201607221711.nupkg).
Nevertheless those commands are language and locale dependent.
In Windows using default en-US settings, date returns `Mon 07/25/2016` and time `09:56 AM`.

When @veikkoeeva tried to run this using fi-FI settings, he got `Dev0165.7.1934` generated, which is invalid according to SemVer (and as you can see it shares nothing of the intended approach in en-US).

Here's some suggestions for making it locale independent: http://stackoverflow.com/questions/203090/how-to-get-current-datetime-on-windows-command-line-in-a-suitable-format-for-us

veikkoeeva:
@jdom Maybe an intermediary fix could be calling a PowerShell script to return the date?

A tip for a workaround suggested by @jdom at Gitter: hardcode the date to `201607221711` (or some other appropriate value) to get the packets build locally.

PR #2014, opened 2016-08-03
Make CreateOrleansPackages.bat work on non en-US locale

benjaminpetit:
Fix for #1984 using powershell to get the date

### DIFF
--- a/src/NuGet/CreateOrleansPackages.bat
+++ b/src/NuGet/CreateOrleansPackages.bat
@@ -44,7 +44,7 @@ if not "%VERSION_BETA%" == "" (
 
 if "%PRERELEASE_BUILD%" == "true" (
     set VERSION_TYPE=Dev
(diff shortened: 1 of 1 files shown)


veikkoeeva:
@benjaminpetit Maybe the command line could be adjusted to make it more robust, such as at http://stackoverflow.com/questions/19335004/how-to-run-a-powershell-script-from-a-batch-file. Have you tested this on other locales? If not, I can check later today.

benjaminpetit:
It worked fine on a fr-FR locale.

I think you are right about `-NoProfile`, but I am not sure if the `-ExecutionPolicy` is required here since we do not run a script?
```
C:\Users\bpetit>powershell get-executionpolicy
Restricted
C:\Users\bpetit>powershell get-date -format "{yyyyMMddHHmm}"
201608040936
```
Anyway I will add them both to be safe, thanks for the suggestion!

jdom:
Thank you!

PR #2016, opened 2016-08-04
Use new Orleans logo

dVakulen:


dVakulen:
Alternative:

ReubenBond:
Your revised version looks better and we shouldn't lose any SEO/discoverability from it. I wonder if those badges could be arranged better.

dVakulen:
Centered: 

chrome_2016-08-04_16-35-45

PR #2020, opened 2016-08-04
Avoid SiloStatus enum boxing

dVakulen:
Using of the <code>Equals</code> on enum causes boxing:

devenv_2016-08-04_21-37-00

Removed all instances of it's usages on <code>SiloStatus</code> enum.


jdom:
Thanks @dVakulen :)

PR #2022, opened 2016-08-04
Revert "Add dynamic adding / removing stream providers functionality."

jason-bragg:
Reverts dotnet/orleans#1966


jason-bragg:
NM, paranoia got the best of me :/

PR #2025, opened 2016-08-07
Assembly Catalog

galvesribeiro:
Initial implementation of `IAssemblyCatalog`.

This is part of the move to support .Net Core on #368 by removing the the dependency on AppDomains.

The default catalog implementation (`AssemblyCatalog`) requires the developer to set the assemblies list by XML `<Assemblies>` or code config. Later on, this implementation can be replaced by using DI.

Some changes were required on codegen in order to support it. So now, there is no scan for assemblies on a specific folder anymore. You have to inform the assembly list that needs to be processed for runtime codegen.


src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -3,104 +3,26 @@
...
-
-        /// <summary>
-        /// Handles <see cref="AppDomain.AssemblyLoad"/> events.
-        /// </summary>
-        /// <param name="sender">The sender of the event.</param>
-        /// <param name="args">The event arguments.</param>
-        private static void OnAssemblyLoad(object sender, AssemblyLoadEventArgs args)
-        {
-            AssemblyProcessor.ProcessAssembly(args.LoadedAssembly);
+            AssemblyProcessor.ProcessAssemblies(catalog.GetAssemblies());
```
ReubenBond: Is there a nice way that we can move this work out of the constructor and into another method?

src/Orleans/Configuration/NodeConfiguration.cs
```diff
@@ -489,6 +494,9 @@ internal void Load(XmlElement root)
                             StartupTypeName = child.GetAttribute("Type");
                         }
                         break;
+                    case "AssemblyCatalog":
+                        Assemblies = ConfigUtilities.ParseAssemblyList(child);
+                        break;
```
jdom: Feels odd to have this return the list of assemblies, and then separately have the runtime read the list to create the catalog, and having 2 mutually exclusive properties to assign

test/TesterInternal/AssemblyLoaderTests.cs
```diff
@@ -10,163 +10,39 @@
...
-                    e.InnerExceptions[1].Message != "Inner Exception #2")
-                {
-                    throw;
-                }
-            }
+            return new AssemblyCatalog()
+                .WithAssembly(@".\TestGrainInterfaces.dll")
+                .WithAssembly(@".\TestGrains.dll")
+                .WithAssembly(@".\TestInternalGrainInterfaces.dll")
+                .WithAssembly(@".\TestInternalGrains.dll");
```
jdom: I thought you were using assembly names, not paths to the DLLs. Nevertheless I see here and in many config files that you are passing paths

test/TesterInternal/ClientConfig_AzureStreamProviders.xml
```diff
@@ -16,4 +16,10 @@
     <Provider Type="Orleans.Providers.Streams.SimpleMessageStream.SimpleMessageStreamProvider" Name="SMSProvider" FireAndForgetDelivery="false"/>
     <Provider Type="Orleans.Providers.Streams.AzureQueue.AzureQueueStreamProvider" Name="AzureQueueProvider" DataConnectionString="UseDevelopmentStorage=true"/>
   </StreamProviders>
+  <Assemblies>
+    <Assembly Path=".\TestGrainInterfaces.dll"/>
+    <Assembly Path=".\TestFSharpInterfaces.dll"/>
+    <Assembly Path=".\Orleans.dll"/>
+    <Assembly Path=".\TestInternalGrainInterfaces.dll"/>
+  </Assemblies>
```
jdom: Do you actually need an entire list, or could this be just refering to the root of them, such as TesterInternal

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -136,8 +141,25 @@ public OutsideRuntimeClient(ClientConfiguration cfg, GrainFactory grainFactory,
...
             {
-                LoadAdditionalAssemblies();
-                
+                AssemblyCatalog = assemblyCatalog;
+
+                if (AssemblyCatalog == null)
+                {
+                    if (config.Assemblies == null ||
+                        config.Assemblies.Count == 0)
+                        throw new InvalidOperationException("No IAssemblyCatalog implementation or assembly list were provided.");
```
jdom: Given that we are now traversing dependencies, it would be nice to also give the option to not specify any assembly, and if that's the case, then it will assume Assembly.GetEntryAssembly() is in the list. So if your host is referencing your grain interfaces project, it would be picked up automatically. It would be nice to verify if there is any grain implementation in the referenced projects though.
This way for most people this migration is non-breaking, unless they happen to be using OrleansHost.exe, where in that case the host doesn't explicitly reference the implementations

jdom:
I wonder whether we should be emulating MVC Core with its [ApplicationParts](https://github.com/aspnet/Mvc/tree/master/src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts) it uses for stuff such as [finding controllers](https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerFeatureProvider.cs) or [TagHelpers](https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/TagHelperFeatureProvider.cs)

Without too much familiarity with their implementation, I believe both systems are trying to do the same thing, so I'd lean to trust that the MVC is very familiar with how the new tooling works (which is the reason we are proactively changing this in preparation to our port, but not really testing it yet in that environment). Maybe it doesn't make sense, but it's worth a small research at least.

src/Orleans/Configuration/NodeConfiguration.cs
galvesribeiro: Lets talk offline about that.

test/TesterInternal/ClientConfig_AzureStreamProviders.xml
galvesribeiro: For both cases it is just old code. You can use the assembly name and add only the root after our last iteration. I'll change it on those files.

src/Orleans/AssemblyLoader/AssemblyLoader.cs
galvesribeiro: I don't understand, where is the constructor there?

src/Orleans/Runtime/OutsideRuntimeClient.cs
galvesribeiro: > Given that we are now traversing dependencies, it would be nice to also give the option to not specify any assembly, and if that's the case, then it will assume Assembly.GetEntryAssembly() is in the list. So if your host is referencing your grain interfaces project, it would be picked up automatically.

Agreed! Will add that.

> It would be nice to verify if there is any grain implementation in the referenced projects though.
This way for most people this migration is non-breaking, unless they happen to be using OrleansHost.exe, where in that case the host doesn't explicitly reference the implementations

This check will fall push us back to the "Magic Scan" that we have today and not to mention that it is not the Catalog/Loader/Processor responsibility to inspect the assembly content IMHO. This filter to detect if there is a grain, stream, etc, is already done by the codegen IIRC...

galvesribeiro: Done. Except the magic IF to handle grain assemblies as diff.

```
if (config.Defaults.Assemblies == null ||
                    config.Defaults.Assemblies.Count == 0)
                    catalog.WithAssembly(Assembly.GetEntryAssembly().FullName);
```

test/TesterInternal/ClientConfig_AzureStreamProviders.xml
galvesribeiro: Done

test/TesterInternal/AssemblyLoaderTests.cs
galvesribeiro: Done

src/Orleans/AssemblyLoader/AssemblyLoader.cs
galvesribeiro: Done

galvesribeiro:
Ok, updated with most of the feedback. 

@jdom, I'm been looking the the AspNet Core parts and tagHelpers as you suggested... I still think that it will fall in the same case as MEF... A bunch of code for small win. I still thinking that now we removed all that bunch of code for scanning and we are close to what people asked which is provide a list of assemblies either by code or XML or, by replacing some sort of catalog implementation (we will add DI support to it later on).

adamhathcock:
This may be relevant: https://github.com/khellang/Scrutor

This is a nice way of scanning assemblies and registering them with `IServiceCollection`  I like using it anyway.

src/Orleans/AssemblyLoader/AssemblyCatalog.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace Orleans.Runtime
+{
+    public class AssemblyCatalog : IAssemblyCatalog
+    {
+        private HashSet<Assembly> catalog = new HashSet<Assembly>();
+        private Logger logger;
+
+        public AssemblyCatalog()
+        {
+            logger = LogManager.GetLogger("AssemblyCatalog", LoggerType.Provider);
```
veikkoeeva: `nameof`?

jdom:
@dotnet-bot test this please
(trying whether Jenkins builds this branch successfully)

sergeybykov:
"Project file does not exist." Hmm...

galvesribeiro:
I think it is .net CLI not installed...

src/Orleans/AssemblyLoader/AssemblyCatalog.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+namespace Orleans.Runtime
+{
+    public class AssemblyCatalog : IAssemblyCatalog
```
veikkoeeva: I wonder if it were useful to separate the handling logic from the traversal logic in creating something like


```
public static IEnumerable<T> AsDepthFirst<T>(this IEnumerable<T> elements, Func<T, IEnumerable<T>> descendantSelector)
{
    var stack = new Stack<T>(elements);
    while(stack.Any())
    {
            var next = stack.Pop();
            yield return next;
            foreach(var descendant in descendantSelector(next))
            {
                stack.Push(descendant);
            }
        }
    }
}
```

Then it would be possible to do something like `rootAssemblies.AsDepthFirst(assembly => assembly.GetReferencedAssemblies().Where(i => !blacklist.Contains(i.Name));`. This likely varies here, taken from the Gitter discussion. This wouldn't certainly cause stack overflow either.

galvesribeiro: @veikkoeeva can you put a PR to this branch? Appreciate that. Thanks

veikkoeeva: @galvesribeiro How long can this branch wait? :-P One the first though this modification would seem sensible, but would require one to suss out the details with APIs. I'm glad if you consider this and find a way, or someone else, but as of myself, well, you'd need to wait a week or two at minimum. Though if this makes sense, I, you, or someone else, could do a PR later.

galvesribeiro: Sure. The way it is today may not be the 10th wonder now but it does what is suppose to do. So I think we can enhance it later. What matter is the interface and its usage. The catalog implementation itself can change anytime.

veikkoeeva: There might be changes to the catalog interface too, but I'd be surprised if they couldn't be backawards compatible if needed.

jdom:
@dotnet-bot test this please
(trying to see whether Jenkins can build with dotnet cli. Most likely it will fail without further configuration)

adamhathcock:
Any chance this is on track to go in?  This would likely help clean up a lot of the .NET Core stuff and i'd still like to help :)

jdom:
Yes, I'm actually really close to pushing some branches related to core to the main repo (see my repo if interested). I found a way to not migrate to dotnet cli for now (and hence we can sill use our msbuild tasks, at least until the new tooling RTMs). I might start reviewing this on Monday, but feel free to review it yourself and accelerate this. Also, if you have experience with asm scanning in aspnet, then your input to whether it applies here too would be useful.

adamhathcock:
Cool.  I'll wait for a pull request or something as hunting around for the code doesn't feel like I'm finding the right thing.

I have two .NET Core projects in prod (one using containers!).  One uses global assembly scanning with `Microsoft.Extensions.DependencyModel` the other just picks assemblies and scans them with Scrutor like I linked above.

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -291,15 +81,15 @@ private static Assembly MatchWithLoadedAssembly(AssemblyName searchFor, IEnumera
...
             }
             return null;
         }
 
         private static Assembly MatchWithLoadedAssembly(AssemblyName searchFor, AppDomain appDomain)
         {
-            return 
+            return
                 MatchWithLoadedAssembly(searchFor, appDomain.GetAssemblies()) ??
                 MatchWithLoadedAssembly(searchFor, appDomain.ReflectionOnlyGetAssemblies());
```
xiazen: is this function still being used ? I think we are dropping using AppDomain to scan assemblies. 
Just a little confused.

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -291,15 +81,15 @@ private static Assembly MatchWithLoadedAssembly(AssemblyName searchFor, IEnumera
...
+                    return assembly;
                 }
             }
             return null;
         }
 
         private static Assembly MatchWithLoadedAssembly(AssemblyName searchFor, AppDomain appDomain)
         {
-            return 
+            return
```
xiazen: is this function still being used ? I think we are dropping using AppDomain to scan assemblies.
Just a little confused.

veikkoeeva: @xiazen I wonder does that mean we support both as per #2145. There was a bit of discussion if the `AppDomain` be shimmed ("polyfilled") at https://gitter.im/dotnet/orleans?at=57b6079287f779f0690a9fca &ndash; also some NETStandard code on assembly loading (that likely doesn't work correctly).

galvesribeiro: @xiazen You are correct. This thing shouldn't be used anymore. I'll get back to it later on. Thanks

xiazen: @veikkoeeva  I think this PR is about removing dependency on AppDomain to move to support .Net Core, which is part of #2145 . That's why I get confused. 

About shimmed `AppDomain`, it might accelate our process on going to .Net Core. But it might also confuse people with AppDomain in .Net. I'm slightly leaning in the direction to not create that shimmed `AppDomain `class. :)

jdom: hehe, funny that this was mentioned, but yes, until Assembly loading gets fixed, I'm introducing a AppDomain shim for now, otherwise there is no way I can get help fixing all the porting holes we currently have, and would have to spend a lot of time fixing every build break in the correct manner myself until I can get help with it. A PR is coming *really* soon.

xiazen: haha. well I'm fine with it being there temporarily, just for an intermediate step for us to go to coreclr.

sergeybykov:
We are using bits and pieces of this PR as part of the 2.0 work, but won't merge it as a whole. So I'm closing it to avoid confusion.

PR #2031, opened 2016-08-08
Remove Addresses new list allocations

dVakulen:
The check for silo validity is already being performed at https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs#L342

### DIFF
--- a/src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
+++ b/src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
@@ -332,19 +332,22 @@ namespace Orleans.Runtime.GrainDirectory
         internal AddressesAndTag LookUpGrain(GrainId grain)
         {
             var result = new AddressesAndTag();
+            IGrainInfo grainInfo;
             lock (lockable)
             {
-                if (partitionData.ContainsKey(grain))
+                if (!partitionData.TryGetValue(grain, out grainInfo))
                 {
-                    result.Addresses = new List<ActivationAddress>();
-                    result.VersionTag = partitionData[grain].VersionTag;
-
-                    foreach (var route in partitionData[grain].Instances.Where(route => IsValidSilo(route.Value.SiloAddress)))
-                    {
-                        result.Addresses.Add(ActivationAddress.GetAddress(route.Value.SiloAddress, grain, route.Key, route.Value.RegistrationStatus));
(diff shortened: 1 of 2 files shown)


jdom:
Change looks good to me.
Not related to your change, but I'm a little bit wary that https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs#L342 was doing the extra filtering for IsValidSilo inside the lock. I would have done that part outside of the lock (calling into an interface from within the lock sounds like a potential for unnecessary contention, since this PR kind of proved that IsValidSilo was being called outside the lock successfully)

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
```diff
@@ -332,19 +332,22 @@ internal void RemoveGrain(GrainId grain)
...
-                    {
-                        result.Addresses.Add(ActivationAddress.GetAddress(route.Value.SiloAddress, grain, route.Key, route.Value.RegistrationStatus));
-                    }
+                    return result;
                 }
             }
+
+            result.Addresses = new List<ActivationAddress>();
+            result.VersionTag = grainInfo.VersionTag;
+            foreach (var route in grainInfo.Instances.Where(route => IsValidSilo(route.Value.SiloAddress)))
```
jdom: hmmm, I'm not sure this is safe, since now the dictionary returned by grainInfo.Instances can be mutated concurrently (see that updates to it are done within a lock).
You'd have to either clone into a list on read (which was what you were initially trying to avoid in this PR, but now it would be for an actually valid reason), or you'd have to change every update of the dictionary into a clone, update and swap operation (which could be trickier to understand the side effects of).
As I mentioned, I'm fine if we don't change this in this PR, and revert to the filtering being done within the lock for now. Still your initial commit makes sense regardless of this additional contention optimization.

dVakulen: Then I'll just revert this commit. I think that accessing of this dictionary is single threaded in most cases, so that lock provides better performance than concurrent data structure. And cost of additional ToList will outweigh all gains from doing filtering out of lock.

jdom:
Thanks @dVakulen !

PR #2038, opened 2016-08-09
remove a redundant sql script added by mistake

shayhatsor:
already double checked with @veikkoeeva


veikkoeeva:
Yep, my bad. Out-of-date load and data generating script I accidentally included.

Issue #637, opened 2015-07-23
Exception in RegisterOrUpdateReminder (bug?) with SQL system store


kucheruk:
When using reminders with sql storage, call RegisterOrUpdateReminder for update reminder causes error:
Could not register reminder "..." to reminder table due to a race. Please try again later.
upd: exception thrown when calling RegisterOrUpdateReminder second time

LocalReminderService.cs:137
```C#
public async Task<IGrainReminder> RegisterOrUpdateReminder(GrainReference grainRef, string reminderName, TimeSpan dueTime, TimeSpan period)
{
    var entry = new ReminderEntry
    {
    ...
    };

    // entry.ETag is null 
    ...
    var newEtag = await reminderTable.UpsertRow(entry);

    //newEtag == null (!)
    if (newEtag != null) 
    {
    ....
    }

    var msg = string.Format("Could not register reminder {0} to reminder table due to a race. Please try again later.", entry);
    logger.Error(ErrorCode.RS_Register_TableError, msg);
    throw new ReminderException(msg);
}
```

```sql
UPDATE [OrleansRemindersTable]    
SET
    [StartTime]             = @startTime,
    [Period]                = @period,
    [GrainIdConsistentHash] = @grainIdConsistentHash
OUTPUT inserted.ETag
WHERE
    [ServiceId] = @serviceId AND @serviceId IS NOT NULL
    AND [GrainId] = @grainId AND @grainId IS NOT NULL
    AND [ReminderName] = @reminderName AND @reminderName IS NOT NULL
    AND ETag = @etag AND @etag IS NOT NULL; --   <---- ?
END
```
Looks like @etag here is always NULL.

veikkoeeva:
@kucheruk That check means the return value from [SqlReminderTable.UpsertRow is null](https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/ReminderService/SqlReminderTable.cs#L79), the call and return value ends up at [OrleansRelationalExtensions.UpsertReminderRowAsync](https://github.com/dotnet/orleans/blob/master/src/Orleans/RelationalStorage/OrleansRelationalExtensions.cs#L395). The reason it can be null is that no row was inserted

[...]

PR #2045, opened 2016-08-10
AWS DynamoDB Reminders

galvesribeiro:

### DIFF
--- a/src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
+++ b/src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
@@ -1,0 +1,261 @@
+﻿using Orleans;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using OrleansAWSUtils.Storage;
+using Amazon.DynamoDBv2.Model;
+using Amazon.DynamoDBv2;
+
+namespace OrleansAWSUtils.Reminders
+{
+    public class DynamoDBReminderTable : IReminderTable
+    {
+        private const string DEPLOYMENT_ID_PROPERTY_NAME = "DeploymentId";
+        private const string GRAIN_REFERENCE_PROPERTY_NAME = "GrainReference";
+        private const string REMINDER_NAME_PROPERTY_NAME = "ReminderName";
+        private const string SERVICE_ID_PROPERTY_NAME = "ServiceId";
+        private const string START_TIME_PROPERTY_NAME = "StartTime";
+        private const string PERIOD_PROPERTY_NAME = "Period";
+        private const string GRAIN_HASH_PROPERTY_NAME = "GrainHash";
+        private const string REMINDER_ID_PROPERTY_NAME = "ReminderId";
+        private const string ETAG_PROPERTY_NAME = "ETag";
+        private const string CURRENT_ETAG_ALIAS = ":currentETag";
+        private const string SERVICE_ID_INDEX = "ServiceIdIndex";
+        private SafeRandom random = new SafeRandom();
+
+        private const string TABLE_NAME_DEFAULT_VALUE = "OrleansReminders";
+        private Logger logger;
+        private DynamoDBStorage storage;
+        private string deploymentId;
+        private Guid serviceId;
+
+        public Task Init(GlobalConfiguration config, Logger logger)
+        {
+            deploymentId = config.DeploymentId;
+            serviceId = config.ServiceId;
+
+            this.logger = logger;
+
+            storage = new DynamoDBStorage(config.DataConnectionStringForReminders, logger);
+            logger.Info(ErrorCode.ReminderServiceBase, "Initializing AWS DynamoDB Reminders Table");
+
+            var secondaryIndex = new GlobalSecondaryIndex
+            {
+                IndexName = SERVICE_ID_INDEX,
+                Projection = new Projection { ProjectionType = ProjectionType.ALL },
+                KeySchema = new List<KeySchemaElement>
+                {
+                    new KeySchemaElement { AttributeName = SERVICE_ID_PROPERTY_NAME, KeyType = KeyType.HASH},
+                    new KeySchemaElement { AttributeName = GRAIN_HASH_PROPERTY_NAME, KeyType = KeyType.RANGE }
+                }
+            };
+
+            return storage.InitializeTable(TABLE_NAME_DEFAULT_VALUE,
+                new List<KeySchemaElement>
+                {
+                    new KeySchemaElement { AttributeName = REMINDER_ID_PROPERTY_NAME, KeyType = KeyType.HASH },
+                    new KeySchemaElement { AttributeName = GRAIN_HASH_PROPERTY_NAME, KeyType = KeyType.RANGE }
+                },
+                new List<AttributeDefinition>
+                {
+                    new AttributeDefinition { AttributeName = REMINDER_ID_PROPERTY_NAME, AttributeType = ScalarAttributeType.S },
+                    new AttributeDefinition { AttributeName = GRAIN_HASH_PROPERTY_NAME, AttributeType = ScalarAttributeType.N },
+                    new AttributeDefinition { AttributeName = SERVICE_ID_PROPERTY_NAME, AttributeType = ScalarAttributeType.S }
+                },
+                new List<GlobalSecondaryIndex> { secondaryIndex });
+        }
+
+        public async Task<ReminderEntry> ReadRow(GrainReference grainRef, string reminderName)
+        {
+            var reminderId = ConstructReminderId(serviceId, grainRef, reminderName);
+
+            var keys = new Dictionary<string, AttributeValue>
+                {
+                    { $"{REMINDER_ID_PROPERTY_NAME}", new AttributeValue(reminderId) },
+                    { $"{GRAIN_HASH_PROPERTY_NAME}", new AttributeValue { N = grainRef.GetUniformHashCode().ToString() } }
+                };
+
+            try
+            {
+                return await storage.ReadSingleEntryAsync(TABLE_NAME_DEFAULT_VALUE, keys, Resolve).ConfigureAwait(false);
+            }
+            catch (Exception exc)
+            {
+                logger.Warn(ErrorCode.ReminderServiceBase,
+                    $"Intermediate error reading reminder entry {Utils.DictionaryToString(keys)} from table {TABLE_NAME_DEFAULT_VALUE}.", exc);
+                throw;
+            }
+        }
+
+        public async Task<ReminderTableData> ReadRows(GrainReference grainRef)
+        {
+            var expressionValues = new Dictionary<string, AttributeValue>
+                {
+                    { $":{SERVICE_ID_PROPERTY_NAME}", new AttributeValue(serviceId.ToString()) },
+                    { $":{GRAIN_REFERENCE_PROPERTY_NAME}", new AttributeValue(grainRef.ToKeyString()) }
+                };
+
+            try
+            {
+                var expression = $"{SERVICE_ID_PROPERTY_NAME} = :{SERVICE_ID_PROPERTY_NAME} AND {GRAIN_REFERENCE_PROPERTY_NAME} = :{GRAIN_REFERENCE_PROPERTY_NAME}";
+                var records = await storage.ScanAsync(TABLE_NAME_DEFAULT_VALUE, expressionValues, expression, Resolve).ConfigureAwait(false);
+
+                return new ReminderTableData(records);
+            }
+            catch (Exception exc)
+            {
+                logger.Warn(ErrorCode.ReminderServiceBase,
+                    $"Intermediate error reading reminder entry {Utils.DictionaryToString(expressionValues)} from table {TABLE_NAME_DEFAULT_VALUE}.", exc);
+                throw;
+            }
+        }
+
+        public async Task<ReminderTableData> ReadRows(uint beginHash, uint endHash)
+        {
+            var expressionValues = new Dictionary<string, AttributeValue>
+                {
+                    { $":{SERVICE_ID_PROPERTY_NAME}", new AttributeValue(serviceId.ToString()) },
+                    { $":Begin{GRAIN_HASH_PROPERTY_NAME}", new AttributeValue { N = beginHash.ToString() } },
+                    { $":End{GRAIN_HASH_PROPERTY_NAME}", new AttributeValue { N = endHash.ToString() } }
+                };
+
+            try
+            {
+                string expression = string.Empty;
+                if (beginHash < endHash)
+                {
+                    expression = $"{SERVICE_ID_PROPERTY_NAME} = :{SERVICE_ID_PROPERTY_NAME} AND {GRAIN_HASH_PROPERTY_NAME} > :Begin{GRAIN_HASH_PROPERTY_NAME} AND {GRAIN_HASH_PROPERTY_NAME} <= :End{GRAIN_HASH_PROPERTY_NAME}";
+                }
+                else
+                {
+                    expression = $"{SERVICE_ID_PROPERTY_NAME} = :{SERVICE_ID_PROPERTY_NAME} AND ({GRAIN_HASH_PROPERTY_NAME} > :Begin{GRAIN_HASH_PROPERTY_NAME} OR {GRAIN_HASH_PROPERTY_NAME} <= :End{GRAIN_HASH_PROPERTY_NAME})";
+                }
+
+                var records = await storage.ScanAsync(TABLE_NAME_DEFAULT_VALUE, expressionValues, expression, Resolve).ConfigureAwait(false);
+
+                return new ReminderTableData(records);
+            }
+            catch (Exception exc)
+            {
+                logger.Warn(ErrorCode.ReminderServiceBase,
+                    $"Intermediate error reading reminder entry {Utils.DictionaryToString(expressionValues)} from table {TABLE_NAME_DEFAULT_VALUE}.", exc);
+                throw;
+            }
+        }
+
+        private static ReminderEntry Resolve(Dictionary<string, AttributeValue> item)
+        {
+            return new ReminderEntry
+            {
+                ETag = item[ETAG_PROPERTY_NAME].N,
+                GrainRef = GrainReference.FromKeyString(item[GRAIN_REFERENCE_PROPERTY_NAME].S),
+                Period = TimeSpan.Parse(item[PERIOD_PROPERTY_NAME].S),
+                ReminderName = item[REMINDER_NAME_PROPERTY_NAME].S,
+                StartAt = DateTime.Parse(item[START_TIME_PROPERTY_NAME].S)
+            };
+        }
+
+        public async Task<bool> RemoveRow(GrainReference grainRef, string reminderName, string eTag)
+        {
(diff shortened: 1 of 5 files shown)


galvesribeiro:
@gabikliot @shayhatsor rebased here. Could you please review it? Thanks

shayhatsor:
LGTM

galvesribeiro:
If anyone else doesn't see any problems I can rebase to resolve the conflicts

gabikliot:
Wait with merging, I haven't reviewed yet.

sergeybykov:
A general comment about `OrleansAWSUtils`.

It is great that we are adding direct and easy support for non-Azure clouds. Not only does this expand the "addressable market" and makes the codebase more portable, this also highlights and supports the important tenet of Orleans - that it runs anywhere you need, the way you need.

When we add similar direct support for GCP, we'll cover the "big three" players.

One thing that I think we need to be conscious of and communicate clearly is the current state of the project. Until somebody runs in production with a piece of tech, it's difficult to be sure that all major issues with it have been addressed.

In that light, I'm thinking that we should explicitly set the version of `OrleansAWSUtils` NuGet package to `1.0.0-beta` or something like that, in order to communicate the state of it and set the expectations accordingly without slowing down iterations on the code. Once somebody confirms that they are using it in production, we'll promote it to the general version. I think we should have done the same with the initial version of `OrleansSQLUtils`, `OrleansConsulUtils`, etc., but we keep learning as we go.

galvesribeiro:
Fully agree with @sergeybykov 

I also would like to have the providers release versions should not be locked on Orleans core anymore so starting with AWSUtils (we should definitively rename that for the "new" naming conventions) keep it as 1.0.0-beta and counting.

There are 3 people ready to use it. Lets see their feedback.

gabikliot:
I fully support the beta version idea.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -0,0 +1,294 @@
...
+            var etagField = new Dictionary<string, AttributeValue>
+                {
+                    { ETAG_INCREMENT, new AttributeValue { N = "1" } }
+                };
+
+            try
+            {
+                if (logger.IsVerbose) logger.Verbose("UpdateRow entry = {0}, etag = {1}", entry.ToString(), entry.ETag);
+
+                var foundETag = await storage.ReadSingleEntryAsync(TABLE_NAME_DEFAULT_VALUE, keys, item =>
```
gabikliot: you don't need the read here, right?
You already have the UpsertEntryAsync in the DynamoDBStorage.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -0,0 +1,294 @@
...
+                {
+                    { SERVICE_ID_PROPERTY_NAME, new AttributeValue(serviceId.ToString()) },
+                    { GRAIN_REFERENCE_PROPERTY_NAME, new AttributeValue( entry.GrainRef.ToKeyString()) },
+                    { PERIOD_PROPERTY_NAME, new AttributeValue(entry.Period.ToString()) },
+                    { START_TIME_PROPERTY_NAME, new AttributeValue(entry.StartAt.ToString()) },
+                    { REMINDER_NAME_PROPERTY_NAME, new AttributeValue(entry.ReminderName) }
+                };
+
+            var extraExpression = $"{ETAG_PROPERTY_NAME} = {ETAG_PROPERTY_NAME} + {ETAG_INCREMENT}";
+            var etagField = new Dictionary<string, AttributeValue>
```
gabikliot: You are supposed to use entry.ETag that was passed, as is, not come up with your own etag.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -0,0 +1,294 @@
...
+
+            var keys = new Dictionary<string, AttributeValue>
+                {
+                    { $"{REMINDER_ID_PROPERTY_NAME}", new AttributeValue(reminderId) },
+                    { $"{GRAIN_HASH_PROPERTY_NAME}", new AttributeValue { N = grainRef.GetUniformHashCode().ToString() } }
+                };
+
+            try
+            {
+                var conditionalValues = new Dictionary<string, AttributeValue> { { CURRENT_ETAG_ALIAS, new AttributeValue { N = eTag } } };
```
gabikliot: here you are using the etag correctly.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -0,0 +1,294 @@
...
+            try
+            {
+                string expression = string.Empty;
+                if (beginHash < endHash)
+                {
+                    expression = $"{SERVICE_ID_PROPERTY_NAME} = :{SERVICE_ID_PROPERTY_NAME} AND {GRAIN_HASH_PROPERTY_NAME} > :Begin{GRAIN_HASH_PROPERTY_NAME} AND {GRAIN_HASH_PROPERTY_NAME} <= :End{GRAIN_HASH_PROPERTY_NAME}";
+                }
+                else
+                {
+                    expression = $"{SERVICE_ID_PROPERTY_NAME} = :{SERVICE_ID_PROPERTY_NAME} AND ({GRAIN_HASH_PROPERTY_NAME} > :Begin{GRAIN_HASH_PROPERTY_NAME} OR {GRAIN_HASH_PROPERTY_NAME} <= :End{GRAIN_HASH_PROPERTY_NAME})";
```
gabikliot: I am not familiar enough with DynamoDB. 
Does it matter how you structure the query? I would expect it does. In Azure Table only PK and RK are sorted and indexed, so it was very important to structure the query to be done in one sorted scan.
But in DynamoDB, they have secondary indices, right?
Don't you need to tell it explicitly what columns to index? I would not index by default all columns, right?
If you do NOT index those columns, I think we need to think about efficiency of that query , same as we did for Azure Table.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -0,0 +1,294 @@
...
+            serviceId = config.ServiceId;
+
+            this.logger = logger;
+
+            storage = new DynamoDBStorage(config.DataConnectionStringForReminders, logger);
+            logger.Info(ErrorCode.MembershipBase, "Initializing AWS DynamoDB Membership Table");
+            return storage.InitializeTable(TABLE_NAME_DEFAULT_VALUE,
+                new List<KeySchemaElement>
+                {
+                    new KeySchemaElement { AttributeName = REMINDER_ID_PROPERTY_NAME, KeyType = KeyType.HASH },
```
gabikliot: what does KeySchemaElement  here mean?

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -0,0 +1,294 @@
...
+
+        public Task Init(GlobalConfiguration config, Logger logger)
+        {
+            deploymentId = config.DeploymentId;
+            serviceId = config.ServiceId;
+
+            this.logger = logger;
+
+            storage = new DynamoDBStorage(config.DataConnectionStringForReminders, logger);
+            logger.Info(ErrorCode.MembershipBase, "Initializing AWS DynamoDB Membership Table");
```
gabikliot: Copy paste: reminders Table.

gabikliot: need different error code.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -0,0 +1,294 @@
...
+                {
+                    { $":{SERVICE_ID_PROPERTY_NAME}", new AttributeValue(serviceId.ToString()) },
+                    { $":{GRAIN_REFERENCE_PROPERTY_NAME}", new AttributeValue(grainRef.ToKeyString()) }
+                };
+
+            try
+            {
+                var expression = $"{SERVICE_ID_PROPERTY_NAME} = :{SERVICE_ID_PROPERTY_NAME} AND {GRAIN_REFERENCE_PROPERTY_NAME} = :{GRAIN_REFERENCE_PROPERTY_NAME}";
+                var records = await storage.ScanAsync(TABLE_NAME_DEFAULT_VALUE, expressionValues, expression,
+                    item => new ReminderEntry
```
gabikliot: wouldn't you want to put that translation code into a func? Its repeated 3 times here.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
shayhatsor: AFAIK, the reminder upsert in orleans should be unconditional and return the new unique etag, this is what's being done here. Only on the delete operation we check the etag. It never made sense to me, I've opened an issue about it a long time ago.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
shayhatsor: In Azure, MSSQL and MYSQL we don't use the incoming etag at all, and the test assumes that. [The etag value isn't populated by Orleans on upsert](https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/ReminderService/LocalReminderService.cs#L120).

galvesribeiro: @shayhatsor this is the mehod that fail on that parallel test if we don't read and do that trick, remember?

@gabikliot are u suggesting dont read the latest etag and trust on the one that come in the object? Or just blindly write?

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: It it the way to describe the PK and RK of a dynamodb table. We need to do that while creating a table

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: Sure. Will do.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: Yes, it index by default only the keys. However, we can't do queries with parts of the field like in Azure (i.e. Substring) but, you are right. I shouls have marked those 2 fields as secondary indexes. Will do that.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
shayhatsor: @galvesribeiro, I remember. but there's a bigger issue here, like i mentioned : [The etag value isn't populated by Orleans on upsert](https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/ReminderService/LocalReminderService.cs#L120). so, in AWS, we read before attempting insert/increment. If you remember, i mentioned you can write with a guid, which will remove the need for the read. You didn't want to have different etag implementations for reminders and other stuff, I can't blame you. I think it's OK either way.

gabikliot: OK, I recall now what Shay is saying about blind upsert. I will review carefully again in the evening.

galvesribeiro: @shayhatsor yeah I remember. Ok. 

@gabikliot take your time.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: I think this is the same case as before...

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: Ok... Adding, the index doesn't do the trick: `KeyConditionExpressions must only contain one condition per key`. So is pointless to make the secondary indexes since they can't be used in the Query operation but instead only on Scan. I don't know if there is a benefit on have the them just for scan since it is only used on Query...

galvesribeiro: Scratch that... The secondary indexes **ARE** considered on the Scan requests and that impacts directly on the performance. So I've added the index on the ServiceId.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: A new enum for all the DynamoDB-based services will be added with corresponding errors on the final PR.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
gabikliot: About secondary indices: index on the ServiceId is good, but maybe not enough. In Azure we had a hierarchical index (achieved via PK sort order): service id first and then grainHash:
https://github.com/dotnet/orleans/blob/master/src/OrleansAzureUtils/Storage/RemindersTableManager.cs#L47
That way you can scale within the service as well: each silo reads its range only, and does not have to scan all reminders for this service id. Can you have global secondary index on 2 keys?

gabikliot:
Re reminders Upsert:  https://github.com/dotnet/orleans/issues/637#issuecomment-124152405
It's good I explained it there. It would take me a significant effort to recall it from the depth of my mind now.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
gabikliot: Ok, I understand the confusion here. The wrong part is that I named thereminderTable.Upsert as upsert. Upsert means: update or insert. What we really need in Reminders table is Unconditionally write. If its new - insert, if not - just overwrite. 
In Azure we indeed implemented "Upsert" as InsertOrReplace, which is correct for reminders but wrong for a general correct Upsert. So this: https://github.com/dotnet/orleans/blob/master/src/OrleansAzureUtils/Storage/AzureTableDataManager.cs#L191
is a wrong name for this method. It should be renamed into WriteUncondionaly, or InsertOrReplcae.

In Dynamo DB case you actually implemented Upsert corrrecly: its conditional on prev etag. So here in Reminders table impl. you SHOULDS not be calling it. You should just pick  a random int for your etag version and write through, overwriting any potential value.
In the future we need to fix the naming and also remove the picking the random value into the reminders service not rely on the table implementation tom pick it.

@shayhatsor , do you agree?

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
gabikliot: Shay, I will open an issue and explain why it is so. The confusion is in naming. Once the naming is solved, I hope it will make sense. 

A preview of the explanation is: in reminders case we prefer availability over consistency. That is: we prefer to risk to have 2 reminders or last overwrite previous one, but minimize the chance of  not creating the reminder at all. Plus we want an API that allows to for sure create a reminder without checking of previous exists. Given that semantical choice (which we can argue if it is correct or not, but given this choice), we have no other choice but write the reminder through. On the other hand, we want to make it much harder to delete a reminder. Thus delete is conditional.
This is a very specific semantics which we chose for reminders, since their goal is to ensure SOME tick eventually happens, which justifies this semantics. I will write more details about it later on.

shayhatsor: Gabi, I believe I understand the motivation and the semantics themselves. I'm just saying that I see no difference between a reminder and any other persistent state.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: @gabikliot ok,understood. Indeed the Azure naming is confusing. In order to fix it, I should use our `PutEntryAsync` on DynamoDB which Create or Replace an item and just do it unconditionally. Will make the update.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: Yes I can add 2 keys. So I'll add the ServiceId as the PK and the GrainHash as the RK.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: Ok... If I do that @gabikliot, this test will fail:

```
        protected async Task RemindersParallelUpsert()
        {
            var upserts = await Task.WhenAll(Enumerable.Range(0, 50).Select(i =>
            {
                var reminder = CreateReminder(MakeTestGrainReference(), i.ToString());
                return Task.WhenAll(Enumerable.Range(1, 5).Select(j => remindersTable.UpsertRow(reminder)));
            }));
            Assert.False(upserts.Any(i => i.Distinct().Count() != 5));
        }
```

galvesribeiro: Done

gabikliot: The parallel upsert test should succeed: reminders have different names there. Name is part of row key.

galvesribeiro: Sure. Added that and it passed.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
galvesribeiro: Done

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
veikkoeeva: Also in https://github.com/dotnet/orleans/issues/637#issuecomment-124152405.

gabikliot:
@galvesribeiro , looks good, please squash and rebase.

galvesribeiro:
@gabikliot @shayhatsor rebased. Thanks for take time to review.

shayhatsor:
@galvesribeiro, I was going to merge this PR but GitHub says I'm not authorized. Does anyone know what's up? @sergeybykov, @jdom ?

jdom:
I do not know what's up. Merge button is enabled for me. Let me know if you still can't and whether you want me to push the button. Otherwise we can wait for Sergey to check what's wrong

shayhatsor:
It's weird, It's still disabled for me. I'll wait to see if it's a temporary GitHub hiccup. No rush at this point, thanks @jdom for the quick reply.

gabikliot:
Same for me, it says I don't have write access to this repository any more.

galvesribeiro:
I never had write access here but now I'm seeing same screen you all here with the gray button.

jdom:
Try now, I think it was my bad, I intended to protect the master branch from force push, but I think I protected it from normal push too. Should be back to normal now

gabikliot:
Yes, it's good now. @shayhatsor you can merge it.

galvesribeiro:
Aha! Now I dont see that button again and only the green checks. Thanks

Issue #2030, opened 2016-08-08
Make Configuration objects bindable to Microsoft.Extensions.Configuration


adamhathcock:
https://github.com/aspnet/Configuration

Basic change:
- Use M.E.Configuration and the binding scenario to load configuration settings from all the file types supported by the ASP.NET Core configuration
- The same configuration objects used now should basically work.
- Allows for more than just XML configuration (I prefer YAML)

I haven't dug too deep but this seems fairly straight forward.

adamhathcock:
If there's basic agreement, I can start on this.

veikkoeeva:
Ping @jdom, @sergeybykov.

jason-bragg:
I've not sufficiently investigated Microsoft.Extensions configuration and logging to offer an informed opinion, but in general I very much prefer integrating with existing abstractions rather than maintaining our own.

As this looks related to #2029, my preference would be to address configuration and dependency injection before tackling logging.

jdom:
I would really like if we move in this direction. Nevertheless, as @jason-bragg mentioned, I believe it's not just a matter of replacing 1 set of configuration objects graph to a different one, but it would probably require a revamp of our config and DI story altogether.
I would really like if we can emulate what AspNet Core is doing today in their config APIs. This way we do not spend countless hours with opinions on what should DI and the config story look like (this is and was always a time sink, as there's a ton of conflicting opinions out there). 
AspNet MVC evolved and improved this aspect on every iteration since its inception, and found consensus after tons of debates. So we should just stand on its shoulders, and benefit on the fact that people are familiar with it.

ReubenBond:
:+1: The more bits we can lift from ASP.NET, the better. Less to learn, more alignment, more developers

adamhathcock:
I'm glad the consensus is just to reuse what ASP.NET Core is doing since I do the same with my projects at work to avoid debate :)

I think this one I can actually start and see where the rabbit hole goes.

adamhathcock:
I started with something ugly and basic.  I guess I want a pre-review before I keep going down this route:
https://github.com/dotnet/orleans/compare/master...adamhathcock:new_config_2?expand=1

Goal:  Make configuration bindable so any file type can be used.  Need to separate logic and data.
Changes:
- Started creating replacement config heirarchy for Grain Clients.
- Remove Xml/File loading
- Added binding tests for xml and json
- Xml format has to change.

Bindable files:
[...]

PR #2047, opened 2016-08-11
[WIP] Using Microsoft.Extensions.Configuration and bindable configs for file formats.

adamhathcock:
Addressing issue: https://github.com/dotnet/orleans/issues/2030

I started with something ugly and basic. I guess I want a pre-review before I keep going down this route.

Goal: Make configuration bindable so any file type can be used. Need to separate logic and data.
Changes:

Started creating replacement config heirarchy for Grain Clients.
- Remove Xml/File loading
- Added binding tests for xml and json
- Xml format has to change.
Bindable files:
https://github.com/adamhathcock/orleans/tree/new_config_2/src/Orleans/Configuration/New

Old xml: https://github.com/adamhathcock/orleans/blob/new_config_2/test/TesterInternal/ClientConfig_StreamProviders.xml
New xml: https://github.com/adamhathcock/orleans/blob/new_config_2/test/TesterInternal/ClientConfig_StreamProviders_New.xml
New json: 
https://github.com/adamhathcock/orleans/blob/new_config_2/test/TesterInternal/ClientConfig_StreamProviders_New.json

Oh yeah, the actual tests I added:https://github.com/adamhathcock/orleans/blob/new_config_2/test/TesterInternal/ConfigTests.cs#L350

The stuff is far from complete but large TODOs that I see:

- Make more bindable configs
- Build out a ClientFactoryBuilder to take file configs and runtime config then once Build is called then the config is immutable. The act of building will do the logic that is current baked all over the config classes.
- ClientFactory will create GrainClient objects
- Add shortcut/helper methods to make the above process easy.

This involves touching a lot of stuff. If M.E.Logging was used then large sections of config would go away (e.g. Trace stuff) I think. But I'd port it over anyway for now.

### DIFF
--- a/src/Orleans/Configuration/New/ClientFactoryBuilder.cs
+++ b/src/Orleans/Configuration/New/ClientFactoryBuilder.cs
@@ -1,0 +1,309 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+using System.Xml;
+using Orleans.Providers;
+using System.Reflection;
+using System.Threading;
+using Orleans.Configuration.New;
+using Orleans.Runtime.Configuration;
+
+namespace Orleans.Runtime.Configuration.New
+{
+    /// <summary>
+    /// Orleans client configuration parameters.
+    /// </summary>
+    public class ClientFactoryBuilder
+    {
+        /// <summary>
+        /// The name of this client.
+        /// </summary>
+        public static string ClientName = "Client";
+        
+        private readonly DateTime creationTimestamp;
+
+        public string SourceFile { get; private set; }
+
+        /// <summary>
+        /// The list fo the gateways to use.
+        /// Each GatewayNode element specifies an outside grain client gateway node.
+        /// If outside (non-Orleans) clients are to connect to the Orleans system, then at least one gateway node must be specified.
+        /// Additional gateway nodes may be specified if desired, and will add some failure resilience and scalability.
+        /// If multiple gateways are specified, then each client will select one from the list at random.
+        /// </summary>
+        public List<IPEndPoint> Gateways { get; } = new List<IPEndPoint>();
+        /// <summary>
+        /// </summary>
+        public int PreferedGatewayIndex { get; set; }
+        
+        /// <summary>
+        /// </summary>
+        public string DNSHostName { get; private set; } // This is a true host name, no IP address. It is NOT settable, equals Dns.GetHostName().
+        /// <summary>
+        /// </summary>
+        public TimeSpan GatewayListRefreshPeriod { get; set; }
+        
+        public LimitManager LimitManager { get; private set; }
+        /// <summary>
+        /// </summary>
+        public ClientConfiguration.GatewayProviderType GatewayProvider { get; set; }
+
+        private static readonly TimeSpan DEFAULT_GATEWAY_LIST_REFRESH_PERIOD = TimeSpan.FromMinutes(1);
+
+        /// <summary>
+        /// </summary>
+        public bool UseAzureSystemStore 
+        { 
+            get { 
+                return GatewayProvider == Configuration.ClientConfiguration.GatewayProviderType.AzureTable 
+                       && !String.IsNullOrWhiteSpace(Settings.SystemStore.DeploymentId) 
+                       && !String.IsNullOrWhiteSpace(Settings.SystemStore.DataConnectionString); 
+            } 
+        }
+
+        /// <summary>
+        /// </summary>
+        public bool UseSqlSystemStore
+        {
+            get
+            {
+                return GatewayProvider == Configuration.ClientConfiguration.GatewayProviderType.SqlServer
+                && !String.IsNullOrWhiteSpace(Settings.SystemStore.DeploymentId)
+                && !String.IsNullOrWhiteSpace(Settings.SystemStore.DataConnectionString);
+            }
+        }
+
+        private bool HasStaticGateways { get { return Gateways != null && Gateways.Count > 0; } }
+        /// <summary>
+        /// </summary>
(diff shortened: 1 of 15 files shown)


src/Orleans/Configuration/New/ClientFactoryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+            if (!string.IsNullOrEmpty(Settings.SystemStore.DataConnectionString))
+            { 
+                if (GatewayProvider == ClientConfiguration.GatewayProviderType.None)
+                {
+                    // Assume the connection string is for Azure storage if not explicitly specified
+                    GatewayProvider = ClientConfiguration.GatewayProviderType.AzureTable;
+                }
+            }
+                    
+                            //ConfigUtilities.ParseTelemetry(child);
```
xiazen: remove this if it's not used.

src/Orleans/Configuration/New/Gateway.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Net;
+using System.Net.Sockets;
+using System.Threading.Tasks;
+using Orleans.Runtime.Configuration;
+
+namespace Orleans.Configuration.New
```
xiazen: Are we sure we are going to call the namespace this? Maybe Orleans.Configuration.vTwo... in case we want to have a v3..

src/Orleans/Configuration/New/ClientFactoryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+
+            GatewayListRefreshPeriod = DEFAULT_GATEWAY_LIST_REFRESH_PERIOD;
+            LimitManager = new LimitManager();
+            ProviderConfigurations = new Dictionary<string, ProviderCategoryConfiguration>();
+        }
+        
+        
+        public ClientFactorySettings Settings { get; private set; }
+
+        public void Build(ClientFactorySettings client)
```
xiazen: call it clientSettings to be more readable.

src/Orleans/Configuration/New/ClientFactoryBuilder.cs
```diff
@@ -0,0 +1,309 @@
...
+            }
+                        
+            if (Settings.SystemStore.SystemStoreType.HasValue)
+            {
+                GatewayProvider = Settings.SystemStore.SystemStoreType.Value;
+            }
+            if (!string.IsNullOrEmpty(Settings.SystemStore.CustomGatewayProviderAssemblyName))
+            {
+                if (Settings.SystemStore.CustomGatewayProviderAssemblyName.EndsWith(".dll"))
+                    throw new FormatException("Use fully qualified assembly name for \"CustomGatewayProviderAssemblyName\"");
```
xiazen: This exception message is confusing. do you want it to use fully qualified assembly name or it shouldn't?

src/Orleans/Configuration/New/Tracing.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <summary>
+        /// The TraceFilePattern attribute specifies the pattern name of a file that trace output should be written to.
+        /// </summary>
+        public string TraceFilePattern
+        { 
+            get { return traceFilePattern; }
+            set
+            {
+                traceFilePattern = value;
+                //ConfigUtilities.SetTraceFileName(this, ClientName, this.creationTimestamp);
```
xiazen: Why is this commented out? did you move this to somewhere else?

xiazen:
Sorry for the nit picking. I understand this is your first pass, and you just want to get some general ideas of the direction to go. I will avoid those nit pickings in the future when you just want a general idea back. 

Couple comments here. 
1. break up super long method to smaller methods and composite them back together, like ClientFactoryBuilder.build(). 
2. Is ClientFactoryBuilder.build supposed to build the client? If yes, then it should have a return type of GrainClient. And also you need to add test for ClientFactoryBuilder to build the client. 
3. If I understand this correctly, ClientFactorySettings serves as a intermediate object between config file and ClientConfiguration. So it's like config file ->config = Microsoft.Extensions.ConfigurationBuilder.Load(config file) -> ClientFactorySettings  = config.bind(ClientFactorySettings). This transferred config file to ClientFactorySettings. But do you have a parser to parse values in ClientFactorySettings to Orleans.Runtime.ClientConfiguratrion? So GrainFactory can use it to init a GrainClient? Would it be a method in ClientFactorySettings  ? 
4. I think if 3 is correct, you should move  ClientConfiguration LocalhostSilo(..) method to ClientFactorySetting class too. 
5. if we go in this direction, do we still want to keep the old way of configuration just to be backwards compatible? Or we keep both way around. 

Just some confusion  I have here and want to confirm with you to make sure we are in the same page.

adamhathcock:
Hi @xiazen 

1) `Build` was only going to be a stop-gap between trying to move code from the old hard-coded parsing to bindable objects (which should be just POCOs).  I agree long methods are horrible.
2) The idea was to put together the GrainClientFactory then it could produce GrainClients.  Maybe it's wrong.
3) You're correct about the idea.  I shouldn't have to write any parser code as eventually all config should be bindable to `ConfigurationBuilder`  Then I could have config in any supported format and be composable across multiple files like `Microsoft.Extensions.Configuration` allows.
4) Eventually it should be moved, yes.
5) I would say remove the old way but I understand that might be too big.

My work situation has changed so I don't be able to continue on with this.  I could keep this here to keep working from or close if you'd like.  Thanks!

xiazen:
I think you can close it if you are not able to continue. 

I have made a copy of your branch and store [it in my repo](https://github.com/xiazen/orleans-xiazen/tree/adam/new_config_2). We might continue work on it on top of your progress if that is ok with you. :) 

Thanks for the effort you made here. Really appreciate it.

adamhathcock:
Glad it helped out.  Can't wait to see what 2.0 is like.

Issue #2029, opened 2016-08-08
Replace Logger with Microsoft.Extensions.Logging


adamhathcock:
https://github.com/aspnet/Logging

Basic change:
- Keep `LogManager` (replace with 2.0 and proper DI?)
- `LogManager` maintains an instance of `ILoggerFactory` which it uses to create `ILogger` instances

Effects:
- Pushes logging concerns to M.E.Logging and all the loggers that already integrate with it and ASP.NET Core
- Logging levels change
- APM methods on the abstract `Logger` have to go somewhere else.  (I feel like these should be on a class called Logger anyway)

Probably more things but I'm new to this code base.  This feels like the right thing to do to me when as moving to .NET Core is a goal.

adamhathcock:
If there's basic agreement, I can start on this.

veikkoeeva:
Ping @jdom, @sergeybykov. Seems related to #2030.

jason-bragg:
I am very much in favor of integrating an existing logging abstraction in lieu of maintaining an Orleans specific version.  Logging is not really our value add..

My preference would be for us to mature our dependency injection, and utilize it for logger implementation injection and move towards some well supported logging abstraction, as suggested here.

I am of the opinion that support for Microsoft.Extensions.Logging can be introduced via dependency injection independent of Orleans logging, after which Orleans logging can be refactored to call through to Microsoft.Extensions.Logging, souley to maintain backwards compatibility, while it is being deprecated.

I guess what I'm suggesting is to use DI and M.E.L to support and alternative logging system then deprecate the current one rather than continuing to maintain it.

adamhathcock:
I'd only start with the static `LogManager` just to be a halfway house as DI isn't fully implemented.

sergeybykov:
Completed in 2.0.

PR #1342, opened 2016-01-27
Injection of grain factory

jason-bragg:
Addresses "Dependency Injection: register IGrainFactory #988"

To inject the grain factory, we need only to add it to the services collection.  However, we can't do this cleanly because the Microsoft.Extensions.DependencyInjection libraries have not been released, so we can't access them directly from within core, and GrainFactory is an internal class that can't be accessed by the DI library.

Given that there was no clean solution, I compromised on the side of simplicity.  As an interim workaround I added a public singleton instance of grain factory as a -HACK- so we could register it in the services collection for DI.  This will be removed when Microsoft.Extensions.DependencyInjection is released and we can get rid of the OrleansDependencyInjection assembly.

Thoughts?

Issue #988, opened 2015-11-04
Dependency Injection: register IGrainFactory


ReubenBond:
Currently I cannot construct objects which require `IGrainFactory` using DI in Orleans because the implementation is not registered.

Example of a failure scenario:
```C#
public HasDeps : Grain, IHasDeps
{
  public HasDeps(SomeDep someDep) { /* ... */ }
}

// Imagine I register SomeDep in the container...
public SomeDep
{
  public SomeDep(IGrainFactory grainFactory) { /* ... */ }
}
```

 Is there a good reason for `IGrainFactory` (and friends?) not to be in the container?

nehmebilal:
Can you register the `IGrainFactory` yourself the same way you register any other dependency?

I think the long term plan is to potentially have multiple instances of the `IGrainFactory` (client connecting to multiple silos or silo connecting to another silo, see #467) in which case you'll need to use named dependencies or to inject some sort of service locator that allows you to retrieve the appropriate `GrainFactory` inside `SomeDep`.

ReubenBond:
This is on the silo, not the client. The `IGrainFactory` which is injected into `SomeDep` should be the same as the one injected into `HasDeps`.

nehmebilal:
How do you know? If the Silo is connecting to another Silo, there will be another instance of `IGrainFactory` that is associated with the second Silo. `SomeDep` is just a class and there is no way to tell if it's looking for currentSilo.GrainFactory or OtherSilo.GrainFactory.

ReubenBond:
It should take the instance from the current lifetime/scope, right? It would be weird if the `IGrainFactory` in `HasDeps` differs from that in `SomeDep` without me explicitly saying it should differ.

How do you propose we actually provide `IGrainFactory` to dependency classes?

jason-bragg:
>"Can you register the  IGrainFactory  yourself "

Not currently, its implementation is internal.

>"If the Silo is connecting to another Silo, there will be another instance of  IGrainFactory  that is associated with the second Silo"

???  This breaks the virtual actor model, does it not ???  I don't want to side track this discussion, but this statement seems to be counter to the fundamentals of the programing model.

At tertiary glance it looks like we can add the grain factory to the list of services (IServiceCollection) in RegisterSystemTypes().  We'll need to get it from the service provider when setting it in the runtime, or removed it from the runtime and always get it from the service provider (and other such cleanup).  Does not look hard...?

nehmebilal:
[...]

Issue #934, opened 2015-10-23
Orleans service provider not available to application layer.


jason-bragg:
A while back an IServiceProvider was introduced into the Silo to allow application developers to inject dependencies into Orleans.  The service provider is initialized with a configurable hook that creates a service provider from a service collection.

From test code:
public class TestStartup
{
  public IServiceProvider ConfigureServices(IServiceCollection services)
  {
    services.AddSingleton<IInjectedService, InjectedService>();
    services.AddTransient<SimpleDIGrain>();
    return services.BuildServiceProvider();
  }
}

This allows application developers to provide Orleans a service provider to use.  However, there is, to my knowledge, not currently a way for application logic to actually use this provider.  It is marked as internal and accessible only to Orleans code.

As it is, if I want to use the same service provider as Orleans uses, I need to provide my own service provider (as in the test code from above) but do it in a way where I can also return the same provider to my application code.  This is peculiarly inconvenient.

1) Is this isolation intended, or just an unfortunate implementation detail of the first pass implementation?
2) If we wanted to expose the Orleans service provider to the application layer, how should this be done?  Singleton?  Attach to static data like Silo? ...?

gabikliot:
Cc @attilah

attilah:
I owe the community a document page, but meanwhile the unit test for the DI would be a great starting point.

Since Orleans provides a precompiled host and usually you get a chance to add your own logic via Bootstrap providers, DI is implemented similar.

You can define a Startup class in the configuration file like:

````
    <Startup Type="UnitTests.General.TestStartup,Tester" />
````

We're utilizing Asp.Net vNext DI implementation, so their rules are our rules.

In this startup type you must define a ConfigureServices method, where you can register your own grains, services, etc. which will be later injected into your grains' ctor when you're using it.

Here is the Startup class from the test:
````
    public class TestStartup
    {
        public IServiceProvider ConfigureServices(IServiceCollection services)
        {
            services.AddSingleton<IInjectedService, InjectedService>();

            services.AddTransient<SimpleDIGrain>();

            return services.BuildServiceProvider();
        }
    }
````

If you are like me and like to use another DI container like Autofac (my preference), in this method you can configure that container and return the provider built by the Autofac adapter.

I hope this helps!

jason-bragg:
[...]

PR #2052, opened 2016-08-12
[WIP] SiloHostBuilder which is like ASP.NET Core WebHostBuilder.

adamhathcock:
Related to: https://github.com/dotnet/orleans/issues/2029 https://github.com/dotnet/orleans/issues/2030

Essentially, I'm trying to copy https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs with `ISiloHostBuilder`

I'd want build the logging and configuration stuff on stop of this interface.

`UseClusterConfiguration` would be replaced with usage of `IConfiguration` from M.E.Configuration
I guess ideally `ConfigureSilo` would be unneccessary too.  It looks like a lot of logic is exposed that ought to be hidden which is why I hid the `InitializeOrleansSilo` and `LoadOrleansConfig` methods.

https://github.com/dotnet/orleans/pull/2047 should rebased on top of this too if people like.


adamhathcock:
I think @jdom did the StartupBuilder which I removed here.

I realize this is destructive so looking for feedback about direction

adamhathcock:
Maybe this kind of stuff should go on the coreclr (2.0?) branch as with this issue: https://github.com/dotnet/orleans/issues/368

jdom:
I'll try to take a look at this very soon. In the meantime, let me summon @attilah, as he is the one who added it in the first place.

adamhathcock:
Thanks @jdom 

I really hope we can do something like this.  I'm open to anything.  I can plow ahead with this as well by moving some things from the Configuration interface (like the providers) so that it's done code-wise via the SiloHostBuilder.  They'd probably be a similar class for building external GrainClients.

src/OrleansHost/WindowsServerHost.cs
```diff
@@ -114,7 +108,8 @@ private int RunImpl(CancellationToken? cancellationToken = null)
         public bool ParseArguments(string[] args)
         {
             string siloName = Dns.GetHostName(); // Default to machine name
-            SiloHost = new SiloHost(siloName);
+            var builder = new SiloHostBuilder();
+            SiloHost = builder.Build(siloName);
```
xiazen: This is confusing to me. Initialization of siloHost should be in its own method, maybe called initSiloHost(args), it shouldn't be inside ParseArguments function. ParseAtguments func should be responsible just for parse args from command line params.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -137,15 +138,29 @@ public string ClusterId
...
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="siloType"></param>
+        /// <param name="config"></param>
+        public Silo(string name, SiloType siloType, ClusterConfiguration config)
+            : this(name, siloType, null, config, null)
+        {
+            var services = new ServiceCollection();
+            StartupBuilder.RegisterSystemTypes(services);
+            Services = services.BuildServiceProvider();
```
xiazen: This might be out of the scope of this PR. but speaking of dependency injection, ideally, Silo class shouldn't be responsible for registerSystemType(services). it would be nice that we decouple registerSystemType with Silo class. 
In this way,
1. we reduce responsibility from Silo class, make it more maintainable. Currently, silo class is responsible for a lot of things, many of those are not necessarily be Silo class's responsibility. 
2. internal system services will be accessible from outside Silo class, we have an issue https://github.com/dotnet/orleans/issues/934 which wants to make system services accessible from application layer. By moving RegisterSystemTypes outside of Silo class, this could potentially be the first step to resolve issue 934.

src/OrleansRuntime/Startup/SiloHostBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+            siloConfigureAction = silo;
+            return this;
+        }
+
+        public SiloHost Build(string name)
+        {
+            var services = new ServiceCollection();
+
+            StartupBuilder.RegisterSystemTypes(services);
+
```
xiazen: Ah I see, you moved RegisterSystemTypes here to part of HostBuilder. ignore my other comment then ;) 

But we do still have the issue #934, expose the system services to application layer grain and potentially other external providers.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -222,7 +238,6 @@ internal Silo(string name, SiloType siloType, ClusterConfiguration config, ILoca
 
             // Configure DI using Startup type
             bool usingCustomServiceProvider;
```
xiazen: Is this variable still needed here? I believe it is not used anywhere after this change.

src/OrleansRuntime/Startup/ISiloHostBuilder.cs
```diff
@@ -0,0 +1,46 @@
...
+using Orleans.Runtime.Configuration;
+using Orleans.Runtime.Host;
+
+namespace Orleans.Runtime.Startup
+{
+    /// <summary>
+    /// A builder for <see cref="SiloHost"/>.
+    /// </summary>
+    public interface ISiloHostBuilder
+    {
```
xiazen: I think LogManager's initialization can be part of this class's responsibility too, instead of Silo class's. See Silo.cs line  195, for the goal of remove responsibility from Silo class.

src/OrleansAzureUtils/Hosting/AzureSilo.cs
```diff
@@ -154,21 +155,20 @@ public bool Start(ClusterConfiguration config, string deploymentId = null, strin
...
-                host = new SiloHost(instanceName);
-                host.LoadOrleansConfig(); // Load config from file + Initializes logger configurations
-            }
-            else
-            {
-                host = new SiloHost(instanceName, config); // Use supplied config data + Initializes logger configurations
+                builder.UseClusterConfiguration(config); // Use supplied config data + Initializes logger configurations
             }
 
+            builder.ConfigureSilo(silo =>
```
jason-bragg: move mega-lambda into it's own function and pass it to config call.

src/OrleansRuntime/Startup/ISiloHostBuilder.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <param name="services">The delegate that can allow a custom IoC container to be used by the Orleans silo.</param>
+        /// <returns></returns>
+        ISiloHostBuilder BuildServiceProvider(Func<IServiceCollection, IServiceProvider> services);
+
+        /// <summary>
+        /// Do some post-creation configuration of the Silo.
+        /// </summary>
+        /// <param name="silo"></param>
+        /// <returns></returns>
+        ISiloHostBuilder ConfigureSilo(Action<SiloHost> silo);
```
jdom: Could you explain what this would do? I had the impression that all configuration would be done in the `ConfigureServices` method (basically by configuring the container registrations). I know that would be hard to achieve with today's internal compononents, but I'd like to understand whether you are heading in that direction or you had something different in mind.

src/OrleansRuntime/Startup/StartupBuilder.cs
```diff
@@ -12,89 +9,13 @@ namespace Orleans.Runtime.Startup
...
-            }
-
-            RegisterSystemTypes(serviceCollection);
-
-            if (usingCustomServiceProvider)
-            {
-                var instance = Activator.CreateInstance(startupType);
-                return servicesMethod.Build(instance, serviceCollection);
-            }
-
```
xiazen: I see you moved the functionality of configure ServiceProvider to ISiloHostBuilder. But I think ISiloHostBuilder just support Asp.Net DJ framework. But originally we support third-party DJ framework too, like Unity etc, and this code block above configure it up. Does this change would prevent us using Third-party DJ framwork?

src/OrleansRuntime/Startup/SiloHostBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Runtime.Configuration;
+using Orleans.Runtime.Host;
+
+namespace Orleans.Runtime.Startup
+{
+    public class SiloHostBuilder : ISiloHostBuilder
+    {
+        private ClusterConfiguration clusterConfiguration;
+        private readonly List<Action<IServiceCollection>> configureServicesDelegates = new List<Action<IServiceCollection>>();
```
xiazen: What's the reason you want to have a list of configureServiceDelegate, wouldn't one be enough?

attilah:
Thanks for this WIP @adamhathcock. It is something like what we were talking about, before on gitter with the others.

I'd personally make one step back and first work out the interfaces and a change spec in an issue before going to a PR, since this change is bigger than that IMHO.

Based on #2029 and #2030 I think we are better if working towards change of hosting from inside-out and not the way around, since it will involve more changes.

For example we're modifying this hosting part, but not how the configuration and logging is done, then we've to touch hosting part again. If we do it from the other way around it would be more approachable.

Removing ```StartupBuilder``` is like cutting a feature, since this is how startup discovery done in Asp.Net (one of the ways).

What is part of hosting in Asp.Net and I also can imagine support within Orleans is the support of Environments.

For all of these to work better I agree with @jason-bragg's comment (https://github.com/dotnet/orleans/issues/2029#issuecomment-238691920) what we need deeper and better DI usage first to leverage these new things thru DI, as it is done in Asp.Net.

src/OrleansHost/WindowsServerHost.cs
adamhathcock: This can be moved.  I was just getting it work.  However, I wouldn't have separate publically callable initialization methods.  The builder should take care of every option before creation.  Once something is created, it is fully initialized and ready to be used.

src/OrleansRuntime/Silo/Silo.cs
adamhathcock: Maybe not...this PR is messy on purpose to keep changes small for this initial "do you like this at all?" review :)

src/OrleansAzureUtils/Hosting/AzureSilo.cs
adamhathcock: This mega lambda exists only because I didn't think of how to do all of what this was doing before the "initialize" was called.  This would definitely be cleaned up if this general direction is liked.

src/OrleansRuntime/Startup/SiloHostBuilder.cs
adamhathcock: It isn't actually!

There could be tons of extension methods that want to configure the service collection.  If it is called more than once it would overwrite.  In the ASP.NET builder code (and my own applications) you need a list to organize different functionality with it's own initialization code (e.g. AddLogging(), AddTelemetry) which all need to add things to the `IServiceCollection`

src/OrleansRuntime/Startup/StartupBuilder.cs
adamhathcock: The addition of the `BuildServiceProvider` method on `SiloHostBuilder` is where you can use the `IServiceCollection` to use a 3rd-party DI container instead of the default one.

adamhathcock:
Btw, is there a list of areas that need to be cleaned up for usage by DI and constructor injection?  I'm scared to just jump in again.

src/OrleansRuntime/Silo/Silo.cs
xiazen: I see! new to the community. still getting used to the work style here.

jason-bragg:
"If the best place to start is removing more statics to get more constructor injection going with the current DI model then I will."

Approach and order are very debatable, and while I may have preferences, as long as we're making progress, I don't have a strong opinion.  Though, I'm not convinced we've a clear understanding of the problems we need to solve to mature the DI system.  I suggest we pick a capability then address problems as they come.

At an earlier point I took a pass at addressing "Dependency Injection: register IGrainFactory" #988, with PR "Injection of grain factory" #1342.  Ultimately this was abandoned because the solution was too hacky and required changes to GrainFactory.  Wiring up this specific feature, end to end (including ensuring it is available from third party containers, and can be injected into at least grains and providers), would be a useful capability that also surfaces shortcomings in our DI system.

This PR seems more related to config, so its somewhat orthogonal, but where configuration deals with DI, and how logging is improved are both influenced by the maturity of our DI integration.

adamhathcock:
@jason-bragg I hear what you're saying.  It's hard for me to see how to piecemeal some of these systems in and to share with the current code.  I was hoping I could more easily see that by attempting to just do some things.

I guess another way to attack this is to wrap the current config and expose more through the builder.  However, there are things that just conflict between the current style and the ASP.NET Core style.  

I'll look at attacking something smaller by seeing what's still static and removing them.  I know LogManager is but that's minor.  I want to understand more what requires AppDomain isolation and remove that requirement as that's a big barrier for .NET Core and this ASP.NET Core style stuff.

sergeybykov:
@adamhathcock 
>I want to understand more what requires AppDomain isolation and remove that requirement as that's a big barrier for .NET Core and this ASP.NET Core style stuff.

I don't believe we require AppDomain isolation for anything other than in the `AssemblyLoader` and the test harness. Unless I misunderstood what you meant.

xiazen:
I guess the general direction to go with DI is to have a DI system where both internal orleans layer and application layer can register service provider to , and it can also auto resolve dependency when initializing objects which has related dependency. 
While we do have a wide range of areas which used DI. Mainly in three areas, one is internal Orleans code, another is application grains and the third is external providers, such as storage providers, streaming providers. 
The DI pattern in internal Orleans code is service locator pattern, which result in that all the client class need to have a dependency on ServiceProvider and pass it through as a param in a chain... but since it's internal framework code, it's not an uncommon pattern to use. So I think we can put this part in a lower priority. **The work in this part is surface internal services to application layer and external providers**

For application layer grain, we are already supporting DI there. So if the work above is finished,  application layer grain would get those internal services automatically and also has a place to register services to. 

External providers should be similar to application layer grains, even though I think they are still using service locator pattern, which needs to be fixed. 

So I think, the first step of the DI system could look like this:
In cluster configuration step, we can add a step of wire all those dependency together, including register all services to a ServiceProvider, and also configure DI for application layer grains and external providers. 
More specifically, it can look like this
We can have a extension method called RegisterSystemTypes(IServiceCollection services) which is a shortcut for users to call to add all services required by internal orleans system. When user implement , they do
 IServiceProvider ConfigureServices(IServiceCollection services){
   service.RegisterSystemTypes();
  services.addOtherUserRequiredServices... 
// more DI configuration...
}

And ConfigureServices is before we spin up the silo host. 

While there's many ways we can do this. One is simply ask user to implement IStartupBuilder
interface IStartUpBuilder:
{
  void IServiceProvider ConfigureStartup(IServiceCollection services); 
} 
The one concern here is the user may forget to call service.RegisterSystemTypes(); then our internal orlean DI wouldn't be wired correctly. It also creates friction for first-time users

Another approach is 
class StartUpBuilder{
   IUserServiceBuilder builder;
   void IServiceProvider ConfigureStartup()
{
  services = builder.ConfigureServiceCollection();
if(system type not registered) services.RegisterSystemTypes();
}
}

interface IUserServiceBuilder: {
  //register services and configure DI
 // if this user is advanced orleans user, he can RegisterSystemTypes here so his application layer grain and external providers can have access to internal services. 
  IServiceCollection ConfigureServiceCollection (IServiceCollection services)
}


This is just me rambling on to get some opinion back. And I agree with you, what I just said is beyond the scope of this PR. I think we can have another PR to do what I just said. And I just came out of this in about one hour. Not much thought. Welcome for opinions.

xiazen:
Hi Adam,

the orleans core team had a LONG discussion about how we want to make progress on DI. And as follows is our opinion. 

Step 1: make internal system types registering through DI container, as a way to make them injectable/usable to application layer, and also as a way to make our DI system better. Here is an  PR #2192  about making GrainFactory injectable as a first step going in this direction.

In the future, we want to register more system types to DI to make our DI system better, and also leverage `Silo `class's responsibility of initializing everything to DI system and SiloHostBuilder.  We may or may not want to make those internal system types injectable to application layer through public interface, depends on the community's needs. 

Step 2: We talked about make internal types replaceable to give orleans user flexibility to inject their own implementation of internal types, as an attempt to in line with the pattern in ASP.Net. We are holding off on that path now, since the community doesn't have this kind of requirement yet. But we did make an effort to make GrainFactory replaceable, if you are interested, you can go look at my repo: https://github.com/xiazen/orleans-xiazen/tree/DI

adamhathcock:
@xiazen that all sounds good to me.  I mostly did this to drive forward discussion and see things get done.

Unfortunately, my work situation has changed and I can't follow through on this but I still want to use Orleans in the future.

PR #2053, opened 2016-08-13
AWS SQS Stream Provider

galvesribeiro:

### DIFF
--- a/src/OrleansAWSUtils/Storage/SQSStorage.cs
+++ b/src/OrleansAWSUtils/Storage/SQSStorage.cs
@@ -126,19 +126,18 @@ namespace OrleansAWSUtils.Storage
         /// Delete the queue
         /// </summary>
         /// <returns></returns>
-        public Task DeleteQueue()
+        public async Task DeleteQueue()
         {
             try
             {
                 if (string.IsNullOrWhiteSpace(queueUrl))
                     throw new InvalidOperationException("Queue not initialized");
-                return sqsClient.DeleteQueueAsync(queueUrl);
+                await sqsClient.DeleteQueueAsync(queueUrl);
             }
             catch (Exception exc)
             {
                 ReportErrorAndRethrow(exc, "DeleteQueue", ErrorCode.StreamProviderManagerBase);
             }
-            return null;
         }
 
         #endregion
@@ -178,10 +177,10 @@ namespace OrleansAWSUtils.Storage
                 if (string.IsNullOrWhiteSpace(queueUrl))
                     throw new InvalidOperationException("Queue not initialized");
 
-                if (count < 1 || count > MAX_NUMBER_OF_MESSAGE_TO_PEAK)
+                if (count < 1)
                     throw new ArgumentOutOfRangeException(nameof(count));
 
-                var request = new ReceiveMessageRequest { QueueUrl = queueUrl, MaxNumberOfMessages = count };
+                var request = new ReceiveMessageRequest { QueueUrl = queueUrl, MaxNumberOfMessages = count <= MAX_NUMBER_OF_MESSAGE_TO_PEAK ? count : MAX_NUMBER_OF_MESSAGE_TO_PEAK };
                 var response = await sqsClient.ReceiveMessageAsync(request);
                 return response.Messages;
             }
--- a/src/OrleansAWSUtils/Storage/SQSStorage.cs
+++ b/src/OrleansAWSUtils/Storage/SQSStorage.cs
@@ -1,0 +1,233 @@
+﻿using Amazon.Runtime;
+using Amazon.SQS;
+using Orleans.Runtime;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Amazon.SQS.Model;
+using Orleans;
+using SQSMessage = Amazon.SQS.Model.Message;
+
+namespace OrleansAWSUtils.Storage
+{
+    /// <summary>
+    /// Wrapper/Helper class around AWS SQS queue service
+    /// </summary>
+    public class SQSStorage
+    {
+        /// <summary>
+        /// Maximum number of messages allowed by SQS to peak per request
+        /// </summary>
+        public const int MAX_NUMBER_OF_MESSAGE_TO_PEAK = 10;
+        private const string AccessKeyPropertyName = "AccessKey";
+        private const string SecretKeyPropertyName = "SecretKey";
+        private const string ServicePropertyName = "Service";
+        private Logger Logger;
+        private string accessKey;
+        private string secretKey;
+        private string service;
+        private string queueUrl;
+        private AmazonSQSClient sqsClient;
+
+        /// <summary>
+        /// The queue Name
+        /// </summary>
+        public string QueueName { get; private set; }
+
+        /// <summary>
+        /// Default Ctor
+        /// </summary>
+        /// <param name="queueName">The name of the queue</param>
+        /// <param name="connectionString">The connection string</param>
+        /// <param name="deploymentId">The cluster deployment Id</param>
+        public SQSStorage(string queueName, string connectionString, string deploymentId = "")
+        {
+            QueueName = string.IsNullOrWhiteSpace(deploymentId) ? queueName : $"{deploymentId}-{queueName}";
+            ParseDataConnectionString(connectionString);
+            Logger = LogManager.GetLogger($"SQSStorage", LoggerType.Runtime);
+            CreateClient();
+        }
+
+        #region Queue Management Operations
+
+        private void ParseDataConnectionString(string dataConnectionString)
+        {
+            var parameters = dataConnectionString.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+
+            var serviceConfig = parameters.Where(p => p.Contains(ServicePropertyName)).FirstOrDefault();
+            if (!string.IsNullOrWhiteSpace(serviceConfig))
+            {
+                var value = serviceConfig.Split(new[] { '=' }, StringSplitOptions.RemoveEmptyEntries);
+                if (value.Length == 2 && !string.IsNullOrWhiteSpace(value[1]))
+                    service = value[1];
+            }
+
+            var secretKeyConfig = parameters.Where(p => p.Contains(SecretKeyPropertyName)).FirstOrDefault();
+            if (!string.IsNullOrWhiteSpace(secretKeyConfig))
+            {
+                var value = secretKeyConfig.Split(new[] { '=' }, StringSplitOptions.RemoveEmptyEntries);
+                if (value.Length == 2 && !string.IsNullOrWhiteSpace(value[1]))
+                    secretKey = value[1];
+            }
+
+            var accessKeyConfig = parameters.Where(p => p.Contains(AccessKeyPropertyName)).FirstOrDefault();
+            if (!string.IsNullOrWhiteSpace(accessKeyConfig))
+            {
+                var value = accessKeyConfig.Split(new[] { '=' }, StringSplitOptions.RemoveEmptyEntries);
+                if (value.Length == 2 && !string.IsNullOrWhiteSpace(value[1]))
+                    accessKey = value[1];
+            }
+        }
+
+        private void CreateClient()
+        {
+            var credentials = new BasicAWSCredentials(accessKey, secretKey);
+            sqsClient = new AmazonSQSClient(credentials, new AmazonSQSConfig { RegionEndpoint = AWSUtils.GetRegionEndpoint(service) });
+        }
+
+        private async Task<string> GetQueueUrl()
+        {
+            try
+            {
+                var response = await sqsClient.GetQueueUrlAsync(QueueName);
+                if (!string.IsNullOrWhiteSpace(response.QueueUrl))
+                    queueUrl = response.QueueUrl;
+
+                return queueUrl;
+            }
+            catch (QueueDoesNotExistException)
+            {
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Initialize SQSStorage by creating or connecting to an existent queue
+        /// </summary>
+        /// <returns></returns>
+        public async Task InitQueueAsync()
+        {
+            try
+            {
+                if (string.IsNullOrWhiteSpace(await GetQueueUrl()))
+                {
+                    var response = await sqsClient.CreateQueueAsync(QueueName);
+                    queueUrl = response.QueueUrl;
+                }
+            }
+            catch (Exception exc)
+            {
+                ReportErrorAndRethrow(exc, "InitQueueAsync", ErrorCode.StreamProviderManagerBase);
+            }
+        }
+
+        /// <summary>
+        /// Delete the queue
+        /// </summary>
+        /// <returns></returns>
+        public Task DeleteQueue()
+        {
+            try
+            {
+                if (string.IsNullOrWhiteSpace(queueUrl))
+                    throw new InvalidOperationException("Queue not initialized");
+                return sqsClient.DeleteQueueAsync(queueUrl);
+            }
+            catch (Exception exc)
+            {
+                ReportErrorAndRethrow(exc, "DeleteQueue", ErrorCode.StreamProviderManagerBase);
+            }
+            return null;
+        }
+
+        #endregion
+
+        #region Messaging
+
+        /// <summary>
+        /// Add a message to the SQS queue
+        /// </summary>
+        /// <param name="message">Message request</param>
+        /// <returns></returns>
+        public async Task AddMessage(SendMessageRequest message)
+        {
+            try
+            {
+                if (string.IsNullOrWhiteSpace(queueUrl))
+                    throw new InvalidOperationException("Queue not initialized");
+
+                message.QueueUrl = queueUrl;
+                await sqsClient.SendMessageAsync(message);
+            }
+            catch (Exception exc)
+            {
+                ReportErrorAndRethrow(exc, "AddMessage", ErrorCode.StreamProviderManagerBase);
+            }
+        }
+
+        /// <summary>
+        /// Get Messages from SQS Queue.
+        /// </summary>
+        /// <param name="count">The number of messages to peak. Min 1 and max 10</param>
+        /// <returns>Collection with messages from the queue</returns>
+        public async Task<IEnumerable<SQSMessage>> GetMessages(int count = 1)
+        {
+            try
+            {
+                if (string.IsNullOrWhiteSpace(queueUrl))
+                    throw new InvalidOperationException("Queue not initialized");
+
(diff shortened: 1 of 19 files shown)


src/OrleansSQLUtils/Storage_SqlServer.sql
```diff
@@ -0,0 +1,421 @@
+﻿
```
dVakulen: Seems like return of the query from https://github.com/dotnet/orleans/pull/2038

src/Orleans/Telemetry/Consumers/FileTelemetryConsumer.cs
```diff
@@ -70,8 +70,7 @@ public void Close()
                         return;
                     }
                     _logOutput.Flush();
-                    _logOutput.Dispose();
```
dVakulen: Revert of the https://github.com/dotnet/orleans/pull/2034/files

test/TesterInternal/Config_SQSStreamProviders.xml
```diff
@@ -0,0 +1,39 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
dVakulen: Probably it's better to use programmatic config instead.

galvesribeiro:
@dVakulen sorry, now it is properly rebased.

galvesribeiro:
@jason-bragg since you was playing with Streams recently, do you mind review this as well? Thanks

src/OrleansAWSUtils/Storage/SQSStorage.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// </summary>
+        public string QueueName { get; private set; }
+
+        /// <summary>
+        /// Default Ctor
+        /// </summary>
+        /// <param name="queueName">The name of the queue</param>
+        /// <param name="connectionString">The connection string</param>
+        /// <param name="deploymentId">The cluster deployment Id</param>
+        public SQSStorage(string queueName, string connectionString, string deploymentId = "")
```
jason-bragg: Separation of concerns - 
Parsing fields out of some configuration data should not be part of this class.  Construct with actual parameters or with interface for configuration class.  Parsing should be done between configuration and this class, not part of this class.

src/OrleansAWSUtils/Storage/SQSStorage.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// Delete the queue
+        /// </summary>
+        /// <returns></returns>
+        public Task DeleteQueue()
+        {
+            try
+            {
+                if (string.IsNullOrWhiteSpace(queueUrl))
+                    throw new InvalidOperationException("Queue not initialized");
+                return sqsClient.DeleteQueueAsync(queueUrl);
```
jason-bragg: await instead of return.  By returning delete, exceptions will not be caught in below catch.

src/OrleansAWSUtils/Storage/SQSStorage.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// <param name="count">The number of messages to peak. Min 1 and max 10</param>
+        /// <returns>Collection with messages from the queue</returns>
+        public async Task<IEnumerable<SQSMessage>> GetMessages(int count = 1)
+        {
+            try
+            {
+                if (string.IsNullOrWhiteSpace(queueUrl))
+                    throw new InvalidOperationException("Queue not initialized");
+
+                if (count < 1 || count > MAX_NUMBER_OF_MESSAGE_TO_PEAK)
```
jason-bragg: suggest capping count at MAX_NUMBER_OF_MESSAGE_TO_PEAK but letting operation continue rather than throwing if too many items are requested.

src/OrleansAWSUtils/Streams/SQSAdapter.cs
```diff
@@ -0,0 +1,56 @@
...
+        public IQueueAdapterReceiver CreateReceiver(QueueId queueId)
+        {
+            return SQSAdapterReceiver.Create(queueId, DataConnectionString, DeploymentId);
+        }
+
+        public async Task QueueMessageBatchAsync<T>(Guid streamGuid, string streamNamespace, IEnumerable<T> events, StreamSequenceToken token, Dictionary<string, object> requestContext)
+        {
+            if (token != null)
+            {
+                throw new ArgumentException("AzureQueue stream provider currebtly does not support non-null StreamSequenceToken.", "token");
```
jason-bragg: AzureQueue?

src/OrleansAWSUtils/Streams/SQSAdapter.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace OrleansAWSUtils.Streams
+{
+    internal class SQSAdapter : IQueueAdapter
+    {
+        protected readonly string DeploymentId;
+        protected readonly string DataConnectionString;
+        private readonly HashRingBasedStreamQueueMapper streamQueueMapper;
```
jason-bragg: concrete type for queue mapper?  Isn't interface sufficient?

src/OrleansAWSUtils/Streams/SQSAdapterFactory.cs
```diff
@@ -0,0 +1,95 @@
...
+            if (!config.Properties.TryGetValue(DataConnectionStringPropertyName, out dataConnectionString))
+                throw new ArgumentException(string.Format("{0} property not set", DataConnectionStringPropertyName));
+            if (!config.Properties.TryGetValue(DeploymentIdPropertyName, out deploymentId))
+                throw new ArgumentException(string.Format("{0} property not set", DeploymentIdPropertyName));
+
+            cacheSize = SimpleQueueAdapterCache.ParseSize(config, CacheSizeDefaultValue);
+
+            string numQueuesString;
+            numQueues = NumQueuesDefaultValue;
+            if (config.Properties.TryGetValue(NumQueuesPropertyName, out numQueuesString))
```
jason-bragg: plz use ProviderConfigurationExtensions.
GetIntProperty

src/OrleansAWSUtils/Streams/SQSAdapterReceiver.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        public QueueId Id { get; private set; }
+
+        public static IQueueAdapterReceiver Create(QueueId queueId, string dataConnectionString, string deploymentId)
+        {
+            if (queueId == null) throw new ArgumentNullException("queueId");
+            if (string.IsNullOrEmpty(dataConnectionString)) throw new ArgumentNullException("dataConnectionString");
+            if (string.IsNullOrEmpty(deploymentId)) throw new ArgumentNullException("deploymentId");
+
+            var queue = new SQSStorage(queueId.ToString(), dataConnectionString, deploymentId);
```
jason-bragg: I don't know naming rules of SQS.  Please ensure queueId.ToString is a valid queue name.  If not we'll need some sort of name 'sanitizer'.

src/OrleansAWSUtils/Streams/SQSBatchContainer.cs
```diff
@@ -0,0 +1,112 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using SQSMessage = Amazon.SQS.Model.Message;
+
+namespace OrleansAWSUtils.Streams
+{
+    [Serializable]
+    internal class SQSBatchContainer : IBatchContainer
```
jason-bragg: This batch container is almost identical to azure queue batch container, except it has a SQSMessage instead of a CloudQueueMessage. 
Plz consider using a common base class.

src/OrleansAWSUtils/Storage/SQSStorage.cs
galvesribeiro: Although it is the kind of change that IMHO is pure cosmetic and just a judgement call since there is no real benefit on that (as you can see on SQL connections which are pure strings), I'll make this change not just here but on the DynamoDBStorage as well, since their connection strings are basically the same. I'll do that in another PR before we release the nuget package itself for AWSUtils among other cosmetic issues.

test/Tester/StreamingTests/SQSClientStreamTests.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public override void Dispose()
+        {
+            var deploymentId = HostedCluster.DeploymentId;
+            base.Dispose();
+            SQSStreamProviderUtils.DeleteAllUsedQueues(SQSStreamProviderName, deploymentId, StorageConnectionString).Wait();
+        }
+
+        [Fact, TestCategory("Functional"), TestCategory("AWS"), TestCategory("Storage"), TestCategory("Streaming")]
+        public async Task SQSStreamProducerOnDroppedClientTest()
```
jason-bragg: Please mark only as AWS.  We can add more categories as test infrastructure improves, but adding "Functional" or "Storage" or "Streaming" will break anyone who doesn't have AWS setup when testing those categories.

test/TesterInternal/ClientConfig_SQSStreamProviders.xml
```diff
@@ -0,0 +1,18 @@
...
+  <Tracing DefaultTraceLevel="Info" TraceToConsole="false" TraceToFile="{0}-{1}.log" BulkMessageLimit="1000">
+    <TraceLevelOverride LogPrefix="Runtime" TraceLevel="Info" />
+    <TraceLevelOverride LogPrefix="Application" TraceLevel="Info" />
+    <TraceLevelOverride LogPrefix="Tetris" TraceLevel="Info" />
+    <TraceLevelOverride LogPrefix="AssemblyLoader" TraceLevel="Warning" />
+  </Tracing>
+  <Statistics MetricsTableWriteInterval="300s" PerfCounterWriteInterval="30s" LogWriteInterval="300s" StatisticsCollectionLevel="Info"/>
+  <Messaging ResponseTimeout="30s" ClientSenderBuckets="8192" MaxResendCount="0" UseMessageBatching="false" MaxMessageBatchingSize="100"/>
+  <StreamProviders>
+    <Provider Type="Orleans.Providers.Streams.SimpleMessageStream.SimpleMessageStreamProvider" Name="SMSProvider" FireAndForgetDelivery="false"/>
```
jason-bragg: SimpleMessageStreamProvider?

jason-bragg: Why do we need this.  Shouldn't we be using programmatic configurations now?

src/OrleansAWSUtils/Storage/SQSStorage.cs
galvesribeiro: Done.

test/Tester/StreamingTests/SQSSubscriptionMultiplicityTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        public override void Dispose()
+        {
+            var deploymentId = HostedCluster.DeploymentId;
+            base.Dispose();
+            SQSStreamProviderUtils.DeleteAllUsedQueues(SQSStreamProviderName, deploymentId, StreamConnectionString).Wait();
+        }
+
+        [Fact, TestCategory("Functional"), TestCategory("AWS"), TestCategory("Storage"), TestCategory("Streaming")]
```
jason-bragg: Please mark only as AWS - For all added test.

test/TesterInternal/StreamingTests/SQSStreamTests.cs
```diff
@@ -0,0 +1,192 @@
...
+using UnitTests.StreamingTests;
+using Orleans.Providers.Streams;
+using System.Collections.Generic;
+using Orleans.Storage;
+
+namespace UnitTests.Streaming
+{
+    public class SQSStreamTests : HostedTestClusterPerTest
+    {
+        internal static readonly FileInfo SiloConfigFile = new FileInfo("Config_SQSStreamProviders.xml");
```
jason-bragg: plz use programmatic config.

src/OrleansAWSUtils/Storage/SQSStorage.cs
galvesribeiro: Done

src/OrleansAWSUtils/Streams/SQSAdapter.cs
galvesribeiro: Ops... The implementation is 99% the same as Azure, so copy and paste miss that...

src/OrleansAWSUtils/Streams/SQSAdapter.cs
galvesribeiro: Humm... In AzureQueueAdapter it uses the concrete type as well... Changed here to `IConsistentRingStreamQueueMapper`

src/OrleansAWSUtils/Storage/SQSStorage.cs
```diff
@@ -0,0 +1,233 @@
...
+{
+    /// <summary>
+    /// Wrapper/Helper class around AWS SQS queue service
+    /// </summary>
+    public class SQSStorage
+    {
+        /// <summary>
+        /// Maximum number of messages allowed by SQS to peak per request
+        /// </summary>
+        public const int MAX_NUMBER_OF_MESSAGE_TO_PEAK = 10;
```
jason-bragg: Why have this?

src/OrleansAWSUtils/Streams/SQSAdapterFactory.cs
galvesribeiro: Done

src/OrleansAWSUtils/Streams/SQSAdapterReceiver.cs
galvesribeiro: Yes it is valid.

> A queue name can have up to 80 characters. The following characters will be accepted: alphanumeric characters, hyphens (-), and underscores (_).

src/OrleansAWSUtils/Streams/SQSBatchContainer.cs
galvesribeiro: All this stream implementation is almost the same as the Azure. In fact, almost all Queue-based implementation will be the same...

I would yes think on a refactory on both Azure and AWS Utils and move common code on a base library but, I think there are too much things in state today to mess with the production code now of Azure. But yes, I agree with you, will make a note here and have a look at it later on.

test/Tester/StreamingTests/SQSClientStreamTests.cs
galvesribeiro: OK

test/Tester/StreamingTests/SQSSubscriptionMultiplicityTests.cs
galvesribeiro: OK

jason-bragg:
@galvesribeiro Thanks for working on this.  Streaming support for AWS is a great addition.

This stream provider looks based off of the AzureQueue stream provider.  Azure queues are unordered and can introduce duplicates, does SQS have the same characteristics?  I ask because ordered queues like Kafka or EventHub should be handled different, and I want to be somewhat confident we're setting up this provider in a way that is appropriate for it's data.

src/OrleansAWSUtils/Storage/SQSStorage.cs
galvesribeiro: As the comment mention, SQS limit to a 10 msg pooling per request. I know it is only used in a single place but, in the Storage provider for example, other reviewers asked to move the limit to a constant.

galvesribeiro:
Yes, SQS don't guarantee order. SQS has mostly the same feature as well as the limitation on Azure Queues and that is why the code is almost copy/pasted from Azure.

I plan to have a refactory later to make a "GenericQueueStreamProvider" where the only thing that would basically change, is the backing Queue.

Something like EventHub and Kafka on AWS, is called Kinesis, and is other provider type that someone else in the community is already working and will should put a PR soon.

Bear with me that the cosmetic changes that you suggested will be address in a final PR that will add more descriptive error codes and some other cosmetic change not just to this PR but to all the other providers as I already agreed with the other reviewers of the AWSUtils.

Thanks for take time to review @jason-bragg, I'll push the commit with the changes in a few minutes.

src/OrleansAWSUtils/Streams/SQSBatchContainer.cs
jason-bragg: Agreed.  Duplicate code sucks, but building a common batch container base class can easily be argued as out of scope for this PR.

src/OrleansAWSUtils/Storage/SQSStorage.cs
jason-bragg: Ah, I thought this was a limit we were introducing.  If this is a backend service limitation, is there a constant in the service library we can use?

src/OrleansAWSUtils/Streams/SQSAdapterReceiver.cs
jason-bragg: KK.  If queueId is created with a queue prefix that does not follow there rules, it could break, but this is safe enough.

src/OrleansAWSUtils/Storage/SQSStorage.cs
galvesribeiro: You mean on AWS SDK? Not it just throw an exception if we pass the limits.

jason-bragg: https://www.youtube.com/watch?v=p76S-x4d5OA

galvesribeiro: lol Yeah... I'm not going comment on that :P Believe me... There are worse things than that ;)

src/OrleansAWSUtils/Streams/SQSBatchContainer.cs
gabikliot: ye, ...., as long as someone DOES promise to go back and refactor. Just saying sucks but lettting it go is not good enough in my view.

galvesribeiro: I'll do it myself. Just need to get this AWS package shipped and will get this refactor. It will probably introduce a common class in the core and those providers will just pass a backend implementation on a single interface. All this stream and Storage providers looks like the same code to me.

galvesribeiro:
@jason-bragg the code config on the tests is the only feedback I intentionally didn't made the changes from your feedback because before we release the beta nuget of AWSUtils, I'll move all the tests to a separated project as @jdom suggested on other thread, so there I'll use code config. Is there anything else? This is the last PR for this bare minimal set of providers since all the other were merged.

Thanks

galvesribeiro:
Thanks @jason-bragg 

Will put another PR with the remaining fixes soon.

sergeybykov:
Thank you, @galvesribeiro, for pulling this off!

PR #1586, opened 2016-03-19
Cached type shallow copy availability

dVakulen:
Addresses https://github.com/dotnet/orleans/issues/1305, 

Checking for ability of shallow copying is taking considerable amount of CPU:
bc0e53dc-e0d8-11e5-9dd9-38540a102571

Benchmark, 3 000 000 repeats:
```ini
BenchmarkDotNet=v0.9.1.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i5-4200M CPU @ 2.50GHz, ProcessorCount=4
Frequency=2435776 ticks, Resolution=410.5468 ns
HostCLR=MS.NET 4.0.30319.34209, Arch=32-bit RELEASE

Type=ShallowCopyableBench  Mode=Throughput  LaunchCount=1  
WarmupCount=1  TargetCount=3  

```
                        Method |        Median |      StdDev | ThreadsCount |
------------------------------ |-------------- |------------ |------------- |
                   CachedClass |   106.5624 ms |   3.1811 ms |            1 |
                 CachedDecimal |   106.4367 ms |   1.8021 ms |            1 |
               CachedImmutable |   104.3393 ms |   1.7869 ms |            1 |
    CachedShallowCopyableClass |   102.3058 ms |  37.8404 ms |            1 |
               CachedValueType |   102.6185 ms |  25.0215 ms |            1 |
                NonCachedClass | 3,847.8685 ms | 116.0603 ms |            1 |
              NonCachedDecimal |   409.7000 ms |   7.1004 ms |            1 |
            NonCachedImmutable | 3,624.7472 ms | 130.7426 ms |            1 |
 NonCachedShallowCopyableClass | 6,778.8531 ms | 433.8992 ms |            1 |
            NonCachedValueType | 3,497.3040 ms |  19.6930 ms |            1 |
                   CachedClass |    65.8086 ms |   4.0328 ms |            3 |
                 CachedDecimal |    61.1600 ms |   2.6875 ms |            3 |
               CachedImmutable |    62.1187 ms |   1.4798 ms |            3 |
    CachedShallowCopyableClass |    63.8821 ms |   1.1557 ms |            3 |
               CachedValueType |    61.7941 ms |   0.1123 ms |            3 |
                NonCachedClass | 2,007.5797 ms |  30.7070 ms |            3 |
              NonCachedDecimal |   157.9320 ms |   1.2766 ms |            3 |
[...]

Issue #694, opened 2015-08-13
pre-emptive multi tasking


ashkan-saeedi-mazdeh:
Currently Orleans uses co-operative multi tasking and it means the grain itself is responsible for releasing the CPU and wait. In runtimes like Erlang VM actors can not execute more than a certain amount of computation at a time so system's responsiveness goes up because of this even if some bad actors/processes want to do a hell lot of work.

I think implementing this in Orleans might require CLR support but am not fully sure about it so felt asking.

Is it possible to implement such thing in Orleans to stop grains by the runtime with their state and then rerun them again after other grains get a chance of doing some work?

gabikliot:
In general, we opted for a co-operative multi tasking since that provides a much better overall system performance. Due to that we can easily run at very high CPU utilization (95% and more), fully utilizing the system.

However, Orleans actually already has a somewhat partial way to support preemptive multitasking. It can be controlled by setting [the number of active threads](https://github.com/dotnet/orleans/blob/master/src/Orleans/Configuration/NodeConfiguration.cs#L116).
By default it is set to num cores, but if you set it to high, like 100 or 1000, you will effectively get an effect of OS time sharing the application threads (and thus grains). That way one grain has much less chance to impact responsiveness . The downside is that the overall system throughput (for well behaved apps) will go down quite a bit and in fact even the responsiveness can go down as a side effect. But for worst case, "untrusted"/potentially buggy code, it will provide some sort if guarantee on responsiveness.

In fact, we explored that option in our paper and you can read about it here: http://research.microsoft.com/pubs/210931/Orleans-MSR-TR-2014-41.pdf, section 5.1.

ashkan-saeedi-mazdeh:
@gabikliot  Well if you want to run each actor on a single thread then yes the cost would be too high but if like Erlang the VM had a mechanism much more lightweight than threads to context switch between them and could put them on pending and reexectue them then the cost would not be that high.
I think you have done it partly in terms of actor isolation, because your runtime translates calls to actor references to messages actors can not reference each other even on the same machine and modify states of each other but if you want to stop one of your actors from executing its code and rerun it later on, CLR should provide the possibility for you or you have to use an OS thread for each actor which is damn expensive.

[...]

PR #2060, opened 2016-08-15
Threading model rework

dVakulen:
Currently Orleans uses SEDA (Staged Execution Design Architecture) implemented with fixed dedicated threads per stage and <code>BlockingCollection</code>'s as queues. However, the performance is highly sensitive to the number of threads in each SEDA stage and queuing delay reportedly dominates the end-to-end latency, and by far exceeds the network latency, as well as processing time in each stage, so in order to achieve lowest possible latency dynamic threads reallocation among the stages is needed. 

On the other hand using <code>BlockingCollection</code> doesn't seems as optimal for Orleans use case: it's synchronization algorithms consumes quite big part of CPU time. 

Results of profiling: 

```
  Including system methods:
```

timeline64_2016-08-15_21-29-55
- top five methods by CPU usage are system ones, and caused by internals of the <code>BlockingCollection</code>
  
  ```
   Excluding system methods:
  ```

timeline64_2016-08-15_21-35-16
- Top 3 again related to the queue usage, and fourth one - <code>AsynchThreadProc</code> is caused by semaphore usage in fixed threads processor.

As solution each stage was implemented as dataflow block, which allows to handle dynamic thread reallocation automatically, plus requires very little synchronization.

Profile results with dataflow stages:

```
   With system methods:
```

timeline64_2016-08-15_21-50-35

```
   And without:
```

timeline64_2016-08-15_21-49-59

All thread synchronization hot spots are gone except Monitor.Enter, which does not caused by usage of Dataflow library. 

Actual benchmark from https://github.com/dotnet/orleans/pull/2003 shows 34 - 50 % (depends on amount of data for processing)  perfomance improvement while consuming 4 - 8 % less CPU. 

All OrleansTaskScheduler tests are passing, except few ones can randomly fail, but same fails can be observed  even more frequently when running against current master. 

Current implementation have some minor todo's, which will be addressed after Orleans internal performance test run, which will also show whether it worthy reviewing. Also in case of need it's possible to implement swappable execution models, but it will be not an easy task.


veikkoeeva:
Very interesting use of TPL Dataflow! Do you have an idea how this would factor in if the networking library would be outsourced?

galvesribeiro:
@veikkoeeva you are stealing my questions everywhere :P I would like to see how it work for example with DotNetty or libUV since they have their own threading model as well...

shayhatsor:
Is this the beginning of an implementation of the threading model found here: http://research.microsoft.com/apps/pubs/default.aspx?id=263907 ?

dVakulen:
@shayhatsor Yes, forgot to include it for context

shayhatsor:
> in order to achieve lowest possible latency dynamic threads reallocation among the stages is needed. 
Can someone please explain how dynamic threads reallocation works ? What are the instigators for reallocation ? Maybe a short use case or example will make it clear.

dVakulen:
@veikkoeeva This won't matter almost at all - I haven't touched the networking pieces. As for own threading models of external libraries - most likely they were built with respect to default thread pool.

veikkoeeva:
@shayhatsor I haven't read the article, but I assume that is a reference to *work stealing*. I believe TPL Dataflow has a a rather tuned system for preserving hot data and instructions on appropriate caches (i.e. reduced cache thrashing) but still having smarts to balance throughput by stealing work from other queues (which is a queue going to another core likely). I don't have examples at hand, but *work stealing* should get you forwards. This isn't a new matter, has come up already in [Available Now: Preview of Project “Orleans” – Cloud Services at Scale (see comments)](https://blogs.msdn.microsoft.com/dotnet/2014/04/02/available-now-preview-of-project-orleans-cloud-services-at-scale/). You may also become off as more well-rounded from links at https://github.com/dotnet/orleans/issues/694.

veikkoeeva:
@dVakulen If you're taking it from that paper, I think I need to take a good read of it. I'm sure you have thought about this more than most before this PR, so this probably turns up to something interesting &ndash; and beneficial.

jdom:
Brilliant, I've been wanting to do something like this for a long time. Haven't looked at the implementation yet, but I definitely will.
Also, I'll immediately pull your changes and kick off some of the load tests that we run nightly.

src/OrleansRuntime/Scheduler/WorkerPoolThread.cs
```diff
@@ -102,8 +103,11 @@ internal WorkerPoolThread(WorkerPool gtp, OrleansTaskScheduler sched, int thread
                 WorkerThreadStatisticsNumber = SchedulerStatisticsGroup.RegisterWorkingThread(Name);
         }
 
+        private readonly ActionBlock<IWorkItem> _executor;
         protected override void Run()
         {
+            // not used anymore
+            Thread.Sleep(int.MaxValue);
```
gabikliot: ??

src/OrleansRuntime/Scheduler/WorkerPoolThread.cs
```diff
@@ -133,9 +137,9 @@ protected override void Run()
 #endif
                         // Get some work to do
                         IWorkItem todo;
-
-                        todo = IsSystem ? scheduler.RunQueue.GetSystem(Cts.Token, maxWorkQueueWait) : 
-                            scheduler.RunQueue.Get(Cts.Token, maxWorkQueueWait);
+                        todo = null;
+                        //todo = IsSystem ? scheduler.RunQueue.GetSystem(Cts.Token, maxWorkQueueWait) :
```
gabikliot: there is no special support any more for high pro system activities?

src/Orleans/Runtime/RuntimeContext.cs
```diff
@@ -5,14 +5,15 @@ namespace Orleans.Runtime
 {
     internal class RuntimeContext
     {
-        public TaskScheduler Scheduler { get; private set; }
+        public TaskScheduler Scheduler { get;  set; }
```
gabikliot: we use InitializeThread instead of making setter public.

src/OrleansRuntime/Messaging/InboundMessageQueue.cs
```diff
@@ -16,22 +17,24 @@ public int Count
             get
             {
                 int n = 0;
-                foreach (var queue in messageQueues)
-                    n += queue.Count;
-                
+                // currently doesn't being used
```
gabikliot: "currently doesn't being used" :  and that is a good reason to retrun 0 instead?

gabikliot:
Good initiative, thank you @dVakulen.

I added some small comments so far. 

>All OrleansTaskScheduler tests are passing, except few ones can randomly fail, but same fails can be observed even more frequently when running against current master.

That is very scary. Catching concurrency bugs and escaped Task context is hard. The number one threat is breaking the single threaded execution model - THE most important runtime guarantee we have.

---

Just to make sure we are all on the same page, in case someone is looking for my feedback to bless that PR, **at this point I am not expressing neither positive nor negative opinion**.
I am also not going to review that seriously before I see a design doc explaining how dataflow works and how it works here for Orleans scheduling. In particular, I am not going to decipher the implementation from the code, like I did for Cancellation Token.
**I also encourage others to take a deep look at this, so we spread the knowledge wider.**
It could also be a good opportunity for @sergeybykov to share the current scheduler design doc that I wrote back in the days. It will help others understand how the current design works.

sergeybykov:
@dVakulen 
>but same fails can be observed even more frequently when running against current master.

We are not seeing this in our tests. Could you share more details please.

Looks intriguing. Thank you for bringing up this idea!

As @gabikliot already alluded to, the scheduler is one of the most critical pieces of the whole codebase, and it has to be 100% correct. We got burned once in the past after rewriting the scheduler, and it took us a lot of effort to get it back into shape. Hence, we definitely need to tread this path very carefully and perform exhaustive code reviews and testing, and then some.

But if we convince ourselves that the code is correct and get the perf benefits, that will be awesome!

src/OrleansRuntime/Scheduler/WorkerPoolThread.cs
dVakulen: >some todo's, which will be addressed after Orleans internal performance test run, which will also show whether it worthy reviewing

As I  previously mentioned this PR needs some polishing which won't affect functionality. I wanted to see full sized performance test results first in order to see where it's going to go.

src/Orleans/Runtime/RuntimeContext.cs
dVakulen: InitializeThread can be called only once per thread.

src/OrleansRuntime/Scheduler/WorkerPoolThread.cs
dVakulen: There is: https://github.com/dotnet/orleans/pull/2060/files#diff-f21e00f434dac5a42463e7cae5b772f7R49. WorkerPoolThread doesn't being used anymore.

dVakulen:
@sergeybykov Actually I can't reproduce those fails either, neither on this PR, nor on master. They were related to task execution sequencing (e.g. Sched_Task_SubTaskExecutionSequencing, ActivationSched_SubTaskExecutionSequencing). Looking at their code I may assume that there was just problems with overload of my PC, due to which the time dedicated for tests to complete wasn't enough. 

Also I surely will add the design doc if the performance tests will show that this initiative is worth pursuing,

jdom:
Hi @dVakulen, sorry it took a little long, as we were having build failures not related to this PR. Results are back, but unfortunately the load tests failed as the runtime did not behave correctly. It's not easy to pinpoint exactly what went wrong, but if you want, I can share some of the silo logs with you so that you can take a closer look.

shayhatsor:
@dVakulen, In light of @jdom's tests conclusions, I suggest:
1) Make sure you can run the same tests locally. If not, maybe there's an option to expose these tests with Jenkins, and you can tell the test-bot to run those one demand.
2) I don't know if it's possible, but try to first replace the smallest part possible and work your way up. It will be a rigorous long process, but I can't see any other way for you to find bugs/inconsistencies. Maybe looking at the tests themselves might give you a hint, but I doubt it.

dVakulen:
Runtime misbehavior was fixed.

jdom:
I kicked off another run with the latest. The tests take 5 hours to complete, so I'll post results later.
@shayhatsor we do agree it would be nice to expose the load tests, and we actually have plans to migrate them, but that's no simple task, although we'll get there eventually.
Nevertheless, these tests would not be cheap for any contributor to run locally, as they use a cluster of 24 physical machines and several agents. Still we assume it could be valuable for people to have access to those and use as a base to load test some of their specific app scenarios.
In the meantime I shared the logs with @dVakulen privately, so I assume that's why he updated this PR with more commits.

jdom:
There are still several failures in the run after the fixes. Sharing the logs with you privately again

dVakulen:
Root cause of the failures found, working on solution.

dVakulen:
The problem was with unfair scheduling of the tasks. Fixed.

jdom:
Just did a run, and looks like tests are now green (cluster is stable). Nevertheless, I do see a *reduction* in overall throughput in the range of 6%-15%

dVakulen:
Thank you, @jdom. Can you please share the logs for this run?

dVakulen:
The cause of the reduction in throughput is reduced CPU usage - 81% on average instead of 95% in normal run. While it can be fixed - the resulting performance in load test will be on par with current master.  I'll try to find out the differences between the tests for now.

sergeybykov:
The test is designed to fully saturate the silos with a large enough pipeline of requests in flight. Unless the load generators run at their capacity (I doubt it, but it's easy to check or increase their number), the drop in CPU utilization might actually point to contentions that prevent the silos from reaching their capacity throughput.

dVakulen:
@sergeybykov Benchmark on which I run tests (https://github.com/dotnet/orleans/pull/2003) completes on this branch in 10 seconds, while on master it takes 20. So even with underutilized CPU I've expected some performance improvement.

dVakulen:
Theoretically latest commit should fix the problem.

src/OrleansRuntime/Scheduler/WorkItemGroup.cs
```diff
@@ -308,12 +309,12 @@ public void Execute()
                         }
 
                         // Check the cancellation token (means that the silo is stopping)
-                        if (thread.CancelToken.IsCancellationRequested)
-                        {
-                            log.Warn(ErrorCode.SchedulerSkipWorkCancelled, "Thread {0} is exiting work loop due to cancellation token. WorkItemGroup: {1}, Have {2} work items in the queue.",
-                                thread.ToString(), this.ToString(), WorkItemCount);
-                            break;
-                        }
+                        //if (thread.CancelToken.IsCancellationRequested)
```
ReubenBond: I'd be more comfortable if this was deleted instead of just commented out

sergeybykov:
@dVakulen 
>@sergeybykov Benchmark on which I run tests (#2003) completes on this branch in 10 seconds, while on master it takes 20. Same proportion is observed when running with multiple silos. So even with underutilized CPU I expected some performance improvement.

But the perf tests that @jdom are end-to-end distributed tests where move moving parts are involved, and silos and load generators run on different physical (8 core) machines. The picture is likely to be very different from running a single process benchmark on a desktop/laptop.

I'm not saying this is wrong. Just being cautious because distributed benchmarking is hard. We need to get to the bottom of these numbers to understand what's going on.

dVakulen:
By the way, comparing current NightlyLoadTest numbers (189000 TPS) with ones mentioned at https://github.com/dotnet/orleans/pull/1586#issuecomment-200561741 - looks like last months perfomance related PR's have made some effect.

sergeybykov:
@dVakulen 
>By the way, comparing current NightlyLoadTest numbers (189000 TPS) with ones mentioned at #1586 (comment) - looks like last months perfomance related PR's have made some effect. 

No doubt - thanks primarily to **your** optimizations. :-)

src/Orleans/Runtime/DedicatedThreadPool.cs
```diff
@@ -0,0 +1,756 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="Helios.Concurrency.DedicatedThreadPool.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+/*
+ * Copyright 2015 Roger Alsing, Aaron Stannard
+ * Helios.DedicatedThreadPool - https://github.com/helios-io/DedicatedThreadPool
```
jthelin: Are there any Orleasns-specific changes in this code?

This `DedicatedThreadPool ` class is available as a NuGet package, which I suggest is the easiest and fastest way to do this reuse.
https://www.nuget.org/packages/Helios.DedicatedThreadPool/

@sergeybykov - FWIW, the original Akka.Net code is Apache-2.0 licensed, so if we really want to copy-paste include this Apache-2 licensed code into the Orleans (MIT) code base then we probably need to get some experts to confirm everything is OK? IANAL and i don't recall what the prevailing advice is about cross-/re-licensing of external OSS code ;)
https://github.com/helios-io/DedicatedThreadPool

It would be so much easier to re-use the NuGet package! ;)

Aaronontheweb: @jthelin the NuGet package just installs the .CS file. No binaries. We wanted to make it easy for consumers to mark the class as `internal` just in case multiple libraries in the same project depend on it.

As far as licensing goes - preserving the original copyright header on the file and including the license text somewhere in the project (

jdom:
@dVakulen, shared some more logs from the latest commits with you privately

src/Orleans/Runtime/DedicatedThreadPool.cs
dVakulen: >Are there any Orleans-specific changes in this code?

Yes, I added ability to prioritize system tasks.

dVakulen:
There's indeed is point of contention., so now I'm working on solution (again :). The approach I'm currently taking is the following: one global work queue with fixed number of executors into which all stages submits work items, the queue itself implements work stealing pattern, so that transition between stages should happen without touching global queue.

dVakulen:
.

dVakulen:
@dotnet-bot test this please

gabikliot:
>one global work queue with fixed number of executors into which all stages submits work items, the queue itself implements work stealing pattern, so that transition between stages should happen without touching global queue.

And now you need to solve the question of priorities: when executor decides to steal the work from the global queue, it needs to decide which work item (of which stage) he takes first. Or broader: what weights are you assigning to how much work that belongs to each stage to steal and execute. It cannot be fifo or round robin, it has to be proportional to the number of items and their execution time per stage.  If you don't do it "number of items and their execution time" proportional, you are back at the original problem, just like described in the paper.

And then you realize you are at square zero: deciding on those prioritizes is exactly like deciding on the sizes of the individual executor thread pools in the current design. Plus, now you lost cache locality.

I wrote and explained that on Gitter when you first asked me about that. We thought a lot about that.

Not saying you cannot necessarily squeeze more perf., but yes saying that fundamentally you are going to arrive to the same solution.

veikkoeeva:
> Not saying you cannot necessarily squeeze more perf., but yes saying that fundamentally you are going to arrive to the same solution.
Nevertheless, this has been a good run thus far and still going, I see. :)

dVakulen:
Please, correct me if I'm wrong, but my vision is the following: In order to achieve minimum possible latency there's a need to either use threads reallocator described in the http://research.microsoft.com/apps/pubs/default.aspx?id=263907 or use single queue (maybe with priorities).  

The cache locality isn't there already due to context switches between the threads allocated to different stages, while in case of single queue and batch processing of items per stage - there will be less switches and better locality. Plus, latter solution even with implemented prioritizing will be much more simple than reallocator based one. Also synchronization greatly affects performance especially in case of not 100% silo load, which in fact more rare in production than 75% load. 

About complexity of work item weights assigning - assuming that each thread in current implementation gets ~same quantum of execution time, and serialization costs compares to N grain methods executions, while latter are being executed in batches of N items - the current execution order can be represented as simple FIFO. 

So I'm thinking that having sync-free work items processing will outweigh all complexity of prioritizing items in the queue.

dVakulen:
@gabikliot You also said that one could try to implement this, in case of having time and that it will be interesting to try an alternative threading architecture.

gabikliot:
Dimitry, about your last comment: yes, I said that and still strongly stand behind that. I am in no way discouraging you from trying to implement another solution. I see now that maybe my comment could have been interpreted that way. That was not my intention. Sorry if it sounded that way. Please do try another approach, by all means. 

I will reply later to you technical questions above.

gabikliot:
1) >The cache locality isn't there already
Of course it is there.
There is a small number of threads, and as long as each thread is only taking from its own queue (which is the case now), we get the each threads executes multiple items until preempted. That is not only true for networking threads, but also for activation worker threads, which executes a number of turns on each activation before taking on another activation.
So of course, of course, there is cache locality now. That was the whole point of staying with thread pool per stage. We also read a lot of papers that tried to optimize SEDA and at the end all agreed that cache locality was a number one factor.

2)  single queue - no, I don't think single queue is good. I think single queue will be bad. I am not going into details here why. I think we discussed that in the past.

>maybe with priorities
Yes, exactly. And how do you set priorities? You need to evaluate the number of items and blocking time and execution time per item per stage.

So what I said in my comment is:
fundamentally I think the solution is the same: "evaluate the number of items and blocking time and execution time per item per stage." Fundamentally, you will have to use the methodology of evaluation and mathematical solution that we developed in the paper. What do you do with the solution now?
You can now either set priorities or have separate queues.
I agree that priorities might be simpler implementation wise. 

But lets please step back. Who said we need to implement that at all? The solution with single queue and priorities may be simpler than tuning number of thread per stage, BUT:
a) it will be more complicated probably than the current code, which does not do either.
2) and who said that we absolutely need to have either of the two?

There are so much more optimizations that we can do, which can help in both static and dynamic case. If you are interested, I can tell you about a lot of ideas, some of them are more open and some are almost 100% sure will save a lot. You never asked. ;-)
I think our time will be spend much better on looking at those optimizations then implementing either of the above two.

>. Also synchronization greatly affects performance especially in case of not 100% silo load, which, in fact, more rare in production than 75% load.

Agree on that. What we should do is run a benchmark at 50% CPU util (maybe also 75%) and measure throughput there (or alternatively fix throughput  and measure CPU). We did that in the paper of course, since throughput at saturation is an OK measure but really not realistic enough measure. That is a bigger comment to Orleans perf. measurement methodology at large. I think we need both: measure at saturation and at fixed util (It can be either at fixed CPU or fixed the rate of requests and measure CPU).

sergeybykov:
@dVakulen Kudos for playing with the scheduling, questioning the current implementation, and trying other ideas. Even though quite a bit of thinking and effort went into this piece already, as @gabikliot indicated, we've always found surprises along the way, and even some mysteries. For example, we've consistently seen significantly different throughput numbers between two racks of seemingly identical servers.

I'm a little bit concerned that we are running our benchmarks on ~6 y/o machines, old OS (for continuity of benchmark numbers), and even older processors. The dedicated HW has given us relative stability of benchmarks compared to running them in the cloud. However, the conclusions we are drawing from those numbers may or may not be correct for latest generations of HW and improvements in SW. Running on Core CLR and cross platform may surprise us yet again.

I think experimentation in this area is important, and may yield significant gains even in unexpected places and with previously discarded ideas.

gabikliot:
I agree Sergey. Concretely, are you suggesting someone in the community build cloud based benchmark, or port our existing one, or we wait until core CLR port is done?

sergeybykov:
We'll likely have to port our distributed benchmarks to cloud sometime soon anyway. I am not suggesting the community should build one. However, based on the recent experience of MSR interns running similar benchmarks in Azure, the numbers there were much less stable than on our dedicated racks.

I'm merely sharing my thoughts that there is a number of variables here and potentially a number of surprises. So we need to be careful with discarding ideas based on current measurements. I understand it's hard to manage, as it's impractical to try to keep more than one or maybe a handful of alternative implementations in the code. Sounds like another potential MSR internship to me. 😄 

Ideally, I'd like to be able to run such benchmarks on a dedicated rack of modern servers with a range of OSs to compare. Core CLR will let us compare Windows to Linuxes soon.

dVakulen:
So I went and did a bit of research and here's what I've found: 

http://matt-welsh.blogspot.com/2010/07/retrospective-on-seda.html - keypoints: 
 - On a lightly or moderately loaded server the worst case context switching overhead becomes a problem and leads to poor cache behavior and greatly increased response time.

- It's better to decouple stages (i.e., code modules) from queues and thread pools (i.e., concurrency boundaries) and it is probably best to group multiple stages within a single "thread pool domain" where latency is critical.

----
https://issues.apache.org/jira/browse/CASSANDRA-10989 - Apache Cassandra's issue on moving away from SEDA to TPC (thread per processor). Relevant noted issues: 
- thread parking/unparking overheads
- extensive context switching (i-/d- caches thrashing)
- less than optimal multiple writer/multiple reader data structures for memtables, partitions, metrics, more
- large number of GC roots, longer TTSP


----
https://issues.apache.org/jira/browse/CASSANDRA-4718 
- batched dequeuing, resulted in appearance of the 
https://github.com/apache/cassandra/blob/cassandra-2.1.0/src/java/org/apache/cassandra/concurrent/SharedExecutorPool.java
 - fixed number of threads per multiple queues. 

----
https://issues.apache.org/jira/browse/CASSANDRA-1632
 - another issue related to the queues lock contention, solution again includes thread local queues. 

----
https://issues.apache.org/jira/browse/CASSANDRA-10993 
- step in attempt to replace queues completely with either finite state machines or chained futures. 

----
https://github.com/LMAX-Exchange/disruptor/blob/master/docs/Disruptor.docx 
- high performance alternative to bounded queues for exchanging data between concurrent threads. 
- Disruptor for a three-stage pipeline is 3 orders of magnitude lower than an equivalent queue-based approach.  In addition, the Disruptor handles approximately 8 times more throughput for the same configuration. 
- It also used in Apache Storm - http://www.michael-noll.com/blog/2013/06/21/understanding-storm-internal-message-buffers/

dVakulen:
So here's a list of possible performance and latency improvements:

1. Actors reassigning described in https://www.microsoft.com/en-us/research/wp-content/uploads/2016/06/eurosys16loca_camera_ready-1.pdf - up to 100% in throughput, and ~20% in latency, needed work amount  - ?. 

2. Thread Reallocation from ActOp paper - ~50% latency, work needed - ?

3. Single queue with workstealing and priorities - 0-100% in performance, depending on the case, and  ~50% in latency,  work - 8+ SP.

4. Port https://github.com/LMAX-Exchange/disruptor/ into .Net and use it with current stages, maybe with shared executors -  0-100% in performance, ~50% in latency, work needed - 13 SP, but implementation can be possibly reused in some other places. 

@sergeybykov Can you please provide approximate estimates for the ActOp related activities? 
@gabikliot Can you please tell about
 >There are so much more optimizations that we can do, which can help in both static and dynamic case.

gabikliot:
Funny, the references you posted - retrospective-on-seda.html and Cassandra (both of which we of course read and much more) - are the ones I used to draw exactly the opposite conclusion from what it seems you are implying.

The retrospective talks about not putting too much stages. Of course, obviously.

>I would only put a separate thread pool and queue in front of a group of stages that have long latency or nondeterministic runtime, such as performing disk I/O. (This approach harkens back to the original Flash event-driven server design that SEDA was inspired by.) This is essentially the design we used in the Pixie operating system.

And that is exactly what we do. We essentially have 3 stages: incoming IO, compute and ougoing IO. So all the internal compute stuff is a single thread pool, exactly like he writes, with aggressive optimized Task inlining where TPL can do it automatically. The IO are separate pools, just like he says, since indeed they have long latency or nondeterministic run time: can sometimes block. Specifically, our socket send IS BLOCKING!

Cassandra: in my view system like Cassandra could clearly be optimized better then general purpose SEDA. The main different is that Cassandra  is just one very specific and very simple workload: receive request, in memory read-write and reply. One workload. 
Orleans is a framework for an infinite number of different, maybe very varying workloads, hosting arbitrary application library code, which can be blocking. Thus it must use general purpose scheduler, which is obviously less optimized then a very specific highly optimized design for very specific workload like  Cassandra .
In fact, Cassandra  is also know for its bad throughout. This even motivated the new Scyla DB. Another point: look at that paper: https://www.usenix.org/node/179748. It shows how aggressively a particular workload can be optimized. But its a different story for a general purpose almost-OS-like scheduler.

dVakulen:
>The cache locality isn't there already
>Of course it is there.

If the queues are filled with enough quantity of items - it's indeed there: each thread processes it's own queue for quantum of time (around 3 ms) issued by OS scheduler, and during that time it can handle quite a batch of items before the context switch. 

But let's look at processing of the single request with initially empty queues - there will be at least 3 context switches (1 per stage). 
And the most common case - average load - some queues will be either empty, or contain items which processing time will be less than OS scheduler time quantum. The thread context switches cost in this case will be much more than on full saturation, plus each thread will spend some time on SpinWait inside BlockingCollection before the giving up . This actually leads to the saturation due to framework underperforming in common case. Synchronization on stage queues (which is the top contention point in the Orleans) just adds additional cost to each item processing. 
Thus currently framework runs on 95% CPU even on the median load, which confuses people, and prevent from potentially running on the reduced CPU frequencies, etc.
I don't think that anyone wants to run production servers on 95% load when there's no spike.

> our socket send IS BLOCKING!


Yes, and according to the information from stackoverflow posted few years ago blocking version is faster than non-blocking one. But if we to outsource the network layer, or change the implementation to async (in case there was a change in the numbers), this won't be a problem.


>The main different is that Cassandra is just one very specific and very simple workload: receive request, in memory read-write and reply. One workload. 

Aren't Orleans the same? Receive request -> non-blocking process -> reply. If there will be blocking processing current implementation will stall on processing stage as well.  

>In fact, Cassandra is also know for its bad throughout.  

Maybe because it's SEDA architecture? Which they are trying to rewrite now. 

Regarding Scylla DB - because sharing of information across cores requires costly locking, Scylla uses a shared-nothing model that shards all requests onto individual cores.
It runs one application thread per core, and depends on explicit message passing, not shared memory between threads. This design avoids slow, unscalable lock primitives and cache bounces.
(http://www.scylladb.com/technology/architecture/)

gabikliot:
OK. Worth trying out.

Orleans is not like Cassandra, we execute application code (grain code), Cassandra just reads/writes a key value.

dVakulen:
@dotnet-bot test netfx please

dVakulen:
The remaining failed tests are time sensitive and sometimes fails on the master branch as well. 
@ReubenBond can you please run perf tests on this?

galvesribeiro:
`DefaultCluster.Tests.TimerTests.TimerOrleansTest.TimerOrleansTest_Migration` is failing every PR. @ReubenBond  already opened an issue for that IIRC.

ReubenBond:
@dVakulen sure thing, apologies for the delay.

I've assigned myself for tracking, but I think @gabikliot is a much more suitable reviewer.

ReubenBond:
@dVakulen your run vs the nightly run which completed shortly prior. Approx 5% improvement.
image

dVakulen:
@ReubenBond Thank you! Is it possible to run the perf tests on 50% of the load? Also, could you please share the logs for that run?

ReubenBond:
@dVakulen I'm not sure how to reduce the load. Maybe @sergeybykov or @gabikliot have an idea?
I sent you the logs via twitter DM

gabikliot:
You can have less clients, or reduce the batch size. It will not be linear with the number of clients or batch size, need to experiment. Try halving the number of clients.

src/Orleans/Runtime/OrleansThreadPool.cs
```diff
@@ -0,0 +1,1476 @@
...
+        private static bool QueueUserWorkItemHelper(WaitCallback callBack, Object state)
+        {
+            bool success = false;
+
+            //
+            // If we are able to create the workitem, we need to get it in the queue without being interrupted
+            // by a ThreadAbortException.
+            //
+            try { }
+            finally
```
chris-eaton: Can I ask why this is done in a finally block? Never seen finally used this way before?

dVakulen: It's done in finally block in order to avoid ThreadAbortException - http://stackoverflow.com/a/2186123/3725200

src/OrleansRuntime/Scheduler/WorkItemGroup.cs
```diff
@@ -453,3 +467,76 @@ private void ReportWorkGroupProblem(string what, ErrorCode errorCode)
...
+        SetupGetter();
+    }
+
+    public static void SetPropertyName(string newName)
+    {
+        _propertyName = newName;
+        SetupGetter();
+    }
+
+    public static Task CurrentTask => _getter();
```
ReubenBond: Is this ever used?

dVakulen: This is not really ready for review yet. This particular code was used for ensuring that current task and task scheduler is corrent during work items executions, and was not removed as I thought that this checks possibly could be hidden behind conditional compilation variable.

ReubenBond:
The latest perf numbers are looking very good!
```
PingLoadTest_LocalReentrant
master:   109087
dVakulen: 146911 (+ 35%)

PingLoadTest_RandomReentrant_MultiSilos 
master:   253968
dVakulen: 460284 (+ 81%)

NightlyLoadTest 
master:   187853
dVakulen: 199186 (+ 6%)

ActivationCollectorStressTest 
master:   191248
dVakulen: 250911 (+ 31%)
```

dVakulen:
Many thanks to @ReubenBond and @jdom for collaboration and patience.  

#### Current status of this PR:
Now that this PR became *big pile of optimizations* the latest [performance test results](https://github.com/dotnet/orleans/pull/2060#issuecomment-297853999) does not clearly show difference that are due to changes for original PR's intention. But few runs of NightlyLoadTest were performed on old cluster without unrelated optimizations, and they showed [5% increase](https://github.com/dotnet/orleans/pull/2060#issuecomment-276227175)   on full load, and up to 20% on ~75% of load, so I'm strongly suspect that at least half of increase in latest run was due to the rework of threading model, and that on average load the difference can be much bigger.

#### Latest view on changes to threading model:

Every work item goes through single thread pool which was mainly borrowed from [.Net implementation](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/ThreadPool.cs). On different workloads usually at least 50% of items for execution are being 
taken from thread local queues. 
The only blocking IO in Orleans is happening on outgoing messages send stage, in order to avoid thread pool starvation on that sends - the number of concurrent sends was limited by using DataFlow block, which also uses shared work queue, so now behavior of that stage is similar to master branch one (if DotNetty is to be integrated into Orleans at some point - this limitation will not be needed). 


#### Continuation sense considerations:
 
This change was written as proof of concept and needs lots of refactoring (work) to be ready for review.
 So now the question is - are that performance results sufficient enough to make such change acceptable and thus whether it makes sense to continue work on it? 

cc @sergeybykov , @gabikliot

ReubenBond:
These are significant perf improvements (huge, even). There are many areas for improvement and you've tackled a good chunk of those here. I support splitting this PR out into per-optimization/per-optimization-class PRs so that we can merge them in more reviewable chunks which can be more easily measured. Does that sound like the right approach to you?

I'm sure others will want to weigh in as well.

In previous load tests (old cluster), I was only running the `NightlyLoadTest` test, but that test seems less sensitive to optimizations than all of the other tests. We only saw ~7% swings in any direction.

dVakulen:
Yes, I'm planning to create separate issues/PRs for those additional optimizations.

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -625,23 +625,61 @@ private bool ShouldLogError(Exception ex)
...
-        /// </summary>
-        /// <param name="message"></param>
-        /// <returns>Resolve when message is addressed (modifies message fields)</returns>
-        private async Task AddressMessage(Message message)
+		// this is a compatibility method for portions of the code base that don't use
+		// async/await yet, which is almost everything. there's no liability to discarding the
+		// Task returned by AsyncSendMessage()
+		internal void SendMessage(Message message, ActivationData sendingActivation = null)
+		{
+			if (TrySendMessageToLocalActivation(message))
```
gabikliot: If I read this code correctly (and correct me please if I am wrong), this code says that when we have a message destined to a grain (no activation picked yet), we first try to see if we have a local activation for that grain and if yes, send to it. Only if not, we go via Address Msg which will lookup in the directory (or local directory cache).

If I did understand it correctly, then this is a very important and critical bug.
If we do that, it means we are not guaranteeing that eventually we will converge to a single activation (in case we had 2 activations for some time). Lets say we have 2 activations (A and B) due to silo failure, then we converged in the directory and left only B but activation A was still on silo X and silo X now due to this code keeps using A and A will forever (or for a long time) stay and serve the requests from X while the rest of the silos will be using B.

That violated our guarantees to eventually converge to 1 activation.
The solution is to always rely only on the directory which is guaranteed to converge (or the directory cache, which is also guaranteed to converge since we check it with the directory).

We had that bug, about 3-4 years ago. You will not believe how costly this bug was  and how long it took us to chase it down.

dVakulen: Yes, this is most questionable additional optimization and your question was my primary concern when doing it, but it does not relates to original PR intention. 
I also saw the comment in  [Catalog's FastLookup method](https://github.com/dVakulen/orleans/blob/master/src/OrleansRuntime/Catalog/Catalog.cs#L1217). This method tries to get information from local ```ActivationDirectory``` cache and is being used by ```RandomPlacementDirector``` , ```StatelessWorkerDirector``` and ```ClientObserversPlacementDirector``` on activation selection path; that's why I decided to use it directly instead of going through multiple awaits in PlacementDirectors logic.

gabikliot:
A good writeup on SEDA:
http://www.scylladb.com/2017/07/06/scyllas-approach-improve-performance-cpu-bound-workloads/

Need to be careful about terminology: the author compares SEDA with NO SEDA and shows how the latter achieves better results. I would call them both SEDA, just the 1st one is "uncontrolled SEDA" (unlimited number of threads and stages) vs. the latter one has a small number of stages (to control latency, as he explains) plus Thread per Core.

Just like Orleans. And what he calls "need fine tuning" at the conclusion is what the paper http://research.microsoft.com/apps/pubs/default.aspx?id=263907 was all about - systematic approach to semi-automatic fine tuning.

dVakulen:
Indeed good article. 
The stages in ScyllaDb is -not quite like Orleans - they are using shared threadpool, while in Orleans each stage owns separate pool. Dataflow blocks is perfect fit for being a synchronous stage - they do the batching and using shared threadpool, but they can not be used for async ones when high degree of concurrency is needed. And as Orleans deals with custom user code - the benefits from batching can be lesser than in ScyllaDb, while L1\L2 cache hits - more important. 
Anyway there's possibility to implement switchable executor services, so current implementation could be supported alogside with new one, and then the performance differences could be easily measured and optimal one choosen for particular workload.

dVakulen:
This PR after series of experiments ended up in unreadable state. The changes will be covered by series of follow-up PR's.

PR #2063, opened 2016-08-15
Add Memory Stream Provider

JingqiaoFu:
Add Memory Stream Provider, Would you mind squashing the commits for me? Thanks.


gabikliot:
Is it mimicking Azure Queue provider or Event Hub? Grain per queue?

src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Orleans.Providers.Streams.Memory
+{
+    public interface IMemoryStreamQueueGrain : IGrainWithGuidKey
+    {
+        Task SetMaxEventCount(int maxEventCount);
```
gabikliot: who calls that?

src/OrleansProviders/Streams/Memory/MemoryStreamProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Orleans.Providers.Streams.Common;
+
+namespace Orleans.Providers.Streams.Memory
+{
+    /// <summary>
+    /// This is a persistent stream provider that uses in-memory grain to queue the events.
+    /// This is primarily for test purposes.
```
gabikliot: Would not an easier and actually better approach is to build an in-memory emulation of AzureQueueManager. It will expose the same set of low level APIs as AzureQueueManager. It will require significantly less new code, as now you are mostly duplicating the other providers.
If you want to model it after EH, you could do the same - in memory EH emulation.
Then you could just slightly refactor the existing EH or AQ to inject the queue manager.

src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
JingqiaoFu: This is supposed to be a helper function for users that limits the size of the queue. It doesn't have to be called during the provider's life cycle.

gabikliot:
I am not sure you if you meant that you addressed my feedback. It does not appear so. In particular, I mean my main feedback about a different design that will build an in memory emulation of AQ or EH, and not the full new adapter.

src/OrleansProviders/Streams/Memory/MemoryStreamProvider.cs
jason-bragg: This provider is mostly modeled after EventHub.  One of the reasons for introducing it was to allow developers to develop against a persistent recoverable stream provider, like EH, without having to have a Hub per developer.  It is not at all unreasonable to have an in-memory mode supported by each persistent stream provider rather than a separate in-memory provider.  The original EventHub provider worked much like you describe, it could write to EventHub partitions or in-memory queues.  That pattern turned out to be more difficult than expected.  The current EventHub stream provider is the most complex provider we have so we're trying not to add any more complexity to it until we complete it's feature set.

Once the EventHub stream provider is complete and more stable, I'd be very open to moving to that model, but to avoid introducing further complexity into the EventHub stream provider we decided to introduce this capability via a new provider.

JingqiaoFu:
Gaby: Sorry about the confusion, the fixes I made was for the internal codeflow feedback. I will talk with Jason regarding your feedback.

gabikliot:
I did not realize @JingqiaoFu is on the core team, I thought he needs my feedback. 
No problem, scratch out my feedback.

src/OrleansProviders/Streams/Memory/MemoryAdapterFactory.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        private Guid GenerateDeterministicGuid(QueueId queueId)
+        {
+            // provider name hash code
+            JenkinsHash jenkinsHash = JenkinsHash.Factory.GetHashGenerator();
+            int providerNameGuidHash = (int)jenkinsHash.ComputeHash(providerName);
+
+            // get queueId hash code
+            uint queueIdHash = queueId.GetUniformHashCode();
```
jdom: in order to trust this uniform hash code, please remove the overloads for `QueueId.GetQueueId` that do not take an explicit hash. The currently have no usages, so it's an easy removal, but will prevent naive usages in the future.

src/OrleansProviders/Streams/Memory/MemoryAdapterReceiver.cs
```diff
@@ -0,0 +1,68 @@
...
+            try
+            {
+                task = queueGrain.Dequeue(maxCount);
+                awaitingTasks.Add(task);
+                eventData = await task;
+                batches = eventData.Select(data => (IBatchContainer) MemoryBatchContainer.FromMemoryEventData<object>(data, ++sequenceId)).ToList();
+            }
+            catch (Exception exc)
+            {
+                throw;
```
jdom: if this catch is just throwing, why are you catching in the first place? Didn't you forget a log call here?

JingqiaoFu: Yes, there should be an error logging call. Will add that.

PR #2065, opened 2016-08-16
Persistent Stream Provider initialization timeout fix.

jason-bragg:
Persistent stream provider can timeout during initialization if many queues are configured.  To address this, we don't await receiver initialization any more.  Receiver initialization was always allowed to fail, in which case re-initialization was necessary.  Now we no longer await it since we don't handle the exception anyway.

### DIFF
--- a/src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
+++ b/src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Orleans.Concurrency;
 using Orleans.Runtime;
@@ -14,7 +15,6 @@ namespace Orleans.Streams
         private const int StreamInactivityCheckFrequency = 10;
 
         private readonly string streamProviderName;
-        private readonly IStreamProviderRuntime providerRuntime;
         private readonly IStreamPubSub pubSub;
         private readonly Dictionary<StreamId, StreamConsumerCollection> pubSubCache;
         private readonly SafeRandom safeRandom;
@@ -32,8 +32,9 @@ namespace Orleans.Streams
         private IDisposable timer;
 
         internal readonly QueueId QueueId;
-        private bool IsShutdown { get { return timer == null; } }
-        private string StatisticUniquePostfix { get { return streamProviderName + "." + QueueId; } }
+        private Task recieverInitTask;
+        private bool IsShutdown => timer == null;
+        private string StatisticUniquePostfix => streamProviderName + "." + QueueId;
 
         internal PersistentStreamPullingAgent(
             GrainId id, 
@@ -49,14 +50,13 @@ namespace Orleans.Streams
 
             QueueId = queueId;
             streamProviderName = strProviderName;
-            providerRuntime = runtime;
             pubSub = streamPubSub;
             pubSubCache = new Dictionary<StreamId, StreamConsumerCollection>();
             safeRandom = new SafeRandom();
             this.config = config;
             numMessages = 0;
 
-            logger = providerRuntime.GetLogger(GrainId + "-" + streamProviderName);
+            logger = runtime.GetLogger(GrainId + "-" + streamProviderName);
             logger.Info(ErrorCode.PersistentStreamPullingAgent_01, 
                 "Created {0} {1} for Stream Provider {2} on silo {3} for Queue {4}.",
                 GetType().Name, GrainId.ToDetailedString(), streamProviderName, Silo, QueueId.ToStringWithHashCode());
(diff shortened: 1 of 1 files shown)


src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -131,27 +131,33 @@ internal class PersistentStreamPullingAgent : SystemTarget, IPersistentStreamPul
...
+            return TaskDone.Done;
         }
 
         public async Task Shutdown()
         {
             // Stop pulling from queues that are not in my range anymore.
             logger.Info(ErrorCode.PersistentStreamPullingAgent_05, "Shutdown of {0} responsible for queue: {1}", GetType().Name, QueueId.ToStringWithHashCode());
-            if (timer != null)
+            Task localrecieverInitTask;
+            if ((localrecieverInitTask = Interlocked.Exchange(ref recieverInitTask, null)) != null)
```
gabikliot: Why do you need interlock?

jason-bragg: Don't strictly need it,  it's just a safer pattern, but in retrospect, I see how it could give the impression that this code needs to be thread safe, when it doesn't.  It just needs to be 'Orleans turn' safe.

Will return to previous null check + set null pattern.

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -81,7 +80,7 @@ internal class PersistentStreamPullingAgent : SystemTarget, IPersistentStreamPul
         /// <param name="queueAdapterCache"></param>
         /// <param name="failureHandler"></param>
         /// <returns></returns>
-        public async Task Initialize(Immutable<IQueueAdapter> qAdapter, Immutable<IQueueAdapterCache> queueAdapterCache, Immutable<IStreamFailureHandler> failureHandler)
+        public Task Initialize(Immutable<IQueueAdapter> qAdapter, Immutable<IQueueAdapterCache> queueAdapterCache, Immutable<IStreamFailureHandler> failureHandler)
```
jdom: Since this method can throw exceptions that are not usage errors (according to the [TAP guidelines#Exceptions](https://msdn.microsoft.com/en-us/library/hh873175(v=vs.110).aspx#Anchor_2) ), it should probably use the `OrleansTaskExtentions.WrapInTask` utility method so that exceptions result in faulted tasks instead of failing to produce a task.

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -81,17 +80,21 @@ internal class PersistentStreamPullingAgent : SystemTarget, IPersistentStreamPul
...
         {
             if (qAdapter.Value == null) throw new ArgumentNullException("qAdapter", "Init: queueAdapter should not be null");
             if (failureHandler.Value == null) throw new ArgumentNullException("failureHandler", "Init: streamDeliveryFailureHandler should not be null");
+            return OrleansTaskExtentions.WrapInTask(() => InitializeInternal(qAdapter.Value, queueAdapterCache.Value, failureHandler.Value));
+        }
 
+        public void InitializeInternal(IQueueAdapter qAdapter, IQueueAdapterCache queueAdapterCache, IStreamFailureHandler failureHandler)
```
jason-bragg: Meant to make this private :/  Will fix.

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -307,11 +325,18 @@ public void RemoveSubscriber_Impl(GuidId subscriptionId, StreamId streamId)
         {
             try
             {
+                Task localrecieverInitTask = recieverInitTask;
```
jdom: nit: typo in several variables and the new field

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -131,27 +134,42 @@ internal class PersistentStreamPullingAgent : SystemTarget, IPersistentStreamPul
...
                 Utils.SafeExecute(tmp.Dispose);
             }
 
+            Task localRecieverInitTask = recieverInitTask;
+            recieverInitTask = null;
+            if (localRecieverInitTask != null)
+            { 
+                try
+                {
+                    await localRecieverInitTask;
```
jdom: could this interfere with our attempt to fail fast due to a (different) silo failure? Maybe we should use `WithTimeout`? I'm not sure what's the right approach

jason-bragg: If we need a timeout for fail fast, we should do it at the top level call, not a each external call.

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
jason-bragg: "i" after "e" except aft....    I hate the english language.

jdom:
I do realize that the neither current nor the new implementation wait to perform some basic initialization checks in order to fail the initialization permanently.
Ideally at some point I'd like to perform the check during initialization for whether the account used or it's credentials are completely invalid (as opposed to some temporary failure to access the external resource), and in that case fail to initialize completely, instead of failing almost silently (other than some messages in the log).

This is something for a different discussion and different PR though, so I'm merging this, as the change looks good and an improvement for what we had.

Issue #2056, opened 2016-08-14
Remove strings from InBuiltTypes under serialization


Drawaes:
There are a large number of method names that are currently linked by passing "MethodName". They should all be changed to nameof(MethodName) to allow compile time checking. I am happy to do the work if you think it is okay to change to this? I see no performance or functionality impact.

jason-bragg:
I see no issue with this.. anyone else?

veikkoeeva:
Feels like a sane move.

jdom:
Resolved by #2067

Issue #1464, opened 2016-02-16
Using ImmutableCollections with Orleans


Eldar1205:
1. I've read about Orleans [Immutable] attribute and Immutable<T> wrapper to tell the framework not do deep-copy some object.
2. I've read in the technical report that deep copying collections is where the performance hit is at.
3. I've read some of Orleans internal code and there are several places I saw List<T> being copied inside a lock to return a copy of an internal list.

All things above can be addressed better by having support for .Net's new Immutable collections (https://msdn.microsoft.com/en-us/library/system.collections.immutable(v=vs.111).aspx) . It's an external dependency but I believe a good one to have, encouraged to use by .Net as a framework.

1. Does Orleans support application code using Immutable collections in messaging and avoid deep-coping them and have efficient serialization for them?
2. Do you think Orleans should use Immutable collections internally instead of synchronizing lists access and copy lists to achieve immutability?

jthelin:
The [Immutable] functionality in Orleans was added long time ago before .NET added their own support for immutability. Now it is "legacy" code ;)

I think it would make sense to add support for System.Collections.Immutable now, and move to mark the Orleans custom [Immutable] support as [Obsolete] / deprecated for eventual removal in a future release.

Eldar1205:
The [Immutable] and Immutable<> could still have their usages for non-collection instances. I think about silo local pure functional operations - they shouldn't suffer from any copying cost. On the other hand, a simple to implement algorithm can detect a class has only readonly fields using reflection (https://msdn.microsoft.com/en-us/library/system.reflection.fieldinfo.isinitonly.aspx) or even better detect this at compile time using Roslyn which I'm only familiar with by name

sergeybykov:
I agree with @Eldar1205 - `Immutable` is a flexible convention to communicate the intent rather than implementation, which has value beyond the set of concrete immutable classes in `System.Collections.Immutable`. I think we need both - automatically recognize types from `System.Collections.Immutable` as immutable and do not deep-copy them, and still support `[Immutable]/Immutable<t>` for arbitrary types and arguments.

I am a bit cautious about jumping and using those types internally though. We need to benchmark such changes, at least those on the hot paths. We've seen with `ConcurrentDictionary` that it's not free performance wise. I wonder what's the price of the immutable collections. Allocations?

gabikliot:
[...]

PR #2067, opened 2016-08-16
#2056 #1464 Adding Immutable Collections and BuiltInTypes Cleanup

Drawaes:
I have added support for Immutable Collections, it assumes that the content of the collection is immutable as well. I think that fits in with the Immutable<> wrapper semantics. I provided hand coded serialization and deserialization as well to avoid the default binary serializer.

I have also removed the hardcoded strings from the BuiltInTypes class and replaced them with build time checking via the "nameof" command.

### DIFF
--- a/src/Orleans/project.json
+++ b/src/Orleans/project.json
@@ -1,6 +1,7 @@
-{
+﻿{
   "dependencies": {
-    "Newtonsoft.Json": "7.0.1"
+    "Newtonsoft.Json": "7.0.1",
+    "System.Collections.Immutable": "1.1.36"
   },
   "frameworks": {
     "net451": {}
--- a/src/NuGet/Microsoft.Orleans.Core.nuspec
+++ b/src/NuGet/Microsoft.Orleans.Core.nuspec
@@ -19,6 +19,7 @@
     <tags>Orleans Cloud-Computing Actor-Model Actors Distributed-Systems C# .NET</tags>
     <dependencies>
       <dependency id="Newtonsoft.Json" version="7.0.1" />
+      <dependency id="System.Collections.Immutable" version="1.1.36" />
     </dependencies>
   </metadata>
   <files>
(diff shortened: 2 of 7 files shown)


src/Orleans/project.json
```diff
@@ -1,6 +1,7 @@
-{
+﻿{
   "dependencies": {
-    "Newtonsoft.Json": "7.0.1"
+    "Newtonsoft.Json": "7.0.1",
+    "System.Collections.Immutable": "1.1.36"
```
jdom: Hmmm, I'm a little hesitant to add this dependency here, as every grain project will be coupled to this too, and there are many ways to get into conflicts, especially since Roslyn uses it too, which runs on many different environments.
Not pushing back on this yet, but can we find another option to avoid this extra dependency in the core?
IMO, Newtonsoft.Json should probably be removed from the core too, but before adding more just for the sake of serialization, let's see if we can think of alternatives. Otherwise by the same token we probably should be adding F# types that are immutable or several others that we know how to serialize.

Drawaes: I tend to have the same reservations here. I am not sure what the best solution is for this? Maybe add a new project which is a DLL to put extended serialization logic in? This would probably allow the Json.Net to be moved out as well?

Drawaes:
Did you dig out the performance testing you did with Immutables on the hot paths. I would like to take a look at it if I could. It might be that recent improvements make them a better option so it might be worth the dependency then.

src/Orleans/project.json
jdom: I think that we could take this dependency for now. We already have some talks of how we might refactor dependencies in the future, so this might be a non-issue then, but for now let's add it, as it is clearly a valuable contribution.
Make sure to add this dependency in `Microsoft.Orleans.Core.nuspec` too

Drawaes:
I backed out the accidental check in for the locking removal I did (unless you wanted that too) and put in the dependency on the nuspec file. I will raise an issue for the lock removal and discuss that before submitting a PR.

jdom:
Thanks for this contribution @Drawaes!

Issue #1059, opened 2015-11-22
Support deep Service Fabric integration


ReubenBond:
[Service Fabric](https://azure.microsoft.com/en-us/services/service-fabric/) is a great platform for application hosting which offers high-density hosting, application upgrade, and state replication.

Many users, such as myself, would rather use Orleans than Service Fabric Reliable Actors. There are many reasons for this, including that Orleans is much more mature, open source, and feature rich.

We have [integration packages](https://www.nuget.org/packages/Microsoft.Orleans.ServiceFabric) for hosting Orleans atop Service Fabric. This integration is simple and effective but does not allow Orleans to gain all of the benefits which Service Fabric has to offer, such as collocated state.

Orleans should be the obvious choice for Virtual Actors in any hosting environment, so there is some additional work to be done which I would like insight into and assistance with.

This issue is for discussion and tracking of that work.
### Basic Integration:
- [x] **Hosting**: Basic stateless, unpartitioned service hosting. Use the existing [NuGet packages](https://www.nuget.org/packages/Microsoft.Orleans.ServiceFabric).
### Integration Required for Collocated State:
- [x] **Gateway Provider**: Implement an `IGatewayListProvider` based on Service Fabric's `NamingService`. Added in #2542
- [x] **Cluster Membership Provider**: Silos need to be discoverable by clients & each other. ~~This should involve creating an `IMembershipTable` implementation which leverages Fabric's `NamingService` or `PropertyService`.~~ This involves creating an `IMembershipOracle` implementation which can receive updates directly from Service Fabric in addition to periodically polling partitions. Added in #2542
- [ ] **Grain Placement**: Orleans should be partitioned and grain placement should deterministically map `GrainId` to partitions. Using consistent hashing means that we do not need to hold/maintain a grain directory. One potential downside to this approach is that we rely on the randomness of GrainIds to balance load across hosts and cannot perform load shedding in a granular fashion. This is mitigated by using a significant number of partitions, allowing Service Fabric to perform this load balancing at the partition level by shifting partitions between nodes based. This change requires only that we configure a new `PlacementStrategy`/`PlacementDirector` which resolves `GrainId` to `SiloAddress` based upon Service Fabric partitions. We can optionally eliminate the Grain Directory when we are using Service Fabric, but keeping the Grain Directory around allows us to use different placement strategies (as long as they are stateless or use a state provider which doesn't require consistent mapping between `GrainId` & Fabric partition.)
[...]

PR #2070, opened 2016-08-17
[WIP] Service Fabric support

ReubenBond:
This is an implementation of deeper Service Fabric support (#1059).
It's not ready for use, so please don't use it. I'm pushing an early revision so that people can see what this might look like. It will change significantly before it's merged and may be split into multiple PRs.

* `OrleansCommunicationListener` + a helper class, `OrleansServiceListener`: Allows users to host a silo in a Stateful or Stateless Reliable Service (Service Fabric parlance). Silos can co-exist with Web APIs, WCF services, etc. The silo will publish an endpoint called "Orleans" which is used when locating the service.
* `OrleansServiceFabricExtensions`: Helper class which adds an `.AddServiceFabricSupport()` extension method to `IServiceCollection` so that the relevant services can be made available via dependency injection.
* `ReliableDictionaryStateProvider`: an `IStorageProvider` which replicates state from the local primary to remote secondary replicas.
* (internal) `ServiceFabricNamingServiceGatewayProvider`: `IMembershipTable` & `IGatewayListProvider` implementation which uses SF's internal strong-consistency Naming Service as the store, eliminating the need for Azure Tables. SF's Naming Service is similar to Zookeeper.
* (internal) `ServiceFabricPartitionPlacementDirector`: Placement director which maps grains to SF partitions (and hence silos) based on ID. This allows state state
* (internal) `PartitionResolver`: Helper class which maps SF partitions to Orleans silo endpoints. Used by the placement provider.
* (internal) `OrleansFabricSilo`: Internal silo management class which hosts the Orleans silo instance and manages its lifecycle in accordance with SF's lifecycle notifications (via `OrleansCommunicationListener`.

I made changes to DI which will not be merged - I am using a delegate to configure DI, but it looks as though we will mirror ASP.NET's system instead. The projects in the Samples folder will be placed in their own solutions (and will be moved around a bit).

Before this is ready to be merged, it is crucial to ensure that the placement director works as advertised and that grains can never migrate to another Service Fabric partition. SF partition counts/names are fixed once deployed, but a single executable can host multiple partitions and SF can migrate partitions between nodes during failover and for load balancing. With Orleans currently having many static classes and being unable to host multiple silos in the same AppDomain, this PR cannot guarantee that a grain is actually on the primary for its partition.  This also lacks tests.


ReubenBond:
If I make a `feature-service-fabric` branch, then will automated tests still run? If so, I will do that so that we can develop this there.

sergeybykov:
I don't believe CI will run for a branch other than `master`. I don't know if that's something that's easy to configure somewhere in Jenkins' config.

src/OrleansServiceFabricUtils/ServiceFabricNamingServiceGatewayProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        private async Task<MembershipTableData> ReadEntries(SiloAddress siloAddress = null)
+        {
+            ReadResults result;
+            do
+            {
+                // Continue attempting to read from the table until a consistent read is made.
+                result = await this.TryReadEntries(siloAddress);
+            }
+            while (!result.IsConsistent);
```
gabikliot: some read may be inconsistent? really? how? I thought the bread and butter of SF and its primary pride was strong consistency.

gabikliot:
>this PR cannot guarantee that a grain is actually on the primary for its partition.

Why?

src/OrleansServiceFabricUtils/ServiceFabricNamingServiceGatewayProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+namespace Microsoft.Orleans.ServiceFabric
+{
+    using System.Fabric;
+    using System.Globalization;
+
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Converters;
+    using Newtonsoft.Json.Linq;
+
+    public class ServiceFabricNamingServiceGatewayProvider : IMembershipTable, IGatewayListProvider
```
gabikliot: The name confused me at first. Its both Table and GW Provider, not just GW provider.

src/OrleansServiceFabricUtils/ReliableDictionaryStateProvider.cs
```diff
@@ -0,0 +1,145 @@
...
+            return Task.FromResult(0);
+        }
+
+        public Logger Log { get; private set; }
+
+        public async Task ReadStateAsync(string grainType, GrainReference grainReference, IGrainState grainState)
+        {
+            var store = await this.GetStore();
+            ConditionalValue<GrainState> result;
+            using (var tx = this.stateManager.CreateTransaction())
```
gabikliot: why do you need a transaction around it?

src/OrleansServiceFabricUtils/ServiceFabricNamingServiceGatewayProvider.cs
ReubenBond: It looks like the Naming Service in particular allows for torn reads. It notifies you when they occur, at least. We will retry until we read a consistent view of the entire table.

gabikliot: I am sure you know that already, but you could have also integrated differently: instead of using NS as a key value store for MBR table, use NS as a full blown membership provider: listen to its notification about silo liveness, read all silos from it, use the SF internal membership basically and fully replace Orleans' MBR. Deeper integration sort of.
That's what I did 3-4 years ago.

ReubenBond:
@gabikliot In its current iteration if there are multiple partition on one node, then it will try to start multiple silos in the same AppDomain, which I believe will result in a failure. My original comment was wrong - I think it will fail to initialize rather than initialize but behave incorrectly.

My original (flawed) thinking was this:
* Multiple SF partitions on one node (overprovisioning is common in SF and regardless, failovers can result in multiple primaries for different partitions on one node)
* Single process per node, regardless of how many partitions. This is SF behavior, maybe it can be changed.
* This PR maps GrainId -> PartitionId -> SiloAddress
* SF can migrate a partition at runtime (eg, failover, load balancing).
* Given a silo could be the primary for multiple partitions, a mapping from GrainId -> PartitionId -> SiloAddress could still map to an active primary (for the wrong partition) even though the intended partition has changed its primary.

In short: need to ensure that a SiloAddress is only ever mapped to one partition. Also need to ensure that if a silo's generation doesn't match the generation specified by the sender, then it rejects the message. Maybe those conditions are already true.

src/OrleansServiceFabricUtils/ServiceFabricNamingServiceGatewayProvider.cs
ReubenBond: Maybe I should rename this to `ServiceFabricNamingServiceClusterMembershipProvider`. The Zookeeper provider also serves the same dual purpose of being a membership table & gateway list provider.

src/OrleansServiceFabricUtils/ReliableDictionaryStateProvider.cs
ReubenBond: `IReliableDictionary`'s methods require a transaction.

src/OrleansServiceFabricUtils/ServiceFabricNamingServiceGatewayProvider.cs
ReubenBond: I may add a version which does what you did, but for now I wanted to adhere to Orleans' membership protocol as tightly as possible.

gabikliot:
>Given a silo could be the primary for multiple partitions, a mapping from GrainId -> PartitionId -> SiloAddress could still map to an active primary (for the wrong partition) even though the intended partition has changed its primary.

gabikliot:
> Given a silo could be the primary for multiple partitions, a mapping from GrainId -> PartitionId -> SiloAddress could still map to an active primary (for the wrong partition) even though the intended partition has changed its primary.

I am  not sure I understand that.

>need to ensure that a SiloAddress is only ever mapped to one partition

Why?

I thought we map grains ids to partitions, and SF maps partitions to silo (service instances) based on latest membership it mainatins and SF guarantees that any partition is mapped to at most single service instance at a given time. That is the strong consistency property that I thought it provides.
What am I missing?

gabikliot:
Found it: https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/Messaging/IncomingMessageAcceptor.cs#L439
We simply drop msg if they are for my silo ip:port but different gen number. super important as part of ensuring "enforcing perfect failure detection".

`SA.Matches():` Gen id is zero for msgs from the client, since client, by design, does not stick to a particular silo but rather to any GW.  That specific design point about client-gw is not actually too fundamental, can be changed, its more about separation of concerns.

jdom:
I can very easily configure Jenkins to build a branch other than `master` if that's what we want here

ReubenBond:
@jdom that would be great. I'd like to wait for the new DI config method before merging this into its own branch, so I'll leave it for now

ReubenBond:
I plan on separating this out into smaller PRs so that we can merge the functionality present in OrleansContrib/OrleansFabricSilo sooner as we incrementally add deeper Service Fabric support.

Kwal:
@ReubenBond do you have any thoughts as to which piece you're going to create a PR for first?

My vote would be for the `IMembershipTable` and `IGatewayListProvider` implementation(s) - I think that'd be a huge win for running in SF and it seems like that piece is fairly straightforward (based on your existing code).

ReubenBond:
@Kwal that's the plan :) 
#2120 first, then membership/gateway, then placement, storage, etc

PR #2074, opened 2016-08-17
Import activation tests

benjaminpetit:
Import activation tests that were not yet migrated to GitHub, with some modifications:
- use `TestCluster`
- switch to Xunit


jdom:
LGTM

PR #2076, opened 2016-08-18
Include test adapter as a packages.config reference

jdom:
Looks like VSO builds do not find the xUnit VS test adapter if it's using a nuget v3 reference (as it searches for adapter inside the packages folder, which is no longer used with project.json).


jdom:
@dotnet-bot test this please

PR #2078, opened 2016-08-18
Persistent stream pulling agent now uses exponential backoff

jason-bragg:
Persistent stream pulling agent now uses exponential backoff when encountering errors when reading from queues.
This prevents agents from spamming backend queue services that are encountering transient errors.
If no data is read, usual retry timer is applied.  Exponential backoff is only triggered when errors occur.

### DIFF
--- a/src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
+++ b/src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
@@ -9,7 +9,8 @@ namespace Orleans.Streams
 {
     internal class PersistentStreamPullingAgent : SystemTarget, IPersistentStreamPullingAgent
     {
-        private static readonly IBackoffProvider DefaultBackoffProvider = new ExponentialBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(30), TimeSpan.FromSeconds(1));
+        private static readonly IBackoffProvider DeliveryBackoffProvider = new ExponentialBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(30), TimeSpan.FromSeconds(1));
+        private static readonly IBackoffProvider ReadLoopBackoff = new ExponentialBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(20), TimeSpan.FromSeconds(1));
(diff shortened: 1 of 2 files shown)


src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -332,80 +334,113 @@ public void RemoveSubscriber_Impl(GuidId subscriptionId, StreamId streamId)
...
-                    {
-                        lastTimeCleanedPubSubCache = now;
-                        CleanupPubSubCache(now);
-                    }
+                    // If read succeeds and there is more data, we continue reading.
+                    // If read succeeds and there is no more data, we breack out of loop
+                    // If read fails, we try again, with backoff policy.
+                    //    This prevents spamming backend queue which may be encountering transient errors.
+                    //    We retry until the operation succeeds or we are shutdown.
+                    bool moreData = await AsyncExecutorWithRetries.ExecuteWithRetries(
```
gabikliot: I there is an overload that accepts success callback, so it will decide on next iteration based on return value. That will save you moreData  check.

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -332,80 +334,113 @@ public void RemoveSubscriber_Impl(GuidId subscriptionId, StreamId streamId)
...
+        /// <param name="rcvr"></param>
+        /// <param name="maxCacheAddCount"></param>
+        /// <returns></returns>
+        private async Task<bool> ReadFromQueue(QueueId myQueueId, IQueueAdapterReceiver rcvr, int maxCacheAddCount)
+        {
+            try
+            {
+                var now = DateTime.UtcNow;
+                // Try to cleanup the pubsub cache at the cadence of 10 times in the configurable StreamInactivityPeriod.
+                if ((now - lastTimeCleanedPubSubCache) >= config.StreamInactivityPeriod.Divide(StreamInactivityCheckFrequency))
```
gabikliot: the cache cleanup code logically does not belong inside the func, but rather outside the loop, or even better in a separate timer. you can improve it later on.

gabikliot:
LGTM.

Issue #2081, opened 2016-08-20
Performance of Serialization Manager and removal of locks for reading


Drawaes:
As I understand it the Serialization Manager will mostly load the type -> Copier/Serializers and Desers at startup time. After that there should be minimal registration. It seems that this is a "hot path" for serialization and looking up the type to find the copier or serializer will be happening a lot. 

There is a lock around every read and while the operation in the lock is normally very quick it is a point of contention. I realize the penalties of immutable dictionaries and concurrent dictionaries have been discussed. I would propose instead the locks are removed for reading, and for writing a Clone and Swap with a lock approach is taken. This might increase the amount of GC, but it should only be up until a silo has seen all types it will need to serialize. 

I have done some testing and with 1 thread doing serialization on my crude test there is basically no difference (expected) but with two and three threads the difference ~17% for 4 threads and 6% for 2 threads (4 core Xeon laptop with 8 hyper threads). The test was just using the RoundTripSerializationForTesting method.

I am happy to make a PR if it is of interest.

ReubenBond:
Fixed - the new serializer is very fast

PR #2082, opened 2016-08-20
Changed the SerializationManager to remove the read locks. #2081

Drawaes:
I changed the locks to only be on the write, using a clone of the dictionary and a swap. This improved performance in a tight loop of ~15% on a 4 thread scenario.


src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -99,9 +99,11 @@ public static object GetUninitializedObjectWithFormatterServices(Type type)
         private static List<IExternalSerializer> externalSerializers;
         private static ConcurrentDictionary<Type, IExternalSerializer> typeToExternalSerializerDictionary;
         private static Dictionary<string, Type> types;
-        private static Dictionary<RuntimeTypeHandle, DeepCopier> copiers;
-        private static Dictionary<RuntimeTypeHandle, Serializer> serializers;
-        private static Dictionary<RuntimeTypeHandle, Deserializer> deserializers;
+        volatile private static Dictionary<RuntimeTypeHandle, DeepCopier> copiers;
+        volatile private static Dictionary<RuntimeTypeHandle, Serializer> serializers;
+        volatile private static Dictionary<RuntimeTypeHandle, Deserializer> deserializers;
```
jdom: Consider using IReadonlyDictionary instead. This way since users can't mutate it by mistake unless they keep the concrete instance around (or downcast), both of which we can avoid.
Also, is the volatile keyword really necessary here? I think you should already be covered with your usage of locks for the clone and swap (lock already creates a memory barrier that forces the value to not be read from the cache).

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -390,31 +395,79 @@ public static void Register(Type t, DeepCopier cop, Serializer ser, Deserializer
                 {
                     if (cop != null)
                     {
-                        lock (copiers)
+                        if (_setupDone)
                         {
-                            DeepCopier current;
-                            if (forceOverride || !copiers.TryGetValue(t.TypeHandle, out current) || (current == null))
+                            lock (copiers)
```
jdom: what is the purpose of the double locking (on `registeredTypes` and then individually on `copiers`)? Is some other method modifying the collections individually?

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -237,6 +239,9 @@ internal static void RegisterBuiltInSerializers()
             externalSerializers = new List<IExternalSerializer>();
             typeToExternalSerializerDictionary = new ConcurrentDictionary<Type, IExternalSerializer>();
             types = new Dictionary<string, Type>();
+            //copiers = new Dictionary<RuntimeTypeHandle, DeepCopier>();
+            //serializers = new Dictionary<RuntimeTypeHandle, Serializer>();
+            //deserializers = new Dictionary<RuntimeTypeHandle, Deserializer>();
```
jdom: commented code is duplicate. Please remove

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -390,31 +395,79 @@ public static void Register(Type t, DeepCopier cop, Serializer ser, Deserializer
...
+                            {
+                                Deserializer currentDeser;
+                                if (forceOverride || !deserializers.TryGetValue(t.TypeHandle, out currentDeser) || (currentDeser == null))
+                                {
+                                    var clone = deserializers.ToDictionary(kv => kv.Key, kv => kv.Value);
+                                    clone[t.TypeHandle] = deser;
+                                    //Put in a memory barrier to ensure the dictionary is fully updated before we switch it to the new location
+                                    System.Threading.Thread.MemoryBarrier();
+                                    deserializers = clone;
+                                }
```
jdom: just create a generic helper method to add an item to a dictionary, and conditionally do a set of a clone and swap. Otherwise you are duplicating all of the logic for each of the collections

src/Orleans/Serialization/SerializationManager.cs
jdom: I believe the double locking was previously being used to avoid issues when reading, but since you removed the locks when reading (as the dictionaries are now effectively immutable) you probably don't need the seconds layer and more granular locking.
Please verify, as you are looking at the code more closely at this point.

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -2307,8 +2382,7 @@ private static void RegisterSerializationProviders(List<TypeInfo> providerTypes)
             externalSerializers.Clear();
             typeToExternalSerializerDictionary.Clear();
             providerTypes.ForEach(
-                typeInfo =>
-                {
+                typeInfo => {
```
jdom: superfluous (and incorrect according to our guidelines) change

Drawaes:
I agree with the helper method. The volatile I will go through it again I think I might be able to remove the volatiles. I left the double locking that was there. But I can rationalise that as well

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -527,18 +622,15 @@ public static void Register(Type type, Type serializerType)
                 {
                     Register(
                         type,
-                        obj =>
-                        {
+                        obj => {
```
jdom: several of these are superfluous but also incorrect according to our guidelines. It's better to keep the diff small, but especially if the changes conflict with our guidelines (as sometimes fixing some styling errors is nice if they don't take the focus of the PR)

jdom:
Great work. I've been willing to remove these unnecessary locks for some time. Other done the few comments, looks good!

src/Orleans/Serialization/SerializationManager.cs
Drawaes: Agreed, apologies I think auto format did that.

src/Orleans/Serialization/SerializationManager.cs
Drawaes: On the volatile keyword, I am worried here that you won't see the reference change. The readers have no lock and no memory barrier. So if a thread has cached the reference, will the lock on another thread be enough to ensure it gets the new reference or could it in theory be cached to the old reference for ever? Also as the serializer and deserializer can only ever be as a pair, is it worth just making it a dictionary<type,Tuple<Serilizer,Deserizlizer>> ?

Drawaes: making i just a Dictionary(TypeInfo, Tuple ( Serializer, Deserializer)) will have the added effect of no one seeing a serializer or deserializer without seeing the other side of the equation. It makes it atomic in nature

jdom: if you are worried about cases that reads need to be up to date entirely and cannot use a stale cache (which will not stay forever), then use Volatile.Read(ref copiers) instead. In my experience, using volatile was almost always the wrong (and lazy) approach. I remember even seeing lots of smart people discussing that introducing the `volatile` keyword in c# was a bad idea because it does not do the intuitive thing. I don't recall the details, but let me know if you'd like me to try to find it. Regardless, Volatile.Read might ease your concerns.

jdom: Re: pair of methods, I guess it's fine to have a tuple if they are always assigned as a pair (which does make sense)

Drawaes:
Okay in reviewing my work to make the changes you have recommended I have found an issue with the existing types (Dictionary(string,type)) variable. It is currently locked on writing but not on reading, so I am going to wrap it in the same semantics.

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -660,6 +538,25 @@ public static void Register(Type type, Type serializerType)
...
+        {
+            V currentValue;
+            if (forceOverride || !currentDictionary.TryGetValue(keyToUpdate, out currentValue) || (currentValue == null))
+            {
+                if (addingComplete)
+                {
+                    var newDictionary = currentDictionary.ToDictionary(kv => kv.Key, kv => kv.Value);
+                    newDictionary[keyToUpdate] = valueToUpdate;
+                    System.Threading.Thread.MemoryBarrier();
+                    currentDictionary = new ReadOnlyDictionary<K,V>(newDictionary);
```
jdom: I did not mean to use an actual ReadOnlyDictionary concrete instance, but just use IReadOnlyDictionary (which is already implemented by Dictionary). Not sure if the perf hit of wrapping in a ReadOnlyDictionary that prevents mutations is really necessary. Having said that, we probably don't mutate this very often, so it will not be in the hot path once the silo is warmed up. I'm not sure if there is a perf hit when reading using this wrapper class, so unless we test that, I'd say it's safer to just assign the Dictionary to the IReadOnlyDictionary field.

jdom:
Awesome, thanks for the changes! I'll merge them tomorrow, after a run through our load and reliability test suite.

Drawaes:
As an aside, it might be worth looking at "unrolling" the internal types, seeing as they are basically static. Wire does something similar, as a bunch of if's is faster than a lookup, and we can pass through the token potentially. But I will look at that another day.

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -98,18 +98,18 @@ public static object GetUninitializedObjectWithFormatterServices(Type type)
...
+        private static IReadOnlyDictionary<RuntimeTypeHandle, Tuple<Serializer,Deserializer>> serializers;
+        private static bool addingComplete;
         private static ConcurrentDictionary<Type, Func<GrainReference, GrainReference>> grainRefConstructorDictionary;
 
+
         private static IExternalSerializer fallbackSerializer;
         private static LoggerImpl logger;
         private static bool IsBuiltInSerializersRegistered;
         private static readonly object registerBuiltInSerializerLockObj = new object();
-        internal static int RegisteredTypesCount { get { return registeredTypes == null ? 0 : registeredTypes.Count; } }
```
jdom: why is this gone? I assume this is used for debugging purposes, and I normally dislike dead code, but was there a legitimate reason to remove it in this PR?

src/Orleans/Serialization/SerializationManager.cs
jdom: no worries, but do you mind reverting these? I think it might cause unnecessary conflicts with our core clr branch, where serialization is a big conflicting point

jdom:
Looks very good, thanks. I'll do 1 last perf run and merge tomorrow.

src/Orleans/Serialization/SerializationManager.cs
Drawaes: I can put it back, it wasn't threadsafe (as it was 2 calls to registeredTypes) and I couldn't find any references, but I can wrap it in a lock and throw it back in.

jdom: hmmm, I see.
how about:
```
internal static int RegisteredTypesCount { get { return registeredTypes == null ? 0 : lock (registeredTypes) registeredTypes.Count; } }
```
This way if registeredTypes is null, the lock still works

Drawaes:
Could do, I was avoiding changing too much code, as I wanted to make the effect of the removal of the locking the focus. I am happy to make that change if it won't effect anything else.

jdom:
Just as an update, I didn't forget about this PR. I'm still investigating why these changes are currently stalling (or delaying to the point of timing out) our Functionals test run.

Drawaes:
Okay, that's odd, is there anything I can help with or look at for you?

Drawaes:
i think I have found one case where there could be a problem
that would be if a copier called the serialize or deserialze
because updating the copier and the serialize and deserialze isn't atomic
so you could update the copier (that calls the serialize method) and you haven't yet put the serializer into the dictionary. Not sure if that is ever done, but it could catch someone out.

Drawaes:
We could fix that by putting the copier, serializer and deserializer into a light wrapper object for the three (a tuple<1,2,3,> is getting a bit messy) and make sure that updates are always atomic

jdom:
But that wouldn't cause a deadlock, correct? Just a failure I suppose. Nevertheless I think that improving it is always good, so go ahead, I'll run it and see if it helps, since these issues are tricky to troubleshoot

Drawaes:
Are they tests in the pack in the git hub? or are they internal only?

jdom:
Yes, these are just the github tests. They are not the Bits, which is what Jenkins (our CI server in github) runs, but also the functionals.
Effectively our build downloads the code, it adds a credentials file (with accounts for azure storage and eventhub) and then runs build.cmd and testall.cmd

Drawaes:
Okay, is a specific set of tests or test? Preferably one I don't need an azure storage, I have an msdn account through work, so I should get around to signing to azure.

Drawaes:
Okay there is a class I can reuse already (well struct actually) SerializerMethods, and its immutable which is what i planed so its perfect

Drawaes:
Changed the locking to a single common lock object to avoid any confusion across the class. I made a single dictionary for type registration, copy,ser,deser to reduce the potential number of dictionary copies. I have included a "SetupStart", "SetupEnd" method on the serializationmanager which would allow you to back to the non-copy non threadsafe mode. My proposal on that is to put that just in at the start of the silo startup codegen, and then just at the end of that codegen phase. This would stop a large amount of reallocation during the initial codegen.

sergeybykov:
@Drawaes, are you still working on this?

Drawaes:
Ahh there are conflicts now. I didn't see that. It passed the tests and was waiting for a review by @jdom

sergeybykov:
@Drawaes Can you rebase it?

Drawaes:
Will do. Is it okay if I do it tomorrow?

sergeybykov:
Of course. :-) We've been holding you for much longer. Sorry about that,

sergeybykov:
@ReubenBond, will this conflict with #2162 or #2165 in any way?
@Drawaes Do you need help with rebasing it? We can do that if you give push access to your branch.

Drawaes:
sure who can I give push access to?

sergeybykov:
@jdom is the Git ninja. Give it to him.

Drawaes:
done

gabikliot:
What is the status of this?

jdom:
We tried rebasing it to latest, but still a few issues remained. Nevertheless we are also working on changing how the discovery of serializer types (and grain invokes, etc) occurs as a separate stage from registering them in SerializationManager, so it's better if we hold it off until then to try to rebase it

Drawaes:
Hi,

What is the status on this? If the serialization manager has settled down and you want me to take a look I am happy to?

ReubenBond:
Hey @Drawaes :) If you take a look at this, I'll make sure to fix any issues we run into with this or other PRs. 

BTW, I only just now noticed your handle is Seaward backwards (I'm a disgrace)

Drawaes:
I think u are the first. I need a prize 4 you!

jdom:
LOL, I always wondered what your handle meant... I always read it like "Dr Awaes" in my head, thinking it was misspelling Dr Awes or something

Drawaes:
I think the highlight was @richorama attempting to pronounce it like an exotic name during a community meeting... I said Tim was fine :)

jdom:
btw, the big SerializationManager changes in #2592 were already merged. Do you think this PR still apply? If so, can you update it to the latest?

Drawaes:
Before I close I will take a quick look. It should be tonight

Drawaes:
Hey sorry I have been slack. This is redundant now.

PR #2085, opened 2016-08-22
[.NET Core] Update xprojs for dotnet instead of DNX.  Update some packages.

adamhathcock:
This now runs on the dotnet RTM build process.

Updated xunit packages for RTM.
Updated Newtonsoft as ASP.NET Core stuff uses 9.0.1 and there are conflicts otherwise.

Was going to update Protobuf from the alpha version but things changed and I have no idea how to regenerate the protobuf code.

### DIFF
--- a/test/TesterInternal/project.json
+++ b/test/TesterInternal/project.json
@@ -33,7 +33,6 @@
     "System.Threading.Tasks.Parallel": "4.0.1",
     "WindowsAzure.Storage": "7.0.0",
     "xunit": "2.2.0-beta2-build3300",
-    "xunit.runner.visualstudio": "2.1.0",
     "Xunit.SkippableFact": "1.2.14"
   },
   "frameworks": {
--- a/src/ClientGenerator/ClientGenerator.xproj
+++ b/src/ClientGenerator/ClientGenerator.xproj
@@ -4,15 +4,16 @@
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
   </PropertyGroup>
-  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />
+  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
   <PropertyGroup Label="Globals">
     <ProjectGuid>3790c530-a187-4de7-9b77-aff493fae7f0</ProjectGuid>
     <RootNamespace>ClientGenerator</RootNamespace>
-    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">..\..\artifacts\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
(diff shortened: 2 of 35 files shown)


test/TesterInternal/project.json
```diff
@@ -29,11 +29,9 @@
     "TestInternalGrainInterfaces": { "target": "project" },
     "TestInternalGrains": { "target": "project" },
     "dotnet-test-xunit": "2.2.0-preview2-build1029",
-    "Newtonsoft.Json": "7.0.1",
     "System.Threading.Tasks.Parallel": "4.0.1",
     "WindowsAzure.Storage": "7.0.0",
     "xunit": "2.2.0-beta2-build3300",
-    "xunit.runner.visualstudio": "2.1.0",
```
jdom: this is needed to see the tests in VS, isn't it?

src/Orleans/project.json
```diff
@@ -1,7 +1,7 @@
 {
   "version": "1.0.0-*",
   "dependencies": {
-    "Newtonsoft.Json": "7.0.1"
+    "Newtonsoft.Json": "9.0.1"
```
jdom: I would split this version change to a separate commit. I was already planning to remove the dependency on Json for the other projects, as it seemed like it was a transitive dependency, so thanks for it.
The reason I'm saying to split it, is that probably removing Json on other projects should also be done in the `master` branch (as we are using NuGet transitive restore there), so separate commits would make conflicts more manageable (since we don't want to miss any other reference upgrades that may happen in this long lived branch). 'master' can just cherrypick this commit

test/TesterInternal/project.json
adamhathcock: This was only required for DNX.  Under the dotnet cli the dotnet-test-xunit package is what's needed

src/Orleans/project.json
adamhathcock: I was going to leave it alone but wasn't sure if you wanted master on the latest or not.

I can revert this stuff but I'm now on vacation until Tuesday.  Will fix then.

adamhathcock:
Reverted the Newtonsoft changes.  Should be covered here: https://github.com/dotnet/orleans/pull/2102

adamhathcock:
I think this is going to be broken until https://github.com/dotnet/orleans/pull/2102 is actually in.

Things fail because of Newtonsoft mismatches.  Orleans should be using 9.0.1 as this is what a lot of the Microsoft.Extensions libraries use.

jdom:
Not sure what you mean with "things fail". I tried building and running the tests and they pass. I do see a warning when building the test project:
```
Compiling tester for .NETFramework,Version=v4.5.1
D:\projects\OrleansVso\test\tester\project.json(37,30): warning NU1012: Dependency conflict. Microsoft.DotNet.ProjectModel 1.0.0-rc3-003121 expected Newtonsoft.Json >= 9.0.1 but received 7.0.1
D:\projects\OrleansVso\test\tester\project.json(37,30): warning NU1012: Dependency conflict. Microsoft.Extensions.DependencyModel 1.0.0 expected Newtonsoft.Json >= 9.0.1 but received 7.0.1
D:\projects\OrleansVso\test\tester\project.json(37,30): warning NU1012: Dependency conflict. Microsoft.Extensions.Testing.Abstractions 1.0.0-preview2-003121 expected Newtonsoft.Json >= 9.0.1 but received 7.0.1
```

Nevertheless it seems to be harmless. Having said that, we will probably update to 9.0.1 in that branch really soon.

jdom:
Thanks @adamhathcock ! I'll clean up and pull what's in master to this branch soon

adamhathcock:
Thanks!  Don't know why I was getting errors with that then.

PR #2087, opened 2016-08-22
Upgrade to ServiceBus 3.3.2 and related changes.

jason-bragg:
- Upgrade to SERVICEBUS 3.3.2
- New EventData keeps eventdata properties and user properties in different bags, which caused PartitionKey property to be lost when data passed through cache. - fixed
- Azure storage read table call was failing, but was never noticed because there was no testing.  Added testing and fixed storage layer.
- Azure clear table call expected single partition.  Didn't work if there were multiple partitions in the table. -fixed.
- EventHub reciever, when reading from 'now', was reading from the last data in the queue, rather than all future data.  -fixed
- EventHub recovery test was intermittently failing because it was using two silos, but poor fault injection requires that grains stay on same silo.  -fixed.

### DIFF
--- a/src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
+++ b/src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
@@ -124,26 +124,10 @@ namespace Orleans.AzureUtils
         public async Task ClearTableAsync()
         {
             IEnumerable<Tuple<T,string>> items = await ReadAllTableEntriesAsync();
-            IEnumerable<IGrouping<string, Tuple<T, string>>> partitions = items.GroupBy(item => item.Item1.PartitionKey);
-            List<Tuple<T,string>> batch = new List<Tuple<T, string>>(AzureTableDefaultPolicies.MAX_BULK_UPDATE_ROWS);
-            foreach (IGrouping<string, Tuple<T, string>> partition in partitions)
-            {
-                foreach (Tuple<T, string> item in partition)
-                {
-                    batch.Add(item);
-                    // delete and clear when we have a full batch
-                    if (batch.Count == AzureTableDefaultPolicies.MAX_BULK_UPDATE_ROWS)
-                    {
-                        await DeleteTableEntriesAsync(batch);
-                        batch.Clear();
-                    }
-                }
-                if (batch.Count != 0)
-                {
-                    await DeleteTableEntriesAsync(batch);
-                    batch.Clear();
-                }
-            }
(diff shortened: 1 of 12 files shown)


src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -123,22 +123,27 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
...
+                {
+                    batch.Add(item);
+                    // delete and clear when we have a full batch
+                    if (batch.Count == AzureTableDefaultPolicies.MAX_BULK_UPDATE_ROWS)
+                    {
+                        await DeleteTableEntriesAsync(batch);
+                        batch.Clear();
+                    }
+                }
+                if (batch.Count != 0)
```
jdom: Is it worth parallelizing? Also, it might be work creating a small extension method for batching, to make it more readable (and also easier to parallelize if that's appropriate.
See http://stackoverflow.com/questions/13731796/create-batches-in-linq for very simple extensions

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubQueueCache.cs
```diff
@@ -243,7 +243,7 @@ protected override void OnPurge(CachedEventHubMessage? lastItemPurged, CachedEve
         {
             if (log.IsInfo && lastItemPurged.HasValue && newestItem.HasValue)
             {
-                log.Info($"CachePeriod: EnqueueTimeUtc: {LogFormatter.PrintDate(lastItemPurged.Value.EnqueueTimeUtc)} to {LogFormatter.PrintDate(newestItem.Value.EnqueueTimeUtc)}, DequeueTimeUtc: {LogFormatter.PrintDate(lastItemPurged.Value.DequeueTimeUtc)} to {LogFormatter.PrintDate(newestItem.Value.DequeueTimeUtc)}");
+                log.Verbose($"CachePeriod: EnqueueTimeUtc: {LogFormatter.PrintDate(lastItemPurged.Value.EnqueueTimeUtc)} to {LogFormatter.PrintDate(newestItem.Value.EnqueueTimeUtc)}, DequeueTimeUtc: {LogFormatter.PrintDate(lastItemPurged.Value.DequeueTimeUtc)} to {LogFormatter.PrintDate(newestItem.Value.DequeueTimeUtc)}");
```
jdom: `if (log.Verbose)` to avoid concatenating if unneeded

PR #2090, opened 2016-08-23
Update HelloWorld sample

xiazen:
Update HelloWorld Sample

- made HelloWorld.SeprarateProcess sln and HelloWorld.SingleProcess sln
- in HelloWorld.SeprarteProcess sln, orleans client and host running in
  two sepaeate process and be managed in its own project
- in HelloWorld.SingleProcess sln, it is the old HelloWorld Sample

In the future, I intend to make a OrleansHost and OrleansClient project template under Microsoft Orleans Tools for Visual Studio extension if the community feels good with this sample. For reasons below:
1. Running client and host in separate process and manage them in different project is the common practice we want users to adapt
2. If 1 is true, we should guide users to use it that way by providing them two separate project templates


Samples/HelloWorld/OrleansClient/OrleansClient.csproj
```diff
@@ -0,0 +1,108 @@
...
+    </Reference>
+    <Reference Include="Microsoft.CodeAnalysis.CSharp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.CodeAnalysis.CSharp.1.0.0\lib\net45\Microsoft.CodeAnalysis.CSharp.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=7.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\packages\Newtonsoft.Json.7.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Orleans, Version=1.3.0.0, Culture=neutral, processorArchitecture=MSIL">
```
xiazen: will update these version to 1.2.3 in my next iteration with other feedbacks I may get

Samples/HelloWorld/OrleansHost/OrleansHost.csproj
```diff
@@ -0,0 +1,125 @@
...
+    <Reference Include="OrleansHost, Version=1.3.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Orleans.OrleansHost.1.3.0\lib\net451\OrleansHost.exe</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="OrleansProviders, Version=1.3.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Orleans.OrleansProviders.1.3.0\lib\net451\OrleansProviders.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="OrleansRuntime, Version=1.3.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Orleans.OrleansRuntime.1.3.0\lib\net451\OrleansRuntime.dll</HintPath>
```
xiazen: will update these orleans version to 1.2.3 in my next iteration with other feedbacks I may get

Samples/HelloWorld/OrleansClient/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+            const int initializeAttemptsBeforeFailing = 5;
+            int attemp = 0;
+            while (true)
+            {
+                try
+                {
+                    GrainClient.Initialize(config);
+                    Console.WriteLine("Client successfully connect to silo host");
+                    var friend = GrainClient.GrainFactory.GetGrain<IHello>(0);
+                    Console.WriteLine("\n\n{0}\n\n", friend.SayHello("Good morning, my friend!").Result);
```
jdom: this should be outside of the initialization retry loop.

Samples/HelloWorld/OrleansHost/OrleansHostWrapper.cs
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            return new SiloArgs(siloName ?? Dns.GetHostName(), deploymentId); // Default to machine name
+        }
+
+        public void PrintUsage()
+        {
+            Console.WriteLine(
+                @"USAGE: 
+    [OrleansHost] [name=<siloName>] [deploymentId=<idString>] [/debug]
```
jdom: concatenate `Assembly.GetExecutingAssembly().Name` instead of hard-coding OrleansHost, since the console app name might be different.
Also, the exe name is not optional, so remove the brackets.

Samples/HelloWorld/OrleansHost/OrleansHostWrapper.cs
```diff
@@ -0,0 +1,171 @@
...
+Where:
+    OrleansHost     - the name of your orleans host console app, default to OrleansHost[.exe]
+    name=<siloName> - Name of this silo in the Config file list (optional)
```
jdom: [...] `in the node configuration override list`

Samples/HelloWorld/OrleansHost/OrleansHostWrapper.cs
```diff
@@ -0,0 +1,171 @@
...
+        public SiloArgs(string siloName, string deploymentId)
+        {
+            this.DeploymentId = deploymentId;
+            this.SiloName = siloName;
+        }
+
+        public SiloArgs()
+        {
+            SiloName = null;
+            DeploymentId = null;
```
jdom: no need to assign these to null

Samples/HelloWorld/OrleansHost/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// Orleans test host
+    /// </summary>
+    public class Program
+    {
+        private static OrleansHostWrapper hostWrapper;
+        static int Main(string[] args)
+        {
+            int returnCode = StartSilo(args);
+
+            Console.WriteLine("Press Enter to terminate...");
```
jdom: if the silo didn't start, you should terminate without waiting for input or calling shutdown

Samples/HelloWorld/OrleansHost/OrleansHostWrapper.cs
```diff
@@ -0,0 +1,171 @@
...
+                try
+                {
+                    siloHost.StopOrleansSilo();
+                    siloHost.Dispose();
+                    Console.WriteLine("Orleans silo '{0}' shutdown.", siloHost.Name);
+                }
+                catch (Exception exc)
+                {
+                    siloHost.ReportStartupError(exc);
+                    Console.Error.WriteLine("{0}:\n{1}\n{2}", exc.GetType().FullName, exc.Message, exc.StackTrace);
```
jdom: `Console.Error.WriteLine(exc);`

Samples/HelloWorld/OrleansHost/OrleansHostWrapper.cs
```diff
@@ -0,0 +1,171 @@
...
+                    siloHost.InitializeOrleansSilo();
+
+                    if (siloHost.StartOrleansSilo())
+                    {
+                        Console.WriteLine("Successfully started Orleans silo '{0}' as a {1} node.", siloHost.Name, siloHost.Type);
+                        return 0;
+                    }
+                    else
+                    {
+                        throw new SystemException(string.Format("Failed to start Orleans silo '{0}' as a {1} node.", siloHost.Name, siloHost.Type));
```
jdom: avoid using SystemException as it is deprecated. You might use OrleansException instead, as it seems appropriate here

Samples/HelloWorld/OrleansHost/OrleansHostWrapper.cs
```diff
@@ -0,0 +1,171 @@
...
+        public void PrintUsage()
+        {
+            Console.WriteLine(
+                @"USAGE: 
+    [OrleansHost] [name=<siloName>] [deploymentId=<idString>] [/debug]
+Where:
+    OrleansHost     - the name of your orleans host console app, default to OrleansHost[.exe]
+    name=<siloName> - Name of this silo in the Config file list (optional)
+    deploymentId=<idString> 
+                    - Which deployment group this host instance should run in (optional)");
```
jdom: - Optionally override the deployment group this host instance should run in (otherwise will use the one in the configuration)

Samples/HelloWorld/OrleansHost/Program.cs
xiazen: Yes it totally make sense that it might not be convenient to wait for user input to terminate a silo. 
But if we terminate the program without user input, the user wouldn't be able to see exception printed to console , and they would get confused on why program ended. We didn't print the exception message to logs either. 

And ShutdownSilo() handles the case when silo didn't start up correctly. Look at OrleansHostWrapper.Stop(). In main method, in my opinion, for the goal of precise and readability, we should just express to people the simple logic: start the silo, wait for your input, then we shut down the silo. If people wants more detailed info, they can dig into one layer down to specific methods to know how those method handle exceptions, edge cases, etc

Samples/HelloWorld/OrleansHost/OrleansHostWrapper.cs
```diff
@@ -0,0 +1,168 @@
...
+            // Default to machine name
+            siloName = siloName ?? Dns.GetHostName();
+            return new SiloArgs(siloName, deploymentId); 
+        }
+
+        public void PrintUsage()
+        {
+            Console.WriteLine(
+                @"USAGE: 
+                Assembly.GetExecutingAssembly().Name [name=<siloName>] [deploymentId=<idString>] [/debug]
```
jdom: hmmm, I did not meant to add a literal string, but to evaluate `Assembly.GetExecutingAssembly().Name` and output that to the console. Just add {0} and the pass it as another argument to WriteLine

xiazen: {0} ? what does it mean? 

How about we just do YouOrleansHostConsoleAppName?

xiazen:
Addressed all feedback except for the two I left comment on. rebased and squashed

Samples/HelloWorld/OrleansHost/OrleansHostWrapper.cs
xiazen: addressed

Samples/HelloWorld/HelloWorld.SeparateProcess.sln
```diff
@@ -0,0 +1,40 @@
...
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "HelloWorldInterfaces", "HelloWorldInterfaces\HelloWorldInterfaces.csproj", "{3BD01348-ACC3-4378-A49E-4D62FB0BD3D0}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "HelloWorldGrains", "HelloWorldGrains\HelloWorldGrains.csproj", "{943794E3-F54E-4678-B225-B45018344AEC}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OrleansClient", "OrleansClient\OrleansClient.csproj", "{05CB5084-F694-4CD1-A4A3-C08ABE8B4476}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OrleansHost", "OrleansHost\OrleansHost.csproj", "{76521F11-81CD-46D0-970F-2F972F67F626}"
```
jdom: should we call the project something different than OrleansHost? Otherwise this will generate an OrleansHost.exe file which can be confused with the pre-built OrleansHost.exe that comes in the OrleansHost nuget package.
Maybe SiloHost?

Samples/HelloWorld/OrleansClient/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+                        Console.WriteLine("Press Enter to terminate...");
+                        Console.ReadLine();
+                        return 1;
+                    }
+                    Thread.Sleep(TimeSpan.FromSeconds(2));
+                }
+            }
+
+            var friend = GrainClient.GrainFactory.GetGrain<IHello>(0);
+            Console.WriteLine("\n\n{0}\n\n", friend.SayHello("Good morning, my friend!").Result);
```
jdom: I understand this is similar to the other hello world, but I always disliked that it confuses people into thinking they must use `.Result` after every grain call from the client.
I'd move this to its own async method such as:
```
private async Task DoClientWork()
{
      var friend = GrainClient.GrainFactory.GetGrain<IHello>(0);
      var response = await friend.SayHello("Good morning, my friend!")
      Console.WriteLine("\n\n{0}\n\n", response);
} 
```
and then just call it from here by using `DoClientWork.Wait()`. It accomplishes a very similar thing, it's just that it's clear that you can (and should) await grain calls. The .Wait or .Result is just a side effect of having the Main entry point not being asynchronous, but not of Orleans programming model, and for real world apps, you will never have all of your client code embedded in the Program.Main method.

xiazen: good point. will do

xiazen: fixed

Samples/HelloWorld/HelloWorld.SeparateProcess.sln
xiazen: We already have SiloHost in the other project, which has client and host in the same process. 

Maybe OrleansSiloHost? OrleansSilo?

xiazen: fixed. named to OrleansSiloHost

xiazen:
any other concerns ? or someone can merge it for me ?

xiazen:
rebased and added readme file and took @jdom 's change too.

jdom:
@dotnet-bot test this please

PR #2096, opened 2016-08-25
Updated Google.Protobuf dependency

seniorquico:
This PR updates the Google.Protobuf dependency to a non-prerelease version.

There is a problem with this PR, though. The "GenerateProtos.cmd" script has a hard coded dependency on the "protoc.exe" utility. In the prerelease versions, "protoc.exe" was included in the Google.Protobuf package. It appears They have split the utility into its own package, Google.Protobuf.Tools.

I am still learning about the changes that project.json brings to projects. In particular, it appears that adding Google.Protobuf.Tools as a dependency won't work as it had previously. The "protoc.exe" utility doesn't appear to be extracted into a packages folder within the solution. It's not clear to me how a tool can be called from a script with this new package structure. Any suggestions are welcome, and I am happy to amend this PR.

### DIFF
--- a/src/.nuget/packages.config
+++ b/src/.nuget/packages.config
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="FSharp.Compiler.Tools" version="4.0.0.1" />
+  <package id="Google.Protobuf.Tools" version="3.0.0" />
   <package id="xunit.runner.console" version="2.1.0" />
   <package id="xunit.runner.visualstudio" version="2.1.0" />
 </packages>
--- a/src/OrleansGoogleUtils/project.json
+++ b/src/OrleansGoogleUtils/project.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "Google.Protobuf": "3.0.0-alpha4"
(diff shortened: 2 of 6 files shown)


gabikliot:
I think we should still upgrade, even if we cannot automate GenerateProtos.cmd script. The user can still download the protoc.exe utility manually and copy it. I added the script  for convenience, but I don't consider it Orleans's issue to automate protobuff tool. It's protobuff issue. If we later find a way to automate - great, but even if we don't we should move to the new nuget version.

For example, we can just manually save the matching protoc.exe into Orleans's tree and put it next to GenerateProtos.cmd script. And upgrade it every time we upgrade the protobuf nuget version.

Please sign cla.

jdom:
you can edit /src/.nuget/packages.config and add the tools package there. Then it will be downloaded to the same `packages` directory as it was being downloaded prior to our move to `project.json`

gabikliot:
Thank you! 
Woukd you be interested in a tighter integration with  Google Cloud Platform, or are you only interested in protobuf?

seniorquico:
I was only interested in testing the Protobuf serialization. My project has an on-premise requirement, so I haven't thought yet about cloud services.

PR #2099, opened 2016-08-27
[WIP] Dependency Injection

xiazen:
- Make orleans service provider available to application layer grain (#934)
- Make GrainFactory injectable ()  #988 

TODO:
1. find real dependency of each class which requires DI, inject the real dependency instead of the whole ServiceProvider
2. look at how ASP.NET handle third party DI container, maybe borrow some of their pattern over. Right now we are using the original method
3.  update external providers to be use DI system, I briefly tried in PersistentStreamProvider this time. 
4. if we are going this way, we will change public class, such as Silo.cs, AzureSilo.cs, which introduce backwards compatibility problem. 
5. if we are going this direction, a lot of files will need to change, mostly tests and external provider classes and its related classes, especially those manager class which is in charge of instantiating the external provider classes . 

This PR is not complete, post this to the community as a possible direction we want to go in fixing DI.

### DIFF
--- a/src/OrleansRuntime/Startup/IStartup.cs
+++ b/src/OrleansRuntime/Startup/IStartup.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Runtime.MembershipService;
+using Orleans.Runtime.ReminderService;
+
+namespace Orleans.Runtime.Startup
+{
+    public interface IStartup
+    {
+        IServiceProvider ConfigureServices(IServiceCollection services);
+    }
+}
--- a/src/OrleansRuntime/Startup/OrleansInternalServices.cs
+++ b/src/OrleansRuntime/Startup/OrleansInternalServices.cs
@@ -1,0 +1,126 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Runtime.MembershipService;
+using Orleans.Runtime.ReminderService;
+
+namespace Orleans.Runtime.Startup
+{
+    // <summary>
+    // extension methods to register services required internal Orleans methods
+    // </summary>
+    public static class OrleansInternalServices
+    {
+
+        // extension methods to register internal orleans required services
+        public static void RegisterSystemTypes(IServiceCollection serviceCollection)
+        {
+            // Register the system classes and grains in this method.
+            // Note: this method will probably have to be moved out into the Silo class to include internal runtime types.
+            IGrainFactory grainFactory = new GrainFactory();
+            serviceCollection.AddSingleton((GrainFactory)grainFactory);
+            //application grain getService by interface type which is public to them, while concret types are internal
+            serviceCollection.AddSingleton<IGrainFactory>(grainFactory);
+            serviceCollection.AddTransient<IMembershipTable, GrainBasedMembershipTable>();
+            serviceCollection.AddTransient<IReminderTable, GrainBasedReminderTable>();
+        }
+
+        //default service provider where only contains internal services, mostly used in tests
+        public static IServiceProvider DefaultServiceProvider()
+        {
+            return ServiceProvider;
+        }
+
+
+        private static IServiceProvider BuildDefaultServiceProvider()
+        {
+            IServiceCollection serviceCollection = new ServiceCollection();
+            RegisterSystemTypes(serviceCollection);
+            return serviceCollection.BuildServiceProvider();
+        }
+
+        private static readonly IServiceProvider ServiceProvider = BuildDefaultServiceProvider();
+
+
+        //use this method for third party DI containers
+        public static IServiceProvider ConfigureStartup(string startupTypeName)
+        {
+            bool usingCustomServiceProvider = false;
+            IServiceCollection serviceCollection = new ServiceCollection();
+            ConfigureServicesBuilder servicesMethod = null;
+            Type startupType = null;
+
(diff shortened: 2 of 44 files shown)


test/TesterInternal/Deployment/AzureSiloTests.cs
```diff
@@ -90,4 +96,13 @@ public IPEndPoint GetIPEndpoint(string endpointName)
...
     }
+
+    public class Startup : IStartup
+    {
+        public IServiceProvider ConfigureServices(IServiceCollection svcCollection)
+        {
+            OrleansInternalServices.RegisterSystemTypes(svcCollection);
+            return svcCollection.BuildServiceProvider();
+        }
+    }
```
xiazen: This is a pattern borrowed from ASP.Net startUp class pattern : https://docs.asp.net/en/latest/fundamentals/startup.html
TODO here to put configuration and initialization in this class, for example, init an AzureSilo.

src/OrleansRuntime/Startup/IStartup.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Runtime.MembershipService;
+using Orleans.Runtime.ReminderService;
+
+namespace Orleans.Runtime.Startup
+{
+    public interface IStartup
```
jason-bragg: Why introduce our own?  Microsoft.AspNetCore.Hosting.IStartup?

src/OrleansTestingHost/TestCluster.cs
```diff
@@ -107,7 +110,7 @@ public void Deploy(IEnumerable<string> siloNames)
         /// Deploys the cluster using the specified configuration and starts the client in-process.
         /// </summary>
         /// <param name="siloNames">Only deploy the specified silos which must also be present in the <see cref="Runtime.Configuration.ClusterConfiguration.Overrides"/> collection.</param>
-        public async Task DeployAsync(IEnumerable<string> siloNames)
+        public async Task DeployAsync(IEnumerable<string> siloNames, IServiceProvider serviceProvider)
```
jason-bragg: Not convinced deploying a service provider is a good idea.  Maybe an IServiceCollection, but a fully populated container sounds like a recipe for badness.

src/OrleansRuntime/Startup/IStartup.cs
xiazen: Mirosoft.AspNetCore.HosttingIStartup has two methods
1. Configure(Microsoft.AspNetCore.Builder.IApplicationBuilder)
2.  ConfigureServices(Microsoft.Extensions.DependencyInjection.IServiceCollection)

we definitely want to borrow method 2, but I'm not sure about method 1, since I IApplicationBuilder is pretty web server builder pattern, here is its interface doc: https://docs.asp.net/projects/api/en/latest/autoapi/Microsoft/AspNetCore/Builder/IApplicationBuilder/

jason-bragg: Fair point.  At first glance it looked like the app builder could be used to build subsystems like streaming or storage, but upon further inspection, yeah, it's clearly for building http request delegates.

src/OrleansTestingHost/TestCluster.cs
xiazen: yes agree here, that's why I have the TODO keyword in Silo constructor on "need to inject real depdendency instead of the whole ServiceProvider"

src/OrleansRuntime/Startup/OrleansInternalServices.cs
```diff
@@ -0,0 +1,126 @@
...
+    public static class OrleansInternalServices
+    {
+
+        // extension methods to register internal orleans required services
+        public static void RegisterSystemTypes(IServiceCollection serviceCollection)
+        {
+            // Register the system classes and grains in this method.
+            // Note: this method will probably have to be moved out into the Silo class to include internal runtime types.
+            IGrainFactory grainFactory = new GrainFactory();
+            serviceCollection.AddSingleton((GrainFactory)grainFactory);
```
jason-bragg: Suggest adding an internal static GrainFactory.Create() call and making the grain factory constructor private to track down all places where internal systems are not getting the grain factory from DI.  This can be done in a second pass, but until we track down all instances where grain factory is internally constructed, the grain factory is not injectable.

src/OrleansTestingHost/TestCluster.cs
xiazen: here should inject the real dependency required by DeployAsync, instead of ServiceProvider. But I'm not sure what's the real dependency is, need to look more into the code. So I had the TODO keyword all over

src/OrleansRuntime/Startup/OrleansInternalServices.cs
xiazen: Yes totally agree with you here. My first pass is super far away from finished. Just an demonstration of where I want to go.

test/TesterInternal/Deployment/AzureSiloTests.cs
```diff
@@ -90,4 +96,13 @@ public IPEndPoint GetIPEndpoint(string endpointName)
 
         public void UnsubscribeFromStoppingNotification(object handlerObject, EventHandler<object> handler) { }
     }
+
+    public class Startup : IStartup
+    {
+        public IServiceProvider ConfigureServices(IServiceCollection svcCollection)
+        {
+            OrleansInternalServices.RegisterSystemTypes(svcCollection);
```
jason-bragg: Why should this register the Orleans services?  Doesn't svcCollection already have them?

xiazen:
close this PR for now based on internal discussion. 
This PR do solved the problem of 
*Make orleans service provider available to application layer grain (#934)
•Make GrainFactory injectable () #988 

But since we are borrowing patterns from ASP.NET, we want to dig more into how ASP.NET handles DI. We will move this PR offline and come back with a solution the core team agreed on.

Issue #2011, opened 2016-08-02
Membership, storage and reminder providers as external extensions


ashkan-saeedi-mazdeh:
I've realized the need to use a different storage provider for my application. 
I'm writing a complete storage provider, IMembershipTable and IReminderTable for CouchBase KVS. 
http://www.couchbase.com/

Should we in general consider moving these to Orleans like Azure, SQL and other cloud providers or keep them outside and just list them in contrib?

Also if in contributions, then should we include them in a visible part of documentation or not? We should really try the initiative of running Orleans everywhere for EVERYONE. As an unfortunate example in CouchBase's manual they mentioned Akka.NET in their async programming with .NET page but not Orleans and we should change this. I got offended :D 

What should be the approach @sergeybykov  @jdom  @ReubenBond  @galvesribeiro  Add others as you see fit.

ashkan-saeedi-mazdeh:
Understood from @jdom  's comment on gitter when talking to @galvesribeiro  that the policy is to go to contributions.

jdom:
Not necessarily. OrleansContrib can be an enabler to reach maturity, if it helps, but it is not the only option. Please read the thread at https://github.com/dotnet/orleans/issues/2005 for clarification on @galvesribeiro's conclusion.

I do believe that regardless of whether it's in the main repo or Orleans contrib, we should be doing a better effort to decouple each provider from the main solution. We should have some sort of test kit (as @veikkoeeva suggested) that we can run our providers against, to guarantee at least a minimal set of behavior from them. Of course each provider can then have additional provider specific tests.
Providers in the main repo should not have any advantage (in terms of how in-the-box they are, or how easy it is to test them) compared to new experimental/external providers.
Also it should be good if they can be versioned and released independently of Orleans core nugets.

veikkoeeva:
[...]

PR #2104, opened 2016-08-31
Run Tester and TesterInternal assemblies in parallel

jdom:
This is now possible since we use TestCluster which assigns random listening ports to the silos, and because each assembly runs in a different AppDomain.
Added -verbose switch to xUnit console runner, so we can see which test is running and can troubleshoot in an easier way when there are issues or the run halts

### DIFF
--- a/src/Test.cmd
+++ b/src/Test.cmd
@@ -20,7 +20,7 @@ if []==[%TEST_FILTERS%] set TEST_FILTERS=-trait "Category=BVT" -trait "Category=
 @echo on
 call "%CMDHOME%\SetupTestScript.cmd" "%OutDir%"
 
-packages\xunit.runner.console.2.1.0\tools\xunit.console %TESTS% %TEST_FILTERS% -xml "%OutDir%/xUnit-Results.xml" -parallel none -noshadow
+packages\xunit.runner.console.2.1.0\tools\xunit.console %TESTS% %TEST_FILTERS% -xml "%OutDir%/xUnit-Results.xml" -parallel assemblies -noshadow -verbose
 set testresult=%errorlevel%
 popd
 endlocal&set testresult=%testresult%


jdom:
BTW, as we start splitting the Tester project into smaller self contained projects (like in #2011), we could drastically reduce the time it takes to run our tests

jdom:
@dotnet-bot test this please
reason: looks like an unrelated test failure

veikkoeeva:
@jdom Tangential to this PR, I see assigning a random port means [this](https://github.com/dotnet/orleans/blob/master/src/OrleansTestingHost/TestClusterOptions.cs#L31). There's the unfortunate possibility to run into collisions with this method. Taking a quick stab it, maybe something like

```
public static int GetEphemeralPort()
{
    using(var portFinderSocket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
    {                
        portFinderSocket.Bind(new IPEndPoint(IPAddress.Loopback, 0));
        return ((IPEndPoint)portFinderSocket.LocalEndPoint).Port;                
    }
}
```
I could turn this into a PR if it is OK? Maybe something like this would useful as a utility function when having developer deployments too.

jdom:
Don't merge this yet, having issues with some Functionals

jdom:
@veikkoeeva I believe that's interesting. 
Just have in mind that you would do that just for the base port, but for every port we assign for each node
Nevertheless, if I'm reading it correctly, that approach binds to the port (not sure how costly that is, but probably negligible) and then disposes the socket and releases the port (temporarily until the silo starts up). Nevertheless the configuration for all the nodes is typically done before we even start any silo, so the logic would have to be careful not to reuse the same port for other silos that are about to be started

jdom:
I'm getting a lot of errors after 1 of the assemblies finishes running (they run fine in parallel, but as soon as 1 completes entirely, then there are a lot of file not found exceptions). Closing for now until I investigate, as I don't want this to be merged by mistake, and I have other high priority tasks to work on before troubleshooting this one

veikkoeeva:
@jdom It could probably return the socket too. I believe giving port zero to the operating system is a standard measure to ask back a free port in the ephemeral range. I think it is undefined as to what happens to the port after releasing it, but for what I know, Windows pretty much iteratres the range forwards so it should take a while before it gives the same port again.

At worst this method has port conflict like the currently used method, but they should be more rare. Let's see if I get into computer to make a PR (perhaps with more explanation if I can find any).

PR #2109, opened 2016-09-02
Remove confusing parameter from AzureSilo.Start

benjaminpetit:
Yes, it is a breaking change.

But currently if you do something like this:
````
var config = AzureSilo.DefaultConfiguration();
config.Globals.DeploymentId = "aca7e5e9105745a0b82be8c5a6b07215";
config.AddAzureTableStorageProvider("AzureStore");
orleansAzureSilo = new AzureSilo();
bool ok = orleansAzureSilo.Start(config);
````

The Start method will partially ignore the DeploymentId that is in the configuration which is confusing.

### DIFF
--- a/src/OrleansAzureUtils/Hosting/AzureSilo.cs
+++ b/src/OrleansAzureUtils/Hosting/AzureSilo.cs
@@ -189,8 +189,10 @@ namespace Orleans.Runtime.Host
 
             // Bootstrap this Orleans silo instance
 
+            // If deploymentId was not direclty provided, take the value in the config. If it is not 
+            // in the config too, just take the DeploymentId from Azure
             if (deploymentId == null)
-                deploymentId = host.Config.Globals.DeploymentId;
+                deploymentId = host.Config.Globals.DeploymentId ?? serviceRuntimeWrapper.DeploymentId;
 
             myEntry = new SiloInstanceTableEntry
             {
--- a/src/OrleansAzureUtils/Hosting/AzureSilo.cs
+++ b/src/OrleansAzureUtils/Hosting/AzureSilo.cs
@@ -127,12 +127,28 @@ namespace Orleans.Runtime.Host
         #region Azure RoleEntryPoint methods
 
         /// <summary>
+        /// Initialize this Orleans silo for execution. Config data will be read from silo config file as normal
+        /// </summary>
+        /// <param name="deploymentId">Azure DeploymentId this silo is running under. If null, defaults to the value from the configuration.</param>
+		/// <param name="connectionString">Azure DataConnectionString. If null, defaults to the DataConnectionString setting from the Azure configuration for this role.</param>
+        /// <returns><c>true</c> is the silo startup was successful</returns>
+        public bool Start(string deploymentId = null, string connectionString = null)
+        {
+            return Start(null, deploymentId, connectionString);
+        }
+
+        /// <summary>
         /// Initialize this Orleans silo for execution
         /// </summary>
+        /// <param name="config">Use the specified config data.</param>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/OrleansAzureUtils/Hosting/AzureSilo.cs [hunk 1]
```diff
@@ -174,6 +169,8 @@ public bool Start(ClusterConfiguration config, string deploymentId = null, strin
 
             // Bootstrap this Orleans silo instance
 
+            var deploymentId = host.Config.Globals.DeploymentId;
```
jdom: >var deploymentId = host.Config.Globals.DeploymentId; [](start = 12, length = 52)

maybe you can keep the method signature as is, and here do:
`deploymentId = deploymentId ?? host.Config.Globals.DeploymentId ?? serviceRuntimeWrapper.DeploymentId;`

This way the change will not be breaking, just doing fallbacks

benjaminpetit: It is still confusing for the caller, I don't think it is bad to have a breaking change in this case. I think we should allow to pass a config or a deploymentId that will override the value from the config, but not both.

jdom: yeah, I also find this very confusing, and think it stems from when Orleans was XML config based only. I think that even today there are still deployments with a static XML configuration, but they want to just override the deploymentId in code. Granted, instead of just calling Start, they could go through the process of manually loading the XML config, then updating the deploymentId, and then calling start.
But I'd be wary about making this breaking change, when we are already thinking that several breaking changes to config/startup are coming in our Orleans 2.0 release.

jdom: btw, looks like my concern is addressed in the latest rebase

src/OrleansAzureUtils/Hosting/AzureSilo.cs
```diff
@@ -142,15 +158,14 @@ public bool Start()
...
         {
+            if (config != null && deploymentId != null)
+                throw new ArgumentException("Cannot use config and deploymentId on the same time");
+
             // Program ident
             Trace.TraceInformation("Starting {0} v{1}", this.GetType().FullName, RuntimeVersion.Current);
 
-            // Check if deployment id was specified
-            if (deploymentId == null)
-                deploymentId = serviceRuntimeWrapper.DeploymentId;
```
jdom: does this mean that if the user has an XML configuration, he must explicitly use serviceRuntimeWrapper directly?

jdom:
The most common use case is that you use Azure's Hosted Service DeploymentId. Overriding is a nice to have.
With the current code in commit fa7a81086d0bb123eb5dde0d4e5d0e6aa80c2051 this is still the case if you use programmatic config (you start with `AzureSilo.DefaultConfiguration()`).
Nevertheless if you use an external XML config, this default is lost, and it will use the hardcoded deploymentId in the XML. This default should be preserved by the PR, regardless of how you handle fallbacks or mutually exclusive overrides.

src/OrleansAzureUtils/Hosting/AzureSilo.cs
benjaminpetit: No, we don't rely on serviceRuntimeWrapper here. If the user has an XML configuration and want to override the deploymentId, he should use `Start(string deploymentId = null, string connectionString = null)`

jdom: sure, but does he have to manually pass in what was previously the default (the hosted service deployment id)?

benjaminpetit: Yes

PR #2110, opened 2016-09-03
Remove Pure attribute usage

jdom:
Code contracts is an additional dependency when targeting .NET Standard that we should not pull into just for this random occurrence.
There are no other usages of code contracts in the solution

### DIFF
--- a/src/Orleans/IDs/GrainId.cs
+++ b/src/Orleans/IDs/GrainId.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Diagnostics.Contracts;
 using System.Globalization;
 using Orleans.Core;
 using Orleans.Serialization;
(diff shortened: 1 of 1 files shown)


ReubenBond:
LGTM. We can pull Contracts back in if we decide to use them more extensively.

xiazen:
LGTM

xiazen:
Reuben is one step ahead of me ~

ReubenBond:
Eep, sorry! I was just waiting for VS to update & figured I'd check some PRs

xiazen:
no need to apologize ! Great speed there ! 👍

veikkoeeva:
Further, what would using these give to Orleans? It looks like Code Contracts, for instance, is in dead-end.

Issue #2117, opened 2016-09-05
Bug while searching in docs and broken links


galvesribeiro:
With the new docfx based docs I found out that there is some bug when you search for something.

To give an example, open the main docs page and search for SQL. Click on the first link which should be http://dotnet.github.io/orleans/Documentation/Advanced-Concepts/Configuring-SQL-Tables.html?q=SQL

Note that docfx add `?q=SQL` query string parameter to the URL. The page is rendered however the code sections (like the XML config in this case) is empty. If you look at the source of the page, the HTML code is there for that section but it isn't rendered. Just remove the querystring parameter and it render properly.

![image](https://cloud.githubusercontent.com/assets/4714040/18257969/36ad36b4-73a3-11e6-83c0-6516270a59c4.png)

I wonder if it is a docfx bug or something with our pages.

Also, there are broken links that return a 404 like in this page http://dotnet.github.io/orleans/Documentation/Runtime-Implementation-Details/Relational-Storage.html on the first paragraph there is a `SQL Tables` link. I don't know if it was a mistake on the migration steps or something on our indexing.

**Note**: it happens on chrome and edge as well.

richorama:
Issue raised with docfx: https://github.com/dotnet/docfx/issues/675

sergeybykov:
Looks like this is fixed now in DocFx.

PR #2118, opened 2016-09-05
Added missing /Documentation/ to some broken links

galvesribeiro:
Related to #2117

### DIFF
--- a/src/Documentation/Runtime-Implementation-Details/Relational-Storage.md
+++ b/src/Documentation/Runtime-Implementation-Details/Relational-Storage.md
@@ -5,7 +5,7 @@ title: Relational Storage
 
 # Relational Storage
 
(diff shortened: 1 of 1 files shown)


ReubenBond:
LGTM, thanks. By the way, are docs still going to gh-pages or are we going to use master?

galvesribeiro:
@ReubenBond as far as I understood, it will be kept on gh-pages and later one when we have the docfx build process automated, it will be pushed to another repo like OrleansDocs or something.

sergeybykov:
BTW, without rerunning DocFx, this change only went to the .md file, but not to the .html. It'll get picked up next time somebody runs DocFx. But it you want a change like this to appear in the docs immediately after the PR is merged, you need to run DocFx to update them.

galvesribeiro:
Ahhh I though we needed merge rights @sergeybykov np. Will do it on next time. Thanks for the tip

sergeybykov:
No big deal. It's more important to improve the 'sources' of the docs.

PR #2124, opened 2016-09-07
Remove message headers & metadata dictionaries 

dVakulen:
Instantiation of the two dictionaries per message is by far the top contributor to the memory pressure in many workloads: 

timeline64_2016-09-07_20-06-07

Plus as it one of the hottest code path in Orleans - usage of the dictionary consumes some chunk of CPU: 

timeline64_2016-09-07_20-11-17

timeline64_2016-09-07_20-14-06

After removal: memory footprint reduced by 4x, serialization time improved by ~10%, average serialized size reduced by ~11% (which can be 11% of whole serialized message in case of message with small payload). 

Adding of the new header will require bigger effort from now, but I think that it's justified.

### DIFF
--- a/src/Orleans/Messaging/Message.cs
+++ b/src/Orleans/Messaging/Message.cs
@@ -282,7 +282,7 @@ namespace Orleans.Runtime
             if (id == null) return false;
 
             // don't set expiration for one way, system target and system grain messages.
-           return Direction != Directions.OneWay && !id.IsSystemTarget && !Constants.IsSystemGrain(id);
+            return Direction != Directions.OneWay && !id.IsSystemTarget && !Constants.IsSystemGrain(id);
         }
 
         public string DebugContext
@@ -312,14 +312,14 @@ namespace Orleans.Runtime
         // Resends are used by the sender, usualy due to en error to send or due to a transient rejection.
         public bool MayResend(IMessagingConfiguration config)
         {
-           return ResendCount < config.MaxResendCount;
+            return ResendCount < config.MaxResendCount;
         }
 
         // Forwardings are used by the receiver, usualy when it cannot process the message and forwars it to another silo to perform the processing
         // (got here due to outdated cache, silo is shutting down/overloaded, ...).
         public bool MayForward(GlobalConfiguration config)
         {
-           return ForwardCount < config.MaxForwardCount;
+            return ForwardCount < config.MaxForwardCount;
         }
 
         /// <summary>
@@ -365,7 +365,7 @@ namespace Orleans.Runtime
             {
                 if (bodyObject != null)
                 {
-                   return bodyObject;
+                    return bodyObject;
                 }
                 try
                 {
@@ -379,7 +379,7 @@ namespace Orleans.Runtime
                         bodyBytes = null;
                     }
                 }
-               return bodyObject;
+                return bodyObject;
             }
             set
             {
@@ -395,12 +395,12 @@ namespace Orleans.Runtime
         {
             if (bytes == null)
             {
-               return null;
+                return null;
             }
             try
             {
                 var stream = new BinaryTokenStreamReader(bytes);
-               return SerializationManager.Deserialize(stream);
+                return SerializationManager.Deserialize(stream);
             }
             catch (Exception ex)
             {
@@ -443,7 +443,7 @@ namespace Orleans.Runtime
             {
                 message.RequestContextData = contextData;
             }
-           return message;
+            return message;
         }
 
         // Initializes body and header but does not take ownership of byte.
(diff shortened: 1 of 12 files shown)


sergeybykov:
Looks promising! Once @jason-bragg reviews, we'll run perf tests.

src/Orleans/Messaging/Message.cs
```diff
@@ -117,84 +105,84 @@ public enum RejectionTypes
             GatewayTooBusy,
         }
 
+        internal HeadersContainer Headers { get; set; } = new HeadersContainer();
+
         public Categories Category
         {
-            get { return GetScalarHeader<Categories>(Header.CATEGORY); }
-            set { SetHeader(Header.CATEGORY, value); }
+            get { lock (Headers) return Headers.Category; }
```
jason-bragg: Why are we locking the headers container?

src/Orleans/Messaging/Message.cs
```diff
@@ -881,5 +788,581 @@ internal void DropExpiredMessage(MessagingStatisticsGroup.Phase phase)
...
+                {
+                    stream.Write(input.@TargetSilo);
+                }
+            }
+
+            [global::Orleans.CodeGeneration.DeserializerMethodAttribute]
+            public static global::System.Object Deserializer(global::System.Type expected,  BinaryTokenStreamReader stream)
+            {
+                var result = new HeadersContainer();
+                global::Orleans.@Serialization.@DeserializationContext.@Current.@RecordObject(result);
```
jason-bragg: nit:  Please remove @s if they are not needed.

src/Orleans/Messaging/Message.cs
```diff
@@ -881,5 +788,581 @@ internal void DropExpiredMessage(MessagingStatisticsGroup.Phase phase)
...
+                }
+            }
+
+            public Directions? Direction
+            {
+                get { return _direction; }
+                set
+                {
+                    _direction = value;
+                    headers = value == null ? headers & ~Header.DIRECTION : headers | Header.DIRECTION;
```
jason-bragg: suggest adding value == default(x) checks to every setter to set/clear flags.  Seem to be doing is selectively.

jason-bragg:
This is great!
GJ!

With the flat properties, I don't think we need to lock the headers any more, but we should double check.
I'm not convinced we need to track the flags as we can build it on the fly by checking to see what properties don't have default values, but I'm not sure it matters that much.

src/Orleans/Messaging/Message.cs
dVakulen: Thought the same, but wanted to confirm.

dVakulen:
Feedback addressed

jdom:
awesome!

sergeybykov:
Let's perf test this before we merge.

jdom:
Run a portion of our load tests and results are good. TPS numbers are slightly up. I'm running another set of tests now where I expect even better relative results (because it's basically just pings, so most of the CPU and memory utilization is about Orleans internals)

jdom:
Ran the other tests, and it's sometimes a little bit better or in the same ballpark as some of our faster runs (as the numbers vary from run to run). So numbers are good for merging this. I think @jason-bragg wanted to do a final review pass since you addressed his comments, and we can merge it. :+1:

sergeybykov:
Thank you, @dVakulen! Great work, as always!

gabikliot:
Can you please share the perf. numbers? We considered that optimization with @alan-geller in the past, and the conclusion was that it will help a bit, but not significantly. I would love to know both single hop and multi hop pings test numbers, to see how off we were.

sergeybykov:
After several test runs over the weekend, it is pretty obvious that we achieved a significant throughput improvement because of this change.

**Before**
PingLoadTest_LocalReentrant: 62268.5, 61773.3, 61470.9
PingLoadTest_RandomReentrant_MultiSilos: 474885.1, 476808.9, 512502.3
NightlyLoadTest: 182312.0, 182842.7, 185517.9

**After**
PingLoadTest_LocalReentrant: 74129.2, 75153.2, 74729.0
PingLoadTest_RandomReentrant_MultiSilos: 567015.7, 554467.1, 537480.2
NightlyLoadTest: 189163.2, 197200.1, 198469.7

Thank again, @dVakulen, for this awesome contribution!

jdom:
Kudos to @dVakulen, he's doing an outstanding job on all of these PRs (not just this one). We had a very big bump in our numbers since we started merging his PRs.

galvesribeiro:
Great work @dVakulen! :) kudos for you! :shipit:

gabikliot:
This is great! 
The saving is per msg, thus the more msgs per silo we have, the larger is the saving overall. Not much in a mixed workload but pretty significant in messaging intensive workload. Very good result.

PR #2127, opened 2016-09-07
Fix #2119 by allowing full uninitialization in SiloHost

### LINKED ISSUE #2119: Allow clean uninitialization of silos which failed to start
Often I see errors amounting to trying to re-initialize some singletons in Orleans which have already been initialized. Examples include `GrainTypeManager` and `UnobservedExceptionHandler`.

Usually when a user reports an error like this, the error is actually something else which caused their silo to fail initialization, but their code retries a few times and hides the original exception further up the log.

Currently it's not clear how to uninitialize a silo so that we can have a clean attempt at starting it.
We should fix this.

While looking into the issue, I see this code in `Silo.Terminate()` and hope someone can explain it to me. Why can we not terminate a silo which is not in the "Running" state?
```C#
if (SystemStatus.Current.Equals(SystemStatus.Stopping) || 
    SystemStatus.Current.Equals(SystemStatus.ShuttingDown) || 
    SystemStatus.Current.Equals(SystemStatus.Terminated))
{
    stopAlreadyInProgress = true;
    // Drop through to wait below
}
else if (!SystemStatus.Current.Equals(SystemStatus.Running))
{
    throw new InvalidOperationException(String.Format("Calling Silo.{0} on a silo which is not in the Running state. This silo is in the {1} state.", operation, SystemStatus.Current));
}
else
{
    if (gracefully)
        SystemStatus.Current = SystemStatus.ShuttingDown;
    else
        SystemStatus.Current = SystemStatus.Stopping;
}
```

If we could just remove that exception, then calls to `Silo.Stop()`/`Silo.Shutdown()` could successfully clean up the static members.

Of course, another way to fix this problem is to use DI pervasively and kill all stateful statics.

ReubenBond:
Fixes #2119 and adds a simple test to demonstrate the issue.

### DIFF
--- a/src/OrleansRuntime/Silo/SiloHost.cs
+++ b/src/OrleansRuntime/Silo/SiloHost.cs
@@ -136,6 +136,7 @@ namespace Orleans.Runtime.Host
         {
             Utils.SafeExecute(UnobservedExceptionsHandlerClass.ResetUnobservedExceptionHandler);
             Utils.SafeExecute(LogManager.UnInitialize);
+            Utils.SafeExecute(GrainTypeManager.Stop);
         }
 
         /// <summary>
--- a/test/Tester/SiloInitializationTests.cs
+++ b/test/Tester/SiloInitializationTests.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Globalization;
+using System.Reflection;
+using Orleans.Runtime.Configuration;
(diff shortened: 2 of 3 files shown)


sergeybykov:
LGTM

gabikliot:
Lgtm.

PR #2133, opened 2016-09-09
Fix synchronization bug in Orleans/Async/BatchWorker 

sebastianburckhardt:
A rare race on the currentWorkCycle field can cause NullPointerException.

The reason is that the delegate does not properly capture the current value of currentWorkCycle under the lock, but the delegate executes after the lock is released, at which point the value of currentWorkCycle may have changed already. 

The fix is to manually capture the current value. This is more efficient anyway. Not sure why I ever used a delegate in the first place.

### DIFF
--- a/src/Orleans/Async/BatchWorker.cs
+++ b/src/Orleans/Async/BatchWorker.cs
@@ -59,7 +59,8 @@ namespace Orleans
         // on the same task scheduler
         private void CheckForMoreWork()
         {
-            Action signalThunk = null;
+            TaskCompletionSource<Task> signal = null;
+            Task task_to_signal = null; ;
 
             lock (this)
             {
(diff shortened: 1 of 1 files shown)


src/Orleans/Async/BatchWorker.cs
```diff
@@ -59,7 +59,8 @@ private void Start()
         // on the same task scheduler
         private void CheckForMoreWork()
         {
-            Action signalThunk = null;
+            TaskCompletionSource<Task> signal = null;
+            Task task_to_signal = null; ;
```
jdom: nit picking: name should be taskToSignal (and also there's an extra ';' )

src/Orleans/Async/BatchWorker.cs
```diff
@@ -86,8 +86,8 @@ private void CheckForMoreWork()
             }
 
             // to be safe, must do the signalling out here so it is not under the lock
-            if (signalThunk != null)
-                signalThunk();
+            if (signal != null)
+                signal.SetResult(task_to_signal);
```
jdom: super nit picking (only if you tackle the other one): no need to have an `if`, just do `signal?.SetResult(taskToSignal);`. In this particular case there can be no race between the if check and the method call, but it's a good practice in general.

jdom:
LGTM :+1: 
I had some very minor comments, so if it's easy to address, I'll wait for it, otherwise I can just merge, is no big deal.

sebastianburckhardt:
Ok, I have made those minor changes.

PR #2134, opened 2016-09-10
CodeGen: Emit & load debug symbols if built in Debug mode.

ReubenBond:
When using runtime code generation, currently assemblies are compiled and loaded without debug symbols.

In order to slightly improve the debugging situation, this PR causes the runtime code generator to emit debug symbols and load them alongside generated assemblies.

**Approach:** The crux of this change simply instructs Roslyn to emit debug symbols (see `compilation.Emit` calls in `CodeGeneratorCommon`) and then passes those debug symbols to `Assembly.Load`.

In order to make this work during test runs, where we are caching generated assemblies to improve performance, we need to alter `ICodeGeneratorCache` to work with symbols as well as just raw assemblies. To achieve this we change uses of `byte[]` to `GeneratedAssembly`, which holds assembly + symbols.

### DIFF
--- a/src/OrleansCodeGenerator/CodeGeneratorCommon.cs
+++ b/src/OrleansCodeGenerator/CodeGeneratorCommon.cs
@@ -46,13 +46,16 @@ namespace Orleans.CodeGenerator
         /// <param name="assemblyName">
         /// The name for the generated assembly.
         /// </param>
+        /// <param name="emitDebugSymbols">
+        /// Whether or not to emit debug symbols for the generated assembly.
+        /// </param>
         /// <returns>
         /// The raw assembly.
         /// </returns>
         /// <exception cref="CodeGenerationException">
         /// An error occurred generating code.
         /// </exception>
-        public static GeneratedAssembly CompileAssembly(GeneratedSyntax generatedSyntax, string assemblyName)
+        public static GeneratedAssembly CompileAssembly(GeneratedSyntax generatedSyntax, string assemblyName, bool emitDebugSymbols)
         {
             // Add the generated code attribute.
             var code = AddGeneratedCodeAttribute(generatedSyntax);
@@ -90,7 +93,7 @@ namespace Orleans.CodeGenerator
                     .WithOptions(options);
 
             var outputStream = new MemoryStream();
-            var symbolStream = RuntimeVersion.IsDebugBuild ? new MemoryStream() : null;
+            var symbolStream = emitDebugSymbols ? new MemoryStream() : null;
             try
             {
                 var compilationResult = compilation.Emit(outputStream, symbolStream);
--- a/src/OrleansCodeGenerator/CodeGeneratorCommon.cs
+++ b/src/OrleansCodeGenerator/CodeGeneratorCommon.cs
@@ -9,6 +9,7 @@ namespace Orleans.CodeGenerator
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
+
     using Orleans;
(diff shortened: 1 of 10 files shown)


sergeybykov:
@dotnet-bot test this please

src/OrleansCodeGenerator/CodeGeneratorCommon.cs
```diff
@@ -87,9 +89,13 @@ public static byte[] CompileAssembly(GeneratedSyntax generatedSyntax, string ass
                     .AddSyntaxTrees(code.SyntaxTree)
                     .AddReferences(assemblies)
                     .WithOptions(options);
+            MemoryStream symbolStream = null;
             using (var stream = new MemoryStream())
+#if DEBUG
+            using (symbolStream = new MemoryStream())
+#endif
```
jdom: Should we try to dynamically determine if it should use debug symbols if the *target* assembly is debuggable instead? Otherwise it will be almost always the case that people use the official nuget packages that were compiled in release mode.
There is a method that we are currently using called `RuntimeVersion.IsAssemblyDebugBuild` that *may* be useful for this test (after making it internal).

ReubenBond: Thanks @jdom, I've updated the PR to expose RuntimeVersion.IsDebugBuild & am using that instead

jdom: What I meant (not sure if it's possible, or even easy, otherwise it might not be worth it), is to generate debug symbols if you are codegening types for an assembly that (or a set of assemblies where at least one) is debuggable.
Also, we shouldn't change the behavior that prints whether Orleans itself is debuggable. Now this will print Debug if the entry assembly is debuggable, instead of whether Orleans itself is.

jdom:
BTW, for PRs where the goal is to merge as a single commit, please avoid rebasing/squashing unless there's merge conflicts, as it is a little harder to follow how the feedback was addressed. We can still squash from the GitHub UI at the time of merge.

src/OrleansCodeGenerator/CodeGeneratorCommon.cs
ReubenBond: Thanks, fixed: I'll emit symbols if any input assembly was built in debug mode & I've undone the behavioral change.

ReubenBond:
> BTW, for PRs where the goal is to merge as a single commit, please avoid rebasing/squashing unless there's merge conflicts, as it is a little harder to follow how the feedback was addressed.

Fair enough. I tend to rebase continually to ensure that the PR stays mergeable & I don't end up in some kind of merge hell. I'll leave changes unsquashed, though.

src/OrleansCodeGenerator/RoslynCodeGenerator.cs
```diff
@@ -89,8 +89,14 @@ public void GenerateAndLoadForAssemblies(params Assembly[] inputs)
             }
 
             var timer = Stopwatch.StartNew();
+            var emitDebugSymbols = false;
             foreach (var input in inputs)
             {
+                if (!emitDebugSymbols)
+                {
+                    emitDebugSymbols = RuntimeVersion.IsAssemblyDebugBuild(input);
```
jdom: we actually do need `|=` as it was before, don't we? Otherwise last assembly will define this value for the entire set

ReubenBond: I've now changed it back, but it's redundant because if emitDebugSymbols was true, then this block would not execute. |= is clearer

jdom: Oh, sorry, you are right, I did miss the if statement... I thought there was none.

PR #2135, opened 2016-09-10
CodeGen: fix generated DeepCopy method to call RecordObject earlier

ReubenBond:
Currently generated `DeepCopy` method incorrectly calls `RecordObject` after copying fields instead of before. This results in circular references involving being incorrectly serialized, resulting in a loss of referential equality.

The fix is to simply emit the call to `RecordObject` before the field-copying calls.

This was not failing in our test suite (`Serialize_CircularReference`) because the assembly holding the recursively defined type has the `[SkipCodeGeneration]` attribute, so a serializer was not being generated and so the fallback serializer was being used.

### DIFF
--- a/src/Orleans/Serialization/SerializationContext.cs
+++ b/src/Orleans/Serialization/SerializationContext.cs
@@ -3,6 +3,8 @@ using System.Collections.Generic;
 
 namespace Orleans.Serialization
 {
+    using System.Runtime.CompilerServices;
+
     /// <summary>
     /// Maintains context information for current thread during serialization operations.
     /// </summary>
@@ -60,7 +62,7 @@ namespace Orleans.Serialization
 
             public override int GetHashCode(object obj)
             {
-                return obj == null ? 0 : obj.GetHashCode();
+                return obj == null ? 0 : RuntimeHelpers.GetHashCode(obj);
             }
         }
 
--- a/src/OrleansCodeGenerator/SerializerGenerator.cs
+++ b/src/OrleansCodeGenerator/SerializerGenerator.cs
@@ -283,12 +283,6 @@ namespace Orleans.CodeGenerator
                                 SF.VariableDeclarator("result")
                                     .WithInitializer(SF.EqualsValueClause(GetObjectCreationExpressionSyntax(type))))));
 
-                // Copy all members from the input to the result.
-                foreach (var field in fields)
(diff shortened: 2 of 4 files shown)


src/Orleans/Serialization/SerializationContext.cs
```diff
@@ -60,7 +62,7 @@ public override bool Equals(object x, object y)
 
             public override int GetHashCode(object obj)
             {
-                return obj == null ? 0 : obj.GetHashCode();
+                return obj == null ? 0 : RuntimeHelpers.GetHashCode(obj);
```
ReubenBond: Not strictly required for this change, but a potential bug. Just as we use `object.ReferenceEquals` above this, we should call `RuntimeHelpers.GetHashCode` rather than `obj.GetHashCode` so that we are not dependent on user-defined implementations of GetHashCode. It's not likely to cause an issue, but it's possible.

test/TesterInternal/Serialization/BuiltInSerializerTests.cs
```diff
@@ -21,6 +21,8 @@
 
 namespace UnitTests.Serialization
 {
+    using TestGrainInterfaces;
+
     /// <summary>
```
ReubenBond: Instead of redefining `CircularTest1`, we use the implementation in `TestGrainInterfaces` because that implementation has a generated serializer whereas this assembly does not, due to `[assembly: SkipCodeGeneration]` attribute.

src/OrleansCodeGenerator/SerializerGenerator.cs
```diff
@@ -304,6 +298,12 @@ private static MemberDeclarationSyntax GenerateDeepCopierMethod(Type type, List<
                         recordObject.Invoke(currentSerializationContext)
                             .AddArgumentListArguments(SF.Argument(originalVariable), SF.Argument(resultVariable))));
 
+                // Copy all members from the input to the result.
+                foreach (var field in fields)
+                {
+                    body.Add(SF.ExpressionStatement(field.GetSetter(resultVariable, field.GetGetter(inputVariable))));
+                }
+
                 body.Add(SF.ReturnStatement(resultVariable));
```
ReubenBond: Meat of the change: call `RecordObject` before copying the fields.

PR #2137, opened 2016-09-11
Optimize deserialization

dVakulen:
Continuation of the https://github.com/dotnet/orleans/pull/2001. 

On running of the serialization benchmark with  [IL based serializer](https://github.com/dotnet/orleans/tree/feature-il-serializer) the CheckLength method again appears at the top of the list:

dottraceview64_2016-09-11_18-57-01

Optimized for most common case of reading of the simple type from ArraySegment of 4 kilobytes size.

Benchmark shows 12% deserialization perfomance improvement.

### DIFF
--- a/src/Orleans/Serialization/BinaryTokenStreamReader.cs
+++ b/src/Orleans/Serialization/BinaryTokenStreamReader.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
 using System.Text;
 using Orleans.CodeGeneration;
@@ -27,6 +28,7 @@ namespace Orleans.Serialization
         private int currentSegmentOffset;
         private int currentSegmentCount;
         private int totalProcessedBytes;
+        private int currentSegmentOffsetPlusCount;
         private readonly int totalLength;
 
         private static readonly ArraySegment<byte> emptySegment = new ArraySegment<byte>(new byte[0]);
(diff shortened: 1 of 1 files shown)


sergeybykov:
Thanks you, @dVakulen!

dVakulen:
There's possibility to optimize serialization module by ~30%, but it will require relatively big efforts, which 
in the light of upcoming replacement of this module doesn't seems justified.

sergeybykov:
>There's possibility to optimize serialization module by ~30%,

The current plan is not to replace our serialization, and only add a replacement for `BinaryFormatter`. So such an optimization may still be relevant. @ReubenBond is actively working on that.

dVakulen:
Yes, @ReubenBond not only implemented IL-based serializer, but also digged into that area for optimizations, so I'll refrain from making changes in it for a while.

ReubenBond:
@dVakulen Optimizations to ILSerializer aren't my primary goal at this stage. My main intention is just to investigate some perf issues when performing nightly stress tests. I ended up writing a bunch of optimization/fixes for it, as you can see in my [il-serializer](https://github.com/ReubenBond/orleans/tree/il-serializer) branch.

If you want to dig into optimizations, then I'm happy to help. One example is a minor perf improvement which I'm not pursuing at this stage is [this `ByteArrayBuilder` optimization](https://github.com/ReubenBond/orleans/commit/97af49d97a11106d798710599c117bcb2220fc78) to avoid calls to `Buffer.BlockCopy` - but it would require that we compile with `/unsafe`, so maybe that makes some people uneasy.

PR #2146, opened 2016-09-13
EventHub receiver metrics reporting abstraction

jason-bragg:
Cleaned up EventHub receiver metrics reporting and put reporting behind an abstraction.


gabikliot:
LGTM.

PR #2147, opened 2016-09-13
Updated ThreadTrackingStatistic to use PerfCounterTelemetryConsumer

galvesribeiro:
More work for #2122

### DIFF
--- a/src/Orleans/Statistics/ThreadTrackingStatistic.cs
+++ b/src/Orleans/Statistics/ThreadTrackingStatistic.cs
@@ -1,7 +1,5 @@
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
-using System.Threading;
 
 namespace Orleans.Runtime
 {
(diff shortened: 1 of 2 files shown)


galvesribeiro:
CC: @jason-bragg / @jdom

src/Orleans/Statistics/ThreadTrackingStatistic.cs
```diff
@@ -163,7 +162,7 @@ public void OnStartProcessing()
                 firstStart = false;
                 if (StatisticsCollector.CollectContextSwitchesStats)
                 {
-                    TrackContextSwitches();
+                    logger.IncrementMetric(metricName);
```
jason-bragg: I don't think this is a good way to trigger logging these metrics.  In general, all we are doing here is telling the telemetry that there is a metric of this name that is being incremented, which is not the case.  We are then relying on the Orleans perf counter metric consumer to detect this metric and perform special processing on it.  This is quite kludgy.

The problem is that we don't have anything to log here, we just need to signal something else to retrieve and log the data, so this is really a signal to an framework extension point.  I suspect that DI could help.

Suggest something like:
1) Create public IThreadTrackingStatistic interface with the public surface of ThreadTrackingStatistic.
2) Move ThreadTrackingStatistic to OrleansTelemetryConsumers.Counters as an implementation of IThreadTrackingStatistic.
3) Add extension function to register ThreadTrackingStatistic type in the service collection for DI.
4 Update Orleans to get IThreadTrackingStatistic form DI.

tsibelman: I second your coment what will happen if I implement another telemetry consumer ? And I will get just a single call to Context Switches/sec, what is it means for me ? 

Obviosly telemetry data coming into interface should make sense to all consumers and not to be custom fit for one specific implementation.

It just one specific example of a more general leaking abstraction in design of IMetricTelemetryConsumer.  Orleans has many metrics constructed from other metrics but IMetricTelemetryConsumer don't have any general interface to implement this.

galvesribeiro:
Hey guys, this PR is outdated with the lots of changes recently made by @jason-bragg 

Should we close it?

jason-bragg:
Yeah, I've a work item to create a tracking issue for this and close it out.  I'll do that today.

jason-bragg:
Closing as alternative solutions are preferred and PR is out of date.
Added tracking issue "Remove performance counters from ThreadTrackingStatistic #2609"

PR #2151, opened 2016-09-14
Cleanup in preparation for .NET Standard

jdom:
These are small fixes in preparation for the .NET Standard projects


jdom:
If it looks good for whoever is merging it, please do not squash into a single commit, since they are all different small fixes

xiazen:
LGTM

PR #2155, opened 2016-09-14
Add build dependency to ServiceFabric utilities project

jdom:
This is so it builds before TesterInternal, and the post build script executes with that assembly already created.
It was breaking our VSO builds


ReubenBond:
I'm not quite sure why this is necessary, there should be no relation between those two projects. What am I missing?

Sorry for breaking VSO

jdom:
@dotnet-bot test this please

jdom:
Don't worry, it's not your fault, is this different CI build we have in VSO. If you look at TesterInternal post build scripts, it ends up generating all the nuget packages when ran in VSO. This was because TesterInternal used to be the project that (transitively) referenced all other projects. By this time, because TesterInternal does not transitively depend on the new ServiceFabric utilities project, then the nuget package creation fails.
So this is not a project reference that I added, just a build dependency for the project

ReubenBond:
Thanks - I see you also fixed the missing x64 lines for `OrleansTelemetryConsumers.Counters`.

Issue #2156, opened 2016-09-15
Type serialization does not handle RuntimeType or un-registered types


ReubenBond:
There are two issues with serialization of `Type` today:

* Currently we register a specialized serializer for `Type` like so:
```C#
Register(
  typeof(Type),
  BuiltInTypes.CopyType,
  BuiltInTypes.SerializeType,
  BuiltInTypes.DeserializeType);
```
At runtime, types are typically an instance of `RuntimeType`, which is an internal class, rather than `Type`, which is an abstract class.

* The type serializer which we register is only able to serialize types which have previously been registered by a call to `SerializationManager.Register(...)`, meaning that we are limited to a subset of types.

The first issue can be fixed by additionally registering the same serializer for `typeof(Type).GetType()` to get the `RuntimeType`.

The latter requires more thought. The easiest solution is to use `type.AssemblyQualifiedName` + `Type.GetType()`. That would serialize assembly version also, which is likely undesirable. Another approach is to use a hybrid of the current model (pre-registered types) with a fallback to `GetType`/`AssemblyQualifiedName`.

ReubenBond:
Currently, type serialization works as follows:

1. If the runtime type of the object being serialized is the same as the type in the type definition, then we write a 1-byte `ExpectedType` token and move on.
2. If the runtime type is a well-known type, such as a primitive, certain common .NET types (DateTime, Guid), or a common Orleans type (GrainId, SiloAddress), then it is encoded using 2 bytes: `SpecifiedType` + a one byte type identifier.
3. Otherwise, the type is encoded as `SpecifiedType` + `NamedType` + full type name (excluding assembly) of the type.

The current approach is generally good, but it does not handle unexpected types well because
* The implementation requires the type to be pre-registered by a call to `SerializationManager.Register`.
* The serialization information doesn't include any way to identify the assembly which a type comes from.

Additionally, if we have a `List<object>` and the runtime type of the object is either `Dog` or `Cat`, the full type name is serialized for each element.

Orleans has a system for serializing cyclic references which works something like this:
* In the serialization context, there is a dictionary which stores the start index of each object which is being serialized.
* When serializing an object, that dictionary is first checked and if the object is present (using reference equality), only a reference to its offset is serialized, otherwise the object is serialized.
* When deserializing an object, the same process happens in reverse: every deserialized object is added to the dictionary as soon as it's instantiated so that future references can lookup the dictionary and point to the existing instance.

[...]

PR #2157, opened 2016-09-15
Serialize Type & RuntimeType using AssemblyQualifiedName

ReubenBond:
This is a fix + test case for #2156.

It is a work-in-progress & is not expected to be committed as-is, since the presence of version & key info in assembly names is not ideal for most users.


sergeybykov:
@ReubenBond What should we do with this one?

PR #2158, opened 2016-09-15
Remove MarshalByRefObject where it is not needed

jdom:
(and not supported in .NET Standard)


jdom:
Reviewed the code and these are not being accessed across AppDomains, and confirmed by running this commit through our entire Functional suite.

PR #2159, opened 2016-09-15
Optimize callbacks timeouts

dVakulen:
Creating and disposing of timer per two-way message costs 0.5 seconds per 1 million of callbacks. 

Linear increment of due date allows using of the queue for storing callbacks, with single timer for checking it. 
In some cases callback instance may be promoted to next GC gen, which wouldn't had happen otherwise, but it's doesn't matter much.  

Local benchmark shows up to 7% throughput increase, but I'm not sure which effect it will have on the real benchmarks.


gabikliot:
I will provide some feedback later on about that.

src/Orleans/Runtime/CallbackData.cs
```diff
@@ -207,5 +205,124 @@ public TimeSpan RequestedTimeout()
...
+                            {
+                                callbacks.Dequeue();
+                                continue;
+                            }
+
+                            if (callback.dueTime < now)
+                            {
+                                callbacks.Dequeue();
+                                callback.OnTimeout();
+                                _onTimeout(callbacks, callback);
```
jdom: is it guaranteed that these callbacks are extremely fast? otherwise the contention in queueLock might cause the CPU to spin very aggressively

src/Orleans/Utils/FastLock.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Orleans.Runtime
+{
+    // About 2x faster than ordinary lock on mostly single threaded locking
+    internal class FastLock
```
jdom: might want to consolidate with `AsyncSerialExecutor.InterlockedExchangeLock` to avoid having 2 similar implementations

src/Orleans/Runtime/CallbackData.cs
```diff
@@ -150,22 +153,18 @@ public void DoCallback(Message response)
...
+            if (timeouter != null)
             {
-                var tmp = timer;
-                if (tmp != null)
+                try
                 {
-                    timer = null;
-                    tmp.Dispose();
+                    callbackTimeouter = null;
+                    timeouter.Dispose();
```
jdom: wouldn't this dispose the timeouter for multiple CallbackData instead of just this one since this is a thread static?

src/Orleans/Runtime/CallbackData.cs
```diff
@@ -69,21 +76,18 @@ public void StartTimer(TimeSpan time)
                 timeSinceIssued.Start();
             }
 
+            if (callbackTimeouter == null) callbackTimeouter = new CallbackTimeouter(config.ResendOnTimeout);
```
jdom: Shouldn't you use `Interlocked.CompareExchange`?

gabikliot:
We definitely did think about using "timer wheel" or something similar to optimize those timers here, so I think the general direction of looking into it is sure worthy, but I would go a bit differently about it.
I think there are approaches:
1) just replace our SafeTimer with simplest .NET Timer. That will allow to optimize out a couple of layers of timer wrappers, which will save both hopefully a lot of memory (there is one timer per msg) and also some extra bookeeping we do in the safe timer. That is a safest option and you will sure gain benefits, there are no tradeoff here.
2) Try something like your queue. But if we go that route, we should **do it appropriately**. What I mean by that is make it a separate class, well tested in isolation, well perf. bechmarked in isolation (compare it to vanilla simple .NET Timer, not to Orleans safe Timer). You are building a general utility here, so lets do it the right way. After that is done and IF it outperforrms .NET Timer, then change CallbackData to use it. 

My intuition is that the 1st approach will get most of the benefits (the reduced cost of GC will be significant I hope), but I might be wrong.

dVakulen:
The 95% of time is spent in starting of the simplest .NET Timer, so replacing SafeTimer with it woun't gain almost anything. I surely did benchmarked quite a few possible ways of optimizing this.. So I'll add benchmarks and tests for the extracted timer wheel.

src/Orleans/Runtime/CallbackData.cs
dVakulen: No, the access here is singlethreaded.

src/Orleans/Runtime/CallbackData.cs
dVakulen: They should be, but it'll be safer to release lock for this activities.

src/Orleans/Utils/FastLock.cs
dVakulen: Agreed

src/Orleans/Runtime/CallbackData.cs
dVakulen: Yes, will fix

src/Orleans/Runtime/CallbackData.cs
jdom: Hmm, OK, but remember this is the thread static field, not just an instance field. I'm a little surprised that access is completely single threaded, but it could be, just wanted to give you a heads up so you can double  check

gabikliot:
Ok, let's put this  timer wheel logic in a separate file, as part of general utilities. It should be usable  and testable in a standalone mode.

dVakulen:
Extracted TimerWheel. 

Benchmark results: 
               Method |        Median |     StdDev |  Repeats |
--------------------- |-------------- |----------- |--------- |
          DotNetTimer | 3,516.5780 ms | 32.5467 ms | 10000000 |
  ParallelDotNetTimer | 4,149.2678 ms | 27.1073 ms | 10000000 |
 TimerWheelCommonCase |   508.0843 ms |  3.2149 ms | 10000000 |
    TimerWheelTimeout | 1,626.1717 ms | 73.9679 ms | 10000000 |
          DotNetTimer | 1,066.7433 ms |  1.2738 ms |  3000000 |
  ParallelDotNetTimer | 1,241.1823 ms |  4.0009 ms |  3000000 |
 TimerWheelCommonCase |   190.7896 ms |  3.6939 ms |  3000000 |
    TimerWheelTimeout |   426.6386 ms | 25.4385 ms |  3000000 |

TimerWheel benchmark doing more work than .Net timer one, but still up to 8 times faster.

dVakulen:
@dotnet-bot test this please
(Test failures is non -related to this PR : System.Net.Sockets.SocketException : An attempt was made to access a socket in a way forbidden by its access permissions)

src/Orleans/Runtime/CallbackData.cs
```diff
@@ -70,20 +93,18 @@ public void StartTimer(TimeSpan time)
             }
 
             TimeSpan firstPeriod = timeout;
-            TimeSpan repeatPeriod = Constants.INFINITE_TIMESPAN; // Single timeout period --> No repeat
             if (config.ResendOnTimeout && config.MaxResendCount > 0)
             {
-                firstPeriod = repeatPeriod = timeout.Divide(config.MaxResendCount + 1);
+                resendPeriod = firstPeriod = timeout.Divide(config.MaxResendCount + 1);
+                DueTime = DateTime.UtcNow.Add(firstPeriod);
```
gabikliot: you can collapse this condition and the n on resend case into one, with single timer wheel var, and only difference via timerWheelShouldReEnqueue/shouldDequeue values. No need for code dup.

src/Orleans/Runtime/CallbackData.cs
```diff
@@ -100,14 +121,16 @@ public void OnTimeout()
             OnFail(msg, error, "OnTimeout - Resend {0} for {1}", true);
         }
 
+        public DateTime DueTime { get; private set; }
```
gabikliot: please put at the top of the class, near other fields.

src/Orleans/Utils/TimerWheel.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        // Crawls through the callbacks and timeouts expired ones
+        private void CheckQueue()
+        {
+            var now = DateTime.UtcNow;
+            while (true)
+            {
+                try
+                {
+                    _queueLock.Get();
```
gabikliot: why not to use a simple lock statement?

src/Orleans/Utils/InterlockedExchangeLock.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            return Interlocked.Exchange(ref lockState, Locked) != Locked;
+        }
+        
+        public void Get()
+        {
+            if (TryGet())
+                return;
+
+            SpinWait spinWait = new SpinWait();
```
gabikliot: why this and  not to use any of the standard .NET sync primitives, like event or mutex or anything else with similar semantics.

src/Orleans/Runtime/CallbackData.cs
```diff
@@ -70,20 +93,18 @@ public void StartTimer(TimeSpan time)
...
-            DisposeTimer();
-            timer = new SafeTimer(TimeoutCallback, null, firstPeriod, repeatPeriod);
 
-        }
-
-        private void TimeoutCallback(object obj)
-        {
-            OnTimeout();
+            DueTime = DateTime.UtcNow.Add(firstPeriod);
+            if (callbacksWheel == null) callbacksWheel = new TimerWheel<CallbackData>(_callbacksWheelCheckPeriod, shouldDequeue: timerWheelShouldDequeue);
```
gabikliot: please use {}, I got confused by reading this.

src/Orleans/Utils/TimerWheel.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Threading.Tasks;
+using Orleans.Runtime;
+
+namespace Orleans.Runtime
+{
+    // Removes the need in timer per item, optimized for singlethreaded consumption. 
+    internal class TimerWheel<T> : IDisposable where T : ITimebound
+    {
+        private const int MaxItemsPerLock = 10;
+        private readonly Queue<T> _entries = new Queue<T>();
```
gabikliot: One negative caveat of your impl. is that in the fast path case, you will be holding  to the element in this queue for full 30 sec, while in old impl. we disposed timer immediately upon succesfull reply. Therefore, you will be holding callback datas for 30 sec, which can increase memory footprint.
The solution is either to support dispose/unregister, or store weak reference to T.

src/Orleans/Utils/TimerWheel.cs
```diff
@@ -0,0 +1,128 @@
...
+        // Crawls through the callbacks and timeouts expired ones
+        private void CheckQueue()
+        {
+            var now = DateTime.UtcNow;
+            while (true)
+            {
+                try
+                {
+                    _queueLock.Get();
+                    for (int i = 0; i < MaxItemsPerLock; i++)
```
gabikliot: why MaxItemsPerLock?

The basic mechanism is to spin all the time, keep checking the queue all the time, without even a sleep in between.
How is that better then what we have now?
The more advanced impl. will maintain list of differences, so that the top of the queue is the earliest and sleep untill that. I am not suggesting to do that, the overhead of maintaining the sorted queue may be larger then the current impl, BUT what is not clear to me is how the suggested always spinning impl. can be better.

gabikliot:
My 2 main pieces of feedback:
a) I am again finding my self trying to decipher the implementation. It should be documented instead.
2) the spinning seems to be very naive impl., and I cant see how it will be better then the current impl.

src/Orleans/Runtime/CallbackData.cs
```diff
@@ -13,30 +15,51 @@ internal interface ITimebound
         /// This method is called by the timer when the time out is reached.
         /// </summary>
         void OnTimeout();
-        TimeSpan RequestedTimeout();
+
+        DateTime DueTime { get; }
```
jdom: I would avoid this, since it could mean that each item could change the due time dynamically, or even have different immutable values, which would not be supported by the timer wheel. Instead you could have this as an implementation detail of the timer wheel (basically instead of having a Queue<ITimebound> have a Queue<Tuple<ITimeboud>> (or even better a struct instead of Tuple, and potentially a weak reference to the ITimebound.
With this change, then you can optimize for the idle case too and avoid having a timer in the first place, and have a never ending async task that does `await Task.Delay(time to next expiration in the queue)` (because DueTime is not dynamic and the queue now can enforce an ever increasing due time) or explicitly delay a few milliseconds to yield the thread a little.

src/Orleans/Utils/TimerWheel.cs
dVakulen: Unfortunately comment about this was lost during timer wheel extracting. On the common path callbacks on average gets completed in order they were queued, the difference may be between local and remote calls, but it doesn't matter much. There's a [check](https://github.com/dotnet/orleans/pull/2159/files#diff-24969f74ffdd4f9fe9cc680c955cc0ceR42) for already fired callback  , thus it will be removed from the queue in tens to hundreds of milliseconds after successful reply. I've considered using weak references, but it adds to much overhead, so it should be an last resort.

src/Orleans/Utils/TimerWheel.cs
dVakulen: I don't expect much contention here, and for such scenarios this kind of lock provides 2x better performance. Plus it allows more fine grained control - for example [executing potentially long function](https://github.com/dotnet/orleans/pull/2159/files#diff-a694ce799337a9585c6bb404e7ca2339R80) outside of the lock.

src/Orleans/Utils/TimerWheel.cs
dVakulen: There is sleep in between - this  logic is being executed in thread pool with period of 70 ms. MaxItemsPerLock was added to allow other thread make progress, but it seems that it's better to remove it.
In order to get this ordered by time with top of the queue is the earliest I've created [two timer wheels ](https://github.com/dotnet/orleans/pull/2159/files#diff-24969f74ffdd4f9fe9cc680c955cc0ceR38) - one for ordinary messages, and one for resendable as their periods [differs](https://github.com/dotnet/orleans/blob/master/src/Orleans/Runtime/CallbackData.cs#L76).

src/Orleans/Utils/InterlockedExchangeLock.cs
dVakulen: Mainly because this provides the best performance for 1 producer - 1 consumer scenarios. For example such construction is used in [ConcurrentQueue](https://github.com/dotnet/corefx/blob/master/src/System.Threading.Tasks.Dataflow/src/Internal/ConcurrentQueue.cs#L567).

src/Orleans/Runtime/CallbackData.cs
dVakulen: Yes, that's the alternative approach that requires using of the weak references which incurs so much overhead (500 ms per 2 million callbacks) so I'm not convinced that it can be faster than current implementation. 

Agreed about making due time immutable.

src/Orleans/Utils/InterlockedExchangeLock.cs
gabikliot: But we don't have 1 producer - 1 consumer. we have multiple producers - 1 consumer . and the consumer is long running one - it scans the whole queue under lock.

dVakulen: We do have 1 producer - 1 consumer because the timer wheels are [ThreadStatic](https://github.com/dotnet/orleans/pull/2159/files#diff-24969f74ffdd4f9fe9cc680c955cc0ceR37).

dVakulen:
Can this be perf tested, please, so that the further direction could be seen?

ReubenBond:
@dVakulen I can run perf tests if you wish. Will this change would have a noticeable impact above the noise?

dVakulen:
@ReubenBond 
>Local benchmark shows up to 7% throughput increase, but I'm not sure which effect it will have on the real benchmarks.

ReubenBond:
@dVakulen Is that 7% measured when making grain calls or in microbenchmarks?

Could you rebase? I'll kick off a test run.
image

dVakulen:
@ReubenBond It was on making grain calls.

ReubenBond:
@dVakulen the test ran at about 50% of the usual TPS - Maybe something was wrong with the way I ran it, so I'll try a full run tonight after the scheduled nightly run is finished

dVakulen:
Thank you, @ReubenBond. Any news on this?

ReubenBond:
Thanks for the poke. The full run had the same issue, but it looks like it was an external problem, so I've kicked off another run.

ReubenBond:
Initial results look good. Here's the headline TPS metric:
Before: 208689 TPS (average of several runs)
After: 223409 TPS
So that's an improvement of 7%.
I'm going to kick off another run to see if it corroborates that improvement.

dVakulen:
@ReubenBond Has another run confirmed the improvement?

ReubenBond:
@dVakulen subsequent runs have shown performance deltas of ~+2% and -8%. The -8% may've been an anomaly. I'm running again. It seems there's a lot of noise :(

dVakulen:
@ReubenBond If there's no clear perf gain, I think it would be better to just close this PR.

dVakulen:
As this one reportedly haven't shown any performance improvements in real benchmarks - closing it.

dVakulen:
It seems that there were some issues with the hardware (https://gitter.im/dotnet/orleans?at=5823d287c2f2cf7275f01d9f, https://github.com/dotnet/orleans/pull/2395#issuecomment-259755359), so let's perf test this once again.

ReubenBond:
The latest run shows a degradation in performance of about 7.5% :(

dVakulen:
Thank you, @ReubenBond. Closing then.

PR #2160, opened 2016-09-16
More .Net Standard compatibility fixes

jdom:


jdom:
Please don't squash and merge, since there are small but different fixes in each commit

src/Orleans/Configuration/ClientConfiguration.cs
```diff
@@ -175,7 +175,11 @@ public ClientConfiguration()
             NetInterface = null;
             Port = 0;
             DNSHostName = Dns.GetHostName();
+#if NETSTANDARD
+            DeploymentId = Environment.MachineName;
+#else
```
ReubenBond: Maybe we should just change it to MachineName in both cases, just because the code is ugly otherwise

ReubenBond: LGTM anyway

jdom: I can do that. Nevertheless in theory there could be a side effect: if several machines didn't override the default, but since they are being run by the same service account, then all of their deployment IDs align. Not sure how important would be not to break that scenario though.
Anybody else have other opinions? or maybe a different default (or even constant) value?

jdom: /cc @sergeybykov

sergeybykov: I think we should set it to empty string in both cases, just like we do for `DataConnectionString`.

jdom: Done, rebased since I wanted to keep the commits separate on not force you to squash when merging. Setting this as the empty string was the only change I did since the last time reviewed

PR #2162, opened 2016-09-16
IL-based fallback serializer

ReubenBond:
This is an implementation of a *fallback* serializer which is intended to replace `BinaryFormatter` as a part of the .NET Core initiative.

This serializer is very similar to the current Roslyn-based serializers except that instead of emitting C# code, it emits IL directly at runtime, allowing it to access private members and types.

The bulk of the logic is in `IlBasedSerializerGenerator`, which I hope is easy enough to follow.

IL generation is abstracted via a simple facade called `IlDelegateBuilder<TDelegate>` which is based on on .NET's `ILGenerator`.
The usage of the class is as follows:
1. Instantiate the builder with the type of delegate being built (eg, `DeepCopier`).
2. Call methods on the builder to add instructions to it.
3. Call `Build()` to create the final delegate.

Please let me know what you think.

### DIFF
--- a/src/Orleans/Serialization/SerializationManager.cs
+++ b/src/Orleans/Serialization/SerializationManager.cs
@@ -138,7 +138,7 @@ namespace Orleans.Serialization
 
         #region Static initialization
 
-        public static void InitializeForTesting(List<TypeInfo> serializationProviders = null, IExternalSerializer fallback = null)
+        public static void InitializeForTesting(List<TypeInfo> serializationProviders = null, TypeInfo fallbackType = null)
         {
             try
             {
@@ -146,7 +146,7 @@ namespace Orleans.Serialization
                 BufferPool.InitGlobalBufferPool(new MessagingConfiguration(false));
                 RegisterSerializationProviders(serializationProviders);
                 AssemblyProcessor.Initialize();
-                fallbackSerializer = fallback ?? GetFallbackSerializer();
+                fallbackSerializer = GetFallbackSerializer(fallbackType);
             }
             catch (ReflectionTypeLoadException ex)
             {
@@ -154,10 +154,10 @@ namespace Orleans.Serialization
             }
         }
 
-        internal static void Initialize(List<TypeInfo> serializationProviders, IExternalSerializer fallback = null)
+        internal static void Initialize(List<TypeInfo> serializationProviders, TypeInfo fallbackType = null)
         {
             RegisterBuiltInSerializers();
-            fallbackSerializer = fallback ?? GetFallbackSerializer();
+            fallbackSerializer = GetFallbackSerializer(fallbackType);
 
             if (StatisticsCollector.CollectSerializationStats)
             {
@@ -1898,13 +1898,22 @@ namespace Orleans.Serialization
             return retVal;
         }
 
-        private static IExternalSerializer GetFallbackSerializer()
+        private static IExternalSerializer GetFallbackSerializer(TypeInfo fallbackType)
         {
+            IExternalSerializer serializer;
+            if (fallbackType != null)
+            {
+                serializer = (IExternalSerializer)Activator.CreateInstance(fallbackType.AsType());
+            }
+            else
+            {
 #if NETSTANDARD
-            var serializer = new IlBasedFallbackSerializer();
+                serializer = new IlBasedFallbackSerializer();
 #else
-            var serializer = new BinaryFormatterSerializer();
+                serializer = new BinaryFormatterSerializer();
 #endif
+            }
+
             serializer.Initialize(logger);
             return serializer;
         }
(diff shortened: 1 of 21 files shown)


sergeybykov:
>         private static IExternalSerializer GetFallbackSerializer(bool useJsonSerializer)

I think it's time to remove JSON as a fallback option. We know it's not robust enough to do the job.

---
Refers to: src/Orleans/Serialization/SerializationManager.cs:1881 in 468005f. [](commit_id = 468005f740a5c9974590d92352a45c8053631447, deletion_comment = False)

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -1894,7 +1890,7 @@ private static IExternalSerializer GetFallbackSerializer(bool useJsonSerializer)
 #if NETSTANDARD
                 throw new OrleansException("Can't use binary formatter as fallback serializer while running on .Net Core");
 #else
-                serializer = new BinaryFormatterSerializer();
+                serializer = new IlBasedFallbackSerializer(new IlBasedSerializers());
```
sergeybykov: Why is this not under `if NETSTANDARD`? Should we have a config flag for choosing `BinaryFormatterSerializer` vs. `IlBasedFallbackSerializer`, so that we can test the latter on full .NET?

ReubenBond:
@sergeybykov wrote:
> I think it's time to remove JSON as a fallback option. We know it's not robust enough to do the job.

I've got a branch where I've cleaned up JsonFallbackSerializer, but I didn't want to pollute this PR with those changes. Should I submit it as a PR shortly after we merge this?

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: For now, how about I change the code to the following?
```C#
if (useJsonSerializer)
{
    serializer = new OrleansJsonSerializer();
}
else
{
#if NETSTANDARD
    serializer = new IlBasedFallbackSerializer(new IlBasedSerializers());
#else
    serializer = new BinaryFormatterSerializer();
#endif
}
```
Then, in a following commit, I'll remove useJsonSerializer and have the correct serializer passed in through the `Initialize`/`InitializeForTesting` method, which we'll choose using a config option.

Alternatively, I can do it all in this PR

ReubenBond:
I've pushed commits which remove the UseStandardSerializer & UseJsonFallbackSerializer options. Many files were touched in the process, so I'd rather not squash them.

ReubenBond:
Fixed to work in vNext sln

jdom:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot test this please

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -137,30 +137,26 @@ public static object GetUninitializedObjectWithFormatterServices(Type type)
...
+                fallbackSerializer = fallback ?? GetFallbackSerializer();
             }
             catch (ReflectionTypeLoadException ex)
             {
                 throw ex.Flatten();
             }
         }
 
-        internal static void Initialize(List<TypeInfo> serializationProviders)
+        internal static void Initialize(List<TypeInfo> serializationProviders, IExternalSerializer fallback = null)
```
ReubenBond: Good question. At the moment I added a commit atop this to select the fallback serializer even on full .NET, but if you prefer a config option then how about I add one which is similar to the `SerializationProviders` config & is represented by a `TypeInfo`?

ReubenBond: The latest commit adds an option to `<Messaging>`. Used like so:
```xml
<Messaging>
  <FallbackSerializationProvider
    type="Orleans.Serialization.IlBasedFallbackSerializer, Orleans"/>
</Messaging>
```

sergeybykov: I htink this is good - it'll allow us to keep any existing tests unchanged and won't affect anybody, unless they explicitly specify the IL one as the fallback.

sergeybykov:
So this is ready for testing? First to prove that there's no impact by default, and then that everything passes with the IL-based serializer enabled?

ReubenBond:
Yep, this is ready for testing. I've kicked off another functionals run. We ran functionals with the IL-serializer enabled & they passed

sergeybykov:
Okay, I'll kick off the load tests tomorrow morning after nightly tests are done.

sergeybykov:
Perf test went fine. Please rebase, and I'll merge it.

galvesribeiro:
Good work on that @ReubenBond, thanks! 👍

ReubenBond:
Rebased, thanks :) Does anyone else want to check out the code before we merge?

PR #2163, opened 2016-09-16
Add ServiceFabric category to TestServiceFabric tests

ReubenBond:
Thanks for pointing this out, @jdom


jdom:
BTW, for the future, you can apply this attribute to the class itself when it applies to all. MSTest didn't allow it, but our custom attribute does, and we did not migrate our existing usages. 
I'll merge this since it does not contradict our usages, we still probably need to do a bigger pass and fix all in 1 go

PR #2164, opened 2016-09-16
Recover types from ReflectionTypeLoadException

ReubenBond:
There are several places where we enumerate all types in an assembly.
When a type cannot be initialized for some reason, a `ReflectionTypeLoadException` is thrown. Types can fail to initialize for a variety of reasons including failure to load referenced types or static constructor exceptions.
Currently we bail out at that point and do not process any types for that assembly.
`ReflectionTypeLoadException` has a `Types` property which contains an array of the types which were successfully loaded.

This change salvages those types which were successfully loaded.

### DIFF
--- a/src/Orleans/CodeGeneration/TypeUtils.cs
+++ b/src/Orleans/CodeGeneration/TypeUtils.cs
@@ -561,7 +561,7 @@ namespace Orleans.Runtime
             }
             catch (Exception exception)
             {
-                if (logger.IsWarning)
+                if (logger!= null && logger.IsWarning)
                 {
                     var message =
                         string.Format(
@@ -569,6 +569,13 @@ namespace Orleans.Runtime
                             assembly.FullName,
                             exception);
                     logger.Warn(ErrorCode.Loader_TypeLoadError_5, message, exception);
+                }
+                
+                var typeLoadException = exception as ReflectionTypeLoadException;
+                if (typeLoadException != null)
+                {
+                    return typeLoadException.Types?.Where(type => type != null).Select(type => type.GetTypeInfo()) ??
+                           Enumerable.Empty<TypeInfo>();
                 }
 
                 return Enumerable.Empty<TypeInfo>();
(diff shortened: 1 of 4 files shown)


src/Orleans/CodeGeneration/TypeUtils.cs
```diff
@@ -561,7 +561,7 @@ public static IEnumerable<TypeInfo> GetDefinedTypes(Assembly assembly, Logger lo
             }
             catch (Exception exception)
```
attilah: Why not just catching `ReflectionTypeLoadException` here and make others bubble up? This way we're swallowing all other exceptions if logging is not turned on to at least Warning level.

src/Orleans/CodeGeneration/TypeUtils.cs
```diff
@@ -561,7 +561,7 @@ public static IEnumerable<TypeInfo> GetDefinedTypes(Assembly assembly, Logger lo
...
             {
-                if (logger.IsWarning)
+                if (logger!= null && logger.IsWarning)
                 {
                     var message =
                         string.Format(
```
attilah: Can we move to use string interpolation when adding new code? It would be less to cleanup at a later stage.

src/Orleans/CodeGeneration/TypeUtils.cs
ReubenBond: The intention was to keep the same behavior. Should non-`ReflectionTypeLoadException` exceptions bubble up?

src/Orleans/CodeGeneration/TypeUtils.cs
ReubenBond: I'll change it, thanks

src/Orleans/CodeGeneration/TypeUtils.cs
jdom: Yeah, I agree with Reuben, there's no reason to have a breaking change with this behavior now. Especially since we'll revamp assembly loading soon.

PR #2165, opened 2016-09-16
(TEST) Use IL serializers instead of Roslyn-generated serializers

ReubenBond:
I don't have access to the testing infrastructure, but if it's easy to kick off a test with this, I think it would be interesting to see:

1.  If it passes functional tests (otherwise maybe there's an issue with #2162 which is worth investigating)
2.  How this fares in stress/perf.

IL based serializer is in PR #2162


ReubenBond:
Closed because this is far from ready & I can run full tests now.

PR #2166, opened 2016-09-16
DI docs

attilah:


jdom:
Hehe, finally! 👍 LGTM
It is getting a tiny bit ahead of the upcoming PR, since #988 is not currently supported (as of today, things might change in a few days, hehe)

src/Documentation/Advanced-Concepts/Dependency-Injection.md
```diff
@@ -0,0 +1,75 @@
...
+Previously DI configuration was specified at the cluster node level, this was changed in the recent release. 
+
+## Configuring from Code
+
+It is possible to tell Orleans what `Startup` type you like to use with code based configuration. There is an extension method named `UseStartup` on the `ClusterConfiguration` class which you can use to do that.
+
+``` csharp
+var configuration = new ClusterConfiguration();
+
+configuration.UseStartup<MyStartup>();
```
jdom: My Startup doesn't match the name used in the XML config, nor the example below. Please make it consistent, by, for example, using the full name so it's clear it is a user type: `configuration.UseStartup<MyApplication.Configuration.Startup>();`

src/Documentation/Advanced-Concepts/Dependency-Injection.md
```diff
@@ -0,0 +1,75 @@
...
+Dependency Injection is currently supported only on the server side within Orleans.
+
+Orleans makes it possible to inject dependencies into application [Grains](../Getting-Started-With-Orleans/Grains.md).
+
+However Orleans supports every container dependent injection mechanisms, one of the most commonly used method is constructor injection.
+
+Theoretically any type can be injected which was previously registered in a [`IServiceCollection`](https://docs.asp.net/projects/api/en/latest/autoapi/Microsoft/Extensions/DependencyInjection/IServiceCollection/index.html) during Silo startup.
+
+Orleans also makes it possible to inject the following service into an application object instance:
+- `IGrainFactory`
```
jdom: I would remove this `IGranFactory` reference for now and we can add it later when it's actually available. Because if it doesn't make the cut for 1.3.0-beta2, then this would not be usable. And it is not usable for people in 1.2.3 yet anyway, so it is safer to just add it later

src/Documentation/Advanced-Concepts/Dependency-Injection.md
jdom: and the extension method is actually `UseStartupType` (we want to change it eventually, but not now)

amccool:
look like I'm a few hours late with #2169

src/Documentation/Advanced-Concepts/Dependency-Injection.md
```diff
@@ -0,0 +1,75 @@
...
+
+**Note**:
+As Orleans is evolving, as of the current plans it will be possible to leverage dependency injection in other application classes as well, like [`StreamProviders`](../Orleans-Streams/Stream-Providers.md). 
+
```
amccool: I think there needs to be mention of the constructor of the grain being the place where the dependency is injected.  That wasn't immediately obvious to developers that are not versed in DI

## Constructors
Orleans still uses the notion of a class as the fundamental block for the grain.  It has a constructor which executed, but normally this constructor needs to be a empty default constructor with no arguments
Orleans uses the constructor to perform constructor injection (see Mark Seeman), where a given dependency is injected into the grain at(really before) activation time.
So to injection a necessary dependency into your grain, give the grain a constructor taking an argument
``` csharp
public MyGrain(ISuperCoolServiceClient client)
{
  this._client = client;
}
public async Task<string> FavoriteIceCream()
{
  return await client.MakeCallToDependency();
}
```
Where `ISuperCoolServiceClient` is a interface to the dependency you are injecting.   but it could be anything, a `int` constant, or a `string` for the name of your current pet

src/Documentation/Advanced-Concepts/Dependency-Injection.md
```diff
@@ -0,0 +1,75 @@
...
+<?xml version="1.0" encoding="utf-8" ?>
+<tns:OrleansConfiguration xmlns:tns="urn:orleans">
+  <tns:Globals>
+    <tns:Startup Type="MyApplication.Configuration.Startup,MyApplication" />
+  </tns:Globals>
+</tns:OrleansConfiguration>
+```
+# Example
+
+Here is a complete example `Startup` class:
```
amccool: would a more complete example startup be helpful?  Tied to a constructor example for clarity.

# Startup Class
Now that your grain is fully dependent on something, its time to fulfill that dependency. A Startup class is the `SEAM` that performs that action. The Startup class is invoked by the silo when it starts up.
Lets create a startup class.  In this case we are using the `Microsoft.Dependency.Injection` classes, but you can use whatever DI container you like, just return its IServiceProvider.
``` csharp
public class Startup
  {
  public IServiceProvider ConfigureServices(IServiceCollection services)
  {
            services.AddSingleton<ISuperCoolServiceClient, SuperCoolServiceClient>(f =>
            {
                var configSection = (SuperCoolServiceConfiguration) ConfigurationManager.GetSection(SuperCoolServiceConfiguration.SectionName);
                var baseUrl = new Uri(configSection.BaseUrl);
                //return the concrete implementation
                return new SuperCoolService(baseUrl);
            });
            
           services.AddTransient<MyGrain>(
                sp => new MyGrain(sp.GetRequiredService<ISuperCoolServiceClient>()));
  }
```

PR #2169, opened 2016-09-17
Dependency Injection Documentation

amccool:
this needs review both on the gh-pages side of things, and the technical accuracy side.   squash requested


amccool:
It appears that I should have been updating the yml and md files in src,  rather then html files.  I am going to close this PR and resubmit.

Issue #1981, opened 2016-07-25
Investigate Need for Tag in GrainDirectory


sebastianburckhardt:
There is quite a bit of code to return version numbers from the code that registers new addresses in the directory (see `AddressAndTag` struct). Instead of just an address, an address plus a version number is returned.

It appears to me that the version tag is not actually used in the codebase.
The only place where it is passed is `DirectoryCache.AddOrUpdate`
but all implementations of `DirectoryCache` that I can find simply ignore it.

Unless there are objections (e.g. secret implementations of DirectoryCache that actually need this version number), I would like to refactor the code to remove it. It will make the code easier and a tiny bit faster.

sebastianburckhardt:
I think there is a space leak in the grain directory. The `GrainInfo` objects are never removed from `GrainDirectory.partitionData` even after all activations have been deactivated. This would appear to create problems for applications that keep creating new grains over time.

gabikliot:
@sebastianburckhardt , I finally remembered to take a look.
The version IS used, by adaptive cache, here:
https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/GrainDirectory/AdaptiveDirectoryCacheMaintainer.cs#L165

You can see the logic directly inspects and relies on that.
What I cannot say safely at this point is if the logic there is correct/necessary/optimal. It has been too many years since I had a deep understanding of this code. From a brief look now, it does seem like it is doing the right thing. AdaptiveCache and AdaptiveDirectoryCacheMaintainer are the default ones, so this code is definitely used.

If you want to revisit this code and maybe find that the version/tag are not required and this code can be safely changed (basically, challenge the algorithm used there), I will be ready to discuss it with you.

sebastianburckhardt:
Ok, I understand now - thanks for clarifying. Not sure how I missed this.

The algorithm seems fine to me... I can see that the tag provides an efficient way to validate that cached directory entries are not out of date. Definitely more efficient than sending and diffing a list of activation ids. 

I agree that using structs instead of tuples could perhaps reduce memory management overhead, but changing this is probably not a high priority.

gabikliot:
Agreed. 
So going to close that issue for now, since s we resolved it.

PR #2170, opened 2016-09-17
Remove contention point in GrainDirectoryPartition

dVakulen:
Due to large amount of operations under the lock <code>LookUpActivations</code> is one of the main points of contention.

### DIFF
--- a/src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
+++ b/src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
@@ -268,18 +268,19 @@ namespace Orleans.Runtime.GrainDirectory
                 return GrainInfo.NO_ETAG;
             }
 
+            IGrainInfo grainInfo;
             lock (lockable)
             {
-                IGrainInfo grainInfo;
                 if (!partitionData.TryGetValue(grain, out grainInfo))
                 {
                     partitionData[grain] = grainInfo = new GrainInfo();
                 }
 
                 grainInfo.AddActivation(activation, silo);
-                if (log.IsVerbose3) log.Verbose3("Adding activation for grain {0}", grain.ToString());
-                return grainInfo.VersionTag;
             }
+
+            if (log.IsVerbose3) log.Verbose3("Adding activation for grain {0}", grain.ToString());
+            return grainInfo.VersionTag;
         }
 
         /// <summary>
--- a/src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
+++ b/src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
@@ -261,16 +261,19 @@ namespace Orleans.Runtime.GrainDirectory
             {
                 return GrainInfo.NO_ETAG;
             }
+
+            IGrainInfo grainInfo;
             lock (lockable)
             {
-                if (!partitionData.ContainsKey(grain))
+                if (!partitionData.TryGetValue(grain, out grainInfo))
                 {
-                    partitionData[grain] = new GrainInfo();
+                    partitionData[grain] = grainInfo = new GrainInfo();
                 }
-                partitionData[grain].AddActivation(activation, silo);
             }
+
+            grainInfo.AddActivation(activation, silo);
             if (log.IsVerbose3) log.Verbose3("Adding activation for grain {0}", grain.ToString());
-            return partitionData[grain].VersionTag;
+            return grainInfo.VersionTag;
         }
 
         /// <summary>
@@ -289,17 +292,19 @@ namespace Orleans.Runtime.GrainDirectory
 
             if (!IsValidSilo(silo))
                 return result;
-            
+
+            IGrainInfo grainInfo;
             lock (lockable)
             {
-                if (!partitionData.ContainsKey(grain))
+                if (!partitionData.TryGetValue(grain, out grainInfo))
                 {
-                    partitionData[grain] = new GrainInfo();
+                    partitionData[grain] = grainInfo = new GrainInfo();
                 }
-                var grainInfo = partitionData[grain];
-                result.Address = grainInfo.AddSingleActivation(grain, activation, silo, registrationStatus);
-                result.VersionTag = grainInfo.VersionTag;
             }
+
+            result.Address = grainInfo.AddSingleActivation(grain, activation, silo, registrationStatus);
+            result.VersionTag = grainInfo.VersionTag;
+
             return result;
         }
 
@@ -364,21 +369,23 @@ namespace Orleans.Runtime.GrainDirectory
         internal AddressesAndTag LookUpActivations(GrainId grain)
         {
             var result = new AddressesAndTag();
+            IGrainInfo graininfo;
             lock (lockable)
             {
-                IGrainInfo graininfo;
-                if (partitionData.TryGetValue(grain, out graininfo))
+                if (!partitionData.TryGetValue(grain, out graininfo))
                 {
-                    result.Addresses = new List<ActivationAddress>();
-                    result.VersionTag = partitionData[grain].VersionTag;
-
-                    foreach (var route in partitionData[grain].Instances)
-                    {
-                        if (IsValidSilo(route.Value.SiloAddress))
-                            result.Addresses.Add(ActivationAddress.GetAddress(route.Value.SiloAddress, grain, route.Key));
-                    }
+                    return result;
                 }
             }
+
+            result.Addresses = new List<ActivationAddress>();
+            result.VersionTag = graininfo.VersionTag;
+            foreach (var route in graininfo.Instances)
+            {
+                if (IsValidSilo(route.Value.SiloAddress))
+                    result.Addresses.Add(ActivationAddress.GetAddress(route.Value.SiloAddress, grain, route.Key));
+            }
+
             return result;
         }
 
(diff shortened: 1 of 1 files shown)


src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
```diff
@@ -261,16 +261,19 @@ internal virtual int AddActivation(GrainId grain, ActivationId activation, SiloA
             {
                 return GrainInfo.NO_ETAG;
             }
+
+            IGrainInfo grainInfo;
```
dVakulen: Also reduced amount of dictionary lookups.

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
```diff
@@ -289,17 +292,19 @@ internal virtual AddressAndTag AddSingleActivation(GrainId grain, ActivationId a
...
                 {
-                    partitionData[grain] = new GrainInfo();
+                    partitionData[grain] = grainInfo = new GrainInfo();
                 }
-                var grainInfo = partitionData[grain];
-                result.Address = grainInfo.AddSingleActivation(grain, activation, silo, registrationStatus);
-                result.VersionTag = grainInfo.VersionTag;
             }
+
+            result.Address = grainInfo.AddSingleActivation(grain, activation, silo, registrationStatus);
```
dVakulen: Previously it was under the lock, if necessary it can be guarded with more fine-grained one.

jdom: I wonder if will actually need this fine grained lock to avoid races. BTW, maybe removing the several unnecessary lookups already reduced the contention?

dVakulen: No, the lookups were pretty fast.

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
```diff
@@ -364,21 +369,23 @@ internal void RemoveGrain(GrainId grain)
...
-                    foreach (var route in partitionData[grain].Instances)
-                    {
-                        if (IsValidSilo(route.Value.SiloAddress))
-                            result.Addresses.Add(ActivationAddress.GetAddress(route.Value.SiloAddress, grain, route.Key));
-                    }
+                    return result;
                 }
             }
+
+            result.Addresses = new List<ActivationAddress>();
```
ReubenBond: Not sure how much this will help, but we could pass `graininfo.Instances.Count` to the constructor here to set the capacity and avoid additional allocations/copies.

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
```diff
@@ -261,16 +261,19 @@ internal virtual int AddActivation(GrainId grain, ActivationId activation, SiloA
...
-                if (!partitionData.ContainsKey(grain))
+                if (!partitionData.TryGetValue(grain, out grainInfo))
                 {
-                    partitionData[grain] = new GrainInfo();
+                    partitionData[grain] = grainInfo = new GrainInfo();
                 }
-                partitionData[grain].AddActivation(activation, silo);
             }
+
+            grainInfo.AddActivation(activation, silo);
```
gabikliot: Exactly. This is simply wrong, since GrainInfo is not thread safe.

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
gabikliot: I don't understand. How can you imagine a single scenario when this lock is not necessary? The grain info is not thread safe.

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
veikkoeeva: If there are no downsides (e.g. I don't know how list allocates extra internally and should this list grow significantly), feels like the correct action to take.

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
dVakulen: Thought so, just wanted to double check

dVakulen:
Returned locking around grain info

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
dVakulen: There were no downsides.

sebastianburckhardt:
Instead of trying to remove locking, how about striping the lock? For example, use 8 locks, and map grains to locks based on last three bits of consistent hash of grain id.

This could go even further... e.g. have 8 copies of the grain directory itself, and use 8 different consistent hashes. Then we would also solve the problem of imbalance on the grain directory.

dVakulen:
@sebastianburckhardt Solution with 8 locks is viable, but it can degrade in case of non-uniform distribution of requests per grain. 

Could you please describe the problem of imbalance on the grain directory?

sebastianburckhardt:
The imbalance is a result of how consistent hashing works. Each silo's directory partition is responsible for a particular range of grain id hashes, and this range is based on the random hashes of the participating silos. Because this is all random, some silos end up with larger ranges while others have fairly small ranges.  I think this imbalance on the directory is worse than what you get for the grain activations (which are distributed uniformly at random over all silos). Not sure though if that is true, need to some math first.

This is probably not in scope for this particular PR, but I think it may be worth thinking about a solution at some point.

gabikliot:
The imbalance in the directory is indeed real.
This can be seen in the periodic statistics that are printed: there is both imbalance in the ring range and also, consequently, in the number of directory entries per silo.
The solution, like Sebastian pointed out, is to use "virtual buckets" - multiple points on the ring per silo.

This is already implemented in ConsistentRingProvider and used in Reminders and streams queue balancer. But the directory was written before ConsistentRingProvider  and was never ported to use it.
So one way to fix it is to carefully re-write the directory to use ConsistentRingProvider. Alternatively we can do what Sebastian suggested, use multiple directory objects with different hashes. It will be essentially the same idea, just implemented slightly differently.
Indeed, I think both are beyond the scope of this PR. 

Interestingly, we have known about that problem for years, but it never actually become something that anyone ever noticed or complained about. It can potentially decrease the impact of new grain activations (lookups will be mostly fine due to caching), but even with that problem the throughput of new activation was good enough so no one ever complained. ;-)

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
```diff
@@ -364,21 +376,58 @@ internal void RemoveGrain(GrainId grain)
         internal AddressesAndTag LookUpActivations(GrainId grain)
```
sergeybykov: Since the overwhelming majority of cases, pretty close to100%, involves a single activation per grain registered in the directory, wouldn't it make sense to optimize that hot path instead of the general but non-existent in reality case of multiple activations registered per grain? For example, would it make sense to add a field to `AddressesAndTag` of type `ActivationAddress` for a single result, so that the `Addresses` fields will pretty much always be set to null and we'd never allocate those lists? Alternatively, we could make lookup methods return a `Tuple<AddressesAndTag, AddressAndTag>`. Although a Tuple would cause another undesirable allocation, so maybe adding a field would be cheaper. C# 7 will eventually allow to return multiple values from a method more naturally.

dVakulen: Thought about that optimization, but wanted to process iteratively.

 As for majority - in commit https://github.com/dotnet/orleans/commit/ed1b4efeb31b6fb57159257f06272752da620b13#diff-10c9ac6653bd494c2edbdb7701be8c90 
method <code>LookUpGrain</code> was renamed into <LookUpActivations>, so that it now may seem to be responsible for looking up of multiple activations, which is usually not the case. 

chrome_2016-09-29_19-16-36

dVakulen:
I'll look at problem of imbalance on the grain directory after resolution of this one.

As for why no one ever complained - it has quite small impact on performance, especially compared to the cost of multiple <code>BlockingCollection</code>'s 

 Lock contention view: 

timeline64_2016-09-29_19-30-12

But without those queues it becomes top point of contention: 

timeline64_2016-09-29_19-30-35

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
sebastianburckhardt: Since we are staring at this code right now anyway: as I remarked in #1981: the tag (some kind of version number of the directory entry for a particular grain) is never actually really needed for anything (I suspect it had a purpose in the past at some point). We can save space and time by removing it everywhere in the code.

gabikliot:
I will look at the tag and will try to remember why it was added. Maybe you can open an issue of looking into it?

sebastianburckhardt:
Ok... I renamed existing issue #1981 to track discussion on the tag.

src/OrleansRuntime/GrainDirectory/GrainDirectoryPartition.cs
sergeybykov: I agree - let's optimize those things iteratively, as separate PRs. I'm running a load test for this, for just in case. Will merge after that.

sergeybykov:
Thank you, @dVakulen!

PR #2172, opened 2016-09-19
Add comments in TestCluster and TestClusterOptions.

benjaminpetit:


src/OrleansTestingHost/TestCluster.cs
```diff
@@ -28,23 +28,43 @@ namespace Orleans.TestingHost
...
         public IReadOnlyList<SiloHandle> SecondarySilos => this.additionalSilos;
 
-        protected readonly List<SiloHandle> additionalSilos = new List<SiloHandle>();
-        protected readonly Dictionary<string, GeneratedAssembly> additionalAssemblies = new Dictionary<string, GeneratedAssembly>();
+        private readonly List<SiloHandle> additionalSilos = new List<SiloHandle>();
 
+        private readonly Dictionary<string, GeneratedAssembly> additionalAssemblies = new Dictionary<string, GeneratedAssembly>();
+
+        /// <summary>
+        /// Client condfiguration to use when initializing the client
```
jdom: typo

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -163,15 +217,11 @@ public static ClientConfiguration BuildClientConfiguration(ClusterConfiguration
...
-            }
-            else
-            {
-                config.ResponseTimeout = clusterConfig.Globals.ResponseTimeout;
-            }
+
+            // Test is running inside debugger - Make timeout ~= infinite
+            config.ResponseTimeout = Debugger.IsAttached
+                ? TimeSpan.FromMilliseconds(1000000)
+                : clusterConfig.Globals.ResponseTimeout;
```
jdom: the comment is not inside the if, so rephrase a little bit

src/OrleansTestingHost/TestCluster.cs
```diff
@@ -28,23 +28,43 @@ namespace Orleans.TestingHost
     /// </remarks>
     public class TestCluster
     {
+        /// <summary>
+        /// Primary silo handle
+        /// </summary>
```
jdom: nit picking for future comments, as I know more are coming: if it's a simple short phrase, keep the comment and XML info in a single line, so that it doesn't unnecessarily occupy more than the actual code.

benjaminpetit: Do you know if this is possible to do this with Resharper?

PR #2173, opened 2016-09-19
Fix comments in SiloHandle, StorageTestConstants, TestingClientOption…

benjaminpetit:
…s, TestingSiloHost and TestingSiloOptions

### DIFF
--- a/src/OrleansTestingHost/TestingSiloHost.cs
+++ b/src/OrleansTestingHost/TestingSiloHost.cs
@@ -78,7 +78,8 @@ namespace Orleans.TestingHost
         /// <summary> GrainFactory to use in the tests </summary>
         public IGrainFactory GrainFactory { get; private set; }
 
-        private Logger logger
+        /// <summary> Get the logger to use in tests </summary>
+        protected Logger logger
         {
             get { return GrainClient.Logger; }
         }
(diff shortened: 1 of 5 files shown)


src/OrleansTestingHost/TestingSiloHost.cs
```diff
@@ -38,29 +38,47 @@ namespace Orleans.TestingHost
...
+        /// <summary> Base port number for the gateway silos </summary>
         public const int ProxyBasePort = 40000;
 
+        /// <summary> Number of silos in the cluster </summary>
         private static int InstanceCounter = 0;
 
+        /// <summary> GrainFactory to use in the tests </summary>
         public IGrainFactory GrainFactory { get; private set; }
 
-        protected Logger logger
```
jdom: please don't change this, since the previous guidance with this TestingSiloHost class used to say you should inherit from it, so this protected logger was used by many. We are deprecating this class soon, no need to do breaking changes in the meantime.

benjaminpetit: I can revert this, but it is not used here? You mean that it could be used outside our codebase?

xiazen:
I did the same thing.... :< without noticing Ben already did it. PR goes by too fast.

PR #2174, opened 2016-09-19
Added XML comments for memory stream provider related files.

JingqiaoFu:
Added XML comments for memory stream provider related files.

### DIFF
--- a/src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
+++ b/src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
@@ -3,12 +3,24 @@ using System.Threading.Tasks;
 
 namespace Orleans.Providers.Streams.Memory
 {
+    /// <summary>
+    /// Interface for In-memory stream queue grain.
+    /// </summary>
     public interface IMemoryStreamQueueGrain : IGrainWithGuidKey
     {
+        /// <summary>
+        /// Set max event count. The total number of events enqueued can not exceed maxEventCount.
+        /// </summary>
         Task SetMaxEventCount(int maxEventCount);
 
+        /// <summary>
+        /// Enqueue an event.
+        /// </summary>
         Task Enqueue(MemoryEventData eventData);
-         
+
+        /// <summary>
+        /// Dequeue an event.
+        /// </summary>
         Task<List<MemoryEventData>> Dequeue(int maxCount);
     }
 }
--- a/src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
+++ b/src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
@@ -19,7 +19,7 @@ namespace Orleans.Providers.Streams.Memory
         Task Enqueue(MemoryEventData eventData);
 
         /// <summary>
-        /// Dequeue an event.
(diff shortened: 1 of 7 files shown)


src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
```diff
@@ -3,12 +3,24 @@
...
         Task SetMaxEventCount(int maxEventCount);
 
+        /// <summary>
+        /// Enqueue an event.
+        /// </summary>
         Task Enqueue(MemoryEventData eventData);
-         
+
+        /// <summary>
+        /// Dequeue an event.
```
jason-bragg: Dequeue up to maxCount events

src/OrleansProviders/Streams/Memory/MemoryAdapterConfig.cs
```diff
@@ -10,17 +10,37 @@ namespace Orleans.Providers.Streams.Memory
...
     {
+        /// <summary>
+        /// Stream provider name.
+        /// </summary>
         public String StreamProviderName { get; private set; }
 
+        /// <summary>
+        /// Total queue count name. Indicates the name of the property in provider config.
+        /// </summary>
         private const string TotalQueueCountName = "TotalQueueCount";
```
jason-bragg: why is this no public?

src/OrleansProviders/Streams/Memory/MemoryAdapterFactory.cs
```diff
@@ -85,16 +135,30 @@ public IQueueAdapterReceiver CreateReceiver(QueueId queueId)
...
+        /// Create a cache for a given queue id
+        /// </summary>
+        /// <param name="queueId"></param>
         public IQueueCache CreateQueueCache(QueueId queueId)
         {
             return new MemoryPooledCache(bufferPool, logger.GetSubLogger("messagecache", "-"));
         }
 
+        /// <summary>
+        /// Aquire delivery failure handler for a queue
```
jason-bragg: acquire?

src/OrleansProviders/Streams/Memory/MemoryPooledCache.cs
```diff
@@ -192,17 +212,32 @@ public void AddToCache(IList<IBatchContainer> messages)
...
+        /// </summary>
+        /// <param name="purgedItems"></param>
         public bool TryPurgeFromCache(out IList<IBatchContainer> purgedItems)
         {
             purgedItems = null;
             return false;
         }
 
+        /// <summary>
+        /// Acquire a stream message cursor.  This can be used to retreave messages from the
```
jason-bragg: retrieve?

src/OrleansProviders/Streams/Memory/MemoryAdapterConfig.cs
JingqiaoFu: the PopulateFromProviderConfig method is the only place that needs the name to get total queue count so far, do you think it might be needed somewhere else?

jason-bragg: Having it public makes it easier for developers that want to specify the value in xml configs to do so.  I'm not saying it's entirely necessary, just that when we've hidden config value names in the past it's made it difficult for end users to configure systems correctly.

Issue #2143, opened 2016-09-12
[WIP] Orleans 2.X


galvesribeiro:
_This issue will be a "root" issue to track the progress on 2.0 milestone following the steps agreed at [Meetup 10](https://youtu.be/_SbIbYkY88o)._ 

The main target on this milestone is the port to [.Net Core](https://github.com/dotnet/coreclr) but other features like enhanced deployment & cluster management in multiple hosting environments, pervasive DI, new persistence layer, etc., but all them basically follow the major goal to be compatible with .Net Core. 

The initial strategy is to keep working on the port on `master` branch by having a separated and temporary solution where we will target `.Net Standard` while using new projects "linking" to the `.cs` files and following the guidelines [here](https://docs.microsoft.com/en-us/dotnet/articles/core/tutorials/target-dotnetcore-with-msbuild) (thanks to @jdom for this briliant idea). The reason for that strategy and not the current one (have a separated branch for that) is to reduce the merge complexity and be able to hotfix and add non-breaking changes to 1.3.x releases while we still working on .Net Core port. We will focus on have any breaking change on hold to 2.0 but that doesn't block us from having new features being released under 1.x.y tag.

The tasks will be distributed across multiple phases defined as **_CRAWL_**, **_WALK_** and **_RUN_**.  So, here is the list of macro-tasks that we should cover in this story and each one of them will be tracked by a separated issue and its sub-issues referenced here. 
- [x] **_CRAWL_**
  - [x] .Net Core Compatible Binaries - #2145
  - [x] Orleans 2.0 (beta is out)
- [ ] **_WALK_**
  - [x] Building on .Net Core 
  - [ ] Testing on .Net Core
  - [ ] Orleans 2.X
- [ ] **_RUN_**
  - [ ] Cross-platform testing
  - [ ] Orleans 2.Y

**Note**: #368 was tracking it before but since most of the issues there were already completed and we moved to a new strategy, from now on, lets keep the discussion about that here.

jason-bragg:
As each of these stages (Crawl, Walk, Run) will include it's own discussions and a number of sub-tasks, would it make sense to open a similar issue for each stage as we enter it?  This is a good high level view, but if this ends up including all of the discussion and details of the full port it will get a bit unruly up in here :)

galvesribeiro:
@jason-bragg yep! That is the idea as we agreed on the meetup. I'm writing the Crawl main issue right now and will refer it here. Gimme a few minutes.

[...]

PR #2177, opened 2016-09-20
Solution for .NET Standard using linked files

jdom:
This new solution contains the working place for the migration effort to .NET Standard and .NET Core (https://github.com/dotnet/orleans/issues/2143).

The `vNext` tree has a separate solution (`src/Orleans.vNext.sln`) with several projects that have links to the code files in the main (non-vNext) solution.
Currently not all of the original projects were migrated to target .NET Standard, and we'll add them as we need and can (for example, we first need to get a few basic test projects up so that we can start asserting the port is working as expected).

We mainly use 2 different compilation constant to differentiate between the 2 solutions (via conditional compilation):
- `NETSTANDARD`: When something is just different between the 2 targets. Long term we want to avoid multi-targeting library projects for full .NET Framework and .NET Standard. So it means that this compilation constant will eventually be dropped, but we do not expect this soon (at least not before the WALK phase defined in issue #2143).
- `NETSTANDARD_TODO`: short term temporary conditional compilation to get the project to build, but requires fixing before that feature is functional. This was mainly added to get the project bootstrapped and compiling so that the community can start contributing these pieces that were temporarily left out.

Things coming soon:
- I will soon add this solution to be built in CI to avoid contributions going to the main solution to break vNext.
- I'll start adding some basic test projects. The ones that require a silo will take some more time, as there are several things needed for this work.
- This should now be ready for people submitting PRs against it, especially with fixes for stuff marked with the `NETSTANDARD_TODO` compilation constant.

See [this comment](https://github.com/dotnet/orleans/pull/2177#issuecomment-248194403) for rationale of this approach.

Note to committers (and reviewers): please do not squash this into a single commit. If there's feedback to address (most likely), I'll add new commits, but when this has been approved an it's ready to be merged, I'll just rebase and clean the commit history, since I tried to track each of the remaining issues into separate commits (which is a nice way to review too, if you want to go 1 commit at a time)


ReubenBond:
@dotnet-bot test this

ReubenBond:
@dotnet-bot test this please

galvesribeiro:
Thanks @jdom! will review tomorrow

ReubenBond:
Is the eventual plan to have just one .sln, or to maintain two? I'm not sure I see the intended path here

galvesribeiro:
@ReubenBond this is temporary so we can work in parallel in `master`. When we have become .Net Standard compatible, we just replace the projects/solution.

ReubenBond:
Hopefully this is a short-lived change. I appreciate the push to move .NET Standard work into the mainline branch - it should help us move much faster.

jdom:
Ideally this will not be needed once we finish the walk phase, and we are confident that we can replace the binaries with the ones targeting .Net standard. So it will be medium-term lived, or as long as it takes us to reach the end of that 2nd phase

vNext/src/Orleans/Orleans.csproj
```diff
@@ -0,0 +1,1047 @@
...
+    <None Include="..\..\..\src\Orleans\Configuration\OrleansConfiguration.xsd">
+      <Link>Configuration\OrleansConfiguration.xsd</Link>
+      <SubType>Designer</SubType>
+    </None>
+    <None Include="project.json" />
+  </ItemGroup>
+  <ItemGroup />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets" />
+  <PropertyGroup>
+    <NuGetTargetMoniker>.NETStandard,Version=v1.5</NuGetTargetMoniker>
```
ReubenBond: (I'm probably mistaken, but) the project itself doesn't seem to be targeting .NET Standard 1.5 directly, it looks like a PCL
image

ReubenBond:
Assuming that we have consensus on the general approach (creating a vNext subdirectory with projects which link to files from the main tree) and that we will clean up the `#if`s before 2.0 (there should be an open issue on it), then LGTM. I'm still a little wary of introducing mess, but maybe the benefit outweighs the cost.

vNext/src/Orleans/Orleans.csproj
jdom: yes, this is by design. It is an official hack explained here: https://docs.microsoft.com/en-us/dotnet/articles/core/tutorials/target-dotnetcore-with-msbuild

jdom: This will be replaced once the final msbuild tooling for .net standard is in place. In the meantime it allows us to do things like compile time codegen with the existing targets, instead of migrating everything to the dotnet CLI tooling which will be completely changed very soon

ReubenBond: That explains it :) Sounds good.

src/Orleans/AssemblyLoader/AssemblyProcessor.cs
```diff
@@ -97,11 +97,14 @@ private static void ProcessAssembly(Assembly assembly)
             {
                 Logger.Verbose3("Processing assembly {0}", assemblyName);
             }
+
+#if !NETSTANDARD
```
jdom: There is no reflection only load, so we won't have this scenario in .net standard (that's why it's not a TODO item at least here).
I'll review tomorrow if there's another reflection only load check that I incorrectly marked as TODO and fix it, thanks. The reality is that previously they were all just NETSTANDARD, and I recently started rebasing and crafting the commits so they add the NETSTANDARD_TODO one instead, so it is very likely that I missed a few where it made sense.

jdom:
I understand, and I also explored many options to avoid this. Nevertheless this ended up being the most workable approach.
A little bit of the rationale behind this approach:
- using csproj allows us to be unblocked even before the final tooling RTMs (otherwise we would have to re-do the tooling once xproj goes back to csproj)
- we also decided to work in the `master` branch instead of creating a separate branch:
  - otherwise most of our contributors would be unaware of it and unknowingly break when submitting PRs for the non-vNext solution (or even if aware, they would need to submit changes against 2 different branches, which is not at all practical).
  - this way we continue to support the good old non-vNext solution until we reach the end of the WALK phase and have a usable 2.0 release.
  - because we support both targets in the same branch, then we require this extra vNext solution, and also the linked files (there were other otions considered, such as using conditionals in the main csproj themselves, but that's probably even more messier, but I could stand corrected).

If getting there would be a week's worth of work, then this intermediate vNext solution would not be needed, but we know that's impossible. We need something that temporarily builds even if it is not functional for the community (and rest of our core team) contributions to start coming, otherwise I'm the single bottleneck to get everything 100% working before we can push to master.

vNext/src/Orleans/Orleans.csproj
jdom: BTW, I'm glad you downloaded it locally for reviewing, as it is hard to get a sense of how it works without it. Let me know if you have any issues building, as this csproj stuff made me do a few hacks (the other one is that project references don't work well with this, so it is using project build dependencies to define the project build order, and then dependent projects reference the built dll in the output path).

ReubenBond: Both trees build fine & BVTs pass in main tree. Now I'm trying to get IL-serializer working on top of it :)

xiazen:
Good job !@jdom . Will take a closer look tomorrow !

galvesribeiro:
@ReubenBond When Julian first came with the new .sln idea I was reluctant about multiple `#if`s myself but the benefits of being on master as he stated worth a lot more IMHO however, I think we can save some of those `#if`s if we bump our framework version to 4.6.1. but we will eventually hit Azure users that will need use startup tasks to install it. This is the only downside of not having a diff branch - stuck on 4.5.1. Many things like `AsyncLocal<T>` for the `RequestContext` are only available on 4.6.1. 

Overall, LGTM. I'll be back soon and start working on that as well.

PR #2179, opened 2016-09-20
Minor fixes to documentation

richorama:
* Include DI page
* Fix link in persistence page

Also, I don't think the HTML was built after previous changes were made.


sergeybykov:
Thank you, @richorama!

PR #2182, opened 2016-09-20
Synchronized Tutorials TOC with the article

svick:
Articles are now on the right level, in the right order and with the right name.


sergeybykov:
Thanks!

PR #2186, opened 2016-09-20
Added XML comments for TelemetryConsumers

jason-bragg:


src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
```diff
@@ -153,20 +163,56 @@ private PerfCounterConfigData GetCounter(string counterName)
 
         #region IMetricTelemetryConsumer Methods
 
+        /// <summary>
+        /// Increment metric.
+        /// </summary>
+        /// <param name="name"></param>
```
jdom: Please remove empty param names or return elements. It's better to not have them than have explicit empty definitions (not having any will remove warnings if that's what you want, but empty items cause more longer term problems, as well as unnecessarily harm readability)

PR #2188, opened 2016-09-20
fix XML warnings in DynamoDBStorage and DynamoDBRemiderTable

xiazen:


src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -181,6 +214,10 @@ private static ReminderEntry Resolve(Dictionary<string, AttributeValue> item)
             }
         }
 
+        /// <summary>
+        /// Aync method to clear the reminder table
+        /// </summary>
+        /// <returns>A task object </returns>
```
jdom: Either drop the entire return, or just say `A task that completes when clearing the table completes`. Nevertheless, that is the standard way of using a task, so I would not add it.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
```diff
@@ -157,6 +183,13 @@ private static ReminderEntry Resolve(Dictionary<string, AttributeValue> item)
             };
         }
 
+        /// <summary>
+        /// Remove one row from the reminder table 
+        /// </summary>
+        /// <param name="grainRef"> specific grain ref to locate the row </param>
+        /// <param name="reminderName"> remider name to locate the row </param>
+        /// <param name="eTag"> e tag </param>
+        /// <returns> A Task object to indicate the result of this aync method </returns>
```
jdom: I'd say `Return true if a row was removed'. The fact that it's a task is not the important part of the return value.

src/OrleansAWSUtils/Reminders/DynamoDBReminderTable.cs
jdom: There are several occurrences, so I mainly added this on a Task and Task<T> so that it reflects what is important to say here, but it is typically not the fact that the type is a Task, unless the Task completes at a special point in time, and is not necessarily trivial.

xiazen: fixed them

PR #2191, opened 2016-09-20
Rebuild docs with latest updates.

attilah:


jdom:
Thanks @attilah, for some reason I guess it's my environment that doesn't include the `Improve my Doc` button, not sure what's wrong.
Regardless, @richorama, do you see any issues with updating our DocFx requirement to 2.5.3? I would assume no, but asking nonetheless, since you are the most familiar with this

PR #2193, opened 2016-09-21
Remove UseStandardSerializer option

ReubenBond:
Removes `UseStandardSerializer` option, which is designed to allow users to use BinaryFormatter instead of Orleans' serializer. BinaryFormatter is not available on .NET Core and users can use BinaryFormatter by providing an `IExternalSerializer` implementation.

Once this is merged, I'll propose a similar PR to remove the JSON fallback serialization option.


sergeybykov:
`UseStandardSerializer` was added in the very beginning of the project when our confidence in the generated serializers was much lower. Should have removed this flag a long time ago, as we pretty much never really used it since.

PR #2199, opened 2016-09-21
removing async from documentation example

skalpin:
Updating home page example to match closer to the one in the github readme


sergeybykov:
Thank you, @skalpin!

Issue #2197, opened 2016-09-21
.NET Standard compat: Avoid usages of CallContext


jdom:
In our .NET Standard solution I temporarily removed the usages of `CallContext` from our `RequestContext` class, as it's not supported in .NET Standard. See https://github.com/dotnet/orleans/commit/772dde721cf89bb799703dadbf0a4cf988f5fca1 for the commit where this was excluded.

We need to find an alternative. Without too much investigation on my end, it seems that the recommended alternative is to use `AsyncLocal<T>` which is available in both .NET Standard and .NET 4.6+. If we go with this approach, we should probably keep this code under conditional compilation for the mean time, as we are not ready to update the dependency for the main solution (non-vNext) to 4.6 just yet (soon though).

In order to open the solution that compiles using .NET Standard (and hence easy to check what is available in that environment), check out [this readme](https://github.com/dotnet/orleans/blob/master/vNext/README-CoreClr-port.md)

PR #2200, opened 2016-09-21
Replace CallContext.LogicalSetData with AsyncLocal

dVakulen:
Addresses #2197. 

A couple of references: [AsyncLocal tests](https://github.com/dotnet/corefx/blob/master/src/System.Threading/tests/AsyncLocalTests.cs), [AsyncLocal backing storage structure](http://referencesource.microsoft.com/#mscorlib/system/threading/executioncontext.cs,189).

### DIFF
--- a/src/Orleans/Runtime/RequestContext.cs
+++ b/src/Orleans/Runtime/RequestContext.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 #if !NETSTANDARD_TODO
 using System.Runtime.Remoting.Messaging;
 #endif
@@ -30,8 +31,6 @@ namespace Orleans.Runtime
     /// </remarks>
     public static class RequestContext
     {
-        // TODO: We can potentially use AsyncLocal<T> instead of the logical call context, which is supported in .NET Standard
-
         /// <summary>
         /// Whether Trace.CorrelationManager.ActivityId settings should be propagated into grain calls.
         /// </summary>
@@ -41,6 +40,11 @@ namespace Orleans.Runtime
         internal const string E2_E_TRACING_ACTIVITY_ID_HEADER = "#RC_AI";
         internal const string ORLEANS_REQUEST_CONTEXT_KEY = "#ORL_RC";
         internal const string PING_APPLICATION_HEADER = "Ping";
+
+#if NETSTANDARD_TODO
+        private static readonly AsyncLocal<Guid> ActivityId = new AsyncLocal<Guid>();
+        private static readonly AsyncLocal<Dictionary<string, object>> CallContextData = new AsyncLocal<Dictionary<string, object>>();
+#endif
 
         /// <summary>
         /// Retrieve a value from the RequestContext key-value bag.
@@ -111,6 +115,13 @@ namespace Orleans.Runtime
                 {
                     activityIdObj = Guid.Empty;
                 }
+                else
+                {
+#if NETSTANDARD_TODO
+                    ActivityId.Value = (Guid) activityIdObj;
+#endif
+                }
+
 #if !NETSTANDARD_TODO
                 Trace.CorrelationManager.ActivityId = (Guid) activityIdObj;
 #endif
@@ -133,10 +144,13 @@ namespace Orleans.Runtime
         {
             Dictionary<string, object> values = GetContextData();
 
-#if !NETSTANDARD_TODO
             if (PropagateActivityId)
             {
-                Guid activityId = Trace.CorrelationManager.ActivityId;
+#if !NETSTANDARD_TODO
+                var activityId = Trace.CorrelationManager.ActivityId;
+#else
+                var activityId = ActivityId.Value;
+#endif
                 if (activityId != Guid.Empty)
                 {
                     values = values == null ? new Dictionary<string, object>() : new Dictionary<string, object>(values); // Create new copy before mutating data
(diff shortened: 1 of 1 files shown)


sergeybykov:
>UnitTests.CancellationTests.GrainCancellationTokenTests.CancellationTokenCallbacksThrow_ExceptionShouldBePropagated

BTW, this test is also failing intermittently in our VSTS builds.

jdom:
@dotnet-bot test this please

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -42,6 +41,11 @@ public static class RequestContext
         internal const string ORLEANS_REQUEST_CONTEXT_KEY = "#ORL_RC";
         internal const string PING_APPLICATION_HEADER = "Ping";
 
+#if NETSTANDARD_TODO
+        private static readonly AsyncLocal<Guid> ActivityId = new AsyncLocal<Guid>();
+        private static readonly AsyncLocal<Dictionary<string, object>> CallContextData = new AsyncLocal<Dictionary<string, object>>();
```
jdom: I'm not entirely convinced with how the ActivityId is being handled here. If I understand correctly, the only way of setting it is by using `E2_E_TRACING_ACTIVITY_ID_HEADER` in the contextData. If that's the case, why is it being extracted to its own AsyncLocal value? A little clarification on how this is expected to work would be very valuable.

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -42,6 +41,11 @@ public static class RequestContext
         internal const string ORLEANS_REQUEST_CONTEXT_KEY = "#ORL_RC";
         internal const string PING_APPLICATION_HEADER = "Ping";
 
+#if NETSTANDARD_TODO
```
jdom: The purpose of `NETSTANDARD_TODO` constant is to have something that needs addressing ASAP. So this PR should probably change these to just `NETSTANDARD`; meaning that it is a different implementation for the other platform, but does not need addressing any further (other than at some point we'll get rid of all `NETSTANDARD` conditional compilation entirely too and have 1 way of doing it, for example when we support 4.6, or move platform-specific code to a platform-specific project, but this is longer term)

src/Orleans/Runtime/RequestContext.cs
dVakulen: Previously <code>Trace.CorrelationManager</code>  was used for storing it after extracting from contextData. It could be stored in <code>CallContextData</code> dictionary as well, saving us from additional variable but on cost of more conditional compilation statements.

src/Orleans/Runtime/RequestContext.cs
dVakulen: Oh, I see. Just wanted to mark it as subject for further complete removal of <code>NETSTANDARD</code> condition. Will rename.

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -111,7 +115,14 @@ public static void Import(Dictionary<string, object> contextData)
                 {
                     activityIdObj = Guid.Empty;
                 }
-#if !NETSTANDARD_TODO
+                else
+                {
+#if NETSTANDARD
```
xiazen: Why add an else here? it alters original logic. The local value in ActivityId would remain unchanged if first `if ` condition succeed.

xiazen:
I have ported RequestContext related tests to vNext in #2274 . will pull your change down tomorrow to see it those tests would succeed.

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -154,24 +167,28 @@ public static void Import(Dictionary<string, object> contextData)
         public static void Clear()
         {
             // Remove the key to prevent passing of its value from this point on
-#if !NETSTANDARD_TODO
+#if !NETSTANDARD
             CallContext.FreeNamedDataSlot(ORLEANS_REQUEST_CONTEXT_KEY);
+#else
+            CallContextData.Value = null;
```
xiazen: is there a way to free the dictionary? or is it ok to leave it to GC?

dVakulen: It is ok to leave it to GC.

jdom:
Thanks, LGTM.
Let's wait for Xiao to run the tests before merging

xiazen:
test RequestContext_ActivityId_ExportToMessage set activityId from the test in order to test export, so do we want to change RequestContext.ActivityId field to internal? for testing?

xiazen:
test `RequestContext_ActivityId_ExportImport` and `RequestContext_ActivityId_ExportToMessage` need to set ActivityId in order to test ExportImport and ExportToMessage. 
I changed RequestContext.ActivityId to internal and revised these two tests, all NonSiloRequired RequestContextTests succeed. But I don't know about silo related. Guess we will find out soon,

So I guess, change RequestContext.ActivityId filed to be internal to test project and I will revise those two tests after this change got check in.

jdom:
Well, actually that is related to the comment I made [a few weeks ago](https://github.com/dotnet/orleans/pull/2200#discussion_r80571942), that I completely forgot. But yeah, it's not clear how end users will set the ActivityId. Internal is no good, otherwise it will only make tests pass, but not really be useful. Maybe either making it public, or even have the getter & setter accessors only set an entry in the request context dictionary, but do not treat it as a separate AsyncValue

src/Orleans/Runtime/RequestContext.cs
xiazen: So you think end user should be able to set ActivityID? If we just want to preserve original logic, we should make it public.

jdom:
Thanks @dVakulen !!

galvesribeiro:
Thanks @dVakulen! Great work! 👍

PR #2201, opened 2016-09-21
Fix bug in GrainFactoryServiceProvider

jdom:
Was causing some EventHub tests to fail because providerRuntime.ServiceProvider was null, and not the providerRuntime itself (which was the path being tested for null). I believe this would have meant that if someone uses the EventHubStreamProvider from the client, they would have hit this bug.

Also replaced `MemoryAdapterFactory` to depend on `IGrainFactory` instead of the `GrainFactory` concrete class and updated hack accordingly. And since IGrainFactory is now currently being injected in the silo side, rely on that first and only fall back if it is not set.

### DIFF
--- a/src/Orleans/Streams/PersistentStreams/PersistentStreamProvider.cs
+++ b/src/Orleans/Streams/PersistentStreams/PersistentStreamProvider.cs
@@ -60,7 +60,7 @@ namespace Orleans.Providers.Streams.Common
             }
             public object GetService(Type serviceType)
             {
-                var service = providerRuntime?.ServiceProvider?.GetService(serviceType);
+                var service = providerRuntime.ServiceProvider?.GetService(serviceType);
                 if (service != null)
                 {
                     return service;
--- a/src/Orleans/Streams/PersistentStreams/PersistentStreamProvider.cs
+++ b/src/Orleans/Streams/PersistentStreams/PersistentStreamProvider.cs
@@ -50,7 +50,7 @@ namespace Orleans.Providers.Streams.Common
 
         public bool IsRewindable { get { return queueAdapter.IsRewindable; } }
 
-        // this is a workaround until we address "Dependency Injection: register IGrainFactory #988"
+        // this is a workaround until an IServiceProvider instance is used in the Orleans client
         private class GrainFactoryServiceProvider : IServiceProvider
         {
             private IStreamProviderRuntime providerRuntime;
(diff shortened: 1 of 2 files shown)


src/Orleans/Streams/PersistentStreams/PersistentStreamProvider.cs
```diff
@@ -60,11 +60,18 @@ public GrainFactoryServiceProvider(IStreamProviderRuntime providerRuntime)
             }
             public object GetService(Type serviceType)
             {
-                if (serviceType == typeof (GrainFactory))
+                var service = providerRuntime?.ServiceProvider?.GetService(serviceType);
```
jason-bragg: can the provider runtime be null because if so, getting the grain factory from it won't work.

src/OrleansProviders/Streams/Memory/MemoryAdapterFactory.cs
```diff
@@ -61,7 +61,7 @@ public void Init(IProviderConfiguration providerConfig, string providerName, Log
             this.providerName = providerName;
             this.queueGrains = new ConcurrentDictionary<QueueId, IMemoryStreamQueueGrain>();
             this.adapterConfig = new MemoryAdapterConfig(providerName);
-            grainFactory = (GrainFactory)serviceProvider.GetService(typeof(GrainFactory));
+            grainFactory = (GrainFactory)serviceProvider.GetService(typeof(IGrainFactory));
```
jason-bragg: We shouldn't need a GrainFactory here.  Should be IGrainFactory.

jdom: oh, yeah, missed the cast, good catch

src/Orleans/Streams/PersistentStreams/PersistentStreamProvider.cs
```diff
@@ -60,11 +60,18 @@ public GrainFactoryServiceProvider(IStreamProviderRuntime providerRuntime)
             }
             public object GetService(Type serviceType)
             {
-                if (serviceType == typeof (GrainFactory))
+                var service = providerRuntime.ServiceProvider?.GetService(serviceType);
```
xiazen: maybe add a null check on providerRuntime too? otherwise providerRuntime.ServiceProvider would throw a NullReferenceException.

xiazen: and also if providerRuntime is null, getting GrainFactory, or any service from it should return null.

jason-bragg: Provider must have a runtime to work.  Null check here wouldn't help.  If we were to add a check we should check the value at provider initialization or at latest service provider creation.

jason-bragg: Just double checked.  Runtime is properly checked at provider initialization.  As a private internal class this service provider can trust that and does not need to check for itself.

xiazen: thanks for checking. This PR looks good to me. Approve it .

xiazen: just testing new feature of github

PR #2202, opened 2016-09-21
Cleanup of XML docs in the Orleans project

sergeybykov:


src/Orleans/Logging/LoggerExtensions.cs
```diff
@@ -56,6 +56,7 @@ public static void Verbose3(this Logger logger, string format, params object[] a
         /// Info is suitable for information that does not indicate an error but that should usually be logged in production.
         /// Info is lower than Warning.
         /// </summary>
+        /// <param name="logger">Terget logger.</param>
```
ReubenBond: nit: typo on Target

src/Orleans/Telemetry/TraceParserUtils.cs
```diff
@@ -70,16 +70,15 @@ public static string PrintProperties(string message, IDictionary<string, string>
                 errorCode);
         }
 
-        /// <summary>
-        /// The method to call during logging to format the log info into a string ready for output.
-        /// </summary>
```
ReubenBond: Should I be reformatting my docs to remove newlines around tags?

src/Orleans/Core/Grain.cs
```diff
@@ -269,9 +269,10 @@ protected Grain()
         /// This constructor is particularly useful for unit testing where test code can create a Grain and replace
         /// the IGrainIdentity, IGrainRuntime and State with test doubles (mocks/stubs).
         /// </summary>
-        /// <param name="state"></param>
         /// <param name="identity"></param>
         /// <param name="runtime"></param>
+        /// <param name="state"></param>
+        /// <param name="storage"></param>
```
jdom: if you are not filling the parameters (or `returns` tag), just remove them altogether. It does get rid of the warning, and causes less maintenance issues in the future (if you add new parameters), as well as avoids overriding explicitly with empty in the case the member is inherited. What does cause a warning though is having a few parameters *documented* but not all

src/Orleans/Telemetry/TraceParserUtils.cs
jdom: When the text is short, it does make sense to avoid multiple lines. This way we reduce the unnecessary comment lines to code lines ratio.
This is not mandatory though, just a nice to have.

src/Orleans/Core/Grain.cs
sergeybykov: Good to know.

src/Orleans/Logging/LoggerExtensions.cs
sergeybykov: Fixed. Thank you.

PR #2203, opened 2016-09-22
Add powershell script to run tests in parallel in build

benjaminpetit:
Using background jobs to start xunit console, one per test assembly.

### DIFF
--- a/src/Parallel-Tests.ps1
+++ b/src/Parallel-Tests.ps1
@@ -3,7 +3,11 @@ param(
     [string] $testFilter,
     [string] $outDir)
 
-$xunitRunner = $(Join-Path $PSScriptRoot 'packages\xunit.runner.console.2.1.0\tools\xunit.console.exe ')
+# If there is multiple xunit packages installed, take the latest one
+$xunitRunner = Get-ChildItem packages -Directory -Filter "xunit.runner.console.*" | 
+                ForEach-Object { Get-ChildItem $_.FullName -Recurse -File -Filter "xunit.console.exe" } | 
+                Sort-Object -Property VersionInfo | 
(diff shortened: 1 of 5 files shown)


src/Parallel-Tests.ps1
```diff
@@ -0,0 +1,32 @@
+param(
+    [string[]] $assemblies,
+    [string] $testFilter,
+    [string] $outDir)
+
+$xunitRunner = $(Join-Path $PSScriptRoot 'packages\xunit.runner.console.2.1.0\tools\xunit.console.exe ')
```
ReubenBond: It's a little long-winded, but if we want version tolerance & platform independence (join-path), then perhaps something like this?
```ps
$xunitRunner = (Get-Item $([System.IO.Path]::Combine($PSScriptRoot, "packages","xunit.runner.console.*","tools", "xunit.console.exe")) | Sort-Object FullName | Select-Object -Last 1).FullName
```

Maybe that's overkill & maybe there's a better way.

jdom:
I'll fix the build error tomorrow

src/Parallel-Tests.ps1
benjaminpetit: Totally overkill :) I think it should be easier to just have a if/else to choose the path when we will be cross platform. Maybe we will not use the same exec or the same version?

I like the idea to get the latest one, will see if I can come up with a quick solution

benjaminpetit: Last commit should implement your suggestions!

benjaminpetit:
@dotnet-bot test this please

benjaminpetit:
@dotnet-bot test this please

PR #2204, opened 2016-09-22
Remove UseJsonFallbackSerializer

ReubenBond:
Follow-on from #2193.

Json.NET is not a suitable fallback serializer. It does not handle the complex types which it needs to handle.

### DIFF
--- a/src/Orleans/Serialization/SerializationManager.cs
+++ b/src/Orleans/Serialization/SerializationManager.cs
@@ -143,7 +143,7 @@ namespace Orleans.Serialization
 
         #region Static initialization
 
-        public static void InitializeForTesting(List<TypeInfo> serializationProviders = null, bool useJsonFallbackSerializer = false)
+        public static void InitializeForTesting(List<TypeInfo> serializationProviders = null)
         {
             try
             {
@@ -151,7 +151,7 @@ namespace Orleans.Serialization
                 BufferPool.InitGlobalBufferPool(new MessagingConfiguration(false));
                 RegisterSerializationProviders(serializationProviders);
                 AssemblyProcessor.Initialize();
-                fallbackSerializer = GetFallbackSerializer(useJsonFallbackSerializer);
+                fallbackSerializer = GetFallbackSerializer();
             }
             catch (ReflectionTypeLoadException ex)
             {
@@ -159,20 +159,14 @@ namespace Orleans.Serialization
             }
         }
 
-        internal static void Initialize(List<TypeInfo> serializationProviders, bool useJsonFallbackSerializer)
+        internal static void Initialize(List<TypeInfo> serializationProviders)
         {
             RegisterBuiltInSerializers();
-
 #if NETSTANDARD
-            if (!useJsonFallbackSerializer)
-            {
-                logger.Warn(ErrorCode.SerMgr_UnavailableSerializer,
-                    "Cann't use binary formatter as fallback serializer while running on .Net Core, will use Json.Net instead");
-            }
-
-            useJsonFallbackSerializer = true;
+            logger.Warn(ErrorCode.SerMgr_UnavailableSerializer,
+                "Cann't use binary formatter as fallback serializer while running on .Net Core, will use Json.Net instead");
 #endif
-            fallbackSerializer = GetFallbackSerializer(useJsonFallbackSerializer);
+            fallbackSerializer = GetFallbackSerializer();
 
             if (StatisticsCollector.CollectSerializationStats)
             {
@@ -1872,22 +1866,13 @@ namespace Orleans.Serialization
             return retVal;
         }
 
-        private static IExternalSerializer GetFallbackSerializer(bool useJsonSerializer)
+        private static IExternalSerializer GetFallbackSerializer()
         {
-            IExternalSerializer serializer;
-            if (useJsonSerializer)
-            {
-                serializer = new OrleansJsonSerializer();
-            }
-            else
-            {
 #if NETSTANDARD
                 throw new OrleansException("Can't use binary formatter as fallback serializer while running on .Net Core");
 #else
-                serializer = new BinaryFormatterSerializer();
+            var serializer = new BinaryFormatterSerializer();
 #endif
-            }
-
             serializer.Initialize(logger);
             return serializer;
         }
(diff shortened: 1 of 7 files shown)


src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -143,36 +138,26 @@ public static object GetUninitializedObjectWithFormatterServices(Type type)
...
                 throw ex.Flatten();
             }
         }
 
-        internal static void Initialize(List<TypeInfo> serializationProviders, bool useJsonFallbackSerializer)
+        internal static void Initialize(List<TypeInfo> serializationProviders)
         {
             RegisterBuiltInSerializers();
-
-#if NETSTANDARD
```
sergeybykov: Don't we still need this #if and the warning on NETSTANDARD?

src/Orleans/Serialization/OrleansJsonSerializer.cs
```diff
@@ -7,16 +7,20 @@
 
 namespace Orleans.Serialization
 {
-    internal class OrleansJsonSerializer : IExternalSerializer
+    using Orleans.Providers;
+
+    public class OrleansJsonSerializer : IExternalSerializer
```
sergeybykov: Why do we need to make it public?

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: Eep, this entire block should be removed. Thanks for pointing this out!

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -2210,40 +2186,5 @@ public SerializerMethods(DeepCopier deepCopy, Serializer serialize, Deserializer
             public Serializer Serialize { get; private set; }
             public Deserializer Deserialize { get; private set; }
         }
-
-        public static JsonSerializerSettings GetDefaultJsonSerializerSettings()
```
sergeybykov: Don't we still need this for other usages of `OrleansJsonSerializer`, e.g. by storage providers?

test/TesterInternal/Serialization/BuiltInSerializerTests.cs
```diff
@@ -65,8 +59,7 @@ public BuiltInSerializerTests(ITestOutputHelper output)
         }
 
         [Theory, TestCategory("Functional"), TestCategory("Serialization")]
-        [InlineData(SerializerToUse.Default)]
```
sergeybykov: What's this change for?

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: Oh, I misread the diff. I'm not sure that we need a warning which the user cannot do anything about. On the other hand, in `GetFallbackSerializer()`, I'm throwing unconditionally on `NETSTANDARD`, which is clearly untenable. What's the other option, though, other than to wait for #2162 for `NETSTANDARD` support.

We could return a fallback serializer which always throws.

src/Orleans/Serialization/OrleansJsonSerializer.cs
```diff
@@ -7,16 +7,20 @@
...
         private static JsonSerializerSettings defaultSettings;
         private Logger logger;
-
+        
         /// <summary>
         /// Returns a configured <see cref="JsonSerializerSettings"/> 
         /// </summary>
         /// <returns></returns>
-        internal static JsonSerializerSettings GetDefaultSerializerSettings()
+        public static JsonSerializerSettings GetDefaultSerializerSettings()
```
sergeybykov: Same question here.

ReubenBond: This method was essentially moved from `SerializationManager`, but it's used by external providers in order to get serialization settings which are suitable for serializing types which we use (grain refs, etc).

test/TesterInternal/Serialization/BuiltInSerializerTests.cs
ReubenBond: It makes it easier to add & test other serializers (eg, `IlBasedSerializer`) without modifying every method.

In the (legitimate?) case where one serializer is intended to fail and another is intended to succeed, then we can use `InlineData` to specify that.

I'm happy to revert the change and simply remove the do-nothing `Fallback` option.

src/Orleans/Serialization/SerializationManager.cs
sergeybykov: Maybe we should leave the warning here, and let `GetFallbackSerializer()` throw right after?

src/Orleans/Serialization/OrleansJsonSerializer.cs
ReubenBond: Users ought to be able to specify external serializers to use, even if only in certain cases.
This class could be used for storage providers, for example. Currently they use bits of this class (even if they were using it via `SerializationManager`).

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: See the other comment about moving the method. If you feel it's better to keep the methods on SerializationManager, then I can move them back. It seems messy, though.

A cleaner approach would be for those storage providers to consume an `IExternalSerializer` for all of their serialization, rather than each implementing support for JSON + Orleans serialization.

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: No problem, change incoming.

ReubenBond:
> It appears as if there's more than just removing the option of using JSON as a fallback serializer here. But maybe I'm just confused.

You're right, I'll split the other changes into a separate PR. They were changes I made while I was cleaning up the JSON fallback serializer option, but they're not as targeted as the title suggests.

src/Orleans/Serialization/OrleansJsonSerializer.cs
sergeybykov: I don't disagree with your reasoning. But doesn't this mean that all storage providers that use `OrleansJsonSerializer` would have to be updated? If so, then I think it is better to make such a change explicit - as a separate PR, and probably targeting 1.4.0, since we are so close to releasing 1.3.0.

ReubenBond: Agreed

ReubenBond:
Updated - it's much slimmer now.
I retained the removal of the redundant `[InlineData(Fallback)]` attribute in `BuiltInSerializerTests`, since it previously had the same behavior as `JsonFallback`, which is now removed.

PR #2206, opened 2016-09-22
Move JSON serialization methods into OrleansJsonSerializer

ReubenBond:
This is a (minorly) breaking change which moves `GetDefaultJsonSerializerSettings` & `UpdateSerializerSettings` out of `SerializationManager` and into `OrleansJsonSerializer`.

It also makes `OrleansJsonSerializer` public so that external providers can reference it.

### DIFF
--- a/src/Orleans/Serialization/OrleansJsonSerializer.cs
+++ b/src/Orleans/Serialization/OrleansJsonSerializer.cs
@@ -7,16 +7,25 @@ using Orleans.Runtime;
 
 namespace Orleans.Serialization
 {
-    internal class OrleansJsonSerializer : IExternalSerializer
+    using Orleans.Providers;
+
+    public class OrleansJsonSerializer : IExternalSerializer
     {
+        public const string UseFullAssemblyNamesProperty = "UseFullAssemblyNames";
+        public const string IndentJsonProperty = "IndentJSON";
         private static JsonSerializerSettings defaultSettings;
         private Logger logger;
 
+        static OrleansJsonSerializer()
+        {
+            defaultSettings = GetDefaultSerializerSettings();
+        }
+
         /// <summary>
-        /// Returns a configured <see cref="JsonSerializerSettings"/> 
+        /// Returns the default serializer settings.
         /// </summary>
-        /// <returns></returns>
-        internal static JsonSerializerSettings GetDefaultSerializerSettings()
+        /// <returns>The default serializer settings.</returns>
+        public static JsonSerializerSettings GetDefaultSerializerSettings()
         {
             var settings = new JsonSerializerSettings
             {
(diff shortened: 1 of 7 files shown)


src/Orleans/Serialization/OrleansJsonSerializer.cs
```diff
@@ -7,16 +7,25 @@
 
 namespace Orleans.Serialization
 {
-    internal class OrleansJsonSerializer : IExternalSerializer
+    using Orleans.Providers;
```
jdom: nit: please put using statements outside

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
```diff
@@ -91,14 +91,21 @@ public class AzureBlobStorage : IStorageProvider
 
         IEnumerable<string> FormatPropertyMessage(IProviderConfiguration config)
         {
-            foreach (var property in new string[] { ContainerNamePropertyName, "SerializeTypeNames", "PreserveReferencesHandling", "UseFullAssemblyNames", "IndentJSON" })
+            var properties = new[]
+            {
+                ContainerNamePropertyName,
+                "SerializeTypeNames",
+                "PreserveReferencesHandling",
+                OrleansJsonSerializer.UseFullAssemblyNamesProperty,
```
jdom: SerializeTypeNames and PreserveReferencesHandling are bogus, aren't they? we don't seem to be using them at all

src/Orleans/Serialization/OrleansJsonSerializer.cs
```diff
@@ -7,16 +7,25 @@
...
     {
+        public const string UseFullAssemblyNamesProperty = "UseFullAssemblyNames";
+        public const string IndentJsonProperty = "IndentJSON";
         private static JsonSerializerSettings defaultSettings;
         private Logger logger;
 
+        static OrleansJsonSerializer()
+        {
+            defaultSettings = GetDefaultSerializerSettings();
+        }
```
jdom: I think the defaultSettings cached value is an old thing and we do not use it anymore. I'm not suggesting we should use it, because mutating this shared object is not expected but possible (and was causing sneaky side-effects, so that's why we stopped using it).

src/Orleans/Serialization/OrleansJsonSerializer.cs
ReubenBond: I'll fix my StyleCop settings. I think they're  fairly close to the defaults.

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
ReubenBond: They seem to be bogus, I was considering removing them but wasn't 100% certain. I found no occurrences of those strings anywhere in our solution.

I'll nix them.

src/Orleans/Serialization/OrleansJsonSerializer.cs
jdom: I got confused, I guess it's OK... It's not configurable now, but that's orthogonal to this PR, merging

PR #2218, opened 2016-09-23
Optimize the scheduler, remove redundant semaphore and interlocked exchange.

gabikliot:
It would be interesting to know if this saves any perf.

We maintain this semaphore and counter only for the case that we inject more threads dynamically. We don't actually do that any more, for years (its a separate discussion why). Did not want to fully remove the thread injection, but sure no need to pay the semaphore and interlocked exchange price, on a very hot path.

It is one of those things that is hard to predict how perf. will be impacted. I won't be surprised if it does not help at all (semaphore was never blocking, it always had enough room), or saves a ton (since it still inc/dec the semaphore on every task)!

### DIFF
--- a/src/OrleansRuntime/Scheduler/OrleansTaskScheduler.cs
+++ b/src/OrleansRuntime/Scheduler/OrleansTaskScheduler.cs
@@ -355,24 +355,6 @@ namespace Orleans.Runtime.Scheduler
             return Pool.DoHealthCheck();
         }
 
-        /// <summary>
-        /// Action to be invoked when there is no more work for this scheduler
-        /// </summary>
-        internal Action OnIdle { get; set; }
-
-        /// <summary>
-        /// Invoked by WorkerPool when all threads go idle
-        /// </summary>
-        internal void OnAllWorkerThreadsIdle()
-        {
-            if (OnIdle == null || RunQueueLength != 0) return;
-
-#if DEBUG
-            if (logger.IsVerbose2) logger.Verbose2("OnIdle");
-#endif
-            OnIdle();
-        }
-
         internal void PrintStatistics()
         {
             if (!logger.IsInfo) return;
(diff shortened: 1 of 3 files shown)


dVakulen:
I think it should help to some extent, for example at https://cloud.githubusercontent.com/assets/5787619/17674725/6f621da0-632f-11e6-95f8-7e69d793a9c3.png the semaphore appears pretty high in profiling results.

gabikliot:
Anyone interested in reviewing/testing this? Maybe triage and assign reviewer?

sergeybykov:
We'll definitely review and test this soon, now that 1.3.0 is branched out and pretty much done.

ReubenBond:
Changes LGTM. As a side-note, the distinction between `InjectMoreWorkerThreads` and `ShouldInjectWorkerThread` is not immediately clear from the names. A more obvious name for the former might be `CanInjectWorkerThreads` or `EnableWorkerThreadInjection`.

sergeybykov:
I'm ready to merge this. But I like @ReubenBond's suggestion to rename `InjectMoreWorkerThreads` to `EnableWorkerThreadInjection`.

gabikliot:
Sure, will rename.

sergeybykov:
Or I can merge it, and submit the renaming as a separate change.

sergeybykov:
Thanks, Gabi!

gabikliot:
No problem. Did it make any change to the numbers?

sergeybykov:
I didn't see any measurable gains.

sergeybykov:
BTW, please use the "Squash and merge" option instead of "Create a merge commit" when merging PRs that consist of a single commit or should be squashed, and "Rebase and merge" (new GitHub feature) for those PRs where we want to preserve the individual commits. These two options avoid merge commits, and keep the history less cluttered. Not a big deal, just nice features that GitHub recently added.

gabikliot:
Will do.

PR #2219, opened 2016-09-23
Remove CancellationTokenCallbacksThrow_ExceptionShouldBePropagated test from BVT/Functional sets

sergeybykov:
Remove CancellationTokenCallbacksThrow_ExceptionShouldBePropagated test from BVT/Functional sets, as it's failing intermittently on a regular basis.

/cc @dVakulen.


dVakulen:
Was going to do the same. Will try to find the root cause.

PR #2220, opened 2016-09-23
CodeGen: Avoid wrapping IGrainMethodInvoker.Invoke body in try/catch

ReubenBond:
When code generation was originally ported from CodeDOM to Roslyn, the initial goal was to keep the emitted code close to the previous generator's.

Currently, generated `IGrainMethodInvoker.Invoke` methods have a try/catch wrapping the body.

This try/catch is needless and this PR removes it.
Example old body:
```C#
global::System.Int32 interfaceId = @request.@InterfaceId;
global::System.Int32 methodId = @request.@MethodId;
global::System.Object[] arguments = @request.@Arguments;
try
{
    if (@grain == null)
        throw new global::System.ArgumentNullException("grain");
    switch (interfaceId)
    {
        case -831689659:
            switch (methodId)
            {
                case -1573596583:
                    return ((global::Orleans.IRemindable)@grain).@ReceiveReminder((global::System.String)arguments[0], (global::Orleans.Runtime.TickStatus)arguments[1]).@Box();
                default:
                    throw new global::System.NotImplementedException("interfaceId=" + -831689659 + ",methodId=" + methodId);
            }
 
        default:
            throw new global::System.NotImplementedException("interfaceId=" + interfaceId);
    }
}
catch (global::System.Exception exception)
{
    return global::Orleans.Async.TaskUtility.@Faulted(exception);
}
```

Example new body:
```C#
global::System.Int32 interfaceId = @request.@InterfaceId;
global::System.Int32 methodId = @request.@MethodId;
global::System.Object[] arguments = @request.@Arguments;
if (@grain == null)
    throw new global::System.ArgumentNullException("grain");
switch (interfaceId)
{
    case -831689659:
        switch (methodId)
        {
            case -1573596583:
                return ((global::Orleans.IRemindable)@grain).@ReceiveReminder((global::System.String)arguments[0], (global::Orleans.Runtime.TickStatus)arguments[1]).@Box();
            default:
                throw new global::System.NotImplementedException("interfaceId=" + -831689659 + ",methodId=" + methodId);
        }

    default:
        throw new global::System.NotImplementedException("interfaceId=" + interfaceId);
}
```

### DIFF
--- a/src/OrleansCodeGenerator/GrainMethodInvokerGenerator.cs
+++ b/src/OrleansCodeGenerator/GrainMethodInvokerGenerator.cs
@@ -228,23 +228,8 @@ namespace Orleans.CodeGenerator
                         grainArgument,
                         SF.LiteralExpression(SyntaxKind.NullLiteralExpression)),
                     SF.ThrowStatement(argumentNullException));
-
-            // Wrap everything in a try-catch block.
-            var faulted = (Expression<Func<Task<object>>>)(() => TaskUtility.Faulted(null));
-            const string Exception = "exception";
-            var exception = SF.Identifier(Exception);
-            var body =
-                SF.TryStatement()
-                    .AddBlockStatements(grainArgumentCheck, interfaceIdSwitch)
-                    .AddCatches(
-                        SF.CatchClause()
-                            .WithDeclaration(
-                                SF.CatchDeclaration(typeof(Exception).GetTypeSyntax()).WithIdentifier(exception))
-                            .AddBlockStatements(
-                                SF.ReturnStatement(
-                                    faulted.Invoke().AddArgumentListArguments(SF.Argument(SF.IdentifierName(Exception))))));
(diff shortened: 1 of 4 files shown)


jdom:
Hmmm, I wonder what would happen if the grain method throws synchronously with a return type of `Task<object>` (which doesn't use await internally for boxing... that might throw synchronously here, and then the try/catch would actually be needed.

Can you add a test similar to `SynchronousExceptionThrownShouldResultInFaultedTask` but with a method call that returns `Task<object>` to verify this?

ReubenBond:
Sure thing, @jdom, check the latest - extended test passes.

`InsideRuntimeClient.Invoke` already surrounds the call with a try/catch

jdom:
Awesome, and thanks for adding the test, because with 4.6 we'll be able to optimize part of this some more, so having coverage would be great

PR #2223, opened 2016-09-24
Connection string fallbacks for testing

jdom:
Refactoring our TestCluster infrastructure to use the `Microsoft.Extensions.Configuration` package to be able to provide fallbacks for our secrets (connection strings) and also other default config settings.
In our different build environments then we can have environmental variable or JSON files with these settings, making it a lot more flexible.


veikkoeeva:
@jdom I can't make time to look into this currently, but `Microsoft.Extensions.Configuration` is what I think is the right direction. The other thing is that the tests could make use of them, which is easier if configuration and environment setup are decoupled from class constructs, i.e. they handle only their own invariants in the constructors. Maybe just my opionion, though.

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -39,20 +49,57 @@ public TestClusterOptions(short initialSilosCount)
...
+        }
+
         /// <summary>
         /// The default initial silos count. See <see cref="TestClusterOptions"/>.
         /// </summary>
-        public static short DefaultInitialSilosCount { get; set; } = 2;
+        public static short GetInitialSilosCount(IConfiguration configuration)
+        {
+                short defaultValue;
+                return short.TryParse(configuration[nameof(ExtendedConfigurationOptions.InitialSilosCount)], out defaultValue)
```
jason-bragg: This is of course supported by the config lib, but I think a more maintainable means of doing this is to use ConfigurationBuilder.Bind to bind configured values directly to instance properties.

src/OrleansTestingHost/Extensions/TestConfigurationExtensions.cs
```diff
@@ -14,37 +14,40 @@ public static class TestConfigurationExtensions
         /// <summary>
         /// This call tweaks the cluster config with settings specific to a test run.
         /// </summary>
-        public static void AdjustForTestEnvironment(this ClusterConfiguration clusterConfig)
+        public static void AdjustForTestEnvironment(this ClusterConfiguration clusterConfig, string dataConnectionStringFallback)
```
jason-bragg: not sure having this override is good approach.  Following this pattern.  Opens the door to other overrides, which over time could make this code very hard to maintain.  Suggest making caller is responsible for setting the connection string in the cluster config.

src/OrleansTestingHost/Extensions/TestConfigurationExtensions.cs
```diff
@@ -14,37 +14,40 @@ public static class TestConfigurationExtensions
...
             }
-
             AdjustProvidersDeploymentId(clusterConfig.Globals.ProviderConfigurations, "DataConnectionString", clusterConfig.Globals.DataConnectionString);
         }
 
         /// <summary>
         /// This call tweaks the client config with settings specific to a test run.
         /// </summary>
-        public static void AdjustForTestEnvironment(this ClientConfiguration clientConfiguration)
+        public static void AdjustForTestEnvironment(this ClientConfiguration clientConfiguration, string dataConnectionStringFallback)
```
jason-bragg: same as other AdjustForTestEnvironment.  Discourage the use of an override.

src/OrleansTestingHost/Extensions/TestConfigurationExtensions.cs
jdom: well, this IS that code you are talking about, the one responsible to setting the connection string in the cluster config. I'll play with it a little bit to see how I can improve it further

src/OrleansTestingHost/TestClusterOptions.cs
jdom: nice, I thought the options stuff was only available if using the DI binding stuff, but you made me look again and found the `Microsoft.Extensions.Configuration.Binder` nuget package, and it has `ConfigurationBinder.Bind` method, which is probably what you were referring to. I'll update it to use this

src/OrleansTestingHost/Extensions/TestConfigurationExtensions.cs
jason-bragg: Ah, I see.  I'll leave this to your discression, but I'd hope we've a better means of overriding the settings than including each setting we want to change in this calls arguements.

xiazen:
Are you going to take a look of this #2047 too? ;) related to Microsoft.Extensions.Configuration and Adam dropped it because that he changed jobs 
Will look at the PR tomorrow.

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -19,22 +19,62 @@ namespace Orleans.TestingHost
...
+
+            /// <summary>
+            /// Default subfolder the the logs
+            /// </summary>
+            public string LogsFolder { get; set; }
+
+            /// <summary>
+            /// Default data connection string to use in tests
+            /// </summary>
+            public string DataConnectionString { get; set; }
```
jason-bragg: If this is mainly about secrets, where is the EventHub info?

jdom:
@veikkoeeva thanks for the suggestion about moving the config stuff to the test projects themselves and decouple the TestCluster from it. It does make sense and tried to do it, but we currently have the `TestingSiloHost` that we are in the process of deprecating, but until we do so, doing that dependency inversion is much trickier. I created #2232 to get us there, and we can iterate once more once it's been fully deprecated.
For now this PR just replaces our current custom hacky approach with using Microsoft.Configuration.Extensions to perform a similar fallback to what we were doing. What you suggest can be the next incremental step. We are also considering having the cluster config itself bindable, so if that happens soon, then we would refactor the current approach regardless.

src/OrleansTestingHost/TestClusterOptions.cs
jdom: EventHub is only in `TestDefaultConfiguration`, since the TestCluster itself does not care about it, only our tests (which is in the same vein as what @veikkoeeva was suggesting, which we should not pollute the main testing helper package with our own internal testing configuration).

src/OrleansTestingHost/Extensions/TestConfigurationExtensions.cs
jdom: We can remove this method once we deprecate `TestingSiloHost`. For now it's code that is shared between `TestClusterOptions` and `TestingSiloHost`. It won't grow in parameter count with more overrides, as we are adding setting more values in the ClusterConfiguration and ClientConfiguration objects directly, and we are less and less relying on this kind of auto-magic replacements (which were most practical with XML configuration, but not that needed with programmatic)

jdom:
@dotnet-bot test this please

jdom:
This PR should be ready for final review (at least it's no longer work in progress until I get more feedback).
I addressed some comments (such as using ConfigurationBinder) and polished the solution some more.

jdom:
Ping, I'd like if this can be reviewed, as it would really ease running functionals

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -33,26 +83,41 @@ public TestClusterOptions()
...
-            this.InitialSilosCount = initialSilosCount;
-            this.BaseSiloPort = ThreadSafeRandom.Next(22300, 30000);
-            this.BaseGatewayPort = ThreadSafeRandom.Next(40000, 50000);
+            this.ExtendedFallbackOptions.InitialSilosCount = initialSilosCount;
         }
 
+
         /// <summary>
-        /// The default initial silos count. See <see cref="TestClusterOptions"/>.
+        /// Create a new TestClusterOptions
```
ReubenBond: Not related to this PR, but if we decide on StyleCop settings, then there's usually a set [constructor doc string](http://stylecop.soyuz5.com/SA1642.html)

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -33,26 +83,41 @@ public TestClusterOptions()
...
+            extendedConfiguration.Bind(fallbackOptions);
+            return fallbackOptions;
+        }
 
         /// <summary>
-        /// Default subfolder the the logs
+        /// Gets or sets fallback options in the case that some configuration settings are not explicitly provided by the user, such as the <see cref="GlobalConfiguration.DataConnectionString"/>
         /// </summary>
-        public static string DefaultLogsFolder { get; set; } = "logs";
+        public FallbackOptions ExtendedFallbackOptions { get; set; }
```
ReubenBond: Could this be made get only?

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -17,14 +19,62 @@ namespace Orleans.TestingHost
     [Serializable]
     public class TestClusterOptions
     {
+        /// <summary>
+        /// Extended options to be used as fallbacks in the case that explicit options are not provided by the user.
+        /// </summary>
+        public class FallbackOptions
+        {
+            /// <summary>
+            /// Is set to true, by default the cluster will output traces in the console
```
ReubenBond: typo: Is -> If

Also, should we be using [the standard](http://stylecop.soyuz5.com/SA1623.html) "Gets or sets a value indicating whether the default cluster will output traces to the console."? I know it seems long-winded, but I wonder if we should decide on style standards for docs, layout, etc. If we do, we can mostly adhere to them automatically using ReSharper + StyleCop + some other plugins (AgentSmith, I think?)

veikkoeeva: I concur.

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -17,14 +19,62 @@ namespace Orleans.TestingHost
...
+            /// </summary>
+            public static IConfiguration DefaultExtendedConfiguration { get; set; } = DefaultConfigurationBuilder().Build();
+        }
+
         private ClusterConfiguration _clusterConfiguration;
         private ClientConfiguration _clientConfiguration;
 
         /// <summary>
-        /// Create a new TestClusterOptions with the default initial silo count. See <see cref="DefaultInitialSilosCount"/>.
+        /// Create a new TestClusterOptions with the default initial silo count. See <see cref="ExtendedFallbackOptions"/>.
```
xiazen: Looks good to me. Except some minor comments. Testing code looks much cleaner now. I like it.

jdom:
Addressed comments related to XML. Had to rebase on top of current master due to conflicts.
When ready to merge, please use the squash option, as the intermediate commits are of no value other than to facilitate the review process.

xiazen:
this pull request broke vNext Orleans.TestingHost project compiling. 

Did a fix in #2253 , in case  you(@jdom) found out and did the same thing as me, :)

PR #2227, opened 2016-09-27
port Serialization test to vNext

xiazen:
Port part of serialization tests to vNext. 

remaining serialization tests cannot be ported over because missing of TestGrainInterface project. @jdom teach me how to create the correct csproj for vNext? 

and also, do we want a non-silo-test-internal, regards of the accessibility to test internal types? More clean this way.

### DIFF
--- a/vNext/test/NonSiloTests/project.json
+++ b/vNext/test/NonSiloTests/project.json
@@ -1,4 +1,4 @@
-{
+﻿{
   "dependencies": {
     "xunit": "2.1.0",
     "xunit.runner.visualstudio": "2.1.0",
--- a/vNext/src/Orleans/Shims/AppDomain.cs
+++ b/vNext/src/Orleans/Shims/AppDomain.cs
@@ -38,7 +38,6 @@ namespace Orleans
             {
                 typeof(Exception).GetTypeInfo().Assembly,
                 typeof(AssemblyProcessor).GetTypeInfo().Assembly,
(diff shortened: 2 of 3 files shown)


test/Tester/SerializationTests/ProtobufSerializationTests/ProtobufSerializationTests.cs
```diff
@@ -10,10 +10,13 @@ namespace UnitTests.Serialization
 {
     public class ProtobufSerializationTests
     {
+#if !NETSTANDARD_TODO
```
ReubenBond: Should this #if be around the whole class if Protobuf doesn't support .NET Core yet?

vNext/test/NonSiloTests/project.json
```diff
@@ -1,5 +1,6 @@
-{
+﻿{
   "dependencies": {
+    "Google.Protobuf": "3.0.0",
```
jdom: why is this? we do not have that google projects migrated yet

xiazen: pulled in for ProtobufSerializationTests. Or should I migrate google now?

jdom: as we talked personally, but replying here in case someone was following, tests should pass in the crawl phase too. The difference is that tests will run in .NET Core too in the walk phase, while now we'll just run them in full .NET Framework referencing the .NET Standard binaries.

PR #2234, opened 2016-09-28
port SchedulerTests and SerializerTests and ConfigTests to vNext

xiazen:
1. Tests not inside !NETSTANDARD_TODO builds and pass successfully
2. For tests which failed , I marked it with !NETSTANDARD_TODO and also its blocker, based on my short investigation
3. @ReubenBond  I port over more serialization tests, I think you will be interested to take a look  I saw your message on Gitter.


vNext/test/NonSiloTests/Orleans.NonSiloTests.csproj
```diff
@@ -79,29 +111,103 @@
...
+    </Content>
+    <Content Include="SerializationTests\ConfigurationTests\ClientConfigurationForSerializer3.xml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="SerializationTests\ConfigurationTests\ClientConfigurationForSerializer4.xml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="SerializationTests\ConfigurationTests\ClientConfigurationForSerializer5.xml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
```
xiazen: I copied over the files, instead of link them, because it need a slightly different config in vNext. See below

vNext/test/NonSiloTests/Orleans.NonSiloTests.csproj
```diff
@@ -79,29 +111,103 @@
...
+    <Content Include="SerializationTests\ConfigurationTests\ClientConfigurationForSerializer.xml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="SerializationTests\ConfigurationTests\ClientConfigurationForSerializer3.xml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="SerializationTests\ConfigurationTests\ClientConfigurationForSerializer4.xml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="SerializationTests\ConfigurationTests\ClientConfigurationForSerializer5.xml">
```
xiazen: I copied over the files, instead of link them, because it need a slightly different config in vNext. See below

vNext/test/NonSiloTests/SerializationTests/ConfigurationTests/ClientConfigurationForSerializer.xml
```diff
@@ -0,0 +1,14 @@
...
+  <Gateway Address="localhost" Port="40000"/>
+  <!-- To turn tracing off, set DefaultTraceLevel="Off" and have no overrides.
+    For the trace log file name, {0} is replaced by "Client" and {1} is the current time. -->
+  <Tracing DefaultTraceLevel="Info" TraceToConsole="false" TraceToFile="{0}-{1}.log" BulkMessageLimit="1000">
+  </Tracing>
+  <Statistics MetricsTableWriteInterval="300s" PerfCounterWriteInterval="30s" LogWriteInterval="300s" WriteLogStatisticsToTable="true" StatisticsCollectionLevel="Info"/>
+  <Messaging ResponseTimeout="30s" ClientSenderBuckets="8192" MaxResendCount="0">
+    <SerializationProviders>
+      <Provider type="UnitTests.Serialization.FakeSerializer, Orleans.NonSiloTests"/>
+    </SerializationProviders>
```
xiazen: Copied over the entire config file because FakeSerializer is in different assembly in vNext. Open to suggestion about 
1. link vs copy
2. port them to another project which share the same assembly name as main solution.

test/Tester/SerializationTests/SerializationTests.ImmutableCollections.cs
```diff
@@ -19,7 +19,8 @@ void RoundTripCollectionSerializationTest<T>(IEnumerable<T> input)
             var output = SerializationManager.RoundTripSerializationForTesting(input);
             Assert.Equal<T>(input,output);
         }
-
+#if !NETSTANDARD_TODO
+    //blocked by Cannot use binary formatter as fallback serializer while running on .Net Core
```
jdom: this is not due to binary formatter, but due to not loading the right serializers, and ends up trying to use the fallback serializer (which is binary formatter). Probably most of these are related to that... There should only be a few tests that use the binary formatter purposely

test/TesterInternal/ConfigTests.cs
```diff
@@ -833,7 +838,8 @@ public void ServerConfig_LoadFrom()
 
             Assert.Equal(filename, Path.GetFileName(config.SourceFile)); // OrleansConfiguration.SourceFile
         }
-
+#if !NETSTANDARD_TODO
+        //blocked by cannot update System.Xml.XmlDocument to 4.0.1 which is compatible with .net framework 4.6
```
jdom: hmmm, very strange. Let's see if we can quickly fix the dependency instead of marking a lot of related parts of the code with TODO

vNext/test/NonSiloTests/SerializationTests/ConfigurationTests/ClientConfigurationForSerializer.xml
jdom: I guess it's OK for now to copy them

test/Tester/SerializationTests/SerializationTests.ImmutableCollections.cs
xiazen: AHHHH~   will take a second look. wasn't looking too much in this path.

xiazen: You are partially right. in InitializeForTesting method, it's trying to GetFallbackSerializer by default. That failed a bunch of tests. I'm gonna mark that time using TODO. and remove others.

xiazen: fixed it :)

test/TesterInternal/ConfigTests.cs
xiazen: come over anytime . 
But the exception is 
Test Name:	UnitTests.ConfigTests.Config_StorageProviders_AzureTable_Default
Test FullName:	UnitTests.ConfigTests.Config_StorageProviders_AzureTable_Default
Test Source:	D:\microsoft\orleans-xiazen\test\TesterInternal\ConfigTests.cs : line 731
Test Outcome:	Failed
Test Duration:	0:00:00.284

Result StackTrace:	
at Orleans.Runtime.Configuration.ProviderConfiguration.LoadProviderConfigurations(XmlElement root, XmlNamespaceManager nsManager, IDictionary`2 alreadyLoaded, Action`1 add)
   at Orleans.Runtime.Configuration.ProviderCategoryConfiguration.Load(XmlElement child) in D:\microsoft\orleans-xiazen\src\Orleans\Configuration\ProviderConfiguration.cs:line 221
   at Orleans.Runtime.Configuration.GlobalConfiguration.Load(XmlElement root) in D:\microsoft\orleans-xiazen\src\Orleans\Configuration\GlobalConfiguration.cs:line 969
   at Orleans.Runtime.Configuration.ClusterConfiguration.LoadFromXml(XmlElement root) in D:\microsoft\orleans-xiazen\src\Orleans\Configuration\ClusterConfiguration.cs:line 99
   at Orleans.Runtime.Configuration.ClusterConfiguration.Load(TextReader input) in D:\microsoft\orleans-xiazen\src\Orleans\Configuration\ClusterConfiguration.cs:line 86
   at Orleans.Runtime.Configuration.ClusterConfiguration.LoadFromFile(String fileName) in D:\microsoft\orleans-xiazen\src\Orleans\Configuration\ClusterConfiguration.cs:line 204
   at UnitTests.ConfigTests.Config_StorageProviders_AzureTable_Default() in D:\microsoft\orleans-xiazen\test\TesterInternal\ConfigTests.cs:line 735
Result Message:	System.IO.FileLoadException : Could not load file or assembly 'System.Xml.XmlDocument, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)

xiazen:
Port over CustomSerializerTests to vNext, the remaining SerializerTests which doesn't depend on silo.  

You can merge this one by squash both commit, and just with the first commit message.

vNext/src/Orleans/Shims/ClassesForTesting/CustomSerializerTestClasses.cs
```diff
@@ -0,0 +1,116 @@
+﻿using System;
+using Orleans.CodeGeneration;
+using Orleans.Serialization;
+
+namespace UnitTests.CustomSerializerTestClasses
```
jdom: what's this for? seems strange to have a shim that is only used for testing? why is this not part of the testing code?

test/TesterInternal/DevTestClientConfiguration.xml
```diff
@@ -1,7 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <ClientConfiguration xmlns="urn:orleans">
   <SystemStore SystemStoreType="SqlServer"
-               DataConnectionString="connection string"/>
+               DataConnectionString="connection string"
+               DeploymentId="deployment id"/>
```
jdom: Let's fix this in a separate PR. Not sure if putting a dummy ID is what we want in this case

vNext/src/Orleans/Shims/ClassesForTesting/CustomSerializerTestClasses.cs
xiazen: This is for CustomSerializerTests class. I added those types to AppDomain.GetAssemblely() method so SerializationManager can map serializer for them. They are in shim folder because they are related to AppDomain. Create them in other project would end up add reference for Orleans project.

test/TesterInternal/DevTestClientConfiguration.xml
xiazen: It is used in config test, the test was just testing loading config from a file works or not. And the test is failing in main solution too. So I just fixed it. 

Let's talk about it tomorrow.

vNext/src/Orleans/Shims/ClassesForTesting/CustomSerializerTestClasses.cs
jdom: Let's look at an alternative tomorrow, basically by modifying the App domain shim itself

xiazen: Updated with an alternative. Like it more? ;)

vNext/src/Orleans/Shims/AppDomain.cs
```diff
@@ -41,16 +39,9 @@ public Assembly[] GetAssemblies()
...
-                typeof(SerializerTestClass2).GetTypeInfo().Assembly,
-                typeof(SerializerTestClass3).GetTypeInfo().Assembly,
-                typeof(SerializerTestClass4).GetTypeInfo().Assembly,
-                typeof(SerializerTestClass5).GetTypeInfo().Assembly,
-                typeof(AsyncObserverArg).GetTypeInfo().Assembly,
-                typeof(StreamSubscriptionHandleArg).GetTypeInfo().Assembly,
-                typeof(GenericArg).GetTypeInfo().Assembly
+                Assembly.Load(new AssemblyName("Orleans.NonSiloTests")),
+                Assembly.Load(new AssemblyName("TestInternalGrains")),
+                Assembly.Load(new AssemblyName("TestGrainInterfaces"))
```
jdom: What if you just have a SetAssemblies in this stub, that way, you can call from your tests something like this before initializing:
`Orleans.AppDomain.SetAssemblies(System.AppDomain.GetAssemblies()`.
This is a temporary hack, I understand, but it will work with other test projects too

xiazen: Yes, that will work too and more flexible for tests to load only Assemblies they are interested in. 

But since this is a hack, we would end up using `SetAssemblies `method to many tests, while we shouldn't ideally. So why not gather all the code change related to this Hack and put it inside AppDomain. So when the hack can be removed, we just fix AppDomain and we are good. IMO

jdom: Sure. Loading the nonsilotests assembly might cause issues when we have other test projects, but we can deal with that later. Perhaps by then we have the assembly loader stuff done already. 
Looks good, I'll do a final review tomorrow and merge

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -145,7 +145,10 @@ public static void InitializeForTesting(List<TypeInfo> serializationProviders =
                 BufferPool.InitGlobalBufferPool(new MessagingConfiguration(false));
                 RegisterSerializationProviders(serializationProviders);
                 AssemblyProcessor.Initialize();
+#if !NETSTANDARD_TODO
+                 //blocked by Cannot use binary formatter as fallback serializer while running on .Net Core
                 fallbackSerializer = GetFallbackSerializer();
+#endif
```
jdom: there's already an #if in the GetFallbackSerializer method itself, please change that to return null if that's what you want to do termporarily.

vNext/src/Orleans/Shims/AppDomain.cs
```diff
@@ -38,6 +38,10 @@ public Assembly[] GetAssemblies()
             {
                 typeof(Exception).GetTypeInfo().Assembly,
                 typeof(AssemblyProcessor).GetTypeInfo().Assembly,
+                // load up classes for testing, until Assembly scanning is supported in vNext
+                Assembly.Load(new AssemblyName("Orleans.NonSiloTests")),
```
jdom: is Orleans.NonSiloTests required? because that's my only concern with this approach. Still, if it is indeed required we can still merge for now and then fix in when we have more test projects, so it doesn't break those.

vNext/test/NonSiloTests/SerializationTests/ConfigurationTests/ClientConfigurationForSerializer.xml
jdom: I think I miss-understood this a little bit, and will be a lot more common (a ton more of these files) when we port silo tests in the upcoming days. I guess we'll start deviating a lot if we copy these files.
I think our best alternatives are:
1. Create an Orleans.NonSiloTests project in the main solution (which I believe it's valuable regardless)
2. port them to another project which share the same assembly name as main solution. (what you suggested before).

I believe 1 is best, what do you think?

jdom: Ok, we decided to do alternative 1, but we'll doing in an upcoming PR soon

PR #2238, opened 2016-09-29
Replace Expression Trees with IL

ReubenBond:
Three commits with the same purpose: replacing usage of [LINQ Expression Tree](https://msdn.microsoft.com/en-us/library/mt654263.aspx) compilation with IL, since [this blog post about .NET Standard](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/) indicates that Expression.Compile will be interpreted (rather than compiled to IL) in upcoming versions of the platform:
> In some cases we’ll emulate their behavior (e.g. interpreting expression trees instead of compiling them) while in other cases we’ll throw (e.g. when compiling regexes).

The mixed use of LINQ Expressions & IL has also been bugging me for a while.

As a part of this work, this change removes `GrainFactory.CastInternal` and instead calls `GrainExtensions.AsReference<T>` as needed. I removed that method because calling it from IL was complicated because one of the parameters was a `Func<IAddressable, GrainReference>`, which is difficult to call from IL (requires generating a type with static fields/methods & passing function pointer to ctor for `Func<..>`) & it wasn't strictly necessary. The result is simpler, in my opinion.

Let me know what you think.

### DIFF
--- a/src/Orleans/Core/GrainCasterFactory.cs
+++ b/src/Orleans/Core/GrainCasterFactory.cs
@@ -1,0 +1,131 @@
+namespace Orleans
+{
+    using System;
+    using System.Linq;
+    using System.Reflection;
+    using System.Reflection.Emit;
+
+    using Orleans.Runtime;
+
+    internal static class GrainCasterFactory
+    {
+        /// <summary>
+        /// The cached <see cref="MethodInfo"/> for <see cref="GrainExtensions.AsWeaklyTypedReference"/>.
+        /// </summary>
+        private static readonly MethodInfo GrainReferenceCastHelperMethodInfo =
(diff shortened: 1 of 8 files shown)


ReubenBond:
Ran functionals, everything looks clear.

sergeybykov:
I kicked off load/reliability tests for just in case. Although I don't expect any impact from this.

sergeybykov:
I see that `TimerOrleansTest_Migration` test failed for some reason, and I was able to repro its failure locally.

sergeybykov:
TimerOrleansTest_Migration failed in nightly functional run also. So it must be unrelated.

PR #2239, opened 2016-09-29
Add TestCategory to SiloInitializationIsRetryableTest

ReubenBond:


jdom:
Hmmm, I didn't know about this test... this needs to be updated to use the new `AppDomainSiloHost`. Oh well

jdom:
Or even simpler have the entire test run in a separate AppDomain with a callback, and just return test status at the end.

PR #2240, opened 2016-09-30
Updated MemoryStreamProvider to support custom message serialization

jason-bragg:
Defaults to using Orleans serialization manager.
This allows the stream provider to be customized to send events that are not natively serializable over Orleans.


src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
```diff
@@ -25,7 +25,6 @@ namespace Orleans.Providers.Streams.Common
     /// <typeparam name="TQueueMessage">Queue specific data</typeparam>
     /// <typeparam name="TCachedMessage">Tightly packed cached structure.  Should only contain value types.</typeparam>
     public class PooledQueueCache<TQueueMessage, TCachedMessage>
-        where TQueueMessage : class
```
xiazen: why remove this?

src/OrleansProviders/Streams/Memory/MemoryAdapterReceiver.cs
```diff
@@ -1,18 +1,19 @@
...
 
-namespace Orleans.Providers.Streams.Memory
+namespace Orleans.Providers
 {
-    internal class MemoryAdapterReceiver : IQueueAdapterReceiver
+    internal class MemoryAdapterReceiver<TSerializer> : IQueueAdapterReceiver
+        where TSerializer : IMemoryMessageBodySerializer, new()
     { 
-        private IMemoryStreamQueueGrain queueGrain;
-        private long sequenceId;
```
xiazen: why is sequenceId removed?

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
jason-bragg: Constraint was not necessary, and I needed to use a structure for the TQueueMessage.  Constraint was originally added because type needed to be allowed to be null, but that was a very early iteration.  No reason for it now.

src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
```diff
@@ -1,26 +1,22 @@
-﻿using System.Collections.Generic;
+﻿
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
-namespace Orleans.Providers.Streams.Memory
+namespace Orleans.Providers
```
JingqiaoFu: What's the thinking of removing .Stream.Memory? GeneratorStreamProvider has .Stream.Generator in its namespace.

src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
```diff
@@ -1,26 +1,22 @@
...
 {
     /// <summary>
     /// Interface for In-memory stream queue grain.
     /// </summary>
     public interface IMemoryStreamQueueGrain : IGrainWithGuidKey
     {
         /// <summary>
-        /// Set max event count. The total number of events enqueued can not exceed maxEventCount.
-        /// </summary>
-        Task SetMaxEventCount(int maxEventCount);
```
JingqiaoFu: So we will not allow customized max queue size?

src/OrleansProviders/Streams/Memory/MemoryMessageData.cs
```diff
@@ -0,0 +1,39 @@
...
+    public struct MemoryMessageData
+    {
+        /// <summary>
+        /// Stream Guid of the event data.
+        /// </summary>
+        public Guid StreamGuid;
+        /// <summary>
+        /// Hash code of namespace. Stores hash for fast and efficient storage.
+        /// </summary>
+        public string StreamNamespace;
```
JingqiaoFu: Since it's not actually using the hash of namespace here, the comment should be changed. However, I think it's reasonable to consider using Hash since MemoryMessageData is also used as cache structure.

src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
jason-bragg: Generally we try to keep namespaces short and simple in Orleans core assemblies.  Usually 'orleans' or 'oreans.runtime'.  Since this is in providers I felt Orleans.providers was appropriate.  The long namespaces in generator was my mistake.   I considered changing it with this PR but felt it was orthogonal, though I do think we should change the generators namespaces as well.

src/OrleansProviders/Streams/Memory/IMemoryStreamQueueGrain.cs
jason-bragg: It's not that we will not allow it, as much as there was no usage of it in this case and no clear way for a user to set it.  For instance, it was not configurable, or directly settable.  With no clear story or tests for this case, I concluded it was best not to have the code around until we need it.  Great thing about source control is that the feature is not really gone, we can restore it when it is needed and/or we have the time to fully flesh out the feature.  Unused and untested code is unnecessary maintenance overhead.

src/OrleansProviders/Streams/Memory/MemoryStreamQueueGrain.cs
```diff
@@ -1,57 +1,37 @@
...
-        /// <summary>
-        /// Type name. Default to the grain name.
-        /// </summary>
-        public const String TypeName = "MemoryStreamQueueGrain";
-
-        /// <summary>
-        /// Event queue. 
-        /// </summary>
-        private Queue<MemoryEventData> eventQueue = new Queue<MemoryEventData>();
+        private readonly Queue<MemoryMessageData> eventQueue = new Queue<MemoryMessageData>();
```
JingqiaoFu: Keep/add comments to avoid compile time warnings.

jason-bragg: The field is private, so it shouldn't generate warnings...  I'll double check.  We don't tend to add comments for private fields unless they are unclear.  This field seems clear, but that's just my opinion.  If you disagree, I'll happily add the comment back.

JingqiaoFu: I remember adding comments based solely on the warnings I got on VS, so I think we might have to do that.

src/OrleansProviders/Streams/Memory/MemoryMessageData.cs
jason-bragg: Comment updated.
Going to stick with string rather than hash, because it would complicate the serialization story and performance is less of a concern for this provider as it's intended for test.  If/when GC because an issues, we will address it.

jason-bragg:
PR updated.
Comments addressed.

Issue #2196, opened 2016-09-21
.NET Standard compat: Avoid Silo being MarshalByRefObject but support running tests


jdom:
In our .NET Standard solution I temporarily removed the `MarshalByRefObject` inheritance from the `Silo` class, as it's not supported in .NET Standard. See https://github.com/dotnet/orleans/commit/0a80ae6ce08dfb5da304cbd64690ece1eecc4b81 for the commit where this was excluded (ignore the test hooks for now, as this work is being done independently and is already in progress by @galvesribeiro).

Nevertheless, the testing infrastructure in the OrleansTestinhHost project (that contains TestCluster) relies on it. This project is still targeting .NET 4.6 in the vNext solution, and is able to use AppDomains, nevertheless it should not rely on the Silo class itself to be the glue between the AppDomains, and instead rely in something like `AppDomainSiloHost` (that can live in this test project) to inherit from `MarshalByRefObject` and use the Silo.

In order to open the solution that compiles using .NET Standard (and hence easy to check what is available in that environment), check out [this readme](https://github.com/dotnet/orleans/blob/master/vNext/README-CoreClr-port.md)

jdom:
I'm grabbing this one. Will have a PR out tomorrow, but wanted to ping this issue to avoid duplicate effort.

jdom:
Implemented in #2241

PR #2241, opened 2016-09-30
Move AppDomain testing stuff to OrleansTestingHost project

jdom:
This implements #2196 
Move AppDomain testing stuff to OrleansTestingHost project. This is because Orleans will be compiled as .NET Standard, while OrleansTestingHost will continue to be .NET Framework for now (so it can use AppDomains). This will evolve after the CRAWL phase to not rely on AppDomains.
Also abstract it away so that it can be replaced with other strategies in the future (such as processes, containers, etc).


src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -1054,6 +1038,16 @@ internal void SetMaxForwardCount_ForTesting(int val)
                 silo.OrleansConfig.Globals.MaxForwardCount = val;
             }
 
+            internal void LatchIsOverloaded(bool overloaded)
+            {
+                this.silo.Metrics.LatchIsOverload(overloaded);
```
sergeybykov: >LatchIsOverload [](start = 34, length = 15)

Typo in the method name?

jdom: Yes, but I didn't touch it to avoid unnecessary changes. I can fix if you prefer

sergeybykov: Either way is fine with me.

jdom: K, let's merge as is then. We can decide to fix later if it doesn't bring braking changes

PR #1918, opened 2016-07-08
Fixes for non-Silo unit testing of  Grain<T> WriteStateAsync()

dsarfati:
During some of our non-silo testing we were encountering null pointer exceptions during WriteStateAsync(). This was due to the fact that the full runtime is not being initialized (since its a test) so there were some methods that were not checking for null values. This should fix all the Grain<T> issues encountered with the non-silo unit testing that was added in 1.2.2

PR #1919, opened 2016-07-08
Non-silo unit test examples in Samples\HelloWorld

dsarfati:
Added example of using the 'GrainCreator' to do non-silo unit tests to Samples\HelloWorld. Migrated HelloWorld.Tests to xUnit and updated to 1.2.2

There were multiple questions on gitter related to how to write these types of tests. @centur

PR #2243, opened 2016-09-30
#1918 from @dsarfati rebased and fixed

sergeybykov:
I rebased #1918, fixed it, so that it builds, and made minor adjustments to the code on top of @dsarfati's commit.

### DIFF
--- a/src/Orleans/Core/GrainStateStorageBridge.cs
+++ b/src/Orleans/Core/GrainStateStorageBridge.cs
@@ -28,20 +28,22 @@ namespace Orleans.Core
             }
            
             this.grainTypeName = grainTypeName;
-           
-            this.baseGrain = statefulGrain as Grain;
             this.store = store;
         }
 
         internal void SetGrain(Grain grain)
         {
+            if(grain == null)
+                throw new ArgumentNullException(nameof(grain));
+
             statefulGrain = grain as IStatefulGrain;
 
-            if (statefulGrain?.GrainState == null)
-            {
-                throw new ArgumentNullException($"{nameof(grain)}", "No grain state object supplied");
-            }
+            if(statefulGrain == null)
+                throw new ArgumentException("Attempt to configure storage bridge for a non-perisstent grain.", nameof(grain));
 
+            if (statefulGrain.GrainState == null)
+                throw new ArgumentException("No grain state object supplied", nameof(grain));
+            
             this.baseGrain = grain;
         }
 
--- a/src/Orleans/Core/GrainStateStorageBridge.cs
+++ b/src/Orleans/Core/GrainStateStorageBridge.cs
@@ -12,11 +12,11 @@ namespace Orleans.Core
     internal class GrainStateStorageBridge : IStorage
     {
         private readonly IStorageProvider store;
-        private readonly IStatefulGrain grain;
-        private readonly Grain baseGrain;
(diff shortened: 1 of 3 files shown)


sergeybykov:
@jdom I think we can merge this, so that @dsarfati doesn't need to do anything. His commit is included here.

jdom:
I think this PR minimizes the side effects of mocking at an odd level. Instead of relying on the concrete `GrainCreator` and concrete `GrainStateStorageBridge` as in https://github.com/dotnet/orleans/pull/1919/files#r70298407 I would strongly recommend that you do not try to use `GrainCreator` for this. Just use your own way of instantiating the grain (which you are partially doing already in that PR, when it comes to generic grains), and setting an explicit mock of IStorage. That way you would isolate your tests better, and also this change would not be needed.
Am I missing a scenario here @dsarfati?

sergeybykov:
I think we should close it then along with #1918, until @dsarfati or somebody else has time to update the code.

dsarfati:
Sorry for not responding back earlier. I was on paternity leave while you guys were making these comments and I never got a chance to respond. I think the concern with the constructor method was that it forces us to have code that is only ever used for testing and left a bad taste. This intention of this PR was to allow us to create the grain using the same constructor that orleans will use. This seems especially important if we are using a DI container. If there is a better approach to this I'm all ears.

dsarfati:
The real issue here is that there is currently no way to call SetStorage since IStatefulGrain is internal. Maybe a better approach would be to make IStatefulGrain a public interface and make GrainGreator internal. That way anyone can create their own methods for instantiating grains for testing (like in my PR) and not be forced to use the GrainCreator to set the IStorage on those grains.

sergeybykov:
@dsarfati, congratulations with the baby! No worries, family is more important. :-)

@ReubenBond @jason-bragg, what do you think here?

jason-bragg:
LGTM but still has some conflicts.  Please update and I'll merge.

sergeybykov:
@jason-bragg What do you think of @jdom's comment above?
>I think this PR minimizes the side effects of mocking at an odd level. Instead of relying on the concrete GrainCreator and concrete GrainStateStorageBridge as in https://github.com/dotnet/orleans/pull/1919/files#r70298407 I would strongly recommend that you do not try to use GrainCreator for this. Just use your own way of instantiating the grain (which you are partially doing already in that PR, when it comes to generic grains), and setting an explicit mock of IStorage. That way you would isolate your tests better, and also this change would not be needed.

dsarfati:
I was able to accomplish the same effect using reflection to set the private variables for Runtime, identity, storage...  it makes me feel dirty but I now have traditional unit tests running without any silos. I was thinking about making this an OrleansContrib project so others could use it. If we decided to include this PR then I will update the tests to use the grain creator instead of reflection.

jason-bragg:
While I agree that this PR does not put us at an ideal end-state, it is an incremental change in the right direction.  As the grain has no direct dependency on the storage provider, only the storage class, this change improves separation of concerns, flexibility, extensibility and testability.  Given a choice between an incremental change in the right direction, or holding off for a more complete refactor, I'd vote for the incremental change.

sergeybykov:
I rebased it and kicked of tests.

sergeybykov:
Functional tests passed.

PR #2245, opened 2016-09-30
port non-silo tests: OrleansRuntime tests, GatewaySelectionTest, CounterTests and CodeGeneratorTests over to vNext

xiazen:


test/TesterInternal/CodeGeneratorTests.cs
```diff
@@ -21,7 +21,10 @@ public void ServiceInterface_IsGrainClass()
             Assert.False(TypeUtils.IsGrainClass(t), t.FullName + " is not grain class");
             t = typeof(Orleans.Runtime.GrainDirectory.RemoteGrainDirectory);
             Assert.False(TypeUtils.IsGrainClass(t), t.FullName + " should not be a grain class");
+#if !NETSTANDARD_TODO
+           //blocked by CachedTypeResolver not coreclr compatible yet
```
jdom: if this compiles even if it doesn't pass the test, then it's better to have a failing test that we know we need fixing, otherwise it's hard to understand what is working and what is not

jdom: Created a PR on your branch, so that we can merge this

xiazen:
merged your change

PR #2246, opened 2016-10-01
Ability to specify interleaving per message type

yevhen:
Hi guys!

As you know Orleankka provides a functional message-based style API for Orleans and all grains have a uniform interface like: 

```cs
interface IActor
{
    Task<object> Receive(object msg);
}
```

To support some advanced scenarios like ability to specify message interleaving in a more granular way (per-message type) which is incredibly valuable feature we extended this interface with additional method:

```cs
interface IActor
{
    Task<object> Receive(object msg);
    [AlwaysInterleave] Task<object> ReceiveReentrant(object msg);
}
```
So when sending a message via actor reference we were selecting one method ot the other depending on the configuration for this message/actor type.

I'm preparing next big and updated release (1.0, ya finally!) and most anticipated feature is to be able to send messages from clients without need to reference server-side libraries (containing actor definitions). Basically, proper client-server separation.

Unfortunately, current implementation of per-message reentrancy requires to have that information on client (so that we can choose either `Receive` or `ReceiveReentrant`). Which is a leak of server-side knowledge and is very illogical.

I haven't found any other way then to extend implementation upstream. It's just a rough sketch not a final code, something to discuss further. I will highly appreciate you help on this issue. Please advice, mesage reentrancy is very important (and used) feature in Orleankka and now this stops me from releasing 1.0 😞

### DIFF
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -382,14 +382,14 @@ namespace Orleans.Runtime
 
 #region Grains
 
-        internal bool IsReentrantGrain(ActivationId running)
+        internal bool IsMessageInterleaves(Message message, ActivationId running)
         {
             ActivationData target;
             GrainTypeData data;
             return TryGetActivationData(running, out target) &&
                 target.GrainInstance != null &&
                 GrainTypeManager.TryGetData(TypeUtils.GetFullName(target.GrainInstanceType), out data) &&
-                data.IsReentrant;
+                data.IsMessageInterleaves(message);
         }
 
         public void GetGrainTypeInfo(int typeCode, out string grainClass, out PlacementStrategy placement, out MultiClusterRegistrationStrategy activationStrategy, string genericArguments = null)
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -382,14 +382,14 @@ namespace Orleans.Runtime
 
 #region Grains
 
-        internal bool IsMessageInterleaves(Message message, ActivationId running)
+        internal bool CanInterleave(ActivationId running, Message message)
         {
             ActivationData target;
             GrainTypeData data;
             return TryGetActivationData(running, out target) &&
                 target.GrainInstance != null &&
                 GrainTypeManager.TryGetData(TypeUtils.GetFullName(target.GrainInstanceType), out data) &&
-                data.IsMessageInterleaves(message);
+                (data.IsReentrant || data.IsMessageAllowedToInterleave(message));
         }
 
         public void GetGrainTypeInfo(int typeCode, out string grainClass, out PlacementStrategy placement, out MultiClusterRegistrationStrategy activationStrategy, string genericArguments = null)
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -6,6 +6,8 @@ using System.Reflection;
 using System.Runtime.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
+
+using Orleans.CodeGeneration;
 using Orleans.GrainDirectory;
 using Orleans.Providers;
 using Orleans.Runtime.Configuration;
@@ -389,7 +391,7 @@ namespace Orleans.Runtime
             return TryGetActivationData(running, out target) &&
                 target.GrainInstance != null &&
                 GrainTypeManager.TryGetData(TypeUtils.GetFullName(target.GrainInstanceType), out data) &&
-                (data.IsReentrant || data.IsMessageAllowedToInterleave(message));
+                (data.IsReentrant || data.MayInterleave((InvokeMethodRequest)message.BodyObject));
         }
 
         public void GetGrainTypeInfo(int typeCode, out string grainClass, out PlacementStrategy placement, out MultiClusterRegistrationStrategy activationStrategy, string genericArguments = null)
--- a/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
+++ b/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
@@ -24,7 +24,7 @@ namespace Orleans.Runtime
         internal Type StateObjectType { get; private set; }
         internal bool IsReentrant { get; private set; }
         internal bool IsStatelessWorker { get; private set; }
-        Func<InvokeMethodRequest, bool> IsMessageAllowedToInterleavePredicate { get; set; }
+        internal Func<InvokeMethodRequest, bool> MayInterleave { get; private set; }
    
         public GrainTypeData(Type type, Type stateObjectType)
         {
@@ -36,7 +36,7 @@ namespace Orleans.Runtime
             GrainClass = TypeUtils.GetFullName(typeInfo);
             RemoteInterfaceTypes = GetRemoteInterfaces(type); ;
             StateObjectType = stateObjectType;
-            IsMessageAllowedToInterleavePredicate = GetMessageAllowedToInterleavePredicate(type);
+            MayInterleave = GetMayInterleavePredicate(type) ?? (_ => false);
         }
 
         /// <summary>
@@ -137,10 +137,10 @@ namespace Orleans.Runtime
         /// </summary>
         /// <param name="grainType">Grain class.</param>
         /// <returns></returns>
-        private static Func<InvokeMethodRequest, bool> GetMessageAllowedToInterleavePredicate(Type grainType)
+        private static Func<InvokeMethodRequest, bool> GetMayInterleavePredicate(Type grainType)
         {
             if (!grainType.GetCustomAttributes<MayInterleaveAttribute>().Any())
-                return req => false;
+                return null;
 
             if (grainType.GetCustomAttributes(typeof(ReentrantAttribute), true).Any())
                 throw new InvalidOperationException(
@@ -166,11 +166,6 @@ namespace Orleans.Runtime
             var predicate = Expression.Lambda<Func<InvokeMethodRequest, bool>>(call, parameter).Compile();
 
             return predicate;
-        }
-
-        internal bool IsMessageAllowedToInterleave(Message message)
-        {
-            return IsMessageAllowedToInterleavePredicate((InvokeMethodRequest) message.BodyObject);
         }
     }
 }
--- a/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
+++ b/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
@@ -2,6 +2,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+
+using Orleans.CodeGeneration;
 using Orleans.Concurrency;
 using Orleans.GrainDirectory;
 using Orleans.MultiCluster;
@@ -21,7 +23,7 @@ namespace Orleans.Runtime
         internal Type StateObjectType { get; private set; }
         internal bool IsReentrant { get; private set; }
         internal bool IsStatelessWorker { get; private set; }
-        Func<object, bool> IsMessageInterleavesPredicate { get; set; }
+        Func<InvokeMethodRequest, bool> IsMessageAllowedToInterleavePredicate { get; set; }
    
         public GrainTypeData(Type type, Type stateObjectType)
         {
@@ -33,7 +35,7 @@ namespace Orleans.Runtime
             GrainClass = TypeUtils.GetFullName(typeInfo);
             RemoteInterfaceTypes = GetRemoteInterfaces(type); ;
             StateObjectType = stateObjectType;
-            IsMessageInterleavesPredicate = _ => false;
+            IsMessageAllowedToInterleavePredicate = _ => false;
         }
 
         /// <summary>
@@ -130,17 +132,18 @@ namespace Orleans.Runtime
         }
 
         // used by Orleankka's bootstrapper
-        internal void SetMessageAlwaysInterleaves(Func<object, bool> predicate)
+        internal void SetMessageAllowedToInterleavePredicate(Func<InvokeMethodRequest, bool> predicate)
         {
             if (predicate == null)
                 throw new ArgumentNullException(nameof(predicate));
 
-            IsMessageInterleavesPredicate = predicate;
+            IsMessageAllowedToInterleavePredicate = predicate;
         }
 
         public bool IsMessageAllowedToInterleave(Message message)
         {
-            return IsMessageInterleavesPredicate(message.BodyObject.GetType());
+            var request = message.BodyObject as InvokeMethodRequest;
+            return request != null && IsMessageAllowedToInterleavePredicate(request);
         }
     }
 }
--- a/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
+++ b/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
@@ -138,9 +138,9 @@ namespace Orleans.Runtime
             IsMessageInterleavesPredicate = predicate;
         }
 
-        public bool IsMessageInterleaves(Message message)
+        public bool IsMessageAllowedToInterleave(Message message)
         {
-            return IsReentrant || IsMessageInterleavesPredicate(message.BodyObject.GetType());
+            return IsMessageInterleavesPredicate(message.BodyObject.GetType());
         }
     }
 }
--- a/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
+++ b/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
@@ -21,8 +21,8 @@ namespace Orleans.Runtime
         internal Type StateObjectType { get; private set; }
         internal bool IsReentrant { get; private set; }
         internal bool IsStatelessWorker { get; private set; }
+        Func<object, bool> IsMessageInterleavesPredicate { get; set; }
    
-     
         public GrainTypeData(Type type, Type stateObjectType)
         {
             var typeInfo = type.GetTypeInfo();
@@ -33,6 +33,7 @@ namespace Orleans.Runtime
             GrainClass = TypeUtils.GetFullName(typeInfo);
             RemoteInterfaceTypes = GetRemoteInterfaces(type); ;
             StateObjectType = stateObjectType;
+            IsMessageInterleavesPredicate = _ => false;
         }
 
         /// <summary>
@@ -126,6 +127,20 @@ namespace Orleans.Runtime
                             typeof(MultiClusterRegistrationStrategy).Name,
                             grainClass.Name));
             }
+        }
+
+        // used by Orleankka's bootstrapper
+        internal void SetMessageAlwaysInterleaves(Func<object, bool> predicate)
+        {
+            if (predicate == null)
+                throw new ArgumentNullException(nameof(predicate));
+
+            IsMessageInterleavesPredicate = predicate;
+        }
+
+        public bool IsMessageInterleaves(Message message)
+        {
+            return IsReentrant || IsMessageInterleavesPredicate(message.BodyObject.GetType());
         }
     }
 }
--- a/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
+++ b/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
@@ -142,8 +142,10 @@ namespace Orleans.Runtime
 
         public bool IsMessageAllowedToInterleave(Message message)
         {
-            var request = message.BodyObject as InvokeMethodRequest;
-            return request != null && IsMessageAllowedToInterleavePredicate(request);
+            if (message.Direction != Message.Directions.Request)
+                return IsReentrant;
+
+            return IsMessageAllowedToInterleavePredicate((InvokeMethodRequest)message.BodyObject);
         }
     }
 }
--- a/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
+++ b/src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
 
 using Orleans.CodeGeneration;
@@ -35,7 +36,7 @@ namespace Orleans.Runtime
             GrainClass = TypeUtils.GetFullName(typeInfo);
             RemoteInterfaceTypes = GetRemoteInterfaces(type); ;
             StateObjectType = stateObjectType;
-            IsMessageAllowedToInterleavePredicate = _ => false;
+            IsMessageAllowedToInterleavePredicate = GetMessageAllowedToInterleavePredicate(type);
         }
 
         /// <summary>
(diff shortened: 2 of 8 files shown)


sergeybykov:
What do you think would be the ideal way to specify per-message reentrancy?

We already have `InvokeMethodOptions` enum with `AlwaysInterleave` in it, and we could easily add another value to it if necessary. `GrainReference.InvokeMethodAsync<T>` already takes a value for `InvokeMethodOptions options`. We just don't expose it in codegen. It should be relatively easy to introduce a per method / per interface attribute that would instruct codegen to add an additional optional parameter, e.g. a bool, that would get translated into a value of `InvokeMethodOptions` passed down to `GrainReference.InvokeMethodAsync<T>`.

Alternatively, we could use `RequestContext` for specifying the reentrancy flag, but IMO that would be ugly.

Are there other options?

yevhen:
Any implementation which doesn't require clients (senders) to know about which message should be sent as reentrant. AFAIU both of suggestions (per method / per interface attribute and `RequestContext') doesn't fit that requirement.

Also, those won't work with F#. Discriminated unions require special handling due to how cases are implemented internally (as static properties!). That's why I've added a function callback,. so I can install an appropriate implementation for C#/F#.

yevhen:
I should probably add some details about upcoming 1.0 client-server implementation in Orleankka. 

Previously, we were having fixed (static) set of grain endpoints and interfaces covering all possible grain configurations (one for worker, one for actor with random placement, with activation count placement, etc) and acting as generic hosts. The actual type of the actor (in the form of its type code) that need to be instantiated had been sent with every request.

Now, in 1.0 we got rid of this in favor of pre-generating (using Roslyn) grain interfaces dynamically. Since, all interfaces have exactly the same shape - this was only about generating interfaces with exactly the same type name. On server-side, since we are having all required information, we amend the generated interface with all neccessary attributes, such as `StatelessWorker`, `Placement`, etc. For client-side - it's enough to just generate an interface with exactly the same name, for Orleans to correctly bind it to server-side implementation. So there is no need to share exactly the same dll with interfaces. It's just enough for client to know server-side actor's type to be able to send him messages.

I don't (and shouldn't) have any server-side information on client. I gues this rules out `RequestContext` and other suggestions. Plus, the added weirdness of handling F# properly doesn't leave any other choice then to add function callback.

sergeybykov:
>Any implementation which doesn't require clients (senders) to know about which message should be sent as reentrant.

What criteria would be used to decide whether to interleave a particular request? Argument type? An expression calculated over argument(s)?

You mentioned in the code that you envision this per type predicate to be installed by a boorstrap provider. Do you see a problem with defining such a predicate as part of the grain class, so that it could be auto-discovered and installed without the need for a bootstrap provider?

sergeybykov:
Or am I completely missing the point?

yevhen:
> What criteria would be used to decide whether to interleave a particular request? Argument type? An expression calculated over argument(s)?

It's argument type for both C# and F#.  Thus, there is a some weirdness wrt to how properly calculate the argument type when argument is F#'s Discriminated Union.

As F# is functional the callback style looks more idiomatic and will cover both C# and F#. In F# this was previously implemented as conventional static function on actor's type. For C# - this was supported via `[ReentrantAttribute(Type msgType)]` which you can put on actor class.

yevhen:
> Do you see a problem with defining such a predicate as part of the grain class, so that it could be auto-discovered and installed without the need for a bootstrap provider?

No, ok for me. I just don't know how to add this to Orleans so it doesn't look innocent to its core design.

yevhen:
I can imagine something similar to how stream filters are defined now, like putting on the method:

```cs
[AlwaysInterleave("SomeClass.SomeStaticCallbackMethod")]
Task SomeMethod(object msg)
```

Where `SomeClass.SomeStaticCallbackMethod` points to method having following signature:

```cs
bool Predicate(Type grainType, object msg);
```

sergeybykov:
>bool Predicate(Type grainType, object msg);

Shouldn't it rather be
bool Predicate(TypeInfo grainType, InvokeMethodRequest request);?

What about F#? Will such approach work for it?

yevhen:
Yes, it will

ReubenBond:
If we could work this into grain-level or silo-level interceptors, would that work for you? I'm not sure that it can be done cleanly

yevhen:
Interceptor would be fine. Anything that calls my predicate with grain type and mwthod argument info will do.

sergeybykov:
Silo wide interceptors require a bootstrap provider or some other way to install them globally. I think we better avoid such a burden here if we can. Hence, grain level interceptors seem like a more convenient pattern to me. However, if you envisions no grain class specific logic in the predicate, then defining a per class predicate may become a burden. 

We could add a method to `IGrainInvokeInterceptor` or introduce a separate interface. `IGrainReentrancyOverride`?

```csharp
public interface IGrainReentrancyOverride
{
    bool CanInterleaveRequest(TypeInfo grainType, InvokeMethodRequest request);
}
```

The scheduler will likely have to treat grain classes with a reentrancy overrides as reentrant (they are handled differently today) but interleave requests only when `CanInterleaveRequest()` returns true. Not sure off the top of my head how difficult such a change to the scheduler will be. @gabikliot might have a better idea. This may be another argument for per grain class predicates, so that the scheduler doesn't have to treat all grain classes as reentrant if a per silo predicate is installed.

yevhen:
Sounds complicated to me but you know better how to make it in a most lean and logical way. I'm ok with grain level interceptors if they can amend how the message is delievered (interleaved or not)

sergeybykov:
I don't know if I know better. :-) Just throwing some thoughts for other people to weigh in.

sergeybykov:
Any other opinions here? @ReubenBond, what do you think? @jdom?

We need to unblock @yevhen.

ReubenBond:
Whatever design we adopt, it cannot break the single-threaded invariant.

If we are going to add a new method/interface, are there any other provisions we'd like in that interface? Just to make sure that we're not making adding a specific feature when we could solve this with a more general approach which might result in a less complicated API in the long run. As an example, we're talking about classifying a message as being `AlwaysInterleave` or not, but messages can also be `ReadOnly` & `Unordered`.

I think my comment about using grain-level interceptors is silly, since it executes too late.

sergeybykov:
Here's a proposal.

1. Add a 
```csharp
bool CanInterleaveRequest(InvokeMethodRequest request);
```
method to `GrainTypeData` class. `GrainTypeData` in its constructor will find if exist all methods of the grain class marked with the `[Interleave("SomeClass.SomeStaticCallbackMethod")]` attributes and build a method ID->callback map. Can put Reflection-related methods to `TypeUtils`.

2. Add a reference to `GrainTypeData` to `ActivationData`, and expose `GrainTypeData.CanInterleaveRequest(InvokeMethodRequest request)` via it.

3. Add to `Dispatcher.CanInterleave` another || with a call to `targetActivation.CanInterleaveRequest`.

Will this be sufficient? Did I miss something?

gabikliot:
Sorry, I am not fully following why this is needed. Can u please briefly summarize what will this achieve?

sergeybykov:
@gabikliot You mean the whole purpose of the change?

In Orleankka, @yevhen implements the Erlang/Akka style model with a single `Receive(object message)` remote method where different message types conceptually map to methods of a grain interface in the traditional approach but the "dispatching" happens within `Receive`.

Now for different message types he needs to have different reentrancy rules, as if for different methods in a grain interface. He wants to do that dynamically via a callback. We could consider expanding the scope of this change, so that `[Interleave]` would also be applicable to traditional grain methods.

@yevhen, is this an accurate description?
@gabikliot, does it make sense now?

yevhen:
> @yevhen, is this an accurate description?

absolutely.

gabikliot:
Sergey, yes I meant the whole purpose.
i did not understand the 2nd paragraph of your description. What callback? Why do we need to Change dispatcher? Always interleave is already applicable to regular grain calls, isn't it? Or is it internal attribute?

sergeybykov:
@yevhen has a single method that we wants to interleave for some requests and not to interleave for others. The decision would be made based on the argument types and potentially other data in the request. Hence, a static `[Interleave]` attribute won't be enough, as the interleaving decision will have to be made on a per-request basis. That's why he need a callback - to make the interleaving decisions for each request. The change to Dispatcher is needed to invoke this callback (if it's defined) in addition to the reentrancy checks it already performs in `CanInterleave`.

gabikliot:
Ohh, I think I understand now.

So alternatively, he could have implemented the queuing himself, in his dispatch logic in the silo interceptor. In this approach, from Orleans perspective all calls can interleave (there will be a single call 'DO' with `AlwaysInterleave`) and @yevhen will decide himself what should be queued and queue it himself in his logic, and what can be invoked directly., and dispatch the queued one when the previous done, etc....
But, that means a lot of logic duplication with the already existing Orleans logic in the Dispatcher, so instead of implementing it all, @yevhen  wants the Dispatcher to be able to call into his callback. Dispatcher will be the execution/queueing/dispatching mechanism, while his code can provide dynamic policy.

Is that a correct description?

yevhen:
@gabikliot ye, close to. Previously, I have these two methods:

```cs
interface IActor : IGrainWithStringKey 
{
    Task<object> Receive(object msg);
    [AlwaysInterleave] Task<object> ReceiveReentrant(object msg);
}
```
so grains we not `Reentrant` in general and when message was being sent through my grain reference wrapper, I was inspecting the type of the message and depending on configuration one method or the other was chosen.

Unfortunately, when sending message from client to server with that approach it means that server configuration (which type of message need to be sent via interleaved method) need to be available on the client, which doesn't seem right.

Also, what's more important - it doesn't work with streams 😞

yevhen:
I kinda solve it by marking all actors as Reentrant and then using AsyncSerialExecutor to execute non-interleaved calls serially but it adds additional overhead to every actor and call. So it would be really cool if  such support is available out-of-the-box from underlying framework. Will make it 100% friendly for message-based style of communication.

sergeybykov:
@yevhen
>Also, what's more important - it doesn't work with streams 😞 

What's the problem with streams?

@gabikliot 
>But, that means a lot of logic duplication with the already existing Orleans logic in the Dispatcher, so instead of implementing it all, @yevhen wants the Dispatcher to be able to call into his callback. Dispatcher will be the execution/queueing/dispatching mechanism, while his code can provide dynamic policy.

Yes, exactly - the callback would supply a kind of a dynamic reentrancy policy.

Do you see any holes in my proposal above? I assume no changes in the scheduler will be required, as we'll leverage the Dispatcher's reentrancy logic, but I'm not 100% sure I'm not missing something.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
```diff
@@ -127,5 +128,19 @@ internal static MultiClusterRegistrationStrategy GetMultiClusterRegistrationStra
...
+        {
+            if (predicate == null)
+                throw new ArgumentNullException(nameof(predicate));
+
+            IsMessageInterleavesPredicate = predicate;
+        }
+
+        public bool IsMessageInterleaves(Message message)
+        {
+            return IsReentrant || IsMessageInterleavesPredicate(message.BodyObject.GetType());
```
jdom: I take it that in Orleankka you do not use the `InvokeMethodRequest` as the BodyObject, otherwise this wouldn't give you what you want

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
```diff
@@ -127,5 +128,19 @@ internal static MultiClusterRegistrationStrategy GetMultiClusterRegistrationStra
                             grainClass.Name));
             }
         }
+
+        // used by Orleankka's bootstrapper
+        internal void SetMessageAlwaysInterleaves(Func<object, bool> predicate)
```
jdom: it would be good to combine this with @sergeybykov's suggestion of being able to define it statically, with an attribute or something on the grain implementation or interface, or even on the argument type as defined in the interfaces that the silo uses. I would prefer on the grain type itself, since the argument type could be re-used across grain implementations, and each have different interleaving implications, or even different grain implementations for the same grain interface might have different needs.

yevhen:
> What's the problem with streams?

In current design grains are not marked as Reentrant so stream subscriptions inside actor are not reentrant as well. If I mark it as Reentrant I'll loose ability to specify reentrancy in a granular way.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
yevhen: > grain implementation or interface

implementation. I think interface is not a good candidate since it's reentrancy is an implementation detail.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
yevhen: that's just spike :) 

Ye, I need a method request argument.

gabikliot:
About streams, I think your approach with marking grain Reentrant and using AsyncSerialExecutor  will work as well. We just need to make sure Interceptor also intercepts grain extension calls. I think it does, but if not we can change it, and then  Reentrant + AsyncSerialExecutor should work.

Regarding Sergey's proposal:
you are talking about some static map (static based on grain classes) that is exposed via ActivationData to the dispatcher. But in the discussion we just had, we were talking about Interceptor providing this callback (somehow) to the dispatcher based on the dynamic data (the actual msg). The 2nd one is more powerful generally, I think. And knowing @yevhen , I suspect he would want the dynamic case. Would the static map be enough?

The alternative is Reentrant + AsyncSerialExecutor (or generally, his own dispatch/queueing logic in the interceptor), which sounds like much more generic to me (albeit a bit less efficient?): dynamic, totally not constrained by anything in the platform.

jdom:
Just to expand a little, the issue with grain interceptor is that they currently get invoked by the time the grain request is being executed, so it's too late to decide how to schedule. 
AsyncSerialExecutor is of course a workaround that has both perf implications, as well as potential unnecessary issues with deactivation (either explicitly due to failure to process a message and decided to suicide, or by external forces when the activation has queued up messages). Not dismissing it though, but useful to weigh pros and cons.
Having the delegate as @yevhen suggested but combining it with static information to know which delegate to use seems like a clean and decent option IMO

gabikliot:
Ok, I reread the thread now. It seems like the static callback would be enough. Correct?

gabikliot:
Sergey, about your suggestion of adding `bool CanInterleaveRequest(InvokeMethodRequest request)`:

Does the callback only need to refer to the current msg only (is this request Alwaysinterleave), or is it relative to what is currently already executing (statefull w.r.t. the current state of the execution). For example, look at how ReadOnly is handled.In the case of the latter, the callback will need to accept the state of the currently executing msgs, which means we need to track this state (which we currently do not do explicitly).


I guess I am still confused by the:
1) dynamic vs static map
2) current request vs. all execution state
3) it sounded like the real issue that bothered @yevhen is that the client needed to reference the `AlwaysInterleave`. In such a case, we can hide it from the interface and leave on the impl. method.
But on the other hand, it looks like he wants it to be based on the msg, so it IS based on client.

I think I am mostly very confused.

yevhen:
Let me explain this with code.

In Orleans you declare separate methods for each message OO-style:

```cs
interface ISomeGrain
{
   Task Foo(int arg);
   [AlwaysInterleave] Bar(string arg)
}
``` 

So you can have specify interleaving granularly per particular method. While in Orleankka all actors have FP-style uniform interface with single `Receive` method and messages are designed as message classes. So the only way to signal whether the message will need to be dispatched interleaved is by marking its runtime type. For C# api we have special attribute for this. Code below:

```cs
// message classes
class Foo { public int Arg; }
class Bar { public string Arg; }

// actor class
[Reentrant(typeof(Bar)] // messages of this type should be dispatched as interleaved
class SomeActor : 
{
    Task<object> Receive(object msg) 
}
```
Basically callback with an argument type would be enough but F# with its implementation of Discriminated Unions, which is what most prople use, adds another complexity wrt to determining the *actual* type of the argument and so passing just `obj.GetType()` is not enough, that's why I'm asking for callback which passes whole argument.

Does it clear now?

> Does the callback only need to refer to the current msg only (is this request Alwaysinterleave), or is it relative to what is currently already executing (statefull w.r.t. the current state of the execution).

no, it's not stateful in regard to previous executions.  Just need to inspect the type of the message that is about to be dispatched and then depending on configuration provided by the developer (the Reentrant attributes mentioned above for C# vesion or conventional predicate for F# version) decided whether it should be dispatched interleaved or not).

I believe streams messages are dispatched via the same Dispatched mechanism and so this method should work (cover) them as well.

gabikliot:
Got it  now!

And the callback will be provided via `[Interleave("SomeClass.SomeStaticCallbackMethod")]` and invoked via reflection directly by the dispatcher, according to Sergey's proposal. So unrelated to interceptors or bootstrap providers.

OK, I think Sergey's proposal should work theoretically. 

The main caveat is **this will be the first time we allow an arbitrary application code be directly invoked in the runtime**. Untill now it was always either in the activation scope (interceptors), or in the providers scope. Stream adapters, all this stuff runs in the providers.
So it opens all the standard questions of: security, what if it throws, what if it starts Tasks, .... a Pandora box.

sergeybykov:
>The main caveat is this will be the first time we allow an arbitrary application code be directly invoked in the runtime. Untill now it was always either in the activation scope (interceptors), or in the providers scope. Stream adapters, all this stuff runs in the providers.
So it opens all the standard questions of: security, what if it throws, what if it starts Tasks, .... a Pandora box.

Sort of. To get to this advanced feature and to abuse it will require some determination and deeper than average understanding of how the runtime works. Will be pretty much invisible to regular users. Personally, I'm not too concerned.

gabikliot:
Ok. Just add a test with throwing and make sure we handle it correctly.
Strictly speaking this is layering violation.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -382,14 +382,14 @@ internal int UnregisterGrainForTesting(GrainId grain)
...
-        internal bool IsReentrantGrain(ActivationId running)
+        internal bool IsMessageInterleaves(Message message, ActivationId running)
         {
             ActivationData target;
             GrainTypeData data;
             return TryGetActivationData(running, out target) &&
                 target.GrainInstance != null &&
                 GrainTypeManager.TryGetData(TypeUtils.GetFullName(target.GrainInstanceType), out data) &&
-                data.IsReentrant;
+                data.IsMessageInterleaves(message);
```
gabikliot: I would leave data.IsReentrant and add OR data.IsMessageAllowedToInterleave.
That way you separate the 2 things, which may be useful in other places.
Plus naming: `IsMessageAllowedToInterleave`.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -382,14 +382,14 @@ internal int UnregisterGrainForTesting(GrainId grain)
 
 #region Grains
 
-        internal bool IsReentrantGrain(ActivationId running)
+        internal bool IsMessageInterleaves(Message message, ActivationId running)
```
gabikliot: lets name that: CanInterleave and order the args like in Dispatcher.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
gabikliot: Please add 2 tests:
one with successful interleaving and one with callback throwing exception . I want to see that the exception is handled correctly. That will only make us define what "correctly" is in this case.

yevhen:
Incorporated all of the feeback and also added tests for stream delivery. Made me realize I haven't done anything in Orleankka to support (make life easier when working with) `Immutable<T>`. Will create another PR.

test/Tester/HostedTestClusterBase.cs
```diff
@@ -14,7 +14,7 @@ public abstract class HostedTestClusterEnsureDefaultStarted : OrleansTestingBase
     {
         protected TestCluster HostedCluster { get; private set; }
 
-        public HostedTestClusterEnsureDefaultStarted(DefaultClusterFixture fixture)
+        public HostedTestClusterEnsureDefaultStarted(BaseTestClusterFixture fixture)
```
jdom: this base class can only be used with the DefaultClusterFixture, as it uses an collection fixture. If you need a non-default cluster, start a new one using IClassFixture. Look at `GrainCancellationTokenTests` for an example, where it doesn't use this base class at all.

test/TestGrains/ReentrantGrain.cs
```diff
@@ -68,6 +70,59 @@ public Task SetSelf(INonReentrantGrain self)
         }
     }
 
+    public class NonReentrantGrainWithMessageInterleavePredicate : Grain, INonReentrantGrainWithMessageInterleavePredicate
```
jdom: Please prefix grain and interface name with Grain

test/TesterInternal/ReentrancyTests.cs
```diff
@@ -342,6 +482,55 @@ public void FanOut_AC_Stress_NonReentrant()
...
+            Name = name;
+            InstallPredicate();
+            return TaskDone.Done;
+        }
+
+        void InstallPredicate()
+        {
+            GrainTypeData grainType;
+            GrainTypeManager.Instance.TryGetData(typeof(NonReentrantGrainWithMessageInterleavePredicate).FullName, out grainType);
+            grainType.SetMessageAllowedToInterleavePredicate(req =>
```
jdom: I thought the plan was to provide an attribute or some other way to statically define the predicate.

Something like `[MessageAllowedToInterleavePredicate("UnitTests.Grains.NonReentrantGrainWithMessageInterleaveGrain.NonReentrantGrainWithMessageInterleavePredicate")]` on the grain implementation itself

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
```diff
@@ -127,5 +130,22 @@ internal static MultiClusterRegistrationStrategy GetMultiClusterRegistrationStra
...
+            if (predicate == null)
+                throw new ArgumentNullException(nameof(predicate));
+
+            IsMessageAllowedToInterleavePredicate = predicate;
+        }
+
+        public bool IsMessageAllowedToInterleave(Message message)
+        {
+            if (message.Direction != Message.Directions.Request)
+                return IsReentrant;
```
yevhen: Right, I was just not sure about other possible message directions. So it's more of a fallback to previous behavior.

test/Tester/HostedTestClusterBase.cs
yevhen: Ok. Sorry, there a lot of implicits in how unit tests are run. Never used xunit before

test/TestGrains/ReentrantGrain.cs
yevhen: Ok.

test/TesterInternal/ReentrancyTests.cs
yevhen: I'm not sure. From the discussion above it was not clear what was the final decision. Should be a class-level attribute for this to work with streams. If that is a general consensus - I'll take this route

test/Tester/HostedTestClusterBase.cs
yevhen: Also, I'm not sure why? It only accesses HostCluster which is defined on BaseTestClassFixture

yevhen: Ok. This xunit stuff is weird ((

yevhen:
So are we about to introduce specia class-levell attribute?

veikkoeeva:
>>    The main caveat is this will be the first time we allow an arbitrary application code be directly invoked in the runtime. Untill now it was always either in the activation scope (interceptors), or in the providers scope. Stream adapters, all this stuff runs in the providers.
    So it opens all the standard questions of: security, what if it throws, what if it starts Tasks, .... a Pandora box.
>
>Sort of. To get to this advanced feature and to abuse it will require some determination and deeper than average understanding of how the runtime works. Will be pretty much invisible to regular users. Personally, I'm not too concerned.

Should be documented with a few lines?

sergeybykov:
>So are we about to introduce special class-level attribute?

My understanding was that the attribute at the minimum would need to be method level. I guess expanding it to also allow it at the class level will be a shortcut for applying the same attribute to each method. In addition that would allow to control interleaving of streaming calls?

yevhen:
Then, we can just create ctor overload for AlwaysInterleave and allow it on a class level. Same logic.

What do you think?

sergeybykov:
>Then, we can just create ctor overload for AlwaysInterleave and allow it on a class level. Same logic.
What do you think?

I'm afraid it will be confusing: "always" wouldn't actually mean always with the predicate. I think it'll be cleaner if we leave `AlwaysInterleave` as is and add a separate attribute applicable to both class and method level. Maybe `ConditionallyInterleave`? Or `MayInterleave`?

yevhen:
Do we really need it to be applicable on method level? I'm trying to think about other use-cases outside Orleankka's uniform interface model and don't see any. I think if we allow it on method level that will be an overengineering.

yevhen:
With Orleans OOP interfaces messages are modeled as methods and so AlwaysIterleave should suffice.

sergeybykov:
>I think if we allow it on method level that will be an overengineering. 

You are probably right. Let's not overengineer. We can always expand to method level later if needed.

yevhen:
Let's finish with naming (the hardest part).

1. MayInterleave
2. CanInterleave
3. InterleavePredicate
4. DynamicInterleave (as opposite to static way of attributing methods with AlwaysInterleave)
5. Other?

sergeybykov:
My vote is for `MayInterleave`. Other options I can think of are `InterleaveIf` or `InterleaveWhen`, but I'm not sure I like them more.

yevhen:
argh, looks like I did wrong rebase. Will redo tomorrow. Ignore, pls

jdom:
Be careful that if you want to re-open the PR eventually, you have to re-open with exactly the same original commits as before closing, and only after that rebase the branch. Otherwise GitHub just thinks they are entirely different branches and want allow you to re-open the same PR.

jdom:
btw, my preference would also be `MayInterleave`

yevhen:
I think it's easier to create new PR.

jdom:
As you prefer. Nevertheless if it was my choice, I'd prefer to re-open this one, as all the discussions happened here. Basically just do the rebase (or re-do) on a separate branch, reopen this, and then force push to this branch

yevhen:
I don't understand why I still get others commits after rebase :(

yevhen:
It would be better to squash changes but for now could be seen here https://github.com/dotnet/orleans/pull/2246/files

I think `CanInterleave` method in GrainTypeData need to be called MayInterleave and probably the field as well to be consistent with MayInterleave attribute name.

sergeybykov:
Looks good to me as is. I kicked off a load test just to confirm there is no impact on perf. But I can't imagine that. Will merge after that unless I hear objections.
>I think CanInterleave method in GrainTypeData need to be called MayInterleave and probably the field as well to be consistent with MayInterleave attribute name.

Did you rather mean `Catalog.CanInterleave`? I'm fine either way with a slight preference for `CanInterleave` because Catalog gives a definitive answer to Dispatcher whether a specific message can interleave or not.

test/TestGrains/ReentrantGrain.cs
```diff
@@ -68,6 +71,84 @@ public Task SetSelf(INonReentrantGrain self)
         }
     }
 
+    [MayInterleave("MayInterleave")]
```
jdom: Does `[MayInterleave(nameof(MayInterleave)]` compile at this scope (or call the method MayInterleavePredicate to make it more explicit that it has hothing to do with the attribute name)? If so, that would be preferred, as people might look at this test to get a sense of how the feature works

sergeybykov:
Perf test looks good.
Thank you, @yevhen!

yevhen:
Cool, thanks! When to expect next minor release? ))

sergeybykov:
1.3.1. is in the works. See #2350.

PR #2249, opened 2016-10-02
Simplify code generation attributes to always use Type

ReubenBond:
Currently code generation attributes such as `[MethodInvoker]` and `[GrainReference]` can be instantiated using either the name of the `IGrain` type which they were generated for, or the `Type` itself.

This removes the string property, which is unnecessary.
Before:
```c#
[MethodInvoker("global::Orleans.IMembershipTableGrain", -1256503757, typeof (global::Orleans.IMembershipTableGrain))]
```
After:
```C#
[MethodInvoker(typeof(IMembershipTableGrain), -1256503757)]
```

Renames `GrainType` to `TargetType`, since the property is used for both generated serializers and grain interfaces.

Removes the unnecessary code analysis suppression, since [CA1813:AvoidUnsealedAttributes](https://msdn.microsoft.com/en-us/library/ms182267.aspx) does not apply to abstract classes.

Edit: this simplifies the logic used when loading support classes, so that it is more obviously correct. Previously the code (inconsistently) checked for both the `GrainType` & `ForGrainType` properties.

### DIFF
--- a/src/Orleans/CodeGeneration/ActivationAttribute.cs
+++ b/src/Orleans/CodeGeneration/ActivationAttribute.cs
@@ -2,36 +2,26 @@ using System;
 
 namespace Orleans.CodeGeneration
 {
-    using Orleans.Runtime;
-
     /// <summary>
     /// For internal (run-time) use only.
     /// Base class of all the activation attributes 
     /// </summary>
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes"), AttributeUsage(System.AttributeTargets.All)]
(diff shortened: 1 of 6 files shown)


src/Orleans/CodeGeneration/ActivationAttribute.cs
```diff
@@ -2,36 +2,26 @@
...
         {
-            ForGrainType = forGrainType;
+            this.TargetType = targetType;
         }
 
         /// <summary>
+        /// Gets the type which this implementation applies to.
         /// </summary>
-        protected GeneratedAttribute() { }
+        public Type TargetType { get; }
```
xiazen: why change name of GrainType to TargetType?

xiazen:
LGTM

PR #2251, opened 2016-10-03
Remove zero length check in Protobuf serializer

pditterline:
The Protobuf serializer currently returns null for a zero-length message. This causes a message with all default values to deserialize as null rather than a default message since Protobuf doesn't serialize
default values. As far as I can tell, there's no case where you would want null here... the tests all pass without the length check.

Removed the length check and added a unit test to demonstrate the issue (my new test fails without the change). Also fixed spelling on the other unit tests. :)


gabikliot:
Good catch, thank you!

PR #2256, opened 2016-10-04
Avoid race condition in scheduler test

jdom:
The non-determinism was in the test itself, not the scheduler

### DIFF
--- a/test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
+++ b/test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
@@ -604,19 +604,32 @@ namespace UnitTests.SchedulerTests
             {
                 TimeSpan waitCheckTime = TimeSpan.FromMilliseconds(150 * ChainLength * NumChains * waitFactor);
 
-                var timeoutLimit = TimeSpan.FromSeconds(1);
                 try
                 {
                     await resultHandles[i].Task.WithTimeout(waitCheckTime);
                 }
                 catch (TimeoutException)
                 {
-                    Assert.True(false, "Result did not arrive before timeout " + timeoutLimit);
+                    Assert.True(false, "Result did not arrive before timeout " + waitCheckTime);
                 }
 
                 bool ok = resultHandles[i].Task.Result;
-                Assert.False(taskChainEnds[i].IsFaulted, "Task chain " + i + " should not be Faulted: " + taskChainEnds[i].Exception);
+                
+                if (!taskChainEnds[i].IsCompleted)
+                {
+                    try
+                    {
+                        // since resultHandle being complete doesn't directly imply that the final chain was completed (there's a chance for a race condition), give a small chance for it to complete.
+                        await taskChainEnds[i].WithTimeout(TimeSpan.FromMilliseconds(10));
+                    }
+                    catch (TimeoutException)
+                    {
+                        Assert.True(false, $"Task chain end {i} should complete very shortly after after its resultHandle");
+                    }
+                }
+
                 Assert.True(taskChainEnds[i].IsCompleted, "Task chain " + i + " should be completed");
+                Assert.False(taskChainEnds[i].IsFaulted, "Task chain " + i + " should not be Faulted: " + taskChainEnds[i].Exception);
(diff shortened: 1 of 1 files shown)


test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
```diff
@@ -604,19 +604,32 @@ public void ActivationSched_SubTaskExecutionSequencing()
             {
```
ReubenBond: Above here we have
```C#
                taskChainEnds[chainNum] = task.ContinueWith(t =>
                {
                    if (t.IsFaulted) throw t.Exception;
                    output.WriteLine("Inside Chain {0} Final Task", chainNum);
                    resultHandles[chainNum].SetResult(true);
                }, scheduler);
```

Should we also call `resultHandles[chainNum]..SetException(t.Exception)` in the `t.IsFaulted` case? To speed failures along.

test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
```diff
@@ -604,19 +604,32 @@ public void ActivationSched_SubTaskExecutionSequencing()
...
                 catch (TimeoutException)
                 {
-                    Assert.True(false, "Result did not arrive before timeout " + timeoutLimit);
+                    Assert.True(false, "Result did not arrive before timeout " + waitCheckTime);
                 }
 
                 bool ok = resultHandles[i].Task.Result;
-                Assert.False(taskChainEnds[i].IsFaulted, "Task chain " + i + " should not be Faulted: " + taskChainEnds[i].Exception);
+                
+                if (!taskChainEnds[i].IsCompleted)
```
ReubenBond: Why not await unconditionally?

test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
```diff
@@ -604,19 +604,32 @@ public void ActivationSched_SubTaskExecutionSequencing()
...
+                {
+                    try
+                    {
+                        // since resultHandle being complete doesn't directly imply that the final chain was completed (there's a chance for a race condition), give a small chance for it to complete.
+                        await taskChainEnds[i].WithTimeout(TimeSpan.FromMilliseconds(10));
```
ReubenBond: If we bumped the timeout to 10s instead, we wouldn't lose correctness, but we might gain reliability.

test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
jdom: nah, there shouldn't be any failures anyway, so no point in optimizing that path

ReubenBond:
@dotnet-bot please test this please thanks

test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
jdom: didn't want to do this, as I wasn't sure about the intentions... otherwise the more correct way would be to get rid of the resultHandle stuff, and just await the chain end...

test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
jdom: since the WithTimeout extension method is more involved than I'd like, I wanted to keep this loop lightweight. But in hindsight it probably doesn't matter, I'll go with your suggestion

test/TesterInternal/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
ReubenBond: I think given there's no bound on the amount of time between resultHandle completion & taskChainEnd completion, it would give us more certainty, especially if we start running many tests in parallel.

Good point about awaiting taskChainEnd - maybe it's related to deadlocks or perhaps something pre .NET 4.5, but I'm not sure.

PR #2259, opened 2016-10-05
Fix error message when type in FallbackSerializationProvider cannot be loaded.

ReubenBond:
Fix error message when type in <FallbackSerializationProvider type="..."/> cannot be loaded.

### DIFF
--- a/src/Orleans/Configuration/MessagingConfiguration.cs
+++ b/src/Orleans/Configuration/MessagingConfiguration.cs
@@ -345,7 +345,7 @@ namespace Orleans.Runtime.Configuration
 
                     var type = ConfigUtilities.ParseFullyQualifiedType(
                         typeName,
-                        "The type specification for the 'type' attribute of the Provider element could not be loaded");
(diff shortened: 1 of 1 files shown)


src/Orleans/Configuration/MessagingConfiguration.cs
```diff
@@ -345,7 +345,7 @@ internal virtual void Load(XmlElement child)
 
                     var type = ConfigUtilities.ParseFullyQualifiedType(
                         typeName,
-                        "The type specification for the 'type' attribute of the Provider element could not be loaded");
+                        "The type specification for the 'type' attribute of the FallbackSerializationProvider element could not be loaded");
```
jdom: It would be very useful to include the typeName also in the exception message

ReubenBond: Good point. I'll fix it for the this & the Provider config above.

jdom:
@dotnet-bot test this please
reason: seemingly unrelated test failure http://dotnet-ci.cloudapp.net/job/dotnet_orleans/job/master/job/innerloop_prtest/239/

Issue #2252, opened 2016-10-04
Remove unnecessary test dependency on OrleansManager


ReubenBond:
OrleansManager is referenced by a bunch of projects in our codebase, as seen below. It's only ever referenced for one thing: `RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID`, which is just a const equal to `1`.

Is there a reason to reference it which I'm not seeing?
Would someone be happy to move this constant elsewhere?

```
<Tester> (10 items)
  MembershipTests (2 items)
    LivenessTests.cs (2 items)
      (42,94) IManagementGrain mgmtGrain = GrainClient.GrainFactory.GetGrain<IManagementGrain>(RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID);
  PSClientTests (2 items)
    PSClientTests.cs (2 items)
      (125,55) getGrainCommand.Parameters.Add("LongKey", RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID);
  ManagementGrainTests.cs (2 items)
    (28,77) mgmtGrain = GrainClient.GrainFactory.GetGrain<IManagementGrain>(RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID);
  OrleansTestingBase.cs (2 items)
    (42,94) IManagementGrain mgmtGrain = GrainClient.GrainFactory.GetGrain<IManagementGrain>(RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID);
  TestUtils.cs (2 items)
    (97,94) IManagementGrain mgmtGrain = GrainClient.GrainFactory.GetGrain<IManagementGrain>(RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID);
<TesterInternal> (22 items)
  ActivationsLifeCycleTests (2 items)
    ActivationCollectorTests.cs (2 items)
      (197,94) IManagementGrain mgmtGrain = GrainClient.GrainFactory.GetGrain<IManagementGrain>(RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID);
  General (4 items)
    ElasticPlacementTest.cs (2 items)
      (302,94) IManagementGrain mgmtGrain = GrainClient.GrainFactory.GetGrain<IManagementGrain>(RuntimeInterfaceConstants.SYSTEM_MANAGEMENT_ID);
    RequestContextTest.cs (2 items)
[...]

PR #2261, opened 2016-10-05
Remove test dependency on OrleansManager (#2252)

ReubenBond:
As per #2252, this inlines the constant id for IManagementGrain and removed test project dependencies on OrleansManager

### DIFF
--- a/src/OrleansManager/OrleansManager.csproj
+++ b/src/OrleansManager/OrleansManager.csproj
@@ -50,7 +50,6 @@
     </Compile>
     <Compile Include="Program.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
-    <Compile Include="RuntimeInterfaceConstants.cs" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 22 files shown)


jdom:
I won't push back if you think it's OK, but how about creating another constant in the relevant project instead of using that magic number (at least in projects where you use that value more than once, for example in Tester (that all other projects reference anyway))? I guess it's not relevant which ID to use, but conceptually it's better to reuse it, and since you are using all 1s, then it feels like it should be a constant somewhere...

ReubenBond:
@jdom I'm happy to make that change, but it wouldn't be 100% accurate, since `OrleansManager` itself only references `Orleans` and it would be a shame to add a new public constant. What if I were to change it to `0` instead of `1`? `1` looks odd to me, anyhow, since I think `0` is more conventional for singletons - but that's the value which was being used.

ReubenBond:
Changed the value to 0

gabikliot:
LGTM.

PR #2262, opened 2016-10-05
Do not trace to the console by default when running tests

jdom:
I believe I unintentionally changed our default to trace to console when running our tests (#2223), and since we are tracing to a log file already, this is unnecessarily increasing our log file, and I believe in some cases it might be failing the build, as errors written to the console are treated as build failures in jenkins (I think it's related to the cryptic error we recently started to see: `Failure to restore a package: Unable to resolve service for type 'System.String' while attempting to activate 'UnitTests.Grains.ExplicitlyRegisteredSimpleDIGrain`).

### DIFF
--- a/test/Tester/DependencyInjectionGrainTests.cs
+++ b/test/Tester/DependencyInjectionGrainTests.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
-using Orleans;
 using Orleans.Runtime;
 using Orleans.Runtime.Configuration;
 using Orleans.TestingHost;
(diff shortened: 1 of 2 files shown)


ReubenBond:
That explains those weird failure messages.
Should this change also be made in TestClusterOptions.cs:43? I see `{ nameof(TraceToConsole), "true" },` there.

jdom:
Not necessarily, this is how it was before my change. By default it logs everywhere to ease troubleshooting in case the test runner uses the output. Nevertheless I don't mind changing the default if that is what we want, since it does log to a file anyway

ReubenBond:
Makes sense, I went and looked at the old behavior and it looks like this restores it.

PR #2264, opened 2016-10-05
Fix some .NET Standard warnings

ReubenBond:
`Marshal.SizeOf(Type)` is obsolete in `System.Runtime.InteropServices` as of version 4.1.0.0, so this PR replaces its usage with a set of if/else statements.

Also fixes a documentation issue with `Type.IsAssignableFrom(Type)` which exists on .NET Standard as `TypeExtensions.IsAssignableFrom(Type, Type)`

### DIFF
--- a/src/Orleans/Serialization/SerializationManager.cs
+++ b/src/Orleans/Serialization/SerializationManager.cs
@@ -1228,45 +1228,22 @@ namespace Orleans.Serialization
 
         private static void WriteEnum(object obj, BinaryTokenStreamWriter stream, Type type)
         {
-            var underlyingType = Enum.GetUnderlyingType(type);
-            var size = Marshal.SizeOf(underlyingType);
-            switch (size)
-            {
-                case 1:
-                    stream.Write(Convert.ToByte(obj));
-                    break;
-                case 2:
-                    stream.Write(Convert.ToInt16(obj));
-                    break;
-                case 4:
-                    stream.Write(Convert.ToInt32(obj));
-                    break;
-                case 8:
-                    stream.Write(Convert.ToInt64(obj));
-                    break;
-                default:
-                    throw new NotSupportedException($"Serialization of type {type.GetParseableName()} is not supported.");
-            }
+            var t = Enum.GetUnderlyingType(type).TypeHandle;
+            if (t.Equals(byteTypeHandle) || t.Equals(sbyteTypeHandle)) stream.Write(Convert.ToByte(obj));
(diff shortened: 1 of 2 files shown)


src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -1228,45 +1228,22 @@ public static void SerializeInner(object obj, BinaryTokenStreamWriter stream, Ty
...
-                    break;
-                default:
-                    throw new NotSupportedException($"Serialization of type {type.GetParseableName()} is not supported.");
-            }
+            var t = Enum.GetUnderlyingType(type).TypeHandle;
+            if (t.Equals(byteTypeHandle) || t.Equals(sbyteTypeHandle)) stream.Write(Convert.ToByte(obj));
+            else if (t.Equals(shortTypeHandle) || t.Equals(ushortTypeHandle)) stream.Write(Convert.ToInt16(obj));
+            else if (t.Equals(intTypeHandle) || t.Equals(uintTypeHandle)) stream.Write(Convert.ToInt32(obj));
+            else if (t.Equals(longTypeHandle) || t.Equals(ulongTypeHandle)) stream.Write(Convert.ToInt64(obj));
+            else throw new NotSupportedException($"Serialization of type {type.GetParseableName()} is not supported.");
```
xiazen: what about 
        private static readonly RuntimeTypeHandle floatTypeHandle = typeof(float).TypeHandle;
        private static readonly RuntimeTypeHandle doubleTypeHandle = typeof(double).TypeHandle;
        private static readonly RuntimeTypeHandle charTypeHandle = typeof(char).TypeHandle;
        private static readonly RuntimeTypeHandle boolTypeHandle = typeof(bool).TypeHandle;

?
Do we serialize them?

ReubenBond: According to the docs [here](https://msdn.microsoft.com/en-us/library/sbbt4032.aspx)
> The approved types for an enum are byte, sbyte, short, ushort, int, uint, long, or ulong.

I tested with char, just to be extra sure:
image

xiazen:
interesting, I tested yesterday with

      ```
  enum Days
        {
            Sat = 's',
            Sun = 'u'
        }
```

And it passed compile. Maybe the char was converted to byte.

xiazen:
But since docs is very clear that char cannot be an enum!  so I'm gonna merge this.

ReubenBond:
I checked your example:
image

PR #2253, opened 2016-10-04
Add Orleans.NonSiloTests to main solution

xiazen:
relink vNext solution with moved files,
    added Tester project to vNext,
    fixed build error introduced by #2223

    fix MainSolution/Orleans.NonSiloTests proj and link projects together

    Move non silo tests files to MainSolution/ Orleans.NonSiloTests project

    add NonSiloTestsIn  to main solution

PR #2266, opened 2016-10-06
.NET Standard support non-serialized fields via NonSerializedAttribute

ReubenBond:
Fields in .NET Standard do not have a `FieldInfo.IsNotSerialized` property but the corresponding `[NonSerialized]` attribute is supported, so currently we are serializing fields which are marked as `[NonSerialized]`.

This PR fixes that.

An upcoming PR adds additional testing for non-serialized fields, such coverage exists today, but only for `IlBasedFallbackSerializer`. The new PR is blocked by #2253 & #2261.

### DIFF
--- a/src/Orleans/Serialization/IlBasedSerializerGenerator.cs
+++ b/src/Orleans/Serialization/IlBasedSerializerGenerator.cs
@@ -190,10 +190,8 @@ namespace Orleans.Serialization
         {
             var result = type.GetAllFields().Where(
                 field =>
-#if !NETSTANDARD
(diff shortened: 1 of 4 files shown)


xiazen:
#2253 will be merged today if nothing wrong happen.

xiazen:
did you build against both main and vNext solution?

ReubenBond:
@xiazen yep, why - is it not working for you? I rebased on master and it seems fine

xiazen:
nah just double check before I merge this. LGTM

PR #2269, opened 2016-10-06
Unload AppDomain in background thread when called from finalizer

jdom:
Unloading an AppDomain is not supported from a finalizer thread: http://stackoverflow.com/questions/4064749/appdomain-unload-throws-in-finalizer
Regardless of AppDomains, expensive operations shouldn't be attempted anyway from the finalizer thread


ReubenBond:
Is it valid to defer a task on an object which is being finalized?

jdom:
hmmm, you might have a good point there...
Maybe I should not do this at all? I'm scared that we might leak resources... we weren't doing this before #2241, but I introduced it just in case, out of paranoia. It turned out that now our tests in the Tester.dll are aborting without completing due to trying to unload an app domain from the finalizer thread...
I think we still need to troubleshoot why we are leaking this, but still thought to try to prevent leaks even when sloppy

ReubenBond:
I think that it's safe to do, but if we wanted to feel extra certain, we could just execute `StopSilo` on a shallow copy of the `AppDomainSiloHandle`, right?

jdom:
How about this approach? basically I made the clean up a per-SiloHandle implementation concern, and in this case we just care to unload the app domain and nothing else, without referencing any members after the finalizer ran

ReubenBond:
Sounds good to me

jdom:
Run all Functional tests for Tester.dll, and the process is no longer aborting, so it should be good to squash and merge

PR #2273, opened 2016-10-07
Make the "No target activation" warning easier to troubleshoot

jdom:
Make the "No target activation" warning easier to troubleshoot in the case of a client disconnection, otherwise it could mislead into thinking that there is no *grain* activation available.


xiazen:
LGTM

PR #2274, opened 2016-10-07
add remaining NonSiloTests to NonSiloTests

xiazen:
and also Moved TestUtil.cs to TestExtension project to avoid circular reference.


test/NonSiloTests/Orleans.NonSiloTests.csproj
```diff
@@ -259,6 +275,18 @@
...
+      <Visible>False</Visible>
+      <ProductName>Microsoft .NET Framework 4.5.1 %28x86 and x64%29</ProductName>
+      <Install>true</Install>
+    </BootstrapperPackage>
+    <BootstrapperPackage Include="Microsoft.Net.Framework.3.5.SP1">
+      <Visible>False</Visible>
+      <ProductName>.NET Framework 3.5 SP1</ProductName>
+      <Install>false</Install>
+    </BootstrapperPackage>
+  </ItemGroup>
```
xiazen: Oh no. I believe I deleted this before I post PR. Guess rebasing ruined it. Will fix it

test/NonSiloTests/Orleans.NonSiloTests.csproj
```diff
@@ -16,6 +16,21 @@
...
+    <UpdateInterval>7</UpdateInterval>
+    <UpdateIntervalUnits>Days</UpdateIntervalUnits>
+    <UpdatePeriodically>false</UpdatePeriodically>
+    <UpdateRequired>false</UpdateRequired>
+    <MapFileExtensions>true</MapFileExtensions>
+    <ApplicationRevision>0</ApplicationRevision>
+    <ApplicationVersion>1.0.0.%2a</ApplicationVersion>
+    <IsWebBootstrapper>false</IsWebBootstrapper>
+    <UseApplicationTrust>false</UseApplicationTrust>
+    <BootstrapperEnabled>true</BootstrapperEnabled>
```
xiazen: Yeah I do install the publish tool. Not sure is it related to that.

xiazen:
run full functional tests : https://microsoftstudios343.visualstudio.com/defaultcollection/Orleans/Orleans%20Team/_build/xaml?_a=summary&buildId=9227&tab=summary failed one tests which could fail intermittently due to SerializationManager is static. Rerun it locally the test succeed.

ReubenBond:
@xiazen I'm seeing these tests fail consistently in the vNext project when run individually. They pass in main sln.
image

src/OrleansSQLUtils/Properties/AssemblyInfo.cs
```diff
@@ -21,4 +21,5 @@
 [assembly: InternalsVisibleTo("UnitTests")]
 [assembly: InternalsVisibleTo("TesterInternal")]
 [assembly: InternalsVisibleTo("UnitTestGrains")]
+[assembly: InternalsVisibleTo("Orleans.NonSiloTests")]
```
jdom: We shouldn't add this. If you are migrating SQL tests, do not migrate them to NonSiloTests because we eventually want them to go to OrleansSqlUtils.Tests (see https://github.com/dotnet/orleans/issues/2011). I'd say either keep them where they are, or start moving to the provider specific test project.

test/NonSiloTests/General/RequestContextTestsNonSiloRequired.cs
```diff
@@ -0,0 +1,393 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Orleans;
+using Orleans.Runtime;
+using Orleans.Serialization;
+using Xunit;
+
+namespace UnitTests.General
+{
+    public class RequestContextTests_Local : IDisposable
```
jdom: since there is no easy way of diffing this, I assume there was no change at all, just moving code around?

test/NonSiloTests/Orleans.NonSiloTests.csproj
```diff
@@ -74,6 +98,18 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="..\TesterInternal\Data\TestDb.mdf">
+      <Link>Data\TestDb.mdf</Link>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+    <Content Include="..\TesterInternal\Data\TestDb_log.ldf">
+      <Link>Data\TestDb_log.ldf</Link>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
```
jdom: Are these needed for something? I remember asking around and nobody knew whay it was needed or whether we shold just kill it

test/NonSiloTests/Orleans.NonSiloTests.csproj
```diff
@@ -168,6 +213,10 @@
       <Project>{6ff2004c-cdf8-479c-bf27-c6bfe8ef93e0}</Project>
       <Name>OrleansRuntime</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\src\OrleansSQLUtils\OrleansSQLUtils.csproj">
+      <Project>{47e9ae37-4eae-48d5-a778-2c80e9a6375f}</Project>
+      <Name>OrleansSQLUtils</Name>
+    </ProjectReference>
```
jdom: same as before, this test project should be able to test the Orleans core without any extensibility providers

test/TestExtensions/TestExtensions.csproj
```diff
@@ -41,10 +41,25 @@
...
+      <Name>OrleansRuntime</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\src\OrleansTestingHost\OrleansTestingHost.csproj">
+      <Project>{40ee3b00-d381-485f-9c69-ff706837deed}</Project>
+      <Name>OrleansTestingHost</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\src\Orleans\Orleans.csproj">
+      <Project>{bc1bd60c-e7d8-4452-a21c-290aec8e2e74}</Project>
+      <Name>Orleans</Name>
+    </ProjectReference>
```
jdom: Since you are adding references to these (which I was arbitrarily avoiding, but I think it's fine to do anyway), we can treat this as the shared project to avoid having NonSiloTests and TesterInternal have to reference Tester.

vNext/test/NonSiloTests/project.json
```diff
@@ -1,5 +1,6 @@
 ﻿{
   "dependencies": {
+    "System.Management.Automation": "6.1.7601.17515",
```
jdom: why is this needed? It might be a legitimate usage for now, but I'd like to understand it before letting it in, as it might mean it won't work in the WALK phase when we run tests in .NET Core.

xiazen:
@ReubenBond  let me double check, it worked for me .

test/NonSiloTests/General/RequestContextTestsNonSiloRequired.cs
xiazen: yes just move things around. since one file contains tests which require a silo and tests which doesn't request a silo

test/TestExtensions/TestExtensions.csproj
xiazen: yes this is the intension.

xiazen: I guess I treat it like a TesterUtils project, since NonSiloTests and TestsInternal will have to reference each other just for the test utils class. Or we can create TestUtils project

src/OrleansSQLUtils/Properties/AssemblyInfo.cs
xiazen: it is for `GarewaySelection_SqlServer` test, in `GatewaySelectionTests`. It uses a `SqlMembershipTable `as a `GatewayListProvider`. Since this is a Gateway selection tests, not a provider specific test, I wasn't sure where to put it. 

Do you still think we should keep it in original project for now?

xiazen: I will move it back to original project and we can talk about this later.

xiazen:
@ReubenBond 
After looking into it, I think it failed by design for now. 

for test SerializationTests_Exception_DotNet, is deserialization for Catalog.DuplicateActivationException supported in vNext now? 

For test ServiceInterface_IsGrainClass, I think it failed because our assembly loading in vNext is not intact. It failed on not able to resolve `Orleans.Runtime.SystemTarget`, see TypeUtils.IsSystemTargetClass method, line 149

For the three tests tesing RequestContext,  it failde due to RequestContext is not fully upgraded to coreclr yet, there's a PR out for fixing that

vNext/test/NonSiloTests/project.json
xiazen: Yes it is needed for DeployTest,  it has dependency on System.Management.Automation

xiazen: This package doesn't have any dependency. So I think it would fine running in ,NET core? https://www.nuget.org/packages/System.Management.Automation

test/NonSiloTests/Orleans.NonSiloTests.csproj
jdom: Ok, weird, so it's needed just for those tests? We can probably avoid it at some point

test/TestExtensions/TestExtensions.csproj
jdom: Well, yeah, I was originally thinking of a separate project, but there's no need to do that, so better keep the unnecesary project count low :) I'm fine with this change

vNext/test/NonSiloTests/project.json
jdom: well, no, it's because it won't work in .NET Core 👅 
This is related to WMI, which is not something we can support outside of Windows... not sure what I'd do with this test... It's probably better to keep it where it is and postpone this decision (as the most likely thing is that we should treat this as a .NET Framework only extension in its own project)... I believe those test don't even work today

xiazen: Ahh, it is needed for PowerShell in the tests. https://msdn.microsoft.com/en-us/library/system.management.automation(v=vs.85).aspx

PowerShell is open source and can run in multi-platform now: https://azure.microsoft.com/en-us/blog/powershell-is-open-sourced-and-is-available-on-linux/

So maybe we got a shot there ?

jdom: but `DeployTest` uses powershell scripts that themselves use WMI, so not really. Also looked at that test class, and all tests are marked as `Skip="FIX me"`, so let's not add this yet for some tests that don't even run.

xiazen: moved DeployTests back to Tester

xiazen: And rebased

test/NonSiloTests/Orleans.NonSiloTests.csproj
```diff
@@ -48,21 +48,43 @@
...
+    <Compile Include="OrleansRuntime\Streams\SubscriptionMarkerTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="..\..\src\Build\GlobalAssemblyInfo.cs">
       <Link>Properties\GlobalAssemblyInfo.cs</Link>
     </Compile>
     <Compile Include="SchedulerTests\OrleansTaskSchedulerAdvancedTests.cs" />
     <Compile Include="SchedulerTests\OrleansTaskSchedulerAdvancedTests_Set2.cs" />
     <Compile Include="SchedulerTests\OrleansTaskSchedulerBasicTests.cs" />
-    <Compile Include="SerializationTests\BuiltInSerializerTests.cs" />
+    <Compile Include="SerializationTests\BondSerializationTests.cs" />
```
jdom: >BondSerializationTests [](start = 41, length = 22)

hmmm, I'm a little sad we are introducing this to the new NonSiloTests project instead of in its own project, but I want to get this merged soon, so I guess it's as easy to move from here than from Tester. In upcoming PRs, please don't bring in new tests for extensions (ie: Bond is not core to Orleans, just an extension) into the new test projects that we'll start creating, as this would just split the currently 2 huge conglomerate of tests into 3, but not really vertical separation.

test/Tester/Tester.csproj
```diff
@@ -146,33 +145,15 @@
...
-    </Content>
+    <Content Include="DeploymentTests\Configuration\Deployment.xml" />
+    <Content Include="DeploymentTests\Configuration\TestAbsolutePaths\Deploy-AbsolutePaths.xml" />
+    <Content Include="DeploymentTests\Configuration\TestAbsolutePaths\OrleansConfiguration.xml" />
+    <Content Include="DeploymentTests\Configuration\TestCleanEmptyOrleans\Clean-EmptyRelativePaths.xml" />
+    <Content Include="DeploymentTests\Configuration\TestCleanEmptyOrleans\OrleansConfiguration.xml" />
+    <Content Include="DeploymentTests\Configuration\TestCleanOrleans\Clean-RelativePaths.xml" />
+    <Content Include="DeploymentTests\Configuration\TestCleanOrleans\OrleansConfiguration.xml" />
+    <Content Include="DeploymentTests\Configuration\TestRelativePaths\Deploy-RelativePaths.xml" />
+    <Content Include="DeploymentTests\Configuration\TestRelativePaths\OrleansConfiguration.xml" />
```
jdom: Please undo these changes entirely, as we need the CopyToOutputDirectory

xiazen: I cannot undo these changes entirely. Move DeploymentTest to NonSiloTests is  part of preview commit. But I can fix it by set it to CopyToOutpytDirectory.

test/NonSiloTests/Orleans.NonSiloTests.csproj
xiazen: I'm not 100% convinced that we should have separate project for providers.  We can give it tags or in its own folder if you think you think it's hard to find all the tests related to external providers. 

The thing is, if we categorize too much, it would be hard to find a place to put tests like BondSerializationTests or test like GatewaySelectionTest, where only one test in those file related to SqlMembershipTable. So in this situation , we would split the tests into two files and one in NonSiloTests, one in  SQLMemebershipTests. It's confusing to find all the tests related to GatewaySeleciton in this case

Issue #2195, opened 2016-09-21
.NET Standard compat: use supported overloads from Socket class


jdom:
In our .NET Standard solution I temporarily removed the functionality in `IncomingMessageAcceptor` as we were using overloads of the Socket class that are not available in .NET Standard. See https://github.com/dotnet/orleans/commit/30b7a4468a7e96f0973875d4e41b321ef4ad91f4 for the commit where this was excluded.

We need to update the code to use supported overloads. For stuff like accepting a new connection or listening (stuff that it's not in the hot path), it is fine to use the overload that returns a Task. Nevertheless for the receive operation we need to be performance conscious. We can potentially use a similar approach as in  https://github.com/dotnet/orleans/commit/02753a3666b44b520d9e7a2f7acdb776e155093d for that particular operation.

In order to open the solution that compiles using .NET Standard (and hence easy to check whether certain overloads are available), check out [this readme](https://github.com/dotnet/orleans/blob/master/vNext/README-CoreClr-port.md)

dVakulen:
As per https://github.com/dotnet/corefx/commit/aed4a2f12d0929d0998058f0a6c8a363d807e4eb and https://github.com/dotnet/corefx/commit/c714efcf986c3a51ce94ff8aeb3b96cfe05ca9c8 it looks like that currently unavailable methods will be back soon.

jdom:
I assume this is not really *soon* though. I guess it will come with .NET Standard 2.0 next year, or do you think this particular change is coming before it? Because this is a big blocker (although probably trivially fixable) that we don't want to let linger for more than a few weeks

dVakulen:
Well, if put that way, it's indeed not soon at all.

dVakulen:
I'll try to handle this.

jdom:
Hi @dVakulen, any chance you could have a PR this week? No pressure, it's just to plan accordingly, otherwise we might just grab it as this piece is very important to get done soon.

dVakulen:
Hi @jdom, yes, I'll make a PR by the end of the week.

dVakulen:
I've made an implementation that can be used for running of the test suite, but have some performance issues which I'd like to fix before submitting of the PR.

jdom:
Any updates or anything we can do to help? I thought it would not be that much work to switch one overload for the other, did you encounter some unknowns or were you trying to re-write the entire thing?
[...]

PR #2275, opened 2016-10-09
IncomingMessageAcceptor sockets change from APM to EAP

dVakulen:
Addresses #2195;


dVakulen:
Reason of changes in IncomingMessageBuffer: <code>SocketAsyncEventArgs</code> that is used  with *Async methods pins byte arrays that are used for receiving, so it's generally not recommended to mess with it's List\<ArraySegment\<byte\>\>, instead it's better to keep parts of incomplete message between receives in separate place (<code>MessagePrefixHandler in this implementation</code>).

gabikliot:
Jason Bragg is a boss on that topic! We tried to do that in the past, and he run into some issues. I am sure he will have something to say about this.

jason-bragg:
Reviewing now, but TryDecodeMessage originally didn't update decodeOffset until the entire message (lengths, headers, and body) had been read, and CalculateKnownMessageSize calculated the total size of the message known at any point in time.  This per message 'read then modify' approach simplified the reasoning about state.  In the new code, we update the decodeOffset every time we read data from the buffer, which means we need to be more careful about tracking correctness at each possible state.

This is not an abjection to the current approach, as I've not finished my review, it is merely an observation that ask be considered.

src/Orleans/Messaging/IncomingMessageBuffer.cs
```diff
@@ -1,44 +1,329 @@
...
+            public List<ArraySegment<byte>> GetPrefix()
+            {
+                return buf;
+            }
+
+            public void Reset()
+            {
+                Count = 0;
+                HasPrefix = false;
+                BufferPool.GlobalPool.Release(buf);
```
jason-bragg: BuildSegmentListWithLengthLimit does not copy the data, it just carves up the list of segments.  This means we're returning buffer to the pool that are still being used in the message we are currently processing.

jason-bragg:
Among my original work with this networking layer, I attempted to port it to using SocketAsyncEventArgs.  This work was abandoned because we found a bug in the .net networking libraries.  When setting a new buffer list per call, the new buffer list would at time be replaced with the old buffer list due to an underlying threading issue.  In general, to use SocketAsyncEventArgs we would need to use the same buffer each time, which is what it looks like you’re attempting to do here.  The issue was reported, but we’ve not received confirmation of a fix or details of which release it is (will be) in.

I raise the above to clarify that if we move to SocketAsyncEventArgs, changing the way we manage our buffers in receive calls is not an optimization issue as much as a necessity for the system to function correctly.

Having said that, this change, to me, looks error prone and difficult to maintain.  I'm still reviewing but I'm not convinced of it's correctness.  I suggest the following instead.

Allocate a fixed size read buffer per connection that is one large contiguous buffer.  The size should be a multiple of the block size.
After each read, copy the buffer into blocks from the buffer pool.
Process the buffers as before.

This brute force yet simple approach will introduce an additional network copy, which should always be avoided if possible in networking layers, but is a simpler change and easier to maintain.  We will need to perf test this change to assess the cost of the additional copy, but with luck the perf gain from moving to SocketAsyncEventArgs will offset the additional buffer copy cost.

Given that we are not de-serializing our messages immediately after reading them (sometimes not at all if they are being forwarded, I don't see a way round an additional data copy.  We either swap the buffers per call, or copy them.

dVakulen:
@jason-bragg  Can you please provide some of your arguments about error proneness and difficulties to maintain? 
Regarding large contiguous buffer - I thought that we want to avoid allocations in LOH, which are usually starts from 85kb objects, while current default buffer size is 128kb. 

And anyway, let's perf test this one before switching to anything else, in order to have numbers for comparison.

src/Orleans/Messaging/IncomingMessageBuffer.cs
```diff
@@ -1,44 +1,329 @@
...
+            var result = new List<ArraySegment<byte>>(readBuffer.Count);
+            ByteArrayBuilder.BuildSegmentListWithLengthLimit(
+                result,
+                prefixHolder.GetPrefix(),
+                decodeOffset,
+                length,
+                out bytesRead);
+
+            length -= bytesRead;
+            decodeOffset += bytesRead;
```
dVakulen: Decode offset now does not make a difference between stored and network buffers. I was thinking about hiding the buffer implementation details in the separate buffer, but this was the most obvious solution.

src/Orleans/Messaging/IncomingMessageBuffer.cs
```diff
@@ -1,44 +1,329 @@
...
+
+            if (length <= 0)
+            {
+                return result;
+            }
+
+            // we've read past the stored message prefix, so dropping it
+            if (bytesRead > 0)
```
dVakulen: Sorry, that's due to a bit cryptic in this place - due to check at https://github.com/dotnet/orleans/pull/2275/files#diff-5bac212740a340ab14eaa8ccde4ab409R222 we'll never reach that point unless all data from stored buffer was red.

src/Orleans/Messaging/IncomingMessageBuffer.cs
dVakulen: Indeed, will fix.

jason-bragg:
@dVakulen 
I apologize for speaking in generalities in my overall comments before publishing the review.  Please see my review for potential issues.
Two main areas of concern are the decode offset management and returning buffers to the pool while still in use.

As for LOH allocations, one per socket connection (IMO) should not be a concern.  If necessary we can introduce an big buffer pool for the receive buffers.  We can use a list of block from the block pool instead, but that's just more moving parts.  If we're using the same buffer through the lifetime of the connection, a single large (pooled if necessary) block would be (marginally) more efficient.  I don't have a strong objection to lots of small (pooled) bocks rather than a single large one (this is an easy future optimization if justifiable), I'm more interested in how the message processing logic and the receive buffer interact.

> "Let's perf test this one before switching to anything else, in order to have numbers for comparison. "
Prior to perf test, please consider my review comments, and add a (temporary) log line indicating that the left over data code path is being hit.  Partial messages being left over in the buffer is not the common case, so it's not hard to have tests run and never hit that path.  Let's make sure that path is being called.  Maybe configure the receive buffer to be a single block just to ensure this get's hammered a lot...?

src/Orleans/Messaging/IncomingMessageBuffer.cs
```diff
@@ -1,44 +1,329 @@
...
+        public virtual bool TryDecodeMessage(out Message msg)
+        {
+            msg = null;
+            if (decodeOffset == receiveOffset)
+            {
+                return false;
+            }
+
+            // Is there enough read into the buffer to continue (at least read the lengths?)
+            if (TryHandleMessageLengthPrefix()) return false;
```
jason-bragg: > I've avoided unnecessary check. 

What unnecessary check?
headerLenght== 0 || bodyLength ==0?  This is a cheaper check than TryHandleMessageLengthPrefix.  In the most common case, both checks are made anyway.  In the less common case, only headerLenght== 0 || bodyLength (the cheaper check) should be necessary.

_correctness_
Improbable but not impossible scenario.  What happens if a read gets 8 bytes, (we read lengths but not  header or body and nothing in holder), then header and body are read in next loop but they are less than 8?  Say they are 6 bytes?

AvailableReadLenght (0 + 6 – 0) < readLength (8)

In this case the header and body are there, but we don’t process them because we are looking for the lengths size.

If we’re not looking for lengths, because they’ve already been read, why are we checking if there is enough data to read them?  While this scenario is not likely, the logic is technically incorrect.

Am I missing something?

jason-bragg:
@dVakulen 
What I'm asking is that you take a step back and look at the problem at a higher level.  I've no doubt you can get this approach to work, but please consider the subtle problems you've encountered so far and imagine the maintainability of this system by an engineer that has less insight into this system (and raw aptitude) than yourself.  Is there a simpler and more maintainable approach?  I suspect there is, but I’m not 100% sure, so I’m asking only that you consider the question.

If we have a fixed network buffer, be it a block list or a contiguous buffer, then we need to process or copy data each loop.  Since our processing works with a list of buffers, we can just copy the network buffer into blocks that append to whatever was left from the previous read, or if we want to be more efficient, just add the network block to the end of the list of remaining blocks, then process them.  When all messages have been processed from a buffer, we keep whatever is left in a block list for the next pass.

Main thing here is that in both cases, the buffer processing (move complicated part) does not care.  We only need to worry about this at the beginning and end of the buffer processing loop.  The current approach of trying to manage two buffer rather than a single (append only) buffer list seems unnecessarily complicated.

WorkingBuffer – buffer list we process messages from.
ReceiverBuffer – buffer or buffer list we use in socket read

_Copy all_
Each network read loop:
	WorkingBuffer.AppendAndCopyBlocks(ReceiverBuffer)
	WorkingBuffer.ProcessMessagesAndAdjustBuffer

_Copy only remaining_
Each network read loop:
	WorkingBuffer.AppendBlocks(ReceiverBuffer) – no copy, just append receive buffer (list) to block list.
	WorkingBuffer.ProcessMessages
	WorkingBuffer = Copy(WorkingBuffer.AllAfter(decodeOffset))

sergeybykov:
I'm checking with the owners of the socket APIs if the bug we encountered with `SocketAsyncEventArgs` got fixed.

I'm a bit concerned about this change ballooning up quickly. Ideally, I think we should try to separate, if possible, the API change for .NET Core compatibility and the bigger and riskier effort of changing how buffers are used. Is it possible to split this PR into two: a smaller one for the former, and another one for the latter?

Longer term, we'd like to 'outsource' this whole layer to an external library like DotNetty anyway. So investments into further optimizing our proprietary solution will be lost the moment we do that. Not a showstopper, of course, just a factor to consider.

dVakulen:
I've considered approach with keeping IncomingMessageBuffer untouched previously, but overestimated costs of buffers copying. Also I'm not quite convinced that solution with managing of two buffers is much more complicated than the current one. And I didn't had in mind optimizing of Orleans proprietary  solution  - just wanted not to drop the performance due to the rewrite. 

With latest commit switched to "Copy All" implementation and added pooling of SocketAsyncEventArgs.

src/OrleansRuntime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -235,153 +330,176 @@ private static void AcceptCallback(IAsyncResult result)
...
+
+            // Creates with incomplete state: IMA should be set before using
+            readEventArgs.UserToken = new ReceiveCallbackContext(null, null, poolWrapper);
+            allocatedSocketEventArgsCounter.Increment();
+            return poolWrapper;
+        }
+
+        private void FreeSocketAsyncEventArgs(SocketAsyncEventArgs args)
+        {
+            var receiveToken = (ReceiveCallbackContext) args.UserToken;
```
jason-bragg: If we're going to reuse the ReceiveCallbackContext, we need to make sure to reset it's buffers.  Any errors that could change the socket feeding a receiver buffer must reset the buffer.

jason-bragg:
@dVakulen 
Much thanks.  I appreciate your patience.  The new pass, to me, seems much simpler and easier to reason about.  GJ

_Some comments:_
I suspect there are still some holes in the error handling, which I've comments on in the latest review pass.

While I suspect I understand the intentions behind the pooled SAEA, I don't think the added complexity is justifiable, as we don't churn through sockets fast enough for it to need optimized.

There is also a rather subtle edge case with the buffers.  If there are any socket errors, the buffers need to be recreated or clear (clearing is cheaper).  If not, a partial message from the previous socket connection could corrupt the incoming messages on the new socket.  This is a very uncommon edge case that none of our tests will catch, so please double check the error handling to ensure that new sockets have clean buffers.

dVakulen:
>Partial messages being left over in the buffer is not the common case, so it's not hard to have tests run and never hit that path. 

I bit late response - actually that code path was being hit quite often, especially with big messages.

src/OrleansRuntime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -235,153 +330,176 @@ private static void AcceptCallback(IAsyncResult result)
...
-                    catch (Exception exception)
+                    catch (Exception ex)
                     {
-                        var socketException = exception as SocketException;
-                        ima.Log.Warn(ErrorCode.Messaging_IMA_NewBeginReceiveException,
-                            String.Format("Exception on new socket during BeginReceive with RemoteEndPoint {0}: {1}",
-                                socketException != null ? socketException.SocketErrorCode.ToString() : "", rcc.RemoteEndPoint), exception);
-                        ima.SafeCloseSocket(sock);
+                        this.Log.Warn(ErrorCode.Messaging_ExceptionReceiveAsync, "Exception trying to process accept from endpoint ", ex);
+                        throw;
```
dVakulen: The catch few line below will handle it and restart the socket. Added restarting in line.

src/OrleansRuntime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -235,153 +330,176 @@ private static void AcceptCallback(IAsyncResult result)
...
-                if (ima.RecordOpenedSocket(sock))
-                {
-                    // And set up the asynch receive
-                    var rcc = new ReceiveCallbackContext(sock, ima);
-                    try
+                    }
+                    catch (SocketException ex)
                     {
-                        rcc.BeginReceive(ReceiveCallback);
+                        Log.Warn(ErrorCode.Messaging_ExceptionReceiveAsync, "Error when processing data received from {0}:\r\n{1}", "Q", ex, sock.RemoteEndPoint);
```
dVakulen: Fixed

src/OrleansRuntime/Messaging/IncomingMessageAcceptor.cs
dVakulen: Fixed

src/Orleans/Messaging/ByteArrayBuilder.cs
```diff
@@ -459,6 +459,31 @@ public static List<ArraySegment<byte>> BuildSegmentList(List<ArraySegment<byte>>
         public static List<ArraySegment<byte>> BuildSegmentListWithLengthLimit(List<ArraySegment<byte>> buffer, int offset, int length)
         {
             var result = new List<ArraySegment<byte>>();
+            int ignore;
+            BuildSegmentListWithLengthLimit(result, buffer, offset, length, out ignore);
+            return result;
+        }
+
+        public static void BuildSegmentListWithLengthLimit(List<ArraySegment<byte>> target, List<ArraySegment<byte>> buffer,
```
dVakulen: No, removed.

src/OrleansRuntime/Messaging/ObjectPool.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Threading.Tasks;
+namespace Orleans.Runtime
+{
+
+    /// <summary>
+    /// Simple object pool Interface.
+    /// Objects allocated should be returned to the pool when disposed.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal interface IObjectPool<T> where T : IDisposable
```
jason-bragg: Good point.  I was thinking we could pool the buffers later, but yeah, why pool only the buffers when we can pool the entire SAEA, which inclused the buffer.  Don't know where my brain was. :)

jason-bragg:
> actually that code path was being hit quite often, especially with big messages. 

That makes sense.  That's probably because the receivebuffer isn't auto expanding any more.  It used to expand up to 1MB, so messages smaller than that would not tend to leave leftover data.  I can see how it would be more common in the new code.

A future optimization may be to track how common it is to fill the entire receive buffer, and swap out the SAEA with one with a bigger buffer if it happens a lot, but those thoughts can wait until the next optimization pass.  If we replace this all with something else (DotNetty?) the concerns will be very different.

Issue #2272, opened 2016-10-07
Adding [PreferLocalPlacement] to grains causes placement timeout exceptions


creyke:
As per the issue title.

When we try and prefer local placement we large numbers of both of the below placement exceptions:

Orleans.Runtime.SiloUnavailableException: The target silo became unavailable for message: NewPlacement Request S192.168.57.169:22222:213551971*grn/BEA129CD/00000000@fc94258e->S192.168.57.192:22222:213552014*grn/2C1481B1/00000000@6f9046bf #30721: global::Orleans.Streams.IStreamConsumerExtension:DeliverImmutable(). Target History is: <S192.168.57.192:22222:213552014:*grn/2C1481B1/00000000:@6f9046bf>. See https://aka.ms/orleans-troubleshooting for troubleshooting help.

Orleans.Runtime.Catalog+DuplicateActivationException: DuplicateActivationException

jdom:
You are seeing this only when you upgrated to 1.3.0-beta2?
You are using the multi-cluster feature, correct? Do you experience the same thing when only 1 cluster is active?
Any more contextual information would be very valuable

sergeybykov:
`DuplicateActivationException` is by design, should be reported as an Info event. That's how racing activation requests are resolved.

The first one is a bit unclear. Is silo 192.168.57.192:22222 actually part of the cluster when this error is logged?

creyke:
No info on previous versions I'm afraid. First time we've used the attribute. Will check whether had joined Monday but we basically ran exactly the same code (other than without the attribute) multiple times both with and without it and only get these errors with it each time.

sebastianburckhardt:
I think this is a bug in `PreferLocalPlacementDirector.OnAddActivation`. Unlike the `RandomPlacementDirector.OnAddActivation`, it does not check if the local silo is actually active, thus it may place a new activation on a silo that is shutting down. This is consistent with the text of the error message (the silo being shut down cannot  create the new instance).  

My proposed fix #2276 is to revert to random placement if the local silo is not active.

sebastianburckhardt:
@creyke, if you want to test if this actually fixes your problem (perhaps it only fixes a symptom) use the branch https://github.com/sebastianburckhardt/orleans/tree/geo-samples-2272.

It's the same as the geo-samples branch but with this fix cherry-picked.

creyke:
Thanks @sebastianburckhardt - we are now doing this.

sergeybykov:
Resolved via #2276.

PR #2276, opened 2016-10-10
Make PreferLocalPlacement revert to RandomPlacement on non-active silos

sebastianburckhardt:
this is intended to fix the problems described in issue #2272.

### DIFF
--- a/src/OrleansRuntime/Placement/IPlacementContext.cs
+++ b/src/OrleansRuntime/Placement/IPlacementContext.cs
@@ -26,6 +26,8 @@ namespace Orleans.Runtime.Placement
 
         SiloAddress LocalSilo { get; }
 
+        SiloStatus LocalSiloStatus { get; }
+
         /// <summary>
         /// Try to get the transaction state of the activation if it is available on this silo
         /// </summary>
--- a/src/OrleansRuntime/Placement/PreferLocalPlacementDirector.cs
+++ b/src/OrleansRuntime/Placement/PreferLocalPlacementDirector.cs
@@ -15,6 +15,10 @@ namespace Orleans.Runtime.Placement
         internal override Task<PlacementResult> 
             OnAddActivation(PlacementStrategy strategy, GrainId grain, IPlacementContext context)
         {
+            // if local silo is not active, revert to random placement
+            if (context.LocalSiloStatus != SiloStatus.Active)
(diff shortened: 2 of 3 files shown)


src/OrleansRuntime/Placement/PreferLocalPlacementDirector.cs
```diff
@@ -15,6 +15,10 @@ internal class PreferLocalPlacementDirector : RandomPlacementDirector
         internal override Task<PlacementResult> 
             OnAddActivation(PlacementStrategy strategy, GrainId grain, IPlacementContext context)
         {
+            // if local silo is not active, revert to random placement
+            if (context.LocalSiloStatus != SiloStatus.Active)
```
jdom: in theory this is behaviorally equivalent to `! context.AllActiveSilos.Contains(context.LocalSilo)`, so you wouldn't need those breaking changes to `IPlacementContext`. I assume you did those because you worried about the efficiency of the alternative, correct?

src/OrleansRuntime/Placement/IPlacementContext.cs
```diff
@@ -26,6 +26,8 @@ internal interface IPlacementContext
 
         SiloAddress LocalSilo { get; }
 
+        SiloStatus LocalSiloStatus { get; }
```
jdom: we recently took our warning count to 0, so be sue to add XML comments for public members

src/OrleansRuntime/Placement/PreferLocalPlacementDirector.cs
sebastianburckhardt: Yes.

IPlacementContext is an internal interface - why would the changes be "breaking"?

src/OrleansRuntime/Placement/IPlacementContext.cs
sebastianburckhardt: Ok, good to know.

Doesn't matter here as far as I can tell, it's all internal.

src/OrleansRuntime/Placement/PreferLocalPlacementDirector.cs
jdom: ah, I thought it was a public interface, since we implement it in some of our VSO tests, but those have access to internals. It's fine then :)

PR #2270, opened 2016-10-06
Add additional Serialization tests

ReubenBond:
This PR:
* Adds `BuiltInSerializerTests` & dependencies to the vNext sln
* Adds additional test coverage for serialization to both solutions

PR #2277, opened 2016-10-11
Remove irrelevant IlBasedSerializerTests methods

ReubenBond:
These tests are irrelevant because IlBasedFallbackSerializer is explicitly intended to be a fallback serializer and not an external serializer (which is what the removed tests test for).

Additionally, this was the source of some non-determinism in test runs, since SerializationManager was not isolated in these tests.

### DIFF
--- a/test/NonSiloTests/SerializationTests/IlBasedSerializerTests.cs
+++ b/test/NonSiloTests/SerializationTests/IlBasedSerializerTests.cs
@@ -1,29 +1,19 @@
 namespace UnitTests.Serialization
 {
-    using System;
-    using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


jdom:
So I see that you removed them entirely instead of removing the "BVT" or "Functional" trait. Is there no value at all in them or you have the coverage for serializing complex types in some other test methods?

ReubenBond:
@jdom those complex types are now being tested in `BuiltInSerializerTests.Serialize_ComplexClass`, which was introduced in #2270.

The test cases are not valid since `IlBasedFallbackSerializer` is explicitly intended only as a fallback serializer.

PR #2278, opened 2016-10-11
Expand serialization benchmarks in OrleansBenchmarks

ReubenBond:
Updates the serialization benchmarks to cover an additional scenario and emit more diagnostics


test/Benchmarks/OrleansBenchmarks/Serialization/SerializationBenchmarks.cs
```diff
@@ -1,58 +1,311 @@
...
+                copy = SerializationManager.RoundTripSerializationForTesting(copy);
             }
+            return copy;
         }
+    }
 
-        [Benchmark]
-        public void DeserializerBenchmark()
+    [Serializable]
+    internal struct SomeStruct
```
ReubenBond: These types are included here and made internal so that they are inaccessible to the code generator. This is in order to force the use of the fallback serializer.

Issue #2283, opened 2016-10-11
Should we automatically retry when we encounter SiloUnavailableException?


benjaminpetit:
Following #2282 

If we detect that a gateway failed, should we retry automatically to "reroute" all the message not yet sent to another gateway?

There is no risk to duplicate request here if we make sure to retry only messages not sent, but we have to preserve call ordering.

gabikliot:
You are talking about msgs queued on the particular GW  in the client (that is, the msg was already addressed via this GW and is now in the GW sent queue, but not pushed via socket yet).

Yes, I think that would be nice. We can indeed save all those msgs. 
Looks at `GatewayConnection.Stop`. 
1) When we call base.Stop(); we should go to base (AsynchQueueAgent), 
2) take all the queued msgs (requestQueue) 
3) clear the msg.TargetSilo and msg.TargetActivation
4) and just resend it via msg center (MsgCenter.SendMessage(msg))

Taking msgs from requestQueue will ensure that the msg was not sent yet (since it is BlockingCollection and is already synchronized/thread safe internally), so you don't need any extra lock or flag on the msg.

benjaminpetit:
Yes, that is this use case I was talking about.

Not sure it will add lot of complexity if we bundle fixes for this and #2282. I will try to send a PR addressing both issues

gabikliot:
I prefer not to bundle fixes.

gabikliot:
Resolved via #2298. If a gateway fails but a client still has queued but yet not sent msg to this gateway, we now re-route them (via a new gateway). If the msg was already sent, we don't reroute it.

Issue #2282, opened 2016-10-11
Some Grain calls throw SiloUnavailableException when they should be successful


benjaminpetit:
Since https://github.com/dotnet/orleans/issues/942, when we detect that a gateway went down, we take all requests not completed that should be sent to this gateway and set SiloUnavailableException as the response.

The problem is that some of these requests might already have been sent and currently being run on a silo. We just should wait normally the response instead of mark them as faulted.

Here is the test I wrote to reproduce the issue: https://github.com/benjaminpetit/orleans/commit/7ef900ab10217f5073ce3f555d7bd9b1f88b0015

The flaw of this test is that I needed to modify the algorithm to choose a Gateway to only use the one pointed by the PreferedGatewayIndex value (when set), I am not sure it is an acceptable change.

PR #2285, opened 2016-10-11
Some Grain calls throw SiloUnavailableException when they should be successful 

benjaminpetit:
Proposed fix for #2282


src/Orleans/Messaging/GatewayManager.cs
```diff
@@ -134,6 +134,10 @@ public Uri GetLiveGateway()
             {
                 lock (lockable)
                 {
+                    if (config.PreferedGatewayIndex >= 0 && config.PreferedGatewayIndex < count)
+                    {
+                        return live[config.PreferedGatewayIndex];
```
jdom: Already talked in person with @benjaminpetit, but we can't do this... He'll have the tests call several times until there's a hide confidence that this path will be hit.

benjaminpetit: Fixed!

gabikliot:
I am not sure about this "fix".

First, there are 2 different issues here: 
1) msg is queued on the sender client (lets talk about client only for now) but was not sent yet vs. was already sent.
2) the gateway through which it was sent failed, but not the silo where the msg was destined to.

Your test checks 2, while the fix (TentativeCount) addresses  1 (if TentativeCount>1, do not break, otherwise [queued  but not sent] dobreak).

Issue 1: I think the issue of queued but not sent is a non issue - there is a short period between when we pick the GW and when we actually sent the msg (this queued period), so I would not address that at all. Just ignore that short period. If the msg is created but GW is not picked yet, we will not route it via dead GW anyway.

Issue 2: GW went down but you still do not want to break the msg. Well, if that is what you want, just don't do OnTargetSiloFail from BreakOutstandingMessagesToDeadSilo in OutsideRuntimeClient..
BUT, the problem is: you don't know on the client who is the real destination. It may be this GW or maybe another silo. So if you pick to not break, you might be waiting the full 30sec since the silo actually did fail.

So anyway, I don't think the prev. behavior is a bug. If you prefer just not to break msgs when GW fails, since the msg may have been re-routed to a different silo, I am OK with that. But then you don't need the TentativeCount. Simply do not break the msg.

benjaminpetit:
Well I added the TentativeCount to be able to break with an exception only when the message was not sent but I think I can find a more elegant way to do it. And after thoughts, it seems that it would be better/easier to fix #2283 and #2285 on the same commit: remove the `BreakOutstandingMessagesToDeadSilo` method and reroute all message not yet sent to another GW when possible.

On the contrary, to me it seems like a bug:

- In client -> silo case: there is a good chance that the grain you were targeting was not hosted on the silo you used as a gateway. Your call has a good chance to be successful, even if the gateway was not gracefully stopped.
- In silo -> silo case: I also think we should not fail and try to reroute the call, even if we were targeting the correct silo, because ownership might have been transferred to another silo.

To me, we should only throw `SiloUnavailableException` when the request was sent to the correct silo and that this silo was not able to process or forward it. Or when the gateway you used was not able to forward the request before stopping. But I am not sure this is doable...

gabikliot:
I agree we should try to resend any msg that we can, but only if we are 100% sure it didn't start processing yet.  For msgs we are not sure, we can't resend. And then, for those msgs we have 2 choices: pessimisticslly break early or optimistically hope to get an answer and if not time out.

The first case ( not resend when we can) is a bug. The second case ( pessimistic or optimistic) is a policy decision. I agree with you that in most cases optimistic makes more sense.

benjaminpetit:
Closing this one because it is partially fixed in #2298. Let's discuss in #2282 if we should fix it for inter silos communication

PR #2288, opened 2016-10-12
Adds native JSON support to MySQL

veikkoeeva:
Adds native JSON support to MySQL

MySQL 5.7.8 added a native JSON type, release notes
https://dev.mysql.com/doc/refman/5.7/en/json.html.

As we currently support only the latest version, this
could be included without regard to older versions. A small
problem is the official packet was released already and the
type there is plain text.

This alter to the table is backwards compatible, but for very
large tables can be time consuming.

### DIFF
--- a/src/OrleansSQLUtils/CreateOrleansTables_MySql.sql
+++ b/src/OrleansSQLUtils/CreateOrleansTables_MySql.sql
@@ -226,6 +226,12 @@ CREATE TABLE Storage
 ) ROW_FORMAT = COMPRESSED KEY_BLOCK_SIZE = 16;
 ALTER TABLE Storage ADD INDEX IX_Storage (GrainIdHash, GrainTypeHash);
 
+-- The following alters the column to JSON format if MySQL is at least of version 5.7.8.
+-- See more at https://dev.mysql.com/doc/refman/5.7/en/json.html for JSON and
+-- http://dev.mysql.com/doc/refman/5.7/en/comments.html for the syntax.
+/*!50708 ALTER TABLE Storage MODIFY COLUMN PayloadJson JSON */;
+
+
 INSERT INTO OrleansQuery(QueryKey, QueryText)
 VALUES
 (


veikkoeeva:
Ping @shayhatsor and @jdom. The tests pass at least on my machine.

veikkoeeva:
Scrap that, the official released happened just now!

shayhatsor:
@veikkoeeva, this is very cool. But I'm not sure we can make everyone upgrade to the latest version. It seems It'd be better to detect the version in this case.

veikkoeeva:
@shayhatsor And the release also went to Internet just a bit too soon for this. I was considering that we have the syntactic issue the deployment scripts so that version `5.7.11-log` (latest) accepts it, but `5.6.32-log` does not, as was found out in the chat.

I could add conditional that checks the version number and deploys JSON to those who support it and the current text form for older. To make this worthwhile, I think we should open an issue also to fix the script syntax to support older version. I remember you had the modification already, so should it be committed to the repo?

shayhatsor:
@veikkoeeva, what I think we should do is:
1) make the current script compatible with MySQL 5.6 (with the simple fix I've mentioned on gitter)
2) test the db version in order to decide about using the new JSON type column.
what do you think ?

veikkoeeva:
@shayhatsor That sounds like a plan. I have to see if it is viable to also check if the deployed database supports JSON, but used the 1.3.0 script and it could alter the column without much of a hassle I'm afraid we can't do it automatically if it consumes processing resources comparable to data stored since it can be a lot of data. In that case what we can do is to update the script and let the users migrate if they see the need for it.

veikkoeeva:
@sergeybykov I believe this could still make it to the 1.4.0 version if there aren't other concerns to this. I noticed some, recored at https://gitter.im/dotnet/orleans?at=587df0406c1635643c3c4309 (sorry, a bit busy, having a break at work).

sergeybykov:
@veikkoeeva We are trying to close 1.4.0 this week. Do you think this will be ready to merge in a few days?

veikkoeeva:
@sergeybykov There isn't more to this JSON issue, I could perhaps figure out more versatile ways of doing that and all other things, but that'd be a lot more code (as far as I can see it). Maybe useful if someone else could run the tests too, @shayhatsor?

The other things pointed in Gitter can be worked out later.

sergeybykov:
Thank you, @veikkoeeva!

PR #2291, opened 2016-10-12
Adds helper methods to ease ADO.NET configuration

veikkoeeva:
The ADO.NET functionality should utilities such
as some other providers (e.g. AzureUtils).

### DIFF
--- a/src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
+++ b/src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
@@ -53,6 +53,8 @@ namespace Orleans.Runtime.Configuration
             connectionString = GetConnectionString(connectionString, config);
             var properties = new Dictionary<string, string> { { AdoNetStorageProvider.DataConnectionStringPropertyName, connectionString } };
 
+            //In the following the call to ToString is documented to return a non-localized string
+            //and internally just a literal so should be quick. See more at https://technet.microsoft.com/fi-fi/library/s802ct92.
             if(serializationFormat == AdoNetSerializationFormat.Json)
             {
                 properties[AdoNetStorageProvider.UseJsonFormatPropertyName] = true.ToString();
--- a/src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
+++ b/src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
@@ -1,10 +1,9 @@
-﻿using Orleans.Runtime.Configuration;
-using Orleans.Storage;
+﻿using Orleans.Storage;
 using System;
 using System.Collections.Generic;
 
 
-namespace OrleansSQLUtils
+namespace Orleans.Runtime.Configuration
 {
     /// <summary>
     /// The format in which the data will serialized to the storage.
--- a/src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
+++ b/src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
@@ -1,0 +1,89 @@
+﻿using Orleans.Runtime.Configuration;
+using Orleans.Storage;
+using System;
+using System.Collections.Generic;
+
+
+namespace OrleansSQLUtils
+{
+    /// <summary>
+    /// The format in which the data will serialized to the storage.
+    /// </summary>
+    public enum AdoNetSerializationFormat
+    {
+        /// <summary>
+        /// The native Orleans binary serialization format.
+        /// </summary>
+        OrleansNativeBinary = 1,
+
+        /// <summary>
+        /// XML.
+        /// </summary>
+        Xml                 = 2,
+
+        /// <summary>
+        /// JSON.
+        /// </summary>
+        Json                = 3
+    }
+
+
+    /// <summary>
+    /// Extension methods for configuration classes specific to <see cref="OrleansSQLUtils"/>.
+    /// </summary>
+    public static class AdoNetConfigurationExtensions
+    {
+        /// <summary>
+        /// Adds a storage provider of type <see cref="AdoNetStorageProvider"/>.
+        /// </summary>
+        /// <param name="config">The cluster configuration object to add provider to.</param>
+        /// <param name="providerName">The provider name.</param>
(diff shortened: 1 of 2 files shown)


veikkoeeva:
I'll see if there's an efficient way to add tests still. I intend this to be ready in a few hours. This is just to open a PR.

src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// </summary>
+        /// <param name="config">The cluster configuration object to add provider to.</param>
+        /// <param name="providerName">The provider name.</param>
+        /// <param name="connectionString">The ADO.NET storage connection string. If none is provided, it will use the same as in the Globals configuration.</param>
+        /// <param name="serializationFormat">The format in which the data will serialized to the storage.</param>
+        public static void AddAdoNetStorageProvider(
+            this ClusterConfiguration config,
+            string providerName = "AdoNetStorage",
+            string connectionString = null,
+            AdoNetSerializationFormat serializationFormat = AdoNetSerializationFormat.OrleansNativeBinary)
```
jdom: do we really want to have Orleans native as the default? I would really discourage this if there is some alternative that works fine. Orleans serializer is not version tolerant, which is a very bad thing for stuff persisted in durable storage. I think in newer providers (ie: azure blob storage) we even stopped providing Orleans binary serializer as an option.

veikkoeeva: It can be changed. Would XML be a neutral choice?

In the future I personally would hope to remove this "string configuration" and provide actual libraries. The default could be what is included by default, such as XML and some Orleans fallback (de)serializer.

Non-version tolerant isn't good at least. The ADO.NET storage has an interceptor mechanism to handle it, but it's not accessible until DI or interface change since Orleans instantiates the class and the application code can't get a reference to it.

veikkoeeva:
There doesn't seem to be a way to fully test this functionality without doing much more work than I can afford currently. It looks like the other such methods aren't holistically tested either. However, I used the Storage sample to check the values are passed correlty and data is saved correctly.

src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
jdom: You might know better, probably XML is common in DBs, but Json seems acceptable too. Since you and @shayhatsor own this area, I'm fine with whatever choice, it's just that I prefer not the non-version tolerant one, as it can be a source of issues that people get bitten with after they have code (and data) in prod.

jdom:
Seems fine for now. I think that once we move to having testing kits as we talked in https://github.com/dotnet/orleans/issues/2011 we should have a lot more silo and non-silo test coverage for every provider, and that will be an easy chance to use the extension methods too.

src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
veikkoeeva: OK. I'll update withint the next 24 hours. Too tired now. :)

veikkoeeva:
@jdom I made XML as the default serialization choice. I also changed the namespace to the root in case there will be other helper methods besides storage related. I rebased out of habit also, hopefully it's OK.

src/OrleansSQLUtils/Storage/AdoNetConfigurationExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+            string providerName = "AdoNetStorage",
+            string connectionString = null,
+            AdoNetSerializationFormat serializationFormat = AdoNetSerializationFormat.Xml)
+        {
+            if(string.IsNullOrWhiteSpace(providerName))
+            {
+                throw new ArgumentNullException(nameof(providerName));
+            }
+
+            connectionString = GetConnectionString(connectionString, config);
```
attilah: IMO It is not a good practice to use an input parameter and modify it inside the body of a method. Same stands for serializationFormat below.

veikkoeeva: Should it be changed? I believe this codebase (and plenty of code in the wild) modifies default value is if it is in illegate state. I don't see the `serializationFormat` being modified.

As you might be aware, this is practically a copy of `AzureConfigurationExtensions`, likely not the only place in this codebase and there's plenty of code elsewhere that does this too.

attilah: Back in the day that's what I have been thought :-) Touching input parameters by modifying their values is a kind of code smell in managed code, but "where I came from" native code it can lead to unwanted, hard to track consequences, for example if you're modifying a pointer what you received as an input parameter.
I'd avoid at all times. Imagine that you're logging the values what you're calling a method with, then you see an error and you've no clue HOW that value got there since that is not what you passed in. If we "allow" this in the codebase it will make it hard to differentiate on a case by case basis that it is OK here to modify it, but there it can lead to hard to track errors.

jdom: I understand that could be an issue in native code, but the reality is that not here in c# when we are dealing with references to immutable objects/structs.
But to add to it, IMO creating a new variable just so that we don't update the reference can cause more maintainability problems than just reusing it (as code could mistakenly use the input parameter instead of the sanitized one). Also for consistency with everywhere else we should let this in and if you feel strongly (I don't, so I'm open to change) we can open an issue and then update all our usages of input sanitization in one go, but not be prescriptive yet on the non-consistent approach until there's consensus.

attilah: Ok.

veikkoeeva: @attilah Here the comments tell the connection string should be given, but if not, for conveniene a global one will be acquired and used. Partial function application would probably work here, but maybe we should keep the configuration like the other storage configuration utilities?

jdom: Since I think this is settled (at least temporarily, but future changes would go in a different issue anyway), I'll merge, since we want to pull this into 1.3.1

Issue #2287, opened 2016-10-12
Silo should forward request when asked to shutdown


benjaminpetit:
If I'm not mistaken, when a silo properly shutdown (not stop) it should:

1. Finish the requests that are being processed on locally activated grains (requests that were started to process before shutdown was called)
2. Then forward the remaining non processed requests to another silo so they can be completed by someone else

It seems that step 2. is not properly done. In this test https://github.com/benjaminpetit/orleans/commit/53631d95df931e6c416170d92bd6b4caeb0b1f45 the `await promisesAfterShutdown` throws `TimeoutException`

Am I missing something?

gabikliot:
Lets check in this test, with properly marked throwing exception if it indeed fails, and then I will take a deeper look. From a brief look I took now, sounds like you are correct, the 2nd call should have succeeded (`HostedCluster.StopSilo` is actually doing shutdown -> confusing. The grain in non reentrant.)

benjaminpetit:
I moved the test in another branch, removed the functional category tag and sent this PR #2292

gabikliot:
In https://github.com/dotnet/orleans/pull/2292#issuecomment-253379088  @benjaminpetit wrote:

I am not 100% sure, but I think we may shutdown too soon the MessageCenter. When I remove the call to SafeExecute(messageCenter.BlockApplicationMessages) from Silo.Terminate(bool gracefully) the test seems to pass everytime.

The test might also fail because of SiloUnavailableException. Hopefully, this will be fixed soon :)

gabikliot:
Good catch, this is indeed most probably the cause.
The msg is queued for rerouting, but now there is a race between this rerouting and the message center shutting down.

The fix is far from trivial.
You don't want just to remove messageCenter.BlockApplicationMessages, cause it has a purpose by itself. Even if you did, it will only shorten the race but not solve it, since as part of rerouting we may need to do directory lookup (almost sure will do, since the grain does not have an activation). That means we need to wait until the rerouting (which involves async activities, such as directory lookup) to fully finish.
To do that properly we will need to expose those rerouted msg from the SafeExecute(() => catalog.DeactivateAllActivations().WaitWithThrow(stopTimeout));, and wait on a Task associated with their rerouting.
Quite an invasive change, across multiple layers.

[...]

PR #2292, opened 2016-10-12
Adding test to reproduce forward issue

benjaminpetit:
Repro for #2287

### DIFF
--- a/test/Tester/Forwarding/ShutdownSiloTests.cs
+++ b/test/Tester/Forwarding/ShutdownSiloTests.cs
@@ -1,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Orleans.Runtime;
+using Orleans.TestingHost;
+using UnitTests.GrainInterfaces;
+using UnitTests.Tester;
+using Xunit;
+
+namespace Tester.Forwarding
+{
+    public class ShutdownSiloTests : TestClusterPerTest
+    {
+        public const int NumberOfSilos = 2;
+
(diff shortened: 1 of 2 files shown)


benjaminpetit:
I am not 100% sure, but I think we may shutdown too soon the MessageCenter. When I remove the call to `SafeExecute(messageCenter.BlockApplicationMessages)` from `Silo.Terminate(bool gracefully)` the test seems to pass everytime.

The test might also fail because of SiloUnavailableException. Hopefully, this will be fixed soon :)

gabikliot:
Good catch, this is indeed most probably the cause.
The msg is queued for rerouting, but now there is a race between this rerouting and the message center shutting down.

The fix is far from trivial.
You don't want just to remove `messageCenter.BlockApplicationMessages`, cause it has a purpose by itself. Even if you did, it will only shorten the race but not solve it, since as part of rerouting we may need to do directory lookup (almost sure will do, since the grain does not have an activation). That means we need to wait until the rerouting (which involves async activities, such as directory lookup) to fully finish.
To do that properly we will need to expose those rerouted msg from the `SafeExecute(() => catalog.DeactivateAllActivations().WaitWithThrow(stopTimeout));`, and wait on a Task associated with their rerouting.
Quite an invasive change, across multiple layers.

Conceptually, I think this is the right thing to do: to be more strict in the silo about different silo-internal shutdown stages and move between them in lock step (I am not suggesting distributed lock step, only within the silo w.r.t. to silo internal activities). But that will not be easy.
Let me know if you want to embark on that journey and we can brainstorm how exactly.

gabikliot:
I moved the discussion into the issue: https://github.com/dotnet/orleans/issues/2287

PR #2295, opened 2016-10-13
Serialization fixes

ReubenBond:
Fixes for 3 serialization issues:

1. A recent change (by myself) resulted in generated `GrainReference` classes being passed through the fallback serializer.
2. Due to an inadequacy in `TypeUtilities.AreInternalsVisibleTo`, serializers were not being generated for types referencing internals from another assembly. For example, `ActivationInfo` in `OrleansRuntime` would not have a serializer generated because it has a field whose type is internal to `Orleans`. The change adds support for `[assembly: InternalsVisibleTo("xxx")]` where "xxx" is the short name of an assembly (eg, `"OrleansRuntime"`, rather than only supporting fully qualified assembly names.
3. The check to determine if a type is accessible for serialization involves recursively checking all serialized fields in that type. The check was deficient because it did not support fields which had predefined serializers (eg, `List<SomeAccessibleType>`).

I've added a test for types which were not having serializers generated.

Note: the commits are independent except for the last one containing the new test case.

### DIFF
--- a/src/OrleansCodeGenerator/SerializerGenerationManager.cs
+++ b/src/OrleansCodeGenerator/SerializerGenerationManager.cs
@@ -105,13 +105,8 @@ namespace Orleans.CodeGenerator
             if (TypeUtils.HasAllSerializationMethods(t)) return false;
 
             // This check is here and not within TypeUtilities.IsTypeIsInaccessibleForSerialization() to prevent potential infinite recursions 
-            var skipSerialzerGeneration = t.GetAllFields()
-                .Any(
-                    field => field.GetCustomAttribute<NonSerializedAttribute>() == null &&
-                        TypeUtilities.IsTypeIsInaccessibleForSerialization(
-                            field.FieldType,
-                            module,
-                            targetAssembly));
+            var skipSerialzerGeneration =
+                t.GetAllFields().Any(field => IsFieldInaccessibleForSerialization(module, targetAssembly, field));
             if (skipSerialzerGeneration)
             {
                 return false;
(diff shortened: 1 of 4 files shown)


src/OrleansCodeGenerator/SerializerGenerationManager.cs
```diff
@@ -121,6 +116,13 @@ internal bool RecordTypeToGenerate(Type t, Module module, Assembly targetAssembl
             return true;
         }
 
+        private static bool IsFieldInaccessibleForSerialization(Module module, Assembly targetAssembly, FieldInfo field)
+        {
+            return field.GetCustomAttribute<NonSerializedAttribute>() == null
+                   && !SerializationManager.HasSerializer(field.FieldType)
```
jdom: do you think that order can affect us? If the serializer was manually crafted and registered in this assembly, would it already be in SerializationManager by the time codegen runs?
I assume it is, but just checking.

ReubenBond: Good question.

* For compile-time codegen, we explicitly call `RegisterBuiltInSerializers` before generating code, so those serializers will be included.
* This will only ever bring in additional serializers, so we don't need to worry that it will result in *less* serializers than what we previously had.
* For run-time assembly, codegen runs *before* `FindSerializationInfo`, so serializers in the same assembly will not be included here. That's a deficiency.

We could run `FindSerializationInfo` before running codegen, that way the serializers are loaded before we rescan the assembly and generate types which don't have serializers yet.

PR #2297, opened 2016-10-13
Updates the Nuget icons to the new one

veikkoeeva:
It looks like two package icons are not updated to
the new one. Also, one image in OrleansContrib,
Microsoft.Orleans.ServiceFabric.Client, still
uses the old one.

See at https://www.nuget.org/packages?q=microsoft.orleans.*.


veikkoeeva:
These might be left like that on purpose. In that case let's just close this issue.

ReubenBond:
Thanks, @veikkoeeva! I didn't change the icon yet, but I updated my OrleansContrib Fabric packages to point to the new Fabric packages :)

ReubenBond:
We could also update the logo in at https://raw.githubusercontent.com/jthelin/jthelin.github.io/master/images/Orleans-nuget.png so that the packages automatically point to it

jthelin:
Thanks for the heads-up @jdom 

The legacy copy of the icon is updated now.
https://www.nuget.org/packages?q=microsoft.orleans.*

PR #2298, opened 2016-10-13
In client, when a gateway connection close reroute not yet sent message to another gateway

benjaminpetit:
This is a fix for #2283. And it come with partial fix for #2282 for free* (it will still throw when occurring between two silos)


benjaminpetit:
@dotnet-bot test this please

src/Orleans/Messaging/GatewayConnection.cs
```diff
@@ -66,7 +67,6 @@ public override void Stop()
             IsLive = false;
             receiver.Stop();
             base.Stop();
-            RuntimeClient.Current.BreakOutstandingMessagesToDeadSilo(Silo);
```
gabikliot: lets please not remove that for now. I am still not convinced that is the right things to do, and as we said: this is unrelated to saving the msgs in the queue.
Once the queue resend is submitted, you can open a new PR with this line removed only and test and we can discuss if this is needed or not.

I am strongly opposing on bundling independent fixes/changes.

src/Orleans/Messaging/GatewayConnection.cs
```diff
@@ -176,6 +176,17 @@ public void Connect()
             }
         }
 
+        protected override void DrainQueueOnStop(IEnumerable<Message> requests)
```
gabikliot: Hmm...., instead of doing the protected msg hack, why not just add a method on the AsynchQueueAgent: DrainQueue and call it from within GatewayConnection.Stop, before calling base.Stop.

gabikliot: And in this method, instead of doing requestQueue.GetConsumingEnumerable() just loop over  the queue, cal Take one at a time, and resend. If AsynchQueueAgent does not have MsgCenter, just pass it as an arg.

test/Tester/MembershipTests/SilosStopTests.cs
```diff
@@ -50,26 +50,5 @@ public override TestCluster CreateTestCluster()
                 Assert.Equal(typeof(SiloUnavailableException), ex.GetBaseException().GetType());
             }
         }
-
-        [Fact, TestCategory("Functional"), TestCategory("Liveness")]
-        public async Task SiloUngracefulShutdown_ClientOutstandingRequestsBreak()
```
gabikliot: why did you remove this test?

gabikliot: Ohh, I see, because of SiloUnavailableException.
As we said above, lets put the SiloUnavailableException back and handle it separately.

benjaminpetit:
@dotnet-bot test this please

benjaminpetit:
@dotnet-bot test this please

src/Orleans/Runtime/AsynchQueueAgent.cs
```diff
@@ -111,6 +112,23 @@ public override void Stop()
...
+            while (true)
+            {
+                T request;
+                try
+                {
+                    request = requestQueue.Take();
+                }
+                catch (InvalidOperationException)
+                {
+                    break; // Queue is empty
```
gabikliot: are you using that style to detect an empty collection since using Count may be very in-efficient? 
maybe better to use TryTake instead?
catching `InvalidOperationException` just does not sounds like a robust pattern.

benjaminpetit: Yes I guess it should be better to just use TryTake. But in this case, should I check that `Stop()` was called before?

benjaminpetit:
Since I put back the SiloUnavailableException when the request was sent but not answered, I don't know how to test this. 'SiloGracefulShutdown_NoFailureOnGatewayShutdown' obviously fail because of that.

gabikliot:
Re the test: you can use either a "watch grain" pattern or rely on memory storage provider.
The tested grain will either increment a counter in the memory storage provider or send a msg to a watch grain. In either case you need to make sure the watch grain or the provider grain are on the silo that is not shutdown, so their state sticks.
And then at the end of the test you query the watch grain or storage provider (via the grain) about the state and make sure the call has happened.
This is at a high level. Jason can show you the details.

gabikliot:
I like the fix, about 10 lines. That's how it should be.
Lets just add the test and then can merge.

benjaminpetit:
Not sure if I understood the proposed test? Make calls to a grain that increment a counter, ignoring possible SiloUnaivailableException that might occurs, and check the value at the end?

gabikliot:
There is a general pattern of how to do those tests: watcher grain.
The tested grainreports all its activities to the watcher grain and you ask the watcher grain from the test about the progress of the tested grain.

In your case, lets say tested grain is A and watcher is B.
Send x to A and A sends it to B. await the call to A in the test.
Send y to A and A sends it to B. await the call in the test. it may throw in your case , ignore.
Ask B if y happened. In your case, it may not happen yet. So instead: ask B if y happened and if not B will resolve the grain call promise WHEN y happens (useTCS for that).

We have examples of that in your tests.

benjaminpetit:
I tried your suggestion, I tried other tests, nothing seems to work. If someone want to take this PR and add a test, feel free to do so, I can't find a valid test. I think we might loses some message when the GW is shutting down.

test/Tester/Forwarding/ShutdownSiloTests.cs
```diff
@@ -43,6 +48,76 @@ public override TestCluster CreateTestCluster()
...
+            const int numberOfGrains = 2 * NumberOfSilos;
+
+            var grains = new List<ICounterGrain>();
+            for (var i = 0; i < numberOfGrains; i++)
+            {
+                grains.Add(await GetCounterGrainOnPrimary());
+            }
+
+            var cancellationTokenSource = new CancellationTokenSource();
+            var promise = CallIncrement(grains, cancellationTokenSource.Token);
```
sergeybykov: Isn't it an issue that `CallIncrement` is called on the main thread instead of on a different one (using `Task.Run()` or something), so that the while loop would be running independently from the main thread?

test/Tester/Forwarding/ShutdownSiloTests.cs
```diff
@@ -43,6 +48,76 @@ public override TestCluster CreateTestCluster()
...
+        {
+            var counter = 0;
+            var tasks = new List<Task>();
+            while (!token.IsCancellationRequested)
+            {
+                foreach (var g in grains)
+                {
+                    tasks.Add(g.IncrementValue());
+                }
+                counter++;
```
sergeybykov: Won't the counter always be equal to `tasks.Count`?

benjaminpetit: `counter == tasks.Count / grains.Count`. But I kept the counter so it is easier to understand

gabikliot:
I am OK with merging this PR as is, even without the test.
I have high confidence that the new code does not do any damage, and most probably does save some messages, so it is only goodness.

As for the test:
1) I would not use cancellation token at all for those kind of tests. since we don't really have any usage of them by anyone (at least I am not aware of), I would trust them, **yet**.
2) the whole testing efforts goes back to the  question I asked initially about the overall goal and my suggestion about testing this: https://github.com/dotnet/orleans/issues/2287#issuecomment-253719306

So my proposal is to merge this (since there is no downside), and then work on further testing if that is important. I can guarantee you we are not 100% loss free in shutdown cases.

PR #2299, opened 2016-10-13
ClientInvokeCallback tests should always reset ClientInvokeCallback

ReubenBond:
We are seeing failures related to this in functional test runs recently.
One of the test cases is not resetting ClientInvokeCallback.

This PR wraps each of those tests in try/finally blocks to reset `GrainClient.ClientInvokeCallback` and also resets it in the classes `Dispose()` method.

### DIFF
--- a/test/TesterInternal/General/RequestContextTest.cs
+++ b/test/TesterInternal/General/RequestContextTest.cs
@@ -422,6 +422,7 @@ namespace UnitTests.General
         public void Dispose()
         {
             Trace.CorrelationManager.ActivityId = Guid.Empty;
+            GrainClient.ClientInvokeCallback = null;
             RequestContext.Clear();
         }
 
--- a/src/Orleans/Core/GrainClient.cs
+++ b/src/Orleans/Core/GrainClient.cs
@@ -312,6 +312,7 @@ namespace Orleans
             }
             outsideRuntimeClient = null;
             grainFactory = null;
+            ClientInvokeCallback = null;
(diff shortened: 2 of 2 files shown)


test/TesterInternal/General/RequestContextTest.cs
```diff
@@ -741,58 +742,84 @@ public void Dispose()
...
+                Assert.Equal(1, callback.TotalCalls); // "Number of callbacks"
+
+                GrainClient.ClientInvokeCallback = null;
+                activityId = await grain.E2EActivityId();
+                Assert.Equal(Guid.Empty, activityId); // "E2EActivityId Call#2"
+                Assert.Equal(1, callback.TotalCalls); // "Number of callbacks - should be unchanged"
+            }
+            finally
+            {
+                GrainClient.ClientInvokeCallback = null;
```
ReubenBond: I wasn't certain if it was - I'll revert those bits.

Clean.cmd
```diff
@@ -0,0 +1,5 @@
+@ECHO .
+@ECHO Cleaning %CD%
+@ECHO .
+FOR /F "tokens=*" %%G IN ('DIR /B /AD /S bin') DO RMDIR /S /Q "%%G"
+FOR /F "tokens=*" %%G IN ('DIR /B /AD /S obj') DO RMDIR /S /Q "%%G"
```
jdom: unrelated change?

PR #2300, opened 2016-10-14
Fix vNext solution build break

jdom:


dVakulen:
Sorry about that, should've try to build vNext solution first.

jdom:
No worries. Unfortunately we are having issues integrating that vNext solution in our CI... that would help catch these issues much earlier

Issue #2289, opened 2016-10-12
When ADO.NET library is failed to load, the error should be clearer


veikkoeeva:
Kenneth [reported in Gitter](https://gitter.im/dotnet/orleans?at=57fdec54a7c618cd2603ebd2) a vague error message, which basically was
```
at Orleans.SqlUtils.DbConnectionFactory.GetFactory(String invariantName)
at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
at Orleans.SqlUtils.DbConnectionFactory.CreateConnection(String invariantName, String connectionString)
at Orleans.SqlUtils.RelationalStorage.<ExecuteAsync>d__14`1.MoveNext()
```
Upon further investigation it appers the reason was the ADO.NET connector library failed to load (i.e. was not found in this case). The code is [here](https://github.com/dotnet/orleans/blob/8d6153f46763a3ec97db483142ce17b9d90dc9c9/src/OrleansSQLUtils/Storage/DbConnectionFactory.cs#L19)). It apperas to be this error message should be clearer and perhaps give direct instructions on where to acquire libraries  (e.g. https://www.nuget.org/packages/MySql.Data/) and or point to other instruction (such as at http://www.telerik.com/blogs/3-ways-to-use-the-sqlite-nuget-with-telerik-data-access).

attilah:
With the fix in place, .Net will throw a usable ConfigurationException with the message: "Failed to find or load the registered .Net Framework Data Provider.".

PR #2303, opened 2016-10-14
Fix for #2289

attilah:


jdom:
LGTM, but I'll let @veikkoeeva or @shayhatsor validate too.
Note to merger (in case it's not me): notice that the commit message is not descriptive, please use squash and rebase option to fix the commit message.

src/OrleansSQLUtils/Storage/DbConnectionFactory.cs
```diff
@@ -13,10 +14,21 @@ internal static class DbConnectionFactory
...
+            }
+
             // Seeks for database provider factory classes from GAC or as indicated by
             // the configuration file, see at <see href="https://msdn.microsoft.com/en-us/library/dd0w4a2z%28v=vs.110%29.aspx">Obtaining a DbProviderFactory</see>.       
 
             DataRow factoryData = DbProviderFactories.GetFactoryClasses().Rows.Find(invariantName);
+
+            if (factoryData == null)
+            {
+                throw new InvalidOperationException($"Can't find database provider factory with '{invariantName}' invariant name. Please check the configuration file to see if the provider is configured correctly or the binaries/assemblies are deployed with the application.");
```
veikkoeeva: Maybe change the part of "binaries/assemblies" to "ADO.NET libraries" as it is more specific about the exact issue.

src/OrleansSQLUtils/Storage/DbConnectionFactory.cs
```diff
@@ -13,10 +14,21 @@ internal static class DbConnectionFactory
 
         private static CachedFactory GetFactory(string invariantName)
         {
+            if (String.IsNullOrWhiteSpace(invariantName))
```
veikkoeeva: More of a discussion point, I see the reasoning using `System.String` instead of the alias when referencing the class functions. Maybe there is something to refactor in this regard if we want to be uniform.

veikkoeeva:
I'm OK with this change otherwise than that history could look nicer (i.e. the commit message). 👍

src/OrleansSQLUtils/Storage/DbConnectionFactory.cs
shayhatsor: I agree, also it should be "application configuration file" instead of "configuration file"

jdom: Oh, I'm sorry, I read your I'm OK with the change, and immediately merged after it, and did not see this comment you added just before I merged

src/OrleansSQLUtils/Storage/DbConnectionFactory.cs
attilah: I agree, we'll definitely need to make a "style changes only" PR ;-) and make everything conform to a standard and have a resharper file also tied to the solution!

PR #2304, opened 2016-10-15
clean up dependency in test projects

xiazen:
- remove Orleans.NonSiloTests dependency on Tester
- remove TesterInternal dependency on Orleans.NonSiloTests
- remove ununsed reference in test projects


test/TesterInternal/TesterInternal.csproj
```diff
@@ -39,22 +39,17 @@
...
-    <Reference Include="System.Transactions" />
     <Reference Include="System.Web.Extensions" />
-    <Reference Include="System.XML" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="ActivationsLifeCycleTests\ActivationCollectorTests.cs" />
     <Compile Include="ActivationsLifeCycleTests\DeactivateOnIdleTests.cs" />
     <Compile Include="CollectionFixtures.cs" />
     <Compile Include="ConnectionStringFixture.cs" />
+    <Compile Include="ErrorInjectionStorageProvider.cs" />
```
xiazen: moved it back to TesterInternal since usage for this class all came from TesterInternal

jdom: I'd like to move MockStorageProvider to TesterInternal too. This mock uses stuff that we want to deprecate as it will not work in .net core (MarshalByRefObject), and it could be tested differently. NonSiloTests doesn't need such a sophisticated mock in order to test the configuration, nevertheless I'm fine with this change temporarily, If you don't, I'll probably remove it soon

---
In reply to: [83520658](https://github.com/dotnet/orleans/pull/2304#discussion_r83520658) [](ancestors = 83520658)

attilah: I also moved that too, I was needed that for my work.

PR #2305, opened 2016-10-15
Removes InconsistentStateException from parallel tests

veikkoeeva:
In ADO.NET tests when parallel test execution became possible,
it was possible for the the hash test grain IDs and types
collide with the ones used in testing 100 parallel insertions
and reads.


jdom:
Thanks @veikkoeeva for the fast turnaround!

PR #2312, opened 2016-10-17
Remove delegate allocation

dVakulen:
On each <code>Catalog</code> local lookup the delegate was allocated.

### DIFF
--- a/src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
+++ b/src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
@@ -19,6 +19,8 @@ namespace Orleans.Runtime.GrainDirectory
         /// list of silo members sorted by the hash value of their address
         /// </summary>
         private readonly List<SiloAddress> membershipRingList;
+
+        private readonly Func<SiloAddress, int, bool, bool> isSiloNextInTheRing; 
         private readonly HashSet<SiloAddress> membershipCache;
         private readonly AsynchAgent maintainer;
         private readonly Logger log;
@@ -86,12 +88,14 @@ namespace Orleans.Runtime.GrainDirectory
         internal readonly CounterStatistic UnregistrationsManyRemoteSent;
         internal readonly CounterStatistic UnregistrationsManyRemoteReceived;
 
-        
         public LocalGrainDirectory(Silo silo)
         {
             log = LogManager.GetLogger("Orleans.GrainDirectory.LocalGrainDirectory");
 
             MyAddress = silo.LocalMessageCenter.MyAddress;
+            isSiloNextInTheRing = (siloAddr, hash, excludeMySelf) => (siloAddr.GetConsistentHashCode() <= hash) &&
+                                                                     (!excludeMySelf || !siloAddr.Equals(MyAddress));
+
             Scheduler = silo.LocalScheduler;
             membershipRingList = new List<SiloAddress>();
             membershipCache = new HashSet<SiloAddress>();
@@ -460,8 +464,12 @@ namespace Orleans.Runtime.GrainDirectory
                 return Seed;
             }
 
-            SiloAddress siloAddress;
+            SiloAddress siloAddress = null;
             int hash = unchecked((int)grainId.GetUniformHashCode());
+
+
+            // excludeMySelf from being a TargetSilo if we're not running and the excludeThisSIloIfStopping flag is true. see the comment in the Stop method.
+            bool excludeMySelf = !Running && excludeThisSiloIfStopping;
 
             lock (membershipCache)
             {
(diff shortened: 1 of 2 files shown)


src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -471,12 +479,17 @@ public SiloAddress CalculateTargetSilo(GrainId grainId, bool excludeThisSiloIfSt
                     return excludeThisSiloIfStopping && !Running ? null : MyAddress;
                 }
 
-                // excludeMySelf from being a TargetSilo if we're not running and the excludeThisSIloIfStopping flag is true. see the comment in the Stop method.
-                bool excludeMySelf = !Running && excludeThisSiloIfStopping; 
-
                 // need to implement a binary search, but for now simply traverse the list of silos sorted by their hashes
```
dVakulen: The binary search will provide benefit only on relatively large deployments (with silos count > 75). Does it makes sense to implement this?

src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -86,12 +88,14 @@ internal class LocalGrainDirectory :
...
         internal readonly CounterStatistic UnregistrationsManyRemoteReceived;
 
-        
         public LocalGrainDirectory(Silo silo)
         {
             log = LogManager.GetLogger("Orleans.GrainDirectory.LocalGrainDirectory");
 
             MyAddress = silo.LocalMessageCenter.MyAddress;
+            isSiloNextInTheRing = (siloAddr, hash, excludeMySelf) => (siloAddr.GetConsistentHashCode() <= hash) &&
+                                                                     (!excludeMySelf || !siloAddr.Equals(MyAddress));
```
jdom: Since this is not using any closure, what is the purpose of having a delegate field instead of a (more readable) proper method?

dVakulen: It's still capturing <code>MyAddress</code>.

jdom: I'm confused? is it really capturing the address. I would expect this code to capture `this` and then accessing the `MyAddress` field, which is no different to calling a proper instance method.

ReubenBond: image

You're right, @jdom. It's a good idea to remove the allocation here, but unless there's something wrong with `membershipRingList.FindLast`, that call can be retained.

Instead, I would suggest creating a `private readonly Predicate<SiloAddress>` field which captures `this` once and passing that to the call here. That way you have no per-call allocations.

jdom: There's 2 things here. Since this is inlining FindLast, then there is not even a need to remove the delegate allocation, as there is no longer a closure being passed to another higher order function.
But even if we decided to continue using FindLast and remove the inlining, what I don't understand is the need for a function field that captures `this`, since that is exactly what an instance method is. So just define the function as a method body instead of in a field that is later assigned.

ReubenBond: @jdom I agree with the first point.
For the second, when you pass an instance method as a delegate, like
```C#
membershipRingList.FindLast(this.SomePredicateMethod)
```
you are allocating a new `Predicate<T>` every time.

dVakulen: Indeed, fixed.

src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -1088,6 +1098,11 @@ internal IRemoteGrainDirectory GetDirectoryReference(SiloAddress silo)
             return InsideRuntimeClient.Current.InternalGrainFactory.GetSystemTarget<IRemoteGrainDirectory>(Constants.DirectoryServiceId, silo);
         }
 
+        private bool IsSiloNextInTheRing(SiloAddress siloAddr, int hash, bool excludeMySelf)
```
jason-bragg: This logic is same as other logic, except with <= rather than >=.
Please consider using one static function with both hashes in arguments, with the hash order changed, rather then two functions with the comparison reversed.

jason-bragg:
In general, I've no objection to this change, but I don't really see much value in it ether.  Unless I'm misunderstanding something, this call path is not nearly hot enough to merit this sort of micro-optimization.  I could well be wrong though.

src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
dVakulen: I think that the resulting logic for this one-liner function will be more complex than now. If this is desired - I can change it.

dVakulen:
@jason-bragg This allocation happens on almost every request message send. On the other hand GC induced by Orleans doesn't takes much CPU, so I don't expect any effect from one or two removed allocations. But in perspective the number of allocs on hot paths could be reduced to noticeable extent.

jason-bragg:
@dVakulen,  Agreed.  I don't expect perf gain from this, but removing allocations and link expressions (notoriously slow) from the hot path can't hurt.

LGTM

sergeybykov:
Thank you, @dVakulen for pushing the throughput numbers higher with nearly every PR!

dVakulen:
@sergeybykov Just curious - is there was any change after https://github.com/dotnet/orleans/pull/2275?

ReubenBond:
@dVakulen there was an approx 6% increase by my reckoning. It's hard to be sure that it was caused by #2275 specifically, particularly because #2295 went in at about the same time.

Issue #2309, opened 2016-10-17
Dependency injection does not find startup class (Orleans V1.2.3)


peter-perot:
I'm using Orleans since March 2015 in production. Those days we used an ugly service locator to resolve dependencies from our grains.

Since DI in Orleans has matured over time I now want to get rid of that service locator. Last week I have been playing around with ASP.NET Core DI and came to the conclusion that it will suffice our requirements (formerly we used Microsoft Unity DI).

I have put all my bootstrapping code into the startup class and tweaked my server-side XML configuration file as described here: https://dotnet.github.io/orleans/Documentation/Advanced-Concepts/Dependency-Injection.html

However: The `ConfigureServices` methods never gets called. Even if I put some intentionally misspelled full qualified class name into the `<Startup>` tag nothing happens (no error message or something like that).

Today I started a simple HelloWorld project from scratch using Orleans 1.2.3 (this is the version our production uses) with the console application silo host trying to get DI to work, but it still does not work: the bootstrapping code is not executed.

Since I have no clue what is missing I started a question at Stackoverflow (http://stackoverflow.com/questions/40046319/dependency-injection-with-orleans-version-1-2-3-not-working), but no answers until now.

Anyone any hint for me?

peter-perot:
Updating to Orleans 1.3.0 did not solve the problem.

My config file:

```xml
<?xml version="1.0" encoding="utf-8"?>
<OrleansConfiguration xmlns="urn:orleans">
  <Globals>
    <Startup Type="Grains.Startup, Grains" />
    <StorageProviders>
      <Provider Type="Orleans.Storage.MemoryStorage" Name="MemoryStore" />
      <!-- To use Azure storage, uncomment one of the following lines: -->
      <!--<Provider Type="Orleans.Storage.AzureTableStorage" Name="AzureStore" DataConnectionString="UseDevelopmentStorage=true" />-->
      <!--<Provider Type="Orleans.Storage.AzureTableStorage" Name="AzureStore" DataConnectionString="DefaultEndpointsProtocol=https;AccountName=MYACCOUNTNAME;AccountKey=MYACCOUNTKEY" />-->
    </StorageProviders>
    <SeedNode Address="localhost" Port="22222" />
  </Globals>
  <Defaults>
    <Networking Address="localhost" Port="22222" />
    <ProxyingGateway Address="localhost" Port="40000" />
[...]

PR #2313, opened 2016-10-18
Issue #2309: The 'Startup' tag must be placed into the 'Defaults' section

peter-perot:
This PR corrects a mistake in the docs. See https://github.com/dotnet/orleans/issues/2309.

### DIFF
--- a/src/Documentation/Advanced-Concepts/Dependency-Injection.md
+++ b/src/Documentation/Advanced-Concepts/Dependency-Injection.md
@@ -45,14 +45,14 @@ configuration.UseStartupType<MyApplication.Configuration.MyStartup>();
 
 ## Configuring via XML
 
-To register your `Startup` class with Orleans you've to add a `Startup` element to the `Globals` section and in the `Type` attribute you've to specify the assembly qualified name for the type.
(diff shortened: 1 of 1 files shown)


ReubenBond:
Thanks, LGTM!

I wonder if you would make one more adjustment & fix the `OrleansConfiguration.xsd` file so that it includes the `<Startup>` section under the `NodeConfiguration` complexType.

That's not on this branch, so I'll merge this anyway.

peter-perot:
@ReubenBond: What's the correct branch for the XSD fix: `master` or `1.3.0`?

Anyway: Is the `Startup` tag only allowed in the `Defaults` section, or can one configure a different startup per node such that the tag is also allowed in the `Override` section?

ReubenBond:
@peter-perot `master`. `Startup` is defined in `NodeConfiguration`, so It ought to be accessible in both `Defaults` and `Override`.

peter-perot:
XSD has been adjusted. See PR #2314.

sergeybykov:
@peter-perot For some reason there's a merge conflict in #2314.

PR #2318, opened 2016-10-19
IL-based serializer improvements

ReubenBond:
Hopefully the way these commits are split out makes it fairly easy to follow what's going on.

* Efficient enum serialization (statically selects the right integral type & directly encodes `stream.Write` & `Enum.ToObject` calls)
* Direct (non-boxing) serialization of simple types 
* Improvements to emitted IL:
  * Select between `call` & `callvirt` OpCode depending on method
  * Select between `Ldsfld` & `Ldfld` depending on whether a field is static or non-static
  * Use optimal OpCodes where applicable. Eg: `Ldarg_0` with no `int` value rather than `Ldarg` + `0`
* Cache result of `Type.GetTypeFromHandle` in static fields for generated IL rather than calling that method every time a `Type` is needed.
* (more) Compact representation of types
* Misc fixes like consolidating & renaming classes + reducing allocations

Here're the perf numbers before & after this change using the `FallbackFullLoop` benchmark method.

**Before:**

        Serializer |      Median |    StdDev | Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
------------------ |------------ |---------- |------ |------ |------ |------------------- |
           Default | 284.7841 us | 9.0201 us | 36.00 |     - |     - |          34,755.93 |
 ILBasedSerializer |  50.5088 us | 1.5490 us |  1.38 |     - |     - |           3,295.73 |

**After: 3x quicker and allocates 1/3 as much memory**

        Serializer |      Median |    StdDev | Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
------------------ |------------ |---------- |------ |------ |------ |------------------- |
 ILBasedSerializer |  17.3400 us | 0.3762 us | 18.00 |     - |     - |           1,310.49 |

## Further work which could be done (but maybe isn't worthwhile yet):

* In-place deserialization: instead of DeserializeInner returning an `object`, make it return `void` & take a `ref` to the destination. Avoids copying & boxing.
* Avoid boxing by making `SerializeInner` generic.
* Improve type serialization by giving commonly used types well-known short identifiers instead of serializing type names.


src/Orleans/Serialization/IlBasedSerializerGenerator.cs
```diff
@@ -1,217 +0,0 @@
-namespace Orleans.Serialization
```
ReubenBond: This file was renamed to ILBasedSerializerGenerator and if you look at the individual commits in git then you'll see that it's renamed and changed, not deleted and added under a different name. For some reason GitHub displays it differently. Sorry :(

ReubenBond:
Functional tests pass

src/Orleans/Serialization/ILSerializerGenerator.cs
```diff
@@ -14,12 +16,54 @@ internal class ILSerializerGenerator
...
         private readonly SerializationManager.DeepCopier immutableTypeCopier = obj => obj;
 
         private readonly ILFieldBuilder fieldBuilder = new ILFieldBuilder();
+        
+        public ILSerializerGenerator()
+        {
+            this.directSerializers = new Dictionary<RuntimeTypeHandle, SimpleTypeSerializer>
+            {
+                [typeof(int).TypeHandle] = new SimpleTypeSerializer(w => w.Write(default(int)), r => r.ReadInt()),
+                [typeof(uint).TypeHandle] = new SimpleTypeSerializer(w => w.Write(default(uint)), r => r.ReadUInt()),
```
jdom: aren't primitive types already covered by the built-in types in the SerializationManager? What's the reason for these in the fallback serializer?

vNext/src/Orleans/project.json
```diff
@@ -7,22 +7,23 @@
...
     "System.Runtime.Serialization.Formatters": "4.0.0-rc3-24212-01",
     "System.Runtime.Serialization.Primitives": "4.1.1",
-    "System.Reflection.Emit.Lightweight": "4.0.1",
-    "System.Diagnostics.FileVersionInfo": "4.0.0",
-    "System.Threading.ThreadPool": "4.0.10"
+    "System.Security.Cryptography.Algorithms": "4.2.0",
+    "System.Threading.Thread": "4.0.0",
+    "System.Threading.ThreadPool": "4.0.10",
+    "System.Xml.XmlDocument": "4.0.1",
+    "System.Xml.XPath.XmlDocument": "4.0.1"
```
jdom: All Orleans dependencies should be copied to the Tester project too (since because of the csproj hack, transitive package dependencies are currently not working).

ReubenBond:
@jdom yep, the perf numbers are in the original post. It's 3x faster and allocates 1/3 as much memory. Pushed fixes - Tester's project.json should be pulling the right deps now.

vNext/src/Orleans/project.json
ReubenBond: fixed in latest

galvesribeiro:
Beautiful black magic! 👍 
Great work! It's the first time I see someone that play with IL and don't make it looks like mandarin to me!

PR #2322, opened 2016-10-19
move over BasicActivationTests

xiazen:
moved TestUtils in Tester to TestExtention , renamed it TesterUtils.

PR #2323, opened 2016-10-19
Move tests in Tester which uses HostedTestClusterEnsureDefaultStarted to vNext

xiazen:
This lives on top of #2322 . will need to wait for #2322 to be pushed and then rebase. 

Move tests in Tester which uses HostedTestClusterEnsureDefaultStarted to vNext Tester project. 

Some tests test persistence, so we will need to port over OrleansProviders project soon.


xiazen:
commit one is equal to #2322 

so just look at commit 2

xiazen:
rebased with latest. looking into failed activation tests :)

PR #2325, opened 2016-10-20
Cleanup of tests

jdom:
Several cleanup tasks and fix up of references.
All in separate commits for easy reviewing (as reviewing the entire set is probably a lot harder).
There's one specifically large and uninteresting that you can skip reviewing (the namespace move).
Please do not squash changes into 1 commit and use the normal merge mode (or rebase at least).


test/TestExtensions/DefaultClusterFixture.cs
```diff
@@ -8,8 +8,11 @@ public class DefaultClusterFixture : BaseTestClusterFixture
         protected override TestCluster CreateTestCluster()
         {
             var options = new TestClusterOptions();
+#if !NETSTANDARD_TODO
+            // waiting for OrleansProviders project to be ported over to vNext
             options.ClusterConfiguration.AddMemoryStorageProvider("Default");
             options.ClusterConfiguration.AddMemoryStorageProvider("MemoryStore");
+#endif
```
jdom: @xiazen  included this change from PR #2322 to slightly reduce your merge-conflict-required fixes.

xiazen: I'm not sure git is smart emough to recognise this . this may still appear as different diff to it. We shouldn't need to do this in the future. The merge conflict is not hard to deal with.

jdom: well, the point is that #if is required to get it to build. This comment was just a heads up to know how to resolve conflicts, not to let git be smart about it and autofix it (otherwise there would be  no need for a github comment :P )

jdom: heh, actually git was smart about it :)
Anyway, pushed a rebased version now that your PR was merged.
Also included the missing collection fixture, otherwise tests that used `HostedTestClusterEnsureDefaultStarted` were not starting the silo cluster

test/TestExtensions/OrleansTestingBase.cs
```diff
@@ -0,0 +1,20 @@
...
+        public Logger logger => GrainClient.Logger;
+
+        protected static IGrainFactory GrainFactory => GrainClient.GrainFactory;
+
+        protected static long GetRandomGrainId()
+        {
+            return random.Next();
+        }
+    }
+}
```
xiazen: why do you removed a lot of comment and method `TestSilosStarted` from this class when you moved it to TestExtensions?

test/TesterInternal/General/GrainPlacementTests.cs
```diff
@@ -290,5 +290,20 @@ private int ActivationCount(IPlacementTestGrain grain, int sampleSize)
             output.WriteLine("PreferLocalPlacement grain is now located on silo {0}", newActual);
             Assert.Equal(expected, newActual);  // "PreferLocalPlacement strategy should not move activations when other non-hosting silo fails."
         }
+
+        private void TestSilosStarted(int expected)
```
xiazen: Ah I see you moved it here. But why? ElasticPlacementTests use this method too?

test/TestExtensions/OrleansTestingBase.cs
jdom: because it was not generic, it was just used in a test (for no particular reason) but it's not logic that we want to put into this simplified project (and would brin unnecessary dependencies with it)

test/TesterInternal/HostedTestClusterBase.cs
```diff
@@ -27,33 +27,4 @@ public virtual void Dispose()
...
     }
-
-    public abstract class HostedTestClusterPerTest : OrleansTestingBase, IDisposable
-    {
-        protected TestCluster HostedCluster { get; private set; }
-
-        public HostedTestClusterPerTest()
-        {
-            TestDefaultConfiguration.InitializeDefaults();
-
```
xiazen: this is the one difference between HosterTestClusterPerTest and TestClusterPerTest 

TestDefaultConfiguration.InitializeDefaults(); is never called in TestClusterPerTest 

since its static ```
method 
        static TestClusterPerTest()
        {
            TestDefaultConfiguration.InitializeDefaults();
        }
```
Now we are not initializing this default configuration for tests using HostedTestClusterPerTest by delete it and replace it 

Maybe some explanation will help

File: test/TesterInternal/StreamingTests/HaloStreamSubscribeTests.cs [hunk 5]
```diff
@@ -72,7 +73,7 @@ public void Dispose()
             }
         }
 
-        [Fact, TestCategory("Functional"), TestCategory("Streaming"), TestCategory("Halo")]
```
xiazen: why remove Streaming tag?

test/TesterInternal/StreamingTests/StreamLifecycleTests.cs
```diff
@@ -63,31 +63,31 @@ public override void Dispose()
...
         }
 
-        [Fact, TestCategory("Functional"), TestCategory("Streaming"), TestCategory("Cleanup")]
+        [Fact, TestCategory("Functional")]
         public async Task StreamCleanup_BadDeactivate()
         {
             await DoStreamCleanupTest_Deactivate(true, false);
         }
 
-        [Fact, TestCategory("Functional"), TestCategory("Streaming"), TestCategory("Cleanup")]
```
xiazen: same here. and also why remove cleanup tag ?

test/TesterInternal/StreamingTests/StreamLimitTests.cs
```diff
@@ -55,21 +53,20 @@ public override TestCluster CreateTestCluster()
...
 
-
         public StreamLimitTests(ITestOutputHelper output)
         {
             this.output = output;
             StreamNamespace = StreamTestsConstants.StreamLifecycleTestsNamespace;
             mgmtGrain = GrainClient.GrainFactory.GetGrain<IManagementGrain>(0);
         }
         
-        [Fact, TestCategory("Streaming"), TestCategory("Limits")]
```
xiazen: same here why remove streaming and limits tag? especially why remove streaming tag?

test/TesterInternal/StreamingTests/StreamLimitTests.cs
```diff
@@ -202,10 +199,10 @@ public StreamLimitTests(ITestOutputHelper output)
 
             // 1 Stream, Max Producers, 1 Consumer
             await Test_Stream_Limits(
-                StreamTestsConstants.SMS_STREAM_PROVIDER_NAME, 
+                SmsStreamProviderName, 
                 1, 1, MaxConsumersPerStream, useFanOut: true);
         }
-        [Fact, TestCategory("Streaming"), TestCategory("Limits")]
+        [Fact]
```
xiazen: hmmm this test doesn't belong to any catalog now? and same question to others.

test/TesterInternal/HostedTestClusterBase.cs
jdom: yes, I'm aware of this difference. It was like that in HostedTestClusterPerTest before but we updated it. The duplicate copy was not updated by mistake I suppose. Regardless, the current approach is correct, as we wanted to only call this once far all fixtures.

xiazen: So TestDefaultConfiguration.InitializeDefaults(); is needed? then why do we remove the one duplicate which has it ?

test/TesterInternal/StreamingTests/HaloStreamSubscribeTests.cs
xiazen: ahhh~ I missed it  .

PR #2326, opened 2016-10-21
Release BinaryTokenStreamWriter buffers after use in more cases.

ReubenBond:
I'm not certain how useful this change is, since I haven't profiled the difference.

There are several cases where we are taking buffers from the buffer pool and not returning them. This PR eliminates several of those cases.

`BinaryTokenStreamWriter.ToByteArray()` allocates a new byte array and copies the result, whereas `ToBytes()` returns the underlying `List<ArraySegment<byte>>` directly. Hence it is safe to free the buffers after the former (assuming no further use), but not after the latter. `ToBytes()` should be called `AsBytes()` or `GetBuffers()` to reduce any confusion.

Adding an overload to `JenkinsHash.ComputeHash` which took a `List<ArraySegment<byte>>` instead of just the current `byte[]` would further help to reduce allocations.

### DIFF
--- a/src/Orleans/IDs/GrainId.cs
+++ b/src/Orleans/IDs/GrainId.cs
@@ -319,7 +319,9 @@ namespace Orleans.Runtime
         {
             var writer = new BinaryTokenStreamWriter();
             writer.Write(this);
-            return writer.ToByteArray();
+            var result = writer.ToByteArray();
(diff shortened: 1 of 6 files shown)


gabikliot:
Yes, we should try to return, whenever we can (given reasonable effort).

PR #2327, opened 2016-10-21
GrainCasterFactory: emit correct call/callvirt opcodes

ReubenBond:
Fixes an issue which @jdom encountered on .NET Core where the incorrect opcode is being used in `GrainCasterFactory`.

Why does a different opcode need to be emitted? That's because of this line:
```C#
private static readonly MethodInfo IsAssignableFromMethodInfo =
            TypeUtils.Method((Type t) => t.IsAssignableFrom(default(Type)));
```

On full .NET, `Type` has a virtual `IsAssignableFrom(Type)` instance method. On .NET Core, that method is a (non-virtual) extension method.

### DIFF
--- a/src/Orleans/Core/GrainCasterFactory.cs
+++ b/src/Orleans/Core/GrainCasterFactory.cs
@@ -129,8 +129,14 @@ namespace Orleans
         /// <param name="method">The method to call.</param>
         private static void EmitCall(ILGenerator il, MethodInfo method)
         {
-            if (method.IsFinal || !method.IsVirtual) il.Emit(OpCodes.Call, method);
-            else il.Emit(OpCodes.Callvirt, method);
+            if (method.IsFinal || !method.IsVirtual)
+            {
+                il.Emit(OpCodes.Call, method);
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


src/Orleans/Core/GrainCasterFactory.cs
```diff
@@ -122,6 +122,17 @@ internal static class GrainCasterFactory
...
         
+        /// <summary>
+        /// Emits a call to the specified method.
+        /// </summary>
+        /// <param name="il">The il generator.</param>
+        /// <param name="method">The method to call.</param>
+        private static void EmitCall(ILGenerator il, MethodInfo method)
+        {
+            if (method.IsFinal || !method.IsVirtual) il.Emit(OpCodes.Call, method);
+            else il.Emit(OpCodes.Callvirt, method);
```
jdom: Please use braces and new lines

ReubenBond: Will do. Are there stylecop rules which we can agree on & use?

jdom:
Thanks for the quick turnaround, @ReubenBond ! :)

PR #2329, opened 2016-10-21
Make basic silo tests run with .NET Standard version of Orleans

jdom:
AppDomain shim loads all assemblies in local folder.
Added all references from Orleans and OrleansRuntime into Tester since because of the csproj hack, it looks like the nuget dependencies are not transitively passed on to Tester

### DIFF
--- a/vNext/src/Orleans/Shims/AppDomain.cs
+++ b/vNext/src/Orleans/Shims/AppDomain.cs
@@ -1,13 +1,16 @@
 ﻿#if NETSTANDARD_TODO
 using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Reflection;
-using Orleans.Runtime;
 
 namespace Orleans
 {
     internal sealed class AppDomain
     {
         public delegate void AssemblyLoadEventHandler(object sender, AssemblyLoadEventArgs args);
+
         public delegate void UnhandledExceptionEventHandler(object sender, UnhandledExceptionEventArgs e);
 
         public static AppDomain CurrentDomain { get; private set; }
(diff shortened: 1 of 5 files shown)


vNext/test/Tester/project.json
```diff
@@ -6,7 +6,31 @@
     "Newtonsoft.Json": "9.0.1",
     "xunit": "2.1.0",
     "xunit.runner.visualstudio": "2.1.0",
-    "Xunit.SkippableFact": "1.2.14"
+    "Xunit.SkippableFact": "1.2.14",
+
+
+
+
+    "System.Collections.Immutable": "1.2.0",
```
jdom: Added a space here to separate the test dependencies (cannot add comments in a json file).
These dependencies are currently only needed because of our csproj hack, but should be gone after we remove the hack, and become normal transitive dependencies

vNext/src/Orleans/Shims/AppDomain.cs
```diff
@@ -31,19 +34,31 @@ private AppDomain()
             //};
         }
 
-        public Assembly[] GetAssemblies()
+        private Lazy<Assembly[]> assembliesList = new Lazy<Assembly[]>(() =>
         {
-            // TODO: replace with IAssemblyCatalog or something like that.
-            Assembly[] assemblies =
+            string path = ".";
+            var assemblyFiles = Directory.EnumerateFiles(path, "*.dll");
```
ReubenBond: Not sure "." is the best way to get the directory which the current assembly resides in - wont this return the files in the working directory?

I find this to be reliable, particularly when dealing with things like shadow copying, etc:
```C#
var path = new Uri((this.GetType().Assembly).CodeBase).LocalPath;
```

jdom: Heh, actually I had that originally (plus unescaping the uri string) but decided to simplify it like this for now, since it's only temporary. I wouldnt mind changing to it again if that's preferable though

PR #2330, opened 2016-10-21
Modified stream types to not use fallback serializer and allow external

Carlm-MS:
- implement serializers for AzureQueueBatchContainer and EventSequenceToken.  They were routed to the fallback serializer which caused them to be incompatible with external serializers
- added unit tests to verify serialization/deserialization with external serializers.

### DIFF
--- a/src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueBatchContainer.cs
+++ b/src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueBatchContainer.cs
@@ -4,9 +4,9 @@ using System.Linq;
 using Microsoft.WindowsAzure.Storage.Queue;
 using Newtonsoft.Json;
 using Orleans.Providers.Streams.Common;
-using Orleans.Runtime;
 using Orleans.Serialization;
 using Orleans.Streams;
+using RuntimeRequestContext = Orleans.Runtime.RequestContext;
 
 namespace Orleans.Providers.Streams.AzureQueue
 {
@@ -14,22 +14,24 @@ namespace Orleans.Providers.Streams.AzureQueue
     internal class AzureQueueBatchContainer : IBatchContainer
     {
         [JsonProperty]
-        private EventSequenceToken sequenceToken;
+        protected EventSequenceToken sequenceToken;
 
         [JsonProperty]
-        private readonly List<object> events;
+        protected List<object> events;
 
         [JsonProperty]
-        private readonly Dictionary<string, object> requestContext;
+        protected Dictionary<string, object> requestContext;
 
         [NonSerialized]
         // Need to store reference to the original AQ CloudQueueMessage to be able to delete it later on.
         // Don't need to serialize it, since we are never interested in sending it to stream consumers.
         internal CloudQueueMessage CloudQueueMessage;
 
-        public Guid StreamGuid { get; private set; }
+        public Guid StreamGuid { get; protected set; }
 
-        public String StreamNamespace { get; private set; }
+        public String StreamNamespace { get; protected set; }
+
+        internal Dictionary<string, object> RequestContext => this.requestContext;
 
         public StreamSequenceToken SequenceToken
         {
@@ -37,7 +39,7 @@ namespace Orleans.Providers.Streams.AzureQueue
         }
 
         [JsonConstructor]
-        private AzureQueueBatchContainer(
+        internal AzureQueueBatchContainer(
             Guid streamGuid, 
             String streamNamespace,
             List<object> events,
@@ -48,7 +50,7 @@ namespace Orleans.Providers.Streams.AzureQueue
             this.sequenceToken = sequenceToken;
         }
 
-        private AzureQueueBatchContainer(Guid streamGuid, String streamNamespace, List<object> events, Dictionary<string, object> requestContext)
+        protected AzureQueueBatchContainer(Guid streamGuid, String streamNamespace, List<object> events, Dictionary<string, object> requestContext)
         {
             if (events == null) throw new ArgumentNullException("events", "Message contains no events");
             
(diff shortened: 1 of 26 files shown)


jason-bragg:
I am concerned that this will change the default binary serialization, as that would break any customer that is currently using these types.  Any existing containers persisted in an azure queue or any tokens persisted by an application will no longer be readable if the default binary serialization changes.

Questions:
Does this change the default binary serialization?  I suspect it does.
How can you test/verify that it does not?

This change needs to be made in a way such that the original default binary format does not change.

Carlm-MS:
Yes - it changes the default serialization and is incompatible with the old version.

The old version was using the fallback serializer which by default is using the .net binary serializer.  This is problematic because the .net serializer doesn't understand nested types that are serialized using IExternalSerializer. This commit is a fix for that issue and also boosts performance a bit.

I will see if I can make the change backward-compatible.

src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueBatchContainer.cs
```diff
@@ -58,6 +62,70 @@ private AzureQueueBatchContainer(Guid streamGuid, String streamNamespace, List<o
...
+        {
+            var typed = untypedInput as AzureQueueBatchContainer;
+            if (typed == null)
+            {
+                throw new SerializationException();
+            }
+
+            writer.Write(typed.StreamGuid);
+            writer.Write(typed.StreamNamespace);
+            SerializationManager.Serialize(typed.sequenceToken, writer);
```
ReubenBond: These should be `SerializeInner` - similarly for the Deserialize calls. Check out the pattern used in auto-generated serializers, particularly for DeepCopy with its call to `SerializationContext.Current.RecordObject(original, result);` immediately after the result is created and for Deserializer with its call to `Serialization.`DeserializationContext.`Current.`RecordObject(result);` immediately after the constructor.

jdom:
Also please update the changelog.md file, especially if it's a breaking change. Consider that the bigger audience for the changelog is people trying to make the decision of whether to upgrade and how their app will be impacted

src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueBatchContainer2.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <returns>The deserialized value</returns>
+        public static object Deserialize(Type expected, BinaryTokenStreamReader reader)
+        {
+            var guid = reader.ReadGuid();
+            var ns = reader.ReadString();
+            var eventToken = SerializationManager.DeserializeInner<EventSequenceToken2>(reader);
+            var events = SerializationManager.DeserializeInner<List<object>>(reader);
+            var context = SerializationManager.DeserializeInner<Dictionary<string, object>>(reader);
+            var deserialized = new AzureQueueBatchContainer2(guid, ns, events, context, eventToken);
+            DeserializationContext.Current.RecordObject(deserialized);
```
ReubenBond: The reason ordering is important in DeepCopy/Deserialize is for cyclic objects. I'm not sure if it's possible for this class to be cyclic, but if so then these methods need to be changed like so:
```
construct object
record object
deserialize & set properties on object
return object
```

It's probably fine as-is, but worth noting.

Carlm-MS: @ReubenBond I'll make the change. Although it's unlikely that a cycle will be introduced, it's possible.

test/NonSiloTests/SerializationTests/StreamTypeSerializationTests.cs
```diff
@@ -0,0 +1,217 @@
...
+using System.Runtime.Serialization.Formatters.Binary;
+using Orleans.Providers.Streams.AzureQueue;
+using Orleans.Providers.Streams.Common;
+using Orleans.Runtime.Host.Providers.Streams.AzureQueue;
+using Orleans.Serialization;
+using Orleans.ServiceBus.Providers;
+using OrleansServiceBus.Providers.Streams.EventHub;
+using Xunit;
+
+namespace UnitTests.Serialization
```
jason-bragg: In the azure queue stream provider, we deserialized the container using
DerializationManager.DeserializeFromByteArray<AzureQueueBatchContainer>(bytes);

My suspicion is that this will work even if we've serialized a AzureQueueBatchContainer2, but can we add a test to verify?

src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueBatchContainer2.cs
```diff
@@ -0,0 +1,111 @@
...
+using Newtonsoft.Json;
+using Orleans.CodeGeneration;
+using Orleans.Providers.Streams.AzureQueue;
+using Orleans.Providers.Streams.Common;
+using Orleans.Serialization;
+
+namespace Orleans.Runtime.Host.Providers.Streams.AzureQueue
+{
+    [RegisterSerializer]
+    internal class AzureQueueBatchContainer2 : AzureQueueBatchContainer
```
jason-bragg: Just a nit pick detail: If we're going to specifically version data can we use naming convention of blag and blargV2 rather than just putting a 2 on it?

sergeybykov:
@Carlm-MS Are you fine with the requested changes?

jason-bragg:
@Carlm-MS, Just touching base to see if there is anything you need from us on this.

Carlm-MS:
@jason-bragg the requested changes seem reasonable.  I'll take care of them.

jdom:
Hi @Carlm-MS, looks like the Orleans.vNext solution is failing to build. Until we migrate entirely to .NET Standard (or at least have final tooling that allows multitargeting) we currently have 2 solutions to build against the different targets, and the vNext one just has links to the files in the main solution. Could you add links to the new files by opening Orleans.vNext,sln?
Let me know if you need help

jason-bragg:
@dotnet-bot test this please

PR #2331, opened 2016-10-22
 upgrade xunit.assert to be 2.1.0 in TestInternalGrains

xiazen:
upgrade xunit.assert to be 2.1.0 in TestInternalGrains to inline with Tester dependency


xiazen:
Thanks for @jdom 's suggestion on upgrade xunit.assert to be 2.1.0 instead of add new dependency to Tester.

Issue #2145, opened 2016-09-12
[CRAWL Phase] .Net Core Compatibility


galvesribeiro:
_This issue will track the **CRAWL** phase of Orleans 2.X milestone as tracked by #2143_

The idea with this task is to basically make Orleans Core compatible with [.Net Core](https://github.com/dotnet/coreclr), by having it target [.Net Standard](https://docs.microsoft.com/en-us/dotnet/articles/standard/library) and have the minimal impact as possible in the current codebase. 

Since .Net Core tooling still unreliable and will (drastically) change before it reach RTM, we are following the guidelines from [here](https://docs.microsoft.com/en-us/dotnet/articles/core/tutorials/target-dotnetcore-with-msbuild)  and in a new temporary solution we will work at `master` branch.

The target is to have Orleans building and running on .Net Full CLR and .Net Core (hence why target .Net Standard) on Windows and have the full test set only for the .Net Full CLR (.Net 4.6.1+) while having basic validation on .Net Core on Windows. 

We will focus on Windows as a baseline but that doesn't mean people can't run tests on Linux/OSX once this work is done. That means that the core test set will focus on .Net Full CLR and basic .Net Core on Windows but you can test your Orleans application on .Net Core supported OSes. The test framework should stay as is. That means the major efforts will not be to update the test kit to multiple OSes and full .Net Core coverage but, if you feel comfortable, we still accept (and really encourage and appreciate!) PRs however, it will take lower priority (and of course efforts on review) from the core team and people from the community working on this port.

Here are the macro tasks and its issues that will be tracked for this phase:
- [x] Add new solution targeting .Net Standard Library 1.5 (`netstandard15`) - #2177 
- [x] Build Orleans.vNext.sln in the CI Build.cmd script - #2353 #2362
- [x] Replace Binary Serializer - #2162
- [x] Move Performance Counters to its own TelemetryConsumer - #2122 
- [x] Use supported overloads from Socket class - #2195 
- [x] Remove usage of named mutexes #2366
- [x] Remove AppDomain usage for assembly loading/scanning - #2025 and in continued progress by @attilah 
- [x] Run non-silo tests (not blocked by Silo not working in memory) - in progress by @xiazen
- [x] Avoid Silo being MarshalByRefObject but support running tests - #2196 
- [x] Avoid test hooks being MarshalByRefObject (a few still pending) - #2333
- [x] Avoid usages of CallContext - #2197 
[...]

PR #2333, opened 2016-10-23
Convert AppDomain test hooks to SystemTarget

jdom:
This is so that we can eventually run the tests without the need to use AppDomains, as they are not available in .NET Core.

This PR is related to #2145

### DIFF
--- a/src/OrleansTestingHost/AppDomainSiloHost.cs
+++ b/src/OrleansTestingHost/AppDomainSiloHost.cs
@@ -129,7 +129,7 @@ namespace Orleans.TestingHost
             if (obj == null) return null;
             if (obj is MarshalByRefObject || obj is ISerializable)
             {
-                // Referernce to the provider can safely be passed across app-domain boundary in unit test process
+                // Reference to the provider can safely be passed across app-domain boundary in unit test process
                 return obj;
             }
             throw new InvalidOperationException(
--- a/src/OrleansTestingHost/AppDomainSiloHost.cs
+++ b/src/OrleansTestingHost/AppDomainSiloHost.cs
@@ -1,8 +1,18 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Runtime.Serialization;
 using Orleans.CodeGeneration;
+using Orleans.Providers;
 using Orleans.Runtime;
 using Orleans.Runtime.Configuration;
+using Orleans.Runtime.GrainDirectory;
+using Orleans.Runtime.Messaging;
+using Orleans.Runtime.Placement;
+using Orleans.Runtime.TestHooks;
+using Orleans.Storage;
 
 namespace Orleans.TestingHost
 {
@@ -18,13 +28,17 @@ namespace Orleans.TestingHost
         public AppDomainSiloHost(string name, Silo.SiloType siloType, ClusterConfiguration config)
         {
             this.silo = new Silo(name, siloType, config);
+            this.silo.InitializeTestHooksSystemTarget();
+            this.AppDomainTestHook = new AppDomainTestHooks(this.silo);
         }
 
         /// <summary> SiloAddress for this silo. </summary>
         public SiloAddress SiloAddress => silo.SiloAddress;
 
-        /// <summary>Gets the Silo test hook (NOTE: this will be removed really soon, and was migrated here temporarily)</summary>
-        public Silo.TestHooks TestHook => silo.TestHook;
+        /// <summary>Gets the Silo test hook</summary>
+        internal ITestHooks TestHook => GrainClient.InternalGrainFactory.GetSystemTarget<ITestHooksSystemTarget>(Constants.TestHooksSystemTargetId, this.SiloAddress);
+
+        internal AppDomainTestHooks AppDomainTestHook { get; }
 
         /// <summary>Methods for optimizing the code generator.</summary>
         public class CodeGeneratorOptimizer : MarshalByRefObject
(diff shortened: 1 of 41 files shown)


src/OrleansTestingHost/AppDomainSiloHost.cs
```diff
@@ -87,4 +101,87 @@ public void Shutdown()
...
+
+        /// <summary>Find the named storage provider loaded in this silo. </summary>
+        internal IStorageProvider GetStorageProvider(string name) => CheckReturnBoundaryReference("storage provider", (IStorageProvider)silo.StorageProviderManager.GetProvider(name));
+
+        private static T CheckReturnBoundaryReference<T>(string what, T obj) where T : class
+        {
+            if (obj == null) return null;
+            if (obj is MarshalByRefObject || obj is ISerializable)
+            {
+                // Referernce to the provider can safely be passed across app-domain boundary in unit test process
```
ReubenBond: nit: typo on Reference

src/OrleansTestingHost/AppDomainSiloHost.cs
```diff
@@ -87,4 +101,87 @@ public void Shutdown()
...
+                return obj;
+            }
+            throw new InvalidOperationException(
+                $"Cannot return reference to {what} {TypeUtils.GetFullName(obj.GetType())} if it is not MarshalByRefObject or Serializable");
+        }
+
+        public IDictionary<GrainId, IGrainInfo> GetDirectoryForTypeNamesContaining(string expr)
+        {
+            var x = new Dictionary<GrainId, IGrainInfo>();
+            foreach (var kvp in ((LocalGrainDirectory)silo.LocalGrainDirectory).DirectoryPartition.GetItems())
```
ReubenBond: Did `Silo.LocalGrainDirectory` change type recently? why the cast? Same goes for `Silo.Catalog`.

src/OrleansTestingHost/AppDomainSiloHost.cs
```diff
@@ -87,4 +101,87 @@ public void Shutdown()
...
+        private readonly SafeRandom random = new SafeRandom();
+
+        internal void BlockSiloCommunication(IPEndPoint destination, double lossPercentage)
+        {
+            if (simulatedMessageLoss == null)
+                simulatedMessageLoss = new ConcurrentDictionary<IPEndPoint, double>();
+
+            simulatedMessageLoss[destination] = lossPercentage;
+
+            var mc = (MessageCenter)silo.LocalMessageCenter;
```
ReubenBond: I wonder if this class should take these dependencies explicitly so the casts can be avoided.

test/TesterInternal/ActivationsLifeCycleTests/DeactivateOnIdleTests.cs
```diff
@@ -273,32 +280,23 @@ public void Dispose()
...
+            secondaryActivation = await testCluster.SecondarySilos[0].TestHook.UnregisterGrainForTesting(grain);
             Assert.Equal(0, primaryActivation + secondaryActivation);
 
 
-            if (doLazyDeregistration)
+            if (lazyDeregistrationDelay > TimeSpan.Zero)
             {
                 // Wait a bit
                 TimeSpan pause = lazyDeregistrationDelay.Multiply(2);
                 logger.Info("Pausing for {0} because DoLazyDeregistration=True", pause);
```
ReubenBond: nit: log line is misleading

src/OrleansTestingHost/AppDomainSiloHost.cs
```diff
@@ -87,4 +101,87 @@ public void Shutdown()
...
+            return CheckReturnBoundaryReference("bootstrap provider", provider);
+        }
+
+        /// <summary>Find the named storage provider loaded in this silo. </summary>
+        internal IStorageProvider GetStorageProvider(string name) => CheckReturnBoundaryReference("storage provider", (IStorageProvider)silo.StorageProviderManager.GetProvider(name));
+
+        private static T CheckReturnBoundaryReference<T>(string what, T obj) where T : class
+        {
+            if (obj == null) return null;
+            if (obj is MarshalByRefObject || obj is ISerializable)
```
jdom: You are probably right. Nevertheless this is old code that I just moved around and plan to deprecate it soon, not add more usages to it, since we want to get away from app domains for testing

src/OrleansTestingHost/AppDomainSiloHost.cs
jdom: In the original code before the move, it was using the private field that didn't need casting. Similarly these are test hooks that were just moved but not migrated yet, and should be migrated soon. Didn't make too much effort to make sure abstractions are kept in place, since they are going away.

src/OrleansTestingHost/AppDomainSiloHost.cs
jdom: similarly, because the code is now moved to a non-nested class, it can no longer access private fields, so the casting would have to either be done in the constructor or here (or expose implementation details that we probably don't want to do for code that should eventually be removed).

test/TestGrains/StatsCollectorGrain.cs
```diff
@@ -0,0 +1,35 @@
...
+            return TaskDone.Done;
+        }
+
+        public Task ReportStatsCalled()
+        {
+            numMetricsCalls++;
+            return TaskDone.Done;
+        }
+
+        public Task<long> GetReportMetricsCallCount()
```
jdom: Took a look, and I still can't make sense why it was ever -1, but still, it does seem that 0 is the right value also according to how the test which uses it is asserting, so fixed

Issue #1263, opened 2016-01-09
TODO: Turn on Code Analysis for the Orleans.dll developer API assembly.


jthelin:
Recording this work item as a relatively easy Up-For-Grabs issue that anyone wanting to get more involved with Orleans project could take on.

TODO: Turn on Code Analysis to run automatically during build of the main developer API in `Orleans.dll` assembly by ticking the "Enable Code Analysis on Build" checkbox in project properties.

Then need to make minor code changes to fix any issues identuified by Code Analysis -- mostly missing IDisposable functions on some classes that use other IDisposable resources.

The main goal is to get a clean pass (zero CA warnings) from "_MIcrosoft Minimal Recommended Ruleset_" when running Build.cmd of Visual Studio build.

A stretch goal would be to use "_MIcrosoft Managed Recommended Ruleset_", but that is a significantly more onerous task requiring more code fixes.

Extra credit: 
`OrleansAzureUtils.dll` is the other "developer facing" assembly containing often use public Orleans APIs. Would be good to turn on Code Analysis for that assembly too, either as separate PR or combined with this one.

leonpw:
I will look into this

sergeybykov:
Resolved via #2335.

PR #2335, opened 2016-10-23
TODO: Turn on Code Analysis for the Orleans.dll developer API assembl…

leonpw:
…y. #1263

### DIFF
--- a/src/Orleans/Async/AsyncLock.cs
+++ b/src/Orleans/Async/AsyncLock.cs
@@ -9,15 +9,15 @@ namespace Orleans
     /// </summary>
     /// <remarks>
     /// (Adapted from http://blogs.msdn.com/b/pfxteam/archive/2012/02/12/10266988.aspx)
-    /// 
+    ///
     /// When programming with <b>async</b>, the <b>lock</b> keyword is problematic:
     /// <list type="bullet">
     ///     <item><b>lock</b> will cause the thread to block while it waits for exclusive access to the critical section of code.</item>
     ///     <item>The <b>await</b> keyword cannot be used within the scope of a <b>lock</b> construct.</item>
     /// </list>
-    /// 
+    ///
     /// It is still useful, at times, to provide exclusive access to a critical section of code. AsyncLock provides semantics
-    /// that correspond to that of a (non-recursive) mutex, while maintining compatibility with the tenets of async programming. 
+    /// that correspond to that of a (non-recursive) mutex, while maintining compatibility with the tenets of async programming.
     /// </remarks>
     /// <example>
     /// The following example implements some work that needs to be done under lock:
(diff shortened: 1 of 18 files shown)


src/Orleans/Async/AsyncLock.cs
```diff
@@ -35,15 +35,16 @@ namespace Orleans
...
-    /// 2) AsyncLock does NOT have to implement the Finalizer function. The underlying resource of SemaphoreSlim will be eventually released by the .NET, 
+    /// 2) AsyncLock does NOT have to implement the Finalizer function. The underlying resource of SemaphoreSlim will be eventually released by the .NET,
     /// when SemaphoreSlim is finalized. Having finalizer for AsyncLock will not speed it up.
     /// 3) LockReleaser is IDisposable to implement the "using" pattern.
     /// 4) LockReleaser does NOT have to implement the Finalizer function. If users forget to Dispose the LockReleaser (analagous to forgetting to release a mutex)
-    /// the AsyncLock wil remain locked, which may potentialy cause deadlock. This is OK, since these are the exact regular mutex semantics - if one forgets to unlcok the mutex, it stays locked. 
+    /// the AsyncLock wil remain locked, which may potentialy cause deadlock. This is OK, since these are the exact regular mutex semantics - if one forgets to unlock the mutex, it stays locked.
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
```
sergeybykov: This is a stop gap solution to avoid going down the rabbit hole of classes that use `AsyncLock` and making them all `IDisposable`, right? I think that's fine. We can open an issue to track removal of it. Low priority IMO.

src/Orleans/Cancellation/GrainCancellationTokenSource.cs
```diff
@@ -1,14 +1,14 @@
-﻿using System;
-using System.Threading.Tasks;
-using Orleans.CodeGeneration;
+﻿using Orleans.CodeGeneration;
 using Orleans.Serialization;
+using System;
+using System.Threading.Tasks;
```
sergeybykov: There was no real reason to flip them, right? It's just some tool that did it automatically?

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -14,12 +6,19 @@
...
 using Orleans.Storage;
 using Orleans.Streams;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
```
sergeybykov: Same here - there was no real reason to flip the using statements, was there?

src/Orleans/Statistics/ThreadCycleStopWatch.cs
```diff
@@ -98,10 +98,53 @@ private static class NativeMethods
...
             [DllImport("Kernel32.dll")]
-            public static extern IntPtr GetCurrentThread();    
+            public static extern IntPtr GetCurrentThread();
+        }
+
+        #region IDisposable Support
+
+        private bool disposedValue = false; // To detect redundant calls
+
+        protected virtual void Dispose(bool disposing)
```
sergeybykov: I wonder if we even need to do anything here? The class doesn't actually hold any native resources that it needs to release. Maybe just suppress the warning with a comment about that instead of adding effective no-op code?

src/Orleans/Statistics/ThreadTrackingStatistic.cs
```diff
@@ -241,5 +242,40 @@ private void TrackContextSwitches()
             }
 #endif
         }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
+
+        protected virtual void Dispose(bool disposing)
```
sergeybykov: Same comment here - consider suppressing the warning instead of adding no-op code. Curious what others think.

src/Orleans/Runtime/OutsideRuntimeClient.cs
ReubenBond: I'm surprised it ordered them this way: the rule is usually "System first, then alphabetically" isn't it? Doesn't really matter either way.

src/Orleans/Cancellation/GrainCancellationTokenSource.cs
leonpw: Yes, some option in my Visual Studio was off. Sorting namespaces with System first.

src/Orleans/Statistics/ThreadCycleStopWatch.cs
leonpw: It does have a IntPtr handle. I'll empty the dispose then.

src/Orleans/Statistics/ThreadTrackingStatistic.cs
leonpw: Idem dito here.

sergeybykov:
Thank you, @leonpw!

PR #2336, opened 2016-10-24
RuntimeType serialization

ReubenBond:
The first commit ensures that the fallback serializer is not available for most of the tests in BuiltInSerializerTests - most tests should pass without the fallback.

The second commit registers a serializer for `RuntimeType` (the runtime type of `Type` 😉), aligns the serializer with the deserializer, and adds a test.

### DIFF
--- a/src/Orleans/Serialization/BuiltInTypes.cs
+++ b/src/Orleans/Serialization/BuiltInTypes.cs
@@ -1637,7 +1637,7 @@ namespace Orleans.Serialization
 
         internal static void SerializeType(object obj, BinaryTokenStreamWriter stream, Type expected)
         {
-            stream.Write(((Type)obj).OrleansTypeName());
+            stream.Write(((Type)obj).OrleansTypeKeyString());
         }
 
         internal static object DeserializeType(Type expected, BinaryTokenStreamReader stream)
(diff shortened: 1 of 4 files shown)


src/Orleans/Serialization/BuiltInTypes.cs
```diff
@@ -1637,7 +1637,7 @@ internal static object CopyDateTimeOffset(object obj)
 
         internal static void SerializeType(object obj, BinaryTokenStreamWriter stream, Type expected)
         {
-            stream.Write(((Type)obj).OrleansTypeName());
+            stream.Write(((Type)obj).OrleansTypeKeyString());
```
sergeybykov: This won't break backward compatibility with data serialized and persisted previously, right?

ReubenBond: It won't break backward compatibility in any case which works today (where `OrleansTypeKeyString()` & `OrleansTypeName()` just happen to produce the same result, eg:
```C#
> typeof(Grain).OrleansTypeName()
"Orleans.Grain"
> typeof(Grain).OrleansTypeKeyString()
"Orleans.Grain" // Simple (non-generic, non-array, etc) types not in "System" will be equal.
> typeof(int).OrleansTypeName()
"System.Int32"
> typeof(int).OrleansTypeKeyString()
"Int32" // Types in System wouldn't work.
```

If you look at `DeserializeType` you'll see it calls `ResolveTypeName` which looks for the type in `types`. If you look for where `types` gets modified, you'll see that the keys are generated using `OrleansTypeKeyString()` and not `OrleansTypeName()` - so the current serialize method is wrong to use `OrleansTypeName()`.

ReubenBond: This PR just makes these two consistent. The types still need to be 'registered' in order for serialization to succeed. Ideally, we would remove that requirement.

jdom:
LGTM. Unless @sergeybykov has a concern, I'll merge it

Issue #2337, opened 2016-10-24
Grain throws NullRefExceptions when using default constructor in unit test scenarios


rrector:
Grain will throw difficult to track down NullReferenceExceptions when the default constructor is used to create grains in a unit test environment. Additionally, version 1.3 now throws NullRef exceptions in this scenario when accessing GrainFactory and ServiceProvider properties where these used to return null.

To aid in debugging, Orleans should provide more descriptive exceptions or expected results when the runtime is not available.

sergeybykov:
Resolved via #2338.

PR #2338, opened 2016-10-24
Provide better handling of null GrainRuntime in Grain code

rrector:
As described in Issue #2337, provide either default behavior or useful exceptions when IGrainRuntime or IGrainIdentity are null either because of using the default constructor or passing null to the constructor that sets them (only applies in the context of unit testing).


ReubenBond:
This seems fine to me, does anyone have any objections to it?

sergeybykov:
Thank you, @rrector!

Issue #467, opened 2015-05-29
Make the GrainClient class non-static


nehmebilal:
Connecting to multiple Orleans Silos from the same client is not possible because the Orleans client class is static and there can be only one of it in a process. Making the `GrainClient` non-static will help solving this issue.

Client code will create an instance of `GrainClient` instead of using the static `GrainClient.Initialize` as follows:

``` csharp
        IGrainClient orleansClient = new GrainClient(“DevTestClientConfiguration.xml”);
        IGrainFactory grainFactory = orleansClient.GrainFactory;
        IHello grain = grainFactory.GetGrain<IHello>(“0”);
```

---

@gabikliot wrote:
TODOs:
- [x] Remove all static mutable fields from client runtime.
- [ ] Static Interners
- [x] `RuntimeClient.Current.SendRequest`
- [ ] Static `TraceLogger` instances
- [ ] Static mutable fields inside `TraceLogger`
- [x] Placement Singletons
- [x] Serializer internal collections
- [ ] All statistics counters

gabikliot:
This is a great in initiative.
Lets try to make progress on that in steps. First step that I would recommend is to identify all places within the runtime that rely on the runtime to be singleton. This goes beyond just GrainClient class. Of course, we won't be able to remove singleton static GrainClient  until we remove all the internal runtime static singletons.

So the 1st one to look is `RuntimeClient`, which has a static `RuntimeClient.Current` of type `IRuntimeClient`.
This `RuntimeClient.Current` is what allowed us to have the same code in multiple places for silo and client runtime. One notable example is `GrainReference.InvokeMethod_Impl` calling `RuntimeClient.Current.SendRequest`.  To fix that we would need to inject some way for all grain references to call `SOMETHING.SendRequest`, which will be different on the client and silo side.

So I suggest we focus on those places first. Once we solved all those, we can look at making a breaking API change to application code, which will be a result of making `GrainClient` class non-static.

ghuntley:
"This is a great in initiative." :+1: agreed.

ReubenBond:
If we want to support multiple grain clients in a given process/AppDomain calling into different clusters, then each `GrainReference` should have a reference to the `IRuntimeClient` for that cluster. That's easy enough. Issues become more apparent when we talk about serializing/deserializing grain references. When a `GrainReference` is sent over the wire or to/from storage, we need some way to associate it with the correct cluster. This could be done either automatically or manually.

[...]

Issue #96, opened 2015-02-07
Support for custom (default) placement strategies


hstuart:
Currently the default placement strategy is rather hardwired and limited and it is only possible to tweak it using the built-in placement strategies. I'd like to propose that we open up PlacementDirectorsManager, PlacementStrategy, and PlacementDirector so grain solutions can opt to create their own placement strategies for more specialised setups. Simultaneously, it would be beneficial to be able to specify a custom default placement strategy in the configuration and have that take effect globally rather than having to go through all the grains and add attributes to them.

Some of the custom placement strategies that might be of interest could be to limit new activations to not overwhelm a fixed machine setup (e.g. in a non-cloud/non-virtual setup), place activations according to expected communication coherence, etc.

sergeybykov:
Let me try to break it into pieces. Correct me if I'm wrong.

1. Make default placement globally configurable.
2. Add ability to add custom placement policies with application code.
3. Placement strategies that would benefit scenarios of fixed deployments.

1 - no-brainer.
2 - seems like a medium size workitem to augment AssemblyLoader load custom strategies and lazily associate them with grain classes that use them.
3. Curious about specific examples and how far PreferLocal can go in these scenarios.

Overall, great suggestions!

gabikliot:
The default placement configuration is already configurable. GlobalPlacement.DefaultPlacementStrategy.

About adding custom placement strategies: it is very easy now. You add a new PlacementStrategy attribute, write the PlacementDirector that implements this strategy and hook it up correctly in the  PlacementDirectorsManager. What else is needed?

New placement strategies - great, we would love to see more.

sergeybykov:
As I understand, the proposal is to bring custom placement strategies to the application level, so that one doesn't need to rebuild Orleans to add a new strategy to use by their grain classes.

My bad - I forgot that DefaultPlacementStrategy is already in config.

hstuart:
I should have been clearer with regards to configuration support, sorry. What I meant was to extend PlacementStrategy's GetDefaultPlacement with assembly/type load support for externally specified assemblies. I agree that it is already exposed in the configuration.

And yes, my proposal is to bring custom placement strategies to the application level, as I can take much more informed choices there than in the general framework, and I would prefer not having to maintain my own fork of Orleans just to get custom placement.

[...]

PR #2345, opened 2016-10-27
Reduce usage of stateful statics by leveraging dependency injection

ReubenBond:
This is the first of several PRs aimed at removing stateful static classes so that we can eventually host multiple grain clients in the same process/AppDomain. See #467.
- `AssemblyProcessor` is no longer static & is now disposable because it hooks the `AppDomain.AssemblyLoad` event (and hence should remove itself from that event on disposal).
- Added a new class, `TypeMetadataCache`, which holds the previously static grain->GrainReference & grain->Invoker lookups in `GrainFactory`
- Removed all statics from `GrainFactory`
- Added a new `IInternalGrainFactory` interface which `GrainFactory` implements.
- Refactored `GrainFactory` methods usage of `GrainFactoryBase.MakeGrainReference_FromType` to remove statics & closure allocations.
- `PlacementDirectorsManager` uses DI to lookup placement directors.
- Removed static `PlacementDirectorsManager.Instance` property.
- Removed `InitializeClass` from `PlacementStrategy` sub classes (using static initializer instead)
- Changed `PlacementDirector` from abstract class to interface.
- Added `IPlacementDirector<TStrategy>` marker interface (implementing `IPlacementDirector`) to aid use in DI (mapping between PlacementStragy & director is resolved using DI rather than hard-coded).
- _`GrainFactory` is now marked as internal_
- Removed many references to `RuntimeClient.Current` in favor of injecting the `IRuntimeClient` via DI
- Added `SerializationTestEnvironment` class to make it easier to set up `SerializationManager`, etc, now that `AssemblyProcessor` is not static.
- Removed static `RegistrarManager.Instance` & replaced with DI.
- Added `DefaultPlacementStrategy` class to represent the default placement strategy.
- Removed static `DeploymentLoadPublisher.Instance` & replaced with DI.
- Added `SiloInitializationParameters` to capture the values which are passed to `Silo` ctor & then passed through to various other classes. This makes it injectable (shouldn't inject plain `SiloAddress` for ex). This class also holds derived values, like the local silo address & node config.
- Most services are registered via DI & typically consume all of their services via DI (rather than static properties).

Fixed a few issues discovered when running functionals, which were otherwise clean. I'm running functionals again. There are some issues which need resolving for load tests (build errors) which I'm sorting through.


src/Orleans/Runtime/GrainInterfaceMap.cs
```diff
@@ -76,8 +76,11 @@ public override string ToString()
         private readonly bool localTestMode;
         private readonly HashSet<string> loadedGrainAsemblies;
 
-        public GrainInterfaceMap(bool localTestMode)
+        private readonly PlacementStrategy defaultPlacementStrategy;
```
veikkoeeva: I wonder would it be useful to make the default a public property. With the idea the default would be visible.

src/Orleans/Runtime/GrainRuntime.cs
```diff
@@ -4,31 +4,45 @@
...
+            ServiceId = globalConfig.ServiceId;
+            SiloIdentity = localSiloDetails.SiloAddress.ToLongString();
             GrainFactory = grainFactory;
             TimerRegistry = timerRegistry;
             ReminderRegistry = reminderRegistry;
             StreamProviderManager = streamProviderManager;
             ServiceProvider = serviceProvider;
         }
 
-        public Guid ServiceId { get; private set; }
```
veikkoeeva: :)

src/Orleans/Serialization/SerializationTestEnvironment.cs
```diff
@@ -0,0 +1,34 @@
+namespace Orleans.Serialization
+{
+    using System.Collections.Generic;
+    using System.Reflection;
+
+    using Orleans.Runtime;
+
+    public class SerializationTestEnvironment
```
veikkoeeva: What is the purpose of this class?

src/OrleansCounterControl/CounterControl.cs
```diff
@@ -177,7 +177,7 @@ private static void RegisterWindowsPerfCounters(bool useBruteForce)
                 if (GrainTypeManager.Instance == null)
                 {
                     var loader = new SiloAssemblyLoader(new Dictionary<string, SearchOption>());
-                    var typeManager = new GrainTypeManager(false, null, loader); // We shouldn't need GrainFactory in this case
+                    var typeManager = new GrainTypeManager(false, loader, DefaultPlacementStrategy.GetDefaultForTesting());
```
veikkoeeva: On the surface this looks fishy without comments. Tests here? What?

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -19,6 +19,8 @@
 
 namespace Orleans.Runtime
 {
+    using Orleans.Runtime.Messaging;
```
veikkoeeva: Should this be before or after `namespace Orleans.Runtime`? I'm note sure if we should maintain consitency of keeping namespace directives on the other scope.

src/OrleansRuntime/Counters/SiloStatisticsManager.cs
```diff
@@ -12,18 +12,23 @@ internal class SiloStatisticsManager
...
             NetworkingStatisticsGroup.Init(true);
-            ApplicationRequestsStatisticsGroup.Init(globalConfig.ResponseTimeout);
+            ApplicationRequestsStatisticsGroup.Init(initializationParams.GlobalConfig.ResponseTimeout);
             SchedulerStatisticsGroup.Init();
             StorageStatisticsGroup.Init();
             runtimeStats = new RuntimeStatisticsGroup();
-            logStatistics = new LogStatistics(nodeConfig.StatisticsLogWriteInterval, true);
-            MetricsTable = new SiloPerformanceMetrics(runtimeStats, nodeConfig);
-            countersPublisher = new CountersStatistics(nodeConfig.StatisticsPerfCountersWriteInterval);
+            this.logStatistics = new LogStatistics(initializationParams.NodeConfig.StatisticsLogWriteInterval, true);
```
veikkoeeva: It feels like these could be give as parameters in the future. As in not creating instances in constructor using `new` and configuration options, but giving the instances via constructor and moving the configurations closer to the source. Hene "unmasking" the process.

src/OrleansRuntime/GrainDirectory/MultiClusterRegistration/RegistrarManager.cs
```diff
@@ -4,25 +4,24 @@
...
-        public static void InitializeGrainDirectoryManager(LocalGrainDirectory router, int numRetriesForGSI)
+        
+        public RegistrarManager(GrainDirectoryPartition directoryPartition, GlobalSingleInstanceActivationMaintainer gsiActivationMaintainer, GlobalConfiguration globalConfig, Logger logger)
         {
-            Instance = new RegistrarManager();
-            Instance.Register<ClusterLocalRegistration>(new ClusterLocalRegistrar(router.DirectoryPartition));
-            Instance.Register<GlobalSingleInstanceRegistration>(new GlobalSingleInstanceRegistrar(router.DirectoryPartition, router.Logger, router.GsiActivationMaintainer, numRetriesForGSI));
+            this.Register<ClusterLocalRegistration>(new ClusterLocalRegistrar(directoryPartition));
+            this.Register<GlobalSingleInstanceRegistration>(
+                new GlobalSingleInstanceRegistrar(
```
veikkoeeva: As in `SiloStatisticsManager`: It feels like these could be give as parameters in the future. As in not creating instances in constructor using new and configuration options, but giving the instances via constructor and moving the configurations closer to the source. Hene "unmasking" the process. If this makes sense, maybe we could add a comment (and issue?) about it. If not, ignore this.

There are probably other places this line of reasoning could be added.

src/OrleansRuntime/Placement/IPlacementDirector.cs
```diff
@@ -0,0 +1,25 @@
+using System.Threading.Tasks;
+
+namespace Orleans.Runtime.Placement
+{
+    /// <summary>
+    /// Interface for placement directors.
+    /// </summary>
+    internal interface IPlacementDirector
+    {
+        Task<PlacementResult> OnSelectActivation(
```
veikkoeeva: Proactively dropping the `Async` suffx for internals. :)

src/OrleansRuntime/Placement/PlacementDirectorsManager.cs
```diff
@@ -1,46 +1,40 @@
...
-            Instance.Register<ActivationCountBasedPlacement, ActivationCountPlacementDirector>();
-
-            var acDirector = (ActivationCountPlacementDirector)Instance.directors[typeof(ActivationCountBasedPlacement)];
-            acDirector.Initialize(globalConfig);
-
-            Instance.defaultPlacementStrategy = PlacementStrategy.GetDefault();
-            Instance.clientObserversPlacementDirector = new ClientObserversPlacementDirector();
+            this.serviceProvider = services;
+            this.defaultPlacementStrategy = defaultPlacementStrategy.PlacementStrategy;
+            this.clientObserversPlacementDirector = new ClientObserversPlacementDirector();
```
veikkoeeva: Would it make sense to move this to a constructor parameter?

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -25,14 +23,15 @@
...
+using Orleans.Runtime.TestHooks;
 using Orleans.Serialization;
 using Orleans.Storage;
 using Orleans.Streams;
 using Orleans.Timers;
-using Orleans.Runtime.TestHooks;
 
 namespace Orleans.Runtime
 {
+    using Microsoft.Extensions.DependencyInjection.Extensions;
```
veikkoeeva: Inner or outer?

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -623,17 +647,17 @@ await scheduler.QueueTask(() => streamProviderUpdateAgent.UpdateStreamProviders(
         private void ConfigureThreadPoolAndServicePointSettings()
         {
 #if !NETSTANDARD_TODO
-            if (nodeConfig.MinDotNetThreadPoolSize > 0)
+            if (LocalConfig.MinDotNetThreadPoolSize > 0)
```
veikkoeeva: What if this is zero? What should happen then? Is it an illegal state?

src/OrleansRuntime/Silo/SiloInitializationParameters.cs
```diff
@@ -0,0 +1,60 @@
+namespace Orleans.Runtime
+{
+    using Orleans.Runtime.Configuration;
```
veikkoeeva: Inner our outer?

src/OrleansTelemetryConsumers.Counters/OrleansPerformanceCounterInstaller.cs
```diff
@@ -35,7 +37,7 @@ public OrleansPerformanceCounterInstaller()
             if (GrainTypeManager.Instance == null)
             {
                 var loader = new SiloAssemblyLoader(new Dictionary<string, SearchOption>());
-                var typeManager = new GrainTypeManager(false, null, loader); // We shouldn't need GrainFactory in this case
+                var typeManager = new GrainTypeManager(false, loader, DefaultPlacementStrategy.GetDefaultForTesting());
```
veikkoeeva: Like in other places, this `.GetDefaultForTesting()` feels a bit of smell that asks for explanation.

src/OrleansRuntime/ConsistentRing/VirtualBucketsRingProvider.cs
```diff
@@ -5,6 +5,8 @@
 
 namespace Orleans.Runtime.ConsistentRing
 {
+    using Orleans.Runtime.Configuration;
```
veikkoeeva: The same note as in `Catalog`. `Should this be before or after namespace Orleans.Runtime? I'm note sure if we should maintain consitency of keeping namespace directives on the other scope.

ReubenBond: Good point. I just let ReSharper/StyleCop do what it wants most of the time and only revert their autoformatting if it's a minor change. I can revert bits if people have a strong opinion.

src/OrleansRuntime/Silo/Silo.cs
ReubenBond: I think the convention of inner is a carry-over from `#include` in C++, but it's the StyleCop way. I can undo

src/OrleansRuntime/Placement/IPlacementDirector.cs
ReubenBond: This is one of the first places I'd like to see us use `ValueTask<T>` when it's available, but I didn't drop the suffix - it was never there :)

src/Orleans/Serialization/SerializationTestEnvironment.cs
ReubenBond: Explained in the PR description, it's just to make it easier to initialize SerializationManager given some parts which were previously static are no longer static.

I hope to get rid of most of this, but in a later PR where SerializationManager is no longer static.

src/Orleans/Runtime/GrainInterfaceMap.cs
ReubenBond: So we would inject GrainInterfaceMap instead of DefaultPlacementStrategy?

src/OrleansCounterControl/CounterControl.cs
ReubenBond: Yep this is supposed to be cleaned up later. I removed the method in the latest push. I replaced it with a sub-class. The placement strategy won't matter here anyhow - unless I have a faulty understanding.

src/OrleansRuntime/Counters/SiloStatisticsManager.cs
ReubenBond: You're right. This PR already ended up being larger in scope than I had hoped, though. It's definitely not supposed to reach the logical conclusion of the direction it starts us in - just lays some groundwork and gives us something which can be reviewed a little more quickly than a big(er) bang PR.

src/OrleansRuntime/Silo/SiloInitializationParameters.cs
ReubenBond: removed - it was not needed

src/OrleansRuntime/Silo/Silo.cs
ReubenBond: fixed

src/OrleansRuntime/GrainTypeManager/SiloAssemblyLoader.cs
```diff
@@ -10,12 +10,19 @@
 
 namespace Orleans.Runtime
 {
+    using Orleans.Runtime.Configuration;
```
ReubenBond: fixed

src/OrleansRuntime/Placement/IPlacementDirector.cs
veikkoeeva: I can give away I'd like to use in some places in the ADO.NET parts, also add `IAsyncEnumerable` and removing reflection parts (which Orleans operations do not use). :)

src/Orleans/Serialization/SerializationTestEnvironment.cs
veikkoeeva: Maybe extend the comment to class documentation? Just in case someone else comes to look at this in interrim.

src/OrleansRuntime/ConsistentRing/VirtualBucketsRingProvider.cs
jdom: Yeah, I was about to comment in other PRs too. You can configure ReSharper to put outside. I don't really care if inside or outside, but I care about consistency, so having all outside should be the way to go for now

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -212,29 +190,112 @@ internal Silo(string name, SiloType siloType, ClusterConfiguration config, ILoca
...
+                    services.TryAddSingleton<MembershipOracle>(
+                        sp =>
+                        sp.GetRequiredService<MembershipFactory>()
+                          .CreateMembershipOracle(sp.GetRequiredService<Silo>(), sp.GetRequiredService<IMembershipTable>()));
+                    services.TryAddExisting<IMembershipOracle, MembershipOracle>();
+                    services.TryAddExisting<ISiloStatusOracle, MembershipOracle>();
+                    services.TryAddSingleton<Func<ISiloStatusOracle>>(sp => () => sp.GetRequiredService<ISiloStatusOracle>());
+                    services.TryAddSingleton<MultiClusterOracleFactory>();
+                    services.TryAddSingleton<LocalReminderServiceFactory>();
+                    services.TryAddSingleton<SiloControl>();
```
jdom: hmmm, registering this and other system targets in the container can be extremely dangerous, since callers should never resolve them as is, and instead should get a reference to the system target via the internal grain factory.
Maybe instead of registering it here (which I assume it's only because you want constructor parameters to be injected at construction time), you can use `ActivatorUtilities.CreateInstance` when registering it as a system target, but never register it as a service.
The thing that would be legal to register in the container is `ISiloControl` which can ask the IInternalGrainFactory to get that system target, but not these concrete objects that should never be used from the container directly.

src/Orleans/Core/GrainFactory.cs
```diff
@@ -37,10 +25,28 @@ public class GrainFactory : IGrainFactory
...
+
+        /// <summary>
+        /// The runtime client.
+        /// </summary>
+        private readonly IRuntimeClient runtimeClient;
+
         // Make this internal so that client code is forced to access the IGrainFactory using the 
         // GrainClient (to make sure they don't forget to initialize the client).
-        internal GrainFactory()
+        public GrainFactory(IRuntimeClient runtimeClient, TypeMetadataCache typeCache)
```
jdom: Having GrainFactory depending on IRuntimeClient seems backwards, as like I mentioned IRuntimeClient IMO should just be a facade on top of other services. I would depend on the other services directly from here. Also, I would not let InsideRuntimeClient create and expose the IGrainFactory service, it should just get it from the provider, but create it somewhere else. Ideally (at some point in the future) there shouldn't be a need for our code to depend on IRuntimeClient, and only external code might use it as a simplified facade (realistically I don't know how practical that is given the current shape of the runtime client)

src/OrleansRuntime/Placement/PlacementDirectorsManager.cs
```diff
@@ -86,5 +83,20 @@ private PlacementDirector ResolveDirector(PlacementStrategy strategy)
             var director = ResolveDirector(strategy);
             return director.OnAddActivation(strategy, grain, context);
         }
+
+        private void ResolveIncludedStrategies()
```
ReubenBond: fixed, thanks

src/OrleansRuntime/Messaging/MessageCenter.cs
```diff
@@ -37,9 +37,13 @@ public bool TryDeliverToProxy(Message msg)
...
         public IMessagingConfiguration MessagingConfiguration { get; private set; }
 
-        public MessageCenter(IPEndPoint here, int generation, IMessagingConfiguration config, ISiloPerformanceMetrics metrics = null)
+        public MessageCenter(SiloInitializationParameters silo, NodeConfiguration nodeConfig, IMessagingConfiguration config, ISiloPerformanceMetrics metrics = null)
         {
-            Initialize(here, generation, config, metrics);
+            this.Initialize(silo.SiloAddress.Endpoint, nodeConfig.Generation, config, metrics);
+            if (nodeConfig.IsGatewayNode)
+            {
+                this.InstallGateway(nodeConfig.ProxyGatewayEndpoint);
```
ReubenBond: Good feedback, there are lots of places where I've been trying to cut the circular deps. I think we should address them in a later PR if possible. In this specific case, it does not look as though there will be an issue.

I think the code would be clearer if I were to inline both Initialize & InstallGateway into the constructor - how about you?

src/Orleans/Serialization/SerializationTestEnvironment.cs
ReubenBond: fixed, thanks

ReubenBond: I still dislike this, though - I will see if I can move it to a common test project.

src/Orleans/Core/GrainAttributes.cs
```diff
@@ -165,7 +165,7 @@ public abstract class PlacementAttribute : Attribute
 
             internal PlacementAttribute(PlacementStrategy placement)
             {
-                PlacementStrategy = placement ?? PlacementStrategy.GetDefault();
+                PlacementStrategy = placement;
```
ReubenBond: Fixed, thanks

src/Orleans/Core/GrainFactory.cs
ReubenBond: I agree that it's not the nicest. `IRuntimeClient` is an internal abstraction, but `IGrainFactory` is an external abstraction which exposes some of `IRuntimeClient`s methods. The issue is that OutsideRuntimeClient provides the implementation of CreateObjectReference & DeleteObjectReference. It holds a mapping of id -> obj internally and uses that mapping in those methods and when dispatching calls to those objects.

I could extract all of those methods into some kind of `ILocalObjectManager` or something. It would take a dep on `IMessageCenter` & logger, but I believe that's it. That way, `OutsideRuntimeClient` & `GrainFactory` would depend on `ILocalObjectManager`. On Silos, we could/should inject an implementation which throws for all methods (currently, InsideRuntimeClient throws InvalidOperationException).

If you'd like I can open a separate issue for that?

src/Orleans/IDs/TypeCodeMapper.cs
```diff
@@ -7,10 +7,10 @@ namespace Orleans.Runtime
     /// </summary>
     internal static class TypeCodeMapper
     {
-        internal static GrainClassData GetImplementation(Type interfaceType, string grainClassNamePrefix = null)
+        internal static GrainClassData GetImplementation(Type interfaceType, IRuntimeClient runtimeClient, string grainClassNamePrefix = null)
         {
             GrainClassData implementation;
-            IGrainTypeResolver grainTypeResolver = RuntimeClient.Current.GrainTypeResolver;
+            IGrainTypeResolver grainTypeResolver = runtimeClient.GrainTypeResolver;
```
ReubenBond: Fixed, kind-of. I can pass `runtimeClient.GrainTypeResolver` directly, but I cannot inject `IGrainTypeResolver` directly into `GrainFactory`, since on clients, `IGrainTypeResolver` is created during initialization (it involves a remote call to a silo).

I agree that `IRuntimeClient` is a bit too broad in scope - we should pare it down.

src/OrleansRuntime/Silo/Silo.cs
ReubenBond: Very good point. I'll use `ActivatorUtilities` as you suggest.

Later on, I'll likely change this for IMembershipOracle so that if the user has injected their own implementation (eg, my ServiceFabric one), then this will not register  the default IMembershipOracle implementation as a SystemTarget. Doesn't matter yet.

src/Orleans/Core/IInternalGrainFactory.cs
```diff
@@ -0,0 +1,51 @@
+namespace Orleans
+{
+    using System;
+
+    using Orleans.Runtime;
+
+    /// <summary>
+    /// The internal grain factory interface.
+    /// </summary>
+    internal interface IInternalGrainFactory : IGrainFactory
```
ReubenBond: We should talk about what issues the abstraction had

src/Orleans/Serialization/IlDelegateBuilder.cs
```diff
@@ -17,6 +17,7 @@ internal class ILDelegateBuilder<TDelegate>
 
         /// <summary>Creates a new instance of the <see cref="ILDelegateBuilder{TDelegate}"/> class.</summary>
         /// <param name="fields">Field builder.</param>
+        /// <param name="fields">The field builder.</param>
```
ReubenBond: ah bad merge, thanks

src/OrleansRuntime/Placement/PlacementDirectorsManager.cs
```diff
@@ -1,46 +1,43 @@
...
+    using System.Collections.Concurrent;
+
+    using Microsoft.Extensions.DependencyInjection;
+
     internal class PlacementDirectorsManager
     {
-        private readonly Dictionary<Type, PlacementDirector> directors = new Dictionary<Type, PlacementDirector>();
-        private PlacementStrategy defaultPlacementStrategy;
-        private ClientObserversPlacementDirector clientObserversPlacementDirector;
+        private readonly ConcurrentDictionary<Type, IPlacementDirector> directors = new ConcurrentDictionary<Type, IPlacementDirector>();
```
ReubenBond: I tend to favor 'obviously not deficient' over 'not obviously deficient', so ConcurrentDictionary whenever I'm not completely certain that something will only ever be accessed in a single-threaded fashion. If you prefer, I can put a lock on it and use regular Dictionary

src/OrleansRuntime/Placement/RandomPlacementDirector.cs
```diff
@@ -4,11 +4,11 @@
 
 namespace Orleans.Runtime.Placement
 {
-    internal class RandomPlacementDirector : PlacementDirector
+    internal class RandomPlacementDirector : IPlacementDirector<RandomPlacement>
     {
         private readonly SafeRandom random = new SafeRandom();
 
-        internal override async Task<PlacementResult> OnSelectActivation(
+        public virtual async Task<PlacementResult> OnSelectActivation(
```
ReubenBond: Because there are subclasses of RandomPlacementDirector which override these methods

src/OrleansRuntime/Silo/Silo.cs
jdom: yeah, that makes sense. They should be registered in the way that the local silo accesses them, and not the system target interface (as an IAddressable) itself

src/Orleans/Serialization/SerializationTestEnvironment.cs
```diff
@@ -0,0 +1,34 @@
...
+            this.GrainFactory = new GrainFactory(null, this.TypeMetadataCache);
+        }
+
+        public void InitializeForTesting(List<TypeInfo> serializationProviders = null, TypeInfo fallbackType = null)
+        {
+            SerializationManager.InitializeForTesting(serializationProviders, fallbackType);
+            this.AssemblyProcessor.Initialize();
+        }
+
+        public static SerializationTestEnvironment Initialize(List<TypeInfo> serializationProviders = null, TypeInfo fallbackType = null)
```
jdom: >List<TypeInfo> serializationProviders = null, TypeInfo fallbackType = null [](start = 62, length = 74)

I see this being called with and explicit Initialize(null, null) most of the times. Shouldn't they use the default optional parameters implicitly instead when tests do not really care about these values at all?

test/TesterInternal/HostedTestClusterBase.cs
```diff
@@ -7,13 +7,16 @@
 
 namespace UnitTests
 {
+    using System.Collections.Generic;
+    using System.Reflection;
```
jdom: apart from being inside the namespace, I believe these are not needed for the change in this file. I did notice similar a change in several other test files.
EDIT: confirmed by opening the code, these references can go away

rapoth:
@ReubenBond Thank you for this PR! 

As per https://github.com/dotnet/orleans/issues/96 to support custom placement strategies through application code, would you also be able to make the internal APIs for PlacementDirectorsManager, PlacementStrategy, and PlacementDirector public?

src/Orleans/Core/GrainFactory.cs
```diff
@@ -56,14 +62,13 @@ internal GrainFactory()
...
+        /// <returns>A reference to the specified grain.</returns>
         public TGrainInterface GetGrain<TGrainInterface>(Guid primaryKey, string grainClassNamePrefix = null) where TGrainInterface : IGrainWithGuidKey
         {
-            return Cast<TGrainInterface>(
-                GrainFactoryBase.MakeGrainReference_FromType(
-                    baseTypeCode => TypeCodeMapper.ComposeGrainId(baseTypeCode, primaryKey, typeof(TGrainInterface)),
-                    typeof(TGrainInterface),
-                    grainClassNamePrefix));
+            Type interfaceType = typeof(TGrainInterface);
+            var implementation = this.GetGrainClassData(interfaceType, grainClassNamePrefix);
```
jason-bragg: As you stated, this makes PR's harder to read because we don't have the 'tooling' in the review process.  Having said that, this is just a readability issues, so I'll not push the issue.  Change if you feel inclined, or leave as is..  Main goal for me was to communicate the (unofficial) standard so you were aware of it in future changes.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -212,29 +190,109 @@ internal Silo(string name, SiloType siloType, ClusterConfiguration config, ILoca
...
-                LocalDataStoreInstance.LocalDataStore = keyStore;
-            }
-
+            
             // Configure DI using Startup type
-            bool usingCustomServiceProvider;
-            Services = StartupBuilder.ConfigureStartup(nodeConfig.StartupTypeName, out usingCustomServiceProvider);
+            this.Services = StartupBuilder.ConfigureStartup(
+                this.LocalConfig.StartupTypeName,
+                (services, usingCustomServiceProvider) =>
```
jason-bragg: I have concerns about adding all of these classes to DI as singletons.
Singletons should be thread safe, as we've not control over who users them.  I do not believer many of these classes are thread safe.
Many of these singletons have very specific initialization logic that must be run before they are usable, and little to no testing done on what happens if they are used before they are initialized.
It is not clear to me which of these should remain singletons.

I don't think these concerns should block the change, I just want us to be aware of the risk we are introducing.

jason-bragg:
I am of the opinion that this is a risky change, but worth it.  In themselves, I think these changes are safe, the risk is more in what they enable.

LGTM

jdom:
@ReubenBond is it OK to merge or you have some extra changes incoming, or test results that you are waiting? It's probably OK, but since it's a big PR I wanted to double-check with you before clicking Merge

PR #2348, opened 2016-10-27
.NET Standard: Remove prerelease BinaryFormatter & fix dependent code

ReubenBond:
The goal here is to remove the pre-release `System.Runtime.Serialization.Formatters` package in the vNext solution. Doing so involves removing/#ifdef-ing uses of `ISerializable`, `BinaryFormatter`, `[Serializable]`, and `[NonSerialized]`.
- Adds a shim for `[Serializable]`
- For the purposes of serialization/codegen, determines whether or not a type is marked as serializable based upon the presence of an attribute called "SerializableAttribute" rather than the concrete `System.SerializableAttribute` type or the `type.IsSerializable` property.
- Similarly for `[NonSerialized]`.
- Because `GlobalConfiguration` & co are no longer serializable from the perspective of .NET's, they cannot traverse AppDomain boundaries, so [Wire](https://github.com/akkadotnet/Wire) is used to serialize them into byte arrays in `OrleansTestingHost` when creating instances of `AppDomainSiloHost`. Note that we could use our own serialization infrastructure, but since `SerializationManager` has some static fields, we should avoid this - otherwise test results might be tainted. We could also try to cleanly uninitialize it and all related classes before constructing the `Silo`.
- Removed pre-release `System.Runtime.Serialization.Formatters` NuGet package.

### DIFF
--- a/src/Orleans/CodeGeneration/GeneratedAssembly.cs
+++ b/src/Orleans/CodeGeneration/GeneratedAssembly.cs
@@ -1,9 +1,11 @@
 namespace Orleans.CodeGeneration
 {
+    using System;
+
     /// <summary>
     /// Represents a generated assembly.
     /// </summary>
-    [System.Serializable]
+    [Serializable]
     public class GeneratedAssembly
     {
         /// <summary>
--- a/src/Orleans/IDs/GuidId.cs
+++ b/src/Orleans/IDs/GuidId.cs
@@ -11,7 +11,10 @@ namespace Orleans.Runtime
     /// </summary>
     [Serializable]
     [Immutable]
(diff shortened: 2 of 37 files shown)


jdom:
@dotnet-bot test this please

sergeybykov:
@ReubenBond Please rebase.

jdom:
Since there is an alternative (using the new stable 4.3.0 version of that package) making we can postpone this decision, as it requires some other hacks in the meantime (such as serializing config to marshal through app domains). I think we should probably close this for now (but keep the branch in case we want to bring it back at some point)

sergeybykov:
I agree.

PR #2360, opened 2016-10-28
EventHubStreamProvider improvements

jason-bragg:
- Block writes on readonly stream providers.
- Support monitoring of initialization and shutdown of EventHub stream provider receivers.
- Support customization of EventHub stream provider direction.
- Fix incomplete EventHubStreamProvider receiver monitoring.  User could not specify custom monitoring.
- Support custom data comparer in EventHubStreamProvider cache.

### DIFF
--- a/src/Orleans/Serialization/ILSerializerGenerator.cs
+++ b/src/Orleans/Serialization/ILSerializerGenerator.cs
@@ -4,7 +4,6 @@ namespace Orleans.Serialization
     using System.Collections.Generic;
     using System.Linq;
     using System.Linq.Expressions;
-    using System.Net;
     using System.Reflection;
 
     using Orleans.Runtime;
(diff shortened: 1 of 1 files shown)


jason-bragg:
This should be included in 1.3.1

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
```diff
@@ -97,18 +121,33 @@ public EventHubMessage(CachedEventHubMessage cachedMessage)
...
+        /// </summary>
         public static readonly ICacheDataComparer<CachedEventHubMessage> Instance = new EventHubDataComparer();
 
-        public int Compare(CachedEventHubMessage cachedMessage, StreamSequenceToken token)
+        /// <summary>
+        /// Compare a cached message with a sequence token to determine if it message is before or after the token
+        /// </summary>
+        /// <param name="cachedMessage"></param>
+        /// <param name="streamToken"></param>
+        /// <returns></returns>
```
jdom: Either add description for the parameter (preferred) or remove them entirely.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
```diff
@@ -44,6 +44,30 @@ public struct CachedEventHubMessage
     public class EventHubMessage
     {
         /// <summary>
+        /// Contructor
+        /// </summary>
+        /// <param name="streamIdentity"></param>
+        /// <param name="partitionKey"></param>
```
jdom: Not sure how important, but I've been suggested to use the [constructor doc string](http://stylecop.soyuz5.com/SA1642.html) here: https://github.com/dotnet/orleans/pull/2223#pullrequestreview-2441730
Also, please add description for the properties (or remove them entirely, but looks like we could use a description here).

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -79,6 +79,10 @@ public class EventHubAdapterFactory : IQueueAdapterFactory, IQueueAdapter, IQueu
         /// Create a queue mapper to map EventHub partitions to queues
         /// </summary>
         protected Func<string[], IEventHubQueueMapper> QueueMapperFactory { get; set; }
+        /// <summary>
+        /// Create a receiver monitor to report performance metrics.
+        /// </summary>
+        protected Func<string, string, Logger, IEventHubReceiverMonitor> ReceiverMonitorFactory { get; set; }
```
jdom: would be useful to explain the type parameters for the Func in the XML document, since the signature without argument names is not self explanatory.

PR #2375, opened 2016-10-31
Fix README Badges

ReubenBond:
The issue stats badges have been broken for a while and they don't seem to be coming back (site is 503'ing). The build status badge is also broken.

We could add build status for .NET Standard, but I feel we should wait until we are almost ready to release it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,8 @@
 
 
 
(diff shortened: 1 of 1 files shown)


gabikliot:
Looks much better now. Thanks Reuben!

PR #2378, opened 2016-10-31
Added queue flow controller that is triggered by silo load shedding.

jason-bragg:


src/OrleansServiceBus/OrleansServiceBus.csproj
```diff
@@ -78,6 +78,10 @@
       <Project>{0054DB14-2A92-4CC0-959E-A2C51F5E65D4}</Project>
       <Name>OrleansProviders</Name>
     </ProjectReference>
+    <ProjectReference Include="..\OrleansRuntime\OrleansRuntime.csproj">
+      <Project>{6FF2004C-CDF8-479C-BF27-C6BFE8EF93E0}</Project>
+      <Name>OrleansRuntime</Name>
+    </ProjectReference>
```
jdom: What's the implication to clients using event hub streams? Not necessarily complaining about not adding the dependency, but if we do, we either need to add a dependency in the nuspec, or dynamically test whether we are running on the silo before assuming that the Orleans.Runtime DLL is deployed (which is a Silo dll).

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -85,7 +85,7 @@ public Task Initialize(TimeSpan timeout)
             {
                 checkpointer = await checkpointerFactory(settings.Partition);
                 cache = cacheFactory(settings.Partition, checkpointer, baseLogger);
-                flowController = new AggregatedQueueFlowController(MaxMessagesPerRead) {cache};
+                flowController = new AggregatedQueueFlowController(MaxMessagesPerRead) {cache, new LoadShedQueueFlowController()};
```
jdom: should this not be incorporated when running in the client?

src/OrleansServiceBus/OrleansServiceBus.csproj
jason-bragg: Orleans runtime was already a dependency in the nuget because of Orleans.providers, so this should not add any extra dependencies.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
jason-bragg: Receivers are never run on the client.  They are silo side only.

src/OrleansRuntime/Streams/LoadShedQueueFlowController.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="loadSheddingLimit"></param>
+        public LoadShedQueueFlowController(int loadSheddingLimit)
+        {
+            if (loadSheddingLimit < 0 || loadSheddingLimit > 100) throw new ArgumentOutOfRangeException(nameof(loadSheddingLimit), "Value must be between 0-100");
+            this.loadSheddingLimit = loadSheddingLimit != 0 ? loadSheddingLimit : int.MaxValue;
+        }
```
jdom: I'm concerned about this ambiguous constructor. Passing an int or a double would have very different behaviors that might be really easy to miss. Maybe you can collapse into a single constructor with a Boolean indicating it's a percentage of the silo load shedding limit or just the actual value?

gabikliot:
LGTM.

jdom:
@sergeybykov please include this one too in 1.3.1

sergeybykov:
Done.

PR #2381, opened 2016-11-02
.NET Standard: port over OrleansAzureUtils to vNext

xiazen:
1. Port over OrleansAzureUtils to vNext 
2. move all tests related to OrleansAzureUtils to TesterAzureUtils project, both in vNext and main solution
3. make sure/fix tests in TesterAzureUtils of vNext solution passes.


src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
```diff
@@ -109,8 +109,12 @@ public AzureQueueDataManager(string queueName, string deploymentId, string stora
                 var myQueue = queueOperationsClient.GetQueueReference(QueueName);
 
                 // Create the queue if it doesn't already exist.
-
+#if !NETSTANDARD_TODO
+                //BeginCreateIfNotExists and EndCreateIfNotExists Not supported yet. 
                 bool didCreate = await Task<bool>.Factory.FromAsync(myQueue.BeginCreateIfNotExists, myQueue.EndCreateIfNotExists, null);
+#else
+                bool didCreate = true;
```
jdom: As we talked in private, it makes more sense to not introduce those #if NETSTANDARD_TODO when the fix is trivial

jdom:
Let's also include tests that cover this functionality

vNext/src/Build.cmd
```diff
@@ -58,7 +58,7 @@ if "%builduri%" == "" set builduri=Build.cmd
 set PROJ=%CMDHOME%\Orleans.vNext.sln
 
 :: Restore nuget packages before building the solution
-"%MSBUILDEXE%" %CMDHOME%\..\..\src\Before.Orleans.sln.targets /p:SolutionPath=%PROJ%
+"%MSBUILDEXE%" %CMDHOME%\..\..\src\Before.Orleans.sln.targets /p:SolutionPath=%PROJ% /v:d
```
jdom: should revert

vNext/src/Orleans/project.json
```diff
@@ -24,6 +24,6 @@
     "System.Xml.XPath.XmlDocument": "4.0.1"
   },
   "frameworks": {
-    "netstandard1.5": { }
+    "netstandard1.5": {}
```
jdom: superfluous change

src/OrleansTestingHost/TestingSiloHost.cs
```diff
@@ -657,9 +657,8 @@ private SiloHandle StartOrleansSilo(Silo.SiloType type, TestingSiloOptions optio
         /// <param name="type">The type of the silo to deploy</param>
         /// <param name="options">The options to use for the silo</param>
         /// <param name="instanceCount">The instance count of the silo</param>
-        /// <param name="shared">The shared AppDomain to use</param>
         /// <returns>A handle to the deployed silo</returns>
-        public static SiloHandle StartOrleansSilo(TestingSiloHost host, Silo.SiloType type, TestingSiloOptions options, int instanceCount, AppDomain shared = null)
+        public static SiloHandle StartOrleansSilo(TestingSiloHost host, Silo.SiloType type, TestingSiloOptions options, int instanceCount)
```
xiazen: AppDomain shared = null never been used, so remove this param,

xiazen:
in vNext: 
Only failing tests in TesterAzureUtils: Tester.AzureUtils.Persistence.PersistenceProviderTests_Local.LoadClassByName

It fails because TypeUtils.ResolveType(String fullname) only can resolve type in Orleans assembly in vNext

In main solution, all tests passes on vso. See test results here: https://microsoftstudios343.visualstudio.com/Orleans/_build/xaml?_a=summary&buildId=10186

src/OrleansAzureUtils/Storage/RemindersTableManager.cs
```diff
@@ -118,55 +118,57 @@ private RemindersTableManager(Guid serviceId, string deploymentId, string storag
...
+            string filterExtendBeginAndEnd = TableQuery.CombineFilters(
+                    TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.GreaterThan, sBegin),
+                    TableOperators.Or,
+                    TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.LessThanOrEqual,
+                        sEnd));
+            string query1 = TableQuery.CombineFilters(
+                filterOnServiceIdStr, TableOperators.And, filterExtendBeginAndEnd);
+
+            var results = await ReadTableEntriesAndEtagsWithFilterStringAsync(query1);
+            return results.ToList();
```
jdom: I understand these are logically equivalent, but I'm pretty sure they are not equivalent in performance. The new approach will probably force Azure to do a full table scan, while the other will do 2 faster partial scans, as well as parallelizing them. Since there is no need to have a single query to do both (even if the new approach would be just twice as slower than the parallel calls, which is probably some more than that), I'd stick to the original approach. If you feel strongly about it, we should have the change in a separate isolated PR anyway, so that it's not lost in the sea of style/new APIs usage fixes.

src/OrleansTestingHost/TestingSiloHost.cs
jdom: Are any of these needed? Otherwise please undo as they seem unrelated to the PR. If there are tests that use TestingSiloHost, we need to update them to TestCluster before porting them to vNext. We already agreed that we are not porting TestingSiloHost to vNext, remember?
I think this unnecessary breaking change could affect some of our VSO tests that use the TestingSiloHost directly. But still, since we are going to deprecate TestingSiloHost, let's avoid it.

src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueBatchContainer.cs
```diff
@@ -88,7 +88,11 @@ internal static CloudQueueMessage ToCloudQueueMessage<T>(Guid streamGuid, String
         {
             var azureQueueBatchMessage = new AzureQueueBatchContainerV2(streamGuid, streamNamespace, events.Cast<object>().ToList(), requestContext);
             var rawBytes = SerializationManager.SerializeToByteArray(azureQueueBatchMessage);
-            return new CloudQueueMessage(rawBytes);
+
+            //new CloudQueueMessage(byte[]) not supported in netstandard, taking a detour to set it
+            var cloudQueueMessage = new CloudQueueMessage(null as string);
+            cloudQueueMessage.SetMessageContent(rawBytes);
+            return cloudQueueMessage;
```
xiazen: that method wasn't loaded in runtime. Maybe it's because the hack we were using. Will try it with the new tooling of Attila's change

src/OrleansAzureUtils/Providers/Streams/AzureQueue/SimpleAzureQueueAdapter.cs
```diff
@@ -71,14 +71,17 @@ public IQueueAdapterReceiver CreateReceiver(QueueId queueId)
             CloudQueueMessage cloudMsg = null;
             if (isBytes)
             {
-                cloudMsg = new CloudQueueMessage(data as byte[]);
-            }else if (isString)
+                //new CloudQueueMessage(byte[]) not supported in netstandard
+                cloudMsg = new CloudQueueMessage(null as string);
+                cloudMsg.SetMessageContent(data as byte[]);
```
xiazen: same top

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -467,57 +437,93 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
         /// </summary>
         /// <param name="predicate">Predicate function to use for querying the table and filtering the results.</param>
         /// <returns>Enumeration of entries in the table which match the query condition.</returns>
+#if !NETSTANDARD
+        //Microsoft.WindowsAzure.Storage.Table.Queryable is not supported in netstandard, so TableQuery with an Expression predicate 
+        // is not supported in netstandard, just keep this method in full dotnet
+        // for full dotnet users consistency
         public async Task<IEnumerable<Tuple<T, string>>> ReadTableEntriesAndEtagsAsync(Expression<Func<T, bool>> predicate)
```
xiazen: hmm.... but we put it in a public class and a public method. Not sure how they use it.

src/OrleansAzureUtils/Storage/RemindersTableManager.cs
xiazen: will do

xiazen:
@jdom about the last comment. The TestingSiloHost is used by some tests in TesterAzureUtils. I didn't remember we talked about not porting it over, but I know it is being deprecated. 

So two option we got here: 
1. don't port it over, remove those tests using it in TesterAzureUtils too and remember which tests are those.
2. port it over, migrate it off in both main and vNext. 

I prefer option 2 so that we don't keep a list somewhere about which tests still need to port over to vNext. 
But if you are strong about option 1, I can do it. 

About your concern of this change would break tests in VSO. I tested with NightlyFunc. It didn't break

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -386,10 +361,9 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
         /// <returns>Enumeration of all entries in the specified table partition.</returns>
         public Task<IEnumerable<Tuple<T, string>>> ReadAllTableEntriesForPartitionAsync(string partitionKey)
         {
-            Expression<Func<T, bool>> query = instance =>
-                instance.PartitionKey == partitionKey;
+            string query = TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.Equal, partitionKey);
```
xiazen: will do

xiazen:
And AppDomain shared = null is never used

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
xiazen: I'm not sure about this. This project is called AzureUtils and this class can be perfectly used as a DAO layer between the application and AzureTable

src/OrleansAzureUtils/Storage/RemindersTableManager.cs
xiazen: But we should definitely investigate the perf difference between those two approach

src/OrleansTestingHost/TestingSiloHost.cs
xiazen: @jdom about the last comment. The TestingSiloHost is used by some tests in TesterAzureUtils. I didn't remember we talked about not porting it over, but I know it is being deprecated. 
So two option we got here: 
1. don't port it over, remove those tests using it in TesterAzureUtils too and remember which tests are those.
2. port it over, migrate it off in both main and vNext. 
I prefer option 2 so that we don't keep a list somewhere about which tests still need to port over to vNext. 
But if you are strong about option 1, I can do it. 
About your concern of this change would break tests in VSO. I tested with NightlyFunc. It didn't break

src/OrleansAzureUtils/Hosting/AzureConfigUtils.cs
```diff
@@ -125,7 +127,8 @@ private static IEnumerable<DirectoryInfo> FindAppDirectoryLocations()
                 if (appRootPath != null)
                     locations.Add(new DirectoryInfo(appRootPath));
             });
-
+#if !NETSTANDARD_TODO
+            //System.Web namespace is deprecated in netstandard, need to find a alternative way to get {ServerRoot}
```
ReubenBond: I'm fine with leaving this as TODO for now, but I also believe we should simply drop the unavailable bits.

src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueBatchContainer.cs
ReubenBond: I'm also blaming "the hack" for the runtime exceptions while trying to port codegen - hopefully @attilah's change fixes everything and restores my faith in .NET Standard

src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
```diff
@@ -140,12 +139,13 @@ public AzureQueueDataManager(string queueName, string deploymentId, string stora
             {
                 // that way we don't have first to create the queue to be able later to delete it.
                 CloudQueue queueRef = queue ?? queueOperationsClient.GetQueueReference(QueueName);
-                var exists = Task<bool>.Factory.FromAsync(queueRef.BeginExists, queueRef.EndExists, null);
+                var exists = queueRef.ExistsAsync();
                 if (await exists)
```
ReubenBond: Doesn't matter, but this could be inlined without losing clarity, or the await could be moved up.

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -356,11 +332,10 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
...
                     string queryString = TableQueryFilterBuilder.MatchPartitionKeyAndRowKeyFilter(partitionKey, rowKey);
                     var query = new TableQuery<T>().Where(queryString);
-                    TableQuerySegment<T> segment = await Task.Factory
-                        .FromAsync<TableQuery<T>, TableContinuationToken, TableQuerySegment<T>>(
-                            tableReference.BeginExecuteQuerySegmented,
-                            tableReference.EndExecuteQuerySegmented<T>, query, null, null);
+                    TableQuerySegment<T> segment = await tableReference.ExecuteQuerySegmentedAsync(query, null); 
+
                     retrievedResult = segment.Results.SingleOrDefault();
+
```
ReubenBond: nit: stylecop would complain about this blank line before a }

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
ReubenBond: @xiazen hopefully no one is using an Orleans-specific Azure util library as a general purpose lib. If there are any Orleans libs based on this, we can help them to implement fixes as issues arise.

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -467,57 +437,93 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
...
+        /// <param name="filter">filter string to use for querying the table and filtering the results.</param>
+        /// <returns>Enumeration of entries in the table which match the query condition.</returns>
+        public async Task<IEnumerable<Tuple<T, string>>> ReadTableEntriesAndEtagsWithFilterStringAsync(string filter)
+        {
+            const string operation = "ReadTableEntriesAndEtags";
+            var startTime = DateTime.UtcNow;
+            try
+            {
+                TableQuery<T> cloudTableQuery = filter == null
+                    ? new TableQuery<T>()
```
ReubenBond: This could be made into a private static readonly field to save on allocations.

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
xiazen: no, no Orleans lib based on this method. Yeah I guess I can remove it.

src/Orleans/Serialization/OrleansJsonSerializer.cs
```diff
@@ -61,10 +64,13 @@ public static JsonSerializerSettings UpdateSerializerSettings(JsonSerializerSett
             if (config.Properties.ContainsKey(UseFullAssemblyNamesProperty))
             {
                 bool useFullAssemblyNames;
+#if !NETSTANDARD_TODO
+                //Newtonsoft.Json.JsonSerializerSettings.set_TypeNameAssemblyFormat not suppported in netstanard
                 if (bool.TryParse(config.Properties[UseFullAssemblyNamesProperty], out useFullAssemblyNames) && useFullAssemblyNames)
                 {
                     settings.TypeNameAssemblyFormat = FormatterAssemblyStyle.Full;
                 }
+#endif
```
xiazen: it was weird for me since the method builds well in compile, but cannot be loaded in runtime. 

Newtonsoft.Json suggest to use TypeNameAssemblyFormatHandling. since TypeNameAssemblyFormat is obsolete
see
https://github.com/JamesNK/Newtonsoft.Json/blob/17f87f53211af9645ab8b6947d79e9f4e0e3c6be/Src/Newtonsoft.Json/JsonSerializerSettings.cs#L185

Will take a closer look on Monday. Just tried to use it in vNext, compiler cannot find TypeNameAssemblyFormatHandling method...

src/OrleansAzureUtils/Hosting/AzureConfigUtils.cs
xiazen: Yeah I'm not sure how useful it is, the comment on it suggested that it is used for finding server root in web roles running in IIS web apps or IIS express.. 

More than happy to remove it if this block of code is obsolete

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -467,57 +437,93 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
...
+            {
+                CheckAlertSlowAccess(startTime, operation);
+            }
+        }
+        /// <summary>
+        /// Read data entries and their corresponding eTags from the Azure table.
+        /// </summary>
+        /// <param name="filter">Predicate function to use for querying the table and filtering the results.</param>
+        /// <returns>Enumeration of entries in the table which match the query condition.</returns>
+        private async Task<IEnumerable<Tuple<T, string>>> ReadTableEntriesAndEtagsWithTableQueryAsync(TableQuery<T> cloudTableQuery)
```
ReubenBond: If we're always returning a list, and especially given this is a private method, we should just return the concrete type: `List<T>` would be preferable. I see that the usages of this method convert it to a list anyway.

I'd understand if you consider this out of scope for this PR, since you're really just converting an existing method which follows this pattern.

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -467,57 +437,93 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
...
+
+                List<T> results = await AsyncExecutorWithRetries.ExecuteWithRetries(
+                    counter => executeQueryHandleContinuations(),
+                    AzureTableDefaultPolicies.MaxTableOperationRetries,
+                    (exc, counter) => AzureStorageUtils.AnalyzeReadException(exc.GetBaseException(), counter, TableName, Logger),
+                    AzureTableDefaultPolicies.TableOperationTimeout,
+                    backoff);
+
+                // Data was read successfully if we got to here                    
+                return results.Select(i => Tuple.Create(i, i.ETag)).ToList();
```
ReubenBond: Isn't `Tuple.Create(i, i.ETag)` a bit redundant?

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -467,57 +437,93 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
...
             {
-                CheckAlertSlowAccess(startTime, operation);
+                // Out of retries...
+                var errorMsg = $"Failed to read Azure storage table {TableName}: {exc.Message}";
+                if (!AzureStorageUtils.TableStorageDataNotFound(exc))
+                {
+                    Logger.Warn(ErrorCode.AzureTable_09, errorMsg, exc);
+                }
+                throw new OrleansException(errorMsg, exc);
             }
```
ReubenBond: The original version of this method had a `finally` block with a call to `CheckAlertSlowAccess`

test/TesterAzureUtils/project.json
```diff
@@ -0,0 +1,16 @@
...
+    "xunit.runner.visualstudio": "2.1.0",
+    "Xunit.SkippableFact": "1.2.14"
+  },
+  "frameworks": {
+    "net451": {}
+  },
+  "runtimes": {
+    "win-anycpu": {},
+    "win": {}
+  }
```
ReubenBond: I know the goalposts are moving, but should we be using the new csproj format, or is @attilah going to convert to that in a separate PR?

vNext/src/OrleansAzureUtils/project.json
```diff
@@ -0,0 +1,40 @@
...
+        },
+        "runtime": {
+          "lib/portable-net45+wp8+win8+wpa/System.Spatial.dll": {}
+        }
+      }
+    }
+  },
+  "dependencies": {
+    "Microsoft.NETCore.Portable.Compatibility": "1.0.1",
+    "NETStandard.Library": "1.6.0",
```
ReubenBond: (not an issue) We typically take a "latest version - 1" approach to dependencies. For .NET Standard, I believe we should adopt the latest version of every package before our initial release, because it's all moving so rapidly.

vNext/test/TesterInternal/App.config
```diff
@@ -0,0 +1,17 @@
...
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
+  </startup>
+  <runtime>
+    <ThrowUnobservedTaskExceptions enabled="false" />
+    <gcServer enabled="true" />
+    <gcConcurrent enabled="false" />
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-9.0.1.0" newVersion="9.0.1.0" />
```
ReubenBond: In other projects above, you're redirecting to `9.0.0.0`

test/TesterAzureUtils/project.json
xiazen: @attilah is going to submit me a PR to fix that ~

src/OrleansAzureUtils/Hosting/ServiceRuntimeWrapper.cs
```diff
@@ -183,11 +183,9 @@ private void Initialize()
...
 
                 // Microsoft.WindowsAzure.ServiceRuntime isn't loaded. We may be running within a web role or not in Azure.
-                // Trying to load by partial name, so that we are not version specific.
-                // Assembly.LoadWithPartialName has been deprecated. Is there a better way to load any version of a known assembly?
-                #pragma warning disable 618
-                assembly = Assembly.LoadWithPartialName("Microsoft.WindowsAzure.ServiceRuntime");
-                #pragma warning restore 618
+#pragma warning disable 618
+                assembly = Assembly.Load(new AssemblyName("Microsoft.WindowsAzure.ServiceRuntime"));
+#pragma warning restore 618
```
xiazen: Or we can do a brutal force scan to load with assembly full name of `Microsoft.WindowsAzure.ServiceRuntimewith ` all its version , and load the latest one. 

tested with Assembly.Load(new AssemblyName("Microsoft.WindowsAzure.ServiceRuntime, Version = 2.7.0.0, Culture = neutral, PublicKeyToken = 31bf3856ad364e35"));

it works....

sergeybykov: The whole point of `ServiceRuntimeWrapper` was to make the code work with any version of `Microsoft.WindowsAzure.ServiceRuntime`. Referencing a specific version will defeat the purpose of it.

xiazen: No not referring to a specific version, but to try load the one version which exists in the app directory through brutal force try load, since LoadWithPartialName is obsolete now. 

The code would look like:
```
List<string> serviceRuntimeFullNameOfAllVersions = []...
foreach(var serviceRuntimeFullname in serviceRuntimeFullNameOfAllVersions )
{
try
{
  Assembly.Load(new AssemblyName(serviceRuntimeFullname ));
  break;
}catch(Exception e)
{
log(e);
}
}
```

xiazen: discussed in person. Will find a more robust solution

xiazen: another solution ... let me know what you think of this. 
```
var roleRootDir = Environment.GetEnvironmentVariable("RoleRoot");
                string osMode = "x64";
                if (!Environment.Is64BitOperatingSystem)
                    osMode = "x86";
                assembly = Assembly.LoadFrom(roleRootDir + "\\base\\" + osMode + "\\Microsoft.WindowsAzure.ServiceRuntime.dll");
```

xiazen: hmm... how do you feel like using.. 

```
var roleRootDir = Environment.GetEnvironmentVariable("RoleRoot");
string osMode = "x64";
 if (!Environment.Is64BitOperatingSystem)
        osMode = "x86";
assembly = Assembly.LoadFrom(roleRootDir + "\\base\\" + osMode + "\\Microsoft.WindowsAzure.ServiceRuntime.dll"); 
```

I tried to look into queryAssemblyInfo from GAC, looks like it only support query with full assembly name also.

jdom: That is not supported in .NET Standard

src/OrleansAzureUtils/Storage/RemindersTableManager.cs
jdom: Why definitely? Just asking, because at least at hindsight it doesn't look like the new approach should be better than the old one. Still, sure, let's discuss in a different PR, but let's not change the behavior implicitly when porting

xiazen: If we can get the query results in one request to AzureTable then it's better, if it doesn't request a full table scan.

src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
```diff
@@ -140,12 +139,13 @@ public AzureQueueDataManager(string queueName, string deploymentId, string stora
...
                 {
-                    await Task.Factory.FromAsync(queueRef.BeginDelete, queueRef.EndDelete, null);
+                    await queueRef.DeleteIfExistsAsync();
```
jdom: good catch. I tried to make sure that every BeginXXX method was converted to XXXAsync exactly, but I missed this one. I agree this should just use DeleteIfExistsAsync only (and only log if it actually deleted something).

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
jdom: Even if the optimization is good, I would actually separate that into it's own PR... This is a big PR that should only do port, but no other changes. This one would be minor and was already called out, so feel free to push back, just wanted to make sure we do not do similar small optimizations in big changes that could go unnoticed, where sometimes they could be silently breaking.

ReubenBond: I spoke with @jdom and I believe we should avoid making this change now

src/OrleansAzureUtils/Hosting/ServiceRuntimeWrapper.cs
attilah: I had a chat with @sergeybykov about my idea to simply just use Type.GetType with assembly qualified name for .Net and .Net Core and if that way we can retrieve a type from ServiceRuntime then we're good if not then we bail out and not trying to load it.

If we're hosted in a cloud service that dll must be loaded by the host environment before invoking us.

@xiazen please try out this alternative.

attilah: @xiazen 

This worked from a cloud service:

var serviceRuntimeType = Type.GetType("Microsoft.WindowsAzure.ServiceRuntime.RoleEnvironment, Microsoft.WindowsAzure.ServiceRuntime", false);

So you need to update ServiceRuntimeWrapper to gather the type this way and it should be all good.

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
xiazen: it was moved up to ReadTableEntriesAndEtagsWithFilterString. But since we are merging these two back together, this doesn't matter anymore. :) 

In one word, it is there.

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
xiazen: I'm not sure I follow. which way are you suggesting?

vNext/test/TesterInternal/AssemblyInfo.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Orleans.CodeGeneration;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("TesterInternal")]
+[assembly: AssemblyDescription("Orleans Internal Tests")]
+[assembly: AssemblyConfiguration("")]
```
xiazen: indeed. :)

xiazen:
submitted new iteration: move to new tooling and fix PR feedback.

Still trying to fix some tests which started to fail after new tooling because of some issues with BinaryRedirect,,  but want to share changes earlier

src/OrleansAzureUtils/Hosting/ServiceRuntimeWrapper.cs
```diff
@@ -183,11 +183,9 @@ private void Initialize()
                 logger.Warn(ErrorCode.AzureServiceRuntime_NotLoaded, msg1);
 
                 // Microsoft.WindowsAzure.ServiceRuntime isn't loaded. We may be running within a web role or not in Azure.
-                // Trying to load by partial name, so that we are not version specific.
-                // Assembly.LoadWithPartialName has been deprecated. Is there a better way to load any version of a known assembly?
-                #pragma warning disable 618
-                assembly = Assembly.LoadWithPartialName("Microsoft.WindowsAzure.ServiceRuntime");
-                #pragma warning restore 618
+#pragma warning disable 618
+                assembly = Assembly.Load(new AssemblyName("Microsoft.WindowsAzure.ServiceRuntime, Version = 2.7.0.0, Culture = neutral, PublicKeyToken = 31bf3856ad364e35"));
```
ReubenBond: why is this version-specific now?

attilah: We had a conversation about this thing. It turns out that they are not changing the version number of this assembly with the new Azure SDK releases. It was not the case before. Since 2.7 (which is 1 year old) it is still the case, so we decided to remove the magic and just depend on this.

src/OrleansAzureUtils/Storage/RemindersTableManager.cs
```diff
@@ -118,53 +118,59 @@ private RemindersTableManager(Guid serviceId, string deploymentId, string storag
...
+                        sEnd));
+            string query1 = TableQuery.CombineFilters(
+                filterOnServiceIdStr, TableOperators.And, TableQuery.GenerateFilterCondition(nameof(ReminderTableEntry.PartitionKey), QueryComparisons.GreaterThan, sBegin));
+            string query2 = TableQuery.CombineFilters(
+                filterOnServiceIdStr, TableOperators.And, TableQuery.GenerateFilterCondition(nameof(ReminderTableEntry.PartitionKey), QueryComparisons.LessThanOrEqual, sEnd));
+
+            var results1 = ReadTableEntriesAndEtagsAsync(query1);
+            var results2 = ReadTableEntriesAndEtagsAsync(query2);
+            IEnumerable<Tuple<ReminderTableEntry, string>>[] results = await Task.WhenAll(results1, results2);
+            return results[0].Concat(results[1]).ToList();
```
ReubenBond: Might be crazy, but I'm pretty sure this is wrong. It's equivalent to a SQL statement like:
```SQL
SELECT * WHERE ParitionKey > "ServiceId_" AND PartitionKey <= "<ServiceId>`" AND PartitionKey >= <LowerBound>
UNION
SELECT * WHERE ParitionKey > "ServiceId_" AND PartitionKey <= "<ServiceId>`" AND PartitionKey <= <UpperBound>​​​
```
Which is equivalent (ignoring duplicates) to:
```SQL
SELECT * WHERE ParitionKey >= "ServiceId_" AND PartitionKey <= "<ServiceId>`"
```

src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
```diff
@@ -140,12 +139,12 @@ public AzureQueueDataManager(string queueName, string deploymentId, string stora
             {
                 // that way we don't have first to create the queue to be able later to delete it.
                 CloudQueue queueRef = queue ?? queueOperationsClient.GetQueueReference(QueueName);
-                var exists = Task<bool>.Factory.FromAsync(queueRef.BeginExists, queueRef.EndExists, null);
-                if (await exists)
+                if (queueRef != null)
```
jdom: this check isn't valid, as the reference cannot be null

vNext/src/OrleansRuntime/error.txt
```diff
@@ -0,0 +1,11 @@
+The process cannot access the file 'D:\microsoft\orleans-xiazen\vNext\src\OrleansRuntime\obj\Debug\netstandard1.5\Generated\OrleansRuntime.codegen.cs' because it is being used by another process.
```
ReubenBond: delete :) We merged a fix so that these are no longer generated, so it wont be a problem any longer

src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
```diff
@@ -140,12 +139,12 @@ public AzureQueueDataManager(string queueName, string deploymentId, string stora
...
                 {
-                    await Task.Factory.FromAsync(queueRef.BeginDelete, queueRef.EndDelete, null);
+                    await queueRef.DeleteIfExistsAsync();
                     logger.Info(ErrorCode.AzureQueue_03, "Deleted Azure Queue {0}", QueueName);
```
jdom: only log the message if delete succeeded (the boolean value returned by DeleteIfExistsAsync)

vNext/src/OrleansRuntime/error.txt
xiazen: will do , sign I must be really messy last Wednesday. This is not good to include in !

vNext/src/OrleansAzureUtils/OrleansAzureUtils.csproj
```diff
@@ -0,0 +1,42 @@
...
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="App.config" />
+    <Compile Include="**\*.cs" />
+    <EmbeddedResource Include="**\*.resx" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="..\Build\GlobalAssemblyInfo.cs" />
+    <Compile Include="..\..\..\src\OrleansAzureUtils\**\*.cs" Exclude="..\..\..\src\OrleansAzureUtils\Properties\*.cs" />
```
ReubenBond: Maybe it's not going to affect this assembly because it doesn't have build-time codegen, but there are other C# files generated by the .NET tooling (I don't know their purpose) which should be excluded. You can exclude `..\..\..\src\OrleansAzureUtils\obj\**\*.cs;..\..\..\src\OrleansAzureUtils\bin\**\*.cs` to get rid of them.

src/OrleansAzureUtils/Storage/AzureTableDataManager.cs
```diff
@@ -503,8 +469,9 @@ public AzureTableDataManager(string tableName, string storageConnectionString, L
 
                     // Data was read successfully if we got to here                    
                     return results.Select(i => Tuple.Create(i, i.ETag)).ToList();
-                }
-                catch (Exception exc)
+
+            }
+            catch (Exception exc)
```
jdom: indenting weirdness

vNext/src/NuGet/Microsoft.Orleans.OrleansAzureUtils.nuspec
```diff
@@ -0,0 +1,36 @@
...
+        <dependency id="WindowsAzure.Storage" version="7.2.1" />
+        <dependency id="System.Net.Requests" version="4.0.11" />
+      </group>
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="OrleansAzureUtils.dll" target="lib\netstandard1.5" />
+    <file src="OrleansAzureUtils.pdb" target="lib\netstandard1.5" />
+    <file src="OrleansAzureUtils.xml" target="lib\netstandard1.5" />
+    <file src="$SRC_DIR$OrleansAzureUtils\**\*.cs" target="src" />
```
ReubenBond: If possible, the src files which are in the obj & bin folders should be explicitly excluded using `exclude="$SRC_DIR$OrleansAzureUtils\obj\**\*.cs;$SRC_DIR$OrleansAzureUtils\bin\**\*.cs;". I'm not sure if this will cause issues, but just in case.

test/TesterAzureUtils/project.json
jdom: this file should be gone now after you rebased

test/TesterInternal/TestStoreGrainState.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Globalization;
+using Orleans;
+using TestExtensions;
+
+[Serializable]
+public class TestStoreGrainState
```
jdom: please put inside some namespace

vNext/src/NuGet/Microsoft.Orleans.OrleansAzureUtils.nuspec
```diff
@@ -0,0 +1,36 @@
...
+    <authors>Microsoft</authors>
+    <owners>Microsoft,Orleans</owners>
+    <projectUrl>https://github.com/dotnet/Orleans</projectUrl>
+    <licenseUrl>https://github.com/dotnet/Orleans#license</licenseUrl>
+    <iconUrl>https://raw.githubusercontent.com/dotnet/orleans/gh-pages/assets/logo_128.png</iconUrl>
+    <summary>
+      Azure Utilities Library of Microsoft Orleans - OrleansAzureUtils.dll
+    </summary>
+    <description>
+      Library of utility types for Microsoft Azure of Microsoft Orleans.
```
ReubenBond: maybe something like "Support for hosting Orleans on Microsoft Azure.", similarly, the summary should say something like "Azure support for Microsoft Orleans"

vNext/src/OrleansRuntime/error.txt
jdom: remove this files

vNext/src/OrleansAzureUtils/OrleansAzureUtils.csproj
xiazen: @attilah  should we apply this to other project in vNext? since I guess this would be a common issue for all the projects. I

test/TesterAzureUtils/TesterAzureUtils.csproj
```diff
@@ -0,0 +1,160 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
ReubenBond: Tester is a weird name and we should never have chosen it. This project should be called "Tests.AzureUtils" or, using the normal convention, "OrleansAzureUtils.Tests". I'm fine with leaving this as-is, we can rename test projects later if we want.

vNext/test/TesterInternal/App.config
jdom: yes, this doesn't seem consistent. I believe Newtonsoft.Json assembly version is always 9.0.0.0 even if the nuget package is 9.0.1

vNext/src/OrleansAzureUtils/OrleansAzureUtils.csproj
ReubenBond: @xiazen I've included it in my `codegen-netstandard` branch

xiazen: You added it to all projects? 
if yes, then no worries, you are the champion ~

src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
xiazen: what if `queue `is null? then `queueOperationsClient.GetQueueReference `will not be called? then `queueRef `could be null.

src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
xiazen: why do we log `ErrorCode.AzureQueue_03` when delete succeed? 

should it log a success message when it succeed and log a failure message when the it failed(the Boolean is false)

jdom: no, it's just showing informational message when it actually deleted something... if the table wasn't there to begin with, nothing is logged

src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
jdom: if `queue` is null, then that's when `queueOperationsClient.GetQueueReference` is called. If it's non-null, it just assigns queue to queueRef

src/OrleansAzureUtils/Storage/AzureQueueDataManager.cs
xiazen: no I mean, should it be
```
 if (await queueRef.DeleteIfExistsAsync())
                {
                    logger.Info(ErrorCode.AzureQueue_03, "Deleted Azure Queue {0}", QueueName);
                }else
{
// log the failure message
}
```

vNext/src/OrleansAzureUtils/OrleansAzureUtils.csproj
ReubenBond: Yep, I did, I added it to everything I could see

vNext/src/OrleansAzureUtils/app.config
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <!-- redirect to Newtonsoft.Json 9.0.0.0 version because WindowsAzure.Storage doesn't play well with 9.0.1 version -->
```
jdom: this comment should no longer be correct, right?

jdom:
Looks good to me, @ReubenBond, if your comments were also addressed, please squash and merge

xiazen:
just tested with SampleApp
should add 
` <PackageTargetFallback>$(PackageTargetFallback);portable-net45+win8+wpa81+wp8</PackageTargetFallback>` to Orleans application's csproj file which take OrleansAzureUtils as an dependency 

should add this to release note I think. a gotcha

xiazen:
@ReubenBond  said LG to him. he told me

xiazen:
include me in release note PR for tech review.

Issue #2382, opened 2016-11-02
Inconsistent serialization of Grain References.


jamescarter-le:
Hello,

  I have found an issue where we cannot send an IGrain inside a wrapper message to a Grain, it falls over when trying to DeepCopy the object:

```
public class MyMessage {
  public ISendingGrain Grain { get; set; }
}

public interface IReceivingGrain : IGrainWithIntegerKey {
  Task ReceiveAGrain(MyMessage message);
  Task<MyMessage> ReturnAGrain(ISendingGrain grain);
}

public interface ISendingGrain : IGrainWithIntegerKey {
  Task DoTest();
}

public SendingGrain : Grain, ISendingGrain {

  public async Task DoTest() {
   var recvGrain = await GrainFactory.GetGrain<IReceivingGrain>(0);

   var message =  await recvGrain.SendAGrain(this); // This works.

   await recvGrain.ReceiveAGrain(new MyMessage {
     Grain = this
   }); // This fails.
  }
}
```
This fails, when trying to DeepCopy MyMessage it complains that 'this', the Grain that is not serializable, instead of serializing it into a IAddressable grain reference.

However, if you do this the other way round, return a MyMessage containing a Grain interface, this does return a GrainReference correctly.

Why would this be, is it intentional, or is it a bug?

jamescarter-le:
Closed by #2383

PR #2383, opened 2016-11-02
Check IAddressable before DeepCopy

jamescarter-le:
Modified CodeGenerator to check if the item being DeepCopied is an IAddressable before DeepCopying, converting to IGrainReference if so.

Details provided in https://github.com/dotnet/orleans/issues/2382

### DIFF
--- a/src/OrleansCodeGenerator/SerializerGenerator.cs
+++ b/src/OrleansCodeGenerator/SerializerGenerator.cs
@@ -706,23 +706,27 @@ namespace Orleans.CodeGenerator
                 }
 
                 // Addressable arguments must be converted to references before passing.
+                // IGrainObserver instances cannot be directly converted to references, therefore they are not included.
                 ExpressionSyntax deepCopyValueExpression;
                 if (typeof(IAddressable).IsAssignableFrom(this.FieldInfo.FieldType)
-                    && this.FieldInfo.FieldType.GetTypeInfo().IsInterface)
+                    && this.FieldInfo.FieldType.GetTypeInfo().IsInterface
+                    && !typeof(IGrainObserver).IsAssignableFrom(this.FieldInfo.FieldType))
                 {
                     var getAsReference = getValueExpression.Member(
                         (IAddressable grain) => grain.AsReference<IGrain>(),
                         this.FieldInfo.FieldType);
 
-                    // If the value is a Grain at runtime, convert it to a strongly-typed GrainReference
-                    // using value.AsReference<TInterface>().
-                    // C#: value is Grain ? value.AsReference<TInterface>() : value;
+                    // If the value is not a GrainReference, convert it to a strongly-typed GrainReference.
+                    // C#: !(value is GrainReference) ? value.AsReference<TInterface>() : value;
                     deepCopyValueExpression =
                         SF.ConditionalExpression(
-                            SF.BinaryExpression(
-                                SyntaxKind.IsExpression,
-                                getValueExpression,
-                                typeof(Grain).GetTypeSyntax()),
+                            SF.PrefixUnaryExpression(
+                                SyntaxKind.LogicalNotExpression,
+                                SF.ParenthesizedExpression(
+                                    SF.BinaryExpression(
+                                        SyntaxKind.IsExpression,
+                                        getValueExpression,
+                                        typeof(GrainReference).GetTypeSyntax()))),
                             SF.InvocationExpression(getAsReference),
                             getValueExpression);
                 }
--- a/src/OrleansCodeGenerator/SerializerGenerator.cs
+++ b/src/OrleansCodeGenerator/SerializerGenerator.cs
@@ -705,12 +705,29 @@ namespace Orleans.CodeGenerator
                     return getValueExpression;
                 }
 
+                ExpressionSyntax deepCopyValueExpression;
+
+                // Addressable arguments must be converted to references before passing.
+                if (typeof(IAddressable).IsAssignableFrom(this.FieldInfo.FieldType)
+                    && this.FieldInfo.FieldType.GetTypeInfo().IsInterface)
+                {
+                    deepCopyValueExpression =
+                        SF.ConditionalExpression(
+                            SF.BinaryExpression(SyntaxKind.IsExpression, getValueExpression, typeof(Grain).GetTypeSyntax()),
+                            SF.InvocationExpression(getValueExpression.Member("AsReference", this.FieldInfo.FieldType)),
+                            getValueExpression);
+                }
+                else
+                {
+                    deepCopyValueExpression = getValueExpression;
+                }
+
                 // Deep-copy the value.
                 Expression<Action> deepCopyInner = () => SerializationManager.DeepCopyInner(default(object));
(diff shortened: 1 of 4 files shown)


gabikliot:
I don't think this will work. You cannot know what is the real strong type of the reference you need to convert to. If you convert to I grain ref, it is a weak reference, without the type, which should not be exposed to the app.  You can do this automatically conversation only if you have the full type, but in this case you don't always have. If you don't, the safer approach is to throw. And in general, less auto magic.

jamescarter-le:
How is that different from passing the Grain via a parameter in the InvokeMethod call?

gabikliot:
In that case we can convert ONLY if the formal declared argument is specified and it is a strongly typed reference.

Basically, we never can, in general, except for a few cases when we can. Should we have instead not converted ever at all? Maybe. Less confusing, but less fluent in important cases. It was a trade-off.

jamescarter-le:
I'm not sure I follow where you are coming from, if we do not do the AsReference() call in the GrainReference generated code to convert the Grain to a Reference we cannot pass Grain's around - what would you do instead, pass the identifier to later be resolved using Grain factory?

That is not possible when passing an IGrainObserver, which currently can be passed, alongside Grains.

Maybe I'm missing an area of understanding, but the check to see if we should call AsReference before DeepCopy ensures the interface is an IAddressable. If it was declared as Object then this code would not get invoked and you would get the original behavior.

Thanks for your feedback, can you suggest an alternative approach to enable passing Grain's inside message structures, as we can do this when returning message structures. Perhaps you would like to disable this feature and not allow the returning of serializable classes containing Grain references?

gabikliot:
What I meant is AsRefernce<I grain type>.

The cleanest way is to just pass in your application code this.AsReference<igraintype>. That way no conversions are needed.

jamescarter-le:
Understood - this works as a workaround. I still don't understand why the Generated Code can't do this for us like it does for arguments?

src/OrleansCodeGenerator/SerializerGenerator.cs
```diff
@@ -705,12 +705,31 @@ public ExpressionSyntax GetGetter(ExpressionSyntax instance, bool forceAvoidCopy
                     return getValueExpression;
                 }
 
+                ExpressionSyntax deepCopyValueExpression;
+
+                // Addressable arguments must be converted to references before passing.
+                if (typeof(IAddressable).IsAssignableFrom(this.FieldInfo.FieldType)
+                    &&
+                    (typeof(Grain).IsAssignableFrom(this.FieldInfo.FieldType) ||
```
ReubenBond: If the field is a subclass of `Grain`, then we should not generate code which will assign a `GrainReference` to it. If the field itself is a `Grain`, then we're dead in the water.

We should only do this if the field type is an `interface` and *not* if it is a `Grain`

Of course, the emitted code below, where you perform the `copy.Field = input.Field is Grain ? input.Field.AsReference<T>() : input.Field`

Also worth noting that you're accessing the input field multiple times. Note that `getValueExpression` can emit a method call in the event that the field is private.

jamescarter-le:
Apologies, I did not squash before pushing.

ReubenBond:
Thank you, @jamescarter-le. Could you please also add a test? Apologies for not mentioning this earlier.

ReubenBond:
I'll add a test and push to your branch

ReubenBond:
@jamescarter-le could you please review the [latest commit](https://github.com/dotnet/orleans/pull/2383/commits/54bbea370808ca7c47c1dd6730978631939fe455) which I pushed to your branch? If it looks good to you, then I will merge once tests complete.

jamescarter-le:
@ReubenBond This looks good - thanks for your assistance.

ReubenBond:
@gabikliot I want to check that PR looks good to you before merging.
Thanks, @jamescarter-le!

gabikliot:
@ReubenBond , I am OK with that, but I think @sergeybykov should take a look. He implemented the automatic cast from `this` to grain reference in the past, and I remember there were issues with doing that when it was deep inside the object. He should have more context I think. If its good with him, its good with me.

sergeybykov:
The following condition doesn't seem right to me.
>```csharp
if (typeof(IAddressable).IsAssignableFrom(this.FieldInfo.FieldType) 
                && this.FieldInfo.FieldType.GetTypeInfo().IsInterface) 
```

If the field value is already a `GrainReference`, wouldn't we end up converting it again?

Also, this operation
>```csharp
(IAddressable grain) => grain.AsReference<IGrain>()
```

returns a weakly typed grain reference, right? How would it get converted back to a strongly typed one? My suspicion is that even though this code will avoid the serialization error, it won't work correctly and might even break scenarios that work today with grain references being passed within data structures.

ReubenBond:
Apologies for the poor formatting, I'm on mobile. W.r.t the condition: we
are checking the static type of the field, not the runtime type of the
value. The check should determine whether or not it's possible that a Grain
could be assigned where a GrainReference would be more appropriate.
Therefore we check that the static type is addressable and an interface.
Does that sound right?

W.r.t the expression, it's just a type safe manner of getting the
MethodInfo for AsReference, rather than using a string. This helps for when
we are trying to find who is using a given method in VS and in the event
that we refactor it. Maybe I'm being pedantic.

sergeybykov:
>Apologies for the poor formatting, I'm on mobile. W.r.t the condition: we
are checking the static type of the field, not the runtime type of the
value. The check should determine whether or not it's possible that a Grain
could be assigned where a GrainReference would be more appropriate.
Therefore we check that the static type is addressable and an interface.
Does that sound right?

It does, but the same will be true if a field of type `IFooGrain` is assigned a value of type `FooGrainReference`, correct? If so, then `AsReference()` will be called on the grain reference instead of serializing it as is.

sergeybykov:
>it's just a type safe manner of getting the
MethodInfo for AsReference, rather than using a string.

Would it generate code to invoke `AsReference<IGrain>` instead of `AsReference<IFooGrain>`?

ReubenBond:
> the same will be true if a field of type IFooGrain is assigned a value of type FooGrainReference, correct? If so, then AsReference() will be called on the grain reference instead of serializing it as is.

The emitted code will check to see if the field value `is Grain` at runtime.

ReubenBond:
Ought to be IFooGrain, the type argument is passed on line 716. I'll add better commenting before we merge

sergeybykov:
Oh, I see now. What about observers? I guess we can ignore system targets because they are internal.

ReubenBond:
I've added a comment in the latest commit, so if it looks good to you then please feel free to squash & merge :)

ReubenBond:
@sergeybykov regarding `IGrainObserver` & `SystemTarget`, I've pushed an update which inverts the logic so that we check `!(value is GrainReference)` instead of `value is Grain`. This should now support `SystemTarget`.

I've also added an additional condition to the guard so that we do not emit this code if the field derives from `IGrainObserver`, since they need to be converted into `GrainReference` explicitly.

sergeybykov:
Thank you, @jamescarter-le (and @ReubenBond)!

PR #2384, opened 2016-11-03
Update Consul-Deployment.md

meenie216:
The change here is only the inclusion of the network and gateway details in the silo config file.

Their absence confused me (easy to do - especially given my inexperience with Orleans).

I've also fixed the internal links in this doco - seems that most of the internal documentation links are missing Documentation in the paths


sergeybykov:
Thank you, @meenie216!

Issue #1383, opened 2016-02-05
Forcefully deactivate grains stuck in a request for too long


sergeybykov:
We've seem this happening multiple times - a grain call gets stuck forever either awaiting a `Task` that never completes or, very rarely, trying to retrieve `.Result` from a resolved `Task`.

The general problem here is that, whatever the reason for a grain to get stuck in a method call, it is forever unavailable for other calls, until the silo gets restarted. Hence the user, session, device the grain represents gets unavailable as the result. We've seen occasionally a grain being unavailable for a couple of days.

The proposal is to add logic for forceful deactivation of such grains either automatically or/and via an explicit command. This would allow a new activation of the stuck grain to be created to process future requests.

The automatic option could be such that we set a configurable threshold for request execution time, e.g. 10 minutes, and automatically deactivate grains that reach it. The explicit command option can be implemented via ManagementGrain

The potential risk of forceful deactivation with incomplete execution of request is that the grain may eventually resume execution of the method if the `Task` or a deadlock it got blocked on gets resolved. This may lead to a violation of the single-threading/single-activation guarantees.

Weighing the risk against the experience of grains getting unavailable in most such cases until the silo is rebooted, I think the pros outweigh the cons here.

ReubenBond:
Given Orleans controls the task scheduling, couldn't we ensure that the continuation is never scheduled in the case that the grain has been forcefully deactivated?

gabikliot:
I think it is important issue and we can definitely do a lot in this direction (and already have a lot of pieces available). Before I provide more details, just wanted to clarify:

>a grain call gets stuck forever awaiting a `Task` that never completes

Here you mean not a grain call, but some arbitrary library call, which does not have an enforced timeout?

> a grain call gets stuck forever ... to retrieve .Result from a resolved Task.

How is that even possible? How a call to `.Result` from a resolved `Task` can block forever?

@ReubenBond , yes of course we can do that, and already do exactly that for deactivated grains that by mistake left not-awaited ContinueWith blocks. What @sergeybykov means, beyond what you wrote, is the case when the grain code is stuck in the middle of some code, not awaiting a task but executing (or blocked on some synchronous system call for example), so you either have to preempt a thread, or deal with the fact that after the code unblocks, it will keep executing and we can't stop it, until it calls await again, at which point we can stop it.

sergeybykov:
>Here you mean not a grain call, but some arbitrary library call, which does not have an enforced timeout?

[...]

PR #2387, opened 2016-11-04
Blocked grains detection and mitigation

benjaminpetit:
Fix for #1383

A bug in a grain implementation can "block" a grain activation. An easy mitigation would be to detect this state when we try to enqueue a new message to this activation, and then destroy it.

This PR implement this workaround. When destroying the current activation, all pending messages will be forwarded to a new activation. The `OnDeactivateAsync` is also called, like a "normal" deactivation.


gabikliot:
I will not be able to review it until next week. So if you need it earlier, go ahead, otherwise I will review it next week. Need to be very careful in this code.

test/TesterInternal/OrleansRuntime/StuckGrainTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                    () => stuckGrain.NonBlockingCall().WithTimeout(TimeSpan.FromMilliseconds(500)));
+            }
+
+            // Wait so the first task will reach with DefaultCollectionAge timeout
+            await Task.Delay(TimeSpan.FromSeconds(3));
+
+            // No issue on this one
+            await stuckGrain.NonBlockingCall();
+
+            Assert.Equal(nonBlockingCalls, await stuckGrain.GetNonBlockingCallCounter());
```
ReubenBond: What would happen if `RunForever` acquired & indefinitely held a lock which `OnDeactivateAsync` also required?

ReubenBond:
I'm not yet very familiar with much of this code and my naivete is probably reflected by these questions.

When `OnDeactivateAsync` is called from the `catalog.DeleteActivations(..)` chain, does it actually execute, given that the activation is blocked?

What should happen if the activation actually blocks the thread (`.Wait()` or spinning forever) rather than returning a task which just never resolves? Maybe that's out of scope

test/TesterInternal/OrleansRuntime/StuckGrainTests.cs
benjaminpetit: Good question. The grain will be then still blocked, and I think another thread will be blocked too.

Another solution could be to just removing the activation from the grain directory, but we will not forward pending request (is this really necessary?)

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -389,11 +389,24 @@ private void EnqueueRequest(Message message, ActivationData targetActivation)
...
+            {
+                case ActivationData.EnqueueMessageResult.Success:
+                    // Great, nothing to do
+                    break;
+                case ActivationData.EnqueueMessageResult.ErrorInvalidActivation:
+                    ProcessRequestToInvalidActivation(message, targetActivation.Address, targetActivation.ForwardingAddress, "EnqueueRequest");
+                    break;
+                case ActivationData.EnqueueMessageResult.ErrorStuckActivation:
+                    // Avoid any new call to this activation
+                    targetActivation.SetState(ActivationState.Deactivating);
```
benjaminpetit: Why not?

test/TesterInternal/OrleansRuntime/StuckGrainTests.cs
sergeybykov: I think we *should not* try to do anything with the stuck activation. It should be enough to remove it from the directory, so that no new requests will be sent to it. If it gets unstuck, it'll just get collected.

I would be nice to try to forward enqueued requests, but it's just a nice to have because chances are they already timed out on the caller side by the time we detected that the activation is stuck.

benjaminpetit:
1. Yes it will, no lock is taken when `OnDeactivateAsync` is called. That means that we will run two method of the grain in "parallel" (not really, see 2.), which can be bad. But we are trying to mitigate a situation that should not happen...
2. It will not work. the EnqueueMessage is not even called for this activation in this case, I am trying to find why

sergeybykov:
Can we do the first step first - just remove stuck activation from the directory?

benjaminpetit:
Updated, and found the original issue for that: #1383

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -389,11 +389,22 @@ private void EnqueueRequest(Message message, ActivationData targetActivation)
...
+            {
+                case ActivationData.EnqueueMessageResult.Success:
+                    // Great, nothing to do
+                    break;
+                case ActivationData.EnqueueMessageResult.ErrorInvalidActivation:
+                    ProcessRequestToInvalidActivation(message, targetActivation.Address, targetActivation.ForwardingAddress, "EnqueueRequest");
+                    break;
+                case ActivationData.EnqueueMessageResult.ErrorStuckActivation:
+                    // Avoid any new call to this activation
+                    catalog.RemoveFromDirectory(targetActivation).Wait();
```
ReubenBond: I know this is annoying, but the `.Wait()` is a little worrisome here, since it can involve making a remote call.
It just bears a little discussion whether or not we want to do it this way or make the entire call chain async.

Maybe the answer is to make it block for now and change it to be async when `ValueTask<T>` is available so that we don't incur the Task overhead

ReubenBond:
LGTM other than the one comment.

src/OrleansRuntime/Catalog/ActivationData.cs
```diff
@@ -483,26 +483,38 @@ public int WaitingCount
...
+                    return EnqueueMessageResult.ErrorInvalidActivation;
                 }
-                // If maxRequestProcessingTime is never set, then we will skip this check
-                if (maxRequestProcessingTime.TotalMilliseconds > 0 && Running != null)
+                if (Running != null)
                 {
-                    // Consider: Handle long request detection for reentrant activations -- this logic only works for non-reentrant activations
                     var currentRequestActiveTime = DateTime.UtcNow - currentRequestStartTime;
-                    if (currentRequestActiveTime > maxRequestProcessingTime)
+                    if (currentRequestActiveTime > CollectionAgeLimit)
```
sergeybykov: I concur - better define a separate config parameter and default it to CollectionAgeLimit.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -205,6 +205,13 @@ internal void Start()
             gcTimer = t;
         }
 
+        internal async Task RemoveFromDirectory(ActivationData activationData)
+        {
+            // Avoid any new call to this activation
+            activationData.SetState(ActivationState.Invalid);
+            await directory.UnregisterAsync(activationData.Address, UnregistrationCause.Force);
```
sergeybykov: No need for this method to be async I think.

sergeybykov:
@gabikliot 
>I think it is very temping to say we are mitigating a concrete scenario, but that comes at expanse of leaving corpses. I prefer to leave liveness bug rather than introducing new safety (correctness) bug.

We are making this tradeoff based on the real life cases we are seeing, and the business impact they have. We just saw a couple of cases with Gear of influential players being locked out for hours and days because of what appears like a bug down in the guts of some native library that somehow corrupted the grain thread. 

We are picking a very conservative default of the activation collection time for detection, 2 hours by default. A legitimate case for an activation to be stuck anywhere near that time is really hard to imagine.

While your assumption of a possible misdetection cannot, of course, be disproved, I consider it highly unlikely because we are leveraging the same warning point that's been there for ages, and we've not seen false positives.

In the worst case scenario, we may violate the single-threading guarantee, which is already not a strong 100% guarantee.

In my assessment, this is a worthwhile tradeoff, weighing the reality of what we are seeing and the business impact vs. the theoretical possibility of violation of the single-threading guarantee.

gabikliot:
I am not against solving this. I am against solving this naively and partially.

>violate the single-threading guarantee which is already not a strong 100% guarantee.

I don't think this is correct. Single-threading guarantee IS 100% correct, as fat as I know.
We remove the activation from the scheduler before activating the new one, so no new turn for the old activation will run.

gabikliot:
Also, isn't the whole detection wrong? We detect for non-reentrant grains the overall time request executes. But did not you want thread stuck time?  That is, the detection should be in the scheduler. Did I misunderstood it?

sergeybykov:
It seems to me that we, at least for now, should explicit limit this to non-reentrant grains. Probably need to add a separate test with a reentrant grain.

I think it would also be prudent to call `Catalog.DeactivateActivations` for the stuck activation before removing it from the directory. It won't do anything for an activation stuck in the running state - will only schedule `DestroyActivationAsync` via `activationData.AddOnInactive()`, which in this case will never happen, but it'll protect us from a mistake or if the activation suddenly getting unstuck.

>Also, isn't the whole detection wrong? We detect for non-reentrant grains the overall time request executes. But did not you want thread stuck time? That is, the detection should be in the scheduler. Did I misunderstood it?

Here we are specifically addressing the case of a request that never (for a long time) completes - a `Task` isn't getting resolved, for example, due to an incorrect use of a TCS. A blocked thread may also manifest the same way, and this fix will make the grain stuck with it available for activation on a different silo. There's potentially a different, complementary, solution for the blocked thread issue in the Scheduler, for example, involving trying to abort the bad thread and adding a new one instead. But that is out of scope of this mitigation.

gabikliot:
I will review on the weekend.

benjaminpetit:
I did not implement the reentrant check because if the grain is reentrant, we will never call `ActivationData.EnqueueMessage` from the dispatcher, the request will be directly handled by `Dispatcher.HandleIncomingRequest`. Unless I am missing something?

I will add a test with a reentrant grain

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -205,6 +205,15 @@ internal void Start()
             gcTimer = t;
         }
 
+        internal void RemoveStuckGrainFromDirectory(ActivationData activationData)
+        {
+            DeactivateActivations(new List<ActivationData>() {activationData}).Ignore();
```
gabikliot: I think a more appropriate here is DeactivateActivationOnIdle. 
We might need to even write a different method, if we find a need for a bit different sequence, but as of now  it is the same sequence. If you don't like the misleading name, lets rename DeactivateActivationOnIdle into DeactivateOneActivation

gabikliot: you may need to refactor it a little bit, to move the  CounterStatistic.FindOrCreate(StatisticNames.CATALOG_ACTIVATION_SHUTDOWN_VIA_DEACTIVATE_ON_IDLE).Increment(); outside, or pass an argument which counter to increment. And add another counter for stuck activations.

gabikliot: Please add comment explaining the logic here: why we call DeactivateActivation and ALSO call directory.UnregisterAsync. (DeactivateActivation  already does directory.UnregisterAsync, so we want to explain that we don't want to wait with UnregisterAsync until activation gets Unstuck).

gabikliot:
I think overall its OK. See some minor comments.

An alternative could be to try deactivate this activation more aggressively: actually try to call OnDeactiavteAsync, and then remove it from the scheduler and fully eliminate it (in addition to all we do now). Since we are NOT targeting here activation stuck on a thread, but merely stuck Task, we CAN safely do all the above and  get rid of this activation completely.

With the current approach you theoretically increasing the chance of duplicate activation. This activation gets unstcuk and may send more msgs from the in-flight method call. Or even if not, we will call its DeactivateAsync, and that can do all kind of bad things, since a) we now have grain that is not in the directory sending msgs b) there is another activation of this grain that is active.

I would personally prefer the approach of fully deactivating it. (there is no data corruption risk with this approach). But it is more complicated code-wise. You will need to write a new DeactivateActivation method with somewhat different order of operation.

I think I am OK with either approach.

sergeybykov:
My personal bias is against trying to invoke `OnDeactiavteAsync` for the following reasons.

1. Code in `OnDeactiavteAsync` might depend on an unresolved `Task` or a data structure that is affected by the stuck method call. We have a chance of getting stuck with the `OnDeactiavteAsync` call itself, since it's asynchronous. 
2. App code does not ever expect `OnDeactiavteAsync` to be invoked prior to execution of other code of the grain class.
3. `OnDeactiavteAsync` is not guaranteed to be called in case of a silo failure, and we advise (maybe not loud enough) against using it for persisting anything. We are mitigating a rather extraordinary situation here, and I think it's no better than a silo failure from the developer expectations perspective.

PR #2395, opened 2016-11-07
Removes deadlocking and corrupted hashing

### LINKED ISSUE #2393: Getting lots of InconsistentStateExceptions with SQL Storage provider
We've run into an issue where we get lots of InconsistentStateExceptions, such as this:

image

I have created a project to reproduce this issue; it is the OrleansInconsistentStateException folder from my [experimental BitBucket repository](https://bitbucket.org/dandago/experimental/src).

Change the connection string in the silo project's OrleansConfiguration.xml.

The client program repeatedly calls an async void method. This is intentional; it simulates the case we ran into where we have event handlers from a third party library firing. We can't await those.

It is possible that you'll get database-level deadlocks instead of InconsistentStateExceptions at times. Rerun it a few times.

Both the InconsistentStateExceptions and database-level deadlocks are problematic. The load isn't particularly large here.

veikkoeeva:
* Removes deadlocking on rapid insert of initial
grain states by rearranging locking functionality.

* Removes a problem that was mainly seen in highly
concurrent grain state hashing where the hash of
the state grain type was corrupted in some minor ratio.
This basically led some grain states to be unretrievable.

* Adds a directive to prohibit lock escalation. This could
become a problem if the Storage table gets large and concurrent
operations exceed the threshold (5000) locks in which case
SQL Server reacts by tacking a table lock for operations. On
a large table this would be very problematic.

* Minor clean up on logging where the grain ID was logged twice
concatenated and testing where "MySQL" was applied to SQL Server tests
and some tests added to ensure some helper classes function as
intended.

Fixes #2393.

### DIFF
--- a/src/OrleansSQLUtils/Storage/Provider/AdoGrainKey.cs
+++ b/src/OrleansSQLUtils/Storage/Provider/AdoGrainKey.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Globalization;
 using System.Text;
 
 
@@ -100,7 +101,7 @@ namespace Orleans.Storage
             string keyExtension = null;
             if(IsLongKey)
             {
-                primaryKey = N1Key.ToString();
+                primaryKey = N1Key.ToString(CultureInfo.InvariantCulture);
                 keyExtension = StringKey;
             }
             else if(IsGuidKey)
@@ -114,7 +115,7 @@ namespace Orleans.Storage
             }
 
             const string GrainIdAndExtensionSeparator = "#";
-            return primaryKey + string.Format($"{primaryKey}{(keyExtension != null ? GrainIdAndExtensionSeparator + keyExtension : string.Empty)}");
+            return string.Format($"{primaryKey}{(keyExtension != null ? GrainIdAndExtensionSeparator + keyExtension : string.Empty)}");
         }
 
 
--- a/test/TesterInternal/StorageTests/Relational/SqlServerStorageTests.cs
+++ b/test/TesterInternal/StorageTests/Relational/SqlServerStorageTests.cs
@@ -19,6 +19,7 @@ namespace UnitTests.StorageTests.Relational
     /// <remarks>To duplicate these tests to any back-end, not just for relational, copy and paste this class,
     /// optionally remove <see cref="RelationalStorageTests"/> inheritance and implement a provider and environment
     /// setup as done in <see cref="CommonFixture"/> and how it delegates it.</remarks>
+    [TestCategory("SqlServer")]
     public class SqlServerStorageTests: RelationalStorageTests, IDisposable, IClassFixture<CommonFixture>
     {
         /// <summary>
@@ -26,16 +27,11 @@ namespace UnitTests.StorageTests.Relational
         /// </summary>
         private const string AdoNetInvariant = AdoNetInvariants.InvariantNameSqlServer;
 
-        /// <summary>
-        /// The category of this back-end vendor.
-        /// </summary>
-        public const string VendorCategory = "SqlServer";
-
 
         public SqlServerStorageTests(CommonFixture commonFixture) : base(AdoNetInvariant, commonFixture)
         {
             //XUnit.NET will automatically call this constructor before every test method run.
-            Skip.If(PersistenceStorageTests == null, $"Persistence storage not available for {VendorCategory}.");
+            Skip.If(PersistenceStorageTests == null, $"Persistence storage not available for SqlServer.");
         }
 
 
@@ -45,7 +41,7 @@ namespace UnitTests.StorageTests.Relational
         }
 
         [SkippableFact]
-        [TestCategory("Functional"), TestCategory("Persistence"), TestCategory(VendorCategory)]
+        [TestCategory("Functional"), TestCategory("Persistence")]
         internal async Task WriteReadCyrillic()
         {
             await PersistenceStorageTests.PersistenceStorage_Relational_WriteReadIdCyrillic();
@@ -53,14 +49,14 @@ namespace UnitTests.StorageTests.Relational
 
 
         [SkippableFact]
-        [TestCategory("Functional"), TestCategory("Persistence"), TestCategory(VendorCategory)]
+        [TestCategory("Functional"), TestCategory("Persistence")]
         internal async Task WriteReadWriteRead100StatesInParallel()
         {
             await Relational_WriteReadWriteRead100StatesInParallel();
         }
 
         [SkippableFact]
-        [TestCategory("Functional"), TestCategory("Persistence"), TestCategory(VendorCategory)]
+        [TestCategory("Functional"), TestCategory("Persistence")]
         internal async Task StorageDataSetGeneric_HashCollisionTests()
         {
             await Relational_HashCollisionTests();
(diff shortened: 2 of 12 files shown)


veikkoeeva:
I believe I'm done here unless someone wishes for something. @shayhatsor, do you want to review?

shayhatsor:
@veikkoeeva, there's only one issue I can think of... maybe we should have the storage provider code decide whether an update or an insert is needed by checking if the local state is null. This will greatly simplify the SQL script for current and future backends. Instead of having a `WriteToStorageKey` we'll have `InsertToStorageKey` and `UpdateStorageKey`. A similar approach is used in [DynamoDBStorageProvider](https://github.com/dotnet/orleans/blob/master/src/OrleansAWSUtils/Storage/Provider/DynamoDBStorageProvider.cs#L206), what do you think?

veikkoeeva:
@shayhatsor That is a good point. The historical reason for this is  aconsideration for some failure scenarios where one has done in-storage things to the table data and also learning how the system works. I don't remember if they hold anymore, so might as simplify this the next round.

src/OrleansSQLUtils/Storage/Provider/AdoGrainKey.cs
```diff
@@ -73,17 +74,18 @@ public byte[] GetHashBytes()
                 bytes = ToGuidKey(N0Key, N1Key).ToByteArray();
             }
 
+            var utf8 = new UTF8Encoding();
```
attilah: what is the reason for newing up an UTF8Encoding class instance?

veikkoeeva: @attilah Do you mean this in particular or all the othes too?

It appears not be thread safe. If I use the static property, there will incorrect results in high-concurrency scenario. I newed up them all just be on the robust side. Should gargage or performance become a problem, there looks like be more fruitful considerations too.

veikkoeeva:
@shayhatsor Now I remember one considreation having it in one, it's `MERGE` too (or UPSERT), but `MERGE` is problematic.

src/OrleansSQLUtils/Storage/Provider/AdoGrainKey.cs
sergeybykov: Would `ThreadLocal` be better?

attilah: @veikkoeeva can we do a repro for the scenario which produces the invalid output? I'm not convinced that this is needed, since the encoding classes are stateless in general.

@sergeybykov  I'd not go the ThreadLocal way since I thing that's not available in .Net Core, I just don't see the real need for this, were you bitten by it before?

attilah: @sergeybykov Ok it was not the ThreadLocal what we were moving away from, I was remembered incorrectly.

shayhatsor:
@veikkoeeva what do you mean by
> having it in one, it's MERGE too (or UPSERT), but MERGE is problematic.

?

veikkoeeva:
@shayhatsor That I remember one of the reasons both the UPDATE and INSERT are in the same script, it is UPSERT, just writte in a way that tries to take advantange of the knowledge that INSERT is a rare occasion and it can be checked with a condition. Only in that case consider a heavier operation is considered. But the operation could be split in two in code already. It might make sense later on, like considering how to do the state clearing part.

shayhatsor:
@veikkoeeva, yes I remember that we also avoided `MERGE` in other parts of the SQL Server script for the same reasons. The reason I mentioned the split idea here is because I thought that it will also remove the need for lock hints and configuration. But after taking a closer looks at the Storage table, I figured out we'll still need the same lock mechanism for `INSERT`s.
The thought that comes to mind is whether we have to handle the duplicate activation case by pessimistically. We can decide that it's OK for two or more rows of the same grain state to exist in the Storage table. In the rare case of duplicate activation, both `INSERT`s will succeed. But still in `InsertToStorageKey` we can perform a `SELECT` and return the number of instances we've found. The storage provider code will throw the `InconsistentStateException` when a number higher than 1 is returned. This will result in all concurrent `InsertToStorageKey` calls to fail, which is OK by me. `UpdateStorageKey`, `ReadStorageKey` and `ClearStorageKey` will need to work only on the "first" returned row.
The reason I'm going so far to avoid locking is because it's very vendor specific. This might also improve the performance of INSERT operations.
Do you think this approach can work ?

test/TesterInternal/StorageTests/Relational/SqlServerStorageTests.cs
```diff
@@ -148,15 +148,15 @@ public void Dispose()
 
 
         [SkippableTheory, ClassData(typeof(StorageDataSetGeneric<string, string>))]
-        [TestCategory("Functional"), TestCategory("Persistence"), TestCategory("MySql")]
+        [TestCategory("Functional"), TestCategory("Persistence"), TestCategory(VendorCategory)]
```
jdom: nit picking, but since you are doing this change in many places: You can move the Persistence and VendorCategory TestCategory attributes to the class level, instead of repeating on a method by method basis (we are slowly moving towards that since we migrated from MSTest). Functional or BVT is OK to apply at the method level even if currently all methods are of the same category, but you never know if a new test method will also be of the same kind.
Just a suggestion, feel free to ignore if there were no other changes required

jdom:
Attila's PR #2403 was just merged. Do you mind rebasing and making sure it works with it, so that we can include it in the real imminent 1.3.1 release? 
Thanks a lot for it

veikkoeeva:
@jdom I rebased and modified the tests. I also tested the new hasher and it looks like working. The situation that tripped me was  https://gitter.im/dotnet/orleans?at=58237b87e462097a302ba4ab. If you look at it, it uses `Encoding.UTF8.GetBytes(thesSring);` to get the bytes and then calls the overload that takes a `byte[]` parameter. I tried also the one that too `string` and it failed equally. I would have assumed this refactored version would also fail on the `string` overload due to the internal `Encoding.UTF.GetBytes` call, but apparently it does not. I have no explanation for this.

veikkoeeva:
@shayhatsor That is true, or the storage provider could sort the results and returns the oldest of the duplicates and soon enough the other one is left with an old version and could be cleaned in an explicit data scrubbing operation.

It would be interesting to examine as to which method would bring the most performance on the table and consider an alternate implementation using an in-memory table. Though I just recently read a few blog posts from experts (in addition to the ones mentioned in the actual ADO.NET storage PR) written just recently that indeed, heap table especially in situation like this is the most efficient -- especially if `DELETE` can be avoided. It is avoided by just nulling data and empty rows can be scrubbed explicitly if it becomes a problem.

Lately thinking about this I have considered if it should be mentioned explicitly in the documentation that using the `Storage` table one needs to consider also other issues. One of them is backup and disaster recovery time, as explained at http://thomaslarock.com/2013/01/database-filegroups-just-like-seatbelts-but-with-less-chafing/. One is also making sure `TempDb` is appropriately tuned and also schemas should be judiciously used. The script tries to provide sensible defaults, but it can only do so much and I'm afraid people might have a bad surprise about this at some point.

Anyway, to recap, this bug had two parts: deadlocking due to table lock in concurrent insert scenario and then the hash corruption case (see the link and image on @jdom note just before). The problem for table locking seemed to be that `holdlock` escalated to the whole table, but taking a bit more time on MSDN I figured out to maintain as close as locking only single rows as possible and actually forcing the index use to make it efficient. Similarly I removed lock escalation to table lock in high-concurrent scenario as a pre-emptive default, as this situation would have deadlocked the table again.

sergeybykov:
@shayhatsor, if you approve this, we could rush it into 1.3.1. We just concluded that there's no perf regression there, and are ready to start the release process as soon as we decide on this one.

shayhatsor:
@sergeybykov, this is approved. @veikkoeeva, thanks for the fix !

sergeybykov:
@veikkoeeva, thanks for digging into it!

Issue #2152, opened 2016-09-14
Errors When deploying to Azure App Services


sudsy:
I am deploying my Orleans App for the first time and encountering errors when deploying to an Azure App Service (Web App).

The trace of my error is at http://pastebin.ca/3715471

Some further information
- The application runs in release build on my local IIS development server
- The application runs under the local Azure Storage Emulator if I configure it as a Web Role
- I have tried to ensure that my dependencies are at the level required by Orleans 
- I have ensured that tracing is turned off on my web application 
- I have ensured that there is no tracing configuration in the Web.Config file
- I have added a reference to Orleans.AzureUtils because it seems to be needed, but I am unsure if this is correct for a Web app
- I have tried prerelease v 1.3 orleans
- Tried setting TraceFileName = null and TraceToConsole false in orleans config

sergeybykov:
I suspect it throw at https://github.com/dotnet/orleans/blob/master/src/Orleans/Logging/LogManager.cs#L168 when the code tries to write an error to the `DefaultTraceListener`. What's not clear to me is why `AzureDriveTraceListener` is throwing the "disposed object" exception. I don't know why `AzureDriveTraceListener` is there to begin with.

Maybe try to empty `Trace.Listeners` before initializing Orleans.

sudsy:
Clearing Trace.Listeners did the trick. Thanks! I have yet to see if there are side effects to this. Once I get a little time back I will try to stand up a plain vanilla Web App and see if it has the same problem.

I have moved on somewhat, now getting the following error, which seems to be some sort of networking issue. I will work on it again tomorrow to see if I can resolve it, but if you have any tips in the meantime they would be appreciated.

```
<Error>
<Message>An error has occurred.</Message>
<ExceptionMessage>
Response did not arrive on time in 00:00:30 for message: Request *cli/ac5c6195@c9a0f6ab->S10.0.0.5:40000:211618952TypeManagerId@S00000011 #6: global::Orleans.Runtime.ITypeManager:GetTypeCodeMap(). Target History is: <S10.0.0.5:40000:211618952:TypeManagerId:@S00000011>.
</ExceptionMessage>
<ExceptionType>System.TimeoutException</ExceptionType>
<StackTrace>
[...]

PR #2396, opened 2016-11-07
Fixes issue #2152

attilah:


ReubenBond:
LGTM, but do you think you could remove the merge commits so I don't have to squash the `error.txt` commit into this?

jdom:
In the future, please be more descriptive in the PR title (and commit). It can additionally reference an issue, but the PR (and merged commit) should be readable regardless of the reference

PR #2400, opened 2016-11-09
ActivationCountPlacementDirector: allow null DeploymentLoadPublisher

ReubenBond:
Small tweak to simplify usage of `ActivationCountPlacementDirector` in our internal load tests

### DIFF
--- a/src/OrleansRuntime/Placement/ActivationCountPlacementDirector.cs
+++ b/src/OrleansRuntime/Placement/ActivationCountPlacementDirector.cs
@@ -52,7 +52,6 @@ namespace Orleans.Runtime.Placement
                     "GlobalConfig.ActivationCountBasedPlacementChooseOutOf is " + globalConfig.ActivationCountBasedPlacementChooseOutOf);
 
             chooseHowMany = globalConfig.ActivationCountBasedPlacementChooseOutOf;
-            deploymentLoadPublisher = deploymentLoadPublisher;
             deploymentLoadPublisher?.SubscribeToStatisticsChangeEvents(this);
         }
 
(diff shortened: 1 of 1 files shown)


src/OrleansRuntime/Placement/ActivationCountPlacementDirector.cs
```diff
@@ -54,8 +52,8 @@ public ActivationCountPlacementDirector(DeploymentLoadPublisher deploymentLoadPu
                     "GlobalConfig.ActivationCountBasedPlacementChooseOutOf is " + globalConfig.ActivationCountBasedPlacementChooseOutOf);
 
             chooseHowMany = globalConfig.ActivationCountBasedPlacementChooseOutOf;
-            this.deploymentLoadPublisher = deploymentLoadPublisher;
-            this.deploymentLoadPublisher.SubscribeToStatisticsChangeEvents(this);
+            deploymentLoadPublisher = deploymentLoadPublisher;
+            deploymentLoadPublisher?.SubscribeToStatisticsChangeEvents(this);
```
jdom: Isn't it possible to pass in an mock or dummy instance of DeploymentLoadPublisher instead in the afected test? Because it feels that without this, this would not work as expected in a real Silo deployment.
If it's too big a change, then I don't feel that strongly about it anyway

src/OrleansRuntime/Placement/ActivationCountPlacementDirector.cs
```diff
@@ -54,8 +52,8 @@ public ActivationCountPlacementDirector(DeploymentLoadPublisher deploymentLoadPu
                     "GlobalConfig.ActivationCountBasedPlacementChooseOutOf is " + globalConfig.ActivationCountBasedPlacementChooseOutOf);
 
             chooseHowMany = globalConfig.ActivationCountBasedPlacementChooseOutOf;
-            this.deploymentLoadPublisher = deploymentLoadPublisher;
-            this.deploymentLoadPublisher.SubscribeToStatisticsChangeEvents(this);
+            deploymentLoadPublisher = deploymentLoadPublisher;
```
gabikliot: this line is weird now, right?

PR #2401, opened 2016-11-09
Azurequeueprovider messagevisibilityconfig

jafletch:
When using the Azure Queue based stream provider, I occasionally see both warnings about queue message deletes being slow coupled with errors related to message deletes failing. The errors appear to be happening because, while waiting to be deleted, the message visibility timeout expires and the message becomes visible in the queue again which renders the poprecipt invalid and causes a 404 on delete. This change allows for configuring the message visibility timeout value to allow for a larger window for both processing and deletes to be completed. Setting the timeout to 2min (vs 30sec default) has eliminated these errors for me, even in cases where I still see the slowness warnings.

### DIFF
--- a/src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueAdapter.cs
+++ b/src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueAdapter.cs
@@ -11,6 +11,7 @@ namespace Orleans.Providers.Streams.AzureQueue
     {
         protected readonly string DeploymentId;
         protected readonly string DataConnectionString;
+        protected readonly TimeSpan? VisibilityTimeout;
         private readonly HashRingBasedStreamQueueMapper streamQueueMapper;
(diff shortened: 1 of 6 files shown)


gabikliot:
We have a set of standalone tests for AzureQueueDataManager. Please add a new test that passes the new messageVisibilityTimeout.

gabikliot:
BTW, if the default visibility is 30 sec, we might be better increasing it to 1 or 2 min by default , to decrease the chance of double delivery (since our delivery timeout is 30 sec by default). Just an idea.

jafletch:
New tests added to provide some coverage for the changes.

Default timeout is 30 seconds per all sources I have found (https://msdn.microsoft.com/en-us/library/azure/dd179474.aspx for example)

I'll defer to you guys' judgement on choice for a new default. As I said before, I am using 2 minutes for my scenario and I don't believe I have seen an error since that change.

jason-bragg:
A default timespan can be added as a natural extension to this.  No need to hold up this change for it.
If there is still interest in a default, please continue the discussion, and we can update system with the conclusion.

PR #2403, opened 2016-11-09
Modify JenkinsHash to be stateless.

attilah:
Modify JenkinsHash to be stateless.

### DIFF
--- a/src/Orleans/IDs/JenkinsHash.cs
+++ b/src/Orleans/IDs/JenkinsHash.cs
@@ -6,29 +6,9 @@ namespace Orleans
     // Based on the version in http://home.comcast.net/~bretm/hash/7.html, which is based on that
     // in http://burtleburtle.net/bob/hash/evahash.html.
     // Note that we only use the version that takes three ulongs, which was written by the Orleans team.
-    public class JenkinsHash
+    public static class JenkinsHash
     {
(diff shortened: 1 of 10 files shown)


jdom:
LGTM

gabikliot:
Why? There was a reason it was done this way.

jdom:
Could you explain more? The reason is not apparent. We started seeing bugs due to multi-threaded usages of this, which of course could be fixed to use thread local (or local) instances instead, but the reality is that there is no apparent reason not to make this class thread safe and stateless, and be less error prone in the usage.

gabikliot:
I will look at this and try to recall.

As a side general note, if you do seek my feedback, please try to provide a bit more details in the PR. The level of details u provided now is good enough, but without it I have no way to know if you found some kind of bug in this code, or it is just a code cleanup, or something else.

jdom:
Agree on that more context should be added by default. I was aware of this context due to an in-person standup meeting, but sometimes I miss context too and have to explicitly request it.

gabikliot:
Looks good, and simpler and even more efficient this way.  I can't see a good reason why we didn't do it this way in the first place.

sergeybykov:
Thanks, @gabikliot, for confirming it.

PR #2404, opened 2016-11-10
Fix broken consul docs links

galvesribeiro:


src/Documentation/Runtime-Implementation-Details/Cluster-Management.md
```diff
@@ -104,7 +104,7 @@ As already mentioned, `MembershipTable` is used as a rendezvous point for silos
...
 
-4.  [Consul IO](https://www.consul.io) - we used [Consul's Key/Value store](https://www.consul.io/intro/getting-started/kv.html) to impelement the membershop table. Refer to [Consul-Deployment](http://dotnet.github.io/orleans/Runtime-Implementation-Details/Consul-Deployment) for more details.
+4.  [Consul IO](https://www.consul.io) - we used [Consul's Key/Value store](https://www.consul.io/intro/getting-started/kv.html) to impelement the membershop table. Refer to [Consul-Deployment](http://dotnet.github.io/orleans/Documentation/Runtime-Implementation-Details/Consul-Deployment) for more details.
```
jdom: Shouldn't this be using a relative path so that docfx builds the right link (and can also validate it exists)? I'm pretty sure we don't normally link to github.io for inter documentation links. @richorama should probably confirm

richorama: Yes, it should be relative and end with a '.md' extension. Docfx will rewrite the URL.

richorama: It should be relative (i.e. can't start with a slash) I think this is correct: `Consul-Deployment.md`

jdom:
Thanks @galvesribeiro. Fix should be fixed once we regenerate the docs

PR #2409, opened 2016-11-11
Update Front-Ends-for-Orleans-Services.md

pdubb29:
I encountered an error following the documentation step by step, this a proposed change to the documentation so other people can avoid the same error.

### DIFF
--- a/src/Tutorials/Front-Ends-for-Orleans-Services.md
+++ b/src/Tutorials/Front-Ends-for-Orleans-Services.md
@@ -58,6 +58,7 @@ namespace WebApplication1
        	   ...
 ```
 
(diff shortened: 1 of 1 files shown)


src/Tutorials/Front-Ends-for-Orleans-Services.md
```diff
@@ -58,6 +58,7 @@ namespace WebApplication1
        	   ...
 ```
 
+The default configuration for the config has the port set to 40000, which your configuration file which we set up earlier in the documentation has been set to 30000
```
sergeybykov: I'm sorry, I'm a bit confused here. Maybe this could be rephrased in a more straightforward way? Or should we instead update the config file to match the default or maybe even not set the gateway port explicitly at all?

PR #2410, opened 2016-11-11
MissingActivation Test renames and new test added to repro a multi silo bug alike.

attilah:
MissingActivation_1 = MissingActivation_WithoutDirectoryLazyDeregistration_MultiSilo
MissingActivation_2 = MissingActivation_WithDirectoryLazyDeregistration_SingleSilo

Added new test to repro the random bug of MissingActivation_2, test name: MissingActivation_WithoutDirectoryLazyDeregistration_MultiSilo_SecondaryFirst.

If we've 2 silos and the grain is created in the secondary for the first time, then we're removing it via TestHooks from all Silos and then the next call to the same grain happens to be in the Primary Silo, then the grain is not healing and every subsequent calls is failing to it.

### DIFF
--- a/test/TesterInternal/ActivationsLifeCycleTests/DeactivateOnIdleTests.cs
+++ b/test/TesterInternal/ActivationsLifeCycleTests/DeactivateOnIdleTests.cs
@@ -9,12 +9,19 @@ using TestExtensions;
 using UnitTests.GrainInterfaces;
 using UnitTests.TestHelper;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace UnitTests.ActivationsLifeCycleTests
 {
     public class DeactivateOnIdleTests : OrleansTestingBase, IDisposable
     {
+        private readonly ITestOutputHelper output;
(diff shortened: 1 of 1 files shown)


test/TesterInternal/ActivationsLifeCycleTests/DeactivateOnIdleTests.cs
```diff
@@ -251,45 +258,120 @@ public void Dispose()
...
+
             for (int i = 0; i < 10; i++)
             {
-                await MissingActivation_Runner(i, directoryLazyDeregistrationDelay);
+                await MissingActivation_Runner(i, lazyDeregistrationDelay);
             }
         }
 
-        private async Task MissingActivation_Runner(int grainId, TimeSpan lazyDeregistrationDelay)
+        [SkippableFact, TestCategory("Functional"), TestCategory("ActivationCollector")]
```
attilah: That's what I meant, thanks, I'm fixing it.

PR #2411, opened 2016-11-11
Update Unit-Testing-Grains.md

pdubb29:
I encountered an error following the documentation step by step, this a proposed change to the documentation so other people can avoid the same error.

### DIFF
--- a/src/Tutorials/Unit-Testing-Grains.md
+++ b/src/Tutorials/Unit-Testing-Grains.md
@@ -46,6 +46,7 @@ namespace Tests
             // Optional.
             // By default, the next test class which uses TestingSiloHost will
             // cause a fresh Orleans silo environment to be created.
+            //this code causes a build error. "Cannot call non-static method from a static context"
             StopAllSilos();
         }
 


src/Tutorials/Unit-Testing-Grains.md
```diff
@@ -46,6 +46,7 @@ namespace Tests
             // Optional.
             // By default, the next test class which uses TestingSiloHost will
             // cause a fresh Orleans silo environment to be created.
+            //this code causes a build error. "Cannot call non-static method from a static context"
             StopAllSilos();
```
jdom: I think instead of stating that the sample is broken, we should fix it.
I believe this should use `StopAllSilosIfRunning()` instead.
Nevertheless, I'm surprised that the docs are so stale. We stopped endorsing `TestingSiloHost` in favor of `TestCluster` some time ago (and we plan to deprecate the former soon), and in fact the samples were also updated: https://github.com/dotnet/orleans/blob/master/Samples/HelloWorld/SiloTests/HelloWorldSiloTests.cs

If you want, we would welcome an update to this doc that shows TestCluster usage instead of TestingSiloHost.

sergeybykov:
@pdubb29, are you interested in fixing the sample?

Issue #2379, opened 2016-11-01
Heterogenous silo support


benjaminpetit:
# Heterogenous silo support
Right now, all silos assume that they all support the same set of grains. We want to remove this limitation.

## Why
Removing this assumption could enable these scenarios: 
- **Controlling grain location**: some people want to fine tune grain placement (some grain might require a GPU for computation, some other requires lot of memory, needs to be close to storage, ect.)
- **In-place upgrade of silos**: support upgrade with different grain type supported (ie: deploy new binary with some grain added, but no modification on already supported grains)

## Out of scope (for now)
- Grain versioning
- Stateless grains
- Implicit streaming subscription
- Multi-cluster support

## Design

1. **Add list of excluded grains by assembly loader in config** I don't think it is a strong requirement, but it will enable easier testing. It will also allow user to deploy the same package on the cluster even if he want to use this feature.
2. **All silos need to compute a union of all supported grains across the cluster**. Each time a silo join or leave the cluster, this will need to be recomputed
3. **Gateways need to expose this global list of supported grains to the client** 
4. **Update all placement directors** to only choose in a list of silos that support the target grain type 

Possible optimizations: 
- Silos should push the list of supported grain to connected clients when it is updated

gabikliot:
Thumbs up @BenjaminGibbs !
long overdue ... @jason-bragg ;-)

benjaminpetit:
Merged!

PR #2443, opened 2016-11-23
Heterogenous silos support 

benjaminpetit:
This is a PR for #2379 

Currently, the GrainInterfaceMap is built on each silo, and is sent to clients when they connect to a gateway.

### Proposed changes 
I tried to divide the PR in logical commits so it is easier to review:
- To test this functionality easily, I added a list of grain types that are ignored by the AssemblyLoader. (bbc118d and f793f2f). 
- Add a notion of “cluster wide” GrainInterfaceMap, that is a merge of all GrainInterfaceMaps built locally on each silo. (568cf00, 4bdc77a and 3aab754)
- Expose this new global GrainInterfaceMap to the clients (3aab754)
- Rebuild the global GrainInterfaceMap when new silos join the cluster (a259b31 and adf6ea1)
- Modify PlacementDirector to only consider “compatible”silos. (83057de and 77e80d8)

### How and when the global GrainInterfaceMap is recalculated
To avoid race conditions and complicated error handling, we choose to refresh the GrainInterfaceMap using a timer: every timer tick, if a silo joined the cluster since the last computation, we will get the local GrainInterfaceMap from the new silo and merge it into the global map. If this fails, it will be retried on the next tick. This process is done on each silo in the cluster.
The drawback of this approach is that a silo that joined the cluster will not be chosen for placement by other silos until this timer fire. I don’t think this is an issue, and the timer interval can be tweaked in configuration.

### Current implementation limitations
- When a new GrainInterfaceMap is computed, it is not pushed to clients. Client will need to disconnect/reconnect to get a refreshed type map.
- The “excluded grain types” should be use for test only

### DIFF
--- a/src/Orleans/Configuration/GlobalConfiguration.cs
+++ b/src/Orleans/Configuration/GlobalConfiguration.cs
@@ -170,6 +170,10 @@ namespace Orleans.Runtime.Configuration
         /// The number of seconds to attempt to join a cluster of silos before giving up.
         /// </summary>
         public TimeSpan MaxJoinAttemptTime { get; set; }
+        /// <summary>
+        /// The number of seconds to refresh the cluster grain interface map
+        /// </summary>
+        public TimeSpan TypeMapRefreshTimeout { get; set; }
         internal ConfigValue<int> ExpectedClusterSizeConfigValue { get; set; }
         /// <summary>
         /// The expected size of a cluster. Need not be very accurate, can be an overestimate.
@@ -470,6 +474,7 @@ namespace Orleans.Runtime.Configuration
         private static readonly TimeSpan DEFAULT_LIVENESS_DEATH_VOTE_EXPIRATION_TIMEOUT = TimeSpan.FromSeconds(120);
         private static readonly TimeSpan DEFAULT_LIVENESS_I_AM_ALIVE_TABLE_PUBLISH_TIMEOUT = TimeSpan.FromMinutes(5);
         private static readonly TimeSpan DEFAULT_LIVENESS_MAX_JOIN_ATTEMPT_TIME = TimeSpan.FromMinutes(5); // 5 min
+        private static readonly TimeSpan DEFAULT_REFRESH_CLUSTER_INTERFACEMAP_TIME = TimeSpan.FromMinutes(1);
         private const int DEFAULT_LIVENESS_NUM_MISSED_PROBES_LIMIT = 3;
         private const int DEFAULT_LIVENESS_NUM_PROBED_SILOS = 3;
         private const int DEFAULT_LIVENESS_NUM_VOTES_FOR_DEATH_DECLARATION = 2;
@@ -522,6 +527,7 @@ namespace Orleans.Runtime.Configuration
             UseLivenessGossip = DEFAULT_LIVENESS_USE_LIVENESS_GOSSIP;
             ValidateInitialConnectivity = DEFAULT_VALIDATE_INITIAL_CONNECTIVITY;
             MaxJoinAttemptTime = DEFAULT_LIVENESS_MAX_JOIN_ATTEMPT_TIME;
+            TypeMapRefreshTimeout = DEFAULT_REFRESH_CLUSTER_INTERFACEMAP_TIME;
             MaxMultiClusterGateways = DEFAULT_MAX_MULTICLUSTER_GATEWAYS;
             BackgroundGossipInterval = DEFAULT_BACKGROUND_GOSSIP_INTERVAL;
             UseGlobalSingleInstanceByDefault = DEFAULT_USE_GLOBAL_SINGLE_INSTANCE;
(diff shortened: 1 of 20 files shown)


src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
```diff
@@ -13,18 +15,26 @@ namespace Orleans.Runtime
...
+        private Dictionary<int, IList<SiloAddress>> supportedSilosByTypeCode;
         private readonly Logger logger = LogManager.GetLogger("GrainTypeManager");
         private readonly GrainInterfaceMap grainInterfaceMap;
         private readonly Dictionary<int, InvokerData> invokers = new Dictionary<int, InvokerData>();
         private readonly SiloAssemblyLoader loader;
         private static readonly object lockable = new object();
+		private readonly PlacementStrategy defaultPlacementStrategy;
 
-        private readonly PlacementStrategy defaultPlacementStrategy;
+        internal IDictionary<SiloAddress, GrainInterfaceMap> GrainInterfaceMapsBySilo
```
ReubenBond: Given the perf implications of using `IDictionary<,>`, I think the concrete type would be preferable.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -200,6 +201,15 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
         /// </summary>
         public Dispatcher Dispatcher { get; }
 
+        public IList<SiloAddress> GetCompatibleSiloList(GrainId grain)
+        {
+            var typeCode = grain.GetTypeCode();
+            var compatibleSilos = GrainTypeManager.GetSupportedSilos(typeCode).Intersect(AllActiveSilos).ToList();
```
ReubenBond: This method has to be cheap, since it's on the warm/hot path. Currently, it is too allocation-happy.

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
benjaminpetit: Yes you are right, completely missed that

src/Orleans/Configuration/GlobalConfiguration.cs
```diff
@@ -170,6 +170,10 @@ public bool PrimaryNodeIsRequired
         /// The number of seconds to attempt to join a cluster of silos before giving up.
         /// </summary>
         public TimeSpan MaxJoinAttemptTime { get; set; }
+        /// <summary>
+        /// The number of seconds to refresh the cluster grain interface map
+        /// </summary>
+        public TimeSpan TypeMapRefreshTimeout { get; set; }
```
benjaminpetit: Agree, I renamed to TypeMapRefreshInterval

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
```diff
@@ -13,18 +15,26 @@ namespace Orleans.Runtime
     internal class GrainTypeManager
     {
         private IDictionary<string, GrainTypeData> grainTypes;
+        private IReadOnlyDictionary<SiloAddress, GrainInterfaceMap> grainInterfaceMapsBySilo;
```
ReubenBond: What I meant in the other comment was to use `Dictionary<,>` here, since it's cheaper to enumerate:
```
    Method |       Mean |    StdDev |  Gen 0 | Allocated |
---------- |----------- |---------- |------- |---------- |
  Concrete | 27.2424 ns | 0.8356 ns |      - |       0 B |
 Interface | 51.0121 ns | 1.1681 ns | 0.0125 |      56 B |
```

src/OrleansRuntime/Placement/IPlacementContext.cs
```diff
@@ -70,6 +70,16 @@ public static string GetGrainTypeName(this IPlacementContext @this, int typeCode
             return grainClass;
         }
 
+        public static string GetGrainTypeNameAndSupportedSilos(this IPlacementContext @this, GrainId grainId, out IList<SiloAddress> supportedSiloAddresses, string genericArguments = null)
```
ReubenBond: Is this method used? I couldn't see any usages

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
benjaminpetit: :)

src/OrleansRuntime/Placement/IPlacementContext.cs
benjaminpetit: Relics of a previous rebase, removed

sergeybykov:
@ReubenBond, please 'Squash and merge' when you all your feedback is taken care of.

src/Orleans/Runtime/GrainInterfaceMap.cs
```diff
@@ -173,6 +209,14 @@ internal bool ContainsGrainInterface(int interfaceId)
             }
         }
 
+        internal bool ContainsGrainImplementation(int typeCode)
```
ReubenBond: No usages of this?

galvesribeiro:
Awesome @benjaminpetit great work! 👍

PR #2447, opened 2016-11-24
Azure queue stream provider silo to silo serialization bug.

jason-bragg:
V2 messaging is Compatible with V1, but V2 containers expect V2 tokens, so we musta always generate v2 tokens.
This should fix the failing azure queue functional tests.


gabikliot:
What is V2?

jason-bragg:
@gabikliot, @Carlm-MS added new versions of EventSequenceToken, EventHubSequenceToken, and AzureQueueBatchContainer witch allow serialization of these types using external serializers, specifically json.  These new types should be binary compatible with the serialized original types by default, to maintain backwards compatibility.  These new types all have the same name as the original type, appended with "V2".

While @Carlm-MS added significant testing around type copy and serialization, grain to grain calls using the new batch container with an old event sequence token were failing.  This mixing of original and new types have caused me to (re)question the original change.  While I'd rather not revert these changes, if an alternate approach is safer, I'm open to suggestions.

PR #2453, opened 2016-11-25
Add "Custom/" prefix for NewRelic metrics

DixonDs:
According to NR docs https://docs.newrelic.com/docs/agents/manage-apm-agents/agent-data/custom-metrics if is required to prefix all custom metrics with `"Custom/"`.


sergeybykov:
Thank you, @DixonDs!

PR #2458, opened 2016-11-29
Port Code Generation to .NET Standard 1.5

ReubenBond:
* Includes a fix by @attilah to move the `*.codegen.cs` file to the build directory and include it dynamically
* Upgrades .NET packages to latest stable versions (`System.Runtime.Serialization.Formatters` was previously on a beta version, now it's at `4.3.0` like the other packages)
* Implemented a new package: `Orleans.PlatformServices` which is used to load assemblies from byte arrays. This packages is 'multi-targeted', meaning that a separate version is compiled for `netstandard1.5` and `net462`. 
* Builds are now output into separate folders for `netstandard1.5` and `net462` rather than a single directory. This allows us to implement the above package, otherwise dlls will clobber each other and strange build errors will be thrown.


ReubenBond:
@dotnet-bot please test this

.gitignore
```diff
@@ -3,6 +3,7 @@
 
 # Tool Runtime Dir
 /vNext/src/[Tt]ools/
+/vNext/src/[Bb]ootstrap/
```
ReubenBond: During the VS build, we output the bootstrap code generator to the Bootstrap folder so that all projects can access it.

src/Orleans/AssemblyLoader/AssemblyProcessor.cs
```diff
@@ -75,7 +75,20 @@ public void Initialize()
...
+                    {
+                        this.ProcessAssembly(generatedAssembly?.Assembly);
+                    }
+                }
+
+                foreach (var generatedAssembly in CodeGeneratorManager.GetGeneratedAssemblies().Values)
+                {
+                    this.ProcessAssembly(generatedAssembly?.Assembly);
+                }
+
```
ReubenBond: in net462 we rely on assembly load events to process generated assemblies. In netstandard1.5, no such event exists, so we pass generated assemblies back from codegen and process them immediately.

src/OrleansCodeGenerator/CodeGeneratorCommon.cs
```diff
@@ -62,7 +62,7 @@ public static GeneratedAssembly CompileAssembly(GeneratedSyntax generatedSyntax,
 
             // Reference everything which can be referenced.
             var assemblies =
-                System.AppDomain.CurrentDomain.GetAssemblies()
+                AppDomain.CurrentDomain.GetAssemblies()
```
galvesribeiro: AppDomain on netstandard?

src/Orleans/CodeGeneration/GeneratedAssembly.cs
```diff
@@ -32,5 +39,14 @@ public GeneratedAssembly(GeneratedAssembly other)
...
         public byte[] DebugSymbolRawBytes { get; set; }
+
+        /// <summary>
+        /// Gets or sets the assembly.
+        /// </summary>
+        public Assembly Assembly
+        {
+            get { return this.assembly; }
+            set { this.assembly = value; }
+        }
```
ReubenBond: The code generator is responsible for loading the generated assembly. This is where the result is stored.

src/OrleansCodeGenerator/RoslynCodeGenerator.cs
```diff
@@ -72,22 +73,24 @@ public void AddGeneratedAssembly(string targetAssemblyName, GeneratedAssembly ge
         /// <summary>
         /// Generates code for all loaded assemblies and loads the output.
         /// </summary>
-        public void GenerateAndLoadForAllAssemblies()
+        public IReadOnlyList<GeneratedAssembly> GenerateAndLoadForAllAssemblies()
```
galvesribeiro: I wonder if worth keep using those interfaces give the recent article we saw few days ago on main chat that mention the (big) overhead to iterate on collection interfaces assuming that we know that there is no other implementations that we will ever return here.

src/OrleansCodeGenerator/RoslynCodeGenerator.cs
```diff
@@ -273,14 +287,42 @@ private static void TryLoadGeneratedAssemblyFromCache(Assembly targetAssembly)
...
+        private static Assembly LoadAssembly(GeneratedAssembly asm)
+        {
+#if ORLEANS_BOOTSTRAP
+            throw new NotImplementedException();
+#elif NETSTANDARD
+            Assembly result;
+            result = Orleans.PlatformServices.PlatformAssemblyLoader.LoadFromBytes(asm.RawBytes, asm.DebugSymbolRawBytes);
+            AppDomain.CurrentDomain.AddAssembly(result);
+            return result;
+#else
```
ReubenBond: This assembly is `netstandard1.5`, but it calls into the multi-targeted `PlatformServices` assembly. I didn't want to mess with the existing .NET 4.6 tree (only vNext tree), so we use `Assembly.Load()` directly in the main tree.

src/OrleansCodeGenerator/RoslynCodeGenerator.cs
```diff
@@ -273,14 +287,42 @@ private static void TryLoadGeneratedAssemblyFromCache(Assembly targetAssembly)
...
+        /// Loads the specified assembly.
+        /// </summary>
+        /// <param name="asm">The assembly to load.</param>
+        private static Assembly LoadAssembly(GeneratedAssembly asm)
+        {
+#if ORLEANS_BOOTSTRAP
+            throw new NotImplementedException();
+#elif NETSTANDARD
+            Assembly result;
+            result = Orleans.PlatformServices.PlatformAssemblyLoader.LoadFromBytes(asm.RawBytes, asm.DebugSymbolRawBytes);
```
galvesribeiro: Do we load Debug symbols in release build as well? (dumb question)

test/Tester/CodeGenTests/GeneratorGrainTest.cs
```diff
@@ -251,5 +252,6 @@ public GeneratorGrainTest()
             var output = await grain.Echo(input);
             Assert.Equal(input, output);
         }
+#endif
```
ReubenBond: CodeGen tests were imported into the tree, but F# projects aren't yet present.

test/Tester/CodeGenTests/GeneratorGrainTest.cs
```diff
@@ -243,6 +243,7 @@ public GeneratorGrainTest()
             Assert.Equal(input, output);
         }
 
+#if !NETSTANDARD_TODO
```
galvesribeiro: F# still off in this release, correct?

vNext/src/Build.cmd
```diff
@@ -36,25 +43,25 @@ call %_dotnet% restore "%CMDHOME%\.nuget\Tools.csproj"
 @echo Build Debug ==============================
 
 SET CONFIGURATION=Debug
-SET OutDir=%~dp0..\Binaries\%CONFIGURATION%
+SET OutputPath=%~dp0..\Binaries\%CONFIGURATION%
```
ReubenBond: By setting `OutputPath` instead of `OutDir`, we end up with target-specific output paths. Overriding `OutDir` instead would leave us with a single output directory for both `net462` and `netstandard1.5`, which wont work anymore because we have multiple versions of some DLLs (namely the platform services dll)

vNext/src/ClientGenerator/BootstrapCodegen.proj
```diff
@@ -0,0 +1,48 @@
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
ReubenBond: @jdom created this project to clean up the way that the bootstrap versions of projects were built. These directives previously existed in `Orleans.csproj`, but were moved here largely for cleanliness

vNext/src/Orleans.PlatformServices/PlatformAssemblyLoader.cs
```diff
@@ -0,0 +1,52 @@
...
+                else
+                {
+                    return AssemblyLoadContext.Default.LoadFromStream(assemblyStream);
+                }
+            }
+#elif NET46
+            return Assembly.Load(assembly, debugSymbols);
+#else
+            throw new NotImplementedException();
+#endif
```
ReubenBond: This is the crux of the platform-specific assembly loading code.

vNext/src/Orleans.PlatformServices/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,15 @@
...
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Orleans.PlatformServices")]
+[assembly: AssemblyDescription("Provides platform-specific implementations of functionality required by Orleans.")]
+
+// Setting ComVisible to false makes the types in this assembly not visible 
+// to COM components.  If you need to access a type in this assembly from 
+// COM, set the ComVisible attribute to true on that type.
+[assembly: ComVisible(false)]
+//[assembly: SkipCodeGeneration]
```
galvesribeiro: Should be removed?

test/Tester/CodeGenTests/GeneratorGrainTest.cs
galvesribeiro: Will we add support to F# on crawl phase?

vNext/src/ClientGenerator/BootstrapCodegen.proj
jdom: well, previously they were distributed in Orleans.csproj, and also each project that was part of the bootstrap had a big section in it for when bootstrap was true that was copy/pasted in each. This is more centralized

src/OrleansCodeGenerator/CodeGeneratorCommon.cs
jdom: it's still just the shim

src/Orleans/AssemblyLoader/AssemblyProcessor.cs
```diff
@@ -75,7 +75,20 @@ public void Initialize()
                 Assembly[] assemblies = AppDomain.CurrentDomain.GetAssemblies();
 
                 // initialize serialization for already loaded assemblies.
-                CodeGeneratorManager.GenerateAndCacheCodeForAllAssemblies();
+                var generated = CodeGeneratorManager.GenerateAndCacheCodeForAllAssemblies();
```
jdom: I believe this is the only place that uses `GenerateAndCacheCodeForAllAssemblies`. If that is indeed the case, just kill that method and pass the collection of assemblies that you just got from the shim, so we reduce the shim's usage.

vNext/src/NuGet/Microsoft.Orleans.OrleansCodeGenerator.Build.nuspec
```diff
@@ -27,18 +27,18 @@
...
-    <file src="OrleansCodeGenerator.pdb" target="tools" />
-    <file src="Microsoft.CodeAnalysis.CSharp.dll" target="tools" />
-    <file src="Microsoft.CodeAnalysis.dll" target="tools" />
+    <file src="net462\ClientGenerator.exe" target="tools" />
+    <file src="net462\ClientGenerator.pdb" target="tools" />
+    <file src="net462\ClientGenerator.exe.config" target="tools" />
+    <file src="net462\Orleans.dll" target="tools" />
+    <file src="net462\Orleans.pdb" target="tools" />
+    <file src="net462\OrleansCodeGenerator.dll" target="tools" />
+    <file src="net462\OrleansCodeGenerator.pdb" target="tools" />
```
jdom: Isn't OrleansCodegenerator (and CodeAnalysis) netstandard1.5 now?

jdom: ah, but it might be pulled into that output folder too by ClientGenerator?

vNext/src/NuGet/Microsoft.Orleans.PlatformServices.nuspec
```diff
@@ -0,0 +1,35 @@
...
+  </metadata>
+  <files>
+    <file src="netstandard1.5\Orleans.PlatformServices.xml" target="lib\netstandard1.5" />
+    <file src="netstandard1.5\Orleans.PlatformServices.dll" target="lib\netstandard1.5" />
+    <file src="netstandard1.5\Orleans.PlatformServices.pdb" target="lib\netstandard1.5" />
+    <file src="net462\Orleans.PlatformServices.xml" target="lib\net462" />
+    <file src="net462\Orleans.PlatformServices.dll" target="lib\net462" />
+    <file src="net462\Orleans.PlatformServices.pdb" target="lib\net462" />
+    <file src="$SRC_DIR$Orleans.PlatformServices\**\*.cs" target="src" />
+  </files>
```
jdom: Not really sure how it works, but shouldn't this have a `ref` folder? I'm basing this naively on `System.Runtime` package contents

ReubenBond: No idea, I'll check it out

jdom: nevermind, just tried to do a dotnet pack on it, and it looks like it's not added. Maybe this folder is only used in libraries that are *part* of .NET Standard, not libraries that are built on top of it. Regardless, looks like it's not needed.

test/Tester/CodeGenTests/GeneratorGrainTest.cs
ReubenBond: Yep, for now. I plan to start porting the F# tests very soon

src/OrleansCodeGenerator/RoslynCodeGenerator.cs
ReubenBond: I'm fine with converting this to a concrete collection. It shouldn't be a big deal - this isn't hot path stuff

ReubenBond:
@dotnet-bot test this please

vNext/src/init-tools.cmd
```diff
@@ -26,6 +26,16 @@ if exist "%BUILD_TOOLS_SEMAPHORE%" (
...
+for /f %%i in ('dotnet --version') do set INSTALLED_VERSION=%%i
+set /p REQUIRED_VERSION=<"%~dp0DotnetCLIVersion.txt"
+
+echo INSTALLED: %INSTALLED_VERSION%
+echo REQUIRED: %REQUIRED_VERSION%
+if '%INSTALLED_VERSION%' == '%REQUIRED_VERSION%' (
+  echo Will use system version of dotnet.
+  exit /b -1
+)
+
```
jdom: Can you do this check outside this script? I'd like to avoid changes to this file as it's an exact copy/paste of https://github.com/dotnet/corefx/blob/master/init-tools.cmd and I'd like us to not fork and then have to maintain it separately from them. BTW, we should probably add a comment in the first line of this file saying that

vNext/src/Build.cmd
```diff
@@ -1,4 +1,4 @@
-@if "%_echo%" neq "on" echo off
+REM @if "%_echo%" neq "on" echo off
```
jdom: lol, instead of this you can actually do:
`set _echo=on'
That was the purpose of that line (copied from corefx repo)

ReubenBond: This is just a temp change - I don't want to merge that commit

jdom:
Nevermind, tests run just fine :)

vNext/src/NuGet/Microsoft.Orleans.OrleansCodeGenerator.Build.nuspec
ReubenBond: So the point is that we want Orleans.dll and OrleansCodeGenerator.dll, etc, for the express purpose of running ClientGenerator.exe, which is a net462 binary - therefore we should take the dlls from that path

Issue #2444, opened 2016-11-23
Expose system targets to application code


sergeybykov:
This idea has been discussed more than once in the past. It has been brought up again in https://github.com/dotnet/orleans/issues/2421#issuecomment-262569302.

>So I've been looking for other solutions as my pattern is really opposing being shoe-horned into Orleans Streams.
My first thought was using the IGrainObserver pattern, with a Grain per EventStore Stream, however it would need to be brought to life on Startup, perhaps using a Reminder, then I would have many Reminders ticking for no reason once the Grain was alive.
However I then dug into Reminders further, and it looks like Reminders are implementing exactly what I am after. They can be load balanced across silos, can invoke Grain operations, be registered from a Grain, callback that grain, are persisted and deal with silo loss/addition.
I think some minor adjustments to a copy of LocalReminderService would work very well for this. Unfortunately it looks like a lot of objects LocalReminderService interacts with are Internal, and the Silo also lacks any hooks to register SystemTarget objects.
This is a pity as it really looks like a good fit for what I need here - do you have any plans to take some Internal classes and expose them? Also - am I missing some hook where I am able to register a SystemTarget? The Service would need to be alive before any Grains are activated otherwise they could call a Target that did not yet exist.

I see at least two potential ways, likely complementary, to expose such functionality to application code.

1. Partitioned system service. Essentially a generalized version of the Reminder service, but with an application system target class instead of `LocalReminderService`, e.g. `GrainService`, and a corresponding equivalent to `ReminderRegistry` proxy class, e.g. `GrainServiceClient`. We only create a single `LocalReminderService` system target per silo today, but it shouldn't be too difficult to make that number configurable. In this approach, callers are completely oblivious to partitioning and routing, and it's the responsibility of the `GrainService` to react to repartitioning events and changes to the ring range it owns.

2. Ad-hoc system targets created at run time by application code. This would be similar to how we create pulling agents for stream providers. Application code, a bootstrap provider or grain code or DI, would explicitly create an app system target, e.g. we could call it a `PinnedGrain`, and pass its reference to whatever app code needs to talk to it. The grain reference will be valid for as long as the silo the system target was created on is alive. After that, it will throw on an attempt to send a message via it. In this approach, it's the responsibility of the code creating `PinnedGrains` to manage them and partition the load across them.

jamescarter-le:
[...]

PR #2459, opened 2016-11-29
Grain Services

jamescarter-le:
I have attempted to implement as advised by Sergey the `GrainService` and `GrainServiceClient` types as described by his Option 1 in Issue #2444 
 
 The surface area of this change is large, because I have added a `IGrainService` type, analogues to `ISystemTarget`. I assumed `ISystemTarget` was to remain `internal`, and so I added GrainService as a direct clone of SystemTarget, except it implements LocalReminderService type RingRange knowledge.
 
 Some notes and assumptions made along the way:
  - `SystemTarget` was to remain internal, required change throughout code-base to deal with this additional base `Addressable` type. Changed:
    - `Silo`
    - `SchedulingContext`
    - `ResponseWorkItem`
    - `RequestWorkItem`
    - `IncomingMessageAgent`
    - `InsideRuntimeClient`
    - `ActivationDirectory`
    - `GrainReference`
    - `UniqueKey`
    - `ActivationId`
    - If we were to make `SystemTarget` public, we could prevent all these changes, and derive `GrainClient` directly from `SystemTarget`
 - `GrainServiceClient` added - `ReminderRegistry` is now derived from this and uses the `GrainService` it resolves through `InternalRuntimeClient`
 - `GrainClient` added
    - Basically a `SystemTarget` crossed with the Start, Stop, RingRange properties of `LocalReminderService`
    - Provides virtual methods for derived types to override.
    - Responsible for subscribing to the `IConsistentRingProvider`, keeping this type `internal`
 - `LocalReminderService` modified
    - Now derives from `GrainService` instead of `SystemTarget`
    - Removed all code moved to `GrainClient`
    - This type now solely handles Reminder specific responsibilities.
    - Uses the virtual methods provided in `GrainClient` to hook into the Start/Stopped flow and respond to Ring Range change events.
  
This change is by someone not fully aware of the total scope of Orleans, and will contain significant issues. It is implemented only to convey an idea of how to proceed to allow developers using Orleans to provide RingRange based services, such as those similar to the Reminder Service.

@sergeybykov - Please could you provide feedback on whether SystemTarget should remain internal or be exposed, meaning GrainService can derive from it and we remove the far reaching changes into the code base, however this would require making further types, GrainReference, UniqueKey and more public also.


sergeybykov:
@jamescarter-le, you beat me by a couple of hours - I was prototyping the change for you - https://github.com/sergeybykov/orleans/tree/grainservice. :-)

I made `SystemTarget` and `ISystemTarget` public to minimize the amount of changes. Please take a look/

I also wanted to prototype an encapsulation option in addition to the inheritance one, to see if it would be cleaner.

jamescarter-le:
@sergeybykov Looks like we both came up with the same in the end. I wasn't sure if you wanted to expose SystemTarget, I can make GrainService inherit from SystemTarget instead.

Would you like to implement the encapsulation option, and keep a private SystemTarget inside GrainService, exposed through ISystemTarget and ISystemTargetBase on GrainService?

Happy to continue to progress this item with your guidance.

sergeybykov:
I looked at the encapsulation option briefly. Looks like it would be much more complicated because of the interface to class mapping that happens al load time, and would have to change to inject such mapping at run time to reflect encapsulation. I think it's not worth it.

I think it's better to expose `SystemTarget` and `ISytemTarget` as public at this point. We may decide to change the naming last minute, as somebody suggested on Gitter, but that would be trivial to do.

So, my suggestion is to take the inheritance route, as it's much less disruptive.

>Also - currently SystemTargets have Constant GrainIds - how would you envisage we deal with this when adding application specific ones?

It seems to me that we need to use type codes for such system targets, to differentiate between different types. I don't remember why they are not used today.

So `UniqueKey.NewSystemTargetKey` would have to change to `public static UniqueKey NewSystemTargetKey(short systemId, long typeCodeData=0);` 
and we'll need a new method `GrainId.GetGrainServiceGrainId(short id, TypeInfo grainServiceType)`, similar to `GetSystemTargetGrainId`, to construct a `GrainId` with the type code and a key.

We can start with creating a single `GrainService` activation of a type per silo, and use key 0 for them. We can later expand to support more than one activation per silo, and assign sequential keys to them: 1,2,3...

On the `GrainServiceClient` side we'll need to wire resolution of the `GrainService` type via the type code map, just like we do for grains. Or we can cheat initially by passing TypeInfo of the target `GrainService` type, and calculate its type code as part of `GrainFactory.GetSystemTarget` call.

This is what comes to my mind. I'm curious if people have simpler/better ideas.

sergeybykov:
@jamescarter-le Let me know if I confused you too much. :-) We'd be happy to help here.

jamescarter-le:
Hi @sergeybykov  - Not at all that is very helpful and I am following you.

I will revert my change and implement again but by changing the access modifier of SystemTarget to public instead of making GrainService distinct.

Happy to go with single grain services per Silo for now, with TypeCode as the differentiator in the GrainId.

Quick question, how are you intending to resolve these grain services inside a Grain?
Currently Reminders are helper methods on a Grain. We could add a GetGrainService<T(IGrainServiceClient)> call to a Grain, which queries the Runtime, however perhaps people would want to get these by DI? If we are expecting people to register them by DI, we could inject the GrainReference for the GrainService into the Grain expecting to consume the service. This should be okay as SystemTarget's cannot be called from Clients anyway according to what I can see.

I should be able to do more work on this over the weekend.

sergeybykov:
>Quick question, how are you intending to resolve these grain services inside a Grain?
Currently Reminders are helper methods on a Grain. We could add a GetGrainService<T(IGrainServiceClient)> call to a Grain, which queries the Runtime, however perhaps people would want to get these by DI? If we are expecting people to register them by DI, we could inject the IGrainServiceClient into the Grain expecting to consume the service. This should be okay as SystemTarget's cannot be called from Clients (Silo Clients) anyway according to what I can see.

I think we don't want to expose this via a method on `Grain` class just yet. We can try adding a `GetGrainService<T>` to `IGrainFactory` first, and return a wrapper over a `GrainReference`.

We need a partitioning key to choose the target silo. We can start with implicitly using `CurrentActivationData.GrainReference` just like `GetReminderService` does today, and later consider if we also want to support other kinds of partitioning keys, for example, to allow clients to call grain services, too.

The reason we need a wrapper is that we'll need to recalculate the target silo for every request or upon any change to the cluster state because that will cause repartitioning of the system targets. `RegisterOrUpdateReminder` calls `GetReminderService` -> `MapGrainReferenceToSiloRing` -> `ConsistentRingProvider.GetPrimaryTargetSilo(hashCode)` on every call for that reason.

jamescarter-le:
So if we return a GrainReference from GrainFactory, then we would be returning the IGrainService (SystemTarget) reference, as the GrainServiceClient is not IAddressable.

In this case - I'm not sure I follow what the GrainServiceClient should be doing - I thought that was its job, to address messages to the correct GrainService? See the GrainServiceClient.GrainService property:

https://github.com/dotnet/orleans/pull/2459/files#diff-3748abb74ef2a327f5a9e859d9ca19d2R15

I understand the to be addressed may move, and thought it was the GrainServiceClient's job to address the calls. Perhaps I've misunderstood.

src/OrleansRuntime/ReminderService/LocalReminderServiceFactory.cs
```diff
@@ -17,14 +17,17 @@ public LocalReminderServiceFactory()
         internal IReminderService CreateReminderService(Silo silo, IGrainFactory grainFactory, TimeSpan iniTimeSpan)
         {
             var reminderServiceType = silo.GlobalConfig.ReminderServiceType;
-            logger.Info("Creating reminder system target for type={0}", Enum.GetName(typeof(GlobalConfiguration.ReminderServiceProviderType), reminderServiceType));
+            logger.Info("Creating reminder grain service for type={0}", Enum.GetName(typeof(GlobalConfiguration.ReminderServiceProviderType), reminderServiceType));
+
+            // At the point this is called the GrainClassData is not available for IReminderService - is this why Constants are used?
```
jamescarter-le: Please advise how to get the TypeCode for the Grain.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -284,6 +286,8 @@ internal Silo(SiloInitializationParameters initializationParams)
                         services.TryAddSingleton<IConsistentRingProvider>(
                             sp => new ConsistentRingProvider(this.initializationParams.SiloAddress));
                     }
+
+                    services.TryAddSingleton<IGrainServiceClient<IReminderService>, GrainServiceClient<IReminderService>>();
```
jamescarter-le: Added to resolve the IReminderService from the InsideRuntimeClient in a common way.
When we add user defined IGrainServices we grab the GrainServiceClient from the DI.

test/Tester/GrainServiceTests/GrainServiceTests.cs
```diff
@@ -0,0 +1,43 @@
...
+            var options = new TestClusterOptions(1);
+            options.ClusterConfiguration.UseStartupType<GrainServiceStartup>();
+
+            return new TestCluster(options);
+        }
+
+        [Fact, TestCategory("BVT"), TestCategory("Functional"), TestCategory("GrainServices")]
+        public async Task SimpleInvokeGrainService()
+        {
+            // We need to get the Silo to create the GrainService instances and register them as SystemTargets.
```
jamescarter-le: I need to find a way to bootstrap creating the GrainService SystemTargets in the Silo on Startup.

src/OrleansRuntime/Services/GrainServiceClient.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+        
+        /// <summary>
+        /// Resolves the Grain Reference invoking this request.
+        /// </summary>
+        protected GrainReference CallingGrainReference
+        {
+            get
+            {
+                return RuntimeClient.Current.CurrentActivationData.GrainReference;
```
ReubenBond: I'm trying to remove statics like this so that we can enable non-static grain clients

jamescarter-le: Thanks Reuben, It looks like IInternalGrainFactory and IRuntimeClient are passed into IServiceProvider so I can take an IServiceProvider in GrainServiceClient constructor as that is a public interface.

I will refactor this later and you can review.

jamescarter-le:
I've added code to construct GrainServices in the Silo, but when I attempt to run the Tests I'm getting:

Exception thrown: 'Orleans.CodeGenerator.CodeGenerationException' in OrleansCodeGenerator.dll

Additional information: (1,18913): error CS0122: 'ITestHooksSystemTarget' is inaccessible due to its protection level

I wonder if someone wouldn't mind having a look at this for me - I can't see where I am attempting to expose anything internal?

jamescarter-le:
Currently the Silo is constructing the Grain Service for each Service using Reflection to locate the constructor, supplied by Type names in the Config.

I'm sure you guys may have a better way of doing this, happy to take feedback.

test/Tester/GrainServiceTests/GrainServiceTests.cs
jamescarter-le: Done.

sergeybykov:
I submitted a PR - https://github.com/jamescarter-le/orleans/pull/2 - with a few minor cleanup type of changes. Otherwise, looks good to me. I wonder what others think.

sergeybykov:
I rebased your branch on top of current master and pushed the result to https://github.com/sergeybykov/orleans/tree/2459-3. Can you use that to update the PR branch?

We are looking to fix the two vNext build errors. I have trouble with the tooling. I asked @attilah to help.

sergeybykov:
I pushed another commit to https://github.com/sergeybykov/orleans/tree/2459-3 to resolve the .NET Core build issue (thanks to @attilah).

Let us know if you are not sure how to integrate the rebased branch into yours.

jamescarter-le:
Apologies I'm short on time at the moment - I will look to get the PR branch merged in shortly.

sergeybykov:
@jamescarter-le If you don't have time, we could merge my rebased branch of this PR with all your commits instead. The commits are what matters in the end anyway. :-)

jamescarter-le:
Hi Sergey, yes sorry happy to you to get the changes in however is easiest
for yourself.

Cheers, James

ReubenBond:
Merged via #2531

PR #2460, opened 2016-11-29
fix test UnitTests.CodeGeneration.CodeGeneratorTests.ServiceInterface_IsGrainClass

xiazen:
fix test UnitTests.CodeGeneration.CodeGeneratorTests.ServiceInterface_IsGrainClass

full name of `Orleans.Runtime.SystemTarget` is "Orleans.Runtime.SystemTarget" instead of "Orleans.Runtime.SystemTarget, OrleansRuntime"

### DIFF
--- a/src/Orleans/CodeGeneration/TypeUtils.cs
+++ b/src/Orleans/CodeGeneration/TypeUtils.cs
@@ -394,29 +394,6 @@ namespace Orleans.Runtime
             return !IsGeneratedType(type);
         }
 
-        public static bool IsSystemTargetClass(Type type)
-        {
-            Type systemTargetType;
-            if (!TryResolveType("Orleans.Runtime.SystemTarget, OrleansRuntime", out systemTargetType)) return false;
-
-            var systemTargetInterfaceType = typeof(ISystemTarget);
-            var systemTargetBaseInterfaceType = typeof(ISystemTargetBase);
-#if !NETSTANDARD
-            if (type.Assembly.ReflectionOnly)
-            {
-                systemTargetType = ToReflectionOnlyType(systemTargetType);
-                systemTargetInterfaceType = ToReflectionOnlyType(systemTargetInterfaceType);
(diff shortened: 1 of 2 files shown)


xiazen:
> The fully qualified name of the type, including its namespace but not its assembly; 

 from [documentation ](https://msdn.microsoft.com/en-us/library/system.type.fullname(v=vs.110).aspx)

src/Orleans/CodeGeneration/TypeUtils.cs
```diff
@@ -397,7 +397,7 @@ public static bool IsGrainClass(Type type)
         public static bool IsSystemTargetClass(Type type)
         {
             Type systemTargetType;
-            if (!TryResolveType("Orleans.Runtime.SystemTarget, OrleansRuntime", out systemTargetType)) return false;
+            if (!TryResolveType("Orleans.Runtime.SystemTarget", out systemTargetType)) return false;
```
benjaminpetit: Why not use something like "typeof(ISystemTarget).FullName"?

jdom: If I remember correctly, because `SystemTarget` is defined in OrleansRuntime and not Orleans

benjaminpetit: Yes, but **I**SystemTarget is in Orleans, and since all SystemTarget should implement this interface, it should be fine, no?

ReubenBond: What if we just moved this method to the test where it's used and then referenced `SystemTarget` directly? Like this:
image

xiazen: if this method's only purpose is for testing, I think it's fine to move it to the test.
Anyone know some context about this method?

xiazen: Will do

xiazen:
updated with feedback..

xiazen:
forgot I shouldn't git amend ..

Issue #2366, opened 2016-10-29
Remove usage of named mutexes


galvesribeiro:
As part of the first test of Orleans in .Net Core (#2145), I tried to run the @jdom Hello World sample project in Linux but it throw `PlatformNotSupportedException`:

```
Named synchronization primitives are not supported on this platform.
Exception = System.PlatformNotSupportedException: Named synchronization primitives are not supported on this platform.
   at System.Threading.EventWaitHandle..ctor(Boolean initialState, EventResetMode mode, String name, Boolean& createdNew, EventWaitHandleSecurity eventSecurity)
   at Orleans.Runtime.Host.SiloHost.StartOrleansSilo(Boolean catchExceptions)
```

That is one of the issues on coreclr repo related to it: https://github.com/dotnet/coreclr/issues/3422

We need to find a workaround for it.

jdom:
BTW,  I used SiloHost in my samole, but could have used Silo directly. 
Still, we probably need to address this, just letting you know that it shouldn't be a blocker

galvesribeiro:
@jdom yes, it is just a blocker on linux. There are other thtings I had to change on your hello world but we can talk offline when you get back.

jdom:
What I meant is this is not a blocker, since you are not required to use SiloHost, so there is a valid workaround. That said, we still need to fix it

galvesribeiro:
Oh! Ok, will change to silo directly and check it out

galvesribeiro:
@jdom just confirmed. Replacing `SiloHost` with `Silo` did the trick: https://twitter.com/galvesribeiro/status/792455412746481664

jdom:
Any takers to fix this one?

galvesribeiro:
@jdom will ping u Monday and I can probably do that after align with you.

galvesribeiro:
@JacobSnyder done on #2462

PR #2462, opened 2016-11-30
Ignore EventWaitHandle when not available

galvesribeiro:
As part of #2145, in order to make the `SiloHost` to start without problem, we have to not rely on `EventWaitHandle` since it is not available on linux because named mutexes are not supported in it.

For the crawl phase, we are just catching `PlatformNotSupportedException`, logging and moving forward. In walk phase, we should implement event signaling for graceful shutdown on linux by using POSIX semaphores or SIGNALs.

This is a fix for #2366.


jdom:
Thanks @galvesribeiro! One step closer to running in linux! :)

galvesribeiro:
@jdom it actually run just fine. The problem is build, which still windows-only due to the codegen.

PR #2466, opened 2016-12-01
Avoid getting duplicate type errors when using runtime codegen.

jdom:
When using runtime codegen in .NET Core, there are some issues with duplicate types being exposed by the .NET Core runtime. We are temporarily using a workaround to avoid linking to those duplicate types that are outside of the exposed façades.

### DIFF
--- a/src/OrleansCodeGenerator/CodeGeneratorCommon.cs
+++ b/src/OrleansCodeGenerator/CodeGeneratorCommon.cs
@@ -72,6 +72,18 @@ namespace Orleans.CodeGenerator
             // Generate the code.
             var options = new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary);
 
+#if NETSTANDARD
+            // CoreFX bug https://github.com/dotnet/corefx/issues/5540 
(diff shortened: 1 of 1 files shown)


src/OrleansCodeGenerator/CodeGeneratorCommon.cs
```diff
@@ -72,6 +72,18 @@ public static GeneratedAssembly CompileAssembly(GeneratedSyntax generatedSyntax,
             // Generate the code.
             var options = new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary);
 
+#if NETSTANDARD
```
ReubenBond: I think this should be NETSTANDARD_TODO

ReubenBond: because we want to make sure we fix it

jdom: hmmm, do you think so? it's not something that we need to address ASAP, is it? We are still depending on this external issue.

ReubenBond: Yeah, it's fine either way really - not going to push back. I'm happy to merge as soon as build passes

PR #2468, opened 2016-12-01
Enable Portable PDBs for generated assembly

attilah:
Make generated code to produce portable PDB to support cross-platform initiatives.


src/OrleansCodeGenerator/CodeGeneratorCommon.cs
```diff
@@ -108,7 +109,10 @@ public static GeneratedAssembly CompileAssembly(GeneratedSyntax generatedSyntax,
             var symbolStream = emitDebugSymbols ? new MemoryStream() : null;
             try
             {
-                var compilationResult = compilation.Emit(outputStream, symbolStream);
+                var emitOptions = new EmitOptions()
+                    .WithDebugInformationFormat(DebugInformationFormat.PortablePdb);
+
+                var compilationResult = compilation.Emit(outputStream, symbolStream, options: emitOptions);
```
jdom: would this work well even if `emitDebugSymbols` is false (and hence `symbolStream` is null)?

PR #2473, opened 2016-12-02
Fix binding redirects for functional tests

jdom:
For when tests are run in a single output folder where runtime codegen DLL is available.


jdom:
I think ideally we should just explicitly reference runtime codegen and have a way to opt out of it, as @ReubenBond also suggested. Issues like these are hard to troubleshoot, because depending on which assemblies just happen to be there, a Silo could behave very differently, and that's what happens when we build every project into a single folder.

xiazen:
LGTM. merging it

Issue #1079, opened 2015-11-25
Documentation for OnActivateAsync?


svick:
When looking into answering [a question on SO](http://stackoverflow.com/q/33920495/41071) I found some mentions of the `OnActivateAsync()` method, but pretty much the only documentation about it I could find is [a short mention in a tutorial](http://dotnet.github.io/orleans/Step-by-step-Tutorials/A-Service-is-a-Collection-of-Communicating-Actors#the-life-of-an-actor) and [few mentions in passing in another tutorial](http://dotnet.github.io/orleans/Getting-Started-With-Orleans/Developing-a-Grain).

Is there some documentation about this method that I'm overlooking? If not, shouldn't there be more of it? Maybe a detailed explanation of the lifecycle of a grain?

sergeybykov:
I think this is a valid point. We don't seem to have a doc section clearly describing the lifecycle of a grain. Do you feel like contributing one?

svick:
I don't feel like I know Orleans enough to write it, so no, sorry.

sergeybykov:
That's fine. :-)

amamh:
@svick 

Doesn't seem like there's anything complicated about it.

Constructor -> ReadStateAsync -> OnActivateAsync

This will only happen if the grain is not already active. Note that being subscribed to a stream keeps the grain active.

ashkan-saeedi-mazdeh:
@amamh  @sergeybykov 
I can write it but is really `OnReadStateAsync` called on activation?

amamh:
Well, I believe that's the whole point of a state. It's going to be read before the grain is activated i.e. restore its previous state. The idea is that a grain/actor conceptually always exists with no lifecycle and that's done by persisting its state, so it can die/wake up without the knowledge of the users. Clients see actors as always "there", that's the point of "virtual" actors.

sergeybykov:
@amamh You are right. One small correction - it is `ReadStateAsync`. It is called on the storage provider if the grain class extends `Grain<T>`.

@ashkan-saeedi-mazdeh Your help will be very much appreciated!

ashkan-saeedi-mazdeh:
@sergeybykov  my pleasure.
@amamh  yeah , OnReadAsync confused me, I thought there is a virtual method which I didn't know about.

jason-bragg:
>I can write it..

@ashkan-saeedi-mazdeh :+1: 
Thanks!

miker1423:
[...]

PR #2478, opened 2016-12-03
Added Grain Life Cycle docs

ashkan-saeedi-mazdeh:
Based on #1079 and the fact that the other guy did not do it and I had a bit of time today, I wrote this.

I don't know if I am fully correct about idle state or not. also I messed up commits, building the docfx caused a lot of unwanted changes. let me know if I should do anything.


src/Documentation/Getting-Started-With-Orleans/Grain-LifeCycle.md
```diff
@@ -0,0 +1,55 @@
...
+## callbacks 
+
+Just like the class which can have constructors and (destructors in some languages), grains have
+`OnActivateAsync` and `OnDeactivateAsync` virtual methods which you can override.
+
+## Different grain types and their differences regarding life cycle
+
+Your grains can be persistent and store their state in storage
+or have no state which is stored in storage. 
+In case of the later, all activations are created equal and there is no difference between them.
```
galvesribeiro: What you mean by that? Even without state a grain is differentiated by its ID/Key, not the state.

src/Documentation/Getting-Started-With-Orleans/Grain-LifeCycle.md
```diff
@@ -0,0 +1,55 @@
...
+- another grain or a client calls a method of a grain
+- the grain gets activated and brought to memory
+  - the grain state will be read from storage if any
+  - if successful then `OnActivateAsync` is called
+- The grain will process some messages
+- The grain remains idle for some time
+- Runtime will call `OnDeactivateAsync`
+- Runtime removes the grain from memory
+
+Now if a Silo gets a shutdown request gracefully, the message cueues of all activated grains will be forwarded to another silo with the grain activation,
```
galvesribeiro: queue ?

galvesribeiro:
I think you can talk about `GetGrain()` and how it get just reference and not actually activate a grain. The activation happens on the target silo when the first message arrive and is made internally by the runtime. The client/caller never have an activation by itself, but the return of `GetGrain()` is a reference/pointer to the the target grain running somewhere else in the cluster hence why the method-shipping pattern used here.

src/Documentation/Getting-Started-With-Orleans/Grain-LifeCycle.md
```diff
@@ -0,0 +1,64 @@
...
+> `ReadStateAsync` is called before calling `OnActivateAsync` and if it fails then the grain will not be activated.
+
+
+## Sequence of events in a grain's life cycle
+
+So the life cycle of a grain is like this
+
+- another grain or a client calls a method of a grain
+- the grain gets activated and brought to memory
+  - the grain state will be read from storage if any
```
shanegrueling: Is this the place the constructor of a grain gets called and dependency injection takes place?

Might be an important part of the life cycle of a grain.

ashkan-saeedi-mazdeh:
@shanegrueling  @galvesribeiro  Improved based on your feedback. Thank you both

src/Documentation/Advanced-Concepts/Activation-Garbage-Collection.md
```diff
@@ -132,6 +132,8 @@ You can also instruct the runtime to deactivate the grain next time it becomes i
 ``` csharp
   protected void DeactivateOnIdle()
 ```
+An activation is considered idle if it is not processing any message at the moment
+and has an empty inbox (i.e. there are no requests waiting to be processed by it).
```
sergeybykov: This is not quite correct - an activation will get deactivated after completing the request it is processing even if it has more requests enqueued (they will get forwarded to a new activation).

Also, terminology wise, we never talk about inboxes. Using a term like that here would be confusing I'm almost sure.

src/Documentation/Getting-Started-With-Orleans/Grain-LifeCycle.md
```diff
@@ -0,0 +1,65 @@
+---
+layout: page
+title: Grain LifeCycle
+---
+
+
+## Grain LifeCycle
+
+We say grains virtually always exist but here we want to describe different stages of a grain's life.
```
sergeybykov: This is confusing verbiage to me. I'm afraid that the analogies, such as "like a class definition" and "instantiated with a `new` keyword in C#" will likely only contribute to the confusion. I think we can be much more straightforward here. For example, something along the following lines.

"Grains are logical entities that always exist, virtually, and have stable logical identities (keys). Application code never creates or destroys grains. Instead, it acts as if all possible grains are always in memory and available for processing requests.

Grains get physically instantiated, activated, by the Orleans runtime automatically on an as-needed to process incoming requests. After a grain has been idle for a certain amount of time, the Orleans runtime automatically removes, deactivates, it from memory.

A physical instances of a grain in memory is called a grain activation. Grain activations are invisible to application code as well as the process of activating and deactivating them. Only the grain itself can be aware of that  - by overriding virtual methods `OnActivateAsync` and `OnDeactivateAsync` that get invoked upon activation and deactivation of the grain respectively.

Over the course of its eternal, virtual, life a grain goes through the cycles of activations and deactivations, staying always available for callers to invoke it, whether it is in memory at the time of the call or not."

src/Documentation/Getting-Started-With-Orleans/Grain-LifeCycle.md
```diff
@@ -0,0 +1,65 @@
...
+The grain's instances are differentiated by their key and a grain with a specific key is either activated or not.
+You don't control this by using construction/destruction keywords and instead Orleans runtime manages this for you.
+
+All possible grains in the whole key space virtually always exist
+and it means whenever you call a method of a grain, if it is not active, it will get activated by the runtime.
+Also after the grain passes a certain amount of time in idle state, the runtime will deactivate the grain (i.e. destroys it).
+A grain is in idle state if it is not currently processing a message and its queue of messages is empty.
+
+When you get a grain reference in a client or in another grain by using `GetGrain`,
+you only get a reference to a grain and you don't get the activation.
```
sergeybykov: Maybe "you only get a proxy object with a logic address (identity) of the grain, but not its physical address."?

src/Documentation/Getting-Started-With-Orleans/Grain-LifeCycle.md
```diff
@@ -0,0 +1,65 @@
...
+When you call a method using the reference,then the grain will be activated in a silo (if it is not already active)
+and your reference sends messages to the activation by calling methods and points to the activation logically.
+
+`GetGrain` itself is not asynchronous, can not fail and only returns the logical reference which points to the activation.
```
sergeybykov: Maybe "`GetGrain` itself is an inexpensive local operation of constructing a proxy object with an embedded identity of the target grain."?

src/Documentation/Getting-Started-With-Orleans/Grain-LifeCycle.md
```diff
@@ -0,0 +1,65 @@
...
+
+> `ReadStateAsync` is called before calling `OnActivateAsync` and if it fails then the grain will not be activated.
+
+
+## Sequence of events in a grain's life cycle
+
+So the life cycle of a grain is like this
+
+- another grain or a client calls a method of a grain
+- the grain gets activated and brought to memory
```
sergeybykov: I would clarify explicitly "if the grain is not activated anywhere in the cluster."
Then I would say that "an instance of the grain class gets created" instead of "grain gets activated and brought to memory" to be more precise.

ashkan-saeedi-mazdeh:
@sergeybykov  I'll apply all of the comments and will push again in 2-3 days.

sergeybykov:
@ashkan-saeedi-mazdeh, thanks so much for doing it!

ashkan-saeedi-mazdeh:
@sergeybykov  I've done the required changes. 

It is a case of the stuff you would do it faster and easier if you did it instead of me but I'll get improved :)

sergeybykov:
@ashkan-saeedi-mazdeh I submitted https://github.com/ashkan-saeedi-mazdeh/orleans/pull/1 with some edits. Please take a look.

ashkan-saeedi-mazdeh:
@sergeybykov  Looked at it and merged. Perfect! Thanks.

ashkan-saeedi-mazdeh:
@sergeybykov  Please close #1079 since I wasn't the issue owner, I couldn't do it.

sergeybykov:
Thank you, @ashkan-saeedi-mazdeh!

PR #2484, opened 2016-12-06
CodeGen: Eliminate unnecessary constraints on generic types

ReubenBond:
Eliminates some unnecessary constraints on the kinds of generic type implementations which codegen will consider implementing serializers for.

The expectation is that this should not hurt code generation performance significantly and that it removes some uneccessary dependencies on other classes from the code generator.

@attilah this is for you.


attilah:
@ReubenBond please LMK about the results of functionals when it finished running.

PR #2485, opened 2016-12-06
Reduce use of globals in SiloProviderRuntime

ReubenBond:
* Removed `SiloProviderRuntime.Instance` global
* Adds a new interface, `ISiloRuntimeClient`, which contains methods moved from `SiloProviderRuntime`.
* `GrainCreator` now always uses `ActivatorUtilities.CreateFactory` instead of `Activator.CreateInstance`, meaning that grain classes with multiple public constructors cannot be constructed unless they are explicitly registered. @jdom are you fine with this? This behavior is consistent with ASP.NET as far as I know.
* Some methods on `SiloProviderRuntime` are now forwarded to `InsideRuntimeClient` where I think they are a more natural fit. We can move them elsewhere if desired.

This is working towards #467. The idea is to make multiple smaller commits to remove globals so that we can implement support for multiple distinct grain clients per AppDomain.

### DIFF
--- a/src/OrleansRuntime/Core/InsideRuntimeClient.cs
+++ b/src/OrleansRuntime/Core/InsideRuntimeClient.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Reflection;
 using System.Reflection.Emit;
 using System.Threading.Tasks;
@@ -10,19 +11,17 @@ using Orleans.CodeGeneration;
 using Orleans.Runtime.Configuration;
 using Orleans.Runtime.ConsistentRing;
 using Orleans.Runtime.GrainDirectory;
-using Orleans.Runtime.Providers;
 using Orleans.Runtime.Scheduler;
 using Orleans.Serialization;
 using Orleans.Storage;
 using Orleans.Streams;
-
 
 namespace Orleans.Runtime
 {
     /// <summary>
     /// Internal class for system grains to get access to runtime object
     /// </summary>
-    internal class InsideRuntimeClient : IRuntimeClient
+    internal class InsideRuntimeClient : ISiloRuntimeClient
     {
         private static readonly Logger logger = LogManager.GetLogger("InsideRuntimeClient", LoggerType.Runtime);
         private static readonly Logger invokeExceptionLogger = LogManager.GetLogger("Grain.InvokeException", LoggerType.Application);
@@ -37,7 +36,6 @@ namespace Orleans.Runtime
         public TimeSpan ResponseTimeout { get; private set; }
         private readonly GrainTypeManager typeManager;
         private readonly Lazy<ISiloStatusOracle> siloStatusOracle;
-        private IGrainTypeResolver grainInterfaceMap;
 
         internal readonly IConsistentRingProvider ConsistentRingProvider;
 
@@ -342,7 +340,7 @@ namespace Orleans.Runtime
                     var request = (InvokeMethodRequest) message.BodyObject;
                     if (request.Arguments != null)
                     {
-                        CancellationSourcesExtension.RegisterCancellationTokens(target, request, logger);
+                        CancellationSourcesExtension.RegisterCancellationTokens(target, request, logger, this);
                     }
 
                     var invoker = invokable.GetInvoker(request.InterfaceId, message.GenericGrainType);
@@ -399,7 +397,7 @@ namespace Orleans.Runtime
         {
             // If the target has a grain-level interceptor or there is a silo-level interceptor, intercept the
             // call.
-            var siloWideInterceptor = SiloProviderRuntime.Instance.GetInvokeInterceptor();
+            var siloWideInterceptor = this.CurrentStreamProviderRuntime.GetInvokeInterceptor();
             var grainWithInterceptor = target as IGrainInvokeInterceptor;
             
             // Silo-wide interceptors do not operate on system targets.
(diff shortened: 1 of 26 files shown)


jdom:
> GrainCreator now always uses ActivatorUtilities.CreateFactory instead of Activator.CreateInstance, meaning that grain classes with multiple public constructors cannot be constructed unless they are explicitly registered. @jdom are you fine with this? This behavior is consistent with ASP.NET as far as I know

I'm on board, I think is the way to go, but @sergeybykov did push back on this last time as it would be a breaking change

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -262,14 +267,7 @@ internal Silo(SiloInitializationParameters initializationParams)
...
                     services.TryAddSingleton<Func<IGrainRuntime>>(sp => () => sp.GetRequiredService<IGrainRuntime>());
-                    if (usingCustomServiceProvider)
-                    {
-                        services.TryAddSingleton<GrainCreator>();
-                    }
-                    else
-                    {
-                        services.TryAddSingleton(sp => new GrainCreator(null, sp.GetRequiredService<Func<IGrainRuntime>>()));
-                    }
+                    services.TryAddSingleton<GrainCreator>();
```
sergeybykov: I'm trying to understand if this is correct. Today we explicitly pass `null` as a first argument to `new GrainCreator`, which triggers it to use `Activator.CreateInstance(type)` for instantiating grain activations. This will make it always use `ActivatorUtilities.CreateFactory(type, Type.EmptyTypes)`.

Why do we need to change this? Seems unrelated to the stated purpose of the PR.

src/OrleansRuntime/Core/InsideRuntimeClient.cs
```diff
@@ -580,23 +578,11 @@ public void ReceiveResponse(Message message)
             }
         }
 
-        public Logger AppLogger
-        {
-            get { return appLogger; }
-        }
+        public Logger AppLogger => appLogger;
```
sergeybykov: In the future, it would help to group such non-essential changes in a separate commit to simplify reviewing. It's pretty easy to do with SourceTree.

src/OrleansRuntime/GrainTypeManager/TypeManager.cs
```diff
@@ -1,33 +1,34 @@
...
+        private readonly ImplicitStreamSubscriberTable implicitStreamSubscriberTable;
 
-        internal TypeManager(SiloAddress myAddr, GrainTypeManager grainTypeManager)
+        internal TypeManager(
+            SiloAddress myAddr,
+            GrainTypeManager grainTypeManager,
+            ImplicitStreamSubscriberTable implicitStreamSubscriberTable)
             : base(Constants.TypeManagerId, myAddr)
         {
+            if (implicitStreamSubscriberTable == null) throw new ArgumentNullException(nameof(implicitStreamSubscriberTable));
```
sergeybykov: A minor thing, but this moved throwing of an ANE to the constructor `TypeManager` while it used to happen only when `GetImplicitStreamSubscriberTable` was called. Today this won't break because the table happens to get initialized inside `StreamingInitialize()` *before* `TypeManager` is instantiated. But this may easily break in the future if we rearrange things or make some of the streaming components pluggable and maybe even optional.

src/OrleansRuntime/Silo/Silo.cs
ReubenBond: I made a note about this above (you can see @jdom's comment on this PR).
This is needed because some tests require access to things which were previously static but are now no longer exposed in a static manner. Therefore, we inject them into the grains used by the tests. The issue is that under the existing setup, constructor injection is not enabled for grains unless the user specifies a startup class - which is weird because even if we don't actually inject anything in our startup class, it enables dependency injection. It's a violation of the principle of least surprise.

src/OrleansRuntime/Core/InsideRuntimeClient.cs
```diff
@@ -820,5 +790,89 @@ public void BreakOutstandingMessagesToDeadSilo(SiloAddress deadSilo)
                 }
             }
         }
+        
+        public StreamDirectory GetStreamDirectory()
+        {
```
ReubenBond: These are originally from `SiloProviderRuntime`

src/OrleansRuntime/GrainTypeManager/TypeManager.cs
ReubenBond: That's true. The NRE is not really needed here. The value won't be null even if it hasn't yet been initialized - DI  is used to construct `ImplicitStreamSubscriberTable` and then `TypeManager` is constructed using it inside `Silo.CreateSystemTargets` like so:
```C#
var implicitStreamSubscriberTable = Services.GetRequiredService<ImplicitStreamSubscriberTable>();
this.RegisterSystemTarget(new TypeManager(this.SiloAddress, this.typeManager, implicitStreamSubscriberTable));
```

Currently, it's initialized in `Silo.DoStart` before the call to `CreateSystemTargets`,
On the client side, it's initialized by calling into the gateway (which hits the `TypeManager` system target).

So I believe this is sound

src/OrleansRuntime/Core/InsideRuntimeClient.cs
ReubenBond: Yep, sorry. I'll shake the habit.

ReubenBond:
Addressed all feedback, functionals should all be passing - 2 were failing for the same reason (multiple constructors), but they're fixed.

src/OrleansRuntime/GrainTypeManager/TypeManager.cs
ReubenBond: In the latest revision, you'll see that this aligns to what Ben is doing since his heterogenous silos PR

sergeybykov:
>EDIT: btw, I undid most of the superfluous/stylistic changes you pointed out @sergeybykov

Oh, sorry, I didn't mean to revert them. Cleanup changes like that are good. It's just easier to review when such changes are in a separate commit. If you still have them around and can easily submit, please do.

PR #2490, opened 2016-12-07
fix failing functional tests

xiazen:
- add binding redirects
- add remove PersistenGrainTests from vNext, since it uses
  AppDomainTestHooks, which isn't fulling working in vNext yet
- update test.cmd


xiazen:
don't merge yet, I want to verify with vso build

xiazen:
fixed most tests failing

PR #2500, opened 2016-12-08
fix vnext functional tests

xiazen:
-get FluentAssertions version consistent to 4.17.0


xiazen:
this should fixed all functional failing in vnext . already tested functional in vso for it

PR #2502, opened 2016-12-08
Fix #2358: Invoke interceptor broken for generic grains

### LINKED ISSUE #2358: Invoke interceptor broken for generic grains
Just upgraded to 1.3.0 and am having issue with generic grains.

An example interface and class that's showing this issue:

```
public interface IGenericTest<T> : IGrainWithIntegerKey
{
    Task<T> PrintType(T obj);
}
```

```
public class GenericTestGrain<T> : Grain, IGenericTest<T>
{
    public Task<T> Print(T obj)
    {
        Debug.WriteLine("TEST");
        return Task.FromResult(obj);
    }
}
```

Also have invoke interceptor setup up, for simplicity looks like this: 

```
providerRuntime.SetInvokeInterceptor((method, request, grain, invoker) =>
{
    return invoker.Invoke(grain, request);
});
```

then using the grain like:

```
var grain = await GrainFactory.GetGrain<IGenericTest<int>>(0); // Runs without error.
await grain.Print(1);
```

Getting the grain seems to be fine, but when I call the method on the grain I get:

```
System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.

at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
at Orleans.InterceptedMethodInvokerCache.GetInterfaceToImplementationMap(Int32 interfaceId, Type implementationType)
at Orleans.InterceptedMethodInvokerCache.CreateInterceptedMethodInvoker(Type implementationType, Int32 interfaceId, IGrainMethodInvoker invoker)
at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
at Orleans.InterceptedMethodInvokerCache.GetOrCreate(Type implementationType, Int32 interfaceId, IGrainMethodInvoker invoker)
at Orleans.Runtime.InsideRuntimeClient.InvokeWithInterceptors(IAddressable target, InvokeMethodRequest request, IGrainMethodInvoker invoker)
at Orleans.Runtime.InsideRuntimeClient.<Invoke>d__57.MoveNext()
```

When I remove the SetInvokeInterceptor call everything is happy. Seems like the code is trying to use the typed interface name as the key instead of the generic.

ReubenBond:
Fixes #2358. 

* Previously there were two methods used to get an id for a grain interface, with slightly different behaviour. That seemed to be erroneous and so `ComputeInterfaceId` was removed in favour of `GetGrainInterfaceId`.
* `GetGrainInterfaceId` must take into account whether or not the grain is a constructed generic. A constructed generic cannot be a grain interface as specified in code. Therefore its generic type definition is used instead.
* The `Type.GetRuntimeInterfaceMap(Type)` method does not take into account inherited interfaces, only the current interface. This would cause interceptors to receive `null` instead of the implementation's `MethodInfo`.
* Mapping between methods on a generic interface and their counterparts on a concrete class is not entirely obvious, but it is now implemented (see `GetGenericInterfaceInvoker`). A different method was used in order to keep differences clear.
* Added an additional test to ensure this case is covered.
* Changed tests to BVTs since they're quick & fairly important.

Thank you for reporting, @dancvogel, and thanks for the prod, @richorama 😄

### DIFF
--- a/src/Orleans/Core/InterceptedMethodInvokerCache.cs
+++ b/src/Orleans/Core/InterceptedMethodInvokerCache.cs
@@ -37,7 +37,7 @@ namespace Orleans
             if (!this.invokers.TryGetValue(implementationType, out invokerMap))
             {
                 // Generate an the invoker mapping using the provided invoker.
-                this.invokers[implementationType] = invokerMap = GetInterfaceToImplementationMap(implementationType, invoker);
+                this.invokers[implementationType] = invokerMap = CreateInterfaceToImplementationMap(implementationType, invoker);
             }
 
             // Attempt to get the invoker for the provided interfaceId.
@@ -57,12 +57,28 @@ namespace Orleans
         /// <param name="implementationType">The implementation type.</param>
         /// <param name="invoker">The grain method invoker.</param>
         /// <returns>The mapped interface.</returns>
-        private static Dictionary<int, InterceptedMethodInvoker> GetInterfaceToImplementationMap(
+        private static Dictionary<int, InterceptedMethodInvoker> CreateInterfaceToImplementationMap(
             Type implementationType,
             IGrainMethodInvoker invoker)
         {
-            // This method creates a map from interfaceId -> intercepted invoker.
-            if (implementationType.IsConstructedGenericType) return GetGenericInterfaceInvoker(implementationType, invoker);
+            if (implementationType.IsConstructedGenericType) return CreateMapForConstructedGeneric(implementationType, invoker);
+            return CreateMapForNonGeneric(implementationType, invoker);
+        }
+
+        /// <summary>
+        /// Creates and returns a map from interface id to intercepted invoker for non-generic or open-generic types.
+        /// </summary>
+        /// <param name="implementationType">The implementation type.</param>
+        /// <param name="invoker">The underlying invoker.</param>
+        /// <returns>A map from interface id to intercepted invoker.</returns>
+        private static Dictionary<int, InterceptedMethodInvoker> CreateMapForNonGeneric(Type implementationType, IGrainMethodInvoker invoker)
+        {
+            if (implementationType.IsConstructedGenericType)
+            {
+                throw new InvalidOperationException(
+                    $"Type {implementationType} passed to {nameof(CreateMapForNonGeneric)} is a constructed generic type.");
+            }
+
             var implementationTypeInfo = implementationType.GetTypeInfo();
             var interfaces = implementationType.GetInterfaces();
 
@@ -102,23 +118,22 @@ namespace Orleans
         }
 
         /// <summary>
-        /// Maps the interfaces of the provided <paramref name="implementationType"/>.
+        /// Creates and returns a map from interface id to intercepted invoker for constructed generic types.
         /// </summary>
-        /// <param name="implementationType">The implementation type, which must be a concrete generic type.</param>
-        /// <param name="invoker">The grain method invoker.</param>
-        /// <returns>The mapped interface.</returns>
-        private static Dictionary<int, InterceptedMethodInvoker> GetGenericInterfaceInvoker(
+        /// <param name="implementationType">The implementation type.</param>
+        /// <param name="invoker">The underlying invoker.</param>
+        /// <returns>A map from interface id to intercepted invoker.</returns>
+        private static Dictionary<int, InterceptedMethodInvoker> CreateMapForConstructedGeneric(
             Type implementationType,
             IGrainMethodInvoker invoker)
         {
-            // This method creates a a map interfaceId -> intercepted invoker.
             // It is important to note that the interfaceId and methodId are computed based upon the non-concrete
             // version of the implementation type. During code generation, the concrete type would not be available
             // and therefore the generic type definition is used.
             if (!implementationType.IsConstructedGenericType)
             {
                 throw new InvalidOperationException(
-                    $"Type {implementationType} passed to {nameof(GetGenericInterfaceInvoker)} is not a constructed generic type");
+                    $"Type {implementationType} passed to {nameof(CreateMapForConstructedGeneric)} is not a constructed generic type");
             }
 
             var genericClass = implementationType.GetGenericTypeDefinition();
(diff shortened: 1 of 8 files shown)


src/Orleans/Core/InterceptedMethodInvokerCache.cs
```diff
@@ -34,74 +32,142 @@ internal class InterceptedMethodInvokerCache
...
         /// </summary>
-        /// <param name="implementationType">The grain type.</param>
-        /// <param name="interfaceId">The interface id.</param>
-        /// <param name="invoker">The invoker.</param>
-        /// <returns>A new <see cref="InterceptedMethodInvoker"/> for the provided values.</returns>
-        private static InterceptedMethodInvoker CreateInterceptedMethodInvoker(
+        /// <param name="implementationType">The implementation type.</param>
+        /// <param name="invoker">The grain method invoker.</param>
+        /// <returns>The mapped interface.</returns>
+        private static Dictionary<int, InterceptedMethodInvoker> GetInterfaceToImplementationMap(
```
jason-bragg: CreateInterfaceToImplementationMap?

src/Orleans/Core/InterceptedMethodInvokerCache.cs
```diff
@@ -34,74 +32,142 @@ internal class InterceptedMethodInvokerCache
...
             IGrainMethodInvoker invoker)
         {
-            // If a grain extension is being invoked, the implementation map must match the methods on the extension
-            // and not the grain implementation.
-            var extensionMap = invoker as IGrainExtensionMap;
-            if (extensionMap != null)
+            // This method creates a map from interfaceId -> intercepted invoker.
+            if (implementationType.IsConstructedGenericType) return GetGenericInterfaceInvoker(implementationType, invoker);
+            var implementationTypeInfo = implementationType.GetTypeInfo();
+            var interfaces = implementationType.GetInterfaces();
```
jason-bragg: nit: plz avoid var unless type is obvious.

src/Orleans/Core/InterceptedMethodInvokerCache.cs
```diff
@@ -34,74 +32,142 @@ internal class InterceptedMethodInvokerCache
...
         /// </summary>
-        /// <param name="interfaceId">The interface id.</param>
-        /// <param name="implementationType">The implementation type.</param>
+        /// <param name="implementationType">The implementation type, which must be a concrete generic type.</param>
+        /// <param name="invoker">The grain method invoker.</param>
         /// <returns>The mapped interface.</returns>
-        private static Dictionary<int, MethodInfo> GetInterfaceToImplementationMap(
-            int interfaceId,
-            Type implementationType)
+        private static Dictionary<int, InterceptedMethodInvoker> GetGenericInterfaceInvoker(
```
jason-bragg: CreateInterfaceToIntercepterInvokerMap?
from comment "This method creates a map interfaceId -> intercepted invoker"

src/Orleans/Core/InterceptedMethodInvokerCache.cs
```diff
@@ -34,74 +32,142 @@ internal class InterceptedMethodInvokerCache
...
-                .Select(i => implementationType.GetTypeInfo().GetRuntimeInterfaceMap(i))
-                .SelectMany(map => map.InterfaceMethods
-                    .Zip(map.TargetMethods, (interfaceMethod, targetMethod) => new { interfaceMethod, targetMethod }))
-                .ToArray();
-
-            // Map the grain interface methods to the implementation methods.
-            return GrainInterfaceUtils.GetMethods(interfaceTypes[interfaceId])
-                .ToDictionary(GrainInterfaceUtils.ComputeMethodId,
-                    m => interfaceMapping.SingleOrDefault(pair => pair.interfaceMethod == m)?.targetMethod);
+            // This method creates a a map interfaceId -> intercepted invoker.
```
jason-bragg: a a

test/Tester/MethodInterceptionTests.cs
```diff
@@ -25,5 +29,38 @@ public class MethodInterceptionTests : HostedTestClusterEnsureDefaultStarted
             result = await grain.SayHello();
             Assert.Equal("Hello", result);
         }
+
+        [Fact]
+        public async Task GenericGrainMethodInterceptionTest()
+        {
+            var grain = GrainFactory.GetGrain<IGenericMethodInterceptionGrain<int>>(
+                0,
+                typeof(GenericMethodInterceptionGrain<>).Namespace + "." + nameof(GenericMethodInterceptionGrain<int>));
```
jason-bragg: This seems very cryptic and unusable.  How would a user ever know how to form this sort of magic?
Also required user know about concrete type (at least it's name).
Are we confident there is not a good way around requiring this hint?

If we're going to require something like this can we at least add some  IGrainFactory extensions for this, something like:
GetGenericGrain<TInterface,TGrainType>(grainId)

If this is out of scope for this PR, we can address this in a separate PR, or open an issue.

jason-bragg:
Some minor comments and a (probably pre-existing) usability issue, but LGTM

src/Orleans/Core/InterceptedMethodInvokerCache.cs
ReubenBond: Good point, thanks

src/Orleans/Core/InterceptedMethodInvokerCache.cs
ReubenBond: Fixed, thanks

src/Orleans/Core/InterceptedMethodInvokerCache.cs
ReubenBond: thanks

test/Tester/MethodInterceptionTests.cs
ReubenBond: It's a subtlety in how grain types are resolved, but I fixed it by having the tricky grain class not inherit from the generic grain class. If we decide it's a priority, I may go and fix some deficiencies in how type resolution works.

ReubenBond:
Thanks for the review, @jason-bragg! I've addressed most of the comments in the latest revision.

PR #2506, opened 2016-12-08
move tests to DefaultCluster.Tests in main and vNext solution 

xiazen:
-add BVT test category to all functionals in DefaultCluster.Tests
-turn on parallel testing in DefaultCluster.Tests
-clean dependency in NonSiloTests proj, remove dependency on AzureUtils and BondUtils


xiazen:
@dotnet-bot test this please

src/Parallel-Tests.ps1
```diff
@@ -23,7 +23,14 @@ foreach ($a in $assemblies)
 {
     $xmlName = 'xUnit-Results-' + [System.IO.Path]::GetFileNameWithoutExtension($a) + '.xml'
     $outXml = $(Join-Path $outDir $xmlName)
-    $cmdLine = $xunitRunner.FullName + ' ' + $a + ' ' + $testFilter + ' -xml ' + $outXml + ' -parallel none -noshadow -verbose' 
+	if([System.IO.Path]::GetFileNameWithoutExtension($a) -eq 'DefaultCluster.Tests')
```
xiazen: DefaultCluster.Tests  don't have to run in parallel, just it make tests run faster. it do have some gain in time. 

Yes good suggestion.

xiazen: actually it does not have real time gain... isn't it supposed to be faster? am I doing it wrong?

jdom: I believe you could just remove the `-parallel none` switch from here, and instead rely on the <dllname>.xunit.runner.json files to specify by themselves how they parallelize the run for their own test assembly.

test/DefaultCluster.Tests/TimeoutTests.cs
```diff
@@ -9,7 +9,7 @@
 using Xunit;
 using Xunit.Abstractions;
 
-namespace UnitTests
+namespace DefaultCluster.Tests
 {
     // if we parallelize tests, this should run in isolation
```
jdom: look at this comment. This is a test that should run in isolation and cannot be parallelized, even if it runs on the default cluster. It should not be moved.

src/OrleansCodeGenerator/GrainReferenceGenerator.cs
```diff
@@ -61,12 +61,13 @@ internal static TypeDeclarationSyntax GenerateClass(Type grainType, Action<Type>
...
-#if !NETSTANDARD
-                    // we could add Serializable attribute if we want, but that would require that the target
-                    // project depends on System.Runtime.Serialization.Formatters, which is currently in preview
+
                     SF.Attribute(typeof(SerializableAttribute).GetNameSyntax()),
+#if !NETSTANDARD_TODO
+                    //ExcludeFromCodeCoverageAttribute became an internal class in netstandard
                     SF.Attribute(typeof(ExcludeFromCodeCoverageAttribute).GetNameSyntax()),
 #endif
+
```
jdom: why is this change bundled with the move?

test/DefaultCluster.Tests/BasicActivationTests.cs
```diff
@@ -11,11 +11,12 @@
 
 #pragma warning disable 618
 
-namespace UnitTests.General
+namespace DefaultCluster.Tests.General
 {
+    [TestCategory("BVT"), TestCategory("Functional")]
```
jdom: were these tests converted to BVTs on purpose? Because they weren't before.
BTW, we should not mark the whole class as BVT or Functional, and have each test be marked with it (as there might be a class that have a few BVTs and some that arent't, and it would be easy to miss when adding new test methods). Nevertheless,  big categories such as "ActivateDeactivate" is fine to add them at class level.

test/DefaultCluster.Tests/ClientAddressableTests.cs
```diff
@@ -7,8 +7,9 @@
 using UnitTests.GrainInterfaces;
 using Xunit;
 
-namespace UnitTests
+namespace DefaultCluster.Tests
 {
+    [TestCategory("BVT")]
```
jdom: Same as before, please don't add this at class level. And was this added by design? because they weren't BVTs before.

jdom: There are a ton of similar instances. I won't mention them all, but please review each of them

src/Parallel-Tests.ps1
xiazen: yeah make sense, I already turned the parallelizeTestCollections on in .xunit.runner.json. maybe that's why it doesn't have any time again with -parallel collection param. 

will revert this file

test/DefaultCluster.Tests/TimeoutTests.cs
xiazen: will move back

src/OrleansCodeGenerator/GrainReferenceGenerator.cs
jdom: I see, cool, was confused by the added TODO  conditional compilation

xiazen: cool? do you are going to push? ;) don't forget my PR!

jdom: I just stopped by since you assigned to me,  but since I can't download locally or even run/check functionals, I'll leave it to someone else on the team. The PR looks good to me as far as I can tell from github

attilah:
Based on the previous comments LGTM.

test/DefaultCluster.Tests/project.json
```diff
@@ -0,0 +1,15 @@
+﻿{
```
ReubenBond: This file shouldn't be needed any more.

vNext/test/TestInternalDtosRefOrleans/TestInternalDtosRefOrleans.csproj
```diff
@@ -0,0 +1,55 @@
...
+    <RootNamespace>UnitTests.DtosRefOrleans</RootNamespace>
+    <AssemblyName>TestInternalDtosRefOrleans</AssemblyName>
+  </PropertyGroup>
+  <PropertyGroup>
+    <TargetFramework>net462</TargetFramework>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <RuntimeIdentifier>win</RuntimeIdentifier>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="..\..\..\test\TestInternalDtosRefOrleans\**\*.cs" Exclude="      ..\..\..\test\TestInternalDtosRefOrleans\obj\**\*.cs;       ..\..\..\test\TestInternalDtosRefOrleans\bin\**\*.cs;       ..\..\..\test\TestInternalDtosRefOrleans\Properties\*.cs" />
```
ReubenBond: The Excludes here lose readability because of the auto-formatting :( maybe the whitespace should just be removed altogether

ReubenBond:
LGTM with minor changes. I'm not a huge fan of the name of the DefaultCluster.Tests project, but it's not a big deal.

I'd consider calling it Orleans.SystemTests or Orleans.FunctionalTests (whereas the other projects should be Orleans.UnitTests or something). That way, it's obvious what belongs in it. Projects which tests specific components like Azure support should be named accordingly, usually <project being tested>.UnitTests or .SystemTests.

Again, though, LGTM wth minor changes

xiazen:
but this test is a subset of functional tests. It's a cluster of tests which uses default cluster option. 
But yeah we need better naming convention. agreed!

test/DefaultCluster.Tests/project.json
xiazen: I believe this is in main repo

vNext/test/TestInternalGrains/TestInternalGrains.csproj
```diff
@@ -17,6 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Compile Include="..\..\..\test\TestInternalGrains\**\*.cs" Exclude="      ..\..\..\test\TestInternalGrains\obj\**\*.cs;       ..\..\..\test\TestInternalGrains\bin\**\*.cs;       ..\..\..\test\TestInternalGrains\Properties\*.cs" />
```
xiazen: I will fix this in next PR

PR #2508, opened 2016-12-12
Reduce usage of statics in IRuntimeClient implementation & usage

ReubenBond:
More of the same. I'm very wary of making any of these PRs large, so this isn't a huge step forward.

Generally speaking, this moves additional members from `IRuntimeClient` to `ISiloRuntimeClient` and entirely removes some members.

This also removes some of the usages of `RuntimeClient.Current`, which is one of the main goals of this series of PRs.

* Introduces an interface, `IGrainTimer` from the class `GrainTimer` in order to cleanly separate existing `IActivationData` interface from its only implementation, `ActivationData` (we could remove `IActivationData` if we don't think it's useful and use `ActivationData` instead)
* `IInvokable` was moved from `Orleans` to `OrleansRuntime` and `GrainTypeManager` was added as a parameter. I felt adding a parameter here was better than requiring that the implementations (`SystemTarget` and `ActivationData`) add an extra field.
* `GrainRuntime` moved to from `Orleans` to `OrleansRuntime`. It's a trivial implementation of `IGrainRuntime` and moving it there lets us implement it using internal interfaces without losing anything.

### DIFF
--- a/src/OrleansRuntime/Core/GrainTimer.cs
+++ b/src/OrleansRuntime/Core/GrainTimer.cs
@@ -6,7 +6,7 @@ using System.Threading.Tasks;
 
 namespace Orleans.Runtime
 {
-    internal class GrainTimer : IDisposable
+    internal class GrainTimer : IGrainTimer
     {
         private Func<object, Task> asyncCallback;
         private AsyncTaskSafeTimer timer;
@@ -16,17 +16,17 @@ namespace Orleans.Runtime
         private int totalNumTicks;
         private static readonly Logger logger = LogManager.GetLogger("GrainTimer", LoggerType.Runtime);
         private Task currentlyExecutingTickTask;
-        private readonly ActivationData activationData;
+        private readonly IActivationData activationData;
 
-        internal string Name { get; private set; }
+        public string Name { get; }
         
         private bool TimerAlreadyStopped { get { return timer == null || asyncCallback == null; } }
 
-        private GrainTimer(Func<object, Task> asyncCallback, object state, TimeSpan dueTime, TimeSpan period, string name)
+        private GrainTimer(IActivationData activationData, Func<object, Task> asyncCallback, object state, TimeSpan dueTime, TimeSpan period, string name)
         {
             var ctxt = RuntimeContext.CurrentActivationContext;
             InsideRuntimeClient.Current.Scheduler.CheckSchedulingContextValidity(ctxt);
-            activationData = (ActivationData) RuntimeClient.Current.CurrentActivationData;
+            this.activationData = activationData;
 
             this.Name = name;
             this.asyncCallback = asyncCallback;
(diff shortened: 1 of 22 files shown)


src/OrleansRuntime/Core/IInvokable.cs
```diff
@@ -0,0 +1,12 @@
+using Orleans.CodeGeneration;
+
+namespace Orleans.Runtime
+{
+    /// <summary>
+    /// Common internal interface for SystemTarget and ActivationData.
+    /// </summary>
+    internal interface IInvokable
+    {
+        IGrainMethodInvoker GetInvoker(GrainTypeManager typeManager, int interfaceId, string genericGrainType = null);
```
jdom: I wonder if it makes sense to pass another `IInvokable` instead of `GrainTypeManager` and make it implement IInvokable. This way you can build a strategy chain instead of forcing a single decorated object of type GrainTypeManager. Just a naive suggestion, and don't know if it even makes sense logically.

ReubenBond: @jdom I thought about that, but figured it would be too confusing to people reading the code. At the bottom of the stack, would you pass null? i.e, would `GrainTypeManager` ignore the `IInvokable` parameter?

src/OrleansRuntime/Core/GrainTimer.cs
```diff
@@ -40,33 +40,35 @@ private GrainTimer(Func<object, Task> asyncCallback, object state, TimeSpan dueT
         }
 
         internal static GrainTimer FromTimerCallback(
-                TimerCallback callback,
-                object state,
-                TimeSpan dueTime,
-                TimeSpan period,
-                string name = null)
+            TimerCallback callback,
```
attilah: argument name mismatch with the method below, what about "timerCallback" ? or something which tells "more" about what is there?

src/OrleansRuntime/ReminderService/LocalReminderService.cs
```diff
@@ -31,7 +31,7 @@ private enum ReminderServiceStatus
         private IRingRange myRange;
         private long localTableSequence;
         private int rangeSerialNumber;
-        private GrainTimer listRefresher; // timer that refreshes our list of reminders to reflect global reminder table
+        private IGrainTimer listRefresher; // timer that refreshes our list of reminders to reflect global reminder table
```
attilah: "timer" suffix for consistent naming?

src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
```diff
@@ -74,20 +80,20 @@ private int RemoveDeadProducers()
         }
 
         /// accept and notify only Active producers.
-        private static bool IsActiveProducer(IStreamProducerExtension producer)
+        private bool IsActiveProducer(IStreamProducerExtension producer)
         {
             var grainRef = producer as GrainReference;
             if (grainRef !=null && grainRef.GrainId.IsSystemTarget && grainRef.IsInitializedSystemTarget)
-                return RuntimeClient.Current.GetSiloStatus(grainRef.SystemTargetSilo) == SiloStatus.Active;
+                return this.siloStatusOracle.GetApproximateSiloStatus(grainRef.SystemTargetSilo) == SiloStatus.Active;
```
attilah: this keyword usage.

src/OrleansRuntime/ReminderService/LocalReminderService.cs
ReubenBond: I didn't change the name of this, but I'll change it if you prefer.

ReubenBond:
Please squash & merge if you're satisfied

galvesribeiro:
Great work!

Issue #469, opened 2015-06-02
Deprecate Code-generated concrete factories


gabikliot:
We currently have 2 ways to use grain factories:
Code-generated concrete factories (`MyGrainFactory.GetFrain()`) and generic factories (`GrainFactory.GetGrain<IMyGrain>()`). The goal is to completely deprecate and eventually eliminate the code-generated concrete factories. Reasons:
1) 2 ways for application to get grain reference - confusing.
2) Code-generated concrete factories are not testable. Generic factories are (#443).
3) Reduce code duplication - the runtime implementation of the 2 different factories has some code duplication.
4) Overall reduce the amount of generated code. Getting rid of code-generated factories is one step in that direction.
5) Generic factories allow to have Grain interface and implementation in same project, while [Code-generated factories do not](https://orleans.codeplex.com/discussions/577719).

There is a number of steps required to make this possible:
- [x] Port all CI tests to use generic factories. Done via #468.
- [x] Use generic factories for system target references inside the runtime. Done via #470.
- [x] Eliminate all usage of reflection calls to concrete factories (there are still some in streaming code). Done via #527.
- [x] Solve the bug of generic factories used inside generic grain (#423). Done via #479.
- [x] Port all our internal tests to use generic factories
- [x] Deprecate code-generated factories.

nehmebilal:
Great!

Also add to the list that code-generated factories don't get renamed during code refactoring. If my GrainA is using GrainB as follows:

```
GrainBFactory.GetGrain("grain_b_id").DoSomething();
```
and I rename GrainB to GrainC using visual studio refactoring, the code will break. It wouldn't if the generic factory was used :)

gabikliot:
This issue is now done! `GetGrain` via code-generated concrete factories was deprecated in #540.

PR #2510, opened 2016-12-13
Inline & remove TypeCodeMapper

ReubenBond:
Minor PR in working towards #469. `TypeCodeMapper` was using `RuntimeClient.Current`, but its usages are sparse and can be inlined for (I believe) greater clarity.


ReubenBond:
Functionals pass (well, no new failures)

PR #2513, opened 2016-12-13
Send all statistics to telemetry consumers

jason-bragg:
Only a hard coded set of statistics were going to telemetry consumers.  Now all non-string statistics are tracked.


jason-bragg:
tagging @galvesribeiro, as he's more familiar with these systems than I.

src/Orleans/Statistics/CounterStatistic.cs
```diff
@@ -33,6 +33,16 @@ public interface ICounter
...
         string GetDisplayString();
         CounterStorage Storage { get; }
+        void TrackMetric(Logger logger);
+    }
+
+    public static class Metric
+    {
+        public static string CreateName(ICounter counter)
+        {
+            return counter.Name + "." + (counter.IsValueDelta ? "Delta" : "Current");
```
galvesribeiro: I wonder if we could cache those names on first use...

src/Orleans/Statistics/CounterStatistic.cs
```diff
@@ -141,7 +151,7 @@ private static CounterStatistic FindOrCreate_Impl(StatisticName name, bool useDe
             }
         }
 
-        static public bool Delete(string name)
+        public static bool Delete(string name)
```
xiazen: the same question here. Just a little confused about why you turn a lot of private method into public. A little explanation will help

galvesribeiro:
For some reason my review was deleted by GitHub o.O

Anyway, the only comment I have is regarding the `Metric.Create()`... I was saying that I wonder if we couldn't cache the metric name instead of concat strings each time it is tracked....

Besides that, LGTM, I should did that before. 

Thanks for doing it @jason-bragg 👍

xiazen:
@galvesribeiro I think I deleted them by accident. I was deleting my false comments...

but anyway, LGTM. I will let @jason-bragg  to decide whether to fix the Metric.Create() or not

galvesribeiro:
@jason-bragg thanks for the latest changes.

Regarding the delta, I always think that deltas and other calculations should be managed by the ones consuming/processing the metric.

LGTM :shipit: !

xiazen:
@galvesribeiro I second you opinion on delta, Orleans should just be responsible for outputting raw data. All the post processing should be managed by the ones consume the data.

galvesribeiro:
@xiazen yup! That was the major principle of bring those APIs for Telemetry. Lets hope at some point everything telemetry-related is flowing that way.

PR #2514, opened 2016-12-14
Fix invocation interception for grain extensions

ReubenBond:
I broke the `PreInvocationCallbackWithStreamTest` test with #2502.
The test checks that `IGrainExtension`s can have their methods intercepted by grains. Previously we were passing a `null` `MethodInfo` parameter to the interceptor, since the correct implementation method could not be found.

#2502 broke that by throwing an exception in the event that an implementation method cannot be found.

This PR fixes it again by mapping the interface method to the extension method in the event that the incoming request is destined for the extension object.

As you can see, it's a very simple fix. Functionals are running.

### DIFF
--- a/src/OrleansRuntime/Core/InsideRuntimeClient.cs
+++ b/src/OrleansRuntime/Core/InsideRuntimeClient.cs
@@ -410,11 +410,25 @@ namespace Orleans.Runtime
                 return invoker.Invoke(target, request);
             }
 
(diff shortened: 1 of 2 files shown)


ReubenBond:
@dotnet-bot please test this ❤️

ReubenBond:
@dotnet-bot please test this

ReubenBond:
@dotnet-bot test this please

ReubenBond:
Tested, working

PR #2515, opened 2016-12-14
Remove usages of Silo.CurrentSilo from GrainDirectoryPartition

ReubenBond:
The priority is to remove `RuntimeClient.Current` since that's more critical for #467, but this PR mostly removes some of the usages of `Silo.CurrentSilo` from `Dispatcher` and `GrainDirectoryPartition` as well as a couple of `RuntimeClient.Current` usages.

* `MembershipFactory` becomes `MembershipTableFactory`, since `IMembershipOracle` is constructed by DI now.
* `Silo` initialization was cleaned up somewhat since `MembershipTableFactory` is now injected into `MembershipOracle` and the `IMembershipTable` implementation is now initialized upon its creation, in `IMembershipOracle.Start()`.
* `ManagementGrain` now uses `IMembershipOracle` instead of `IMembershipTable` when getting silo statuses, but still requires `IMembershipTable` for reading membership table data (of course). This means that it is more applicable in circumstances where no `IMembershipTable` exists, eg under the (upcoming) deep Service Fabric integration.
* @sebastianburckhardt Multi-cluster initialization was altered to make it more friendly to injection and to clean up `Silo` initialization. This means that it is responsible for initializing channels from configuration itself during `Start()`, like so:
```C#
gossipChannels = await this.channelFactory.CreateGossipChannels();
```
* `MultiClusterOracleFactory` was renamed to `MultiClusterGossipChannelFactory` to reflect its new purpose.

@sebastianburckhardt I see failures in the `ThreeClusterBattery` and `FourClusterBattery` tests both in master and on my branch. Is there something I need to do to test those? Other GeoCluster tests are passing


src/OrleansRuntime/MembershipService/MembershipTableFactory.cs
```diff
@@ -0,0 +1,115 @@
...
+            this.serviceProvider = serviceProvider;
+            logger = LogManager.GetLogger(nameof(MembershipTableFactory), LoggerType.Runtime);
+        }
+
+        internal async Task<IMembershipTable> GetMembershipTable()
+        {
+            if (membershipTable != null) return membershipTable;
+            using (this.initializationLock.LockAsync())
+            {
+                if (membershipTable != null) return membershipTable;
```
ReubenBond: Double-checked locking is used to avoid unnecessary work here, so that this method can be called cheaply

src/OrleansRuntime/MembershipService/MembershipTableFactory.cs
```diff
@@ -0,0 +1,115 @@
...
+                    result = AssemblyLoader.LoadAndCreateInstance<IMembershipTable>(
+                        globalConfig.MembershipTableAssembly,
+                        this.logger);
+                }
+                else
+                {
+                    throw new NotImplementedException("No membership table provider found for LivenessType=" + livenessType);
+                }
+
+                await WaitForTableToInit(result);
```
ReubenBond: The table is now initialized before it's returned, whereas that was previously the responsibility of Silo. Now it is called from the `MembershipOracle.Start()` method.

src/OrleansRuntime/MultiClusterNetwork/MultiClusterGossipChannelFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+        private readonly GlobalConfiguration globalConfig;
+        private readonly Logger logger;
+
+        public MultiClusterGossipChannelFactory(GlobalConfiguration globalConfig)
+        {
+            this.globalConfig = globalConfig;
+            logger = LogManager.GetLogger("MultiClusterGossipChannelFactory", LoggerType.Runtime);
+        }
+
+        internal async Task<List<IGossipChannel>> CreateGossipChannels()
```
ReubenBond: This method hasn't meaningfully changed. The only difference is that it no longer takes the `Silo` parameter, since the configuration is injected instead.

src/OrleansRuntime/MembershipService/MembershipTableFactory.cs
sebastianburckhardt: Why do we need a lock here?  I though initialization is called only once and during a very specific time (MembershipOracle.Start).

ReubenBond: It's also called from `ManagementGrain`, but I agree that the lock should have zero contention. It was done for correctness even though the current usage pattern doesn't require it. I can change it if you wish

src/OrleansRuntime/MultiClusterNetwork/MultiClusterGossipChannelFactory.cs
sebastianburckhardt: makes sense to me.

sebastianburckhardt:
@ReubenBond, don't worry about the failing `ThreeClusterBattery`  and  `FourClusterBattery`. There are some known bugs in there that I haven't gotten around to fixing yet.

ReubenBond:
@dotnet-bot please test this

ReubenBond:
@dotnet-bot test this please

ReubenBond:
Thanks, @sebastianburckhardt :)

Fixed an issue in `ManagementGrain`: some tests were failing in CI (passing locally) which I believe was due to a race condition whereby `ISiloStatusOracle.GetApproximateSiloStatuses()` had not yet observed the most up-to-date set of silos. Fixed by explicitly calling `IMembershipGrain.GetHosts()` (as we do in master) when the `ISiloStatusOracle` is `MembershipOracle` (and therefore uses an `IMembershipTable` for membership info.

src/OrleansRuntime/Core/ManagementGrain.cs
```diff
@@ -27,13 +46,17 @@ public override Task OnActivateAsync()
...
-            var mTable = await GetMembershipTable();
-            var table = await mTable.ReadAll();
-            
-            var t = onlyActive ? 
-                table.Members.Where(item => item.Item1.Status == SiloStatus.Active).ToDictionary(item => item.Item1.SiloAddress, item => item.Item1.Status) :
-                table.Members.ToDictionary(item => item.Item1.SiloAddress, item => item.Item1.Status);
-            return t;
+            // If the status oracle isn't MembershipOracle, then it is assumed that it does not use IMembershipTable.
+            // In that event, return the approximate silo statuses from the status oracle.
+            if (!(this.siloStatusOracle is MembershipOracle)) return this.siloStatusOracle.GetApproximateSiloStatuses(onlyActive);
```
jdom: Shouldn't the branching logic (or actually not really branching, but the right strategy in the concrete class) be in the oracle itself?

sergeybykov:
@ReubenBond, please rebase and resolve the conflict.

ReubenBond:
@sergeybykov rebased and added another (minor) commit. I'll run functionals once more, too

PR #2520, opened 2016-12-14
Add doc page on heterogeneous silos

benjaminpetit:
I added a small documentation page on heterogeneous silos. I also regenerate the documentation with docfx, it seems that it was not done in the previous commits


src/Documentation/Advanced-Concepts/Heterogeneous-Silos.md
```diff
@@ -0,0 +1,39 @@
+---
+layout: page
+title: Heterogeneous silos
+---
+
+# Heterogeneous silos
+On a given cluster, silos can support a different set of grain types:
```
ReubenBond: Is Heterogeneous.png used? Could we also checkin the src file for it (assuming it was generated from some other file)?

benjaminpetit: Yes the file is used in the Heterogenous page?

I can add the visio file if you want, but it is not a very complicated diagram

benjaminpetit: In the first commit there is the png in the src/ dir, and in the second one it is the 'html'version (check the file list, it will not appear in GitHub diff since).

Anyway I added the visio file also

ReubenBond: I can see it's included manually in the HTML, but I don't see the image included *here* in the source. Am I mistaken?

benjaminpetit: It doesn't appear in the raw view, but when you click on "Change the file using the online editor" you can see it. Bizarre.

PR #2521, opened 2016-12-15
Added Lebara to "Who Is Using Orleans"

DixonDs:
Since it is used in production at Lebara for some time now


gabikliot:
Thank you.

PR #2525, opened 2016-12-16
Fix type assertion in AdaptiveDirectoryCacheMaintainer

ReubenBond:
Fixes an error in `AdaptiveDirectoryCacheMaintainer` which was performing a type check which would never succeed. `TValue` in the below code is always `IReadOnlyList<Tuple<SiloAddress, ActivationId>>`, but the type check assumed it was `List<Tuple<SiloAddress, ActivationId>>`.

See the `//TODO:` below for the source of the error.

This removes the check and pushes the responsibility onto the caller.

### DIFF
--- a/src/OrleansRuntime/GrainDirectory/AdaptiveDirectoryCacheMaintainer.cs
+++ b/src/OrleansRuntime/GrainDirectory/AdaptiveDirectoryCacheMaintainer.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
 using System.Threading;
 using Orleans.Runtime.Scheduler;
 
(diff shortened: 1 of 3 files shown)


src/OrleansRuntime/GrainDirectory/AdaptiveDirectoryCacheMaintainer.cs
```diff
@@ -14,14 +13,19 @@ internal class AdaptiveDirectoryCacheMaintainer<TValue> : AsynchAgent
 
         private readonly AdaptiveGrainDirectoryCache<TValue> cache;
         private readonly LocalGrainDirectory router;
+        private readonly Func<List<ActivationAddress>, TValue> updateFunc;
```
jason-bragg: Update function is not a great name, but can't think of a better one.
This pattern, over all, is awful, but not sure how to fix it, atm. :/
No change requested..  just making noises as bits of my soul die.

jason-bragg:
LGTM

src/OrleansRuntime/GrainDirectory/AdaptiveDirectoryCacheMaintainer.cs
ReubenBond: By the way, I don't think the pattern here is *that* bad.

jason-bragg: As for the name, it seems to be a translation function.  It translates the list of activation addresses to whatever the cache is storing.  Since there are no assumptions may about the cached data there is no general way to translate, hence the need for a function.  While this is not horrible, it definitely smells.  Maybe using some sort of functor, like dictionaries use comparers, provided as a generic might be a cleaner approach.  AdaptiveDirectoryCacheMaintainer<TValue,TTranslator>?  In any case, this pattern at least deserves a look at for refactoring, but I understand that such is beyond the scope of this bugfix.

Issue #2427, opened 2016-11-21
Embed an HTTP endpoint into silos


sergeybykov:
Adding an embedded HTTP endpoint similar to https://github.com/OrleansContrib/OrleansHttp to silo code I think would bring several benefits.

* Simplify learning Orleans, as a separate frontend layer won't be required for invoking a grain from a browser or another app.
* Simplify local development by not requiring a `GrainClient` for testing simple enough scenarios.
* Simplify access to `ManagementGrain` by not requiring a client app like `OrleansManager`. This could be leveraged for managing or troubleshooting production deployments by remoting into the cluster or accessing the internal HTTP endpoint some other way.
* The endpoint could be exposed to the internet via a TLS-secured façade.
* For certain applications exposing the built-in HTTP endpoint via a load balancer may be sufficient without a separate frontend layer.

Looking at `OrleansHttp`, my impression is it could be used almost as-is if we remove the need for a bootstrap provider, and instead initialize the endpoint as part of the silo startup sequence.

galvesribeiro:
Although I'm still look for a future where the client protocol would be documented and open, this is definitively a very good starting point to remove the needless external client where it is just a HTTP endpoint forwarding requests to grains.

The endpoint controllers could be easily generated by codegen and spin up on silo startup as you mentioned.

For scenarios where streams or observers are required, some websocket (maybe SignalR) endpoint can be exposed as well.

creyke:
+1

sebastianburckhardt:
I agree that not having to go through a client could be advantage, also for performance (particularly _response latency_).

If the client is on the silo, it saves one hop. Also, silos are better equipped than clients: they have a grain directory cache, and they have a sophisticated scheduler.

When we did performance experiments for the planned reactivity support (project Bruges with @ticup) the response latency inside grains was much better than inside clients.

richorama:
I think this is an exciting possibility.

Some thoughts:

* __URI scheme.__ It's important to get the URI scheme right. I don't think OrleansHttp quite got it right with regard to compound keys. You should also be able to put parameters in the body of an HTTP POST to enable larger variable values.
* __Performance.__ I think performance is a second order problem, I'm fairly certain that the clever people in the community will find a way for it to go fast. (perhaps using libuv???).
* __Encodings.__ OrleansHttp uses JSON, but additional encodings (XML?) could be added later by observing the `Content-Accept` request header.
[...]

PR #2526, opened 2016-12-16
added http api from orleans contrib

richorama:
Added an HTTP API to the silo, to allow grains to be called on any HTTP verb.

The uri scheme is as follows: 

```
http://localhost:8080/grain/{FULL_NAME_FOR_GRAIN_INTERFACE}/{PRIMARY_KEY}/{METHOD}?ARGNAME=ARGVALUE
```

For example:

```
http://localhost:8080/grain/UnitTests.GrainInterfaces.ISimpleGrain/0/SetA?a=42
```

Some notes:

1. The endpoint is disabled by default.
1. It defaults to port 8080, be can be configured to listen on any port.
1. A username and password can be configured to allow basic auth.
1. Generic grains are currently not support.
1. Compound keys are currently not supported.
1. Method overloading is not supported.
1. Classname prefixes are supported by appending to the grain interface type with a colon : `MyNamespace.IMyGrain:ClassPrefix`
1. Responses are always JSON encoded.
1. Argument values are JSON encoded (`"hello world"` is valid JSON :¬)
1. Reflection is used to make the grain call. This is all cached, but I'm sure @ReubenBond can figure out a better way of implementing this.
1. The API is implemented directly on OWIN, using [Nowin](https://github.com/Bobris/Nowin) as the host. Attempts were made to use WebAPI, but these failed.
1. The HTTP plumbing (i.e the `Router` class) is designed to be (a bit) generic, allowing other services to use it, and register additional routes in the future, such as streaming, swagger, etc...  That's why I went with `grain` at the start of the path.


richorama:
related to #2427

richorama:
Is the build failure related to this PR? The failed tests seem unrelated.

sergeybykov:
@dotnet-bot Test this please

richorama:
I think I've broken 'FastKill' will fix.

src/OrleansRuntime/Http/ExtensionMethods.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Microsoft.Owin;
+using Newtonsoft.Json;
+using System;
+using System.Threading.Tasks;
+
+namespace Orleans.Runtime
+{
+    internal static class ExtensionMethods
+    {
+        public static Task ReturnJson(this IOwinContext context, object value)
```
ReubenBond: I think everything HTTP related should be in a separate library if possible - that's not to say that changes to the core won't be needed, just that they should be general-purpose changes, rather than HTTP-specific.

richorama: I'm not opposed to moving the HTTP pieces to a separate assembly, I was following the advice here: https://github.com/dotnet/orleans/issues/2427#issuecomment-263644610

galvesribeiro:
Yeah, I agree that we should have just the abstractions in it, like @attilah did with the DI, and at config time you inject the proper API implementation. If we were to stick with something as a direct dependency, I would try have ASP.Net Core (even on full framework) WebAPI on it...

richorama:
@attilah see branch here for attempt with ASP.NET WebAPI on Orleans Http: https://github.com/OrleansContrib/OrleansHttp/tree/asp.net

It didn't seem to support async actions on the controller (i.e. the resulting JSON was an incomplete task rather than the result value).

tsibelman:
Passing arguments in query string has limitations on size of the parameters, I suggest using POST request with parameters inside the body

sergeybykov:
@attilah is currently looking to see if we can refactor this to use Kestrel or WebAPI.

tsibelman:
Nice, we using HttpListener right now but we also want to move over Kestler. We using Orleans to create a microservice architecture where each service has it own cluster of Orleans silos. In our solution developers of micro services asked to define subset of grain interfaces using special attribute, this way developers are encouraged to think about what they want to expose as a public API of the service instead of exposing all of their grains.

galvesribeiro:
@tsibelman I'm doing the same here. However, we still doing the client as an external webAPI thing... Would be good to get this work done.

tsibelman:
@galvesribeiro what do you mean by external webApi ? External to orleans silo like on separate machine ?

galvesribeiro:
External API as in an external separate machine.

richorama:
@sergeybykov @attilah Thanks. I attempted to get it to work with self hosted WebAPI (.NET 4.6) but it didn't seem to support async. This build is failing because of lack of .NET Core support, and I'm not sure what to do. (there also seems to be a preference to keep this in contrib).

Help is appreciated.

sergeybykov:
>there also seems to be a preference to keep this in contrib

I'm of the opposite opinion - I think we need a built-in default easy to get going HTTP endpoint in silos. Similar to Node. The need to run clients as a separate tier, I'm afraid, is a barrier to adoption, especially for simple in-house, POC like scenarios.

tsibelman:
@sergeybykov  You maybe interested to know that our company is now starting the process of open sourcing our Infrastructure, particularly our custom Orleans host that, allows exposing selected grains via HTTP protocol.

sergeybykov:
@tsibelman This is great! Please share the link when it's published. @attilah is looking at the options at the moment, and learning from what you guys implemented will definitely help.

tsibelman:
@sergeybykov it took us some time but we finally published our framework https://github.com/gigya/microdot , we have step by step guide in the wiki and sample solution in separate repository https://github.com/gigya/microdot-samples

sergeybykov:
@tsibelman Fantastic! We need to add links to the docs.

tsibelman:
@sergeybykov great to hear

sergeybykov:
We are considering taking a slightly different approach. We'll use this PR as a reference, but implement it without the use of Reflection. Hence, I'm closing this PR.

Thank you, @richorama!

snarlynarwhal:
What's the status regarding an embedded HTTP API? Has this been implemented yet? Or should I use [OrleansHTTP](https://github.com/OrleansContrib/OrleansHttp)?

sergeybykov:
>What's the status regarding an embedded HTTP API?

We still hope to include it in 2.0.0. But it's at the bottom of the priority list.

snarlynarwhal:
Okay thanks for the update!

PR #2531, opened 2016-12-19
Rebased version of #2459: Grain Services by @jamescarter-le

sergeybykov:
This is a rebased version of #2459 to make it mergeable to latest master.

PR #2534, opened 2016-12-20
Removed unused ClientType from GrainServiceConfiguration

sergeybykov:
A follow-up to #2459 and #2531. Removed unused ClientType property from GrainServiceConfiguration.

@jamescarter-le, any objections?


ReubenBond:
We should also remove serviceName and replace its usage with ServiceType. It might otherwise give the impression that a consumer can have multiple services with identical types but different names

jamescarter-le:
The intention was someone should be able to have multiple services with
different names of the same type.

I was not able to finish the implementation in time. I had intended to set
up the ServiceProvider to register the IGrainServiceClient but at the point
the Grain Services are constructed there is no access to register types.

Happy to remove it for now and I can modify it later on, or skip that if
it's not useful.

Issue #2533, opened 2016-12-20
AsyncExecutorWithRetries consumes all memory


rrector:
If AsyncExecutorWithRetries is used with an infinite number of retries, and every retry fails, the asynchronous tail recursion in ExecuteWithRetriesHelper will cause an indefinitely long chain of Tasks/continuations to be created until all memory is consumed.

If infinite retries is a requirement, ExecuteWithRetriesHelper should probably be modified to use a loop instead of recursion.

sergeybykov:
@rrector, sounds like you have a fix in mind. A PR?

ReubenBond:
Fixed via #2535, thanks @rrector :)

PR #2535, opened 2016-12-20
Make AsyncExecutorWithRetries non-recursive

rrector:
The recursion in AsyncExecutorWithRetries is converted by the generated async state machine to a task/continuation. Each level of the recursion generates a new continuation on the current Task to the next retry iteration. If infinite retries are specified, and the operation fails on each try, then new tasks/continuations are created indefinitely. They as well as all intermediate state (e.g. exceptions attached to previous tasks) are rooted and evade GC until either one of the retries succeeds or all virtual memory is consumed. #2533

This change does the retries in a loop instead of using recursion, allowing intermediate tasks to examined and released for GC.

### DIFF
--- a/src/Orleans/Async/AsyncExecutorWithRetries.cs
+++ b/src/Orleans/Async/AsyncExecutorWithRetries.cs
@@ -4,8 +4,6 @@ using Orleans.Runtime;
 
 namespace Orleans
 {
-    using System.Runtime.ExceptionServices;
-
     /// <summary>
     /// This class a convinent utiliity class to execute a certain asyncronous function with retires, 
     /// allowing to specify custom retry filters and policies.
@@ -13,7 +11,6 @@ namespace Orleans
     internal static class AsyncExecutorWithRetries
     {
         public static readonly int INFINITE_RETRIES = -1;
-        private static readonly Func<Exception, int, bool> retryAllExceptionsFilter = (Exception exc, int i) => true;
 
         /// <summary>
         /// Execute a given function a number of times, based on retry configuration parameters.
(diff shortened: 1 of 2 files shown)


src/Orleans/Async/AsyncExecutorWithRetries.cs
```diff
@@ -114,75 +111,71 @@ internal static class AsyncExecutorWithRetries
...
+                            retry = retryExceptionFilter(exc, counter);
                     }
-                    else
+
+                    if (!retry)
                     {
-                        TimeSpan delay = onErrorBackOff.Next(counter);
-                        await Task.Delay(delay);
-                        return await ExecuteWithRetriesHelper(function, callCounter, maxNumSuccessTries, maxNumErrorTries, maxExecutionTime, startExecutionTime, retryValueFilter, retryExceptionFilter, onSuccessBackOff, onErrorBackOff);
+                        throw;
```
ReubenBond: This changed from `ExceptionDispatchInfo.Capture(exception).Throw()`

src/Orleans/Async/AsyncExecutorWithRetries.cs
```diff
@@ -114,75 +111,71 @@ internal static class AsyncExecutorWithRetries
...
+
+                    if (callCounter < maxNumSuccessTries || maxNumSuccessTries == INFINITE_RETRIES) // -1 for infinite retries
                     {
-                        return await ExecuteWithRetriesHelper(function, callCounter, maxNumSuccessTries, maxNumErrorTries, maxExecutionTime, startExecutionTime, retryValueFilter, retryExceptionFilter, onSuccessBackOff, onErrorBackOff);
+                        if (retryValueFilter != null)
+                            retry = retryValueFilter(result, counter);
                     }
-                    else
+
+                    if (retry)
```
ReubenBond: This block of code could be moved to just before the `while`, allowing you to eliminate the similar block inside the `catch`.

rrector: Not really - this uses the onSuccessBackoff func, while the catch uses the onErrorBackoff func

ReubenBond: Ah, I see - good point :)

ReubenBond:
LGTM, thank you!

ReubenBond:
I'll run a full functional pass first to make sure, but otherwise I'm happy to merge.

ReubenBond:
Thank you, @rrector!

Issue #2429, opened 2016-11-21
Sql Server RegisterOrUpdateReminder deadlock


mtdaniels:
I'm seeing sql deadlocks occasionally when RegisterOrUpdateReminder is called.

We use reminders somewhat heavily to make grain work durable (if, for some reason, the grain is killed while performing work, the reminder will reactivate the grain, and work will continue).

I believe the upsert is creating the deadlock because the insert upgrades the lock scope from a single row to the key range.  So, two transactions are able to get row locks that are in the same range, but when they try to insert (because neither currently exist), they deadlock.

I saw this: https://github.com/dotnet/orleans/pull/2395 - but it doesn't look like that modified the reminders implementation.

I have worked around the deadlock issue by adding retries, but this is obviously just masking the underlying problem.

sergeybykov:
@mtdaniels Did you mean to point to https://github.com/dotnet/orleans/pull/2395 rather?

I wonder what @veikkoeeva and @shayhatsor think.

mtdaniels:
Indeed I did - sorry about that

veikkoeeva:
@mtdaniels Do you mean reminders or storage deadlocks? It's entirely plausible reminders deadlock and it could be greatly mitigated (and likely removed altogether), there's no optimizations done to the queries and the table from what they originally were. The locking could be eased and it would greatly benefit using indexes to filter the where clauses.

mtdaniels:
@veikkoeeva - reminders deadlock.

veikkoeeva:
@mtdaniels Would it be OK to consider if you alter the table with some judicious use of indexes and possibly augment the queries and if it helps, make a PR here? :) I can work on this too, but I'm in a bind for about two weeks before I can take on this.

Indeed, the reminder queries are very plain, unoptimized and painful to watch. I apologize.

veikkoeeva:
You can alter the tables practically on the fly. Though currently you have to restart silos if you alter the queries, since currently the silos reload queries only then.

veikkoeeva:
@mtdaniels May I ask, how many remidners you have, by the way? If you hundreds of millions, we should do more extensive modifications. I think doable still, but adding a few simple indexes wouldn't be the most efficient way to go. :)

mtdaniels:
Typically not very many exist at a single time (on the order of hundreds).  There is somewhat high throughput (1,500/minute).  The majority of them are deleted before they ever fire.

[...]

PR #2556, opened 2017-01-02
add indexes to MySql tables

bronsh:
Improve MySql performance by using indexes

### DIFF
--- a/src/OrleansSQLUtils/CreateOrleansTables_MySql.sql
+++ b/src/OrleansSQLUtils/CreateOrleansTables_MySql.sql
@@ -42,7 +42,8 @@ CREATE TABLE OrleansMembershipVersionTable
 	Timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
 	Version INT NOT NULL DEFAULT 0,
 
-	CONSTRAINT PK_OrleansMembershipVersionTable_DeploymentId PRIMARY KEY(DeploymentId)
+	CONSTRAINT PK_OrleansMembershipVersionTable_DeploymentId PRIMARY KEY(DeploymentId),
+	KEY IX_Deployment_Version(DeploymentId, Version)
 );
 
 -- Every silo instance has a row in the membership table.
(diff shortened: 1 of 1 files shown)


sergeybykov:
@shayhatsor and @veikkoeeva are the SQL wizards. I wonder what they think.

veikkoeeva:
@bronsh I'm curious to know if you have analysis to share?

Here's what I think: `OrleansMembershipTable` and `OrleansMembershipVersionTable` have fairly low row count, so most likely they're cached in memory and indexing won't have a perceptible advantage or disadvantage. Indexing `OrleansRemindersTable` like that might help or be counterproductive. I think it's a fair assumption there are not many distinct `serviceId`s, so the index will likely be ignored as it's really discriminating the rows (it's maintenance costs, though). Indexing `GrainHash` will most likely help, but I'd be delighted to see someone share insight on `SELECT/UPDATE/INSERT` statistics. That aside, beneficial candinates would be also `GrainId` and `ReminderName`, but they're potentially a bit fattier indices and there are other issues too...

... Which leads to https://github.com/dotnet/orleans/issues/2429. In general adding indexes have merit and adding them judiciously make a positive effect, but to scale up reminders (and stats), the constructs would need a bit overhauling, I believe.

sergeybykov:
Should we close this for now?

veikkoeeva:
@sergeybykov, @bronsh Indices on `GrainId` and `ReminderName` could be beneficial and perhaps indexing the hash range, but indexing `ServiceId` might be (I haven't measured) hurtful because maintenance isn't free and it doesn't really filter the set effectively. Also `OrleansMembership*Table` likely is always quite shallow (i.e. in memory and processed quickly), so without further measuring, at least for me it's difficult to tell if it has impact or not.

Having sane defaults are useful, but I'm not sure these proposed measures.

sergeybykov:
@veikkoeeva I'm trying to decide what we should do here. Keep the PR open until there are perf benefits measurements?

sergeybykov:
I'll close it for now. Please reopen when we figure out the benchmarking strategy.

Issue #2547, opened 2016-12-27
[Proposal] Non-static SerializationManager changes


ReubenBond:
In order to complete the work required for non-static grain clients (#467), it would be very helpful to make `SerializationManager` also non-static so that we can deserialize grain references correctly.

`GrainReferences` need to have access to services provided by Orleans in order to function (for example, they need to make remote calls). Therefore, I think it's a good idea to change the way that serialization and deserialization work. I'm proposing these changes and want your feedback:

* Serialization methods accept an `ISerializationContext`/`IDeserializationContext`/`ICopyContext` in their signature. The methods will all look something like this:
```C#
public interface IExternalSerializer
{
    void Initialize(Logger logger);
    bool IsSupportedType(Type itemType);
    object DeepCopy(object source, ICopyContext context);
    void Serialize(object item, ISerializationContext context, Type expectedType);
    object Deserialize(Type expectedType, IDeserializationContext context);
}
```

`SerializationContext` & `DerserializationContext`:
* Add a `Stream` property which includes the `BinaryTokenStreamWriter`/`BinaryTokenStreamReader`.
* Add a `ServiceProvider` property of type `IServiceProvider`.
* Add a `SerializationManager` property for convenient access to `SerializationManager` (could also be accessed via DI).
* Are pooled rather than being thread-static.

The implications of this are:
* `SerializationManager` can be made non-static
* Serializers can access services provided by Orleans (or the user, since it's the same container)
* Serializers will need to be adjusted to satisfy the new signature (i.e, this is a **breaking change**), but I will create Roslyn Analyzers to ensure that breakages aren't hidden or inexplicable: ![Roslyn Analyzer Prototype](http://i.imgur.com/koq7g2t.gif)

Alternatives:
* Keep method signatures the same and pass those context objects inside of `BinaryTokenStreamReader` and `BinaryTokenStreamWriter`. I think this is messy and myopic.
* Add an `IOnDeserialized` interface which accepts an `IServiceProvider`. I played with this. It will help for non-static grain clients, but not with making `SerializationManager` non-static.
* Others?

Thoughts?

jdom:
I agree with adding a context, and avoid thread-statics, static singletons, and I would even avoid the IServiceProvider for getting the SerializationManager (while doing serialization, the serialization manager is a very important piece of the context and shouldn't be hidden by the service provider, which should be a last resort for edge cases).

[...]

PR #2558, opened 2017-01-03
Initial work to make SerializationManager non-static (#2547)

ReubenBond:
This is an implementation of #2547, & is related to #467.

This change moves `SerializationContext.Current` & `DeserializationContext.Current` to `SerializationManager`, keeps them static, but makes them private. It continues to use static `SerializationManager` methods in the generated code, but it passes in all of the context information required - so we can begin to remove global state from those methods down the track.

### DIFF
--- a/src/Orleans/Messaging/Message.cs
+++ b/src/Orleans/Messaging/Message.cs
@@ -1188,12 +1188,12 @@ namespace Orleans.Runtime
             {
                 HeadersContainer input = (HeadersContainer)untypedInput;
                 var headers = input.GetHeadersMask();
-                var stream = context.StreamWriter;
-                stream.Write((int)headers);
+                var writer = context.StreamWriter;
+                writer.Write((int)headers);
                 if ((headers & Headers.CACHE_INVALIDATION_HEADER) != Headers.NONE)
                 {
                     var count = input.CacheInvalidationHeader.Count;
-                    stream.Write(input.CacheInvalidationHeader.Count);
+                    writer.Write(input.CacheInvalidationHeader.Count);
                     for (int i = 0; i < count; i++)
                     {
                         WriteObj(context, typeof(ActivationAddress), input.CacheInvalidationHeader[i]);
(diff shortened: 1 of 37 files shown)


src/Orleans/Messaging/Message.cs
```diff
@@ -601,7 +606,7 @@ private List<ArraySegment<byte>> Serialize_Impl(out int headerLengthOut, out int
             {
                 BufferPool.GlobalPool.Release(headerBytes);
             }
-            headerBytes = headerStream.ToBytes() as List<ArraySegment<byte>>;
+            headerBytes = context.Stream.ToBytes() as List<ArraySegment<byte>>;
             int headerLength = headerBytes.Sum(ab => ab.Count);
```
attilah: This was not a change in this PR, but what about using context.Stream.CurrentOffset instead of the allocating enumerator in the Sum Linq extension method?

It is doable for both header and body here.

ReubenBond: It's a bit complicated to do this for the body, since it's not always set directly from a stream. We could avoid allocations by writing a helper function.

I consider this out of scope for this PR, though.

src/Orleans/Serialization/SerializationContext.cs
```diff
@@ -3,6 +3,27 @@
...
+        /// <param name="original">Original object.</param>
+        /// <param name="copy">Copy object that will be the serialized form of the original.</param>
+        void RecordCopy(object original, object copy);
+
+        object CheckObjectWhileCopying(object raw);
+    }
+
+    public interface ISerializationContext
+    {
+        BinaryTokenStreamWriter Stream { get; }
```
ReubenBond: Sounds good to me, I'll rename them.

src/Orleans/Messaging/Message.cs
attilah: Yes I saw that too, but I was able to overcome that restriction in an "else" branch above, but sure we can exclude it in this PR, not a blocking for it and I saw another place in Message.cs which can be enhanced.

src/Orleans/Messaging/Message.cs
```diff
@@ -1177,7 +1188,7 @@ public static void Serializer(object untypedInput, ISerializationContext context
             {
                 HeadersContainer input = (HeadersContainer)untypedInput;
                 var headers = input.GetHeadersMask();
-                var stream = context.Stream;
+                var stream = context.StreamWriter;
```
attilah: The variable name left as-is, below in ```BinaryFormatterSerializer``` it has the proper name of ```reader``` and ```writer```.

ReubenBond: ok, renamed them all to either 'reader' or 'writer'

PR #2561, opened 2017-01-03
Skip assemblies that target wrong platform or are non-CLR assemblies

Carlm-MS:


src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -319,9 +319,16 @@ private bool ReflectionOnlyLoadAssembly(string pathName, out Assembly assembly,
...
                 if (SimulateReflectionOnlyLoadFailure)
                     throw NewTestUnexpectedException();
-                
+
                 assembly = Assembly.ReflectionOnlyLoadFrom(pathName);
             }
+            catch (BadImageFormatException)
+            {
+                // skip assemblies that are don't target the correct platform or are non-clr assemblies.
+                complaints = null;
```
ReubenBond: Would a complaint be useful in this case?

ReubenBond:
One comment, otherwise LGTM

src/Orleans/AssemblyLoader/AssemblyLoader.cs
Carlm-MS: I'll update the review. 

It seems that BadImageFormatException in the context of ReflectionOnlyLoad only captures scenarios where a native assemblies is inspected in a managed context.  It does not capture scenarios where a managed assembly is targeting x86 and loaded in a 64-bit process - the reflectionOnlyLoad succeeds but the assembly load fails with a bad image exception.

I'll peek at the PE header to see if the assembly is targeting the correct platform.

ReubenBond: Interesting. You could possibly work out the compatibility by checking both the loaded assembly's `AssemblyName`'s `ProcessorArchitecture` against `Environment.Is64BitProcess`

image

attilah:
Before ae merge it in we need to discuss it IMHO, since this is one of the problems I addressed in my AssemblyLoader work.

ReubenBond:
@attilah should we discuss on this PR then?

Carlm-MS:
The updated commit checks for compatibility of machine type to the currently running executable image before trying to reflection-only load the assembly.

attilah:
@Carlm-MS thanks for updating the PR it looks much consumable without the code that we're getting into the PE and COFF handling stuff.

LGTM.

Issue #2552, opened 2016-12-28
[Question] Unexpected error serializing message - Collection was modified


mhertis:
Can you guys help me figure it out why I'm getting exceptions below. It seems that when I push (or consume?) message to the stream, serialization fails because dictionary was modified?

`Runtime.Messaging.GatewaySender/GatewaySiloSender_1	100.104.2.49:11111]	Unexpected error serializing message Request S100.104.2.49:30000:0*grn/5B91ACDA/00000000+WDC:87a3d5d3-d4e1-62b1-a05e-11bf2de91f72@15a52896->*cli/467c4d8e@eb8000bcd70c748e #7382800: global::Orleans.Streams.IStreamConsumerExtension:DeliverImmutable() on the gateway	`

```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Enumerator.MoveNext()
   at Orleans.Serialization.BuiltInTypes.SerializeDictionary[K,V](Object original, BinaryTokenStreamWriter stream, Type expected)
   at Orleans.Serialization.SerializationManager.SerializeInner(Object obj, BinaryTokenStreamWriter stream, Type expected)
   at Orleankka.OrleansCodeGenNomnio_Unullo_Core_Contracts_MessageEnvelopeSerializer`1.Serializer(Object untypedInput, BinaryTokenStreamWriter stream, Type expected)
   at Orleans.Serialization.SerializationManager.SerializeInner(Object obj, BinaryTokenStreamWriter stream, Type expected)
   at Orleans.Serialization.BuiltInTypes.SerializeStringObjectDictionary(Object original, BinaryTokenStreamWriter stream, Type expected)
   at Orleans.Serialization.SerializationManager.SerializeMessageHeaderDictHelper(Dictionary`2 headers, BinaryTokenStreamWriter stream)
   at Orleans.Serialization.SerializationManager.SerializeMessageHeaders(Dictionary`2 headers, BinaryTokenStreamWriter stream)
   at Orleans.Runtime.Message.Serialize_Impl(Int32& headerLengthOut, Int32& bodyLengthOut)
   at Orleans.Runtime.Messaging.Gateway.GatewaySender.Send(Message msg, Socket sock)
```

attilah:
Cannot be the case that the dictionary you're passing receives new items?!

gabikliot:
[...]

PR #2568, opened 2017-01-05
Add comment to generated DeepCopier code when [Immutable] is being used.

ReubenBond:
In response to #2552, this adds a comment which makes it more obvious why a DeepCopy method is just returning its input.

An example of the output is this:
```C#
public static global::System.Object DeepCopier(global::System.Object original)
{
    // No deep copy required since Orleans.TableVersion is marked with the [Immutable] attribute.
    return original;
}
```


gabikliot:
Thanks Reuben, looks great. Sergey already merged.

PR #2569, opened 2017-01-05
[WIP] OrleansDockerUtils -> Docker and Docker Swarm Clusters for Orleans

galvesribeiro:
This PR brings support to Docker and Docker Swarp cluster for Orleans.

It is based on @ReubenBond's work on SF Membership Oracle but using the Docker APIs.

The design of this Membership Oracle is to not have a MBR just as SF doesn't. It works by communicating with Docker Daemon or Swarm endpoints using https://github.com/Microsoft/Docker.DotNet which is a wrapper over Docker APIs, and leveraging the `Labels` feature of docker. 

The idea is to group a set of containers by `deploymentId` on the same host (Docker Daemon) or across multiple hosts (Swarm cluster). Whenever a client start, the  `DockerGatewayProvider` connects to Docker and query for the current list of containers which are a silo, and are in the same `deploymentId` so it can build the gateway list. After that first list, it set a timer to refresh that list from docker. The same happens at the silo side which basically find the other silos by the  `deploymentId`.

Basically there is no persistence at all neither MBR tables since all the containers can be found by the labels and when a container dies, it is automatically removed from docker list so it will not return anymore.

The hosting model is basically any container engine which implement Docker APIs (i.e. Docker Daemon, Docker Swarm, etc.). There is no change on how things are deployed or a new OrleansHost for that. All people need to do is to set the appropriate labels on the container which has the Orleans application code and everything will just works as long as the container is able to reach over network the Docker API host (i.e. the container host or Swarm manager).

There are two things that I'll implement after this PR:

1. Listen to docker stream of events and add/remove silos live -> the streams are very unstable yet and not available on production docker engines.
2. At runtime apply a label to the container mentioning that it is dead -> due to a blocker on docker for windows, commit changes to a running container isn't available yet.

This is a WIP. I was able to run manual tests here and made 2 containers and a client talk perfectly. The current code is totally usable. I've simulated adding and removing silos/containers to the cluster and they join and leave just fine.

I'm now working on get a unit test project with some `DockerFile`s so all people will need to run the tests is have Docker installed.

I appreciate any feedback while I'm working in the tests so I can address them quickly.

### DIFF
--- a/src/OrleansDockerUtils/DockerMembershipOracle.cs
+++ b/src/OrleansDockerUtils/DockerMembershipOracle.cs
@@ -1,0 +1,147 @@
+﻿using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Orleans.Docker
+{
+    internal class DockerMembershipOracle : IMembershipOracle
+    {
+        private readonly Dictionary<SiloAddress, SiloEntry> silos = new Dictionary<SiloAddress, SiloEntry>();
+        private readonly ILocalSiloDetails localSiloDetails;
+        private readonly Logger log;
+        private readonly GlobalConfiguration globalConfig;
+
+        // Cached collection of active silos.
+        private volatile Dictionary<SiloAddress, SiloStatus> activeSilosCache;
+
+        // Cached collection of silos.
+        private volatile Dictionary<SiloAddress, SiloStatus> allSilosCache;
+
+        private Timer timer;
+        private DateTime lastRefreshTime;
+
+        /// <summary>
+        /// Initialize a new instance of the <see cref="DockerMembershipOracle"/> class
+        /// </summary>
+        /// <param name="localSiloDetails">The silo which this instance will provide membership information for.</param>
+        /// <param name="globalConfig">The cluster configuration.</param>
+        /// <param name="loggerFactory">The logger factory.</param>
+        public DockerMembershipOracle(
+            ILocalSiloDetails localSiloDetails,
+            GlobalConfiguration globalConfig,
+            Func<string, Logger> loggerFactory)
+        {
+            this.log = loggerFactory("MembershipOracle");
+            this.localSiloDetails = localSiloDetails;
+            this.globalConfig = globalConfig;
+            this.silos[SiloAddress] = new SiloEntry(SiloStatus.Created, SiloName);
+        }
+
+        /// <summary>
+        /// Status of this silo.
+        /// </summary>
+        public SiloStatus CurrentStatus => GetApproximateSiloStatus(SiloAddress);
+
+        /// <summary>
+        /// Address of this silo.
+        /// </summary>
+        public SiloAddress SiloAddress => localSiloDetails.SiloAddress;
+
+        /// <summary>
+        /// Name of this silo.
+        /// </summary>
+        public string SiloName => this.localSiloDetails.Name;
+
+        public Task BecomeActive()
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool CheckHealth(DateTime lastCheckTime)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IReadOnlyList<SiloAddress> GetApproximateMultiClusterGateways()
+        {
+            throw new NotImplementedException();
+        }
+
+        public SiloStatus GetApproximateSiloStatus(SiloAddress siloAddress)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Dictionary<SiloAddress, SiloStatus> GetApproximateSiloStatuses(bool onlyActive = false)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool IsDeadSilo(SiloAddress silo)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool IsFunctionalDirectory(SiloAddress siloAddress)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task KillMyself()
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task ShutDown()
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task Start()
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task Stop()
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool SubscribeToSiloStatusEvents(ISiloStatusListener observer)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool TryGetSiloName(SiloAddress siloAddress, out string siloName)
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool UnSubscribeFromSiloStatusEvents(ISiloStatusListener observer)
+        {
+            throw new NotImplementedException();
+        }
+
+        private class SiloEntry
+        {
+            public SiloEntry(SiloStatus status, string name)
+            {
+                Status = status;
+                Name = name;
+            }
+
+            public SiloStatus Status { get; }
+            public string Name { get; }
+
+            /// <summary>
+            /// Gets or sets a value indicating whether this entry was updated.
+            /// </summary>
+            public bool Refreshed { get; set; }
+        }
+    }
+}
--- a/src/OrleansDockerUtils/DockerMembershipOracle.cs
+++ b/src/OrleansDockerUtils/DockerMembershipOracle.cs
@@ -7,7 +7,6 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Orleans.Docker
@@ -32,8 +31,7 @@ namespace Microsoft.Orleans.Docker
         private readonly ConcurrentDictionary<ISiloStatusListener, ISiloStatusListener> subscribers =
             new ConcurrentDictionary<ISiloStatusListener, ISiloStatusListener>();
         private readonly object updateLock = new object();
-        private readonly Dictionary<SiloAddress, SiloEntry> silos = new Dictionary<SiloAddress, SiloEntry>();
-        private readonly TimeSpan refreshPeriod = TimeSpan.FromSeconds(30);
+        private readonly Dictionary<SiloAddress, SiloEntry> silos = new Dictionary<SiloAddress, SiloEntry>();        
         private readonly ILocalSiloDetails localSiloDetails;
         private readonly Logger log;
         private readonly GlobalConfiguration globalConfig;
@@ -47,10 +45,7 @@ namespace Microsoft.Orleans.Docker
 
         // Cached collection of multicluster gateways.
         private volatile List<SiloAddress> multiClusterSilosCache;
-
-        private Timer timer;
-        private DateTime lastRefreshTime;
-
+        
         /// <summary>
         /// Initialize a new instance of the <see cref="DockerMembershipOracle"/> class
         /// </summary>
@@ -78,7 +73,7 @@ namespace Microsoft.Orleans.Docker
         /// <returns><see langword="true"/> if the participant is healthy, <see langword="false"/> otherwise.</returns>
         public bool CheckHealth(DateTime lastCheckTime)
         {
-            return lastRefreshTime.Add(refreshPeriod + refreshPeriod) > DateTime.UtcNow;
+            return resolver.LastRefreshTime > DateTime.UtcNow;
         }
 
         public IReadOnlyList<SiloAddress> GetApproximateMultiClusterGateways()
@@ -187,14 +182,7 @@ namespace Microsoft.Orleans.Docker
         /// </summary>
         public Task Start()
         {
-            timer = new Timer(
-                self => ((DockerMembershipOracle)self).RefreshAsync().Ignore(),
-                this,
-                refreshPeriod,
-                refreshPeriod);
-
             resolver.Subscribe(this);
-            RefreshAsync().Ignore();
             UpdateStatus(SiloStatus.Joining);
 
             return TaskDone.Done;
@@ -362,24 +350,6 @@ namespace Microsoft.Orleans.Docker
             }
         }
 
-        private async Task RefreshAsync()
-        {
-            try
-            {
-                await resolver.Refresh();
-
-                lastRefreshTime = DateTime.UtcNow;
-            }
-            catch (Exception exception)
-            {
-                log?.Warn(
-                    (int)ErrorCode.Docker_MembershipOracle_ExceptionRefreshingSilos,
-                    "Exception refreshing silos.",
-                    exception);
-                throw;
-            }
-        }
-
         /// <summary>
         /// Clears the cached data.
         /// </summary>
@@ -442,9 +412,7 @@ namespace Microsoft.Orleans.Docker
         }
 
         private void StopInternal()
-        {
-            timer?.Dispose();
-            timer = null;
+        {            
             resolver.Unsubscribe(this);
         }
 
--- a/src/OrleansDockerUtils/DockerMembershipOracle.cs
+++ b/src/OrleansDockerUtils/DockerMembershipOracle.cs
@@ -373,8 +373,8 @@ namespace Microsoft.Orleans.Docker
             catch (Exception exception)
             {
                 log?.Warn(
-                    (int)ErrorCode.Docker_MembershipOracle_ExceptionRefreshingPartitions,
-                    "Exception refreshing partitions.",
+                    (int)ErrorCode.Docker_MembershipOracle_ExceptionRefreshingSilos,
+                    "Exception refreshing silos.",
                     exception);
                 throw;
             }
--- a/src/OrleansDockerUtils/DockerMembershipOracle.cs
+++ b/src/OrleansDockerUtils/DockerMembershipOracle.cs
@@ -1,47 +1,19 @@
-﻿using Orleans.Runtime;
+﻿using Microsoft.Orleans.Docker.Models;
+using Microsoft.Orleans.Docker.Utilities;
+using Orleans;
+using Orleans.Runtime;
 using Orleans.Runtime.Configuration;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Orleans.Docker
 {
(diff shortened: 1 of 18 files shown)


src/OrleansDockerUtils/OrleansDockerExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (parameters.TryGetValue("DaemonEndpoint".ToUpper(), out username))
+            {
+                parameters.TryGetValue("Password".ToUpper(), out password);
+
+                credentials = new BasicAuthCredentials(username, password);
+            }
+
+            if (parameters.TryGetValue("Certificate".ToUpper(), out certificate))
+            {
+                if (!File.Exists(certificate)) throw new FileNotFoundException("Unable to find certificate file");
```
ReubenBond: Windows users might be confused about this, since they're used to having a cert store and referring to certs using thumbprints rather than file names.

src/OrleansDockerUtils/OrleansDockerExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+                .ToDictionary(_ => _.First().ToUpper(), _ => _.Last());
+
+            string dockerDaemonEndpoint = string.Empty;
+            string username = string.Empty;
+            string password = string.Empty;
+            string certificate = string.Empty;
+
+            Credentials credentials = null;
+
+            if (parameters.TryGetValue("DaemonEndpoint".ToUpper(), out username))
```
ReubenBond: DaemonEndpoint is the username?

The logic in this method isn't straightforward because if the user specifies DaemonEndpoint as well as Certificate, the credentials will first be set to BasicAuthCredentials and then to CertificateCredentials.

src/OrleansDockerUtils/DockerSiloResolver.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            this.deploymentId = deploymentId;
+            this.dockerClient = dockerClient;
+            log = loggerFactory?.Invoke(nameof(DockerSiloResolver));
+        }
+
+
+        public async Task Refresh()
+        {
+            var result = await Task.WhenAll((await dockerClient.Containers.ListContainersAsync(
```
ReubenBond: `await Task.WhenAll(await ...)`?

src/OrleansDockerUtils/DockerMembershipOracle.cs
```diff
@@ -0,0 +1,468 @@
...
+
+        private readonly ConcurrentDictionary<ISiloStatusListener, ISiloStatusListener> subscribers =
+            new ConcurrentDictionary<ISiloStatusListener, ISiloStatusListener>();
+        private readonly object updateLock = new object();
+        private readonly Dictionary<SiloAddress, SiloEntry> silos = new Dictionary<SiloAddress, SiloEntry>();
+        private readonly TimeSpan refreshPeriod = TimeSpan.FromSeconds(30);
+        private readonly ILocalSiloDetails localSiloDetails;
+        private readonly Logger log;
+        private readonly GlobalConfiguration globalConfig;
+        private readonly IDockerSiloResolver resolver;
```
ReubenBond: Just to be sure, which parts of this file differ from the Service Fabric oracle other than this type?

gabikliot:
Can you provide a bit more details, describing how exactly you integrate with docker, where is the MBR table or who and how membership Oracle is provided, what is the hosting model, ... So people can understand without reading the code.

src/OrleansDockerUtils/OrleansDockerExtensions.cs
galvesribeiro: Yeah I agree. But this is the way Linux users deal with certs. The cert stores are not implemented for linux yet so I rather stick with this filename for now.

src/OrleansDockerUtils/OrleansDockerExtensions.cs
galvesribeiro: No, it is the Docker Daemon or Docker Swarm endpoint uri like `http://10.0.0.1:2375` or `tcp://localhost:2375`

src/OrleansDockerUtils/Utilities/ErrorCode.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Microsoft.Orleans.Docker.Utilities
+{
+    internal enum ErrorCode
+    {
+        Runtime = 100000,
+        DockerBase = Runtime + 4400,
+        Docker_GatewayProvider_ExceptionNotifyingSubscribers = DockerBase + 1,
+        Docker_GatewayProvider_ExceptionRefreshingGateways = DockerBase + 2,
+        Docker_MembershipOracle_ExceptionNotifyingSubscribers = DockerBase + 3,
+        Docker_MembershipOracle_ExceptionRefreshingPartitions = DockerBase + 4
```
galvesribeiro: Ops... Bad copy and paste... Removing

src/OrleansDockerUtils/DockerSiloResolver.cs
galvesribeiro: Done

src/OrleansDockerUtils/DockerSiloResolver.cs
```diff
@@ -0,0 +1,91 @@
...
+                })).Select(c => dockerClient.Containers.InspectContainerAsync(c.ID)));
+
+            lock (updateLock)
+            {
+                silos = result.Select(_ => 
+                {
+                    return new DockerSiloInfo(_.Config.Hostname,
+                        new IPEndPoint(IPAddress.Parse(_.NetworkSettings.Networks.First().Value.IPAddress), 
+                            int.Parse(_.Config.Labels[DockerLabels.SILO_PORT])),
+                        new IPEndPoint(IPAddress.Parse(_.NetworkSettings.Networks.First().Value.IPAddress),
```
galvesribeiro: Done

src/OrleansDockerUtils/DockerMembershipOracle.cs
galvesribeiro: Nothing is different actually. Just this resolver interface... Like I said yesterday, too much boilerplate code... We should later refactory this as a common abstract class or something in another PR.

galvesribeiro:
@gabikliot sorry, I was actually hoping just @ReubenBond jump in since he just implemented the Service Fabric Membership Oracle and this implementation is almost the same. But yes, I'll edit the PR and include more details. 

Regarding MBR, there is no MBR table at all, just as in SF. I'll detail more on the PR text. Thanks

src/OrleansDockerUtils/OrleansDockerExtensions.cs
ReubenBond: The variable is called `username`

src/OrleansDockerUtils/OrleansDockerExtensions.cs
ReubenBond: I'm fine with that, I was just pointing out that the name might be a cause for confusion with some users. A name like `CertificatePath` or `CertFile` might be clearer.

src/OrleansDockerUtils/OrleansDockerExtensions.cs
galvesribeiro: Oh! now I saw the problem. It was a change before the push which led to this bug... the string on the key is wrong... It should be `"Username"`. Fixed.

ReubenBond:
Documentation on this would be great - I need to document "Hosting Orleans on Service Fabric", too.

galvesribeiro:
Yes, the docs is the PR right after this one @ReubenBond 

Unfortunately we sill need to write docs in a separated PR to use the `gh-pages` branch (unless GH find a way to multibranch PR!)

sergeybykov:
This is a great effort!

Let's open an issue to discuss the design instead of mixing it here with code review feedback.

I'm trying to comprehend it, and have some basic initial questions.

1. How does it handle generations (epochs) of silos? I see the code that parses it from `DockerLabels.GENERATION` : `var generation = int.Parse(container.Config.Labels[DockerLabels.GENERATION]);`, but not where it is getting set. If the silo process restarts within a container, it will have a new generation. How does Docker know that?

2. >due to a blocker on docker for windows, commit changes to a running container isn't available yet.

Does this mean that there is no way to mark a silo as dead?

3. In general, Is there an implied equivalency here of "container is up" and "silo is running"? Can't we get into a situation that a silo is unresponsive, e.g. it's messaging queues are or threads are blocked, but Docker will consider the container healthy and keep it in the list of container/silos on the cluster?

gabikliot:
@galvesribeiro, I know how SF works and also what does it mean from Orleans perspective. I was asking about Docker Swarm. I am looking into more high level description of how it is used, both hosting and MBR, like here: http://dotnet.github.io/orleans/Documentation/Runtime-Implementation-Details/Consul-Deployment.html

For example, you can see here the long thread we went through when designing Consul based solution: https://github.com/dotnet/orleans/pull/1267
You can see all the questions that were asked, all the options we considred.

So can we please open an issue and discuss the design, like @sergeybykov suggested. Ideally, I think we prefer doing that BEFORE code is submitted for review, so that people can actually comprehend the code, after the design is ready. Without seeing the design, its hard to give any valuable feedback.

gabikliot:
Also would be interesting to ask why Swarm and not other Docker clustering technologies. Is there a specific reason, or customer who asked for that, or just a first one we picked without any particular reason?

galvesribeiro:
@sergeybykov 

> Let's open an issue to discuss the design instead of mixing it here with code review feedback.

Ok will do in a minute.

> How does it handle generations (epochs) of silos? I see the code that parses it from but not where it is getting set. If the silo process restarts within a container, it will have a new generation. How does Docker know that?

Docker containers are immutable. In no way a same container or its main process can restart. It is discarded and start a fresh one so the generation is pointless in this case. I'll describe in the issue how I do that.

> Does this mean that there is no way to mark a silo as dead

If the silo process is frozen or crash, the container is killed and a new one must be spawned. 1 container has only 1 main process which in this case is the silo process. So if it dies, the whole container die. More on that in the issue.

@gabikliot 

I agree, let me open an issue and link here in a minute.

> Also would be interesting to ask why Swarm and not other Docker clustering technologies. Is there a specific reason, or customer who asked for that, or just a first one we picked without any particular reason?

You can use any docker clustering platform as long as they respect Docker APIs and they do (almost). Swarm is by far Today's most used technology for Docker clustering and the one that comes from Docker Inc. If you look at the core, there is no explicit mention tie to Swarm. The same code that talk with a single Docker Daemon talk to Swarm. Its orthogonal. Just a matter of change endpoints. I'll dive into details in the issue.

Anyway, linking the issue here soon. Thanks for the comments.

galvesribeiro:
Btw, sidenote... I didn't added the .Net core projects yet, will do in the end since the code dont change. The tests I'm building some easy way to run from xUnit with a `DockerFile` so we can spin up containers on any dev machine.

src/OrleansDockerUtils/DockerSiloResolver.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            this.deploymentId = deploymentId;
+            this.dockerClient = dockerClient;
+            log = loggerFactory?.Invoke(nameof(DockerSiloResolver));
+        }
+
+
+        public async Task Refresh()
+        {
+            var containerList = await dockerClient.Containers.ListContainersAsync(
```
gabikliot: can you please replace  var containerList with the actual type, so its more readable.

src/OrleansDockerUtils/DockerSiloResolver.cs
```diff
@@ -0,0 +1,102 @@
...
+                                {LABEL_FILTER,
+                                    new Dictionary<string, bool>
+                                        {
+                                            {$"{DockerLabels.DEPLOYMENT_ID}={deploymentId}", true},
+                                            {$"{DockerLabels.IS_DOCKER_SILO}", true}
+                                        }
+                                },
+                            }
+                });
+            var inspectionResult = await Task.WhenAll(containerList.Select(c => dockerClient.Containers.InspectContainerAsync(c.ID)));
```
gabikliot: same comment about var here.

src/OrleansDockerUtils/DockerMembershipOracle.cs
```diff
@@ -0,0 +1,468 @@
...
+            activeSilosCache = null;
+            allSilosCache = null;
+            multiClusterSilosCache = null;
+        }
+
+        /// <summary>
+        /// Updates the status of this silo.
+        /// </summary>
+        /// <param name="status">The updated status.</param>
+        private void UpdateStatus(SiloStatus status)
```
gabikliot: UpdateStatus does not really do anything, it does not propagate the new status of this silos remotely to other silo. We need that in Orleans. There is a difference between Starting/Active, or Terminating/Stopping/Dead.

src/OrleansDockerUtils/DockerMembershipOracle.cs
```diff
@@ -0,0 +1,468 @@
...
+                    }
+                }
+            }
+        }
+
+        private async Task RefreshAsync()
+        {
+            try
+            {
+                await resolver.Refresh();
```
gabikliot: This is a bit weird. Oracle refreshes the docker resolver which notifies its subscribers (this Oracle) back about  the MBR. Cycle of dependences? Or did I misunderstand it.

gabikliot:
Thank you for the detailed explanation @galvesribeiro .
Can you please provide a bit more details (not too much, but still a bit) about Swarm membership. Just a bit background, to make sure the models are compatible.

galvesribeiro:
> Can you please provide a bit more details (not too much, but still a bit) about Swarm membership. Just a bit background, to make sure the models are compatible.

@gabikliot Sure, but sorry, I din't understand. You mean the *Docker Swarm membership implementation from Docker Inc.* or the the Docker Oracle Membership I implemented **using** Swarm for Orleans?

src/OrleansDockerUtils/DockerSiloResolver.cs
galvesribeiro: Done

src/OrleansDockerUtils/DockerSiloResolver.cs
galvesribeiro: Done

src/OrleansDockerUtils/DockerMembershipOracle.cs
galvesribeiro: Yeah you are right. I copy that over from the SF implementation. I just refactored the timer out of both gateway and oracle. Now it is inside the resolver and this dependency don't happen anymore. Thanks for notice that.

src/OrleansDockerUtils/DockerMembershipOracle.cs
galvesribeiro: I don't understand @gabikliot. What you mean by "it does not do anything"? 

```
foreach (var subscriber in subscribers.Values)
                {
                    subscriber.SiloStatusChangeNotification(SiloAddress, status);
                }
```

It is notifying all the registered subscribers about the new status on this method. I don't see your point here, sorry. Can you elaborate?

gabikliot: It does not propagate the new status of this silo to other silos (yes, it does update local subscribers).

When a silo goes lets say to terminating state (just an example), we need to notify all other silos about it. Its not enough to notify all local components. If we can't do distributed silo states, that's a major gap that we are opening compared to the main MBR Oracle and we need to think through the implications.

gabikliot:
I meant the former:  "Docker Swarm membership implementation from Docker Inc."

Mostly not how it is implemented, but what it provides to the users of that protocol. What are the semantics. Like does it notify or do you have to pool it, how fast it detects, does it have an option to specify health check ....

That knowledge will help review the latter (" Docker Oracle Membership I implemented using Swarm for Orleans").

galvesribeiro:
@gabikliot I'm updating the issue related to this PR and will ping you there in a minute.

src/OrleansDockerUtils/DockerMembershipOracle.cs
galvesribeiro: Ok, then we have the same problem in SF implementation because it does the same thing there. o.O

gabikliot: The SF solution is not merged yet, so not really a good source of example to  follow, at least until merged. 
We need to solve that, right?

ReubenBond: The propagation of silo status changes happens in [FabricMembershipOracle.cs:L348](https://github.com/dotnet/orleans/pull/2542/files#diff-287349e1f70e4eabfc71ecef5e59735bR342). The only states which are supported are `Active` and `Dead`, but they satisfy the distinction for `SiloStatus.IsTerminating()`.

Do you think this is an issue, @gabikliot?

galvesribeiro:
@gabikliot I've updated #2571 with more info related of health and membership. 

About the Swarm membership implementation (or any other orchestration engine for containers like Mesos, DC/OS and Kubernetes), there nothing that can help us in this case. Swarm has its own Raft implementation for its manager, but it is restricted to Swarm usage. The other orchestration engines have their own and suffer of the same problem. They only care about their cluster, and not the containers. What they care is only if the container is up or down. Nothing more.

src/OrleansDockerUtils/DockerMembershipOracle.cs
gabikliot: This is a somewhat critical question I think so it may deserve its own issue.
If we only need 2 states for SF livenes, then maybe we need only 2 for the regular case as well? Maybe we can simplify the silo state machine and get rid of all other states and stay with just 2?
One need to look at what we are getting out if the other states. I wonder what @sergetbykov thinks. Would like to hear his opinion.

One thing is sure in my eyes: we should try to have similar silo state machines in all livenes cases, and not just ditch states in some cases cause we can't implementat them. If u go that route, it may work, or more likely you will be supporting diverging implementaion and at the end paying higher cost.

Alternatively: when I wrote the first integration with SF(WF at a time) 3-4 years ago, it was similar to your code now, and in addition my plan was to use their build in key value store to store silo states. I didn't implemenent it, but maybe it can work.

 If u go that route, them yet another alternative is to ditch SF livenes and just use the key value store to implement MBR table. That is exactly what we did for Consul. It works, so here is an advantage. And Consul is comparable to SF, so why not? More choices.:-)

gabikliot:
Does swarm have key value store? To store silo states or implement MBR table?

galvesribeiro:
@gabikliot no, it doesn't. And that is why the whole point of discussion here. If we have it, we would be implementing the `IMembershipTable` like in Consul and Zookeeper. :(

galvesribeiro:
Guys I'm closing this issue since it is not required anymore. See here: https://github.com/dotnet/orleans/issues/2571#issuecomment-276229777

Thanks for all the feedback.

PR #2577, opened 2017-01-09
Simplify initialization of DI container

ReubenBond:
Simple change to pre-construct & add services to the DI container before passing it to the startup builder, rather than passing a lambda to add the services to it.

Since we're constructing the container and passing it to the user, 'TryAdd' was misleading, so we're calling 'Add' now.


attilah:
LGTM, nice cleanup! ;-)

PR #2581, opened 2017-01-11
Add flag for test to assume that all silos in the cluster share the same set of grain types at startup

benjaminpetit:
Add flag for test to assume that all silos in the cluster share the same set of grain types at startup

Some tests are failing because all silos didn't refreshed the cluster interface map in time before the test started. 

To avoid introducing delay in the test startup, we assume that all silos not in the cluster type map support the same set of types as the local one.

### DIFF
--- a/src/Orleans/Configuration/GlobalConfiguration.cs
+++ b/src/Orleans/Configuration/GlobalConfiguration.cs
@@ -436,7 +436,7 @@ namespace Orleans.Runtime.Configuration
 
         public int ActivationCountBasedPlacementChooseOutOf { get; set; }
 
-        public bool AssumeAllSilosEqualsForTesting { get; set; }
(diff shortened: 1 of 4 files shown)


src/Orleans/Configuration/GlobalConfiguration.cs
```diff
@@ -436,6 +436,7 @@ internal void SetReminderServiceType(ReminderServiceProviderType reminderType)
 
         public int ActivationCountBasedPlacementChooseOutOf { get; set; }
 
+        public bool AssumeAllSilosEqualsForTesting { get; set; }
```
sergeybykov: Nit: Maybe `AssumeHomogenousSilosForTesting`?

benjaminpetit: Yes it is better

PR #2582, opened 2017-01-11
Support messages over the max CloudQueueMessage size

Carlm-MS:
when sending a large number of events or events of sufficient size, the max size of an azure queue message is exceeded.  This commit does the following

1. Adds a config paramto the azure queue adapter to support large messages
2. Splits messages into chunks that can be sent as cloud queue messages
3. Reassembles the chunks in azure queue batch containers

### DIFF
--- a/src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueAdapterReceiver.cs
+++ b/src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueAdapterReceiver.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Microsoft.WindowsAzure.Storage.Queue;
 using Orleans.AzureUtils;
 using Orleans.Runtime;
+using Orleans.Runtime.Host.Providers.Streams.AzureQueue;
 using Orleans.Streams;
 
 namespace Orleans.Providers.Streams.AzureQueue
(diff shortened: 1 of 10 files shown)


src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueAdapterReceiver.cs
```diff
@@ -79,10 +82,38 @@ public Task Initialize(TimeSpan timeout)
...
+                        }
+
+                        if (!messageSegments.TryGetValue(segment.Guid, out segmentList))
+                        {
+                            segmentList = new List<MessageSegment>();
+                            messageSegments.Add(segment.Guid, segmentList);
+                        }
+
+                        segmentList.Add(segment);
+                        if (segment.Count == segmentList.Count)
```
jason-bragg: how do we know we're not getting duplicate segments?  Azure queue will redeliver messages that have not been removed from the queue, so duplicates can show up, correct?

src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueAdapterReceiver.cs
```diff
@@ -18,24 +19,26 @@ internal class AzureQueueAdapterReceiver : IQueueAdapterReceiver
...
-        public static IQueueAdapterReceiver Create(QueueId queueId, string dataConnectionString, string deploymentId, TimeSpan? messageVisibilityTimeout = null)
+        public static IQueueAdapterReceiver Create(QueueId queueId, string dataConnectionString, string deploymentId, bool hasLargeMesageSupport = false, TimeSpan? messageVisibilityTimeout = null)
         {
             if (queueId == null) throw new ArgumentNullException("queueId");
             if (String.IsNullOrEmpty(dataConnectionString)) throw new ArgumentNullException("dataConnectionString");
             if (String.IsNullOrEmpty(deploymentId)) throw new ArgumentNullException("deploymentId");
-            
             var queue = new AzureQueueDataManager(queueId.ToString(), deploymentId, dataConnectionString, messageVisibilityTimeout);
-            return new AzureQueueAdapterReceiver(queueId, queue);
+            return new AzureQueueAdapterReceiver(queueId, queue, hasLargeMesageSupport);
```
jason-bragg: How are the original CloudQueueMessages deleted from the queue in MessagesDeliveredAsync?

sergeybykov:
Did you include vNext/src/blah.txt by mistake or there's a reason for that?

Carlm-MS:
It was a mistake.

sergeybykov:
My general concern here is that this would add a layer of complexity and an implicit protocol on top of the basic Azure Queues functionality. Conceptually, the same could be done on top of any other queuing service to overcome their message size limits. So it doesn't strike me as an Azure Queues specific extension.

A simpler, and arguably more popular, pattern we've seen being employed is a queue+blob combination where a reference to the blob containing the actual message is enqueued. Although arguably it involves more roundtrips which may increase latency.

My concern about the extra complexity is primarily around the various possible failure cases that the protocol would have to handle correctly, for example, failing to dequeue individual messages, repartitioning and transfer responsibility for a queue between silos in between processing of two messages that are parts of the same larger message, violation of message ordering when messages are reenqueued after a failure to process, etc., etc. It seems to me it would be hard to guarantee that the protocol robustly handles all of such cases, as they are not easy to simulate.

My opinion is we better leave the Azure Queues provider simple. Support for larger messages can be done either via the queue+blob approach or/and as a separate stream provider, thus leaving the existing one untouched other than maybe making some of its methods virtual for subclassing.

The ideal case to me would be if this could be done as a 'meta-provider' that would wrap an arbitrary stream provider to extend its message size limit (similar to how `ShardedStorageProvider` does that over arbitrary persistence providers). But I'm not sure if this is easily achievable.

Carlm-MS:
@sergeybykov I've opened up an issue for this.

PR #2583, opened 2017-01-11
move remaining tests to vNext

xiazen:
-move remaining tests which can be moved to vNext
-summarize context around tests which are not ready to be moved to vNext
yet

summary as follows , excluded tests for now
    ..\..\..\test\TesterInternal\CounterControlProgTests.cs; -- CounterControl project is not in vnext
    ..\..\..\test\TesterInternal\GeoClusterTests\\*.cs;  -- uses TestingSiloHost which is being deprecated ATM
    ..\..\..\test\TesterInternal\HostedTestClusterBase.cs; -- uses TestingSiloHost which is being deprecated ATM
    ..\..\..\test\TesterInternal\OrleansHostProgTests.cs;  -- OrleansHost proj is not in vNext
    ..\..\..\test\NonSiloTests\AssemblyLoaderTests.cs -- AssemblyLoading complication in vNext
    ..\..\..\test\Tester\SiloInitializationTests -- Orleans.Runtime.Host.SiloHost is not MarshalByRefObject in vNext, but AppDomain needed it to be

    ..\..\..\test\TesterInternal\General\BootstrapProvidersTests.cs; -- migrate off AppDomainTestHook #2743
    ..\..\..\test\TesterInternal\StorageTests\PersistenceGrainTests.cs; -- migrate off AppDomainTestHook #2743
    ..\..\..\test\TesterInternal\StreamingTests\StreamPubSubReliabilityTests.cs  -- migrate off AppDomainTestHook #2743


test/TesterInternal/General/RequestContextTest.cs
```diff
@@ -169,22 +181,33 @@ public void Dispose()
...
+#else
             Trace.CorrelationManager.ActivityId = activityId;
+#endif
            Assert.Null(RequestContext.Get(RequestContext.E2_E_TRACING_ACTIVITY_ID_HEADER));
             Guid result = await grain.E2EActivityId();
             Assert.Equal(activityId,  result);  // "E2E ActivityId not propagated correctly"
             RequestContext.Clear();
 
+#if NETSTANDARD
+            RequestContext.ActivityId.Value = nullActivityId;
```
ReubenBond: I wonder, could these be unified so that setting RequestContext.ActivityId.Value works on Full .NET, too?

xiazen:
@dotnet-bot test this please

test/TesterInternal/General/RequestContextTest.cs
xiazen: the way RequestContext.ActivityId is implemented in net core currently is not supported in full .net. 

I think @dVakulen looked at this problem in #2200 before, maybe he has some thought on this problem.

dVakulen: We need 4.6 version of full framework in order to use AsyncLocal.

sergeybykov: We are switching to .NET 4.6.2 in 1.5.0.

xiazen: I suggest we leave this as is. And after we switch to 4.6. I will do the clean up

jdom: agreed, we still need to think about back-compat implications of this. We don't want to blindly switch to AsyncLocal. We did on .NET Core because we didn't have an alternative, but not sure what it would mean for the 1.5 release that still only targets .NET.

jdom: Although agree with the original pain point. I think our tests can use a wrapper that just has a NETSTANDARD_TODO wrapper in front of it, so we don't need to have conditional compilation ALL over the place, just in 1 place in our test project.

vNext/test/DefaultCluster.Tests/DefaultCluster.Tests.csproj
```diff
@@ -15,7 +15,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="..\..\..\test\DefaultCluster.Tests\**\*.cs" Exclude="..\..\..\test\DefaultCluster.Tests\FSharpGrainTests.cs;..\..\..\test\DefaultCluster.Tests\obj\**\*.cs;..\..\..\test\DefaultCluster.Tests\bin\**\*.cs;..\..\..\test\DefaultCluster.Tests\Properties\*.cs" />
```
jdom: this goes back and forth... let's keep it without line feeds, as some editors tend to dislike them and and XML friendly line feeds.

src/Orleans/CodeGeneration/TypeUtils.cs
```diff
@@ -491,18 +491,6 @@ public static bool IsGrainMethodInvokerType(Type type)
...
-        {
-            return Type.GetType(fullName, true);
-        }
-
-        public static bool TryResolveType(string fullName, out Type type)
-        {
-            type = Type.GetType(fullName, false);
-            return type != null;
-        }
-#else
```
jdom: This is a non-trivial change. I'd like to understand a little bit more of what prompted it, because this wouldn't just affect tests, but runtime behavior.

vNext/test/DefaultCluster.Tests/DefaultCluster.Tests.csproj
xiazen: ok

src/Orleans/CodeGeneration/TypeUtils.cs
xiazen: without this, the test cannot resolve type in another assembly. 

ASAIK, this two methods rely on CachedTypeResolver, which I also enabled in netstandard. I read through the class, I think the class itself doesn't have complication in netstandard. it's its dependency which relates to assembly loading which has complication. 

Maybe tell me more of what runtime behavior it will effect

test/TesterInternal/General/RequestContextTest.cs
xiazen: Yes I can do a wrapper. 
As for using AsyncLocal, I think the back-compat issue here is how to set ActivityId from 1.5 on.
I think we are not questioning the reliability of using AsyncLocal to store ActivityId, right? 
I would vote to keep ActivityId as part of RequestContext, which means to set ActivityId like:
`RequestContext.ActivityId.Value = nullActivityId`,
instead of using 
`Trace.CorrelationManager.ActivityId = nullActivityId`

it seems to me is a better design, even though it is a breaking change.

Or we can do `RequestContext.ActivityId = nullActivityId`, by modifying our current implementation of `RequestContext.ActivityId`

src/Orleans/CodeGeneration/TypeUtils.cs
xiazen: as for runtime, it will enable runtime to resolve types in another assembly , like in full dotnet do.

jdom:

jdom:

Issue #2519, opened 2016-12-14
Client Observer Reference doesn't work on second silo


brianqcgames:
This is behavior I am seeing in 1.3.1

Setup:
I am running 2 silos using zookeeper for membership
I am running 1 client using zookeeper for gateway discovery
I see in the logs that both silos record the open socket from the client
A grain (TestGrain) is created that holds an observer reference to an object in the client. The grain is usually created in silo 1 because of timing.
The grain is persisting the reference to redis for temporary storage.
I kill silo 1 (or whatever silo is holding the TestGrain object).
If the client re-activates TestGrain the state appears to be properly re-created from redis including the observer reference.
If TestGrain then tries to send a notification to the client using the observer reference it will sometimes fail with the error: 
Discarding OneWay rejection for message OneWay S127.0.0.1:11112:219365805*grn/E5E4FD49/00000000@4a6885bf->*cli/d85ac3a0f126053a-7b70-462e-8ac5-a2d8ffd9a2e5 #178: global::QCServicesGrainInterfaceLibrary.ITestGrainObserver:OnSimpleNotification(). Exc = No activation for client *cli/d85ac3a0

What is strange is it doesn't always fail and I can't find a consistent way to make it fail. My hypothesis is that it fails when the silo that was the "first" silo a client connects to is the one that was shut down but I can't reliably produce that scenario or tell it is actually happening because I am not sure how the client gateway connections work.

brianqcgames:
Tagging @benjaminpetit  as Sergey mentioned he was looking at client-gateway communication recently.

benjaminpetit:
Do you shutdown `Silo 1` properly, or do you kill it?

I think it will happen if `Silo 1` was the owner of the `GrainDirectoryPartition`: when this silo is killed, this partition is lost. I am not sure that a client-gateway communication could help here.

Could you provide a unit test for this? If not I will try to add one.

brianqcgames:
It was a kill. I was trying to simulate silo failure.

So the GrainDirectoryPartition only lives on a single silo? 

I guess the larger question is how resilient should we expect Orleans to be in the event of catastrophic silo failure. Are there other cases where data living on a single silo can can cause failures throughout the cluster?

I will see about getting a unit test built after the holidays.

Thanks

benjaminpetit:
The `GrainDirectoryPartition` is distributed across all silos in the cluster.

In a 2-silo cluster, if one is killed you lose (assuming both silos are under the same load):
[...]

PR #2587, opened 2017-01-11
Trigger registration of clients connected to the gateways in the directory when a silo is dead

benjaminpetit:
From #2519 

When a silo stop is dead, we loose all data that was on its directory, including client entries. We already have a mechanism that refresh these entries every `ClientRegistrationRefresh` in `ClientObserverRegistrar`.

To improve reliability in this case, `ClientObserverRegistrar` we could listen to membership change, and force a refresh of the clientId entries that are connected to the gateway, without waiting for the timer to fire.


gabikliot:
I already reviewed that, LGTM. Good improvement @benjaminpetit !

Issue #2591, opened 2017-01-13
[Breaking Change] Improve serializer registration


ReubenBond:
Since we will not be supporting static `SerialzationManager.Register(...)` method, we need to change the signature of the method used to register custom serializers so that it takes some kind of context (eg, `ISerializerRegistrar`) so that serializers can be registered.

In a local branch, I've made the following changes:
* `[Serializer(typeof(TargetType))]` automatically registers the type it's attached to as the serializer for `TargetType`. This includes when `TargetType` is an open generic.
* Generated serializers no longer have `Register` methods or the associated attribute.
* No additional class is generated for generic types since the Register method is no longer needed. i.e, no "master registerer" class.
* The `Register` method in types marked with `[RegisterSerializer]` must accept a parameter of type `ISerializerRegistrar`.

So `[RegisterSerializer]` is rarely required. The one place it's required in our codebase is for the tests which register the same serializer for multiple different JSON types (`JObject`, `JArray`, etc) but that could just as easily be accomplished using an `IExternalSerializer` implementation. We could even remove `[RegisterSerializer]` altogether, but I'm not suggesting that yet.

Does this all sound fine?

Related to #467 

EDIT: oh, and the `[RegisterSerializer]` change is a good candidate for something to add to a code analyzer

ReubenBond:
This was implemented in #2611, which has been merged. Eventually the `Register` method can be removed, but it's simply marked as obsolete for now.

ReubenBond:
This is a 'breaking change', as the title says, but it's not a very breaking change and only affects a small number of users hand has a simple fix.

PR #2592, opened 2017-01-13
Non-static SerializationManager

ReubenBond:
Disclaimer: this is a work in progress. It needs work, more functional testing, and to be split into smaller, more mergable PRs. `SerialziationManager` is a core part of our system and `GrainReference` is serializable, so this PR is necessarily going to be large

Implements changes from #2591 & [#467](https://github.com/dotnet/orleans/issues/467#issuecomment-255074892) (with changes).

When using an external serializer (eg, JSON.NET, ProtoBuf), `GrainReference` instances must be bound via a call to `IGrainFactory.Bind(IAddressable)` before they can be used. When using the Orleans serialization framework, this happens automatically.

Once this is ready, we will be close to completing #467. Only some minor additional work is required to complete the task, basically changing the GrainClient class itself.

Note: this is a breaking change.

EDIT: to be clear, I don't expect this code to be reviewed in detail while I've got the work-in-progress tag on it. When it's ready, I'll remove the tag and update this PR.


xiazen:
great work @ReubenBond !

src/ClientGenerator/ClientGenerator.cs
```diff
@@ -105,15 +106,23 @@ private bool CreateGrainClient(CodeGenOptions options)
...
             }
 
+            var services = new ServiceCollection();
+            services.AddSingleton<AssemblyProcessor>();
+            services.AddSingleton(sp => sp);
+            services.AddSingleton<IMessagingConfiguration>(new GlobalConfiguration());
+            services.AddSingleton<RoslynCodeGenerator>();
+            services.AddSingleton<SerializationManager>();
+            var serviceProvider = services.BuildServiceProvider();
+            var codeGenerator = serviceProvider.GetRequiredService<RoslynCodeGenerator>();
```
jason-bragg: This seems like a quite peculiar way of constructing a RoslynCodeGenerator.  Would you comment on your reasoning?

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -79,23 +80,22 @@ internal class AssemblyLoader
             return discoveredAssemblyLocations;
         }
 #endif
-        public static T TryLoadAndCreateInstance<T>(string assemblyName, Logger logger) where T : class
+        public static T TryLoadAndCreateInstance<T>(string assemblyName, IServiceProvider serviceProvider, Logger logger) where T : class
```
jason-bragg: Suggestion - feel free to ignore, as this can be done later.
Suggest introducing a generic type factory interface and passing it in rather than the entire IServiceProvider.  Something like:

interface ITypeFactory
{
    T Create<T>(Type concreteType)  where T : class;
}

An impl of this that calls the activator utilities with the service provider can be passed in making this call path more specific and not depend on DI.

Again, this can be done as a minor refactor later..

src/Orleans/CodeGeneration/CodeGeneratorManager.cs
```diff
@@ -30,19 +30,26 @@ internal static class CodeGeneratorManager
...
         /// <summary>
         /// The code generator cache.
         /// </summary>
         private static ICodeGeneratorCache codeGeneratorCacheInstance;
 
+        private readonly IServiceProvider serviceProvider;
+
+        public CodeGeneratorManager(IServiceProvider serviceProvider)
+        {
+            this.serviceProvider = serviceProvider;
```
jason-bragg: ITypeFactory again.  In this case, it could also be registered in the service provider to allow it to be injected into the code generator manager

src/ClientGenerator/ClientGenerator.cs
attilah: Oh yeah, it is not what DI is for ;-) totally agreeing with @jason-bragg!

src/Orleans/Configuration/ClientConfiguration.cs
```diff
@@ -137,6 +138,11 @@ public enum GatewayProviderType
         public bool StatisticsWriteLogStatisticsToTable { get; set; }
         public StatisticsLevel StatisticsCollectionLevel { get; set; }
 
+        /// <summary>
+        /// Gets or sets the service configuration delegate.
+        /// </summary>
+        public Func<IServiceCollection, IServiceProvider> ServiceConfigurator { get; set; } = services => services.BuildServiceProvider();
```
jason-bragg: In a general sense this is functionality not configuration.  Does this belong here?

src/Orleans/Core/GrainClient.cs
```diff
@@ -327,7 +331,7 @@ public static Logger Logger
             get
             {
                 CheckInitialized();
-                return RuntimeClient.Current.AppLogger;
+                return ((IRuntimeClient)outsideRuntimeClient).AppLogger;
```
jason-bragg: Why is cast necessary here?

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -116,28 +128,39 @@ private GrainReference(GrainId grainId, string genericArgument, SiloAddress syst
...
-        internal static GrainReference NewObserverGrainReference(GrainId grainId, GuidId observerId)
+        internal static GrainReference NewObserverGrainReference(GrainId grainId, GuidId observerId, IRuntimeClient runtimeClient)
         {
-            return new GrainReference(grainId, null, null, observerId);
+            return new GrainReference(grainId, null, null, observerId, runtimeClient);
         }
 
         #endregion
 
+        public void Bind(IGrainFactory factory)
```
jason-bragg: I don't see where this is used?  Concerned that a grain reference read from storage may not be bound prior to being used in a grain call.

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -620,26 +646,26 @@ public static GrainReference FromKeyString(string key)
...
                 GuidId observerId = GuidId.FromParsableString(observerIdStr);
-                return NewObserverGrainReference(GrainId.FromParsableString(grainIdStr), observerId);
+                return NewObserverGrainReference(GrainId.FromParsableString(grainIdStr), observerId, null);
             }
             else if (systemTargetIndex >= 0)
             {
                 grainIdStr = trimmed.Substring(grainIdIndex, systemTargetIndex - grainIdIndex).Trim();
                 string systemTargetStr = trimmed.Substring(systemTargetIndex + (SYSTEM_TARGET_STR + "=").Length);
                 SiloAddress siloAddress = SiloAddress.FromParsableString(systemTargetStr);
-                return FromGrainId(GrainId.FromParsableString(grainIdStr), null, siloAddress);
```
jason-bragg: if these have null runtime, who ensures they get bound?

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -109,21 +96,45 @@ public IStreamProviderRuntime CurrentStreamProviderRuntime
...
+            services.AddSingleton<StreamProviderManager>();
+            services.AddFromExisting<IStreamProviderManager, StreamProviderManager>();
+            services.AddSingleton<CodeGeneratorManager>();
+
+            this.ServiceProvider = cfg.ServiceConfigurator(services);
+
+            this.InternalGrainFactory = this.ServiceProvider.GetRequiredService<IInternalGrainFactory>();
+            this.ClientStatistics = this.ServiceProvider.GetRequiredService<ClientStatisticsManager>();
+            this.SerializationManager = this.ServiceProvider.GetRequiredService<SerializationManager>();
+            this.messageFactory = this.ServiceProvider.GetService<MessageFactory>();
```
jason-bragg: As before, peculiar construction pattern.

src/ClientGenerator/ClientGenerator.cs
ReubenBond: What's peculiar about it?

jason-bragg: Seems peculiar to construct a container to construct a class rather than just constructing the class directly.

src/Orleans/Configuration/ClientConfiguration.cs
ReubenBond: I should remove it from this PR and we can talk about introducing it later. It's a good hook to have, and I consider it configuration since it's used only as part of the configuration phase and is the same signature as our Startup.ConfigureServices method (& same as ASP.NET), but it's not serializable.

src/Orleans/Core/GrainClient.cs
ReubenBond: OutsideRuntimeClient explicitly implements IRuntimeClient for AppLogger

src/Orleans/Runtime/GrainReference.cs
ReubenBond: I should remove this method and add it as an extension method in a separate PR. There's a `#warning` you can search for to see where the runtime client is set during deserialization. Currently, references to foreign clusters will always be deserialized as though they are references to the cluster whose SerializationManager is deserializing it.

src/Orleans/Runtime/GrainReference.cs
ReubenBond: There's a #warning on this method in my local branch - it needs to be fixed. Trying to use a grain ref which is not bound will throw a well-typed exception

src/Orleans/Runtime/OutsideRuntimeClient.cs
ReubenBond: I don't think so, consider that:
* We prefer to have just one instance of this class
* This class relies on some services in the container
* Some services in the container rely on this class

src/Orleans/Runtime/GrainReference.cs
ReubenBond: I still need to work out the best approach. The simplest is to add a 'GetGrainFromKeyString(string)` method to `IGrainFactory` or a new service and to move this method there.

src/ClientGenerator/ClientGenerator.cs
ReubenBond: There's some ugliness around this, by the way, since we don't link OrleansRuntime against the code generator and we have this `CodeGeneratorManager` type which tries to perform the linking at runtime

src/Orleans/Runtime/OutsideRuntimeClient.cs
jason-bragg: As with earlier comment, I'm not suggesting this is wrong or should be changed, just trying to understand the reasoning for this pattern.
In this case I can at least see that the service provider is kept around for future use, but it seems to only be needed for the StreamProviderManager.
Suggest container construction be separated and passed to the OutsideRuntimeClient constructor for testability reasons.  Can even be a static factory function if needed.

static IServiceProvider DefaultServiceProvider()
{
    var services = new ServiceCollection(ClientConfiguration cfg);
    ....
}

public OutsideRuntimeClient(IServiceProvider sp, ClientConfiguration cfg, bool secondary) 
{
    ...
}

public OutsideRuntimeClient(ClientConfiguration cfg, bool secondary = false)
    : this(DefaultServiceProvider(cfg), cfg, secondary)
{
    ...
}

ReubenBond: > Suggest container construction be separated and passed to the OutsideRuntimeClient constructor for testability reasons.

Yeah, I'm not entirely happy with the way the container is constructed today - either here or in Silo.cs (also, it's not consistent, why is it not constructed in `InsideRuntimeClient` if here it's in `OutsideRuntimeClient`, seems illogical, right?)

I see it as a post-PR topic, since we should change both and move towards a 'SiloBuilder'/'ClientBuilder' approach. I'm certainly not firm on that, but I'd like to have a separate discussion on what the root of our system is. Currently, I'd say it's the `IRuntimeClient` implementation, but `IRuntimeClient` is internal and so users cannot make use of it.

jason-bragg: Understood.  Huge PR, so I'm quite ok with pushing this to a later refactor.  Just wanted to make sure I understood why it was this way.  Having said that, the more of this that you can structure well in this pass, the more well formed the code will be if we don't get back to this immediately.  I'm not confident we've a good record of doing cleanup passes yet.

src/Orleans/Core/GrainExtensions.cs
```diff
@@ -62,6 +62,16 @@ public static TGrainInterface Cast<TGrainInterface>(this IAddressable grain)
             return RuntimeClient.Current.InternalGrainFactory.Cast<TGrainInterface>(grain);
         }
 
+        /// <summary>
+        /// Binds the grain reference to the provided <see cref="IGrainFactory"/>.
+        /// </summary>
+        /// <param name="grain">The grain reference.</param>
+        /// <param name="grainFactory">The grain factory.</param>
+        public static void BindGrainReference(this IAddressable grain, IGrainFactory grainFactory)
```
ReubenBond: This is a convenience method to allow grains to be bound to the grain factory. As you can see, it's trivial. Should it be removed?

ReubenBond:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot test this please

src/OrleansProviders/Streams/Memory/MemoryMessageBody.cs
```diff
@@ -29,16 +30,28 @@ public interface IMemoryMessageBodySerializer
     /// Default IMemoryMessageBodySerializer
     /// </summary>
     [Serializable]
-    public class DefaultMemoryMessageBodySerializer : IMemoryMessageBodySerializer
+    public class DefaultMemoryMessageBodySerializer : IMemoryMessageBodySerializer, IOnDeserialized
```
jason-bragg: After looking at this a bit more, I think a better solution would be for the [de]serialization calls in the IMemoryMessageBodySerializer to take a serialization manager as an argument, and for the GetEvents call in the IBatchContainer to also take a serialization manager.  The stream consumer extension getting the messages from the batch container would need to pass in a serialization manager.  This, IMO, is a more straight forward approach.

ReubenBond: @jason-bragg I'm happy with that approach. I didn't want to change the interface, since it implies that it's supposed to be serializer agnostic.

I don't want to have to add `IOnDeserialized` just for this.

jason-bragg: Understood.  I'd also prefer a solution that didn't alter the public surface, as this will be a breaking change, but I'm more open to breaking changes in the streaming system as that is still one of the newer systems which is, imo, still in flux.  We can also probably count the number of users that implemented custom IBatchContainers on one hand...

ReubenBond:
@dotnet-bot test this please

src/Orleans/AssemblyLoader/AssemblyLoader.cs
ReubenBond: I think we should leave this for later, as we discussed. I've removed all new abstractions from this PR so that it's more focused on the main goal.

ReubenBond:
I want to draw attention to some changes this PR introduces.

For example, these two methods are being added to a very important interface, `IGrainFactory`. What do people think of this?
```C#
/// <summary>
/// Binds the provided grain reference to this instance.
/// </summary>
/// <param name="grain">The grain reference.</param>
void BindGrainReference(IAddressable grain);

/// <summary>
/// Creates a grain reference from a storage key string.
/// </summary>
/// <param name="key">The key string.</param>
/// <returns>The newly created grain reference.</returns>
GrainReference GetGrainFromKeyString(string key);
```

The former allows a grain reference which was deserialized by an external serializer to be 'bound' to a particular instance of Orleans. I'm not so averse to this method, even if it seems to expose the sausage making process a little.

The latter replaces the currently static `GrainReference.FromKeyString(string key)` method. It's used in storage providers. I'm not entirely happy with it. Part of me feels that introducing a new interface, like `IGrainReferenceFactory` or some such.

I'd like to hear some opinions.

ReubenBond:
I added a new `IGrainReferenceConverter` interface specifically for `GetGrainFromStringKey`.

I consider this ready, assuming no one feels strongly about any of the decisions made here

ReubenBond:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot test this please

src/Orleans/Messaging/Message.cs
```diff
@@ -433,7 +433,7 @@ private Message(Categories type, Directions subtype)
         // Caller must clean up bytes
         public Message(List<ArraySegment<byte>> header)
         {
-            var context = new DeserializationContext
+            var context = new DeserializationContext(RuntimeClient.Current.InternalGrainFactory)
```
ReubenBond: Clearly this isn't the desired end-state, but these ctors are changed in the subsequent SerializationManager PR

src/Orleans/Runtime/GrainReference.cs
ReubenBond: This is resolved - no nulls.

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -687,6 +714,11 @@ protected GrainReference(SerializationInfo info, StreamingContext context)
             if (String.IsNullOrEmpty(genericArg))
                 genericArg = null;
             genericArguments = genericArg;
+
+#if !NETSTANDARD
+            var deserializationContext = context.Context as IDeserializationContext;
+            deserializationContext?.GrainFactory.BindGrainReference(this);
+#endif
```
ReubenBond: While netstandard1.5 has a StreamingContext type, it's a stub and serves no real purpose other than to exist

src/Orleans/Serialization/BinaryFormatterSerializer.cs
```diff
@@ -58,7 +61,10 @@ public void Serialize(object item, ISerializationContext context, Type expectedT
                 return;
             }
 
-            var formatter = new BinaryFormatter();
+            var formatter = new BinaryFormatter
+            {
+                Context = new StreamingContext(StreamingContextStates.All, context)
```
ReubenBond: to allow GrainReference inside of BinaryFormatter serialized objects, we pass the context. Technically it's not needed while serializing, but I want to be consistent.

src/Orleans/Serialization/DeserializationContext.cs
```diff
@@ -57,6 +63,9 @@ public void RecordObject(object obj)
         }
 
         /// <inheritdoc />
+        public IGrainFactory GrainFactory => this.grainFactory ?? (this.grainFactory =  RuntimeClient.Current?.InternalGrainFactory);
```
ReubenBond: This is removed in a subsequent PR

src/Orleans/Serialization/SerializationContext.cs
```diff
@@ -7,6 +7,11 @@ namespace Orleans.Serialization
     public interface ICopyContext
     {
         /// <summary>
+        /// Gets the serialization manager.
+        /// </summary>
+        SerializationManager SerializationManager { get; }
```
jdom: should you also add IServiceProvider to this and ISerializationContext, just to have symmetry with IDeserializionContext? I don't really care, but my OCD found this asymmetry a little odd.

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -110,64 +112,57 @@ public static object GetUninitializedObjectWithFormatterServices(Type type)
...
-            {
-                throw ex.Flatten();
-            }
         }
 
-        internal static void Initialize(List<TypeInfo> serializationProviders, TypeInfo fallbackType = null)
+        internal SerializationManager(IServiceProvider serviceProvider, List<TypeInfo> serializationProviders = null, TypeInfo fallbackType = null)
         {
+            this.serializationContext = new ThreadLocal<SerializationContext>(() => new SerializationContext(this));
+            this.deserializationContext = new ThreadLocal<DeserializationContext>(() => new DeserializationContext(this));
```
jdom: I'm not *too* worried that we are passing a reference to `this` in this constructor (although it could also be done in a factory method, but it's OK regardless), but could we at least make this assignment at the end of the constructor, so that we make sure nobody uses the field before the object is done constructing.

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -604,13 +599,13 @@ internal static void FindSerializationInfo(Type type)
                                     typeInfo.Name,
                                     assembly.GetName().Name);
 
-                            var register = typeInfo.GetMethod("Register", Type.EmptyTypes);
+                            var register = typeInfo.GetMethod("Register", new[] {typeof(SerializationManager)});
                             if (register != null)
```
jdom: Since you made this change to the expected signature (breaking change, please make a note in the changelog!), you need to also update the Warning message you output when `register` is null.

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -924,26 +919,27 @@ internal static DeepCopier GetCopier(Type t)
...
+            var context = this.serializationContext.Value;
+            context.Reset();
+            
             Stopwatch timer = null;
             if (StatisticsCollector.CollectSerializationStats)
             {
                 timer = new Stopwatch();
                 timer.Start();
-                Copies.Increment();
+                context.SerializationManager.Copies.Increment();
```
jdom: accessing the SerializationManager instance from the `context` is a little odd from here. Could it be possible that they differ? If it's not possible, then maybe just assert it in the beginning and then use `this` instead?

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -2215,6 +2215,14 @@ private static void RegisterSerializationProviders(List<TypeInfo> providerTypes)
                 });
         }
 
+        public void Dispose()
+        {
+            IDisposable disposable = this.serializationContext;
+            if (disposable != null) disposable.Dispose();
+            disposable = this.deserializationContext;
+            if (disposable != null) disposable.Dispose();
```
jdom: if you are scared about race conditions, just do:
```
this.serializationContext?.Dispose();
this.deserializationContext?.Dispose();
```

test/DefaultCluster.Tests/GrainReferenceTest.cs
```diff
@@ -113,6 +113,8 @@ private void TestGrainReferenceSerialization(int id, bool resolveBeforeSerialize
             
             this.GrainFactory.BindGrainReference(other);
 
+            this.GrainFactory.BindGrainReference(other);
+
```
jdom: merge conflict (duplicate Bind)

src/Orleans/Streams/Internal/StreamConsumerExtension.cs
```diff
@@ -5,13 +5,14 @@
...
 using Orleans.Runtime;
+using Orleans.Serialization;
 
 namespace Orleans.Streams
 {
     internal interface IStreamSubscriptionHandle
     {
         Task<StreamHandshakeToken> DeliverItem(object item, StreamSequenceToken currentToken, StreamHandshakeToken handshakeToken);
-        Task<StreamHandshakeToken> DeliverBatch(IBatchContainer item, StreamHandshakeToken handshakeToken);
+        Task<StreamHandshakeToken> DeliverBatch(IBatchContainer item, StreamHandshakeToken handshakeToken, SerializationManager serializationManager);
```
jdom: It's a little annoying that this and `IBatchContainer` take a dependency on SerializationManager, where *I believe* for most implementors this is unused, but it's still a dependency that flows through every stream provider, factory, adaptors, caches, etc (although to be fair, I scanned them pretty fast, so I'm not sure if there's a better alternative).
I suppose that as we use DI more and more, we can probably get rid of these dependencies being injected in methods, and only add them the the relevant implementations that do need them. Nevertheless, even if temporarily, it strikes me as this is a big change that touches a lot of files, but we are already talking that the streaming infrastructure shouldn't have used SerializationManager, and that we should start endorsing a more version tolerant serializer (as this **is** persisted, even if for little time in some cases). Anyway, just pointing out, in case you can think of a better alternative

ReubenBond: I'll take another look at it - I don't like it either, but it was discussed with Jason at the time. The core issue is that many of these types are `[Serializable]` and some of the important generic types have the `new()` constraint, eg:
```C#
public class MemoryStreamProvider<TSerializer> : PersistentStreamProvider<MemoryAdapterFactory<TSerializer>>
    where TSerializer : IMemoryMessageBodySerializer, new()
```

src/Orleans/Messaging/IncomingMessageBuffer.cs
```diff
@@ -154,7 +161,11 @@ public bool TryDecodeMessage(out Message msg)
             List<ArraySegment<byte>> body = ByteArrayBuilder.BuildSegmentListWithLengthLimit(readBuffer, bodyOffset, bodyLength);
             
             // build message
-            msg = new Message(header);
+            var context = new DeserializationContext(this.serializationManager)
```
sergeybykov: These are not new allocations, just lifted up from inside constructor of `Message`, right? If that's the case then it's not a concern for this change. Should we consider (outside of this PR) using an object pool instead, as this is a hot path?

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -36,13 +38,7 @@ internal class PersistentStreamPullingAgent : SystemTarget, IPersistentStreamPul
...
         private string StatisticUniquePostfix => streamProviderName + "." + QueueId;
 
-        internal PersistentStreamPullingAgent(
-            GrainId id, 
-            string strProviderName,
-            IStreamProviderRuntime runtime,
-            IStreamPubSub streamPubSub,
-            QueueId queueId,
-            PersistentStreamProviderConfig config)
+        internal PersistentStreamPullingAgent(GrainId id, string strProviderName, IStreamProviderRuntime runtime, IStreamPubSub streamPubSub, QueueId queueId, PersistentStreamProviderConfig config, SerializationManager serializationManager)
```
sergeybykov: Nit: the multi-line version was more readable.

src/Orleans/Streams/Internal/StreamConsumerExtension.cs
jdom: hmmm, but that's promising, if there's only a few places where the serialization manager is used, then perhaps we can drop the `new()` constraint on them, and use `ActivatorUtilities` to new them up, and only the ones that actually care about SerializationManager (which we currently have, but expect to get rid at some point) can actually get injected with it in their constructors...
It's probably harder than that in practice, but something to explore...

src/Orleans/Messaging/IncomingMessageBuffer.cs
ReubenBond: These are new. I'll see if I can allocate these once per message buffer - depending on what the access patterns are like.

Agree that we should consider an object pool.

ReubenBond: We can reduce the allocations below the number of allocations in the existing code with a little bit of reorganization.

src/Orleans/Streams/Internal/StreamConsumerExtension.cs
```diff
@@ -31,14 +32,16 @@ internal class StreamConsumerExtension : IStreamConsumerExtension
         private readonly ConcurrentDictionary<GuidId, IStreamSubscriptionHandle> allStreamObservers; // map to different ObserversCollection<T> of different Ts.
         private readonly Logger logger;
         private readonly bool isRewindable;
+        private readonly IRuntimeClient runtimeClient;
```
jdom: I dont see any usages of this. Is it just leftover from previous iterations?

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -19,6 +20,7 @@ internal class PersistentStreamPullingAgent : SystemTarget, IPersistentStreamPul
         private readonly Dictionary<StreamId, StreamConsumerCollection> pubSubCache;
         private readonly SafeRandom safeRandom;
         private readonly PersistentStreamProviderConfig config;
+        private readonly SerializationManager serializationManager;
         private readonly Logger logger;
```
jdom: This is no longer relevant either, correct?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubBatchContainer.cs
```diff
@@ -116,5 +125,28 @@ internal static EventData ToEventData<T>(Guid streamGuid, String streamNamespace
             }
             return eventData;
         }
+
+        [SerializerMethod]
+        private static void Serialize(object obj, ISerializationContext context, Type expected)
+        {
+            var input = (EventHubBatchContainer) obj;
+            SerializationManager.SerializeInner(input.eventHubMessage, context, typeof(EventHubMessage));
+        }
```
jdom: shouldn't you be using the SM from the context?

src/OrleansAzureUtils/Providers/Streams/AzureQueue/IAzureQueueDataAdapter.cs
```diff
@@ -17,12 +17,12 @@ public interface IAzureQueueDataAdapter
         /// <summary>
         /// Creates a cloud queue message from stream event data.
         /// </summary>
-        CloudQueueMessage ToCloudQueueMessage<T>(Guid streamGuid, string streamNamespace, IEnumerable<T> events, Dictionary<string, object> requestContext);
+        CloudQueueMessage ToCloudQueueMessage<T>(SerializationManager serializationManager, Guid streamGuid, string streamNamespace, IEnumerable<T> events, Dictionary<string, object> requestContext);
```
jdom: Similar to what you did with other streaming parts, can this be avoided and instead make it a constructor dependency of the concrete instances (and also remove the new() constraint)?

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
```diff
@@ -9,10 +7,10 @@
 
 namespace Orleans.Providers
 {
-    [Serializable]
     internal class MemoryBatchContainer<TSerializer> : IBatchContainer
-        where TSerializer : IMemoryMessageBodySerializer, new()
+        where TSerializer : IMemoryMessageBodySerializer
     {
+        private readonly IMemoryMessageBodySerializer serializer;
```
jdom: you dropped the  [NonSerialized] attribute. Do you need to add it back or was it by design? Would this cause upgrade issues?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubBatchContainer.cs
ReubenBond: This does do that: SerializeInner method is currently static - it takes a context

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
ReubenBond: I kept it because that type will serialize itself correctly now - if using our serializer

ReubenBond: I think it's generally bad to serialize services, though.

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
```diff
@@ -9,10 +7,10 @@
 
 namespace Orleans.Providers
 {
-    [Serializable]
     internal class MemoryBatchContainer<TSerializer> : IBatchContainer
```
jason-bragg: Unclear how this will work?  Batch containers are the way we send sets of events to consumers.  If these are not serializable, how will they be sent?
If they are serialized, how is the serializer assigned?

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
jdom: hmmm, this can be a little tricky... I missed that... supplying your own serializer might not be straightforward. Still I like it a little bit better than supplying SM on every call, but in that case not sure if I like it better than supplying IServiceProvider on every call, or at the very least have an explicit interface for binding it after deserializing (bind to an SP, not to SM). it can be that this interface is optional, and the binding call happens only if the serializer implements it

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
jason-bragg: Julian explained what's going on here.
This solution is -ok- but imo worse than the previous approach.
The problem is that all IBatchContainers, and all interfaces included in the stream adapter pattern, are application extension points.  They are how developers write new persistent stream providers.  In the previous approach, we had to change quite a few internal classes to get the serialization manager to the batch container, but those were all our code.  In this pattern we don't have to do that (hence less code changes) but the application layer code needs to be more complicated as it needs to figure out some trick to getting the serialization manager, like we've done in DefaultMemoryMessageBodySerializer.  We've not only made the serialization manager no longer global, we've made it the application layer's problem to figure out how to get it, instead of passing it to systems that need it.

ReubenBond: I'll add the [Serializable] attribute back. Note that [Serializable] is an attribute which hints to us that a serializer should be generated for some type, but also gives permission to BinaryFormatter to serialize that type. I want to avoid the latter wherever possible, since we don't have enough control over it so that we can inject context in cases like this - where we're serializing services instead of just POCOs

> We've not only made the serialization manager no longer global, we've made it the application layer's problem to figure out how to get it, instead of passing it to systems that need it.

The answer is that they should be constructed by DI. The only types which should be instantiated via `new` are POCOs and types which aren't shared (via serialization or otherwise)

ReubenBond: BTW, the issue I'm having now is precisely surrounding that context - there's no single public entry point into our system that I can pass to a type during deserialization.

What I'd want is some kind of `IOrleansRuntime` which has an `IGrainFactory`, `SerializationManager`, and probably `IServiceProvider`. I added an `IRuntimeClient` property to `SerializationManager` as an entry point into the system, but it's ugly (and a circular reference)

src/OrleansAzureUtils/Providers/Streams/AzureQueue/IAzureQueueDataAdapter.cs
ReubenBond: Will do

ReubenBond:
Please note the newly introduced type, `ISerializerContext` (distinct from `ISerializationContext`), I'm happy to change the name.

The purpose of it is to provide context to foreign serializers so that they can continue to offer a high-fidelity experience.

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
jdom: to sum up a very long discussion we had in private chat, I think this extensibility point is too hacky, since it requires the *end user* to implement the hack if he wants to provide a different IMemoryMessageBodySerializer type.
IMO, we could have something that we call after MemoryBatchContainer is deserialized that augments it with the IServiceProvider (either [OnDeserialized] if it reliably works or something we explicitly call, since we control the usage of this class) and the it can use ActivatorUtilities to new up the serializer.
Alternatively, I agree that the previous approach that Jason suggested is a little more explicit on how to extend it, although I would pass IServiceProvider instead of SerializationManager, so alternative implementations can avoid using SerializationManager, which was probably a bad choice for the default serialization mechanism.

ReubenBond: I opted to add a new `IOnDeserialized` interface, since it's simpler and cleaner than adding an attribute. It's in the latest revision with a couple of tests for the inbuilt serializers.

jdom:
>                                     "Type {0} from assembly {1} has the RegisterSerializer attribute but no public static void Register method",

I'm not sure if I had a comment on a previous PR or it was here and got lost, but this exception message needs an update, especially since it's a sneaky breaking change that would not break at compile time.

---
Refers to: src/Orleans/Serialization/SerializationManager.cs:634 in 975bdbb. [](commit_id = 975bdbb4fbfb5e7d8504df67bd6880a16803bbc2, deletion_comment = False)

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
```diff
@@ -60,5 +57,10 @@ public bool ShouldDeliver(IStreamIdentity stream, object filterData, StreamFilte
         {
             return true;
         }
+
+        void IOnDeserialized.OnDeserialized(ISerializerContext context)
+        {
+            this.serializer = ActivatorUtilities.GetServiceOrCreateInstance<TSerializer>(context.ServiceProvider);
```
jdom: Awesome, this is what I was advocating for. Nevertheless, for performance reasons, I think there is a method in ActivatorUtilities that allows you to create an object factory and cache it, and avoid reflection in subsequent calls, since this is going to be called extremely frequently for these same types

src/OrleansProviders/Streams/Memory/MemoryMessageBody.cs
```diff
@@ -29,26 +30,35 @@ public interface IMemoryMessageBodySerializer
...
+        /// <inheritdoc />
         public MemoryMessageBody Deserialize(ArraySegment<byte> bodyBytes)
         {
-            return SerializationManager.DeserializeFromByteArray<MemoryMessageBody>(bodyBytes.ToArray());
+            return serializationManager.DeserializeFromByteArray<MemoryMessageBody>(bodyBytes.ToArray());
+        }
+
+        void IOnDeserialized.OnDeserialized(ISerializerContext context)
+        {
+            this.serializationManager = context.SerializationManager;
```
jdom: correct me if I'm wrong, but with the latest, this class is never (de)serialized, so the IOnDeserialized interface is not needed here. It will always be constructed using ActivatorUtilities. In theory you could also drop the [Serializable] attribute from it (it wasn't being used before either). Unless you added this in case someone was *incorrectly* using it and serializing it for a different purpose?

jdom:

src/OrleansProviders/Streams/Memory/MemoryMessageBody.cs
ReubenBond: > Unless you added this in case someone was incorrectly using it and serializing it for a different purpose?

precisely :)

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
ReubenBond: The updated method would look like this:
```C#
this.serializer = context.ServiceProvider.GetService<TSerializer>() ?? (TSerializer)ObjectFactory(context.ServiceProvider, null);
```
Where ObjectFactory is
```C#
private static readonly ObjectFactory ObjectFactory = ActivatorUtilities.CreateFactory(
    typeof(TSerializer),
    null);
```

ReubenBond:
How does this sound for an updated error message?
```C#
"Type {0} from assembly {1} has the RegisterSerializer attribute but no public static void Register(SerializationManager sm) method."
```

jdom:
Message looks good. We *might* want to assert if the old signature was used instead, and provide a more informative message, but I'm fine not doing this extra work.

ReubenBond:
I think we should make Roslyn Analyzers for these cases - it's a much better experience

src/OrleansProviders/Streams/Memory/MemoryBatchContainer.cs
jdom: sounds good, although I'd make this field Lazy<ObjectFactory> so that exceptions flow to the right place, and not on type initialization.

jdom: BTW, I'd call this field serializerFactory, not ObjectFactory.

ReubenBond: Sure thing

src/OrleansProviders/Streams/Memory/MemoryAdapterFactory.cs
```diff
@@ -79,6 +72,7 @@ public void Init(IProviderConfiguration providerConfig, string name, Logger log,
 
             // 10 meg buffer pool.  10 1 meg blocks
             bufferPool = new FixedSizeObjectPool<FixedSizeBuffer>(adapterConfig.CacheSizeMb, () => new FixedSizeBuffer(1 << 20));
+            this.serializer = ActivatorUtilities.CreateInstance<TSerializer>(this.serviceProvider);
```
jdom: hmmm, this is creating the serializer differently than in MemoryBatchContainer...

ReubenBond: Will consolidate

src/OrleansProviders/Streams/Memory/MemoryAdapterFactory.cs
```diff
@@ -79,6 +77,9 @@ public void Init(IProviderConfiguration providerConfig, string name, Logger log,
 
             // 10 meg buffer pool.  10 1 meg blocks
             bufferPool = new FixedSizeObjectPool<FixedSizeBuffer>(adapterConfig.CacheSizeMb, () => new FixedSizeBuffer(1 << 20));
+
+            this.serializer = this.serviceProvider.GetService<TSerializer>() ??
+                              (TSerializer)ObjectFactory.Value(this.serviceProvider, null);
```
jdom: nit: I'd avoid duplication (which should always match) by extracting this to an internal static CreateSerializer(IServiceProvider) method, and then the MemoryBatchContainer<TSerializer> just calls MemoryAdapterFactory<TSerializer>.CreateSerializer

ReubenBond: I figured duplication was cheaper, but I'll change it.

ReubenBond: Does the latest commit look like what you're after?

jason-bragg:
LG2M

src/OrleansProviders/Streams/Memory/MemoryAdapterFactory.cs
jdom: absolutely, thanks

jdom:
Woohooo!! this is in! 🎉  Thanks @ReubenBond for bearing with us! :)

ReubenBond:
I'm so happy :D Thanks for putting up with it - it was a long road and a big PR, but this lays the groundwork for more modularity as well as #467

attilah:
Shipit! :shipit:

sergeybykov:
This was yet another epic thing by Reuben!

PR #2594, opened 2017-01-15
Fix event sourcing project on vNext solution

galvesribeiro:
- Moved the event sourcing project to the right path on vNext solution.
- Added missing .nuspec for vNext
- Added missing test


attilah:
LGTM, thanks @galvesribeiro!

Issue #1600, opened 2016-03-22
The Unit-Testing-Grains.md file uses MSTest, and doesn't mention any other test framework.


moswald:
While MSTest isn't _deprecated_, it's usage isn't exactly in vogue at this time. Most new code seems to be tested with xUnit 2.

A preferred step-by-step document would use xUnit in the example code, and it would at least mention that other frameworks such as NUnit are possible.

sergeybykov:
@moswald Good point. Would you like to take a stab at updating https://github.com/dotnet/orleans/blob/gh-pages/Step-by-step-Tutorials/Unit-Testing-Grains.md?

mikkelbu:
Shouldn't this issue be closed, since #2595 was merged?

moswald:
Oh, yep. Jeez, I filed this awhile ago and then promptly forgot about following up. I have a private branch that I guess I don't need anymore. 😅

PR #2595, opened 2017-01-15
Update Unit-Testing-Grains to feature XUnit

robrodi:
Quick fix for #1600.  Samples are now ```Fact```s rather than ```TestClass``` and ```TestMethod```.  Also renamed the test method to more closely follow the xUnit naming conventions


robrodi:
Btw, in the future, this class should inherit from ```TestCluster```. rather than ```TestingSiloHost```.

attilah:
LGTM, thanks @robrodi!

PR #2596, opened 2017-01-15
Serialization/CodeGen: Avoid generating Register methods

ReubenBond:
* When we encounter a type, eg `SerializerType`, with a `[Serializer(typeof(TargetType))]` attribute, we register the type as the serializer for `TargetType`. I.e, `SerializationManager.Register(typeof(TargetType), typeof(SerializerType))`
* Avoid generating `Register` methods
* Avoid adding `[RegisterSerializer]` attribute
* Avoid generating non-generic "master registerer" class for generic types

This code was extracted from #2592

This is a non-breaking change

### DIFF
--- a/src/Orleans/Serialization/SerializationManager.cs
+++ b/src/Orleans/Serialization/SerializationManager.cs
@@ -573,9 +573,15 @@ namespace Orleans.Serialization
 
             try
             {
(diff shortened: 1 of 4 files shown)


attilah:
LGTM.

Issue #2600, opened 2017-01-17
Exception trying to send an exception response


galvesribeiro:
In 2.0 TP, whenever an exception happens and must be serialized to send to the remote caller this exception happen:

```

Value cannot be null.
Parameter name: meth
Exception = System.ArgumentNullException: Value cannot be null.
Parameter name: meth
   at System.Reflection.Emit.DynamicILGenerator.Emit(OpCode opcode, MethodInfo meth)
   at Orleans.Runtime.InsideRuntimeClient.CreateExceptionPrepForRemotingMethod()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at Orleans.Runtime.InsideRuntimeClient.PrepareForRemoting(Exception exception)
   at Orleans.Runtime.InsideRuntimeClient.SafeSendExceptionResponse(Message message, Exception ex)
```
And
```
[2017-01-16 21:27:44.610 GMT    11      WARNING 101532  InsideRuntimeClient     192.168.1.80:22222] Exception trying to send an exception response: Value cannot be null.
```

Reuben found out that this is the offending code:

```
var methodInfo = typeof(Exception).GetMethod(
                "PrepForRemoting",
                BindingFlags.Instance | BindingFlags.NonPublic);
```

Looks like `Exception.PrepForRemoting` is not available in .Net Core.

cc: @ReubenBond

ReubenBond:
Note that this is on .NET Core. Does this happen using 2.0TP on Full .NET, too, @galvesribeiro?

galvesribeiro:
Just confirmed, full framework has it, .Net Core doesn't.

galvesribeiro:
Implemented on #2601

PR #2601, opened 2017-01-17
Small fix on PrepareToRemote #2600

galvesribeiro:
Fix for #2600

### DIFF
--- a/src/OrleansRuntime/Core/InsideRuntimeClient.cs
+++ b/src/OrleansRuntime/Core/InsideRuntimeClient.cs
@@ -77,7 +77,7 @@ namespace Orleans.Runtime
         public Catalog Catalog { get; private set; }
 
         public SiloAddress MySilo { get; private set; }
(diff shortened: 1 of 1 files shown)


src/OrleansRuntime/Core/InsideRuntimeClient.cs
```diff
@@ -473,6 +473,10 @@ private void SafeSendResponse(Message message, object resultObject)
             var methodInfo = typeof(Exception).GetMethod(
                 "PrepForRemoting",
                 BindingFlags.Instance | BindingFlags.NonPublic);
+
+            if (methodInfo == null)
+                return new Func<Exception, Exception>(exc => exc);
```
ReubenBond: We generally use cast delegates instead of using `new`.
Do you mind removing adding a comment here?

galvesribeiro:
@ReubenBond The changes were made.

PR #2604, opened 2017-01-17
add new test assembly to test.cmd

xiazen:
add missing App.config


ReubenBond:
Looks good. If BVTs pass then we can merge. Thanks

PR #2608, opened 2017-01-18
Fix building in VS2017 RC2.

attilah:
Apparently when building in VS things are still different, so building the projects one-by-one was re-added to proj file.

### DIFF
--- a/vNext/src/ClientGenerator/BootstrapCodegen.proj
+++ b/vNext/src/ClientGenerator/BootstrapCodegen.proj
@@ -1,39 +1,79 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />
-    <PropertyGroup>
-      <ExcludeCodegenConstants>$(DefineConstants);EXCLUDE_CODEGEN;ORLEANS_BOOTSTRAP;NETSTANDARD;NETSTANDARD_TODO</ExcludeCodegenConstants>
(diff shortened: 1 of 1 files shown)


benjaminpetit:
Tested on my machine and it works!

vNext/src/ClientGenerator/BootstrapCodegen.proj
```diff
@@ -1,39 +1,79 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />
```
galvesribeiro: You are using a very old SDK... There is no need for those imports anymore.

In fact, if you mix the projects, VS and MSBuild will load assemblies in a wrong way which will mess everything...

attilah: That did not change, for a reason. ToolsVersion is set to 15.0 to make sure latest tools are used. For CSProj, VBProj, etc you're right, that's what I changed when moved to new tooling.

I've a some thoughts about how to change the bootstrapping but removing commons import and having an Sdk reference on the project node  is now working in itself.

galvesribeiro: Hummm... I'm having a very hard time to keep old and need project definitions on the same solution. MSBuild seems doesn't like it. I've discussing with the SDK team and the only workaround was upgrade everything or downgrade everything. Anyway, was just a though.

attilah: There is the key you're saying: "same solution"...this does not have to do anything with that.

Issue #2477, opened 2016-12-03
EventHubStreamProviders exceptions


jason-bragg:
Running into these exceptions in some production services using the EventHubStreamProvider.  These exceptions seem to only occur when pulling agents are being shutdown, so they are transient and have no application impact, but should still be addressed.

System.NullReferenceException: Object reference not set to an instance of an object.
   at Orleans.ServiceBus.Providers.EventHubAdapterReceiver.<GetQueueMessagesAsync>d__19.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Orleans.Streams.PersistentStreamPullingAgent.<ReadFromQueue>d__35.MoveNext()

UnobservedExceptionsHandlerClass Caught an UnobservedTaskException event sent by [Id=3202, Status=Faulted]. Exception = 
Exc level 0: System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread.
Exc level 1: Orleans.Runtime.OrleansException: SystemTarget PullingAgentSystemTarget/FF/f0f62a85 not active on this silo. Msg=Response S100.117.42.37:11111:218312555*grn/47A87F83/40f3ae87@f124a5a1->S100.117.42.37:11111:218312555PullingAgentSystemTarget/FF/f0f62a85@Sf0f62a85 #5336990: global::Orleans.Streams.IStreamConsumerExtension:GetSequenceToken()

sergeybykov:
Resolved via #2612.

PR #2612, opened 2017-01-19
Harden persistent stream provider pulling agent.

jason-bragg:
-Reduced duplicate logging during errors.
-Fixed null reference exception that occur durring shutdown.

Addresses "EventHubStreamProviders exceptions #2477"


ReubenBond:
LGTM once the conflict with master is resolved

PR #2613, opened 2017-01-19
BugFix: EventHubQueueCache failing to write checkpoints on purge.

jason-bragg:
When logging was added to EventHubQueueCache purge the base class purge function, which triggers checkpoint logic, was not called.  This broke periodic checkpointing.


sergeybykov:
@dotnet-bot test this please.

PR #2621, opened 2017-01-21
Removed message batching configuration settings.

jason-bragg:


ReubenBond:
Did this setting only exist for pub/sub?

jason-bragg:
No, these here a network settings that enabled batching in the low-level networking library.

This batching was prototyped years back, but didn't show performance gains, so was never really used.
When I optimized the networking layer about a year back, I added batched read on the reading side.  This was enabled all the time (no config switch) and was not compatible with the original batching code.  

It was not clear if we’d remove the original batching code or fix it.  Eventually we removed it, but missed the config settings.

I noticed these settings being used in tests, which clued me into the fact that we’d missed them.  This hurt my feelings.  This PR makes me feel better. 😊

ReubenBond:
Great :) Thanks for clearing it up. I was thinking about batching again yesterday, since I'm playing with the Raft implementation & want a way to efficiently host many little concurrent replica sets in a cluster

jason-bragg:
I want to batch our sends as well, but I also want to replace our networking layer.  My dreams don't get along.

ReubenBond:
I think we can make them work, but a new networking layer can be responsible for batching, or it can be a component in the networking pipeline

PR #2622, opened 2017-01-21
Removed SimpleQueueCache dependecy on EventSequenceToken

jason-bragg:
SimpleQueueCache has a hard dependency on EventSequenceToken.  This is bad, as it's intended to be a generic cache and should not be dependent on specific batch containers or sequence tokens.

The original compromise was made to manage cursor location in the cache, which was originally, unset, set, or idle.  To help manage cursor position for idle cursors, we'd needed to know the sequence token of the next message.  We used the state of the EventSequenceToken to get the next sequence token, which introduced the dependency.

This change treats unset and idle cursors the same, with a special case for cursors set in the future.

### DIFF
--- a/src/OrleansProviders/Streams/Common/SimpleCache/SimpleQueueCache.cs
+++ b/src/OrleansProviders/Streams/Common/SimpleCache/SimpleQueueCache.cs
@@ -8,7 +8,7 @@ namespace Orleans.Providers.Streams.Common
     internal class CacheBucket
     {
         // For backpressure detection we maintain a histogram of 10 buckets.
-        // Every buckets recors how many items are in the cache in that bucket
+        // Every bucket records how many items are in the cache in that bucket
         // and how many cursors are poinmting to an item in that bucket.
         // We update the NumCurrentItems when we add and remove cache item (potentially opening or removing a bucket)
         // We update NumCurrentCursors every time we move a cursor
@@ -40,7 +40,6 @@ namespace Orleans.Providers.Streams.Common
     public class SimpleQueueCache : IQueueCache
     {
         private readonly LinkedList<SimpleQueueCacheItem> cachedMessages;
-        private StreamSequenceToken lastSequenceTokenAddedToCache;
         private readonly int maxCacheSize;
         private readonly Logger logger;
         private readonly List<CacheBucket> cacheCursorHistogram; // for backpressure detection
@@ -109,6 +108,7 @@ namespace Orleans.Providers.Streams.Common
                 cacheCursorHistogram.RemoveAt(0); // remove the last bucket
             }
             purgedItems = allItems;
+            Log(logger, "TryPurgeFromCache: purged {0} items from cache.", purgedItems.Count);
             return true;
         }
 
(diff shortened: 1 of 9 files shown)


jason-bragg:
@gabikliot may also want to ponder this, as the original compromise never set well with him, rightfully so.

ReubenBond:
Certainly, and he has much more insight into these APIs than I do. I only assigned myself because I was rebooting my PC (audio issues) and didn't want to lose track :P

src/OrleansProviders/Streams/Common/SimpleCache/SimpleQueueCacheCursor.cs
```diff
@@ -24,16 +24,16 @@ public class SimpleQueueCacheCursor : IQueueCacheCursor
 
         internal bool IsSet => Element != null;
 
-        internal void Reset(StreamSequenceToken token)
+        internal void Set(LinkedListNode<SimpleQueueCacheItem> item)
         {
-            Element = null;
-            SequenceToken = token;
+            Element = item;
+            SequenceToken = item?.Value.SequenceToken;
```
gabikliot: why would item ever be null? Its better to have the contract more narrow (set on non null value, reset on null value) instead of not well specified interface.
If you are unsure, add NullArgException.

jason-bragg: Good catch.  Null is not valid in this case.

jason-bragg:
Feedback and bug found with further testing addressed.

src/OrleansProviders/Streams/Common/SimpleCache/SimpleQueueCache.cs
```diff
@@ -244,40 +242,24 @@ internal bool TryGetNextMessage(SimpleQueueCacheCursor cursor, out IBatchContain
...
             batch = cursor.Element.Value.Batch;
-            
-            // Advance to next:
+
+            // If we are at the end of the cache unset cursor and move offset one forward
             if (cursor.Element == cachedMessages.First)
             {
-                // If we are at the end of the cache unset cursor and move offset one forward
-                ResetCursor(cursor, ((EventSequenceToken)cursor.SequenceToken).NextSequenceNumber());
+                UnsetCursor(cursor, null);
```
ReubenBond: Should this be null? It seems like the code accommodates for a null sequence token, treating it as a default, but I'm not 100% sure.

jason-bragg: Yes.  This is an important part of the change.  We used to always need a sequence token, either of the next message we expect or of some future sequence token we expect to deliver from.  This required that we be able to compute the next sequence token given the current sequence token (the NextSequenceNumber call) which is not functionality supported by sequence tokens.

Now we only expect to have a sequence token if the cursor is currently active or if we're waiting for a future event.  If an active cursor reaches the most recent item in the cache, and has no more items to deliver, it becomes unset and waits for more data, which is signaled by the Refresh call which now takes a new sequence token to start from (rather than starting from the computed one originally left in the cursor).

ReubenBond: That makes sense, thanks for the explanation

ReubenBond:
LGTM. I'll give Gabi a chance to have another look at it

gabikliot:
Looks good @jason-bragg !

Issue #2551, opened 2016-12-28
Access to grain client connection status


mhertis:
> If a client loses connection to the cluster, all of the existing producers and consumers on that client will be removed from the pubsub system, even if the client does not unsubscribe. @jason-bragg 

For this reason we need to know client connection status, some kind of a signal/event on connection change (i.e. is lost, reestablished, ..) so that client can resubscribe to the stream.

There was also short conversation on this topic on [gitter](https://gitter.im/dotnet/orleans?at=5858397bc895451b75f77b42).

sergeybykov:
@benjaminpetit, can we easily do something like this?

benjaminpetit:
Sorry for the delay, I am looking into it

benjaminpetit:
We could expose public events when a gateway connect/disconnect, and keep a counter of opened connections. This way, when the counter go back to 0, we know that we are no longer connected to the cluster.

It seems that using `NetworkingStatisticsGroup` we already have this mechanism: when a gateway connects, it calls `NetworkingStatisticsGroup.OnOpenedGatewayDuplexSocket()`` and `NetworkingStatisticsGroup.OnClosedGatewayDuplexSocket` when the gateway is disconnected.

Should we refactor `NetworkingStatisticsGroup` to fit our need or should we add propoer event in the `GrainClient`?

jason-bragg:
some background discussion on this can be found in issue 1421. https://github.com/dotnet/orleans/issues/1421

sergeybykov:
>Should we refactor NetworkingStatisticsGroup to fit our need or should we add propoer event in the GrainClient?

I think we need to add support for an optional callback that we'd invoke upon client disconnect.

sergeybykov:
Resolved via #2628.

PR #2628, opened 2017-01-23
Client cluster disconnection

benjaminpetit:
Fix for "Access to grain client connection status" #2551 

Now `ProxiedMessageCenter` keeps a counter of opened gateway connection. When this counter reach 0, we send an event to the client to notify that the client might have been disconnected from the cluster.

I think we still can have race condition like this:
- Client connected to GW1 (counter set to 1)
- GW1 connection lost (but no update to the counter yet)
- GW2 connection established (but no update to the counter yet)
- GW2 updates the opened connection counter (counter set to 2)
- GW1 updates the opened connection counter (counter set to 1)

But the windows seems small enough for me, so this fix seems still useful for now

### DIFF
--- a/src/Orleans/Core/GrainClient.cs
+++ b/src/Orleans/Core/GrainClient.cs
@@ -381,6 +381,15 @@ namespace Orleans
             return RuntimeClient.Current.CurrentStreamProviderManager.GetProvider(name) as Streams.IStreamProvider;
         }
 
+        public delegate void ConnectionToClusterLostHandler(object sender, EventArgs e);
+
+        public static event ConnectionToClusterLostHandler ClusterConnectionLost;
+
+        internal static void NotifyClusterConnectionLost()
+        {
+            ClusterConnectionLost?.Invoke(null, null);
+        }
+
         internal static IList<Uri> Gateways
         {
             get
--- a/src/Orleans/Messaging/ProxiedMessageCenter.cs
+++ b/src/Orleans/Messaging/ProxiedMessageCenter.cs
@@ -83,6 +83,7 @@ namespace Orleans.Messaging
         public SiloAddress MyAddress { get; private set; }
         public IMessagingConfiguration MessagingConfiguration { get; private set; }
         private readonly QueueTrackingStatistic queueTracking;
+        private int numberOfConnectedGateways = 0;
 
         public ProxiedMessageCenter(ClientConfiguration config, IPAddress localAddress, int gen, GrainId clientId, IGatewayListProvider gatewayListProvider)
         {
@@ -423,6 +424,19 @@ namespace Orleans.Messaging
                 throw new ArgumentException("Handshake client ID can only be updated  with a geo client.", nameof(clientId));
 
             ClientId = clientId;
+        }
+
+        internal void OnGatewayConnectionOpen()
+        {
+            Interlocked.Increment(ref numberOfConnectedGateways);
+        }
+
+        internal void OnGatewayConnectionClosed()
+        {
+            if (Interlocked.Decrement(ref numberOfConnectedGateways) == 0)
+            {
+                GrainClient.NotifyClusterConnectionLost();
+            }
         }
 
         public void Dispose()
--- a/src/Orleans/Messaging/ProxiedMessageCenter.cs
+++ b/src/Orleans/Messaging/ProxiedMessageCenter.cs
@@ -373,7 +373,10 @@ namespace Orleans.Messaging
         /// </summary>
         public void Disconnect()
         {
-            throw new NotImplementedException("Disconnect");
+            foreach (var connection in gatewayConnections.Values)
+            {
+                connection.Stop();
+            }
         }
 
         /// <summary>
--- a/test/Tester/ClientConnectionTests/ClientDisconnectionEventTests.cs
+++ b/test/Tester/ClientConnectionTests/ClientDisconnectionEventTests.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Orleans;
+using Orleans.Runtime;
+using Orleans.TestingHost;
+using TestExtensions;
+using UnitTests.GrainInterfaces;
+using Xunit;
+
+namespace Tester.ClientConnectionTests
+{
+    public class ClientDisconnectionEventTests : TestClusterPerTest
+    {
+        public override TestCluster CreateTestCluster()
+        {
+            return new TestCluster(new TestClusterOptions(2));
+        }
+
+        [Fact, TestCategory("BVT")]
+        public async Task EventSendWhenDisconnectedFromCluster()
+        {
+            var runtime = (OutsideRuntimeClient) RuntimeClient.Current;
+
+            var semaphore = new SemaphoreSlim(0, 1);
+            GrainClient.ClusterConnectionLost += (sender, args) => semaphore.Release();
+
+            // Burst lot of call, to be sure that we are connected to all silos
+            for (int i = 0; i < 100; i++)
+            {
+                var grain = GrainFactory.GetGrain<ITestGrain>(i);
+                await grain.SetLabel(i.ToString());
+            }
+
+            runtime.Disconnect();
+
+            Assert.True(await semaphore.WaitAsync(TimeSpan.FromSeconds(10)));
+        }
+    }
+}
--- a/src/Orleans/Messaging/GatewayConnection.cs
+++ b/src/Orleans/Messaging/GatewayConnection.cs
@@ -77,8 +77,7 @@ namespace Orleans.Messaging
             }
             if (s == null) return;
 
-            SocketManager.CloseSocket(s);
-            NetworkingStatisticsGroup.OnClosedGatewayDuplexSocket();
+            CloseSocket(s);
         }
 
         // passed the exact same socket on which it got SocketException. This way we prevent races between connect and disconnect.
@@ -101,13 +100,11 @@ namespace Orleans.Messaging
             }
             if (s != null)
             {
-                SocketManager.CloseSocket(s);
-                NetworkingStatisticsGroup.OnClosedGatewayDuplexSocket();
+                CloseSocket(s);
             }
             if (socket2Disconnect == s) return;
 
-            SocketManager.CloseSocket(socket2Disconnect);
-            NetworkingStatisticsGroup.OnClosedGatewayDuplexSocket();
+            CloseSocket(socket2Disconnect);
         }
 
         public void MarkAsDead()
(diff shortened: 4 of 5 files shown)


test/Tester/ClientConnectionTests/ClientDisconnectionEventTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public override TestCluster CreateTestCluster()
+        {
+            return new TestCluster(new TestClusterOptions(2));
+        }
+
+        [Fact, TestCategory("BVT")]
+        public async Task EventSendWhenDisconnectedFromCluster()
+        {
+            var runtime = (OutsideRuntimeClient) RuntimeClient.Current;
```
ReubenBond: I'm trying to get rid of `RuntimeClient.Current`. Could you grab this using DI instead?

gabikliot:
I think you are raising a very good point here @benjaminpetit .
The described window race is indeed short, but that is not the point.

First, the window may not be so short on the cluster itself. The client may see a reconnect shorty after, but from the cluster perspective it may have been longer (the period of time when there is zero silos knowing about that client may be longer then the time this client thinks he is not connected).
Second, for the type of problems we looking this callback to fix, length of disconnect may be irrelevant. If the problem is once there is a disconnect, even a short one, we loose some client specific data in the cluster, it does not matter how long it is disconnected. The data is lost.

Which means that we need to better understand what this callback is used for.
No doubt it may be usefull for some scenarios. But it may not be enough for others. I am not convinced we really need to exposue it at all. The only scenario I am aware of that needs it is the deleted persistent steam subscription problem/bug.
We discussed that at lenght in the past with @jason-bragg , and I am still of the opinion that instead of adding this callback we should simply fix the bug. Jason knows how to, we discussed that privately. I say: just fix it, and no need to expose this callback at all. 

Exposing this callback, seems to me like a violation of Orleans's principle of higher level abstractions: pushing the complexity to the client instead of solving it in the runtime. yes, our customers want to be exposed to the guts of the system, since they are used not to trust runtimes and "solve it themselves". Orleans always used to stand against that.

So to summarize: I recommend solving the problem that requires that callback and not exposing the callback at all.

sergeybykov:
The scenarios where this problem surfaced have been primarily (my interpretation) about small deployments of 1-2-3 silos where the likelihood of a client losing connection to all of them is higher than with a larger deployment. Client stream subscriptions being deleted is one but not necessarily the only manifestation of this behavior. Even in a simple scenario of all silos getting rebooted it will take the client a while to reconnect to the gateways.

While adding an even like this does expose some of the guts of the system, in my view it is better than having a 'black box' client in such a case unable to communicate with the cluster. If later we provide better guarantee for client stream subscriptions and handle cluster reboots in a more expedient way, we can always remove the event if it's not needed anymore.

gabikliot:
@sergeybykov , can you describe at least one scenario when this callback will be used by the client code (except the deleted  client stream subscriptions handling)?

For deleted  client stream subscriptions handling it maybe useful, but we should solve it internally in a similar way to how client code could solve it given this callback. It will be:
a) a much better solution, because we might be able to solve it better/more efficiently, since we have more access to more runtime internals.
b) we will solve it for all clients.

Except for that case, what are other usages? We we cant come with a real, requested and justified usage, we shouldn't be adding it, right?

sergeybykov:
@gabikliot The simplest scenario I can think of is "the cluster got shut down". This is more likely to happen with small deployments of 1-2 silos. The way this situation will manifest on the client, correct me if I'm wrong, is a series of seemingly transient failures to deliver messages. An event like this would provide a more clear indication of a loss of connection to the cluster.

Similarly, if the cluster gets rebooted, depending on timing and whether or not new silos will reuse the same addresses as old ones, it will take a while for the client to rediscover the gateways to connect to. By default we requery gateways every minute. Again, not as much of a problem for larger deployments where it is less likely that all of the gateways will disappear at the same time. An event like this would 1) provide a clear indication of the situation; and 2) allow the app to reinitialize the client faster.

Today, it would also help with the cases of stream consumers/producers and observers getting disconnected. Even if and when we eliminate those possibilities, I still see value of the event because of the reasons I listed.

gabikliot:
I think the only usage you specified is informational: client will know. It can log it, it can monitor it. But what can it do with this info to help his situation, except informational?
You wrote: reinitialize faster. How?

sergeybykov:
>You wrote: reinitialize faster. How?

For example, it can start invoking `Initialize()` every 1-3-5 seconds. In the process it can send alerts based on the app rules, e.g. after 3 consequently failed attempts to reconnect. The important part I think is that it will have a way to explicitly execute such a logic in a non-ambiguous situation.

gabikliot:
That wouldn't work right? Initialize will throw already initialized.

sergeybykov:
>That wouldn't work right? Initialize will throw already initialized.

It would have to uninitialized first, you are right. Unless we change that.

gabikliot:
The runtime can do the same: if disconnected, periodically try to reconnect. And it can do it much better: it has an access to gateway provider, so it can react faster and more intelligently. Plus of course it will solve it for all customers.

So back to my question: is there anything the client can do with this, which we cannot do better? Except for  informational, monitoring, alerting...

jason-bragg:
> is there anything the client can do with this, which we cannot do better? Except for informational, monitoring, alerting...

I don't know.  What we know is at least two internal systems we've written need to know this, so there is reason to believe other systems built on top of Orleans may well also need this signal.  Users will build services and other features on top of the actor model and we can't anticipate their needs.

I view the disconnected signal as an information signal, and don't see any reason to hide this signal from the users.

For the immediate purposes, by exposing it publicly it allows developers of stream providers (not just the internally developed ones) to clean up whatever state they have, but also allows other advanced feature sets like Transactions, Indexing, and reactive queries (all of which are being considered) to be developed as extensions to Orleans without need of internal access.

IMO, it's a question of extensibility; not every advanced feature should be tightly integrated, requiring internal access and changes.

gabikliot:
My recommendation is to delay exposing extensibility points, until we know for sure they are needed and cannot be solved within the runtime. Extensibility, in general, is an anti-pattern in Orleans. From day one we wanted "to lead the developer down the garden path", which means not leaving him too much choices to make mistakes. You can read about it here: https://github.com/akka/akka-meta/blob/master/ComparisonWithOrleans.md and also in this famous and Sergey's favorite picture of screws vs. car. Orleans is a car, not a set of tools. We were against turning Orleans into CORBA or WCF. What does that means? Deciding for the developer. Orleans is a **highly opinionated** model, not a general extensability framework.

I am not against extensibility per-se. I am against it without a reason. Both scenarios Sergey mentioned: fixing lost stream subscription and re-initialize, are better done by the runtime. In fact, as @benjaminpetit  and I explained, application cannot correctly fix the lost stream subscriptions (the subscription state could be lost while client not getting a "disconnected" callback at all). It should be fixed by the runtime.

Alerting/monitoring - these are valid scenarios and better provided via dedicated monitoring/alerting support. I think we know have Telemetry providers. We can hook up this signal there.

When the actual scenario from Transactions, Indexing, and reactive queries comes up, lets re-consider. 

Most importantly - we should fix the lost stream subscription bug asap (instead of providing hooks for application to try to fix it themselves), clearly showing **customer dedication** and **doing the right thing**.

sergeybykov:
I think this particular PR isn't a good example to argue about closeness vs. extensibility. I agree with @jason-bragg:
>I view the disconnected signal as an information signal, and don't see any reason to hide this signal from the users.

It does not change the programming model or gets us any closer to CORBA/WCF. While we support other facilities like tracing and telemetry, and we could do more in the runtime, I see no reason to hide such a signal from developer.

In general, we've been drifting toward more openness and extensibility with fetures like interceptors, DI, etc., as we've learned about all kinds of scenarios people out there find for Orleans. There's still a balance to strike here, so that we don't turn Orleans into a bucket of nuts and bolts. But I don't believe this specific change brings any such danger.

gabikliot:
OK, I disagree, but I am OK with that decision.

I still do encourage you to fix the streaming bug instead of telling clients to fix it themselves by using this signal. They will not be able to fix it correctly. And even if they could, we should fix it for everyone.

src/Orleans/Core/GrainClient.cs
```diff
@@ -381,6 +381,15 @@ public static Streams.IStreamProvider GetStreamProvider(string name)
             return RuntimeClient.Current.CurrentStreamProviderManager.GetProvider(name) as Streams.IStreamProvider;
         }
 
+        public delegate void ConnectionToClusterLostHandler(object sender, EventArgs e);
+
+        public static event ConnectionToClusterLostHandler ClusterConnectionLost;
+
+        internal static void NotifyClusterConnectionLost()
+        {
+            ClusterConnectionLost?.Invoke(null, null);
```
gabikliot: null in both?
Isn't it against the events API convention? Why using event than?

src/Orleans/Messaging/ProxiedMessageCenter.cs
```diff
@@ -425,6 +429,19 @@ internal void UpdateClientId(GrainId clientId)
...
+        internal void OnGatewayConnectionOpen()
+        {
+            Interlocked.Increment(ref numberOfConnectedGateways);
+        }
+
+        internal void OnGatewayConnectionClosed()
+        {
+            if (Interlocked.Decrement(ref numberOfConnectedGateways) == 0)
+            {
+                GrainClient.NotifyClusterConnectionLost();
```
gabikliot: should catch potential exception, log and recover. This calls into client code.

src/Orleans/Core/GrainClient.cs
```diff
@@ -381,6 +381,15 @@ public static Streams.IStreamProvider GetStreamProvider(string name)
             return RuntimeClient.Current.CurrentStreamProviderManager.GetProvider(name) as Streams.IStreamProvider;
         }
 
+        public delegate void ConnectionToClusterLostHandler(object sender, EventArgs e);
```
gabikliot: If we provide "lost" event, we will most surely need to provide "gained back" event and count. Otherwise, from pure API design perspective this looks half baked. In particular, the rule you should apply is: would it pass MSDN API review?

sergeybykov:
>@sergeybykov , I think you jumped the gun with this review. The code still needs to be solid.

You are probably right. I was under the [false] impression the code has been reviewed enough already by others.

src/Orleans/Core/GrainClient.cs
sergeybykov: Not necessarily. If the recommendation is to reinitialize the client upon a connection loss, then there is no need to a "gained back" event I think.

src/Orleans/Messaging/ProxiedMessageCenter.cs
jason-bragg: +1
May also want to consider exposing this signal in a thread safe way.. possibly via task completion?

src/Orleans/Core/GrainClient.cs
sergeybykov: `sender` should conceptually be the `GrainClient` instance. Except `GrainClient` is still static. Once we make it non-static, we should send a reference to it here. It will especially make sense for the scenarios with multiple clients connecting to different clusters. The question is what to pass before then. As ugly as it looks, I think maybe `null` as a `sender` is fine in these temporary circumstances.

Is there something meaningful we'd pass as `EventArgs`?

src/Orleans/Messaging/ProxiedMessageCenter.cs
sergeybykov: Client is free-threaded. So it's not a concern.

src/Orleans/Core/GrainClient.cs
jason-bragg: If we want to pass the client when it's no longer static, and plan to do so, can we add a comment or todo noting this.  Seems like something very easy to overlook in the future.

PR #2629, opened 2017-01-23
Fix "*.exe.config" file usage in nuspec files. Fixes #2616.

### LINKED ISSUE #2616: OrleansHost.exe.config is not placed in output folder in consuming project
When investigating the #2614 issue it turned out that OrleansHost.exe.config is not getting copied when the user builds a project which references the ```Microsoft.Orleans.OrleansHost``` package.

I think that this is an error in the given nuspec file and we need to use ```contentFiles```.

attilah:
- Modify Nuget packages to copy executable and other tool files to output directory.
- Fix file order in nuspec files (cosmetic).

### DIFF
--- a/src/OrleansManager/Microsoft.Orleans.OrleansManager.targets
+++ b/src/OrleansManager/Microsoft.Orleans.OrleansManager.targets
@@ -3,7 +3,7 @@
   <Target Name="CopyOrleansManagerTools" BeforeTargets="AfterBuild">
     <ItemGroup>
       <FilesToCopy Include="$(MSBuildThisFileDirectory)..\tools\OrleansManager.exe.config" />
-      <FilesToCopy Include="$(MSBuildThisFileDirectory)..\tools\ClientConfiguration.xml" />
+      <FilesToCopy Include="$(MSBuildThisFileDirectory)..\tools\ClientConfiguration.xml" Condition="!Exists('$(OutDir)ClientConfiguration.xml')" />
     </ItemGroup>
 
     <Copy SourceFiles="@(FilesToCopy)" DestinationFolder="$(OutDir)" />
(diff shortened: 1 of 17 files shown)


src/OrleansManager/Microsoft.Orleans.OrleansManager.targets
```diff
@@ -0,0 +1,21 @@
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Target Name="CopyOrleansManagerTools" BeforeTargets="AfterBuild">
+    <ItemGroup>
+      <FilesToCopy Include="$(MSBuildThisFileDirectory)..\tools\OrleansManager.exe.config" />
+      <FilesToCopy Include="$(MSBuildThisFileDirectory)..\tools\ClientConfiguration.xml" />
```
ReubenBond: Is there any chance that this config will override user config?

attilah: I don't know. We only supply this in a Nuget for users who want no hassle usage, like OrleansHost.exe. The Nuget is designed to run from Output directory, but Nuget was not copying that file.
With the current setup the user can overwrite this file in an AfterBuild or later target if needed, but I don't think that anyone runs it from output directory in a live environment ;-)

ReubenBond: I wonder if a conditional copy is doable

attilah: We can do that too. I'll talk with @sergeybykov what is he thinking about this.

ReubenBond: Sounds good

Issue #1609, opened 2016-03-24
Storage providers have to throw an InconsistentStateException upon a write conflict


Eldar1205:
A grain's `WriteStateAsync`leads to `GrainStateStorageBridge.WriteStateAsync` which reports any failure as `OrleansException`, when state conflicts should be reported as InconsistentStateException, according to [documentation](http://dotnet.github.io/orleans/Getting-Started-With-Orleans/Grain-Persistence). Additionally, the `AzureTableStorage` doesn't throw `InconsistentStateException` on state conflicts, since it delegates to `GrainStateTableDataManager.Write` which doesn't throw that exception. This looks like a bug in the code, not in the documentation. Do you think it should be fixed, or keep it like that since existing applications may already rely on the exceptions thrown today?

lmagyar:
As I see, the other problem is that ```InconsistentStateException``` is not handled by the framework at all.
See also #1420

Eldar1205:
IMO the application should be able to response first before Orleans, but Orleans should respond if the application doesn't. Think of that as convention over configuration. The fact the documented exception isn't what's thrown from the (probably) most used built-in storage provider is a bug IMO and should be handled regardless of what to do with the situation.

sergeybykov:
I think the desired behavior here is for storage providers that support eTags to throw an `InconsistentStateException` as per the documentation. Once such an exception is thrown from `WriteStateAsync`, grain code has a chance to catch and handle it. If grain code allows the exception to escape, silo should deactivate the activation and deliver the exception to the caller just like any other exception thrown from a grain method call.

shayhatsor:
@sergeybykov, regarding your previous comment:
> If grain code allows the exception to escape, silo should deactivate the activation and deliver the exception to the caller just like any other exception thrown from a grain method call.

i think that deactivation is a good idea only in special exceptions like `InconsistentStateException`. In other cases, this incurs a performance penalty of deactivation,activation and read of state from backing store.

gabikliot:
@shayhatsor , yes, we were talking here about behavior in the case of `InconsistentStateException`.

sergeybykov:
@shayhatsor @gabikliot Yes, this is only for `InconsistentStateException`.

shayhatsor:
@sergeybykov, if I understand correctly, this issue can be closed and a new issue, labeled enhancement, should be opened for deactivating stateful grains when an InconsistentStateException is thrown.

Eldar1205:
[...]

PR #2630, opened 2017-01-23
Azure table storage throws InconsistentStateException

jason-bragg:
Azure table storage throws InconsistentStateException upon a write conflict.

See issue "Storage providers have to throw an InconsistentStateException upon a write conflict #1609"

### DIFF
--- a/src/Orleans/Core/GrainStateStorageBridge.cs
+++ b/src/Orleans/Core/GrainStateStorageBridge.cs
@@ -92,7 +92,7 @@ namespace Orleans.Core
                 string errMsgToLog = MakeErrorMsg(what, errorOccurred);
                 store.Log.Error((int) ErrorCode.StorageProvider_WriteFailed, errMsgToLog, errorOccurred);
                 // If error is not specialization of OrleansException, wrap it
-                if (!errorOccurred.GetType().IsInstanceOfType(typeof(OrleansException)))
+                if (!(errorOccurred is OrleansException))
                 {
                     errorOccurred = new OrleansException(errMsgToLog, errorOccurred);
                 }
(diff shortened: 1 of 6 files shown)


src/Orleans/Core/GrainStateStorageBridge.cs
```diff
@@ -91,7 +91,11 @@ public GrainStateStorageBridge(string grainTypeName, IStatefulGrain grain, IStor
 
                 string errMsgToLog = MakeErrorMsg(what, errorOccurred);
                 store.Log.Error((int) ErrorCode.StorageProvider_WriteFailed, errMsgToLog, errorOccurred);
-                errorOccurred = new OrleansException(errMsgToLog, errorOccurred);
+                // If error is not specialization of OrleansException, wrap it
+                if (!errorOccurred.GetType().IsInstanceOfType(typeof(OrleansException)))
```
ReubenBond: Would it be better to use `!(errorOccurred is OrleansException)`?

src/OrleansAzureUtils/Providers/Storage/AzureTableStorage.cs
```diff
@@ -176,10 +177,20 @@ public Task Close()
                 await tableDataManager.Write(record);
                 grainState.ETag = record.ETag;
             }
+            catch (StorageException exc)
+            {
+                Log.Error((int)AzureProviderErrorCode.AzureTableProvider_WriteError,
+                    $"Error Writing: GrainType={grainType} Grainid={grainReference} ETag={grainState.ETag} to Table={tableName} Exception={exc.Message}", exc);
+                if (exc.IsUpdateConditionNotSatisfiedError())
+                {
+                    throw new InconsistentStateException("Conditions not satisfied", exc);
```
ReubenBond: Since users will see this exception, maybe the message should be more descriptive, like "Conflict while writing grain state: precondition check failed. This likely indicates a duplicate activation." - I don't know. We could also consider including some context, like what's in the log message.

ReubenBond:
@dotnet-bot test this please

Issue #2528, opened 2016-12-17
Unable to serialize new .NET package exceptions


kogir:
Packages I depend on have begun migrating to what I think is "netstandard"? In any case, they now require tiny little micro-packages for each .NET namespace. This would be merely annoying, except the latest versions of these namespace level packages contain types which are no longer `[Serializable]`.

It appears that many exceptions Orleans used to successfully relay from grains to my client applications can no longer be serialized out of the box, and it's not clear how I should add support for serializing them.

Here's a test project demonstrating my problem, and two now unserializable exception types that have been appearing in my logs: https://github.com/realartists/orleans-serialization-failure

I've tried `KnownType`, `KnownAssembly`, and `GenerateSerializer` attributes without success. `KnownType` and `KnownAssembly` silently fail to generate a serializer, only failing at runtime, and `GenerateSerializer` fails during build.

I tried adding a custom serializer for `Exception`, but it's never called for derived types, even as a fallback.

I must be missing something obvious.

I'm using the 1.3.1 nuget package. The specific exceptions I'm seeing are both `Orleans.Runtime.OrleansException`:

_No copier found for object of type System.Net.Http.HttpRequestException. Perhaps you need to mark it [Serializable] or define a custom serializer for it?_
```
Orleans.Serialization.SerializationManager.DeepCopierHelper(Type t, Object original):640
 Orleans.Serialization.SerializationManager.DeepCopyInner(Object original):106
 Orleans.Serialization.SerializationManager.DeepCopy(Object original):41
 Orleans.Runtime.InsideRuntimeClient.SafeSendExceptionResponse(Message message, Exception ex):0
... [My grain code]
```
_No copier found for object of type System.Net.Http.UnsupportedMediaTypeException. Perhaps you need to mark it [Serializable] or define a custom serializer for it?_
```
Orleans.Serialization.SerializationManager.DeepCopierHelper(Type t, Object original):640
 Orleans.Serialization.SerializationManager.DeepCopyInner(Object original):106
 Orleans.Serialization.SerializationManager.DeepCopy(Object original):41
 Orleans.Runtime.InsideRuntimeClient.SafeSendExceptionResponse(Message message, Exception ex):0
```

sergeybykov:
I believe this is already resolved in master via #2162 (has to be explicitly configured for now). Will be included in v1.4.0. Can you build latest source and try with that?

kogir:
Oh, great. I'll try that out and let you know. Thanks!

[...]

Issue #2310, opened 2016-10-17
Failsafe Exception Serialization


ReubenBond:
We currently rely on all serialized types to be present during deserialization. This means that the assembly for any type transferred over the wire must be present on the receiving end. This is fine for most cases, but can be tricky for 'unexpected' types, such as exceptions.

Users would like to be able to see basic information about which exceptions are being thrown and from where, even if they are unable to deserialize the concrete exception. We can accomplish this using the `ILBasedSerializer`, adopting a special scheme for serializing Exceptions whereby base class fields are serialized first, like so:
`<FullTypeName><Length><ExceptionBaseClassFields><RemainingFields>`

Using this scheme, we can fall back to deserializing into `Exception` or an `OrleansFallbackException` class and retain the original `Message` and `StackTrace` properties along with the name of the original exception type.

I've prototyped this and added a test case. I can make it PR ready.

Thoughts? Should I pursue this?

jdom:
I think it's valuable. Main cases is when the client does not have all the types that the silos use, but also the upcoming heterogeneous deployments, where not all silos are equal.
Unofortunately this will only work when ILBasedSerializer is used as the fallback, and not when using Binary Formatter, but still, I think it's good.
BTW, if I would state a preference, I'd call it something like RemoteNonDeserializableException (or something as non-ambiguous of that sort) and not make it inherit from OrleansException, since these are typically related to protocol.

gabikliot:
+1

sergeybykov:
What about non-exception types? Do we throw a meaningful exception if, say, a `Task<object>` method tries to return an object of a type not present on the caller's side?

ReubenBond:
> Unfortunately this will only work when ILBasedSerializer is used as the fallback, and not when using Binary Formatter, but still, I think it's good.

Maybe we should switch to ILBasedSerializer :) We could default to ILBasedSerializer for exceptions and keep using BinaryFormatter for other exotic types.

> I'd call it something like RemoteNonDeserializableException (or something as non-ambiguous of that sort) and not make it inherit from OrleansException

Noted, lgtm.

> What about non-exception types? Do we throw a meaningful exception if, say, a Task<object> method tries to return an object of a type not present on the caller's side?

At the moment, a `TypeLoadException` will be thrown from both `ILBasedSerializer` and `BinaryFormatterSerializer`. It should contain the original type name, like so:
[...]

PR #2633, opened 2017-01-25
Failsafe Exception serialization (#2310)

ReubenBond:
This is an implementation of failsafe Exception serialization.

Features:
* Serializes Exceptions so that they can be deserialized even if the original exception type is not available.
* A fallback type, `RemoteNonDeserializableException`, is used in place of the original exception type when the original is not available. In this case, the message and stack trace are preserved, but any additional fields on sub-classes of `Exception` cannot be inspected.
* Exceptions are round-tripped without any loss of fidelity, regardless of whether or not one of the parties does not have access to the original exception type.

Caveats:
* Currently it does not support referential transparency for the exception itself.
* This is an extension to `ILBasedSerializer` and therefore will only be enabled if that is the configured `FallbackSerializationProvider`. This it the default on .NET Core, but not Full .NET.

Let me know if I missed something.

xref #2528

### DIFF
--- a/src/Orleans/Serialization/ILBasedExceptionSerializer.cs
+++ b/src/Orleans/Serialization/ILBasedExceptionSerializer.cs
@@ -1,4 +1,6 @@
-﻿namespace Orleans.Serialization
+﻿using System.Runtime.Serialization;
+
+namespace Orleans.Serialization
 {
     using System;
     using System.Collections.Concurrent;
@@ -145,7 +147,16 @@
                 {
                     exception.AdditionalData = innerContext.StreamReader.ReadBytes(additionalDataLength);
                 }
-                
+
+                // If a particular type is expected, but the actual type is not available, throw an exception to avoid 
+                if (expectedType != null && !expectedType.IsAssignableFrom(typeof(RemoteNonDeserializableException)))
+                {
+                    throw new SerializationException(
+                        $"Unable to deserialize exception of unavailable type {exception.OriginalTypeName} into expected type {expectedType}. " +
+                        $" See {nameof(Exception.InnerException)} for recovered information.",
+                        exception);
+                }
+
                 result = exception;
             }
             
--- a/src/Orleans/Serialization/ILBasedExceptionSerializer.cs
+++ b/src/Orleans/Serialization/ILBasedExceptionSerializer.cs
@@ -1,0 +1,231 @@
+﻿namespace Orleans.Serialization
+{
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Reflection;
+
+    /// <summary>
+    /// Methods for serializing instances of <see cref="Exception"/> and its subclasses.
+    /// </summary>
+    internal class ILBasedExceptionSerializer
+    {
+        private static readonly Type ExceptionType = typeof(Exception);
+
+        /// <summary>
+        /// The collection of serializers.
+        /// </summary>
+        private readonly ConcurrentDictionary<Type, SerializationManager.SerializerMethods> serializers =
+            new ConcurrentDictionary<Type, SerializationManager.SerializerMethods>();
+
+        /// <summary>
+        /// The field filter used for generating serializers for subclasses of <see cref="Exception"/>.
+        /// </summary>
+        private readonly Func<FieldInfo, bool> exceptionFieldFilter;
+
+        /// <summary>
+        /// The serializer used as a fallback when the concrete exception type is unavailable.
+        /// </summary>
+        /// <remarks>
+        /// This serializer operates on <see cref="RemoteNonDeserializableException"/> instances, however it 
+        /// includes only fields from <see cref="Exception"/> and no sub-class fields.
+        /// </remarks>
+        private readonly SerializationManager.SerializerMethods fallbackBaseExceptionSerializer;
+        
+        /// <summary>
+        /// The serializer generator.
+        /// </summary>
+        private readonly ILSerializerGenerator generator;
+
+        private readonly TypeSerializer typeSerializer;
+
+        public ILBasedExceptionSerializer(ILSerializerGenerator generator, TypeSerializer typeSerializer)
+        {
+            this.generator = generator;
+            this.typeSerializer = typeSerializer;
+
+            // Exceptions are a special type in .NET because of the way they are handled by the runtime.
+            // Only certain fields can be safely serialized.
+            this.exceptionFieldFilter = field =>
+            {
+                // Any field defined below Exception is acceptable.
+                if (field.DeclaringType != ExceptionType) return true;
+
+                // Certain fields from the Exception base class are acceptable.
+                return field.FieldType == typeof(string) || field.FieldType == ExceptionType;
+            };
+
+            // When serializing the fallback type, only the fields declared on Exception are included.
+            // Other fields are manually serialized.
+            this.fallbackBaseExceptionSerializer = this.generator.GenerateSerializer(
+                typeof(RemoteNonDeserializableException),
+                field =>
+                {
+                    // Only serialize base-class fields.
+                    if (field.DeclaringType != ExceptionType) return false;
+
+                    // Certain fields from the Exception base class are acceptable.
+                    return field.FieldType == typeof(string) || field.FieldType == ExceptionType;
+                },
+                fieldComparer: ExceptionFieldInfoComparer.Instance);
+
+            // Ensure that the fallback serializer only ever has its base exception fields serialized.
+            this.serializers[typeof(RemoteNonDeserializableException)] = this.fallbackBaseExceptionSerializer;
+        }
+
+        public void Serialize(object item, ISerializationContext outerContext, Type expectedType)
+        {
+            var outerWriter = outerContext.StreamWriter;
+            
+            var actualType = item.GetType();
+
+            // To support loss-free serialization where possible, instances of the fallback exception type are serialized in a
+            // semi-manual fashion.
+            var fallbackException = item as RemoteNonDeserializableException;
+            if (fallbackException != null)
+            {
+                this.ReserializeFallback(fallbackException, outerContext);
+                return;
+            }
+            
+            // Write the concrete type directly.
+            this.typeSerializer.WriteNamedType(actualType, outerWriter);
+
+            var innerContext = new SerializationContext
+            {
+                StreamWriter = new BinaryTokenStreamWriter()
+            };
+
+            // Serialize the exception itself.
+            var methods = this.GetSerializerMethods(actualType);
+            methods.Serialize(item, innerContext, null);
+
+            // Write the serialized exception to the output stream.
+            outerContext.StreamWriter.Write(innerContext.StreamWriter.CurrentOffset);
+            outerContext.StreamWriter.Write(innerContext.StreamWriter.ToBytes());
+        }
+
+        public object Deserialize(Type expectedType, IDeserializationContext outerContext)
+        {
+            var outerReader = outerContext.StreamReader;
+
+            var typeKey = TypeSerializer.ReadTypeKey(outerReader);
+            
+            // Read the serialized payload.
+            var length = outerReader.ReadInt();
+            var innerBytes = outerReader.ReadBytes(length);
+            
+            var innerContext = new DeserializationContext
+            {
+                StreamReader = new BinaryTokenStreamReader(innerBytes)
+            };
+            
+            object result;
+
+            // If the concrete type is available and the exception is valid for reconstruction,
+            // reconstruct the original exception type.
+            var actualType = this.typeSerializer.GetTypeFromTypeKey(typeKey, throwOnError: false);
+            if (actualType != null)
+            {
+                // Deserialize into the concrete type.
+                var methods = this.GetSerializerMethods(actualType);
+                result = methods.Deserialize(null, innerContext);
+            }
+            else
+            {
+                // Since the concrete type is unavailable, deserialize into the fallback type.
+
+                // Read the Exception fields.
+                var exception = (RemoteNonDeserializableException)this.fallbackBaseExceptionSerializer.Deserialize(null, innerContext);
+                exception.OriginalTypeName = typeKey.GetTypeName();
+
+                // If there is additional data, store it for later serialization.
+                var additionalDataLength = length - innerContext.StreamReader.CurrentPosition;
+                if (additionalDataLength > 0)
+                {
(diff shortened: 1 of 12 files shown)


galvesribeiro:
Can't we serialize the original exception types as a dictionary so people have an alternative to the properties? At least if those properties are primitive types... That would be good.

ReubenBond:
@galvesribeiro We could serialize those other fields, but I don't think it's worthwhile for the trouble it brings, and most exceptions give enough context in their `Message` + `StackTrace`. Remember: this is just a fallback for when the original type isn't available, a slightly degraded experience is to be expected.

ReubenBond:
@dotnet-bot test this please

sergeybykov:
@dotnet-bot test this please

src/Orleans/Serialization/ILBasedExceptionSerializer.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        /// <summary>
+        /// Returns a copy of the provided instance.
+        /// </summary>
+        /// <param name="original">The object to copy.</param>
+        /// <param name="context">The copy context.</param>
+        /// <returns>A copy of the provided instance.</returns>
+        public object DeepCopy(object original, ICopyContext context)
+        {
+            return original;
```
jason-bragg: No copy?

src/Orleans/Serialization/RemoteNonDeserializableException.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace Orleans.Serialization
+{
+    using System;
+    using System.Text;
+
+    /// <summary>
+    /// Represents an exception which cannot be fully deserialized.
+    /// </summary>
+    [Serializable]
+    public class RemoteNonDeserializableException : Exception
```
jason-bragg: All Orleans exceptions should inherit from OrleansException

src/Orleans/Serialization/ILBasedExceptionSerializer.cs
ReubenBond: Exceptions should all be immutable

src/Orleans/Serialization/ILBasedExceptionSerializer.cs
```diff
@@ -0,0 +1,231 @@
...
+{
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Reflection;
+
+    /// <summary>
+    /// Methods for serializing instances of <see cref="Exception"/> and its subclasses.
+    /// </summary>
+    internal class ILBasedExceptionSerializer
```
ReubenBond: That name is a bit long. I could call it `CilBasedExceptionSerializer`, but I think that's harder to read. This also entails renaming the existing IL* classes

src/Orleans/Serialization/RemoteNonDeserializableException.cs
ReubenBond: Will fix

jason-bragg:
Question about problem space:
What is the difference between an exception not being [de]serializable, and any other object.  I'm questioning why we are special casing exceptions here rather than any serialization issue.  I'm asking because it seems like we could catch the serialization issues at a higher level and produce our own serialization exception without needing specific logic built into a serializer, which may not even be the one a customer is using..?

ReubenBond:
@jason-bragg good questions.
> What is the difference between an exception not being [de]serializable, and any other object. I'm questioning why we are special casing exceptions here rather than any serialization issue.

The difference is that when an exception is thrown, it can be any exception, whereas other objects cannot be replaced by our own special 'failsafe object', since that would result in cast exceptions. We could try to apply the same technique to the entire type hierarchy and deserialize to the best possible match, but that has the downside of not being round-trippable because sub-class data would be lost - unless we were to synthesize types at runtime in order to provide some field to stick that extension data on. It's doable, but far from simple.

The point of this PR is to provide a better experience than just "some serialization exception occurred".

> I'm asking because it seems like we could catch the serialization issues at a higher level and produce our own serialization exception without needing specific logic built into a serializer, which may not even be the one a customer is using..?

Fair point. Currently, we have two fallback serialziers, ILBased and BinaryFormatter. When fallback seralizers are used, we do not serialize our own type information (it's the responsibility of the fallback serializer to handle that - which is good), so we would have to change this behavior in order to be able to reconstruct any useful information for users.

jason-bragg:
>other objects cannot be replaced by our own special 'failsafe object'

>The point of this PR is to provide a better experience than just "some serialization exception occurred".

I think this is a question of what user experience we want to support.  I see little value in a 'failsafe' object either for exceptions or application objects.  Exception handling will either be general, in which any 'serialization exception' would be sufficient, or specific in which a failsafe would add no value.  IMO, if the application layer returns an exception or data that is not serializable, this is an application layer bug, and our responsibility here is to provide the user with the information necessary to address the issue.  Attempts to build some sort of fallback that may allow them to account for such issues, seems of limited value.  I'm not saying we shouldn't do this change, I'm just questioning the value vs. the complexity.  I really don't know if we want to be in the business of compensating for application level bugs.  Providing good monitoring, diagnostics and reporting, yes, but beyond that..?

jason-bragg:
Never Mind - I caught up with the original thread.  Makes sense for a fallback serializer to special case exceptions to help application engineers.  Still suspect that this would be better handled at a higher level and not require special handling in the fallback serializer, but I trust the judgment of those with deeper insight into the problem.

galvesribeiro:
@jason-bragg I agree with the majority of your points. Mask the real exception in this case could lead to a false impression.

However, I was the one complaining to @ReubenBond that we are not serializing exceptions properly and the reason for that, is because I was with an Azure SDK exception too deep (inside my grain constructor related to ES code) and since that exception type was not available at the client side, it took me too much time banging my head to find out the proper exception because I was getting Orleans WARNINGs saying that it can´t deserialize the exception. 

With that in mind, I think @ReubenBond's approach on this fallback will be useful, and it will only be hit in case the client never have the type to be deserialized.

galvesribeiro:
Also, it is not just a matter of handling it... I agree it is a bug in application layer, but it help a lot at development time to have at least a string message which includes the content of the original exception so the developer can easily figure out what is going on.

src/Orleans/Serialization/RemoteNonDeserializableException.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace Orleans.Serialization
+{
+    using System;
+    using System.Text;
+
+    /// <summary>
+    /// Represents an exception which cannot be fully deserialized.
+    /// </summary>
+    [Serializable]
+    public class RemoteNonDeserializableException : OrleansException
```
jason-bragg: I don't think this should have anything internal in it as it prevents other external serializers from developing comparable feature sets.

src/Orleans/Serialization/RemoteNonDeserializableException.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// Gets the type name of the original <see cref="Exception"/> represented by this instance.
+        /// </summary>
+        public string OriginalTypeName { get; internal set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the original type of this exception can be reconstructed during
+        /// deserialization.
+        /// </summary>
+        internal bool CanAttemptReconstruction { get; set; }
```
jason-bragg: How do we know we can attempt reconstruction?  If we do, how do we know we can populate the exception correctly?  Seems sketchy.

jason-bragg: What test cases should we have for this?  Are there any constraints for this behavior?

src/Orleans/Serialization/RemoteNonDeserializableException.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Gets or sets the additional data deserialized alongside this 
+        /// </summary>
+        internal byte[] AdditionalData { get; set; }
```
jason-bragg: Additional data deserialized alongside this?  Like what?

src/Orleans/Serialization/RemoteNonDeserializableException.cs
```diff
@@ -0,0 +1,78 @@
...
+    using System;
+    using System.Text;
+
+    /// <summary>
+    /// Represents an exception which cannot be fully deserialized.
+    /// </summary>
+    [Serializable]
+    public class RemoteNonDeserializableException : OrleansException
+    {
+        public RemoteNonDeserializableException() { }
```
jason-bragg: I'd think we'd at least need original exception's message, type, and call stack..?

jason-bragg:
@galvesribeiro, Agreed.  My questions and comments are regarding the goal.  I think the goal should be to provide valuable information to a developer to help them track down the problem in their application, rather than to compensate or account for that error.

galvesribeiro:
Exactly! That is what I tried to explain, @jason-bragg. In the end, this fallback exception serializer would do nothing more than capture the *unknown* exception info (i.e. type, message, stack) and put inside a *known* exception type which can go over the wire and help developer to track the problem easily.

src/Orleans/Serialization/RemoteNonDeserializableException.cs
ReubenBond: We get that - all fields from the base `Exception` class are serialized and deserialized correctly

src/Orleans/Serialization/RemoteNonDeserializableException.cs
ReubenBond: Ah, that field should have been removed. It's not used and isn't needed. Thanks for pointing this out.

kogir:
I look forward to removing my current hack (included below to show an example of the motivation):

```C#
namespace RealArtists.ShipHub.CloudServices.OrleansSilos {
  using System;
  using System.Collections.Concurrent;
  using System.Threading.Tasks;
  using Orleans.Providers;
  using Orleans.Serialization;

  public class ShipHubBootstrapProvider : IBootstrapProvider {
    public string Name { get; private set; }

    private ConcurrentDictionary<Type, bool> _exceptionSerializability = new ConcurrentDictionary<Type, bool>();

    public Task Close() {
      return Task.CompletedTask;
    }

    public Task Init(string name, IProviderRuntime providerRuntime, IProviderConfiguration config) {
      Name = name;

      // Force exceptions to be serializable
      providerRuntime.SetInvokeInterceptor(async (method, request, grain, invoker) => {
        try {
          return await invoker.Invoke(grain, request);
        } catch (Exception ex) {
          var exType = ex.GetType();
          bool serializable;

          if (!_exceptionSerializability.TryGetValue(exType, out serializable)) {
            serializable = Attribute.IsDefined(exType, typeof(SerializableAttribute));
            _exceptionSerializability.TryAdd(exType, serializable);
          }

          if (serializable) {
            throw;
          } else {
            throw new Exception(ex.ToString());
          }
        }
      });

      return Task.CompletedTask;
    }
  }
}
```

src/Orleans/Serialization/RemoteNonDeserializableException.cs
ReubenBond: Any fields defined in subclasses of Exception are included here. For example, in `TableStorageUpdateConditionNotSatisfiedException`, the `GrainId`, `GrainType`, and `TableName` fields will be serialized here. I've updated the comment.

src/Orleans/Serialization/RemoteNonDeserializableException.cs
jason-bragg: I meant we should pass them into this constructor when we construct the object directly, but after further investigation, I noticed we never really do that... NM :)

src/Orleans/Serialization/ILBasedExceptionSerializer.cs
jason-bragg: Ur call.  Just a nit pick detail think.  No strong opinion on this.   :)

ReubenBond:
@jason-bragg this PR doesn't help users paper over application bugs, it just helps with debugging. Clients shouldn't be required to load EntityFramework, Azure Storage, or other dlls just so they can see transient exceptions.

Take the `TableStorageUpdateConditionNotSatisfiedException` exception you recently added, for example. It includes an `InnerException` field for the actual exception. Without this PR, that inner exception will cause a serialization error on the client unless they've loaded the Azure storage DLLs.

You can argue that grains should never throw, but that's a question of development philosophy rather than a definite bug.

src/Orleans/Serialization/ILBasedExceptionSerializer.cs
jason-bragg: Understood - All good.

src/Orleans/Serialization/RemoteNonDeserializableException.cs
ReubenBond: That's fair. I'll open it up.

src/Orleans/Serialization/RemoteNonDeserializableException.cs
jason-bragg: Quite suspicious of the value of this.  It's internal, so applications can't even use it.  Data is packed in binary which application will likely not know what to do with.  What if one of the property types is not serializable?  A property bag may have some value, as at least that could contain named properties that the application layer can inspect.  I'd drop for now as type, message, stack should be suffice, but just my opinion.

ReubenBond: @jason-bragg I understand the skepticism - the value is questionable, but it does add value. This data is serialized regardless, but this property allows the exception to be safely round-tripped through nodes which cannot deserialize it fully, which is a possible (rare) case.

It allows the serializer to maintain the property that an exception of type X and an exception of type `RemoteNonDeserializableException` representing that original exception have an identical wire representation.

jason-bragg:
> This PR doesn't help users paper over application bugs, it just helps with debugging.

Great! Perfect +1

> Clients shouldn't be required to load EntityFramework, Azure Storage, or other dlls just so they can see transient exceptions.

Agreed, but grain interfaces and the exceptions they throw are a contract in a distributed system, so it is an application error bug if they are throwing exceptions that can't be transmitted and understood by the caller.  We can do better (as this PR does) at communicating that this is happening and aid them in correcting the issue, but we can't avoid this constraint.

ReubenBond:
> Agreed, but grain interfaces and the exceptions they throw are a contract in a distributed system, so it is an application error bug if they are throwing exceptions that can't be transmitted and understood by the caller. We can do better (as this PR does) at communicating that this is happening and aid them in correcting the issue, but we can't avoid this constraint.

That's fair, I agree with that.

jdom:
:)

PR #2637, opened 2017-01-26
EventHubs Stream provider ported to netstandard

galvesribeiro:
As part of #2145 this PR add EventHub stream provider support to vNext solution supporting `netstandard1.5`.

The new `netstandard1.5` project rely on the new nugets [Microsoft.Azure.EventHubs](https://www.nuget.org/packages/Microsoft.Azure.EventHubs/0.0.4-preview) and [Microsoft.Azure.EventHubs.Processor](https://www.nuget.org/packages/Microsoft.Azure.EventHubs.Processor/0.0.4-preview) which will deprecate the old ones once they are released. There were namespace changes and a bit of API changes on those nugets and they were addressed with some `#if`s among other small changes. 

For now those nugets were only used at vNext and should only be updated on master when we merge the solutions.

cc: @jason-bragg

### DIFF
--- a/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
+++ b/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
@@ -3,8 +3,12 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+#if NETSTANDARD
+using Microsoft.Azure.EventHubs;
+#else
 using Microsoft.ServiceBus;
 using Microsoft.ServiceBus.Messaging;
+#endif
 using Orleans.Providers;
 using Orleans.Providers.Streams.Common;
 using Orleans.Runtime;
@@ -108,7 +112,15 @@ namespace Orleans.ServiceBus.Providers
             adapterSettings = new EventHubStreamProviderSettings(providerName);
             adapterSettings.PopulateFromProviderConfig(providerConfig);
             hubSettings = adapterSettings.GetEventHubSettings(providerConfig, serviceProvider);
+#if NETSTANDARD
+            var connectionStringBuilder = new EventHubsConnectionStringBuilder(hubSettings.ConnectionString)
+            {
+                EntityPath = hubSettings.Path
+            };
+            client = EventHubClient.CreateFromConnectionString(connectionStringBuilder.ToString());
+#else
             client = EventHubClient.CreateFromConnectionString(hubSettings.ConnectionString, hubSettings.Path);
+#endif
 
             if (CheckpointerFactory == null)
             {
@@ -203,7 +215,11 @@ namespace Orleans.ServiceBus.Providers
                 throw new NotImplementedException("EventHub stream provider currently does not support non-null StreamSequenceToken.");
             }
             EventData eventData = EventHubBatchContainer.ToEventData(streamGuid, streamNamespace, events, requestContext);
-            return client.SendAsync(eventData);
+#if NETSTANDARD
+            return client.SendAsync(eventData, streamGuid.ToString());
+#else
+            return client.SendAsync(eventData); 
+#endif
         }
 
         /// <summary>
@@ -243,9 +259,14 @@ namespace Orleans.ServiceBus.Providers
 
         private async Task<string[]> GetPartitionIdsAsync()
         {
+#if NETSTANDARD
+            EventHubRuntimeInformation runtimeInfo = await client.GetRuntimeInformationAsync();
+            return runtimeInfo.PartitionIds;
+#else
             NamespaceManager namespaceManager = NamespaceManager.CreateFromConnectionString(hubSettings.ConnectionString);
             EventHubDescription hubDescription = await namespaceManager.GetEventHubAsync(hubSettings.Path);
-            return hubDescription.PartitionIds;
+            return hubDescription.PartitionIds; 
+#endif
         }
     }
 }
--- a/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
+++ b/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
@@ -91,7 +91,7 @@ namespace Orleans.ServiceBus.Providers
             {
                 checkpointer = await checkpointerFactory(settings.Partition);
                 cache = cacheFactory(settings.Partition, checkpointer, baseLogger);
-                flowController = new AggregatedQueueFlowController(MaxMessagesPerRead) {cache, LoadShedQueueFlowController.CreateAsPercentOfLoadSheddingLimit()};
+                flowController = new AggregatedQueueFlowController(MaxMessagesPerRead) { cache, LoadShedQueueFlowController.CreateAsPercentOfLoadSheddingLimit() };
                 string offset = await checkpointer.Load();
                 receiver = await CreateReceiver(settings, offset, logger);
                 monitor.TrackInitialization(true);
@@ -129,7 +129,7 @@ namespace Orleans.ServiceBus.Providers
                 var watch = Stopwatch.StartNew();
                 messages = (await receiver.ReceiveAsync(maxCount, ReceiveTimeout))?.ToList();
                 watch.Stop();
-                
+
                 monitor.TrackRead(true);
                 monitor.TrackMessagesRecieved(messages?.Count ?? 0, watch.Elapsed);
             }
@@ -266,7 +266,7 @@ namespace Orleans.ServiceBus.Providers
             }
 #endif
             // if we have a starting offset or if we're not configured to start reading from utc now, read from offset
-            if (!partitionSettings.Hub.StartFromNow || 
+            if (!partitionSettings.Hub.StartFromNow ||
                 offset != EventHubConstants.StartOfStream)
             {
                 logger.Info("Starting to read from EventHub partition {0}-{1} at offset {2}", partitionSettings.Hub.Path, partitionSettings.Partition, offset);
@@ -286,7 +286,10 @@ namespace Orleans.ServiceBus.Providers
             }
 #if NETSTANDARD
             PartitionReceiver receiver = client.CreateReceiver(partitionSettings.Hub.ConsumerGroup, partitionSettings.Partition, offset, offsetInclusive);
-            receiver.PrefetchCount = partitionSettings.Hub.PrefetchCount.Value;
+
+            if (partitionSettings.Hub.PrefetchCount.HasValue)
+                receiver.PrefetchCount = partitionSettings.Hub.PrefetchCount.Value;
+
             return receiver;
 #else
             return await consumerGroup.CreateReceiverAsync(partitionSettings.Partition, offset, offsetInclusive); 
--- a/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
+++ b/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
@@ -131,7 +131,7 @@ namespace Orleans.ServiceBus.Providers
                 watch.Stop();
                 
                 monitor.TrackRead(true);
-                monitor.TrackMessagesRecieved(messages != null ? messages.Count : 0, watch.Elapsed);
+                monitor.TrackMessagesRecieved(messages?.Count ?? 0, watch.Elapsed);
             }
             catch (Exception ex)
             {
@@ -219,12 +219,9 @@ namespace Orleans.ServiceBus.Providers
 
                 // clear cache and receiver
                 IEventHubQueueCache localCache = Interlocked.Exchange(ref cache, null);
-#if NETSTANDARD
-                PartitionReceiver
-#else
-                EventHubReceiver  
-#endif
-                    localReceiver = Interlocked.Exchange(ref receiver, null);
+
+                var localReceiver = Interlocked.Exchange(ref receiver, null);
+
                 // start closing receiver
                 Task closeTask = TaskDone.Done;
                 if (localReceiver != null)
@@ -269,12 +266,8 @@ namespace Orleans.ServiceBus.Providers
             }
 #endif
             // if we have a starting offset or if we're not configured to start reading from utc now, read from offset
-            if (!partitionSettings.Hub.StartFromNow ||
-#if NETSTANDARD
-                offset != PartitionReceiver.StartOfStream)
-#else
-                offset != EventHubConsumerGroup.StartOfStream) 
-#endif
+            if (!partitionSettings.Hub.StartFromNow || 
+                offset != EventHubConstants.StartOfStream)
             {
                 logger.Info("Starting to read from EventHub partition {0}-{1} at offset {2}", partitionSettings.Hub.Path, partitionSettings.Partition, offset);
             }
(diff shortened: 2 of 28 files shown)


src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -108,7 +112,15 @@ public virtual void Init(IProviderConfiguration providerCfg, string providerName
             adapterSettings = new EventHubStreamProviderSettings(providerName);
             adapterSettings.PopulateFromProviderConfig(providerConfig);
             hubSettings = adapterSettings.GetEventHubSettings(providerConfig, serviceProvider);
+#if NETSTANDARD
+            var connectionStringBuilder = new EventHubsConnectionStringBuilder(hubSettings.ConnectionString)
```
jason-bragg: I understand the need for these conditional computation blocks, but we should only use them if necessary.  ServiceBus libraries often add new capabilities prior to removing (deprecating) older ones.  With that in mind, it's possible that this pattern is already supported in the current nugets.  If it is, we can just move to the new pattern without the conditional code.  Please confirm that new patterns are not already supported in cases where we use conditional code.

galvesribeiro: Sure, will double check those.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -202,7 +219,12 @@ public Task MessagesDeliveredAsync(IList<IBatchContainer> messages)
 
                 // clear cache and receiver
                 IEventHubQueueCache localCache = Interlocked.Exchange(ref cache, null);
-                EventHubReceiver localReceiver = Interlocked.Exchange(ref receiver, null);
+#if NETSTANDARD
+                PartitionReceiver
```
jason-bragg: Would using statement to define type be cleaner?

galvesribeiro: Sure, will do that.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -224,30 +246,56 @@ public Task MessagesDeliveredAsync(IList<IBatchContainer> messages)
...
             if (partitionSettings.Hub.PrefetchCount.HasValue)
             {
                 consumerGroup.PrefetchCount = partitionSettings.Hub.PrefetchCount.Value;
             }
+#endif
             // if we have a starting offset or if we're not configured to start reading from utc now, read from offset
-            if (!partitionSettings.Hub.StartFromNow || offset != EventHubConsumerGroup.StartOfStream)
+            if (!partitionSettings.Hub.StartFromNow ||
+#if NETSTANDARD
+                offset != PartitionReceiver.StartOfStream)
```
jason-bragg: Please verify PartitionReceiver.StartOfStream maps to "-1" as EventHubConsumerGroup.StartOfStream used to.  This is a persisted value in some cases.

galvesribeiro: https://github.com/Azure/azure-event-hubs-dotnet/blob/8774bc4afbfbe44f1722ee648cbaca7ae07edede/src/Microsoft.Azure.EventHubs/PartitionReceiver.cs#L28

Yes it is. I confirmed before.

jason-bragg: Thanks.  Paranoia strikes deep :)

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
galvesribeiro: Done

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
galvesribeiro: They really changed. Everything that is on the old nugets that could be saved I did. They dont have those new patterns there even on the latest version of the old package...

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
galvesribeiro: @jason-bragg I'll revert this change now. I did the using but I found out that this Type is used later on in the same file and the `using` will not help since they use the same type but the property is in different places which is this case:

```
#if NETSTANDARD
                offset != PartitionReceiver.StartOfStream)
#else
                offset != EventHubConsumerGroup.StartOfStream) 
#endif
```
So, I prefer to stick with those `#if`s there and later on, when we decide to merge the trees we revisit it (which we will do anyway to remove the other `#if`s).

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -224,30 +246,56 @@ public Task MessagesDeliveredAsync(IList<IBatchContainer> messages)
...
+                offset = partitionInfo.LastEnqueuedOffset;
                 offsetInclusive = false;
                 logger.Info("Starting to read latest messages from EventHub partition {0}-{1} at offset {2}", partitionSettings.Hub.Path, partitionSettings.Partition, offset);
             }
-            return await consumerGroup.CreateReceiverAsync(partitionSettings.Partition, offset, offsetInclusive);
+#if NETSTANDARD
+            return client.CreateReceiver(partitionSettings.Hub.ConsumerGroup, partitionSettings.Partition, offset, offsetInclusive);
+#else
+            return await consumerGroup.CreateReceiverAsync(partitionSettings.Partition, offset, offsetInclusive); 
+#endif
```
jdom: Is CreateReceiver a non-blocking operation? Otherwise please use async version.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCheckpointer.cs
```diff
@@ -85,7 +94,11 @@ private Task Initialize()
         public void Update(string offset, DateTime utcNow)
         {
             // if offset has not changed, do nothing
-            if (string.Compare(entity.Offset, offset, StringComparison.InvariantCulture)==0)
+#if NETSTANDARD
+            if (string.Compare(entity.Offset, offset, StringComparison.Ordinal) == 0)
+#else
+            if (string.Compare(entity.Offset, offset, StringComparison.InvariantCulture) == 0) 
+#endif
```
jdom: IMO, just migrate here, don't use conditional compilation

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCheckpointer.cs
```diff
@@ -22,7 +26,12 @@ public class EventHubCheckpointer : IStreamQueueCheckpointer<string>
         /// <summary>
         /// Indicates if a checkpoint exists
         /// </summary>
-        public bool CheckpointExists => entity != null && entity.Offset != EventHubConsumerGroup.StartOfStream;
+        public bool CheckpointExists => entity != null && entity.Offset !=
+#if NETSTANDARD
+            PartitionReceiver.StartOfStream;
+#else
+            EventHubConsumerGroup.StartOfStream; 
+#endif
```
jdom: third time I see this, should you create a constant with the conditional compilation, and then just have these different portions of the code just point to our constant? It does make sense logically regardless of the different versions of the nuget, if this is not JUST an opaque constant to us, let's just hold on to it.

test/ServiceBus.Tests/CollectionFixtures.cs
```diff
@@ -0,0 +1,13 @@
+﻿using TestExtensions;
+using Xunit;
+
+namespace UnitTests
+{
+    // Assembly collections must be defined once in each assembly
+    [CollectionDefinition("DefaultCluster")]
+    public class DefaultClusterTestCollection : ICollectionFixture<DefaultClusterFixture> { }
```
jdom: Is this one used at all for this assembly?

vNext/test/ServiceBus.Tests/ServiceBus.Tests.csproj
```diff
@@ -0,0 +1,83 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jdom: This DLL should be added to vNext/src/Test.cmd

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -121,11 +127,11 @@ public Task Initialize(TimeSpan timeout)
             try
             {
                 var watch = Stopwatch.StartNew();
-                messages = (await receiver.ReceiveAsync(maxCount, ReceiveTimeout)).ToList();
+                messages = (await receiver.ReceiveAsync(maxCount, ReceiveTimeout))?.ToList();
```
galvesribeiro: Yes, I caugh null reference exceptions while porting and found out that yes, it can now return null. Here is the source https://github.com/Azure/azure-event-hubs-dotnet/blob/8774bc4afbfbe44f1722ee648cbaca7ae07edede/src/Microsoft.Azure.EventHubs/Amqp/AmqpPartitionReceiver.cs#L119

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -202,7 +219,12 @@ public Task MessagesDeliveredAsync(IList<IBatchContainer> messages)
 
                 // clear cache and receiver
                 IEventHubQueueCache localCache = Interlocked.Exchange(ref cache, null);
-                EventHubReceiver localReceiver = Interlocked.Exchange(ref receiver, null);
+#if NETSTANDARD
+                PartitionReceiver
+#else
+                EventHubReceiver  
+#endif
```
jason-bragg: Not that we really have coding standards, but in general we don't use var unless the type is obvious, for code readability reasons.  I'd understand ignoring that in this case, but it's nice to know @galvesribeiro is not being lazy :)

galvesribeiro: I always got someone going against `var` whenever I use it on my PRs :) Will change that.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
galvesribeiro: Yep, non-blocking.
https://github.com/Azure/azure-event-hubs-dotnet/blob/8774bc4afbfbe44f1722ee648cbaca7ae07edede/src/Microsoft.Azure.EventHubs/Amqp/AmqpEventHubClient.cs#L49

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -224,30 +246,56 @@ public Task MessagesDeliveredAsync(IList<IBatchContainer> messages)
...
+            EventHubClient client = EventHubClient.CreateFromConnectionString(connectionStringBuilder.ToString());
+#else
+            EventHubClient client = EventHubClient.CreateFromConnectionString(partitionSettings.Hub.ConnectionString, partitionSettings.Hub.Path); 
+
             EventHubConsumerGroup consumerGroup = client.GetConsumerGroup(partitionSettings.Hub.ConsumerGroup);
             if (partitionSettings.Hub.PrefetchCount.HasValue)
             {
                 consumerGroup.PrefetchCount = partitionSettings.Hub.PrefetchCount.Value;
             }
+#endif
```
galvesribeiro: It is one case where the API moved... Now it is a property on `PartitionReceiver` that I need to set before return it. Done

galvesribeiro:
Feedback implemented. Thanks

galvesribeiro:
@jason-bragg / @jdom Can someone point me out why this NonSilo test fail? Looks unrelated to the PR...

galvesribeiro:
@dotnet-bot test this please

galvesribeiro:
NVM... More issues on Jenkins...

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -224,30 +243,54 @@ public Task MessagesDeliveredAsync(IList<IBatchContainer> messages)
...
                     await client.GetPartitionRuntimeInformationAsync(partitionSettings.Partition);
-                offset = patitionInfo.LastEnqueuedOffset;
+                offset = partitionInfo.LastEnqueuedOffset;
                 offsetInclusive = false;
                 logger.Info("Starting to read latest messages from EventHub partition {0}-{1} at offset {2}", partitionSettings.Hub.Path, partitionSettings.Partition, offset);
             }
-            return await consumerGroup.CreateReceiverAsync(partitionSettings.Partition, offset, offsetInclusive);
+#if NETSTANDARD
+            PartitionReceiver receiver = client.CreateReceiver(partitionSettings.Hub.ConsumerGroup, partitionSettings.Partition, offset, offsetInclusive);
+            receiver.PrefetchCount = partitionSettings.Hub.PrefetchCount.Value;
```
jdom: watch out, as in the previous code, PrefetchCount might have been null. Look at how it was before:
```
	 if (partitionSettings.Hub.PrefetchCount.HasValue)
              {	
                  consumerGroup.PrefetchCount = partitionSettings.Hub.PrefetchCount.Value;		                  consumerGroup.PrefetchCount = partitionSettings.Hub.PrefetchCount.Value;
              }	
```

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubConstants.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    internal class EventHubConstants
+    {
+        public readonly static string StartOfStream =
+#if NETSTANDARD
+            Microsoft.Azure.EventHubs.PartitionReceiver.StartOfStream;
+#else
+            Microsoft.ServiceBus.Messaging.EventHubConsumerGroup.StartOfStream;
+#endif
+    }
```
jdom: Instead of creating a new class, I would create this as an internal field in `EventHubCheckpointer` class. Don't feel very strongly about it, so if there are no other changes required before the merge, feel free to ignore

galvesribeiro: It is used on 2 files... I think have this constant is OK in a separated class. However, I'll make it internal.

galvesribeiro: Duh, it is already internal, so unless there are other objections, I'll leave it as is.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
galvesribeiro: yeah, just confirmed that https://github.com/Azure/azure-event-hubs-dotnet/blob/8774bc4afbfbe44f1722ee648cbaca7ae07edede/src/Microsoft.Azure.EventHubs/PartitionReceiver.cs#L63

Added a if or leave the default (300) value.

PR #2640, opened 2017-01-26
Checkpointer tests wrong checkpointer connection string - bugfix

jason-bragg:
Somewhere along the line the azure storage connection string used in the checkpointer was changed to the eventhub connection string.
This change sets it back.


galvesribeiro:
@jason-bragg it is fixed on my PR already.

jason-bragg:
@galvesribeiro, understood, but i'd like to get this in for 1.4 which we're locking down this week.  It's unclear if #2637 will be in for 1.4..?  This is a small no-brainer fix, so no reason I can think of to not fast-path it in.

PR #2641, opened 2017-01-26
VSIX maintenance and fixes.

attilah:
- Revert icons for templates.
- Fix FSharp template to use Nugets.
- Remove unused references.


sergeybykov:
What bothers me a bit is that we have a bunch of different .ico files with the same name - Orleans.ico. I think we should rename them to CSharpInterfaces.ico, VBImplementation.ico, etc. to prevent future confusion.

attilah:
Ok, let me update it then. I'll fix the PNG names too in the gh-pages to keep them in sync.

PR #2642, opened 2017-01-26
GrainServices are now Started by the Silo on Startup

jamescarter-le:
Previously GrainServices which were added never called Startup, Reminder service worked because it is being created and started up discretely.

### DIFF
--- a/test/Tester/GrainServiceTests/CustomGrainService.cs
+++ b/test/Tester/GrainServiceTests/CustomGrainService.cs
@@ -47,19 +47,19 @@ namespace Tester
             
         }
 
-        private bool m_Started = false;
-        private bool m_StartedInBackground = false;
-        private bool m_Init = false;
+        private bool started = false;
+        private bool startedInBackground = false;
+        private bool init = false;
 
         public override Task Init(IServiceProvider serviceProvider)
         {
-            m_Init = true;
+            init = true;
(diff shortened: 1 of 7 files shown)


ReubenBond:
@jamescarter-le is `GrainService.Init(IServiceProvider serviceProvider)` ever used? I see no references to it.

jamescarter-le:
@ReubenBond Looks like I've missed that, I will add it to the pull request.

Thanks

test/Tester/GrainServiceTests/CustomGrainService.cs
```diff
@@ -33,6 +47,22 @@ public CustomGrainService(IGrainIdentity id, Silo silo, IGrainServiceConfigurati
             
         }
 
+        private bool m_Started = false;
```
ReubenBond: nit: the convention used in this codebase it to name private fields in lower-camelcase with no prefix

jamescarter-le:
Apologies I had followed that for the GrainServices, Resharper got involved and I forgot 👍

ReubenBond:
@jamescarter-le no worries - we ought to include some kind of code style settings file.
LGTM, we can merge when tests pass.

jamescarter-le:
Thanks @ReubenBond

Issue #2456, opened 2016-11-29
Revisit azure queue serialization.


jason-bragg:
To support external serializers (specifically json) we merged PRs “Modified stream types to not use fallback serializer and allow external #2330“ and “Azure queue stream provider silo to silo serialization bug fix. #2447”.

While these changes are diligent about verifying that default serialization continues to work to ensure backwards compatibility, the complexity introduced by having two versions of the types that can intermix is significant.  These changes may be perfectly safe, but reasoning about them for confidence in their current state and maintaining them in the future is non-trivial.
There is also a question as to whether the backwards compatibility pattern necessitates the upgrade of services reading from the queue prior to upgrading services writing to the queue.

I am sufficiently concerned about the above to request a reconsideration of this pattern.
In some other stream providers (MemoryStreamProvider, EventHubStreamProvider) we support application layer overrides of the serialization to allow developers to adapt the stream provider to their own data formats.  I am of the opinion that it is safer to introduce a similar solution to the Azure queue stream provider, with the default being the original data formatting, then introduce an alternative serializer that supports the custom serialization needs intended by the pull requests mentioned above.  This would strongly protect existing consumers using the azure queue stream provider, allow new consumers to opt into using the external serializer, and enable application developers to introduce their own application specific solutions as needed.

Tagging: @Carlm-MS, @gabikliot

gabikliot:
Of course I support NOT having 2 different types that do the same. I support consolidation. Not sure how I missed the original PR that introduced this duplication.

If you only need to support binary and json, then just add a config option to stream provider. If more formats are needed, provide a way to override.

Carlm-MS:
The initial commit was made to address a bug in the azure queue stream provider.  The bug was that the stream provider was serializing the event token and batch container using the fallback serializer.  This caused any types that implement the orleans serialization methods and/or are serialized using external serializers to throw exceptions if those types also are not decorated with the SerializableAttribute.  If they were decorated with the SerializableAttribute then they too would end up being serialized using the fallback serializer.

So - even though the provider was using the serialization manager for serialization, the result was that the payload was always using the fallback serializer for the entire payload -or- serialization exceptions were being thrown for every message.

The introduction of the V2 types were not in the initial commit but was added by request in order to maintain backward compatibility. If all of the tests use the V2 classes, then they will function normally.  There are a lot of tests around the V2 classes deserializing the original ones. The original tokens shouldn't be used in the V2 class because it'll then use the fallback serializer.

[...]

PR #2643, opened 2017-01-26
Added a data adapter to azure queue stream provider

jason-bragg:
The data adapter allows users to use the new v2 versions of sequence token and batch container without risking changes to the original versions, which are being used in production services.
The adapter also allows service developers to control how data is structured in azure queues, enabling the streaming of azure queue events of legacy systems or systems integrated with other non-orleans services.

This change addresses "Revisit azure queue serialization. #2456"

### DIFF
--- a/src/OrleansAWSUtils/Streams/SQSBatchContainer.cs
+++ b/src/OrleansAWSUtils/Streams/SQSBatchContainer.cs
@@ -17,7 +17,7 @@ namespace OrleansAWSUtils.Streams
     internal class SQSBatchContainer : IBatchContainer
     {
         [JsonProperty]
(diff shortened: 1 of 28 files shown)


attilah:
Do we really want to call it ```EventSequenceTokenV2 ``` ?

sergeybykov:
`EventSequenceTokenV2` was introduced in `https://github.com/dotnet/orleans/pull/2330`. This PR is just an improvement follow-up.

attilah:
The question still stands ;-)

sergeybykov:
Sure. I'm just trying to get this question out of scope of this PR and 1.4 release. ;-)

PR #2654, opened 2017-01-30
Reduce use of statics. Prep for non-static SerializationManager

ReubenBond:
Simple changes which were extracted from #2592.

The changes are very basic. Some of the changes only make sense if you evaluate them with the understanding that they are in preparation for making SerializationManager non-static, so keep that in mind.

Most of the changes are simply passing fixtures into test classes.

### DIFF
--- a/src/OrleansAzureUtils/Storage/Exceptions.cs
+++ b/src/OrleansAzureUtils/Storage/Exceptions.cs
@@ -6,7 +6,7 @@ using Orleans.Runtime;
 namespace Orleans.Storage
 {
     /// <summary>
-    /// Exception thrown when a storage provider detects an Etag inconsistency when attemptiong to perform a WriteStateAsync operation.
+    /// Exception thrown when a storage provider detects an Etag inconsistency when attempting to perform a WriteStateAsync operation.
     /// </summary>
     [Serializable]
     public class TableStorageUpdateConditionNotSatisfiedException : InconsistentStateException
@@ -27,7 +27,7 @@ namespace Orleans.Storage
             : base(errorMsg, storedEtag, currentEtag, storageException)
         {
             this.GrainType = grainType;
-            this.GrainId = grainId;
+            this.GrainId = grainId.ToKeyString();
             this.TableName = tableName;
         }
 
@@ -48,7 +48,7 @@ namespace Orleans.Storage
         /// <summary>
         /// Id of grain
         /// </summary>
-        public GrainReference GrainId { get; }
+        public string GrainId { get; }
 
         /// <summary>
         /// Type of grain that throw this exception
@@ -101,7 +101,7 @@ namespace Orleans.Storage
             : base(info, context)
         {
             this.GrainType = info.GetString("GrainType");
-            this.GrainId = GrainReference.FromKeyString(info.GetString("GrainId"));
+            this.GrainId = info.GetString("GrainId");
             this.TableName = info.GetString("TableName");
         }
 
@@ -111,7 +111,7 @@ namespace Orleans.Storage
             if (info == null) throw new ArgumentNullException(nameof(info));
 
             info.AddValue("GrainType", this.GrainType);
-            info.AddValue("GrainId", this.GrainId.ToKeyString());
+            info.AddValue("GrainId", this.GrainId);
             info.AddValue("TableName", this.TableName);
             base.GetObjectData(info, context);
         }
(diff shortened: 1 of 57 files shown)


ReubenBond:
@dotnet-bot test this please

src/OrleansCodeGenerator/SerializerGenerationManager.cs
```diff
@@ -117,7 +117,7 @@ internal bool RecordTypeToGenerate(Type t, Module module, Assembly targetAssembl
             return true;
         }
 
-        private static bool IsFieldInaccessibleForSerialization(Module module, Assembly targetAssembly, FieldInfo field)
```
benjaminpetit: >static [](start = 16, length = 6)

Really needed? Or just in preparation of a next PR?

ReubenBond: It's prep, see the call to `SerializationManager.HasSerialzier`? We know a field can be accessed for serialization if we already have a serializer for it

src/OrleansAzureUtils/Storage/Exceptions.cs
```diff
@@ -48,7 +48,7 @@ public class TableStorageUpdateConditionNotSatisfiedException : InconsistentStat
         /// <summary>
         /// Id of grain
         /// </summary>
-        public GrainReference GrainId { get; }
+        public string GrainId { get; }
```
sergeybykov: >public string GrainId { get; } [](start = 8, length = 30)

Why do we need to switch from `GrainReference` to `string` here? Seems unrelated to elimination of statics.

sergeybykov:
Except for the question above LGTM.

test/NonSiloTests/General/Identifiertests.cs
```diff
@@ -12,19 +12,20 @@
...
     {
         private readonly ITestOutputHelper output;
+        private readonly TestEnvironmentFixture environment;
         private static readonly Random random = new Random();
 
         class A { }
         class B : A { }
         
-        public Identifiertests(ITestOutputHelper output)
+        public Identifiertests(ITestOutputHelper output, TestEnvironmentFixture fixture)
```
jdom: same, no need to add the fixture to the constructor if you do not intend to access it from the tests

src/OrleansAzureUtils/Storage/Exceptions.cs
ReubenBond: It's because GrainReferences internally need a reference to `IRuntimeClient`, which BinaryFormatterSerializer can't provide. We need that info when calling `FromKeyString` below.

src/OrleansRuntime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -401,7 +402,7 @@ private SocketAsyncEventArgs GetSocketReceiveAsyncEventArgs(Socket sock)
             return saea.SocketAsyncEventArgs;
         }
 
-        private static SaeaPoolWrapper CreateSocketReceiveAsyncEventArgsPoolWrapper()
+        private SaeaPoolWrapper CreateSocketReceiveAsyncEventArgsPoolWrapper()
```
ReubenBond: It is, but not in this PR

test/NonSiloTests/General/InternalSerializationTests.cs
```diff
@@ -15,14 +15,17 @@ namespace UnitTests.Serialization
...
+    [Collection(TestEnvironmentFixture.DefaultCollection)]
     public class InternalSerializationTests
     {
+        private readonly TestEnvironmentFixture fixture;
+
         /// <summary>
         /// Initializes the system for testing.
         /// </summary>
-        public InternalSerializationTests()
+        public InternalSerializationTests(TestEnvironmentFixture fixture)
```
ReubenBond: These will all be used, since SerializationManager will be on that fixture

test/DefaultCluster.Tests/GrainReferenceTest.cs
```diff
@@ -117,33 +117,32 @@ private static void TestGrainReferenceSerialization(int id, bool resolveBeforeSe
...
-            Assert.Equal(grain,  otherGrain);  // "Deserialized grain reference equality is preserved"
-            int res = otherGrain.GetA().Result;
+            Assert.NotNull(other);
+            Assert.Equal(grain,  other);  // "Deserialized grain reference equality is preserved"
+            int res = other.GetA().Result;
             Assert.Equal(id,  res);  // "Returned values from call to deserialized grain reference"
         }
 
-        private static object DotNetSerialiseRoundtrip(object obj)
+        private static T DotNetSerialiseRoundtrip<T>(T obj)
```
ReubenBond: fixed

PR #2658, opened 2017-01-30
Fixed #2643 Added a data adapter to azure queue stream provider by @jason-gragg

sergeybykov:
This is an attempt to fix #2643, which has a build failure in vNext.


src/OrleansAzureUtils/Providers/Streams/AzureQueue/AzureQueueStreamProvider.cs
```diff
@@ -4,8 +4,16 @@ namespace Orleans.Providers.Streams.AzureQueue
 {
     /// <summary>
     /// Persistent stream provider that uses azure queue for persistence
+    /// WARNING: This version is maintained for compatability purposes.  New services should use AzureQueueStreamProviderV2 as it supports external serializers.
```
benjaminpetit: >WARNING: This version is maintained for compatability purposes.  New services should use AzureQueueStreamProviderV2 as it supports external serializers [](start = 8, length = 151)

Should we mark this class as obsolete?

ReubenBond:
@sergeybykov the exception in tests:
```
Orleans.CodeGenerator.CodeGenerationException : Found GenerateSerializerAttribute for type global::Orleans.Providers.Streams.AzureQueue.AzureQueueBatchContainerV2, but code could not be generated. Ensure that the type is accessible.
```

Ought to go away if build-time code generation is enabled for the Azure utils project

src/OrleansAzureUtils/OrleansAzureUtils.csproj
```diff
@@ -114,25 +114,11 @@
     <None Include="project.json" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <!-- Begin Orleans: Without these lines the project won't build properly -->
```
ReubenBond: This change should be un-reverted so that the code generator can operate at build-time. The runtime code generator cannot access those internal types.

src/OrleansAzureUtils/OrleansAzureUtils.csproj
```diff
@@ -114,25 +114,11 @@
...
-  <PropertyGroup>
-    <OrleansProjectType>Server</OrleansProjectType>
```
ReubenBond: these particular lines aren't needed, though - we don't use this property anymore

src/OrleansAzureUtils/OrleansAzureUtils.csproj
sergeybykov: I just pushed with that commit removed. :-)

Issue #2659, opened 2017-01-31
Override MethodId for generated GrainReferences


ReubenBond:
In order to support the versioning story which @benjaminpetit is working on, we should support explicitly specifying a method's `MethodId` so that it remains stable when it's renamed or when one of its parameter types are renamed.

The result would look like this:
```C#
[TypeCodeOverride(12324)] // We should rename this to InterfaceId
public interface IAggregatorGrain : IGrainWithGuidKey
{
  [MethodId(1)] // this would be added.
  Task<int> Sum(int[] values);
}
```

EDIT: ~~the~~ my intention is to add similar attributes for fields/properties to support stable serialization, too. So we could either have a bunch of attributes for types/methods/fields/assemblies, or we could have just one: `[Id(x)]` (which is what Bond call their attribute)

ReubenBond:
Implemented via #2660

PR #2660, opened 2017-01-31
Allow overriding MethodId using [MethodId(id)] on interface methods

ReubenBond:
Implements #2659.

This slightly improves the upgrade story by removing the restriction that method names and method parameter type names cannot be changed without breaking existing clients/silos.

Users can add the `[MethodId(x)]` attribute to methods to override the default algorithm for generating ids.

### DIFF
--- a/src/Orleans/CodeGeneration/GrainInterfaceUtils.cs
+++ b/src/Orleans/CodeGeneration/GrainInterfaceUtils.cs
@@ -126,6 +126,9 @@ namespace Orleans.CodeGeneration
         
         public static int ComputeMethodId(MethodInfo methodInfo)
         {
+            var attr = methodInfo.GetCustomAttribute<MethodIdAttribute>(true);
+            if (attr != null) return attr.MethodId;
+
             var strMethodId = new StringBuilder(methodInfo.Name + "(");
             ParameterInfo[] parameters = methodInfo.GetParameters();
             bool bFirstTime = true;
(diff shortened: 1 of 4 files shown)


sergeybykov:
LGTM.

src/Orleans/CodeGeneration/GrainInterfaceUtils.cs
```diff
@@ -126,6 +126,9 @@ public static bool IsStatelessWorker(MethodInfo methodInfo)
         
         public static int ComputeMethodId(MethodInfo methodInfo)
         {
+            var attr = methodInfo.GetCustomAttribute<MethodIdAttribute>(true);
+            if (attr != null) return attr.MethodId;
```
dVakulen: What will happen if attribute with same MethodId will be applied to multiple methods of interface?

benjaminpetit: Compilation will fail because we generate a switch case in the generated `GrainReference`.

Maybe we could add simple check in `CodeGeneratorCommon.GenerateGrainInterfaceAndMethodSwitch`, but I don't know how to surface the error, but I think we can merge as is.

ReubenBond: Precisely, @benjaminpetit. We could add this as a Roslyn Analyzer, but I figured that since it will fail compilation when the user adds a duplicate, it's safe to leave this unaddressed.

galvesribeiro: Sorry I'm late here... Just curious... It will fail compilation only while using build-time codegen... What about runtime codegen? Will we need to rely analyzers?

ReubenBond: @galvesribeiro it will fail the build at runtime ;)

PR #2665, opened 2017-02-02
Rename IRuntimeClient.Identity to CurrentActivationIdentity

ReubenBond:
`IRuntimeClient.Identity` is a confusing property, since the name implies it's the identity of the runtime client (which might be interpreted as client/silo id).


sergeybykov:
This whole thing makes little sense on the client side: `CurrentActivationAddress`, `CurrentSilo`, `Identity`. Maybe there's a better way to refactor the code to eliminate the confusion of the notions of "activation" and "current silo" on the client?

ReubenBond:
`CurrentActivationIdentity` is used solely for logging. I've removed `CurrentSilo` from `IRuntimeClient` in the latest & have made `CurrentActivationIdentity` more consistent between the outside & inside runtimes.

PR #2666, opened 2017-02-02
Add RequestContextUtil (prep for non-static SerializationManager)

ReubenBond:
Please review the latest commit only and of course avoid merging before the prior commits are merged (#2654 & #2665).

Since `RequestContext` has an implicit dependency on `SerializationManager`, changes will need to be made to support #467.

This PR injects a seemingly useless class, `RequestContextUtil` which will later hold an instance of non-static `SerializationManager` (as can be seen in #2592).

It also adds a client-side dependency on `Microsoft.Extensions.DependencyInjection`. This needs to happen sooner or later. DI is not exposed to consumers by this PR.

### DIFF
--- a/src/Orleans/Runtime/RequestContextUtil.cs
+++ b/src/Orleans/Runtime/RequestContextUtil.cs
@@ -1,0 +1,12 @@
+using System.Collections.Generic;
+
+namespace Orleans.Runtime
+{
+    public class RequestContextUtil
+    {
+        public Dictionary<string, object> Export()
+        {
+            return RequestContext.Export();
+        }
+    }
+}
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -1,6 +1,5 @@
 using System;
 using Orleans.GrainDirectory;
-using Orleans.Runtime;
 
 namespace Orleans
 {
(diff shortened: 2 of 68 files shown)


ReubenBond:
The goal here is to break up #2592. If this strategy is alright, then I'll do the same for `MessageFactory` and continue rebasing #2592 as chunks are split out. It's a tedious process, but hopefully it's easier to review these individual pieces.

jdom:
Hmmm, not too fond of the RequestContextUtil name. After I see what it does, I'll try to help with a better name

src/Orleans/Runtime/RequestContextUtil.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+
+namespace Orleans.Runtime
+{
+    public class RequestContextUtil
```
sergeybykov: I'm trying to wrap my head around the purpose of `RequestContextUtil`. What is `Export()` supposed to do eventually? Deep-copy `RequestContext`?

sergeybykov:
`RequestContextUtil` -> `RequestContextManager`?

src/Orleans/Runtime/RequestContextUtil.cs
ReubenBond: @sergeybykov that's right, so it needs a reference to SerializationManager. Currently the class is just a stub

ReubenBond:
Sergey's suggestion is good, imo

ReubenBond:
I think I'll close this PR and just pass SerializationManager around where needed, instead.

PR #2651, opened 2017-01-28
improve JournaledGrain API, add better tests/examples

sebastianburckhardt:
Based on discussion in #2598.

Added `RetrieveConfirmedEvents` method to JournaledGrain, which can read a desired segment of the event log.

Added many more tests/examples as the first step to better documentation. These demonstrate a diverse collection of scenarios, including several configuration options. 

They are located in folders:
- Tester.EventSourcing
- TestGrains.EventSourcing 
- TestGrainInterfaces.EventSourcing

Issue #2598, opened 2017-01-16
Provide docs for Log Consistency Providers and split test


galvesribeiro:
First of all, thank you @sebastianburckhardt for the great work!

After #1854 got merged I tried play with it but was unable to get it to work. I was unable to find among the tons of tests what are the files related to `OrleansEventSourcing` project. The configuration and requirements is what is hardest to find.

I would suggest a simple page with the minimum required configuration to get it to work. 

Also, would be good if that project has its own test project. I can do it myself if you point out what exactly is related to this project inside the test projects.

Thank you Sebastian! :)

sebastianburckhardt:
No doubt: the very next step is to add documentation and samples, both for the geo-distribution use case and for the simpler, non-geodistributed event sourcing. I will create a new PR shortly.

The only things already there right now are 
- a couple unit tests in Tester/EventSourcingTests. These demonstrate a few aspects of the API, but not much about the configuration.
- Several tests for log-consistency providers in TesterInternal/GeoClusterTests. These are unlikely to be of much interest unless you are planning to write your own log-consistency provider, which is probably not the first thing you want to try.

In a separate branch, I already have three geo-distributed samples, including information about how to configure and run them under Windows Azure. I will add them to the current Orleans samples.

galvesribeiro:
Thank you for promptly reply @sebastianburckhardt 

Yes, I think the first thing you can do, is just a very simple page that allow people that are already familiar with ES (including the old code) to configure Orleans to use your new provider mode.

After that, with time, you can elaborate more samples and a more complete doc page. That will unlock right now people (like me) to use it and provide feedback.

Also, I don't see in that code where we can replay events. For example, someone introduced a new grain type or an external system that behave as a projection. How can the events be replayed? That would be good to have in the "Hello world ES".

Again, thanks for your great work!

sebastianburckhardt:
Maybe I can give you enough information right here, to unblock you now before I get to writing the documentation.

To get the equivalent of the previous ES support in Orleans, for a class like in the JournaledPerson example at `TestGrains/EventSourcing/JournaledPersonGrain.cs`, you would use something like the following configuration

```csharp
    [StorageProvider(ProviderName = "WhateverStorageProviderYouWantToUse")]
    [LogConsistencyProvider(ProviderName = "LogStorage")]
    public class JournaledPersonGrain : JournaledGrain<PersonState,IPersonEvent>, IJournaledPersonGrain
```
with corresponding provider configuration

```xml
[...]

PR #2667, opened 2017-02-03
event-sourcing documentation

sebastianburckhardt:
A first draft of the documentation as promised in #2598.

This is not yet complete, as it does not yet discuss the samples which I will add soon.

Also, we most likely want to not merge this until #2651 is completed in case there are more changes to the API.

However, I thought it may be good to at least start this PR as it may be quite helpful to everyone involved to have some coherent documentation.

### DIFF
--- a/src/Documentation/Event-Sourcing/Configuration.md
+++ b/src/Documentation/Event-Sourcing/Configuration.md
@@ -1,0 +1,69 @@
+---
+layout: page
+title: Configuration
+---
+
+# Configuration
+
+## Configuring Project References
+
+### Grain Interfaces
+
+As before, interfaces depend only on the `Microsoft.Orleans.Core` package, because the grain interface is independent of the implementation. 
+
+### Grain Implementations
+
+JournaledGrains need to derive from `JournaledGrain<S,E>` or `JournaledGrain<S>`, which is defined in the `Microsoft.Orleans.EventSourcing` package. 
+
+### Log-Consistency Providers
+
+We currently include three log-consistency providers (for state storage, log storage, and custom storage). All three are contained in the `Microsoft.Orleans.EventSourcing` package as well. Therefore, all Journaled Grains already have access to those. For a description of what these providers do and how they differ, see [Included Log-Consistency Providers](LogConsistencyProviders.md).
+
+## Cluster Configuration
+
+Log-consistency providers are configured just like any other Orleans providers.
+For example, to include all three providers (of course, you probably won't need all three), add this to the `<Globals>` element of the configuration file:
+
+```xml
+<LogConsistencyProviders>
+  <Provider Type="Orleans.EventSourcing.StateStorage.LogConsistencyProvider" Name="StateStorage" />
+  <Provider Type="Orleans.EventSourcing.LogStorage.LogConsistencyProvider" Name="LogStorage" />
+  <Provider Type="Orleans.EventSourcing.CustomStorage.LogConsistencyProvider" Name="CustomStorage" />
+</LogConsistencyProviders>
+```
+
+The same can be achieved programmatically. Assuming the project contains the `Microsoft.Orleans.EventSourcing` package, and `config` is a `ClusterConfiguration` object:
+
+```csharp
+using Orleans.Runtime.Configuration; \\ pick up the necessary extension methods
+
+config.AddLogStorageBasedLogConsistencyProvider("LogStorage");
+config.AddStateStorageBasedLogConsistencyProvider("StateStorage");
+config.AddCustomStorageBasedLogConsistencyProvider("CustomStorage");
+```
+
+## Grain Class Attributes
+
+Each journaled grain class can be configured to use a log-consistency provider, and a storage provider of choice. These choices are independent.
+
+
+### LogConsistencyProvider Attributes
+
+By default, a journaled grain uses the built-in `StateStorage` provider. This provider persists the latest state snapshot, but not the events themselves. We can choose a different provider by adding a `[LogConsistencyProvider(ProviderName=...)]` attribute to the grain class, and give the name of the provider as configured by the Cluster Configuration. For example, the following attribute means that for `ChatGrain` instances, we use the `LogStorage` provider, which was configured to use `Orleans.EventSourcing.LogStorage.LogConsistencyProvider`. This provider always persists the entire event sequence (log).
+
+```csharp
+[LogConsistencyProvider(ProviderName = "LogStorage")]
+public class ChatGrain : JournaledGrain<XDocument, IChatEvent>, IChatGrain { ... }
+```
+
+### StorageProvider Attributes
+
+Some consistency providers (including both the LogStorage and StateStorage providers) in turn rely on a StorageProvider to communicate with storage. As usual, a default storage provider is used if none is explicitly specified. But we can add one. For example, we could specify
+
+```csharp
+[LogConsistencyProvider(ProviderName = "LogStorage")]
+[StorageProvider(ProviderName = "AzureBlobStorage")]
+public class ChatGrain : JournaledGrain<XDocument, IChatEvent>, IChatGrain { ... }
+```
+
+
--- a/src/Documentation/Event-Sourcing/Configuration.md
+++ b/src/Documentation/Event-Sourcing/Configuration.md
@@ -31,11 +31,10 @@ For example, to include all three providers (of course, you probably won't need 
   <Provider Type="Orleans.EventSourcing.CustomStorage.LogConsistencyProvider" Name="CustomStorage" />
 </LogConsistencyProviders>
 ```
-
 The same can be achieved programmatically. Assuming the project contains the `Microsoft.Orleans.EventSourcing` package, and `config` is a `ClusterConfiguration` object:
 
 ```csharp
-using Orleans.Runtime.Configuration; \\ pick up the necessary extension methods
+using Orleans.Runtime.Configuration; // pick up the necessary extension methods
 
 config.AddLogStorageBasedLogConsistencyProvider("LogStorage");
 config.AddStateStorageBasedLogConsistencyProvider("StateStorage");
@@ -44,21 +43,21 @@ config.AddCustomStorageBasedLogConsistencyProvider("CustomStorage");
 
 ## Grain Class Attributes
 
-Each journaled grain class can be configured to use a log-consistency provider, and a storage provider of choice. These choices are independent.
+Each journaled grain class must have a `LogConsistencyProvider` attribute to specify the log-consistency provider. Some providers additionally require a `StorageProvider` attribute.
 
 
 ### LogConsistencyProvider Attributes
 
-By default, a journaled grain uses the built-in `StateStorage` provider. This provider persists the latest state snapshot, but not the events themselves. We can choose a different provider by adding a `[LogConsistencyProvider(ProviderName=...)]` attribute to the grain class, and give the name of the provider as configured by the Cluster Configuration. For example, the following attribute means that for `ChatGrain` instances, we use the `LogStorage` provider, which was configured to use `Orleans.EventSourcing.LogStorage.LogConsistencyProvider`. This provider always persists the entire event sequence (log).
+To specify the log-consistency provider, add a `[LogConsistencyProvider(ProviderName=...)]` attribute to the grain class, and give the name of the provider as configured by the Cluster Configuration. For example:
 
 ```csharp
-[LogConsistencyProvider(ProviderName = "LogStorage")]
-public class ChatGrain : JournaledGrain<XDocument, IChatEvent>, IChatGrain { ... }
+[LogConsistencyProvider(ProviderName = "CustomStorage")]
+public class ChatGrain : JournaledGrain<XDocument, IChatEvent>, IChatGrain, ICustomStorage { ... }
 ```
 
 ### StorageProvider Attributes
 
-Some consistency providers (including both the LogStorage and StateStorage providers) in turn rely on a StorageProvider to communicate with storage. As usual, a default storage provider is used if none is explicitly specified. But we can add one. For example, we could specify
+Some log-consistency providers (including `LogStorage` and `StateStorage`) use a standard StorageProvider to communicate with storage. This provider is specified using a separate `StorageProvider` attribute, as follows:
 
 ```csharp
 [LogConsistencyProvider(ProviderName = "LogStorage")]
@@ -66,4 +65,14 @@ Some consistency providers (including both the LogStorage and StateStorage provi
 public class ChatGrain : JournaledGrain<XDocument, IChatEvent>, IChatGrain { ... }
 ```
 
+## Default Providers
 
+It is possible to omit the `LogConsistencyProvider` and/or the `StorageProvider` attributes, if a default is specified in the configuration. This is done by using the special name `Default` for the respective provider. For example:
+```xml
+<LogConsistencyProviders>
+  <Provider Type="Orleans.EventSourcing.LogStorage.LogConsistencyProvider" Name="Default" />
+</LogConsistencyProviders>
+<StorageProviders>
+  <Provider Type="Orleans.Storage.MemoryStorage" Name="Default" />
+</StorageProviders>
+```
--- a/src/Documentation/Event-Sourcing/GrainStateAPI.md
+++ b/src/Documentation/Event-Sourcing/GrainStateAPI.md
@@ -99,7 +99,7 @@ This guarantees that the given sequence of events is written to storage atomical
 
 ## Retrieving the Event Sequence
 
-The following method allows the application to retrieve a specified segment of the sequence of all confirmed events:
+The following method from the base `JournaledGrain` class allows the application to retrieve a specified segment of the sequence of all confirmed events:
 
 ```csharp
 Task<IReadOnlyList<EventType>> RetrieveConfirmedEvents(int fromVersion, int toVersion)
@@ -107,6 +107,15 @@ Task<IReadOnlyList<EventType>> RetrieveConfirmedEvents(int fromVersion, int toVe
 
 However, it is not supported by all log consistency providers. If not supported, or if the specified segment of the sequence is no longer available, a `NotSupportedException` is thrown. 
 
+To retrieve all events up to the latest confirmed version, one would call 
+```csharp
+await RetrieveConfirmedEvents(0, Version);
+```
+
+Only confirmed events can be retrieved: an exception is thrown if `toVersion` is larger than the current value of the property `Version`.
+
+Since confirmed events never change, there are no races to worry about, even in the presence of [multiple instances](MultiInstance.md) or [delayed confirmation](MultiVersion.md). However, in such situations, it is possible that the value of the property `Version` is larger by the time the `await` resumes than at the time `RetrieveConfirmedEvents` is called, so it may be advisable to save its value in a variable. See also the section on [Concurrency Guarantees](MultiVersion.md).
+ 
 
 
 
--- a/src/Documentation/Event-Sourcing/GrainStateAPI.md
+++ b/src/Documentation/Event-Sourcing/GrainStateAPI.md
@@ -1,0 +1,112 @@
+---
+layout: page
+title: JournaledGrain API
+---
+
+# JournaledGrain Basics
+
+Journaled grains derive from `<JournaledGrain<StateType,EventType>`, with the following type parameters:
+
+* The `StateType` represents the state of the grain. It must be a class with a public default constructor.  
+* `EventType` is a common supertype for all the events that can be raised for this grain, and can be any class or interface. 
+
+All state and event objects should be serializable (because the log-consistency providers may need to persist them, and/or send them in notification messages). 
+
+For grains whose events are POCOs (plain old C# objects),  `JournaledGrain<StateType>` can be used as a shorthand for `JournaledGrain<StateType,Object>`.
+
+## Reading the Grain State
+
+To read the current grain state, and determine its version number, the JournaledGrain has properties
+
+```csharp
+GrainState State { get; }
+int Version { get; }
+```
+
+The version number is always equal to the total number of confirmed events, and the state is the result of applying all the confirmed events to the initial state. The initial state, which has version 0 (because no events have been applied to it), is determined by the default constructor of the GrainState class.
+
+_Important:_ The application should never directly modify the object returned by `State`. It is meant for reading only. Rather, when the application wants to modify the state, it must do so indirectly by raising events.
+
+## Raising Events
+
+Raising events is accomplished by calling the `RaiseEvent` function. For example, a grain representing a chat can raise a `PostedEvent` to indicate that a user submitted a post:
+
+```csharp
+RaiseEvent(new PostedEvent() { Guid = guid, User = user, Text = text, Timestamp = DateTime.UtcNow });
+```
+
+Note that `RaiseEvent` kicks off a write to storage access, but does not wait for the write to complete. For many applications, it is important to wait until we have confirmation that the event has been persisted. In that case, we always follow up by waiting for `ConfirmEvents`:
+
+```csharp
+RaiseEvent(new DepositTransaction() { DepositAmount = amount, Description = description });
+await ConfirmEvents();
+```
+
+Note that even if you don't explicitly call `ConfirmEvents`, the events will eventually be confirmed - it happens automatically in the background. For more discussion on this topic, see [Immediate vs. Delayed Confirmation](MultiVersion.md).
+
+## State Transition Methods
+
+The runtime updates the grain state _automatically_ whenever events are raised. There is no need for the application to explicitly update the state after raising an event. However, the application still has to provide the code that specifies _how_ to update the state in response to an event. This can be done in two ways.
+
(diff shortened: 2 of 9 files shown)


sergeybykov:
I agree - let's finish #2651 first. It seems almost done to me.

src/Documentation/Event-Sourcing/Configuration.md
```diff
@@ -0,0 +1,69 @@
...
+  <Provider Type="Orleans.EventSourcing.StateStorage.LogConsistencyProvider" Name="StateStorage" />
+  <Provider Type="Orleans.EventSourcing.LogStorage.LogConsistencyProvider" Name="LogStorage" />
+  <Provider Type="Orleans.EventSourcing.CustomStorage.LogConsistencyProvider" Name="CustomStorage" />
+</LogConsistencyProviders>
+```
+
+The same can be achieved programmatically. Assuming the project contains the `Microsoft.Orleans.EventSourcing` package, and `config` is a `ClusterConfiguration` object:
+
+```csharp
+using Orleans.Runtime.Configuration; \\ pick up the necessary extension methods
```
jdom: use `//` for comments

src/Documentation/Event-Sourcing/Configuration.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+## Grain Class Attributes
+
+Each journaled grain class can be configured to use a log-consistency provider, and a storage provider of choice. These choices are independent.
+
+
+### LogConsistencyProvider Attributes
+
+By default, a journaled grain uses the built-in `StateStorage` provider. This provider persists the latest state snapshot, but not the events themselves. We can choose a different provider by adding a `[LogConsistencyProvider(ProviderName=...)]` attribute to the grain class, and give the name of the provider as configured by the Cluster Configuration. For example, the following attribute means that for `ChatGrain` instances, we use the `LogStorage` provider, which was configured to use `Orleans.EventSourcing.LogStorage.LogConsistencyProvider`. This provider always persists the entire event sequence (log).
```
jdom: Isn't this an odd default for people looking to use event sourcing?

src/Documentation/Event-Sourcing/Configuration.md
```diff
@@ -0,0 +1,69 @@
...
+
+```csharp
+[LogConsistencyProvider(ProviderName = "LogStorage")]
+public class ChatGrain : JournaledGrain<XDocument, IChatEvent>, IChatGrain { ... }
+```
+
+### StorageProvider Attributes
+
+Some consistency providers (including both the LogStorage and StateStorage providers) in turn rely on a StorageProvider to communicate with storage. As usual, a default storage provider is used if none is explicitly specified. But we can add one. For example, we could specify
```
jdom: which one is the default when not specified?

src/Documentation/Event-Sourcing/Configuration.md
jdom: btw, can we emulate storage provider's behavior and not provide a default that could turn into a nasty breaking change in the future? Basically, you MUST be explicit

src/Documentation/Event-Sourcing/GrainStateAPI.md
```diff
@@ -0,0 +1,112 @@
...
+```csharp
+protected override void TransitionState(State state, EventType @event)
+{
+   // code that updates the state
+}
+```
+
+The transition methods are assumed to have no side effects other than modifying the state object, and should be deterministic (otherwise, the effects are unpredictable).  If the transition code throws an exception, that exception is caught and included in a warning in the Orleans log, issued by the log-consistency provider.  
+
+When, exactly, the runtime calls the transition methods depends on the chosen log consistency provider and its configuration. It is best for applications not to rely on a particular timing, except when specifically guaranteed by the log consistency provider.
```
jdom: Does it mean that there could be race conditions if I raise an event, don't wait for it to be confirmed, and then start handling the next request?

src/Documentation/Event-Sourcing/GrainStateAPI.md
```diff
@@ -0,0 +1,112 @@
...
+
+This guarantees that the given sequence of events is written to storage atomically. Note that since the version number always matches the length of the event sequence, raising multiple events increases the version number by more than one at a time.
+
+
+## Retrieving the Event Sequence
+
+The following method allows the application to retrieve a specified segment of the sequence of all confirmed events:
+
+```csharp
+Task<IReadOnlyList<EventType>> RetrieveConfirmedEvents(int fromVersion, int toVersion)
```
jdom: this method exists in the JournaledGrain base class?

src/Documentation/Event-Sourcing/Configuration.md
galvesribeiro: Yeah that was my first question when I was asking for docs for the ES work... I was looking something more close to GetEventSource or StreamStone, which store events, and not the snapshot... That made me confuse...

src/Documentation/Event-Sourcing/GrainStateAPI.md
galvesribeiro: I think he added this recently and the PR wasn't merged yet. I asked for a way to get all the events (at least the confirmed ones) so I can replay them or republish them in a stream so other consumers can for example re-create existent (or create new) projections.

src/Documentation/Event-Sourcing/Configuration.md
sebastianburckhardt: StateStorage is currently the default, because it's automatic snapshotting is more robust (LogStorage becomes terribly large over time as the number of events grows).

src/Documentation/Event-Sourcing/Configuration.md
sebastianburckhardt: Yes, it is probably better to not provide a default here. An implicit choice is confusing. We can change that.

src/Documentation/Event-Sourcing/GrainStateAPI.md
sebastianburckhardt: Yes. See the documentation page on "immediate vs. delayed confirmation".

src/Documentation/Event-Sourcing/GrainStateAPI.md
sebastianburckhardt: Yes, I added that because @galvesribeiro asked for it.

Neftedollar: Will this method have  version with datetime or timespans or another timebased parameters?
like
```csharp
Task<IReadOnlyList<EventType>> RetrieveConfirmedEvents(DateTime fromVersion, DateTime toVersion)
```

sebastianburckhardt: Probably not, at least for now. 

There are a lot of queries that could potentially be useful... but we probably don't want to build that complexity into the basic API. Adding such support makes it more complex to implement new providers.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Orleans.EventSourcing.**StateStorage**.LogConsistencyProvider
+
+This provider stores *grain state snapshots*, using a standard storage provider that can be configured independently. 
+
+The data that is kept in storage is an object that contains both the grain state (as specified by the first type parameter to `JournaledGrain`) and some meta-data (the version number, amd a special tag that is used to avoid duplication of events when storage accesses fail).
+
+Since the entire grain state is read/written every time we access storage, this provider is not suitable for objects whose grain state is very large. 
+
+This provider does not support `RetrieveConfirmedEvents`: it cannot retrieve the events from storage because the events are not persisted.
```
sebastianburckhardt: The initial draft of the event sourcing API made some restrictive assumptions 

1. Grains do only one thing with the events: use them to update the grain state. 
2. The grain state and transition function change very rarely, e.g. due to bug fixes; in all those situations, the grain is deactivated before the change

However, I think this is a bit too narrowly conceived. For example, in the following situations, a grain may need to access the event sequence (parts of it or all of it):

- the grain is maintaining some set of views that can change dynamically. 
- the grain wants to react to each event in some special way (in addition to just updating the grain state). For example, it may want to forward the event to some client.

Neither of these can be done nicely using just the initial draft of the API, but can be done very cleanly if we additionally support the `RetrieveConfirmedEvents`. For example, the grain code can handle `OnStateChanged`, and then use `RetrieveConfirmedEvents` to get access to new events in the sequence. It all works very robustly together (e.g. no surprises due to races, because we validate that grain code accesses only the confirmed part of the log, which does not change). 


It is not difficult to implement `RetrieveConfirmedEvents` in the provider: the requested segment of the event sequence either exists (in which case we can fetch and return it) or it doesn't (in which case we throw `NotSupportedException`, as documented). For all true event sourcing solutions, the log exists. For the `StateStorage` style of provider, it does not.

I don't think it is a terribly leaky abstraction, its abstraction level matches the rest of the API. 

Of course it is leaky in the sense that it can expose limitations of the provider (e.g. the provider may have thrown away the log, or may take a long time fetching it from cold storage). But I don't think this is something we should try to avoid at all cost... it seems not reasonable to me to expect that all providers work perfectly with all applications. If we require that, then all providers have to be super-heavily engineered. For example, it would be impossible to use light-weight providers that take snapshots immediately, all the time, and discard the events quickly.

About the snapshotting: we probably need a separate discussion on that, and write some higher-quality provider that actually does it. I never quite understood the details of how this is supposed to work. How does the provider know when it has to replay the log because the transitions have changed? After it fails to successfully deserialize the snapshot?

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Orleans.EventSourcing.**LogStorage**.LogConsistencyProvider
+
+This provider stores *the complete event sequence as a single object*, using a standard storage provider that can be configured independently.
+
+The data that is kept in storage is an object that contains a `List<EventType> object`, and some meta-data (a special tag that is used to avoid duplication of events when storage accesses fail).
+
+This provider does support `RetrieveConfirmedEvents`. All events are always available and kept in memory.
```
sebastianburckhardt: Yes, this provider is not a great one. It is very similar to the initial draft of event sourcing in Orleans, where the grain state was defined as List<Event>.

The entire List<Event> is kept in memory because every time we read/write to/from storage, we have to read/write the entire list. 

 I think I say somewhere not to use this provider in production unless the number of events remains very small.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
```diff
@@ -0,0 +1,54 @@
...
+
+Since the whole event sequence is read/written every time we access storage, this provider is _not suitable for use in production_, unless the event sequences are guaranteed to remain pretty short. The main purpose of this provider is to illustrate the semantics of the event sourcing, and for samples/testing environments.
```
sebastianburckhardt: Agreed, as said in earlier comments, we probably DO need a better provider to get to the point where this is a "real" event sourcing solution.

sebastianburckhardt: Opened #2723 and assigned to myself.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
galvesribeiro: > About the snapshotting: we probably need a separate discussion on that, and write some higher-quality provider that actually does it. I never quite understood the details of how this is supposed to work. 

IMHO snapshot is not something that the provider must care about. The current snapshot is the current Grain State object with all the events already applied. If someone need to push this snapshot to somewhere else, just send the object to it. There is nothing the provider should do about it.

> After it fails to successfully deserialize the snapshot

Like I said, we shouldn't serialize the snapshot. It is a transient state in memory of all the events applied to it.

> How does the provider know when it has to replay the log because the transitions have changed?

The events should be replay before `OnActivateAsync()` and applied to the in-memory State object so it can build the current Snapshot of the state. 

Whenever an even is raised, it is applied by the provider into the State object, stored into the event store (not regular grain state storage) and optionally a callback implemented into the grain code can be fired (i.e. `OnEventRaised()`) so the grain can for example publish this event to a stream so other grains can consume and react to it (i.e. to generate projections).

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
galvesribeiro: > The entire List is kept in memory because every time we read/write to/from storage, we have to read/write the entire list.

The problem here is that you are using our traditional Grain State storage to store all events and keeping then in memory. This will definitively don't scale. What we need is to persist into a real Event Store (like GES or StreamStone), read the stream when activating, build the state project and don't keep the event list in memory. Whenever people need to read events (i.e. to replay/republish them) you provide a give API like `GetEvents(start, end)` do whatever they need, and discard it. Only the grain state is kept in memory as a snapshot of all events applied.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
galvesribeiro: We need the provider to be based on real event stores, not use current Orleans grain state...

sebastianburckhardt: I don't think it is a either/or decisions. Yes, of course we want to support event stores (#2711).

Still, I think it is possible, and would be quite nice, to support efficient event sourcing on top of standard storage providers (#2723). We may have to solve a few issues related to that.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
sebastianburckhardt: I think the term 'snapshot' can mean two different things:
- the latest grain state, which is a view obtained from applying all the events. This notion of snapshot is already built into the API.
- a performance optimization to reduce grain activation time, that caches some grain states (not necessarily the most recent version) in persistent storage, so it can replay from there instead of the beginning of time. This is not currently implemented by any of our providers, but can be added.

I think @jdom and I were referring to the latter.

galvesribeiro: > a performance optimization to reduce grain activation time, that caches some grain states (not necessarily the most recent version) in persistent storage, so it can replay from there instead of the beginning of time. This is not currently implemented by any of our providers, but can be added.

This in ES terms is called `Checkpoint`. That makes you persist a snapshot of the stream into a full object as a checkpoint so you can read from it and apply the following events. That is something IIRC the event stores (GES/SS) already support. If it doesn't, then it is a possible candidate to be stored in a grain state.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
galvesribeiro: I'm not sure it will be really efficient... The current storage provider save and read the whole state from storage every time. Event Stores read/write each record... By definition they are diff... Idk how you plan to get it working. Think that you can have hundreds of events per entity...

sebastianburckhardt: Yes, to make this work, we need to use the IStorageProvider differently than what it was originally intended for... if you look at IStorageProvider, it is essentially a key-value store, but special-cased to the scenario at hand, where (1) there is a single key per grain (derived in some way from Grain type , Grain Reference) and (2) the value is the grain state. 

But really, I think many of the providers don't care much about this specific structure... they could store any value for any key, and give us e-tag semantics. And that is good enough to implement event sourcing efficiently. 

Effectively we would be re-purposing the same storage providers so they store log segments and checkpoints for us, instead of just grain state.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
jdom: > the grain is maintaining some set of views that can change dynamically.

Typically replaying of the events does not happen at normal operation time. When you need to create a different view or process past events slightly differently, then that is typically done when the new version of the app is deployed. You won't go grain by grain as they get normally activated (you might though), but instead have a startup process that goes through the entire log and re-process it in some way... That's why you might need to tap into the specific storage semantics, instead of the narrow `ReadConfirmedEvents` that each grain would be able to access to once it gets activated.


> the grain wants to react to each event in some special way (in addition to just updating the grain state). For example, it may want to forward the event to some client.

Again, something that you can configure at the provider level to guarantee atomicity/transactions. Adding this method by itself can lead to hard to solve data consistency issues.

> About the snapshotting: we probably need a separate discussion on that, and write some higher-quality provider that actually does it. I never quite understood the details of how this is supposed to work. How does the provider know when it has to replay the log because the transitions have changed? After it fails to successfully deserialize the snapshot?

I guess this is partially answered by my first answer... this is not something that is typically done at runtime... You might have a deployment startup process that just deletes all the old snapshots for an entity and then starts replaying all events. We can help at the infrastructure level by adding the functions to each of the providers that allow snapshotting, by allowing reading the entire history, etc, but probably not by creating an abstraction for every provider, as they can all behave differently or be optimized for different needs, and hence the primary reason why people might choose to use one storage over the other. Also, it might be useful to build a mechanism to replay all events on the grains themselves, but at the same time be mindful that we are replaying the entire history so that we can optimize the access to storage. Nevertheless, I haven't thought about solutions to this in the context of Orleans yet. I'm only starting to look at what we have in our repo. Agreed that a separate discussion is needed

sebastianburckhardt: > Typically replaying of the events does not happen at normal operation time
I was originally assuming that as well, but I now think that it really depends on what kind of system you want to build. I don't think it's always true. More importantly, I don't think enforcing this makes anybody's life easier.

> Adding this method by itself can lead to hard to solve data consistency issues.
I dont' see why. The confirmed events never change.

> You won't go grain by grain as they get normally activated (you might though), but instead have a startup process that goes through the entire log and re-process

This seems like an advanced scenario. There is currently no way to write such a provider in a general way (since it requires knowing all the grains and how the grain state was changed).

I think we should support the simplest solution first: for each grain, when grain is loaded, restore from most recent checkpoint found (if possible) or otherwise (no checkpoint found or checkpoint is no longer valid due to some changes) replay entire log. I don't really see a problem with this.

The only thing needed to support this is some way to know if a checkpoint is invalid, e.g. because the grain state has changed (e.g. different fields), or the user has changed the transition function (e.g. fixed a bug). I am not sure there is an easy way to detect this automatically - clearly, a deserialization failure would indicate that the checkpoint is useless. Perhaps this is enough, as users can plug in their own deserialization code and use an explicit versioning mechanism.

galvesribeiro: > This seems like an advanced scenario. There is currently no way to write such a provider in a general way (since it requires knowing all the grains and how the grain state was changed).

I agree... Use cases for that, like @jdom mentioned, is if I add a new system that want to leverage all history of an already in place system. Today, what I do (since I'm using SS and not the storage provider from Orleans), what I do is make a console App that read whole stream of events up to a certain part directly from ATS with StreamStone, and publish then to a stream which the new system will consume and generate their projections.

Re. Checkpoints... I think at user level you should have a `WriteCheckpointAsync()` so the user decide when to create a checkpoint for a given stream. 

The major problem with the current storage APIs is that it is unreadable outside Orleans. And that is why you need to expose APIs like the `GetConfirmedEvents()` as I requested so we can support the scenario I mentioned here first. If we have a *open* storage, like SS does, or use GES, people can easily read whole stream and publishe somewhere else.  I'll post my API suggestion for ES storage in #2723 later on.

sebastianburckhardt: One thing that we can do pretty easily is to combine the approach for the ES storage interface (#2723) with  a configuration option for automatic checkpoints via IStorageProvider. 

Basically, the user (in addition to specifying an event store) also (optionally) specifies a standard storage provider, and gives a configuration option on how often they want check points, e.g. "CheckPointDistance=100".

We can then automatically save these checkpoints - these checkpoints then serve as optimizations to reduce load time, i.e. if not found or not deserializable, we just go back to replaying the full event sequence.

galvesribeiro: That is perfect usage for the `IStorageProvider` :D

I like it! Also, I would make the automatic checkpoint optional.

src/Documentation/Event-Sourcing/MultiVersion.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. Confirm all `RaiseEvent` calls using `ConfirmEvents` before the grain method returns.
+
+1. Make sure tasks returned by `RaiseConditionalEvent` complete before the grain method returns.
+
+1. Avoid  `[Reentrant]` or `[AlwaysInterleave]` attributes, so only one grain call can be processed at a time.
+
+If we follow these rules, it means that after an event is raised, no other grain code can execute until the event has been written to storage. Therefore, it is impossible to observe inconsistencies between the version in memory and the version in storage. While this is often exactly what we want, it also has some decided disadvantages.
+
+* if the connection to a remote cluster or to storage is temporarily interrupted, then the grain becomes unavailable: effectively, the grain cannot execute any code while it is stuck waiting to confirm the events, which can take an indefinite amount of time (the confirmation protocol never times out).
```
jdom: why wouldn't it time out? This is access to storage (at least in the ES scenario), there should be a way to timeout...

sebastianburckhardt: All the consistency providers automatically retry.

src/Documentation/Event-Sourcing/MultiVersion.md
```diff
@@ -0,0 +1,60 @@
...
+
+* when handling a lot of of updates, confirming them one at a time can be very inefficient.
```
jdom: I'd add: when handling a lot of of updates, confirming them one at a time can be very inefficient *if the individual grain instance is expected to have high throughput.*

src/Documentation/Event-Sourcing/MultiVersion.md
```diff
@@ -0,0 +1,60 @@
...
+```
+Also, since the state returned by the `State` property does not include the effect of unconfirmed events, there is an alternative property 
+
+```csharp
+StateType TentativeState { get; }
+```
+
+which returns a "tentative" state, obtained from "State" by applying all the unconfirmed events. The tentative state is essentially a "best guess" at what will likely become the next confirmed state, after all unconfirmed events are confirmed. However, there is no guarantee that it actually will, because the grain may fail, or because the events may race against other events and lose, causing them to be canceled (if they are conditional) or appear at a later position in the sequence than anticipated (if they are unconditional). 
+
+Just like reacting to state changes via `OnStateChanged`, grains can react to changes in the tentative state:
```
jdom: I would combine the `Subscribing to Changes` section from the previous doc with this one, as they talk exactly of the same thing, except with the tentative state. Also, the `Concurrency` section below is extremely important for the `OnStateChanged` method that was explained in the previous document, so having them side by side makes more sense. You could just link from the previous doc to this one so that the reader could navigate in any way.

src/Documentation/Event-Sourcing/Configuration.md
sebastianburckhardt: I improved this section - it describes the configuration mechanism first, then how to specify defaults separately.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
```diff
@@ -0,0 +1,54 @@
...
+# Built-In Log-Consistency Providers
+
+The `Microsoft.Orleans.EventSourcing` package includes several log-consistency providers that cover basic scenarios suitable to get started, and allow some extensibility.
+
+
+### Orleans.EventSourcing.**StateStorage**.LogConsistencyProvider
+
+This provider stores *grain state snapshots*, using a standard storage provider that can be configured independently. 
+
+The data that is kept in storage is an object that contains both the grain state (as specified by the first type parameter to `JournaledGrain`) and some meta-data (the version number, amd a special tag that is used to avoid duplication of events when storage accesses fail).
```
sebastianburckhardt: ok. 

I have added a paragraph with a bit more discussion on RetrieveConfirmedEvents useage.

src/Documentation/Event-Sourcing/LogConsistencyProviders.md
sebastianburckhardt: I think this discussion is now tracked in #2711.

src/Documentation/Event-Sourcing/MultiVersion.md
sebastianburckhardt: I clarified the text.

src/Documentation/Event-Sourcing/MultiVersion.md
sebastianburckhardt: Yes.

src/Documentation/Event-Sourcing/MultiVersion.md
sebastianburckhardt: I will put these three subsections into a separate page.

src/Documentation/Event-Sourcing/Configuration.md
jdom: I'm pretty sure that is not how storage providers work. I think we use the name "Default" but it has no special meaning to Orleans, you still need to explicitly mark your grain to use that storage provider. Log providers should probably do the same

---
In reply to: [101595540](https://github.com/dotnet/orleans/pull/2667#discussion_r101595540) [](ancestors = 101595540)

Issue #2668, opened 2017-02-03
Bug: Azure Blob Storage provider leaves grain in inconsistent state after ClearStateAsync is called and loses first follow-up write of the state to the same grain in certain conditions


centur:
The bug is around Etag being set on ClearStateAsync call and it's a bit tricky to reproduce (code in provider is a bit convoluted). The key point is to re-use grain instance ( while it's still in memory

Reproduction steps:
1. Create a simple grain with Write\Clear state calls (see DebugGrain in the end)

2. Invoke it like this:
```C#
var grain = GrainClient.GrainFactory.GetGrain<IDebugGrain>(DateTime.UtcNow.Ticks.ToString());

await grain.Write("First write is sucessful");
var isCleaned = await grain.CleanState();

await grain.Write("Second write - doomed to be lost");
var secondData = await grain.Read(); // This reads because grain is in memory =) but there is no state saved
```

What happens is actually an interesting thing ( this why we were missing this bug for a while in our custom copy of BlobStorageProvider):

1. Empty state read, `grainState.Etag == null`
2. First payload is written,  `grainState.Etag == <BlobEtag>`
3. First payload is cleared, `grainState.Etag == <BlobEtag>` HERE is where our Etag is invalid.
4. Second payload is trying to be written. **Write FAILS** because there is an incorrect Etag (anyone can check the blob storage - there will be NO data with `Second write - doomed to be lost` payload )
This happens because of [AzureBlobStorage.cs: L248](https://github.com/dotnet/orleans/blob/5ede8d848a469adcb36846f607aff80b1cda205a/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs#L248)
5. Consequent READ succeeds because  the grain activation is still in memory (this is how bug eluded us for sooo long).

What's also interesting is the code in the next write which fails to write BUT fixes the state. Here is the current method annotated with interesting comments

```C#
   public async Task WriteStateAsync(string grainType, GrainReference grainId, IGrainState grainState)
        {
            var blobName = GetBlobName(grainType, grainId);
            try
            {
                if (this.Log.IsVerbose3) this.Log.Verbose3((int)AzureProviderErrorCode.AzureBlobProvider_Storage_Writing, "Writing: GrainType={0} Grainid={1} ETag={2} to BlobName={3} in Container={4}", grainType, grainId, grainState.ETag, blobName, container.Name);

                var json = JsonConvert.SerializeObject(grainState.State, jsonSettings);

[...]

PR #2669, opened 2017-02-03
Fix: assign correct value to grainState.Etag when ClearStateAsync is called

centur:
Fixes issue #2668


sergeybykov:
Thank you, @centur!

Issue #1352, opened 2016-01-29
Generic Grain Methods


Joshua-Ferguson:
I created an Orleans grain with a generic method call and I noticed that codegen failed. I'm not sure if it should fail on generics (intentional for performance reasons, or user error?). I created a [PR](https://github.com/Joshua-Ferguson/orleans/pull/1) on a fork to show the changes I made to get generics working. But I'm not sure the changes are made are the best way to do it (really just wanted to be able to test that it could be done). I got the type info < T > passed where I could and used < dynamic > everywhere else.

```
public interface ISecretNinjaTestGrain : IGrainWithIntegerKey
{
    Task<int> FakeGeneric<T>();
    Task<int> GenericArg<T>(T data);
    Task<T> SimpleGeneric<T>();
    Task<int> SimpleGenericInteger<T>();
    Task<int> SimpleInteger();
    Task<T> VeryGeneric<T>(T data);
    Task<string> MultiGeneric<T, T2, T3>(T t, T2 t2, T3 t3);
    Task<T2> BreakDance<T, T2, T3>(T t, T2 t2, T3 t3);
    Task<T> LessSimpleGeneric<T>(T one, T two);
}

public class SecretNinjaTestGrain : Grain, ISecretNinjaTestGrain
{
    public Task<int> SimpleInteger() => Task.FromResult(3);
    public Task<int> SimpleGenericInteger<T>() => Task.FromResult(3);
    public Task<T> SimpleGeneric<T>() => Task.FromResult(default(T));
    public Task<T> LessSimpleGeneric<T>(T one, T two) => Task.FromResult(default(T));
    public Task<int> FakeGeneric<T>() => Task.FromResult(3);
    public Task<int> GenericArg<TDifferent>(TDifferent data) => Task.FromResult(3);
    public Task<T> VeryGeneric<T>(T data) => Task.FromResult(data);
    public Task<string> MultiGeneric<T, T2, T3>(T t, T2 t2, T3 t3) => Task.FromResult($"{t}{t2}{t3}");
    public Task<T2> BreakDance<T, T2, T3>(T t, T2 t2, T3 t3) => Task.FromResult(default(T2));
}
```

I was able to make calls to all these methods.

Joshua-Ferguson:
I was a little premature. The test I wrote only looked at the Task object for LessSimpleGeneric and BreakDance and not the return value, so they throw a null object exception if you try to get the Result. But MultiGeneric works just fine.

[...]

PR #2670, opened 2017-02-03
Support for grains with generic methods

ReubenBond:
This PR adds support for generic methods on grains in a pay-for-what-you-use fashion, so any additional cost is only incurred for generic methods. The cost should be quite low anyway because the invoker stub is generated using (simple) IL.

Supports methods with constraints, overloads, generic methods inside generic grains, upcasting arguments, grain observers, and grain extensions (last one not tested). We can address anything I missed as they come.

This is what a generated `GrainReference` method looks like for a generic method:
```C#
public global::System.Threading.Tasks.Task<global::System.Type[]> @GetTypesInferred<T, U>(T @t, U @u, global::System.Int32 @v)
{
    return base.@InvokeMethodAsync<global::System.Type[]>(-1221167187, new global::System.Object[]{typeof (T), typeof (U), @t, @u, @v});
}
```
Notice that it passes the static types of the arguments in the arguments array before the actual arguments.

The other end is more complicated: in the `IGrainMethodInvoker`, we add a static `GenericMethodInvoker` field for each generic method and use that for invoking the methods using runtime codegen (IL). So a snippet of the generated invoker looks like this:
```C#
switch (interfaceId)
{
    case 447824211:
        switch (methodId)
        {
            case 1337175495:
                return GetTypesExplicitT_U_V.@Invoke(@grain, arguments);
            case -105247224:
                return GetTypesInferredT_U_V.@Invoke(@grain, arguments);
            case -1221167187:
                return GetTypesInferredT_U.@Invoke(@grain, arguments);
```

With the fields declared something like this:
```C#
private static readonly global::Orleans.CodeGeneration.GenericMethodInvoker GetTypesExplicitT_U_V = new global::Orleans.CodeGeneration.GenericMethodInvoker(typeof (global::Tester.CodeGenTests.IGrainWithGenericMethods), "GetTypesExplicit", 3);
```

We use these fields to ensure that everything is as efficient as can be: no additional allocations are made on the hot path (any allocs are made only once).

The generated IL for the invokers looks like this:
```
Ldarg_0 // Load the grain argument
Castclass <grain interface type>

// foreach arg:
Ldarg_1 // Load the arguments array
Ldc_I4 <arg index + number of type params>
Ldelem_Ref // Load the argument as a reference (since it's an object array)
CastClass <argument type> // Or Unbox_Any <type> if it's a value type.

Call <concrete method>

// If return type is void
Call TaskDone.Done
// Else if return type is Task
Call PublicOrleansTaskUtils.Box(Task)
// Else if return type is Task<T>
Call PublicOrleansTaskUtils.Box(Task<T>)

Ret
```

Thanks, @Joshua-Ferguson, for your PR which I took inspiration from to create this :) You led the way. I had figured it was too difficult to implement and just not worth it.

Implements #1352
cc @@marianogenovese & @centur (since you both asked about this functionality recently)

### DIFF
--- a/src/Orleans/CodeGeneration/GenericMethodInvoker.cs
+++ b/src/Orleans/CodeGeneration/GenericMethodInvoker.cs
@@ -1,0 +1,164 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Orleans.Runtime;
+using Orleans.Serialization;
+
+namespace Orleans.CodeGeneration
+{
+    public class GenericMethodInvoker : IEqualityComparer<object[]>
+    {
+        private static readonly ConcurrentDictionary<Type, MethodInfo> BoxMethods = new ConcurrentDictionary<Type, MethodInfo>();
+        private static readonly Func<Type, MethodInfo> CreateBoxMethod = GetBoxMethod;
+        private static readonly MethodInfo GenericMethodInvokerDelegateMethodInfo =
+            TypeUtils.Method((GenericMethodInvokerDelegate del) => del.Invoke(null, null));
+        private static readonly ILFieldBuilder FieldBuilder = new ILFieldBuilder();
+
+        private readonly MethodInfo genericMethodInfo;
+        private readonly Type grainInterfaceType;
+        private readonly int typeParameterCount;
+
+        private readonly ConcurrentDictionary<object[], GenericMethodInvokerDelegate> invokers;
+        private readonly Func<object[], GenericMethodInvokerDelegate> createInvoker;
+
+        delegate Task<object> GenericMethodInvokerDelegate(IAddressable grain, object[] arguments);
+
+        public GenericMethodInvoker(Type grainInterfaceType, string methodName, int typeParameterCount)
+        {
+            this.grainInterfaceType = grainInterfaceType;
+            this.typeParameterCount = typeParameterCount;
+            this.invokers = new ConcurrentDictionary<object[], GenericMethodInvokerDelegate>(this);
+            this.genericMethodInfo = GetMethod(grainInterfaceType, methodName, typeParameterCount);
+            this.createInvoker = this.CreateInvoker;
+        }
+        
+        public Task<object> Invoke(IAddressable grain, object[] arguments)
+        {
+            var invoker = this.invokers.GetOrAdd(arguments, this.createInvoker);
+            return invoker(grain, arguments);
+        }
+
+        private GenericMethodInvokerDelegate CreateInvoker(object[] arguments)
+        {
+            // First, create the concrete method which will be called.
+            var typeParameters = arguments.Take(this.typeParameterCount).Cast<Type>().ToArray();
+            var concreteMethod = this.genericMethodInfo.MakeGenericMethod(typeParameters);
+
+            // Next, create a delegate which will call the method on the grain, pushing each of the arguments,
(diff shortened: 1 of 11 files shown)


src/Orleans/Serialization/IlDelegateBuilder.cs
```diff
@@ -81,6 +77,62 @@ public void LoadArgument(ushort index)
...
+                    this.il.Emit(OpCodes.Ldc_I4_5);
+                    break;
+                case 6:
+                    this.il.Emit(OpCodes.Ldc_I4_6);
+                    break;
+                case 7:
+                    this.il.Emit(OpCodes.Ldc_I4_7);
+                    break;
+                case 8:
+                    this.il.Emit(OpCodes.Ldc_I4_8);
```
sergeybykov: Is this because these opcodes are faster than `Emit(OpCodes.Ldc_I4_S, (byte) value)`?

src/Orleans/CodeGeneration/GenericMethodInvoker.cs
```diff
@@ -0,0 +1,164 @@
...
+            throw new ArgumentException($"Could not find Box method for type {returnType}");
+        }
+
+        bool IEqualityComparer<object[]>.Equals(object[] x, object[] y)
+        {
+            if (ReferenceEquals(x, y)) return true;
+            if (ReferenceEquals(x, null)) return false;
+            if (ReferenceEquals(null, y)) return false;
+
+            if (x.Length < this.typeParameterCount || y.Length < this.typeParameterCount) return false;
```
sergeybykov: >this.typeParameterCount [](start = 27, length = 23)

I don't understand what `this.typeParameterCount` has to do with the equality check for x vs. y. I would expect the condition to be `x.Length != y.Length` instead.

src/Orleans/CodeGeneration/GenericMethodInvoker.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        int IEqualityComparer<object[]>.GetHashCode(object[] obj)
+        {
+            if (obj == null || obj.Length == 0) return 0;
+            unchecked
+            {
+                // Only consider the type parameters.
+                var result = 0;
+                for (var i = 0; i < this.typeParameterCount && i < obj.Length; i++)
```
sergeybykov: >this.typeParameterC [](start = 36, length = 19)

Similarly here - why `this.typeParameterCount`?

src/Orleans/Serialization/IlDelegateBuilder.cs
ReubenBond: Yep, and they're smaller. It's just an optimization, so I can remove it if preferred.

src/Orleans/CodeGeneration/GenericMethodInvoker.cs
ReubenBond: In practice we expect x & y to have the same lengths, but this equality comparison is only checking the prefix of those arrays, because the intention is to check the type parameters only & not the entire contents. The array is specified as  type params followed by method arguments

src/Orleans/CodeGeneration/GenericMethodInvoker.cs
ReubenBond: If we included other values in our comparison, the hash code would be different for almost every call to a method, so we would generate IL very frequently. The goal is to generate IL once for each variant of a generic method.

ReubenBond:
Added more comments and gave the code a minor refactor for clarity.

PR #2671, opened 2017-02-03
Add three multi-cluster samples 

sebastianburckhardt:
These are the three samples that we have had out for a while at
https://github.com/sebastianburckhardt/orleans/tree/geo-samples

Since we now have all the functionality in master, I think we should move these samples into the Samples/ folder so they are in the same place as all the other ones.

All three samples are based originally on AzureWebSample, but run three Orleans clusters (the VS solution contains three Azure deployments). They can be run directly from VisualStudio, using the Azure Simulator, and will display three different windows, one per deployment. Users can then interact with these windows and observe what happens.

The samples are:

- **HelloGeoSample** demonstrates [GlobalSingleInstance] and [OneInstancePerCluster].

- **ReplicatedChatGrainSample** demonstrates multi-instance JournaledGrain, supporting disconnected operation.

- **ReplicatedEventSample** demonstrates two different types of clusters (front, back), with a replicated JournaledGrain that is modified in the back cluster, and has notifications automatically sent to the front clusters.


sergeybykov:
The folder for ReplciatedChatGrainSample is names QueuedGrainSample. We better avoid the possible confusion and rename the folder.

sebastianburckhardt:
Yes, good catch.
I already renamed the project files and namespaces, but forgot the folder!

sergeybykov:
>add NuGet.exe to samples 

Nuget.exe shouldn't and doesn't need to be checked in.

sergeybykov:
Nor should orleans.codegen.cs files be there. They need to be referenced in the .csproj files though.

sebastianburckhardt:
If I don't check in NuGet.exe and orleans.codegen.cs, the solutions do not compile. Perhaps there is some other way, but I don't know it (I remember trying  a lot of other things till it finally worked).

The content of the orleans.codegen.cs files does not matter since they are overwritten during compilation and ignored by .gitignore.  But compilation fails if they are missing. Therefore, I check in an empty file.

attilah:
Maybe you're using an outdated Orleans.SDK.targets file, I've changed it to DO NOT need these empty files or be part of the solution. We're adding them to compilation dynamically upon task execution.

sergeybykov:
@dotnet-bot test this please.

jdom:
I'll download this branch and submit a PR for you with Sergey's suggestions

jdom:
Created https://github.com/sebastianburckhardt/orleans/pull/10 for it.
Please do not merge this. After the PR is applied, we can **squash** this, but please do not merge, otherwise there will be big binaries added and removed that will remain in the git history for the end of times

sebastianburckhardt:
Looks great. I tried all the samples, they work for me. Thanks for cleaning this up!

Samples/ReplicatedChatGrainSample/QueuedGrainSample.Grains/ChatGrain.cs
```diff
@@ -0,0 +1,83 @@
...
+using ReplicatedChatGrainSample.Interfaces;
+using Orleans;
+using Orleans.Providers;
+using Orleans.Concurrency;
+using Orleans.EventSourcing;
+
+namespace ReplicatedChatGrainSample.Grains
+{
+    [Serializable]
+    public class ChatState
```
jdom: Does this class even need to be serializable? Because I tried locally by removing this attribute and it worked. If it doesn't need to be, I can remove the attribute, as it would be less confusing... I wasn't expecting this to have to be serializable, unless the provider implements snapshotting or something like that.
I'm in the process of doing more cleaning, so I can remove it myself if you say this isn't needed

jdom:
Created another PR with some more cleanup tasks: https://github.com/sebastianburckhardt/orleans/pull/11

PR #2675, opened 2017-02-06
Extract MessageFactory from Message class.

ReubenBond:
Prep for non-static SerializationManager.

Purpose is to give ourselves a way to inject `SerializationManager`/`RequestContextUtil` when creating messages. We could add an extra field to `Message`, but I believe it's best to keep POCOs as data-only (no services), since that makes serialization/deserialization obvious.

Only the last commit is unique to this PR: the others have their own PRs.


sergeybykov:
@ReubenBond Please rebase. Otherwise, it's ready to be merged.

ReubenBond:
rebased

PR #2678, opened 2017-02-06
DynamoDB silo identity fix

chrisbrook83:
(@galvesribeiro - as per chat on Gitter)

When using DynamoDB for the membership store, I was seeing `ConditionalCheckFailedException` errors when trying to update the `IAmAlive` field in the database. 

This was because the key being used to locate the record always had a value of "-0-0" rather than the expected "_IP-Port-Generation_". This was because it was being built using the `Address`, `Port`, and `Generation` values on the `SiloInstanceRecord` object, but on this code path these values are not assigned. Only the `SiloIdentity` property gets set by the `MembershipOracle` (see `OnIAmAliveUpdateInTableTimer`).

I have added new unit tests for the `GetKeys` call on `SiloInstanceRecord`, and fixed the issue.

I also added integration tests for all the membership stores, to exercise `UpdateIAmAlive`, as it looked like this was not being tested. **Note, I've not been able to run all these tests, as I don't have the various databases set up locally**.

### DIFF
--- a/test/AWSUtils.Tests/AWSUtils.Tests.csproj
+++ b/test/AWSUtils.Tests/AWSUtils.Tests.csproj
@@ -129,6 +129,7 @@
     <Compile Include="CollectionFixtures.cs" />
     <Compile Include="LivenessTests.cs" />
     <Compile Include="MembershipTests\DynamoDBMembershipTableTest.cs" />
+    <Compile Include="MembershipTests\SiloInstanceRecordTests.cs" />
     <Compile Include="Reminder\DynamoDBRemindersTableTests.cs" />
     <Compile Include="StorageTests\AWSTestConstants.cs" />
     <Compile Include="StorageTests\Base_PersistenceGrainTests_AWSStore.cs" />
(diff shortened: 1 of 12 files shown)


test/AWSUtils.Tests/AWSUtils.Tests.csproj
```diff
@@ -129,6 +129,7 @@
     <Compile Include="CollectionFixtures.cs" />
     <Compile Include="LivenessTests.cs" />
     <Compile Include="MembershipTests\DynamoDBMembershipTableTest.cs" />
+    <Compile Include="MembershipTests\SiloInstanceRecordTests.cs" />
```
galvesribeiro: Please add this test into the `vNext/test/AWSUtils.Tests`... It is a link to this file so the test can run in .net core. I've added it here and it pass without problem.

galvesribeiro:
@chrisbrook83 LGTM. Thanks for doing that. I've cloned and ran all tests for all the providers affected by this change in containers. All they pass. 

The only comment is to add your new test on `AWSUtils.Test` at the vNext solution.  

Good work!

Let's wait someone from the core team to look at it and merge if there is no objections.

Thanks!

test/AWSUtils.Tests/AWSUtils.Tests.csproj
chrisbrook83: All done... Thanks - I'd not spotted that other sln

sergeybykov:
Thank you, @chrisbrook83!

PR #2679, opened 2017-02-06
DynamoDB Gateway Provider fix: was including non-gateway silos

chrisbrook83:
The DynamoDB gateway provider was not filtering out silos with a proxy-port of 0.

All other providers only consider Active silos, sharing the current DeploymentId, with a non-zero proxy port (e.g. https://github.com/dotnet/orleans/blob/master/src/OrleansAzureUtils/Storage/OrleansSiloInstanceManager.cs#L235)

Added test coverage and fixed using additional predicate on query.

I've not been able to re-run all the membership provider tests for other providers locally, but have examined the code for all of them, and they all appear to include this correct check.

### DIFF
--- a/src/OrleansAWSUtils/Membership/DynamoDBGatewayListProvider.cs
+++ b/src/OrleansAWSUtils/Membership/DynamoDBGatewayListProvider.cs
@@ -45,9 +45,15 @@ namespace Orleans.Runtime.MembershipService
             var expressionValues = new Dictionary<string, AttributeValue>
             {
                 { $":{SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME}", new AttributeValue(deploymentId) },
-                { $":{SiloInstanceRecord.STATUS_PROPERTY_NAME}", new AttributeValue { N =  INSTANCE_STATUS_ACTIVE } }
+                { $":{SiloInstanceRecord.STATUS_PROPERTY_NAME}", new AttributeValue { N = INSTANCE_STATUS_ACTIVE } },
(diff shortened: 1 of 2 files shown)


src/OrleansAWSUtils/Membership/DynamoDBGatewayListProvider.cs
```diff
@@ -45,9 +45,15 @@ public Task InitializeGatewayListProvider(ClientConfiguration conf, Logger logge
             var expressionValues = new Dictionary<string, AttributeValue>
             {
                 { $":{SiloInstanceRecord.DEPLOYMENT_ID_PROPERTY_NAME}", new AttributeValue(deploymentId) },
-                { $":{SiloInstanceRecord.STATUS_PROPERTY_NAME}", new AttributeValue { N =  INSTANCE_STATUS_ACTIVE } }
+                { $":{SiloInstanceRecord.STATUS_PROPERTY_NAME}", new AttributeValue { N = INSTANCE_STATUS_ACTIVE } },
+                { $":{SiloInstanceRecord.PROXY_PORT_PROPERTY_NAME}", new AttributeValue { N = "0"} }
```
sergeybykov: Shouldn't this rather be `N != "0"`. Apologies if I misunderstood the syntax.

galvesribeiro: @sergeybykov this only means we are setting the attribute as a Number to 0 so it can be used later in the real query. It is not the query itself.

sergeybykov:
Thank you, @chrisbrook83!

PR #2682, opened 2017-02-07
Reduce allocations in usages of ClosureWorkItem

ReubenBond:
By creating a couple of specialised versions of `ClosureWorkItem` to handle specific usages, we can reduce allocations by around 1 or 2 per invocation.

* Change the name property from a `Func<string>` to a `string` to avoid allocating a closure.
* Introduce `AsyncClosureWorkItem` for simple `Func<Task>` cases.
* Introduce `AsyncClosureWorkItem<T>` for `Func<Task<T>>` cases.

### DIFF
--- a/src/Orleans.Runtime/Core/Dispatcher.cs
+++ b/src/Orleans.Runtime/Core/Dispatcher.cs
@@ -173,7 +173,7 @@ namespace Orleans.Runtime
                                             nonExistentActivation), exc);
                                 }
                             },
-                            () => "LocalGrainDirectory.UnregisterAfterNonexistingActivation"),
+                            "LocalGrainDirectory.UnregisterAfterNonexistingActivation"),
                             catalog.SchedulingContext);
 
                         ProcessRequestToInvalidActivation(message, nonExistentActivation, null, "Non-existent activation");
--- a/src/Orleans.Runtime/Messaging/IncomingMessageAgent.cs
+++ b/src/Orleans.Runtime/Messaging/IncomingMessageAgent.cs
@@ -164,7 +164,7 @@ namespace Orleans.Runtime.Messaging
                     if (targetActivation != null) targetActivation.DecrementEnqueuedOnDispatcherCount();
                 }
             },
-            () => "Dispatcher.ReceiveMessage"), context);
+            "Dispatcher.ReceiveMessage"), context);
         }
     }
 }
--- a/src/Orleans.Runtime/Scheduler/ClosureWorkItem.cs
+++ b/src/Orleans.Runtime/Scheduler/ClosureWorkItem.cs
@@ -1,14 +1,15 @@
 using System;
 using System.Reflection;
+using System.Threading.Tasks;
 
 namespace Orleans.Runtime.Scheduler
 {
     internal class ClosureWorkItem : WorkItemBase
     {
         private readonly Action continuation;
-        private readonly Func<string> nameGetter;
+        private readonly string name;
 
-        public override string Name { get { return nameGetter==null ? "" : nameGetter(); } }
+        public override string Name => this.name ?? GetMethodName(this.continuation);
 
         public ClosureWorkItem(Action closure)
         {
@@ -21,10 +22,10 @@ namespace Orleans.Runtime.Scheduler
 #endif
         }
 
-        public ClosureWorkItem(Action closure, Func<string> getName)
+        public ClosureWorkItem(Action closure, string name)
         {
             continuation = closure;
-            nameGetter = getName;
+            this.name = name;
 #if TRACK_DETAILED_STATS
             if (StatisticsCollector.CollectGlobalShedulerStats)
             {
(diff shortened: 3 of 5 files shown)


gabikliot:
>Change the name property from a Func<string> to a string to avoid allocating a closure.

src/OrleansRuntime/Scheduler/ShedulerExtensions.cs
```diff
@@ -7,64 +7,23 @@ internal static class SchedulerExtensions
...
-                    {
-                        await taskFunc();
-                        resolver.TrySetResult(true);
-                    }
-                    catch (Exception exc)
-                    {
-                        resolver.TrySetException(exc);
-                    }
-                };
-            scheduler.QueueWorkItem(new ClosureWorkItem(() => asyncFunc().Ignore(), () => activityName), targetContext);
```
ReubenBond: This is where the closure was being allocated for the `Func<string>`. The others should have been optimized away.

src/OrleansRuntime/Scheduler/ShedulerExtensions.cs
```diff
@@ -7,64 +7,23 @@ internal static class SchedulerExtensions
...
-            return resolver.Task;
+            var workItem = new AsyncClosureWorkItem(taskFunc);
+            scheduler.QueueWorkItem(workItem, targetContext);
+            return workItem.Task;
         }
 
         internal static Task QueueNamedTask(this OrleansTaskScheduler scheduler, Func<Task> taskFunc, ISchedulingContext targetContext, string activityName = null)
         {
-            var resolver = new TaskCompletionSource<bool>();
-            Func<Task> asyncFunc =
```
ReubenBond: By including the body of this closure in `AsyncClosureWorkItem`, we can avoid an additional allocation here.

src/OrleansRuntime/Scheduler/ShedulerExtensions.cs
ReubenBond: We also allocate a closure here where we capture `asyncFunc`

src/OrleansRuntime/Scheduler/ClosureWorkItem.cs
```diff
@@ -46,23 +47,113 @@ public override void Execute()
...
+                await this.continuation();
+                this.completion.TrySetResult(true);
+            }
+            catch (Exception exception)
+            {
+                this.completion.TrySetException(exception);
+            }
+        }
+
+        public override WorkItemType ItemType => WorkItemType.Closure;
```
ReubenBond: I didn't feel we needed to introduce a new enum value here.

src/OrleansRuntime/Scheduler/ShedulerExtensions.cs
gabikliot: There was a very good reason it was done this way (with Func<string>)
The string was different per activation, and we ended up with a huge number of allocated big strings, and it stressed GC. I even remember @rrector  reporting this. He spend a lot of time chasing the GC thrashing problem, some 4 years ago, maybe.
The code has changed since then obviously, so it is possible that the passed string is not activation dependent any more. But I would be much more careful just removing this.

At least, at least I suggest breaking this PR into multiple smaller ones.

ReubenBond: @gabikliot thanks for pointing that out. If you trace the execution, you can that there's one use of this method, and (eventually) it's from `GrainTimer` where it's always set to `Name` (a property on that class).

Digging further backwards, `GrainTimer.Name` is set to a non-constant value in a couple of cases:
* Reminders, where it's set to the reminder name.
* Streams (`PersistentStreamPullingAgent` and `PersistentStreamPullingManager`), where it's set to the value of `CurrentActivationContext.Name` + "Timer" << so this one allocates.

In this case, we're already allocating the string and holding onto it for a longer duration than the scheduled task.

It's possible I missed something, though.

gabikliot: Reuben, I was talking about all the usages of func string in work items, not just that one.

ReubenBond: Sorry, Gabi, I misunderstood. I'm only changing `ClosureWorkItem` and the other ctor calls all pass a const string. `InvokeWorkItem`, etc, can keep doing what they're doing. I didn't see a good reason to change those (although I see a curiosity in the way `OnActivationCompletedRequest` is called in the catch block: it seems it should be scheduled rather than run inline.)

gabikliot: I was talking both about ALL usages if closure work item, and also other work items. But here we discusdjng closure. There are multiple calls to it. As I said, the code evolved, so there might not be a problem with what u changed, but I would recommend separating this change only and making sure it's not holding for strings in gen 2.

Why do u think on completed request should be scheduled?

ReubenBond: > I was talking both about ALL usages if closure work item

So am I - there are 3 execution paths which pass *any* value for the name, two are constants and one is not. For the one which is not constant, the only usage of *that* code path sets it to a field value - `GrainTimer.Name` and the `GrainTimer` must outlive the work item since the work item holds a reference to it.

> Why do u think on completed request should be scheduled?

I just expected both paths (success or failure) to proceed in the same way. Instead, one is executed inline and the other is executed via the task scheduler. Looking closer I can see that it's correct & required to keep the activation's message pump running.

ReubenBond:
Functionals look good.

ReubenBond:
@dotnet-bot test this please

src/OrleansRuntime/Scheduler/ClosureWorkItem.cs
```diff
@@ -46,23 +47,113 @@ public override void Execute()
...
 
         #endregion
 
-        public override string ToString()
+        internal static string GetMethodName(Delegate action)
         {
-            var detailedName = string.Empty; 
-            if (nameGetter == null) // if NameGetter != null, base.ToString() will print its name.
+            var continuationMethodInfo = action.GetMethodInfo();
+            return string.Format(
```
attilah: Why don't we set the formatted value in ctor and use a private getter to return the name? all the props are readonly and set during ctor, am I missing something?

ReubenBond: because hopefully we never use it and can save the cost

ReubenBond:
Is anyone able to review this? Alternatively we can close it.

ReubenBond:
rebased

Issue #2605, opened 2017-01-18
NullRef thrown when pushing items into a SMS via the .Net TaskScheduler


Scottlr:
Currently, if you try to invoke a grain method when on the .Net TaskScheduler, you get a nice exception thrown saying that's the case. However, trying to push something onto a SMS using the .Net TaskScheduler throws a NullReferenceException- should there be a more meaningful exception thrown here?

Exception thrown when pushing an item onto a SMS via the .Net TaskScheduler
![image](https://cloud.githubusercontent.com/assets/7180731/22056421/f1ac2100-dd57-11e6-8579-4d9b3b470273.png)

Exception thrown when invoking grain method via the .Net TaskScheduler
![image](https://cloud.githubusercontent.com/assets/7180731/22056469/284144d4-dd58-11e6-84e8-6874131cf3b6.png)

Reproducable steps
- Wrap Stream.OnNextAsync() in a Task.Run.

PR #2683, opened 2017-02-07
Throw explicit exception when using streams not in Orleans Context.

gabikliot:
Fixes https://github.com/dotnet/orleans/issues/2605


sergeybykov:
Thank you, Gabi!

Issue #2359, opened 2016-10-28
Migrate samples to 2.1


sergeybykov:
The samples (https://github.com/dotnet/orleans/tree/master/Samples) are on 1.2.3 now. Shouldn't be to difficult to migrate them to 1.3.0.

serbrech:
Here is the list of the sample projects to track progress :

- [ ] Adventure
- [ ] AdventureDotNetCli
- [ ] AzureWebSample
- [ ] Chirper
- [ ] FSharpHelloWorld
- [ ] GPSTracker
- [ ] HelloWorld
- [ ] Presence
- [ ] ServiceFabric
- [ ] StorageProviders
- [ ] TicTacToe
- [ ] Tutorial.Minimal
- [ ] TwitterSentiment
- [ ] UnitTesting.Minimal
- [ ] VBHelloWorld

It also looks like I'll need to update the visual studio version because of tooling dlls referencing vs11 paths.

serbrech:
Could you take a quick look at the package.config here? more package had updates available than I expected https://github.com/serbrech/orleans/commit/c02f48b0d234f0cbd1050ab66ad46d69386391c9
see in particular the System.*
Reason for that is the new dependency on System.Collection.Immutbale, which suggested an update to all the other ones too.  
Just want to be sure before I proceed with all the other sample projects

attilah:
Thanks for taking this issue! ;-)

As a rule of thumb we require the lowest dependency version from all packages, so please - if for other dependencies it is not a requirement - don't upgrade to "latest and greatest" :-) stick to the versions of packages what Orleans is using in its current master branch (which always will be the next release).

serbrech:
Oh, that's not closed yet though, I'll keep working through the different Samples.
@attilah

attilah:
I don't know why was this closed :-O

jdom:
Thanks a lot for doing this. BTW, we plan to release 1.4.0 really soon though, does it make sense to wait 1 week until it's released and update to it instead?
Otherwise, you can also run `Build.cmd`, and that would generate 1.4.0 packages locally so you can reference them (there is now a nuget.config that will automatically pick them up). Check out HelloGeoSample for an example doing this.

serbrech:
after struggling with the first couple projects, I know exactly which packages need to get updates. So I'll try to write a script to run the apropriate nuget updates commands, and the subsequent packages.config edits.
That should cover 90% of the updates, and it will help for future releases too.

[...]

PR #2686, opened 2017-02-07
Update AzureWebSample to orleans 1.3.1

serbrech:
Part of #2359

### DIFF
--- a/Samples/AzureWebSample/HelloEnvironmentGrains/HelloEnvironmentGrains.csproj
+++ b/Samples/AzureWebSample/HelloEnvironmentGrains/HelloEnvironmentGrains.csproj
@@ -37,11 +37,15 @@
       <HintPath>..\packages\Newtonsoft.Json.7.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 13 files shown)


serbrech:
I'll still update the azure packages :)

serbrech:
I tried to update the target framework to the latest (4.6.2), but then I got some other inconsitencies and well, it doesn't bring much.

jdom:
You probably ran into issues with deploying to azure, as the previous family version did not support 4.6.2 out of the box. Fortunately, I see that the new OS family 5 does support it, finally! (from https://docs.microsoft.com/en-us/azure/cloud-services/cloud-services-guestos-update-matrix)
Nevertheless, I agree that there is no point in updating it to 4.6.2 anyway at this point...

jdom:
For these, I think you can remove the `Orleans.Server` and `OrleansHost` package references from the silo project, since we are not really using the pre-built OrleansHost.exe (you can also remove the `appdir-placeholder.txt` file if it's there too).
The reason I noticed this, is from this other sample we are adding, which apparently was copied from AzureWebSample: https://github.com/sebastianburckhardt/orleans/pull/11

jdom:
Thank you, @serbrech !

PR #2687, opened 2017-02-07
Tweak timeouts to improve test reliability

ReubenBond:
A timeout of 1 second assumes the test host isn't slow or overloaded.

### DIFF
--- a/test/NonSiloTests/SchedulerTests/OrleansTaskSchedulerBasicTests.cs
+++ b/test/NonSiloTests/SchedulerTests/OrleansTaskSchedulerBasicTests.cs
@@ -416,12 +416,12 @@ namespace UnitTests.SchedulerTests
                 }
             }), context);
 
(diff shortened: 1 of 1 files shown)


sergeybykov:
1 minute timeout may be too lenient and potentially slow down test run execution if those tests start timing out. Not very likely though. Maybe 5 seconds instead?

ReubenBond:
@sergeybykov having a small timeout means that this test is checking both functional and non-functional aspects of the code, but the latter is non-deterministic. If we have targets for how quickly something should run in an arbitrary environment which we have no control over or insight into, then we should separate those requirements out into a separate test which can test exactly those targets. Such a tests should check "task is scheduled and executed within 1s at least 99.99% of the time"

sergeybykov:
I'm fine merging it as is. The chance of them failing and slowing down a test run is fairly small.

PR #2692, opened 2017-02-07
Reduce SchedulingContext allocs for grain activations

### LINKED ISSUE #2677: [Question] Why is a new SchedulingContext created for each grain call?
This question has been on the back of my mind for a while now: Why is a new `SchedulingContext` created for each grain call?

Could we move the `ISchedulingContext` implementation to `ActivationData` either by making that class implement it itself or by including it as a property? `SystemTarget` does something like this. The current implementation seems a bit allocation-heavy.

Perhaps @gabikliot or @sergeybykov or someone else can shed some light?

ReubenBond:
By implementing `ISchedulingContext` in `ActivationData` we can remove at least one allocation per call.

I did not perform the same optimization for `SystemTarget` at this stage because there are comparatively very few `SystemTarget`s and we already allocate just one `SchedulingContext` per `SystemTarget` (i.e, not per-call).

I chose to implement this explicitly in `ActivationData` rather than adding a `SchedulingContext` member because it's simple and results in less GC pressure without adding any fields.

Fixes #2677.

### DIFF
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -380,8 +380,7 @@ namespace Orleans.Runtime
         /// <param name="activation"></param>
         public void RegisterMessageTarget(ActivationData activation)
         {
-            var context = new SchedulingContext(activation);
-            scheduler.RegisterWorkContext(context);
+            scheduler.RegisterWorkContext(activation.SchedulingContext);
             activations.RecordNewTarget(activation);
             activationsCreated.Increment();
         }
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -398,7 +398,7 @@ namespace Orleans.Runtime
             ActivationCollector.TryCancelCollection(activation);
             activationsDestroyed.Increment();
 
-            scheduler.UnregisterWorkContext(new SchedulingContext(activation));
+            scheduler.UnregisterWorkContext(activation.SchedulingContext);
 
             if (activation.GrainInstance == null) return;
 
@@ -877,7 +877,7 @@ namespace Orleans.Runtime
 
                     await scheduler.RunOrQueueTask(() =>
                         result.StorageProvider.ReadStateAsync(grainType, grainRef, state),
-                        new SchedulingContext(result));
+                        result.SchedulingContext);
                     
                     sw.Stop();
                     StorageStatisticsGroup.OnStorageActivate(result.StorageProvider, grainType, result.GrainReference, sw.Elapsed);
@@ -1135,7 +1135,7 @@ namespace Orleans.Runtime
                 foreach (var activation in list)
                 {
                     var activationData = activation; // Capture loop variable
-                    var task = scheduler.RunOrQueueTask(() => CallGrainDeactivateAndCleanupStreams(activationData), new SchedulingContext(activationData));
+                    var task = scheduler.RunOrQueueTask(() => CallGrainDeactivateAndCleanupStreams(activationData), activationData.SchedulingContext);
                     tasks2.Add(new Tuple<Task, ActivationData>(task, activationData));
                 }
                 var asyncQueue = new AsyncBatchedContinuationQueue<ActivationData>();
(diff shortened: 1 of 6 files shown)


ReubenBond:
@dotnet-bot test this please

ReubenBond:
Functionals look good

gabikliot:
Reuben, my preference would be that you simply store the context inside the activation data, just like system target does now. Instead of having it to implement the interface. This separation was by design. Activation is not a context. Activation has a context.

Beyond the  abstract question of polymotphism and design, from a practical perspective this will be not intrusive and much smaller then your current fix and also compatible with other places that have this context.

ReubenBond:
I won't push back. My reasoning was that we expect a large number of activations, so if we can greatly reduce the number of objects the GC has to track, then that's a good thing. I felt that that was more important than the conceptual clarity of the 'has a' relationship.

I feel that the current `SchedulingContext` class is a poor-man's discriminated union at the moment, too, resulting it being larger than needed. It's a union type, so it has fields for everything even when it's invalid for 2/3 of those fields to be used on any given instance.

gabikliot:
Well, ui

gabikliot:
Well, your original question was in the narrow context of why allocate on every call instead of per ad. Now you are broadening it to the question of abstraction, union, .... 
I prefer not to mix and match issues. For the original question, as we agreed in the original issues, let's mimic system target and do a member field.

If you are really after that, we can revisit it in a separate issue. Not sure why you would be after that. There are so much more places to save much more costly allocations. I can tell you where. And this particular abstraction if a union is the least of our problems. It may not be strictly optimal, but it does the job right.

ReubenBond:
In the issue, I mentioned the approach I took in the current revision as an option:
> Could we move the `ISchedulingContext` implementation to `ActivationData` either **by making that class implement it itself** or by including it as a property?

I'd have given `SystemTarget` the same treatment, but I didn't want to mix it in with this PR (and it's lower pri).

What are the other places which have costlier allocations? I've been removing per-call (hot path) allocs mostly and mostly leaving out-of-band paths.

Anyway, I'll go & change the implementation.

gabikliot:
Yes, let's finish that issue and then open another issue to discuss other places for allocation optimization.

ReubenBond:
ok, updated.

gabikliot:
I will take a look this weekend.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -398,7 +398,7 @@ public void UnregisterMessageTarget(ActivationData activation)
             ActivationCollector.TryCancelCollection(activation);
             activationsDestroyed.Increment();
 
-            scheduler.UnregisterWorkContext(new SchedulingContext(activation));
+            scheduler.UnregisterWorkContext(activation.SchedulingContext);
```
gabikliot: this looks good.
Why not do the same in RegisterMessageTarget: use activation.SchedulingContext instead of var context = new SchedulingContext(activation);

ReubenBond: Fixed, thanks. Now of the 3 `SchedulingContext` ctors, 2 are used once (`SystemTarget` & `ActivationData`) and one is unused (for 'system thread').

I could remove the latter for hygiene, but I rather leave it for now as it looks like it may prove itself useful soon.

gabikliot:
Julian, not a big deal, but next time, if I am consistently reviewing and providing feedback on the cl, please allow me to review the fixes and tge later iterations and decide if it is ready to merge or not.

gabikliot:
Reuben, system thread is not currently used, but it's a place holder for an idea for future improvement I have. So let's keep it for now, until/unless we decide not to implement it. I will share the ideas in the next few days.

ReubenBond:
@gabikliot agreed, there are a few areas where I think it will become useful, so I've left it in.

jdom:
@gabikliot sorry, my bad. I thought that when you added `this looks good.` you meant the entire PR was ready, and you just had a minor comment but were mostly done with it. Once Reuben addressed the small concern, I thought it was safe to merge (kind of like the Signed Off approach we used in CodeFlow).
Nevertheless, next time I'll wait for you to either merge or explicitly mark the PR as `Approved`.

gabikliot:
No worries.

PR #2695, opened 2017-02-08
Suppress noise build warnings

sergeybykov:
Suppress a set of warnings stemming from us having deprecated `AzureQueueStreamProvider`.
Suppress a false positive warning about not disposing a timer.

### DIFF
--- a/src/Orleans/Statistics/RuntimeStatisticsGroup.cs
+++ b/src/Orleans/Statistics/RuntimeStatisticsGroup.cs
@@ -233,11 +233,11 @@ namespace Orleans.Runtime
 
         public void Stop()
         {
-            if (cpuUsageTimer != null)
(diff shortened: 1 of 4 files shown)


src/Orleans/Statistics/RuntimeStatisticsGroup.cs
```diff
@@ -254,7 +254,7 @@ public void Dispose()
             numberOfInducedGCsPF.Dispose();
             largeObjectHeapSizePF.Dispose();
             promotedFinalizationMemoryFromGen0PF.Dispose();
-            this.cpuUsageTimer?.Dispose();
+            cpuUsageTimer?.Dispose();
```
ReubenBond: my bad... if we don't use the safe navigation operator then the warning will go away - the analyzer is dated.

PR #2699, opened 2017-02-08
Create SLN and Project files for master and VS2017 RC4.

attilah:
- Create SLN and Project files for master and VS2017 RC4.
- F# is not included now, since new tooling does not yet support building for net451.
- Cleanup AssemblyInfo.cs for some projects.
- Cleanup unused file or filename with typo.
- Orleans.2017.Full.sln is a temporary file, will be removed, but now should be preserved as a checkpoint for later PRs.

First step to multitargeting and removal of vNext folder.

Now all files has a .2017 suffix to not to collide with VS2015 files for v1.x


src/.nuget/packages.config
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="FSharp.Compiler.Tools" version="4.0.1.20" />
-  <package id="Google.Protobuf.Tools" version="3.0.0" />
+  <package id="Google.Protobuf.Tools" version="3.2.0" />
```
jdom: Is this required by this PR?

src/Orleans/Orleans.2017.csproj
```diff
@@ -0,0 +1,91 @@
...
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(BuildNumber)' == ''">
+      <BuildNumber>1</BuildNumber>
+    </PropertyGroup>
+  </Target>
+  <Target Name="LogBuildProperties">
+    <Message Importance="high" Text="[Build Properties]" />
+    <!-- Show values of some useful properties -->
+    <Message Importance="high" Text="TeamProject=$(TeamProject)&#xD;&#xA;SolutionRoot=$(SolutionRoot)&#xD;&#xA;SourcesDirectory=$(SourcesDirectory)&#xD;&#xA;BinariesDirectory=$(BinariesDirectory)&#xD;&#xA;BinariesRoot=$(BinariesRoot)&#xD;&#xA;BuildDirectory=$(BuildDirectory)&#xD;&#xA;BuildProjectFolderPath=$(BuildProjectFolderPath)&#xD;&#xA;MSBuildProjectDirectory=$(MSBuildProjectDirectory)&#xD;&#xA;OutputPath=$(OutputPath)&#xD;&#xA;OutDir=$(OutDir)&#xD;&#xA;DropLocation=$(DropLocation)&#xD;&#xA;PackagesDirectory=$(PackagesDirectory)&#xD;&#xA;BuildNumber=$(BuildNumber)&#xD;&#xA;builduri=$(builduri)&#xD;&#xA;BuildUri=$(BuildUri)&#xD;&#xA;MSBuildForwardPropertiesFromChild=$(MSBuildForwardPropertiesFromChild)&#xD;&#xA;BuildingInsideVisualStudio=$(BuildingInsideVisualStudio)&#xD;&#xA;IsDesktopBuild=$(IsDesktopBuild)" />
+  </Target>
```
jdom: Is this of any value? It used to contain properties that were useful long time ago for our custom build, but probably not relevant that we are moving to a more standard build... Might still be given the bootstrapping logic, just for you to check

jdom:
Do all version for the nuget references align with the ones in project.json?
Is the plan to later make this solution multi-target and get rid of vNext entirely?

src/OrleansPSUtils/OrleansPSUtils.2017.csproj
```diff
@@ -0,0 +1,59 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup Label="NuGet">
+    <IsPackable>true</IsPackable>
+    <PackageId>Microsoft.Orleans.OrleansPSUtils</PackageId>
```
benjaminpetit: For this project (and maybe some other), at build time, we need to put the build output into a subfolder. We have this in the original csproj
```` xml
  <PropertyGroup Condition="'$(OutDir)' != ''">
    <OutDir>$(OutDir)\Orleans</OutDir>
    <OutputPath>$(OutDir)</OutputPath>
  </PropertyGroup>
````

benjaminpetit: See #1990 and #2068

src/Orleans/Orleans.2017.csproj
attilah: Step by step. I didn't want to go all in, focused on project file migration.

src/OrleansPSUtils/OrleansPSUtils.2017.csproj
attilah: I've other issues with the PS extension, since here we're using s 3rd party nuget package which contains a powershell dll...

With the new csproj's we don't have same set of variables set, I'll check it later.

This PR is the first step, not a complete replacement (mainly because of F#).

attilah:
@jdom yes, I moved back the versions to be in sync with project.json

src/OrleansPSUtils/OrleansPSUtils.2017.csproj
attilah: I'll move the powershell module to the official nuget package: [Microsoft.PowerShell.5.ReferenceAssemblies/](https://www.nuget.org/packages/Microsoft.PowerShell.5.ReferenceAssemblies)

ReubenBond:
> Is the plan to later make this solution multi-target and get rid of vNext entirely?

I hope so :)

src/OrleansPSUtils/OrleansPSUtils.2017.csproj
galvesribeiro: I don't remember now why I used that 3rd party library. I think it was suggested by someone from PS team by that time because of some limitation on the official nuget. However, I've already spoke with the PS team again and I'll port that extension to use the new xplat powershell nugets.

attilah: That's what I did, updated to 5.x which is the first xplat version. We should do that for  vNext and master too.

On the other hand I'd like to have the PSUtils installable from nuget, so we need a nuget package too. I started that work in the 2017 csprojs, but since I'm not a PS guru I don't know the exact packaging needs.

src/.nuget/packages.config
jdom: There is a `GenerateProtos.cmd` that may no longer be linked, but also needs updating to this 3.2.0 version if we truly require it

src/OrleansPSUtils/OrleansPSUtils.2017.csproj
galvesribeiro: Nuget package for PSUtils? Unless you plan to make something like `dotnet ef` which is a *tool* package that is distributed by Nuget packages but only work inside the project directory context (i.e. not a global tool), I don't see why you would need that.

As far as I know, the only official PS Cmdlet distribution model is thru https://www.powershellgallery.com/ and unofficially, from Chocolatey.

Can you elaborate more on why you need a Nuget for it? The gallery deployment is something I was going to arrange with @jdom a while ago and make it pass thru the same sign process that other packages goes thru.

galvesribeiro:
I think in general, we should have 2 sets of references here. 

1. (Almost) The same as in project.json, so we can publish 1.x.y official releases from the same projects
2. All the way up to latest nugets on vNext build.

We are far behind of all nugets to avoid break people on main release and I think we had a debate a while ago where we agreed that vNext/2.0.0 would include a mass upgrade on all the dependencies. Many dependencies like EventHubs and Azure Storage ones will force us to go latest anyway in order to be `netstandard` compatible, so I would suggest move to the latest on vNext before publish it and set a bar there and then we can follow that same policy of latest - 1 again but this time, without having too old nugets.

src/.nuget/packages.config
attilah: Ok, reverted based on feedback.

PR #2700, opened 2017-02-08
Cleanup new multicluster samples

jdom:
Cleanup code
Remove dead code, OrleansHost reference and license
Rename IChatGrainInterface

Also removed unneeded Serializable attribute from ChatState (as it is event sourced). @sebastianburckhardt please confirm if this is actually the case, as if it is uneeded, it will avoid confusion (I was scratching my head on why this would be needed, unless this is actually doing snapshotting for optimization, which I believe isn't).


sebastianburckhardt:
It actually *is* doing snapshotting.

Since there is no attribute for the consistency provider, it uses the default consistency provider, which is `Orleans.EventSourcing.StateStorage.LogConsistencyProvider`, always stores the latest snapshot. The events themselves are not persisted (but they are sent around in notification messages). 

Not sure why this still works when you remove the [Serializable] attribute... maybe the storage provider can somehow serialize it anyway. Or maybe the storage accesses are failing and you can't see it because the sample does not monitor for exceptions.

jdom:
Yeah, it might be that exceptions are not shown. Thanks, I'll revert that commit

jdom:
Done, this is ready to be merged. It's only the cleanup commit, without the removal of Serializable

PR #2703, opened 2017-02-08
Update Using-Immutable-to-Optimize-Copying.md

ReubenBond:
Updated to reflect the current state of the code


ReubenBond:
I think we should merge this doc with the Serialization doc

jdom:
I'm fine if we decide to merge the 2 docs. Nevertheless, merging these changes as they make sense regardless.

PR #2706, opened 2017-02-09
Remove remaining references to RuntimeClient.Current from test code.

ReubenBond:


ReubenBond:
@dotnet-bot test this please

src/OrleansTestingHost/TestCluster.cs
```diff
@@ -415,7 +421,8 @@ public void InitializeClient()
             GrainClient.Initialize(clientConfig);
             this.GrainFactory = GrainClient.GrainFactory;
             this.InternalGrainFactory = this.GrainFactory as IInternalGrainFactory;
-            this.StreamProviderManager = RuntimeClient.Current.CurrentStreamProviderManager;
+            this.ServiceProvider = GrainClient.ServiceProvider;
+            this.StreamProviderManager = this.ServiceProvider.GetRequiredService<IRuntimeClient>().CurrentStreamProviderManager;
```
jdom: Isn't the StreamProviderManager stored directly in the service provider? As I mentioned before, I see IRuntimeClient as a glorified SP facade, with some extra features. But stuff like the stream provider manager looks like it can come directly from the SP.

ReubenBond: It is on the silo, but not on the client. Should I change that?

jdom:
@dotnet-bot test this please

src/OrleansTestingHost/TestCluster.cs
jdom: if it's a small change, sure, otherwise this is a good first step. No need to hold the merge for something like this, as this aspect is not worse than before (and the rest is all advantageous)

jdom:
@dotnet-bot test this please

PR #2712, opened 2017-02-09
Fix DI registrations for ServiceFabric

ReubenBond:
Fixes issues with the Service Fabric providers. Reported by @Kwal (thanks!)

### DIFF
--- a/src/Orleans/Runtime/OutsideRuntimeClient.cs
+++ b/src/Orleans/Runtime/OutsideRuntimeClient.cs
@@ -116,6 +116,7 @@ namespace Orleans
             services.AddFromExisting<IInternalGrainFactory, GrainFactory>();
             services.AddFromExisting<IGrainReferenceConverter, GrainFactory>();
             services.AddSingleton<MessageFactory>();
+            services.AddSingleton<Func<string, Logger>>(LogManager.GetLogger);
             this.ServiceProvider = services.BuildServiceProvider();
             this.InternalGrainFactory = this.ServiceProvider.GetRequiredService<IInternalGrainFactory>();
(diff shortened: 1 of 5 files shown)


src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -116,6 +116,7 @@ public OutsideRuntimeClient(ClientConfiguration cfg, bool secondary = false)
             services.AddFromExisting<IInternalGrainFactory, GrainFactory>();
             services.AddFromExisting<IGrainReferenceConverter, GrainFactory>();
             services.AddSingleton<MessageFactory>();
+            services.AddSingleton<Func<string, Logger>>(LogManager.GetLogger);
```
jdom: This method was only available to internals, do we want to open it up to the public? I'm fine with it, just highlighting this change

ReubenBond: Yeah, I think it's better this way. It's not very discoverable, but this signature makes sense to me unless we want to maybe add some explicit logging provider interface, like `ILogManager`. We eventually may, but this is a conservative option for now, IMO. What do you think?

jdom: I think it's OK for now... Clients are probably not using the Olreans logger anyway, and this opens it up for extensions to the runtime such as this.
Yes, I think we will revamp it... IMO we should use the Microsoft.Extensions.Logging package and get rid of our abstractions...

ReubenBond: That would be much better - more alignment with ASP.NET

PR #2713, opened 2017-02-09
[1.4.0] Fix ServiceFabric support + Samples

ReubenBond:
xref #2712


jdom:
Can you include #2671 and #2700 since they are related to 1.4.0?

ReubenBond:
ok, how does this look?

Issue #2710, opened 2017-02-09
Orleans test host failed to load Microsoft.SqlServer.Types


saitonakamura:
When trying to add `Microsoft.SqlServer.Types` nuget package to Orleans host and therefore load and use it in grains it fails with this errors

```
Could not load file or assembly 'file:///c:\users\mikhail_bashurov\documents\visual studio 2015\Projects\TestOrleansSqlServerTypes\Host\bin\Debug\SqlServerTypes\x64\msvcr100.dll' or one of its dependencies. The module was expected to contain an assembly manifest.
Exception = System.BadImageFormatException: Could not load file or assembly 'file:///c:\users\mikhail_bashurov\documents\visual studio 2015\Projects\TestOrleansSqlServerTypes\Host\bin\Debug\SqlServerTypes\x64\msvcr100.dll' or one of its dependencies. The module was expected to contain an assembly manifest.
File name: 'file:///c:\users\mikhail_bashurov\documents\visual studio 2015\Projects\TestOrleansSqlServerTypes\Host\bin\Debug\SqlServerTypes\x64\msvcr100.dll'
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoadAssemblyName(AssemblyName assemblyRef, Evidence assemblySecurity, RuntimeAssembly reqAssembly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoadFrom(String assemblyFile, Evidence securityEvidence, Byte[] hashValue, AssemblyHashAlgorithm hashAlgorithm, Boolean forIntrospection, Boolean suppressSecurityChecks, StackCrawlMark& stackMark)
[...]

PR #2714, opened 2017-02-10
Fix native PE files check before assembly loading. Fixes: #2710

attilah:


src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -361,40 +358,71 @@ private bool ReflectionOnlyLoadAssembly(string pathName, out Assembly assembly,
...
+                                processorArchitecture = ProcessorArchitecture.X86;
+
+                            var isManaged = isPureIL || processorArchitecture == ProcessorArchitecture.MSIL;
 
-            if (peKind.HasFlag(PortableExecutableKinds.NotAPortableExecutableImage) ||
-                peKind.HasFlag(PortableExecutableKinds.Unmanaged32Bit))
+
+                            var isLoadable = (isPureIL && processorArchitecture == ProcessorArchitecture.MSIL) ||
+                                                 (Environment.Is64BitProcess && processorArchitecture == ProcessorArchitecture.Amd64) ||
+                                                 (!Environment.Is64BitProcess && processorArchitecture == ProcessorArchitecture.X86);
```
ReubenBond: An amd64 process cannot load x86 assemblies?

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -361,40 +358,71 @@ private bool ReflectionOnlyLoadAssembly(string pathName, out Assembly assembly,
...
+            catch (UnauthorizedAccessException ex)
             {
-                complaints = new[] { $"The assembly {assembly.FullName} was compiled for {machine} but the current exe was compiled for {this.exeInfo.Value.MachineType}" };
                 return false;
             }
-
-            return true;
+            catch (MissingMethodException ex)
+            {
+                complaints = new[] { "MissingMethodException occured. Please try to add a BindingRedirect for System.Collections.ImmutableCollections to the App.config file to correct this error." };
```
ReubenBond: This is unfortunate. Is there no way we can deterministically fix this?

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -361,40 +358,71 @@ private bool ReflectionOnlyLoadAssembly(string pathName, out Assembly assembly,
...
+                return false;
+            }
+            catch (Exception ex)
+            {
+                complaints = new[] { ex.ToString() };
```
ReubenBond: I wonder if we should use `LogFormatter.PrintException(ex)` here

ReubenBond:
LGTM after the comments are addressed

src/Orleans/project.json
```diff
@@ -2,7 +2,8 @@
   "dependencies": {
     "Microsoft.Extensions.DependencyInjection": "1.0.0",
     "Newtonsoft.Json": "7.0.1",
-    "System.Collections.Immutable": "1.1.37"
+    "System.Collections.Immutable": "1.1.37",
+    "System.Reflection.Metadata": "1.2.0"
```
jdom: do you need to update the nuspec too?

src/Orleans/AssemblyLoader/AssemblyLoader.cs
attilah: Nope BadImageFormatException happen, we already guarded against it.

src/Orleans/AssemblyLoader/AssemblyLoader.cs
attilah: No, it maybe not will happen, since this catch was there for this explicit library (not going into details), but left here as a safeguard.

src/Orleans/AssemblyLoader/AssemblyLoader.cs
attilah: Are we doing that other places? this was not that, I can add of course, but complaints are printed otherwise I think.

src/Orleans/project.json
attilah: Good catch

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -361,40 +360,71 @@ private bool ReflectionOnlyLoadAssembly(string pathName, out Assembly assembly,
...
+                            var isPureIL = (peReader.PEHeaders.CorHeader.Flags & CorFlags.ILOnly) != 0;
+
+                            if (peReader.PEHeaders.PEHeader.Magic == PEMagic.PE32Plus)
+                                processorArchitecture = ProcessorArchitecture.Amd64;
+                            else if ((peReader.PEHeaders.CorHeader.Flags & CorFlags.Requires32Bit) != 0 || !isPureIL)
+                                processorArchitecture = ProcessorArchitecture.X86;
 
-            if (peKind.HasFlag(PortableExecutableKinds.NotAPortableExecutableImage) ||
-                peKind.HasFlag(PortableExecutableKinds.Unmanaged32Bit))
+                            var isManaged = isPureIL || processorArchitecture == ProcessorArchitecture.MSIL;
```
jdom: `isManaged` is not being used

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -361,40 +360,71 @@ private bool ReflectionOnlyLoadAssembly(string pathName, out Assembly assembly,
...
+                        }
+                        else
+                        {
+                            complaints = new[] { $"The file {fileName} does not contain any CLR metadata, probably it is a native file." };
+                            return false;
+                        }
+                    }
+                }
+            }
+            catch (IOException ex)
```
jdom: warnings for `ex` being defined in all these catch blocks, but not used

jdom:
Cleaned up warnings in https://github.com/dotnet/orleans/pull/2719

PR #2715, opened 2017-02-10
Improve support for generic grains with generic state

ReubenBond:
Current support for generic grains with generic state assumes that the type parameter to the grain is also the type parameter of the state class.

This PR relaxes that assumption by navigating the concrete grain type's hierarchy until it reaches the `Grain<TState>` level and extracting the concrete value of `TState`.

/cc @lwansbrough who [reported this issue on Gitter](https://gitter.im/dotnet/orleans?at=589d05abaa800ee52c75e6de) - thank you, Loch!


ReubenBond:
@dotnet-bot test this please

Issue #2232, opened 2016-09-27
Help wanted: Replace all usages of TestingSiloHost with TestCluster


jdom:
Our plan is to make `TestingSiloHost` obsolete as it was completely superseded by the more flexible `TestCluster`. Nevertheless, we have several tests in the `TesterInternal` project that are still using it and the declarative config that need to be converted.
We should do a pass and remove all references to the TestingSiloHost class.
In order to do that, those tests would have to stop using the XML configuration, and use programmatic configuration.
For reference, PR #1643 is where the Tester project was migrated to use TestCluster, and specifically this commit from it: https://github.com/dotnet/orleans/pull/1643/commits/d931e6a418eb3ec119c1a254d18b90967fe595f1

nikhilgupta23:
@jdom I am not familiar with this project but will still like to contribute by solving this issue (as part of Hacktoberfest). Is there something else that needs to be taken care of or a simple replace of all instances in https://github.com/dotnet/orleans/tree/master/test/TesterInternal will do? Thanks.

jdom:
Hi, thanks for offering help. I updated the main description with some info about the move, plus reference info:
> In order to do that, those tests would have to stop using the XML configuration, and use programmatic configuration.
For reference, PR #1643 is where the Tester project was migrated to use TestCluster, and specifically this commit from it: https://github.com/dotnet/orleans/pull/1643/commits/d931e6a418eb3ec119c1a254d18b90967fe595f1

amccool:
how much do the specifics like the port numbers matter in these XML configs?
case to point
https://github.com/dotnet/orleans/blob/master/test/TesterInternal/Config_BootstrapProviders.xml#L26
shouldn't we want to use the TestCluster defaults?

jdom:
They do not matter, and specifically for ports, it matters that they are not the same for each cluster, which is one of the benefits of moving to TestCluster, where you get the default config and you just override some specific settings that are relevant to just that test fixture

amccool:
is this:

```
new TestingSiloHost(new TestingSiloOptions
                {
                    StartPrimary = true,
                    StartSecondary = false,
                })
```

the same as

`new TestClusterOptions(initialSilosCount:1);`

jdom:
yes, almost equivalent. BTW, did you start this work then?

jdom:
[...]

PR #2718, opened 2017-02-10
another removal of TestingSiloHost

amccool:
removed another usage of TestingSiloHost for #2232


test/TesterInternal/GeoClusterTests/BasicLogTestGrainTests.cs
```diff
@@ -10,28 +13,39 @@
...
+                options.ClusterConfiguration.AddMemoryStorageProvider("Default");
+                options.ClusterConfiguration.AddMemoryStorageProvider("MemoryStore");
+                options.ClusterConfiguration.AddAzureTableStorageProvider("AzureStore");
 
+                //AdjustConfig = cfg => LogConsistencyProviderConfiguration.ConfigureLogConsistencyProvidersForTesting(TestDefaultConfiguration.DataConnectionString,cfg)
+                options.ClusterConfiguration.AddAzureTableStorageProvider();
+                options.ClusterConfiguration.AddStateStorageBasedLogConsistencyProvider();
+                options.ClusterConfiguration.AddLogStorageBasedLogConsistencyProvider();
+                options.ClusterConfiguration.AddCustomStorageInterfaceBasedLogConsistencyProvider("CustomStorage");
+                options.ClusterConfiguration.AddCustomStorageInterfaceBasedLogConsistencyProvider("CustomStoragePrimaryCluster");
```
jdom: Looks like `ConfigureLogConsistencyProvidersForTesting` was setting `"A"` as the primary cluster here. Not sure if that was being used for these tests, but it's a difference

amccool: Good catch on the props.   I will add that for consistency.  perhaps @sebastianburckhardt could validate if its needed.

I will remove the comment as well and push

jdom:
Thanks @amccool ! We are slowly getting rid of TestingSiloHost :)

PR #2720, opened 2017-02-11
Choose more reasonable default when activating JournaledGrain

sebastianburckhardt:
While writing documentation, I noticed another mistake in the JournaledGrain API, which is likely to confuse many users unless fixed.

Previously, the grain did NOT wait until the latest state is loaded on activation. Users had to override OnActivateAsync to make sure the grain has loaded the latest version before it starts executing calls.

*But this is exactly backwards*: the simple, common scenario is that we always wait for the load to complete before the grain starts taking calls; the advanced, uncommon scenario is to skip the wait (there are very few cases where you would want to do this).

**So this PR swaps the default behavior to be more in line with standard expectations.**

 Now, we wait by default during activation; but users can still override OnActivateAsync if they want a different behavior, e.g. do something before waiting, or do something after waiting, or skip the wait completely.

### DIFF
--- a/src/OrleansEventSourcing/JournaledGrain.cs
+++ b/src/OrleansEventSourcing/JournaledGrain.cs
@@ -168,6 +168,16 @@ namespace Orleans.EventSourcing
 
 
         /// <summary>
+        /// By default, upon activation, the journaled grain waits until it has loaded the latest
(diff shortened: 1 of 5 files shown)


jdom:
Thanks, this is more inline with the existing storage provider feature

PR #2721, opened 2017-02-12
Construct providers using DI

ReubenBond:
This ought to fix functional tests, which I recently broke (pretty bad at git).

Providers which are loaded from external libraries should be given access to the IoC container so that they  can access other services without resorting to globals

### DIFF
--- a/vNext/test/ServiceBus.Tests/ServiceBus.Tests.csproj
+++ b/vNext/test/ServiceBus.Tests/ServiceBus.Tests.csproj
@@ -23,7 +23,7 @@
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="FluentAssertions" Version="4.0.0" />
+    <PackageReference Include="FluentAssertions" Version="4.18.0" />
     <PackageReference Include="xunit" Version="2.1.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.1.0" />
     <PackageReference Include="Xunit.SkippableFact" Version="1.2.14" />
--- a/src/Orleans/AssemblyLoader/AssemblyLoader.cs
+++ b/src/Orleans/AssemblyLoader/AssemblyLoader.cs
@@ -7,6 +7,7 @@ using System.Reflection;
 using System.Reflection.PortableExecutable;
 #endif
 using System.Text;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Orleans.Runtime
 {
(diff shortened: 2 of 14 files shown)


vNext/test/ServiceBus.Tests/ServiceBus.Tests.csproj
```diff
@@ -23,7 +23,7 @@
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="FluentAssertions" Version="4.0.0" />
+    <PackageReference Include="FluentAssertions" Version="4.18.0" />
```
jdom: I noticed this dependency once... are we even using it? If it's 1 codefile that uses it, I would prefer to just get rid of it.
I liked and used FluentAssertions, but I prefer consistency... Not necessarily to be done in this PR

ReubenBond: I'm not sure, but I saw relevant errors in vNext tests. I made this change & it fixed them

ReubenBond:
There are some errors in newly added EventSourcingTests, but those have been failing consistently for a while.

The other failures are outputting "No gateways available" & I'm not sure what the issue there is.

jdom:
Question is: can I merge this or are you investigating the gateway errors (or at least running those tests locally) to rule out new issues?

xiazen:
Those No Gateway available failures can be fixed by using TestClusterPerTest in Grain_AzureBlobStore tests. I believe it is caused by Grain_AzureBlobStore tests interfered with each other... 

Basically if Grain_AzureBlobStore_SiloRestart failed, then other tests will have no silo to connect to. 

But I haven't look into why Grain_AzureBlobStore_SiloRestart would fail, it doesn't fail if run independently

xiazen:
But I don't think it is related to Reuben's change

ReubenBond:
Oh, thanks @xiazen! Should I make the `TestClusterPerTest` change here?

jdom:
nah, just rule out if the change is a standard non-deterministic failure, or due to your change... Did you try running those tests locally to see if they pass? If they do, then we can fix the non-deterministic tests separately

xiazen:
I made the TestClusterPerTest  change in my other PR . But my PR is not in a good state to review yet. you can go ahead to fix it if you want :)

ReubenBond:
I've added the change here. A VS 2017 update yesterday (I think) destroyed my local Storage Emulator (it's gone?) ... so I need to work out how to fix that.
image

xiazen:
? just use the real azure storage? I usually use our TestSecrects.json which is used in our load tests... just copy that file to your GitHub repo from load test VSO..

jdom:
TBH, I don't know if moving to a slower testing approach is the way to go... Now it creates a new cluster for each test, which takes CONSIDERABLY longer than before... that's why I suggested not to fix it in this PR, and keep it isolated... if the failures were not related to your change, then I would separate the fix.

ReubenBond:
I can ditch the commit and we can look at it in another PR - either way is fine with me. I agree that the slowness is an issue. We should address that for other tests, too.

jdom:
Yes, please remove the commit and I'll run those tests locally. If they pass, I'll merge

ReubenBond:
done

jdom:
They indeed pass (as was the most likely scenario). Merged

PR #2724, opened 2017-02-13
Bugfix - EventHubSequenceToken not used consistently

jason-bragg:
EventSequenceToken was being used instead of EventHubSequenceToken.

### DIFF
--- a/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
+++ b/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
@@ -138,9 +138,7 @@ namespace Orleans.ServiceBus.Providers
         public int Compare(CachedEventHubMessage cachedMessage, StreamSequenceToken streamToken)
         {
             var realToken = (EventSequenceToken)streamToken;
-            return cachedMessage.SequenceNumber != realToken.SequenceNumber
-                ? (int)(cachedMessage.SequenceNumber - realToken.SequenceNumber)
-                : 0 - realToken.EventIndex;
+            return (int)(cachedMessage.SequenceNumber - realToken.SequenceNumber);
         }
 
         /// <summary>
@@ -237,7 +235,7 @@ namespace Orleans.ServiceBus.Providers
         /// <returns></returns>
         public virtual StreamSequenceToken GetSequenceToken(ref CachedEventHubMessage cachedMessage)
         {
-            return new EventSequenceTokenV2(cachedMessage.SequenceNumber, 0);
+            return new EventHubSequenceTokenV2("", cachedMessage.SequenceNumber, 0);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
```diff
@@ -138,9 +138,7 @@ public class EventHubDataComparer : ICacheDataComparer<CachedEventHubMessage>
         public int Compare(CachedEventHubMessage cachedMessage, StreamSequenceToken streamToken)
         {
             var realToken = (EventSequenceToken)streamToken;
-            return cachedMessage.SequenceNumber != realToken.SequenceNumber
-                ? (int)(cachedMessage.SequenceNumber - realToken.SequenceNumber)
-                : 0 - realToken.EventIndex;
+            return (int)(cachedMessage.SequenceNumber - realToken.SequenceNumber);
```
ReubenBond: `EventHubSequenceTokenV2` also has an `EventIndex`. Is that not relevant for comparison?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
```diff
@@ -237,7 +235,7 @@ protected virtual IBatchContainer GetBatchContainer(EventHubMessage eventHubMess
         /// <returns></returns>
         public virtual StreamSequenceToken GetSequenceToken(ref CachedEventHubMessage cachedMessage)
         {
-            return new EventSequenceTokenV2(cachedMessage.SequenceNumber, 0);
+            return new EventHubSequenceTokenV2("", cachedMessage.SequenceNumber, 0);
```
ReubenBond: Some static analysis tools will complain if `""` is used instead of `string.Empty`.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
jason-bragg: No, and the previous logic contained an extremely rare bug.  The event index is index of an item in a batch, and the cache only deals with full batches.  So by checking the index, we would not recognize that this is in fact the right batch but instead go to the previous batch.  Normally this is fine as duplicates are allowed, however, in the rare case that this batch (the right batch) is the last item in the cache, we'd fail to find a batch even when one was available.  Absurdly rare edge case, but wrong none-the-less.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
jason-bragg: will change to string.empty.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
ReubenBond: That makes sense, thanks for the explanation

PR #2725, opened 2017-02-13
More straightforward implementation of AssumeHomogenousSilosForTesting

benjaminpetit:
Not sure what I had in mind when I wrote the first implementation...

### DIFF
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -209,12 +209,12 @@ namespace Orleans.Runtime
 
         public IList<SiloAddress> GetCompatibleSiloList(GrainId grain)
         {
-            // For test only: if we have silos taht are not yet in the Cluster TypeMap, we assume that they are compatible
+            // For test only: if we have silos that are not yet in the Cluster TypeMap, we assume that they are compatible
             // with the current silo
             if (this.config.AssumeHomogenousSilosForTesting)
                 return AllActiveSilos;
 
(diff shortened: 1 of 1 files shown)


src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -209,23 +209,16 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
 
         public IList<SiloAddress> GetCompatibleSiloList(GrainId grain)
         {
-            var typeCode = grain.GetTypeCode();
+            // For test only: if we have silos taht are not yet in the Cluster TypeMap, we assume that they are compatible
+            // with the current silo
+            if (this.config.AssumeHomogenousSilosForTesting)
+                return AllActiveSilos;
+
+                var typeCode = grain.GetTypeCode();
```
ReubenBond: The indentation makes it look like this var is a part of the if block

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -209,23 +209,16 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
 
         public IList<SiloAddress> GetCompatibleSiloList(GrainId grain)
         {
-            var typeCode = grain.GetTypeCode();
+            // For test only: if we have silos taht are not yet in the Cluster TypeMap, we assume that they are compatible
```
ReubenBond: s/taht/that

PR #2726, opened 2017-02-14
Cherry-pick two more fixes for 1.4.0

sergeybykov:
Bugfix - EventHubSequenceToken not used consistently (#2724)
Support grains with generic state where the state param do not match the grain type params (#2715)


jdom:
Reviewed https://github.com/dotnet/orleans/pull/2714 and it should be safe to pull in too (as well as it's cleanup one: https://github.com/dotnet/orleans/pull/2719)

PR #2728, opened 2017-02-14
fix typo in NewKey parameter check

danielkrainas:
Low fruit but I noticed this as I was checking out the code

### DIFF
--- a/src/Orleans/IDs/UniqueKey.cs
+++ b/src/Orleans/IDs/UniqueKey.cs
@@ -114,7 +114,7 @@ namespace Orleans.Runtime
             // in the string representation of a key, we grab the least significant half of n1.
             // therefore, if n0 is non-zero and n1 is 0, then the string representation will always be
             // 0x0 and not useful for identification of the grain.
-            if (n1 == 0 && n1 != 0)
+            if (n1 == 0 && n0 != 0)
                 throw new ArgumentException("n0 cannot be zero unless n1 is non-zero.", "n0");
             if (category != Category.KeyExtGrain && category != Category.GeoClient && keyExt != null)
                 throw new ArgumentException("Only key extended grains can specify a non-null key extension.");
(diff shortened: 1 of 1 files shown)


src/Orleans/IDs/UniqueKey.cs
```diff
@@ -114,7 +114,7 @@ private static UniqueKey NewKey(ulong n0, ulong n1, Category category, long type
             // in the string representation of a key, we grab the least significant half of n1.
             // therefore, if n0 is non-zero and n1 is 0, then the string representation will always be
             // 0x0 and not useful for identification of the grain.
-            if (n1 == 0 && n1 != 0)
+            if (n1 == 0 && n0 != 0)
                 throw new ArgumentException("n0 cannot be zero unless n1 is non-zero.", "n0");
```
jdom: If I read the exception message correctly, then the comparison should be inverted (essentially `n0 == 0 && n1 != 0`). Either that, or the exception message also needs fixing.

danielkrainas: I was working off the comment since it stated the logic behind the check. I'll fix the message.

jdom: TBH, I'm not familiar with it at this point to know whether the code should implement what's in the comment or in the exception message.

sergeybykov: It seems to me this whole check is wrong, and that it only survived because of the incorrect condition that is always `fasle`. It is totally legal for both `n0` and `n1` to be 0. For example, the very next overload `NewKey(long longKey, ` does
```csharp
var n1 = unchecked((ulong)longKey);
return NewKey(0, n1, category, typeData, keyExt);
```

where longKey can be zero.

Unless I'm overlooking something, I think we should remove this check altogether.

danielkrainas:
I removed the check altogether per @sergeybykov 's comment.

jdom:
Thank you @danielkrainas for finding this inconsistency and fixing it :)

PR #2734, opened 2017-02-15
Fix a cleanup issue in TestCluster after a failure

jdom:
Cleanup logic threw another exception that ended up masquerading the real exception and potentially leaking an initialized silo

### DIFF
--- a/src/OrleansTestingHost/TestCluster.cs
+++ b/src/OrleansTestingHost/TestCluster.cs
@@ -455,7 +455,7 @@ namespace Orleans.TestingHost
                 }
                 catch (Exception)
                 {
-                    this.additionalSilos.AddRange(siloStartTasks.Where(t => t.Exception != null).Select(t => t.Result));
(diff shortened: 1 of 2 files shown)


src/OrleansTestingHost/TestCluster.cs
```diff
@@ -455,7 +455,7 @@ public void InitializeClient()
                 }
                 catch (Exception)
                 {
-                    this.additionalSilos.AddRange(siloStartTasks.Where(t => t.Exception != null).Select(t => t.Result));
+                    this.additionalSilos.AddRange(siloStartTasks.Where(t => t.Exception == null).Select(t => t.Result));
```
xiazen: synced offline, we will add a logic to clean up silos which failed to start to enhance this fix.

ReubenBond: @xiazen in this pr or a later pr? I'll leave it to you two

jdom: Done

PR #2736, opened 2017-02-16
Orleans HelloWorld sample targeting .NET Core

jdom:
Sample using Orleans 2.0 (preview) to run a hello world in .NET Core.

### DIFF
--- a/Samples/HelloWorld.NetCore/NuGet.Config
+++ b/Samples/HelloWorld.NetCore/NuGet.Config
@@ -1,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="Orleans-local-vNext" value="..\..\vNext\Binaries\Debug\" />
+    <remove key="Orleans-local" />
+  </packageSources>
+</configuration>
--- a/Samples/HelloWorld.NetCore/Readme.md
+++ b/Samples/HelloWorld.NetCore/Readme.md
@@ -1,8 +1,30 @@
 # HelloWorld.NetCore
 Orleans HelloWorld sample targeting .NET Core
 
+## Building the Sample
 If you are trying to use a version of Orleans that is not yet released, first run `Build.cmd netstandard` from the root directory of the repository.
 
 Note: If you need to reinstall packages (for example, after making changes to the Orleans runtime and rebuilding the packages), just manually delete all Orleans packages from `(rootfolder)/Samples/{specific-sample}/packages/` and re-run `Build.cmd netstandard`. You might sometimes also need to clean up the NuGet cache folder.
 
 You can then compile as usual, build solution.
+
+```
+dotnet restore
+dotnet build
+dotnet publish
+```
+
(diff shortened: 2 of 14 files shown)


Samples/HelloWorld.NetCore/Readme.md
```diff
@@ -0,0 +1,30 @@
...
+You can then compile as usual, build solution.
+
+```
+dotnet restore
+dotnet build
+dotnet publish
+```
+
+## Running the Sample
+It's important to publish the apps before running them, as Orleans loads several assemblies via reflection and they need to be colocated.
```
jdom: I'd like to make it easier in the future if possible, but unfortunately right now it requires that you Publish the app first

ReubenBond: That's unfortunate :( What can we do about this? Reflect on the entry exe and load all referenced assemblies?

jdom:

Samples/HelloWorld.NetCore/Readme.md
jdom: I don't know at this point. I think the current issue came from OrleansProviders, as it was not referenced, but for some reason the client needed it... I need to investigate a little bit more. In the meantime, I don't think we should hold off from merging the sample itself because of an issue with Orleans. We can iterate more, and it's easier for others if this is in the main repo.

Samples/HelloWorld.NetCore/NuGet.Config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="Orleans-local-vNext" value="..\..\vNext\Binaries\Debug\" />
+    <remove key="Orleans-local" />
```
attilah: After clear we don't need Orleans-local removal I think.
What about adding the myget feed here to me it F5 runnable without building Orleans?

Samples/HelloWorld.NetCore/Readme.md
attilah: AssemblyLoading FTW :-)

jdom: Yeah, although what is odd is that I believe it required the client console app to have OrleansProviders, and that is not right in the first place, unless you are actually expecting to use stream providers or something like that. Nevertheless, I'll investigate later and separately from this PR (or someone else can investigate too if this PR is merged, as it's less friction that way)

Samples/HelloWorld.NetCore/NuGet.Config
jdom: Yeah... it was after playing with it for a while. I did notice it before pushing and decided to keep it, as ideally I would just want to remove the clear, but it did not work at least with my local setup.
Agreed on adding MyGet feed, it's just that it's outdated, and currently I noticed that the build scripts for nuget packages were modified, so if we build locally, it doesn't append a version that is higher than what was published to myget 2 months ago. I actually wanted to sync up with you on that.

PR #2737, opened 2017-02-16
Change CodeGen classes to be non-static

ReubenBond:
In preparation for non-static grain clients and non-static SerializationManager, we need to also make this class non-static.

Simple change, functionals pass.

### DIFF
--- a/src/Orleans/AssemblyLoader/AssemblyLoader.cs
+++ b/src/Orleans/AssemblyLoader/AssemblyLoader.cs
@@ -82,7 +82,7 @@ namespace Orleans.Runtime
             return discoveredAssemblyLocations;
         }
 #endif
-        public static T TryLoadAndCreateInstance<T>(string assemblyName, Logger logger) where T : class
+        public static T TryLoadAndCreateInstance<T>(string assemblyName, Logger logger, IServiceProvider serviceProvider) where T : class
         {
             try
             {
@@ -90,19 +90,18 @@ namespace Orleans.Runtime
                 var foundType =
                     TypeUtils.GetTypes(
                         assembly,
-                        type =>
-                        typeof(T).IsAssignableFrom(type) && !type.GetTypeInfo().IsInterface
-                        && type.GetConstructor(Type.EmptyTypes) != null, logger).FirstOrDefault();
+                        type => typeof(T).IsAssignableFrom(type) && !type.GetTypeInfo().IsInterface,
+                        logger).FirstOrDefault();
                 if (foundType == null)
                 {
                     return null;
                 }
 
-                return (T)Activator.CreateInstance(foundType, true);
+                return (T)ActivatorUtilities.CreateInstance(serviceProvider, foundType);
             }
-            catch (FileNotFoundException exception)
+            catch (FileNotFoundException)
             {
-                logger.Warn(ErrorCode.Loader_TryLoadAndCreateInstance_Failure, exception.Message, exception);
+                logger.Info(ErrorCode.Loader_TryLoadAndCreateInstance_Failure, $"Failed to find assembly {assemblyName} to create instance of {typeof(T)}.");
                 return null;
             }
             catch (Exception exc)
--- a/src/Orleans/CodeGeneration/CodeGeneratorManager.cs
+++ b/src/Orleans/CodeGeneration/CodeGeneratorManager.cs
@@ -1,3 +1,5 @@
+using Microsoft.Extensions.DependencyInjection;
+
 namespace Orleans.CodeGeneration
 {
     using System;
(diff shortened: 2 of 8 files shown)


src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -82,27 +82,26 @@ internal class AssemblyLoader
...
                 }
 
-                return (T)Activator.CreateInstance(foundType, true);
+                return (T)ActivatorUtilities.CreateInstance(serviceProvider, foundType);
             }
-            catch (FileNotFoundException exception)
+            catch (FileNotFoundException)
             {
-                logger.Warn(ErrorCode.Loader_TryLoadAndCreateInstance_Failure, exception.Message, exception);
+                logger.Info(ErrorCode.Loader_TryLoadAndCreateInstance_Failure, $"Failed to find assembly {assemblyName} to create instance of {typeof(T)}.");
```
ReubenBond: This should never have been a warning. The method is explicitly called TryLoad..., and it looks very ugly in user's logs (it spews red text on startup). Hopefully it's fine to fix this here instead of creating a new PR.

src/Orleans/CodeGeneration/CodeGeneratorManager.cs
```diff
@@ -30,20 +32,27 @@ internal static class CodeGeneratorManager
...
+
         /// <summary>
         /// Loads the code generator on demand
         /// </summary>
-        public static void Initialize()
+        public void Initialize()
         {
-            codeGeneratorInstance = LoadCodeGenerator();
-            codeGeneratorCacheInstance = codeGeneratorInstance as ICodeGeneratorCache;
+            this.codeGeneratorInstance = this.serviceProvider.GetService<IRuntimeCodeGenerator>() ?? this.LoadCodeGenerator();
```
jdom: what's the purpose of this fallback?

src/Orleans/AssemblyLoader/AssemblyLoader.cs
jdom: Breaking change. We need to update changelog for these changes on how we load stuff.

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -82,27 +82,26 @@ internal class AssemblyLoader
...
-                        && type.GetConstructor(Type.EmptyTypes) != null, logger).FirstOrDefault();
+                        type => typeof(T).IsAssignableFrom(type) && !type.GetTypeInfo().IsInterface,
+                        logger).FirstOrDefault();
                 if (foundType == null)
                 {
                     return null;
                 }
 
-                return (T)Activator.CreateInstance(foundType, true);
+                return (T)ActivatorUtilities.CreateInstance(serviceProvider, foundType);
```
jdom: Breaking change. We need to update changelog for these changes on how we load stuff.

src/Orleans/CodeGeneration/CodeGeneratorManager.cs
```diff
@@ -107,9 +116,9 @@ public static void AddGeneratedAssembly(string targetAssemblyName, GeneratedAsse
...
         /// </summary>
         /// <returns>The code generator.</returns>
-        private static IRuntimeCodeGenerator LoadCodeGenerator()
+        private IRuntimeCodeGenerator LoadCodeGenerator()
         {
-            IRuntimeCodeGenerator result = AssemblyLoader.TryLoadAndCreateInstance<IRuntimeCodeGenerator>(CodeGenAssemblyName, Log);
+            IRuntimeCodeGenerator result = AssemblyLoader.TryLoadAndCreateInstance<IRuntimeCodeGenerator>(CodeGenAssemblyName, Log, this.serviceProvider);
             if (result == null)
             {
                 Log.Warn(
```
jdom: Similar to the other change you made, shouldn't this be an Info?

src/Orleans/CodeGeneration/CodeGeneratorManager.cs
ReubenBond: If the user injected a code generator, that should be used instead of loading the default assembly.

src/Orleans/AssemblyLoader/AssemblyLoader.cs
ReubenBond: This is in-line with the other change which does the same thing, so it's more consistent now

jdom: Yup, I agree, I'm definitely not pushing back, just that we need to document it (as I believe we didn't document it back then when you made the first change in #2721)

PR #2738, opened 2017-02-16
GrainReferences hold a reference to a RuntimeClient instance

ReubenBond:
One of the key pieces for non-static clients. GrainReference instances must hold a reference to the runtime.

There are a few methods needed on `IGrainFactory` and `IInternalGrainFactory` to support these grain references, most important is the ability to bind a `GrainReference` instance to the runtime.

### DIFF
--- a/src/Orleans/Core/Exceptions.cs
+++ b/src/Orleans/Core/Exceptions.cs
@@ -219,7 +219,7 @@ namespace Orleans.Runtime
         private static string CreateMessage(GrainReference grainReference)
         {
             return $"Attempted to use a GrainReference which has not been bound to the runtime: {grainReference.ToDetailedString()}." +
-                   " Use the Bind method to bind this reference to the runtime.";
+                   $" Use the {nameof(IGrainFactory)}.{nameof(IGrainFactory.BindGrainReference)} method to bind this reference to the runtime.";
         }
 
         internal GrainReferenceNotBoundException(string msg) : base(msg) { }
--- a/src/Orleans/Core/Exceptions.cs
+++ b/src/Orleans/Core/Exceptions.cs
@@ -207,5 +207,29 @@ namespace Orleans.Runtime
         { }
 #endif
     }
+
+    /// <summary>
+    /// Indicates that a <see cref="GrainReference"/> was not bound to the runtime before being used.
+    /// </summary>
+    [Serializable]
+    public class GrainReferenceNotBoundException : OrleansException
+    {
+        internal GrainReferenceNotBoundException(GrainReference grainReference) : base(CreateMessage(grainReference)) { }
+
+        private static string CreateMessage(GrainReference grainReference)
+        {
+            return $"Attempted to use a GrainReference which has not been bound to the runtime: {grainReference.ToDetailedString()}." +
+                   " Use the Bind method to bind this reference to the runtime.";
+        }
+
+        internal GrainReferenceNotBoundException(string msg) : base(msg) { }
+        internal GrainReferenceNotBoundException(string message, Exception innerException) : base(message, innerException) { }
+
+#if !NETSTANDARD
+        protected GrainReferenceNotBoundException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
+        { }
+#endif
+    }
 }
 
--- a/src/Orleans/Core/IInternalGrainFactory.cs
+++ b/src/Orleans/Core/IInternalGrainFactory.cs
@@ -47,5 +47,21 @@ namespace Orleans
         /// A reference to <paramref name="grain"/> which implements <paramref name="interfaceType"/>.
         /// </returns>
         object Cast(IAddressable grain, Type interfaceType);
+
+        /// <summary>
+        /// Gets a reference to the grain with the provided id.
+        /// </summary>
+        /// <typeparam name="TGrainInterface">The grain reference interface type.</typeparam>
+        /// <param name="grainId">The grain id.</param>
+        /// <returns>A reference to the grain with the provided id.</returns>
+        TGrainInterface GetGrain<TGrainInterface>(GrainId grainId);
+
+        /// <summary>
+        /// Gets a reference to the grain with the provided id.
+        /// </summary>
+        /// <param name="grainId">The grain id.</param>
+        /// <param name="genericArguments">The generic type arguments.</param>
+        /// <returns>A reference to the grain with the provided id.</returns>
+        GrainReference GetGrain(GrainId grainId, string genericArguments = null);
     }
 }
--- a/src/Orleans/Core/IInternalGrainFactory.cs
+++ b/src/Orleans/Core/IInternalGrainFactory.cs
@@ -54,7 +54,7 @@ namespace Orleans
         /// <typeparam name="TGrainInterface">The grain reference interface type.</typeparam>
         /// <param name="grainId">The grain id.</param>
         /// <returns>A reference to the grain with the provided id.</returns>
-        TGrainInterface GetGrain<TGrainInterface>(GrainId grainId);
+        TGrainInterface GetGrain<TGrainInterface>(GrainId grainId) where TGrainInterface : IAddressable;
 
         /// <summary>
         /// Gets a reference to the grain with the provided id.
--- a/src/Orleans/Runtime/GrainReference.cs
+++ b/src/Orleans/Runtime/GrainReference.cs
@@ -54,6 +54,7 @@ namespace Orleans.Runtime
         /// </summary>
         protected internal readonly SiloAddress SystemTargetSilo;
 
+        [NonSerialized]
         private IRuntimeClient runtimeClient;
 
         /// <summary>
--- a/src/Orleans/Runtime/GrainReference.cs
+++ b/src/Orleans/Runtime/GrainReference.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Concurrent;
-using System.Reflection;
 using System.Runtime.Serialization;
 using System.Text;
 using System.Threading.Tasks;
@@ -15,8 +14,6 @@ namespace Orleans.Runtime
     [Serializable]
     public class GrainReference : IAddressable, IEquatable<GrainReference>, ISerializable
     {
-        private static readonly Action<Message, TaskCompletionSource<object>> ResponseCallbackDelegate = ResponseCallback;
-
         private readonly string genericArguments;
         private readonly GuidId observerId;
         
@@ -41,12 +38,23 @@ namespace Orleans.Runtime
         
         private bool HasGenericArgument { get { return !String.IsNullOrEmpty(genericArguments); } }
 
+        private IRuntimeClient RuntimeClient
+        {
+            get
+            {
+                if (this.runtimeClient == null) throw new GrainReferenceNotBoundException(this);
+                return this.runtimeClient;
+            }
+        }
+
         internal GrainId GrainId { get; private set; }
 
         /// <summary>
         /// Called from generated code.
         /// </summary>
         protected internal readonly SiloAddress SystemTargetSilo;
+
+        private IRuntimeClient runtimeClient;
 
         /// <summary>
         /// Whether the runtime environment for system targets has been initialized yet.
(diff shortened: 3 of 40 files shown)


src/Orleans/Runtime/GrainReference.cs
```diff
@@ -531,14 +551,15 @@ protected internal static object DeserializeGrainReference(Type t, IDeserializat
...
             var genericArg = reader.ReadString();
-            if (String.IsNullOrEmpty(genericArg))
+            if (string.IsNullOrEmpty(genericArg))
                 genericArg = null;
-
+            
             if (expectObserverId)
             {
-                return NewObserverGrainReference(id, observerId);
+                return NewObserverGrainReference(id, observerId, Runtime.RuntimeClient.Current);
```
ReubenBond: Currently, the global `RuntimeClient.Current` is used. In a later PR, that will be changed to retrieve the `IRuntimeClient` from the context.

ReubenBond:
@dotnet-bot test this please

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -118,27 +128,37 @@ private GrainReference(GrainId grainId, string genericArgument, SiloAddress syst
         /// </summary>
         /// <param name="other">The reference to copy.</param>
         protected GrainReference(GrainReference other)
-            : this(other.GrainId, other.genericArguments, other.SystemTargetSilo, other.ObserverId) { }
+            : this(other.GrainId, other.genericArguments, other.SystemTargetSilo, other.ObserverId, other.RuntimeClient) { }
```
jdom: since the `RuntimeClient` property throws if unbound, is this by design? or did you intend to copy the field instead without throwing?

src/Orleans/Core/Exceptions.cs
```diff
@@ -207,5 +207,29 @@ protected ClientNotAvailableException(SerializationInfo info, StreamingContext c
...
+    /// </summary>
+    [Serializable]
+    public class GrainReferenceNotBoundException : OrleansException
+    {
+        internal GrainReferenceNotBoundException(GrainReference grainReference) : base(CreateMessage(grainReference)) { }
+
+        private static string CreateMessage(GrainReference grainReference)
+        {
+            return $"Attempted to use a GrainReference which has not been bound to the runtime: {grainReference.ToDetailedString()}." +
+                   " Use the Bind method to bind this reference to the runtime.";
```
jdom: I see that the `GrainReference.Bind` method is internal Maybe this should point to `{nameof(IGrainFactory)}.{nameof(BindGrainRereference)}` instead?

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -15,8 +14,6 @@ namespace Orleans.Runtime
     [Serializable]
     public class GrainReference : IAddressable, IEquatable<GrainReference>, ISerializable
     {
-        private static readonly Action<Message, TaskCompletionSource<object>> ResponseCallbackDelegate = ResponseCallback;
```
jdom: I think the point of this was to avoid many allocations of the cast to Action. Perhaps removing the static keyword is good enough? Or would the allocations per GrainReference instance would be so few that this optimization doesn't make any more sense?

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -378,7 +398,7 @@ private void CallClientInvokeCallback(InvokeMethodRequest request)
         }
 
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
-        private static void ResponseCallback(Message message, TaskCompletionSource<object> context)
+        private void ResponseCallback(Message message, TaskCompletionSource<object> context)
```
jdom: what's non-static about this method? Am I not seeing a field access here? Maybe mark them with `this.`

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -531,14 +551,21 @@ protected internal static object DeserializeGrainReference(Type t, IDeserializat
...
-                return NewObserverGrainReference(id, observerId);
+                result = NewObserverGrainReference(id, observerId, null);
+            }
+            else
+            {
+                result = FromGrainId(id, null, genericArg, silo);
             }
-            return FromGrainId(id, genericArg, silo);
+
+            context.GrainFactory?.BindGrainReference(result);
```
jdom: could there be gotchas with this? I presume not, but maybe in a multi-cluster scenario or something of that sort?

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -41,13 +38,24 @@ public class GrainReference : IAddressable, IEquatable<GrainReference>, ISeriali
...
+        }
+
         internal GrainId GrainId { get; private set; }
 
         /// <summary>
         /// Called from generated code.
         /// </summary>
         protected internal readonly SiloAddress SystemTargetSilo;
 
+        private IRuntimeClient runtimeClient;
```
sergeybykov: >private IRuntimeClient runtimeClient; [](start = 8, length = 37)

Doesn't this field need to be marked as `[NonSerialized]`?

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -137,13 +137,13 @@ public static object GetUninitializedObjectWithFormatterServices(Type type)
...
         private static SerializationContext CurrentSerializationContext
-            => serializationContext ?? (serializationContext = new SerializationContext());
+            => serializationContext ?? (serializationContext = new SerializationContext(null));
         
         [ThreadStatic]
         private static DeserializationContext deserializationContext;
 
         private static DeserializationContext CurrentDeserializationContext
-            => deserializationContext ?? (deserializationContext = new DeserializationContext());
+            => deserializationContext ?? (deserializationContext = new DeserializationContext(null));
```
jdom: these look error prone... why are they used? Isn't it better to return null if unset?

src/Orleans/Core/IInternalGrainFactory.cs
```diff
@@ -47,5 +47,21 @@ TGrainInterface GetSystemTarget<TGrainInterface>(GrainId grainId, SiloAddress de
...
         /// </returns>
         object Cast(IAddressable grain, Type interfaceType);
+
+        /// <summary>
+        /// Gets a reference to the grain with the provided id.
+        /// </summary>
+        /// <typeparam name="TGrainInterface">The grain reference interface type.</typeparam>
+        /// <param name="grainId">The grain id.</param>
+        /// <returns>A reference to the grain with the provided id.</returns>
+        TGrainInterface GetGrain<TGrainInterface>(GrainId grainId);
```
jdom: no type constraints at all? Maybe call it something different then, such as GetReference or GetAddressableReference?

sergeybykov (approved):

test/NonSiloTests/Serialization/BuiltInSerializerTests.cs
```diff
@@ -727,9 +729,9 @@ public void Serialize_Uri_Multithreaded(SerializerToUse serializerToUse)
         [InlineData(SerializerToUse.NoFallback)]
         public void Serialize_GrainReference(SerializerToUse serializerToUse)
         {
-            InitializeSerializer(serializerToUse);
+            var environment = InitializeSerializer(serializerToUse);
             GrainId grainId = GrainId.NewId();
-            GrainReference input = GrainReference.FromGrainId(grainId);
+            GrainReference input = environment.InternalGrainFactory.GetGrain(grainId);
 
             object deserialized = OrleansSerializationLoop(input);
```
jdom: you probably want to assert that once deserialized, it is bound to the runtime client that is in the context? Or are you asserting that somewhere else?

src/Orleans/Runtime/GrainReference.cs
ReubenBond: Yes, that would be best, thank you.

src/Orleans/Core/Exceptions.cs
```diff
@@ -207,5 +207,29 @@ protected ClientNotAvailableException(SerializationInfo info, StreamingContext c
         { }
 #endif
     }
+
+    /// <summary>
+    /// Indicates that a <see cref="GrainReference"/> was not bound to the runtime before being used.
+    /// </summary>
+    [Serializable]
+    public class GrainReferenceNotBoundException : OrleansException
```
ReubenBond: My understanding was that all of our exceptions should inherit from it. Maybe we should have a 'TransientException' base class if there is supposed to be any different behaviour for transient exceptions.

src/Orleans/Core/Exceptions.cs
ReubenBond: Thanks, that's better

src/Orleans/Core/Exceptions.cs
jdom: Asked Sergey directly, and apparently we don't have that distinction... all exceptions that we generate typically inherit from OrleansException, so we are good here.

src/Orleans/Runtime/GrainReference.cs
ReubenBond: You're totally right, I'll fix it.

src/Orleans/Runtime/GrainReference.cs
ReubenBond: Good catch. It was originally the field, but I changed it to the property for testing and didn't revert it. I'll change that.

src/Orleans/Runtime/GrainReference.cs
ReubenBond: Ah this is one of those things which changes in a subsequent PR (non-static SerializationManager). In this case, it's `message.BodyObject` which will change. My bad. I can revert those changes and push them into the later PR, but I wanted to extract as much as possible from that PR since it's too heavy for people to review.
image

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: No, this is fine - it will be set to RuntimeClient.Current upon usage if it wasn't set here (because of static type initialization order)

This is fixed in the subsequent PR for non-static serialization manager:
image

src/Orleans/Runtime/GrainReference.cs
jdom: Ah, that would explain it, do as you prefer if you want to revert it to keep the change relevant to where it's needed, but I don't care. The reason it caught my attention is because of the undo of the other optimization for ResponseCallbackDelegate.

src/Orleans/Serialization/SerializationManager.cs
jdom: Well, type initialization order is the "error prone" part I'm worried about. If this is fixed in the upcoming PR, then I don't mind, it's just that I'd rather make it throw explicitly than having a context that sometimes works and sometimes doesn't

test/NonSiloTests/Serialization/BuiltInSerializerTests.cs
ReubenBond: I'll add the check. There are other tests which indirectly test this by making a call on the deserialized grain

src/Orleans/Core/IInternalGrainFactory.cs
ReubenBond: I'll add an `IAddressable` constraint

jdom: Still, later on I saw usages of this not just for grains, but extensions and system targets... maybe it's better to remove the word Grain from the method name anyway

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: It's not error prone, passing `null` there isn't an issue because `DeserializationContext` will coalesce the value with `RuntimeClient.Current` upon usage. It's ugly and I don't like it, but the fix is in the PR queue

jdom: ah, yes, I see it now. Carry on, nothing to see here.

src/Orleans/Runtime/GrainReference.cs
ReubenBond: I've re-added that optimization in the latest commit

src/Orleans/Core/IInternalGrainFactory.cs
ReubenBond: I'm fine with that, the issue is that those things are all 'grain references' in our system, so the logical name might be 'GetGrainReference', but that's no better than GetGrain. Still, I'm fine calling it `GetAddressableReference` if you want to make it explicit - it's just that the distinction is a little shaky

jdom: it's OK... you are right, this is an overload of other methods with the same dilemma. Let's keep as is.

jamescarter-le:
@ReubenBond What is the recommended way of creating a GrainReference now that FromKeyString has gone internal? My storage provider serializers relied on this method to rehydrate a GrainReference from the ToKeyString method.

PR #2742, opened 2017-02-18
Update Hello World sample to Orleans 1.4.0

jdom:


jdom:
@dsarfati, you might want to check if the fix I've done in NonSiloTests is what you would expect, as I just mainly made it build and pass, but didn't stop to think if there's a better approach

dsarfati:
@jdom This looks good to me. Im working on building out a bigger test suite that I'm hoping to include in Orleans Contrib. Right now I support xunit logger, mockable streams, mockable storage, & mockable probe grains. I have been modeling the test framework after the [Akka Testkit](http://doc.akka.io/docs/akka/current/scala/testing.html)

Samples/HelloWorld/SiloTests/packages.config
```diff
@@ -1,10 +1,35 @@
...
+  <package id="System.Linq" version="4.1.0" targetFramework="net451" />
+  <package id="System.Linq.Expressions" version="4.1.0" targetFramework="net451" />
+  <package id="System.Reflection" version="4.1.0" targetFramework="net451" />
+  <package id="System.Reflection.Metadata" version="1.2.0" targetFramework="net451" />
+  <package id="System.Resources.ResourceManager" version="4.0.1" targetFramework="net451" />
+  <package id="System.Runtime" version="4.1.0" targetFramework="net451" />
+  <package id="System.Runtime.Extensions" version="4.1.0" targetFramework="net451" />
+  <package id="System.Runtime.InteropServices" version="4.1.0" targetFramework="net451" />
+  <package id="System.Threading" version="4.0.11" targetFramework="net451" />
+  <package id="System.Threading.Tasks" version="4.0.11" targetFramework="net451" />
```
xiazen: are those updated by visual studio tooling? this is a lot of new packages..
otherwise looks good to me

jdom: Yes, unfortunately, unless we move to the new VS2017 tooling or to packages.json for the nuget packages, we must include all the dependencies of our dependencies, and since we took System.Collections.Immutable 1.1.37 (main offender), there was an explosion of dependencies.
We prefer to keep the samples so that they can be opened in older versions of VS (especially the hello world), so there's not a lot we can do, I'm afraid.

jdom:
That's awesome @dsarfati, looking forward to seeing that. Please ping me when you have something to show

dsarfati:
@jdom Just put up the initial bit of code. Let me know what you think. [Link](https://github.com/dsarfati/OrleansUnitTestSilo). Ill be adding docs and a real ReadMe soon.

PR #2746, opened 2017-02-20
Improve GrainReference deserialization for BinaryFormatter & Json.NET

ReubenBond:
BinaryFormatter and Json.NET have a facility to pass context to types being deserialized. This PR plumbs `IGrainFactory` through that context and makes use of it during deserialization of GrainReference.

Originally I was intending to plumb some new `IExternalSerializerContext`, but figured it's best to be lightweight - we can add some new interface later if needed.

Unfortunately, while `StreamingContext` does exist on .NET Core, it's defined as a struct with no members. In other words, it's totally useless and exists only to keep method signatures consistent with Full Fx.

Functionals pass.

### DIFF
--- a/src/Orleans/Runtime/GrainReference.cs
+++ b/src/Orleans/Runtime/GrainReference.cs
@@ -726,12 +726,24 @@ namespace Orleans.Runtime
                 genericArg = null;
             genericArguments = genericArg;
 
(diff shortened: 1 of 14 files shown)


ReubenBond:
@dotnet-bot test this please

PR #2752, opened 2017-02-21
Add a NuGet package for OrleansGoogleUtils.dll

sergeybykov:
Add a NuGet package for OrleansGoogleUtils.dll.
Remove unnecessary reference to OrleansRuntime.dll.


jdom:
Is the package information also in the csproj for vNext (and in the csprojs that @attilah created for the main solution too)?

sergeybykov:
Update the nuspec to match the properties in vNext\src\OrleansGoogleUtils\OrleansGoogleUtils.csproj.

jdom:
Ah, I missed that this was already there in vNext. We should probably drop the reference to OrleansRuntime in vNext also

Issue #2751, opened 2017-02-21
Serializing State classes with null grain references fail in 1.4


ashkan-saeedi-mazdeh:
As @ReubenBond  and I discussed when I want to serialize a State using WriteStateAsync, if the state class contains a reference to a grain interface say `IPlayer` which was a `GrainWithStringKey` , the serializer (the generated code) throws an exception `ArgumentNullException` when calling `AsReference`

The test failed on a relatively lohng running project which was ok on previous Orleans versions. I can post more info if needed

ashkan-saeedi-mazdeh:
@sergeybykov  I rolled back to 1.3.1 but isn't this important enough to trigger a 1.4.x ?

ReubenBond:
This is fixed in #2756, but I agree that we should consider 1.4.1 given this and the DI issue which was also fixed.

sergeybykov:
@ashkan-saeedi-mazdeh Yes, we'll include the fix in 1.4.1.

PR #2756, opened 2017-02-21
Fix AsReference() in generated code for null values

### LINKED ISSUE #2750: Having a null grain reference as a property on a GrainState object results in an exception
Having a Grain state object fails to persist if a grain Interface property is unset (`null`)

Solution is to set EventGrainState.Metadata.Owner = to a grain reference.

**Code:**
```csharp
 public class EventGrainState
    {
        public bool IsInitialized { get; set; }
        public ObjectMetadata Metadata { get; set; } = new ObjectMetadata();
        public string Name { get; set; }
        public EventData Data { get; set; } = new EventData();
        public TimeSpan Interval { get; set; }
    }
   public class ObjectMetadata
    {
        public DateTime CreatedDate { get; internal set; }
        public DateTime ModifiedDate { get; internal set; }
        public IUserGrain Owner { get; internal set; }
    }
    public class EventData
    {
        public string EventType { get; set; }
        public string Parameters { get; set; }
    }
```

**Stack Trace:**

```Cannot pass null as an argument to AsReference
Parameter name: grain
   at Orleans.GrainExtensions.AsReference[TGrainInterface](IAddressable grain)
   at NRC.Services.Scheduler.Grains.OrleansCodeGenNRC_Services_Scheduler_Grains_ObjectMetadataSerializer.DeepCopier(Object original)
   at Orleans.Serialization.SerializationManager.DeepCopyInner(Object original)
   at NRC.Services.Scheduler.Grains.OrleansCodeGenNRC_Services_Scheduler_Grains_EventGrainStateSerializer.DeepCopier(Object original)
   at Orleans.Serialization.SerializationManager.DeepCopyInner(Object original)
   at Orleans.OrleansCodeGenOrleans_GrainStateSerializer`1.DeepCopier(Object original)
   at Orleans.Serialization.SerializationManager.<>c__DisplayClass61_0.<Register>b__0(Object obj)
   at Orleans.Serialization.SerializationManager.DeepCopyInner(Object original)
   at Orleans.Serialization.SerializationManager.DeepCopierHelper(Type t, Object original)
   at Orleans.Serialization.SerializationManager.DeepCopyInner(Object original)
   at Orleans.Serialization.SerializationManager.DeepCopy(Object original)
   at Orleans.Runtime.GrainReference.InvokeMethodAsync[T](Int32 methodId, Object[] arguments, InvokeMethodOptions options, SiloAddress silo)
   at Orleans.Storage.OrleansCodeGenMemoryStorageGrainReference.WriteStateAsync(String grainType, String grainId, IGrainState grainState)
   at Orleans.Storage.MemoryStorage.<WriteStateAsync>d__21.MoveNext()
```

ReubenBond:
Fix #2750 & #2751

### DIFF
--- a/src/OrleansCodeGenerator/SerializerGenerator.cs
+++ b/src/OrleansCodeGenerator/SerializerGenerator.cs
@@ -595,12 +595,12 @@ namespace Orleans.CodeGenerator
                     && this.FieldInfo.FieldType.GetTypeInfo().IsInterface
                     && !typeof(IGrainObserver).IsAssignableFrom(this.FieldInfo.FieldType))
                 {
-                    var getAsReference = getValueExpression.Member(
+                    var getAsReference = getValueExpression.ConditionalMember(
                         (IAddressable grain) => grain.AsReference<IGrain>(),
                         this.FieldInfo.FieldType);
 
                     // If the value is not a GrainReference, convert it to a strongly-typed GrainReference.
-                    // C#: !(value is GrainReference) ? value.AsReference<TInterface>() : value;
+                    // C#: !(value is GrainReference) ? value?.AsReference<TInterface>() : value;
                     deepCopyValueExpression =
                         SF.ConditionalExpression(
                             SF.PrefixUnaryExpression(
--- a/src/OrleansCodeGenerator/SerializerGenerator.cs
+++ b/src/OrleansCodeGenerator/SerializerGenerator.cs
@@ -595,23 +595,27 @@ namespace Orleans.CodeGenerator
                     && this.FieldInfo.FieldType.GetTypeInfo().IsInterface
                     && !typeof(IGrainObserver).IsAssignableFrom(this.FieldInfo.FieldType))
                 {
-                    var getAsReference = getValueExpression.ConditionalMember(
+                    var getAsReference = getValueExpression.Member(
                         (IAddressable grain) => grain.AsReference<IGrain>(),
                         this.FieldInfo.FieldType);
 
                     // If the value is not a GrainReference, convert it to a strongly-typed GrainReference.
-                    // C#: !(value is GrainReference) ? value?.AsReference<TInterface>() : value;
+                    // C#: (value == null || value is GrainReference) ? value : value.AsReference<TInterface>()
                     deepCopyValueExpression =
                         SF.ConditionalExpression(
-                            SF.PrefixUnaryExpression(
-                                SyntaxKind.LogicalNotExpression,
-                                SF.ParenthesizedExpression(
+                            SF.ParenthesizedExpression(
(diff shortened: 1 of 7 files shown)


src/OrleansCodeGenerator/SerializerGenerator.cs
```diff
@@ -595,12 +595,12 @@ public ExpressionSyntax GetGetter(ExpressionSyntax instance, ExpressionSyntax se
...
                     && !typeof(IGrainObserver).IsAssignableFrom(this.FieldInfo.FieldType))
                 {
-                    var getAsReference = getValueExpression.Member(
+                    var getAsReference = getValueExpression.ConditionalMember(
                         (IAddressable grain) => grain.AsReference<IGrain>(),
                         this.FieldInfo.FieldType);
 
                     // If the value is not a GrainReference, convert it to a strongly-typed GrainReference.
-                    // C#: !(value is GrainReference) ? value.AsReference<TInterface>() : value;
+                    // C#: !(value is GrainReference) ? value?.AsReference<TInterface>() : value;
```
jdom: I assume this is just explanatory, but it really doesn't require any c# 6 related tooling in the dev or runtime environment, correct?

ReubenBond: No, it does. Do we still support pre-6?

jdom: *I believe* we support users building their Orleans apps without c#6... Orleans development with pre-6 is not supported.

jdom: but it's not like I tried it or anything since we started using c# 6 features... so maybe I'm under a wrong assumption.

ReubenBond:
@dotnet-bot test this please

src/OrleansCodeGenerator/SerializerGenerator.cs
ReubenBond: I don't mind either way. I can change the implementation to perform a manual null check, no worries :)

I'll make the change and submit a revision

ReubenBond:
@dotnet-bot test this please

src/OrleansCodeGenerator/SerializerGenerator.cs
jdom: Ah, I missed the update, is the change already done in what's in the branch now?

ReubenBond:
@jdom I've pushed the update - apologies for the delay (was doing SM PR).

Sample of generated code:
```C#
(input.@Writer == null || input.@Writer is global::Orleans.Runtime.GrainReference) ? input.@Writer : input.@Writer.@AsReference<global::UnitTests.GrainInterfaces.IMultifacetWriter>()
```

PR #2759, opened 2017-02-22
Wait (for await)

amccool:
should need an await before accessing the result

### DIFF
--- a/src/Documentation/Advanced-Concepts/Powershell-Client.md
+++ b/src/Documentation/Advanced-Concepts/Powershell-Client.md
@@ -113,7 +113,7 @@ Takes no parameters and when called, if the `GrainClient` is initialized will gr
 
 Wrapper around `GrainClient.GrainFactory.GetGrain<T>()` and its overloads.
 
-The mandatory parameter is `-GrainType` and the `-XXXKey` for the current Grain key types supported by Orleans (`string`, `Guid`, `long`) and also the `-KeyExrension` that can be used on Grains with compound keys.
+The mandatory parameter is `-GrainType` and the `-XXXKey` for the current Grain key types supported by Orleans (`string`, `Guid`, `long`) and also the `-KeyExtension` that can be used on Grains with compound keys.
 
 This Cmdlet return a grain reference of the type passed by as parameter on `-GrainType`.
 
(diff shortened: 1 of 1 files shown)


src/Documentation/Advanced-Concepts/Powershell-Client.md
```diff
@@ -146,7 +150,7 @@ PS> Stop-GrainClient
 
 ```
 
-Thats it for now. We plan to update this page as we introduce more Cmdlets like use Observers, Streams and other Orleans core features more natively on Powershell.
+That is it for now. We plan to update this page as we introduce more Cmdlets like use Observers, Streams and other Orleans core features more natively on Powershell.
```
ReubenBond: Could we also change the last sentence to: We plan to update this page as we introduce Cmdlets for other Orleans features such as Observers and Streams. ?

ReubenBond:
Thanks, @amccool :) Could you also remove the blank lines at the top of each powershell snippet?
image

It's rendering poorly on DocFx

src/Documentation/Advanced-Concepts/Powershell-Client.md
amccool: so just removing the `Thats it for now` blurb?

ReubenBond: @amccool no, that part is fine - I just meant to replace the sentence after it

src/Documentation/Advanced-Concepts/Powershell-Client.md
```diff
@@ -137,7 +133,11 @@ PS> $grainId = [System.Guid]::Parse("A4CF7B5D-9606-446D-ACE9-C900AC6BA3AD")
 
 PS> $grain = Get-Grain -GrainType $grainInterfaceType -GuidKey $grainId
 
-PS> $message = $grain.SayHelloTo("Gutemberg").Result
+PS> $sayHelloTask = $grain.SayHelloTo("Gutemberg")
+
+PS> $sayHelloTask.Wait();
+
+PS> $message = $sayHelloTask.Result
```
jdom: not sure I understand this change. Unless Wait() behaves differently in Powershell, it is doing a blocking call until the result is calculated, and then you just access the pre-calculated Result. Why not just call Result, which will block until the result is calculated too? Seems easier to have 1 line than 3.
Of course, if you want to parallelize, then that's another scenario, and you might want to store each task separately, but not necessarily use Wait() either, just Task.WaitAll.

amccool: I will try to double check.  BTW it looks like psgallery does not have the module published

PR #2769, opened 2017-02-23
Consolidate Newtonsoft.Json version

### LINKED ISSUE #2766: Microsoft.Orleans.ServiceFabric.dll references Newtonsoft.Json v9 but its NuGet does not have it as a dependency

ReubenBond:
Fixes #2766.

The Service Fabric NuGet packages were missing the Newtonsoft.Json dependency.

The core Orleans package was using version 7.0.1 in net451 but 9.0.1 for netstandard, so this consolidates that, too.

### DIFF
--- a/src/OrleansVSTools/OrleansVSTools/OrleansVSTools.csproj
+++ b/src/OrleansVSTools/OrleansVSTools/OrleansVSTools.csproj
@@ -202,8 +202,8 @@
       <Link>Packages\Microsoft.Extensions.DependencyInjection.Abstractions.1.0.0.nupkg</Link>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
-    <Content Include="..\packages\Newtonsoft.Json.7.0.1\Newtonsoft.Json.7.0.1.nupkg">
-      <Link>Packages\Newtonsoft.Json.7.0.1.nupkg</Link>
+    <Content Include="..\packages\Newtonsoft.Json.9.0.1\Newtonsoft.Json.9.0.1.nupkg">
+      <Link>Packages\Newtonsoft.Json.9.0.1.nupkg</Link>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
     <Content Include="..\packages\System.Collections.4.0.11\System.Collections.4.0.11.nupkg">
--- a/src/OrleansVSTools/OrleansVSTools/OrleansVSTools.csproj
+++ b/src/OrleansVSTools/OrleansVSTools/OrleansVSTools.csproj
@@ -202,8 +202,8 @@
       <Link>Packages\Microsoft.Extensions.DependencyInjection.Abstractions.1.0.0.nupkg</Link>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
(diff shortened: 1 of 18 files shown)


sergeybykov:
Microsoft.Orleans.ServiceFabric has a dependency on Microsoft.Orleans.Core which already has a dependency on Newtonsoft.Json 7.0.1. Would this create a multi-version problem? Shouldn't we move Microsoft.Orleans.Core to 9.0.1 instead?

jdom:
Did a quick scan and found a few of these that would need to be updated (there might be other kind of references that I didn't check):
```
      <dependentAssembly>
        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-7.0.0.0" newVersion="7.0.0.0" />
      </dependentAssembly>
```

jdom:

jdom:
Hehe, I didn't expect you to update all the samples, since they use older Orleans nuget packages, and not the latest Orleans that is on master. Nevertheless, it's a safe thing to do, assuming you did this with NuGet itself, so it added or updated ALL relevant binding redirects. Was this done with the nuget tool or find & replace?

ReubenBond:
@jdom I did a search and replace, I believe it was comprehensive, though, I made sure not to clobber Azure Storage libs, which were also at version 7 in some instances

jdom:
What I'm worried about is not really the replace part, since it looks correct, but updating to 9.0.1 might cause a few additions of binding redirects if there weren't any before. If it's too much unnecessary work to try them out (most likely), I'd suggest that we revert updating the samples.  I think there's a plan to update all samples to Orleans 1.4.0 regardless

ReubenBond:
@jdom reverted - we can update them later

src/OrleansVSTools/OrleansVSTools/OrleansVSTools.csproj
```diff
@@ -202,8 +202,8 @@
       <Link>Packages\Microsoft.Extensions.DependencyInjection.Abstractions.1.0.0.nupkg</Link>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
-    <Content Include="..\packages\Newtonsoft.Json.7.0.1\Newtonsoft.Json.7.0.1.nupkg">
-      <Link>Packages\Newtonsoft.Json.7.0.1.nupkg</Link>
+    <Content Include="..\packages\Newtonsoft.Json.9.0.1\Newtonsoft.Json.9.0.1.nupkg">
+      <Link>Packages\Newtonsoft.Json.9.0.1.nupkg</Link>
```
jdom: If we update the VSTools project (which is still 1.4.0), then we are missing updating binding redirects in the template projects (or does the host project automatically adds redirects when the template unfolds?). Same as before, either don't update these tools at all, or do the full change. It's probably easier to just revert, as we don't tend to keep the OrleansVSTools up to date until we are ready to release, but you are free to choose 👅

ReubenBond: much easier to revert for now :)

attilah: LoL :-) I've an animgif in my mind ;-)

jdom:

PR #2771, opened 2017-02-23
[WIP]MemoryPubSub and its tests

xiazen:
MemoryPubsub is designed for the scenario where producer and consumer prefer to use local streaming infra for performance reasons. 
The current design provides the functionality to always provide producer and consumer local streaming infra to use. But also has various concerns such as persistence and consistency through different silos. It only fits in limited cases in current design. We need to solve the persistence issue to make it fit in broader use cases. 

the pattern is still under discussion. see #2811


src/Orleans/Providers/ClientProviderRuntime.cs
```diff
@@ -170,6 +173,8 @@ public IStreamPubSub PubSub(StreamPubSubType pubSubType)
                     return grainBasedPubSub;
                 case StreamPubSubType.ImplicitOnly:
                     return implictPubSub;
+                case StreamPubSubType.MemoryBaseOnly:
+                    return memoryPubSub;
```
jason-bragg: Other pubsub systems are global, memory is per stream provider, and silo only, so:
- Client should throw error if memory pub sub is requested
- Silo should create new memory pubsub each request.

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -106,6 +107,7 @@ public OutsideRuntimeClient(ClientConfiguration cfg, bool secondary = false)
             }
             
             var services = new ServiceCollection();
+            services.AddSingleton<MemoryPubSub>();
```
jason-bragg: Per stream provider, so transient..?

src/Orleans/Streams/PubSub/MemoryPubSub.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Orleans.Runtime;
+
+namespace Orleans.Streams.PubSub
+{
+    internal class MemoryPubSub : IStreamPubSub
+    {
+        private MemoryPubSubStore store;
```
jason-bragg: Not sure why this class wraps the store.  Seems like a simple pass-through?  Other pubsub systems are pass-through because they call to their global state (grain, or implicit subscription table).  Unclear why this one should do this though.

src/Orleans/Streams/PubSub/PubSubPublisherState.cs
```diff
@@ -14,17 +14,18 @@ internal class PubSubPublisherState : IEquatable<PubSubPublisherState>
         [JsonProperty]
         public StreamId Stream;
         [JsonProperty]
-        public GrainReference producerReference; // the field needs to be of a public type, otherwise we will not generate an Orleans serializer for that class.
+        public GrainReference producerReference => producerExtension as GrainReference; // the field needs to be of a public type, otherwise we will not generate an Orleans serializer for that class.
```
jason-bragg: why did you swap these?  I think this will break when persisted.

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
```diff
@@ -0,0 +1,258 @@
...
+            {
+                throw new OrleansException($"{this.GetType().Name} : Remove consumer failed");
+            }
+        }
+
+        public ICollection<PubSubPublisherState> Producers => producers.Keys;
+        public ICollection<PubSubSubscriptionState> Consumers => consumers.Keys;
+    }
+    
+    internal class MemoryPubSubStore
```
jason-bragg: Unless I'm mistaken, this is mostly the logic from the rendezvous grain, correct?  Can this logic be shared?

src/Orleans/Streams/PubSub/PubSubPublisherState.cs
xiazen: the old way would cast producerReference to be null if the producerExtension is not a GrainReference, which should be avoided. Since we do need the original producerExtension object in Producer getter

xiazen: see the change in this class's constructor

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
xiazen: Yes. Some of the logic can be shared. there's structure difference between MemoryPubSub and rendezvous grain. rendezvous grain keep track info for one stream, and MemoryPubsub keep track  info for one streamProvider. 
So this is non-trivia change. I will create an issue for this and come back

src/Orleans/Streams/PubSub/MemoryPubSub.cs
jason-bragg: I see value in this now.  NM :)

src/Orleans/Streams/PubSub/PubSubSubscriptionState.cs
```diff
@@ -45,6 +46,7 @@ internal enum SubscriptionStates
         {
             SubscriptionId = subscriptionId;
             Stream = streamId;
+            consumerExtension = streamConsumer;
             consumerReference = streamConsumer as GrainReference;
             state = SubscriptionStates.Active;
```
xiazen: @ReubenBond  hey Reuben, can changes in this file affect persistence? as in, when serializer serialize entity from persistence storage, the serialization would be problematic before and after this change?

xiazen: specific to Consumer property

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
```diff
@@ -0,0 +1,258 @@
...
+
+        public ISet<PubSubSubscriptionState> RegisterProducer(StreamId streamId, string streamProvider,
+            IStreamProducerExtension streamProducer)
+        {
+            var publisherState = new PubSubPublisherState(streamId, streamProducer);
+            MemoryPubSubStreamSubscriptions subscriptions = new MemoryPubSubStreamSubscriptions();
+            try
+            {
+                subscriptions.AddProducer(publisherState);
+                subscriptions = store.AddOrUpdate(streamId, subscriptions, (id, streamSubscriptions) =>
```
jason-bragg: GetOrAdd?

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
```diff
@@ -0,0 +1,258 @@
...
+            producers.AddOrUpdate(state, byte.MinValue, (publisherState, b) => b);
+        }
+
+        public void RemoveProducer(PubSubPublisherState state)
+        {
+            byte b;
+            bool re = producers.TryRemove(state, out b);
+            if (!re)
+            {
+                throw new OrleansException($"{this.GetType().Name} : Remove producer failed");
```
jason-bragg: Do we want to throw here?  Not sure, just asking, but it seems like the goal would be to ensure that the producer was no longer in the pubsub system, which is the case if false was returned.

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
```diff
@@ -0,0 +1,258 @@
...
+            consumers.AddOrUpdate(state, byte.MinValue, (subscriptionState, b) => b);
+        }
+
+        public void RemoveConsumer(PubSubSubscriptionState state)
+        {
+            byte b;
+            bool re = consumers.TryRemove(state, out b);
+            if (!re)
+            {
+                throw new OrleansException($"{this.GetType().Name} : Remove consumer failed");
```
jason-bragg: same as with removing producers, I can see logging a warning, but throwing seems a bit much.

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
```diff
@@ -0,0 +1,258 @@
...
+            {
+                logger.Error(ErrorCode.Stream_RegisterProducerFailed, $"Failed to register a stream producer.  Stream: {streamId}, Producer: {streamProducer}", exc);
+                throw;
+            }
+            return subscriptions.Consumers.Where(c => !c.IsFaulted).ToSet();
+        }
+
+        public void UnregisterProducer(StreamId streamId, string streamProvider, IStreamProducerExtension streamProducer)
+        {
+            var publisherState = new PubSubPublisherState(streamId, streamProducer);
```
jason-bragg: in general we want to delay allocation until needed.  This could be deferred until the store.TryGetValue check, correct?

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
xiazen: this is copy the logic from GrainBasedPubSub on throwing exception when register/unregister failed. If we don't throw exception, I think it's better to give an return value of false , or something similar to indicate the failure. 

I think this really depend on what orleans application it is written for. If this Orleans application is mainly about streaming, then I think the user will want to get a return value from runtime to the application layer about the failure. If it is for APM, then it is less concerning to indicate success or failure here. 

I think it's the runtime's job to return the result, the method should indicate result on success or failure, more than just logging it. The application layer can always ignore the failure if too annoy.

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
xiazen: Same above

src/Orleans/Streams/PubSub/MemoryPubSubStore.cs
xiazen: Between different Orleans applications, we apply the most common denominator

sergeybykov:
@jason-bragg and I discussed that we might won't to avoid the whole memory pub-sub thing by adding a simple application callback instead.

xiazen:
what kind of application call back? is it an alternative to support the scenario which MemPubsub is trying to support ?

xiazen:
@jason-bragg explained to me the application call back. IMO, it doesn't solve the problem MemPuBSub has. I'm confused on what problem you want to solve by using the application call back. 

To me, it is almost an add on at ProgrammaticSubscribe API. And it is not related to MemPubsub. Maybe we can talk more about it.

xiazen:
close for no use case

PR #2772, opened 2017-02-24
Fix [NonSerialized] on .NET Core

ReubenBond:
Our implementation which made our NonSerializedAttribute check build on .NET Core was not correct.

This fix is quite simple - and I'm silly for not knowing it earlier. Instead of looking for the `[NonSerialized]` attribute on fields, we check for the `FieldAttributes.NotSerialized` flag on the `FieldInfo.Attributes` property.

Our tests don't run on .NET Core, but I tested this method manually using a .NET Core project.

This should fix #2597 

cc @galvesribeiro. You ran into this issue when porting ClientGenerator to .NET Core - we saw some fields in the generated code which should not have been there.

PR #2778, opened 2017-02-24
Simplify instructions for running HelloWorld.NetCore

jdom:
Removed `publish` instructions, since #2772 fixed an issue with serialization that caused the need for that side effect.
Also fixed cascading NuGet.config inconsistencies, to pick up the right nuget packages from the local build.
Created build scripts for PowerShell and Bash to simplify the process even more, for people new to .NET Core.


jdom:
Thanks @galvesribeiro for sending me the Bash script!

PR #2781, opened 2017-02-25
Test project names cleanup

jdom:
Did some renames of some projects, as an attempt to consolidate naming patterns.
Nevertheless, we still need to reach an agreement and consolidate, because we currently have to different naming convention for tests, and both of them are used in more or left half of the tests each.
Current conventions:
 - `Tester` as preffix for tests and `Test` as preffix for grain projects
 - `.Tests` as suffix

The ones in this PR as projects that didn't fit any of those conventions.


ReubenBond:
Any idea what's causing this test failure?
```

**Comment by jdom** [comment]
I've seen that test randomly failing, we need to do something about it

**Comment by jdom** [comment]
@dotnet-bot test this please

**Comment by xiazen** [comment]
@ReubenBond @jdom would it possible that the reason for that test to fail is : whatever previous TestCluster used is not terminated yet, so the new test connected to the old TestCluster, which doesn't have an PubSubStore configured => hence the failure.

**Comment by xiazen** [comment]
Does TestCluster gracefully turn off whenever the test collection finished?

File: src/Orleans.sln [hunk 1]
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26020.0
```
xiazen: yay. thumb up

jdom:
@xiazen yes, that's my assumption, that a previous test somehow leaked a silo so the port is being used, as I have seen that in the past using TestingSiloHost, but it could as well be a parallel test using the same port too. Nevertheless the TestCluster infrastructure uses random ports for the silos, so it would be pretty hard, although not impossible that they collide. But since it is possible, over over the weekend I put out #2779 to minimize that chance. Hopefully we won't see this error again

xiazen:
great! Thanks for the improvement on #2779 
And this LGTM. will merge it.

Issue #2774, opened 2017-02-24
IndexOutOfRangeException in StringEncodedWriteVector.FlipBit


Liversage:
I have tried to upgrade my project from 1.3.1 to 1.4.0. However, I'm currently having problems with `IndexOutOfRangeException` being thrown by [`StringEncodedWriteVector.FlipBit`](https://github.com/dotnet/orleans/blob/70fc5c4029d091060ba548c53bf7427bf0187df6/src/OrleansEventSourcing/Common/StringEncodedWriteVector.cs#L39). This happens because both `writeVector` and `Replica` are empty strings thus `pos = writeVector.IndexOf(Replica)` becomes 0 and then `writeVector[pos - 1]` throws the exception.

Eventually, the exception is caught in [`PrimaryBasedLogViewAdaptor.UpdatePrimary`](https://github.com/dotnet/orleans/blob/70fc5c4029d091060ba548c53bf7427bf0187df6/src/OrleansEventSourcing/Common/PrimaryBasedLogViewAdaptor.cs#L722) where the comment states the following:

> this should never happen - we are supposed to catch and store exceptions 
> in the correct place (LastPrimaryException or notification trackers)

Obviously, there is something wrong in my code. However, it seems that whatever is wrong triggers some unexpected errors in Orleans that are hard to understand.

While I came here to report the unexpected way that `FlipBit` can fail when passed two empty strings any suggestions why this happens in the first place are welcome. I have modified my code based on the updated documentation on event sourced grains and it is not obvious to me what I'm doing wrong. One thing that is a bit special is that this error happens during silo startup where a bootstrap provider calls some event sourced grains.

sergeybykov:
I hope @sebastianburckhardt knows what might be going on here.

sebastianburckhardt:
The `Replica` is never supposed to be null.  It is initialized from the `Silo.ClusterId` which is currently determined as follows:

```csharp
public string ClusterId
{
   get { return GlobalConfig.HasMultiClusterNetwork ? GlobalConfig.ClusterId : GlobalConfig.DeploymentId; } 
}
```
.
So the likely root cause is that `GlobalConfig.DeploymentId` is null for some reason (or  `GlobalConfig.ClusterId ` if you are using the multi-cluster network). Not sure how that can happen... I thought this is being checked somewhere.

Maybe your configuration is not setting the DeploymentId?

(Regardless, we obviously need a better error message). Thanks for looking into this.

Liversage:
[...]

PR #2786, opened 2017-02-27
fix missing check for empty deployment id

sebastianburckhardt:
This is a fix for the bug observed in #2774.

The problem was that cluster-id was picking deployment-id as a default, but deployment-id can be null or empty. This PR adds a check to catch this and use an appropriate default in all cases.

### DIFF
--- a/src/OrleansRuntime/Silo/Silo.cs
+++ b/src/OrleansRuntime/Silo/Silo.cs
@@ -130,8 +130,13 @@ namespace Orleans.Runtime
         /// <summary> Get the id of the cluster this silo is part of. </summary>
         public string ClusterId
         {
-            get { return GlobalConfig.HasMultiClusterNetwork ? GlobalConfig.ClusterId : GlobalConfig.DeploymentId; } 
+            get {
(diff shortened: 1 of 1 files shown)


src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -130,9 +130,14 @@ public bool HasMultiClusterNetwork
...
         public string ClusterId
         {
-            get { return GlobalConfig.HasMultiClusterNetwork ? GlobalConfig.ClusterId : GlobalConfig.DeploymentId; } 
+            get {
+                var configuredId = GlobalConfig.HasMultiClusterNetwork ? GlobalConfig.ClusterId : GlobalConfig.DeploymentId;
+                return string.IsNullOrEmpty(configuredId) ? CLUSTER_ID_DEFAULT : configuredId; 
+            } 
         }
 
+        private const string CLUSTER_ID_DEFAULT = "A"; // if no id is configured, we pick a nonempty default.
```
sebastianburckhardt: fair enough.

PR #2787, opened 2017-02-27
Fix test issue that created client configuration prematurely

jdom:
When using TestClusterOptions it is best to first fully initialize the cluster configuration and then client one, because it will copy stuff from the former on first access.
In this case, the connection string is one of the things it copies, so it is unnecessary. But because of it is was failing to copy the membership config, and the tests were sometimes getting `No gateway available`


jdom:
@dotnet-bot test netfx

jdom:

ReubenBond:
I see, so it would be best to perform that setting after configuring ClusterConfiguration.

LGTM - setting the DataConnectionString there was not needed, it was the `providerProperties` above which mattered.

Issue #2788, opened 2017-02-27
Add a timeout for client to gateway connection


benjaminpetit:
(and maybe for Silo -> Silo connections)

We do not inforce any timeout value for the `Socket.Connect` in https://github.com/dotnet/orleans/blob/master/src/Orleans/Messaging/GatewayConnection.cs#L162

It seems that it can take more than 20 seconds, which is too much: since we try to connect 2 times, it can take more than 40 seconds to declare the GW dead on the client side.

I think we should add a timeout for this operation so the client isn't stuck for too long.

benjaminpetit:
Fixed with #2791

PR #2791, opened 2017-02-27
Add a timeout for socket connection

benjaminpetit:
Fix for #2788 

It seems that the value by default is around 20 seconds.

I added a method in `SocketManager` to be able to specify a timeout from the configuration, and use it for client->gw and silo->silo socket connect.

### DIFF
--- a/src/Orleans/Configuration/MessagingConfiguration.cs
+++ b/src/Orleans/Configuration/MessagingConfiguration.cs
@@ -13,6 +13,10 @@ namespace Orleans.Runtime.Configuration
     public interface IMessagingConfiguration
     {
         /// <summary>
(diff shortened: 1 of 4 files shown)


xiazen:
aww.  'Socket' does not contain a definition for 'BeginConnect' and no extension method 'BeginConnect' accepting a first argument of type 'Socket' could be found (are you missing a using directive or an assembly reference?) [D:\j\workspace\netstandard-w---f16d272e\vNext\src\Orleans\Orleans.csproj]

without much investigation , I think in netstandard, Socket may not have an method of BeginConnect. You need to find an alternative for that

benjaminpetit:
Great... I Have to use the over complicated ConnectAsync for that...

benjaminpetit:
@dotnet-bot Test this please

gabikliot:
LGTM.

Issue #2741, opened 2017-02-18
Programmable stream subscribe API


xiazen:
## Problem Space
Currently in Orleans streaming space, there’s two subscribe models supported. One is ImplicitSubscribe, which define the subscription map for orleans streaming infrastructure in compile time. Another one is ExplicitSubscribe, which rely on the consumer grain to be activated to Subscribe and attach its observer to streaming infrastructure. These two model serves great so far and covers most of the use cases in Orleans streaming. But we recently got a new use case, which is the drive for this new feature in Orleans streaming. 
The new use case is:
1.	I want to dynamically configure consumer’s subscribe/unsubscribe in runtime
2.	I don’t want to use ExplicitSubscribe model, which requires application to activate the consumer grains to build the subscription map. I want to be able to define the subscription map outside grain context(without async grain call), for efficiency concern. For example, when application boostraping. 

**EDIT**: About case **(3.	Since my consumers and producers exists in the same silo, I want they to use local streaming infra, for efficiency concern. )**, it is a new scenario essentially which is not related to the subscribing problems currently in the system. This case is more about performance improvement on streaming, but with certain constraints and complications. We want to tinker more on it. **So remove this case from this issue for now.** 

For this new use case, neither of orleans streaming system’s current subscribe models can support it. This drove us to design this new feature for streaming. As follows it the design for this feature. 
## New Component 
•	`IStreamSubscriptionManager`: the public interface which application use to configure subscription programmablly.  User can access `StreamSubscriptionManager ` through stream provider. It will not support adding or removing implicit subscriptions.  Those subscriptions will be query only.

## New Pattern
1.	Application configure subscription map using `IStreamSubscriptionManager 
`2.	Consumer grain attach its observer using `ResumeAsync` in its grain logic. 

## Common Questions
1 What would a quick remove/re-add pair do to an active grain? Would it just stop receiving events and then start receiving them again?
In the current design, it would simply stop delivering events to the consumer, then resume delivering events.  The effects of this ‘interruption’ would depend on the stream provider implementation.  For SMS, any messages sent between the remove/add, could be lost.  For recoverable stream providers like EventHub, when the re-added subscription is being established the handshake with the grain will inform the streaming system where to start delivering from, and if the data is still in the cache no data loss should occur. 

## API Design

IStreamSubscriptionManager

```
public interface IStreamSubscriptionManager
    {
        Task<StreamSubscription> AddSubscription(IStreamIdentity streamId, GrainReference grainRef);
[...]

PR #2796, opened 2017-02-28
Programmatic streaming subscribe API

xiazen:
for issue #2741 
adding IStreamSubscriptionManager
adding StreamSubscriptionManagerExtensions
adding tests for Programmatic Subscibe

### DIFF
--- a/src/Orleans/Streams/Core/StreamSubscriptionHandle.cs
+++ b/src/Orleans/Streams/Core/StreamSubscriptionHandle.cs
@@ -13,7 +13,7 @@ namespace Orleans.Streams
     {
         public abstract IStreamIdentity StreamIdentity { get; }
 
-        public string ProviderName;
+        public abstract string ProviderName { get; }
         /// <summary>
         /// Unique identifier for this StreamSubscriptionHandle
         /// </summary>
--- a/src/Orleans/Streams/Core/StreamSubscriptionHandle.cs
+++ b/src/Orleans/Streams/Core/StreamSubscriptionHandle.cs
@@ -13,6 +13,7 @@ namespace Orleans.Streams
     {
         public abstract IStreamIdentity StreamIdentity { get; }
 
+        public string ProviderName;
         /// <summary>
         /// Unique identifier for this StreamSubscriptionHandle
         /// </summary>
--- a/src/Orleans/Streams/Core/StreamSubscriptionManager.cs
+++ b/src/Orleans/Streams/Core/StreamSubscriptionManager.cs
@@ -9,33 +9,39 @@ namespace Orleans.Streams.Core
 {
     internal class StreamSubscriptionManager: IStreamSubscriptionManager
     {
+        private readonly string name;
         private readonly IStreamPubSub streamPubSub;
-        private readonly string streamProviderName;
-        public StreamSubscriptionManager(IStreamPubSub pubsub, string streamProviderName)
+        public StreamSubscriptionManager(IStreamPubSub streamPubSub, string name)
         {
-            this.streamPubSub = pubsub;
-            this.streamProviderName = streamProviderName;
+            this.streamPubSub = streamPubSub;
+            this.name = name;
         }
+
         public async Task<StreamSubscription> AddSubscription(IStreamIdentity streamIdentity, GrainReference grainRef)
         {
             var consumer = grainRef.AsReference<IStreamConsumerExtension>();
-            var streamId = StreamId.GetStreamId(streamIdentity.Guid, this.streamProviderName, streamIdentity.Namespace);
+            var streamProviderName = this.name;
+            var streamId = StreamId.GetStreamId(streamIdentity.Guid, streamProviderName, streamIdentity.Namespace);
             var subscriptionId = streamPubSub.CreateSubscriptionId(
                 streamId, consumer);
-            await streamPubSub.RegisterConsumer(subscriptionId, streamId, this.streamProviderName, consumer, null);
-            var newSub = new StreamSubscription(subscriptionId.Guid, this.streamProviderName, streamId, grainRef.GrainId);
+            await streamPubSub.RegisterConsumer(subscriptionId, streamId, streamProviderName, consumer, null);
+            var newSub = new StreamSubscription(subscriptionId.Guid, streamProviderName, streamId, grainRef.GrainId);
             return newSub;
         }
 
         public async Task RemoveSubscription(IStreamIdentity streamId, Guid subscriptionId)
         {
-            await streamPubSub.UnregisterConsumer(GuidId.GetGuidId(subscriptionId), (StreamId)streamId, this.streamProviderName);
+            var streamProviderName = this.name;
+            await streamPubSub.UnregisterConsumer(GuidId.GetGuidId(subscriptionId), (StreamId)streamId, streamProviderName, false);
         }
 
         public Task<IEnumerable<StreamSubscription>> GetSubscriptions(IStreamIdentity streamIdentity)
         {
-            var streamId = StreamId.GetStreamId(streamIdentity.Guid, this.streamProviderName, streamIdentity.Namespace);
+            var streamProviderName = this.name;
+            var streamId = StreamId.GetStreamId(streamIdentity.Guid, streamProviderName, streamIdentity.Namespace);
             return streamPubSub.GetAllSubscriptions(streamId).ContinueWith(subs => subs.Result.AsEnumerable());
         }
     }
+
 }
+
--- a/src/Orleans/Streams/Core/StreamSubscriptionManager.cs
+++ b/src/Orleans/Streams/Core/StreamSubscriptionManager.cs
@@ -32,7 +32,7 @@ namespace Orleans.Streams.Core
         public async Task RemoveSubscription(IStreamIdentity streamId, Guid subscriptionId)
         {
             var streamProviderName = this.name;
-            await streamPubSub.UnregisterConsumer(GuidId.GetGuidId(subscriptionId), (StreamId)streamId, streamProviderName, false);
+            await streamPubSub.UnregisterConsumer(GuidId.GetGuidId(subscriptionId), (StreamId)streamId, streamProviderName);
         }
 
         public Task<IEnumerable<StreamSubscription>> GetSubscriptions(IStreamIdentity streamIdentity)
--- a/src/Orleans/Streams/Core/StreamSubscriptionManager.cs
+++ b/src/Orleans/Streams/Core/StreamSubscriptionManager.cs
@@ -1,0 +1,41 @@
+﻿using Orleans.Runtime;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Orleans.Streams.Core
+{
+    internal class StreamSubscriptionManager: IStreamSubscriptionManager
+    {
+        private readonly IStreamPubSub streamPubSub;
+        private readonly string streamProviderName;
+        public StreamSubscriptionManager(IStreamPubSub pubsub, string streamProviderName)
+        {
+            this.streamPubSub = pubsub;
+            this.streamProviderName = streamProviderName;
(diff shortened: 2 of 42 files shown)


src/Orleans/Streams/Internal/StreamConsumer.cs
```diff
@@ -40,6 +40,27 @@ public StreamConsumer(StreamImpl<T> stream, string streamProviderName, IStreamPr
...
+        public async Task<StreamSubscriptionHandle<T>> AttachObserver(IAsyncObserver<T> observer, GuidId subscriptionId)
+        {
+            if (observer is GrainReference)
+                throw new ArgumentException("On-behalf subscription via grain references is not supported. Only passing of object references is allowed.", "observer");
+
+            if (logger.IsVerbose) logger.Verbose("Attach Observer={0} ", observer);
+            await BindExtensionLazy();
+            var subriptionHandle = myExtension.SetObserver(subscriptionId, stream, observer, null, null);
+            try
+            {
```
xiazen: This should be removed

xiazen: Already registered

test/Tester/StreamingTests/SampleStreamingTests.cs
```diff
@@ -104,6 +106,44 @@ public SampleStreamingTests(string streamProvider, Logger logger, TestCluster cl
...
+            var consumers = new List<IStateless_ConsumerGrain>();
+            var subResultGrain = this.cluster.GrainFactory.GetGrain<Tester.StreamingTests.ISubscribeResultGrain>(
+                Tester.StreamingTests.ProgrammaticSubcribingBoostrapper.SubscribeResultGrainId);
+            var subscriptions = await subResultGrain.GetSubscriptionResult();
+            foreach (var sub in subscriptions)
+            {
+                var consumer = this.cluster.GrainFactory.GetGrain<IStateless_ConsumerGrain>(sub.GrainId.PrimaryKey);
+                consumers.Add(consumer);
+                tasks.Add(consumer.AttachObserver(streamId, StreamNamespace, sub.StreamProviderName, GuidId.GetGuidId(sub.SubscriptionId)));
+            }
```
xiazen: AttachObserver should be in grain's onactivate logic to take advantage of programmatic streaming benefits. I will think of something to revise this test

src/Orleans/Streams/Core/IAsyncObservable.cs
```diff
@@ -25,6 +26,7 @@ public interface IAsyncObservable<T>
         /// </returns>
         Task<StreamSubscriptionHandle<T>> SubscribeAsync(IAsyncObserver<T> observer);
 
+        Task<StreamSubscriptionHandle<T>> AttachObserver(IAsyncObserver<T> observer, GuidId subscriptionId);
```
jason-bragg: The current pattern for attaching observer for existing subscription is to use StreamSubscriptionHandle.Resume.  I don't think we should change that.

src/Orleans/Streams/Core/StreamSubscriptionManager.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Orleans.Streams.Core
+{
+    internal class StreamSubscriptionManager: IStreamSubscriptionManager
+    {
+        private IStreamPubSub streamPubSub;
+        private StreamPubSubType type;
+        private string streamProviderName;
+        //potential memory optimization on this
+        private ConcurrentDictionary<Guid, StreamSubscription> allSubscriptions;
+        public StreamSubscriptionManager(IStreamPubSub pubsub, StreamPubSubType t, string streamProviderName)
```
jason-bragg: I would prefer the StreamSubscriptionManager not have special case logic for each pubsub type.  Let's discuss how to avoid this.

src/Orleans/Streams/Core/StreamSubscriptionManager.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Orleans.Streams.Core
+{
+    internal class StreamSubscriptionManager: IStreamSubscriptionManager
+    {
+        private IStreamPubSub streamPubSub;
+        private StreamPubSubType type;
+        private string streamProviderName;
+        //potential memory optimization on this
+        private ConcurrentDictionary<Guid, StreamSubscription> allSubscriptions;
```
jason-bragg: I don't think we need allSubscriptions.  PubSub system is authoritative on what subscriptions exist, so we shouldn't track this separately.

src/Orleans/Streams/Core/StreamSubscriptionManager.cs
xiazen: why is this? I think we talked about ban AddSub and RemoveSub on ImplicitSubscribe type.

src/Orleans/Streams/Core/StreamSubscriptionManager.cs
xiazen: then we will need to **add** one method to IStreamPubSub interface to ask it to 
Task<List<GuidId>> GetAllSubscriptions(StreamId streamId);

I like this alternative more, you want me to go with this?

xiazen:
Adding new feature, so move it back to work in progress. will remove soon

src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
```diff
@@ -276,14 +277,23 @@ private void RemoveProducer(PubSubPublisherState producer)
...
             if (State.Consumers.Any(c => c.IsFaulted && c.Equals(subscriptionId)))
                 throw new FaultedSubscriptionException(subscriptionId, streamId);
 
             try
             {
+                if (!observerRemoved)
+                {
+                    //remove observer from consumer extension if not removed yet
+                    //this would happen when subscription removed by calling StreamSubscriptionManager
+                    var tasks = State.Consumers.Where(c => c.Equals(subscriptionId))
```
xiazen: I'm thinking about dropping onSubscriptionRemove behavior support. I don't see many users will need it and it will require us to call the consumer grain from here , which may result in an long blocking call and time out UnregisterConsumer call

xiazen: @jason-bragg what do you think? 

And the PR is ready to review now !

jason-bragg: I'll review today.  As for the onSubscriptionRemove, I agree it probably won't get much use, but I do think it's an important signal.  The problem I see with it is that I'm not clear how to wire it up.  I'm ok with leaving it off, but I suggest creating an Issue so we can revisit it.

src/Orleans/Streams/Internal/StreamConsumerExtension.cs
```diff
@@ -63,50 +67,90 @@ internal StreamSubscriptionHandleImpl<T> SetObserver<T>(GuidId subscriptionId, S
...
+        {
+            this.onSubscriptionAddAction = onAdd;
+            this.onSubscriptinRemoveAction = onRemove;
+            return TaskDone.Done;
+        }
+
+        public async Task<bool> RemoveObserver(GuidId subscriptionId, StreamId streamId)
         {
             IStreamSubscriptionHandle ignore;
+            if(this.onSubscriptinRemoveAction != null)
```
jason-bragg: Hmm, I think this will have some unexpected side effects.  Let's leave the remove action out and revisit it in a future pass.

src/Orleans/Streams/Internal/StreamConsumerExtension.cs
```diff
@@ -63,50 +67,90 @@ internal StreamSubscriptionHandleImpl<T> SetObserver<T>(GuidId subscriptionId, S
             }
         }
 
-        internal bool RemoveObserver(GuidId subscriptionId)
+        public Task OnSubscriptionChange<T>(Func<StreamSubscriptionHandle<T>, Task> onAdd, Func<string, IStreamIdentity, Guid, Task> onRemove = null)
+        {
+            this.onSubscriptionAddAction = onAdd;
```
jason-bragg: This is another case where having a typed handle (there have been others) creates problems.  Raising the question, do we want to bit the bullet and change the handle to not be typed.
If we keep the handle typed (probably best for now) I see two problems with this:
1) What happens if someone needs to process streams of different types, they only get one callback?
2) By storing these actions this way, it creates the need to do the reflection magic later, on each call.
I suggest creating an abstraction, something like ISubscriptionChangeHandler, which have onAdd and OnRemove signatures.  Create a generic implementation of this for type specific actions, then add a map of the type to the ISubscriptionChangeHandler for each type specific actions registered.  When we have a subscription change we call the OnAdd or OnRemove on all of the registered handlers.

src/Orleans/Streams/Internal/StreamConsumerExtension.cs
xiazen: I agree

src/Orleans/Streams/Internal/StreamConsumerExtension.cs
xiazen: thanks for the suggestion ! I do like the ISubscriptionChangeHandler for handling the first question.  I will 

not sure how is it going to solve the reflection issue yet. But I assume it is something like what we did for StreamSubscriptionHandler<T> and IStreamSubscriptionHandler. Will try prototype it

test/TestGrains/ProgrammaticSubscribe/Stateless_ConsumerGrain.cs
```diff
@@ -0,0 +1,159 @@
...
+            this.consumerObservers2.Add(observer);
+            this.consumerHandles2.Add(newhandle);
+        }
+    }
+
+    public class Jerk_ConsumerGrain : Grain, IJerk_ConsumerGrain
+    {
+        internal Logger logger;
+        private int onAddCalledCount;
+        private int onAdd2CalledCount;
```
xiazen: this two variable should be removed

xiazen: they are not used

src/Orleans/Streams/Core/StreamSubscriptionHandle.cs
```diff
@@ -13,6 +13,7 @@ public abstract class StreamSubscriptionHandle<T> : IEquatable<StreamSubscriptio
     {
         public abstract IStreamIdentity StreamIdentity { get; }
 
+        public string ProviderName;
         /// <summary>
```
xiazen: @ReubenBond I want to pull the expert here to analyse the situation for me. Adding a string ProviderName to StreamSubscriptionHandle, would this influence backward compatibility of StreamSubscriptionHandle serialization ?

ReubenBond: yes, it will - if you want to be able to deserialize old stream subscription handles correctly, then you need to find some place which is unused in the old version and use that to indicate that this is a newer version.

Abstract classed can never be serialized, so you need to look an implementation. Eg: `isRewindable` is a `bool`, so it will be serialized as a `byte` (`SerializationTokenType.True` or `SerializationTokenType.False`). You can somehow use the unused bits of that byte (probably by changing it to a byte-backed enum). I'm not suggesting it's a good option, but it's an option.

src/Orleans/Streams/Internal/StreamSubscriptionHandleImpl.cs
```diff
@@ -16,11 +16,12 @@ internal class StreamSubscriptionHandleImpl<T> : StreamSubscriptionHandle<T>, IS
         private IAsyncObserver<T> observer;
         [NonSerialized]
         private StreamHandshakeToken expectedToken;
-
+        private string providerName;
```
ReubenBond: This will still not be backwards compatible, since it adds a new field. I would recommend making a get-only property which returns `this.streamImpl.ProviderName`

xiazen:
read for review ~ beep beep

PR #2807, opened 2017-03-03
Prep non-static client

ReubenBond:
These are preparatory commits to clean up a few things before the main non-static client PR. I still need to fix a few things before that PR is ready

The commits ought to be unrelated to each other, so it might be easier to review commit-by-commit.


jdom:
still reviewing, sent some minor comments as I'm preemted, but i'll get back to it later

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
```diff
@@ -10,61 +10,26 @@ namespace OrleansTelemetryConsumers.Counters
     /// <summary>
     /// Telemetry consumer that writes metrics to predefined performance counters.
     /// </summary>
-    public class OrleansPerfCounterTelemetryConsumer : IMetricTelemetryConsumer
+    internal class OrleansPerfCounterTelemetryConsumer : IMetricTelemetryConsumer
```
jdom: Why was this made internal? wouldn't this mean that users cannot use programmatic config to add this telemetry consumer?

ReubenBond: Since it needs SiloAssemblyLoader in the ctor, but that's internal. I'm not certain what the best solution is here. Need to work something out...

src/Orleans/GrainDirectory/GlobalSingleInstanceRegistration.cs
```diff
@@ -11,27 +11,9 @@ namespace Orleans.GrainDirectory
     [Serializable]
     internal class GlobalSingleInstanceRegistration : MultiClusterRegistrationStrategy
     {
-        private static GlobalSingleInstanceRegistration singleton;
+        private static readonly string[] EmptyList = new string[0];
```
ReubenBond: I chose an empty array because the consumer (or at least some) call `.ToList()` on it, but I think it doesn't really make a difference, so I'll use Enumerable.Empty

ReubenBond:
The failing message test was because I had directly exposed headersLength and bodyLength in the `out` params and then used the same variable for them in the test (`dummy`), so the method was overriding itself. I undid that change so the `out` params cannot affect the behavior of the method. Test passes.

ReubenBond:
@jdom I implemented the workaround we discussed for the perf counter telemetry consumer

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
```diff
@@ -12,23 +12,22 @@ namespace OrleansTelemetryConsumers.Counters
     /// </summary>
     internal class OrleansPerfCounterTelemetryConsumer : IMetricTelemetryConsumer
```
jdom: this should now be public again

ReubenBond: ah thanks

PR #2809, opened 2017-03-03
fixed ServiceFabricBase

jmblakl:
The value of ServiceFabricBase defined in ErrorCodes.cs for the OrleansServiceFabricUtils is different and collides with the values defined in Orleans Logging ErrorCodes.cs


ReubenBond:
Thanks :)

PR #2810, opened 2017-03-03
Add missing imports to make solution build in VS2017 too.

attilah:
Add missing Microsoft.Common.props import to projects, this fixes the error of building the solution from VS2017.


xiazen:
LGTM. just the thing I needed. Let me pull down your branch and test it before I merge...

PR #2816, opened 2017-03-06
Added section into Advanced-Concepts\Serialization.html regarding StyleCop

chris-eaton:
Added a help section detailing how to tell StyleCop to ignore Orleans codegen files


chris-eaton:
Updated the PR to include MD changes

jdom:
Hi @chris-eaton, 1.4.0 no longer includes the orleans.codegen.cs file in the project (in fact, you must delete it if you still have it). Does StyleCop try to analyze generated code that is just in the intermediate folder when building? Or is this a side effect of upgrading to 1.4.0 but not deleting that orleans.codegen.cs file?

chris-eaton:
@jdom Hey.  Yeah this code prevents StyleCop from Analysing the generated code thats put into the intermediate folder whilst building.  Unfortunately StyleCop seems to analyse it even though its not part of the solution and generated on build!

Just to be clear, the orleans.codegen.cs has been deleted and the warnings specifically reference a new grains.codegen.cs thats in Debug\Obj [that may not quite be right as im working off memory here, but its definitely in a debug folder]

jdom:
Looks like we might be able to rename the file so that stylecop excludes it. Something like `assemblyname.orleans.generated.cs` will do the trick. Then no workaround is needed by the end user. I believe that's a more appropriate fix

ReubenBond:
At some point in time, StyleCop (any maybe other tools) would ignore files with an `/// <auto-generated />` comment near the top. Maybe we need to add that

jdom:
Yes, I was thinking of adding both: the filename change and the comment

jdom:
I did the filename change in https://github.com/dotnet/orleans/pull/2872 . We can tackle adding the comment also in a future iteration, but I believe this PR should be superseded by it, correct? Can we close it?

PR #2817, opened 2017-03-06
Update Application-Bootstrap-within-a-Silo.md

akosradler:
pointing out the issue when a blocking call is made inside Init


sergeybykov:
Thank you, @akosradler!

PR #2819, opened 2017-03-06
Allow null MultiClusterRegistrationStrategy in GrainTypeData

ReubenBond:
Some usages of `GrainTypeData` are from outside of the silo, where a sensible `MultiClusterRegistrationStrategy` is not available.

For example, it's used in CounterControl.


jdom:

Issue #1780, opened 2016-05-23
Initialize GrainClient and Silo in the same process.


lyjdamzwf:
Hi guys:
For some reasons, I need to get a Grain out of a Grain context but in the Silo process.
So I try to initialize a GrainClient, but it raised a exception below:

> 127.0.0.1:12111]        OutsideRuntimeClient.ConstructorReset(): client Id *cli/
> 74c55794
> Initialization failed. Calling SetUnobservedExceptionHandler the second time.

I tracked the code it caused of the Silo has invoked the SetUnobservedExceptionHandler so the initialization of GrainClient failed.

How can I fix the problem?

Plus, my requirement is when the process receives the shutdown signal, the silo process should notify all users (Grains) offline which are active, but I couldn't get a Grain/StreamProvider in the shutdown callback, so are there any other ways to solve the problem?

sergeybykov:
`GrainClient` has to be in a separate app domain or process from a silo. Currently, they cannot coexist.

>Plus, my requirement is when the process receives the shutdown signal, the silo process should notify all users (Grains) offline which are active, but I couldn't get a Grain/StreamProvider in the shutdown callback, so are there any other ways to solve the problem?

Your process shutdown callback needs simply needs to call `Silo.Shutdown()`. This will trigger a graceful shutdown of the silo, as part of which each activated grain will get deactivated, and its `OnDeactivateAsync()` method will get called.

Do you need something different?

lyjdamzwf:
@sergeybykov Actually, I want a proprocessing before Silo.Shutdown, like notify some messages to  All Grains (simply say is broadcasting in a normal method), and this broadcasting logic codes can't be put in the Grain.OnDeactivateAsync(), since the Silo is shuttting down, some other grains may be already OnDeactivateAsync so that I couldn't use it.

That's all, have I explained clearly? :)

sergeybykov:
@lyjdamzwf Sorry about missing your comment. This isn't supported today. There is no way to send messages from the same app domain with a silo. As a workaround, you could initialize a client in a separate app domain, and signal it to send messages upon a shutdown.

shlomiw:
I also have a scenario when I want to subscribe to events (Orleans Streams) in a common code for both Orleans Clients and Orleans Silos.

i.e. - I have a common Cache manager (not Grain), which is used in all my processes, both clients and silos. I'd like it to be able to subscribe to 'clear' event via Orleans Stream. 
[...]

PR #2822, opened 2017-03-07
Non-static GrainClient

ReubenBond:
This PR implements the majority of the remaining items in #467. With this PR merged, an end-user can
1. Call into multiple separate clusters from a client
2. Call into multiple other clusters from within one cluster

The core abstraction which this introduces is `IClusterClient`, which implements `IGrainFactory`. It is implemented by the public class `ClusterClient`. `ClusterClient` has static `Create` methods which match the existing `GrainClient.Initialize` methods with the major difference that they create and do not start the client. A client is started via a call to `Start`, which is asynchronous.

This does not implement support for automatically binding `GrainReference`s to their originating cluster, but there is an `IGrainFactory.BindGrainReference(ref)` method for that purpose - so users aren't completely stuck. The plan is to implement those kinds of features later on, probably leveraging the `clusterId` concept from GeoClusters.

Calling from one grain into multiple other clusters:
```C#
public class ConnectorGrain : Grain, IConnectorGrain
{
    private readonly IClusterClient clientOne = ClusterClient.Create(
        new ClientConfiguration
        {
            Gateways = new List<IPEndPoint>(new[] {new IPEndPoint(IPAddress.Loopback, 11112)}),
            PreferedGatewayIndex = 0
        });

    private readonly IClusterClient clientTwo = ClusterClient.Create(
        new ClientConfiguration
        {
            Gateways = new List<IPEndPoint>(new[] {new IPEndPoint(IPAddress.Loopback, 22222)}),
            PreferedGatewayIndex = 0
        });

    public override async Task OnActivateAsync()
    {
        await base.OnActivateAsync();
        await this.clientOne.Start();
        await this.clientTwo.Start();
    }

    public async Task<string> Three()
    {
        var one = this.clientOne.GetGrain<IOneGrain>(Guid.NewGuid());
        var two = this.clientTwo.GetGrain<ITwoGrain>(Guid.NewGuid());
        return await one.One() + " + " + await two.Two();
    }
}
```
Similarly, for standalone clients:
```C#
var clientOne = ClusterClient.Create(CreateConfig(11112));
var clientTwo = ClusterClient.Create(CreateConfig(22222));
var clientThree = ClusterClient.Create(CreateConfig(33332));

await Task.WhenAll(
    clientOne.Start(),
    clientTwo.Start(),
    clientThree.Start());

var one = clientOne.GetGrain<IOneGrain>(Guid.NewGuid());
var two = clientTwo.GetGrain<ITwoGrain>(Guid.NewGuid());
var three = clientThree.GetGrain<IConnectorGrain>(Guid.NewGuid());

while (!Console.KeyAvailable)
{
    Console.WriteLine($"{one}: {await one.One()}");
    Console.WriteLine($"{two}: {await two.Two()}");
    Console.WriteLine($"{two}: {await three.Three()}");

    await Task.Delay(TimeSpan.FromSeconds(1));
}
```


ReubenBond:
I've tagged this as a work-in-progress for now. The biggest change intended for the final version is to make the client startup asynchronous

ReubenBond:
Alright, the `IClusterClient.Start()` method in the latest revision is async :)

jdom:
What is the reasoning for having to call Start? Should we separate the creation of the ClusterClient into something that returns an already initialized client? Close(Async) still makes sense, kind of like an async Dispose method, but it sounds like initializing shouldn't be done on the ClusterClient. What do you think?
On a separate note, we were talking with Jason that we could avoid having a static Create method on it, and separate it into ClusterClientBuilder or something of the sort, with a configuration a little bit closer to asp.net (or EF, since it's a client library). Of course this can be after the PR is merged.

ReubenBond:
For example, consider this flow:
* Create client
* Wire up event handlers (cluster disconnect, client request, maybe some custom ones from the service provider)
* Start the client

That flow has no chance of race conditions, whereas the following does
* Create & start the client
* Wire up handlers

Additionally, I believe creation should be synchronous, whereas connecting to a cluster is async.

I'd be happy if we had a ClientBuilder, but I figured that change should coincide with a similar change on the silo side and at the same time, we should allow the user to provide their own DI container (and inject their own services).

src/Orleans/Core/ClusterClient.cs
```diff
@@ -0,0 +1,324 @@
...
+
+            set { this.runtimeClient.SetResponseTimeout(value); }
+        }
+
+        /// <inheritdoc />
+        public Action<InvokeMethodRequest, IGrain> ClientInvokeCallback
+        {
+            get { return this.runtimeClient.ClientInvokeCallback; }
+            set { this.runtimeClient.ClientInvokeCallback = value; }
+        }
```
jdom: These seem like initialization concerns, so if moving to a separate builder, they should go there

src/Orleans/Core/ClusterClient.cs
```diff
@@ -0,0 +1,324 @@
...
+        /// </summary>
+        /// <param name="configuration">The configuration.</param>
+        /// <returns>A newly instantiated <see cref="ClusterClient"/>.</returns>
+        public static ClusterClient Create(ClientConfiguration configuration) => new ClusterClient(configuration);
+
+        /// <inheritdoc />
+        public async Task Start()
+        {
+            this.ThrowIfDisposed();
+            using (await this.initLock.LockAsync())
```
jdom: all awaits in this class should use `.ConfigureAwait(false)`

src/Orleans/Core/ClusterClient.cs
```diff
@@ -0,0 +1,324 @@
...
+            {
+                await this.runtimeClient.Start();
+                this.state = LifecycleState.Started;
+            }
+        }
+
+        /// <inheritdoc />
+        public void Stop()
+        {
+            this.Stop(gracefully: true).Wait();
```
jdom: This one for example would deadlock in some environments if .ConfigureAwait(false) is not used.

src/OrleansPSUtils/StartGrainClient.cs
```diff
@@ -42,51 +44,78 @@ protected override void ProcessRecord()
...
+            return config;
+        }
+
         protected override void StopProcessing()
         {
-            GrainClient.Uninitialize();
+            var client = this.GetClient();
+            client?.Stop();
+            client?.Dispose();
+            this.SetClient(null);
```
jdom: should SetClient always be called even with exceptions? (ie: in a finally block, or moving it before stopping?)

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -153,7 +153,7 @@ public OutsideRuntimeClient(ClientConfiguration cfg, bool secondary = false)
 
                 if (!secondary)
                 {
-                    UnobservedExceptionsHandlerClass.SetUnobservedExceptionHandler(UnhandledException);
+                    UnobservedExceptionsHandlerClass.TrySetUnobservedExceptionHandler(UnhandledException);
```
jdom: Unfortunately that we still do this, especially on the client, where Orleans is not the runtime, just a small client library. We shouldn't be the ones hooking up logging for any unobserved exception that could occur in the frontend.
Nevertheless, is trying to set it and skip otherwise the right choice? Could there be side effects if I start clientA, then clientB, and then dispose clientA?

src/OrleansManager/Program.cs
```diff
@@ -12,6 +12,8 @@ class Program
     {
         private static IManagementGrain systemManagement;
 
+        private static IInternalClusterClient client;
```
jdom: Maybe stop and dispose before exiting the process?

ReubenBond: How's this?
```C#
using (client = (IInternalClusterClient) new ClientBuilder().LoadConfiguration().Build())
{
    client.Start().Wait();

    // ...

    switch (command)
    {
        case "grainstats":
            PrintSimpleGrainStatistics(restWithoutOptions);
            break;
            // ...
    }
                
    client.Stop();
}
```

src/OrleansPSUtils/StartGrainClient.cs
ReubenBond: Fixed, thanks.

src/Orleans/Runtime/OutsideRuntimeClient.cs
ReubenBond: I think it's benign, and I'd happily remove it. Here's the implementation of that handler:
```C#
private void UnhandledException(ISchedulingContext context, Exception exception)
{
    logger.Error(ErrorCode.Runtime_Error_100007, String.Format("OutsideRuntimeClient caught an UnobservedException."), exception);
    logger.Assert(ErrorCode.Runtime_Error_100008, context == null, "context should be not null only inside OrleansRuntime and not on the client.");
}
```
Should I remove it?

src/Orleans/Core/ClusterClient.cs
ReubenBond: You're right, thanks.

src/Orleans/Core/ClusterClient.cs
ReubenBond: fixed

src/Orleans/Core/ClusterClient.cs
```diff
@@ -0,0 +1,324 @@
...
+            add
+            {
+                this.ThrowIfDisposed();
+                this.runtimeClient.ClusterConnectionLost += value;
+            }
+
+            remove
+            {
+                this.ThrowIfDisposed();
+                this.runtimeClient.ClusterConnectionLost -= value;
```
ReubenBond: I'm not sure what the norm is, but as of the latest commit, this event is no longer public (I kept it on `IInternalClusterClient`). I've removed the throw.

src/Orleans/Core/ClusterClient.cs
ReubenBond: Done, kept as internal for easier testing.

src/Orleans/Core/ClusterClient.cs
```diff
@@ -0,0 +1,324 @@
...
+            Created,
+            Started,
+            Disposed
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ClusterClient"/> class.
+        /// </summary>
+        /// <param name="configuration">The client configuration.</param>
+        public ClusterClient(ClientConfiguration configuration)
```
ReubenBond: done

test/TesterInternal/ActivationsLifeCycleTests/DeactivateOnIdleTests.cs
```diff
@@ -353,23 +361,35 @@ public void Dispose()
...
+                    testCluster.Primary.GetTestHook(this.testCluster.InternalGrainFactory)
+                        .UnregisterGrainForTesting(grainReference);
             int secondaryActivation = 0;
 
             if (isMultipleSilosPresent)
             {
-                secondaryActivation = await testCluster.SecondarySilos[0].TestHook.UnregisterGrainForTesting(grainReference);
+                secondaryActivation =
+                    await
+                        testCluster.SecondarySilos[0].GetTestHook(this.testCluster.InternalGrainFactory)
```
jdom: API-wise, it could be nicer if this was testCluster.GetTestHook(SecondarySilos[0]).XXX, since the internal grain factory is an implementation detail, and also there are some test hooks that can be useful for end users who do not have InternalsVisibleTo in order to get ahold of this internal class.

jdom: Hmm, GetTestHook is internal for now anyway, so the 2nd part of my comment doesn't apply (yet). Still, it would be better to avoid this implementation detail

galvesribeiro: I thought we were moving away from hooks...

test/TesterInternal/GeoClusterTests/TestingClusterHost.cs
```diff
@@ -300,26 +300,26 @@ public void StopAllClusters()
...
 
-        private readonly List<AppDomain> activeClients = new List<AppDomain>();
+        private readonly List<ClientWrapperBase> activeClients = new List<ClientWrapperBase>();
 
 
         // The following is a base class to use for creating client wrappers.
-        // We use ClientWrappers to load an Orleans client in its own app domain. 
         // This allows us to create multiple clients that are connected to different silos.
-        public class ClientWrapperBase : MarshalByRefObject {
+        public class ClientWrapperBase {
```
jdom: Should this be IDisposable?

src/Orleans/Runtime/AsynchAgent.cs
```diff
@@ -250,9 +250,11 @@ public override string ToString()
             return Name;
         }
 
+        internal static SystemStatus CurrentSystemStatus { get; set; } = SystemStatus.Unknown;
+
         private static void LogStatus(Logger log, string msg, params object[] args)
         {
-            if (SystemStatus.Current.Equals(SystemStatus.Creating))
+            if (CurrentSystemStatus.Equals(SystemStatus.Creating))
```
jdom: This will not behave the same way as before, since you only set it to Creating once and then it stays like that after Creating actually finished. I'm not suggesting to keep them exactly in sync, but if you aren't doing that, you could at least set this when it transitions to Creating and transition off Creating. Also, if you are not keeping them up to date after that, I would just have a boolean flag, like IsStarting

src/Orleans/Core/ClientBuilder.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.CodeGeneration;
+using Orleans.Messaging;
+using Orleans.Providers;
+using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using Orleans.Serialization;
+using Orleans.Streams;
+
+namespace Orleans
```
jdom: Should these builder classes be moved to Orleans.Configuration?

test/DefaultCluster.Tests/BasicActivationTests.cs
```diff
@@ -186,12 +186,13 @@ public void BasicActivation_MultipleGrainInterfaces()
...
             List<Task> promises = new List<Task>();
-            TimeSpan prevTimeout = this.Client.ResponseTimeout;
+            var client = (IInternalClusterClient) this.Client;
+            TimeSpan prevTimeout = client.ResponseTimeout;
             try
             {
                 // set short response time and ask to do long operation, to trigger expired msgs in the silo queues.
                 TimeSpan shortTimeout = TimeSpan.FromMilliseconds(1000);
-                this.Client.ResponseTimeout = shortTimeout;
+                client.ResponseTimeout = shortTimeout;
```
jdom: if this is the only reason why ResponseTimeout was kept in the interface, I suggest you just use the service provider from within this tests to get to the relevant component and tweak it there, since it's not something we want to encourage (even if these tests do a little bit of white-box testing)

src/Orleans/Runtime/AsynchAgent.cs
ReubenBond: I'll change it to `IsStarting`. Longer-term, we should probably remove this.

src/Orleans/Core/ClientBuilder.cs
ReubenBond: Maybe there's a better namespace than the root, but I configuration seems a bit wrong.
ASP.NET's `WebHostBuilder` is in `Microsoft.AspNetCore.Hosting`, so I could put it in a similar namespace.

test/TesterInternal/GeoClusterTests/TestingClusterHost.cs
ReubenBond: I'll add it. It's being disposed either way

test/NonSilo.Tests/ClientBuilderTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <summary>
+        /// Tests that a builder can not be used to build more than one client.
+        /// </summary>
+        [Fact]
+        public void ClientBuilder_DoubleBuildTest()
+        {
+            var builder = new ClientBuilder();
+            using (builder.Build())
+            {
+                Assert.Throws<InvalidOperationException>(() => builder.Build());
```
jdom: is there a technical reason not to allow this or just philosophical? Because if the latter, maybe it is fine to re-use the same builder multiple times. If there is a technical why this doesn't work, then I don't mind for now

test/TesterInternal/ActivationsLifeCycleTests/DeactivateOnIdleTests.cs
jdom: nope, we got rid of AppDomain-specific test hooks, but not test hooks in general

src/Orleans/Core/ClusterClient.cs
```diff
@@ -143,31 +164,36 @@ public static ClusterClient Create(FileInfo configFile)
         public static ClusterClient Create(ClientConfiguration configuration) => new ClusterClient(configuration);
 
         /// <inheritdoc />
-        public void Start()
+        public async Task Start()
         {
-            lock (this.initLock)
+            this.ThrowIfDisposed();
+            using (await this.initLock.LockAsync())
```
jdom: probably everything related to start and stop in this class also requires .ConfigureAwait(false), since the most likely scenario is that it will still be using the parent context

test/TesterInternal/ActivationsLifeCycleTests/DeactivateOnIdleTests.cs
ReubenBond: I'll add `GetTestHooks(SiloHandle)`  as an extension method in `OrleansTestingHost` for `IClusterClient`, so it will look like `client.GetTestHooks(silo)`

src/Orleans/Core/IInternalClusterClient.cs
```diff
@@ -11,5 +13,26 @@ internal interface IInternalClusterClient : IClusterClient, IInternalGrainFactor
...
+        /// The action receives an <see cref="InvokeMethodRequest"/> with details of the method to be invoked, including InterfaceId and MethodId,
+        /// and a <see cref="IGrain"/> which is the GrainReference this request is being sent through
+        /// </summary>
+        /// <remarks>This callback method should return promptly and do a minimum of work, to avoid blocking calling thread or impacting throughput.</remarks>
+        ClientInvokeCallback ClientInvokeCallback { get; set; }
+
+        /// <summary>
+        /// Event fired when connection to the cluster is lost.
+        /// </summary>
+        event ConnectionToClusterLostHandler ClusterConnectionLost;
```
ReubenBond: They're only there for tests and GrainClient compat. I can remove `ResponseTimeout` outright fairly easily using DI to access the `OutsideRuntimeClient` in the relevant test and in `GrainClient`.

I could use the same trick for the other two. Do you feel that's best?

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -90,47 +90,37 @@ public IStreamProviderRuntime CurrentStreamProviderRuntime
...
         {
-            if (cfg == null)
+            this.handshakeClientId = GrainId.NewClientId();
+            tryResendMessage = TryResendMessage;
+            unregisterCallback = msg => UnRegisterCallback(msg.Id);
+            callbacks = new ConcurrentDictionary<CorrelationId, CallbackData>();
+            localObjects = new ConcurrentDictionary<GuidId, LocalObjectData>();
+        }
+
+        internal void ConsumeServices(IServiceProvider services)
```
ReubenBond: Maybe, but there are a lot of circular dependencies back to IRuntimeClient, so this was the easiest approach. I'll see what it takes to sever those dependencies.

src/Orleans/Core/GrainClient.cs
```diff
@@ -21,6 +21,13 @@ namespace Orleans
     public delegate void ConnectionToClusterLostHandler(object sender, EventArgs e);
 
     /// <summary>
+    /// The delegate called before every request to a grain.
+    /// </summary>
+    /// <param name="request">The request.</param>
+    /// <param name="grain">The grain.</param>
+    public delegate void ClientInvokeCallback(InvokeMethodRequest request, IGrain grain);
```
ReubenBond: I moved the declaration from ClusterClient.cs to here, where it can live alongside the event handler definition, but I could also move it to its own file.

test/NonSilo.Tests/ClientBuilderTests.cs
ReubenBond: No reason other than this is how ASP.NET does it - so I copied their behavior.

src/Orleans/Core/ClusterClient.cs
ReubenBond: That's an out-dated commit, sorry - I changed this to use ConfigureAwait(false) after you pointed it out yesterday.

src/Orleans/Core/ClientBuilder.cs
jdom: Sure, maybe some other, although Hosting doesn't make much sense for the client library either.
Don't necessarily feel strong about Configuration, but what's the reason you don't like it? Effectively it's how you configure the client, right? (and eventually we'll fully configure it using the builder, other than the current `ClientConfiguration` object

src/Orleans/Core/GrainClient.cs
jdom: Yeah, I was looking at a commit where this was defined twice simultaneously, not just a move. But a later commit removed one of them, so it's OK

src/Orleans/Core/ClientBuilder.cs
ReubenBond: To me, Configuration is for specifying configuration rather than consuming it, that's all. I don't feel strongly about it. I felt that putting it in the core namespace made it more obvious as a starting point for users.

src/Orleans/Core/GrainClient.cs
ReubenBond: oh oops!

src/Orleans/Core/ClusterClient.cs
```diff
@@ -0,0 +1,324 @@
...
+        /// <returns>A newly instantiated <see cref="ClusterClient"/>.</returns>
+        public static ClusterClient Create(ClientConfiguration configuration) => new ClusterClient(configuration);
+
+        /// <inheritdoc />
+        public async Task Start()
+        {
+            this.ThrowIfDisposed();
+            using (await this.initLock.LockAsync())
+            {
+                await this.runtimeClient.Start();
```
ReubenBond: I added the disposal check after the lock (thanks). I'll add a double-start check.

src/Orleans/Core/ClusterClient.cs
jdom: yup, this one is indeed outdated. Nevertheless I see several in OusideRuntimeClient that should be .ConfigureAwait(false) also. If you look at [GrainClient.InternalInitialize](https://github.com/dotnet/orleans/pull/2822/files#diff-9f940665838d22872b8984d950096293L194) it would enqueue the initialization work in a background thread. We are no longer doing that, so we either do it too, or be more careful to not propagate the initialization context, and not do any blocking calls in that thread (but I suppose you already did that).

ReubenBond: I think the awaits in OutsideRuntimeClient are fine, since the public Start method does have ConfigureAwait(false) - what do you think?

test/DefaultCluster.Tests/BasicActivationTests.cs
ReubenBond: Fixed

src/Orleans/Core/ClusterClient.cs
jdom: Not really, having it in the parent means it will not capture it at that point to continue in the context once the nested continuations finish.
But even without nesting, and just having a flat execution path, adding .ConfigureAwait(false) on the first call is not good enough, because it might continue synchronously (inline), so it didn't capture the context *explicitly*, but it is *still* running in that context.

ReubenBond: Fixed

src/Orleans/Core/GrainClient.cs
```diff
@@ -229,7 +229,7 @@ private static void DoInternalInitialize(IInternalClusterClient clusterClient)
                         // before we make a call that makes RuntimeClient.Current not null
                         isFullyInitialized = false;
                         client = clusterClient;  // Keep reference, to avoid GC problems
-                        client.Start().Wait();
+                        client.Connect().Wait();
```
jdom: sorry, for not bringing this up earlier, but I believe a good practice when we must do a blocking (either .Wait() or .Result) is to use `.GetAwaiter().GetResult()`, so that at least the exception is no longer an AggregateException, which would only be useful if you had more than 1 exception (for example if you call Task.WhenAll).
That's what I remember at least, and doing a quick search I get stuff like this: https://github.com/aspnet/Security/issues/59

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -237,7 +237,7 @@ private void UnhandledException(ISchedulingContext context, Exception exception)
 
         public async Task Start()
         {
-            await StartInternal().ConfigureAwait(false);
+            await Task.Run(this.StartInternal).ConfigureAwait(false);
```
jdom: Want to move the original comment from GrainClient.InternalInitialize?
// We deliberately want to run this initialization code on .NET thread pool thread to escape any
// TPL execution environment and avoid any conflicts with client's synchronization context

src/Orleans/Core/GrainClient.cs
ReubenBond: Good point, will fix

src/Orleans/Runtime/OutsideRuntimeClient.cs
ReubenBond: Will do

test/NonSilo.Tests/ClientBuilderTests.cs
galvesribeiro: I asked myself the same thing when I was reading that line... I guess the only usage to have it support multiple `Build()` calls, is if behave like a stateless factory... Idk, for me its ok to leave the exception as is...

galvesribeiro:
Great work @ReubenBond! Looking forward to have my tests back on .Net Core :P

ReubenBond:
I'll run functionals again before merge

ReubenBond:
Ok, functionals look good

galvesribeiro:
:shipit:

jdom:
Thanks! Awesome! Reuben, can you merge to vso/master?

ReubenBond:
@jdom huge thanks for reviewing this! I'll do the merge (there's a commit needed to avoid breakages)

lmagyar:
What is the best practice to use these `myClientBuilder.Build();` and `await myClusterClient.Connect();` methods?

As I see, these are not "cheap" calls, is the suggested `OnActivateAsync()` is a good place? Or should we use some global/singleton instance and inject it in the grain ctor???

Should we handle somewhere the `ClusterConnectionLost` events, or is this relevant only for the explicit stream subscriptions? If we should handle these events, as I see we have to re-Build and re-Connect everything again, is it correct?

sergeybykov:
>As I see, these are not "cheap" calls, is the suggested OnActivateAsync() is a good place? Or should we use some global/singleton instance and inject it in the grain ctor???

Are you considering this for the case of grains talking to grains in another cluster?

The primary use case here is clients (frontends) connecting to multiple clusters. There the cost is ~the same as it was with GrainClient.Initialize().

lmagyar:
Yes, I figured out the client story and would like to clarify the grain-to-grain version, maybe the inter-cluster streaming if possible. As I understand this would be useful for connecting Orleans uServices directly. I'm just preparing for a presentation about Orleans's features (covering even the planned features), that's why I'm asking.

ReubenBond:
@lmagyar this is not a tested/supported scenario. The `ClientBuilder` is designed for external use (i.e, outside of the context of a grain). Anecdotally, I found that it works for me when I only use the `IClusterClient` to make grain calls, but I am not sure that it will work correctly when streams or observers are involved. For example, the `IClusterClient` does not have access to the grain's scheduler to schedule those out-of-band calls on. This is possible to implement, but it's not implemented today, and certainly it is not a tested or supported scenario.

shlomiw:
@ReubenBond - I'm on 1.5.0-rc. is it possible to create IClusterClient inside silo and use it to call the same silo grains and/or subscribe streams? I was hoping to use it in my shared common libs for both the client and the silo (i.e. listening to general events in the system).

I've tried but without success.

galvesribeiro:
@shlomiw yes, it is possible. You can have multiple `IClusterClient` instances inside and outside a Silo.

shlomiw:
@galvesribeiro - the `IClusterClient` instantiation inside a silo does work, but I have now a race condition.
As I've seen, you can Connect the client to the silo only when the silo is ready to get requests.
That makes sense. But my issue that I need this Client for other contexts, i.e. a grain which uses my library which in turn uses the Client. But a call to this grain can precede the Client initialization.
I also use this shared IClusterClient for Streams subscriptions and events publishing.

To sum-up an example:
- I have a common library of Cache which is shared among both Clients and Silos.
- To initialize one of the Cache items - it needs to call a specific grain.
- My common library uses IClusterClient to invoke grains, this Client must be initialized when the app starts.
- The Silo needs to initialize this library, and therefore creates a IClusterClient which connect to the silo itself. This can be done, as I seen, only after the silo is up and ready to handle requests. I couldn't initialize the Client inside a Bootstrap Provider.
- I get a race condition of a Grain handling a request and then trying to fetch the mentioned Cache item, but it fails since the IClusterClient wasn't initialized yet.

I can try to solve it with coordinations mechanisms, but I was wondering if there's a better approach. And it would be best to use the shared IClusterClient.

BTW - my previous solution, before 1.5.0, was having different DI to the shared library exposing IGrainFactory, Streams subscriptions etc. For the IGrainFactory inside a silo - I just captured a context of one of the grains. But it is very hacky.

What do you think? thanks in advance!

galvesribeiro:
TL;DR; If you are inside the target silo, there is no point on use `IClusterClient` to connect to itself... You may have an abstraction on your common library (i.e. an interface) and then initialize it with the proper implementation (inside a Silo use `GrainClient` and on a client `IClusterClient`). 

When the silo starts up, the `GrainClient` is an internal instance of [`ISiloRuntimeClient`](https://github.com/dotnet/orleans/blob/b9bad19cbf58c0570f590cc2eaceace366183325/src/OrleansRuntime/Core/ISiloRuntimeClient.cs) which is already initialized and available to grain code. 

In other hand, your client (assuming you are on 1.5.x and using the new non-static client from this PR) will use `IClusterClient` implementation which is available for external clients (it also works if you are inside a Silo but connecting to an external cluster). 

There is no DI today (yet) on our client side unfortunately...

shlomiw:
@galvesribeiro  - I can't figure out how to use `GrainClient` inside a silo. See also #1780. That why I thought `IClusterClient` to the rescue.
What am I missing?

galvesribeiro:
Sorry @shlomiw. Please look at my updated reply. It is `GrainFactory` the property in the grain which is the actual client.

shlomiw:
Ok, I understand. I hoped I could use IClusterClient to share the code also in the silo.
I'm going to revert it back to what it was:
Via bootstrap provider call a grain, which in turns inject the IGrainFactory and also IStreamProvider to my shared DI.
Note that this grain must be kept alive, otherwise it'll loose the stream subscriptions when it is being deactivated.

galvesribeiro:
> Note that this grain must be kept alive, otherwise it'll loose the stream subscriptions when it is being deactivated

Why? The subscriptions are persistent as far as I know. If a grain is deactivated and have an active subscription, it just need to call `Resume()` inside `OnActivateAsync()` in order to receive the message. Why you think it is not persistent?

I would have a generic interface in your case. Something like this:

```
public interface IOrleansClient
{
      IGrainFactory GetFactory();
}
```

Both `IClusterClient` and `GrainFactory` implement `IGrainFactory`. Your shared library should return the `GrainFactory` instance inside a silo or `IClusterClient` in case it is accessing an external Cluster.

shlomiw:
@galvesribeiro - first of all - thanks a lot for your quick replies! it's not trivial :)

1. Keeping the grain alive - I use this Grain for general streaming purposes, I need this stream all the time. The grain is only a way to access the stream inside the silo. Nobody calls it, only the bootstrap provider. So after a while it'll be deactivated and the stream is lost (until re-activation of the grain), I never said it's not persistent :) so I just keep it alive with timer that calls itself (using `AsReference().KeepAlive()`, and and inside `KeepAlive()` calling `DelayDeactivation()`, I think it's robust enough..).

2. Yep - I did something similar :) and put inside the interface `IGrainFactory` and `IStreamProvider`.

Again, it would be best if you'd allow `IClusterClient` inside the silo, maybe with specific initialization for the silo context. It would be the most elegant solution.

Thanks again!

jdom:
I would indeed recommend you do not create an IClusterClient to itself. We do have intentions to make this scenario a lot simpler (basically being able to call the silo outside of the context of a grain), but in the meantime you should avoid creating a real client against itself just for this.

Also, be aware that what you are doing with the timers is not robust enough to keep the grain activated forever. Since this is just a normal grain registered in the directory, if the silo that owns that partition of the directory dies, then the activation will realize it's no longer in the directory and be deactivated. @ReubenBond is currently looking into improving this for normal grains.
Maybe the alternative is to use GrainServices? I'm not familiar enough with them to know for sure. 
Alternatively you can consider using StatelessWorkers, which are not registered in the directory, and are always local to the caller (and besides the name, they are not required to be stateless).

shlomiw:
@jdom - you're completly right! I have to use a local grain since it's needed per Silo. StatelessWorker is a good idea.
Thanks for pointing that out!

galvesribeiro:
@shlomiw the GrainService as @jdom suggested is also a good option if you need one per silo.

shlomiw:
@ReubenBond  - maybe #3362 will solve my above issue and be the elegant solution I was hoping for. To share a common infra code with IClusterClient also for the silo process.

PR #2826, opened 2017-03-08
Fix build on VS 2017

ReubenBond:
VS 2017 complains that no platform is specified in the project.json file for these two projects and fails to build.


ReubenBond:
This is not needed - I'm not sure why VS was complaining earlier, but it works directly after a `git clean -xdf`.

PR #2829, opened 2017-03-08
Fix null reference exception in simple queue cache.

jason-bragg:
Simple queue cache was not rate limiting correctly, leading to messages being removed outside the expected cache purge behavior.  This led to null references in cache cursors under load.

Rate limiting fixed..
Code block which removed messages from cache outside the context of the expected purge behavior was removed.

### DIFF
--- a/src/OrleansProviders/Streams/Common/SimpleCache/SimpleQueueCache.cs
+++ b/src/OrleansProviders/Streams/Common/SimpleCache/SimpleQueueCache.cs
@@ -79,12 +79,7 @@ namespace Orleans.Providers.Streams.Common
         /// </summary>
         public virtual bool IsUnderPressure()
         {
-            if (cachedMessages.Count == 0) return false; // empty cache
-            if (Size < maxCacheSize) return false; // there is still space in cache
-            if (cacheCursorHistogram.Count == 0) return false;    // no cursors yet - zero consumers basically yet.
-            // cache is full. Check how many cursors we have in the oldest bucket.
-            int numCursorsInLastBucket = cacheCursorHistogram[0].NumCurrentCursors;
-            return numCursorsInLastBucket > 0;
+            return cacheCursorHistogram.Count >= NUM_CACHE_HISTOGRAM_BUCKETS;
         }
 
 
@@ -291,6 +286,8 @@ namespace Orleans.Providers.Streams.Common
         private void Add(IBatchContainer batch, StreamSequenceToken sequenceToken)
         {
             if (batch == null) throw new ArgumentNullException(nameof(batch));
+            // this should never happen, but just in case
+            if (Size >= maxCacheSize) throw new CacheFullException();
 
             CacheBucket cacheBucket;
             if (cacheCursorHistogram.Count == 0)
@@ -319,18 +316,6 @@ namespace Orleans.Providers.Streams.Common
 
             cachedMessages.AddFirst(new LinkedListNode<SimpleQueueCacheItem>(item));
             cacheBucket.UpdateNumItems(1);
-
-            if (Size > maxCacheSize)
-            {
-                //var last = cachedMessages.Last;
-                cachedMessages.RemoveLast();
-                var bucket = cacheCursorHistogram[0]; // same as:  var bucket = last.Value.CacheBucket;
-                bucket.UpdateNumItems(-1);
-                if (bucket.NumCurrentItems == 0)
-                {
-                    cacheCursorHistogram.RemoveAt(0);
-                }
-            }
         }
 
         internal static void Log(Logger logger, string format, params object[] args)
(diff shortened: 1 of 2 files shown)


src/OrleansProviders/Streams/Common/SimpleCache/SimpleQueueCache.cs
```diff
@@ -79,12 +79,7 @@ public SimpleQueueCache(int cacheSize, Logger logger)
         /// </summary>
         public virtual bool IsUnderPressure()
         {
-            if (cachedMessages.Count == 0) return false; // empty cache
-            if (Size < maxCacheSize) return false; // there is still space in cache
-            if (cacheCursorHistogram.Count == 0) return false;    // no cursors yet - zero consumers basically yet.
-            // cache is full. Check how many cursors we have in the oldest bucket.
-            int numCursorsInLastBucket = cacheCursorHistogram[0].NumCurrentCursors;
-            return numCursorsInLastBucket > 0;
+            return cacheCursorHistogram.Count >= NUM_CACHE_HISTOGRAM_BUCKETS;
```
gabikliot: Can u please explain why u had to change that function and how it will work now? Why the current code is correct? Not saying it's not, just want u to explain the rationale and how it will work now.

jason-bragg: The expected behavior of the cache is to report when it is 'under pressure' with the IsUnderPressure call.  When the cache is under pressure, it is not ready to receive more messages.  When it is not under pressure it reports how many messages it can accept with the GetMaxAddCount call.

In the previous version of the code, the cache was considered 'not under pressure' as long as the message count in the cache was less than the max cache size "Size < maxCacheSize", and GetMaxAddCount always reported 1/10th of the max cache size.  This made it possible for the cache to report that it had more capacity than it actually did.

For instance, if it had 199 messages in a cache with size 200, it would report that it was not under pressure (199<200) but report that it had capacity for 20 message (200/10), when it actually only had capacity for 1 (200-199).

When this happened, more messages would be added to the cache than it could handle, which was not supposed to be the case.  This problem was handled by a code block in the "add" call which removed messages from the back of the cache, which is not desirable, as the expected means of removing messages from the cache is the purge logic (TryPurgeFromCache).

This change considers the cache 'under pressure' whenever the cache as all of it's available histogram buckets in use (each holding 1/10th of the cache capacity).  This is a coarse check as the most recent bucket may not be entirely full, so the cache will report 'under pressure' when somewhere between 90-100% full.

This properly prohibits more messages from being added to the cache than it has capacity to support.

The solution could be more tight by varying the available capacity reported by the GetMaxAddCount call, but this logic is sufficient and simple.  I tend to favor simplicity unless more complex solutions are warranted.

gabikliot: Based on your description it sounds like there was simply a bug in get max add count. I  am in favor of fixing it and not changing the rest of the code, including not changing the is under pressure. It also sounds a trivial fix. We have the max size and the current size.

jason-bragg: Can you provide your reasoning why the previous IsUnderPressure logic was better?
I don't consider minimizing the changes as a high priority.  I'd prioritize simplicity and maintainability, and am of the opinion that the code is now simpler and more maintainable.
While figuring out what the problem was, understanding the importance of each check (within IsUnderPressure) added to cognitive load.  Once I understood the conditions and realized the checks were not necessary, I saw no reason to keep them.

gabikliot: So this is now a separate open question. You want to simplify or even improve the pressure detection. I need to think about that. Maybe. But the issue here is the bug in max count. The two issues are separate and not directly related. Let's fix the bug and separately think about pressure detection.

jason-bragg: This change addresses the bug.
It addresses it by modifying IsUnderPressure to report correctly given the GetMaxAddCount.

The back-off logic relies on both of these calls, and requires them to function in tandem.  Which call was behaving poorly is a matter of interpretation.  I've opted to address the problem by simplifying the more confusing of the two.

You're suggestion is to correct the bug by making GetMaxAddCount more complicated (admittedly trivially so) rather than simplifying the IsUnderPressure.

jafletch: While I don't have enough background on the code comment on the rightness or wrongness of the chosen fix here, I can tell you that the fix does appear to work. I was the one who originally brought the null ref to Jason's attention and requested this fix. I have it running now on my PPE system and I can confirm that the null ref I was seeing is now gone with this fix.

sergeybykov:
I'll merge this, so that we can include the fix in 1.4.1.

PR #2830, opened 2017-03-08
Allow echoing build script commands

jdom:
If the `_echo` environment variable is set (regardless of its value), then echo is not turned off. We were already doing it in some of the scripts, but not all. This is the exact same mechanism that http://github.com/dotnet/corefx uses

### DIFF
--- a/src/Build.cmd
+++ b/src/Build.cmd
@@ -1,29 +1,20 @@
-@REM NOTE: This script must be run from a Visual Studio command prompt window
-
-@setlocal
-@ECHO off
+@if not defined _echo @echo off
+setlocal
 
 SET CMDHOME=%~dp0.
-if "%VisualStudioVersion%" == "" (
+if not defined VisualStudioVersion (
     @REM Try to find VS command prompt init script
-    where /Q VsDevCmd.bat
-    if ERRORLEVEL 1 (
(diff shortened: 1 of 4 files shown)


src/Build.cmd
```diff
@@ -1,29 +1,20 @@
-@REM NOTE: This script must be run from a Visual Studio command prompt window
-
-@setlocal
-@ECHO off
+@if not defined _echo @echo off
+setlocal
 
 SET CMDHOME=%~dp0.
-if "%VisualStudioVersion%" == "" (
+if not defined VisualStudioVersion (
```
ReubenBond: I didn't know this syntax existed

src/Test.cmd
```diff
@@ -1,5 +1,6 @@
-@setlocal
-@ECHO off
+@if not defined _echo @echo off
+setlocal
+ECHO off
```
ReubenBond: I'm confused by this, it looks like you're disabling echo unconditionally

jdom: oops, forgot to remove the original echo off

src/Build.cmd
jdom: yup, me neither, found out when copying this functionality from the corefx repo

Issue #2808, opened 2017-03-03
Inconsistent InterfaceId for concrete implementation of generic grain interface


benjaminpetit:
### Issue

With concrete implementations of generic grain interface, we have a discrepancy between the InterfaceId  value stored in the  `GrainInterfaceMap` and the value used in the generated `GrainReference`

### Example

If we have a grain interface like this:
``` csharp
public interface IGenericGrain<T, U> : IGrainWithIntegerKey { ... }
```

Orleans will generate this `GrainReference`:
```csharp
internal class OrleansCodeGenGenericGrainReference<T, U> : global::Orleans.Runtime.GrainReference, global::UnitTests.GrainInterfaces.IGenericGrain<T, U> {
    ...
    protected override global::System.Int32 InterfaceId
    {
	get { return 1634994201; }
    }
    ...
}
```
Here the value `1634994201` comes from `GrainInterfaceUtils.GetGrainInterfaceId(typeof(IGenericGrain<,>))`

If we have this grain implementation on the silo:
``` csharp
class ConcreteGrainWithGenericInterfaceOfIntFloat : Grain, IGenericGrain<int, float> { ... }
```
at runtime, the silo will construct the `GrainInterfaceMap`, but to get the value of the InterfaceId it will call
```GrainInterfaceUtils.GetGrainInterfaceId(typeof(IGenericGrain<int, float>))``` which will return `1985271655`

So on this silo, in the `GrainInterfaceMap` we will have an entry for `InterfaceId=1985271655`, but the `InvokeMethodRequest` sent by the `GrainReference` will contain `InterfaceId=1634994201` which will NOT be in the `GrainInterfaceMap`

It seems that this is not an issue since we don't use the `InterfaceId` that comes from the `InvokeMethodRequest` (yet!)

~~~I think that we have two ways of fixing this:~~~
- ~~~Always use the "generic" id when adding an entry to the `GrainInterfaceMap`~~~
- ~~~Or compute the real interface id at runtime for `GrainReference` is this is a generic~~~

benjaminpetit:
In summary:

- We use the computed interface id of the **open** generic for the value stored in the generated GrainReference and  the generated MethodInvoker
- We use the computed id of the **closed** generic for the value stored in the GrainInterfaceMap

So far this seems logic:
- All grain implementations (open or close) will use the same GrainReference/MethodInvoker
[...]

PR #2833, opened 2017-03-09
When adding a concrete implementation of a generic interface, add also the generic version in the GrainInterfaceMap

benjaminpetit:
See #2808 for more explanations.

The fix is simple: add the "open generic" version of the interface in the `GrainInterfaceMap` so we can lookup information with the value returned by the `GrainReference` and the `InvokeMethodRequest` that are used by this implementation.

### DIFF
--- a/src/Orleans/Runtime/GrainInterfaceMap.cs
+++ b/src/Orleans/Runtime/GrainInterfaceMap.cs
@@ -133,10 +133,9 @@ namespace Orleans.Runtime
             lock (this)
             {
                 var interfaceId = GrainInterfaceUtils.GetGrainInterfaceId(iface);
-                var ifaceCompleteName = TypeUtils.GetFullName(iface);
-                var grainInterface = TypeUtils.GetRawClassName(ifaceCompleteName);
+                var interfaceName = TypeUtils.GetRawClassName(TypeUtils.GetFullName(iface));
                 var grainTypeInfo = grain.GetTypeInfo();
-                var grainClass = TypeUtils.GetFullName(grainTypeInfo);
+                var grainName = TypeUtils.GetFullName(grainTypeInfo);
                 var isGenericGrainClass = grainTypeInfo.ContainsGenericParameters;
                 var grainTypeCode = GrainInterfaceUtils.GetGrainClassTypeCode(grain);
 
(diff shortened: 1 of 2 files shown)


src/Orleans/Runtime/GrainInterfaceMap.cs
```diff
@@ -127,49 +128,68 @@ internal void AddMap(GrainInterfaceMap map)
...
                 if (localTestMode)
                 {
+                    var assembly = grainTypeInfo.Assembly.CodeBase;
                     if (!loadedGrainAsemblies.Contains(assembly))
                         loadedGrainAsemblies.Add(assembly);
                 }
             }
         }
 
+        private GrainInterfaceData AddOrGetGrainInterfaceData(Type iface, bool isGenericGrainClass)
```
ReubenBond: Nit: better to name this method GetOrAddXXX, since it only adds if it cannot get

benjaminpetit: I can change that

src/Orleans/Runtime/GrainInterfaceMap.cs
```diff
@@ -127,49 +128,68 @@ internal void AddMap(GrainInterfaceMap map)
...
+
+            // Add entry to mapping iface string -> data
+            var interfaceTypeKey = GetTypeKey(iface, isGenericGrainClass);
+            typeToInterfaceData[interfaceTypeKey] = grainInterfaceData;
+
+            // If we are adding a concrete implementation of a generic interface
+            // add also the latter to the map: GrainReference and InvokeMethodRequest 
+            // always use the id of the generic one
+            if (iface.IsConstructedGenericType)
+                GetOrAddGrainInterfaceData(iface.GetGenericTypeDefinition(), true);
```
ReubenBond: The meaning of the `isGenericGrainClass` parameter becomes a little strange here. This is correct, from what I can tell, but it's not obvious from the method signature that we should be passing true here. In effect, the `true` just informs the method that the "type key" should be a generic type definition (if it's a generic interface). An alternative would be to change the logic in `GetTypeKey` so that the if-statement reads:
`if (typeInfo.IsGenericTypeDefinition || (isGenericGrainClass && typeInfo.IsGenericType)`.

Does that make sense, or am I reasoning about it wrong?

PR #2834, opened 2017-03-09
Move every type to it's own file within Orleans.dll.

attilah:
This is the basis of creating a better abstraction project. Only file move was done with Resharper and verified afterwards, no any changes by hand made.


attilah:
It is not the time to put this in in this shape as of team meeting, closing without merging.

Issue #2644, opened 2017-01-26
Orleans Grains versioning


benjaminpetit:
# Orleans Grains versioning
Currently we assume that for a given Grain implementation, it will be the same across all the cluster. 

## Why
This is an issue when deploying a new version of a Grain with some method added/modified (signature of body). #2379 removed the assumption that all silos support the same set of types, but we need to go further and enable a scenario where we can have multiple grain versions for a given TypeCode in the cluster.

## Out of scope (for now)
-	Grain state versioning
-	We still assume that Orleans binaries are compatibles across versions

## Design
So far, we explored two different scenarios. There are not incompatibles, we might want to implement both on long term.
### 1. Versioning by adding interfaces
In this option, we consider the grain interface (contract) as immutable (as it is right now). You cannot add/remove method from an already implemented and deployed interface. To add new method, you must introduce a new interface. 
```` csharp
// In “V1” and “V2” silos
public interface IMyGrain : IGrainWithIntegerKey
{
    Task Method1();
}
// In “V1” only
public class MyGrain : Grain, IMyGrain
{
    public Task Method1() { … }
}
// In “V2” only
public interface IMyGrainV2 : IMyGrain
{
    Task Method2();
}
// In “V2” only
public class MyGrain : Grain, IMyGrain2
{
    public Task Method1() { … }
    public Task Method2() { … }
}
````
Pros:
-	“Natural” solution for service versioning

Cons:
-	After n versions, you have n different interfaces for the same grain in your codebase
-	Not possible to build/use “upgrade rules” in/from Orleans

### 2. Adding version attribute on interfaces
In this option, interfaces are not immutable anymore: we add an attribute on interface to specify their version when they change:
```` csharp
// In “V1” 
[Version(1)]
public interface IMyGrain : IGrainWithIntegerKey
{
    Task Method1();
}
// In “V2”
[Version(2)]
public interface IMyGrain : IGrainWithIntegerKey
{
    Task Method1();
    Task Method2();
}
````
Pros:
-	“Lighter” than the previous option
- We can have a default value for version number for all grain interfaces (bump this number for all new deployment to avoid compatibility issues), but let developer the possibility to override it
-	We could easily expose automatic upgrade/downgrade rules in Orleans (mix V1 and V2 for new placements for A/B testing, authorize direct calls from a V1 Grain to a V2 but not a V3, ect.)

Cons:
-	Developer must be careful when grain implementation is changed: he must decide whether the version number on the interface need to be changed or not

[...]

PR #2837, opened 2017-03-10
Grain interface versioning

benjaminpetit:
This is a first step for rolling upgrade: enable grain versioning.

In this PR we implemented the solution 2 described in issue #2644: adding a version attribute on Grain interfaces. 

## Proposed changes
The developer can specify the version with the `VersionAttribute` on the grain interfaces. Currently, when sending a request to a grain, we only send the target `GrainId` along with the `InterfaceId` and the `MethodId`: now we add the interface version information to this request.

It will be the responsibility of the silo to decide if an existing activation can handle a request of a given version, and in a case of a new activation what version to use.

I tried to divide the PR in logical commits, so it is easier to review:
- The "codegen part" is in b7fc911 and b84de04: I added the `Version` atribute and the `InterfaceVersion` member in `IGrainMethodInvoker` and `GrainReference`
- Then in 88562c9 I updated `GrainInterfaceMap` and `GrainTypeManager` to store/surface the version attribute
- b520310 is a simple refactor of `IPlacementDirector`: to place a new grain we also need now the requested grain interface along with its version
- In 37f634b and a3041d3 I implemented simple rules to always place new activations on silo with the latest version, and disable current one if a request cannot be processed by the current activation
-  b27d903 is adding tests for this feature. The setup might appear complicated, but I need to start two silos with different assemblies.

## Limitations
For now, we assume that all new implementations are backward compatibles, and that all new activations should be placed on silos that have the latest version of the interface. In next PR(s) we will extract this logic into “policies” and add some configurable options.


benjaminpetit:
@dotnet-bot test this please

@ReubenBond yes currently it is not supported since grain extensions are not in the GrainInterfaceMap. We could handle a separate logic for that, but I am not sure we really need versioning on extensions? I don't have enough experience with extension to be sure.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -215,15 +215,16 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
         {
             // For test only: if we have silos that are not yet in the Cluster TypeMap, we assume that they are compatible
             // with the current silo
             if (this.config.AssumeHomogenousSilosForTesting)
                 return AllActiveSilos;
 
-            var typeCode = grain.GetTypeCode();
-            var compatibleSilos = GrainTypeManager.GetSupportedSilos(typeCode).Intersect(AllActiveSilos).ToList();
+            var typeCode = target.GrainId.GetTypeCode();
+            var versions = new[] {GrainTypeManager.GetAvailableVersions(target.InterfaceId).Max()};
```
sergeybykov: If we always pass a array with a single element in it, what's the point of passing an array?

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -370,7 +375,20 @@ public void HandleIncomingRequest(Message message, ActivationData targetActivati
         {
             lock (targetActivation)
             {
-                if (targetActivation.State == ActivationState.Invalid)
+                if (targetActivation.Grain.IsGrain)
+                {
+                    var request = ((InvokeMethodRequest)message.GetDeserializedBody(this.serializationManager));
```
sergeybykov: Because interface version is part of `InvokeMethodRequest`, we have to deserialize message body before we can decide what to do with the message, potentially to forward it to another silos. In such a case this serialization is wasteful.

Could we instead put interface version into message headers to avoid deserialization here? That would also keep messages backward and forward compatible.

Going one step further (out of scope of this PR), should we consider moving `InterfaceId` and `MethodId` from `InvokeMethodRequest` to headers, and leave only arguments in the body?

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -637,8 +655,10 @@ internal void SendMessage(Message message, ActivationData sendingActivation = nu
             // second, we check for a strategy associated with the target's interface. third, we check for a strategy associated with the activation sending the
             // message.
             var strategy = targetAddress.Grain.IsGrain ? catalog.GetGrainPlacementStrategy(targetAddress.Grain) : null;
+            var request = message.GetDeserializedBody(this.serializationManager) as InvokeMethodRequest;
```
sergeybykov: Another instance of potentially wasteful deserialization?

src/OrleansRuntime/Catalog/Catalog.cs
benjaminpetit: When implementing " policies" later, we could return a list of version candidates, not only one. But yes, for now we only use the latest one.

src/OrleansRuntime/Core/Dispatcher.cs
benjaminpetit: Interface version alone will not be sufficient, we will need to move also `InterfaceId`. We don't use the `MethodId` for placement, so it could make sense to leave it here.

If we move `InterfaceId` and `InterfaceVersion` in headers, we will deserialize two additional intergers for each message received. Here we only deserialize the whole body for new placement and when the target activation is about to process the request (needed just after anyway, unless the current activation is incompatible).

So it is hard to know which implementation will be more efficient, as it will largely depends on the use case (lot of new activations with big params vs not so much new activations with small params)

THe best solution seems to move `InterfaceId` and `InterfaceVersion` into a new structure, included in the message payload, that would be deserialized only if needed.

src/OrleansRuntime/Core/Dispatcher.cs
benjaminpetit: Likely more wasteful than the other instance, since there is little chance the new activation will be placed on the current silo (vs activation about to process the request)

src/Orleans/CodeGeneration/IGrainMethodInvoker.cs
```diff
@@ -11,6 +11,8 @@ public interface IGrainMethodInvoker
         /// <summary> The interface id that this invoker supports. </summary>
         int InterfaceId { get; }
 
+        ushort InterfaceVersion { get; }
```
benjaminpetit: Good point, will test that

src/Orleans/Serialization/BuiltInTypes.cs
```diff
@@ -1850,6 +1852,7 @@ internal static void SerializeInvokeMethodRequest(object obj, ISerializationCont
         internal static object DeserializeInvokeMethodRequest(Type expected, IDeserializationContext context)
         {
             int iid = context.StreamReader.ReadInt();
+            ushort iVersion = context.StreamReader.ReadUShort();
```
benjaminpetit: Not sure of that, better invoke @ReubenBond for this :)

ReubenBond: We can't tell if there's more data left, but we could narrow argCount to a `ushort` and put `iVersion` where the most significant bits currently are

src/OrleansRuntime/Core/Dispatcher.cs
ReubenBond: There are optimizations which could be made by deferring deserialization of the args until the invoker: currently we use and `object[]`, which means that every arg has its type serialized in full. By deferring deserialization, we can take advantage of the expected argument types. It would take a little work, but if you want to make changes there anyway then it might be worth it.

Bear in mind that `InvokeMethodRequest` is a public type, though

src/Orleans/CodeGeneration/IGrainMethodInvoker.cs
benjaminpetit: From my testing with our serializer, there is a real beneif in using ushort, it takes less space. I checked in `BuiltInTypes.SerializeInvokeMethodRequest` the byte array that resulted from the serialization.

src/Orleans/Serialization/BuiltInTypes.cs
benjaminpetit: Yes we could but... Do we need to make it forward compatible right now?

ReubenBond: If you don't do something like that, then you wont be able to read `InvokeMethodRequest` instances from older silos. I'm not saying it needs to be done, just how it could be done.

src/OrleansRuntime/Core/Dispatcher.cs
ReubenBond: This kind of improvement would be easy if we had a type like Bond's `IBonded` for deferred deserialization

sergeybykov:
I suggest we consider a compromise solution here: add a boolean flag to headers to indicate requests with non-zero version, and condition execution of the added logic on that. Later, as a separate optimization, we can refactor the body deserialization logic in such a way that we could cheaply retrieve request 'metadata' (interface ID, version, method ID) without eagerly deserializing request arguments.

benjaminpetit:
@dotnet-bot test this please

src/Orleans/Messaging/Message.cs
```diff
@@ -1106,6 +1127,9 @@ public static void Serializer(object untypedInput, ISerializationContext context
                 if ((headers & Headers.IS_NEW_PLACEMENT) != Headers.NONE)
                     writer.Write(input.IsNewPlacement);
 
+                if ((headers & Headers.IS_USING_IFACE_VERSION) != Headers.NONE)
```
ReubenBond: No need to change it, but this is a bit redundant: the flag is only set in headers if the bool is true, so we don't need to waste a byte & a read/write on it.

src/Orleans/Runtime/GrainInterfaceMap.cs
```diff
@@ -83,6 +85,16 @@ internal IList<int> SupportedGrainTypes
             get { return implementationIndex.Keys.ToList(); }
         }
 
+        internal IList<GrainClassData> SupportedGrainClassData
+        {
+            get { return implementationIndex.Values.ToList(); }
```
ReubenBond: generally not a fan of these 'allocate-per-access' properties

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -637,8 +650,12 @@ internal void SendMessage(Message message, ActivationData sendingActivation = nu
             // second, we check for a strategy associated with the target's interface. third, we check for a strategy associated with the activation sending the
             // message.
             var strategy = targetAddress.Grain.IsGrain ? catalog.GetGrainPlacementStrategy(targetAddress.Grain) : null;
+            var request = message.IsUsingIfaceVersion
+                ? message.GetDeserializedBody(this.serializationManager) as InvokeMethodRequest
+                : null;
+            var target = new PlacementTarget(message.TargetGrain, request?.InterfaceId ?? 0, request?.InterfaceVersion ?? 0);
```
ReubenBond: not a fan of these hot-path allocs, is there any way we can avoid them?

src/Orleans/Runtime/GrainInterfaceMap.cs
benjaminpetit: They are not used in our code. I just kept them for debugging purpose. I could remove them, so no one can call them later in "real" code...

src/OrleansRuntime/Core/Dispatcher.cs
benjaminpetit: Not sure here if it would be preferable to use a `struct` instead. Would be the compiler clever enough to "inline" that?

But I definitely don't like to have function with too many parameters

src/Orleans/Runtime/GrainInterfaceMap.cs
ReubenBond: Removal is good, that way they don't get accidentally used

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -272,6 +272,17 @@ protected virtual int InterfaceId
         /// <summary>
         /// Implemented in generated code.
         /// </summary>
+        protected virtual ushort InterfaceVersion
+        {
+            get
+            {
+                throw new InvalidOperationException("Should be overridden by subclass");
+            }
```
jdom: Sometimes in the code we just new up a GrainReference. Could it conflict with this new feature?

src/Orleans/Serialization/BuiltInTypes.cs
```diff
@@ -1836,8 +1836,10 @@ internal static void SerializeInvokeMethodRequest(object obj, ISerializationCont
             var request = (InvokeMethodRequest)obj;
 
             context.StreamWriter.Write(request.InterfaceId);
+            context.StreamWriter.Write(request.InterfaceVersion);
```
jdom: InvokeMethodRequest should in theory only be used for messaging, and not persistence. So do we care if it's not back compat, since the entire change itself is not backcompat to for side by side deployment?

src/Orleans/Serialization/BuiltInTypes.cs
jdom: As I mentioned in another comment, do we care about it? Deploying different versions of Orleans itself side by side is not supported. This PR enables deploying 2 different *app* versions side by side, as long as they both use the same version of Orleans that enables versioning.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -215,15 +215,26 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
                 return AllActiveSilos;
 
-            var typeCode = grain.GetTypeCode();
-            var compatibleSilos = GrainTypeManager.GetSupportedSilos(typeCode).Intersect(AllActiveSilos).ToList();
+            var typeCode = target.GrainId.GetTypeCode();
+            IReadOnlyList<SiloAddress> silos;
+            if (target.InterfaceVersion > 0)
+            {
+                var versions = new[] { GrainTypeManager.GetAvailableVersions(target.InterfaceId).Max() };
+                silos = GrainTypeManager.GetSupportedSilos(typeCode, target.InterfaceId, versions);
```
jdom: IMO, the default safe choice would be to use the version that was requested, not the maximum version

src/OrleansTestingHost/TestCluster.cs
```diff
@@ -278,6 +287,8 @@ public SiloHandle StartAdditionalSilo()
                 baseSiloPort, 
                 baseGatewayPort);
 
+            configurationModification?.Invoke(nodeConfig);
+
```
jdom: These are remnants of old code, but you are not really using the overload of StartAdditionalSilo that takes an action, right?

test/Tester/HeterogeneousSilosTests/UpgradeTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                    assemblyGrainsV2Dir = new DirectoryInfo(@"..\" + AssemblyGrainsV2Vs + target);
+                }
+            }
+
+            this.options = new TestClusterOptions(2);
+            options.ClusterConfiguration.Globals.AssumeHomogenousSilosForTesting = false;
+            options.ClusterConfiguration.Globals.TypeMapRefreshInterval = refreshInterval;
+            options.ClientConfiguration.Gateways.RemoveAt(1); // Only use primary gw
+
+            StartSiloV1().Wait(TimeSpan.FromSeconds(10));
```
jdom: 10 seconds is probably too aggressive for our build servers

test/Tester/HeterogeneousSilosTests/UpgradeTests.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            // Start a new silo with V2
+            await StartSiloV2();
+
+            // New activation should be V2
+            for (var i = numberOfGrains; i < numberOfGrains * 2; i++)
+            {
+                var grain = GrainFactory.GetGrain<IVersionUpgradeTestGrain>(i);
+                Assert.Equal(2, await grain.GetVersion());
+            }
```
jdom: Shouldn't you validate that the range 0-99 are still using version 1?

test/Tester/HeterogeneousSilosTests/UpgradeTests.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            // Stop the V2 silo
+            await StopSiloV2();
+
+            // Now all activation should be V1
+            for (var i = numberOfGrains * 2; i < numberOfGrains * 3; i++)
+            {
+                var grain = GrainFactory.GetGrain<IVersionUpgradeTestGrain>(i);
+                Assert.Equal(1, await grain.GetVersion());
+            }
```
jdom: validate that all (0-299) are now in v1?

src/Orleans/Runtime/GrainReference.cs
benjaminpetit: This property is only used when `InterfaceId` is used (and this property also throws `InvalidOperationException("Should be overridden by subclass")` by default), so no, it will not conflict

test/Tester/HeterogeneousSilosTests/UpgradeTests.cs
benjaminpetit: It worked fine in Jenkins build for now, let's keep this value?

src/OrleansRuntime/Catalog/Catalog.cs
benjaminpetit: It will change in a next PR

benjaminpetit: About comments on the serialization/deserialization of the interface version, I don't understand what you want me to do here

src/Orleans/Serialization/BuiltInTypes.cs
benjaminpetit: I don't think we ever guaranteed that two versions of Orleans were compatible, why starts now?

jdom: Exactly my point, we shouldn't care about back compat for this value that shouldn't be persisted, only used for messaging.

test/Tester/HeterogeneousSilosTests/UpgradeTests.cs
```diff
@@ -87,17 +87,18 @@ public UpgradeTests()
...
 
             // New activation should be V2
-            for (var i = numberOfGrains; i < numberOfGrains * 2; i++)
+            for (var i = 0; i < numberOfGrains * 2; i++)
             {
                 var grain = GrainFactory.GetGrain<IVersionUpgradeTestGrain>(i);
-                Assert.Equal(2, await grain.GetVersion());
+                var expectedVersion = i < numberOfGrains ? 1 : 2;
+                Assert.Equal(expectedVersion, await grain.GetVersion());
             }
```
jdom: Sorry, but just add 2 loops to make the assert blatantly clear. Also checkout the stale code comment preceding it

src/Orleans/Messaging/Message.cs
```diff
@@ -263,6 +263,15 @@ public bool IsNewPlacement
             }
         }
 
+        public bool IsUsingIfaceVersion
+        {
+            get { return Headers.IsUsingIfaceVersion; }
```
jdom: The abbreviation comment actually applied to all the IsUsingIfaceVersion properties also, and IS_USING_INTERFACE_VERSION.

src/Orleans.sln
```diff
@@ -229,6 +229,16 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BondUtils.Tests", "..\test\
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Consul.Tests", "..\test\Consul.Tests\Consul.Tests.csproj", "{D964FE9E-008A-4EDB-BBD7-8C57185C7A18}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Versions", "Versions", "{2C188181-1B01-4E77-A83D-F6A2EA855BE9}"
```
sergeybykov: Consider if it's possible to reduce the number of added test projects by combining grain interfaces and implementations and potentially putting one of the sets into existing test projects.

src/OrleansRuntime/Core/Dispatcher.cs
sergeybykov: Struct seems like a good option to me.

---
In reply to: [107806329](https://github.com/dotnet/orleans/pull/2837#discussion_r107806329) [](ancestors = 107806329)

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
```diff
@@ -16,14 +16,16 @@ internal class GrainTypeManager
...
+        private Dictionary<int, List<SiloAddress>> supportedSilosByTypeCode;
         private readonly Logger logger = LogManager.GetLogger("GrainTypeManager");
         private readonly GrainInterfaceMap grainInterfaceMap;
         private readonly Dictionary<int, InvokerData> invokers = new Dictionary<int, InvokerData>();
         private readonly SiloAssemblyLoader loader;
         private readonly SerializationManager serializationManager;
         private readonly MultiClusterRegistrationStrategyManager multiClusterRegistrationStrategyManager;
 		private readonly PlacementStrategy defaultPlacementStrategy;
+        private Dictionary<int, HashSet<ushort>> interfaceVersions;
+        private Dictionary<int, Dictionary<ushort, List<SiloAddress>>> supportedSilosByInterface;
```
sergeybykov: `interfaceVersions` appears unused.

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
```diff
@@ -143,11 +145,36 @@ internal void SetInterfaceMapsBySilo(Dictionary<SiloAddress, GrainInterfaceMap>
...
-        internal IList<SiloAddress> GetSupportedSilos(int typeCode)
+        internal IReadOnlyList<SiloAddress> GetSupportedSilos(int typeCode)
         {
             return supportedSilosByTypeCode[typeCode];
         }
 
+        internal IReadOnlyList<SiloAddress> GetSupportedSilos(int typeCode, int ifaceId, IList<ushort> versions)
+        {
+            var result = new List<SiloAddress>();
+
```
sergeybykov: Can we cache the lists and invalidate the cache upon a cluster state change. This will allocate on every placement?

src/Orleans.sln
benjaminpetit: I combined interfaces with their grain implementations. Even if I could merge V1 in the TestGrains, I prefer to keep two projects for grains because it will enable more testing scenario later (deploy v2, then deploy v1)

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
benjaminpetit: Yes we can, but it is a bit hard to cache since we are taking a list of versions as an input. I think I can refactor to take only one and add caching.

jdom:
@dotnet-bot test this please

benjaminpetit:
@dotnet-bot test this please

@jdom it seems that some AsyncPipelineTests aire failing in netfx-functional/
netstandard-win-functional

jdom:
Moving the conversation here:
@benjaminpetit do you know why this test is failing? https://ci.dot.net/job/dotnet_orleans/job/master/job/netfx-bvt_prtest/13/testReport/junit/Tester.HeterogeneousSilosTests/UpgradeTests/UpgradeSeveralQueuedRequestsTest/

BTW, these tests should probably not be BVT, but just Functional. Otherwise please mark them as SlowBVT instead, and also apply the Functional category.

benjaminpetit:
Timing issue I guess. I can increase the waiting time and move categories like you said if you want

jdom:
Yes, please, let's make it pass consistently, and also not run them with BVTs, but just functionals for now

benjaminpetit:
I just submitted #2906

PR #2843, opened 2017-03-14
Minimal disambiguation of service types over DependencyInjection.Abstractions

jason-bragg:
Since we're limited to features provided by Microsoft.Extensions.DependencyInjection.Abstractions, neither we nor our users that don't use more advanced DI, have a means of acquiring a service from a service provider by name or key.  This PR introduces a minimal implementation for this.

To register a keyed or named service in the service collection:

                // add keyed services
                services.AddTransient<Zero>().WithKey<int, IValue<int>, Zero>(0);
                services.AddTransient<One>().WithKey<int, IValue<int>, One>(1);

                // add named services
                services.AddTransient<This>().WithName<IValue<string>, This>("this");
                services.AddTransient<That>().WithName<IValue<string>, That>("that");

To acquire these from a service provider:

                // get by key
                IValue<int> intValue = ServiceProvider.GetServiceByKey<int,IValue<int>>(0);

                // get by name
                IValue<string> stringValue = ServiceProvider.GetServiceByKey<int,IValue<string>>("that");

### DIFF
--- a/src/Orleans/Utils/IKeyedServiceCollection.cs
+++ b/src/Orleans/Utils/IKeyedServiceCollection.cs
@@ -1,0 +1,37 @@
+﻿
+using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.Runtime
+{
+    /// <summary>
+    /// Collection of services that can be disambiguated by key
+    /// </summary>
+    public interface IKeyedServiceCollection<in TKey, out TService>
+        where TService : class
+    {
+        TService GetService(TKey key);
+    }
+
+    public static class KeyedServiceExtensions
+    {
+        /// <summary>
+        /// Acquire a service by key.
+        /// </summary>
+        public static TService GetServiceByKey<TKey, TService>(this IServiceProvider services, TKey key)
+            where TService : class
+        {
+            IKeyedServiceCollection<TKey, TService> collection = services.GetService<IKeyedServiceCollection<TKey, TService>>();
+            return collection?.GetService(key);
+        }
+
+        /// <summary>
+        /// Acquire a service by name.
+        /// </summary>
+        public static TService GetServiceByName<TService>(this IServiceProvider services, string name)
+            where TService : class
+        {
+            return services.GetServiceByKey<string,TService>(name);
+        }
+    }
+}
--- a/test/Tester/DependencyInjectionDisambiguationTests.cs
+++ b/test/Tester/DependencyInjectionDisambiguationTests.cs
@@ -1,0 +1,63 @@
+﻿
+using System;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+using Orleans.Runtime;
+
+namespace UnitTests.General
+{
+    [TestCategory("DI")]
+    public class DependencyInjectionDisambiguationTests
+    {
+        [Fact, TestCategory("BVT"), TestCategory("Functional")]
+        public void DisambiguateByKeyTest()
+        {
+            IServiceProvider services = ConfigureServices();
+            int actual0 = services.GetServiceByKey<int, IValue<int>>(0).Value;
+            Assert.StrictEqual(0, actual0);
+            int actual1 = services.GetServiceByKey<int, IValue<int>>(1).Value;
+            Assert.StrictEqual(1, actual1);
+        }
+
+        [Fact, TestCategory("BVT"), TestCategory("Functional")]
+        public void DisambiguateByNameTest()
+        {
+            IServiceProvider services = ConfigureServices();
+            string actualThis = services.GetServiceByName<IValue<string>>("this").Value;
+            Assert.StrictEqual("this", actualThis);
+            string actualThat = services.GetServiceByName<IValue<string>>("that").Value;
+            Assert.StrictEqual("that", actualThat);
+        }
+
+        private interface IValue<out TValue>
+        {
+            TValue Value { get; }
+        }
+
+        private class SomeValue<TValue> : IValue<TValue>
+        {
+            public TValue Value { get; set; }
+        }
+
+        private class ValueServiceCollection<TValue> : IKeyedServiceCollection<TValue, IValue<TValue>>
+        {
+            public IValue<TValue> GetService(TValue name)
+            {
+                return new SomeValue<TValue> { Value = name };
+            }
+        }
+
+        private IServiceProvider ConfigureServices()
+        {
+            IServiceCollection services = new ServiceCollection();
+
+            // add services by Key;
+            services.AddSingleton<IKeyedServiceCollection<int, IValue<int>>, ValueServiceCollection<int>>();
+
+            // add named services
+            services.AddTransient<IKeyedServiceCollection<string, IValue<string>>, ValueServiceCollection<string>>();
+
+            return services.BuildServiceProvider();
+        }
+    }
+}
--- a/src/Orleans/Orleans.csproj
+++ b/src/Orleans/Orleans.csproj
@@ -182,6 +182,7 @@
     <Compile Include="Streams\Providers\StreamProviderUtils.cs" />
     <Compile Include="Streams\PubSub\StreamSubscriptionManagerExtensions.cs" />
     <Compile Include="Utils\Factory.cs" />
+    <Compile Include="Utils\IKeyedServiceCollection.cs" />
     <Compile Include="Utils\ReferenceEqualsComparer.cs" />
     <Compile Include="Serialization\ReflectedSerializationMethodInfo.cs" />
     <Compile Include="Streams\Core\StreamIdentity.cs" />
--- a/src/Orleans/Utils/ServiceCollectionExtensions.cs
+++ b/src/Orleans/Utils/ServiceCollectionExtensions.cs
@@ -2,7 +2,7 @@ using System.Linq;
 
 using Microsoft.Extensions.DependencyInjection;
 
-namespace Orleans
+namespace Orleans.Runtime
 {
     /// <summary>
     /// Extension methods for <see cref="IServiceCollection"/>.
--- a/test/Tester/Tester.csproj
+++ b/test/Tester/Tester.csproj
@@ -48,6 +48,7 @@
     <Compile Include="ClientConnectionTests\ClientDisconnectionEventTests.cs" />
     <Compile Include="ClientConnectionTests\GatewayConnectionTests.cs" />
     <Compile Include="CollectionFixtures.cs" />
+    <Compile Include="DependencyInjectionDisambiguationTests.cs" />
     <Compile Include="EventSourcingTests\AccountGrainTests.cs" />
     <Compile Include="EventSourcingTests\ChatGrainTests.cs" />
     <Compile Include="EventSourcingTests\CountersGrainPerfTests.cs" />


src/Orleans/Utils/IKeyedServiceCollection.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        TService GetService(TKey key);
+    }
+
+    public static class KeyedServiceExtensions
+    {
+        /// <summary>
+        /// Acquire a service by key.
+        /// </summary>
+        public static TService GetServiceByKey<TKey, TService>(this IServiceProvider services, TKey key)
```
jason-bragg: Suggest TKey be last generic argument.  May be more intuitive, as user probably think about what they want to get, then what key they use to get it.

Keep generic argument order uniform across all calls.

ReubenBond: TKey being the first generic argument is more common, though

src/Orleans/Utils/IKeyedServiceDescriptor.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        public static IServiceCollection SupportKeyedServices(this IServiceCollection collection)
+        {
+            return collection.AddSingleton(typeof(IKeyedServiceCollection<,>), typeof(KeyedServiceCollection<,>));
+        }
+
+        /// <summary>
+        /// Associates a key with a specialization of a service type.
+        /// </summary>
+        public static IServiceCollection WithKey<TKey, TService, TSpecialization>(this IServiceCollection collection, TKey key)
```
jason-bragg: WithKey and WithName look nice, but we've no means of enforcing that they actually conform to the previously added item.  The below will compile:

services.AddTransient<Zero>().WithKey<int, IValue<string>, That>("that");

Which doesn't really make sense.  Do we want to be more explicit about what we're doing here, at the cost of readability?  Something like:

services.AddTransient<Zero>();
services.AddKeyForService<IValue<string>, Zero, int>(0);

src/Orleans/Utils/IKeyedServiceCollection.cs
jason-bragg: I'm good with that.  I'm dyslexic so I don't trust my ordering judgment.  What is intuitive to me rarely works for others.

For instance, I think ok button should be on the right and cancel should be on the left, like GitHub does it, which is wrong by most UI design rules. :)

src/Orleans/Utils/IKeyedServiceDescriptor.cs
ReubenBond: `services.WithKey<string, ISomeService, SomeService>("la");` is a bit weird to me, `AddKeyed` sounds more right. Of course, there's still no lifecycle in there (and there ought to be ), so maybe `AddKeyedSingleton` is best. Below would then become `AddNamedSingleton`. 

What do you think?

src/Orleans/Utils/IKeyedServiceDescriptor.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public KeyedServiceCollection(IServiceProvider services)
+        {
+            this.keyedServices = services.GetServices<IKeyedServiceDescriptor<TKey, TService>>().ToArray();
+            this.services = services;
+        }
+
+        public TService GetService(TKey key)
+        {
+            IKeyedServiceDescriptor<TKey, TService> keyedService = this.keyedServices.FirstOrDefault(s => s.Equals(key));
```
ReubenBond: Correctness here would be more obvious if `TKey` was exposed as an accessible property and used for the comparison rather than implementing `IEquatable<TKey>`.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
jason-bragg: This call adds only  a mapping between the key and the service type, so the lifecycle is that of whatever one registered the service type with.

I just noticed that the displayed text is missing some of the original post,  source still has it, but displayed source does not.. I'll try to fix.

jason-bragg: Ok, kinda fixed, as best I can.
So the pattern is to register one's service, then register the key mapping.
We can do both in a single call, but we'd need to support all of the existing patterns for adding a service (scope, transient, singleton, + all overrides).  That may be better experience though.
Something like:

services.AddTransientServiceWithKey<int, IValue, Zero>(0);

jason-bragg: I just realized you commented on the same line as I did, but weren't responding to my comment. My misunderstanding.  Please consider my comments with the context of my comment on this same line.  The intended pattern was not intended to be:

services.WithKey<string, ISomeService, SomeService>("la"); 

It was intended to be used like:
services.AddTransient< SomeService >.WithKey<string, ISomeService, SomeService>("la");

But as you can see, the api doesn't enforce this usage, just enables it, which I suspected may be unclear, and you're comment has confirmed.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
```diff
@@ -0,0 +1,85 @@
...
+            return keyedService == null
+                ? default(TService)
+                : services.GetService(keyedService.Service) as TService;
```
ReubenBond: The problem with this approach is that it cannot distinguish between two `MemoryStorageProvider` instances with different names, so I cannot expect the following to behave correctly (it wil:
```C#
services.SupportKeyedServices();
services.WithName<IStorageProvider, AzureTableStorage>("users");
services.WithName<IStorageProvider, AzureTableStorage>("pages");
var provider = services.BuildServiceProvider();
var users = provider.GetServiceByKey<IStorageProvider>("users");
```
`users` will actually be the last registered `IStorageProvider`.

Perhaps I'm wrong.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        /// <summary>
+        /// Associates a key with a specialization of a service type.
+        /// </summary>
+        public static IServiceCollection WithKey<TKey, TService, TSpecialization>(this IServiceCollection collection, TKey key)
+            where TService : class
+            where TSpecialization : class, TService
+        {
+            return collection.AddSingleton<IKeyedServiceDescriptor<TKey, TService>>(sp => new KeyedServiceDescriptor<TKey, TService, TSpecialization>(key));
```
ReubenBond: This extends the lifecycle of `TService` to a singleton, even if it's transitive or scoped.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        /// <summary>
+        /// Associates a name with a specialization of a service type.
+        /// </summary>
+        public static IServiceCollection WithName<TService, TSpecialization>(this IServiceCollection collection, string name)
```
ReubenBond: We should also have overloads which take an instance and a factory in addition to these methods which take an implementation type.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
jason-bragg: yes, but this is just the descriptor, their is no need for it to be anything other than singleton, because it only maps the key to the specific type.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
ReubenBond: This model only allows one service of a given implementation to be meaningfully registered. I figure there's benefit in registering multiple instances of the same implementation - they may have different configuration or different semantics based on their key. Eg, maybe there would be two schedulers/loggers/something, one for "system" services and one for "application" services.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
ReubenBond: Oh, it doesn't, because the descriptor doesn't memoize the service, my bad.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
ReubenBond: I agree that the fluent approach doesn't work for creating non-trivial registrations in IServiceCollection. It would be nice to get an Autofac-like API, but this is what we've got.

src/Orleans/Utils/IKeyedServiceDescriptor.cs
jason-bragg: In the above code, the users will always get null, as no azure storage provider was ever registered.  As for the problem of mapping to multiple instance of the same specialization, yes, out of the box, this is not supported.  This only disambiguates between different specializations of a base service type.

For storage providers, we'd planned to simply put the existing manager behind the IKeyedServiceCollection interface, and use the existing managers.

Question, do we need disambiguation of separate instances of a service type, rather than specializations of that type?  I was not of the impression that other DI containers (autofac, unity) supported this, am I mistaken?

src/Orleans/Utils/IKeyedServiceDescriptor.cs
jason-bragg: A couple times in the design discussion an 'autofac' like api was requested, and I asked for specific suggestions because 'make it like this thing over here' as a requirement or feature request has bit me too many times.  I'd be happy to revisit these calls, sincerely, but if there is a pattern you'd like me to consider I'll expect specifics.

jdom: I thought the plan was to not tackle how you would register these in a generic manner (at least from our discussion, it looked like you didn't want to get there yet) but focus on how to resolve. For the registering, I'm certain that just doing type mapping will not be enough, we need to allow configuring the registered named service, which is the current prototype I'm working on.

src/Orleans/Utils/IKeyedServiceCollection.cs
ReubenBond: > For instance, I think ok button should be on the right and cancel should be on the left, like GitHub does it, which is wrong by most UI design rules. :)

I find that weird, too: 'Ok' is like 'Next' and I consider right forwards usually. Samsung phones have the back button on the right, too :(

jdom:
As I mentioned in one of the threads (but adding a general discussion comment), I'm not sure we should tackle a generic way to register things at this point, as at least the proposal only tackles registering types, and not configured instances, where named services is likely where they are more useful. Resolving it's fine to generalize (which defines the shape of the keyed service collection), but I currently don't agree with the way to build the collection.

src/Orleans/Utils/IKeyedServiceCollection.cs
```diff
@@ -0,0 +1,37 @@
+﻿
+using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans
```
jdom: this namespace looks too generic to put extension methods for configuration

test/Tester/DependencyInjectionDisambiguationTests.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        private readonly Fixture fixture;
+
+        public class Fixture : BaseTestClusterFixture
+        {
+            protected override TestCluster CreateTestCluster()
+            {
+                var options = new TestClusterOptions();
+                options.ClusterConfiguration.UseStartupType<MyStartup>();
+                return new TestCluster(options);
```
jdom: feels a little too much to spin up a cluster just to test these extensions that are agnostic to Orleans.

test/Tester/DependencyInjectionDisambiguationTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        private class SomeValue<TValue> : IValue<TValue>
+        {
+            public TValue Value { get; set; }
+        }
+
+        private class ValueServiceCollection<TValue> : IKeyedServiceCollection<TValue, IValue<TValue>>
+        {
+            public IValue<TValue> GetService(TValue name)
+            {
+                return new SomeValue<TValue> { Value = name };
```
jdom: I understand this, as we've been talking a lot about it lately, but it probably makes more sense and it will be closer to our future usage if you pre-allocate a few keyed objects while constructing the ValueServiceCollection, and they this one just accesses them.
then you just register the singletons by passing an initialized collection

src/Orleans/Utils/IKeyedServiceCollection.cs
jason-bragg: Most things in utils are either Orleans or Orleans.runtime.
Most Orleans public apis,are in these namespaces.

test/Tester/DependencyInjectionDisambiguationTests.cs
jason-bragg: Thought it would be good to test this working in a grain, but fair point.  Test can be much simpler.  Will simplify

src/Orleans/Utils/IKeyedServiceCollection.cs
attilah: It depends on the nature of the UI piece (window, message/popup box, etc.). But yeah, Ok, Cancel is the "right" order form LTR, but default is not always the Ok. :D

src/Orleans/Utils/IKeyedServiceDescriptor.cs
attilah: 👍 for @jdom, I miss the ability to have instances registered with a key, not just a type, this is why I and @ReubenBond suggested on the thread the we should store a TKey, ServiceDescriptor and not TKey, TService.

TService does not carry enough information. With this minimal extension we can't have 2 different configured providers of the same provider type for example.

jason-bragg:
Ready for review

test/Tester/DependencyInjectionDisambiguationTests.cs
jdom: I still find this test that doean't pre-allocate a fixed number of keyed services a little bit misleading, but it's a minor test issue.

src/Orleans/Utils/IKeyedServiceCollection.cs
jdom: Even if not completely right either, I would prefer we put this in Orleans.Runtime as other utilities. Putting it in the root namespace would mean that this extension method should be discoverable by all usages of Orleans, where in fact this is not Orleans' specific, and we probably don't expect users to care about it, only extension implementations would.

sergeybykov: I agree that it's better to keep the top `Orleans` namespace small and not have optional things there.

galvesribeiro:
Great work! That is why I always said that those AspNet abstractions would eventually limit us somehow and should be avoided and let people use whatever they want. The same applies to configuration and logging.

ReubenBond:
@galvesribeiro but we need to use *something*, so while I would prefer that everyone use Autofac, we needed to pick something which everyone can work with.

galvesribeiro:
@ReubenBond I think the best of the worlds is to have a very simple interface and let people implement it in whatever container they want... depending on those asp.net interfaces is a big problem IMHO

ReubenBond:
@galvesribeiro but we need to also have a default container to use. We have a simple interface and we do let users provide their own implementation - that's part of the problem. The issue is that semantics are not strong enough and so we need to perform some of the heavy lifting which we'd rather the container perform.

galvesribeiro:
IDK... I still think that would worth more have a very simple container default container implementation of the common interface and let people replace it as they wish. Rely on ASP.Net things still problematic to me. If that was at least something from the core .net framework, ok...

jason-bragg:
To support application developers choice of container, we chose to limit ourselves to an abstraction layer.  Rather than develop our own abstraction layer, we chose to use asp.net's, with the following expectations (hopes?)"

- The popularity of the abstraction will provide us integration with the most popular containers with little effort.
- Due to the popularity of the abstraction, users will be familiar with it and aware of it's limitations and integration issues.
- Over time the abstraction will be maintained and improved with little to no effort from us.

While the solution is not perfect, for the investment of development time, it is, IMO, far better than introducing, developing, testing, popularizing (integrating with existing containers and familiarizing users with it) and maintaining our own abstraction layer.

PR #2848, opened 2017-03-14
Update and cleanup VSIX

attilah:
Update and cleanup VSIX, build with new SDK (no install needed).
No VSPackage inside VSIX built.


src/Build.cmd
```diff
@@ -2,6 +2,38 @@
...
+
+SET VSWHERE_REMOTE_PATH=https://github.com/Microsoft/vswhere/releases/download/1.0.55/vswhere.exe
+SET VSWHERE_LOCAL_PATH=%CMDHOME%\vswhere.exe
+
+if NOT exist "%VSWHERE_LOCAL_PATH%" (
+  echo Downloading vswhere.exe from %VSWHERE_REMOTE_PATH%.
+  powershell -NoProfile -ExecutionPolicy unrestricted -Command "$retryCount = 0; $success = $false; do { try { (New-Object Net.WebClient).DownloadFile('%VSWHERE_REMOTE_PATH%', '%VSWHERE_LOCAL_PATH%'); $success = $true; } catch { if ($retryCount -ge 6) { throw; } else { $retryCount++; Start-Sleep -Seconds (5 * $retryCount); } } } while ($success -eq $false);"
+  if NOT exist "%VSWHERE_LOCAL_PATH%" (
+    echo ERROR: Could not install vswhere correctly.
+    exit /b 1
```
jdom: I wouldn't exit, I'd try to fallback to using the env variables. In fact, I would first check the variables, and if they aren't there, then you use vswhere. This is faster when the env variables are defined AND allows us to selectively pick the VS tooling we want to use if we don't want VSWHERE to randomly pick one of the installations.
So we don't *depend* on the variables being there, but if they are, we'll honor them

attilah: With VS2017 there are NO env vars set for you, it is because they support SxS installations, so the RECOMMENDED way is to use vswhere to select the one you need. Later if we require to select a specific version we'll be able to add filtering criteria on vswhere output, which is the recommended way. So it is not a best practice anymore to depend on env vars (finally).

attilah: As for the exit, I'll modify to be a fallback, but it a download fails, then...also we're only downloading if it is not there, so speed should not be an issue. We're doing the same for vNext already, but there we're downloading megs, not just 1 binary.

src/Build.cmd
```diff
@@ -2,6 +2,38 @@
...
+  )
+)
+
+set pre=Microsoft.VisualStudio.Product.
+set ids=%pre%Community %pre%Professional %pre%Enterprise %pre%BuildTools
+for /f "usebackq tokens=1* delims=: " %%i in (`%VSWHERE_LOCAL_PATH% -latest -products %ids% -requires Microsoft.Component.MSBuild`) do (
+  if /i "%%i"=="installationPath" set VS2017InstallDir=%%j
+)
+
+if NOT "%VS2017InstallDir%" == "" (
```
jdom: btw, you can use this syntax that is a little bit more readable:
` if defined VS2017InstallDir`

jdom:
Like the changes, but I tried it locally and seems to have messed up my VS2015 installation, even after fully removing it

jdom:
Scratch my last comment. The environment might have already been broken since the last time I installed the broken VSIX. I missed the last step in your blog post, but after doing that, then VS2015 worked again, and the VSIX works fine when I install it :)

jdom:

PR #2851, opened 2017-03-16
Use ILocalSiloDetails abstraction

jdom:
Minor cleanup to use the abstraction (interface) instead of the concrete object when it's not required

### DIFF
--- a/src/OrleansRuntime/Silo/Silo.cs
+++ b/src/OrleansRuntime/Silo/Silo.cs
@@ -222,7 +222,7 @@ namespace Orleans.Runtime
             var services = new ServiceCollection();
             services.AddSingleton(this);
             services.AddSingleton(initializationParams);
-            services.AddSingleton<ILocalSiloDetails>(initializationParams);
+            services.AddFromExisting<ILocalSiloDetails, SiloInitializationParameters>();
             services.AddSingleton(initializationParams.ClusterConfig);
             services.AddSingleton(initializationParams.GlobalConfig);
             services.AddTransient(sp => initializationParams.NodeConfig);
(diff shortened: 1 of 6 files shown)


src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -222,7 +222,7 @@ internal Silo(SiloInitializationParameters initializationParams)
             var services = new ServiceCollection();
             services.AddSingleton(this);
             services.AddSingleton(initializationParams);
```
ReubenBond: maybe we should remove this line and only have the one registration:
`services.AddSingleton<ILocalSiloDetails>(initializationParams);`

ReubenBond:
Generally LGTM - I agree that we should do this. I think we should remove the need to ever use the concrete instance. The biggest blocker was `SiloType`, but it seems as though that could be removed, given it's typically set via `NodeConfiguration` like so: `Type = NodeConfig.IsPrimaryNode ? Silo.SiloType.Primary : Silo.SiloType.Secondary;`

In other cases, maybe it makes some sense to add the configuration members to `ILocalSiloDetails`?

jdom:
This was some minor cleanup as I was trying to find how deep the usage of ClusterConfiguration is (for my SiloBuilder exploration). So I'd lean towards the opposite and use the configuration less, not more. I didn't remove the registration of the concrete class, as it's being injected in a few other places where they access more stuff than what's on the interface. Eventually I agree that we should remove it, especially with the SiloBuilder work, but it just wasn't my goal here

ReubenBond:
Yep, we can clean those pieces up later and break configuration up into smaller, more targeted pieces.

Issue #2845, opened 2017-03-14
little bug


edison0621:
In Orleans.EventSourcing.LogStorage
![the question](https://cloud.githubusercontent.com/assets/3294321/23894533/724271ce-08dd-11e7-8710-0e6f6c807a0c.png)

in addition, not only in this file

sergeybykov:
Resolved via #2853.

Thank you, @edison0621!

edison0621:
you're welcome

PR #2853, opened 2017-03-16
Fix typos in format strings

sergeybykov:
Fix for #2845.

### DIFF
--- a/src/OrleansEventSourcing/LogStorage/LogViewAdaptor.cs
+++ b/src/OrleansEventSourcing/LogStorage/LogViewAdaptor.cs
@@ -279,7 +279,7 @@ namespace Orleans.EventSourcing.LogStorage
             /// <inheritdoc/>
             public override string ToString()
             {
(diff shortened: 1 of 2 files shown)


src/OrleansEventSourcing/LogStorage/LogViewAdaptor.cs
```diff
@@ -279,7 +279,7 @@ protected class UpdateNotificationMessage : INotificationMessage
             /// <inheritdoc/>
             public override string ToString()
             {
-                return string.Format("v{0} ({1} updates by {2}) etag={2}", Version, Updates.Count, Origin, ETag);
+                return string.Format("v{0} ({1} updates by {2}) etag={3}", Version, Updates.Count, Origin, ETag);
```
jdom: I'm OK checking this in, nevertheless for string format calls (where you are eagerly formatting the string, which is not always the case when logging), it's better to replace them with interpolated strings: `$"v{Version} ({Updates.Count} updates by {Origin}) etag={ETag}"`

PR #2856, opened 2017-03-17
Add support for Scoped services

jdom:
Add support for scoped services per grain activation (a new container scope is created for each grain activation, and disposed when it's deactivated).

This is similar to how scoped services work in ASP.NET for every request, but in our case it's per grain activation. 
There's [a lot of documentation about lifetimes](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection#service-lifetimes-and-registration-options) already, but to sum up some of the benefits/changes in Orleans:
- We now support DI registrations with  `ServiceLifetime.Scoped`
- Services with `Scoped` lifetime act as singletons within that same scope, so everything that is resolved by the grain activation will refer to the same instance.
- Introduced the `IGrainActivationContext` scoped service, which can be injected in grain constructors, that allows access to grain activation-specific information, such as the identity, so that other scoped services or unit tests can access that information without the need for complex manipulations and mocking of the grain implementation itself.
- `Scoped` and `Transient` objects that are resolved by the scoped container are then automatically disposed (if they implement `IDisposable`) after the grain activation is deactivated.

This is just the minimum viable code to support a slew of scenarios. In the future though, I intend to add more stuff in the `IGrainActivationContext` itself, similar to the `RequestContext` in ASP.NET, to enable several cross-cutting concerns, such as adding a property bag `Items`, or access to the grain activation lifecycle so that other scoped services could initialize (asynchronously) once the grain is activated, access to the grain instance itself, etc.


jdom:
It's probably easier to review commit by commit.
If merging, please preserve the separate commits.
I did not run load tests yet to see if this extra allocation per activation (the scoped service provider) impacts throughput, but I would doubt that it would be considerable.

sergeybykov:
@dotnet-bot test this please.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -933,10 +933,10 @@ private void FastKill()
             UnobservedExceptionsHandlerClass.ResetUnobservedExceptionHandler();
 
             SafeExecute(() => this.SystemStatus = SystemStatus.Terminated);
-            SafeExecute(LogManager.Close);
             SafeExecute(() => AppDomain.CurrentDomain.UnhandledException -= this.DomainUnobservedExceptionHandler);
             SafeExecute(() => this.assemblyProcessor?.Dispose());
             SafeExecute(() => (this.Services as IDisposable)?.Dispose());
+            SafeExecute(LogManager.Close);
```
ReubenBond: Not that it changes in this PR, but I believe this hinders our ability to host multiple silos per AppDomain. Of course, the larger issue is the global `LogManager`, but perhaps we could just flush logs here and keep it open?

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -716,17 +720,18 @@ private void CreateGrainInstance(string grainTypeName, ActivationData data, stri
...
                 //for stateful grains, install storage bridge
                 if (grain is IStatefulGrain)
                 {
                     SetupStorageProvider(grainType, data);
 
                     var storage = new GrainStateStorageBridge(grainType.FullName, data.StorageProvider);
-
-                    grain = grainCreator.CreateGrainInstance(grainType, data.Identity, stateObjectType, storage);
+                    // TODO: avoid double creation!
+                    grain = grainCreator.CreateGrainInstance(data.ServiceProvider, grainType, data.Identity, stateObjectType, storage);
```
ReubenBond: Could we fix this TODO in this PR? it seems like `grain` created above is used only for the type check. Could we not replace that check with `typeof(IStatefulGrain).IsAssignableFrom(grainType)`?

src/OrleansRuntime/Silo/Silo.cs
jdom: good point, but do you think I should do this in this PR on in one that only deals with the statics? I'm fine either way

ReubenBond: No, it's not really relevant to this PR - just something for us to think about when we make those changes.

src/OrleansRuntime/Catalog/Catalog.cs
jdom: yes, I can certainly do that. I still wanted to have a perf run with the same allocation to compare apples to apples. But on further thought, I don't think that's worse it, nor it preserves the apples (as it uses the scoped SP to build the instance that will be discarded, and I don't know if it's as cheap as before, probably it is, but it's no reason to not fix it now).

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -1205,6 +1210,9 @@ private Task DestroyActivations(List<ActivationData> list)
                         directory.InvalidateCacheEntry(activationData.Address);
 
                         RerouteAllQueuedMessages(activationData, null, "Finished Destroy Activation");
+
+                        activationData.ServiceScope?.Dispose();
+                        activationData.ServiceScope = null;
```
ReubenBond: Would it be better to make `IActivationData` an `IDisposable` and keep this logic there?

jdom: I don't know for sure, but since we are creating the ServiceScope externally and then assigning it, then I think it's fine to also Dispose externally. Otherwise I think I'd have to change the logic so that the ActivationData itself creates the scope. Thoughts? is it worse it?

ReubenBond: That's fair. The grain/activationdata creation flow is a little convoluted. It's more complicated to have some things internal and some things external, but it's alright.

src/Orleans/Core/Grain.cs
```diff
@@ -39,7 +36,7 @@ protected IGrainFactory GrainFactory
         /// </summary>
         protected IServiceProvider ServiceProvider 
         {
-            get { return Runtime?.ServiceProvider; }
+            get { return Data.ServiceProvider ?? Runtime?.ServiceProvider; }
```
ReubenBond: This will only ever lead to errors, right? I believe we should only use `Data.ServiceProvider`. If that's `null` then we ought to allow the application to crash.

src/OrleansRuntime/Catalog/ActivationData.cs
```diff
@@ -321,6 +322,8 @@ public GrainId Identity
 
         public ActivationAddress Address { get; private set; }
 
+        public IServiceProvider ServiceProvider => this.ServiceScope?.ServiceProvider;
```
ReubenBond: As before, I think an NRE is better than a null.

src/Orleans/Core/Grain.cs
jdom: I kept it with a fallback similar to some other code I've seen when people created the grains with some factory for their unit tests. I could revert it though, as you are right that I don't know the implications. Still, it doesn't seem unsafe, does it? I'm not ever disposing this. only the scope if set.
In the future, I also envision a way to configure it so that users are not forced to use scoped containers if they don't require it. I could delay this kind of change until then if you prefer.

jdom: BTW, injecting the IServiceProvider should return this same instance (whether scopes are used or not), as if scopes are used, it would return the scoped SP, and if they aren't, it would return the main SP, that's why I thought the fallback seemed reasonable (even though it will never fallback until we provide the option to avoid scopes).

ReubenBond: > Still, it doesn't seem unsafe, does it?

I think it's safe, but it's difficult to verify that this will always be safe, which is why I suggest the failure option. Additionally, we should probably remove `ServiceProvider` from `IGrainRuntime` since it's no longer valid: IGrainRuntime is designed for grains, but its `ServiceProvider` has a different scope.

src/OrleansRuntime/Catalog/ActivationData.cs
jdom: similar to my other answer, I kept it this way so that you can still new up grains in your unit tests without using an actual Silo. Not sure if it's a good idea, but that PR was merged, so I assume people liked that option to not throw. I could make it throw if that's preferred

src/Orleans/Core/Grain.cs
ReubenBond: > until we provide the option to avoid scopes

I think we should avoid that option for two reasons:
1. Simplicity
2. Allows us to provide services (eg, facets) which are scoped to an activation

src/OrleansRuntime/Catalog/ActivationData.cs
ReubenBond: Which PR was that? Maybe we should provide a util for creating grains without a silo which respects scoping

src/Orleans/Core/Grain.cs
jdom: I don't think facets will work that way anyway, there will probably still be a singleton service that acts as a factory for the facets, and they will not be stored in the container by scoping it, otherwise you wouldn't be able to have 2 facets of the same type (but with different name) in the same grain.

BTW, MVC does allow a way to opt out of scoping per request...

src/OrleansRuntime/Catalog/ActivationData.cs
jdom: I believe it was https://github.com/dotnet/orleans/pull/2338 (and also https://github.com/dotnet/orleans/pull/1918 is related to this same philosophy).

Yup, a test kit could be useful but I would definitely not buy into that for this PR :)

src/Orleans/Core/Grain.cs
ReubenBond: Alright, in that case we can leave it. I generally prefer we opt for simplicity & correctness over tweakability. Grain Activations also live much longer than MVC Controllers tend to, so there's a difference there w.r.t cost of adding a scope (per-request vs (relatively) long-lived).

src/OrleansRuntime/Catalog/ActivationData.cs
ReubenBond: Right, not for this PR

jdom:
Created separate PRs for the commits not 100% related to scoped services (see #2875 and #2876)

jdom:
Upgraded to the latest DI package reference and I'm now using validateScopes=true, which the MVC team said it should have been the default, and might be the default in 2.0. See https://youtu.be/x-C-CNBVTaY?t=37m39s

jdom:
@dotnet-bot test this please

jdom:

src/Orleans/Core/Grain.cs
jdom: BTW, I missed the comment of IGrainRuntime having a service locator. I think you are right, we should have an IGrainRuntime scoped service or get rid of most of it, since it seems to be a facade for getting to other services, but in theory you can inject them via DI too. This interface is currently a little odd to as it you could in theory deactivate another grain with it, so it's dangerous to have it be public.

jdom:
Thanks for reviewing.
RE VSIX: yeah, although we don't update OrleansVSTools all the time. It still builds 1.4, so updating the reference there means it won't build neither 1.4 nor 1.5.

jdom:
@dotnet-bot test netfx-bvt

src/Orleans/Core/Grain.cs
jdom: The latest commits have `IGrainActivationContext` as a scoped service that grains can use to access other scoped services, and eventually bind to grain activation lifecycle events (these are not part of the activation context yet), so that there could be additional infrastructure attached to grains (ie: storage adapters, timers, schedulers, or other infrastructure components that require some lifetime coupling to the grain instance)

jdom:
The latest commits have `IGrainActivationContext` as a scoped service that grains can use to access other scoped services, and eventually bind to grain activation lifecycle events (these are not part of the activation context yet), so that there could be additional infrastructure attached to grains (ie: storage adapters, timers, schedulers, or other infrastructure components that require some lifetime coupling to the grain instance)

centur:
@jdom I checked the PR but really it's way above our current level of DI usage (and as we migrated from very early versions, the level is around ground zero ) so can't really make any meaningful comments. It looks alright for me as I can see there will be more interfaces we would be able to inject, including interfaces with narrowly scoped context (the first thought is AppInsights TelemetryClient where we can initialize client instance with some context that is common for a specific grain).

jdom:
The more I think about it, I believe that this should go in in 1.5.0 as well as adding a couple extra properties to `IGrainActivationContext` to make it A LOT more useful for end users. In particular add:
```
 /// <summary>Gets the instance of the grain associated with this activation context.</summary>
Grain GrainInstance { get; }
 /// <summary>Gets a key/value collection that can be used to share data within the scope of this grain activation.</summary>
IDictionary<object, object> Items { get; }
```

Thoughts?

src/NuGet/Microsoft.Orleans.Core.nuspec
```diff
@@ -21,7 +21,7 @@
       <dependency id="Newtonsoft.Json" version="9.0.1" />
       <dependency id="System.Collections.Immutable" version="1.3.1" />
       <dependency id="System.Reflection.Metadata" version="1.4.2" />
-      <dependency id="Microsoft.Extensions.DependencyInjection" version="1.1.0" />
+      <dependency id="Microsoft.Extensions.DependencyInjection" version="1.1.1" />
```
jason-bragg: I think if we're to move from 1.1.0 to 1.1.1 we should do that across the entire code base and in a separate pr.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -1259,6 +1259,8 @@ private Task DestroyActivations(List<ActivationData> list)
                                 grainCreator.Release(activationData, grainInstance);
                             }
                         }
+
+                        activationData.Dispose();
```
jason-bragg: Don't we want to dispose even if exceptions are thrown?

src/NuGet/Microsoft.Orleans.Core.nuspec
jdom: Well, the point is that there are no breaking changes, but the reason to do it here, is that in 1.1.1 they solved a race condition bug that ONLY happens if you are using scopes. So there's no value of updating *unless* we use scopes, and this PR without the update should not be merged *unless* we upgrade to 1.1.1 and avoid the race.

src/OrleansRuntime/Catalog/Catalog.cs
jdom: Sure, moved to a separate try block that includes both grain release (which should in theory NEVER throw otherwise the infrastructure has a nasty bug).
Disposing the objects in the scoped container shouldn't throw either, but because it's userland, I wouldn't just blindly trust that it will never throw :)

jdom:

jdom:
Reason: Unrelated failure. This is ready to be merged

PR #2857, opened 2017-03-17
Cherry-picked fixes from master to 1.4.1 + VSIX

sergeybykov:
Includes a number of fixes from master and VSIX redone to work properly in VS 2015 and 2017.


jdom:
What does #2833 bring? Just curious, as I thought there were no side-effects if that bug is there, and is only in preparation for the versioning story

jdom:
Please include #2859, as this won't work on build agents that do not have VS2017 installed. That PR fixes it, as it was broken by #2820.

benjaminpetit:
Yes Julian is right, #2833 was just in preparation for versioning

sergeybykov:
Removed #2833.
Added #2859.

PR #2804, opened 2017-03-02
Fix stress issues with a test

jdom:
* Avoids stressing the storage on the INSERT call by limiting parallelization on the first write.
* Avoids starving the thread pool while doing concurrent updates, but still force the store to be stressed a little bit.

Fixes #2790

Issue #2790, opened 2017-02-27
SqlServerStorageTests.WriteReadWriteRead100StatesInParallel test failing


jdom:
Getting the following error when running the functionals lately:
`WriteReadWriteRead100StatesInParallel [0:37.528] Failed: Orleans.Storage.InconsistentStateException: Version conflict (WriteState): ServiceId=00000000-0000-0000-0000-000000000000 ProviderName=System.Data.SqlClient_StorageProvider GrainType=UnitTests.StorageTests.Relational.TestDataSets.GrainTypeGenerator+TestGrainGenericWithGuidKey'1[[System.Double, mscorlib]] GrainId=33900 ETag=.`

@veikkoeeva / @shayhatsor do you know of anything that changed in the relational code that might be causing this?
Full stacktrace below:

```
Orleans.Storage.InconsistentStateException
Version conflict (WriteState): ServiceId=00000000-0000-0000-0000-000000000000 ProviderName=System.Data.SqlClient_StorageProvider GrainType=UnitTests.StorageTests.Relational.TestDataSets.GrainTypeGenerator+TestGrainGenericWithGuidKey`1[[System.Double, mscorlib]] GrainId=33900 ETag=.
   at Orleans.Storage.AdoNetStorageProvider.<WriteStateAsync>d__43.MoveNext() in D:\projects\orleans\src\OrleansSQLUtils\Storage\Provider\AdoNetStorageProvider.cs:line 449
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at UnitTests.StorageTests.Relational.CommonStorageTests.<Store_WriteRead>d__9`1.MoveNext() in D:\projects\orleans\test\TesterInternal\StorageTests\CommonStorageTests.cs:line 157
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
[...]

PR #2860, opened 2017-03-17
Avoid parallelization on INSERT

jdom:
This would hopefully fix the non-deterministic failure on Relational_WriteReadWriteRead100StatesInParallel once and for all.


jdom:
Related to #2804 and #2790

PR #2861, opened 2017-03-18
Skip Azure Storage tests if environment is not set

jdom:
Some tests require real Azure Storage to be configured (not the emulator), and they are skipped otherwise.


jdom:
Validated in our Functional CI run that these tests continue to run and pass when Azure environment is correctly configured. This should be ready to merge.

PR #2863, opened 2017-03-18
Small changes to the Minimal Orleans Application tutorial

miker1423:


src/Tutorials/Minimal-Orleans-Application.md
```diff
@@ -220,6 +221,21 @@ At this point we should be able to run Host. A console should appear with a larg
 Within the appropriate bin directory (Debug/Release) there will be a number of log files showing this information as well.  The amount and method of logging is configurable.
 
 
+## Troubleshooting
+- System.SystemException: 'Failed to start Orleans silo 'YOUR-HOST-NAME' as a Secondary node':
+    - Problem: The silo is trying to start but does not find a the specified file to get the configuration.
+    - Solution 1: Set the property "Copy to Output Directory" to "Copy if newer", to set it, right click on the configuration file, then click on properites.
+    - Solution 2: Add the folowing task to the csproj of the silo to copy your XML file to the output folder.
```
jdom: IMO solution 1 should be more than enough, as I can't see a scenario where that wouldn't work and would warrant the more complex workaround.

miker1423: Well I thought if someone was not using VS but that's true, that case is rare.

sergeybykov:
Thank you, @miker1423!

PR #2868, opened 2017-03-21
Fix binding redirects so that SkippableFact works in vNext

jdom:
I noticed that since we merged #2861, a lot of tests fail to be discovered in the vNext solution.
This was due to a missing binding redirect. This change should make SkippableFact test methods discoverable again.


ReubenBond:
@dotnet-bot test this please

Issue #2840, opened 2017-03-12
ReflectionTypeLoadException when loading my Grain, resolved with weird typeof


shlomiw7:
Hi,

I get the following exception:
```
[2017-03-12 18:36:24.434 GMT     1	WARNING	101723	CodeGenerator	127.0.0.1:11111]	AssemblyLoader encountered an exception loading types from assembly 'NG.Server.Game, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null': System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
   at System.Reflection.RuntimeModule.GetTypes(RuntimeModule module)
   at System.Reflection.RuntimeAssembly.get_DefinedTypes()
   at Orleans.Runtime.TypeUtils.GetDefinedTypes(Assembly assembly, Logger logger)	
Exc level 0: System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
   at System.Reflection.RuntimeModule.GetTypes(RuntimeModule module)
   at System.Reflection.RuntimeAssembly.get_DefinedTypes()
   at Orleans.Runtime.TypeUtils.GetDefinedTypes(Assembly assembly, Logger logger)
Exc level 1: System.TypeLoadException: Could not load type 'NG.Server.Game.Player.PlayerGrain' from assembly 'NG.Server.Game, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

And I can't access the PlayerGrain.

But if I add in the InitSilo method:
`var t = typeof(PlayerGrain);`

It's green again and I can access the PlayerGrain normally.
Do you have any idea why?

ReubenBond:
@shlomiw7 Here's what I believe is happening:
* The silo enumerates all types whenever an assembly is loaded.
* When a type is enumerated, its static initializer is executed, which requires that all types directly referenced by that type are loaded. Hence the assembly for each of those types must be loaded. This is done automatically.
* When an assembly is loaded under the `ReflectionOnly` context (as is used in some parts of silo initialization), it is not able to implicitly load other assemblies.
* The type enumeration therefore fails at this point.

I'll improve logging and determine if we can avoid these exceptions.

Does this exception stop the silo from starting and prevent you from accessing `PlayerGrain`?

shlomiw7:
[...]

PR #2869, opened 2017-03-21
ProviderTypeLoader: do not enumerate types in ReflectionOnly assembly.

ReubenBond:
Also improve logging for errors in `TypeUtils.GetDefinedTypes(...)` so that loader exceptions are made apparent.

Intention is to make debugging issues like #2840 more obvious

### DIFF
--- a/src/Orleans/Providers/ProviderTypeLoader.cs
+++ b/src/Orleans/Providers/ProviderTypeLoader.cs
@@ -87,6 +87,14 @@ namespace Orleans.Providers
 
         private static void ProcessNewAssembly(object sender, AssemblyLoadEventArgs args)
         {
+#if !NETSTANDARD
+            // If the assembly is loaded for reflection only avoid processing it.
+            if (args.LoadedAssembly.ReflectionOnly)
+            {
+                return;
+            }
+#endif
+
             // We do this under the lock to avoid race conditions when an assembly is added 
             // while a type manager is initializing.
             lock (managers)
--- a/src/Orleans/CodeGeneration/TypeUtils.cs
+++ b/src/Orleans/CodeGeneration/TypeUtils.cs
@@ -538,7 +538,8 @@ namespace Orleans.Runtime
             {
                 if (logger != null && logger.IsWarning)
                 {
-                    var message = $"AssemblyLoader encountered an exception loading types from assembly '{assembly.FullName}': {exception}";
+                    var message =
+                        $"Exception loading types from assembly '{assembly.FullName}': {LogFormatter.PrintException(exception)}.";
                     logger.Warn(ErrorCode.Loader_TypeLoadError_5, message, exception);
                 }
                 


src/Orleans/Providers/ProviderTypeLoader.cs
```diff
@@ -87,6 +87,14 @@ private void ProcessAssemblyLocally(Assembly assembly)
 
         private static void ProcessNewAssembly(object sender, AssemblyLoadEventArgs args)
         {
+#if !NETSTANDARD
+            // If the assembly is loaded for reflection only avoid processing it.
+            if (args.LoadedAssembly.ReflectionOnly)
```
jdom: The linked issue doesn't look like it was related to the ProviderTypeLoader. What led you to here? 
Also, I think that because this is wired to the `AppDomain.AssemblyLoad` event, it should not be getting called when a reflection-only assembly is loaded (I might be wrong, because there's too much black magic with this context, but I think that's what I remember).

ReubenBond: I thought I had replied to this, but apparently not.

This was just me enumerating usages of `GetDefinedTypes` to see where it was being potentially used incorrectly. `AssemblyLoad` is called when asms are loaded `ReflectionOnly` - I tested in a clean project.

jdom: OK, and I assume it still works correctly with custom providers that are loaded via reflection, right? We scan for IProvider in a separate step, but by the time it reaches here it should already be fully loaded?

ReubenBond: When a provider dll is loaded as non-ReflectionOnly, it will call this again with that asm.

Eventually, we need to overhaul AssemblyLoaded and avoid ReflectionOnly

jdom: I meant, we would only fully load the assembly once we first did a pass via reflection-only and found that there is at least one type that implements IProvider (unless I'm not remembering correctly). That step is separate from this, or is this the step that actually scans the reflection only assemblies and decides whether to fully load it?

ReubenBond: The code path which is called hits `RegisterProviderType` which would fail for a ReflectionOnly type.

jdom: Ok, sounds good. I checked and the code I was referring to happens before this is hit, so it's as expected. BTW, I was referring to: https://github.com/dotnet/orleans/blob/430a1de3c1eb1105a31133117ea4f06d2e1d01d2/src/Orleans/Runtime/OutsideRuntimeClient.cs#L224 and https://github.com/dotnet/orleans/blob/7cf2035e868e81b90767d7aff748405d01145117/src/OrleansRuntime/GrainTypeManager/SiloAssemblyLoader.cs#L73 (which first scans assemblies using reflection-only, and if it finds something that implements IProvider, it fully loads them)

PR #2870, opened 2017-03-21
Remove Silo.CurrentSilo global variable.

ReubenBond:


src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -98,7 +98,8 @@ internal class LocalGrainDirectory :
             ISiloStatusOracle siloStatusOracle,
             IMultiClusterOracle multiClusterOracle,
             IInternalGrainFactory grainFactory,
-            GrainTypeManager grainTypeManager)
+            Func<GrainDirectoryPartition> grainDirectoryPartitionFactory,
```
jdom: I'm a little on the fence with using Func<T> as factory methods all throughout, as they don't really disambiguate whether it gets the T, or creates it. In this particular case, you explicitly want something that creates it.
Subjectively, to me it feels that Func<T> just retrieves the T (you might get a different one if config changes for example, but doesn't create a new transient object every single time it's called).
Still, in a practical sense, it would be equivalent, and avoids having to define `Factory<T>`, `Factory<T1,T2>`, etc.
I believe @jason-bragg had a similar concern and started a thread about it not long ago.
Thoughts?

src/OrleansRuntime/GrainDirectory/MultiClusterRegistration/RegistrarManager.cs
```diff
@@ -11,32 +11,15 @@ namespace Orleans.Runtime.GrainDirectory
     internal class RegistrarManager
     {
         private readonly GrainTypeManager grainTypeManager;
+        private readonly object registrarLock = new object();
+        private readonly IServiceProvider serviceProvider;
 
-        private readonly Dictionary<Type, IGrainRegistrar> registrars = new Dictionary<Type, IGrainRegistrar>();
+        private Dictionary<Type, IGrainRegistrar> registrars = new Dictionary<Type, IGrainRegistrar>();
```
jdom: I'd use IReadOnlyDictionary as the field type, since the intent is to always clone, update and swap.

src/OrleansRuntime/Catalog/GrainCreator.cs
```diff
@@ -104,7 +106,7 @@ public Grain CreateGrainInstance(Type grainType, IGrainIdentity identity, Type s
             var logger = (factory as ILogConsistencyProvider)?.Log ?? storageProvider?.Log;
```
jdom: This logic was always very weird for me, but it gets even more weird by making the contract official with a long Func<T1,T2,T3> that both parties need to know of. What if you just remove the need for passing the logger, and use the SP to assign one?

src/OrleansRuntime/LogConsistency/ProtocolServices.cs
```diff
@@ -107,73 +121,65 @@ internal ProtocolServices(Grain gr, Logger log, IMultiClusterRegistrationStrateg
...
         public string MyClusterId
         {
             get
             {
-                return Silo.CurrentSilo.ClusterId;
+                return this.globalConfig.ClusterId;
             }
         }
 
+        private string ClusterDisplayName => this.ClusterDisplayName;
```
jdom: Doesn't this cause a stack overflow?

src/OrleansRuntime/LogConsistency/ProtocolServices.cs
```diff
@@ -21,45 +19,61 @@ namespace Orleans.Runtime.LogConsistency
...
-        private readonly IInternalGrainFactory grainFactory;
-
-        public IMultiClusterRegistrationStrategy RegistrationStrategy { get; private set; }
-
-        private Grain grain;   // links to the grain that owns this service object
+        private static readonly object[] EmptyObjectArray = new object[0];
 
         // pseudo-configuration to use if there is no actual multicluster network
-        private static MultiClusterConfiguration PseudoMultiClusterConfiguration;
+        private static readonly Interner<string, MultiClusterConfiguration> PseudoMultiClusterConfigurations = new Interner<string, MultiClusterConfiguration>();
```
jdom: why do you need an interner? if used, the PseudoMultiClusterConfiguration will be a single one... just a config with the current clusterId... this is not multi-valued. Or am I misunderstanding its use?

src/OrleansRuntime/Messaging/GatewayAcceptor.cs
```diff
@@ -10,18 +11,21 @@ internal class GatewayAcceptor : IncomingMessageAcceptor
...
+        private readonly GlobalConfiguration globalConfig;
 
         internal GatewayAcceptor(
             MessageCenter msgCtr,
             Gateway gateway,
             IPEndPoint gatewayAddress,
             MessageFactory messageFactory,
-            SerializationManager serializationManager)
+            SerializationManager serializationManager,
+            GlobalConfiguration globalConfig)
```
jdom: in theory ClusterId and HasMultiClusterNetwork properties could be moved to ILocalSiloDetails, so no need to reference the uber global config (which we want to get rid of eventually with the SiloBuilder)

src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
ReubenBond: I'd be fine with defining Factory<T>, etc.

src/OrleansRuntime/Catalog/GrainCreator.cs
ReubenBond: I can use SP to pass in a logger factory to ProtocolServices instead

src/OrleansRuntime/LogConsistency/ProtocolServices.cs
ReubenBond: This class is created per-grain and the config needs to be per-clusterId, so I added the interner for correctness in cases where more than one cluster is hosted in a process (mostly testing).

ReubenBond: The interner is mostly used as a weak dictionary

src/OrleansRuntime/LogConsistency/ProtocolServices.cs
ReubenBond: yes, thanks

src/OrleansRuntime/Messaging/GatewayAcceptor.cs
ReubenBond: I was thinking we should probably move to some kind of "IMultiClusterConfig" eventually - and likewise for other aspects of the configuration

src/OrleansRuntime/LogConsistency/ProtocolServices.cs
ReubenBond: I can remove this and we can allocate a new config per-grain - I don't mind

jdom: ah, good point. It's fine then

src/Orleans/Configuration/GlobalConfiguration.cs
```diff
@@ -217,21 +219,35 @@ public bool PrimaryNodeIsRequired
...
+            {
+                var configuredId = this.HasMultiClusterNetwork ? this.clusterId : this.DeploymentId;
+                return string.IsNullOrEmpty(configuredId) ? DefaultClusterId : configuredId;
+            }
+
+            set
+            {
+                this.clusterId = value;
+            }
+        }
```
ReubenBond: This logic needed to go somewhere. It previously lived in `Silo`, which is a bad place for it (i.e, consumers were accessing `Silo.CurrentSilo` to get to it).

I could put it somewhere else, like in an extension method or some new object

jdom: I guess it's fine, when I first added the comment I thought this was just new logic, didn't know it was moved from somewhere else. I agree we should eventually move it somewhere else, since the config is no longer some config holder, but it's fine for now, since we'll revamp the whole config object anyway.

src/OrleansRuntime/LogConsistency/ProtocolServices.cs
```diff
@@ -107,73 +121,65 @@ internal ProtocolServices(Grain gr, Logger log, IMultiClusterRegistrationStrateg
...
         public string MyClusterId
         {
             get
             {
-                return Silo.CurrentSilo.ClusterId;
+                return this.globalConfig.ClusterId;
             }
         }
 
+        private string ClusterDisplayName => this.pseudoMultiClusterConfiguration == null ? string.Empty : " " + this.MyClusterId;
```
jdom: I think it's fine, although in many places you replaced `this.pseudoMultiClusterConfiguration == null` with `!this.MulticlusterEnabled`

ReubenBond: I'll change it, it's cleaner as `this.MultiClusterEnabled ? " " + this.MyClusterId : string.Empty`

ReubenBond:
@dotnet-bot test this please

jdom:
@dotnet-bot test this please

ReubenBond:
@jdom I get the feeling that the build hosts are overloaded right now, which is why we are seeing failures.

ReubenBond:
@dotnet-bot test this please

src/OrleansRuntime/Scheduler/WorkerPoolThread.cs
```diff
@@ -328,7 +330,7 @@ internal void CheckForLongTurns()
             // Note that we only do this if the current load is reasonably low and the current thread
             // count is reasonably small.
             if (!pool.ShouldInjectWorkerThread ||
-                (Silo.CurrentSilo == null || !(Silo.CurrentSilo.Metrics.CpuUsage < MAX_CPU_USAGE_TO_REPLACE))) return;
+                !(this.performanceMetrics.CpuUsage < MAX_CPU_USAGE_TO_REPLACE)) return;
```
jdom: should this take into account that performanceMetrics can be null? (since I saw some path that you explicitly new up with null, although not sure whether it will hit this line in that path).

jdom:
I'm ready to merge, can you resolve conflicts? (a merge commit is fine, since I'm going to squash anyway).
Ping @jason-bragg / @attilah for last call review before the merge, since they expressed interest in the `Factory<T>` approach.

src/Orleans/Utils/Factory.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace Orleans
+{
+    /// <summary>
+    /// Creates an instance of <typeparamref name="TInstance"/>.
+    /// </summary>
+    /// <typeparam name="TInstance"></typeparam>
+    /// <returns></returns>
+    public delegate TInstance Factory<out TInstance>();
```
attilah: Generally speaking for parameterized factories it makes sense, but the this Factory registration is nothing different from a Transient registration, since DI will always return a new instance and accepts no parameters.

What I am missing from this factory delegate pattern is that I can't see why we don't include the ```ServiceProvider``` as a default first parameter in all the delegates to make it more friendly with DI containers. This way none of the factory functions have a way to use ```ActivatorUtilities.CreateInstance``` to pass in a ```IServiceProvider``` instance.

Since DI Abstractions already having the ```ObjectFactory``` delegate via the ```ActivatorUtilities.CreateFactory``` method, it can be a nice addition to it, hence we should call the same ```ObjectFactory``` instead of ```Factory``` ?

jdom: ObjectFactory takes an `IServiceProvider`. Factory<T> is a curried and typed version of it. Check out FactoryUtilities to see that it's just that.
Basically makes the usage of the factory explicit at the consuming level. You just tell it you want to have a factory of an object, you don't care whether it uses SP, ActivatorUtilities, or the new keyword under the hood, you just care to be explicit that you want something to construct new instances of T on demand.

attilah: After @jdom pointed to me to FactoryUtilities that is exactly what I was looking for (and missed somehow). Renaming to ```ObjectFactory``` can help if we submit this as a PR to the DI Abstraction repo, but it is not a big deal if we leave it as ```Factory```.

jdom: yeah, naming-wise I don't mind, at some point I was going to also suggest calling it `ObjectFactory<T>`, but since it's a different shape altogether, I prefer not to be confused with it. Still, I wouldn't mind a different name either if we want to have something less ambiguous than `Factory<T>`, but I'd vote no for `ObjectFactory<T>` because of that reason

jason-bragg: I don't have a _strong_ opinion on this, but my preference would be to request a factory interface from DI rather than a delegate.  My reasons for this preference are:

- Factory classes can support a wider set of behaviors, including statistics tracking and state-full factories like object pools.
- Factory implementations can be constructed like any other DI type so they can have access to the container, configuration, and other application or framework classes.
- Factory interfaces can be decorated to stack application layer pooling, monitoring or other behaviors, making them very customizable by environment/scenario (test, load, production)
- Existing systems can be easily hidden behind factory interfaces, making integration easier.

Something like:

    public interface IFactory<out TType>
        where TType : class
    {
        TType Create();
    }

jdom: The only issue with the suggested example, is that it doesn't support passing parameters for creating, as we need in this PR. Sure, we could in theory have `IFactory<out TInstance, in TParam1, in TParam2>` "overloads" so that the `Create` can take parameters, but at that point I believe we are just using types to emulate its functional counterpart. 
The functional approach also covers each of the items you listed. In fact if you look at `FactoryUtilities` many of those items are currently implemented, and the constructed object is being built using the container to fulfill the dependencies, but you could also do composition even easier than how you do it with the decorator pattern.
Won't push back if the class approach that's the consensus, but my preference goes to the functional approach in that case, since it's easy to map `Func<T>` to `Factory<T>` with less ceremony, and one is implicitly castable to the other if the end-user cares about that.
BTW, having a single interface and `Create` take `params object[]` is out of the question, as it doesn't make the requirements explicit.

jason-bragg: >"overloads"

Understood, example factory was just intended to communicate the idea, factories with args are possible.

> The functional approach also covers each of the items you listed.

In an absolute sense, yes, but not, imo, in an OO intuitive way.  By passing factories as function, we're exposing this capability in a functional pattern where state is not expected.  That is, it is not expected, or arguably correct, for the factories to have state.  By having the factories exposed clearly as objects themselves, it is understood that they are allowed to be statefull.

ReubenBond: >- Factory classes can support a wider set of behaviors, including statistics tracking and state-full factories like object pools.
>- Factory implementations can be constructed like any other DI type so they can have access to the container, configuration, and other application or framework classes.
>- Factory interfaces can be decorated to stack application layer pooling, monitoring or other behaviors, making them very customizable by environment/scenario (test, load, production)
>- Existing systems can be easily hidden behind factory interfaces, making integration easier.

These points are all supported using delegates - or am I still not fully awake?

So can delegates - that's what happens here. We are registering a `Func<IServiceProvider, Factory<...>>` with DI. That registration can be transient, scoped, or singleton. You can do whatever you want there

Delegates can be decorated, too, by combinators.

jason-bragg: @ReubenBond,  Understood, my original points were poorly stated as they implied that delegates could not do these things when they can, it's just less intuitive that they can or should.  As stated, I don't have a _strong_ opinion on this.  If the use of delegates is the common solution to this problem in most dependency injection systems (making the pattern familiar to most DI users), and our delegates do not conflict (name wise) with other similar patterns, I've no objections to this pattern.  Alternate preferences, but not objections :)

ReubenBond: @jason-bragg the use of delegates is the de-facto standard: http://docs.autofac.org/en/latest/resolve/relationships.html

ReubenBond: If no one is adamant that we should use a different approach, then I'm content with merging as-is. I'm happy to change it if that's how the wind is blowing, though.

PR #2865, opened 2017-03-20
Handle absolute path in IntermediateOutputPath to address issue 2864

andhesky:

Issue #2864, opened 2017-03-20
Microsoft.Orleans.OrleansCodeGenerator.Build.targets breaks if IntermediateOutputPath is an absolute path


andhesky:
IntermediateOutputPath is a relative path by default and the targets file assumes this is always true when it sets the output path of the codegen to "$(ProjectDir)$(IntermediateOutputPath)$(TargetName).codegen.cs". If IntermediateOutputPath is a full path, then the logic breaks. and the output path for codegen is set to an invalid path, for example "C:\enlistment\project\c:\enlistment\project\intermediate\target.codegen.cs"

sergeybykov:
Resolved via #2865 and #2871.

PR #2871, opened 2017-03-21
Handle absolute path in IntermediateOutputPath to address issue #2864 in vNext

jdom:
This is the same change as in #2865, but in the vNext tree.


jdom:

PR #2872, opened 2017-03-21
Rename codegen file to be excluded from analyzers

jdom:
StyleCop and potentially other analyzers will skip files that are contain .g. or .generated. in the name.


jdom:
I included the commit from #2871, otherwise it would conflict with this one. We should probably merge that one first

ReubenBond:
The only other place I could see a change might be relevant is to remove the superfluous orleans.codegen.cs line in the .gitignore, which is harmless.

jdom:
Hehe, yeah, I actually had that and force pushed to remove it, since the samples are based on older versions of Orleans, so the .gitignore still apply. Sure, I could create a new .gitignore at the Samples folder level if you think it's worth it.

PR #2873, opened 2017-03-22
add SlowConsumingPressureMonitor 

xiazen:
mostly finished coding, still need to test  it. submit for early feedback


xiazen:
Discussed with Jason offline, decided to take a simpler but less absolute approach. will shelve this commit. But still update the new approach to the same PR.

xiazen:
switched to a simpler approach, added one programmatic option to configure EventHubQueueCache with the new monitor. 

another option can be through EventHubSettings

test/ServiceBus.Tests/TestStreamProviders/EHStreamProviderWithSlowConsumingPressureDetecting.cs
```diff
@@ -0,0 +1,33 @@
...
+        public class AdapterFactory : EventHubAdapterFactory
+        {
+            public AdapterFactory()
+            {
+                CacheFactory = CreateQueueCache;
+            }
+
+            private IEventHubQueueCache CreateQueueCache(string partition, IStreamQueueCheckpointer<string> checkpointer, Logger log)
+            {
+                var bufferPool = new FixedSizeObjectPool<FixedSizeBuffer>(adapterSettings.CacheSizeMb, () => new FixedSizeBuffer(1 << 20));
```
ReubenBond: Just to make it blatantly obvious, maybe have a const for this value called `Megabyte` or `BlockSize` with a comment saying it's 1MB

xiazen: good idea. will do it !

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubQueueCache.cs
```diff
@@ -162,7 +228,7 @@ internal class AveragingCachePressureMonitor
         public AveragingCachePressureMonitor(double flowControlThreshold, Logger logger)
         {
             this.flowControlThreshold = flowControlThreshold;
-            this.logger = logger.GetSubLogger("flowcontrol", "-");
+            this.logger = logger.GetSubLogger("flowcontrol-averaging-cache-pressure", "-");
```
jason-bragg: Suggestion:
As we're moving more towards an asp.net logging model, I suggest using the typename here rather than custom name.
Also, delilimiter argument is optional, and "-" is default, so it should not be needed.
so:
this.logger = logger.GetSubLogger(GetType().Name);

xiazen: will do

src/OrleansServiceBus/Providers/Streams/EventHub/ICachePressureMonitor.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace OrleansServiceBus.Providers.Streams.EventHub
+{
+    public interface ICachePressureMonitor
+    {
+        void RecordCachePressureContribution(double cachePressureContribution);
+
+        bool IsUnderPressure(DateTime utcNow);
+    }
+
+    internal class AggregatedCachePressureMonitor : List<ICachePressureMonitor>, ICachePressureMonitor
```
jason-bragg: Suggest AggregatedCachePressureMonitor be public and placed in it's own file.

test/ServiceBus.Tests/TestStreamProviders/EHStreamProviderWithSlowConsumingPressureDetecting.cs
```diff
@@ -0,0 +1,34 @@
...
+using Orleans.Streams;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ServiceBus.Tests.TestStreamProviders
+{
+    public class EHStreamProviderWithSlowConsumingPressureDetecting : PersistentStreamProvider<EHStreamProviderWithSlowConsumingPressureDetecting.AdapterFactory>
```
jason-bragg: Suggest test run against default stream provider, configured to use SlowConsumingPressureMonitor, rather than custom adapter.  That is the preferred user scenario, so the one we should ensure works and is maintained going forward.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -140,7 +140,20 @@ public virtual void Init(IProviderConfiguration providerCfg, string providerName
             {
                 var bufferPool = new FixedSizeObjectPool<FixedSizeBuffer>(adapterSettings.CacheSizeMb, () => new FixedSizeBuffer(1 << 20));
                 var timePurge = new TimePurgePredicate(adapterSettings.DataMinTimeInCache, adapterSettings.DataMaxAgeInCache);
-                CacheFactory = (partition,checkpointer,cacheLogger) => new EventHubQueueCache(checkpointer, bufferPool, timePurge, cacheLogger, this.SerializationManager);
+                if (adapterSettings.SlowConsumingMonitorThreshold > 0)
```
jason-bragg: We tend to use nullable values for optional settings.  Suggest SlowConsumingMonitorThreshold be nullable rather than using it's value to detect if it's set or not.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -140,7 +140,20 @@ public virtual void Init(IProviderConfiguration providerCfg, string providerName
...
+                {
+                    CacheFactory = (partition, checkpointer, cacheLogger) =>
+                    {   var cache = new EventHubQueueCache(checkpointer, bufferPool, timePurge, cacheLogger, this.SerializationManager);
```
jason-bragg: suggest minimizing code in conditional statement.  Only thing that depend on the condition is if the SlowConsumingPressureMonitor is added. so

    var cache = new .....
    if(..)
    {
        add monitor
    }
    return cache;

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubQueueCache.cs
```diff
@@ -148,7 +159,62 @@ public bool TryGetNextMessage(object cursorObj, out IBatchContainer message)
 
     }
 
-    internal class AveragingCachePressureMonitor
+    public class SlowConsumingPressureMonitor : ICachePressureMonitor
```
jason-bragg: Suggest AveragingCachePressureMonitor be broke into it's own file, and added in the default cache creation logic, like other ICachePressureMonitors.

xiazen:
all good suggestions. will address them

test/ServiceBus.Tests/SlowConsumingTests/EHSlowConsumingTests.cs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+
+        public EHSlowConsumingTests(Fixture fixture)
+        {
+            this.fixture = fixture;
+        }
+
+        [Fact, TestCategory("EventHub"), TestCategory("Streaming"), TestCategory("Functional")]
+        public async Task EHSlowConsuming_ShouldFavorSlowConsumer()
```
xiazen: @jdom I think I need to skip this test if azure is not set up correctly in Jenkins? 

I got this when running in Jenkins:
MESSAGE:
System.AggregateException : One or more errors occurred.\r\n---- System.TypeInitializationException : The type initializer for 'ServiceBus.Tests.SlowConsumingTests.EHSlowConsumingTests' threw an exception.\r\n-------- System.ArgumentNullException : Value cannot be null.\r\nParameter name: dataConnectionString\r\n---- The following constructor parameters did not have matching fixture data: Fixture fixture
+++++++++++++++++++
STACK TRACE:
----- Inner Stack Trace #1 (System.TypeInitializationException) ----- at ServiceBus.Tests.SlowConsumingTests.EHSlowConsumingTests.Fixture.CreateTestCluster() in D:\j\workspace\netfx-functio---224c30ba\test\ServiceBus.Tests\SlowConsumingTests\EHSlowConsumingTests.cs:line 57 at TestExtensions.BaseTestClusterFixture..ctor() in D:\j\workspace\netfx-functio---224c30ba\test\TestExtensions\BaseClusterFixture.cs:line 30 ----- Inner Stack Trace ----- at Orleans.ServiceBus.Providers.EventHubCheckpointerSettings..ctor(String dataConnectionString, String table, String checkpointNamespace, Nullable`1 persistInterval) in D:\j\workspace\netfx-functio---224c30ba\src\OrleansServiceBus\Providers\Streams\EventHub\ICheckpointerSettings.cs:line 65 at ServiceBus.Tests.SlowConsumingTests.EHSlowConsumingTests..cctor() in D:\j\workspace\netfx-functio---224c30ba\test\ServiceBus.Tests\SlowConsumingTests\EHSlowConsumingTests.cs:line 45 ----- Inner Stack Trace #2 (Xunit.Sdk.TestClassException) -----

This test pass when I have connection string set up locally

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/AggregatedCachePressureMonitor.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <summary>
+        /// If any mornitor in this aggregated cache monitor group is under pressure, then return true
+        /// </summary>
+        /// <param name="utcNow"></param>
+        /// <returns></returns>
+        public bool IsUnderPressure(DateTime utcNow)
+        {
+            bool isUnderPressure = false;
+            
+            this.ForEach(monitor =>
```
jason-bragg: nit:  Link statement would be much cleaner.

return this.Any(monitor => monitor.IsUnderPressure(utcNow);

xiazen: will do

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -143,7 +143,27 @@ public virtual void Init(IProviderConfiguration providerCfg, string providerName
             {
                 var bufferPool = new FixedSizeObjectPool<FixedSizeBuffer>(adapterSettings.CacheSizeMb, () => new FixedSizeBuffer(1 << 20));
                 var timePurge = new TimePurgePredicate(adapterSettings.DataMinTimeInCache, adapterSettings.DataMaxAgeInCache);
-                CacheFactory = (partition,checkpointer,cacheLogger) => new EventHubQueueCache(checkpointer, bufferPool, timePurge, cacheLogger, this.SerializationManager);
+                CacheFactory = (partition, checkpointer, cacheLogger) =>
+                {
+                    var cache = new EventHubQueueCache(checkpointer, bufferPool, timePurge, cacheLogger, this.SerializationManager);
```
jason-bragg: nit:  We generally try to keep lambdas small, when they get large, we tend to put them into their own call.  This may be better as a CreateCacheFactory(..) function.

test/ServiceBus.Tests/SlowConsumingTests/EHSlowConsumingTests.cs
jason-bragg: It's likely the Eventhub connection string.  We don't run eh tests in Jenkins yet (to my knowledge).

test/ServiceBus.Tests/SlowConsumingTests/EHSlowConsumingTests.cs
```diff
@@ -0,0 +1,178 @@
...
+using System.Threading.Tasks;
+using TestExtensions;
+using UnitTests.GrainInterfaces;
+using UnitTests.Grains.ProgrammaticSubscribe;
+using Xunit;
+
+namespace ServiceBus.Tests.SlowConsumingTests
+{
+    [Collection(TestEnvironmentFixture.DefaultCollection)]
+    public class EHSlowConsumingTests : OrleansTestingBase, IClassFixture<EHSlowConsumingTests.Fixture>
```
jason-bragg: For targeted tests of the persistent stream provider infrastructure (kinda like this), we tend to use the generator stream provider, as it allows better control of the data flow as well as eliminates external dependencies like EventHub.  We can test this using EH, I'm just suggesting that the stream generator may be better in the long run.

jdom:
@xiazen this failure is a real one (not timing related, which we are fixing): https://ci.dot.net/job/dotnet_orleans/job/master/job/netfx-functional_prtest/10/testReport/junit/ServiceBus.Tests.SlowConsumingTests/EHSlowConsumingTests/EHSlowConsuming_ShouldFavorSlowConsumer/
Did you miss a SkippableFact or something of the sort?

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/AveragingCachePressureMonitor.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <summary>
+    /// Cache pressure monitor whose back pressure algorithm is based on averaging pressure value
+    /// over all pressure contribution
+    /// </summary>
+    public class AveragingCachePressureMonitor : ICachePressureMonitor
+    {
+        /// <summary>
+        /// Default flow control threshold
+        /// </summary>
+        public static readonly double DefaultThreashold = 1 / 3;
```
jdom: typo in *threshold*

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubStreamProviderSettings.cs
```diff
@@ -17,6 +17,37 @@ public class EventHubStreamProviderSettings
         public string StreamProviderName { get; }
 
         /// <summary>
+        /// SlowConsumingMonitorFlowControlThresholdName
+        /// </summary>
+        public const string SlowConsumingMonitorFlowControlThresholdName = "SlowConsumingMonitorFlowControlThreshold";
```
jdom: nameof(SlowConsumingMonitorFlowControlThreshold)
Avoid hardcoding strings in several places in this class. We were probably doing this before we upgraded to c# 6, but there's no reason to do more hardcoding now

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/SlowConsumingPressureMonitor.cs
```diff
@@ -0,0 +1,104 @@
...
+        public SlowConsumingPressureMonitor(Logger logger)
+            : this(DefaultFlowControlThreshold, DefaultCheckPeriod, logger)
+        { }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="checkPeriod"></param>
+        /// <param name="logger"></param>
+        public SlowConsumingPressureMonitor(TimeSpan checkPeriod, Logger logger)
```
jdom: checkPeriod is a confusing name. It took me some time by reading and re-reading the code to understand what it does, because it's not the time period between checks.
Maybe it should be pressure window size? Or something else not related to check time?

test/ServiceBus.Tests/Streaming/EHSlowConsumingTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using TestExtensions;
+using Xunit;
+
+namespace ServiceBus.Tests.Streaming
+{
+    [TestCategory("EventHub"), TestCategory("Streaming")]
+    [Collection(TestEnvironmentFixture.DefaultCollection)]
+    class EHSlowConsumingTests : OrleansTestingBase, IClassFixture<EHSlowConsumingTests.Fixture>
```
jdom: is this class dead code? I don't see any tests here, and it's named the same as the other tests

test/ServiceBus.Tests/SlowConsumingTests/EHSlowConsumingTests.cs
```diff
@@ -0,0 +1,178 @@
...
+                EHConsumerGroup, EHPath));
+
+        private static readonly EventHubCheckpointerSettings CheckpointerSettings =
+            new EventHubCheckpointerSettings(TestDefaultConfiguration.DataConnectionString,
+                EHCheckpointTable, CheckpointNamespace, TimeSpan.FromSeconds(1));
+
+
+        private readonly Fixture fixture;
+
+        public class Fixture : BaseTestClusterFixture
```
jdom: You should override CheckPreconditionsOrThrow and throw SkipException if the connection string is null.
Then in the test class' constructor you must call fixture.EnsurePreconditionsMet()
Also mark all of the tests with SkippableFact.

See usages of EnsurePreconditionsMet if you want some examples.

test/ServiceBus.Tests/SlowConsumingTests/EHSlowConsumingTests.cs
jdom: Correct, although we should start running EH tests in functionals in VSO, but skip the tests when the connection string is not set. I added some info on how to do it

xiazen: yeah I agree . We should set up the EH related config in VSO and start to run EH tests. 

Although this test do run successfully in VSO. so I think some EH related config is set up in VSO. But I'm not sure how much of the EH related config is set up in VSO. where can I check?

test/ServiceBus.Tests/Streaming/EHSlowConsumingTests.cs
xiazen: Wow I didn't notice this ! there's another EHSlowConsumingTests.cs  on the top... That's the real one. I will delete this one. Thanks for noticing

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/SlowConsumingPressureMonitor.cs
xiazen: I will think of a better name. PressureWindowPeriod? size is too general for me. let me think

test/ServiceBus.Tests/SlowConsumingTests/EHSlowConsumingTests.cs
jdom: yes, configuration is already set up in VSO, it's just that currently none of the EH tests are marked as Functional, so they just don't run because of that. I'd start adding these tests using the Skippable approach, and then we can look at adding more of the other EH tests using the same approach

xiazen: will do. I can add skippable to this test. A lot of EH tests run successfully locally. I think we can star add them to VSO and skip them in Jenkins.

test/ServiceBus.Tests/SlowConsumingTests/EHSlowConsumingTests.cs
xiazen: The test is also testing if the monitor config can be set correctly through EventHubProviderSettings class, so I'd prefer use real EventHubStreamProvider

jdom:
LGTM

xiazen:
@jdom the Dispose() method was called. That's why I added that Boolean check. maybe there's a bug somewhere or I used OrleansTestBase in an incorrectly way?

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/AveragingCachePressureMonitor.cs
```diff
@@ -0,0 +1,93 @@
...
+                return;
+            }
+
+            double pressure = accumulatedCachePressure / cachePressureContributionCount;
+            bool wasUnderPressure = isUnderPressure;
+            isUnderPressure = pressure > flowControlThreshold;
+            // If we changed state, log
+            if (isUnderPressure != wasUnderPressure)
+            {
+                logger.Info(isUnderPressure
```
jason-bragg: I think the info level version of this log message should be in the aggregate pressure monitor, and the logging in the specific monitors should be verbose.
The reason for this is that a specific monitor may become healthy, but reading will still be off until all monitors are healthy.  So having per monitor logging will make it very hard for users to decipher, from logs, if a queue should be reading or not.  These details may be useful in verbose logging, as they would indicate what monitor is stopping reading, but this level of detail is less important, imo, than the overall flow control state.

xiazen: will do . valid concern.

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/SlowConsumingPressureMonitor.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        public bool IsUnderPressure(DateTime utcNow)
+        {
+            //if any pressure contribution in current period is bigger than flowControlThreshold
+            //we see the cache is under pressure
+            bool underPressure = this.biggestPressureInCurrentPeriod > this.FlowControlThreshold;
+            if (this.isUnderPressure != underPressure)
+            {
+                this.isUnderPressure = underPressure;
+                logger.Verbose(this.isUnderPressure
```
jdom: how about logging as Info when under pressure, and as Verbose when not? Otherwise the aggregating monitor might not log the necessary information, right?

jason-bragg: It logs only when there is a state change.
under pressure => not under pressure or not under pressure => under pressure.
In both cases, this is not necessarily authoritative of the behavior, so users shouldn't care about this under normal logging conditions.

xiazen: ouch. I submitted a commit to address this. but under a second thought, I think Jason might be right.

jdom: Sure, on a state change makes more sense. But still it would be nice to know the details for when an individual monitor starts throttling. Maybe in the individual monitor, log as Info only when it transitions to under pressure when it wasn't (but not necessarily log as Info when it goes back to not under pressure, at that point the aggregated monitor logger can be enough).

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/SlowConsumingPressureMonitor.cs
```diff
@@ -0,0 +1,111 @@
...
+                //at the end of each check period, reset biggestPressureInCurrentPeriod
+                this.nextCheckedTime = utcNow + this.PressureWindowSize;
+                this.biggestPressureInCurrentWindow = 0;
+                //if at the end of the window, pressure clears out, log
+                if(this.wasUnderPressure && !underPressure)
+                    logger.Verbose($"Message ingestion is healthy. BiggestPressureInCurrentPeriod: {biggestPressureInCurrentWindow}, Threshold: {FlowControlThreshold}");
+                this.wasUnderPressure = underPressure;
+            }
+
+            return this.wasUnderPressure;
```
xiazen: in this logic, monitor will stay underPressure for a whole window if a slow consumer appears, while biggestPressureInCurrentWindow will be reset to 0. So at the end of this window, if still underPressure, then underPressure stays for another whole window. Otherwise, monitor become notUnderPressure. 

run through real life case,
monitor will be notUnderPressure at the beginning-> a slow consumer appears ->monitor become underPressure, and stay at underPressure for a whole window -> window ends, if slow consumer cleared out in this window, monitor become notUnderPressure. 

welcome to run through more scenario with me, and let me know if this algorithm is conservative enough

xiazen: or if you want something even more conservative, such as :
 
```
public bool IsUnderPressure(DateTime utcNow)
        {
            //if any pressure contribution in current period is bigger than flowControlThreshold
            //we see the cache is under pressure
            bool underPressure = this.biggestPressureInCurrentPeriod > this.FlowControlThreshold;

            if (underPressure)
            {
                //if under pressure, extend the nextCheckedTime, make sure wasUnderPressure is true for a whole window  
                this.wasUnderPressure = underPressure;
                this.nextCheckedTime = utcNow + this.PressureWindowSize;
                logger.Verbose($"Ingesting messages too fast. Throttling message reading. BiggestPressureInCurrentPeriod: {biggestPressureInCurrentPeriod}, Threshold: {FlowControlThreshold}");
                this.biggestPressureInCurrentPeriod = 0;
            }
            else
            {
                if (this.nextCheckedTime < utcNow)
                {
                    //at the end of each check period, reset biggestPressureInCurrentPeriod
                    this.nextCheckedTime = utcNow + this.PressureWindowSize;
                    this.biggestPressureInCurrentPeriod = 0;
                    this.wasUnderPressure = underPressure;
                    if(!this.wasUnderPressure)
                        logger.Verbose($"Message ingestion is healthy. BiggestPressureInCurrentPeriod: {biggestPressureInCurrentPeriod}, Threshold: {FlowControlThreshold}");
                }
            }

            return this.wasUnderPressure;
        }
```

isUnderPressure will be true until slow consumer catch up to above threshold, and be healthy for a whole window

xiazen:
@dotnet-bot test this please

xiazen:
functionals do succeed for me in vso .

jdom:
It might be a non-deterministic test failing

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/SlowConsumingPressureMonitor.cs
jason-bragg: I think the logic works.  It immediately stops reading if there is a cursor past the threshold, but if currently under pressure or no cursor is past the threshold, it only checks again every window.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -257,6 +260,29 @@ private EventHubAdapterReceiver GetOrCreateReceiver(QueueId queueId)
...
+            FixedSizeObjectPool<FixedSizeBuffer> bufferPool, TimePurgePredicate timePurge)
+        {
+            var cache = new EventHubQueueCache(checkpointer, bufferPool, timePurge, cacheLogger, this.SerializationManager);
+            if (adapterSettings.AveragingCachePressureMonitorFlowControlThreshold.HasValue)
+            {
+                var avgMonitor = new AveragingCachePressureMonitor(adapterSettings.AveragingCachePressureMonitorFlowControlThreshold.Value, cacheLogger);
+cache.AddCachePressureMonitor(avgMonitor);
+            }
+            if (adapterSettings.SlowConsumingMonitorPressureWindowSize.HasValue
+            || adapterSettings.SlowConsumingMonitorFlowControlThreshold.HasValue)
```
jason-bragg: What happens if only one of these are set?

src/OrleansServiceBus/Providers/Streams/EventHub/CachePressureMonitors/SlowConsumingPressureMonitor.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Threading.Tasks;
+
+namespace Orleans.ServiceBus.Providers
+{
+    /// <summary>
+    /// Pressure monitor which is in favor of the slow consumer in the cache
+    /// </summary>
+    public class SlowConsumingPressureMonitor : ICachePressureMonitor
+    {
+        private static TimeSpan DefaultPressureWindowSize = TimeSpan.FromMinutes(1);
```
jason-bragg: Defaults should probably be public, so users can specify them in configurations if they want to be explicit.

PR #2874, opened 2017-03-22
[1.4.1] More non-breaking fixes

jdom:


jdom:
For the record, please don't squash these changes. Cherrypicking changes for a patch release is the only exception where I think Rebase is the right choice. Otherwise an old-school merge commit is good enough too

ReubenBond:
LGTM. I'll rebase, since cherrypicked commits are effectively rebases anyhow

PR #2875, opened 2017-03-22
Avoid double creation of activation if is IStatefulGrain

jdom:


jdom:
Spin off to its own PR from https://github.com/dotnet/orleans/pull/2856

PR #2876, opened 2017-03-22
Improve service cleanup

jdom:
- Dispose service provider when shutting down Silo and Client (which in turns disposes all disposable services in it).
- Close log after fully shutting down.
- Do not throw when calling Stop on AsynchQueueAgent after it was disposed.


jdom:
Spin off to its own PR from #2856

jdom:

PR #2877, opened 2017-03-22
Downgrade Newtosoft.Json dependency in test projects to 7.0.1

sergeybykov:
This is to simplify distributed tests for that branch by making dependency on Newtosoft.Json uniform.


ReubenBond:
Why not upgrade everything to 9.0.1? I'm sure some dependency we pull in will require that version soon if not already.

sergeybykov:
This is for 1.4.1 branch where we depend on 7.0.1. I'm only downgrading test projects to be consistent with the core binaries.

ReubenBond:
Ah, I didn't see the branch, that makes sense, thanks

PR #2806, opened 2017-03-02
Add missing entries to app.config for OrleansHost.exe. Fixes #2795

attilah:
Add missing entries to app.config. This fix makes it possible to run OrleansHost.exe without manual modifications when added from ```Microsoft.Orleans.Server``` nuget package to a project.
Fixes #2795.

PR #2878, opened 2017-03-23
Revert binding redirects in OrleanHost.exe.config

sergeybykov:
The original change (#2806) broke our distributed tests because of the version mismatch. I suggest we revert it for now.

This reverts commit f5d6ee15ca2e637a42e7b78443fdb7d36b0e0a40.

### DIFF
--- a/src/OrleansHost/app.config
+++ b/src/OrleansHost/app.config
@@ -29,14 +29,6 @@
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-9.0.0.0" newVersion="9.0.0.0" />
       </dependentAssembly>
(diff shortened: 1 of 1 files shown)


attilah:
Sure, revert this change for now to not to break the distributed tests.

jdom:
Well, this is exactly the reason why I mentioned that band-aiding OrleansHost.exe is a bad idea if it can be avoided, because we can't dictate what others will do in their own hosts (even if using OrleansHost.exe or their own). If it is at all possible, Orleans itself shouldn't use conflicting versions of any 3rd party library.

attilah:
What about removing OrleansHost package as it is and modify it to be a Source only package with a ready to run host for 1.5?

sergeybykov:
That's something we need to consider for 1.5. Though we need to get 1.4.1 out first, and the tests currently rely on OrleansHost.exe for historical reasons.

PR #2879, opened 2017-03-23
Fix functionals 

xiazen:
Technically, consumer grain which use implicit subscribe shouldn't call UnSubscribeAsync to trigger implicitPubsub.UnRegisterConsumer. But throw a exception is a bit harsh, so change implicitPubsub UnRegisterConsumer to preserve its original logic


jdom:
I'll let @jason-bragg review this, as I don't know whether it's due to a breaking change or what

PR #2890, opened 2017-03-27
Run functionals without Azure requirement

jdom:
- Remove Azure dependency on some tests that do not require it
- Have a way to entirely opt out from using Azure (either real service or emulator) if the DataConnectionString is set to empty.
- Add some SkippableFact attribute that were missing from tests that require Azure.

### DIFF
--- a/test/TesterAzureUtils/Streaming/HaloStreamSubscribeTests.cs
+++ b/test/TesterAzureUtils/Streaming/HaloStreamSubscribeTests.cs
@@ -11,6 +11,7 @@ using TestExtensions;
 using UnitTests.GrainInterfaces;
 using UnitTests.StreamingTests;
 using Xunit;
+using Tester.AzureUtils;
 
 namespace UnitTests.HaloTests.Streaming
 {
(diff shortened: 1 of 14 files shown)


test/TesterAzureUtils/Streaming/HaloStreamSubscribeTests.cs
```diff
@@ -63,18 +67,19 @@ public HaloStreamSubscribeTests(Fixture fixture)
...
             HostedCluster = fixture.HostedCluster;
+            fixture.EnsurePreconditionsMet();
         }
-        
+
         public void Dispose()
         {
-            var deploymentId = this.HostedCluster.DeploymentId;
+            var deploymentId = this.HostedCluster?.DeploymentId;
             if (_streamProvider != null && _streamProvider.Equals(AzureQueueStreamProviderName))
```
ReubenBond: Above you check that deploymentId isn't null, but not here.  Should the check be NullOrWhitespace?

jdom: Sure, I can do that. I suppose you meant check if not null... as this is never whitespace assuming HostedCluster is set

PR #2904, opened 2017-03-30
Limit test parallelization to avoid timing issues in few core machines

jdom:
Limit parallelism to 3 test assemblies at a time.
Display test results as they come in, useful to see partial results as the test run finishes each of the test assemblies without having to wait for the entire run to finish.

### DIFF
--- a/vNext/Parallel-Tests.ps1
+++ b/vNext/Parallel-Tests.ps1
@@ -7,18 +7,18 @@ $maxDegreeOfParallelism = 3
 $failed = $false
 
 function Receive-CompletedJobs {
-    $anyFailures = $false
+    $succeeded = $true
     foreach($job in (Get-Job | Where-Object { $_.State -ne 'Running' }))
     {
         Receive-Job $job -AutoRemoveJob -Wait | Write-Host
 
         if ($job.State -eq 'Failed') { 
-            $anyFailures = $true
+            $succeeded = $false
             Write-Host -ForegroundColor Red 'Failed: ' $job.Name '('$job.State')'
         }
         Write-Host ''  
     }
-    return $anyFailures
+    return $succeeded
 }
 
 # If there is multiple xunit packages installed, take the latest one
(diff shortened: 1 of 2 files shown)


vNext/Parallel-Tests.ps1
```diff
@@ -21,16 +39,28 @@ $ExecuteCmd =
...
+    Start-Job $ExecuteCmd -ArgumentList $cmdLine -Name $([System.IO.Path]::GetFileNameWithoutExtension($a)) | Out-Null
+    Write-Host ''
 }
 
-Get-Job | Wait-Job | Receive-Job
+# Wait for all jobs to complete and results ready to be received
+Wait-Job * | Out-Null
 
-if (Get-Job -State Failed)
+if (Receive-CompletedJobs) { $failed = $true }
```
ReubenBond: Maybe this fn should coalesce the state and return that - otherwise this line is a unclear at first.

jdom: what do you mean? Not too familiar with PowerShell, so that's the easiest way I could make it work, but I'm open to suggestions if you'll bear with me

jdom: Talked in person. Inverted the result value of Receive-CompletedJobs to make it a little more intuitive

jdom:
@benjaminpetit do you know why this test is failing? https://ci.dot.net/job/dotnet_orleans/job/master/job/netfx-bvt_prtest/13/testReport/junit/Tester.HeterogeneousSilosTests/UpgradeTests/UpgradeSeveralQueuedRequestsTest/

jdom:
@dotnet-bot test netfx-bvt

jdom:
Kicked off a few times, and all runs seemed to pass (except for 1 failing BVT unrelated to timing). Also, it doesn't appear to have slowed down noticeably the total execution time, so I guess having 3 parallel test runs was a good balance.
In VSO, which our agents are many-core, and they still appear to execute as fast.

PR #2909, opened 2017-04-01
refactor Programmatic Subscribe API

xiazen:
- removing IStreamProvider.SetOnSubscriptionChangeAction
- Add IOnSubscriptionActioner
-clean isRewindable logic from StreamConsumerExtension, StreamConsumerExtension shouldn't hold a isRewindable param.  isRewindable is in StreamProvider concept space, not StreamConsumerExtension concept space
-also removed isRewindable from StreamSubscriptionHandleImpl constructor, since its another param StreamImpl knows about it !


src/Orleans/Streams/Internal/StreamSubscriptionHandleImpl.cs
```diff
@@ -39,7 +38,7 @@ public StreamSubscriptionHandleImpl(GuidId subscriptionId, StreamImpl<T> streamI
             this.observer = observer;
             this.streamImpl = streamImpl;
             this.filterWrapper = filterWrapper;
-            this.isRewindable = isRewindable;
+            this.isRewindable = streamImpl.IsRewindable;
             if (IsRewindable)
```
xiazen: let me know if you want me to remove Is(is)Rewindable member from StreamSubscriptionHandleImpl class, since I don't think it is used in other classes.

src/Orleans/Streams/Core/IOnSubscriptionActioner.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Orleans.Streams.Core
+{
+    public interface IOnSubscriptionActioner
+    {
```
xiazen: let me know if you have better naming option !

jason-bragg: IStreamSubscriptionObserver?

Also needs comments.

src/Orleans/Streams/Core/IOnSubscriptionActioner.cs
```diff
@@ -0,0 +1,35 @@
...
+        Task OnAdd<T>(StreamSubscriptionHandle<T> handle);
```
jason-bragg: OnAdd sounds like it's being told to add, when it's being notified that a subscription has already been added.
OnNewSubscription?  OnSubscriptionAdded?  OnSubscribed?

src/Orleans/Streams/Internal/StreamConsumerExtension.cs
```diff
@@ -149,15 +129,17 @@ public Task<StreamHandshakeToken> DeliverImmutable(GuidId subscriptionId, Stream
...
+                // if no observer attached to the subscription, checkif the grain reference is IOnSubscriptionChangeActioner
+                if (this.onSubscriptionActioner != null)
                 {
                     //if the onAddAction attached an observer to the subscription
                     var streamProvider = this.providerRuntime.ServiceProvider
                                 .GetService<IStreamProviderManager>()
                                 .GetStreamProvider(streamId.ProviderName);
-                    await handler.InvokeOnAdd(streamId, subscriptionId, isRewindable, streamProvider);
+                   
+                    Type constructedType = typeof(OnSubscriptionActionInvoker<>).MakeGenericType(batch.Value.GetType());
```
jason-bragg: This is not safe.  In our test we tend to use heterogeneous streams, but most streams in the real world are streams of some event base type, with specific events being among a set of subclasses.  Like a stream of fruits, consisting of apples, oranges, and bananas.  This will create a stream of the first concrete event type, not the base type.

Two possible solutions I can foresee, though more may exist.
- Implement IOnSubscriptionActioner<T> for each type one wants to subscribe to.
- Untyped stream subscription handle, with a Resume<T>(..) call so users can specify the type they want to subscribe from.

xiazen: by " create a stream of the first concrete event type, not the base type." do you mean create a StreamSubscriptionHandle of the first concrete event type, not the base type?

xiazen: understood, will fix it

xiazen:
functionals succeed in VSO.

xiazen:
@dotnet-bot test this please

src/Orleans/Streams/Core/IStreamSubscriptionObserver.cs
```diff
@@ -7,29 +7,8 @@
 
 namespace Orleans.Streams.Core
 {
-    public interface IStreamSubscriptionObserver<T>
+    public interface IStreamSubscriptionObserver<T> 
     {
         Task OnNewSubscription(StreamSubscriptionHandle<T> handle);
```
jason-bragg: nit:  I know we've been bouncing around with naming, so feel free to ignore this, but now that the ideas have settled some, I think OnSubscribed here makes more sense.  It's basically telling the grain that they've been subscribed to the stream.  OnNewSubscription sound more like a notification of something that is happening, rather than something that has already happened.

src/Orleans/Streams/Internal/StreamSubscriptionChangeHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+                }
+            }
+            observerProxy = proxy;
+            return proxy != null;
+        }
+    }
+
+    internal interface IStreamSubscriptionObserverProxy 
+    {
+        Task InvokeOnNewSubscription(StreamId streamId, GuidId subscriptionId, IStreamProvider streamProvider);
```
jason-bragg: nit:  I don't think "invoke" prefix is useful here.

src/Orleans/Streams/Internal/StreamSubscriptionChangeHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /// <summary>
+    /// Decorator class for a IStreamSubscriptionObserver<T>. Created mainly to avoid using reflection to invoke
+    /// methods on IStreamSubscriptionObserver<T>.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal class StreamSubscriptionObserverProxy<T> : IStreamSubscriptionObserver<T>, IStreamSubscriptionObserverProxy
+    {
+        private IStreamSubscriptionObserver<T> observer => this.SubscriptionObserver as IStreamSubscriptionObserver<T>;
```
jason-bragg: suggest a hard cast on the property setter rather than an 'as' on each read.  If someone uses this wrong I think a hard exception as early as possible is better than a silent null value that could surface as a null reference exception far from the cause of the problem.

src/Orleans/Streams/Internal/StreamSubscriptionChangeHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+        public Task OnNewSubscription(StreamSubscriptionHandle<T> handle)
+        {
+            return this.observer.OnNewSubscription(handle);
+        }
+
+        public Task InvokeOnNewSubscription(StreamId streamId, GuidId subscriptionId, IStreamProvider streamProvider)
+        {
+            var stream = streamProvider.GetStream<T>(streamId.Guid, streamId.Namespace) as StreamImpl<T>;
+            var handle = new StreamSubscriptionHandleImpl<T>(subscriptionId, stream);
+            return this.observer.OnNewSubscription(handle);
```
jason-bragg: should probably call to classes own OnNewSubscription which calls the observer's.  That way the observer's is called in only one place and if we need to add logging or statistics we only have one entry point.

src/Orleans/Streams/Internal/StreamSubscriptionChangeHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// <param name="streamId"></param>
+        /// <param name="messageType"></param>
+        /// <returns></returns>
+        public async Task HandleNewSubscription(GuidId subscriptionId, StreamId streamId, Type messageType)
+        {
+            // if no observer attached to the subscription, check if the grain is a IStreamSubscriptionObserver
+            IStreamSubscriptionObserverProxy observerProxy;
+            if (TryGetStreamSubscriptionObserverProxyForType(messageType, out observerProxy))
+            {
+                var streamProvider = this.providerRuntime.ServiceProvider
```
jason-bragg: doesn't seem like we need providerRuntime  or the service provider here, we only need the stream provider manager, so why not take only that in the constructor?

src/Orleans/Streams/Internal/StreamSubscriptionHandleImpl.cs
xiazen: nvm.... handle is a serializable. remove it might affect serialization

PR #2913, opened 2017-04-04
Versioning placement & compatibility directors/strategies

benjaminpetit:
Following #2837

## Proposed change
This PR I implemented configurable director/policies for versioning..

I implemented two basics compatibility and two basics versioning directors. it would be easy to inject custom directors in the future using dependency injection (like it should be possible for `PlacementDirector`)

For compatibility:
- Backward only compatible (default value)
- All versions compatible

For new placement:
- Always the minimum compatible version (default value)
- Always the latest compatible available version

More directors can be added later.

## Limitations
For now, the strategies are chosen at startup, the values or loaded from configuration. A future PR will allow changing these at runtime

### DIFF
--- a/src/Orleans/Configuration/GlobalConfiguration.cs
+++ b/src/Orleans/Configuration/GlobalConfiguration.cs
@@ -11,6 +11,8 @@ using Orleans.Providers;
 using Orleans.Storage;
 using Orleans.Streams;
 using Orleans.LogConsistency;
+using Orleans.Versions.Compatibility;
+using Orleans.Versions.Placement;
 
 namespace Orleans.Runtime.Configuration
 {
@@ -449,6 +451,10 @@ namespace Orleans.Runtime.Configuration
 
         public string DefaultPlacementStrategy { get; set; }
 
+        public VersionCompatibilityStrategy DefaultVersionCompatibilityStrategy { get; set; }
+		
+		public VersionPlacementStrategy DefaultVersionPlacementStrategy { get; set; }
+
         public TimeSpan DeploymentLoadPublisherRefreshTime { get; set; }
 
         public int ActivationCountBasedPlacementChooseOutOf { get; set; }
@@ -593,6 +599,8 @@ namespace Orleans.Runtime.Configuration
 
             ProviderConfigurations = new Dictionary<string, ProviderCategoryConfiguration>();
             GrainServiceConfigurations = new GrainServiceConfigurations();
+            DefaultVersionCompatibilityStrategy = BackwardCompatible.Singleton;
+			DefaultVersionPlacementStrategy = MinimumVersionPlacement.Singleton;
         }
 
         public override string ToString()
--- a/src/Orleans/Orleans.csproj
+++ b/src/Orleans/Orleans.csproj
@@ -183,6 +183,12 @@
     <Compile Include="Streams\PubSub\StreamSubscriptionManagerExtensions.cs" />
     <Compile Include="Utils\Factory.cs" />
     <Compile Include="Utils\IKeyedServiceCollection.cs" />
+    <Compile Include="Versions\Compatibility\AllVersionsCompatible.cs" />
+    <Compile Include="Versions\Compatibility\BackwardCompatible.cs" />
+    <Compile Include="Versions\Compatibility\ICompatibilityDirector.cs" />
+    <Compile Include="Versions\Placement\LatestVersion.cs" />
+    <Compile Include="Versions\Placement\MinimumVersion.cs" />
+    <Compile Include="Versions\Placement\IPlacementVersionDirector.cs" />
     <Compile Include="Utils\ReferenceEqualsComparer.cs" />
     <Compile Include="Serialization\ReflectedSerializationMethodInfo.cs" />
     <Compile Include="Streams\Core\StreamIdentity.cs" />
@@ -438,6 +444,7 @@
   <ItemGroup>
     <None Include="project.json" />
   </ItemGroup>
+  <ItemGroup />
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Target Name="GenerateVersionNumber" BeforeTargets="VersionInit" Condition="'$(BuildingInsideVisualStudio)' != 'true'" Outputs="$(SolutionDir)Build\Version.txt">
     <PropertyGroup Condition="'$(BuildNumber)' == ''">
--- a/src/Orleans/Versions/Compatibility/ICompatibilityDirector.cs
+++ b/src/Orleans/Versions/Compatibility/ICompatibilityDirector.cs
@@ -1,0 +1,18 @@
+﻿using System;
+
+namespace Orleans.Versions.Compatibility
+{
+    public interface IVersionCompatibilityDirector
+    {
+        bool IsCompatible(ushort requestedVersion, ushort actualVersion);
+    }
+
+    public interface IVersionCompatibilityDirector<TStrategy> : IVersionCompatibilityDirector where TStrategy : VersionCompatibilityStrategy
+    {
+    }
+
+    [Serializable]
+    public abstract class VersionCompatibilityStrategy
+    {
+    }
+}
--- a/src/Orleans/Versions/Placement/IPlacementVersionDirector.cs
+++ b/src/Orleans/Versions/Placement/IPlacementVersionDirector.cs
@@ -1,0 +1,20 @@
+﻿using Orleans.Versions.Compatibility;
+using System;
+using System.Collections.Generic;
+
+namespace Orleans.Versions.Placement
+{
+    public interface IVersionPlacementDirector
+    {
+        ushort GetSuitableVersion(ushort requestedVersion, IReadOnlyList<ushort> availableVersions, IVersionCompatibilityDirector versionCompatibilityDirector);
+    }
+
+    public interface IVersionPlacementDirector<TPolicy> : IVersionPlacementDirector where TPolicy : VersionPlacementStrategy
+    {
+    }
+
+    [Serializable]
+    public abstract class VersionPlacementStrategy
+    {
+    }
+}
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -18,8 +18,11 @@ using Orleans.Runtime.GrainDirectory;
 using Orleans.Runtime.Messaging;
 using Orleans.Runtime.Placement;
 using Orleans.Runtime.Scheduler;
+using Orleans.Runtime.Versions;
 using Orleans.Serialization;
 using Orleans.Storage;
+using Orleans.Versions.Compatibility;
+using Orleans.Versions.Placement;
 
 namespace Orleans.Runtime
 {
@@ -151,6 +154,7 @@ namespace Orleans.Runtime
         private readonly SerializationManager serializationManager;
 
         private readonly MultiClusterRegistrationStrategyManager multiClusterRegistrationStrategyManager;
+        private readonly CachedVersionDirectorManager versionDirectorManager;
 
         public Catalog(
             ILocalSiloDetails localSiloDetails,
@@ -165,7 +169,8 @@ namespace Orleans.Runtime
             PlacementDirectorsManager placementDirectorsManager,
             MessageFactory messageFactory,
             SerializationManager serializationManager,
-            MultiClusterRegistrationStrategyManager multiClusterRegistrationStrategyManager)
+            MultiClusterRegistrationStrategyManager multiClusterRegistrationStrategyManager,
+			CachedVersionDirectorManager versionDirectorManager)
             : base(Constants.CatalogId, messageCenter.MyAddress)
         {
             LocalSilo = localSiloDetails.SiloAddress;
@@ -180,11 +185,12 @@ namespace Orleans.Runtime
             this.nodeConfig = nodeConfig;
             this.serializationManager = serializationManager;
             this.multiClusterRegistrationStrategyManager = multiClusterRegistrationStrategyManager;
+            this.versionDirectorManager = versionDirectorManager;
 
             logger = LogManager.GetLogger("Catalog", Runtime.LoggerType.Runtime);
             this.config = config.Globals;
             ActivationCollector = new ActivationCollector(config);
-            this.Dispatcher = new Dispatcher(scheduler, messageCenter, this, config, placementDirectorsManager, grainDirectory, messageFactory, serializationManager);
+            this.Dispatcher = new Dispatcher(scheduler, messageCenter, this, config, placementDirectorsManager, grainDirectory, messageFactory, serializationManager, versionDirectorManager.CompatibilityDirectorManager);
             GC.GetTotalMemory(true); // need to call once w/true to ensure false returns OK value
 
             config.OnConfigChange("Globals/Activation", () => scheduler.RunOrQueueAction(Start, SchedulingContext), false);
@@ -226,7 +232,7 @@ namespace Orleans.Runtime
             IReadOnlyList<SiloAddress> silos;
             if (target.InterfaceVersion > 0)
             {
-                var version = GrainTypeManager.GetAvailableVersions(target.InterfaceId).Max();
+                var version = versionDirectorManager.GetSuitableVersion(target.InterfaceId, target.InterfaceVersion);
                 silos = GrainTypeManager.GetSupportedSilos(typeCode, target.InterfaceId, version);
             }
             else
--- a/src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
+++ b/src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
@@ -168,7 +168,7 @@ namespace Orleans.Runtime
             return supportedSilosByInterface[ifaceId].Keys.ToList();
         }
 
-        internal int GetLocalSupportedVersion(int ifaceId)
+        internal ushort GetLocalSupportedVersion(int ifaceId)
         {
             return grainInterfaceMap.GetInterfaceVersion(ifaceId);
         }
--- a/src/OrleansRuntime/Silo/Silo.cs
+++ b/src/OrleansRuntime/Silo/Silo.cs
@@ -39,7 +39,12 @@ using Orleans.Storage;
 using Orleans.Streams;
 using Orleans.Timers;
 using Orleans.MultiCluster;
+using Orleans.Runtime.Versions;
+using Orleans.Runtime.Versions.Compatibility;
 using Orleans.Streams.Core;
+using Orleans.Versions.Compatibility;
+using Orleans.Versions.Placement;
+using Orleans.Runtime.Versions.Placement;
 
 namespace Orleans.Runtime
 {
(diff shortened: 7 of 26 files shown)


src/Orleans/Configuration/GlobalConfiguration.cs
```diff
@@ -593,6 +599,8 @@ internal GlobalConfiguration()
 
             ProviderConfigurations = new Dictionary<string, ProviderCategoryConfiguration>();
             GrainServiceConfigurations = new GrainServiceConfigurations();
+            DefaultVersionCompatibilityStrategy = BackwardCompatible.Singleton;
+			DefaultVersionPlacementStrategy = MinimumVersionPlacement.Singleton;
```
jdom: nit: there is weird indenting here and in the property declaration.
UPDATE: TABS!! don't want to start a religious war about tabs vs spaces, since I don't care, but I care about consistency, so let's continue to use spaces.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -165,7 +169,8 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
             PlacementDirectorsManager placementDirectorsManager,
             MessageFactory messageFactory,
             SerializationManager serializationManager,
-            MultiClusterRegistrationStrategyManager multiClusterRegistrationStrategyManager)
+            MultiClusterRegistrationStrategyManager multiClusterRegistrationStrategyManager,
+			CachedVersionDirectorManager versionDirectorManager)
```
jdom: nit: weird indenting

src/Orleans/Versions/Compatibility/ICompatibilityDirector.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace Orleans.Versions.Compatibility
+{
+    public interface IVersionCompatibilityDirector
+    {
+        bool IsCompatible(ushort requestedVersion, ushort actualVersion);
```
jdom: `actualVersion` is confusing, I had to look at the usage to understand what it is. Maybe `versionInUse` or something like that?

jdom:
I'm glad the default placement strategy was changed to not always pick the latest. Neverhtless, I'd prefer that default is not the minimum, but instead by default try to use the requestedVersion. So new clients will effectively pick the new version (if the instance is not yet activated), while older clients will by default continue to talk to activate the previous version, even when all versions are compatible to each other.

benjaminpetit:
@jdom I think I wasn't clear in my description. And maybe there is a bug in my implementation now I think of.

Currently, the "minimum compatible version" is implemented like this:

`` availableVersions.Where(v => versionCompatibilityDirector.IsCompatible(requestedVersion, v)).Min(); ``

So it should take by default the requested version if available, if not the closest upper version. But maybe I should fallback to the smaller version like my description said.

src/Orleans/Configuration/GlobalConfiguration.cs
```diff
@@ -449,6 +451,10 @@ internal void SetReminderServiceType(ReminderServiceProviderType reminderType)
 
         public string DefaultPlacementStrategy { get; set; }
 
+        public VersionCompatibilityStrategy DefaultVersionCompatibilityStrategy { get; set; }
```
jdom: why are these "Default"? Is there a way to override the default?

jdom:
@benjaminpetit well but `availableVersions.Where(v => versionCompatibilityDirector.IsCompatible(requestedVersion, v)).Min()` uses the versionCompatibilityDirector, so it will not use the requestVersion by default in all cases, such as when the compatibility director is the one that says that every version is compatible with every other. At that point, it will ALWAYS use the minimum version, regardless of what was the requested one.

jdom:
Basically the default I'd like is to activate the requested version, or if no longer available, fallback to either the closest higher compatible version, or the closest lower compatible version (either configurable or just pick one that makes sense), or if you don't like the "closest" behavior, then the fallback can be either minimum or latest at that point. Nevertheless, since we try to activate the request version, it kind of makes sense to fall back to the closest compatible version

src/Orleans/Configuration/GlobalConfiguration.cs
benjaminpetit: In the future yes

src/Orleans/Configuration/GlobalConfiguration.cs
benjaminpetit: Yeah I think my VS2017 is set to use tabs instead of spaces... will fix that, sorry

ReubenBond: @jdom maybe we should check in a `.editorconfig` file or something. I'm surprised VS didn't automatically detect the norm, though.

src/Orleans/Versions/Compatibility/ICompatibilityDirector.cs
ReubenBond: alternatively `loadedVersion` or `currentVersion`

jdom:
Just talked in person with Ben. I'm suggesting that the default (again, what will  be used in most scenarios, but you can select other policies or add your custom one) should be the one that tesnds to break less, since testing for back compat is extremely difficult for end users.

So imagine you have versions 1, 2 and 3 deployed. If a request comes for version 1, and there are active silos with version 1, then it should activate there, since it's the most tested scenario for the developer.
If silos with version 1 are shutting down, and a client requests version 1, then it makes sense to activate the closest higher version (since it's more likely that the next higher version will be compatible with the previous). Even though technically v3 should be compatible with v1, it's likely that the most tested scenario for the developer is that v2 is compatible to v1, and v3 is compatible to v2, but not necessarily tested the hell out of the compatibility between v1 and v3.

So in a summary, the default placement IMO should be:
- Activate the requested version if active silos are available
- Fall back to activating the closest higher compatible version
- Fall back to activating the closest lower compatible version (this one can fail if the compatibility director says that compatibility is only backwards, which is fine, and an activation exception is thrown).

sergeybykov:
>Activate the requested version if active silos are available

My concern is that unless this logic is applied proportionally to the fraction of silos with that version, we can easily overload those silos. In the extreme case, the last standing silo with v1 will end up processing all v1 requests even if there's plenty of v2/v3 silos in the cluster.

jdom:
Yes, but if you are shutting down v1, then it makes sense that you are also shutting down clients that request v1 also. The thing is that the overloading possibility applies to any placement director that only takes versions into account (regardless if is is the minimum, maximum or whatever version), it's just that IMO this is the scenario less likely to break in terms of real compatibility between versions (which I iteratate that it's extremely hard for developers to test in their apps).
If we keep this default, we could also provide a variation that takes how many silos of each are available, but I'd argue that if you are shutting down v1, you are very likely to be shutting down clients in v1 also, so this might prove to be a premature optimization.

benjaminpetit:
> My concern is that unless this logic is applied proportionally to the fraction of silos with that version, we can easily overload those silos

Using "Latest Version" you will have the same issue when a new silo version starts. 

> Yes, but if you are shutting down v1, then it makes sense that you are also shutting down clients that request v1 also

Not necessary. Let's say you are pushing a simple bugfix silo-side: you don't want to redeploy client in this case.

I think the best default value would be: get the requested version, or the closest compatible upper version. If not, fallback to a lowest one, like described by @jdom

When strategy change at runtime would be implemented, I expect the dev to use this rule by default, start new silo with the latest version, wait some time so the cluster stabilize, then change the strategy to "Always latest". Then shutdown previous version silo.

Using this method allow developer to run some "smoke test calls" again the new version without creating "real" new activations on the latest version. And it avoids overloading at silo start/shutdown.

benjaminpetit:
So I made some modifications:
- Added a new version director: place new activations on all compatible silos, regardless of the version order. For example if in your cluster you have 10 silos, 2 v2 and 8 v1, 20% of the new activation will go on v2, 80% will go on v1. This is the new default director for now
- Added some caching
- Sort he returned list of silos by the director

src/OrleansRuntime/Catalog/Catalog.cs
jdom: Indenting is still weird

src/OrleansRuntime/Versions/CachedVersionDirectorManager.cs
```diff
@@ -0,0 +1,64 @@
...
+            return this.grainTypeManager.GetSupportedSilos(typeCode, ifaceId, versions);
+        }
+
+        private CachedVersionDirector GetDirector(int ifaceId)
+        {
+            var version = this.VersionPlacementDirectorManager.GetDirector(ifaceId);
+            var compat = this.CompatibilityDirectorManager.GetDirector(ifaceId);
+            var availableVersions = this.grainTypeManager.GetAvailableVersions(ifaceId);
+
+            return new CachedVersionDirector(version, compat, availableVersions);
```
jdom: not sure I understand the need for the CachedVersionDirector class, since the CachedVersionDirectorManager is already caching the results too... Why these 2 layers of caching?
is it because why might have many matches for (ifaceId,requestVersion) but typeCode might be the differentiator, so it make sense to have director.GetSuitableVersion cached for that pair?

src/Orleans/Versions/Placement/IPlacementVersionDirector.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Orleans.Versions.Compatibility;
+using System;
+using System.Collections.Generic;
+
+namespace Orleans.Versions.Placement
+{
+    public interface IVersionPlacementDirector
```
jdom: >IVersionPlacementDirector [](start = 21, length = 25)

Filename doesn't match class name (it's inverted)

src/Orleans/Versions/Placement/IPlacementVersionDirector.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        IReadOnlyList<ushort> GetSuitableVersion(ushort requestedVersion, IReadOnlyList<ushort> availableVersions, IVersionCompatibilityDirector versionCompatibilityDirector);
```
jdom: This API does not allow for weighted decisions... (like instead of randomly placing in all suitable silos, favor either the higher version if there is a big enough deployment rollout of that new version).
I'm fine that we don't over engineer this until we actually provide some of those directors in-the-box, but since it's an extension point, just wanted to raise awareness so we make a conscious choice not to support those for now.

test/Tester/HeterogeneousSilosTests/UpgradeTests/UpgradeTests.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace Tester.HeterogeneousSilosTests.UpgradeTests
+{
+    [TestCategory("Versioning"), TestCategory("ExcludeXAML"), TestCategory("SlowBVT"), TestCategory("Functional")]
+    public class MinimumVersionTests : UpgradeTestsBase
+    {
+        protected override VersionPlacementStrategy VersionPlacementStrategy => MinimumVersionPlacement.Singleton;
+        protected override VersionCompatibilityStrategy VersionCompatibilityStrategy => BackwardCompatible.Singleton;
+        
+        [Fact]
+        public Task AlwaysCreateActivationWithMinimumVersionTest()
```
jdom: >Test [](start = 60, length = 4)

please remove Test suffix from test methods

test/Tester/HeterogeneousSilosTests/UpgradeTests/UpgradeTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        protected override VersionCompatibilityStrategy VersionCompatibilityStrategy => BackwardCompatible.Singleton;
+
+        [Fact]
+        public Task AlwaysCreateActivationWithLatestVersionTest()
+        {
+            return Step1_StartV1Silo_Step2_StartV2Silo_Step3_StopV2Silo(2);
+        }
+
+        [Fact]
+        public Task UpgradeProxyCallNoPendingRequestTest()
```
jdom: >UpgradeProxyCallNoPendingRequest [](start = 20, length = 32)

Can you explain a little bit more what you are asserting? especially since the assert is in external call

test/Tester/HeterogeneousSilosTests/UpgradeTests/UpgradeTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            return ProxyCallNoPendingRequest(2);
```
jdom: >2 [](start = 45, length = 1)

consider using named parameters in all of these, such as ProxyCallNoPendingRequest(expectedVersion: 2)

src/Orleans/Configuration/GlobalConfiguration.cs
sergeybykov: I suggest we remove "Version" from the names to make them shorter or at least replace it with "Grain".

src/Orleans/Versions/Placement/IPlacementVersionDirector.cs
sergeybykov: Maybe IVersionSelector instead?

src/OrleansRuntime/Versions/CachedVersionDirectorManager.cs
benjaminpetit: You are correct, I did that because `typeCode` might be the differenciator.

But I think I might have over-engineered that. This level of caching is only useful if you have one interface for multiple `typeCode`. And you have to get  A LOT of Grain that share the interface for this cahing to be useful. I will just remove it.

src/Orleans/Versions/Placement/IPlacementVersionDirector.cs
benjaminpetit: Seems a better name yes

src/Orleans/Orleans.csproj
```diff
@@ -445,6 +452,7 @@
   <ItemGroup>
     <None Include="project.json" />
   </ItemGroup>
+  <ItemGroup />
```
ReubenBond: nit nit: superfluous

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -171,7 +174,8 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
             MultiClusterRegistrationStrategyManager multiClusterRegistrationStrategyManager,
             IStreamProviderRuntime providerRuntime,
             IStreamProviderManager providerManager,
-            IServiceProvider serviceProvider)
+            IServiceProvider serviceProvider,
+			CachedVersionSelectorManager versionSelectorManager)
```
ReubenBond: nit nit: some spacing weirdness here - looks like the second line is using tabs instead of spaces.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -280,6 +285,16 @@ internal Silo(SiloInitializationParameters initializationParams)
             services.AddSingleton<DefaultPlacementStrategy>();
             services.AddSingleton<ClientObserversPlacementDirector>();
 
+            // Versions
+            services.AddSingleton<VersionSelectorManager>();
```
ReubenBond: How can this be constructed using DI? It has two public constructors

src/OrleansRuntime/Versions/CachedVersionSelectorManager.cs
```diff
@@ -0,0 +1,54 @@
...
+            this.grainTypeManager = grainTypeManager;
+            this.VersionSelectorManager = versionSelectorManager;
+            this.CompatibilityDirectorManager = compatibilityDirectorManager;
+            this.getSilosFunc = GetSuitableSilosImpl;
+        }
+
+        public IReadOnlyList<SiloAddress> GetSuitableSilos(int typeCode, int ifaceId, ushort requestedVersion)
+        {
+            var key = Tuple.Create(typeCode, ifaceId, requestedVersion);
+            return suitableSilosCache.GetOrAdd(key, getSilosFunc);
```
ReubenBond: It doesn't have a GetOrAdd method yet, but you could use CachedReadConcurrentDictionary instead here and have cheaper access times while retaining thread safety (and no need to reassign the dictionary when resetting the cache.

`ResetCache` below could use `.Clear()`

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
```diff
@@ -151,24 +149,27 @@ internal IReadOnlyList<SiloAddress> GetSupportedSilos(int typeCode)
...
-            var key = Tuple.Create(typeCode, ifaceId, version);
-            return supportedSilosCache.GetOrAdd(
-                key,
-                tuple =>
-                {
-                    var silosWithTypeCode = supportedSilosByTypeCode[typeCode];
-                    return supportedSilosByInterface[ifaceId][version].Intersect(silosWithTypeCode).ToList();
-                });
+            var result = new List<SiloAddress>();
+            foreach (var version in versions)
```
ReubenBond: I'm sure I sound like a broken record, always complaining about allocations, but if `versions` used the concrete type instead of an interface, then this might not allocate an enumerator.

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
```diff
@@ -151,24 +149,27 @@ internal IReadOnlyList<SiloAddress> GetSupportedSilos(int typeCode)
...
+            {
+                var silosWithTypeCode = supportedSilosByTypeCode[typeCode];
+                var silosWithCorrectVersion = supportedSilosByInterface[ifaceId][version].Intersect(silosWithTypeCode).ToList();
```
ReubenBond: Is there a reason to call ToList if passing it to AddRange anyway?

src/OrleansRuntime/Silo/Silo.cs
ReubenBond: It does work, so I suppose it's fine. Did the behaviour of DI construction change in `Microsoft.Extensions.DependencyInjection` v1.1.0?

benjaminpetit: I suppose it takes the first and it works... I will merge the ctor, not needed to have two here anyway

src/OrleansRuntime/Versions/CachedVersionSelectorManager.cs
benjaminpetit: Didn't saw this class earlier, you added recently. I will look into that

src/OrleansRuntime/GrainTypeManager/GrainTypeManager.cs
benjaminpetit: I agree but I would argue that we call this method only once per type/iface/version since it is cached at an upper level. I prefer to use a strict param since no modification should be done to that list.

src/Orleans/Utils/CachedReadConcurrentDictionary.cs
```diff
@@ -133,6 +134,25 @@ public void Add(TKey key, TValue value)
...
+        /// <returns>The value for the key. This will be either the existing value for the key if the key is already in the dictionary, or the new value if the key was not in the dictionary.</returns>
+        public TValue GetOrAdd(TKey key, Func<TKey, TValue> valueFactory)
+        {
+            TValue value;
+
+            if (this.GetReadDictionary().TryGetValue(key, out value))
+                return value;
+
+            value = valueFactory(key);
+            Add(key, value);
```
ReubenBond: This should be a TryAdd, and if the TryAdd fails then it should call TryGet and return the result

ReubenBond:
LGTM with one comment - the current GetOrAdd implementation can throw under concurrent access

PR #2914, opened 2017-04-04
Updated "What's New in Orleans" with 1.4.0, 1.4.1, and Orleans FSM meetup

sergeybykov:
…etup.


src/Documentation/What's-new-in-Orleans.md
```diff
@@ -5,6 +5,70 @@ title: What's new in Orleans
 
 # What's new in Orleans?
 
+## [v1.4.1](https://github.com/dotnet/orleans/releases/tag/v1.4.1) March 27th 2017
+
+### Release notes
```
jdom: should we continue duplicating the release notes in so many places? Can't we just link to the release or the changelog?

sergeybykov: Good question. I think it's one of taste. I personally like to be able to scroll through the history of releases without the need to click back and forth. But that's just me. Maybe others don't care, and a link to release is already enough.

Issue #2799, opened 2017-03-01
OrleansSqlUtil missing System.Data.SqlClient


amccool:
after upgrading `Microsoft.Orleans.OrleansSqlUtils` from 1.3.1 to 1.4.0 receive this error.

`01:41:38Z [Runtime.Scheduler.WorkerPoolThread/3] Error XXXX.vshost.exe 0: [2017-03-01 01:41:38.120 GMT 24 ERROR 100646 MembershipOracle 172.18.4.160:11111] !!!!!!!!!! MembershipFailedToStart
Exc level 0: System.InvalidOperationException: Unable to find and/or load a candidate assembly for 'System.Data.SqlClient' invariant name.`

this error was resolved by nuget'ing in System.Data.SqlClient 4.3.0

Seems that the nuspec needs to include the dependency?   Yes this is SqlServer, and I dont know what you want to do about MySql, Postgres, etc, dependencies

veikkoeeva:
@amccool Orleans moved in 1.4.0 to be more .NET Core compatible and one of the changes in the framework involves the functionality and availability of [DbProviderFactories](https://msdn.microsoft.com/en-us/library/system.data.common.dbproviderfactory.aspx) (see https://github.com/dotnet/corefx/issues/4571). 

Now the libraries need to be explic explicitly included in the search path, i.e. Orleans binaries directory. Before that running on .NET Full framework they were implicitly available with the framework and the factory knew how to probe the right paths (application binaries, GAC, the configuration from local config, machine config etc.). In the .NET Core scenario it might not make sense to assume one using ADO.NET would have a SQL Server backend or would want to include the libraries for that by default.

It was @galvesribeiro that did the actual port, so he might still remember the exact details. :)

sergeybykov:
We could add a `Microsoft.Orleans.OrleansSqlUtils.MSSQL` package that would add SQL Server specific dependencies like `System.Data.SqlClient`. I'm not sure if it's worth it though.

galvesribeiro:
I though about that... But those .MSSQL, .Postgres, .MySQL packages, would have nothing more than a reference to the underlying DB client package and maybe hold the .sql script to create the structure... Does it worth maintain a package just for it?

sergeybykov:
>Does it worth maintain a package just for it?

That's exactly what I'm not sure about. It's a bit of a burden. But if it would simplify life of a number of people that use those databases...maybe it is worth throwing a .nuspec for that.

[...]

PR #2915, opened 2017-04-04
Add a note for System.Data.SqlClient

sergeybykov:


src/Documentation/Getting-Started-With-Orleans/Grain-Persistence.md
```diff
@@ -134,6 +134,9 @@ A quick way to test this is to (see in the aforementioned the few commented line
 The ADO.NET persistence has functionality to version data and define arbitrary (de)serializers with arbitrary application rules and streaming, but currently
 there is no method to expose them to application code. More information in [ADO.NET Persistence Rationale](#ADONETPersistenceRationale).
 
+__NB for Microsoft SQL Server users:__
+For the ADO.NET Storage Provider to work with SQL Server, you need to add [System.Data.SqlClient](https://www.nuget.org/packages/System.Data.SqlClient/) NuGet package to your project.
```
jdom: Doesn't a similar thing apply for MySQL and other supported SQL provider implementations?

sergeybykov: I don't know. The only issue that got brought up so far is #2799.

jdom: Let's summon @veikkoeeva or @shayhatsor

veikkoeeva: Yes. This is the case for all providers. It used to be that for SQL Server the provider was included by default in the framework, whereas for others one had to get it from somewhere already (e.g. from Nuget). The problem might be that the SQL Server users haven't accustomed to consider this thing, so they should perhaps be explicitly mentioend. Though it doesn't hurt to mention that in general too. So maybe mention the client needs to acquired by some means, Nuget is a good source, and in .NET Core that applies also to SQL Server.

shayhatsor: I also for mentioning that a nuget reference is needed for all ado.net providers. Mentioning only SQL Server might mislead others to think it's the only one that requires a nuget package.

sergeybykov: Can you suggest specific verbiage and links for non-SQL Server cases?

shayhatsor: Of course, AFAIK we currently officially support MySQL and the ADO.Net provider is on NuGet: https://www.nuget.org/packages/MySql.Data

jdom: I also see the https://www.nuget.org/packages/Npgsql/ (PostgreSQL) being referenced in our tests project, nevertheless, no tests actually use it, so not sure if it works.
Also, there's these provider connection factories in our code: https://github.com/dotnet/orleans/blob/2c0cb3c11c86ee9b4e3299aedeeeab885a39825d/src/OrleansSQLUtils/Storage/DbConnectionFactory.cs#L23
Nevertheless, that doesn't mean they actually work, since there are no tests for them.
Probably just mentioning SQL Server and MySql packages is enough, and people can extrapolate from there.

shayhatsor: Orleans will work with any ADO.NET provider as long as the back-end has been set up with the required tables and the required meta-data has been added in the code. Out of the box we provide setup scripts for SQL Server, MySQL and PostgreSQL. These are found [here](https://github.com/dotnet/orleans/tree/master/src/OrleansSQLUtils) and are the only ones that should be considered officially supported. Correcting my previous comment about supporting only SQL Server and MySQL. Nevertheless, I agree with @jdom about tests, I was under the impression that PostgreSQL tests were added. I wonder if @xclayl can say it's woroking.

jdom: My bad, looks like there are some PostgreSQL tests in our repo (which we don't run in CI anyway, as we don't run MySql tests either, but they are still there). There is no subclass of [RelationalStorageTests](https://github.com/dotnet/orleans/blob/016c0752dec649274f2cf3f2535a662b97618b07/test/TesterSQLUtils/StorageTests/Relational/RelationalStorageTests.cs) which seems to be the newest tests that were added. That was the only thing I looked before when I said there were no tests for Postgre, because I saw only subclasses for SQL Server and MySql.

veikkoeeva: About verbiage. Perhaps just as @shayhatsor outlines, something along these lines (don't use this as such :))
> Any database with ADO.NET provider is supported with any features set up in the provided scripts. At minimum the scripts support the membership provider (PostgreSql), SQL Server and MySQL also support membership, reminder and storage provider parts. The scripts can be found at the Nuget package (http://www.nuget.org/packages/Microsoft.Orleans.OrleansSqlUtils/, directory xyz). The ADO.NET providers corresponding the currently supported scripts are at https://www.nuget.org/packages/MySql.Data, https://www.nuget.org/packages/Npgsql/ and at https://www.nuget.org/packages/System.Data.SqlClient/.

Mentioning lists with links is helpful, but the trouble is naturally to keep them updated. It appears to be a bit so also there are newbies using relational DBs and they have inhererently more knobs and things to tune (especially in enterprise space) that might cause confusion. I would like to also mention other, more general thing, as to how could one setup a database and especially for the storage provider users on how to secure data, disaster recovery scenarios, hot/cold data (https://technet.microsoft.com/en-us/windows-server-docs/storage/storage-spaces/storage-spaces-direct-overview) perhaps and so on. A bit of a trouble is that I'm not currently working with this stuff, so I wouldn't like pretend offering "solid advice" in that regard. Perhaps it's changing, though. :)

About the ADO.NET issue... The PostgreSql tests are for membership only, since that's what's being used. The setup logic would need to overhauled a bit, since it's in two places -- storage and elsewhere -- and the setup is different in both of them. 

I would prefer to unify them and use constructor injection via DI throughout (doable also in XUnit.NET). There is a distinction there about setting up the test state invariants one normally sets class invariants, in a constructor, and then setting up environment, which I should be ensured is set up before the construction logic of any of the individual test classes -- maybe thinkinking so that the setup logic is shared amongst multiple test classes (and they may not need to share a common base class). I remember this distinction was a bit difficult to achieve with the approach of protected inheritance. The storage part went a bit into this direction, but I ran out of time to do it (indeed, the idea was also push a bit towards independent test kits). Anyhow, it's tangential to this discussion.

xclayl: That's correct, the PostgreSQL implementation doesn't include persistence.

Issue #2916, opened 2017-04-04
Issues in Orleans due to clock skew/time drifting


benjaminpetit:
Currently if all silos/client in the cluster are not synchronized, it can cause issues. 

I will try to list all usage of DateTime that are sensitive to clock skews;
- Message expiration: issue and proposed solution(s) here: #1524 
- Membership death votes: votes can expire, so it can lead to a situation where a dead silo is never declared dead by the probing silos.
- Reminder: we store the time when the reminder was supposed to tick in the first time. If the clock skew is big enough, it can delay this first tick.

I will update the list of we found other problematic usages.

Issue #1524, opened 2016-03-04
Expire messages based on arrival time instead of send time


sergeybykov:
To avoid wasteful execution of requests that already timed out on the caller side, we expire and drop messages that have their send timestamp set more than the call timeout (by default 30 seconds) from now. This logic assumes that all silos and clients have clocks synchronized within a reasonable threshold of a few seconds.

We saw an interesting outage recently when clocks of a number of machines for whatever reason got set ~50 seconds forward, and hours later reset back. This caused lots of messages to expire upon arrival and their calls to time out.

The proposal is to 
1. Timestamp messages upon arrival to silo.
2. Expire messages based on their arrival timestamp instead of their send timestamp.
3. Log a warning every time a message would have expire based on the send timestamp to help detect significant clock drifts and to ease troubleshooting of such incidents.

gabikliot:
Ohh, really? You run into that? Eventually? Again? :-) 

Expiring only on arrival time will be safe, but you will expire less msgs, and in the stress case **with multi hop grain calls* you might be really expiring much less that what you could have otherwise, running a higher risk of overload.

I remember I had a full design fleshed out how to detect not-synced clocks. It was basically doing a handshake between any pair of silos and client to gateway upon first connection and validating that clocks are synced. If not synced, it could alert, or fail to start one of them, or not use expiration.
This handshake could later be extended to do more validation, for example to validate that all silos and clients are configured in the same way (for setting that must match). We already have a place for such handshake in the code.

Or alternatively, you could move to use a cloud provider that really knows how to sync its clocks. :-)

sergeybykov:
Yep, a bunch of timeouts out of the blue at 4:30 in the morning. :-)

What's unclear to me is how likely it is for a message to be delayed before it is sent vs. in the queues on the destination silo. Sending queues in theory are independent from grain performance (other than sharing the CPUs with them). The receiving end seems much more sensitive to grain, especially non-reentrant, being slow for external reasons, e.g. their IO dependencies. Without any data to back it up I *suspect* that the risk of performing wasteful work because of expiring base on arrival time is relatively small.

I have no idea what cloud provider you are talking about. I only know two of them. :-)

gabikliot:
The real problem is multi hop calls. For that you need the original first send time.

Yes, only 2 providers: GCP and AWS.

sergeybykov:
[...]

PR #2922, opened 2017-04-07
Change how message expiration is handled

benjaminpetit:
Currently in Orleans, clients and silos will, by default for almost all grain calls, drop messages that expired based on the value of  `Message.Expiration`. This logic assumes that all silos and clients have clocks synchronized within a reasonable threshold of a few seconds.

If there is too much discrepancy between silos/clients clocks, this can caused lots of messages to expire upon arrival and their calls to time out. (see #1524 and #2916)

To remove this clock synchronization requirement, I implemented a slight variation of the solution proposed by @rrector here https://github.com/dotnet/orleans/issues/1524#issuecomment-193002012

- We now store the "how much remaining time to live this message has" instead of the "expected date when the message should be discarded". This value is set originally by the caller.
- When the `Message` constructor is called, we store the "local creation date" (this value is local ie not serialized and not send to other clients/silos)
- Each time we access this value, for serialization or to check if the call should be dropped, we readjust it by comparing the current date time to the "local creation date".

The benefits of this solution compared to the original proposition in #1524 is that:

- we take into account the "routing" time
- from my experimentation way more easy/elegant/less impact
- caller can drop incoming response like today, without having to manage two "expiration time"


benjaminpetit:
Maybe we should rename `Message.Expiration` to something like `Message.TimeToLive` ?

src/Orleans/Messaging/Message.cs
```diff
@@ -278,13 +280,27 @@ public ResponseTypes Result
...
 
-        public DateTime? Expiration
+        public TimeSpan? Expiration
         {
-            get { return Headers.Expiration; }
+            get
+            {
+                var elapsed = DateTime.UtcNow - localCreationTime;
+                Headers.Expiration = Headers.Expiration - elapsed;
+                return Headers.Expiration;
```
ReubenBond: If I access this property 100 times in succession, the header will expire much more quickly than it should, or am I misunderstanding the code?

ReubenBond:
+1 for renaming `Expiration` to `TimeToLive`

LGTM other than the comment.

Side note, maybe for another day: we could handle sender-side expiration more efficiently by creating a linked list of CallbackData objects and using a single timer instead of creating a timer for every message.

src/Orleans/Messaging/Message.cs
benjaminpetit: Yup I need to update the `localCreationTime` too

jdom:
:+1: on renaming to TimeToLive, as Expiration is misleading now

test/DefaultCluster.Tests/DropMessageTests.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            foreach (var task in tasks)
+            {
+                try
+                {
+                    await task;
+                }
+                catch (TimeoutException)
+                {
+                }
```
jdom: Please do:
`await Assert.ThrowsAsync<TimeoutException(() => task);`
This way you assert that the exception is indeed thrown

src/Orleans/Messaging/Message.cs
```diff
@@ -62,6 +62,8 @@ public string TargetHistory
 
         private object bodyObject;
 
+        private DateTime localCreationTime;
```
jdom: Add the `[NonSerialized]` attribute for this field.

src/Orleans/Messaging/Message.cs
```diff
@@ -278,13 +280,29 @@ public ResponseTypes Result
...
-        public DateTime? Expiration
+        public TimeSpan? Expiration
         {
-            get { return Headers.Expiration; }
+            get
+            {
+                var now = DateTime.UtcNow;
+                var elapsed = now - localCreationTime;
+                Headers.Expiration = Headers.Expiration - elapsed;
+                localCreationTime = now;
```
jdom: since you are doing this, maybe the field should be renamed?
How about `localTimeToLiveStartTime`? (since you are also renaming the Expiration property to TimeToLive).

gabikliot:
Why do we need to re adjust local creation time on every access and not only once when created?
How does it take into account routing time?

benjaminpetit:
> Why do we need to re adjust local creation time on every access and not only once when created?
How does it take into account routing time?

Yes, we could no re adjust the local creation time if the `Expiration` member was purely calculated like this
``` charp
public TimeSpan? Expiration
{
    get
    {
        var elapsed = DateTime.UtcNow - localCreationTime;
        return Headers.Expiration - elapsed;
    }
    set { Headers.Expiration = value; }
}
```
Issue is `HeadersContainer.Expiration` isn't updated before serialization. But now I realize that we should move this code in `HeadersContainer` so it is really updated before sending it: if we don't call `Message.IsExpired` or `Message.Elapsed` the value is NOT updated. I should fix that.

By "routing time" I wanted to say "time spent in gateway and silos before finding the correct target silo"

test/TesterInternal/TimeoutTests.cs
```diff
@@ -84,5 +85,37 @@ public void Timeout_LongMethod()
...
+                await Assert.ThrowsAsync<TimeoutException>(() => secondCall);
+            }
+            catch
+            {
+                output.WriteLine(firstCall.IsFaulted ? $"firstCall: faulted" : $"firstCall: {firstCall.Result}");
+                output.WriteLine(secondCall.IsFaulted ? $"secondCall: faulted" : $"secondCall: {secondCall.Result}");
+                throw;
+            }
+
+            await Task.Delay(TimeSpan.FromSeconds(5));
```
jdom: nit: I'd make this `TimeSpan.FromSeconds(5)` a variable and use it here and in firstCall, otherwise it's not really clear why you are waiting.

test/NonSilo.Tests/SerializerTests/MessageSerializerTests.cs
```diff
@@ -32,6 +33,63 @@ public void MessageTest_BinaryRoundTrip()
...
+            bodyList.Add(new ArraySegment<byte>(body));
+            var context = new DeserializationContext(this.fixture.SerializationManager)
+            {
+                StreamReader = new BinaryTokenStreamReader(headerList)
+            };
+            var deserializedMessage = new Message
+            {
+                Headers = SerializationManager.DeserializeMessageHeaders(context)
+            };
+            deserializedMessage.SetBodyBytes(bodyList);
```
jdom: wow, looks like a tremendous amount of code just to test serialization :P Anyways, if it's how it's typically done, go for it

gabikliot:
Since you changed the type, this makes this change not backward compatible. New and old versions will fail to deserialze in the cluster that uses gradual upgrade. Are you concerned about that?

gabikliot:
Also, maybe I misread the code (reading on phone) but it looks like you never update the time to live field? That means if the msg spent 25 sec on silo 1 and then arrives to silo 2, you will restart the time to live to 30 sec.

benjaminpetit:
> Since you changed the type, this makes this change not backward compatible. New and old versions will fail to deserialze in the cluster that uses gradual upgrade. Are you concerned about that?

Not really since for versioning we introduced other breaking changes.

> Also, maybe I misread the code (reading on phone) but it looks like you never update the time to live field? That means if the msg spent 25 sec on silo 1 and then arrives to silo 2, you will restart the time to live to 30 sec.

The time to live field is updated on serialization of the `HeaderContainer` (since the getter recompute the actual value)

benjaminpetit:
@dotnet-bot test this please

jdom:
LGTM. @gabikliot do you have some followup concerns or can we merge this?

PR #2924, opened 2017-04-07
getNodeConfig is not available in client side

xiazen:
disable it in client side

### DIFF
--- a/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
+++ b/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
@@ -52,7 +52,6 @@ namespace Orleans.ServiceBus.Providers
         private string[] partitionIds;
         private ConcurrentDictionary<QueueId, EventHubAdapterReceiver> receivers;
         private EventHubClient client;
-        private Lazy<Func<NodeConfiguration>> getNodeConfig;
         /// <summary>
         /// Gets the serialization manager.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -118,8 +117,7 @@ public virtual void Init(IProviderConfiguration providerCfg, string providerName
             serviceProvider = svcProvider;
             receivers = new ConcurrentDictionary<QueueId, EventHubAdapterReceiver>();
             this.SerializationManager = this.serviceProvider.GetRequiredService<SerializationManager>();
-            this.getNodeConfig = svcProvider.GetRequiredService<Func<NodeConfiguration>>();
-
+            this.getNodeConfig = new Lazy<Func<NodeConfiguration>>(() => this.serviceProvider.GetRequiredService<Func<NodeConfiguration>>());
```
jdom: seems like a very odd way of making it not required. Can't you just use `serviceProvider.GetService` instead of `GetRequiredService`?

jason-bragg: Just providing some context here, it is required, just not on the client.  Initialization happens in Orleans clients and Orleans silos, but receivers (which require this) are only created on the silo.

xiazen: +1. sorry I didn't put on some context on the description, which makes the PR very confusing

jdom: Still, if you use `GetService`, you can check that it's not null when you want to create a receiver. Using Lazy is still a very odd way of achieving this.

jdom: And you are even storing the serviceProvider instance. In that case, why don't you just delay when you call GetRequiredService to the point where this is actually required (when creating the receiver)?

xiazen: yes I can do that too! great advice. we must be smoking something bad when we did this

xiazen:
@dotnet-bot test this please.

xiazen:

xiazen:

jdom:

xiazen:
poke poke, @jdom @jason-bragg

PR #2925, opened 2017-04-07
Remove warning in GatewayManager

benjaminpetit:
It seems that VS complains because `GatewayManager.gatewayRefreshTimer` is not properly disposed. Switching "to "explicit null check" seems to remove the warning.

### DIFF
--- a/src/Orleans/Messaging/GatewayManager.cs
+++ b/src/Orleans/Messaging/GatewayManager.cs
@@ -253,8 +253,8 @@ namespace Orleans.Messaging
 
         public void Dispose()
         {
-            if (gatewayRefreshTimer != null)
-                gatewayRefreshTimer.Dispose();
+            var timer = gatewayRefreshTimer;
+            if (timer != null) timer.Dispose();
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/Orleans/Messaging/GatewayManager.cs
```diff
@@ -253,7 +253,8 @@ private void UpdateLiveGatewaysSnapshot(IEnumerable<Uri> currentKnownGateways, T
 
         public void Dispose()
         {
-            gatewayRefreshTimer?.Dispose();
+            if (gatewayRefreshTimer != null)
+                gatewayRefreshTimer.Dispose();
```
jdom: potential race condition. Opinions @ReubenBond since you added this?
I would prefer to just ignore the warning explicitly. Hopefully they'll fix this at some point (see https://github.com/dotnet/roslyn-analyzers/issues/291)

ReubenBond: IIRC, the analyzer will understand the

```C#
var timer = gatewayRefreshTimer;
if (timer != null) timer.Dispose();
```

Pattern, so we should use it

jdom: yup, that was what I suggested in an email to him. I'd be fine with either (copy value to variable, or explicit supression of warning), but prefer not to risk race conditions

Issue #2844, opened 2017-03-14
Json as Event Hubs message serialization format


DixonDs:
Is it possible to use JSON serializer for serializing Event Hubs stream messages? We are trying to use the Event Hubs stream provider for communication between different clusters (with different codebase and purpose), and using default binary serializer puts usual compatibility issues of the binary format which is something we would like to avoid by using other serializer.

xiazen:
I think it is possible depends on how complicated your event hub message is, but I will  @jason-bragg  and @ReubenBond since they are the two people who are most familiar with the event hub and serializing.

jason-bragg:
To my knowledge no effort has been made to ensure that the eventhub stream provider works with json serialization.

The eventhub adapter does, however, support customization of how data is written to and read from eventhub.  We have consumers that use this feature, so it has been demonstrated to work, but it's not very convenient. 

If you need to control format of written data you'll need to override the QueueMessageBatchAsync in the adapter (implemented in EventHubAdapterFactory).

For custom deserialization of eventhub messages one will need to implement a specialization of the EventHubDataAdapter and configure the adapter to use it. We don't have a full example of this in our codebase, but a somewhat related example can be found in the StreamPerPartitionEventHubStreamProvider in the ServiceBus.Tests assembly. This stream provider uses a custom EventHubDataAdapter, similar to what you'll need to do, but the example only customizes how streamId is derived from the eventhub data, while you'll need to override the GetBatchContainer to control format of read data.

I understand that customizing the adapter in these ways is not desirable, and I hope to spend more time improving this, but in it's current state, it's 'possible' but not 'pretty'.

DixonDs:
@jason-bragg 
>If you need to control format of written data you'll need to override the QueueMessageBatchAsync in the adapter (implemented in EventHubAdapterFactory).

It will help if that method is virtual

gabikliot:
What if he just installs an external serializer and implements json ser for his stream data types? Wouldn't it just work?

DixonDs:
@gabikliot Not really, because the Event Hubs provider uses the private nested class `Orleans.ServiceBus.Providers.EventHubBatchContainer.Body` as a message wrapper. However, as a temporary workaround, I will try to define an external serializer for this private class. There are two problems with this approach even if it works:
1) Obviously it relies on the internal implementation details which can change at any moment
2) The serializer can be set globally, so it will be not possible to have two Event Hubs providers with different serialization formats registered at the same time

[...]

PR #2930, opened 2017-04-11
Add extension points to EventHubAdapterFactory

DixonDs:
Related to https://github.com/dotnet/orleans/issues/2844. In the current form, it is quite problematic to extend `EventHubAdapterFactory`, so these changes should make it easier a bit

### DIFF
--- a/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
+++ b/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
@@ -28,26 +28,32 @@ namespace Orleans.ServiceBus.Providers
         /// Orleans logging
         /// </summary>
         protected Logger logger;
+
         /// <summary>
         /// Framework service provider
         /// </summary>
         protected IServiceProvider serviceProvider;
+
         /// <summary>
         /// Provider configuration
         /// </summary>
         protected IProviderConfiguration providerConfig;
+
         /// <summary>
         /// Stream provider settings
         /// </summary>
         protected EventHubStreamProviderSettings adapterSettings;
+
         /// <summary>
         /// Event Hub settings
         /// </summary>
         protected IEventHubSettings hubSettings;
+
         /// <summary>
         /// Checkpointer settings
         /// </summary>
         protected ICheckpointerSettings checkpointerSettings;
+
         private IEventHubQueueMapper streamQueueMapper;
         private string[] partitionIds;
         private ConcurrentDictionary<QueueId, EventHubAdapterReceiver> receivers;
@@ -79,22 +85,26 @@ namespace Orleans.ServiceBus.Providers
         /// <summary>
         /// Creates a message cache for an eventhub partition.
         /// </summary>
-        protected Func<string, IStreamQueueCheckpointer<string>, Logger, IEventHubQueueCache> CacheFactory { get; set; }
+        protected IEventHubQueueCacheFactory CacheFactory { get; set; }
+
         /// <summary>
         /// Creates a parition checkpointer.
         /// </summary>
         protected Func<string, Task<IStreamQueueCheckpointer<string>>> CheckpointerFactory { get; set; }
+
         /// <summary>
         /// Creates a failure handler for a partition.
         /// </summary>
         protected Func<string, Task<IStreamFailureHandler>> StreamFailureHandlerFactory { get; set; }
+
         /// <summary>
         /// Create a queue mapper to map EventHub partitions to queues
         /// </summary>
         protected Func<string[], IEventHubQueueMapper> QueueMapperFactory { get; set; }
+
         /// <summary>
         /// Create a receiver monitor to report performance metrics.
-        ///   Arguments are EventHub path, EventHub partition, and logger. 
+        ///   Arguments are EventHub path, EventHub partition, and logger.
         ///   Factory funciton should return an IEventHubReceiverMonitor.
         /// </summary>
         protected Func<string, string, Logger, IEventHubReceiverMonitor> ReceiverMonitorFactory { get; set; }
@@ -141,12 +151,7 @@ namespace Orleans.ServiceBus.Providers
 
             if (CacheFactory == null)
             {
-                var bufferPool = new FixedSizeObjectPool<FixedSizeBuffer>(adapterSettings.CacheSizeMb, () => new FixedSizeBuffer(1 << 20));
-                var timePurge = new TimePurgePredicate(adapterSettings.DataMinTimeInCache, adapterSettings.DataMaxAgeInCache);
-                CacheFactory = (partition, checkpointer, cacheLogger) =>
-                {
-                    return CreateCacheFactory(partition, checkpointer, cacheLogger, bufferPool, timePurge);
-                };
+                CacheFactory = CreateCacheFactory(adapterSettings);
             }
 
             if (StreamFailureHandlerFactory == null)
@@ -232,7 +237,7 @@ namespace Orleans.ServiceBus.Providers
 #if NETSTANDARD
             return client.SendAsync(eventData, streamGuid.ToString());
 #else
-            return client.SendAsync(eventData); 
+            return client.SendAsync(eventData);
 #endif
         }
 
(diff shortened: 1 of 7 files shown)


src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -283,6 +283,12 @@ private EventHubAdapterReceiver GetOrCreateReceiver(QueueId queueId)
             return cache;
         }
 
+        protected virtual IEventHubQueueCache CreateCache(IStreamQueueCheckpointer<string> checkpointer,
```
jason-bragg: The pattern for customizing the cache is to provide one's own CacheFactory.  That's why we check to see if it is null before assigning it in the initialization, so a specialization can provide it's own.  Having this virtual function as an alternate incomplete extension point, imo, confuses the pattern.  I concede that assigning a CacheFactory is probably not the best approach, but I don't think we should have multiple extensibility patterns.

DixonDs: I think in this case, it is better to extract the whole cache factory method into a separate class with its own extension points. For instance, in my case, I want to reuse the logic assigning `AveragingCachePressureMonitor` etc and just extend existing `IEventHubQueueCache` implementation which is `EventHubQueueCache`.

jason-bragg: That sounds like a good idea.

We have an abstraction for that, IQueueAdapterCache, but I'm not really sure it's a good fit.  Due to the tighter coupling of the receiver and the cache (in this stream provider), we've not really used it as intended.

We also have a factory pattern in DI now (see src\Orleans\Utils\Factory.cs).  The issue with using DI in these cases has been the fact that we can't assume that every stream provider wants the same caching patterns.  We do have a DI disambiguation pattern (see IKeyedServiceCollection) now, which can be used to get a cache factory for a specific stream provider, but that's not really been thought through yet.

xiazen: You can extend existing IEventHubQueueCache implementation by implementing a class inheriting EventHubQueueCache. EventHubQueueCache is an abstract class. Or does that not work for you? 
I think we are not debating extension through inheritance is better or through a factory pattern is better. It's if the factory fulfils your needs, imo we shouldn't introduce another pattern to bring more confusion.

DixonDs: @xiazen I don't have a problem extending `EventHubQueueCache`, but I don't want to duplicate logic creating `FixedSizeObjectPool<FixedSizeBuffer>`, `TimePurgePredicate` and then the whole `CreateCacheFactory` method

xiazen: Ah I see what you want to achieve here. You want to inject your extended version of EventHubQueueCache, but still reuse the set up logic on the cache, such as time predicate etc. 

And also it can be used for use case when application wants to use default set up on the EventHubQueueCache, but just add some adjustment on top of that, such as add more pressure monitor,etc. 

I can see another alternative which might be less confusing for users:  instead of making this `CreateCache` a method, we can  make it a Func, just like CreateCacheFactory func. 

Another alternative which I like better. Since we are intending to add more extension point to `CreateCacheFactory` . Maybe it's a good idea to encapsulate all those functionality into a CreateCacheFactory class, and user can extend that CreateCacheFactory class by inheritance or decorating. This class will be responsible for create EventHubQueueCache and configure it. 


``

xiazen: We can provide a default implementation of that class, and user can extend it through inheritance or decorating. In this way, we can make EventHubAdapterFactory (a already complicated class) simpler and cleaner, and bring more flexibility in the future to CreateCacheFactory class if we want to do more customization. 

We can also do a ICreateCacheFactory interface, and provide a default implementation for it. I will leave it to you to decide pros or cons on those alternatives. 

I like what you are intending to do in this PR. But I'd like to see it be done in a less confusing way.

DixonDs:
@jason-bragg @xiazen Please, have a look on the latest changes where I created `IEventHubQueueCacheFactory` per discussion

src/OrleansServiceBus/Providers/Streams/EventHub/IEventHubQueueCacheFactory.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Orleans.Runtime;
+using Orleans.Streams;
+
+namespace Orleans.ServiceBus.Providers
+{
+    /// <summary>
+    /// Factory responsible for creating a message cache for an EventHub partition.
+    /// </summary>
+    public interface IEventHubQueueCacheFactory
```
jason-bragg: Consider an Initialize call that takes the EventHubStreamProviderSettings, and SerializationManager rather than passing them through constructor.  That way the cache factory can be a configured type in the EventHubStreamProviderSettings rather than requiring one to subclass the factory and override the CreateCacheFactory.  Though the virtual function is ok, just strange having a factory factory in a factory.. feels a bit wild-westy.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubQueueCacheFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+        private readonly SerializationManager _serializationManager;
+        private readonly FixedSizeObjectPool<FixedSizeBuffer> _bufferPool;
+        private readonly TimePurgePredicate _timePurge;
+
+        public EventHubQueueCacheFactory(EventHubStreamProviderSettings providerSettings, SerializationManager serializationManager)
+        {
+            _providerSettings = providerSettings;
+            _serializationManager = serializationManager;
+
+            _bufferPool = new FixedSizeObjectPool<FixedSizeBuffer>(_providerSettings.CacheSizeMb, () => new FixedSizeBuffer(1 << 20));
```
jason-bragg: If this is supposed to be the 'customization point' for cache creation, I suggest a virtual CreatePool function as well.  Pool specialization is not uncommon.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubQueueCacheFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+            _timePurge = new TimePurgePredicate(_providerSettings.DataMinTimeInCache, _providerSettings.DataMaxAgeInCache);
+        }
+
+        public IEventHubQueueCache CreateCache(string partition, IStreamQueueCheckpointer<string> checkpointer, Logger cacheLogger)
+        {
+            var cache = CreateCache(checkpointer, cacheLogger, _bufferPool, _timePurge, _serializationManager);
+            if (_providerSettings.AveragingCachePressureMonitorFlowControlThreshold.HasValue)
```
jason-bragg: suggest "AddPressurceMonitors" virtual function to specialize function, for those that want to customize the pressure monitoring.

src/OrleansServiceBus/Providers/Streams/EventHub/IEventHubQueueCacheFactory.cs
DixonDs: I personally don't like `Initialize` approach since in fact it violates encapsulation (every method call moves an object from one valid state to other valid state) - after ctor call but before `Initialize`, the object is in an inconsistent state. Then, it is mentioned in https://blogs.msdn.microsoft.com/orleans/2017/04/12/de-inventing-the-wheel/ that the goal is to move away from 2-step initialization

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubQueueCacheFactory.cs
DixonDs: Since currently it is created in ctor, a virtual method won't really work. Instead, I've added an alternative ctor allowing to pass a custom pool.

src/OrleansServiceBus/Providers/Streams/EventHub/IEventHubQueueCacheFactory.cs
jason-bragg: Fair point.  The cost of this pattern is that it requires one to subclass the adapter factory, but I guess that's kinda its purpose.  My hope is to replace all of this with a builder some day.
One can dream!? :)

test/ServiceBus.Tests/TestStreamProviders/StreamPerPartitionEventHubStreamProvider.cs
```diff
@@ -19,27 +22,36 @@ namespace ServiceBus.Tests.TestStreamProviders.EventHub
...
             }
 
-            private IEventHubQueueCache CreateQueueCache(string partition, IStreamQueueCheckpointer<string> checkpointer, Logger log)
+            public IEventHubQueueCache CreateCache(string partition, IStreamQueueCheckpointer<string> checkpointer, Logger cacheLogger)
             {
-                if (timePurgePredicate != null)
-                {
-                    timePurgePredicate = new TimePurgePredicate(adapterSettings.DataMinTimeInCache, adapterSettings.DataMaxAgeInCache);
-                }
+                var timePurgePredicate = new TimePurgePredicate(adapterSettings.DataMinTimeInCache, adapterSettings.DataMaxAgeInCache);
```
jason-bragg: no need to allocate this per cache.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -79,22 +85,26 @@ public class EventHubAdapterFactory : IQueueAdapterFactory, IQueueAdapter, IQueu
         /// <summary>
         /// Creates a message cache for an eventhub partition.
         /// </summary>
-        protected Func<string, IStreamQueueCheckpointer<string>, Logger, IEventHubQueueCache> CacheFactory { get; set; }
+        protected IEventHubQueueCacheFactory CacheFactory { get; set; }
```
jason-bragg: For compatability, please leave this as a func.
I'm fine with encapsulating the cache factory logic in it's own class that can be specialized and replaced, but the minimal behavior that needs replaces is this factory func, no reason to force developers to use our cache factory class or interface.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubQueueCacheFactory.cs
```diff
@@ -0,0 +1,76 @@
...
+using Orleans.Serialization;
+using Orleans.Streams;
+
+namespace Orleans.ServiceBus.Providers
+{
+    public class EventHubQueueCacheFactory : IEventHubQueueCacheFactory
+    {
+        private readonly EventHubStreamProviderSettings _providerSettings;
+        private readonly SerializationManager _serializationManager;
+        private readonly IObjectPool<FixedSizeBuffer> _bufferPool;
```
jason-bragg: This pattern assumes that the same buffer pool is used per cache, which is the default, but not the only supported pattern.
We support a buffer pool:
- Per silo, which can be done here by passing the same buffer pool to each cache factory for all providers.
- Per stream provider, which is the default, and what this implementation does by default.
- Per cache, which this does not support.

I'm not necessarily suggesting you change this, as this is a default utility class, and to that end, it does provide utility, I'm only raising the point that this class will not be useful to those who need a block pool per queue, which is not an uncommon pattern.

test/ServiceBus.Tests/TestStreamProviders/StreamPerPartitionEventHubStreamProvider.cs
```diff
@@ -19,27 +22,36 @@ namespace ServiceBus.Tests.TestStreamProviders.EventHub
...
+            private readonly SerializationManager serializationManager;
+            private readonly Logger log;
 
-            public AdapterFactory()
+            public CacheFactory(EventHubStreamProviderSettings adapterSettings, SerializationManager serializationManager, Logger log)
             {
-                CacheFactory = CreateQueueCache;
+                this.adapterSettings = adapterSettings;
+                this.serializationManager = serializationManager;
+                this.log = log;
```
jason-bragg: why are we passing in the log at cach factory creation time and using the same log per cache, even though a log is provided for each cach in the create cache call?

jason-bragg:
Much thanks @DixonDs!

Issue #2764, opened 2017-02-22
Request for Location Aware Grain Placement – Grain placement strategy


sarmut:
**Goal**: Add an additional placement strategy called LocationAwarePlacement where the activation of the grain type should be placed subject to a given list of node locations inside a silo.

**Motivation**: In our system, we have a collection of files distributed across several servers. Each file is replicated in 3 different servers for availability.  Orleans silo is running in the same cluster.  For any kind of operations that involve processing the file, it makes sense to activate a grain that processes a particular file on the same server where it is being held to provide data-compute locality. 

**Suggested Path**: When calling GetGrain<>(), inject the desired silo locations for the grain activation.  Orleans can pick the silo from this given list.  If none of the desired locations are not possible due to heavy load, it can throw an exception.

This approach does not require changing the existing PlacementStrategy and PlacementDirector interfaces. However, we need to turn the two interfaces from internal to public. 

**Challenges**:  Orleans placement related classes are marked ‘internal’.  This needs to open for injection from outside.

**Sample Solution:**

A sample solution could encode the file identifier as part of the grain identifier so that when asked for an activation location, the placement director could do the following:

1.	Get the file identifier from the grain identifier
2.	Retrieve the replica locations for the file
3.	Choose the activation locations based on the replica locations

The director prefers to activate the grain on the same machine as one of the replica, but may choose another machine when the constraint cannot be satisfied (e.g., workloads).

**Related Threads:**

- https://github.com/dotnet/orleans/issues/96 
- https://github.com/dotnet/orleans/issues/1123 
- https://github.com/dotnet/orleans/issues/2345

sergeybykov:
>Suggested Path: When calling GetGrain<>(), inject the desired silo locations for the grain activation.

`GetGrain()` doesn't perform activation placement. It happens upon a first call to the target grain.

It seems to me this requires support for custom placement strategy & director. Internally, the director will pick a target silos for each new activation. The question I have here is if a static configuration (supplied at silo startup) will be enough for the custom placement director. Or if it'll need updates about location of the resources at runtime. The former, obviously, is much simpler.

rapoth:
Thanks @sergeybykov.

While I agree static configuration for the director is a bit easy, my impression is that in our current situation, there are several reasons why a static configuration might not work well:

- Continuously arriving files are distributed and replicated across servers active at that point in time
- If a server is getting full or having any health issues, files could be moved around
- Servers are assumed to fail randomly so the list of active servers is a function of time

[...]

PR #2932, opened 2017-04-11
Support for custom placement strategies and directors

sergeybykov:
`IPlacementDirector` today defines two methods: `OnAddActivation` and `OnSelectActivation`. The latter is used only for a couple of special cases, and custom placement policies only need to be concerned with the former. `OnSelectActivation` depends on a number of internal types, while `OnAddActivation` does not.

I split the two operations between two interfaces: public `IPlacementDirector` for custom (and built-in) placement directors to implement and  internal `IActivationSelector` for those special internal (for now) cases. A few more types needed to be made public, but those were easy.

This should enable #2764 and #96.

Placement strategies for now have to be serializable. We can look at removing this requirement later.


src/Orleans/Core/IGrainIdentity.cs
```diff
@@ -12,8 +12,12 @@ public interface IGrainIdentity
...
-        long GetPrimaryKeyLong(out string keyExt);
+	    bool IsClient { get; }
+
+		int TypeCode { get; }
+
+		long GetPrimaryKeyLong(out string keyExt);
 
         Guid GetPrimaryKey(out string keyExt);
-    }
+	}
```
ReubenBond: Nit: most of the codebase uses spaces for indentation

src/OrleansRuntime/Placement/PlacementDirectorsManager.cs
```diff
@@ -106,19 +106,27 @@ private IActivationSelector ResolveSelector(PlacementStrategy strategy, bool add
...
+
+            foreach (var strategy in placementStrategies)
+                this.ResolveDirector(strategy);
+            
+            var selectorStrategies = new PlacementStrategy[]
             {
-                var director = this.ResolveDirector(strategy);
-	            if (director is IActivationSelector)
-		            this.ResolveSelector(strategy);
-            }
```
jdom: what was wrong with this approach? it was a nice one to avoid issues if more built in strategies support selection in the future. The only thing that was missing was passing `true` to `ResolveSelector`, right?
Anyway, I think it's OK regardless, even if my preference is the previous approach.

src/Orleans/Placement/PlacementTarget.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public int InterfaceId { get; }
+
+        public ushort InterfaceVersion { get; }
+
+        public bool IsClient => GrainIdentity.IsClient;
+
+        public PlacementTarget(IGrainIdentity grainIdentity, int interfaceId, ushort interfaceVersion)
+        {
+			GrainIdentity = grainIdentity;
```
jdom: be careful with tabs in several places

src/OrleansRuntime/Placement/PlacementDirectorsManager.cs
sergeybykov: The problem was rather subtle. `RandomPlacementDirector` implements both `IPlacementDirector` and `IActivationSelector`. `ActivationCountPlacementDirector` extends `RandomPlacementDirector`, but it only really implements `IPlacementDirector`, though the check `is IActivationSelector` returns `true` for it. That was causing a problem.

I considered splitting `RandomPlacementDirector` into separate classes for placement director and activation selector, but thought it would cause even more code churn. Can be done as a separate change later.

src/Orleans/Placement/PlacementTarget.cs
sergeybykov: Weird. Must be different setting on my desktop. I fixed them, but Git seems to be ignoring the changes.

test/TestGrainInterfaces/CustomPlacement.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace UnitTests.GrainInterfaces
+{
+	public interface ICustomPlacementTestGrain : IGrainWithGuidKey
+	{
+		Task<string> GetRuntimeInstanceId();
+	}
+
+
+	public enum CustomePlacementScenario
```
benjaminpetit: very nit: no 'e' after "Custom"

src/OrleansRuntime/Placement/IPlacementRuntime.cs
```diff
@@ -39,15 +33,15 @@ internal interface IPlacementContext
         void GetGrainTypeInfo(int typeCode, out string grainClass, out PlacementStrategy placement, out MultiClusterRegistrationStrategy strategy, string genericArguments = null);
     }
 
-    internal static class PlacementContextExtensions
+	internal static class PlacementContextExtensions
```
anjaneya17: Minor: rename to PlacementRunTimeExtensions?

src/OrleansRuntime/Placement/PlacementDirectorsManager.cs
```diff
@@ -39,63 +43,90 @@ private IPlacementDirector ResolveDirector(PlacementStrategy strategy)
...
+			IActivationSelector result;
+			var strategyType = strategy.GetType();
+			if (!this.selectors.TryGetValue(strategyType, out result) && addIfDoesNotExist)
+			{
+				var directorType = typeof(IActivationSelector<>).MakeGenericType(strategyType);
+				result = (IActivationSelector)this.serviceProvider.GetRequiredService(directorType);
+				this.selectors[strategyType] = result;
+			}
+
+			return result ?? defaultActivationSelector;
```
anjaneya17: Not very familiar with the code. But basic question: In line number 54 result is assigned to this.selectors[strategyType]. In line 57 if result is null --- defaultActivationSelector is returned. Should we update the this.selectors[strategyType] to default if result is null?

sergeybykov: >Not very familiar with the code.

Me neither. :-) `result` can be null if `TryGetValue` returns `false`. That's the case when we fall back to the default  selector.

jdom:
There's several other tabs that were introduced. Do you want me to fix them? I have push access to your repo

sergeybykov:
@jdom By all means. Thank you.

jdom:
Done, I forced pushed to the `Addressed review comments` commit, since I'd like to merge without squashing

sergeybykov:
Perfect.

jdom:

PR #2934, opened 2017-04-12
Eviction strategy and fix Low traffic queue bug

xiazen:
Fix low traffic queue share pool with high traffic queue bug in EventhubQueueCache… 
-increase TimeBasedPurge frequency on EventHubQueueCache
-for EventHubQueueCache, don't purge the block when message is still valid, based on its TimePurgePredicate
-prevent EventHubQueueCache to purge if it's under pressure, to ensure its rewindability

### DIFF
--- a/src/OrleansProviders/Streams/Common/PooledCache/FixedSizeBuffer.cs
+++ b/src/OrleansProviders/Streams/Common/PooledCache/FixedSizeBuffer.cs
@@ -77,8 +77,13 @@ namespace Orleans.Providers.Streams.Common
         /// </summary>
         public override void SignalPurge()
         {
-            count = 0;
             purgeAction?.Invoke(this);
+        }
+
+        /// <inheritdoc />
+        public override void OnResetState()
+        {
+            count = 0;
             purgeAction = null;
         }
     }
--- a/src/OrleansProviders/Streams/Common/PooledCache/ICacheDataAdapter.cs
+++ b/src/OrleansProviders/Streams/Common/PooledCache/ICacheDataAdapter.cs
@@ -46,19 +46,9 @@ namespace Orleans.Providers.Streams.Common
         StreamPosition GetStreamPosition(TQueueMessage queueMessage);
 
         /// <summary>
-        /// Given a purge request, indicates if a cached message should be purged from the cache
+        /// Should be set to OnBlockAllocated method of the cache's EvicationStrategy
         /// </summary>
-        /// <param name="cachedMessage"></param>
-        /// <param name="newestCachedMessage"></param>
-        /// <param name="purgeRequest"></param>
-        /// <param name="nowUtc"></param>
-        /// <returns></returns>
-        bool ShouldPurge(ref TCachedMessage cachedMessage, ref TCachedMessage newestCachedMessage, IDisposable purgeRequest, DateTime nowUtc);
-
-        /// <summary>
-        /// Assignable purge action.  This is called when a purge request is triggered.
-        /// </summary>
-        Action<IDisposable> PurgeAction { set; }
+        Action<IDisposable> OnBlockAllocated { set; }
     }
 
     /// <summary>
--- a/src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
+++ b/src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
@@ -24,18 +24,16 @@ namespace Orleans.Providers.Streams.Common
     /// </summary>
     /// <typeparam name="TQueueMessage">Queue specific data</typeparam>
     /// <typeparam name="TCachedMessage">Tightly packed cached structure.  Should only contain value types.</typeparam>
-    public class PooledQueueCache<TQueueMessage, TCachedMessage>
+    public class PooledQueueCache<TQueueMessage, TCachedMessage>: IPurgeObservable<TCachedMessage>
         where TCachedMessage : struct
     {
         // linked list of message bocks.  First is newest.
         private readonly LinkedList<CachedMessageBlock<TCachedMessage>> messageBlocks;
-        private readonly ConcurrentQueue<IDisposable> purgeQueue;
         private readonly CachedMessagePool<TQueueMessage, TCachedMessage> pool;
         private readonly ICacheDataAdapter<TQueueMessage, TCachedMessage> cacheDataAdapter;
         private readonly ICacheDataComparer<TCachedMessage> comparer;
         private readonly Logger logger;
-        private int itemCount;
-
+        
         /// <summary>
         /// Cached message most recently added
         /// </summary>
@@ -63,11 +61,18 @@ namespace Orleans.Providers.Streams.Common
         }
 
         /// <summary>
-        /// Called with the newest item in the cache and last item purged after a cache purge has run.
-        /// For ordered reliable queues we shouldn't need to notify on every purged event, only on the last event 
-        ///   of every set of events that get purged.
+        /// Cached message count
         /// </summary>
-        public Action<TCachedMessage?,TCachedMessage?> OnPurged { get; set; }
+        public int ItemCount { get
+            {
+                int count = 0;
+                foreach (var block in this.messageBlocks)
+                {
+                    count += block.ItemCount;
+                }
+                return count;
+            }
+        }
 
         /// <summary>
         /// Pooled queue cache is a cache of message that obtains resource from a pool
@@ -93,7 +98,6 @@ namespace Orleans.Providers.Streams.Common
             this.comparer = comparer;
             this.logger = logger.GetSubLogger("messagecache", "-");
             pool = new CachedMessagePool<TQueueMessage, TCachedMessage>(cacheDataAdapter);
-            purgeQueue = new ConcurrentQueue<IDisposable>();
             messageBlocks = new LinkedList<CachedMessageBlock<TCachedMessage>>();
         }
 
@@ -288,8 +292,6 @@ namespace Orleans.Providers.Streams.Common
                 throw new ArgumentNullException("message");
             }
 
-            PerformPendingPurges(dequeueTimeUtc);
-
             StreamPosition streamPosition;
             // allocate message from pool
             CachedMessageBlock<TCachedMessage> block = pool.AllocateMessage(message, dequeueTimeUtc, out streamPosition);
@@ -297,72 +299,23 @@ namespace Orleans.Providers.Streams.Common
             // If new block, add message block to linked list
             if (block != messageBlocks.FirstOrDefault())
                 messageBlocks.AddFirst(block.Node);
-            itemCount++;
-
-            PerformPendingPurges(dequeueTimeUtc);
 
             return streamPosition;
         }
 
         /// <summary>
-        /// Record that we need to purge all messages associated with purge request in the next purge cycle
+        /// Remove oldest message in the cache, remove oldest block too if the block is empty
         /// </summary>
-        /// <param name="purgeRequest"></param>
-        public void Purge(IDisposable purgeRequest)
+        public void RemoveOldestMessage()
         {
-            purgeQueue.Enqueue(purgeRequest);
-        }
-
-        private void PerformPendingPurges(DateTime nowUtc)
-        {
-            IDisposable purgeRequest;
-            if (purgeQueue.IsEmpty) return;
-            while (purgeQueue.TryDequeue(out purgeRequest))
+            this.messageBlocks.Last.Value.Remove();
+            CachedMessageBlock<TCachedMessage> lastCachedMessageBlock = this.messageBlocks.Last.Value;
+            // if block is currently empty, but all capacity has been exausted, remove
+            if (lastCachedMessageBlock.IsEmpty && !lastCachedMessageBlock.HasCapacity)
             {
-                int currentItemCount = itemCount;
-                PerformBlockPurge(purgeRequest, nowUtc);
-                ReportBlockPurge(currentItemCount);
+                lastCachedMessageBlock.Dispose();
+                this.messageBlocks.RemoveLast();
             }
-        }
-
-        private void PerformBlockPurge(IDisposable purgeRequest, DateTime nowUtc)
-        {
-            TCachedMessage neweswtMessageInCache = messageBlocks.First.Value.NewestMessage;
-            TCachedMessage? lastMessagePurged = null;
-            while (!IsEmpty)
-            {
-                var oldestMessageInCache = messageBlocks.Last.Value.OldestMessage;
-                if (!cacheDataAdapter.ShouldPurge(ref oldestMessageInCache, ref neweswtMessageInCache, purgeRequest, nowUtc))
-                {
-                    break;
-                }
-                lastMessagePurged = oldestMessageInCache;
-                messageBlocks.Last.Value.Remove();
-                itemCount--;
-                CachedMessageBlock<TCachedMessage> lastCachedMessageBlock = messageBlocks.Last.Value;
-                // if block is currently empty, but all capacity has been exausted, remove
-                if (lastCachedMessageBlock.IsEmpty && !lastCachedMessageBlock.HasCapacity)
-                {
-                    lastCachedMessageBlock.Dispose();
-                    messageBlocks.RemoveLast();
-                }
-            }
-
-            if (lastMessagePurged.HasValue)
-            {
-                OnPurged?.Invoke(lastMessagePurged, Newest);
-            }
-
-            purgeRequest.Dispose();
-        }
-
-        private void ReportBlockPurge(int startingItemCount)
-        {
-            if (IsEmpty)
-            {
-                logger.Verbose("BlockPurged: cache empty");
-            }
-            logger.Verbose($"BlockPurged: PurgeCount: {startingItemCount - itemCount}, CacheSize: {itemCount}");
         }
 
         private enum CursorStates
--- a/src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
+++ b/src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
@@ -63,6 +63,11 @@ namespace Orleans.Providers.Streams.Generator
         private class ExplicitEvictionStrategy : IEvictionStrategy<CachedMessage>
         {
             private FixedSizeBuffer currentBuffer;
+            private Queue<FixedSizeBuffer> purgedBuffers;
+            public ExplicitEvictionStrategy()
+            {
+                this.purgedBuffers = new Queue<FixedSizeBuffer>();
+            }
             public IPurgeObservable<CachedMessage> PurgeObservable { set; private get; }
 
             public Action<CachedMessage?, CachedMessage?> OnPurged { get; set; }
@@ -86,6 +91,19 @@ namespace Orleans.Providers.Streams.Generator
                     lastMessagePurged = oldestMessageInCache;
                     this.PurgeObservable.RemoveOldestMessage();
                 }
+
+                //return purged buffer to the pool. except for the current buffer.
+                //if purgeCandidate is current buffer, put it in purgedBuffers and free it in next circle
+                var purgeCandidate = purgeRequest as FixedSizeBuffer;
+                this.purgedBuffers.Enqueue(purgeCandidate);
+                while (this.purgedBuffers.Count > 0)
+                {
+                    if (this.purgedBuffers.Peek() != this.currentBuffer)
+                    {
+                        this.purgedBuffers.Dequeue().Dispose();
+                    }
+                    else { break; }
+                }  
             }
 
             public void OnBlockAllocated(IDisposable newBlock)
@@ -98,11 +116,6 @@ namespace Orleans.Providers.Streams.Generator
             private bool ShouldPurge(ref CachedMessage cachedMessage, ref CachedMessage newestCachedMessage, IDisposable purgeRequest)
             {
                 var purgedResource = (FixedSizeBuffer)purgeRequest;
-                // if we're purging our current buffer, don't use it any more
-                if (this.currentBuffer != null && this.currentBuffer.Id == purgedResource.Id)
-                {
-                    currentBuffer = null;
-                }
                 return cachedMessage.Payload.Array == purgedResource.Id;
             }
 
--- a/src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
+++ b/src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
@@ -71,12 +71,12 @@ namespace Orleans.Providers.Streams.Generator
             public void PerformPurge(DateTime utcNow, IDisposable purgeRequest)
             {
                 //if the cache is empty, then nothing to purge, return
-                if (this.PurgeObservable.ItemCount == 0)
+                if (this.PurgeObservable.IsEmpty)
                     return;
                 var itemCountBeforePurge = this.PurgeObservable.ItemCount;
                 CachedMessage neweswtMessageInCache = this.PurgeObservable.Newest.Value;
                 CachedMessage? lastMessagePurged = null;
-                while (this.PurgeObservable.ItemCount != 0)
+                while (!this.PurgeObservable.IsEmpty)
                 {
                     var oldestMessageInCache = this.PurgeObservable.Oldest.Value;
                     if (!ShouldPurge(ref oldestMessageInCache, ref neweswtMessageInCache, purgeRequest))
--- a/src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
+++ b/src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
@@ -79,7 +79,7 @@ namespace Orleans.Providers.Streams.Generator
                 while (this.PurgeObservable.ItemCount != 0)
                 {
                     var oldestMessageInCache = this.PurgeObservable.Oldest.Value;
-                    if (ShouldPurge(ref oldestMessageInCache, ref neweswtMessageInCache, purgeRequest))
+                    if (!ShouldPurge(ref oldestMessageInCache, ref neweswtMessageInCache, purgeRequest))
                     {
                         break;
                     }
--- a/src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
+++ b/src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
@@ -67,30 +67,8 @@ namespace Orleans.Providers.Streams.Generator
 
             public Action<CachedMessage?, CachedMessage?> OnPurged { get; set; }
 
-            public void PerformPurge(DateTime utcNow)
-            {
-                //do nothing, purge in GeneratorPooledCache is triggered by block pool and conduct by OnFreeBlockRequest method
-            }
-
-            public void OnBlockAllocated(IDisposable newBlock)
-            {
-                var newBuffer = newBlock as FixedSizeBuffer;
-                this.currentBuffer = newBuffer;
-                this.currentBuffer.SetPurgeAction(this.OnFreeBlockRequest);
-            }
-
-            private bool ShouldPurge(ref CachedMessage cachedMessage, ref CachedMessage newestCachedMessage, IDisposable purgeRequest)
-            {
-                var purgedResource = (FixedSizeBuffer)purgeRequest;
-                // if we're purging our current buffer, don't use it any more
-                if (this.currentBuffer != null && this.currentBuffer.Id == purgedResource.Id)
-                {
-                    currentBuffer = null;
-                }
-                return cachedMessage.Payload.Array == purgedResource.Id;
-            }
-
-            private void OnFreeBlockRequest(IDisposable purgeRequest)
+            //Explicitly purge all messages in purgeRequestBlock
+            public void PerformPurge(DateTime utcNow, IDisposable purgeRequest)
             {
                 //if the cache is empty, then nothing to purge, return
                 if (this.PurgeObservable.ItemCount == 0)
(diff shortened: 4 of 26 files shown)


xiazen:

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeBuffer.cs
```diff
@@ -69,17 +75,24 @@ public bool TryGetSegment(int size, out ArraySegment<byte> value)
...
-        /// Reset state and calls purge with its buffer, indicating that any segments using that address are no longer valid.
+        /// Trigger purge action on purge cancdidate buffer
         /// </summary>
-        public override void SignalPurge()
+        public override void SignalPurge(IDisposable purgeCandidate)
         {
-            count = 0;
-            purgeAction?.Invoke(this);
-            purgeAction = null;
+            purgeAction?.Invoke(purgeCandidate);
```
jason-bragg: please keep purgeAction = null for safety.
Under no circumstances should purge action ever be called more than once.

jdom:

src/Orleans/Properties/AssemblyInfo.cs
```diff
@@ -44,3 +44,4 @@
 [assembly: InternalsVisibleTo("Tester.SQLUtils")]
 [assembly: InternalsVisibleTo("Tester.ZooKeeperUtils")]
 [assembly: InternalsVisibleTo("AWSUtils.Tests")]
+[assembly: InternalsVisibleTo("ServiceBus.Tests")]
```
jdom: Is this absolutely needed? One of @jason-bragg's goals when writing this provider was to make sure this adaptor could in theory be built outside Orleans repo entirely.

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeBuffer.cs
```diff
@@ -69,17 +75,24 @@ public bool TryGetSegment(int size, out ArraySegment<byte> value)
             }
             value = new ArraySegment<byte>(buffer, count, size);
             count += size;
+            this.SegmentCount++;
             return true;
         }
 
         /// <summary>
-        /// Reset state and calls purge with its buffer, indicating that any segments using that address are no longer valid.
+        /// Trigger purge action on purge cancdidate buffer
```
jdom: nit: typo in candidate

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeObjectPool.cs
```diff
@@ -52,16 +52,25 @@ public override T Allocate()
             T obj;
             lock (locker)
             {
-                // if we've used all we are allowed, signal object it has been purged and should be returned to the pool
-                if (usedObjects.Count >= maxObjectCount)
+                // just send signal to EventhubQueueCaches to ask them do a TimeBasedPurge
```
jdom: nit: unless this is no longer generally usable, the comment about EventHub doesn't apply here, since it might be used for other purposes (in theory)

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeObjectPool.cs
```diff
@@ -75,8 +84,25 @@ public override void Free(T resource)
         {
             lock (locker)
             {
+                if(this.usedObjects.Contains(resource))
+                    this.usedObjects.Remove(resource);
```
jdom: Contains enumerates, and Remove enumerates once again. Just Remove directly without checking if it's there

src/OrleansProviders/Streams/Common/PooledCache/ICacheDataAdapter.cs
```diff
@@ -59,6 +60,24 @@ public interface ICacheDataAdapter<in TQueueMessage, TCachedMessage>
...
+        /// Handler for when a cachedMessage got purged
+        /// </summary>
+        /// <param name="cachedMessage"></param>
+        void SignalMessagePurged(ref TCachedMessage cachedMessage);
+
+        /// <summary>
+        /// Determine action to make on a purgeRequest
+        /// </summary>
+        /// <param name="purgeRequest"></param>
+        void ConsultPurgeRequest(IDisposable purgeRequest);
```
jdom: I have absolutely zero idea of what this is supposed to do, by looking at the method name or comments.

src/OrleansProviders/Streams/Common/PooledCache/ICacheDataAdapter.cs
```diff
@@ -59,6 +60,24 @@ public interface ICacheDataAdapter<in TQueueMessage, TCachedMessage>
...
+
+        /// <summary>
+        /// CachePressureMonitor used on its cache
+        /// </summary>
+        ICachePressureMonitor CachePressureMonitor { set; }
```
jdom: should CachePressureMonitor and PurgeAction both be initialized in tandem? If so, consider creating a method that initializes both

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
```diff
@@ -23,6 +23,31 @@ public interface IObjectPool<T>
...
         /// <param name="resource"></param>
         void Free(T resource);
+
+        /// <summary>
+        /// Try to find in-use object in this pool by Id 
+        /// </summary>
+        /// <param name="Id"></param>
+        /// <param name="inUseObject"></param>
+        /// <returns></returns>
+        bool TryFindInUseObject(object Id, out T inUseObject);
```
jdom: nit: casing of Id.
Also please add code comments for parameters too, especially since these are really cryptic

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
```diff
@@ -23,6 +23,31 @@ public interface IObjectPool<T>
...
+    }
+
+    internal static class PurgeRequest
+    {
+        /// <summary>
+        /// Empty purge request which is just to inform cache with TimeBasedPurgePredicate to do 
+        /// a time based purge
+        /// </summary>
+        public static IDisposable EmptyPurgeRequest = EmptyDisposable.Instance.Value;
```
jdom: readonly

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
```diff
@@ -23,6 +23,31 @@ public interface IObjectPool<T>
...
+
+        public class EmptyDisposable : IDisposable
+        {
+            public readonly static Lazy<EmptyDisposable> Instance =
```
jdom: what's the purpose of having a Lazy here, if you then have a static field that reads the value immediately?

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
```diff
@@ -41,12 +66,18 @@ public abstract class PooledResource<T> : IDisposable
...
+        /// <summary>
         /// If this object is to be used in a fixed size object pool, this call should be
         ///   overridden with the purge implementation that returns the object to the pool.
         /// </summary>
-        public virtual void SignalPurge()
+        public virtual void SignalPurge(IDisposable purgeCandidate)
         {
-            Dispose();
+            if(purgeCandidate == this)
+                Dispose();
```
jdom: and if it's not  equal to this, should it do a no-op?

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
```diff
@@ -41,12 +66,18 @@ public abstract class PooledResource<T> : IDisposable
...
+
+        /// <summary>
         /// If this object is to be used in a fixed size object pool, this call should be
         ///   overridden with the purge implementation that returns the object to the pool.
         /// </summary>
-        public virtual void SignalPurge()
+        public virtual void SignalPurge(IDisposable purgeCandidate)
         {
-            Dispose();
+            if(purgeCandidate == this)
```
jdom: should you be using `ReferenceEquals` instead?

src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
```diff
@@ -131,8 +133,22 @@ public StreamPosition GetStreamPosition(GeneratedBatchContainer queueMessage)
                 return new StreamPosition(new StreamIdentity(queueMessage.StreamGuid, queueMessage.StreamNamespace), queueMessage.RealToken);
             }
 
+#region purge constulting related logic
```
jdom: nit: typo

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
```diff
@@ -267,36 +270,6 @@ public virtual StreamPosition GetStreamPosition(EventData queueMessage)
...
-
-            return ShouldPurgeFromResource(ref cachedMessage, purgedResource) || timePurage.ShouldPurgFromTime(timeInCache, relativeAge);
-        }
-
-        private static bool ShouldPurgeFromResource(ref CachedEventHubMessage cachedMessage, FixedSizeBuffer purgedResource)
-        {
-            // if message is from this resource, purge
-            return cachedMessage.Segment.Array == purgedResource.Id;
-        }
-
```
jdom: why this block code move? it's hard to diff. Are there any changes to this code?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubStreamProviderSettings.cs
```diff
@@ -69,7 +69,12 @@ public class EventHubStreamProviderSettings
         /// CacheSizeMb setting name.
         /// </summary>
         public const string CacheSizeMbName = "CacheSizeMb";
-        private const int DefaultCacheSizeMb = 128; // default to 128mb cache.
+
+        /// <summary>
+        /// DefaultCacheSizeMb
+        /// </summary>
+        public readonly static int DefaultCacheSizeMb = 128; // default to 128mb cache.
```
jdom: what was the issue with this being a const?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubStreamProviderSettings.cs
```diff
@@ -84,7 +89,12 @@ public int CacheSizeMb
         /// DataMinTimeInCache setting name.
         /// </summary>
         public const string DataMinTimeInCacheName = "DataMinTimeInCache";
-        private static readonly TimeSpan DefaultDataMinTimeInCache = TimeSpan.FromMinutes(5);
+
+        /// <summary>
+        /// DefaultDataMinTimeInCache
```
jdom: not very helpful

src/Orleans/Properties/AssemblyInfo.cs
xiazen: That's needed because I need to init the serialization env for the tests, specifically to BufferPool.InitGlobalBufferPool(new MessagingConfiguration(false));  , which is used by dataAdapter to serialize queue message into cache message body bits. And BufferPool.InitGlobalBufferPool(new MessagingConfiguration(false)); is an internal method.

xiazen: BufferPool is internal , so I kind of need this

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeObjectPool.cs
xiazen: this is generally usable, just Eventhub cache will handle this in its own manner. and other cache will handle it as before. But I agree the comments can be better. But I kind want comment here to explain to how EventHUb will handle this in its own manner to reduce confusion for other engineers to look at

src/OrleansProviders/Streams/Common/PooledCache/ICacheDataAdapter.cs
xiazen: it means, looking at this purgeRequest, make decision about it. The decision has two options currently, one is ignore it(the purge request is handled in other manner), second is dispose the purge request (which is dispose the block represented by the purge request)

src/OrleansProviders/Streams/Common/PooledCache/ICacheDataAdapter.cs
xiazen: yes, I definitely considered encapsulate those purge related methods into one class.

but the Typed param in ICacheDataAdapter is not making this direction super pretty. But I can definitely improve this by thinking of other alternatives or improvements.

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
xiazen: You prefer ID, over Id?

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
xiazen: hmm, you are correct.... I was .... not thinking of how static works  on this....

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
xiazen: == will do that in this circumstance, but ReferenceEquals is definitely safer. will do

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
xiazen: no-op? what is that? do nothing? it is already do nothing when not equal

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
xiazen: it's been removed. this is the part which caused the low traffic queue bug. 
this code basically tell: no matter whether this message is old enough to be purged by TimeBasedPurgePredicate or not, purge it as long as it belongs to this block.  While the last block in block pool doesn't mean all messages on it should be time based purge.

src/Orleans/Properties/AssemblyInfo.cs
jason-bragg: What happens if InitGlobalBufferPool is not called?
If we need test setup of the buffer pool, I suggest we add it to OrleansTestingHost, so it's available to all test harnesses without the need to add internal access.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubDataAdapter.cs
xiazen: just to be clear, it's ShouldPurgeFromResource func got removed. this is not a valid ShouldPurge logic for TimeBasedPrgePredicate.

src/OrleansProviders/Streams/Common/PooledCache/ICacheDataAdapter.cs
jdom: just to be clear, I'm not suggesting a separate class, just a single method to initialize the 2 things that seems to be initialized in tandem. I didn't originally like a set-only method, but less if there's multiple setters that need to be called just to have the object fully initialized

src/OrleansProviders/Streams/Common/PooledCache/ICacheDataAdapter.cs
jdom: well, the point is not just to get an answer in this thread, but to improve either the method name to make it less ambiguous or at least make the code comment clearer (I'd prefer both, although I understand coming up with a good name is also hard)

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
jdom: no, I meant parameters need to be camel case. so in this case `id`

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
jdom: correct, that's what no-op means. So I was asking whether that was by design. Because if it breaks an invariant, then it's safer to throw than no-op.

xiazen:
discussed offline with Jason, decide to add more improvement to cache system in this pass as related change. 
add work-in-progress tag

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
```diff
@@ -63,11 +61,18 @@ public class PooledQueueCache<TQueueMessage, TCachedMessage>
         }
 
         /// <summary>
-        /// Called with the newest item in the cache and last item purged after a cache purge has run.
-        /// For ordered reliable queues we shouldn't need to notify on every purged event, only on the last event 
-        ///   of every set of events that get purged.
+        /// Cached message count
         /// </summary>
-        public Action<TCachedMessage?,TCachedMessage?> OnPurged { get; set; }
+        public int ItemCount { get
```
jason-bragg: This is relatively expensive, and doe not look necessary.  All we need to know is if the cache is empty, correct?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCacheEvictionStrategy.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// <inheritdoc />
+        public void OnBlockAllocated(IDisposable newBlock)
+        {
+            var newBuffer = newBlock as FixedSizeBuffer;
+            this.inUseBuffers.Enqueue(newBuffer);
+            newBuffer.SetPurgeAction(this.OnFreeBlockRequest);
+        }
+
+        /// <inheritdoc />
+        public void PerformPurge(DateTime nowUtc)
```
jason-bragg: how does this get called?

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
xiazen: fixed ! as we discussed

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeBuffer.cs
```diff
@@ -77,8 +77,13 @@ public bool TryGetSegment(int size, out ArraySegment<byte> value)
         /// </summary>
         public override void SignalPurge()
         {
-            count = 0;
             purgeAction?.Invoke(this);
+        }
+
+        /// <inheritdoc />
+        public override void OnResetState()
```
jason-bragg: why does this file need to change?

src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
```diff
@@ -58,13 +60,65 @@ public bool Equals(CachedMessage cachedMessage, IStreamIdentity streamIdentity)
             }
         }
 
+        private class ExplicitEvictionStrategy : IEvictionStrategy<CachedMessage>
```
jason-bragg: This looks like the same logic as ExplicitEvictionStrategy in memory provider.  If so we should make common, not duplicate.

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeBuffer.cs
xiazen: because now SignalPurge doesn't mean the buffer will be returning to the pool, so reset state in SignalPurge is not valid anymore. I need to clean it up in FixedSizeBuffer.Dispose method

src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
xiazen: I thought about that. but typed param made it harder to make common. Since we need to know what type of CachedMessage is to know its buffer ID

jason-bragg: Ah,  I see.  We can look at reducing code duplication later if/when needed.  Shouldn't block this PR.

src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
```diff
@@ -58,13 +60,65 @@ public bool Equals(CachedMessage cachedMessage, IStreamIdentity streamIdentity)
...
+
+            public Action<CachedMessage?, CachedMessage?> OnPurged { get; set; }
+
+            //Explicitly purge all messages in purgeRequestBlock
+            public void PerformPurge(DateTime utcNow, IDisposable purgeRequest)
+            {
+                //if the cache is empty, then nothing to purge, return
+                if (this.PurgeObservable.ItemCount == 0)
+                    return;
+                var itemCountBeforePurge = this.PurgeObservable.ItemCount;
```
jason-bragg: itemCountBeforePurge not used right? Here or in memory version

src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
```diff
@@ -58,13 +60,65 @@ public bool Equals(CachedMessage cachedMessage, IStreamIdentity streamIdentity)
...
+                CachedMessage neweswtMessageInCache = this.PurgeObservable.Newest.Value;
+                CachedMessage? lastMessagePurged = null;
+                while (this.PurgeObservable.ItemCount != 0)
```
jason-bragg: !Empty?

src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
```diff
@@ -99,7 +153,10 @@ private ArraySegment<byte> SerializeMessageIntoPooledSegment(GeneratedBatchConta
                 {
                     // no block or block full, get new block and try again
                     currentBuffer = bufferPool.Allocate();
-                    currentBuffer.SetPurgeAction(PurgeAction);
+                    if (this.OnBlockAllocated == null)
+                        throw new OrleansException("Eviction strategy's OnBlockAllocated is not set for current data adapter, this will affect cache purging");
```
jason-bragg: I'm not sure we should consider this an error at this time.  Remember that queues can be shut down and moved between silos.  Until we've tested that this new code behaves well in those scenarios, let's not assume that data adapters -must- always be wired up to eviction strategies.  The risk here is only that a block does not return to the pool, which is not good, whereas if we throw it could lead to data loss.  Logging a warning would be better than a throw here, if we're concerned about this possibility.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCacheEvictionStrategy.cs
```diff
@@ -0,0 +1,142 @@
...
+                {
+                    break;
+                }
+                lastMessagePurged = oldestMessageInCache;
+                itemsPurged++;
+                this.PurgeObservable.RemoveOldestMessage();
+            }
+            var itemCountAfterPurge = itemCountBeforePurge - itemsPurged;
+
+            //purge finished, time to conduct follow up actions
```
jason-bragg: these 'follow up' actions should only be performed if itemsPurged != 0, correct?

jason-bragg: which means we only need to compute itemCountBeforePurge if itemsPurged != 0.

jason-bragg: in fact itemCountBeforePurge and itemCountAfterPurge only need to be calculated in ReportPurge and only if log level is verbose, correct?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCacheEvictionStrategy.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        private void OnFreeBlockRequest(IDisposable block)
+        {
+            var purgeCandidate = block as FixedSizeBuffer;
+            //free all blocks before purgeCandidate and purgeCandidate
+            if (this.purgedBuffers.Contains(purgeCandidate))
+            {
+                while (this.purgedBuffers.Peek() != purgeCandidate)
+                    this.purgedBuffers.Dequeue().Dispose();
```
jason-bragg: shouldn't we dequeue and dispose every buffer and break after we dispose the purgeCandidate?  This logic will leave the purgeCandidate in the queue, which will double dispose it.

test/TesterInternal/OrleansRuntime/Streams/PooledQueueCacheTests.cs
```diff
@@ -77,12 +77,64 @@ public bool Equals(TestCachedMessage cachedMessage, IStreamIdentity streamIdenti
...
+            private FixedSizeBuffer currentBuffer;
+            public IPurgeObservable<TestCachedMessage> PurgeObservable { set; private get; }
+
+            public Action<TestCachedMessage?, TestCachedMessage?> OnPurged { get; set; }
+
+            //Explicitly purge all messages in purgeRequestBlock
+            public void PerformPurge(DateTime utcNow, IDisposable purgeRequest)
+            {
+                //if the cache is empty, then nothing to purge, return
+                if (this.PurgeObservable.ItemCount == 0)
```
jason-bragg: !empty?

test/TesterInternal/OrleansRuntime/Streams/PooledQueueCacheTests.cs
```diff
@@ -77,12 +77,64 @@ public bool Equals(TestCachedMessage cachedMessage, IStreamIdentity streamIdenti
...
+                    return;
+                var itemCountBeforePurge = this.PurgeObservable.ItemCount;
+                TestCachedMessage neweswtMessageInCache = this.PurgeObservable.Newest.Value;
+                TestCachedMessage? lastMessagePurged = null;
+                while (this.PurgeObservable.ItemCount != 0)
```
jason-bragg: !empty?

src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
xiazen: yes, I guess I just remember to clean that logic in EHCache. forgot other cache.

src/OrleansProviders/Streams/Generator/GeneratorPooledCache.cs
xiazen: this is definitely a valid concern. But my concern is, onBlockAllocated only got set once, if messed up there, all newly allocated buffer will lose the hook to return to the pool. So the memory will be eaten up eventually. Basically fail early vs fail later. 

Maybe a little explain on why queue shifting will affect eviction strategy set up will help. eviction strategy is a dependency of cache creation. EHQueueCache take a IEvicationStrategy param in constructor. In what scenario will eviction strategy exists without the cache?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCacheEvictionStrategy.cs
xiazen: hmm itemCountBeforePurge is a snapshot of the count. We need to store it before actual purge performing, which is how we know if any items got purged right? How do I predict the future and decide to compute itemCountBeforePurge or not?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCacheEvictionStrategy.cs
xiazen: You are right , good catch !

xiazen:

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCacheEvictionStrategy.cs
```diff
@@ -125,18 +129,18 @@ private bool ShouldPurge(ref CachedEventHubMessage cachedMessage, ref CachedEven
...
             if (this.purgedBuffers.Contains(purgeCandidate))
             {
                 while (true)
                 {
-                    var purgedBuffer = this.purgedBuffers.Dequeue();
-                    if (purgedBuffer == purgeCandidate)
-                    {
-                        purgedBuffer.Dispose();
+                    var purgedBuffer = this.purgedBuffers.Peek();
+                    if (purgedBuffer == this.currentBuffer)
```
xiazen: @jason-bragg  went with your suggestion, since my approach will introduce bug in some extreme rare cases

xiazen:
poke poke

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubCacheEvictionStrategy.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            //if nothing purged, then no buffer was purged
+            //if items got purged, then potencially some buffer was purged
+            if (itemsGotPurged)
+            {
+                var IdOfLastPurgedBuffer = lastMessagePurged.Value.Segment.Array;
+                // IdOfLastBufferInCache will be null if cache is empty after purge
+                var IdOfLastBufferInCache = oldestMessageInCache.HasValue ? oldestMessageInCache.Value.Segment.Array : null;
+                //all buffer older than LastPurgedBuffer should be purged 
+                while (this.inUseBuffers.Peek().Id != IdOfLastPurgedBuffer)
```
xiazen: it is not possible here for inUseBUffers to be empty, since itemsGotPurged is true here. If there's items got purged, then definitely inUseBufffers is not empty. 
But you are right, for maintenance, we should have a empty check here

xiazen:

xiazen:
@dotnet-bot test this please

PR #2941, opened 2017-04-14
Remove [RegisterSerializer] and support for correspondig Register method

### LINKED ISSUE #2610: Remove SerializationManager.Register method
Currently, when `SerializationMananger` is scanning types in `FindSerializationInfo`, it checks types for the `[RegisterSerializer]` attribute and calls their `Register` method if one is found.

Since `Register` is a static, parameterless method, implementations currently call into the static `SerializationManager.Register` method to register serialization delegates.

With `SerializationManager` becoming non-static, a change needs to be made here and I see two options:
1. Remove `[RegisterSerializer]` and `Register` altogether and instead rely on `[Serializer(typeof(TargetType))]` attributes like we do in our generated code.
1. Change the signature of the `Register` method so that it takes a parameter of type `SerializationManager` or some purpose-built interface like so:
```C#
public interface ISerializerRegistrar
{
    void Register(
        Type type,
        SerializationManager.DeepCopier copier,
        SerializationManager.Serializer serializer,
        SerializationManager.Deserializer deserializer);
}
```

I'm in favor of the 1st option. What do you think?

ReubenBond:
Removes the deprecated & obsolete `[RegisterSerializer]` attribute and support for the corresponding `Register` method.

Also fixed the formatting in GrainAttributes.cs since I dislike always having to undo the auto-formatting whenever I make a change.

Fixes #2610

### DIFF
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -296,16 +296,6 @@ namespace Orleans
         public sealed class DeserializerMethodAttribute : Attribute
         {
         }
-
-        /// <summary>
-        /// Used to make a class for auto-registration as a serialization helper.
-        /// </summary>
-        [AttributeUsage(AttributeTargets.Class)]
-        [Obsolete(
-            "[RegisterSerializer] is obsolete, please use [Serializer(typeof(TargetType))] instead. Note that the signature of Register has changed to 'void Register(SerializationManager sm)'.")]
-        public sealed class RegisterSerializerAttribute : Attribute
-        {
-        }
     }
 
     namespace Providers
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -181,18 +181,20 @@ namespace Orleans
         {
             public RandomPlacementAttribute() :
                 base(RandomPlacement.Singleton)
-            { }
+            {
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


jdom:
Looks like we implemented option 2 of #2610 (and didn't release it yet, as in wasn't in 1.4.1) but now we are going all the way with option 1. Is my understanding correct?

ReubenBond:
@jdom the attribute is marked as Obsolete in v1.4.1: https://github.com/dotnet/orleans/blob/v1.4.1/src/Orleans/Core/GrainAttributes.cs#L288 

We:
* Obviated the `Register` method via the `[Serializer(...)]` attribute (second part of option 1)
* Obsoleted `[RegisterSerializer]` 
* Allowed `Register` to still work for the time being, but with a necessary modification to the signature < maybe we shouldn't have done this and made the obsoletion message an error instead of a warning.

So now we're doing the first part of option 1.

jdom:
Sure, that makes sense, I was just saying that this part:
> Allowed Register to still work for the time being, but with a necessary modification to the signature < maybe we shouldn't have done this and made the obsoletion message an error instead of a warning.

was actually never released. If you look at the [code in v1.4.1](https://github.com/dotnet/orleans/blob/v1.4.1/src/Orleans/Serialization/SerializationManager.cs#L617), it was still looking for a parameterless `Register` method. I wonder why we didn't go all the way with option 1, since by looking at the issue, it was agreed in January. Still, looks good, merging, I just wanted to understand

ReubenBond:
1.4.1 still had static `SerializationManager` which means registration didn't need to break.

PR #2942, opened 2017-04-14
remove TestEnvironmentFixture annotation from EH tests

xiazen:
TestEnvironmentFixture might interfere with OrleansTestingBase. 
It maybe the reason causing EHSlowConsumingTests.cs  to fall intermittently. Since removing it make EHSlowConsumingTests.cs  succeed 6 times in a row in VSO(main and vNext). I used it by mimic other EH tests. 

Not sure if other EH tests needs to use TestEnvironmentFixture . But TestEnvironmentFixture is designed to set up serialization env for non-silo tests. So I removed TestEnvironmentFixture  for other EH tests too. Stop me if they actually need it.


ReubenBond:
I don't see why it would cause the failures, but that attribute shouldn't be there anyway. I think this was my fault.

Issue #2167, opened 2016-09-16
Suggestion: uniform hash based grain placement policy


Eldar1205:
I’d like to add a new grain placement policy that places grains on silos based on their uniform hash, same way reminders are balanced. The placement policy should allow to specify per grain type whether to force grain activations re-balancing on silo join/leave or not, alternatively instead of re-balancing activations a simpler route could be taken to force deactivation of grains that should be re-balanced, and let them be re-activated by a message an the placement policy will put them in their new silo. 

The motivation for this is using Orleans reminders as a way to implement evenly distributed scheduled jobs executions, where grains are the job executors that are kept **always active** so state could be preserved between consequent job executions in memory, and if needed also persisted. Reminders are evenly distributed between silos based on associated grains uniform hash, but the grains they activate aren't necessarily on same silo and if the grains are always active then when new silos join they get their share of the reminders, but not their share of the grain activations doing the actual work. Having the grains forcibly balanced based on their uniform hash will achieve the necessary elasticity of jobs execution, and will also be more efficient since there's no TCP messaging between reminders and their grains - both are placed on silos based on the grain's uniform hash. I thought about using stateless workers with reminders for that purpose, but this has several disadvantages:
1. After reminders are re-balanced there'll be a time window where the stateless worker is activated on two silos (at least) - the old and new owners of the reminder. Depending on the job, this may be undesirable, and for the specific scenario I have it's highly undesirable.
2. Stateless workers can't leverage Orleans persistence ignorance for grains, which is amazingly useful.
3. Grains that aren't stateless workers could also receive messages from other grains, such as commands to change job execution parameters or stop job entirely. With stateless workers it won't work as expected if the message sender isn't on same silo as the stateless worker.

I like to have your assistance with the following:
1. Explanation on how to contribute to Orleans. I couldn’t infer a step-by-step process on how to start contributing, and I explored the Orleans contribution guide. I know my way around Git, but have never contributed to project on Github, so if there are links I should follow and/or some conventions/methodologies specific to Orleans, I’ll appreciate if you could share.
2. Suggestions on how to approach the implementation, in particular which existing classes/interfaces I should look at for reference since I'll need to use them or to get ideas on how implement the new policy.
3. Suggestions on how to test this contribution, in particular which existing tests I should look at for reference. 
4. Suggestions on how I should document the new placement policy's behaviour and usage scenarios.

[...]

PR #2944, opened 2017-04-16
Add support for hash-based grain placement

sergeybykov:
Add `HashBasedPlacement` placement strategy.
Add `HashBasedPlacementAttribute` for assigning `HashBasedPlacement` to grain classes.
Expose `GetUniformHashCode` via `IGrainIdentity`.
Add `HashBasedPlacementDirector` that leverages `GetUniformHashCode` to calculate hashes of `GrainIdentity.IdentityString` of target grains.

This partly implements #2167, and provides a way to collocate grains of different classes with same keys.

### DIFF
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -191,19 +191,9 @@ namespace Orleans
         [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
         public sealed class HashBasedPlacementAttribute : PlacementAttribute
         {
-            // for testing
-            internal bool SortSiloList { get; }
-
-            public HashBasedPlacementAttribute() :
+			public HashBasedPlacementAttribute() :
                 base(HashBasedPlacement.Singleton)
             {}
-
-            // for testing
-            internal HashBasedPlacementAttribute(bool sortSiloList = false)
-                : base(sortSiloList ? new HashBasedPlacement(true) : HashBasedPlacement.Singleton)
-            {
-                this.SortSiloList = sortSiloList;
-            }
         }
 
         /// <summary>
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -186,6 +186,27 @@ namespace Orleans
         }
 
         /// <summary>
+        /// Marks a grain class as using the <c>HashBasedPlacement</c> policy.
+        /// </summary>
+        [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+        public sealed class HashBasedPlacementAttribute : PlacementAttribute
+        {
+            // for testing
+            internal bool SortSiloList { get; }
+
+            public HashBasedPlacementAttribute() :
+                base(HashBasedPlacement.Singleton)
+            {}
+
+            // for testing
+            internal HashBasedPlacementAttribute(bool sortSiloList = false)
+                : base(sortSiloList ? new HashBasedPlacement(true) : HashBasedPlacement.Singleton)
+            {
+                this.SortSiloList = sortSiloList;
+            }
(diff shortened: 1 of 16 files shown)


sergeybykov:
@dotnet-bot test this please.

src/Orleans/Core/GrainAttributes.cs
```diff
@@ -186,6 +186,27 @@ public sealed class RandomPlacementAttribute : PlacementAttribute
...
+        {
+            // for testing
+            internal bool SortSiloList { get; }
+
+            public HashBasedPlacementAttribute() :
+                base(HashBasedPlacement.Singleton)
+            {}
+
+            // for testing
+            internal HashBasedPlacementAttribute(bool sortSiloList = false)
```
ReubenBond: If silos are not sorted, each silo may determine that grain X belongs to a different silo, which in my view goes against the idea of using hash-based placement (may as well use random). Or is there a natural sort order which all silos converge on but which we cannot easily determine in our tests?

I believe that the placement used here should leverage the consistent ring balancer so that placement decisions converge and so that each grain is collocated with any grain services responsible for it.

sergeybykov:
@dotnet-bot test netfx

sergeybykov:
@dotnet-bot test netfx bvt

src/Orleans/Core/GrainAttributes.cs
sergeybykov: It's `GrainTypeManager.GetSupportedSilos` that caches results.

ReubenBond: Sounds good to me.

ReubenBond:
@dotnet-bot test this

ReubenBond:
@dotnet-bot test this please

sergeybykov:
#2913 will guarantee stable converging sorted silos lists returned by `GetCompatibleSilos` it is a dependency for this PR.

test/TestInternalGrains/HashBasedPlacementGrain.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Orleans.Placement;
+
+namespace UnitTests.Grains
+{
+    [HashBasedPlacement(true)]
```
jdom: since your only test is using `sortSiloList: true` (which is the default), should you kill the code that adds internal members and confusing branches for the case that you are not sorting the lists?

jdom: Nevermind, it looks like passing `true` is actually not the default. In any case, is this still valid? Is it possible to have tests without any test-specific branching logic for this?

sergeybykov: I was going to check if I can remove the test-only sorting now. I might have to change the test grains to return `SiloAddress` instead of a string to make sure client sorts silos the same way as they do themselves.

test/Tester/Placement/CustomPlacementTests.cs
```diff
@@ -40,6 +41,7 @@ public CustomPlacementTests(Fixture fixture)
 
             // sort silo IDs into an array
             this.silos = fixture.HostedCluster.GetActiveSilos().Select(h => h.SiloAddress.ToString()).OrderBy(s => s).ToArray();
+            this.siloAddresses = fixture.HostedCluster.GetActiveSilos().Select(h => h.SiloAddress).OrderBy(s => s.ToString()).ToArray();
```
benjaminpetit: NIT: You don't need to call `.ToString()` since `SiloAddress` implements `IComparable`

sergeybykov: `SiloAddress.CompareTo(SiloAddress)` performs comparison is an somewhat odd way that I found not good for testing - it compares generations first.

PR #2946, opened 2017-04-16
Improvements in Observers docs.

AwsomeCode:
Variable name mismatch and return type was not define for Subscribe & UnSubscribe methods


sergeybykov:
Thank you, @AwsomeCode!

PR #2947, opened 2017-04-16
Grammer fix

Stackingit:
Just making it flow better so that it does not get in the way of a good read :)


sergeybykov:
Thank you, @Stackingit!

Issue #2745, opened 2017-02-19
Deadlock using AdoNetStorageProvider in OrleansSqlUtils 1.4.0


Daemon6666:
I get deadlock using AdoNetStorageProvider when Silo starts.
Code from RelationalStorage in ReadAsync method:
`var ret = ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).GetAwaiter().GetResult().Item1`
I use postgresql database.
Maybe it occurs because of ActivationTaskScheduler is used

veikkoeeva:
@Daemon6666 PostgreSQL does not have a storage provider yet. If you indeed mean the storage provider, most likely the reason to your problem is that.

On a further note, it's unclear to me why it would deadlock instead of throwing an error. It should throw [here](https://github.com/dotnet/orleans/blob/master/src/OrleansSQLUtils/Storage/Provider/AdoNetStorageProvider.cs#L188). It uses `.ConfigurationAwait(false)` and at that place the `Task` has been awaited already. There might be case the PostgreSQL provider runs in the same thread as the caller (see the little fix for MySQL in the `RelationalStorage` code), but it shouldn't have bearing in this case, I believe.

Daemon6666:
I implemented postgresql scripts for grains state storage. They let me use AdoNetStorageProvider (invariant=npgsql)

veikkoeeva:
@Daemon6666 It's difficult tell what the reason could be. It seem the storage provider for SQL Server and MySQL are functional. PostgreSQL has membership provider, which is functional, and one could conclude from the PostgreSQL ADO.NET drivers work OK. Have you tried the PostgreSQL membership provider and see if it works for your?

Daemon6666:
Why did you use `GetAwaiter().GetResult().Item1` on async method instead off  `await ExecuteAsync(...);`
`            //It is certain the result is already ready here and can be collected straight away. Having a truly asynchronous result collection
            //IAsyncEnumerable<TResult> or equivalent method ought to be used. Taking the result here without async-await saves for generating
            //the async state machine.
            var ret = ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).GetAwaiter().GetResult().Item1;`

veikkoeeva:
[...]

PR #2949, opened 2017-04-16
Use non-blocking I/O in relational storage providers

Roman-Blinkov:
Hi!
While playing with PostreSQL membersip provider I found issue with provider initialization.
I prepared demo code in separate branch where you can check issue https://github.com/Roman-Blinkov/orleans/commits/PostreSQL (use \vNext\Orleans.vNext.sln and edit connection string in \Samples\HelloWorld.NetCorePostgreSQL\src\SiloHost\OrleansConfiguration.xml )

P.S. Similar but closed thread https://github.com/dotnet/orleans/issues/2745

### DIFF
--- a/src/OrleansSQLUtils/Storage/RelationalStorage.cs
+++ b/src/OrleansSQLUtils/Storage/RelationalStorage.cs
@@ -133,7 +133,7 @@ namespace Orleans.SqlUtils
         ///}).ConfigureAwait(continueOnCapturedContext: false);                
         /// </code>
         /// </example>
-        public Task<IEnumerable<TResult>> ReadAsync<TResult>(string query, Action<IDbCommand> parameterProvider, Func<IDataRecord, int, CancellationToken, Task<TResult>> selector, CancellationToken cancellationToken = default(CancellationToken), CommandBehavior commandBehavior = CommandBehavior.Default)
+        public async Task<IEnumerable<TResult>> ReadAsync<TResult>(string query, Action<IDbCommand> parameterProvider, Func<IDataRecord, int, CancellationToken, Task<TResult>> selector, CancellationToken cancellationToken = default(CancellationToken), CommandBehavior commandBehavior = CommandBehavior.Default)
         {
             //If the query is something else that is not acceptable (e.g. an empty string), there will an appropriate database exception.
             if(query == null)
@@ -146,11 +146,7 @@ namespace Orleans.SqlUtils
                 throw new ArgumentNullException("selector");
             }
 
-            //It is certain the result is already ready here and can be collected straight away. Having a truly asynchronous result collection
-            //IAsyncEnumerable<TResult> or equivalent method ought to be used. Taking the result here without async-await saves for generating
-            //the async state machine.
-            var ret = ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).GetAwaiter().GetResult().Item1;
-            return Task.FromResult(ret);
+            return (await ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).ConfigureAwait(false)).Item1;
         }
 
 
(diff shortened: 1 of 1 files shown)


sergeybykov:
@dotnet-bot test netfx-functional netstandard-win-functional please

src/OrleansSQLUtils/Storage/RelationalStorage.cs
```diff
@@ -146,11 +146,7 @@ public Task<IEnumerable<TResult>> ReadAsync<TResult>(string query, Action<IDbCom
                 throw new ArgumentNullException("selector");
             }
 
-            //It is certain the result is already ready here and can be collected straight away. Having a truly asynchronous result collection
-            //IAsyncEnumerable<TResult> or equivalent method ought to be used. Taking the result here without async-await saves for generating
-            //the async state machine.
-            var ret = ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).GetAwaiter().GetResult().Item1;
```
jdom: wow, I'm really surprised about this. Looking at the body of `ExecuteAsync` it looks like the comment here is not true. It is not certain that the result is already here. In fact, it's almost guaranteed that the result is not here, and that this code will block the thread.
@veikkoeeva do you agree with this fix, or can you shed some light on why it was like this before?

veikkoeeva: Let the fix in. I don't remember the reasoning in the other thread, but it must be wrong if this fixes the problem. :)

jdom: Well, we still have other test failures that we must sort our, but wanted your confirmation first. Maybe the new failures are because it is now inserting so much faster given that it's non-blocking (previously if it was blocking the Orleans threads, it would have been limiting concurrency).

jdom:
Good, so we all agree that this needs to be fixed, and it doesn't just affect PostgreSQL, but every relational provider, as blocking I/O is very harmful in Orleans: it limits concurrency tremendously if it happens to block some of the few important Orleans threads (it's not like blocking some non-important background thread, where it could starve resources, but in general is not going to take the silo to a halt).

I haven't investigated the test failures, but given the nature of the fix, I'd lean to think that this is similar failure as we had in #2790, but that now that we do non-blocking I/O it manifests in several other tests.
@Roman-Blinkov are you able to take a look?

jdom:
BTW, I'm confused with the title. Does this really affect membership also, or just storage providers? Or does it indirectly affect liveness (membership) because using relational storage takes the silo to a halt, so the silos are being voted dead?

jdom:
BTW, I got some IMs from @veikkoeeva which might help you troubleshoot, or at least workaround the issue in the tests themselves:

> 1) The parallel reminders tests ending up in deadlocking situation look like a problem in that the operation just flood the table. Partially the reason here is that the table design and queries are quite inefficient.
> 2) https://ci.dot.net/job/dotnet_orleans/job/master/job/netstandard-win-functional_prtest/109/testReport/junit/UnitTests.StorageTests.SQLAdapter/RelationalStoreTests/CancellationToken_SqlServer_Test/
This looks like there is a problem in how the cancellation is checked.
> 3) https://ci.dot.net/job/dotnet_orleans/job/master/job/netstandard-win-functional_prtest/109/testReport/junit/UnitTests.MembershipTests/SqlServerMembershipTableTests/MembershipTable_SqlServer_UpdateRowInParallel/ is also "normal" in the sense that the primary key guards against duplicate insertion. It would be great if this could be used also in storage table, but it doesn't seem possible to do efficiently.

> About the test problems, I'm in a rough spot to test them, unfortunately.
Oh, regarding 2), in ADO.NET storage tests there are some tests (having the word 'Huge') in them, which test streaming several megabytes large blobs to the storage. Which leads to conclude together with the error message it's about how a succesful cancellation was checked.

Roman-Blinkov:
Sorry,
I don't plan do further investigation bcz heavy load

jdom:
The original Postgre cause that initiated all this investigation is being looked at here: https://github.com/npgsql/npgsql/issues/1130
BTW, this is probably exacerbated when using some of those blocking methods in the `RelationalStorage` class, but it's not the cause of their issue (as if they were not interfering with the scheduler nor context, even blocking calls should not deadlock, just make stuff in Orleans slower). For example @lwansbrough was able to repro this without using these blocking methods in his own custom npgsql storage provider

sergeybykov:
This PR was subsumed by #3163. Closing it.

PR #2952, opened 2017-04-17
Serialize Type instances without requiring registration

ReubenBond:
Currently, Types need to be registered by `SerializationManager.FindSerializationInfo(..)` before they can be serialized over the wire.

This PR removes that requirement.

Includes a utility class, `RuntimeTypeNameFormatter` which takes a `Type` and returns a string suitable for use with `Type.GetType` which does not include assembly metadata other than assembly name.

If it's preferred, I can remove the `CachedReadConcurrentDictionary`. In my benchmarks I find that it's significantly better than either `Dictionary` + `lock` or `ConcurrentDictionary` for our use-cases.

This can be merged without squashing.

### DIFF
--- a/src/Orleans/Serialization/BuiltInTypes.cs
+++ b/src/Orleans/Serialization/BuiltInTypes.cs
@@ -9,6 +9,7 @@ using System.Reflection;
 using Orleans.CodeGeneration;
 using Orleans.Concurrency;
 using Orleans.Runtime;
+using Orleans.Utilities;
 
 namespace Orleans.Serialization
 {
@@ -1643,12 +1644,12 @@ namespace Orleans.Serialization
 
         internal static void SerializeType(object obj, ISerializationContext context, Type expected)
         {
-            context.StreamWriter.Write(((Type)obj).OrleansTypeKeyString());
+            context.StreamWriter.Write(RuntimeTypeNameFormatter.Format((Type)obj));
         }
 
         internal static object DeserializeType(Type expected, IDeserializationContext context)
         {
-            return context.SerializationManager.ResolveTypeName(context.StreamReader.ReadString());
+            return Type.GetType(context.StreamReader.ReadString());
         }
 
         internal static object CopyType(object obj, ICopyContext context)
--- a/src/Orleans/Utils/CachedReadConcurrentDictionary.cs
+++ b/src/Orleans/Utils/CachedReadConcurrentDictionary.cs
@@ -1,0 +1,203 @@
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Orleans.Utilities
+{
+    /// <summary>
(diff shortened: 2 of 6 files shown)


src/Orleans/Utils/CachedReadConcurrentDictionary.cs
```diff
@@ -0,0 +1,203 @@
...
+        /// <summary>
+        /// Approximate number of reads which did not hit the cache since it was last invalidated.
+        /// This is used as a heuristic that the dictionary is not being modified frequently with respect to the read volume.
+        /// </summary>
+        private int cacheMissReads;
+
+        /// <summary>
+        /// Cached version of <see cref="dictionary"/>.
+        /// </summary>
+        private volatile Dictionary<TKey, TValue> readCache;
```
jdom: Is there a real need for `volatile` here? Just want to reduce the complexity if it's not.
Because it looks like if you remove volatile, you still can read your writes, and if another thread wants to access it, it probably doesn't matter that the cache was stale (otherwise more attention needs to be paid to when you invalidate the cache, which in this case it looks like is after you modify it, so there are still chances that the cache is stale when read from another thread, even with the volatile).
Or am I missing some edge case?

src/Orleans/Serialization/BuiltInTypes.cs
```diff
@@ -1643,12 +1644,12 @@ internal static object CopyDateTimeOffset(object obj, ICopyContext context)
...
         internal static void SerializeType(object obj, ISerializationContext context, Type expected)
         {
-            context.StreamWriter.Write(((Type)obj).OrleansTypeKeyString());
+            context.StreamWriter.Write(RuntimeTypeNameFormatter.Format((Type)obj));
         }
 
         internal static object DeserializeType(Type expected, IDeserializationContext context)
         {
-            return context.SerializationManager.ResolveTypeName(context.StreamReader.ReadString());
+            return Type.GetType(context.StreamReader.ReadString());
```
jdom: I assume you chose to use `Type.GetType` as opposed to `TypeUtils.ResolveType` since the type is always qualified with the assembly name, hence assembly loading is not an issue?

jdom (approved):

src/Orleans/Serialization/BuiltInTypes.cs
ReubenBond: That's right

src/Orleans/Utils/CachedReadConcurrentDictionary.cs
ReubenBond: I'll remove it, it's not needed.

ReubenBond: I fixed it, squashed, and force-pushed so that this can be merged as 3 commits

ReubenBond:
@dotnet-bot test netfx-functional netstandard-win-functional please

ReubenBond:

attilah:
LGTM!

PR #2953, opened 2017-04-18
Add caching in Catalog for GetCompatibleSilos

benjaminpetit:
* Add a caching layer in `Catalog.GetCompatibleSilos` so it is not needed to call the `GrainTypeManager` if we already computed the result
* Sort the list returned by this method (needed for #2944)

### DIFF
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
@@ -154,6 +155,7 @@ namespace Orleans.Runtime
         private readonly TimeSpan maxWarningRequestProcessingTime;
         private readonly SerializationManager serializationManager;
         private readonly MultiClusterRegistrationStrategyManager multiClusterRegistrationStrategyManager;
+        private ConcurrentDictionary<Tuple<int, int, ushort>, List<SiloAddress>> compatibleSilosCache;
 
         public Catalog(
             ILocalSiloDetails localSiloDetails,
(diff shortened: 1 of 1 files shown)


src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -225,28 +228,39 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
-            IReadOnlyList<SiloAddress> silos;
-            if (target.InterfaceVersion > 0)
-            {
-                var version = GrainTypeManager.GetAvailableVersions(target.InterfaceId).Max();
-                silos = GrainTypeManager.GetSupportedSilos(typeCode, target.InterfaceId, version);
-            }
-            else
+            var key = Tuple.Create(target.GrainIdentity.TypeCode, target.InterfaceId, target.InterfaceVersion);
+
+            return this.compatibleSilosCache.GetOrAdd(key, tuple =>
```
ReubenBond: This needlessly allocates a closure to capture `this`. It would be preferable to either use `TryGet` or to allocate the closure once in the constructor and keep it in a field.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -225,28 +228,39 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
+                {
+                    silos = GrainTypeManager.GetSupportedSilos(typeCode);
+                }
 
-            var compatibleSilos = silos.Intersect(AllActiveSilos).ToList();
-            if (compatibleSilos.Count == 0)
-                throw new OrleansException($"TypeCode ${typeCode} not supported in the cluster");
+                var compatibleSilos = silos
+                    .Intersect(AllActiveSilos)
+                    .OrderBy(s => s.ToString())
```
ReubenBond: SiloAddress is already `IComparable`, so is this needed?

benjaminpetit:
I guess I need to sleep more....

This implementation cannot be reliable, since we reset the cache when `Catalog.SiloStatusChangeNotification` is called, but the `GrainTypeManager` is maybe not refreshed yet. We need to move this caching layer in the `GrainTypeManager`. It will be done in #2913

PR #2954, opened 2017-04-19
Allow Service Fabric FabricClient to be injected

### LINKED ISSUE #2940: Service Fabric Gateway Provider should accept custom FabricClient
Users need the ability to provide a customized `FabricClient` to the gateway provider so that they can configure it with certificates, endpoints, etc.

When Service Fabric support was originally integrated, we did not have any dependency injection on the client. Now we do, so we can fix this issue.

ReubenBond:
Fixes #2940.

* Check that a gateway provider is correctly configured only if it's not injected (enables injecting gateway providers)
* Configure Service Fabric support via `IClientBuilder`. The previous method was not actually valid since we didn't have client-side DI when it was implemented. It was just there as an example/stub for future use :)

There will be a separate PR which updates the samples to demonstrate this and make use of `ClientBuilder`, that relies on v1.5.0 NuGet packages (which don't exist yet)

### DIFF
--- a/src/Orleans/Messaging/GatewayProviderFactory.cs
+++ b/src/Orleans/Messaging/GatewayProviderFactory.cs
@@ -22,6 +22,8 @@ namespace Orleans.Messaging
 
         internal IGatewayListProvider CreateGatewayListProvider()
         {
+            this.cfg.CheckGatewayProviderSettings();
+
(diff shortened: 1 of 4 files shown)


ReubenBond:

PR #2955, opened 2017-04-19
Update Service Fabric sample

ReubenBond:
This relies on #2954 and Orleans 1.5.0 packages, so it will not build as-is.

Updates the Service Fabric sample to demonstrate injecting `FabricClient` as well as using `ClientBuilder`.

Also updates dependencies in that sample and makes them consistent (they were quite skewed).

Gist of the change is this:
```C#
var serviceName = new Uri("fabric:/StatelessCalculatorApp/StatelessCalculatorService");
var client = new ClientBuilder()
  .UseConfiguration(new ClientConfiguration())
  .AddServiceFabric(serviceName)
  .ConfigureServices(
    services =>
    {
      // Some deployments require a custom FabricClient, eg so that cluster
      // endpoints and certificates can be configured.
      // A pre-configured FabricClient can be injected.
      var fabricClient = new FabricClient();
      services.AddSingleton(fabricClient);
    })
  .Build();
```


jdom:
too bad you went through the trouble of updating to 1.5.0 today, but the update to 4.6.1 will change a lot of dependencies that will need to be re-aligned

ReubenBond:
That's fine, @jdom, we can let this sit until then

ReubenBond:
Updated. Some work was needed to suport .NET 4.6.1, including #2976

ReubenBond:
This is ready for review/merging

jdom:
Should it target the released 1.5.0-beta1 instead or is this fine? 
I'll merge, we can update it separately if we decide it's a good idea

ReubenBond:
Hm, I think it's fine to leave it as-is - even if 1.5 doesn't exist on NuGet yet. It will exist soon enough

Issue #1420, opened 2016-02-10
Deactivate grains upon write conflict


ReubenBond:
Currently Orleans does not deactivate grains when a call to `WriteStateAsync` fails with a write conflict. In such cases, a grain may fail indefinitely and exceptions would need to be handled by user code.

The correct behavior in this case (in my opinion) is to force a grain deactivation so that any transient state is cleared and any persistent state is read from storage on the receipt of the next message.

I believe we should either throw a well-known exception from storage providers in this case and have Orleans handle the exception with a call to `DeactivateOnIdle` before rethrowing, or require storage providers to handle this case themselves. I prefer the former.

gabikliot:
@reubenbond, I agree.
I clearly remember we discussed that in the past, either privately with Sergey and Jason, or on GH.
I think the design was to throw a known Orleans specific conflict exception from storage provider, the grain can catch and recover if it wants, but if not and the exception escapes, the runtime deactivates the grain and returns the exception to the caller.

Eldar1205:
Hi, I want to add a comment regarding possible consequences with de-activating a grain due to state conflict. With Azure worker role deployment, conflict may be due to swap-upgrade where production & staging slot work at the same time against same storage account. It may be important for the grain to remain active because it runs a low-frequency timer that is critical and shouldn't wait for re-activation. I suggest to at least have a configurable flag whether to de-activate grains on state conflict, that is by default false in order to keep backwards comparability for Orleans users relying on current behaviour.

lmagyar:
See also #1609

ashkan-saeedi-mazdeh:
I think as @gabikliot  said, it only should deactivate if the user lets the exception escape and not handle it since in many valid cases you might be able to tolerate some write failures or want to reread state and try to write again (i.e. if it is a conflict, you might want to handle it yourself).

ReubenBond:
Resolved by #2959

jdom:
Also fixed the Azure table and blob storage providers to always throw `InconsistentStateException` when there are e-Tag conflicts: https://github.com/dotnet/orleans/pull/2971

PR #2959, opened 2017-04-20
DeactivateOnIdle after InconsistentStateException

ReubenBond:
Fix for #1420

When a storage provider throws an `InconsistentStateException` (or a derivative of that exception) which is not handled by a grain, the grain will be deactivated so that state can lazily be reactivated with freshly-read state upon next use.


src/OrleansRuntime/Core/InsideRuntimeClient.cs
```diff
@@ -325,6 +326,12 @@ public void SniffIncomingMessage(Message message)
                         invokeExceptionLogger.Warn(ErrorCode.GrainInvokeException,
                             "Exception during Grain method call of message: " + message, exc1);
                     }
+
+                    if (exc1 is InconsistentStateException && target is Grain targetGrain)
```
jdom: This is using a C# 7 feature. I don't think we made the jump to stop supporting vs 2015 yet (I think the idea is to do it right after 1.5)

src/OrleansRuntime/Core/InsideRuntimeClient.cs
```diff
@@ -325,6 +326,12 @@ public void SniffIncomingMessage(Message message)
                         invokeExceptionLogger.Warn(ErrorCode.GrainInvokeException,
                             "Exception during Grain method call of message: " + message, exc1);
                     }
+
+                    if (exc1 is InconsistentStateException && target is Grain)
+                    {
+                        this.DeactivateOnIdle(((Grain)target).Data.ActivationId);
```
jdom: should we log something, given that it's a new (and potentially surprising) behavior?

jdom:
hmmm, the current failure is a genuine issue:
```
Test: DefaultCluster.Tests.StorageTests.MemoryStorageProviderTests.MemoryStorageGrainEnforcesEtagsTest (from DefaultCluster.Tests)
MESSAGE:
Assert.Throws() Failure\r\nExpected: typeof(Orleans.Storage.InconsistentStateException)\r\nActual: (No exception was thrown)
```

This is because the `MemoryStorageProvider` expects the `MemoryStorageGrain` to throw `InconsistentStateException` when eTags are violated, so that it bubbles up to the grain who's using the storage provider.
We can easily fix this by having the `MemoryStorageGrain` not throw that exception, but instead throw something else, that then gets converted by the `MemoryStorageProvider` into an `InconsistentStateException`.

src/OrleansRuntime/Core/InsideRuntimeClient.cs
ReubenBond: Yes, that sounds like a good idea. INFO level?

jdom: Sure

src/Orleans/Utils/WrappedException.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace Orleans.Runtime.Internal
+{
+    /// <summary>Exception used to communicate with the storage provider, so that it throws this exception to its caller.</summary>
+    [Serializable]
+    public class WrappedException : OrleansException
```
jason-bragg: By doing it this way, we're adding to the Orleans public surface, imo, unnecessarily.
This behavior seems specific to the storage provider.
I don't see this as something that should be a general Orleans exception.

jdom: Hehe, I had exactly the same concern and went back and forth with several options. I can make this internal and add 1 InternalsVisibleTo to OrleansProviders (because this logic is spread through 3 projects)

jdom: BTW, the memory storage grain itself is also an implementation detail of this provider that is public, so that was my encouragement to go with this option, but I certainly prefer making this new exception internal

jason-bragg: Internal/public is one dimension, but my main point is that, imo, this should be considered part of the memory storage provider implementation.  As it's an internal signal between the grain and the storage provider, it's not a proper 'Orleans exception', so should not inherit from that base class, and should not be in Orleans utils, but instead, packaged with the memory storage provider.

jdom: Well, but that's my point, I understand it's an implementation detail, but that details is currently spread through 3 different projects: 
- `IMemoryStorageGrain`: defined in `Orleans`
- `MemoryStorageGrain`: defined in `OrleansRuntime`
- `MemoryStorage`: defined in `OrleansProviders`

Anyway, agree on not inheriting from OrleansException, and will make this internal (and add InternalsVisiblesTo) and move to a storage specific namespace

jdom: BTW, OrleansProviders does NOT reference OrleansRuntime. Originally I attempted to add this exception as a nested type inside MemoryStorageGrain, but the MemoryStorage did not have access to it. Also, can't create a nested type inside the IMemoryStorageGrain (cannot nest types in interfaces), so my only option is to create a full blown type in Orleans itself... but I'll make it internal now

PR #875, opened 2015-10-05
Avoid unnecessary use of TCS

jdom:
Removes the unnecessary wrapping of a `TaskCompletionSource` into another `TaskCompletionSource`.

For some reason 1 test (`Generic_SimpleGrainControlFlow`) started deadlocking after the change (and ending in a `TimeoutException` after 30 seconds). The test was doing the wrong thing in the first place (blocking on `.Result`), nevertheless because the `OutsideRuntimeClient` executes the response callbacks in the same thread as where it is doing the message pump, then the `.Result` resulted in the message pump not running. I'm not sure why this was not the case before the change, so I'm worried that this could have breaking changes for people that were also doing blocking waits (albeit wrong).
This could be fixed by executing the callback inside a `Task.Run()` call, but that could bring bad perf implications for the people writing the code correctly and not blocking (although not sure if this path only occurs in unit tests or in an actual Orleans client, as that could be dangerous too if you are for example doing CPU intensive operations with the response, such as deserializing complex objects).

This PR is mainly a talking point, but it would be fine not to commit to master if the deadlocking I started seeing before my 2nd commit is a valid concern.

PR #2960, opened 2017-04-20
Remove conversion of resulting task via task competition source

dVakulen:
Removes unneeded conversion via ```TaskCompetitionSource``` that was supposedly used for avoiding deadlocks in client code. At first I wanted to add ```TaskCreationOptions.RunContinuationsAsynchronously```, but the tests seems to pass even without it. 

And another possible optimization - task unboxing (which invokes costly ```ContinueWith``` on non-synchronous path) can also be removed if to use generics on grain method invocation path, as can be seen in https://github.com/dotnet/orleans/pull/2060/commits/6bf22d11343e37d203c12e1fef8f43b6f32e1754#diff-64f4df81b54d326e17440c81957e905cR347 . 

Only latest commit matters, other will be squashed automatically on merge anyway.


dVakulen:
Seems that it is an duplicate of https://github.com/dotnet/orleans/pull/875

ReubenBond:
@dVakulen I think now is the time to remove ConvertViaTcs, since we just moved to 4.6. See Gabi's comment here: https://github.com/dotnet/orleans/pull/1084#issuecomment-160387369

dVakulen:
@ReubenBond I meant that this PR moslty duplicating #875 and it'll probably be more ocrrect to just reopen that one.

PR #2961, opened 2017-04-20
Use AsyncLocal for RequestContext

dVakulen:
As target framework version was upgraded to .Net 4.6.1 now it's possible to remove NETSTANDARD ifdefs from ```RequestContext```.


jdom:
I think this is a good idea. Nevertheless, I'm not so sure about no longer honoring `Trace.CorrelationManager.ActivityId` since that one is not just an implementation detail, and people set that directly.
What are your thoughts about it?

dVakulen:
As ```AsyncLocal``` activity id is the only way of storing it while running on .Net core - it should become the default, and on full .Net we can use ```Trace.CorrelationManager.ActivityId``` in case of empty ```AsyncLocal``` one.

dVakulen:
@jdom Added commit that does fallback to ```Trace.CorrelationManager.ActivityId``` in case of empty ```AsyncLocal``` one :   https://github.com/dotnet/orleans/pull/2961/commits/c4b726b804646878d6f4c6d896ff4620bc9c7a0c .
Will add tests if thats acceptable approach.

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -114,17 +115,21 @@ public static void Import(Dictionary<string, object> contextData)
...
 
-#if NETSTANDARD
-                ActivityId.Value = (Guid) activityIdObj;
-#else
-                Trace.CorrelationManager.ActivityId = (Guid)activityIdObj;
+                ActivityId.Value = (Guid)activityIdObj;
+
+                object legacyActivityIdObj;
+#if !NETSTANDARD
+                if ((Guid)activityIdObj == Guid.Empty && contextData.TryGetValue(E2_E_TRACING_LEGACY_ACTIVITY_ID_HEADER, out legacyActivityIdObj))
```
jdom: I would do this inside the previous `if`, before assigning it to `Guid.Empty`

dVakulen: Right, unneeded lookup on common case.

jdom:
Awesome, thanks! BTW, please add a test for it, as you suggested

dVakulen:
Done.

dVakulen:
@dotnet-bot test this please; fails unrelated to the PR changes

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -112,19 +113,23 @@ public static void Import(Dictionary<string, object> contextData)
...
                     activityIdObj = Guid.Empty;
-                }
 
-#if NETSTANDARD
-                ActivityId.Value = (Guid) activityIdObj;
-#else
-                Trace.CorrelationManager.ActivityId = (Guid)activityIdObj;
+                    object legacyActivityIdObj;
+#if !NETSTANDARD
+                    if (contextData.TryGetValue(E2_E_TRACING_LEGACY_ACTIVITY_ID_HEADER, out legacyActivityIdObj))
```
jdom: The reason for the test failures is that `contextData` can be null, so this throws

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 #if !NETSTANDARD
 using System.Runtime.Remoting.Messaging;
 #else
```
jdom: check this ifdefing (you are adding System.Threading two times, one outside the `if`, and one inside the `else`)

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -112,19 +113,23 @@ public static void Import(Dictionary<string, object> contextData)
...
+                    object legacyActivityIdObj;
+#if !NETSTANDARD
+                    if (contextData.TryGetValue(E2_E_TRACING_LEGACY_ACTIVITY_ID_HEADER, out legacyActivityIdObj))
+                    {
+                        Trace.CorrelationManager.ActivityId = (Guid)legacyActivityIdObj;
+                    }
 #endif
+                }
+
+                ActivityId.Value = (Guid)activityIdObj;
```
jdom: Most likely this is a non-issue, but could this cause unnecessary perf impact if the user is never setting the ActivityId in his code (but we set it always to Guid.Empty)? Probably not a concern in terms of memory pressure, but not sure about how expensive is to lift this value when scheduling tasks.
/cc @ReubenBond

dVakulen: It's not expensive from memory standpoint, but ```ExecutionContext``` is optimized for no values case, so flowing of that value will indeed take some CPU time, but I doubt that it could be seen on performance tests. Anyway, fixed.

src/Orleans/Runtime/RequestContext.cs
dVakulen: Fixed

src/Orleans/Runtime/RequestContext.cs
dVakulen: I wonder if that cryptic error message from test runs can be fixed.

jdom:
@dotnet-bot test netfx-bvt
Reason: unrelated failure

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -92,39 +88,43 @@ public static void Set(string key, object value)
...
                 {
-                    activityIdObj = Guid.Empty;
-                }
-
-#if NETSTANDARD
-                ActivityId.Value = (Guid) activityIdObj;
-#else
-                Trace.CorrelationManager.ActivityId = (Guid)activityIdObj;
+                    object legacyActivityIdObj = null;
+#if !NETSTANDARD
```
jdom: super-nit: I assume this generates a warning in NETSTANDARD, since you are initializing `legacyActivityIdObj` but never using it. Just move it inside the `#if`

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -92,39 +88,43 @@ public static void Set(string key, object value)
...
+                    if (contextData?.TryGetValue(E2_E_TRACING_LEGACY_ACTIVITY_ID_HEADER, out legacyActivityIdObj) == true)
+                    {
+                        Trace.CorrelationManager.ActivityId = (Guid) legacyActivityIdObj;
+                    }
 #endif
+                }
+                else
+                {
+                    ActivityId.Value = (Guid)activityIdObj;
```
jdom: I suppose this is enough not to set ActivityId if unused. Do we want to set it to Guid.Empty if `ActivityId.Value` is already set to something else though?
I believe the answer is no, and we should keep the code as-is, but wanted to ask you in case you did not consider that case and you actually have a different opinion about it.

dVakulen: I strongly suspect that it will be empty in that cases, as local values ```ExecutionContext``` does not flow up to parent task as can be seen at https://github.com/dotnet/corefx/blob/b04757c307eee37dcc2b887f0b5ba08d7e866cc1/src/System.Threading/tests/AsyncLocalTests.cs#L255 and ```RequestContext``` is being imported only in tasks that is created by Orleans per request, thus ensuring that non-empty value can not end up in that place.  (+ in some places in streaming, which I didn't reviewed thoroughly, but they also seems to be safe).  But to be on completely safe side, as retrieving of empty value does not incurs much cost [value retrieving code](https://github.com/dotnet/coreclr/blob/c85c19baef5edf867c3a9a217cbe235a892d0c1e/src/mscorlib/shared/System/Threading/ExecutionContext.cs#L238), [values container class that is optimized for 
 common case of no values](https://github.com/dotnet/coreclr/blob/c85c19baef5edf867c3a9a217cbe235a892d0c1e/src/mscorlib/shared/System/Threading/AsyncLocal.cs#L128) I've just added check for non-empty async local activity id value to set it to default value.

src/Orleans/Runtime/RequestContext.cs
dVakulen: Moved

src/Orleans/Runtime/RequestContext.cs
jdom: Hehe, we think alike. I did the same analysis as you and draw the same conclusion: it's probably empty in most cases since it doesn't flow up, but still it wouldn't harm to do the check (although I would have been fine if we didn't do the check too, since it's a very rare edge condition).

jdom:
LGTM, I'll merge when the tests complete

src/Orleans/Runtime/RequestContext.cs
```diff
@@ -36,13 +32,13 @@ public static class RequestContext
 
         internal const string CALL_CHAIN_REQUEST_CONTEXT_HEADER = "#RC_CCH";
         internal const string E2_E_TRACING_ACTIVITY_ID_HEADER = "#RC_AI";
+        internal const string E2_E_TRACING_LEGACY_ACTIVITY_ID_HEADER = "#L_RC_AI";
         internal const string ORLEANS_REQUEST_CONTEXT_KEY = "#ORL_RC";
```
jdom: This is no longer used. But no issue, I'll delete it afterwards, merging for now

test/Tester/TestUtils.cs
```diff
@@ -114,20 +114,12 @@ public static class RequestContextTestUtils
     {
         public static void SetActivityId(Guid id)
         {
-#if NETSTANDARD
             RequestContext.ActivityId.Value = id;
```
jdom: Since we are now officializing this approach (previously it was just a temp workaround for .NET Standard, so I didn't nit pick on it), it would be nice to not expose the AsyncLocal field, but instead have a wrapper property, so this would look like:
`RequestContext.ActivityId = id`

Same with the CallContextData field.

jdom: BTW, looking closer, should we just dump the ActivityId as a separate AsyncLocal field? Looks like we are putting it in the same dictionary as CallContextData when flowing it, and there is no inherent reason to have it separate. The reason some of it was separate, is that we were also augmenting and flowing the `Trace.CorrelationManager.ActivityId`, which is an external property that people might have already been using.

We can of course have a property in RequestContext so that people don't have to use a well-known header for it, such as:

```
        /// <summary>Gets or sets an activity ID that can be used for correlation</summary>
        public Guid ActivityId
        {
            get { return (Guid)(Get(E2_E_TRACING_ACTIVITY_ID_HEADER) ?? Guid.Empty); }
            set { Set(E2_E_TRACING_ACTIVITY_ID_HEADER, value); }
        }
```

jdom: Created a PR against your branch (https://github.com/dVakulen/orleans/pull/8), let me know what you think. I believe it greatly simplifies the ActivityID API (and impl), especially for end users of both .NET and .NET Standard.

dVakulen: Yeah, the reason it was extracted into separate variable was to [ save us from more conditional compilation statements.](https://github.com/dotnet/orleans/pull/2200#discussion_r80574713).

jdom:
Thanks a lot @dVakulen!

dVakulen:
Thank you too, @jdom. 

As side note - while it's indeed cool to receive thumb ups, I don't quite get why this one has got so much  of them (compared to other PR's).

ReubenBond:
@dVakulen the results from, eg #2060 look amazing (up to 81% perf improvement!), I think the thumbs up here are because we've each personally dealt with differences between .NET Standard & full .NET - there were some hard times there... so we're just so happy to see convergence.

jdom:
yup, thumbs up are because we are starting to converge to a single source tree for 2.0. The 3 of us that gave thumbs up are the ones that have being dealing with the differences in the codebases for several months. So after the 4.6.1 merge we were already planning to do this work (and other NETSTANDARD_TODO ones), so you coming and fixing it for us is all goodness. 
All PRs are useful. Some are not in our main priority, so while good, they still add work for us in the form of reviews and testing, that we have to prioritize high to keep the contributions coming. This one is good, but also aligns with our priority, so it removes work from us. Sure, we have to review it, but you did take the time to do most of the heavy lifting and testing.

PR #2962, opened 2017-04-20
Fix various unhandled exceptions happening during client close.

attilah:
These came up during the debugging of client disconnect:

- AsynchQueueAgent.QueueRequest: NullReferenceException
- AsynchQueueAgent.RunNonBatching: NullReferenceException
- GatewayConnection.PrepareMessageForSend: NullReferenceException
- ProxiedMessageCenter.Disconnect: Collection was modified
- ProxiedMessageCenter.Stop: Collection was modified

### DIFF
--- a/src/Orleans/Messaging/GatewayConnection.cs
+++ b/src/Orleans/Messaging/GatewayConnection.cs
@@ -189,7 +189,7 @@ namespace Orleans.Messaging
         protected override bool PrepareMessageForSend(Message msg)
         {
             // Check to make sure we're not stopped
-            if (Cts.IsCancellationRequested)
+            if (Cts!=null && Cts.IsCancellationRequested)
             {
                 // Recycle the message we've dequeued. Note that this will recycle messages that were queued up to be sent when the gateway connection is declared dead
                 MsgCenter.SendMessage(msg);
--- a/src/Orleans/Messaging/GatewayConnection.cs
+++ b/src/Orleans/Messaging/GatewayConnection.cs
@@ -188,8 +188,13 @@ namespace Orleans.Messaging
 
         protected override bool PrepareMessageForSend(Message msg)
         {
+            if (Cts == null)
+            {
+                return false;
+            }
+
             // Check to make sure we're not stopped
-            if (Cts!=null && Cts.IsCancellationRequested)
+            if (Cts.IsCancellationRequested)
             {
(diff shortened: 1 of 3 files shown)


src/Orleans/Runtime/AsynchQueueAgent.cs
```diff
@@ -30,7 +30,10 @@ public void QueueRequest(T request)
                 queueTracking.OnEnQueueRequest(1, requestQueue.Count, request);
             }
 #endif
-            requestQueue.Add(request);
+            if (requestQueue != null)
+            {
+                requestQueue.Add(request);
```
ReubenBond: For this one, it might be better to invert the logic and add a comment. Something like:
```C#
// Avoid adding to the queue during shutdown.
if (requestQueue == null) return;
```

attilah: I invert it, but how is it better? for readability?

The reason from my side was that since this is the last statement in that method an early return doesn't benefit anything.

I'm moving the null check to the beginning to not to execute the stats code in case the ```requestQueue``` is null, because that can falsify the statistics ;-)

jdom: I think it's better due to potential race conditions too :)
But yeah, even better now that it doesn't even attempt to touch statistics.

src/Orleans/Messaging/GatewayConnection.cs
```diff
@@ -189,7 +189,7 @@ public void Connect()
         protected override bool PrepareMessageForSend(Message msg)
         {
             // Check to make sure we're not stopped
-            if (Cts.IsCancellationRequested)
+            if (Cts!=null && Cts.IsCancellationRequested)
```
jdom: This is not doing the right thing now I believe.
If it's null, it should probably just return, because it's past anything it can do. But at the very least, if it's null, it should treat as canceled, not as not canceled as it's doing now.

attilah: As we discussed in person, I'm adding the return on null, since no additional code should be executed in that case.

src/Orleans/Runtime/AsynchQueueAgent.cs
ReubenBond: Simply for understandability - the null check is odd, so the comment mentions when it might be null.

PR #2971, opened 2017-04-24
Throw InconsistenStateException in Azure storage providers

jdom:
Update AzureTableStorage and AzureBlobStorage to consistently throw InconsistentStateException when there are conflicting eTags when updating the store.
- Update AzureTableStorage provider to throw InconsistentStateException also on Clear (it was only doing it on Write).
- Update AzureBlobStorage provider to throw InconsistentStateException on both Write and Clear (previously it was just bubbling up the StorageException).

### DIFF
--- a/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
+++ b/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
@@ -207,8 +207,8 @@ namespace Orleans.Storage
 
                 var blob = container.GetBlockBlobReference(blobName);
 
-                await DoConditionalUpdate(() => blob.DeleteIfExistsAsync(DeleteSnapshotsOption.None, AccessCondition.GenerateIfMatchCondition(grainState.ETag), null, null),
-                    blobName, grainState.ETag).ConfigureAwait(false);
+                await DoOptimisticUpdate(() => blob.DeleteIfExistsAsync(DeleteSnapshotsOption.None, AccessCondition.GenerateIfMatchCondition(grainState.ETag), null, null),
+                    blob, grainState.ETag).ConfigureAwait(false);
 
                 grainState.ETag = null;
 
@@ -228,8 +228,10 @@ namespace Orleans.Storage
         {
             try
             {
-                await DoConditionalUpdate(() => blob.UploadTextAsync(json, Encoding.UTF8, AccessCondition.GenerateIfMatchCondition(grainState.ETag), null, null),
-                    blob.Name, grainState.ETag).ConfigureAwait(false);
+                await DoOptimisticUpdate(() => blob.UploadTextAsync(json, Encoding.UTF8, AccessCondition.GenerateIfMatchCondition(grainState.ETag), null, null),
+                    blob, grainState.ETag).ConfigureAwait(false);
+
+                grainState.ETag = blob.Properties.ETag;
             }
             catch (StorageException exception) when (exception.IsContainerNotFound())
             {
@@ -239,11 +241,9 @@ namespace Orleans.Storage
 
                 await WriteStateAndCreateContainerIfNotExists(grainType, grainId, grainState, json, blob).ConfigureAwait(false);
             }
-
-            grainState.ETag = blob.Properties.ETag;
         }
 
-        private async Task DoConditionalUpdate(Func<Task> updateOperation, string blobName, string currentETag)
+        private static async Task DoOptimisticUpdate(Func<Task> updateOperation, CloudBlob blob, string currentETag)
         {
             try
             {
@@ -251,7 +251,7 @@ namespace Orleans.Storage
             }
             catch (StorageException ex) when (ex.IsPreconditionFailed())
             {
-                throw new InconsistentStateException($"Blob storage condition not Satisfied.  BlobName: {blobName}, Container: {container.Name}, CurrentETag: {currentETag}", "Unkown", currentETag, ex);
+                throw new InconsistentStateException($"Blob storage condition not Satisfied.  BlobName: {blob.Name}, Container: {blob.Container?.Name}, CurrentETag: {currentETag}", "Unkown", currentETag, ex);
             }
         }
     }
--- a/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
+++ b/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
@@ -189,33 +189,30 @@ namespace Orleans.Storage
                 var blob = container.GetBlockBlobReference(blobName);
                 blob.Properties.ContentType = "application/json";
 
-                var containerNotFound = false;
                 try
                 {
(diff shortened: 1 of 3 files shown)


jdom:
Comes after https://github.com/dotnet/orleans/pull/2959

jdom:
Removed dependency on #2959 so that it can be merged in isolation and in whatever order

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
```diff
@@ -189,33 +189,30 @@ private static string GetBlobName(string grainType, GrainReference grainId)
...
-                            Encoding.UTF8,
-                            AccessCondition.GenerateIfMatchCondition(grainState.ETag),
-                            null,
-                            null).ConfigureAwait(false);
+                    await DoConditionalUpdate(() => blob.UploadTextAsync(json, Encoding.UTF8, AccessCondition.GenerateIfMatchCondition(grainState.ETag), null, null),
+                        blobName, grainState.ETag).ConfigureAwait(false);
                 }
                 catch (StorageException exception)
                 {
+                    if (exception.IsPreconditionFailed())
```
ReubenBond: Won't this always be false given it would otherwise be caught in `DoConditionalUpdate`?

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
```diff
@@ -189,33 +189,30 @@ private static string GetBlobName(string grainType, GrainReference grainId)
...
                     if (this.Log.IsVerbose3) this.Log.Verbose3((int)AzureProviderErrorCode.AzureBlobProvider_ContainerNotFound, "Creating container: GrainType={0} Grainid={1} ETag={2} to BlobName={3} in Container={4}", grainType, grainId, grainState.ETag, blobName, container.Name);
                     await container.CreateIfNotExistsAsync().ConfigureAwait(false);
 
-                    await blob.UploadTextAsync(
-                        json,
-                        Encoding.UTF8,
-                        AccessCondition.GenerateIfMatchCondition(grainState.ETag),
-                        null,
-                        null).ConfigureAwait(false);
+                    await DoConditionalUpdate(() => blob.UploadTextAsync(json, Encoding.UTF8, AccessCondition.GenerateIfMatchCondition(grainState.ETag), null, null),
```
ReubenBond: I wonder if recursion would be cleaner

ReubenBond: Well, it might be cleaner, but it would also result in serializing the state twice unless we had a helper function (eg, a local function), but leaving it as-is is probably best.

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
jdom: Oops, yes, forgot to remove this

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
jdom: yeah, I briefly consider that too, but prefer to do this in case it would recurse forever due to an issue with the service, but there reality is that it's very unlikely. I added a commit that makes it recursive.

jdom: BTW, local functions are in c#7, so I used a normal named method.
In theory I could have recursed the entire method, even if it meant serializing the state twice, since this is a very extreme edge case that would probably never happen (since we create the container on init, and it's not common that someone will come and delete the container while Orleans is running). But went for the proper way that covers the edge case, since it wasn't much harder than not doing it like this.

src/OrleansAzureUtils/Providers/Storage/AzureTableStorage.cs
```diff
@@ -237,6 +227,18 @@ public Task Close()
             }
         }
 
+        private async Task DoConditionalUpdate(Func<Task> updateOperation, string grainType, GrainReference grainReference, string tableName, string currentETag)
```
jason-bragg: static?

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
```diff
@@ -261,5 +223,36 @@ private static string GetBlobName(string grainType, GrainReference grainId)
...
+                await DoConditionalUpdate(() => blob.UploadTextAsync(json, Encoding.UTF8, AccessCondition.GenerateIfMatchCondition(grainState.ETag), null, null),
+                    blob.Name, grainState.ETag).ConfigureAwait(false);
+            }
+            catch (StorageException exception) when (exception.IsContainerNotFound())
+            {
+                // if the container does not exist, create it, and make another attempt
+                if (this.Log.IsVerbose3) this.Log.Verbose3((int)AzureProviderErrorCode.AzureBlobProvider_ContainerNotFound, "Creating container: GrainType={0} Grainid={1} ETag={2} to BlobName={3} in Container={4}", grainType, grainId, grainState.ETag, blob.Name, container.Name);
+                await container.CreateIfNotExistsAsync().ConfigureAwait(false);
+
+                await WriteStateAndCreateContainerIfNotExists(grainType, grainId, grainState, json, blob).ConfigureAwait(false);
```
jason-bragg: return WriteStateAndCreateContainerIfNotExists...

Reason is that after exception we duplicate work.  While this is ok now, because it's just a duplication of the same state change that was just done in the successful WriteStateAndCreateContainerIfNotExists, it may not always be ok.  For instance, what would happen if someone started tracking a successful write statistic at the end of this call.

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
```diff
@@ -261,5 +223,36 @@ private static string GetBlobName(string grainType, GrainReference grainId)
...
+            }
+
+            grainState.ETag = blob.Properties.ETag;
+        }
+
+        private async Task DoConditionalUpdate(Func<Task> updateOperation, string blobName, string currentETag)
```
jason-bragg: nit:  DoConditionalUpdate does not seem like a good name for this call.  It name seems to imply that it only updates on some condition..

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
jdom: good point, meant to include the eTag assignment only if it didn't throw (inside the try block).

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
jdom: DoOptimisticUpdate?

jdom:
Please, when merging, do a squash, and only include the comments from the first commit

Issue #2966, opened 2017-04-24
ExecuteWithRetriesHelper swallows the original exception


DixonDs:
Checking `AsyncExecutorWithRetries.ExecuteWithRetriesHelper`, it seems that if the original action throws exceptions again and again, we will reach timeout and then `TimeoutException` will be thrown. In this case, we won't have the original exception anywhere it logs.

Can we rethrow the last exception if any instead of `TimeoutException`?

PR #2972, opened 2017-04-25
Improve ExecuteWithRetries helper

DixonDs:
This change serves multiple purposes:

1. Fixes https://github.com/dotnet/orleans/issues/2966
2. <s>Using exception filters instead of rethrowing the original exception keeps the stack unwound (http://www.thomaslevesque.com/2015/06/21/exception-filters-in-c-6/)</s>
3. <s>Executes the underlying task with timeout to quit faster.</s>
4. <s>In the previous version, we check the timeout after the delay, but I think it is better to do it vice versa - in that case we won't wait the extra delay (which could be significant in case of an exponential retry, for instance) if there is timeout already.</s>
5. <s>Eliminates code duplicates with helper private methods.</s>

### DIFF
--- a/src/Orleans/Async/AsyncExecutorWithRetries.cs
+++ b/src/Orleans/Async/AsyncExecutorWithRetries.cs
@@ -5,7 +5,7 @@ using Orleans.Runtime;
 namespace Orleans
 {
     /// <summary>
-    /// This class a convinent utiliity class to execute a certain asyncronous function with retires, 
+    /// This class a convinent utiliity class to execute a certain asyncronous function with retires,
     /// allowing to specify custom retry filters and policies.
     /// </summary>
     internal static class AsyncExecutorWithRetries
@@ -61,7 +61,7 @@ namespace Orleans
         /// Execute a given function a number of times, based on retry configuration parameters.
         /// </summary>
         /// <param name="function">Function to execute</param>
-        /// <param name="maxNumSuccessTries">Maximal number of successful execution attempts. 
+        /// <param name="maxNumSuccessTries">Maximal number of successful execution attempts.
         /// ExecuteWithRetries will try to re-execute the given function again if directed so by retryValueFilter.
         /// Set to -1 for unlimited number of success retries, until retryValueFilter is satisfied.
         /// Set to 0 for only one success attempt, which will cause retryValueFilter to be ignored and the given function executed only once until first success.</param>
@@ -116,18 +116,6 @@ namespace Orleans
 
             do
             {
-                retry = false;
-
-                if (maxExecutionTime != Constants.INFINITE_TIMESPAN && maxExecutionTime != default(TimeSpan))
-                {
-                    DateTime now = DateTime.UtcNow;
-                    if (now - startExecutionTime > maxExecutionTime)
-                    {
-                        throw new TimeoutException(
-                            $"ExecuteWithRetries has exceeded its max execution time of {maxExecutionTime}. Now is {LogFormatter.PrintDate(now)}, started at {LogFormatter.PrintDate(startExecutionTime)}, passed {now - startExecutionTime}");
-                    }
-                }
-
                 int counter = callCounter;
 
                 try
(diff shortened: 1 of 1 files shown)


gabikliot:
I wonder: this is an internal class, can not to be used by application code. So why does anyone care to extend/improve it? And for internal usage, it is good enough to do what it was meant to do.

DixonDs:
@gabikliot I care because we were observing timeout issues in `PersistentStreamPullingAgent` and all we've seen in logs was `TimeoutException` from `ExecuteWithRetries`. That is something I reported as https://github.com/dotnet/orleans/issues/2966 and while preparing PR to fix it I realized that in the same time there are other things that could be improved like using `WithTimeout` to drop a hanging consumer. 

>I wonder: this is an internal class, can not to be used by application code. So why does anyone care to extend/improve it?

It may be just wording, but it makes it sound like contributions to Orleans internals are not welcomed.

src/Orleans/Async/AsyncExecutorWithRetries.cs
```diff
@@ -116,72 +116,87 @@ internal static class AsyncExecutorWithRetries
...
-                    }
-                }
-
                 int counter = callCounter;
 
                 try
                 {
                     callCounter++;
-                    result = await function(counter);
+                    result = await function(counter).WithTimeout(GetTimeout(startExecutionTime, maxExecutionTime));
```
ReubenBond: We may need to consider tweaking some values if this is the semantics we want. It seems like the right choice, but perhaps `AzureTableDefaultPolicies.TableOperationTimeout` needs to be increased above 3 seconds. This shouldn't break anything either way

jason-bragg: I understand this change and see the value in it, but I'm still a bit concerned by it.  It is a behavioral change in which successful operations can be returned as timeout errors.  It is an improvement in the callers control of how much time is spent retrying, but at a cost.

As the current timeout behavior has not demonstrated itself to be flawed or insufficient, I submit that we should stick with it until we've a need to change it.

src/Orleans/Async/AsyncExecutorWithRetries.cs
```diff
@@ -116,72 +116,87 @@ internal static class AsyncExecutorWithRetries
...
-
-                        if (delay.HasValue)
-                        {
-                            await Task.Delay(delay.Value);
-                        }
+                        await DelayBeforeRetryAsync(onSuccessBackOff, counter);
                     }
                 }
-                catch (Exception exc)
+                catch (Exception exc) when (!(exc is TimeoutException) && ShouldRetry(counter, maxNumErrorTries,
```
jason-bragg: A check for a timeout exception here is not behaviorally safe, as the underlying application layer operation may throw a timeout exception, and we'd normally be allowed to retry.  This is likely a common error that would no longer be retry-able.

jason-bragg:
To address #2966, I suggest we keep the last thrown exception, and if a timeout exception is thrown, we include the last thrown exception as the inner exception of the timeout.

It is also possible for callers to log exceptions as they occur in the provided retryExceptionFilter.

jason-bragg:
@DixonDs In case their is any uncertainty, we very much appreciate and encourage your contributions.  I'm quite confident @gabikliot did not intend to discourage community improvements, but I'll leave it to him to clarify his comments.

gabikliot:
Yes, of course I did not discourage community improvements. Sorry if it may have sounded like that.
My concern, in this particular PR, is exactly the same as Jason expressed above in his review.  It is not clear to me that the change may or may not break the other internal code that uses it. The change introduces a behavioral change and may not be safe. Therefore now this potential improvement need to be compared against the risk and the added cost of making sure nothing else will break.

If it was improving a public API, the trade-off would have been different, since the advantage (of improving the code for any future customers) will likely out weight the extra cost of validating that the internals don't break, and even the risk of breaking them.

DixonDs:
I've updated the PR to have minimal changes to fix https://github.com/dotnet/orleans/issues/2966 with no other changes to be introduced

jason-bragg:
With the exception of a couple of behavioral changes, I liked the previous version better, as it cleaned up the code more, but this is probably less risky.

There is still a behavioral change, in that the previous version only ever through a timeout exception, but I walked through the usages and nothing seems to be checking explicitly for the timeout exception, and in fact some seem to expect other exceptions, so, while this is a behavioral change, I think it's probably a good one.

Thanks @DixonDs!

@DixonDs, as for the difficulty you encountered identifying the exceptions being thrown in the streaming retry logic, I'll look into improving that.  Thanks for letting us know about it.

gabikliot:
Thanks @DixonDs !

PR #2974, opened 2017-04-25
Set stream subscription faulting default to false.

jason-bragg:


xiazen:
LGTM, will merge it after test finished

xiazen:

PR #2976, opened 2017-04-26
Fix OrleansCodeGenerator.Build.nuspec for .NET 4.6.1 upgrade

ReubenBond:


jdom:
Perhaps we should do IlMerge on these? Let's keep an out for issues in build-time codegen during the beta.

Issue #2968, opened 2017-04-24
Faulted consumers in PubSubSubscriptionState


DixonDs:
We see that the size of `PubSubSubscriptionState` is growing from time to time with consumers with `state=1` (`SubscriptionStates.Faulted`). It looks like after a consumer is marked as faulted, it is never returned in `GetAllSubscriptionHandles` so it is not possible to unsubscribe (i.e. remove that consumer from the pubsub state). 

What's the best practice here? Can we remove those fault consumers other than tampering with stored `PubSubSubscriptionState` manually?

xiazen:
I don't think you can remove those fault subscriptions in an elegant way currently. 

You can explicitly remove a healthy subscription from pubsub using IStreamSubscriptionManager.RemoveSubscription. But I don't think you can remove a fault subscription. 

I wonder why fault subscription wasn't cleaned at the time when the subscription was marked Faulted. What purpose are they kept around for? Maybe @jason-bragg knows more about the context there.

DixonDs:
@xiazen 
>Just to clear the water a little bit, Fault state is on a subscription, not a consumer. There's not a concept of Fault state on a consumer. A consumer can have multiple subscriptions.

Yes, thanks for correction.

The whole thing with faulted subscriptions is unclear for me. It looks like the behavior is different for implicit and explicit subscriptions. Correct me if I am wrong, but for explicit subscriptions, once they marked as faulted, they will never recover as they are ignored in most methods like retrieving the subscriptions list etc. But for implicit subscriptions, after a silo restart, previously faulted subscriptions will be used again to deliver messages. If that is correct, does it mean that implicit subscriptions are more fault-tolerant and the preferable way to subscribe to streams?

ashkan-saeedi-mazdeh:
By reading this I got a question as well.
What is the exact conditions that a subscription assumed to be faulty and the state change is triggered?

IIRC Sergey said if a subscription can not receive a message , it will be considered faulty after some retires but that was for clients which this definition is logical for, For grains which subscribe to streams, it should not go to that state.

Chances are high that my assumptions are based overlooking facts since I have no experience making streaming systems.

jason-bragg:
A faulted subscription occurs when a delivery failure occurs on an explicitly subscribed stream and the stream failure handler's ShouldFaultSubsriptionOnError returns true.

Implicit subscriptions cannot be faulted.

[...]

PR #2978, opened 2017-04-26
Add ability to remove faulted subscription

xiazen:
regards to #2968, add ability to remove faulted subscription from pubsub

### DIFF
--- a/src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
+++ b/src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
@@ -280,9 +280,6 @@ namespace Orleans.Streams
         public async Task UnregisterConsumer(GuidId subscriptionId, StreamId streamId)
         {
             counterConsumersRemoved.Increment();
-            if (State.Consumers.Any(c => c.IsFaulted && c.Equals(subscriptionId)))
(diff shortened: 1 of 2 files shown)


xiazen:

src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
```diff
@@ -281,7 +281,8 @@ private void RemoveProducer(PubSubPublisherState producer)
         {
             counterConsumersRemoved.Increment();
             if (State.Consumers.Any(c => c.IsFaulted && c.Equals(subscriptionId)))
-                throw new FaultedSubscriptionException(subscriptionId, streamId);
+                logger.Warn(ErrorCode.Stream_RemoveFaultedSubscription,
```
jason-bragg: Originally faulted subscriptions were intended to be permanent, but it should probably be ok to remove them via the subscription management interfaces or by unsubscribing.  That being the case, I don't think we even need to check (and log) in this case.  So we can just remove this entire if block.

xiazen: yes boss

jason-bragg:
@dotnet-bot test netfx-bvt  netstandard-win-bvt  please

Issue #1747, opened 2016-05-10
ImplicitStreamSubscription via Regex(or other)


amccool:
When multiple namespaces need to be subscribed to it would be nice to be able to use a RegEx or some function.

`[ImplicitStreamSubscription(RegEx=@"(?<pets>cats) and dogs|(?<pets>pigs) and whistles")]`

sergeybykov:
Resolved via #2988.

PR #2988, opened 2017-05-01
Allow complex filters for ImplicitStreamSubscriptionAttribute

DixonDs:
My attempt to address https://github.com/dotnet/orleans/issues/1747

### DIFF
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -350,7 +350,7 @@ namespace Orleans
     }
 
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
-    public sealed class ImplicitStreamSubscriptionAttribute : Attribute
+    public class ImplicitStreamSubscriptionAttribute : Attribute
     {
         public IStreamNamespacePredicate Predicate { get; }
 
@@ -367,6 +367,11 @@ namespace Orleans
         public ImplicitStreamSubscriptionAttribute(Type predicateType)
         {
             Predicate = (IStreamNamespacePredicate)Activator.CreateInstance(predicateType);
+        }
+
+        public ImplicitStreamSubscriptionAttribute(IStreamNamespacePredicate predicate)
+        {
+            Predicate = predicate;
         }
     }
 }
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -1,5 +1,6 @@
 using System;
 using Orleans.GrainDirectory;
+using Orleans.Streams;
 
 namespace Orleans
 {
@@ -22,7 +23,7 @@ namespace Orleans
         /// <para>
         /// This is an advanced feature and should not be used unless the implications are fully understood.
         /// That said, allowing request interleaving allows the run-time system to perform a number of optimizations
-        /// that may significantly improve the performance of your application. 
+        /// that may significantly improve the performance of your application.
         /// </para>
         /// </summary>
         [AttributeUsage(AttributeTargets.Class)]
@@ -41,7 +42,7 @@ namespace Orleans
 
         /// <summary>
         /// The StatelessWorker attribute is used to mark grain class in which there is no expectation
-        /// of preservation of grain state between requests and where multiple activations of the same grain are allowed to be created by the runtime. 
+        /// of preservation of grain state between requests and where multiple activations of the same grain are allowed to be created by the runtime.
         /// </summary>
         [AttributeUsage(AttributeTargets.Class)]
         public sealed class StatelessWorkerAttribute : Attribute
@@ -66,7 +67,7 @@ namespace Orleans
         /// The AlwaysInterleaveAttribute attribute is used to mark methods that can interleave with any other method type, including write (non ReadOnly) requests.
         /// </summary>
         /// <remarks>
-        /// Note that this attribute is applied to method declaration in the grain interface, 
+        /// Note that this attribute is applied to method declaration in the grain interface,
         /// and not to the method in the implementation class itself.
         /// </remarks>
         [AttributeUsage(AttributeTargets.Method)]
@@ -75,11 +76,11 @@ namespace Orleans
         }
 
         /// <summary>
-        /// The MayInterleaveAttribute attribute is used to mark classes 
+        /// The MayInterleaveAttribute attribute is used to mark classes
         /// that want to control request interleaving via supplied method callback.
         /// </summary>
         /// <remarks>
-        /// The callback method name should point to a public static function declared on the same class 
+        /// The callback method name should point to a public static function declared on the same class
         /// and having the following signature: <c>public static bool MayInterleave(InvokeMethodRequest req)</c>
         /// </remarks>
         [AttributeUsage(AttributeTargets.Class)]
@@ -138,8 +139,8 @@ namespace Orleans
 
         /// <summary>
         /// This attribute indicates that instances of the marked grain class
-        /// will have an independent instance for each cluster with 
-        /// no coordination. 
+        /// will have an independent instance for each cluster with
+        /// no coordination.
         /// </summary>
         public class OneInstancePerClusterAttribute : RegistrationAttribute
         {
@@ -328,7 +329,7 @@ namespace Orleans
         /// Specifying [Orleans.Providers.LogConsistencyProvider] property is recommended for all grains that derive
         /// from ILogConsistentGrain, such as JournaledGrain.
         /// If no [Orleans.Providers.LogConsistencyProvider] attribute is  specified, then the runtime tries to locate
-        /// one as follows. First, it looks for a 
+        /// one as follows. First, it looks for a
         /// "Default" provider in the configuration file, then it checks if the grain type defines a default.
         /// If a consistency provider cannot be located for this grain, then the grain will fail to load into the Silo.
         /// </para>
@@ -346,18 +347,26 @@ namespace Orleans
                 ProviderName = Runtime.Constants.DEFAULT_LOG_CONSISTENCY_PROVIDER_NAME;
             }
         }
-
     }
 
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
     public sealed class ImplicitStreamSubscriptionAttribute : Attribute
     {
-        public string Namespace { get; private set; }
+        public IStreamNamespacePredicate Predicate { get; }
 
-        // We have not yet come to an agreement whether the provider should be specified as well.
+        public ImplicitStreamSubscriptionAttribute()
+        {
+            Predicate = new AllStreamNamespacesPredicate();
+        }
+
         public ImplicitStreamSubscriptionAttribute(string streamNamespace)
         {
-            Namespace = streamNamespace;
+            Predicate = new ExactMatchStreamNamespacePredicate(streamNamespace.Trim());
+        }
+
+        public ImplicitStreamSubscriptionAttribute(Type predicateType)
+        {
+            Predicate = (IStreamNamespacePredicate)Activator.CreateInstance(predicateType);
         }
     }
 }
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Text.RegularExpressions;
 using Orleans.GrainDirectory;
 using Orleans.Streams;
 
@@ -372,6 +373,15 @@ namespace Orleans
         public ImplicitStreamSubscriptionAttribute(IStreamNamespacePredicate predicate)
         {
             Predicate = predicate;
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+    public sealed class RegexImplicitStreamSubscriptionAttribute : ImplicitStreamSubscriptionAttribute
+    {
+        public RegexImplicitStreamSubscriptionAttribute(string pattern) 
+            : base(new RegexStreamNamespacePredicate(new Regex(pattern)))
+        {
         }
     }
 }
--- a/src/Orleans/Core/GrainAttributes.cs
+++ b/src/Orleans/Core/GrainAttributes.cs
@@ -350,35 +350,67 @@ namespace Orleans
         }
     }
 
+    /// <summary>
+    /// The [Orleans.ImplicitStreamSubscription] attribute is used to mark grains as implicit stream subscriptions.
+    /// </summary>
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
     public class ImplicitStreamSubscriptionAttribute : Attribute
     {
+        /// <summary>
+        /// Gets the stream namespace filter predicate.
+        /// </summary>
         public IStreamNamespacePredicate Predicate { get; }
 
+        /// <summary>
+        /// Used to subscribe to all stream namespaces.
+        /// </summary>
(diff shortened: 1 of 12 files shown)


src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
```diff
@@ -51,31 +58,39 @@ internal void InitImplicitStreamSubscribers(IEnumerable<Type> grainClasses)
...
             return result;
         }
 
+        private HashSet<int> GetOrAddImplicitSubscribersSet(string streamNamespace)
+        {
+            HashSet<int> entry;
+            if (!table.TryGetValue(streamNamespace, out entry))
+            {
+                entry = FindImplicitSubscribers(streamNamespace);
+                table[streamNamespace] = entry;
```
jason-bragg: Since table is built at runtime, rather than at start, it needs to be a concurrent dictionary, correct?  Or is their some other concurrency guarantee that I'm missing?

src/Orleans/Streams/Predicates/IStreamNamespacePredicate.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Orleans.Streams
+{
+    public interface IStreamNamespacePredicate
+    {
+        bool IsMatch(string streamNamespace);
```
jason-bragg: If we're going to support custom filtering, I suggest that we widen the filter to include stream full stream id, that is stream guid, namespace, and provider name.  Something like:

       bool IsMatch(string streamProviderName, IStreamIdentity streamIdentity);

src/Orleans/Core/GrainAttributes.cs
```diff
@@ -346,18 +347,26 @@ public LogConsistencyProviderAttribute()
                 ProviderName = Runtime.Constants.DEFAULT_LOG_CONSISTENCY_PROVIDER_NAME;
             }
         }
-
     }
 
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
     public sealed class ImplicitStreamSubscriptionAttribute : Attribute
```
jason-bragg: Suggest introduction of
abstract class ImplicitStreamSubscriptionBaseAttribute<TPredicate> where TPredicate :  IStreamNamespacePredicate. new()

with ImplicitStreamSubscriptionAttribute : ImplicitStreamSubscriptionBaseAttribute<ExactMatchStreamNamespacePredicate>

So developers can introduce their own attributes with arguments (which simply providing a type does not allow).  For instance
RegExImplicitStreamSubscriptionAttribute : ImplicitStreamSubscriptionBaseAttribute<RegExMatchStreamNamespacePredicate> with constructors that take a RegEx expression.

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
```diff
@@ -9,10 +10,12 @@ namespace Orleans.Streams
     internal class ImplicitStreamSubscriberTable
     {
         private readonly Dictionary<string, HashSet<int>> table;
+        private readonly List<Tuple<IStreamNamespacePredicate, int>> predicates;
```
jason-bragg: This class needs be serializable.  I don't think this change takes that into account.

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
DixonDs: @jason-bragg  Yes, you are right. There was `ConcurrentDictionary` initially, but then we dealing with some serializer codegen issue I reverted it back to `Dictionary` and forgot to change it back

src/Orleans/Core/GrainAttributes.cs
DixonDs: @jason-bragg I've added another ctor to `ImplicitStreamSubscriptionAttribute` and removed `sealed` to allow this kind of extensibility. I didn't go with the abstract base class you suggested, because 1) attributes cannot be generic thus you cannot put generic restrictions on the predicate type like `where TPredicate : IStreamNamespacePredicate, new()` 2) having a non-abstract attribute with a ctor accepting the predicate type seems to be useful on its own (see `FilteredImplicitSubscriptionGrain` in tests) so that you don't have to implement your own attribute just to pass the predicate type. 

So in the end, it looked like there was no extra benefit of that base attribute type.

src/Orleans/Streams/Predicates/IStreamNamespacePredicate.cs
DixonDs: @jason-bragg The only problem with this that it limits the cachebility of the subscribers in `ImplicitStreamSubscriberTable` since you have to recalculate that list per stream ID + namespace + provider name. With stream IDs having dynamic nature, I guess, it might have significant impact.

What do you think the performance penalty will be in this case? If it is insignificant, I will be happy to make this change.

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
DixonDs: @jason-bragg As far I understand, it should work as long as `IStreamNamespacePredicate` implementations are serializable. At least it is what I've seen when debugging the tests.

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
```diff
@@ -230,38 +210,23 @@ private IStreamConsumerExtension MakeConsumerReference(IInternalGrainFactory gra
...
+        private static IList<IStreamNamespacePredicate> GetPredicatesFromAttributes(Type grainClass)
         {
             if (!TypeUtils.IsGrainClass(grainClass))
             {
-                throw new ArgumentException(string.Format("{0} is not a grain class.", grainClass.FullName), "grainClass");
+                throw new ArgumentException($"{grainClass.FullName} is not a grain class.", nameof(grainClass));
             }
 
-            var attribs = grainClass.GetTypeInfo().GetCustomAttributes<ImplicitStreamSubscriptionAttribute>(inherit: false);
+            var attribs = grainClass.GetTypeInfo().GetCustomAttributes<ImplicitStreamSubscriptionAttribute>(false);
```
xiazen: I think this should be GetCustomAttributes<ImplicitStreamSubscriptionAttribute>(true) ? instead of false? since user can  inherit ImplicitStreamSubscriptionAttribute  now?

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
```diff
@@ -163,57 +173,27 @@ private bool HasImplicitSubscription(string streamNamespace, int grainIdTypeCode
...
-        /// <param name="namespaces">Namespaces instances of the grain class should subscribe to.</param>
-        /// <exception cref="System.ArgumentException">
-        /// No namespaces specified.
-        /// or
-        /// Duplicate specification of namespace "...".
-        /// </exception>
-        private void AddImplicitSubscriber(Type grainClass, ISet<string> namespaces)
+        /// <param name="streamNamespace">The stream namespace to find subscribers too.</param>
+        /// <returns></returns>
+        private HashSet<int> FindImplicitSubscribers(string streamNamespace)
```
xiazen: nit: This method is finding type code of implicit subscribers right? FindImplicitSubscriberTypeCode may be less confusing.

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
```diff
@@ -163,57 +173,27 @@ private bool HasImplicitSubscription(string streamNamespace, int grainIdTypeCode
                 return false;
             }
 
-            HashSet<int> entry;
-            return (table.TryGetValue(streamNamespace, out entry) && // if we don't have implictit subscriptions for this namespace, fail out
-                    entry.Contains(grainIdTypeCode));                 // if we don't have an implicit subscription for this type of grain on this namespace, fail out
+            HashSet<int> entry = GetOrAddImplicitSubscribersSet(streamNamespace);
```
xiazen: same here: GetOrAddImplicitSubscriberTypeCodes will be a less confusing name

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
```diff
@@ -51,31 +59,33 @@ internal void InitImplicitStreamSubscribers(IEnumerable<Type> grainClasses)
...
+                    throw new InvalidOperationException(
+                        $"Internal invariant violation: generated duplicate subscriber reference: {consumer}, subscriptionId: {subscriptionGuid}");
+                }
+                result.Add(subscriptionGuid, consumer);
             }
-
             return result;
         }
 
+        private HashSet<int> GetOrAddImplicitSubscribersSet(string streamNamespace)
```
xiazen: nit: This method is get or add type codes of implicit subscribers right? GetOrAddImplicitSubscriberTypeCodes will be a less confusing name

src/Orleans/Streams/Predicates/IStreamNamespacePredicate.cs
xiazen: nit on naming : bool Match(string namespace). IsMatch is weird to me grammatically.

xiazen: I think we should be careful here to introduce filtering on streamId  into implicit subscribe. Aside from the performance penalty DixonDs brought up, it conflicts with the nature of explicit subscribe too. Because of the dynamic nature of streamId,  filtering/subscribe to a stream fits more naturally into the territory of explicit subscribe. Adding the same feature into implicit subscribe may bring more confusion to users , potentially bugs in edge cases. I think we should defer this feature until specifically asked by users

src/Orleans/Streams/Predicates/ExactMatchStreamNamespacePredicate.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace Orleans.Streams
+{
+    [Serializable]
+    internal class ExactMatchStreamNamespacePredicate : IStreamNamespacePredicate
```
xiazen: I think we should also have a RegexStreamNameSpacePredicate 
``` 
        internal class RegexStreamNameSpacePredict
        {
            private readonly Regex regex;

            public RegexStreamNameSpacePredict( Regex regex)
            {
                this.regex = regex;
            }

            bool Match(string streamNameSpace)
            {
                return this.regex.IsMatch(streamNameSpace);
            }
        }
```
This will be a very common IStreamNamespacePredicate which many users will be interested in

xiazen: why is ExactMatchStreamNamespacePredicate  internal? it should be public to external users right?

src/Orleans/Streams/Predicates/IStreamNamespacePredicate.cs
xiazen: ignore this... I found out Regex is doing the same naming too. So IsMatch is good with me

src/Orleans/Streams/Predicates/ExactMatchStreamNamespacePredicate.cs
DixonDs: @xiazen It is used in `ImplicitStreamSubscriptionAttribute`, so I didn't see a point to expose it externally, since the users can just use `ImplicitStreamSubscriptionAttribute` with ctor accepting the stream namespace.

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
DixonDs: @xiazen Correct me if I am wrong but `inherit: true` doesn't have anything to do with the attribute inheritance, but rather the target type inheritance marked by the attribute. `GetCustomAttributes<ImplicitStreamSubscriptionAttribute>` returns not only `ImplicitStreamSubscriptionAttribute` but all its descendants - from docs: "The type of attribute to search for. Only attributes that are assignable to this type are returned."

src/Orleans/Streams/Predicates/ExactMatchStreamNamespacePredicate.cs
DixonDs: @xiazen Shall I add also `RegexImplicitStreamSubscriptionAttribute` in this case since this predicate requires a parameter to be passed?

xiazen: Yes I think you are correct! Although I'm not sure it is the correct name. 
RegexStreamNameSpacePredict didn't change the nature of ImplicitSubscribe, which is subscribe through annotation on grain class. It just change the underlying mechanism of name space filtering. 
But I cannot think of a better name. I guess I'm fine with the name. If you can come up with a better one then it is great!

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
xiazen: I think you are correct. I misread the documentation as 
```
inherit
Type: System.Boolean
true to inspect the descendants of element; otherwise, false. 
```
The correct doc is `true to inspect the ancestors of element; otherwise, false. `

src/Orleans/Core/GrainAttributes.cs
```diff
@@ -374,4 +375,13 @@ public ImplicitStreamSubscriptionAttribute(IStreamNamespacePredicate predicate)
             Predicate = predicate;
         }
     }
+
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+    public sealed class RegexImplicitStreamSubscriptionAttribute : ImplicitStreamSubscriptionAttribute
```
xiazen: Add xml comment.

src/Orleans/Streams/Predicates/RegexStreamNamespacePredicate.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Text.RegularExpressions;
+
+namespace Orleans.Streams
+{
+    [Serializable]
+    public class RegexStreamNamespacePredicate : IStreamNamespacePredicate
+    {
```
xiazen: Actually I found 0 warnings. But I think it still make sense to add some xml comment on the new predicate and RegexImplicitStreamSubscriptionAttribute . But other than that. I Approve!

Thanks for the work ! It is great

xiazen:
@dotnet-bot test this please

DixonDs:
@jason-bragg is it ready for merging?

src/Orleans/Streams/PubSub/ImplicitStreamSubscriberTable.cs
```diff
@@ -8,11 +10,13 @@ namespace Orleans.Streams
     [Serializable]
     internal class ImplicitStreamSubscriberTable
     {
-        private readonly Dictionary<string, HashSet<int>> table;
+        private readonly ConcurrentDictionary<string, HashSet<int>> table;
```
jason-bragg: Table is now just a cache, and is dynamically updated at runtime, it no longer should be serialized as part of the ImplicitStreamSubscriberTable.
The predicates are ultimately the authority and the only data that needs serialized.
Or am I missing something?

jason-bragg:
Sorry for delayed review.. One last thing, then I think it's good.
Thanks a lot!

jason-bragg:
Thanks @DixonDs 
LG2M
Will merge once tests are complete, if @xiazen doesn't beat me to it :)

PR #2989, opened 2017-05-02
Event hub generator stream provider

xiazen:
Add EventhubGeneratorStreamProvider for unit test and load test scenario. 

It can receive commands from managementGrain to set up stream generator on a stream to mimic a producer behavior. And also commands from managementGrain to stop producing on a stream

It also has good extensibility to configure different stream data generator with


src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/IEventDataGenerator.cs
```diff
@@ -0,0 +1,178 @@
...
+                    this.logger.Info($"Stop producing data on stream {streamId.Namespace}-{streamId.Guid.ToString()}.");
+                }
+            });
+        }
+        public bool TryReadEvents(int maxCount, TimeSpan waitTime, out IEnumerable<EventData> events)
+        {
+            var eventDataList = new List<EventData>();
+            this.generators.ForEach(generator => {
+                IEnumerable<EventData> eventData;
+                if (generator.TryReadEvents(maxCount, waitTime, out eventData))
```
jason-bragg: won't this read maxcount * generater.count events rather than just maxcount events?

xiazen: You are definitely correct..... I apologize

xiazen: This can do maxEvent/numOfGenerator

jason-bragg:
It can, but that assumes each generator returns that amount.  I'd suggest getting one event from each generator until maxCount is reached, or asking for maxcount/numOfGenerator and counting the events returned until maxEvents are accumulated or generators create no more events.

xiazen:
maxCount is the cap count for events reading from the eventhub partition, not the exact count for every read. But yeah we can do getting one event from each generator. If it reaches maxCount, then good, otherwise we still deliver.

PR #2990, opened 2017-05-02
ClientGenerator packs OrleansCodeGenerator.Build package

jdom:
- ClientGenerator.csproj generates the nuget package directly
- Remove meta package and scripts that build the built-time codegen package separately
- Allow setting nuget version from msbuild arguments


vNext/src/ClientGenerator/ClientGenerator.csproj
```diff
@@ -18,21 +24,73 @@
...
+    </ProjectReference>
   </ItemGroup>
+
+  <ItemGroup>
+    <None Include="..\Orleans.SDK.targets">
+      <Pack>true</Pack>
+      <PackagePath>build\$(PackageId).targets</PackagePath>
+      <Visible>false</Visible>
+    </None>
+    <None Include="$(OutDir)/ClientGenerator.exe.config">
```
ReubenBond: It's a little unfortunate that these need to be included manually instead of the references being automatically included in the package, but so be it, I suppose.

jdom: yes, I think there is room for improvement... Ideally we can use a build property that already has all the files that end up in the output folder. I couldn't find it, but at least I preserved the old behavior (plus including PlatformServices, which for some reason wasn't being included before, not sure how it was working)

jdom:
Seems that this is an issue for others too: https://github.com/NuGet/Home/issues/4360

Once we start multi-targeting ClientGenerator, we should probably drop IsTool, and instead use the approach suggested by @AArnott in that thread

Issue #2893, opened 2017-03-28
Enable grain methods with void return


ReubenBond:
Internally, Orleans has the capability to support void-returning remote method calls. These methods are fire-and-forget and there is no natural way for the caller to determine whether or not the message was delivered and processed.

Currently, void methods are only enabled for `IGrainObserver` interfaces.

I propose we expose this functionality for all users.
The alternative today is to call `.Ignore()` on a task a user doesn't care about, but this still incurs the performance penalty of scheduling a continuation.

Does anyone have objections?

dVakulen:
Just referencing some discussion from gitter on this topic - https://gitter.im/dotnet/orleans?at=56b074f6dc33b33c75494db0

dVakulen:
Also cost of continuation scheduling is negligible compared to the price of sending unneeded response message back to the caller.

gabikliot:
Orleans in an RPC system. Rpc by definition is two way. Try suggesting to add a one way messaging to grpc for example.
It's paradigm mismatch. Observers were a mistake in my view and should be discouraged instead of extending them.

sergeybykov:
I can argue both sides here. :-) 

On the one hand, I believe our async RPC model encourages writing safer code with proper error handling, introduces implicit natural backpressure, and eliminates the need for error prone correlation of one-way requests and responses. On the other hand, there are legitimate cases where unreliable one-way messages are all the app needs, and sending empty responses is wasteful. Coincidentally, I was in a discussion with Minecraft about one such design recently.

I think there are two loosely connected parts of the picture here.

1) On the caller's side, a void method call to a grain looks indistinguishable from making call to a `Task`-based method and `Ignore`'ing the returned `Task`, just shorter and cleaner.

2) On the grain side, a `void` method I believe is undesirable because it doesn't fit the async execution model, doesn't allow to make async calls from it, etc. And `async void` isn't a solution. It's a problem that I think we don't want to touch with a 6ft pole. It was meant for very specific use cases, and grain methods aren't even close to those.

I think the ideal case would be if we could allow `void` methods in grain interfaces, but then somehow make them `Task` methods on the grain class that implements it. Unfortunately, the basic interface implementation rules don't allow that.

[...]

PR #2993, opened 2017-05-03
Support one-way grain calls using [OneWay] method attribute

ReubenBond:
This is an implementation of my take on #2893. I don't necessarily intend for this to be implemented, but maybe it demonstrates what's required and how it might look.

When a grain interface has a method is marked with `[OneWay]`, that method will be completed synchronously from the caller's perspective and the remote silo will not send a response message.

An alternative implementation could use a different return type as a marker, like `IOneWay` or `OneWayCall`.

### DIFF
--- a/src/OrleansCodeGenerator/GrainReferenceGenerator.cs
+++ b/src/OrleansCodeGenerator/GrainReferenceGenerator.cs
@@ -208,14 +208,14 @@ namespace Orleans.CodeGenerator
 
                     if (isOneWayTask)
                     {
-                        if (!method.ReturnType.IsInstanceOfType(TaskDone.Done))
+                        if (method.ReturnType != typeof(Task))
                         {
                             throw new CodeGenerationException(
                                 $"Method {grainType.GetParseableName()}.{method.Name} is marked with [{nameof(OneWayAttribute)}], " +
-                                $"but has a return type which is not assignable from {TaskDone.Done.GetType()}");
+                                $"but has a return type which is not assignable from {typeof(Task)}");
                         }
 
-                        var done = typeof(TaskDone).GetNameSyntax(true).Member((object _) => TaskDone.Done);
+                        var done = typeof(Task).GetNameSyntax(true).Member((object _) => Task.CompletedTask);
                         body.Add(SF.ReturnStatement(done));
                     }
                 }
--- a/src/OrleansCodeGenerator/GrainReferenceGenerator.cs
+++ b/src/OrleansCodeGenerator/GrainReferenceGenerator.cs
@@ -11,6 +11,7 @@ namespace Orleans.CodeGenerator
     using Microsoft.CodeAnalysis.CSharp.Syntax;
     using Orleans.CodeGeneration;
     using Orleans.CodeGenerator.Utilities;
+    using Orleans.Concurrency;
     using Orleans.Runtime;
     using GrainInterfaceUtils = Orleans.CodeGeneration.GrainInterfaceUtils;
     using SF = Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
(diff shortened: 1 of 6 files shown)


src/OrleansCodeGenerator/GrainReferenceGenerator.cs
```diff
@@ -203,6 +205,19 @@ private static MemberDeclarationSyntax[] GenerateInvokeMethods(Type grainType, A
...
 
                     body.Add(SF.ExpressionStatement(invocation));
+
+                    if (isOneWayTask)
+                    {
+                        if (!method.ReturnType.IsInstanceOfType(TaskDone.Done))
+                        {
+                            throw new CodeGenerationException(
+                                $"Method {grainType.GetParseableName()}.{method.Name} is marked with [{nameof(OneWayAttribute)}], " +
+                                $"but has a return type which is not assignable from {TaskDone.Done.GetType()}");
```
jdom: why are you using TaskDone.Done instead of just `typeof(Task)`? Am I missing and edge case? Is it because we might want to support `AsyncValue`/ `TaskAwaiter`/`INotifyCompletion` in te future so you are protecting from those changes?

src/OrleansCodeGenerator/GrainReferenceGenerator.cs
```diff
@@ -191,7 +192,8 @@ private static MemberDeclarationSyntax[] GenerateInvokeMethods(Type grainType, A
                 var options = GetInvokeOptions(method);
 
                 // Construct the invocation call.
-                if (method.ReturnType == typeof(void))
+                var isOneWayTask = method.GetCustomAttribute<OneWayAttribute>() != null;
+                if (method.ReturnType == typeof(void) || isOneWayTask)
```
jdom: I was going to suggest to also support void returning methods only if they are marked OneWay, so that if the grain implementation doesn't do any asynchronous work, that both the caller and implementation benefit of not involving tasks in their usage.
I assume this is not enough and codegen would need to be updated to allow that path, correct?

jdom:
I think this approach makes a lot of sense, as @ashkan-saeedi-mazdeh makes a good point that turning a 2-way method into a 1-way invocation that the caller is responsible to do is not very common (although still a possibility, so @dVakulen's suggestion shouldn't be discarded, and could be implemented too).
Many times the API needs to be designed and the implementation be made very aware that it is a one way call.

The only (minor) drawback I see with this particular implementation is that the attribute is not really part of the usage pattern, so the caller might not be aware that this is a one-way call. Having a marker return type would make that explicit. Additionally, depending on the shape of that return type, it could allow the caller not to have to await this ficticious always-completed task, and the implementer to decide whether it runs synchronously without allocating any Task or return a value that has a Task.

ReubenBond:
@jdom I fully agree with your points. We cannot use generalized Task-like types yet, can we? I guess the user could use C# 7 and then have access to async/await for any compliant type. So I'm happy if we don't merge this if we prefer to go for a custom awaitable.

src/OrleansCodeGenerator/GrainReferenceGenerator.cs
```diff
@@ -191,7 +192,8 @@ private static MemberDeclarationSyntax[] GenerateInvokeMethods(Type grainType, A
                 var options = GetInvokeOptions(method);
 
                 // Construct the invocation call.
-                if (method.ReturnType == typeof(void))
+                var isOneWayTask = method.GetCustomAttribute<OneWayAttribute>() != null;
```
sergeybykov: Shouldn't we constrain it even further to only allow methods returning `Task` to be marked `[OneWay]`, and throw an error if a method returning `Task<T>` has the attribute? I mean to stress that no value is coming back.

ReubenBond: You're right - that would be better. If we decide that this PR is desirable then I'll make that change.

ReubenBond: I've made the change in the latest revision

ReubenBond:
Pushed an update to restrict the type to `Task` and use `Task.CompletedTask` instead of `TaskDone.Done`

ashkan-saeedi-mazdeh:
I highly agree with @jdom  about a marker type as return type which means the caller will see it explicitly. sorry for duplicate answer here and on the issue. Just thought that it might be good that it is both here and there.

PR #2995, opened 2017-05-03
Changed TaskDone.Done to Task.CompletedTask

DixonDs:
Since the code is migrated to .Net 4.6.1 now, it seems there is no point to use non-standard `TaskDone.Done` instead of `Task.CompletedTask` anymore. I marked it as deprecated and changed all usages to `Task.CompletedTask`.

Feel free to close this PR without merging if you feel that this change has too little value.


jdom:
Thanks @DixonDs!

Issue #2997, opened 2017-05-04
Stack overflow from ILBasedExceptionSerializer


josiahdecker:
When using the ILBasedSerializer as the fallback serializer an exception that has a circular reference to itself will cause a stack overflow.  A notable example of this is Microsoft.WindowsAzure.Storage.StorageException from the azure storage sdk.  

It looks like the root of the issue is in ILBasedExceptionSerializer.Serialize.  A new SerializationContext is created when serializing an exception, this defeats the check for back-references in the SerializationManager.SerializeInner method allowing for infinite recursion.

Partial stack tracing showing the loop with the StorageException type:
[storage_ex_stack.txt](https://github.com/dotnet/orleans/files/977204/storage_ex_stack.txt)

ReubenBond:
@josiahdecker is it possible you could provide a repro? Thanks :)

josiahdecker:
Repro here
https://github.com/josiahdecker/OrleansExceptionRepro

ReubenBond:
@josiahdecker thanks again - I'll start on a fix now

ReubenBond:
@josiahdecker I've opened PR #2999 with a fix

ReubenBond:
Fixed by #2999, which will be included in the upcoming 1.5.0 release

PR #2999, opened 2017-05-04
Support exceptions with reference cycles in ILBasedExceptionSerializer

ReubenBond:
The [initial PR](https://github.com/dotnet/orleans/pull/2633#issue-203027973) for failsafe exception serialization came with the caveat that it did not support self-referential exceptions.

Unfortunately, this deficiency has affected @josiahdecker (see #2997), so it should be fixed.

This PR fixes the issue by adding support for nested `IDeserializationContext` instances. It's quite simple:
* Instead of reading position from the stream reader/writer, read it from a property on the context.
* Add an extension method to nest contexts, where the nested context is created with a specified offset from its parent.

### DIFF
--- a/src/Orleans/Serialization/DeserializationContext.cs
+++ b/src/Orleans/Serialization/DeserializationContext.cs
@@ -45,18 +45,18 @@ namespace Orleans.Serialization
     public static class DeserializationContextExtensions
     {
         /// <summary>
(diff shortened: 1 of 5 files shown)


ReubenBond:
I consider this ready for review, if anyone has time.

PR #3001, opened 2017-05-08
First iteration of versioning documentation

benjaminpetit:
I am sure we can add a lot more than that, feel free to suggest some modifications


src/Documentation/Grain-Versioning/Backward-compatibility-guidelines.md
```diff
@@ -0,0 +1,122 @@
...
+{
+  // Method inherited from V1
+  Task MyMethod(int arg, obj o);
+}
+```
+
+## Avoid changing existing method logic
+
+It can seems obvious, but you should be very careful when changing the body of an existing method.
+Unless you are fixing a bug, it is better to just add a new method if you need to modify the code.
```
ReubenBond: We should include a motivating example of why - what can go wrong here?

ReubenBond:
Great! Could you also include rebuilt docs in a separate commit?

jdom:
Let's iterate first... it's easier to review and fix this way, since it might require updates

src/Documentation/Grain-Versioning/Backward-compatibility-guidelines.md
```diff
@@ -0,0 +1,122 @@
...
+```
+``` cs
+[Version(2)]
+public interface IMyGrain : IGrainWithIntegerKey
+{
+  // Method inherited from V1
+  Task MyMethod(int arg);
+
+  // New method added in V2
+  Task MyNewMethodMethod(int arg, obj o);
```
jdom: nit: `Method` twice in the name is odd. I see it in many places, so it might have been by design, but I don't really understand it.

src/Documentation/Grain-Versioning/Backward-compatibility-guidelines.md
```diff
@@ -0,0 +1,122 @@
...
+    // Method inherited from V1
+    [Obsolete]
+    Task MyMethod(int arg);
+
+    // New method added in V2
+    Task MyNewMethodMethod(int arg, obj o);
+  }
+  ```
+
+2. When you are __sure__ that no V1 calls are made, deploy V3 with V1 method removed
```
jdom: When you are __sure__ that no V1 calls are made (effectively V1 is no longer deployed in the running cluster), deploy V3 with V1 method removed

src/Documentation/Grain-Versioning/Grain-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Enable versioning
+By default, grains are not versioned. You can version grain by using the VersionAttribute on the grain interface:
+
+``` cs
+[Version(X)]
+public interface IVersionUpgradeTestGrain : IGrainWithIntegerKey {}
+```
+
+Where `X` is the version of the grain interface.
```
jdom: Where `X` is the version number of the grain interface, which is typically monotonically increasing.

src/Documentation/Grain-Versioning/Grain-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Grain version compatibility and placement
+When a call from a versioned grain arrive in a cluster:
+-	If no activation exists, a compatible activation will be created
+-	If an activation exists:
+ - If the current one is not compatible, it will be deactivated and new compatible will be created
+ - If the current one is compatible, the call will be handled normally
```
jdom: indenting is off. Please don't use tabs here, but add spaces before the `-` for nested lists

src/Documentation/Grain-Versioning/Grain-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+
+By default:
+-	All versioned grains are supposed to be backward-compatible only (see [backward compatibility guidelines](Backward-compatibility-guidelines.md)). That means that a v1 grain can make calls to a v2 grain, but a v2 grain cannot call a v1. This default behavior can be changed with `GlobalConfiguration.DefaultCompatibilityStrategy`
+-	When multiple versions exist in the cluster, the new activation will be randomly placed on a compatible silo. This default behavior can be changed with `GlobalConfiguration.DefaultVersionSelectorStrategy`
```
jdom: you talk about defaults here, but there's a missing gap saying that there are configurable policies to define what *Compatible* is and what is the new placement policy.

src/Documentation/Grain-Versioning/Grain-versioning.md
ReubenBond: Do we support the case where X2 < X1? I believe X must be monotonically increasing.

jdom: I said typically, since that's what we expect more often. Although you could of course do targeted bug fixes that break this, or play with the compatibility policies

src/Documentation/Grain-Versioning/Azure-deployment.md
```diff
@@ -0,0 +1,49 @@
...
+
+When using this configuration, "old" clients will be able to talk to activations
+on both versions of silos. Newer clients and silos will only trigger new activations
+on newer silos.
+
+![Rolling gif](rolling.gif)
+
+### Using a staging environment
+
+In this method you deploy a new cspkg in the Staging environment before doing
```
sergeybykov: This paragraph is very Azure specific without mentioning it. Should it be more generic?

src/Documentation/Grain-Versioning/Grain-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+> Grain state it out of the scope.
+
+## Overview
+On a given cluster, silos can support different versions of a grain type.
+![Cluster with different versions of a grain](version.png)
+In this example the client and Silo{1,2,3} were compiled with grain interface `A` version 1. Silo 4 was compiled with `A` version 2.
+
+## Limitations:
+-	No versioning on stateless worker
+-	Streaming agents are not versioned
```
sergeybykov: Maybe "Streaming interfaces" instead? SMS that doesn't use pulling agents uses the same interfaces.

src/Documentation/Grain-Versioning/Backward-compatibility-guidelines.md
benjaminpetit: Lazy copy-paste :p

src/Documentation/Grain-Versioning/Azure-deployment.md
benjaminpetit: The title of the pahe is "Deploy new versions in Azure" it was not intended to be generic. Should I do instead a generic page?

src/Documentation/Grain-Versioning/Grain-versioning.md
benjaminpetit: Theoratically, you could start at version 999 and decrease the number for each new version, but it doesn't seems very natural to me. But since we could do it anyway (rollback previous version for instance) I think the suggestion from @jdom is valid

src/Documentation/Grain-Versioning/Backward-compatibility-guidelines.md
```diff
@@ -0,0 +1,122 @@
...
+  // Method inherited from V1
+  // Do not change the body
+  Task MyMethod(int arg)
+  {
+    SomeSubRoutine(arg);
+  }
+
+  // New method added in V2
+  Task MyNewMethodMethod(int arg)
+  {
```
xiazen: nit, an extra "Method" here too

src/Documentation/Grain-Versioning/Grain-versioning.md
xiazen: Maybe elaborate a little more on "Streaming agents are not versioned". Do you mean the public streaming interface, like IObserver, IObservable does not support versioning?  or streaming infrastructure doesn't support rolling upgrading, so if Orleans streaming has a breaking change in between two versions, and doing a rolling upgrade from Orleans v1 to v2, you will be facing failures?

src/Documentation/Grain-Versioning/Backward-compatibility-guidelines.md
```diff
@@ -0,0 +1,122 @@
...
+  ```
+
+2. When you are __sure__ that no V1 calls are made, deploy V3 with V1 method removed
+  ``` cs
+  [Version(3)]
+  public interface IMyGrain : IGrainWithIntegerKey
+  {
+    // New method added in V2
+    Task MyNewMethodMethod(int arg, obj o);
+  }
```
xiazen: nit: MethodMethod

src/Documentation/Grain-Versioning/Grain-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+``` cs
+[Version(X)]
+public interface IVersionUpgradeTestGrain : IGrainWithIntegerKey {}
+```
+
+Where `X` is the version of the grain interface.
+
+## Grain version compatibility and placement
+When a call from a versioned grain arrive in a cluster:
+-	If no activation exists, a compatible activation will be created
```
xiazen: "a compatible activation will be created " does not have very clear meaning to me. If there's multiple compatible versions, do we choose a random one or do we choose a minimum one or we chose based on the  VersionSelectorStrategy? I think more explanation on the behavior will be helpful to readers

xiazen: I'm not that familiar with versioning. I guess that makes me the perfect reader for this PR :D. But my question can seem out of blue I guess.

src/Documentation/Grain-Versioning/Grain-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+## Grain version compatibility and placement
+When a call from a versioned grain arrive in a cluster:
+- If no activation exists, a compatible activation will be created
+- If an activation exists:
+  - If the current one is not compatible, it will be deactivated and new compatible will be created (see [version selector strategy](Version-selector-strategy.md))
+  - If the current one is compatible, the call will be handled normally (see [compatible grains](Compatible-grains.md))
+
+By default:
+- All versioned grains are supposed to be backward-compatible only (see [backward compatibility guidelines](Backward-compatibility-guidelines.md)). That means that a v1 grain can make calls to a v2 grain, but a v2 grain cannot call a v1. This default behavior can be changed with `GlobalConfiguration.DefaultCompatibilityStrategy`
+- When multiple versions exist in the cluster, the new activation will be randomly placed on a compatible silo. This default behavior can be changed with `GlobalConfiguration.DefaultVersionSelectorStrategy`
```
jdom: you talk about defaults here and the previous paragraph, but there's a missing gap saying that there are configurable policies to define what Compatible is/means and what is the placement policy. Then you can say that these are the default policies

src/Documentation/Grain-Versioning/Compatible-grains.md
```diff
@@ -0,0 +1,22 @@
...
+
+In this document will be described the notion of "backward compatible" and "fully compatible".
+
+## Backward compatible
+
+A grain interface version Vn is said to be backward compatible with Vm if:
+
+  - The name of the interface didn't change (or the overridden typecode)
+  - All public methods present in the Vm version are in the Vn version. __It is important that
+    the signatures of the methods inherited from Vm are not modified__.
```
jdom: a note about message serialization would be useful too, since if you use our magic serializer, just by modifying a single private field in a class part of the object graph you send to a grain can make the serialization to break

src/Documentation/Grain-Versioning/Version-selector-strategy.md
```diff
@@ -0,0 +1,40 @@
...
+## Minimum version
+
+Using this strategy, the version of the new activation will always be the requested or the
+minimum compatible version.
+
+For example if we have 2 versions of a given grain interface, V2, V3, all fully 
+compatibles:
+
+  - If the request was made from a V1 client/silo, the new activation will be a V2
+  - If the request was made from a V3 client/silo, the new activation will be a V3
```
jdom: Is this right? You said that all are fully compatible (not back compat), so wouldn't that mean that the last scenario would activate a V2 grain?

src/Documentation/Grain-Versioning/Compatible-grains.md
```diff
@@ -0,0 +1,22 @@
+# Compatible grains
+
+In this document will be described the notion of "backward compatible" and "fully compatible".
+
+## Backward compatible
```
jdom: This section is confusing, because it seems you are talking about what to have in mind to create versions that are back compat or fully back and forward compat, but it's not cleat that apart from these tips, there is a policy that can be configured and will actually be used. 
Also I would tie/link this with the `Version selector strategy` doc.

src/Documentation/Grain-Versioning/Version-selector-strategy.md
benjaminpetit: You are right, I made so many changes that I forgot what they do :)

src/Documentation/Grain-Versioning/Compatible-grains.md
```diff
@@ -1,22 +1,50 @@
...
+### Definition
 
 A grain interface version Vn is said to be backward compatible with Vm if:
 
   - The name of the interface didn't change (or the overridden typecode)
   - All public methods present in the Vm version are in the Vn version. __It is important that
-    the signatures of the methods inherited from Vm are not modified__.
+    the signatures of the methods inherited from Vm are not modified__: since Orleans use
+    an internal built-in serializer, modifying/renaming a field (even private) can make the
+    serizliation to break.
```
jdom: typo in serizlization

src/Documentation/Grain-Versioning/Compatible-grains.md
```diff
@@ -1,22 +1,50 @@
...
+
+### Definition
 
 A grain interface version Vn is said to be backward compatible with Vm if:
 
   - The name of the interface didn't change (or the overridden typecode)
   - All public methods present in the Vm version are in the Vn version. __It is important that
-    the signatures of the methods inherited from Vm are not modified__.
+    the signatures of the methods inherited from Vm are not modified__: since Orleans use
+    an internal built-in serializer, modifying/renaming a field (even private) can make the
```
jdom: I would clarify that this is prescriptive guidance you should follow when using the Backward compatible policy. It is not clear from reading this that following these tips is separate from configuring the policy.
Reading this it might look like Orleans would infer at runtime which policy to use depending on the deployed code and how compatible it is from version to version.

src/Documentation/Grain-Versioning/Backward-compatibility-guidelines.md
```diff
@@ -47,6 +47,34 @@ public interface IMyGrain : IGrainWithIntegerKey
 }
 ```
 
+Example of what can happen if you just rename the parameter names, let's say
+that we have the two following interface version deployed in the cluster:
+``` cs
```
jdom: Add `**NOTE**: you should not do this change in your code, as it's an example of a bad practice that leads to very bad side-effects.`

src/Documentation/Grain-Versioning/Compatible-grains.md
```diff
@@ -0,0 +1,50 @@
...
+## Backward compatible (default)
+
+### Definition
+
+A grain interface version Vn is said to be backward compatible with Vm if:
+
+  - The name of the interface didn't change (or the overridden typecode)
+  - All public methods present in the Vm version are in the Vn version. __It is important that
+    the signatures of the methods inherited from Vm are not modified__: since Orleans use
+    an internal built-in serializer, modifying/renaming a field (even private) can make the
```
jdom: I would clarify that this is prescriptive guidance you should follow when using the Backward compatible policy. It is not clear from reading this that following these tips is separate from configuring the policy.
Reading this it might look like Orleans would infer at runtime which policy to use depending on the deployed code and how compatible it is from version to version.

src/Documentation/Grain-Versioning/Compatible-grains.md
```diff
@@ -0,0 +1,50 @@
...
+    serizliation to break.
```
jdom: typo in serizlization

src/Documentation/Grain-Versioning/Compatible-grains.md
benjaminpetit: I added a phrase in the introduction to reflect this

PR #1901, opened 2016-07-04
Do not require explicitly registering grains in ServiceCollection

jdom:
This is similar to MVC, where the framework is in charge of determining which concrete controller to resolve, and the user is not required to explicitly register the controller, only the dependencies used by the controller.
Grains in Orleans behave in a very similar way where as you can't explicitly register the mapping grain interface -> grain implementation in the container, and instead Orleans determines which concrete Grain implementation to instantiate. The user should not be responsible for registering all the grain implementations that were also already discovered by Orleans. This is boilerplate and non-intuitive (as you were forced to register the concrete type, not the mapping with the interface, which is not common when using DI and especially when you are implementing an interface for consumer to use).

This supersedes #1841, as there is no "fallback" if the registration isn't there, it is instead how it is supposed to work.

EDIT: I made it exactly like MVC's implementation, so that GrainCreator will not even consider explicit registrations for grains (or controllers in the MVC case).

Classes in MVC that are relevant to controller activation:
- [DefaultControllerActivator](https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerActivator.cs)
- [TypeActivatorCache](https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/Internal/TypeActivatorCache.cs)
- [ActivatorUtilities](https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.ActivatorUtilities.Sources/ActivatorUtilities.cs) and its [façade](https://github.com/aspnet/DependencyInjection/blob/master/src/Microsoft.Extensions.DependencyInjection.Abstractions/ActivatorUtilities.cs)

PR #3002, opened 2017-05-08
Add extensibility point to replace the grain activator

jdom:
This will instantiate grains preserving the current behavior (which is using `ActivatorUtilities`, similar to MVC
s `DefaultControllerActivator`), but allows replacing this service with a custom activator.
This enables less constraining integration with some 3rd party containers, or even manually instantiating grains.

In the future, I envision the `IGrainActivationContext` instance be resolved as a scoped service as defined in #2856.

### DIFF
--- a/src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
+++ b/src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Globalization;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.Runtime
+{
+    /// <summary>
+    /// <see cref="IGrainActivator"/> that uses type activation to create grains.
+    /// </summary>
+    public class DefaultGrainActivator : IGrainActivator
+    {
+        private readonly Func<Type, ObjectFactory> createFactory = type => ActivatorUtilities.CreateFactory(type, Type.EmptyTypes);
+        private readonly ConcurrentDictionary<Type, ObjectFactory> typeActivatorCache = new ConcurrentDictionary<Type, ObjectFactory>();
+
+        /// <inheritdoc />
+        public virtual object Create(IGrainActivationContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+
+            var grainType = context.GrainType;
+
+            if (grainType == null)
+            {
+                throw new ArgumentException(
+                    string.Format(
+                        CultureInfo.InvariantCulture, "The '{0}' property of '{1}' must not be null.",
+                        nameof(context.GrainType),
+                        nameof(IGrainActivationContext)));
+            }
+
+            var serviceProvider = context.ActivationServices;
+            var activator = this.typeActivatorCache.GetOrAdd(grainType, this.createFactory);
+            var grain = activator(serviceProvider, arguments: null);
+            return grain;
+        }
+    }
+}
--- a/src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
+++ b/src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
@@ -38,5 +38,25 @@ namespace Orleans.Runtime
             var grain = activator(serviceProvider, arguments: null);
             return grain;
         }
+
+        /// <inheritdoc />
+        public virtual void Release(IGrainActivationContext context, object grain)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (grain == null)
+            {
+                throw new ArgumentNullException(nameof(grain));
+            }
+
+            var disposable = grain as IDisposable;
+            if (disposable != null)
+            {
+                disposable.Dispose();
+            }
+        }
     }
 }
--- a/src/OrleansRuntime/Catalog/GrainCreator.cs
+++ b/src/OrleansRuntime/Catalog/GrainCreator.cs
@@ -96,5 +96,10 @@ namespace Orleans.Runtime
 
             ((ILogConsistentGrain)grain).InstallAdaptor(factory, state, grainType.FullName, storageProvider, svc);
         }
+
+        public void Release(IGrainActivationContext context, object grain)
+        {
+            this.grainActivator.Release(context, grain);
+        }
     }
 }
--- a/src/OrleansRuntime/Catalog/GrainCreator.cs
+++ b/src/OrleansRuntime/Catalog/GrainCreator.cs
@@ -1,63 +1,51 @@
 ﻿using System;
-using System.Collections.Concurrent;
-using Microsoft.Extensions.DependencyInjection;
 using Orleans.Core;
 using Orleans.LogConsistency;
 using Orleans.Storage;
 using Orleans.Runtime.LogConsistency;
 using Orleans.GrainDirectory;
-using Orleans.Serialization;
-using Orleans.Runtime.MultiClusterNetwork;
-using Orleans.Runtime.Configuration;
 
 namespace Orleans.Runtime
 {
     /// <summary>
-    /// Helper class used to create local instances of grains.
+    /// Helper class used to create local instances of grains. In the future this should be opened up for extension similar to ASP.NET's ControllerFactory.
     /// </summary>
     internal class GrainCreator
     {
+        private readonly IGrainActivator grainActivator;
+
         private readonly Lazy<IGrainRuntime> grainRuntime;
-
-        private readonly IServiceProvider services;
-
-        private readonly Func<Type, ObjectFactory> createFactory;
-
-        private readonly ConcurrentDictionary<Type, ObjectFactory> typeActivatorCache = new ConcurrentDictionary<Type, ObjectFactory>();
         
         private readonly Factory<Grain, IMultiClusterRegistrationStrategy, ProtocolServices> protocolServicesFactory;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="GrainCreator"/> class.
         /// </summary>
-        /// <param name="services">Service provider used to create new grains</param>
+        /// <param name="grainActivator">The activator used to used to create new grains</param>
         /// <param name="getGrainRuntime">The delegate used to get the grain runtime.</param>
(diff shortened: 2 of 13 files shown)


jdom:
cc @Eldar1205 
Eldar, would this be a suitable extension point for you? For usage, check the new tests

jdom:
Reason: unrelated failure: https://ci.dot.net/job/dotnet_orleans/job/master/job/netfx-functional_prtest/236/testReport/junit/NonSilo.Tests.UnitTests.SerializerTests/MessageSerializerTests/MessageTest_TtlUpdatedOnSerialization/

Eldar1205:
Thanks @jdom, this will work for me, some questions/comments:

1. I think the name IGrainActivator is misleading, grain activation is an Orleans term much more broad than grain instance resolution. I suggest the name IGrainResolutionBehavior or some other name not mixed with existing Orleans concepts.

2. Why does the IGrainActivator receive an object and not a Grain? I find it better for compile time safety since all grains should inherit Grain. In particular it's important to distinguish between grain references on Orleans client side and actual Grain instances on Orleans cluster side.

src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Concurrent;
+using System.Globalization;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.Runtime
+{
+    /// <summary>
+    /// <see cref="IGrainActivator"/> that uses type activation to create grains.
+    /// </summary>
+    public class DefaultGrainActivator : IGrainActivator
```
ReubenBond: I was about to suggest that maybe `IGrainFactory` was a more suitable name. Nevermind... 😕 would `IGrainCreator` make sense, since this doesn't actually call the activation lifecycle method on the grain?

src/OrleansRuntime/Catalog/GrainCreator.cs
```diff
@@ -109,5 +96,10 @@ public Grain CreateGrainInstance(Type grainType, IGrainIdentity identity, Type s
 
             ((ILogConsistentGrain)grain).InstallAdaptor(factory, state, grainType.FullName, storageProvider, svc);
         }
+
+        public virtual void Release(IGrainActivationContext context, object grain)
```
ReubenBond: Does this need to be virtual? Seems like a leftover from a previous revision.

src/OrleansRuntime/Catalog/IGrainActivationContext.cs
```diff
@@ -0,0 +1,20 @@
...
+    public interface IGrainActivationContext
+    {
+        /// <summary>Gets the .NET type of the grain activation instance.</summary>
+        Type GrainType { get; }
+
+        /// <summary>Gets the identity of the grain activation.</summary>
+        IGrainIdentity GrainIdentity { get; }
+
+        /// <summary>Gets the <see cref="IServiceProvider"/> that provides access to the grain's service container.</summary>
+        IServiceProvider GrainServices { get; }
```
ReubenBond: Potential confusion between this name and the feature known as GrainServices

src/OrleansRuntime/Catalog/GrainCreator.cs
jdom: This was by design, since I expect most custom implementations will inherit from the default one and call the base in several occasions, similar to the one in our tests. This copies the DefaultControllerActivator from MVC

src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        private readonly Func<Type, ObjectFactory> createFactory = type => ActivatorUtilities.CreateFactory(type, Type.EmptyTypes);
```
jason-bragg: Isn't this all that is needed?  A hook to create the grain factory?

Func<Type, object[] parameters, ObjectFactory>?

Get this one thing from DI, and nothing else needs to change, right?

jdom: Unfortunately not, since the need is to also perform some cleanup tasks after the instance is done with. Also, this was part of the work started in #1901 where we wanted to emulate MVC's controller activation strategy, which to be honest, make a lot of sense for us too.
This matches 1 to 1 with [DefaultControllerActivator](https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerActivator.cs) (except that they have an extra layer of abstraction for ITypeActivatorCache which I didn't add here, as it doesn't add any value yet)

src/OrleansRuntime/Catalog/IGrainActivationContext.cs
jdom: Hmmm, can't think of a better name that would stay within the familiarity with MVC naming conventions. Perhaps `GrainActivationServices` (which is correct, I just didn't use it because it felt too verbose)?

src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
jdom: GrainFacory (or GrainActivationFactory) would be more appropriate to the abstraction that will replace our current GrainCreator, as it matches [DefaultControllerFactory](https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerFactory.cs), which in our case can have stuff like the facet system we are planning in the future (MVC does a VERY similar thing to our facets).
Maybe `IGrainActivationCreator`? I don't know, none seem right, but at least IGrainActivator matches MVC naming, so we at least had familiarity on our side, but I'm definitely open to change this

jdom:
@Eldar1205 
1. re: naming: touched on that on another comment. I agree naming is not ideal, but stated why I still find it preferable to some other options. Nevertheless I'm open to change it if we can find a better one. Let's continue that discussion on the other comment to unify threads.
2. The reason for `object` instead of `Grain` is that it does leave the door open to have grains that do not inherit from Grain in the future, but at the same time there is not a lot that is gained from typing it to Grain anyway. Also, it matches the MVC extensibility point called IControllerActivator.

src/OrleansRuntime/Catalog/IGrainActivationContext.cs
attilah: I just not fond of loosing the "Provider" suffix, but as we talked ApplicationServices maybe a better name?

src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
jason-bragg: Concerns:
Insufficient design - Grain lifecycle needs work, I think we all agree, but it's also very important to get it right.  I am of the opinion that upfront design is necessary.  I do not suspect that it is sufficient or desirable to simply apply the mvc controller pattern to this because they look similar.
I advocate that we do the design work necessary to address grain lifecycle needs, then, after due diligence, if the resulting design is comparable to MVC's controller pattern, then we apply it as it is more familiar to users.  This ensures we're applying the right pattern not forcing the solution into a pattern that is already familiar.
Churn - The issue that Eldar is experiencing is in part due to the fact that we've been working on this system piecemeal, creating churn.  I advocate up front design to make fewer and more complete changes with less churn.  Making a minimal change to address the previous braking change is churn, but not as much as performing this change and the risk of further changes due to lack of upfront design.
Risk - We want to get this in for 1.5, which is already in beta, which means this change will not get as much community testing as the rest of 1.5, and this is critical code.  A minimal change is far less risky.

These are my concerns, however I will happily defer to the teams decision if most of those interested in this change do not share my concerns.

src/OrleansRuntime/Catalog/IGrainActivationContext.cs
jdom: well, that's the point, it's not the same as ApplicationServices (actually it currently is the same instance, but this will definitely change once we add scoped services in #2856). This will have the IServiceProvider scoped to this particular grain activation, similar to MVC's context.RequestServices, which is a scoped instance separate from their ApplicationServices.

Eldar1205:
Hi,

Regarding object being open to not have Grain as base class, you can always
change to object later. Until then I think the type safety gained using
Grain will benefit coherence and testability for both application code and
framework code.
Regarding interface name, aligning with MVC shouldnt be on expanse of
Orleans terminilogy consistency, and the term grain activation is very
consistent so far. Same for grain factory. I like the
IGrainActivationCreator or even more plain simple IGrainInstanceCreator.

src/OrleansRuntime/Catalog/GrainCreator.cs
ReubenBond: GrainCreator is internal, though

jdom:

jdom:
I split the PR into 2 commits (to be merged without squashing), so that the Release part that disposes grain instances is separate, and we could in theory pull just the creation part into 1.4.2 without any behavioral change in comparison to 1.4.1.

I also renamed `IGrainActivationContext.GrainServices` to `IGrainActivationContext.ActivationServices` to make it less ambiguous with the GrainServices feature.

I did not rename `IGrainActivator`, since we spent a long time with Reuben and couldn't come up with a better name. We considered `IGrainActivationActivator` (this one would have full parity with MVC), `IGrainActivationCreator`, `IGrainResolutionBehavior`. If you have a better suggestion, please include it, otherwise we can rename later (which is unfortunate, but practical).

src/OrleansRuntime/Catalog/GrainCreator.cs
jdom: oh yeah, I thought this comment was on DefaultGrainActivator, my bad. Fixed

jason-bragg:
@jdom prototyping addressed most of my lingering concerns.

ReubenBond:
@dotnet-bot test this please

ReubenBond:
image
Not sure why CI is failing here

jdom:
@dotnet-bot test netfx-bvt
The java stacktrace is because Jenkins was restarted

jdom:
@dotnet-bot test this please
Reason: unrelated failures. Now CI should be ready to run without non-deterministic failures

Issue #3000, opened 2017-05-08
Fail to unsubscribe from SMS Provider streams


richcorless:
Hi,

Within my application I use streams to broadcast changes to other grains to avoid having to keep a cache of grain IDs. This works very well in normal operation, however, I am currently looking to make the system as robust as possible so I want it to recover from an ungraceful shutdown (i.e. kill the processes). 

The basic set up is an Orleans server process and an Orleans client process that acts as a gateway which endpoints connect to via WCF to use the system. There is a grain per connected endpoint that is used to cache the services the endpoint has available and this needs to be cleaned up after an ungraceful shutdown if the endpoint disconnected while the Orleans server was down.

When the Orleans client connects once more after recovery, it notifies which endpoints are still connected, if any are no longer connected then the grains need to have a stream unsubscribed. During this process, the OnActivateAsync method executes for the endpoint grain where it resumes the stream subscription without issue, it then calls to cancel the grain (clear state, etc...) to properly unsubscribe from it and I get:

> Extension not installed on grain OrleansStreamTest.Grains.PublisherGrain attempting to invoke type Orleans.Streams.OrleansCodeGenStreamProducerExtensionMethodInvoker from invokable Orleans.Runtime.ActivationData
Exception = Orleans.Runtime.GrainExtensionNotInstalledException: Extension not installed on grain OrleansStreamTest.Grains.PublisherGrain attempting to invoke type Orleans.Streams.OrleansCodeGenStreamProducerExtensionMethodInvoker from invokable Orleans.Runtime.ActivationData

I have a test application that replicates the issue, code snippets below. Configuration wise I am using the Azure Storage Emulator for the GrainStore and the PubSubStore and Simple Message Streams. With this test application I have also found that it affects streams used only by grains, streams used by clients and/or grains unsubscribe with no issues.

**Publisher Grain**
```
public class PublisherGrain : Grain, IPublisherGrain
{
	private IAsyncStream<double> clientStream;

	private IAsyncStream<double> grainStream;

	public override Task OnActivateAsync()
	{
		var streamProvider = this.GetStreamProvider("SMSProvider");
		this.clientStream = streamProvider.GetStream<double>(Guid.Empty, "TestClientStream");
		this.grainStream = streamProvider.GetStream<double>(Guid.Empty, "TestGrainStream");
		return base.OnActivateAsync();
	}

	public async Task PublishAsync(double data)
	{
		Console.WriteLine($"Sending data: {data}");
[...]

PR #3003, opened 2017-05-09
make pubsub more robust when facing ungraceful shutdown of producers

xiazen:
fix for #3000 

The cause of this bug : producer was shut down ungracefully (silo process was killed) ->consumer grain wants to unsubscribe on the stream the producer was producing on ->   unsubscribe request went to pubsub system and pubsub system tries to activate procuder grain and try to notify its ProducerExtension of the removed subscriber -> producer grain wakes up without the ProducerExtension, because the extension only installed when producer calls onNextAsync, so the application gets a GrainExtensionNotInstalledException. 

Fix: ignore the exception in this case. Producer will lost its subscriber list when deactivate and will repopulate its consumer list when it tries to produce again after activating, so no need to clean it.

### DIFF
--- a/src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
+++ b/src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
@@ -216,7 +216,7 @@ namespace Orleans.Streams
                         continue;
                     }
 
-                    tasks.Add(NotifyProducer(producer, subscriptionId, streamId, streamConsumer, filter));
+                    tasks.Add(NotifyProducerOfNewSubscriber(producer, subscriptionId, streamId, streamConsumer, filter));
                 }
 
                 Exception exception = null;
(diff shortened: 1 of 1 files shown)


src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
```diff
@@ -436,16 +436,35 @@ public Task<List<StreamSubscription>> GetAllSubscriptions(StreamId streamId, ISt
...
+        {
+            try
+            {
+                await producer.Producer.RemoveSubscriber(subscriptionId, streamId);
+            }
+            catch (GrainExtensionNotInstalledException)
+            {
+                RemoveProducer(producer);
+            }
+            catch (ClientNotAvailableException)
```
xiazen: @jason-bragg not entirely familiar with this exception, I copied it from NotifyProducerOfNewSubscriber. But do you know in which case this exception will throw?

jason-bragg: This is the correct behavior in this case.
This will be thrown when an Orleans client is producing events on a stream and loses connection to the cluster.
This logic was added to the add consumer, but not to the remove consumer when it should apply to both.

jason-bragg:
LT2M
Will hold merge until tomorrow to give time for others to review.
Thanks @richcorless for reporting this and @xiazen for fixing it so fast.
jbragg

Issue #2571, opened 2017-01-05
OrleansDockerUtils -> Docker and Docker Swarm Clusters for Orleans


galvesribeiro:
This issue is for discussion about the design of `OrleansDockerUtils` so we can add native support to docker containers as Orleans Silos. The implementation is on PR #2569 

## Goal

The idea behind this effort is to enable an Orleans Silo to run in containers with minimal efforts by leveraging both Orleans' provider model and Docker APIs. This PR introduces a `DockerMembershipOracle` and a `DockerGatewayProvider` following the design of #2542 but instead of use SF services, use Docker APIs.

## Some context

Without going too deep, a brief explanation about common topics on docker ecosystem so the reviewers feels more comfortable with it.

### Docker Daemon
This is the Docker process running in the underlying OS (`dockerd` on Linux and OSX and `dockerd.exe` on Windows). This process is responsible to manage the containers and provide a way (using `libcontainer`) to talk to underlying OS' kernel in order to provide the container runtime environment. Each docker host (in production or even your development machine and regardless if single-node or clustered) has one instance of this process. This process also expose a API so 3rd party developers can integrate/automate container management. 

### Docker Swarm
Swarm is the most used clustering (and orchestration) technology for Docker and is created by Docker Inc. Swarm implement almost 100% of the Docker APIs which means, that Docker CLI (and API) commands issued to swarm has (almost) no difference to the ones issued to a single node Docker Daemon. While in Swarm mode, commands are issue to the Swarm Manager nodes and it apply the request actions across the cluster of Docker Daemons properly.

### Docker Compose
*YAML* file describing grouped/correlated services/containers and the dependencies between each other as long as networking and volume mapping settings. This file can be pushed to a single Docker Daemon or to a Swarm cluster and will deploy the whole set of containers, networks and whatever artifact described in it. It is the most used composition tech for Docker containers nowadays.

### Container
A container is basically an isolation share of the host kernel with its own processes. The PID 1 of each container is the command passed when container start or described in the `Dockerfile` (the container image description file). In our case it would be our Silo host application. If for whatever reason that process crash or return/exit, the container will also exit.

### Labels
Docker allow one to add metadata to container and/or images thru labels. They are essentially key/value pairs. It can be anything as long as it can be string represented (key or value) and `JSON.stringfy()`(only value). You can even apply a label without a value. 

## Design

[...]

PR #3004, opened 2017-05-09
Initial work on Orleans-Docker docs

galvesribeiro:
This PR add some guidance on how to deploy Orleans to Docker. 

Initially, it explain how to create a project, add Docker support and debugger capabilities. 

Swarm is described in this document. Kubernetes will be added later on.


src/Documentation/Advanced-Concepts/Docker-Deployment.md
```diff
@@ -0,0 +1,611 @@
...
+
+## Deploying Orleans solutions to Docker
+Deploying Orleans to Docker can be tricky given the way Docker orchestrators and clustering stacks was designed. The most complicated thing is to understand the concept of *Overlay Network* from Docker Swarm and Kubernets Networking model. 
+
+Docker containers and networking model were designed to run mostly stateless and immutable containers. So, spin up a cluster running node.js or nginx applications, is pretty easy. However, if you try to use something more elaborated, like a real clustered or distributed application (like Orleans-based ones) you will eventually have troubles setting it up. It is possible, but not as simples as web-based applications.
+
+Docker clustering consist of putting together multiple hosts to work as a single pull of resources managed using a *Container Orchestrator*. *Docker Inc.* provide **Swarm** as their option for Container Orchestration while *Google* has **Kubernetes** (aka **K8s**). There are other Orchestrators like **DC/OS**, **Mesos**, etc., but in this document we will talk about Swarm and K8s as they are more widely used.
```
stephenlautier: "single pull" - i think you mean single pool*

stephenlautier: "single pull" - i think you mean single pool*

galvesribeiro: Ops!

src/Documentation/Advanced-Concepts/Docker-Deployment.md
```diff
@@ -0,0 +1,611 @@
...
+
+# Docker Swarm
+
+Docker clustering stack is called **Swarm** and you can find more by reading its [documentation here](https://docs.docker.com/engine/swarm/).
+
+To run this article in a `Swarm` cluster, you don't have any extra work. When you run `docker-compose up -d` in a `Swarm` node, it will schedule containers based on the configured rules. The same applies to other Swarm-based services like [Docker Datacenter](https://www.docker.com/enterprise-edition), [Azure ACS](https://azure.microsoft.com/en-us/services/container-service/) (in Swarm mode), [AWS ECS Container Service](https://aws.amazon.com/ecs/) and so on. All you need to do is to deploy your `Swarm` cluster before deploy your **dockerized** Orleans application.
+
+# Google Kubernetes (K8s)
+
+Comming soon...
```
stephenlautier: typo Coming*

galvesribeiro: Done. Thanks

src/Documentation/Advanced-Concepts/Docker-Deployment.md
```diff
@@ -0,0 +1,615 @@
...
+```
+
+`docker-compose.yml` (production):
+```yml
+version: '3.1'
+
+services:
+  orleans-client:
+    image: orleans-client
+    depends_on:
```
stephenlautier: not entirely sure, but do you think this is really needed as it doesnt seem to be supported within swarm mode; or perhaps only for `docker-compose up`?

> The depends_on option is ignored when deploying a stack in swarm mode with a version 3 Compose file.
https://docs.docker.com/compose/compose-file/#dependson

galvesribeiro: Ha! That is new. Didn't checked that update. It used to work... Will check with the Docker folks what should be used now. Thanks for point out.

galvesribeiro: Ok, just checked... It is only for `stack`. You still be able to use it with compose. Nevertheless, you are not required to use `depends_on`. It was just to illustrate the dependency so I think I'll stick with it there. I probably can create another section later (like the Kubernetes one) that show how to leverage stack. 

To give an idea, most of the hosted Docker services like in Azure or AWS, still use the older versions of the engine, which mean, we still don't have `stack` available there.

But again, thanks for point that out.

src/Documentation/Advanced-Concepts/Docker-Deployment.md
```diff
@@ -0,0 +1,615 @@
...
+orleansdocker_orleans-silo_9     tail -f /dev/null   Up 
+```
+
+> **Note**: The `Command` column on those examples are showing the `tail` command just because we are using the debugger container. If we were in production, it would be showing `dotnet OrleansSilo.dll` for example.
+
+# Docker Swarm
+
+Docker clustering stack is called **Swarm** and you can find more by reading its [documentation here](https://docs.docker.com/engine/swarm/).
+
+To run this article in a `Swarm` cluster, you don't have any extra work. When you run `docker-compose up -d` in a `Swarm` node, it will schedule containers based on the configured rules. The same applies to other Swarm-based services like [Docker Datacenter](https://www.docker.com/enterprise-edition), [Azure ACS](https://azure.microsoft.com/en-us/services/container-service/) (in Swarm mode), [AWS ECS Container Service](https://aws.amazon.com/ecs/) and so on. All you need to do is to deploy your `Swarm` cluster before deploy your **dockerized** Orleans application.
```
stephenlautier: in order to deploy on swarms, it seems to be a better approach to use 
```
docker stack deploy -c docker-compose.yml <name>
```
when using v3 compose files, perhaps just to mention it - what do you think?

src/Documentation/Advanced-Concepts/Docker-Deployment.md
stephenlautier: welcome, and thanks for clarifying 👍

src/Documentation/Advanced-Concepts/Docker-Deployment.md
galvesribeiro: Yeah, like I mentioned in your previously comment. It must have `stack` support on the engine. I'll add a comment there for it now. Thanks

stephenlautier: sorry posted this before saw the reply

galvesribeiro: Np, thanks :)

src/Documentation/Advanced-Concepts/Docker-Deployment.md
```diff
@@ -0,0 +1,617 @@
...
+The way we found to workaround this *limitation*, is to manually add a gateway yourself to the client configuration like this:
+
+```csharp
+var hostEntry = await Dns.GetHostEntryAsync("orleans-silo");
+var ip = hostEntry.AddressList[0];
+config.Gateways.Add(new IPEndPoint(ip, 10400));
+```
+
+The string `orleans-silo` here, is essentially the **service name** used in `docker-compose`, which we will dig into more details later on (keep reading!). It will resolve to one of the containers' ips inside that same service (like a Load Balancer). In a docker-compose project, a service is a group of multiple (exactly equal) instances of a give container. For example, we will have a service for our `OrleansSilo` project. Within that service, we will be able to scale the service up and down horizontally and adding each of those containers to the Orleans cluster automatically respecting Orleans Membership Protocol. More on that later.
+
```
gabikliot: I was looking, but did not fin anything  for "later".
I am puzzled how swarm clustering works along side Orleans's Membership. 
If Orleans still creates a membership table in Azure, why GW is configured with IP and not pointed to the table? 

Can you please explain?

galvesribeiro: @gabikliot because you are inside a Compose/Swarm project, the sane way to refer to another service, is by using the service name. Also, if the client is outside the same compose project of the silo, the IPs from the GW are not registered on Azure Table. We discussed before changes on the way membership record IPs on the tables, and everyone was against changes in MBR so the only safe way is to get the service IP, which works as a NLB and span across the cluster. It is a docker internal IP, not public, and works for the purpose of finding any of the GWs.

gabikliot: Let me see if I understand.
So if MBR registered the gw ip in the table, that ip is not routable from outside, so it is useless.
Instead, from outside, one has to go via load balancer end point ( which is what nlb stands for).
Correct?

So u just need to associate that end point with all gateway silos, make sure they are part of this load balanced group, and then it is a static end point for the clients to use. The Client will send to the end point, and the load balancer will route to one of the gateways. There is no way in this architecture to let client pick a gateway, be sticky to a particular gw for particular grains, etc...

Did I understand it correctly?

galvesribeiro: Yes. The point is that you don't need to register the GW IP into the load balancer. The container's IP will be always behind the service IP, which is the load balancer.

So yes, you are correct.

MBR for this case, is only useful for the cluster, not the client.

I'm still investigating Kubernetes, which is way more complicated. Will see how its pod network works.

gabikliot:
Makes sense.

I still think one Could register internal gw ip in the table, as long as it is routable from outside. But there is no need to, as you said, since we can use the service end point.

galvesribeiro:
Yeah, I had a big debate with others from core team on whether we should register an internal IP + router IP in MBR or not. This very same case happens when you have firewalls/routers between OrleansClient and the cluster and this firewall/router only expose 1 single IP or a different IP address for each silo. Like you already know, we tried make a MBR provider for Docker but we fail because among other things, the way we register those IPs is different from what the majority of scenarios expect to be.

In the end, I decided stop fighting and with some updates on Docker I ended up with this solution which work perfectly fine. Even in case of gateway container failure, the service endpoint is notified and the next retry will in fact redirect to another container.

gabikliot:
I was not aware of this discussion.
If it was on GH, I would have had a chance to chime in.

> we tried make a MBR provider for Docker but we fail

I do remember you tried. I remember that you tried to make it similar to Service Fabric, but what was not clear to me to then is if Swarm indeed provides the membership service (the list of active containers) in a way that we could feed in Orleans.

Conceptually, I don't see a problem with exposing the actual IP and not the service end point, IF they are indeed routable and secure. Orleans does its own load balancing between gateways, so it does  not strickly need the load balanced end point.

I think its similar in spirit to the same high level discussion of why not use storage always and write stateless services and never do in memory state (and have to route to this state, etc)... Load balanced end point is great, but not a panacea. One should not force everyone always to use it. If we have our won solution that works better for us, why not.

galvesribeiro:
@gabikliot https://github.com/dotnet/orleans/issues/2571

And yes, I agree that would be awesome if we have our own membership. Maybe we can have it Kubernetes, I don't know, still need to check it. 

For now, this document and the sample will unblock a lot of people trying to use Orleans and Docker with Swarm. They are eagerly asking for it and since I was the one who got it to work and is actively working with Docker daily, I decided to write it and shed some light for people. We can always revisit it later if/when we have a membership solution for it.

gabikliot:
+1 on unblocking first.

sergeybykov:
Thank you very very much, @galvesribeiro! This is just the beginning.

galvesribeiro:
Yup! Will come up with more later. Just want to make sure people got unblocked and I stop to repeat same thing on Gitter 😄

src/Documentation/Advanced-Concepts/Docker-Deployment.md
```diff
@@ -0,0 +1,617 @@
...
+            running = false;
+        }
+
+        static async Task InitializeOrleans()
+        {
+            var config = new ClientConfiguration();
+            config.DeploymentId = "Orleans-Docker";
+            config.PropagateActivityId = true;
+            var hostEntry = await Dns.GetHostEntryAsync("orleans-silo");
```
beckjin: When I use service name "orleans-silo", have a error.

Unhandled Exception: System.AggregateException: One or more errors occurred. (No such device or address)

It's my docker-compose file:
```
version: '3'

services:
  orleans-silo:
    image: hub.xxx.com/beck/orleans/silo-test:0.0.1
    volumes:
      - /usr/local/beck/docker/orleans/test/siloservice/config:/app/Config
~                                                                             
```

RainingNight:
Any 2.x docker sample?

PR #3014, opened 2017-05-11
Add statistics to eh stream provider ecosystem

xiazen:
Main approach is described in this PR. But still need to write tests. get this PR out now for early feedback


src/Orleans/Properties/AssemblyInfo.cs
```diff
@@ -26,6 +26,7 @@
 [assembly: InternalsVisibleTo("OrleansProviders")]
 [assembly: InternalsVisibleTo("OrleansHost")]
 [assembly: InternalsVisibleTo("OrleansAzureUtils")]
+[assembly: InternalsVisibleTo("OrleansServiceBus")]
```
jason-bragg: What is internals access needed for?

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    public interface ICacheMonitor
+    {
+        #region Event driven metrics
+
+        /// <summary>
+        /// Track message added to the cache, and newest and oldest messsges among them 
+        /// </summary>
+        /// <param name="messageAdded"></param>
+        void TrackMessageAdded(long messageAdded);
```
jason-bragg: TrackMessagesAdded?  These track message counts added and removed correct?

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Monitor track block pool related metrics
+    /// </summary>
+    public interface IObjectPoolMonitor
+    {
+        #region Event driven metrics
+        /// <summary>
+        /// Track every time cache allocate memory from block pool
+        /// </summary>
+        /// <param name="memoryInByte"></param>
+        void TrackMemroyAllocatedByCache(int memoryInByte);
```
jason-bragg: Object pools don't know who is allocating, so the 'ByCache' is not useful.

jason-bragg: Also I think we should report this by object count, not memory size.  Object pools rarely know the size of what their pooling.
If this is specifically for a block pool, we can include memory size, but if this is for object pools in general, size probably won't work.

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
```diff
@@ -41,6 +47,11 @@ public abstract class PooledResource<T> : IDisposable
         public IObjectPool<T> Pool { set { pool = value; } }
 
         /// <summary>
+        /// Pooled resource size
+        /// </summary>
+        public virtual int SizeInByte { get {return Marshal.SizeOf(this);} }
```
jason-bragg: shouldn't be calculating these sizes per call at runtime.  As from other post, I think we should avoid size in the general case.

src/Orleans/Properties/AssemblyInfo.cs
xiazen: for SaferTimer

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
xiazen: messaged removed were tracked in TrackMessagePurged

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
xiazen: this is the default implementation, which is here just for not returning null. PooledResource's child class which we are actually using in cache system all override this method to a much cheaper method,

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
xiazen: hmm, I'm not sure how much value is object count by itself. Object count is a Orleans concept, which users may not understand that much. While memory is a universal concept most users understand and care about

src/OrleansServiceBus/Providers/Streams/EventHub/StatisticMonitors/DefaultEventHubReceiverMonitor.cs
```diff
@@ -0,0 +1,87 @@
...
+            logger.TrackMetric("ReadException", exception == null ? 0 : 1, this.logProperties);
+        }
+
+        /// <summary>
+        /// Tracks messages read and time taken per successful read.  Tracked per successful partition read operation.
+        /// </summary>
+        /// <param name="count">Messages read.</param>
+        /// <param name="oldest"></param>
+        /// <param name="newest"></param>
+        public void TrackMessagesRecieved(long count, TimeSpan? oldest, TimeSpan? newest)
```
DixonDs: Typo here and below in the metric name: "Recieved" -> "Received"

xiazen: Thanks !

src/Orleans/Properties/AssemblyInfo.cs
jason-bragg: We went through great efforts to ensure that the EHSP used public capabilities and was a working example of how external users could develop stream providers for other queues.  Basically if we can't develop a persistent recoverable without the need of internals than their is something wrong with our public surface.

xiazen: I totally agree. To avoid this reference, I think there's two solution:
1. short term: maintain a separate timer in EHSP
2. long term: package the utility class which is highly shareable, such as safe timer, to a separate nugget, so we won't wire unnecessary reference

xiazen: 3.use a third party timer. Is there a timer class in .net which can be configured to fire events periodically ? it's totally possible that it is there but I just don't know.

xiazen: or I can just use .net Timer class.... 
What's the history of SafeTimer? Its description makes me think it is better than Timer....

xiazen:
still need to add cache pressure monitor related metrics and tests . work in progress

xiazen:
ready for review

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
jason-bragg: TrackMessageAdded sounds like it's called when adding a single message, but this is called when we add multiple, right?  Suggest TrackMessagesAdded.

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Track message purged from the cache, and the newest and oldest messages among them
+        /// </summary>
+        /// <param name="messagePurged"></param>
+        void TrackMessagePurged(long messagePurged);
```
jason-bragg: Same as TrackMessageAdded, think we should the plural, TrackMessagesPurged

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
```diff
@@ -0,0 +1,67 @@
...
+        #region periodically report discrete metrics
+
+        /// <summary>
+        /// Periodically report cache status metrics
+        /// </summary>
+        /// <param name="oldestMessageAge">Relative age between oldest message and newest message in cache</param>
+        /// <param name="oldestMessageEnqueueTimeToNow">Difference between host utcNow and oldest message's enqueue time in cache </param>
+        /// <param name="newestMessageEnqueueTimeToNow">Difference between host utcNow and newest message enqueue time in cahce</param>
+        /// <param name="totalMessageCount"></param>
+        void ReportMessageStatistics(TimeSpan? oldestMessageAge, TimeSpan? oldestMessageEnqueueTimeToNow, TimeSpan? newestMessageEnqueueTimeToNow,
```
jason-bragg: The message age makes sense as a timespan, but oldest and newest enqueue times should probably be datetimes in Utc.  Using a delta between utc now will always be sketchy due to clock skew, and if the application needs this delta, it can compute it itself using the provided datetimes.

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Monitor track block pool related metrics
+    /// </summary>
+    public interface IObjectPoolMonitor
+    {
+        #region Event driven metrics
+        /// <summary>
+        /// Track every time cache allocate memory from block pool
+        /// </summary>
+        /// <param name="blockCount"></param>
+        void TrackObjectAllocatedByCache(int blockCount);
```
jason-bragg: If this is called every allocation what is the blockCount?

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        #region Periodically report block pool status 
+
+        /// <summary>
+        /// Periodically report block pool status
+        /// </summary>
+        /// <param name="totalBlocks">Total size of block pool</param>
+        /// <param name="freeBlocks">Free memory in block pool which is available for cache to allocate</param>
+        /// <param name="claimedBlocks">Memory claimed by cache in this block pool</param>
+        void Report(long totalBlocks, long freeBlocks, long claimedBlocks);
```
jason-bragg: In general, we should avoid the term 'blocks' in this case, because at this level we're dealing with objects.  Because objects are blocks in the streaming system doesn't imply that they will always be blocks in all object pools.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -102,10 +103,20 @@ public class EventHubAdapterFactory : IQueueAdapterFactory, IQueueAdapter, IQueu
...
-        ///   Factory funciton should return an IEventHubReceiverMonitor.
+        /// Factory funciton should return an IEventHubReceiverMonitor.
         /// </summary>
-        protected Func<string, string, Logger, IEventHubReceiverMonitor> ReceiverMonitorFactory { get; set; }
+        protected Func<EventHubReceiverMonitorDimentions, Logger, IEventHubReceiverMonitor> ReceiverMonitorFactory { get; set; }
+        /// <summary>
+        /// Create a cache monitor to report performance metrics.
+        /// Factory funciton should return an ICacheMonitor.
+        /// </summary>
+        protected Func<EventHubCacheMonitorDimentions, Logger, ICacheMonitor> CacheMonitorFactory;
```
jason-bragg: We've delegated cache creation responsibility to the IEventHubQueueCacheFactory, which is also responsible for the buffer pool.  Given that, I think customization of the monitoring for these systems should also be delegated to the IEventHubQueueCacheFactory.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -174,15 +178,15 @@ public Task Initialize(TimeSpan timeout)
             TimeSpan oldest = dequeueTimeUtc - messages[0].EnqueuedTimeUtc;
             TimeSpan newest = dequeueTimeUtc - messages[messages.Count - 1].EnqueuedTimeUtc;
 #endif
-            monitor.TrackAgeOfMessagesRead(oldest, newest);
+            monitor?.TrackMessagesReceived(messages.Count, oldest, newest);
```
jason-bragg: Since we're changing this, let's report cleaner information.   We originally reported the age, but that's suspect and the monitor can always calculate that if needed from the enqueue times.  Let's report the enqueue times rather than the relative ages.

src/OrleansServiceBus/Providers/Streams/EventHub/IEventHubQueueCache.cs
```diff
@@ -47,5 +47,10 @@ public interface IEventHubQueueCache : IQueueFlowController, IDisposable
         /// Send purge signal to the cache, the cache will perform a time based purge on its cached messages
         /// </summary>
         void SignalPurge();
+
+        /// <summary>
+        /// Cache monitor for current cache
+        /// </summary>
+        ICacheMonitor CacheMonitor { get; }
```
jason-bragg: Who uses this property?  The monitoring seems like an implementation detail, why should any outside system know or care?

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeObjectPool.cs
```diff
@@ -33,8 +33,11 @@ public class FixedSizeObjectPool<T> : ObjectPool<T>
         /// </summary>
         /// <param name="poolSize"></param>
         /// <param name="factoryFunc"></param>
-        public FixedSizeObjectPool(int poolSize, Func<T> factoryFunc)
-            : base(factoryFunc, poolSize)
+        /// <param name="blockPoolId">Unique Id for current block pool</param>
+        /// <param name="monitor"></param>
+        /// <param name="monitorWriteInterval"></param>
+        public FixedSizeObjectPool(Func<T> factoryFunc, string blockPoolId, int poolSize, IObjectPoolMonitor monitor = null, TimeSpan? monitorWriteInterval = null)
```
jason-bragg: If at all possible, I think we should avoid block pools having specific ids.  What is the blockPoolId for?

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
```diff
@@ -23,6 +24,11 @@ public interface IObjectPool<T>
         /// </summary>
         /// <param name="resource"></param>
         void Free(T resource);
+
+        /// <summary>
+        /// Unique id of current blockId
+        /// </summary>
+        string Id { get; }
```
jason-bragg: What is this for?

src/OrleansProviders/Streams/Common/PooledCache/ObjectPool.cs
```diff
@@ -14,12 +16,24 @@ public class ObjectPool<T> : IObjectPool<T>
         private const int DefaultPoolCapacity = 1 << 10; // 1k
         private readonly Stack<T> pool;
         private readonly Func<T> factoryFunc;
+        private long totalBlocks;
```
jason-bragg: as with monitor, I think we should avoid the term 'blocks' at this layer.

src/OrleansProviders/Streams/Common/PooledCache/ObjectPool.cs
```diff
@@ -52,7 +82,15 @@ public virtual T Allocate()
...
         {
+            this.monitor?.TrackObjectReleasedFromCache(1);
             pool.Push(resource);
         }
+
+        private void ReportObjectPoolStatistics(object state)
+        {
+            var freeBlocks = this.pool.Count;
+            long claimedBlocks = this.totalBlocks - freeBlocks;
+            this.monitor.Report(this.totalBlocks, freeBlocks, claimedBlocks);
```
jason-bragg: I think availableBlocks may be more clear than free blocks.  Free usually means returned to the OS.

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
```diff
@@ -97,8 +93,16 @@ public PooledQueueCache(ICacheDataAdapter<TQueueMessage, TCachedMessage> cacheDa
...
             this.comparer = comparer;
             this.logger = logger.GetSubLogger("messagecache", "-");
+            this.itemCount = 0;
             pool = new CachedMessagePool<TQueueMessage, TCachedMessage>(cacheDataAdapter);
             messageBlocks = new LinkedList<CachedMessageBlock<TCachedMessage>>();
+            this.cacheMonitor = cacheMonitor;
+
+            if (this.cacheMonitor != null && cacheMonitorWriteInterval.HasValue)
+            {
+                var timer = new Timer(this.ReportCacheMessageStatistics, null, cacheMonitorWriteInterval.Value, cacheMonitorWriteInterval.Value);
```
jason-bragg: not keeping timer around?  Won't this turn it off when object is garbage collected?  Also, no means of stopping it should the agent responsible for this be shutdown.  May want to consider making the cache disposable and disposing it when receiver is shutdown..  though disposables are annoying :/

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
```diff
@@ -120,6 +124,24 @@ public object GetCursor(IStreamIdentity streamIdentity, StreamSequenceToken sequ
             return cursor;
         }
 
+        private void ReportCacheMessageStatistics(object state)
+        {
+            if (this.IsEmpty)
+            {
+                this.cacheMonitor.ReportMessageStatistics(null, null, null, this.ItemCount);
```
jason-bragg: may as well hardcode 0 for item count here.

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
jason-bragg: all 3 of these values can be calculated in the monitor if enqueue datetimes are reported, but we don't have any way of reporting how long the oldest message has been in the cache, which I think is an important piece of information.  I had initially thought that was what oldestMessageAge was..

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Responsible for monitoring cache related metrics
+    /// </summary>
+    public interface ICacheMonitor
+    {
+        #region Event driven metrics
+
+        /// <summary>
+        /// Track cache pressure metrics when cache pressure monitor encounter a status change
+        /// </summary>
+        void TrackCachePressureMonitorStatusChange(string pressureMonitorType, bool underPressure, double? cachePressureContributionCount, double? currentPressure);
```
jason-bragg: Threshold may be useful here, I know it doesn't change, but mapping pressureMonitorType to threshold for reporting purposes may be tedious for users, should they want to know what the threshold was that was passed.  The threshold can be important of the severity of the underlying monitoring, a little over a threshold may be expected, but a lot might be a reason to throw alerts.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
```diff
@@ -174,15 +178,15 @@ public Task Initialize(TimeSpan timeout)
...
 
             foreach (EventData message in messages)
             {
                 StreamPosition streamPosition = cache.Add(message, dequeueTimeUtc);
                 batches.Add(new StreamActivityNotificationBatch(streamPosition.StreamIdentity.Guid,
                     streamPosition.StreamIdentity.Namespace, streamPosition.SequenceToken));
             }
-
+            cache.CacheMonitor?.TrackMessageAdded(messages.Count);
```
jason-bragg: Suggest adding an Add(List<EventData> messages) to the IEventHubQueueCache, rather than exposing the monitor.

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// Track message purged from the cache, and the newest and oldest messages among them
+        /// </summary>
+        /// <param name="messagePurged"></param>
+        void TrackMessagePurged(long messagePurged);
+
+        /// <summary>
+        /// Track new memory allocated by the cache
+        /// </summary>
+        /// <param name="memoryInByte"></param>
+        void TrackMemoryAllocated(int memoryInByte);
```
jason-bragg: Not necessarily opposed to tracking the memory in the cache, it's just not clear how valuable this is and the metric can be easily misunderstood.

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
```diff
@@ -63,15 +65,7 @@ public class PooledQueueCache<TQueueMessage, TCachedMessage>: IPurgeObservable<T
...
-        public int ItemCount { get
-            {
-                int count = 0;
-                foreach (var block in this.messageBlocks)
-                {
-                    count += block.ItemCount;
-                }
-                return count;
-            }
+        public int ItemCount { get { return this.itemCount; }
```
jason-bragg: Do we need a public property for this?

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
xiazen: will do

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
xiazen: will do

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
xiazen: I originally think enqeueuTime as Datetime make more sense, until I saw that in original EventHubReceiverMonitor it was reporting enqueueTimeToNow instead of enqueueTime. Do you remember why is it? Maybe customer ask for it when EventHubReceiverMonitor is written? 

To calculate how long the message is in the cache, we need to report dequeueTime as well. So the interface will be like 
void ReportMessageStatatistics(TimeSpan? oldestMessageAge, DateTime? newestMessageEnqueueTime, DateTime? newestMessageDequeueTime, DateTime? oldestMessageEnqueueTime, DateTime? oldestMessageDequeueTime )

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="newestMessageEnqueueTimeToNow">Difference between host utcNow and newest message enqueue time in cahce</param>
+        /// <param name="totalMessageCount"></param>
+        void ReportMessageStatistics(TimeSpan? oldestMessageAge, TimeSpan? oldestMessageEnqueueTimeToNow, TimeSpan? newestMessageEnqueueTimeToNow, 
+            long totalMessageCount);
+
+        /// <summary>
+        /// report total cache size
+        /// </summary>
+        /// <param name="totalCacheSizeInByte"></param>
+        void ReportCacheSize(long totalCacheSizeInByte);
```
xiazen: it is because of eventhub cache infra's implementation detail. I cannot report both of those in the same place. And they are conceptually separate as well.

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
xiazen: it is 1.  I don't want to make it void TrackObjectAllocatedByCache(); that's exposing our internal implementation detail: every time object pool allocate, we only allocate one. And less extensible : what if we changed the implementation, and now the object pool can allocate multiple at the same time?

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
xiazen: I will change block to object when it relates to ObjectPool

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
xiazen: make sense.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
xiazen: will do

src/OrleansServiceBus/Providers/Streams/EventHub/IEventHubQueueCache.cs
xiazen: I will follow your suggestion below to fix this

src/OrleansProviders/Streams/Common/PooledCache/FixedSizeObjectPool.cs
xiazen: will change block to object, when it is related to object pool.

I  used block mainly because users are more familiar with that term.

src/OrleansProviders/Streams/Common/PooledCache/IObjectPool.cs
xiazen: it is one dimension for object pool metrics. but I guess it can just stay in ObjectPoolMonitorDimemsions.

src/OrleansProviders/Streams/Common/PooledCache/ObjectPool.cs
xiazen: will do availableObjects

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
xiazen: ?? I'm confused by your text. I understand and agree that I should dispose it when receiver is shutting down. What does the text before that mean?  I think it is good if the timer is GCed when the obejct is GCed, no?

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
xiazen: nah, hardcode is not good. what if there's a bug in this.isEmpty , it returns true even when it is not empty?

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
xiazen: but threshold is not a metrics. and the user knows the threshold, they set it. 
It is the same as message age, we are not reporting DataMinAgeInCache as well.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterReceiver.cs
xiazen: will do

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
xiazen: why the metric can be easily misunderstood? 
Maybe TrackMemoryNewlyAllocated?

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
jason-bragg: >Do you remember why is it?

Original logic was a bit too closely aligned with pre-existing logging in 343, where this provider was ported from.  We need provide the information they need to maintain their reporting, but within those constraints we can/should improve this.

I think all they'd need is:
void ReportMessageStatatistics(TimeSpan? oldestMessageAge, DateTime? newestMessageEnqueueTimeUtc, DateTime? DateTime? oldestMessageEnqueueTimeUtc)
Where oldestMessageAge is the delta between the now and the time the oldest message in the cache was dequeued, and newestMessageEnqueueTimeUtc/oldestMessageEnqueueTimeUtc are the eventhub enqueue times of the oldest and newest message in the cache.

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
jason-bragg: I understand, and agree to a degree that they are conceptually separate, but totalCacheSizeInByte is more meaningful if aligned with message count.

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
xiazen: I will look it up when I got to the office. I guess something is using it since it is public before.

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
jason-bragg: This monitor is related to IObjectPool, not the cache.  IObjectPool only allocates one object at a time, so a count here doesn't really make sense.  I don't think it's a question of exposing implementation details as much as providing monitoring appropriate for the class being monitored.

Along the same line of reasoning, TrackObjectAllocated is more appropriate.  We should not be making assumptions about who is requesting the allocation.

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
jason-bragg: This entire monitoring interface is related to IObjectPool, correct?

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
jason-bragg: If the timer was a member of the class like this.timer  - new Timer(..), it would be kept around until the cache went away, but we don't store a reference of it anywhere, so it will be cleaned up by garbage collection, right?  Once that happens the timer will be turned off.  Unless I'm misunderstanding how/when objects are garbage collected..

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
jason-bragg: Fair enough.  I also made that comment before I reviewed the ItemCount logic.  I thought it actually tried to do work (like it used to).

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
jason-bragg: I understand.  Threshold is a configuration, not a metric, however it can be important to mentoring, as I monitoring, it can influence severity, and the only means a service developer would have of getting at that info, should we not pass it in, would be to map the pressureMonitorType to the threshold.

These monitoring hooks expose metrics, but they will likely be used for both metric reporting and critical logging, especially for alerting purposes.

As for the user knowing these values, a developer intimate with the system may well, but a maintainer looking at logs will likely not.  Pls keep in mind some services use many different stream providers, possibly with many different thresholds, so the likelihood of an engineer knowing these values may be low.

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
jason-bragg: > why the metric can be easily misunderstood.

This is the memory allocated to the cache, not necessarily the memory use, nor does it necessarily accurately account for the size of the messages used by the cache.

For instance if the cache reports having 1k messages and using 1mb of memory, a maintainer may reseasonlby conclude that the average message size is 1k, but the 1mb includes the message payload, and other evenhub data we track plus space lost by fragmentation, so the actual message size may be half that.  This is not a big deal really, I was just pointing out that we've a metric with limited utility (IMO) that is also somewhat ambiguous.  I'd not include it for these reasons, but I won't protest if you feel differently.

>TrackMemoryNewlyAllocated

I think current name is fine.

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
xiazen: agree with the latter point. but I think it still make sense of keep the blockCount around for the sake of extensibility.

src/OrleansProviders/Streams/Common/Monitors/IObjectPoolMonitor.cs
xiazen: I meant, I will keep it as blocks if it appears in cache monitors. or other steaming specific monitors. There maybe not one .

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
xiazen: hmm, I think one engineer will understand the cache size doesn't equal to messagecount*messagesize. Since it lies in the concept of cache. When a system allocate a cache, it allocate a size of memory based on the estimated load, instead of actual load. AFAIK, none of the cache providers provide a solution which allocate memory to exact amount of the actual load

src/OrleansProviders/Streams/Common/PooledCache/PooledQueueCache.cs
xiazen: it is used in EvictionStrategy.

src/OrleansProviders/Streams/Common/Monitors/ICacheMonitor.cs
xiazen: But there's already a concept of message age = message.enqueueTime - newestMessage.enqueueTime in time based purge logic , which is configured through DataMAxAgeInCache. We don't want to introduce another version of message age. 

So I think the interface will look like:
void ReportMessageStatatistics(TimeSpan? oldestMessageAge, DateTime? newestMessageEnqueueTime, DateTime? oldestMessageEnqueueTime, DateTime? oldestMessageDequeueTime )

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -297,8 +281,8 @@ protected virtual IEventHubQueueCacheFactory CreateCacheFactory(EventHubStreamPr
             var globalConfig = this.serviceProvider.GetRequiredService<GlobalConfiguration>();
             var nodeConfig = this.serviceProvider.GetRequiredService<NodeConfiguration>();
             var eventHubPath = hubSettings.Path;
-            var sharedDimentions = new EventHubMonitorAggregationDimentions(globalConfig, nodeConfig, eventHubPath);
```
xiazen: there's many fix typo diff , mostly dimention -> dimension

xiazen:
@dotnet-bot test netfx-bvt netfx-functional

xiazen:

PR #3028, opened 2017-05-19
Update unit test tutorial to use TestCluster

wattsm:
Hello,

I noticed when working through the [unit testing tutorial](https://dotnet.github.io/orleans/Tutorials/Unit-Testing-Grains.html) that it uses `TestingSiloHost` which is [marked as obsolete](https://github.com/dotnet/orleans/pull/2919).

I have rewritten this tutorial to use `TestCluster`, and also added a bit of detail about mocking. I have tried to incorporate as much of the relevant content from the previous version as possible.

Hope this helps. Feed / suggestions / requests more than welcome

### DIFF
--- a/src/Tutorials/Unit-Testing-Grains.md
+++ b/src/Tutorials/Unit-Testing-Grains.md
@@ -7,58 +7,7 @@ title: Unit Testing Grains
 
 This tutorial shows how to unit test your grains to make sure they behave correctly.
 There are two main ways to unit test your grains, and the method you choose will depend on the type of functionality you are testing.
-You can use a mocking framework like [Moq](https://github.com/moq/moq) to mock parts of the Orleans runtime that your grain interacts with, or you can use the  `Microsoft.Orleans.TestingHost` NuGet package to create test silos for your grains.
-
-## Using Mocks
-
-Orleans makes it possible to mock many parts of system, and for a lot of scenarios this is the easiest way to unit test grains. 
-This approach does have limitations, and may require that grains include code used only by your unit tests.
-
-For example, let us imagine that the grain we are testing interacts with other grains.
-In order to be able to mock those other grains we also need to mock the `GrainFactory` member of the grain under test.
-By default `GrainFactory` is a normal `protected` property, but most mocking frameworks require properties to be `public` and `virtual` to be able to mock them.
-So the first thing we need to do is make `GrainFactory` both `public` and `virtual` property:
-
-```csharp
-public new virtual IGrainFactory GrainFactory 
-{
-    get { return base.GrainFactory; }
-}
-```
-
-Now we can create our grain outside of the Orleans runtime and use mocking to control the behaviour of `GrainFactory`:
-
-```csharp
-using System;
-using System.Threading.Tasks;
-using Orleans;
-using Xunit;
-using Moq;
-
-namespace Tests 
-{
-    public class WorkerGrainTests 
-    {
-        [Fact]
-        public async Task RecordsMessageInJournal() 
-        {
-            var data = "Hello, World";
-
-            var journal = new Mock<IJournalGrain>();            
-
-            var worker = new Mock<WorkerGrain>();
-            myGrain.Setup(x => x.GrainFactory.GetGrain<IJournalGrain>(It.IsAny<Guid>())).Returns(journal.Object);
-
-            await worker.DoWork(data)
-
-            journal.Verfiy(x => x.Record(data), Times.Once());
-        }
-    }
-}
-```
-
-Here we create our grain under test, `WorkerGrain`, using Moq which means we can then override the behaviour of the `GrainFactory` so that it returns a mocked `IJournalGrain`. 
-We can then verify that our `WorkerGrain` interacts with the `IJournalGrain` as we expect.
+The `Microsoft.Orleans.TestingHost` NuGet package can be used to create test silos for your grains, or you can use a mocking framework like [Moq](https://github.com/moq/moq) to mock parts of the Orleans runtime that your grain interacts with.
 
 ### Using TestCluster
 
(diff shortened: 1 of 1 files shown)


src/Tutorials/Unit-Testing-Grains.md
```diff
@@ -5,81 +5,159 @@ title: Unit Testing Grains
...
 # Unit Testing Grains
 
-This tutorial shows you how to write unit tests for your grains to make sure they are behaving correctly.
-For a distributed application written in Orleans, you'll need load tests and integration tests as well but here we'll only focus on unit tests.
+This tutorial shows how to unit test your grains to make sure they behave correctly.
+There are two main ways to unit test your grains, and the method you choose will depend on the type of functionality you are testing.
+You can use a mocking framework like [Moq](https://github.com/moq/moq) to mock parts of the Orleans runtime that your grain interacts with, or you can use the  `Microsoft.Orleans.TestingHost` NuGet package to create test silos for your grains.
 
-Orleans makes it possible to mock many of its parts ([example](https://github.com/dotnet/orleans/tree/master/Samples/UnitTesting.Minimal)) but here we only focus on simply running grains in test silos.
+## Using Mocks
```
jdom: Even though it's nice this is doable, I would prioritize TestCluster higher (and hence put it first in the docs).
Also with this approach different people will want to use different mocking frameworks (I love Moq btw). I hope that at some point we put more effort on having a better experience without requiring to mock out complex parts of Orleans yourself (but in the meantime, you might want to link to https://github.com/OrleansContrib/OrleansTestKit).

src/Tutorials/Unit-Testing-Grains.md
```diff
@@ -5,81 +5,159 @@ title: Unit Testing Grains
...
 
-The steps are
+Orleans makes it possible to mock many parts of system, and for a lot of scenarios this is the easiest way to unit test grains. 
+This approach does have limitations, and may require that grains include code used only by your unit tests.
```
jdom: This is not the only limitation. It has several, especially with regards to scheduling, reentrancy, serialization, which will behave very differently in the silo than directly on the framework. I would highlight this.

src/Tutorials/Unit-Testing-Grains.md
```diff
@@ -5,81 +5,159 @@ title: Unit Testing Grains
...
-            Assert.AreEqual(expected, reply);
+            Assert.Equal("Hello, World", greeting);
         }
     }
-}   
+}
+```
 
+Due to the overhead of starting an in-memory cluster you may wish to create a `TestCluster` and reuse it among multiple test cases.
+For example this can be done using Xunit's class or collection fixtures (see [https://xunit.github.io/docs/shared-context.html](https://xunit.github.io/docs/shared-context.html) for more details).
```
jdom: even thought namespace is `Xunit` due to .NET design guidelines, when talking about the test framework you should probably refer to it as `xUnit`.

src/Tutorials/Unit-Testing-Grains.md
```diff
@@ -5,81 +5,159 @@ title: Unit Testing Grains
...
+    }
+}
+```
 
-- Keep in mind that the silo will not be restarted after each test method so if your methods need a fully clean grain then make sure that they'll use different grain IDs.
-- The `TestingSiloHost` class has multiple utility methods for starting and stopping silos which can be used for programmatic simulation of some failure situations.
-- Starting and stopping the silo for each test case takes time and is not a good idea unless you really need it for a test.
-- The two created silos will be named "Primary" and "Secondary". It is important to note that the notion of a Primary silo is only used in Testing. We use the Primary silo to host the system store in a special `MembershipTableGrain` system grain, to allow bootstrap. In production there is no primary and secondary silos: all silos are equal and behave in a peer-to-peer fashion.
-- Configurations of the environment are based on `TestSiloOptions` and `TestClientOptions` objects which can be passed to the constructor of the base class.
+Xunit will call the `Dispose` method of the `ClusterFixture` type when all tests have been completed and the in-memory cluster silos will be stopped.
```
jdom: this is the other only occurrence of `Xunit` that should be `xUnit`

src/Tutorials/Unit-Testing-Grains.md
```diff
@@ -5,81 +5,159 @@ title: Unit Testing Grains
...
+        }
 
-## Remarks
+        [Fact]
+        public async Task SaysHelloCorrectly()
+        {            
+            var hello = _cluster.GrainFactory.GetGrain<IHelloGrain>(Guid.NewGuid());
+            var greeting = await hello.SayHell();
+
+            cluster.StopAllSilos();
```
wattsm: Well spotted, whoops.

src/Tutorials/Unit-Testing-Grains.md
wattsm: Gets me every time 👍

jdom:
Awesome, thanks a lot!

shlomiw:
@wattsm at:
```
public class ClusterFixture : IDisposable 
{
    public ClusterFixture()
    {
        this.Cluster = new TestCluster();
        this.Cluster.Deploy();
    }

    public void Dispose()
    {
        this.Cluster.Dispose();
    }

    public TestCluster Cluster { get; private set; }
}
```

`this.Cluster.Dispose()` should be `this.Cluster.StopAllSilos`?

wattsm:
@shlomiw You're right, it should. I've updated my branch - can the change be merge from this PR or does it require a new one?

jdom:
Dispose does indeed stop all silos also, so it's OK, right?

wattsm:
@jdom `TestCluster` does not implement `IDisposable`.

jdom:
ah, hehe, I see, my bad. Then yeah, a new PR would be needed since this was merged.
Thanks to both of you :)

PR #3034, opened 2017-05-22
Suggestion of new feature.

skyaxl:


sergeybykov:
Thank you. That's a good suggestion. In fact such a request has been brought up multiple times before. I would just rephrase it for better clarity to something like the following:

6. **Secure communication between silos and clients**
* Add support for secure communication mode with certificates used for encryption of messages.

sergeybykov:
@skyaxl I incorporated your suggestion (after rephrasing it) in #3197. So I'm closing this one. Thanks again!

PR #3036, opened 2017-05-23
Support CultureInfo as built-in type

DixonDs:
I know that each and every type cannot be added as a built-in type for Orleans serialization, but I think `CultureInfo` deserves that as widely used system type. Json.Net also supports it out of the box meaning that most storage providers in Orleans already support `CultureInfo` for the grain state serialization. So let's have it supported for messaging as well.

### DIFF
--- a/test/NonSilo.Tests/SerializationTests/SerializationTests.DifferentTypes.cs
+++ b/test/NonSilo.Tests/SerializationTests/SerializationTests.DifferentTypes.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using Orleans.CodeGeneration;
 using Orleans.Serialization;
 using TestExtensions;
@@ -120,6 +121,7 @@ namespace UnitTests.Serialization
         }
 
 #if !NETSTANDARD_TODO
+
         [Fact, TestCategory("BVT"), TestCategory("Functional"), TestCategory("Serialization")]
         public void SerializationTests_RecursiveSerialization()
         {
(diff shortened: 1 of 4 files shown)


test/NonSilo.Tests/SerializationTests/SerializationTests.DifferentTypes.cs
```diff
@@ -131,6 +133,28 @@ public void SerializationTests_RecursiveSerialization()
...
+
+            foreach (var cultureInfo in input)
+            {
+                var output = this.fixture.SerializationManager.RoundTripSerializationForTesting(cultureInfo);
+                Assert.Equal(cultureInfo, output);
+            }
+        }
+
+        [Fact, TestCategory("BVT"), TestCategory("Functional"), TestCategory("Serialization")]
+        public void SerializationTests_CultureInfoList()
```
ReubenBond: This test isn't necessary, since it really just tests whether or not List<T> is serialized correctly

galvesribeiro:
Good work @DixonDs! Thanks

test/NonSilo.Tests/SerializationTests/SerializationTests.DifferentTypes.cs
DixonDs: @ReubenBond Well, only this test was failing before the code changes, that's why I kept it

ReubenBond: Ah, I see. If these tests were put into BuiltInSerializerTests.cs where we disable the fallback serializer, then the first test would be sufficient.

PR #3038, opened 2017-05-23
First iteration of merging the source trees

attilah:
- Update F# to 4.1.17 - the compiler and core versions are aligned by the F# team
- Update xUnit to 2.2.0
- Add F# projects to Orleans.2017 Solution (only dotnet.exe build can build it, excluded when building in VS, hence 2 projects are not loading)
- Update VSWhere to latest version
- Remove single output folder requirement for build
- Update 2017 project files to share common properties, enable assemblyinfo generation and nuget packing, removed unused things
- Create 2017 version of Build and Test batch files
- Enable /m for Build.2017.cmd


Directory.Build.props
```diff
@@ -0,0 +1,46 @@
...
+    <PackageTags>Orleans Cloud-Computing Actor-Model Actors Distributed-Systems C# .NET</PackageTags>
+    <PackageReleaseNotes></PackageReleaseNotes>
+    <RepositoryUrl>https://github.com/dotnet/Orleans</RepositoryUrl>
+    <RepositoryType>git</RepositoryType>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+  </PropertyGroup>
+
+  <!-- For Debug builds generated a date/time dependent version suffix -->
+  <PropertyGroup Condition=" '$(Configuration)'=='Debug' ">
+    <VersionSuffix Condition=" '$(VersionSuffix)'=='' ">dev</VersionSuffix>
```
ReubenBond: How does this work? I see that `VersionSuffix` is set above if it's not set

jdom:

Parallel-Tests.2017.ps1
```diff
@@ -0,0 +1,66 @@
+param(
```
jdom: is this file any different from Parallel-Tests.ps1?

src/OrleansCounterControl/app.manifest
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <asmv1:assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv1="urn:schemas-microsoft-com:asm.v1" xmlns:asmv2="urn:schemas-microsoft-com:asm.v2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-  <assemblyIdentity version="1.2.0.0" name="OrleansCounterControl.exe"/>
+  <assemblyIdentity version="1.5.0.0" name="OrleansCounterControl.exe"/>
```
jdom: Is this manifest of any value since it wasn't updated in a long time? /cc @jason-bragg

Directory.Build.props
```diff
@@ -0,0 +1,46 @@
...
+    <VersionSuffix Condition=" '$(VersionDateSuffix)'!='' ">$(VersionSuffix)-$(VersionDateSuffix)</VersionSuffix>
+  </PropertyGroup>
+
+  <!-- Until we removing old csproj files we need to exclude assembly attributes from AssemblyInfo.cs files -->
+  <PropertyGroup>
+    <DefineConstants>$(DefineConstants);EXCLUDE_ASSEMBLYINFO</DefineConstants>
```
jdom: ><DefineConstants>$(DefineConstants);EXCLUDE_ASSEMBLYINFO</DefineConstants> [](start = 4, length = 74)

If we are using the old src/Orleans.sln, will VS2017 use the Directory.Build.props and hence pick up these settings?

test/Consul.Tests/App.config
```diff
@@ -4,11 +4,5 @@
     <ThrowUnobservedTaskExceptions enabled="false" />
     <gcServer enabled="true" />
     <gcConcurrent enabled="false" />
-    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
-      <dependentAssembly>
-        <assemblyIdentity name="System.Diagnostics.DiagnosticSource" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-4.0.1.0" newVersion="4.0.1.0" />
-      </dependentAssembly>
-    </assemblyBinding>
```
jdom: Are you sure about this removal? I remember about a month ago this was driving me crazy as it is not autogenerated, but required at runtime

test/Tester/App.config
```diff
@@ -16,9 +16,6 @@
       </providers>
     </roleManager>
   </system.web>
```
jdom: Lol, this sounds fishy for this test project

jdom:
> <?xml version="1.0" encoding="utf-8" ?>

I noticed several App.config files added. Shouldn't they be added to src/Orleans.sln also?

---
Refers to: test/TesterSQLUtils/App.config:1 in 3f07fd9. [](commit_id = 3f07fd9ddf97bf41b594b08e0422013ccb0a12f5, deletion_comment = False)

test/TestGrainInterfaces/TestGrainInterfaces.2017.csproj
```diff
@@ -1,35 +1,27 @@
...
-  <ItemGroup>
     <ProjectReference Include="..\..\src\Orleans\Orleans.2017.csproj" />
     <ProjectReference Include="..\..\src\OrleansRuntime\OrleansRuntime.2017.csproj" />
     <ProjectReference Include="..\TestInterfaces\TestInterfaces.2017.csproj" />
   </ItemGroup>
+  <ItemGroup Condition="'$(IncludeFSharp)'=='true'">
+    <ProjectReference Include="..\TestFSharpInterfaces\TestFSharpInterfaces.2017.fsproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="FSharp.Core" Version="4.1.17" />
```
jdom: should this be inside the conditional?

jdom: hmm, I guess probably not, as I believe we have some tests against FSharp basic types, regardless of whether our FSharp grain projects are being built

---
In reply to: [118301010](https://github.com/dotnet/orleans/pull/3038#discussion_r118301010) [](ancestors = 118301010)

test/Consul.Tests/App.config
ReubenBond: @jdom didn't we work out that there is a property we can specify in the project to ensure that these *are* generated for DLLs? I think that's why

Directory.Build.props
attilah: It was something we already have in the master. that variable is consumed by default props/targets files.

Parallel-Tests.2017.ps1
attilah: Nope, but to not to have common files with current master build/test scripts having a "copy" makes easier the last cleanup.
In the vNext folder we followed the same pattern. The file is no different.

src/OrleansCounterControl/app.manifest
attilah: I think not updating it was an oversight? but needed for UAC prompt.

Directory.Build.props
attilah: For that we don't need this to be picked up, that's why it is an exclusion and not inclusion. We had this in master for a long time and it worked.

test/Tester/App.config
attilah: Agree, did not know what was the reason we're having it there, not touched ;-)

test/Consul.Tests/App.config
attilah: We're covering the grounds for binding redirect generation. If this is needed then we maybe have to have a test for it to see. I'd say it is safe.

jdom: @ReubenBond this was an issue in that particular PR, where I moved everything to auto-generated, but this one was still an issue

Directory.Build.props
jdom: what I meant is: will src/Orleans.sln use Directory.Build.props? If so, then this is still an issue

attilah: No, but we don't need to exclude F# grain stuff in Orleans master, since F# is supported when using the old csproj files. Nothing changed around here.

jdom:
@dotnet-bot test this please

jdom:
Reverting the change in #3060 for now, since it's conflicting with the non-2017 solution

Issue #3043, opened 2017-05-24
potential problem with generated serializer code


danvanderboom:
Using Orleans 1.4.1, I am getting these exceptions.

> **System.InvalidOperationException: Collection was modified; enumeration operation may not execute.**  
>    at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)  
>    at System.Collections.Generic.List`1.Enumerator.MoveNextRare()  
>    at Orleans.Serialization.BuiltInTypes.SerializeList[T](Object obj, ISerializationContext context, Type expected)  
>    at Orleans.Serialization.SerializationManager.SerializeInner(Object obj, ISerializationContext context, Type expected)  
>    at NotificationServer.Model.OrleansCodeGenNotificationServer_Model_WellSubscriptionEventSerializer.Serializer(Object untypedInput, ISerializationContext context, Type expected) in **C:\Source\myproject\NotificationServer\NotificationServer.Model\obj\Debug\NotificationServer.Model.orleans.g.cs:line 414**  
>    at Orleans.Serialization.SerializationManager.SerializeInner(Object obj, ISerializationContext context, Type expected)  
>    at Orleans.Serialization.BuiltInTypes.SerializeImmutable[T](Object original, ISerializationContext context, Type expected)  
>    at Orleans.Serialization.SerializationManager.SerializeInner(Object obj, ISerializationContext context, Type expected)  
>    at Orleans.Serialization.BuiltInTypes.SerializeInvokeMethodRequest(Object obj, ISerializationContext context, Type expected)  
>    at Orleans.Serialization.SerializationManager.SerializeInner(Object obj, ISerializationContext context, Type expected)  
>    at Orleans.Serialization.SerializationManager.Serialize(Object raw, BinaryTokenStreamWriter stream)  
>    at Orleans.Runtime.Message.Serialize_Impl(Int32& headerLengthOut, Int32& bodyLengthOut)  
>    at Orleans.Runtime.Messaging.Gateway.GatewaySender.Send(Message msg, Socket sock)

The line of code referred to is part of the generated code file (*.g.cs). Line 414 of it says:
```
global::Orleans.Serialization.SerializationManager.@SerializeInner(input.@WellIDs, context, typeof (global::System.Collections.Generic.List<global::System.Guid>));
```

Here is the whole generated class:
```
[...]

PR #3048, opened 2017-05-25
Ensure items are copied before yielding the thread in SMS stream provider

ReubenBond:
In #3043, @danvanderboom highlighted an issue in the current SMS stream provider whereby the thread can be yielded before an item is copied.

The cleanest fix to this which minimizes overhead and doesn't add unwanted complexity is to eagerly copy the item before we know the thread will be yielded.

### DIFF
--- a/src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducer.cs
+++ b/src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducer.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
 using Orleans.Runtime;
+using Orleans.Serialization;
 using Orleans.Streams;
 
 namespace Orleans.Providers.Streams.SimpleMessageStream
@@ -10,8 +12,13 @@ namespace Orleans.Providers.Streams.SimpleMessageStream
     {
         private readonly StreamImpl<T>                  stream;
         private readonly string                         streamProviderName;
+
+        [NonSerialized]
+        private readonly SerializationManager serializationManager;
+
         [NonSerialized]
         private readonly IStreamPubSub                  pubSub;
+
         [NonSerialized]
         private readonly IStreamProviderRuntime         providerRuntime;
         private SimpleMessageStreamProducerExtension    myExtension;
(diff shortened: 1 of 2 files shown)


src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducer.cs
```diff
@@ -86,7 +95,12 @@ internal class SimpleMessageStreamProducer<T> : IInternalAsyncBatchObserver<T>
             if (isDisposed) throw new ObjectDisposedException(string.Format("{0}-{1}", GetType(), "OnNextAsync"));
 
             if (!connectedToRendezvous)
+            {
+                // In order to avoid potential concurrency errors, synchronously copy the input before yielding the
+                // thread. DeliverItem below must also be take care to avoid yielding before copying for non-immutable objects.
+                item = (T)this.serializationManager.DeepCopy(item);
```
jason-bragg: Only needed if optimizeForImmutableData is false right?
item = optimizeForImmutableData ? item : (T)this.serializationManager.DeepCopy(item);

xiazen: I think optimizeForImmutableData is for configuring sending remote message to consumer grain, right? And this is before that step, the preparation step just to call its own prducerExtension to deliver this item, which is not something optimizeForImmutableData is designed to configure.

xiazen: this is just to handle the case found in #3034. So I think this is good. Just in case my words is confusing , I will state it very clear.

jason-bragg: Agreed, is this probably an unnecessary optimization.

xiazen: what do you mean? are you suspecting it won't fix the bug in #3034 or people shouldn't use Producer like in #3034.

ReubenBond: @jason-bragg you're right, but that information isn't available at this level of abstraction - is it?

jason-bragg: @ReubenBond looks like it is, but again, my suggestion is probably a premature optimization...  Signed off on change.

xiazen: sorry, Jason, I see what you mean now. yes it is an optimization !

sergeybykov:
@dotnet-bot test this

PR #3052, opened 2017-05-25
fix bug on ability to turn off AvgPressureMonitor using EventHubStreamProviderSettings

xiazen:
Since AvgPressureMonitorThreshold has a default value 
and EventHubStreamProvider repopulate its setting from IProviderConfig when initializing,  so just set it to null won't turn it off, it would just set it back to its default value. 

This will fix it.

### DIFF
--- a/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubStreamProviderSettings.cs
+++ b/src/OrleansServiceBus/Providers/Streams/EventHub/EventHubStreamProviderSettings.cs
@@ -46,7 +46,7 @@ namespace Orleans.ServiceBus.Providers
         /// User can turn it off by setting this value to null
         /// </summary>
         public double? AveragingCachePressureMonitorFlowControlThreshold = AveragingCachePressureMonitor.DefaultThreshold;
-
+        private const double CachePressureMonitorOffThreshold = 1.0;
         /// <summary>
         /// EventHubSettingsType setting name.
         /// </summary>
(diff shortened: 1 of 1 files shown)


jason-bragg:
What is the expected behavior?
As I understood it, it was an optional configuration, that if not set would use to the default.
We now want to be able to disable it, which is new, right?

This change, unless I'm misunderstanding it, will change the behavior such that if it's not configured, it will be turned off, am I understanding this correctly?

This is a breaking change, as this will disable the flow control for all systems currently using the default.

I think what we need is a way for users to disable the Averaging cache pressure flow control, should they choose to.  One option is to set it to a 'disable' setting as you've done, but expose it at the configuration level.  For instance

So we need some way for application developers to configure it to be off?

    public const double CachePressureOff = -1.0; 

and application configures does something like

    myEhspSettings.AveragingCachePressureMonitorFlowControlThreshold = CachePressureOff;

And in code we check:

    if (flowControlThreshold <= 0) 
        disable flow contorl
    else 
        set flow control

xiazen:
I apologize for the brief description, probably too short to explain this scenario. 
Offline chat with @jason-bragg to clear out the confusion. Post it as below in case someone has the same concern. 
This is not a breaking change, if the user doesn't explicitly specify AvgPressureMonitorTHreshold to null, then the stream provider would go on with its default value, which is the original/historical logic. 
and set the flowControlThreshold to -1 is a little confusing because it is not a valid value for flowControlThreshold , while 1 is and it technically means turning it off. So set it to one make more sense.

PR #3010, opened 2017-05-10
Reduce number of created tasks per grain request

dVakulen:
[Not for merge yet] 

One of the optimizations from https://github.com/dotnet/orleans/pull/2060

Currently ```OrleansTaskScheduler``` wraps each work item (invoke, response, etc) in task to ensure that activity is executed via an appropriate Task scheduler. Such behavior (creation of lots of tasks and their ```TryExecute``` overhead) can be pretty costly.

This PR reduces number of tasks created from ~3 per request to ~1 per activation lifetime while preserving  all the guarantees from http://dotnet.github.io/orleans/Documentation/Runtime-Implementation-Details/Scheduler.html. This is done via *a bit* hacky solution which most likely won't end up merged, but can be used to observe overhead of current implementation. 

If performance testing of this one will show that the overhead is indeed impactful -  
I'll describe some alternatives which provides a bit lower performance gains, but could be potentially merged.

PR #3054, opened 2017-05-25
[DNM] WorkItemGroup to use single task per activation lifetime

ReubenBond:
DoNotMerge

### DIFF
--- a/src/Orleans/Runtime/RuntimeContext.cs
+++ b/src/Orleans/Runtime/RuntimeContext.cs
@@ -5,7 +5,7 @@ namespace Orleans.Runtime
 {
     internal class RuntimeContext
     {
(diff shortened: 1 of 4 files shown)


gabikliot:
@reubenbond, can you explain? Also the hack with the IL.

ReubenBond:
@gabikliot apologies - this is merely a copy of @dVakulen's #3010 PR, but rebased so that I can build it on our CI. Unfortunately our load tests currently require us to create a build using the dotnet/orleans repo, which is why I had to open this dummy PR. It's something we're hoping to fix soon. Dmitry's change isn't ready to go in (and certainly we don't want those kinds of hacks with reflection if we can avoid it), but he just wanted some perf numbers from our cluster.

PR #3055, opened 2017-05-26
Enable runtime strategy change for versioning

benjaminpetit:
# Enable strategy version change at runtime
This PR enable the change at runtime of the version strategies (version selector and compatibility). Before this, strategy values are read from the Configuration and cannot be change at runtime.

## Why

The idea is that the developer might want to run for a given time two versions side-by-side and enable/disable manually the "version migration".

## Proposed changes

The strategy changes are done via the IManagementGrain, OrleansManager will be extended to support this feature.

### How strategy changes work

  1. a call to change strategy is made to the IManagementGrain for the cluster
  2. the new strategy is stored to a dedicated storage. It is needed to have such storage to survive a silo death/reboot
  3. the new strategy is then broadcasted to all silos in the cluster. If one silo misses the message or is unable to processes it now, the next TypeMap timer refresh should read the correct value that was stored previously

When a silo join the cluster, it will first use the values from the Configuration as default strategies. Then it will try to read the specific rules that were previously stored.

### Version strategy storage
The developer can implement his own storage with DI, by implementing the interface IVersionStore.

We provide a default implementation that internally use a stateful grain, configured with the default storage provider. If no default storage provider exists in the configuration, this feature is disabled.

This default implementation is unique per DeploymentId: if the DeploymentId changes, all previously stored values will not be used.


src/Orleans/SystemTargetInterfaces/IVersionManagement.cs
```diff
@@ -0,0 +1,39 @@
+using System.Threading.Tasks;
+using Orleans.Versions.Compatibility;
+using Orleans.Versions.Selector;
+
+namespace Orleans
+{
+    public interface IVersionManagement
```
ReubenBond: maybe `IVersionManager`?

src/OrleansRuntime/Core/ManagementGrain.cs
```diff
@@ -277,6 +316,34 @@ public Task<object[]> SendControlCommandToProvider(string providerTypeFullName,
...
+                silos,
+                s => applyFunc(GetSiloControlReference(s)));
+            try
+            {
+                await Task.WhenAll(actionPromises);
+            }
+            catch (Exception)
+            {
+                // ignored: silos that failed to set the new strategy will reload it from the storage
+                // in the futur.
```
ReubenBond: nit: future

src/OrleansRuntime/Versions/GrainVersionStore.cs
```diff
@@ -0,0 +1,89 @@
...
+using Orleans.Versions.Compatibility;
+using Orleans.Versions.Selector;
+
+namespace Orleans.Runtime.Versions
+{
+    internal class GrainVersionStore : IVersionStore
+    {
+        private readonly IInternalGrainFactory grainFactory;
+        private readonly string deploymentId;
+        private IVersionStoreGrain StoreGrain => this.grainFactory.GetGrain<IVersionStoreGrain>(this.deploymentId);
```
ReubenBond: What's the purpose of deploymentId here? I'm missing something

benjaminpetit: It was a cheap way to ensure that we don't use rules that were set with an older deploymentId.

benjaminpetit:

src/OrleansRuntime/Versions/GrainVersionStore.cs
ReubenBond: Got it - lgtm. As an aside: shouldn't grain storage be isolated by deployment id anyway? Different deployment = different logical system. It's not something enforced by the system, though, so maybe that's the issue.

PR #3058, opened 2017-05-26
SMS: avoid copy in OnNextAsync if user has specified immutable data optimization

ReubenBond:
Fixes a minor test breakage introduced in #3048.

A test was relying on data not being copied if the optimization for immutable data was set. In reality this isn't an issue, since we don't guarantee consumers and producers are on the same silo, but the fix is trivial. I didn't realize previously that this field was available, my mistake.

### DIFF
--- a/src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducer.cs
+++ b/src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducer.cs
@@ -96,9 +96,13 @@ namespace Orleans.Providers.Streams.SimpleMessageStream
 
             if (!connectedToRendezvous)
             {
-                // In order to avoid potential concurrency errors, synchronously copy the input before yielding the
-                // thread. DeliverItem below must also be take care to avoid yielding before copying for non-immutable objects.
-                item = (T)this.serializationManager.DeepCopy(item);
+                if (!this.optimizeForImmutableData)
+                {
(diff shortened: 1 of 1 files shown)


src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducer.cs
```diff
@@ -96,9 +96,13 @@ internal class SimpleMessageStreamProducer<T> : IInternalAsyncBatchObserver<T>
...
             if (!connectedToRendezvous)
             {
-                // In order to avoid potential concurrency errors, synchronously copy the input before yielding the
-                // thread. DeliverItem below must also be take care to avoid yielding before copying for non-immutable objects.
-                item = (T)this.serializationManager.DeepCopy(item);
+                if (!this.optimizeForImmutableData)
+                {
+                    // In order to avoid potential concurrency errors, synchronously copy the input before yielding the
+                    // thread. DeliverItem below must also be take care to avoid yielding before copying for non-immutable objects.
+                    item = (T) this.serializationManager.DeepCopy(item);
```
xiazen: @jason-bragg  you are correct after all. 💃

ReubenBond:

jdom:

jdom:
@jason-bragg please review asap so this goes in, since the current code is failing a test in most of our unrelated PRs

Issue #3057, opened 2017-05-26
invalid RuntimeContext in ServiceFabric silo on range changenotification


rikbosch:
Hi,

 We are experiencing some trouble connecting to our cluster running in service fabric. Looking in the silo logs, I see the following exception:

```
[2017-05-26 12:59:16.482 GMT     6    ERROR      103006    EquallyDividedRangeRingProvider    10.0.0.4:20032]    !!!!!!!!!! Local IGrainRingRangeListener Orleans.Streams.ConsistentRingQueueBalancer has thrown an exception when was notified about RangeChangeNotification about old <MultiRange: Size=x68811895, %Ring=40.822%> new <MultiRange: Size=x47EE99B4, %Ring=28.098%>.    
Exc level 0: System.InvalidOperationException: Trying to send a message Request S10.0.0.4:20032:233499509-> #15:  on a silo not from within grain and not from within system target (RuntimeContext is not set to SchedulingContext) RuntimeContext.Current=null TaskScheduler.Current=System.Threading.Tasks.ThreadPoolTaskScheduler
   at Orleans.Runtime.InsideRuntimeClient.SendRequestMessage(GrainReference target, Message message, TaskCompletionSource`1 context, Action`2 callback, String debugContext, InvokeMethodOptions options, String genericArguments)
   at Orleans.Runtime.GrainReference.InvokeMethod_Impl(InvokeMethodRequest request, String debugContext, InvokeMethodOptions options)
   at Orleans.Runtime.GrainReference.InvokeMethodAsync[T](Int32 methodId, Object[] arguments, InvokeMethodOptions options, SiloAddress silo)
   at Orleans.Streams.OrleansCodeGenStreamQueueBalanceListenerReference.QueueDistributionChangeNotification()
   at Orleans.Streams.ConsistentRingQueueBalancer.RangeChangeNotification(IRingRange old, IRingRange now)
   at Orleans.Runtime.ConsistentRing.EquallyDividedRangeRingProvider.RangeChangeNotification(IRingRange old, IRingRange now, Boolean increased)
```

orleans v1.5.0-beta

It looks like a TPL issue, where the notification from ```FabricQueryManager.RegisterPartitionChangeHandler``` comes back on a different context, but I'm not sure..

ReubenBond:
@rikbosch thanks for reporting. I believe you are correct that in your assessment. I'm looking into a fix now.

sergeybykov:
Resolved via #3059.

PR #3059, opened 2017-05-26
Service Fabric: execute cluster update notifications on an Orleans thread

ReubenBond:
Fix for #3057, in which @rikbosch is experiencing threading issues caused by the `FabricMembershipOracle` not executing cluster status notifications on an Orleans thread.

Proposed solution is to capture the `OrleansTaskScheduler` in the `IMembershipOracle.Start` call and later use it to schedule notifications.

### DIFF
--- a/Samples/ServiceFabric/Stateless/StatelessCalculatorService/Program.cs
+++ b/Samples/ServiceFabric/Stateless/StatelessCalculatorService/Program.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using System.Fabric;
 using System.Threading;
+using Grains;
 using Microsoft.ServiceFabric.Services.Runtime;
 
 namespace StatelessCalculatorService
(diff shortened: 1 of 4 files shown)


ReubenBond:
@dotnet-bot test netfx-bvt netfx-functional

ReubenBond:

ReubenBond:
@dotnet-bot test this please

gabikliot:
@ReubenBond , I have a suggestion about simplifyig sinchronization.
Instead of using lock in OnUpdate, I would reccomend mutating the oracle state on the oracle context. That way you don't need the lock at all and use only one synchronization procedures: the single threaded nature of the system target. Mixing 2 synch procedures is always error prone.
Just wrap the OnUpdate callback in a call via the context.

Same for cached list: instead of locking and copying on request from the caller, copy when the master list changes and just replace the cached pointer. No need in  a lock at all.

That is what I do in the table oracle.

ReubenBond:
@gabikliot the oracle isn't a `SystemTarget` since this only has access to the public API. This could perhaps inherit from `GrainService` and gain access to single-threaded execution / `SystemTarget` that way, but we would still need to explicitly schedule this out-of-band notification on its context.

ReubenBond:

ReubenBond:

odd failures https://ci.dot.net/job/dotnet_orleans/job/master/job/netfx-functional_prtest/348/

Issue #3056, opened 2017-05-26
AzureQueueStreamProvider errors after trying to publish message


charleszipp:
I am exploring orleans for a potential alternative to ServiceFabric actors. I am testing orleans 1.4.1 on service fabric. I have been researching the streams.
I was able to get the SMSProvider working but, the same publishing and subscription does not seem to work with AzureQueue. I am getting the following error. Also, the message is never received by the grain. No exception is thrown by the client producing the message.
```
ERROR 103317 PullingAgentSystemTarget/FF/921360bb-AZQProvider 127.0.0.1:30036] !!!!!!!!!! Ignored RegisterAsStreamProducer Error Exc level 0: System.ArgumentException: Trying to register non active IStreamProducerExtension: SystemTarget:PullingAgentSystemTarget/FF/921360bb/S127.0.0.1:30036:233499092 Parameter name: streamProducer
```

I feel like i may be missing something in the setup or, the persistent streams require a few more steps than the SMSProvider.

Stateless Service hosting the silo
```csharp
internal sealed class OrleansStatelessService : StatelessService
{
    public OrleansStatelessService(StatelessServiceContext context)
        : base(context)
    { }

    public ClusterConfiguration GetClusterConfiguration()
    {
        var config = new ClusterConfiguration();

        config.Globals.ReminderServiceType = GlobalConfiguration.ReminderServiceProviderType.ReminderTableGrain;
        config.Globals.DataConnectionString = "UseDevelopmentStorage=true";
        config.Defaults.StartupTypeName = typeof(ClusterStartup).AssemblyQualifiedName;
        config.AddAzureTableStorageProvider("Default");
        config.AddAzureTableStorageProvider("PubSubStore");
        config.AddAzureQueueStreamProvider("AZQProvider", deploymentId: Regex.Replace(this.Context.ServiceName.PathAndQuery.Trim('/'), "[^a-zA-Z0-9_]", "").ToLower());
            
        LogManager.LogConsumers.Add(new EventSourceLogger());
        return config;
    }

    protected override IEnumerable<ServiceInstanceListener> CreateServiceInstanceListeners()
    {
        ClusterStartup.Service = this;
        return new[] { OrleansServiceListener.CreateStateless(this.GetClusterConfiguration()) };
    }

    protected override async Task RunAsync(CancellationToken cancellationToken)
    {
        while (true)
        {
            cancellationToken.ThrowIfCancellationRequested();
[...]

PR #3061, opened 2017-05-27
Service Fabric: Mark removed silos as Dead indefinitely

ReubenBond:
Instead of setting a no-longer-present silo's status to `SiloStatus.None`, leave a tombstone declaring it `Dead`.

Related to #3056

### DIFF
--- a/src/OrleansServiceFabricUtils/FabricMembershipOracle.cs
+++ b/src/OrleansServiceFabricUtils/FabricMembershipOracle.cs
@@ -96,7 +96,8 @@ namespace Microsoft.Orleans.ServiceFabric
 
                 // Take all the active silos if their count does not exceed the desired number of gateways
                 var maxSize = this.globalConfig.MaxMultiClusterGateways;
-                result = new List<SiloAddress>(this.silos.Keys);
+                var gateways = this.silos.Where(entry => entry.Value.Status == SiloStatus.Active).Select(entry => entry.Key);
+                result = new List<SiloAddress>(gateways);
 
                 // Restrict the length to the maximum size.
                 if (result.Count > maxSize)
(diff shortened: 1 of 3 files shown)


gabikliot:
@ReubenBond , did you look at trying to reuse MBR Oracle Data? It "is supposed" to abstract the oracle data from how the data is produced. It has all the old-new data manipulations, subscribers, notifications, ...
So in theory you should be able to use it just as is.
Since I only had one oracle, its probably not so clean and maybe some small code changes will be needed.
It can help you reduce code dup.

ReubenBond:
Currently the Service Fabric support does not have access to Orleans internals, which does make it a little harder to write but has the benefit of ensuring that we can write these kinds of providers without privileged access.

gabikliot:
We already made a bunch and a bunch of internal classes public. This train has left the station long time ago. Just do the same. Its in the right namespace.

ReubenBond:
I don't consider "that ship has sailed" to be sufficient reasoning for opening the flood gates. Even though more classes have been opened (which is a good thing), they haven't been opened without some kind of consideration.

I'll look into this further - it's a significant change to move to this data structure since the SF oracle doesn't use `MembershipEntry` at its core like the other oracle does. Either I change `MembershipOracleData` to use a slimmed down structure (maybe a common interface), or I create `MembershipEntry` instances which have most of their fields empty (eg, everything to do with fault/update domains and silo health)

gabikliot:
Ok, forget the opening the flood gates. Bad idea 
If it requires large refactoring, it's probably not worth it.
I would also be very careful not to break an existing Oracle.

You can ignore my suggestion.

ReubenBond:
Thanks, @gabikliot, I will still look into it and see if there's a simple way to unify the two. Thanks also for checking the PR

ReubenBond:
Anyone able to review this?

PR #3067, opened 2017-05-30
Fix provider type for ADO.Net storage provider in config example

wattsm:
The type name in the configuration example appears to be incorrect - I could find no type named `Orleans.SqlUtils.StorageProvider.SqlStorageProvider` in the `Microsoft.Orleans.SqlUtils` package. I believe the correct type name is `Orleans.Storage.AdoNetStorageProvider` and have updated the example.


veikkoeeva:
@wattsm That is correct.

sergeybykov:
Thank you, @wattsm!

PR #3069, opened 2017-05-31
Fix resolve ip address.

yartat:
Fix resolve ip address.
We have a problem when configuring silo hosts. If I setup in hosts file network name (for example 'orleans') with IP 127.0.0.1 and set it as Address in SeedNode, Networking or ProxyingGateway the load configuration will trow `ArgumentException`. The host name 'orleans' is a valid host with valid loopback IP in this case.
So this fix checks any valid host name from hosts file, DNS entries (including case whena DNS responds 'localhost' or '127.0.0.1'). In additional I add parsing IP addresses (IPv4 and IPv6). So it possible to set not only 127.0.0.1, but any valid IP address.

### DIFF
--- a/src/Orleans/Configuration/ClusterConfiguration.cs
+++ b/src/Orleans/Configuration/ClusterConfiguration.cs
@@ -419,58 +419,49 @@ namespace Orleans.Runtime.Configuration
 
         internal static async Task<IPAddress> ResolveIPAddress(string addrOrHost, byte[] subnet, AddressFamily family)
         {
-            var loopback = (family == AddressFamily.InterNetwork) ? IPAddress.Loopback : IPAddress.IPv6Loopback;
+            var loopback = family == AddressFamily.InterNetwork ? IPAddress.Loopback : IPAddress.IPv6Loopback;
 
-            if (addrOrHost.Equals("loopback", StringComparison.OrdinalIgnoreCase) ||
-                addrOrHost.Equals("localhost", StringComparison.OrdinalIgnoreCase) ||
-                addrOrHost.Equals("127.0.0.1", StringComparison.OrdinalIgnoreCase))
+            // IF the address is an empty string, default to the local machine
+            if (string.IsNullOrEmpty(addrOrHost))
             {
-                return loopback;
+                addrOrHost = Dns.GetHostName();
             }
-            else if (addrOrHost == "0.0.0.0")
+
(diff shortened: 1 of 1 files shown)


src/Orleans/Configuration/ClusterConfiguration.cs
```diff
@@ -419,58 +419,55 @@ public string ToString(string siloName)
...
+            if (string.IsNullOrEmpty(addrOrHost))
+            {
+                addrOrHost = Dns.GetHostName();
+            }
 
-            if (addrOrHost.Equals("loopback", StringComparison.OrdinalIgnoreCase) ||
-                addrOrHost.Equals("localhost", StringComparison.OrdinalIgnoreCase) ||
-                addrOrHost.Equals("127.0.0.1", StringComparison.OrdinalIgnoreCase))
+            // Fix StreamFilteringTests_SMS tests
+            if (addrOrHost.Equals("loopback", StringComparison.OrdinalIgnoreCase))
```
jason-bragg: So we lost 'localhost'?

127.0.0.1 and 0.0.0.0 are covered by IPAddress.TryParse, but I don't see how ''localhost' is covered..

jason-bragg: I apologize for not being clear the first time I commented.

What I was getting at was that we specifically handled certain cases, and I wanted to ensure that we continued to do so.
Those cases were:
0.0.0.0 - which returned IPAddress.Any, and is currently handled by the IPAddress.TryParse.
127.0.0.1 - which returned loopback, and is currently handled by the IPAddress.TryParse.
"loopback" - which returned loopback, and is currently doing the same, with the last commit.
"localhost" - which returned loopback, and is not being handled as previously.

I understand that I'm being a bit of a stickler here, but we've many services using this, so we need to ensure we're not breaking any existing users without good cause.

yartat: > So we lost 'localhost'?
The function `Dns.GetHostAddressesAsync(addrOrHost)` returns loopback IP in case  `addrOrHost == "localhost"`. It works if OS contains loopback interface (all current OS).

jason-bragg: Ah, thanks!

jason-bragg:
Thanks for the fix, and quick turnaround on comments!

Issue #3064, opened 2017-05-29
DeepCopier with ObservableCollection suddenly requires all class graph to be marked as [Serialized]


shlomiw:
Hi,

I've changed a property of a class from List<> to ObservableCollection<>.
This class is EF entity and it has a deep graph of classes connected.

I'm sending this class via grain, and I had no issues before the change.
Now I'm getting SerializationException: 'Type .. in Assembly .. is not marked as serializable'. And I need to do it to *all* the different classes types in the object graph.

Can we avoid setting all these classes as [Serializable]?

Thanks in advance,
Shlomi

Attaching a screenshot of the exception:
![image](https://cloud.githubusercontent.com/assets/1025170/26555235/c8055db0-449c-11e7-94fb-0dc1a7052703.png)

ReubenBond:
Hi @shlomiw, here's what's happening: `ObservableCollection<T>` does not have a defined serializer, therefore the fallback serializer (BinaryFormatter) is being used. BinaryFormatter requires that classes be marked as`[Serializable]`.

To fix it, follow the instructions in the [Serialization documentation](http://dotnet.github.io/orleans/Documentation/Advanced-Concepts/Serialization.html) for creating a custom serializer for `ObservableCollection<T>`. Note that the class defined under 'Method 3' of that documentation can be an open generic type (eg `MyObservableCollectionSerializer<T>`) - just make sure to mark it as a serializer for that type: `[Serializer(typeof(ObservableCollection<>))]`.

Let me know if that helps

shlomiw:
@ReubenBond - thank you very much! I'll try it and update you.

shlomiw:
@ReubenBond - I used method 3 and it works :)
Thanks!!

shlomiw:
@ReubenBond - one small question: 
IsOrleansShallowCopyable is internal, but could be useful for custom serializers. What about exposing it? maybe via SerializationManager.IsOrleansShallowCopyable(Type t) (or <T>)

ReubenBond:
@shlomiw DeepCopy is a no-op in the case that a type is shallow-copyable. In which cases would `IsShallowCopyable` be useful to you? We can consider exposing it

shlomiw:
@ReubenBond for my custom serializer I've used the code of Orleans List<T> serialization as a reference.
From Orleans.Serialization.BuiltInTypes:

```
        internal static object DeepCopyList<T>(object original, ICopyContext context)
        {
            var list = (List<T>)original;

[...]

PR #3077, opened 2017-06-01
Exposing internal IsOrleansShallowCopyable to for external custom serializers 

shlomiw:
Hi,
As discussed with @ReubenBond in #3064 , I'm exposing `IsOrleansShallowCopyable` for external custom serializer.

Background: I've created a custom serializer for `ObservableCollection`, and copied the logic of copying `List<T>` from `BuiltInTypes.DeepCopyList()`, it uses this method to allow quick shallow copy.

### DIFF
--- a/src/Orleans/Serialization/SerializationManager.cs
+++ b/src/Orleans/Serialization/SerializationManager.cs
@@ -1045,20 +1045,6 @@ namespace Orleans.Serialization
                 ". Perhaps you need to mark it [Serializable] or define a custom serializer for it?");
         }
 
-        /// <summary>        
-        /// Returns <see langword="true"/> if instances of the provided type can be safely shallow-copied;
-        /// otherwise <see langword="false"/>, indicating that instances must instead be deep-copied.
-        /// </summary>
-        /// <param name="type">The type to check.</param>
-        /// <returns>
-        /// <see langword="true"/> if instances of the provided type can be safely shallow-copied; otherwise
-        /// <see langword="false"/>, indicating that instances must instead be deep-copied.
-        /// </returns>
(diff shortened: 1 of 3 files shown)


src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -1045,6 +1045,16 @@ private object DeepCopierHelper(Type t, object original, ICopyContext context)
                 ". Perhaps you need to mark it [Serializable] or define a custom serializer for it?");
         }
 
+        /// <summary>        
+        /// This method exposes internal IsOrleansShallowCopyable to allow external custom serializers to check if is it ok to shallow copy objects when deep copying.               
+        /// </summary>
+        /// <param name="type">The type to check.</param>
+        /// <returns>Returns true if ok to shallow copy.</returns>
+        public static bool IsTypeOrleansShallowCopyable(Type type)
```
ReubenBond: `IsShallowCopyable(Type)` would be shorter & still not ambiguous, what do you think?

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -1045,6 +1045,16 @@ private object DeepCopierHelper(Type t, object original, ICopyContext context)
                 ". Perhaps you need to mark it [Serializable] or define a custom serializer for it?");
         }
 
+        /// <summary>        
+        /// This method exposes internal IsOrleansShallowCopyable to allow external custom serializers to check if is it ok to shallow copy objects when deep copying.
```
ReubenBond: `Returns true if the provided type can be safely shallow-copied; otherwise false, indicating that the type must instead be deep-copied.`

shlomiw:
Thanks you very much :)

PR #3081, opened 2017-06-02
Wrap Blob/Table Storage Conflict error into InconsistentStateException

DixonDs:
We observe the following scenario from time to time - the grain writes its state for the first time, the server returns `ServerTimeoutError` (no eTag returned), then the next time the grain tries to write its state, the server returns `Conflict` because the first insert actually happened.

I think it should follow the same pattern as with Precondition Failed and use the logic added in https://github.com/dotnet/orleans/pull/2959

### DIFF
--- a/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
+++ b/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
@@ -249,9 +249,9 @@ namespace Orleans.Storage
             {
                 await updateOperation.Invoke().ConfigureAwait(false);
             }
-            catch (StorageException ex) when (ex.IsPreconditionFailed())
+            catch (StorageException ex) when (ex.IsPreconditionFailed() || ex.IsConflict())
             {
-                throw new InconsistentStateException($"Blob storage condition not Satisfied.  BlobName: {blob.Name}, Container: {blob.Container?.Name}, CurrentETag: {currentETag}", "Unkown", currentETag, ex);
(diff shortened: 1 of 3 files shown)


jdom:
I think you are right, this might be a reasonable path. I'll wait for someone else to validate too.
Nevertheless, if we go this route, I believe AzureBlobStorage provider should also be fixed. I'd recommend you find references to IsPreconditionFailed and apply the same logic to all callers.

gabikliot:
So conflict is retuned when trying the insert and the row exists? Vs. precondition failed on mismatched update?

If yes, then the fix looks correct to me.
Agree with Julian to apply it to blob too.

ReubenBond:
If there's no useful distinction between the two error codes, perhaps they should be merged into a single method to ensure that no discrepancies can sneak in.

DixonDs:
Shall we also treat 500 (Internal Server Error) in the same way? The actual update may or may not be failed in this case, but it also means that we haven't got the updated eTag back, so the grain state in some sense is corrupted as well.

jdom:
Not really, because we do not know whether the state is inconsistent or not, we should not make those assumptions at the infrastructure level. Users should be able to catch that and react accordingly if they wish to do so. 
Granted, it would be nice to normalize exceptions so that the user doesn't get at StorageException directly, and instead could look at the InnerException to dig deeper, but try to have the storage provider as a less leaky abstraction. Still 500 is not InconsistentStateException. All that said, please don't tackle this in the same PR if we decide to go forward with something like that

PR #3082, opened 2017-06-02
Blocked grains in "Deactivating" state detection and mitigation 

benjaminpetit:
A Grain can be blocked on the Deactivating state. The following PR extends the logic implemented in #2387 to detect such case.


benjaminpetit:

gabikliot:
Do we have the same for blocking on activation?

benjaminpetit:
> Do we have the same for blocking on activation?
No, I didn't want to add this logic yet, since we already have a complex process in `Catalog.InitActivation`. I can it in the future, in another PR since I think it will involve more changes

Issue #2000, opened 2016-07-29
Allow multiple silo interceptors to be set


richorama:
The current Orleans interceptor API provides methods for getting and setting interceptors:

``` c#
InvokeInterceptor GetInvokeInterceptor();
void SetInvokeInterceptor(InvokeInterceptor interceptor);
```

The problem is that a second call to `SetInvokeInterceptor` will overwrite the original interceptor. This means that two bootstrap providing could be competing for a single interceptor, unless the developer observes any previously set interceptor.

I propose adding a new method to add an interceptor to an internally maintained stack. The runtime code will then run each of the interceptors in the stack in turn.

This would allow easier registration of multiple interceptors.

ReubenBond:
Each interceptor wraps the invocation of the next interceptor, similar to an OWIN middleware.
```C#
request => {
  // do stuff
  var result = next();
  // do stuff with result
  return result;
}
```

Does ASP.NET 5 use a similar model to OWIN?

galvesribeiro:
Yes. Asp.Net Core has the composable middleware behavior and we could use that as example

Eldar1205:
Why not simply use the existing interceptor in the execution of the new interceptor?

richorama:
@Eldar1205 yes, this is a valid solution, however it's not intuitive. You can't tell from the API that you need to wrap any existing interceptor, which leads to interceptors overwriting each other.

Eldar1205:
I think the API leads to realization there is one interceptor since there
is a Get which returns a single interceptor
בתאריך 9 בספט' 2016 11:40,‏ "Richard Astbury" <notifications@github.com>
כתב:

ReubenBond:
Alright, time to make some progress on this issue. From an API perspective, how does this look:
```C#
// Make this internal/private:
// InvokeInterceptor GetInvokeInterceptor();

void AddInvokeInterceptor(InvokeInterceptor interceptor);
void RemoveInvokeInterceptor(InvokeInterceptor interceptor);
```

The configured `InvokeInterceptor` will form a chain and be executed in sequence as described above. There'll be no guarantee on ordering of interceptors - they will be executed in insertion order.

Are there cases where we need to be able to guarantee an ordering?

If we have general consensus, then I'll go ahead and implement this

veikkoeeva:
[...]

PR #3083, opened 2017-06-02
Support multiple request interceptors

ReubenBond:
Implements #2000 and marks the existing interface methods as obsolete.

Interceptors are configured by registering `InvokeInterceptor` instances with the DI container in the `ConfigureServices` startup method, like so:
```C#
services.AddSingleton<InvokeInterceptor>((method, request, target, invoker) =>
{
    RequestContext.Set(InterceptedPropertyKey, 1);
    return invoker.Invoke(target, request);
});
```

The interceptor set via `IProviderRuntime.SetInvokeInterceptor(...)` is executed last.

Future work: add an extension method to the future SiloBuilder to make configuration more discoverable/obvious.


ReubenBond:
@jdom I considered that, but there are a couple of reasons why I didn't take that route.
```C#
Task<object> InvokeInterceptor(
  MethodInfo targetMethod,
  InvokeMethodRequest request,
  IGrain target,
  IGrainMethodInvoker invoker);
```
The last argument is `IGrainMethodInvoker`:
```C#
public interface IGrainMethodInvoker
{
    int InterfaceId { get; }
    ushort InterfaceVersion { get; }
    Task<object> Invoke(IAddressable grain, InvokeMethodRequest request);
}
```
Note no invoker is passed and no `MethodInfo` is passed.

So these APIs do not compose well. Maybe it would be better if `InvokeInterceptor` was recursively defined like:
```C#
Task<object> InvokeInterceptor(
  MethodInfo targetMethod,
  InvokeMethodRequest request,
  IGrain target,
  InvokeInterceptor invoker);
```

We could rethink these APIs to remove the impedance mismatch. The mismatch causes more allocations than I would like, too :P

src/OrleansRuntime/Core/InsideRuntimeClient.cs
```diff
@@ -357,11 +373,13 @@ private Task<object> InvokeWithInterceptors(IAddressable target, InvokeMethodReq
...
             // call.
-            var siloWideInterceptor = this.CurrentStreamProviderRuntime.GetInvokeInterceptor();
             var grainWithInterceptor = target as IGrainInvokeInterceptor;
 
             // Silo-wide interceptors do not operate on system targets.
-            var hasSiloWideInterceptor = siloWideInterceptor != null && target is IGrain;
+#pragma warning disable 618
+            var hasSiloWideInterceptor = target is IGrain && (this.CurrentStreamProviderRuntime?.GetInvokeInterceptor() != null ||
+#pragma warning restore 618
+                                          this.siloInterceptors.Count > 1);
```
ReubenBond: It's quite unfortunate. How would it look if I treated it separately? I'm not sure

ReubenBond:
To be clear: I would be happy for these APIs to be fixed so that we can implement this feature more cleanly. The mismatch in APIs is why I put this implementation off for so long - the logic in `InvokeWithInterceptors` makes me uncomfortable because I tried to make these features pay-for-what-you-use and that resulted in complexity.

ReubenBond:
Looking more at ASP.NET Core for inspiration/alignment. As pointed out in your Filters link, @jdom, they have this [`ActionExecutingContext`](https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/ActionExecutingContext.cs) type. That type looks like our `InvokeMethodRequest`, except it also has a `Result` property.
This is then plumbed through to [`IAsyncActionFilter.OnActionExecutionAsync`](https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IAsyncActionFilter.cs): `Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next);`. 
Where `ActionExecutionDelegate` is defined as:
```C#
public delegate Task<ActionExecutedContext> ActionExecutionDelegate();
```
The reason for the return value there is that MVC supports both async and sync action filters. The context is passed from the async delegate into the (old) sync filter. It's not needed for us.

So mapping this to our world, we might have:
```C#
public delegate Task MethodInvocationDelegate();

public delegate Task MethodInvocationInterceptor(MethodInvocationContext context, MethodInvocationDelegate next);

public class MethodInvocationContext
{
    public IAddressable Grain { get; } // Maybe this class would wrap `InvokeMethodRequest`
    public MethodInfo Method { get; }
    public object[] Arguments { get; }
    public object Result { get; set; } 
}
```

After awaiting `next()` in an `MethodInvocationInterceptor`, `context.Result` will be set to whatever the method returns. `context.Result` is mutable, so it can be modified after awaiting `next()` anyway.

How does this API look?

jdom:
Can you clarify something of the proposed API: would people just register their `MethodInvocationInterceptor` delegates in DI or would we still have a base interface and class similar to `IAsyncActionFilter`?

ReubenBond:
I'm not sure, @jdom I feel that delegates are clear, but we could always allow interfaces too and convert between the two - method signature would be the same.

ReubenBond:
@jdom pushed an update - let me know what you think of the latest revision. It needs to be squashed.

src/OrleansRuntime/Core/GrainMethodInvoker.cs
```diff
@@ -0,0 +1,183 @@
...
+            }
+#pragma warning restore 618
+
+            if (stage == numInterceptors + 2)
+            {
+                stage++;
+
+                // Deprecated silo-level invoker, if present.
+                var grain = this.Grain as IGrain;
+                if (grain != null && deprecatedInvokeInterceptor != null)
```
jdom: Is the order correct? I would have imagined that the deprecated silo level interceptor came before the grain level interceptor

src/OrleansRuntime/Core/GrainMethodInvoker.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        /// <inheritdoc />
+        public object[] Arguments => request.Arguments;
+        
+        /// <inheritdoc />
+        public object Result { get; set; }
+
+        /// <inheritdoc />
+        public async Task Invoke()
+        {
```
jdom: If any of the interceptor sets the Result property, it should short-circuit the state machine, right?

src/Orleans/Providers/IProviderRuntime.cs
```diff
@@ -89,4 +92,48 @@ public interface ILogConsistencyProviderRuntime : IProviderRuntime
     /// <returns>The result of invocation, which will be returned to the client.</returns>
     public delegate Task<object> InvokeInterceptor(
         MethodInfo targetMethod, InvokeMethodRequest request, IGrain target, IGrainMethodInvoker invoker);
+
+    /// <summary>
+    /// The delegate invoked to execute the request.
+    /// </summary>
+    /// <returns>A <see cref="Task"/> which must be awaited before processing continues.</returns>
+    public delegate Task MethodInvocationDelegate();
```
jdom: Just a thought, but should we call them Filters (or GrainCallFilters or something like that) now that it fully aligns with action filters?

src/OrleansRuntime/Core/GrainMethodInvoker.cs
ReubenBond: you're right - good point

src/OrleansRuntime/Core/GrainMethodInvoker.cs
ReubenBond: I'm not sure. Is that more or less surprising to the user? I think more. If the user wants to short-circuit evaluation, they can by not calling `await context.Invoke()`

src/Orleans/Providers/IProviderRuntime.cs
ReubenBond: I was wondering the same thing yesterday. I'm amenable to the name.

ReubenBond: In latest, I've called them GrainCallFilters

ReubenBond:
@Eldar1205 I'm interested in your feedback (and others) about the proposed interface. If you or your team would like to take a look at this PR (eg, the tests) and tell me what you think, I would appreciate it.

src/Orleans/Core/IGrainCallContext.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Gets the arguments for this method invocation.
+        /// </summary>
+        object[] Arguments { get; }
+
+        /// <summary>
+        /// Invokes the request.
+        /// </summary>
+        Task Invoke();
```
jdom: in the future we *may* consider `ValueTask` for this, right?

src/Orleans/Core/IGrainInvokeInterceptor.cs
```diff
@@ -1,11 +1,18 @@
...
+using System.Threading.Tasks;
+using Orleans.CodeGeneration;
+
 namespace Orleans
 {
-    using System.Reflection;
-    using System.Threading.Tasks;
-    using Orleans.CodeGeneration;
-
+    [Obsolete("Use IMethodInvocationInterceptor instead. This interface may be removed in a future release.")]
```
jdom: This string missed the rename to IGrainCallFilter

ReubenBond: ahhh thanks

src/Orleans/Core/IGrainCallContext.cs
ReubenBond: Yes, that's a good idea.

src/Orleans/Core/IGrainInvokeInterceptor.cs
```diff
@@ -1,11 +1,18 @@
...
     public interface IGrainInvokeInterceptor
     {
         Task<object> Invoke(MethodInfo method, InvokeMethodRequest request, IGrainMethodInvoker invoker);
     }
+
+    public interface IGrainCallFilter
```
jdom: either move to a different file (preferable) or at least rename the file, since this is now the important interface for interception

ReubenBond: will do, also documented it.

test/TestGrains/MethodInterceptionGrain.cs
```diff
@@ -97,6 +98,53 @@ public Task<object> Invoke(MethodInfo methodInfo, InvokeMethodRequest request, I
...
+    {
+        private const string Key = GrainCallFilterTestConstants.Key;
+        private IGrainCallContext context;
+
+        public async Task<string> Execute(bool early, bool mid, bool late)
+        {
+            if (late)
+            {
+                this.context.Arguments[2] = false;
+                await this.context.Invoke();
```
jdom: I understand you are probably showing how to break this, but this is really not how filters should be used, right? And since it's the only example of usage in a grain, it might be misleading. Do you want to create a separate test/example grain that doesn't do all these edgy stuff and just plays along with the happy path?

ReubenBond: I'll refactor the tests for the other grains which use the deprecated method and put a note here.

test/TestGrains/MethodInterceptionGrain.cs
```diff
@@ -97,6 +98,53 @@ public Task<object> Invoke(MethodInfo methodInfo, InvokeMethodRequest request, I
...
 
         public Task<int> GetBestNumber() => Task.FromResult(38);
+    }
+
+    public class GrainCallFilterTestGrain : Grain, IGrainCallFilterTestGrain, IGrainCallFilter, IGrainInvokeInterceptor
+    {
+        private const string Key = GrainCallFilterTestConstants.Key;
+        private IGrainCallContext context;
+
+        public async Task<string> Execute(bool early, bool mid, bool late)
```
jdom: grain method named called Execute is hard to immediately know if this is related to interception or not, since Invoke and Execute both sound like infrastructure methods. Maybe rename to `DoStuff` or something that's clear it's a domain-level method

test/Tester/GrainCallFilterTests.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// <returns>A <see cref="Task"/> representing the work performed.</returns>
+        [Fact]
+        public async Task GrainCallFilter_Order_Test()
+        {
+            var grain = this.GrainFactory.GetGrain<IGrainCallFilterTestGrain>(random.Next());
+
+            // This grain method reads the context and returns it
+            var context = await grain.GetRequestContext();
+            Assert.NotNull(context);
+            Assert.Equal("grain!", context);
```
jdom: lol, this is kind of clever, although it's probably much more readable if this would end up concatenating "123456"....

src/OrleansRuntime/Core/GrainMethodInvoker.cs
```diff
@@ -0,0 +1,182 @@
...
+            // Execute each stage in the pipeline. Each successive call to this method will invoke the next stage.
+            // Stages which are not implemented (eg, because the user has not specified an interceptor) are skipped.
+            var numFilters = filters.Count;
+            if (stage < numFilters)
+            {
+                // Call each of the specified interceptors.
+                var systemWideFilter = this.filters[stage];
+                stage++;
+                await systemWideFilter.Invoke(this);
```
jdom: when filters throw (either synchronously or return a faulted Task), will everything work fine? I couldn't find any coverage for that scenario (but tests are a little bit confusing, so there is a chance there is coverage and I'm missing it)

jdom:
@richorama might be a good candidate to check whether it would still support the Dashboard in a good way (especially WRT to configuration)

test/TestGrains/MethodInterceptionGrain.cs
ReubenBond: Will rename to `CallWithBadInterceptors`

test/Tester/GrainCallFilterTests.cs
ReubenBond: ok, sure - makes sense

src/OrleansRuntime/Core/GrainMethodInvoker.cs
ReubenBond: I'll add coverage.

ReubenBond:
Added more test coverage & improved test execution time. You may notice I also added a new test class, that's because I had previously hijacked interception tests to check that `[MethodId(x)]` overrides were working. I've separated those into separate tests.

Eldar1205:
Hi Reuben, I've looked at the tests to see the new API and I really like
the introduction of IGrainCallContext with Invoke() method to simplify
invoking the next interceptor and the properties to access the current
method call context. I do find some names misleading:
- IGrainCallContext reminds of the .Net call context, e.g.
LogicalCallContext. I would prefer the name IGrainMethodCallContext or even
better IGrainMethodInvocationContext, but IGrainCallContext is also fine.
- IGrainCallFilter reminds of the WebApi filters, where I assume the
inspiration is from since the API doesn't necessarily filter anything,
however WebApi filters looked very differently, and the API reminds more of
the WebApi message handlers and Owin middleware mentioned previously, so I
would prefer the name IGrainCallMiddleware or IGrainCallHandler or stick
with well-known interceptor term and call it IGrainCallInterceptor. At any
case, I find the term 'filter' misleading, though I understand if you don't
like to change it due to the amount of code docs written with the term
'filter'.

I'll also note that I'm one of those that dislike the usage of
IServiceCollection as a way to hook behavior into Orleans, but it seems
like the current trend so I'll accept it hoping the online documentation
would list of all the hook points injectable via IServiceCollection, since
it's not the most discoverable. The extension methods to IServiceCollection
make it much better, but one still needs to have the correct 'using' to
find them without using tools like ReSharper.

2017-06-13 20:57 GMT+03:00 Reuben Bond <notifications@github.com>:

ReubenBond:
@Eldar1205 thank you, much appreciated :)
The 'filter' name is based on [`IAsyncActionFilter`](https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IAsyncActionFilter.cs) from MVC. The motivation behind aligning with ASP.NET / MVC is to reduce the number of concepts people need to learn when moving between the two worlds.
As for naming, I'm open to suggestion - I'll discuss with @jdom & others, too.

The extension methods are in the `Orleans` namespace, so they should be easily discoverable. I'll make sure the docs are clear.

richorama:
@jdom good idea, I'll take a look shortly.

test/TestGrains/MethodInterceptionGrain.cs
```diff
@@ -97,6 +128,62 @@ public Task<object> Invoke(MethodInfo methodInfo, InvokeMethodRequest request, I
...
+            //
+
+            this.context = ctx;
+            var value = RequestContext.Get(Key) as string;
+            if (value != null) RequestContext.Set(Key, value + '4');
+            await ctx.Invoke();
+            this.context = null;
+        }
+
+        public async Task<object> Invoke(MethodInfo method, InvokeMethodRequest request, IGrainMethodInvoker invoker)
```
jdom: can you implement the interfaces explicitly? It's still a little confusing without going back and forth to know which are these methods for, especially since these test grains implement both the new and the deprecated way of intercepting.
Also, it's probably makes sense logically too. These are cross-cutting concerns not related to the domain, so having them implemented explicitly instead of publicly makes sense as a good example. Not that the behavior will be any different though

src/OrleansRuntime/Core/GrainMethodInvoker.cs
jdom: As we talked in person, I meant some more coverage with the filters throwing themselves too, and whether that's captured and inserted in the context, or it just bubbles up the async delegate chain. Looks like the exceptions bubble up through the chain, and there is no `context.Exception` like in MVC filters. I'm fine to not imitate this as long as we have a small note in our documentation mentioning this difference.
I think I'm with you that this extra property in MVC instead of just bubbling the exception might have been a side effect of the original synchronous filters that did not work or were composed with async constructs. Even if there is a different valid reason, then we could evolve this approach to catch the exception and add the extra property, but I'm fine with what we have for now, as that would be an additive change.

ReubenBond: How's this?
```C#
// Grain code
        public Task FilterThrows() => Task.CompletedTask;

// Filter code
            if (methodInfo.Name == nameof(FilterThrows))
            {
                throw new MyDomainSpecificException("Filter THROW!");
            }

// Test
        public async Task GrainCallFilter_FilterThrows_Test()
        {
            var grain = this.fixture.GrainFactory.GetGrain<IMethodInterceptionGrain>(random.Next());
            
            var exception = await Assert.ThrowsAsync<MethodInterceptionGrain.MyDomainSpecificException>(() => grain.FilterThrows());
            Assert.NotNull(exception);
            Assert.Equal("Filter THROW!", exception.Message);
        }
```

jdom: It is fine. It is not asserting the difference with MVC itself (that we propagate the exception as is from filter to filter, and from grain to filter), to know if we ever align with them, but I don't really think it's much worth it either.
This test does assert the most important aspect, which is that we eventually bubble up the exception to the grain client/caller, regardless of how it was handled by the filtering logic.

ReubenBond:
Fixed my silly throw expressions which broke CI

jdom:
I'll wait a couple of hours in case there is more feedback, otherwise I'll merge it with the current approach and naming

richorama:
I like the approach. 👍

richorama:
I assume I can register an interceptor from a bootstrap provider?

ReubenBond:
@richorama no, once the container is baked you can no longer register interceptors.

*However*, you could register some custom interceptor which has the ability to be modified at runtime in the container and then later manipulate it. Does that make sense?

richorama:
@ReubenBond yeah, I can work around that. In my case I need to grab the Task Scheduler in the bootstrap provider, and pass it over to the interceptor.

PR #3090, opened 2017-06-05
Mock out eventhub from EventDataGeneratorStreamProvider

xiazen:
So that we don't need to configure anything eventhub specific to use EventDataGeneratorStreamProvider.

### DIFF
--- a/src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
+++ b/src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
@@ -115,7 +115,7 @@ namespace Orleans.ServiceBus.Providers.Testing
             /// <param name="svcProvider"></param>
             public override void Init(IProviderConfiguration providerCfg, string providerName, Logger log, IServiceProvider svcProvider)
             {
-                this.CheckpointerFactory = partition => NoOpCheckpointer.Create();
+                this.CheckpointerFactory = partition => Task.FromResult<IStreamQueueCheckpointer<string>>(NoOpCheckpointer.Instance);
                 this.EventHubReceiverFactory = this.EHGeneratorReceiverFactory;
                 this.ehGeneratorSettings = new EventHubGeneratorStreamProviderSettings(providerName);
                 this.ehGeneratorSettings.PopulateDataGeneratingConfigFromProviderConfig(providerCfg);
--- a/src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
+++ b/src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
@@ -52,34 +52,18 @@ namespace Orleans.ServiceBus.Providers.Testing
         }
 
         /// <summary>
-        /// Configure the start of eventhub partition range. EventDataGeneratorStreamProvider would generate the same set of partitions by filling in numbers from range
-        /// start to range end, when initializing. For example, if rangeStart at 1 and rangeEnd at 5, the generated partitions will be  partition-1, partition-2, partition-3, 
-        /// partition-4, partiton-5
+        /// Configure eventhub partition count wanted. EventDataGeneratorStreamProvider would generate the same set of partitions based on the count, when initializing. 
+        /// For example, if parition count set at 5, the generated partitions will be  partition-0, partition-1, partition-2, partition-3, partiton-4
         /// </summary>
-        public int EventHubPartitionRangeStart = DefaultEventHubPartitionRangeStart;
+        public int EventHubPartitionCount = DefaultEventHubPartitionCount;
         /// <summary>
         /// Default EventHubPartitionRangeStart
         /// </summary>
-        public const int DefaultEventHubPartitionRangeStart = 1;
+        public const int DefaultEventHubPartitionCount = 4;
         /// <summary>
         /// EventHubPartitionRangeStartName
         /// </summary>
-        public const string EventHubPartitionRangeStartName = nameof(EventHubPartitionRangeStart);
-
-        /// <summary>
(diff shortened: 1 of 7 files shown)


src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -305,7 +309,11 @@ private EventHubAdapterReceiver MakeReceiver(QueueId queueId)
                 this.EventHubReceiverFactory);
         }
 
-        private async Task<string[]> GetPartitionIdsAsync()
+        /// <summary>
+        /// Get partition Ids from eventhub
+        /// </summary>
+        /// <returns></returns>
+        protected virtual async Task<string[]> GetPartitionIdsAsync()
```
xiazen: Need to either make GetPartitionIdsAsync virtual or CreateAdapter virtual so that EventDataGeneratorStreamProvider won't connect event hub and ask for partition info, but instead use our mocked partition ids.

src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
```diff
@@ -64,6 +99,17 @@ public void PopulateDataGeneratingConfigFromProviderConfig(IProviderConfiguratio
         public void WriteDataGeneratingConfig(Dictionary<string, string> properties)
         {
             properties.Add(StreamDataGeneratorTypeName, this.StreamDataGeneratorType.AssemblyQualifiedName);
+            properties.Add(EventHubPartitionRangeStartName, this.EventHubPartitionRangeStart.ToString());
+            properties.Add(EventHubPartitionRangeEndName, this.EventHubPartitionRangeEnd.ToString());
+        }
+
+        public static string[] GenerateEventHubPartitions(int rangeStart, int rangeEnd)
```
jason-bragg: Unclear why we need this level of control.  Isn't it sufficient to simply declare how many partitions we want, rather than having a start and end?

src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
```diff
@@ -64,6 +99,17 @@ public void PopulateDataGeneratingConfigFromProviderConfig(IProviderConfiguratio
         public void WriteDataGeneratingConfig(Dictionary<string, string> properties)
         {
             properties.Add(StreamDataGeneratorTypeName, this.StreamDataGeneratorType.AssemblyQualifiedName);
+            properties.Add(EventHubPartitionRangeStartName, this.EventHubPartitionRangeStart.ToString());
```
jason-bragg: ensure start <= end?

src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/NoOpCheckpointer.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// So a checkpointer with no ops should be enough.
+    /// </summary>
+    public class NoOpCheckpointer : IStreamQueueCheckpointer<string>
+    {
+        public static Task<IStreamQueueCheckpointer<string>> Create()
+        {
+            return Task.FromResult<IStreamQueueCheckpointer<string>>(new NoOpCheckpointer());
+        }
+
+        public bool CheckpointExists => false;
```
jason-bragg: Should probably return true here, as false will trigger system to try to diligently create a checkpoint.

src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
xiazen: yeah we can have that.

src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
xiazen: let's just do partition amount

src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
```diff
@@ -87,12 +115,28 @@ public class AdapterFactory : EventHubAdapterFactory, IControllable
             /// <param name="svcProvider"></param>
             public override void Init(IProviderConfiguration providerCfg, string providerName, Logger log, IServiceProvider svcProvider)
             {
+                this.CheckpointerFactory = partition => NoOpCheckpointer.Create();
```
jason-bragg: Since NoOpCheckpointer is stateless, we should be able to use the same instance for all.  NoOpCheckpointer.Instance rather than NoOpCheckpointer.Create()?

test/ServiceBus.Tests/SlowConsumingTests/EHSlowConsumingTests.cs
```diff
@@ -109,7 +71,7 @@ public EHSlowConsumingTests(Fixture fixture)
             seed = new Random();
         }
 
-        [SkippableFact, TestCategory("Functional")]
+        [Fact, TestCategory("Functional")]
```
jason-bragg: Tests can now be BVT, right?  No more external dependencies?

xiazen: this test can be 30 sec long. a bit too long for BVT. and we are running both functionals and bvt in Jenkins now.

src/OrleansServiceBus/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorStreamProvider.cs
xiazen: will do

PR #3092, opened 2017-06-06
Correct ClusterFixture's Dispose() method in unit testing tutorial

wattsm:
As per the comments on #3028 fix the `Dispose()` method on `ClusterFixture` so that it calls `StopAllSilos()`.


jdom:
Thanks again!

PR #3096, opened 2017-06-06
Fix client side EHStreamProvider init

xiazen:
GlobalConfig and NodeConfig is not available in client side. So those two shouldn't be required.


jason-bragg:
Was this causing any issues?  Create cache factory should never be called on the client, so I'd think the current code should be fine?  I ask because when these types are needed, they -are- required..?

xiazen:
yes the function CreateCacheFactory itself is called when init. Remember we wrap all the Cache related customization into EventHubQueueCacheFactory, and CreateCacheFactory is called to create the EventHubQueueCacheFactory

jason-bragg:
Ah, I see.
My preference would be to not initialize things on the client that are not needed their, rather than run the logic without the necessary data.
I don't see an easy way of doing that off the top of my head, but some effort on this now will likely save us some bugs down the line.

xiazen:
I will look into it.

jdom:
Ping. Is there any progress with this? Since this is currently broken. I'd prefer to just merge this PR if no progress will be made very soon

PR #3103, opened 2017-06-07
Improve cache purge logic to avoid bad behavior in massive delivery

xiazen:
scenario when MessageAge related settings misconfigured


xiazen:
Not for review yet. I want to sent it to load test first.

PR #3107, opened 2017-06-08
Redact more connection strings when logging

jdom:
Mask DynamoDB and ServiceBus connection strings, as well as SAS tokens in Azure Storage.

### DIFF
--- a/src/Orleans/Configuration/ConfigUtilities.cs
+++ b/src/Orleans/Configuration/ConfigUtilities.cs
@@ -525,48 +525,36 @@ namespace Orleans.Runtime.Configuration
         /// <summary>
         /// Prints the the DataConnectionString,
         /// without disclosing any credential info
-        /// such as the Azure Storage AccountKey or SqlServer password.
+        /// such as the Azure Storage AccountKey, SqlServer password or AWS SecretKey.
         /// </summary>
-        /// <param name="dataConnectionString">The connection string to print.</param>
+        /// <param name="connectionString">The connection string to print.</param>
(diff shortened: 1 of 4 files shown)


src/Orleans/Configuration/ConfigUtilities.cs
```diff
@@ -525,48 +525,36 @@ internal static string IStatisticsConfigurationToString(IStatisticsConfiguration
...
-            }
-            var sqlConnectionInfo = sqlConnectionString;
-            // Remove any Azure account keys from connection string info written to log files
-            int keyPos = sqlConnectionInfo.LastIndexOf("Password=", StringComparison.OrdinalIgnoreCase);
-            if (keyPos > 0)
+            // Remove any secret keys from connection string info written to log files
+            foreach (var secretKey in secretKeys)
             {
-                sqlConnectionInfo = sqlConnectionInfo.Remove(keyPos) + "Password=<--SNIP-->";
+                int keyPos = connectionInfo.IndexOf(secretKey, StringComparison.OrdinalIgnoreCase);
```
sergeybykov: Why not `.Contains(secretKey)` instead of `IndexOf`?

jdom: because I'm using the index value later on to truncate the string. BTW, I didn't change the logic compared to the previous approach, other than I also validate if the index is `>= 0` (it was just `> 0` before, so a secret could escape if it was the first thing in the connection string)

jdom:

jdom:

Issue #3094, opened 2017-06-06
Control over how Orleans behaves on CTRL+C


wattsm:
The [Orleans Silo listens for CTRL+C](https://github.com/dotnet/orleans/blob/master/src/OrleansRuntime/Silo/Silo.cs#L526) and starts a fast exit when it detects it. This can cause problems if you use something like TopShelf to host a silo and are running it as a console application.

For example, if you configure TopShelf to gracefully shutdown the silo on exit:

```c#
service.WhenStopped(host => {

  host.ShutdownOrleansSilo();
  GC.SuppressFinalize(host);

});
```

When running as a Windows service this code will behave as you'd expect. However, if you run this as a console application then when you press CTRL+C to stop TopShelf Orleans will also respond to the event.

This causes intermittent exceptions on shutdown, which sometimes can leave the console application in a state which requires Task Manager / Visual Studio to kill it. I believe this is caused by the call to `host.ShutdownOrleansSilo()` happening after the silo begins its own fast exit process.

Example logs:

```
[2017-06-06 14:28:46.579 GMT    12	WARNING	100502	Catalog	127.0.0.1:11111]	UnregisterManyAsync 1 failed.	
Exc level 0: System.InvalidOperationException: Grain directory is stopping
   at Orleans.Runtime.GrainDirectory.LocalGrainDirectory.CheckIfShouldForward(GrainId grainId, Int32 hopCount, String operationDescription)
   at Orleans.Runtime.GrainDirectory.LocalGrainDirectory.UnregisterOrPutInForwardList(IEnumerable`1 addresses, UnregistrationCause cause, Int32 hopCount, Dictionary`2& forward, List`1 tasks, String context)
   at Orleans.Runtime.GrainDirectory.LocalGrainDirectory.<UnregisterManyAsync>d__108.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Orleans.Runtime.Scheduler.SchedulerExtensions.<>c__DisplayClass1_0.<<QueueTask>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #3109, opened 2017-06-08
Add flag to disable FastKill on CTRL-C

benjaminpetit:
Fix for #3094

### DIFF
--- a/src/Orleans/Configuration/GlobalConfiguration.cs
+++ b/src/Orleans/Configuration/GlobalConfiguration.cs
@@ -461,6 +461,8 @@ namespace Orleans.Runtime.Configuration
 
         public bool AssumeHomogenousSilosForTesting { get; set; }
 
+        public bool FastKillOnCancelKeyPress { get; set; }
+
         /// <summary>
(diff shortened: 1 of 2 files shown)


ReubenBond:
I believe we should just get out of the Ctrl+C business altogether, but especially in `Silo.cs` - what are the advantages for it?

sergeybykov:
I believe the original reason to intercept Ctrl+C was to speed up cluster reconfiguration by having the silo going down mark itself dead in the membership table, so that there is less of a delay and fewer failing requests while failure of the node is being detected by the membership protocol. I'm not sure how frequently this happens in production deployments, but it is useful during development when deployment ID is reused between multiple test runs.

sergeybykov:

jdom:
I'm with Reuben, that this is not a concern of the Silo itself. Perhaps of SiloHost or some Console-specific host. But I'm OK with the change for now, but it should definitely be moved out once we move to Silo Builder.

bboyle1234:
https://github.com/dotnet/orleans/issues/4151

PR #3114, opened 2017-06-10
Use Server GC and other improvements in Azure sample app

jdom:
- Use Server GC in Worker Role process.
- Validate Azure storage configuration in WorkerRole.OnStart.
- Use programmatic config to configure a new provider.
- Update dependency to Orleans 1.4.2


Samples/AzureWebSample/OrleansAzureSilos/OrleansAzureSilos.csproj
```diff
@@ -145,7 +140,13 @@
     <None Include="app.config">
       <SubType>Designer</SubType>
     </None>
+    <Content Include="GCSettingsManagement.ps1">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
attilah: nit: We usually use the PreserveNewest option everywhere, just in case.

PR #3119, opened 2017-06-12
 Fast path for message addressing

dVakulen:
Multiple awaits on target selection path can be quite costly. These awaits can be avoided by firstly trying to lookup locally known directory information for a target grain. Assuming that hit ratio of local lookups is usually >90% the expected throguhput increase is ~5%.

If this is acceptable approach - the todo will be addressed after performance testing.


dVakulen:
Performance tests conducted by @ReubenBond showed the following throughput changes:

NightlyLoadTest 0%
ActivationCollectorStressTest	 +11%
PingLoadTest_LocalReentrant	 +10%
PingLoadTest_RandomReentrant_MultiSilos +25%

So, repeating the question from above - is such approach is acceptable (Try to fast lookup in local directory cache and 
use placement director's synchronous activation select in case of success) ?

сс @sergeybykov , @gabikliot 



Thanks,
Dmytro

gabikliot:
The results make sense. Multi silo test has the most cross silo messaging.
I will look at the cl tonight and reply here.

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -636,9 +637,32 @@ private bool ShouldLogError(Exception ex)
...
+        private bool TryAddressMessageFast(Message message)
+        {
+            var targetAddress = message.TargetAddress;
+            if (targetAddress.IsComplete) return true;
+
+            AddressesAndTag addressesAndTag;
+
+            if (catalog.FastLookup(message.TargetGrain, out addressesAndTag))
+            {
+                // todo: consolidate with placement directors
```
gabikliot: yes, you will need to consolidate it with placement. 
IPlacementRuntime has a method FastLookup.
you also need to carefully look at AddressMessage and make sure you use the same logic: for strategy, request and target.
and use SetTargetPlacement.

gabikliot:
I like this direction.

Inn general, we want to minimize/eliminate usages of awaits and Tasks from the hot paths.
It is still OK (and preferable) to use Tasks in the control plane logic: all the system targets, ..
But for data plane path - messaging stack, scheduler (implementation of the scheduler), serialization, we want to avoid Tasks.
I remember long time ago I counted 4 Tasks on the messaging path. Ideally, it would be zero. Of course, need to balance with added code complexity.
In some places maybe Tasks can be replaced with CW if not fully eliminated.

Again, its important not to go on a crusade to eliminate all Tasks from the runtime. This will be totally unnecessary and counter productive. But eliminating them from a small number of the strategically chosen place will be an important improvement I think.

dVakulen:
Thank you for the review, Gabriel, will address the todo soon.

dVakulen:
The single test that sometimes fails is ClientIdPartitionDataRebuildTests.ReconstructClientIdPartitionTest_Request introduced in https://github.com/dotnet/orleans/pull/2587. while I'm looking for the fix, this PR is ready for review.

dVakulen:
The original names were returned back.

ReubenBond:
The Lookup method looks like a prime candidate for ValueTask<T>

src/OrleansRuntime/Placement/StatelessWorkerDirector.cs
```diff
@@ -12,6 +12,26 @@ internal class StatelessWorkerDirector : IPlacementDirector<StatelessWorkerPlace
...
+            return Task.FromResult(SelectActivationCore(strategy, target, context));
+        }
+
+        public bool TrySelectActivationSynchronously(
+            PlacementStrategy strategy, GrainId target, IPlacementRuntime context, out PlacementResult placementResult)
+        {
+            placementResult = SelectActivationCore(strategy, target, context);
+            // stateless workers are always local, and  request for creation of a new activation will be created by PlacementDirectorsManager 
+            // in case of null returned by SelectActivationCore  
+            return true;
```
gabikliot: should not you return false if placementResult is null?

gabikliot: Looked more carefully. I think you need to return false on null. 
SelectOrAddActivation will be triggered by null in the prev code, but in your new code if Try returns true, you never even call SelectOrAddActivation.

maybe its better to change Try to return the placement and not bool, to be similar to async select. it will make this kind of reasoning easier?

gabikliot: if I am correct here and it should have indeed been return false on null, then I am very worried that we dont have any unit test about stateless workers that failed on this change. Can you please think about a test which would have failed in this case (if I am indeed correct about the bug). Without such a test, I don't feel comfortable abut this change.

dVakulen:
In reply to https://github.com/dotnet/orleans/pull/3119#discussion_r122999232 The result of TrySelectActivationSynchronously are being checked for both sucess and null here -
 https://github.com/dotnet/orleans/pull/3119/files#diff-d5b7c57a0099011410228c91429924afR696 . I just wanted to preserve semantics of previous version, but agreed that it's better to return false in case of null result.

src/OrleansRuntime/Placement/StatelessWorkerDirector.cs
gabikliot: OK about Try with bool. Lets not do any more changes about OnSelectActivation now, in this CL.

gabikliot:
The code looks good to me now.
The extra improvement in the multi silo ping test of 25% justifies a little bit of the extra complexity in the placement directors I think.

If others don't have comments, this can be merged. After all tests pass of course.

sergeybykov:

src/OrleansRuntime/Placement/IActivationSelector.cs
```diff
@@ -9,10 +10,13 @@ internal interface IActivationSelector
     {
         Task<PlacementResult> OnSelectActivation(
             PlacementStrategy strategy, GrainId target, IPlacementRuntime context);
+
+        bool TrySelectActivationSynchronously(
```
ReubenBond: Could this be made a single method with a return value of `ValueTask<PlacementResult>` instead of two methods?

dVakulen: It'll require adding System.Threading.Tasks.Extensions package, which most likely would be out of scope of this PR. Plus, the problem is not the return type of ```Task``` but ```await```s in method body which causes allocation of state machine with its costs.

dVakulen:
The tests has passed without any change, but ReconstructClientIdPartitionTest_Request test may still fail sometimes due to race between refresh of the clientId entries that are connected to the gateway by ```SiloStatusChangeNotification``` and breaking outstanding promises to stopped silo. Probably it could be marked as skippable.

dVakulen:
No futher changes are currently meant to be done in this PR, so I think it can be merged after scheduled performance test completion

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -597,32 +598,60 @@ private void ResendMessageImpl(Message message, ActivationAddress forwardingAddr
...
                 {
-                    logger.Error(ErrorCode.Dispatcher_SelectTarget_Failed,
-                        String.Format("SelectTarget failed with {0}", ex.Message),
-                        ex);
+                    logger.Error(ErrorCode.Dispatcher_SelectTarget_Failed, $"SelectTarget failed with {ex.Message}", ex);
                 }
+
                 MessagingProcessingStatisticsGroup.OnDispatcherMessageProcessedError(message, "SelectTarget failed");
                 RejectMessage(message, Message.RejectionTypes.Unrecoverable, ex);
+            };
```
jdom: What's the purpose of allocating this delegate with closures for every call? Can't you just define an actual method that you call from both places where failures occur, but doesn't allocate unnecessarily?

src/OrleansRuntime/Core/Dispatcher.cs
```diff
@@ -597,32 +598,60 @@ private void ResendMessageImpl(Message message, ActivationAddress forwardingAddr
...
+
+            Func<Task, Task> transportMessageAfterAddressing = async addressMessageTask =>
```
jdom: similar with this one, it's probably better to define this method and avoid allocating closures. This one bothers me a little less than the previous though, since it is always called, where as the other one allocates and will not even be used in the most common cases

dVakulen: Initially they were inline methods, but it's not yet supported in Orleans, so converted to ordinary lambdas. While they will allocate on each call - that objects will end up collected in gen 0,  quite unlikely will affect benchmarks\real world usages, and I do believe, that such kind of methods is better from support perspective, as they do not pollute the internal api of class. And the codebase maybe will be upgraded to support such methods before next release.

ReubenBond: If we could upgrade to C# 7.0 then we could make use of local functions as well as ValueTask. Do you know what we need to do in order to support that, @jdom? I think the .NET Standard build already supports the newer features.

src/OrleansRuntime/Placement/IActivationSelector.cs
ReubenBond: In that case, you can have a single method on the API and implement it with a fast-path branch returning synchronously and an async branch which calls an async method and wraps its Task into a ValueTask. If we raise the C# level to 7.0 then we can use a local function for that or make use of async for Task-like types.

src/OrleansRuntime/Core/Dispatcher.cs
jdom: I think that if we completely rule out incorporating this improvement in 1.5 final (most likely anyway), then it would be fine to do that.
Still, I wouldn't hold off from merging this because of that, it could be done separately.

gabikliot:
are we going to merge that? do we have new performance results?

sergeybykov:
We need to perf-test this before merging. We were a little bit backed up with finishing 1.5.0 and revamping of the build/test process.

sergeybykov:
@dotnet-bot test netstandard-win-bvt,netfx-functional,netstandard-win-functional

sergeybykov:
@dotnet-bot test netfx-functional netstandard-win-functional

sergeybykov:

dVakulen:
Also currently this branch contains first version of PR, in order to confirm initial perf results.

sergeybykov:

sergeybykov:

dVakulen:
For some reason results of initial performance testing couldn't be reproduced. Now multiple runs of testing on this branch shows up to 7% increase on NightlyLoadTest ( 190447 vs 176772 on master)  8% on ActivationCollectorStressTest (208096 vs 191245 on master), 5% on PingLoadTest_LocalReentrant and 0% on PingLoadTest_RandomReentrant_MultiSilos. Still seems pretty decent, and possibly justifying additional code complexity.

dVakulen:
Currently branch at state of https://github.com/dotnet/orleans/pull/3119#issuecomment-309986027 comment.

sergeybykov:
>Still seems pretty decent, and possibly justifying additional code complexity.

I agree with that, unless there's a strong pushback on the complexity.

gabikliot:
I support that too (accepting the PR).

sergeybykov:
I ran a few more perf tests, and they look positive on ping test and neutral on load test. Hence, I'll merge it.

Thank you, @dVakulen!

dVakulen:
Thank you, @gabikliot for the review.

gabikliot:
Thank you!

PR #3125, opened 2017-06-14
configure .net garbage collection for .net core

stephenlautier:
- update with .NET Core config
- fix a grammer issue


sergeybykov:
Thank you, @stephenlautier!

PR #3127, opened 2017-06-14
Upgrade WindowsAzure.ServiceBus to 4.1.0

### LINKED ISSUE #2910: Update to WindowsAzure.ServiceBus 4.0.0
Any issues if OrleansServiceBus is updated for WindowsAzure.ServiceBus 4.0.0? If that's fine, I could prepare a PR, though it is as trivial as changing the dependency version.

### LINKED ISSUE #3062: EventHub provider and latest version of Microsoft Azure Service Bus  v4.1.1 (WindowsAzure.ServiceBus nuget)
Just want to flag this issue for the future release or anyone who stepped on the same problem as we did and make error text search-able.

`Microsoft.Orleans.OrleansServiceBus` v1.4.1 is not compartible with `WindowsAzure.ServiceBus` major version upgrade v4. (v.4.1.1 as per today)
By definition `Microsoft.Orleans.OrleansServiceBus` 1.4.1 depends on `WindowsAzure.ServiceBus`  v3.3.2.

It's safe to upgrade dependance to the latest v3 ( v3.4.6 as per today), but if you have to upgrade `WindowsAzure.ServiceBus`  to a major v4 version - it breaks EH Stream provider and you'll see this error in your logs:

```
WaWorkerHost.exe Error: 0 :
[2017-05-29 04:39:01.833 GMT    39    ERROR      103303    PullingAgentSystemTarget/FF/c8a2d58b-BullclipEventHub    127.0.0.1:11111]    !!!!!!!!!! QueueAdapterReceiver bullclipeventhub-0-0x00000000 failed to Initialize.    
Exc level 0: System.MissingMethodException: Method not found: 'System.Threading.Tasks.Task`1<Microsoft.ServiceBus.Messaging.EventHubReceiver> Microsoft.ServiceBus.Messaging.EventHubConsumerGroup.CreateReceiverAsync(System.String, System.String, Boolean)'.
   at Orleans.ServiceBus.Providers.EventHubAdapterReceiver.<CreateReceiver>d__26.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Orleans.ServiceBus.Providers.EventHubAdapterReceiver.CreateReceiver(EventHubPartitionSettings partitionSettings, String offset, Logger logger)
   at Orleans.ServiceBus.Providers.EventHubAdapterReceiver.<Initialize>d__18.MoveNext()
   at Orleans.OrleansTaskExtentions.<SafeExecute>d__1.MoveNext()
   at Orleans.OrleansTaskExtentions.<LogException>d__0.MoveNext() (View on Seq)
```

Simple fix for this is to downgrade back to major v3 version.

jdom:
Fixes #2910
Fixes #3062

### DIFF
--- a/src/NuGet/Microsoft.Orleans.OrleansServiceBus.nuspec
+++ b/src/NuGet/Microsoft.Orleans.OrleansServiceBus.nuspec
@@ -22,7 +22,7 @@
       <dependency id="Microsoft.Orleans.OrleansRuntime" version="$version$" />
       <dependency id="Microsoft.Orleans.OrleansProviders" version="$version$" />
       <dependency id="Microsoft.Orleans.OrleansAzureUtils" version="$version$" />
-      <dependency id="WindowsAzure.ServiceBus" version="3.3.2" />
+      <dependency id="WindowsAzure.ServiceBus" version="4.1.0" />
     </dependencies>
   </metadata>
   <files>
--- a/src/OrleansServiceBus/OrleansServiceBus.2017.csproj
+++ b/src/OrleansServiceBus/OrleansServiceBus.2017.csproj
@@ -15,7 +15,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.WindowsAzure.ConfigurationManager" Version="3.1.0" />
-    <PackageReference Include="WindowsAzure.ServiceBus" Version="3.3.2" />
+    <PackageReference Include="WindowsAzure.ServiceBus" Version="4.1.0" />
     <PackageReference Include="WindowsAzure.Storage" Version="7.0.0" />
   </ItemGroup>
   <ItemGroup>
--- a/src/OrleansServiceBus/project.json
+++ b/src/OrleansServiceBus/project.json
@@ -1,7 +1,7 @@
 ﻿{
   "dependencies": {
     "Microsoft.WindowsAzure.ConfigurationManager": "3.1.0",
-    "WindowsAzure.ServiceBus": "3.3.2",
+    "WindowsAzure.ServiceBus": "4.1.0",
     "WindowsAzure.Storage": "7.0.0"
   },
   "frameworks": {
--- a/test/ServiceBus.Tests/ServiceBus.Tests.csproj
+++ b/test/ServiceBus.Tests/ServiceBus.Tests.csproj
@@ -119,7 +119,9 @@
   <ItemGroup>
     <None Include="App.config" />
     <None Include="project.json" />
-    <None Include="ServiceBus.Tests.xunit.runner.json" />
+    <None Include="ServiceBus.Tests.xunit.runner.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
   </ItemGroup>
   <ItemGroup>
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />


jdom:
`WindowsAzure.ServiceBus` nuget package is supported in .NET Framework only. Currently we are on 3.3.2, and the latest is 4.1.2
The vNext (2.0) solution is referencing the `Windows.Azure.EventHubs` package, which is expected to supersede the other ServiceBus library (notice that it’s only for EventHub, and not every other service in the ServiceBus umbrella, and I’m not entirely sure the service behavior is 100% back compat yet, I didn't go through that review for this PR). This version is 1.0.0 and it’s .NET Standard compatible.

To keep it safe, we decided with @jason-bragg to upgrade the `WindowsAzure.ServiceBus` package in 1.5, and not switch to the new .NET Standard library yet as we were forced to do in vNext/2.0.
From looking at the changes, it looks like upgrading to 4.1.0 sounds reasonable. I did the upgrade and every EventHub functional tests seem to be working. I’m not taking advantage of any of the new features, you can leverage them later on if they make sense.

Here is the changelog since 3.3.2 (I had to gather it from each individual package release notes, so I’m pasting here a consolidated changelog):

```
WindowsAzure.ServiceBus changelog

4.1.2
• EventHub/IoTHub: modified URL suffix check for IoTHub specific environments.

4.1.1
• EventProcessorHost: fixed an issue that ReceiverRuntimeInfo in the PartitionContext is not updated correctly.

4.1.0
• EventHubs: fixed a race which can cause a stuck on receive call. 
• EventHubs: fixed an issue which can result in timeout in send calls intermittently.

4.0.0
• Major version update - this release contains breaking changes in CreateReceiver API (Event Hubs only). 
• EventHubs: add a performance counter for AMQP redirected connections. 
• EventHubs: support specifying an identifier of a receiver when creating the receiver. 
• EventHubs: EventHubReceiver.ReceiverRuntimeInfo property - provides receiver related runtime information which includes last sequence number, offset, and enqueued time. To enable the feature, set the value of EventHubConsumerGroup.EnableReceiverRuntimeMetric property to true before creating a receiver. 
• EventHubs: EventDataBatch class: a helper class for creating a batch of EventData objects for SendBatch/SendBatchAsync API. 
• Relay: fix a timeout propagation issue. 
• Relay: fix a regression where opening a listener deadlocks when called on a thread with ambient SynchronizationContext.

3.4.6
• General: fix web socket connect timeout issue. 
• EventHubs: fix null reference exception which can happen during receive calls if a client is closed. 
• EventHubs: improve tracing for receiver redirection scenario.

3.4.5
• General: fix product information in dll Properties. 
• Relay: If the Listener's control channel gets disconnected 3+ times in 10 minutes or less enable heartbeats every 105 seconds to keep the channel from being considered idle. 
• Relay: Fixes an issue where a Listener could stop reconnecting without Faulting or raising the Offline event. 
• Relay: WebSocket and WebStream listener rendezvous using DNS entries instead of Direct IP addresses.

3.4.3
• EventHub/Messaging: fix thread blocking issue which can happen if connectivity mode is set to auto. 
• EventHub/Messaging: fix auto-redirection fallback mechanism which can cause connectivity issue. 
• EventHub/Messaging: AMQP SSL protocols - enable TLS 1.1/1.2 
• EventHub/Messaging: fix a transient error  which can happen while calling runtime operations. 
• Messaging: fix a timeout/transient error when calling AcceptSession API. 
• Messaging: added client telemetry (product/.NET framework/platform version).

3.4.2
• EventHub/Messaging: fix a bug in the client side performance counter where a connection redirect can cause a change in port number which will cause counter instances to have the wrong endpoint format as keys. 
• EventHub/Messaging: Add app.Config switch for toggling entity level performance counters. To toggle the switch without code, ensure the key "Microsoft.ServiceBus.DisableEntityLevelPerformanceCounters" is specified in the application config: 
    <?xml version="1.0" encoding="utf-8" ?> 
    <configuration> 
      <appSettings> 
        <add key="Microsoft.ServiceBus.DisableEntityLevelPerformanceCounters" value="true"/> 
      </appSettings> 
    </configuration>
                
3.4.1
Messaging: fix a bug in the netmessaging protocol where an invalid Sas token can sometimes cause a message sender to leak memory from connection handle, even after sender is closed. (This does not affect Amqp protocol)

3.4.0
• EventHub/Messaging: fixed a bug where some of the Amqp session error are not mapped to MessagingCommunicationException. These are now returned as MessagingCommunicationException.
• EventHub: Add ability to receive from EndOfStream. You can now create receiver using EventHubConsumerGroup.EndOfStream as the offset and the receiver will start receiving from the current last offset (at the time of receiver creation). 
• EventHub: Fixed a bug where max link credit may not have been set to the correct upper bound in some cases when user use the PrefetchSizeInBytes feature.
```

PR #3128, opened 2017-06-14
Cleanup stream producers on dead or missing silos.

jason-bragg:
Catch rejected messages to producers and remove the producer if those rejections come from dead, dying, or missing silos.

This should address #3056?

### DIFF
--- a/src/Orleans/Runtime/SiloStatus.cs
+++ b/src/Orleans/Runtime/SiloStatus.cs
@@ -41,5 +41,13 @@
         {
             return siloStatus == SiloStatus.ShuttingDown || siloStatus == SiloStatus.Stopping || siloStatus == SiloStatus.Dead;
         }
+
+        /// <summary>
+        /// Return true if silo is terminating or missing.
+        /// </summary>
+        public static bool IsUnavailable(this SiloStatus siloStatus)
+        {
+            return siloStatus.IsTerminating() || siloStatus == SiloStatus.None;
+        }
     }
 }
--- a/src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
+++ b/src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
@@ -160,7 +160,7 @@ namespace Orleans.Streams
             {
                 foreach (PubSubPublisherState producerState in producers)
                 {
-                    tasks.Add(HandleProducerErrors(producerState, producerState.Producer.AddSubscriber(subscriptionId, streamId, streamConsumer, filter)));
+                    tasks.Add(ExecuteProducerTask(producerState, producerState.Producer.AddSubscriber(subscriptionId, streamId, streamConsumer, filter)));
                 }
 
                 Exception exception = null;
@@ -370,7 +370,7 @@ namespace Orleans.Streams
 
                 // Notify producers about unregistered consumer.
                 List<Task> tasks = State.Producers
-                    .Select(producerState => HandleProducerErrors(producerState, producerState.Producer.RemoveSubscriber(subscriptionId, streamId)))
+                    .Select(producerState => ExecuteProducerTask(producerState, producerState.Producer.RemoveSubscriber(subscriptionId, streamId)))
                     .ToList();
                 await Task.WhenAll(tasks);
                 //if producers got removed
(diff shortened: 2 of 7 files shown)


ReubenBond:
Does this fix #3056?

src/Orleans/Runtime/SiloStatus.cs
```diff
@@ -41,5 +41,13 @@ public static bool IsTerminating(this SiloStatus siloStatus)
         {
             return siloStatus == SiloStatus.ShuttingDown || siloStatus == SiloStatus.Stopping || siloStatus == SiloStatus.Dead;
         }
+
+        /// <summary>
+        /// Return true if silo is terminating or missing.
+        /// </summary>
+        public static bool IsUnavailable(this SiloStatus siloStatus)
+        {
+            return siloStatus.IsTerminating() || siloStatus == SiloStatus.None;
```
ReubenBond: So it's fine if this returns true for silos which we have not been informed about via membership yet?

jason-bragg:
@ReubenBond, Yes, this should address #3056?
Will add to description.

src/Orleans/Runtime/SiloStatus.cs
jason-bragg: In a general sense, I think that's ok.
In this specific case it should be.
This check is only made if we can't talk to the other silo.  If it is there but just not in our membership service yet, then we should be able to talk to it.

src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
```diff
@@ -442,21 +369,35 @@ public Task<List<StreamSubscription>> GetAllSubscriptions(StreamId streamId, ISt
...
+        {
+            if (State.Producers.Count == 0 && State.Consumers.Count == 0) // + we already know that numProducers == 0 from previous if-clause
+            {
+                await ClearStateAsync(); //State contains no producers or consumers, remove it from storage
+                return true;
+            }
+            return false;
+        }
+
+        private async Task HandleProducerErrors(PubSubPublisherState producer, Task producerTask)
```
xiazen: ExecuteProducerTask maybe a better name.

src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
```diff
@@ -466,20 +407,19 @@ public Task<List<StreamSubscription>> GetAllSubscriptions(StreamId streamId, ISt
...
-        }
-
-        /// <summary>
-        /// Try clear state will only clear the state if there are no producers or consumers.
-        /// </summary>
-        /// <returns></returns>
-        private async Task<bool> TryClearState()
-        {
-            if (State.Producers.Count == 0 && State.Consumers.Count == 0) // + we already know that numProducers == 0 from previous if-clause
+            catch (OrleansMessageRejectionException)
```
ReubenBond: Let's ensure that `SiloUnavailableException` is also handled here, since it signifies that a silo was determined to be dead by membership while some request was outstanding.

src/OrleansRuntime/Streams/PubSub/PubSubRendezvousGrain.cs
```diff
@@ -466,20 +407,19 @@ public Task<List<StreamSubscription>> GetAllSubscriptions(StreamId streamId, ISt
...
+                if (grainRef == null || grainRef.GrainId.IsSystemTarget && siloStatusOracle.GetApproximateSiloStatus(grainRef.SystemTargetSilo).IsUnavailable())
+                {
+                    RemoveProducer(producer);
+                }
+                else // otherwise, throw
+                {
+                    throw;
+                }
             }
-            return false;
```
ReubenBond: There's a case where Silo A messages Silo B (who does not know about Silo A yet), Silo A dies. Since Silo B never knew about Silo A, Service Fabric membership will never inform it of A's demise. SF membership works based upon progressive consistent snapshots of cluster state and does not include a history of all silos which ever existed.

Anyway: I believe this is fine because eventually the call will fail in `PrepareMessageForSend` under the `lastConnectionFailure` clause as long as messages happen in quick enough succession.

This is worth testing, perhaps via a test which uses a mocked membership oracle.

xiazen:

xiazen:

jason-bragg:

jason-bragg:

Issue #3129, opened 2017-06-14
Convert IMembershipTableGrain into a SystemTarget


ReubenBond:
Currently the `IMembershipTableGrain` is special in that it is registered by the primary using the special-purpose `Catalog.CreateSystemGrain(...)` method.

This all only exists for the purpose of `IMembershipTableGrain` which is only valid for test scenarios - it's not reliable in the distributed systems sense of the word.

Ideally, we could remove that method and the notion of a system grain.

PR #3130, opened 2017-06-15
Replace DuplicateActivationException with a strongly-typed return value

### LINKED ISSUE #2733: Rename or remove DuplicateActivationException to prevent confusion
`DuplicateActivationException` doesn't mean that a duplicate activation has been created. It is just a means to resolve a race between two or more silos trying concurrently to register activations for the same grain. However, the very name of the exception class and the message of it misleads people to think that they get duplicate activations.

I suggest we rename the exception class to prevent this confusion from happening.

`RegistrationRaceException`? `ActivationAlreadyRegisteredException`? `ExistingActivationException`?

ReubenBond:
This PR resolves #2733.

The fix focuses on the logic in `Catalog.InitActivation(...)` and how it interacts with `Catalog.RegisterActivationInGrainDirectoryAndValidate(...)`. Specifically, it removes the `DuplicateActivationException` thrown by the latter and caught by the former. Doing that required slightly reworking the control flow in `InitActivation` so that exceptions aren't the only means of error handling.

* Move recovery logic in `InitActivation` from `catch` to a new method which is called from the `finally` block.
* Return a new `ActivationRegistrationResult` struct from `RegisterActivationInGrainDirectoryAndValidate` which replaces `DuplicateActivationException`.
* When an activation fails to register, return early from `InitActivation` so that the `finally` block can clean up the activation.

One change worth noting: `InitActivation` will no longer throw when there is a duplicate activation. The activation returned by `GetOrCreateActivation` (which calls into `InitActivation`) will eventually have its `ForwardingAddress` fixed to point to the existing activation. Note that the result of `InitActivation` is only used in one specific case: when creating the `IMembershipTableGrain`. I opened #3129 to discuss converting that into a SystemTarget instead of routing it through the directory.

### DIFF
--- a/src/Orleans/Statistics/StatisticNames.cs
+++ b/src/Orleans/Statistics/StatisticNames.cs
@@ -219,7 +219,7 @@ namespace Orleans.Runtime
         public static readonly StatisticName CATALOG_ACTIVATION_SHUTDOWN_VIA_DIRECT_SHUTDOWN                        = new StatisticName("Catalog.Activation.Shutdown.ViaDirectShutdown");
         public static readonly StatisticName CATALOG_ACTIVATION_SHUTDOWN_VIA_DEACTIVATE_STUCK_ACTIVATION            = new StatisticName("Catalog.Activation.Shutdown.ViaDeactivateStuckActivation");
         public static readonly StatisticName CATALOG_ACTIVATION_NON_EXISTENT_ACTIVATIONS                            = new StatisticName("Catalog.Activation.NonExistentActivations");
-        public static readonly StatisticName CATALOG_ACTIVATION_DUPLICATE_ACTIVATIONS                               = new StatisticName("Catalog.Activation.DuplicateActivations");
+        public static readonly StatisticName CATALOG_ACTIVATION_CONCURRENT_REGISTRATION_ATTEMPTS                    = new StatisticName("Catalog.Activation.ConcurrentRegistrationAttempts");
 
         // Dispatcher
         public static readonly StatisticName DISPATCHER_NEW_PLACEMENT                                               = new StatisticName("Dispatcher.NewPlacement");
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -28,59 +28,6 @@ namespace Orleans.Runtime
 {
     internal class Catalog : SystemTarget, ICatalog, IPlacementRuntime, ISiloStatusListener
     {
-        /// <summary>
-        /// Exception to indicate that the activation would have been a duplicate so messages pending for it should be redirected.
-        /// </summary>
-        [Serializable]
-        internal class DuplicateActivationException : Exception
-        {
-            public ActivationAddress ActivationToUse { get; private set; }
-
-            public SiloAddress PrimaryDirectoryForGrain { get; private set; } // for diagnostics only!
-
-            public DuplicateActivationException() : base("DuplicateActivationException") { }
-            public DuplicateActivationException(string msg) : base(msg) { }
-            public DuplicateActivationException(string message, Exception innerException) : base(message, innerException) { }
-
-            public DuplicateActivationException(ActivationAddress activationToUse)
-                : base("DuplicateActivationException")
-            {
-                ActivationToUse = activationToUse;
-            }
-
-            public DuplicateActivationException(ActivationAddress activationToUse, SiloAddress primaryDirectoryForGrain)
-                : base("DuplicateActivationException")
-            {
-                ActivationToUse = activationToUse;
-                PrimaryDirectoryForGrain = primaryDirectoryForGrain;
-            }
-
-#if !NETSTANDARD
-            // Implementation of exception serialization with custom properties according to:
-            // http://stackoverflow.com/questions/94488/what-is-the-correct-way-to-make-a-custom-net-exception-serializable
-            protected DuplicateActivationException(SerializationInfo info, StreamingContext context)
-                : base(info, context)
-            {
-                if (info != null)
-                {
-                    ActivationToUse = (ActivationAddress) info.GetValue("ActivationToUse", typeof (ActivationAddress));
-                    PrimaryDirectoryForGrain = (SiloAddress) info.GetValue("PrimaryDirectoryForGrain", typeof (SiloAddress));
-                }
-            }
-
-            public override void GetObjectData(SerializationInfo info, StreamingContext context)
-            {
-                if (info != null)
-                {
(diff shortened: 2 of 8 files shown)


src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -567,130 +514,176 @@ private void SetupActivationInstance(ActivationData result, string grainType, st
...
+                if (logger.IsVerbose) logger.Verbose("InitActivation is done: {0}", activation.Address);
             }
             catch (Exception ex)
             {
-                lock (activation)
+                // Store and re-throw the exception.
+                exception = ex;
+                throw;
+            }
+            finally
```
jdom: why is this in the finally block? Why not both in the catch block as well as in the `if (!registrationResult.IsSuccess)` part?

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -567,130 +514,176 @@ private void SetupActivationInstance(ActivationData result, string grainType, st
...
+                switch (initStage)
+                {
+                    case ActivationInitializationStage.Register: // failed to RegisterActivationInGrainDirectory
+
+                        // Failure!! Could it be that this grain uses single activation placement, and there already was an activation?
+                        activation.ForwardingAddress = registrationResult.ExistingActivationAddress;
+                        if (activation.ForwardingAddress != null)
+                        {
+                            CounterStatistic
+                                .FindOrCreate(StatisticNames.CATALOG_ACTIVATION_DUPLICATE_ACTIVATIONS)
```
jdom: Should the `CATALOG_ACTIVATION_DUPLICATE_ACTIVATIONS` also be renamed? it's scary if you see that number go up, but we concluded this is harmless under normal operation, right?

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -567,130 +514,176 @@ private void SetupActivationInstance(ActivationData result, string grainType, st
...
+                                        "{3}" +
+                                        "Full activation address is {4}. We have {5} messages to forward.",
+                                        address,
+                                        activation.ForwardingAddress,
+                                        activation.GrainInstanceType,
+                                        primary != null
+                                            ? "Primary Directory partition for this grain is " + primary + ". "
+                                            : String.Empty,
+                                        address.ToFullString(),
+                                        activation.WaitingCount);
```
jdom: probably interpolating this string will be more readable, right?

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -567,130 +514,176 @@ private void SetupActivationInstance(ActivationData result, string grainType, st
...
-                                logger.Warn(ErrorCode.Runtime_Error_100064,
-                                    String.Format("Failed to RegisterActivationInGrainDirectory for {0}.",
-                                        activation), ex);
-                                // Need to undo the registration we just did earlier
-                                if (activation.IsUsingGrainDirectory)
+                                else if (logger.IsVerbose)
                                 {
-                                    scheduler.RunOrQueueTask(() => directory.UnregisterAsync(address, UnregistrationCause.Force),
-                                        SchedulingContext).Ignore();
+                                    logger.Verbose(ErrorCode.Catalog_DuplicateActivation, logMsg);
```
jdom: no need to check for IsVerbose, since you are not formatting strings or incurring any penalty by just calling `logger.Verbose`  directly, which already checks for IsVerbose

src/OrleansRuntime/Catalog/Catalog.cs
ReubenBond: I prefered to keep it in one place, but I can do both if you prefer.

src/OrleansRuntime/Catalog/Catalog.cs
ReubenBond: This is harmless, yes - maybe it bears renaming

ReubenBond:

src/OrleansRuntime/Catalog/Catalog.cs
ReubenBond: `public static readonly StatisticName CATALOG_ACTIVATION_DUPLICATE_ACTIVATIONS                               = new StatisticName("Catalog.Activation.DuplicateActivations");`

Potential names? `Catalog.Activation.RegistrationRaces`, `Catalog.Activation.ConcurrentRegistrationAttempts`? I prefer the second one, it's precise.

ReubenBond:

src/OrleansRuntime/Catalog/Catalog.cs
ReubenBond: If the latter is chosen, we should add a line to the release log:
> Renamed the `Catalog.Activation.DuplicateActivations` counter to `Catalog.Activation.ConcurrentRegistrationAttempts` to more accurately reflect what it tracks and its benign nature.

src/Orleans/Statistics/StatisticNames.cs
```diff
@@ -219,7 +219,7 @@ internal class StatisticNames
         public static readonly StatisticName CATALOG_ACTIVATION_SHUTDOWN_VIA_DIRECT_SHUTDOWN                        = new StatisticName("Catalog.Activation.Shutdown.ViaDirectShutdown");
         public static readonly StatisticName CATALOG_ACTIVATION_SHUTDOWN_VIA_DEACTIVATE_STUCK_ACTIVATION            = new StatisticName("Catalog.Activation.Shutdown.ViaDeactivateStuckActivation");
         public static readonly StatisticName CATALOG_ACTIVATION_NON_EXISTENT_ACTIVATIONS                            = new StatisticName("Catalog.Activation.NonExistentActivations");
-        public static readonly StatisticName CATALOG_ACTIVATION_DUPLICATE_ACTIVATIONS                               = new StatisticName("Catalog.Activation.DuplicateActivations");
+        public static readonly StatisticName CATALOG_ACTIVATION_CONCURRENT_REGISTRATION_ATTEMPTS                    = new StatisticName("Catalog.Activation.ConcurrentRegistrationAttempts");
```
jdom: @jason-bragg / @sergeybykov any concerns with this rename?
(do we need to put it in release notes?)

ReubenBond: My comment got swallowed by the diff, but:

"
If the latter is chosen, we should add a line to the release log:
> Renamed the `Catalog.Activation.DuplicateActivations` counter to `Catalog.Activation.ConcurrentRegistrationAttempts` to more accurately reflect what it tracks and its benign nature.

"

IMO it's best to mention it

sergeybykov: I think it's fine. It's never been a critical thing to monitor. So I don't expect it to be a problem.

PR #3136, opened 2017-06-16
Expose available versions information in placement context

benjaminpetit:
Now the developer can create custom placement director that take into account the version distribution support in the silos in the cluster.

With this modification we could for example create a placement directors that do not create grain of version x if there is less than 20% of silos that support it, ...


jdom:
Can you add an entry in the non-breaking improvements section of the changelog (after merging with latest) that reads:
`Expose available versions information in placement context #3136`

test/Versions/TestVersionGrains2/TestVersionGrains2.csproj
```diff
@@ -70,6 +73,10 @@
       <Project>{BC1BD60C-E7D8-4452-A21C-290AEC8E2E74}</Project>
       <Name>Orleans</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\TestGrainInterfaces\TestGrainInterfaces.csproj">
+      <Project>{3972213F-189B-41D4-90FE-38D513C1106D}</Project>
+      <Name>TestGrainInterfaces</Name>
+    </ProjectReference>
```
jdom: You'll need to add this to the vNext projects, as well as the ones with the .2017 suffix in the main tree.

benjaminpetit:

jdom:

test/TestGrainInterfaces/VersionAwarePlacementDirector.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        public override int GetHashCode()
+        {
+            return GetType().GetHashCode();
+        }
+    }
+
+    public class VersionAwarePlacementDirector : IPlacementDirector<VersionAwarePlacementStrategy>
+    {
+        private readonly Random random = new Random();
```
ReubenBond: Should this be SafeRandom instead?

benjaminpetit: SafeRamdom is not public. And since it is a director only used in a test, that does not create activation in parallel, I did not reimplemented it

ReubenBond: ah no problem - sounds fine to me. Because our tests are sometimes used as examples, maybe we should just include a comment saying "don't do this"

PR #3140, opened 2017-06-16
Make EHStatisticMonitorTests more predictable

xiazen:
EHStatictiscMonitorTests fail intermittently, because the test is taking a way to trigger cache pressure changing, which depends a lot on timing. Change this way to a more predictable model, which is  directly injecting cache pressure status.


jdom:

xiazen:
ping @jason-bragg

PR #3152, opened 2017-06-21
Make StreamQueueBalancer Pluggable

xiazen:
- Make StreamQueueBalancer pluggable,

- Make it possible for the injected StreamQueueBalancer to be built on top of an async service, such as external storage, or a lease manager.


src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
```diff
@@ -8,27 +8,32 @@ namespace Orleans.Streams
...
     /// notifies any listeners (<code>IStreamQueueBalanceListener</code>) of changes to the distribution of queues.
     /// </summary>
-    internal interface IStreamQueueBalancer
+    public interface IStreamQueueBalancer
     {
         /// <summary>
+        /// Async Initialize logic
+        /// </summary>
+        /// <returns></returns>
+        Task Initialize();
```
xiazen: add hock for async initialize logic, such as load queue balancer from an async storage service, or connect async lease manager for initialize logic. 

The same reason for other methods to be async

src/Orleans/Streams/Providers/IStreamProviderRuntime.cs
```diff
@@ -99,6 +99,9 @@ public class PersistentStreamProviderConfig
         public const string SILO_MATURITY_PERIOD = "SiloMaturityPeriod";
         public static readonly TimeSpan DEFAULT_SILO_MATURITY_PERIOD = TimeSpan.FromMinutes(2);
 
+        //optional param. user need to set this when they use Custom StreamQueueBalancer
+        public const string QUEUE_BALANCER_FACTORY_NAME = nameof(QueueBalancerFactoryName);
+        public string QueueBalancerFactoryName { get; set; }
```
xiazen: User needs to firstly set QueueBalancerFactoryName and set BalancerType to be CustomBalancer, secondly inject their `IStreamQueueBalancerFactory` with the set name to DI to be able to use this feature.

test/ServiceBus.Tests/PlugableQueueBalancerTests/LeaseManagerGrain.cs
```diff
@@ -0,0 +1,106 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Tester.StreamingTests
+{
+    //one lease manager grain per stream provider, so its key is stream provider name
+    public interface ILeaseManagerGrain : IGrainWithStringKey
```
xiazen: Just used in tests.
The test is a small probe work to prototype LeaseManager based queue balancer, mainly to make sure it can access most dependency object it needs to access.

test/ServiceBus.Tests/PlugableQueueBalancerTests/LeaseManagerGrain.cs
```diff
@@ -0,0 +1,106 @@
...
+        private static DateTime UnAssignedLeaseTime = DateTime.MinValue;
+        private Dictionary<QueueId, DateTime> queueLeaseToRenewTimeMap;
+        private ISiloStatusOracle siloStatusOracle;
+        private IStreamQueueMapper queueMapper;
+        private ClusterConfiguration clusterConfiguration;
+        public override Task OnActivateAsync()
+        {
+            this.siloStatusOracle = base.ServiceProvider.GetRequiredService<ISiloStatusOracle>();
+            this.clusterConfiguration = base.ServiceProvider.GetRequiredService<ClusterConfiguration>();
+            this.queueMapper = base.ServiceProvider.GetServiceByName<IStreamQueueMapper>(this.GetPrimaryKeyString());
```
xiazen: "LeaseManager being grain" has been discussed as a possibility. I want to make sure LeaseManager**Grain**BasedQueueBalancer can be an option later. 

Just want to bring up the concern that, if we go with "Lease manager being a grain" option, assign it complex dependency will not be intuitive with this PR's interface design. 

As an example: I found it difficult for LeaseManagerGrain to access StreamQueueMapper, which intuitively should be its dependency. This is a pretty hacky way to access it. We can think of other more elegant way in the future.

test/ServiceBus.Tests/PlugableQueueBalancerTests/PluggableQueueBalancerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            new PersistentStreamProviderConfig();
+
+        private readonly Fixture fixture;
+
+        public class Fixture : BaseTestClusterFixture
+        {
+            protected override TestCluster CreateTestCluster()
+            {
+                var options = new TestClusterOptions(siloCount);
+                ProviderSettings.EventHubPartitionCount = totalQueueCount;
```
xiazen: Use EventDataGeneratorStreamProvider mainly because it is a very configurable provider, I can assign any queue I like through config

xiazen:

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
```diff
@@ -8,27 +8,32 @@ namespace Orleans.Streams
     /// The stream queue balancer is responsible for load balancing queues across all other related queue balancers.  It
     /// notifies any listeners (<code>IStreamQueueBalanceListener</code>) of changes to the distribution of queues.
     /// </summary>
-    internal interface IStreamQueueBalancer
+    public interface IStreamQueueBalancer
     {
         /// <summary>
+        /// Async Initialize logic
```
ReubenBond: Something like "Initializes this instance", perhaps

test/ServiceBus.Tests/PlugableQueueBalancerTests/PluggableQueueBalancerTests.cs
jason-bragg: Would prefer if these tests where in a test harness so we could run them against other persistent providers.
Initial test should probably use memory stream provider, It allows the same queue flexibility, while also allowing us to do this in a sharable harness.

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
xiazen: hmm, I'm not sure I follow. What else instance will this be initializing? it is initializing itself. So "this instance" is a bit redundant to me

ReubenBond: yep, just the doc comments. There are many `Initialize` methods in Orleans, so I wanted them to be aligned. Some examples from our code:
```C#
        /// <summary>
        /// Initializes the external serializer. Called once when the serialization manager creates 
        /// an instance of this type
        /// </summary>
        void Initialize(Logger logger);

    /// <summary>
    /// Initializes the client runtime from the standard client configuration file.
    /// </summary>
    public static void Initialize()

        /// <summary>
        /// Initializes this instance.
        /// </summary>
        public void Initialize()
```

We don't use StyleCop, but standard MS StyleCop rules check that constructor doc comments all follow a pattern like this:
```C#
/// <summary>
/// Initializes a new instance of the <see cref="AssemblyProcessor"/> class.
/// </summary>
```

src/Orleans/Streams/PersistentStreams/StreamQueueBalancerType.cs
```diff
@@ -39,5 +39,10 @@ public enum StreamQueueBalancerType
         /// That is, it does not rebalance queues based on dymanic changes in the cluster Membership.
         /// </summary>
         StaticClusterConfigDeploymentBalancer,
+
+        /// <summary>
+        /// Use user configured custom queue balancer
+        /// </summary>
+        CustomBalancer
```
ReubenBond: It would result in a larger change, but I have strong preference to fetching the balancer from DI and only using this type if no balancer has been added to the container. The NamedServices DI feature can hopefully be used so that different providers can have different balancers.

test/ServiceBus.Tests/PlugableQueueBalancerTests/PluggableQueueBalancerTests.cs
xiazen: yeah make sense !

src/Orleans/Streams/PersistentStreams/StreamQueueBalancerType.cs
ReubenBond: Check out `ReminderTableFactory` for an example of how this is done today. It will have to change a bit for named services, but it's registered in `Silo` as:
```C#
services.AddSingleton<ReminderTableFactory>();
services.AddSingleton<IReminderTable>(sp => sp.GetRequiredService<ReminderTableFactory>().Create());
```

If the user injects their own `IReminderTable`, then that will be used instead.
Ideally we would only add our default one if they have not added their own - maybe @jdom knows a better way?

src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
```diff
@@ -156,22 +161,22 @@ public bool SubscribeToQueueDistributionChangeEvents(IStreamQueueBalanceListener
...
+                    return Task.FromResult(false);
                 }
                 queueBalanceListeners.Add(observer);
-                return true;
+                return Task.FromResult(true);
             }
         }
 
-        public bool UnSubscribeToQueueDistributionChangeEvents(IStreamQueueBalanceListener observer)
+        public Task<bool> UnSubscribeToQueueDistributionChangeEvents(IStreamQueueBalanceListener observer)
```
ReubenBond: We should rename this: `UnsubscribeFromQueueBalancingEvents` and similar for the subscribe method

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
```diff
@@ -37,7 +42,7 @@ internal interface IStreamQueueBalancer
     /// It should be implemented by components interested in stream queue load balancing.
     /// When change notification is received, listener should request updated list of queues from the queue balancer.
     /// </summary>
-    internal interface IStreamQueueBalanceListener : IAddressable
+    public interface IStreamQueueBalanceListener : IAddressable
```
ReubenBond: Does this need to be IAddressable?

src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
xiazen: why do you favor QueueBalancing over QueueDistribution? queue balancing is a verb, and queue distribution is a norm which describe current queue distribution state, right? seems to me , it make more sense to use QueueDistribution

src/Orleans/Streams/PersistentStreams/StreamQueueBalancerType.cs
xiazen: @jason-bragg  has similar concern on that configuration is a bit redundant. He suggested something which convinced me, and I think you will like it. 

He suggested remove this type, and make StreamQueueBalancerType.cs a list of static string, which list the build in queue balancers. so the runtime can resolve the queue balancer using named service. And when user want to inject their own queue balancer, they can configure PersistenStreamProviderConfig.BalancerType to be their custom queue balancer type by giving it a name. And we can remove the other QueueBalancerFactoryName config . 

I think it is similar to what you suggested. Hope this make sense, or you will see in my PR update ;)

xiazen: As to add balancers to DI. I suspect the value of it. Objects are put into DI because it needs to be available in multiple places. But for balancers, it got created once per stream provider and pass to pulling agent manager, and it's done. I don't see where DI comes to play here. 

And adding balancer to DI doesn't play well with its life cycle. every stream provider create balancer at the step of initializing pulling agent manager, with a different set of params per stream provider. This is hard to access and differentiate in the step of building DI container 

It make sense to me to add BalancerFactory to DI as a named service, because that's the most convenient and familiar pattern for Orleans user to configure a custom service now.

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
xiazen: it was IAddressable before. But you are right, this limits listener to be only IAddressable.

I wonder is there historical reason for it @jason-bragg ?

src/Orleans/Streams/PersistentStreams/StreamQueueBalancerType.cs
```diff
@@ -1,19 +1,22 @@
 namespace Orleans.Streams
 {
-    public enum StreamQueueBalancerType
+    /// <summary>
+    /// Built-in stream queue balancer type which is supported natively in orleans
+    /// </summary>
+    public static class BuiltInStreamQueueBalancerType
```
xiazen: Changing the name of this class will make it a breaking change. But I slightly prefer this name, which has clearer indication, especially now we support pluggable queue balancer now. But I will be fine if you want to change it back

src/OrleansRuntime/Streams/QueueBalancer/IStreamQueueBalancerFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Orleans.Streams
+{
+    public interface IStreamQueueBalancerFactory
+    {
```
xiazen: still think this should be an interface, instead of a factory delegate... 
I think factory as a delegate will be good if it is part of a bigger class's property, it is light weight that way. But that is not true in this PR. 
User inject a factory instance, instead of a factory delegate fits more into OOP patterns

xiazen:
updated with fixes for PR feedback

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
```diff
@@ -8,27 +8,32 @@ namespace Orleans.Streams
...
+        /// Initialize this instance
+        /// </summary>
+        /// <returns></returns>
+        Task Initialize();
+        /// <summary>
         /// Retrieves the latest queue distribution for this balancer.
         /// </summary>
         /// <returns>Queue allocated to this balancer.</returns>
-        IEnumerable<QueueId> GetMyQueues();
+        Task<IEnumerable<QueueId>> GetMyQueues();
```
jason-bragg: Why did you make these tasks?

src/Orleans/Streams/PersistentStreams/StreamQueueBalancerType.cs
jason-bragg: let's leave the enum, and have this class expose string for the enums.  That way legacy code/tests can still use these enum values.

In the config object we can add a 'string QueueBalancer' property and change the BalancerType to be a passthrough that setts the  QueueBalancer'.  The property in the property bag would remain "BalancerType"

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
xiazen: I explained it in my comments before: to make it possible for the queue balancer to be built on top of an async service,  such as an async lease manager, or a storage service.

src/Orleans/Streams/PersistentStreams/StreamQueueBalancerType.cs
xiazen: ?? if you are worrying about breaking change, only by changing the name would make this a breaking change. Change it from a  enum class to a static class with strings won't make it a breaking change. By breaking change I mean: user will have compiling errors from it. 

And I don't fully understand the logic behind an extra QueueBalancer setting. To avoid breaking change? Then back to my original question. 

I guess my question is, what is your concern? if you are worrying about breaking change, then I should just change the name back, right?

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerFactory.cs
```diff
@@ -21,7 +22,7 @@ internal class StreamQueueBalancerFactory
         /// <param name="siloMaturityPeriod">Maturity Period of a silo for queue rebalancing purposes</param>
         /// <returns>Constructed stream queue balancer</returns>
         public static IStreamQueueBalancer Create(
-            StreamQueueBalancerType balancerType,
+            string balancerType,
             string strProviderName,
             ISiloStatusOracle siloStatusOracle,
```
jason-bragg: ISiloStatusOracle, ClusterConfiguration, and IStreamProviderRuntime, can be obtained by DI in the factory's constructor.  We should probably remove them from the Create call.

src/Orleans/Streams/PersistentStreams/StreamQueueBalancerType.cs
jason-bragg: I was concerned about breaking changes.  It seems like we could maintain the old enums and add support for strings, but in retrospect, it's probably best to have only one supported way (strings) and eat the cost of a breaking change.

I withdraw my suggestion.

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
jason-bragg: Ah, sorry, I missed that.  Now I understand.
Ok, so we'd considered this in the original pattern, and chose for this to be synchronous.  The reason for this is that this is a push+pull pattern.  We notify the IStreamQueueBalanceListeners once we get a change, so we already have the change at that point, so there is no reason to be asynchronous.  This is actually important, because the agent is reentrant and giving up it's turn to get the queues could have unexpected side effects.  Even the subscribe and unsubscribe calls where expected to be local, because the queue balancer is expected to be per silo, we don't want it to be distributed and asynchronous.

src/Orleans/Streams/PersistentStreams/StreamQueueBalancerType.cs
xiazen: but it is not a breaking change if we change this class's name back. If we want to avoid breaking change, then we need to change the class name back, which is what I try to say. 
People use this as `BalancerType = StreamQueueBalancerType.WhateverType. ` after this change, object in both sides of the "=" will become string, so it won't cause compile problems.

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerFactory.cs
xiazen: not a bit fan of service locator pattern. but seems like that's the orleans style now. we use it in grain public API too..... 
so remove those and add IServiceProvider param?

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
xiazen: But if we make this synchronous, for example, how would the balancer get its updated queues in a rebalancing scenario? for example, one silo shut down, current balancer need to locate more queues on it from the async leasing manager. How would this work as a injected balancer? 

The whole pluggable queue balancer feature is very limited. You cannot plug in a very powerful balancer without changing the internal runtime.

xiazen: will change it back

xiazen: Thanks for the heads up

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerFactory.cs
jason-bragg: My thinking here is to avoid passing anything that can be obtained by DI, passing only the things that the queue balancer manager alone has, like balancer type, provider name, and maturity period.  The container can be obtained from DI, right?

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
jason-bragg: May need to note in the comments, that GetMyQueues, SubscribeToQueueDistributionChangeEvents, and UnSubscribeFromQueueDistributionChangeEvents will likely be called in the IStreamQueueBalanceListener's thread so they need to be thread safe.

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerFactory.cs
ReubenBond: yes, but we don't need the container either.

src/Orleans/Streams/Providers/IStreamProviderRuntime.cs
```diff
@@ -91,20 +91,25 @@ public class PersistentStreamProviderConfig
         public static readonly TimeSpan DEFAULT_STREAM_INACTIVITY_PERIOD = TimeSpan.FromMinutes(30);
 
         public const string QUEUE_BALANCER_TYPE = "QueueBalancerType";
-        public const StreamQueueBalancerType DEFAULT_STREAM_QUEUE_BALANCER_TYPE = StreamQueueBalancerType.ConsistentRingBalancer;
+        //default balancer type if ConsistentRingBalancer
+        public static Type DEFAULT_STREAM_QUEUE_BALANCER_TYPE = null;
```
jason-bragg: Don't think we need this.. default would be null anyway, right?

src/Orleans/Streams/PersistentStreams/IPersistentStreamPullingAgent.cs
```diff
@@ -13,7 +13,7 @@ internal interface IPersistentStreamPullingAgent : ISystemTarget, IStreamProduce
 
     internal interface IPersistentStreamPullingManager : ISystemTarget
     {
-        Task Initialize(Immutable<IQueueAdapter> queueAdapter);
+        Task Initialize(Immutable<IQueueAdapter> queueAdapter, Immutable<IStreamQueueMapper> queueMapper);
```
jason-bragg: We can get the mapper from the adapter, why pass in the additional arg?

src/OrleansRuntime/Streams/QueueBalancer/ConsistentRingQueueBalancer.cs
```diff
@@ -6,20 +6,26 @@
 
 namespace Orleans.Streams
 {
-    internal class ConsistentRingQueueBalancer : IAsyncRingRangeListener, IStreamQueueBalancer
+    public class ConsistentRingQueueBalancer : IAsyncRingRangeListener, IStreamQueueBalancer
```
jason-bragg: y make this, and all that it requires, public.  This is can/should remain internal, right?

src/Orleans/Streams/PersistentStreams/IPersistentStreamPullingAgent.cs
xiazen: U are fast ! wait for 30 mins and come back ! I just submitted this. haven't clean up yet.

xiazen: sorry for the confusion

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerFactory.cs
```diff
@@ -21,7 +22,7 @@ internal class StreamQueueBalancerFactory
         /// <param name="siloMaturityPeriod">Maturity Period of a silo for queue rebalancing purposes</param>
         /// <returns>Constructed stream queue balancer</returns>
         public static IStreamQueueBalancer Create(
```
jason-bragg: Thought we were getting rid of this class?  No longer needed right?  Queue manager should be able to get balancer form DI by type, then initialize it directly.

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerType.cs
```diff
@@ -1,19 +1,24 @@
+using System;
+
 namespace Orleans.Streams
 {
-    public enum StreamQueueBalancerType
+    /// <summary>
+    /// Built-in stream queue balancer type which is supported natively in orleans
+    /// </summary>
+    public static class StreamQueueBalancerType
```
jason-bragg: This is where we can define the default queue balancer.

src/Orleans/Streams/Providers/IStreamProviderRuntime.cs
xiazen: mainly for less breaking change, so user can still do 
`providerSettings.BalancerType = PersistentStreamProviderConfig.DEFAULT_STREAM_QUEUE_BALANCER_TYPE`, when they configure stream provider. 
Although this seems like an unusual usage pattern, since without this line it will still be the default value. So I don't have strong option against removing this

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerType.cs
xiazen: Not sure I follow. Let's talk in person tomorrow

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerFactory.cs
xiazen: see my latest commit. We decided to shift from "configure balancer through a factory" to "configure balancer through its init method", as a more straight-forward alternative. so this factory interface is removed. 
Concern for a keyed factory is, it  split configuration into two places: factory class and keyed service config in DI, which IMO is not an very comprehensive pattern to users. And the constraits we put on the interface is not enough to avoid bad user code.  It's easy for user to misunderstand this feature as "inject an factory as a singleton and do a case-switch inside its create method to generate the correct balancer"

xiazen: And I assume we were talking about `IStreamQueueBalancerFactory `interface, because this diff lie on the internal `StreamQueueBalancerFactory `class, which doesn't implement `IStreamQueueBalancerFactory `and is a completely different thing. 

And if we are indeed talking about the internal `StreamQueueBalancerFactory`, I'm not opposed to create this internal factory through DI, just pointing out that this internal factory class is a static class with a static method and used just in one place, not sure how valuable we is that change.

src/OrleansRuntime/Streams/QueueBalancer/StreamQueueBalancerFactory.cs
xiazen: This is the internal factory class which is always there, not the public factory interface. 

I kept it because our built in queue balancer need to be configured with dependency which is not in DI or in the queueBalancer.init method param, such as `IDeploymentConfiguration `, `IConsistentRingProviderForGrains`. 

We can remove this class by adding `IDeploymentConfiguration `, `IConsistentRingProviderForGrains` into DI. But adding IDeploymentConfiguration to DI is non-trivia. One of its concrete class is ServiceRuntimeWrapper, which is in AzureUtils. I'm not familiar with how to inject a service to DI in an azure silo

src/Orleans/Streams/Providers/IStreamProviderRuntime.cs
```diff
@@ -91,20 +91,25 @@ public class PersistentStreamProviderConfig
...
 
-
         public TimeSpan GetQueueMsgsTimerPeriod { get; set; } = DEFAULT_GET_QUEUE_MESSAGES_TIMER_PERIOD;
         public TimeSpan InitQueueTimeout { get; set; } = DEFAULT_INIT_QUEUE_TIMEOUT;
         public TimeSpan MaxEventDeliveryTime { get; set; } = DEFAULT_MAX_EVENT_DELIVERY_TIME;
         public TimeSpan StreamInactivityPeriod { get; set; } = DEFAULT_STREAM_INACTIVITY_PERIOD;
-        public StreamQueueBalancerType BalancerType { get; set; } = DEFAULT_STREAM_QUEUE_BALANCER_TYPE;
+
+        /// <summary>
+        /// The queue balancer type for your stream provider. If you are using a custom queue balancer by injecting IStreamQueueBalancer as a transient service into DI,
```
jason-bragg: probably should specify 'transient', we don't really care about the scoping.

xiazen: I specified transient at the end. scroll 💃

xiazen:
@dotnet-bot test this please

PR #3158, opened 2017-06-23
Fix NullReferenceException when no LogConsistencyProvider attribute is provided

Liversage:
If a `LogConsistencyProvider` is configured with the name `Default` then a `JournaledGrain` without a `LogConsistencyProviderAttribute` will use that provider. However, a bug in a logging statement resulted in a `NullReferenceException`. The name of the provider was retrieved from the variable `attr` that is `null` when no attribute is specified.

The fix is to get the name of the provider from the provider itself.

It took me some time to pinpoint the location of this bug in the source code. Interestingly, when I opened the file in Visual Studio ReSharper flagged the exact statement I had located by code inspection with a green squiggly to indicate a possible `NullReferenceException`.


sergeybykov:

sergeybykov:
Thank you, @Liversage!

Issue #3141, opened 2017-06-17
Flag Postgresql as synchronous in RelationalStorage to avoid deadlocking


veikkoeeva:
This is related to https://github.com/dotnet/orleans/pull/2949. It appears the Npsql provider has similar problems to MySQL one, the fix is [to flag it likewise](https://gitter.im/dotnet/orleans?at=59445bb6f31c8ced0c42e583). More information at findings at Gitter discussion at https://gitter.im/dotnet/orleans?at=59445bb6f31c8ced0c42e583.

jdom:
I think this is a good safe approach until https://github.com/npgsql/npgsql/issues/1130 is fixed.

PR at https://github.com/dotnet/orleans/pull/3164

Issue #3019, opened 2017-05-14
Tests fail for both SqlServer and Postgres


xclayl:
SQL Server Error (for the init test in this case):

Test Name:	UnitTests.SqlStatisticsPublisherTests.SqlServerStatisticsPublisherTests.SqlStatisticsPublisher_SqlServer_Init
Test FullName:	UnitTests.SqlStatisticsPublisherTests.SqlServerStatisticsPublisherTests.SqlStatisticsPublisher_SqlServer_Init
Test Source:	 : line 0
Test Outcome:	Failed
Test Duration:	0:00:00.001

Result StackTrace:	
at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at UnitTests.SqlStatisticsPublisherTests.SqlStatisticsPublisherTestsBase..ctor(ConnectionStringFixture fixture, TestEnvironmentFixture environment)
   at UnitTests.SqlStatisticsPublisherTests.SqlServerStatisticsPublisherTests..ctor(ConnectionStringFixture fixture, TestEnvironmentFixture environment)
----- Inner Stack Trace -----
   at UnitTests.SqlStatisticsPublisherTests.StatisticsPublisherProviderRuntime.get_ServiceProvider()
   at Orleans.Providers.SqlServer.SqlStatisticsPublisher.<Init>d__16.MoveNext()
Result Message:	
System.AggregateException : One or more errors occurred.
---- System.NotImplementedException : The method or operation is not implemented.

The Postgres test error is very similar:

Test Name:	UnitTests.SqlStatisticsPublisherTests.PostgreSqlStatisticsPublisherTests.SqlStatisticsPublisher_PostgreSql_Init
Test FullName:	UnitTests.SqlStatisticsPublisherTests.PostgreSqlStatisticsPublisherTests.SqlStatisticsPublisher_PostgreSql_Init
Test Source:	 : line 0
Test Outcome:	Failed
Test Duration:	0:00:00.001

Result StackTrace:	
at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
[...]

PR #3163, opened 2017-06-23
Adds to Roman's fix the test fixes

veikkoeeva:
This works so that

- Fixes cancellation test by unwrapping the real exception

The reason for the cancellation test failure is due to the test arrangement where the .Wait wraps the exception into an AggregateException and the real exception needs to be unwrapped first.

- Lowers database stress level to avoid deadlocking in tests (maybe not needed).

The test failures were due to deadlocks in the database where reads and writes were waiting for the same resource. As an intermediary resolution the test stress level was lowered to be more in line with the other persistence storage test.

Indexing would greatly improve the situation, but even better fix would be to rearrange the table. Indexing can be applied by anyone running a system without correctness problems with regards to Orleans, so maybe the right way to refactor wouldn't be to introduce "default indexing" (as it depends on hardware deployment too) but rearrange the table and index the rearranged table.

- Fixes a problem in how `static ThreadLocal<Random>` could potentially be seeded multiple times with the same value.

- Adds some `.ConfigureAwait(false)` calls to tests as a measure of "best practice", not strictly needed, but maybe it's more benefit to the codebase than to leaving "to the next time" when some visits it.

Closes #2949.

### DIFF
--- a/src/AdoNet/Shared/Storage/RelationalStorage.cs
+++ b/src/AdoNet/Shared/Storage/RelationalStorage.cs
@@ -150,7 +150,7 @@ namespace Orleans.Statistics.AdoNet.Storage
         ///}).ConfigureAwait(continueOnCapturedContext: false);                
         /// </code>
         /// </example>
-        public Task<IEnumerable<TResult>> ReadAsync<TResult>(string query, Action<IDbCommand> parameterProvider, Func<IDataRecord, int, CancellationToken, Task<TResult>> selector, CancellationToken cancellationToken = default(CancellationToken), CommandBehavior commandBehavior = CommandBehavior.Default)
+        public async Task<IEnumerable<TResult>> ReadAsync<TResult>(string query, Action<IDbCommand> parameterProvider, Func<IDataRecord, int, CancellationToken, Task<TResult>> selector, CancellationToken cancellationToken = default(CancellationToken), CommandBehavior commandBehavior = CommandBehavior.Default)
         {
             //If the query is something else that is not acceptable (e.g. an empty string), there will an appropriate database exception.
             if(query == null)
@@ -163,11 +163,7 @@ namespace Orleans.Statistics.AdoNet.Storage
                 throw new ArgumentNullException("selector");
             }
 
-            //It is certain the result is already ready here and can be collected straight away. Having a truly asynchronous result collection
-            //IAsyncEnumerable<TResult> or equivalent method ought to be used. Taking the result here without async-await saves for generating
-            //the async state machine.
-            var ret = ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).GetAwaiter().GetResult().Item1;
-            return Task.FromResult(ret);
+            return (await ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).ConfigureAwait(false)).Item1;
         }
 
 
@@ -192,7 +188,7 @@ namespace Orleans.Statistics.AdoNet.Storage
         /// }).ConfigureAwait(continueOnCapturedContext: false);                
         /// </code>
         /// </example>
-        public Task<int> ExecuteAsync(string query, Action<IDbCommand> parameterProvider, CancellationToken cancellationToken = default(CancellationToken), CommandBehavior commandBehavior = CommandBehavior.Default)
+        public async Task<int> ExecuteAsync(string query, Action<IDbCommand> parameterProvider, CancellationToken cancellationToken = default(CancellationToken), CommandBehavior commandBehavior = CommandBehavior.Default)
         {
             //If the query is something else that is not acceptable (e.g. an empty string), there will an appropriate database exception.
             if(query == null)
@@ -200,11 +196,7 @@ namespace Orleans.Statistics.AdoNet.Storage
                 throw new ArgumentNullException("query");
             }
 
-            //It is certain the result is already ready here and can be collected straight away. Having a truly asynchronous result collection
-            //IAsyncEnumerable<TResult> or equivalent method ought to be used. Taking the result here without async-await saves for generating
-            //the async state machine.
-            var ret = ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, (unit, id, c) => Task.FromResult(unit), cancellationToken, commandBehavior).GetAwaiter().GetResult().Item2;
-            return Task.FromResult(ret);
+            return (await ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, (unit, id, c) => Task.FromResult(unit), cancellationToken, commandBehavior).ConfigureAwait(false)).Item2;
         }
 
         /// <summary>
--- a/src/AdoNet/Shared/Storage/RelationalStorage.cs
+++ b/src/AdoNet/Shared/Storage/RelationalStorage.cs
@@ -225,7 +225,7 @@ namespace Orleans.Statistics.AdoNet.Storage
                     results.Add(obj);
                 }
 
-                await reader.NextResultAsync(cancellationToken);
+                await reader.NextResultAsync(cancellationToken).ConfigureAwait(false);
                 ++resultSetCount;
             }
 
--- a/test/TesterInternal/RemindersTest/ReminderTableTestsBase.cs
+++ b/test/TesterInternal/RemindersTest/ReminderTableTestsBase.cs
@@ -29,7 +29,7 @@ namespace UnitTests.RemindersTest
         protected IOptions<StorageOptions> storageOptions;
         protected IOptions<AdoNetOptions> adoNetOptions;
         protected const string testDatabaseName = "OrleansReminderTest";//for relational storage
-        
+
         protected ReminderTableTestsBase(ConnectionStringFixture fixture, TestEnvironmentFixture clusterFixture, LoggerFilterOptions filters)
         {
             fixture.InitializeConnectionStringAccessor(GetConnectionString);
@@ -78,7 +78,13 @@ namespace UnitTests.RemindersTest
             var upserts = await Task.WhenAll(Enumerable.Range(0, 5).Select(i =>
             {
                 var reminder = CreateReminder(MakeTestGrainReference(), i.ToString());
-                return Task.WhenAll(Enumerable.Range(1, 5).Select(j => remindersTable.UpsertRow(reminder)));
+                return Task.WhenAll(Enumerable.Range(1, 5).Select(j =>
+                {
+                    return RetryHelper.RetryOnExceptionAsync<string>(5, RetryOperation.Sigmoid, async () =>
+                    {
+                        return await remindersTable.UpsertRow(reminder);
+                    });
+                }));
             }));
             Assert.DoesNotContain(upserts, i => i.Distinct().Count() != 5);
         }
@@ -89,7 +95,7 @@ namespace UnitTests.RemindersTest
             await remindersTable.UpsertRow(reminder);
 
             var readReminder = await remindersTable.ReadRow(reminder.GrainRef, reminder.ReminderName);
-            
+
             string etagTemp = reminder.ETag = readReminder.ETag;
 
             Assert.Equal(JsonConvert.SerializeObject(readReminder), JsonConvert.SerializeObject(reminder));
@@ -113,7 +119,12 @@ namespace UnitTests.RemindersTest
             await Task.WhenAll(Enumerable.Range(1, iterations).Select(async i =>
             {
                 GrainReference grainRef = MakeTestGrainReference();
-                await remindersTable.UpsertRow(CreateReminder(grainRef, i.ToString()));
+
+                await RetryHelper.RetryOnExceptionAsync<Task>(10, RetryOperation.Sigmoid, async () =>
+                {
+                    await remindersTable.UpsertRow(CreateReminder(grainRef, i.ToString()));
+                    return Task.CompletedTask;
+                });
             }));
 
             var rows = await remindersTable.ReadRows(0, uint.MaxValue);
@@ -125,7 +136,7 @@ namespace UnitTests.RemindersTest
             Assert.Equal(rows.Reminders.Count, iterations);
 
             var remindersHashes = rows.Reminders.Select(r => r.GrainRef.GetUniformHashCode()).ToArray();
-            
+
             SafeRandom random = new SafeRandom();
 
             await Task.WhenAll(Enumerable.Range(0, iterations).Select(i =>
--- a/test/TesterInternal/RemindersTest/ReminderTableTestsBase.cs
+++ b/test/TesterInternal/RemindersTest/ReminderTableTestsBase.cs
@@ -75,7 +75,7 @@ namespace UnitTests.RemindersTest
 
         protected async Task RemindersParallelUpsert()
         {
-            var upserts = await Task.WhenAll(Enumerable.Range(0, 50).Select(i =>
+            var upserts = await Task.WhenAll(Enumerable.Range(0, 5).Select(i =>
             {
                 var reminder = CreateReminder(MakeTestGrainReference(), i.ToString());
                 return Task.WhenAll(Enumerable.Range(1, 5).Select(j => remindersTable.UpsertRow(reminder)));
--- a/test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
+++ b/test/TesterInternal/MembershipTests/MembershipTableTestsBase.cs
@@ -352,7 +352,7 @@ namespace UnitTests.MembershipTests
 
             TableVersion newTableVer = tableData.Version.Next();
 
-            var insertions = Task.WhenAll(Enumerable.Range(1, 20).Select(i => membershipTable.InsertRow(data, newTableVer)));
+            var insertions = Task.WhenAll(Enumerable.Range(1, 20).Select(async i => { try { return await membershipTable.InsertRow(data, newTableVer); } catch { return false; } }));
 
             Assert.True((await insertions).Single(x => x), "InsertRow failed");
 
@@ -367,7 +367,7 @@ namespace UnitTests.MembershipTests
                     TableVersion tableVersion = updatedTableData.Version.Next();
 
                     await Task.Delay(10);
-                    done = await membershipTable.UpdateRow(updatedRow.Item1, updatedRow.Item2, tableVersion);
+                    try { done = await membershipTable.UpdateRow(updatedRow.Item1, updatedRow.Item2, tableVersion); } catch { done = false; }
                 } while (!done);
             })).WithTimeout(TimeSpan.FromSeconds(30));
 
@@ -389,10 +389,10 @@ namespace UnitTests.MembershipTests
             MembershipEntry newEntry = CreateMembershipEntryForTest();
             bool ok = await membershipTable.InsertRow(newEntry, newTableVersion);
             Assert.True(ok);
-            
-            
+
+
             var amAliveTime = DateTime.UtcNow;
-            
+
             // This mimics the arguments MembershipOracle.OnIAmAliveUpdateInTableTimer passes in
             var entry = new MembershipEntry
             {
--- a/test/TesterInternal/RetryHelper.cs
+++ b/test/TesterInternal/RetryHelper.cs
@@ -1,0 +1,89 @@
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace UnitTests
+{
+    /// <summary>
+    /// Predeterimed retry operation, used with <see cref="RetryHelper"/>.
+    /// </summary>
+    /// <remarks>A very crude system for current tests.</remarks>
+    public static class RetryOperation
+    {
+        /// <summary>
+        /// Procudes sigmoid shaped delay curve.
+        /// </summary>
+        /// <param name="retryAttempt">The current retry attempt.</param>
+        /// <returns>Delay for the next attemp to retry.</returns>
+        /// <remarks>This function has not received deep scrutiny.</remarks>
+        public static TimeSpan Sigmoid(int retryAttempt)
+        {
+            const int MaxDurationInSeconds = 50;
+            return TimeSpan.FromMilliseconds(Convert.ToInt32(Math.Round((1 / (1 + Math.Exp(-retryAttempt + 3))) * MaxDurationInSeconds)) * 1000);
+        }
+
+
+        /// <summary>
+        /// Produces a linear delay curve in increments of 10 ms.
+        /// </summary>
+        /// <param name="retryAttempt">The current retry attempt.</param>
+        /// <remarks>This function has not received deep scrutiny.</remarks>
+        public static TimeSpan LinearWithTenMilliseconds(int retryAttempt)
+        {
+            return TimeSpan.FromMilliseconds(retryAttempt * 10);
+        }
+    }
+
+
+    /// <summary>
+    /// A simple retry helper to make testing more robust.
+    /// </summary>
+    internal static class RetryHelper
+    {
+        /// <summary>
+        /// A scaffolding function to retry operations according to parameters.
+        /// </summary>
+        /// <typeparam name="TResult">The result type of the function to retry.</typeparam>
+        /// <param name="maxAttempts">Maximum number of retry attempts.</param>
+        /// <param name="retryFunction">The function taking the current retry parameter that provides the time to wait for next attempt.</param>
+        /// <param name="operation">The operation to retry.</param>
+        /// <param name="cancellation">The cancellation token.</param>
+        /// <returns>The result of the retry.</returns>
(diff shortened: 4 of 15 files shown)


sergeybykov:

src/OrleansSQLUtils/Storage/RelationalStorage.cs
```diff
@@ -146,11 +146,7 @@ public Task<IEnumerable<TResult>> ReadAsync<TResult>(string query, Action<IDbCom
                 throw new ArgumentNullException("selector");
```
shayhatsor: I think that it'd be better to have a `TaskCompletionSource` and use `SetException`.

src/OrleansSQLUtils/Storage/RelationalStorage.cs
```diff
@@ -175,19 +171,15 @@ public Task<IEnumerable<TResult>> ReadAsync<TResult>(string query, Action<IDbCom
         /// }).ConfigureAwait(continueOnCapturedContext: false);                
         /// </code>
         /// </example>
-        public Task<int> ExecuteAsync(string query, Action<IDbCommand> parameterProvider, CancellationToken cancellationToken = default(CancellationToken), CommandBehavior commandBehavior = CommandBehavior.Default)
+        public async Task<int> ExecuteAsync(string query, Action<IDbCommand> parameterProvider, CancellationToken cancellationToken = default(CancellationToken), CommandBehavior commandBehavior = CommandBehavior.Default)
         {
             //If the query is something else that is not acceptable (e.g. an empty string), there will an appropriate database exception.
             if(query == null)
             {
                 throw new ArgumentNullException("query");
```
shayhatsor: I think that it'd be better to have a `TaskCompletionSource` and use `SetException`.

src/OrleansSQLUtils/Storage/RelationalStorage.cs
veikkoeeva: @shayhatsor Can you elaborate?

src/OrleansSQLUtils/Storage/RelationalStorage.cs
veikkoeeva: @shayhatsor Can you elaborate?

src/OrleansSQLUtils/Storage/RelationalStorage.cs
shayhatsor: We can leave this method not async and just retun a task with an exception

src/OrleansSQLUtils/Storage/RelationalStorage.cs
shayhatsor: We can leave this method not async and just retun a task with an exception

src/OrleansSQLUtils/Storage/RelationalStorage.cs
jdom: But from what I see here, it does require to await anyway, so the method can't be synchronous.
Also, using FromException is hit and miss with regards to preserving a stack trace, never had good experiences with that.
Having said all that, if there was a chance to make this method sync (which initially it doesn't look like), throwing ArgumentException is the only permitted exception to be thrown synchronously according to TAP design guidelines, as it is considered a [usage error](https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap#exceptions) that shouldn't be handled, but the code should be fixed instead.

veikkoeeva: @jdom, @shayhatsor Maybe we can entertain optimizations chances later?

jdom: I agree, was just commenting that it might not even be an optimization to be done in the future. Still, @shayhatsor is the right guy to approve this. 
I saw a lot of tests are failing after this change, though

veikkoeeva: @jdom That's true for that specific comment (though in the spirit of optimizations there's at least `ValueTask` and maybe `IAsyncEnumeable`. But wait a moment, do you mean even after my purported fixes there are regressions that shouldn't exist? I see there's https://ci.dot.net/job/dotnet_orleans/job/master/job/netfx-functional_prtest/461/testReport/UnitTests.MembershipTests/SqlServerMembershipTableTests/MembershipTable_SqlServer_UpdateRowInParallel/ for instance and it's flagged as regression. I can't reproduce that, but it could be due to performance problems, though primary key violation could easily indicate something else too (though in that case I suppose there would be a retry).

veikkoeeva: @jdom I see there are others (at least now that I rebased). Let me examine them...

veikkoeeva: @jdom Statistics tests fail to `AggregateException` and the inner exception is `NotImplementedException`. Statistics is implemented, so this is probably something else with the setup (that could be vastly simplified, I have code to do constructor injection in XUnit.NET, for instance). Statistics tests fail

The second set of tests that fail is persistence tests, but here again there isn't a clear reason other than that inconsistent write is reported whereas there should be. Looking at the tests, these shouldn't have anything special in way of data, but they use `Func<IInternalGrainFactory, GrainReference> getGrainReference` which is different from 
```
internal Tuple<GrainReference, GrainState<TestState1>> GetTestReferenceAndState(long grainId, string version)
{
    return Tuple.Create(this.grainFactory.GetGrain(GrainId.GetGrainId(UniqueKey.NewKey(grainId, UniqueKey.Category.Grain))), new GrainState<TestState1> { State = new TestState1(), ETag = version });
}
```
to generate a reference. It feels like grain reference shouldn't be the case to `InconsistentStateException` as such, but I wonder if the case could be such that on write-read cycles a wrong reference is combined with a state and hence a subsequent update cycle won't locate a reference in the storage, fails to update it and hence an exception is reported.
InconsistentStateException

I write these as a head's up since I'm on the road today, so to speak, and likely won't have many minutes to put time besides this. I'm not sure about the following week (I have little time at least).

veikkoeeva: All right, in `SqlStatisticsPublisher` the line `this.grainReferenceConverter = providerRuntime.ServiceProvider.GetRequiredService<IGrainReferenceConverter>();` throws `NotImplementedException`. I don't know why, though, I'm gearing to train here. :)

veikkoeeva: Another quick plunge on the failing relational storage tests, here are the three first grain reference IDs from the first run:
1. {*grn/3/00000000+úÎhùBìóxWÚCôÃép}
2. {*grn/3/00000000+ðènûùlæøþáýøçüÁ}
3. {*grn/3/00000000+ðènûùlæøþáýøçüÁ}

The second and third ones are the same, so the test fails. This is taken from `StorageDataSetGenericHuge_Json_WriteRead`, which uses `StorageDataSetGeneric`, which in turn use `RandomUtilities.GetRandomGrainReference` go generate the reference. The thing is, these are deterministically not random now. I don't know the reason. Quickly spelunkig through the code, my hazard quess would be line `(Func<IInternalGrainFactory, GrainReference>)(grainFactory => RandomUtilities.GetRandomGrainReference<TGrainKey>(grainFactory)),` maybe return just being the same instance and hence returning the same value. This is just a hazard guess, but it's midnight and if someone has a fresh pair of eyes (@jdom, @ReubenBond?), it'd be great.

veikkoeeva: By the way, even if it should be really small chance to generate the same ID randomly, there's such a chance. Originally the idea was to test also whatver strange IDs one has from a given character set, just in case to see there isn't problems in there. One way forwards would to just take, say, two of that just numerated set, copy the third one from either one and change a few latters. This way the test set is always generic.

sergeybykov:
@dotnet-bot test netfx-bvt

sergeybykov:

src/OrleansSQLUtils/Storage/RelationalStorage.cs
veikkoeeva: @ReubenBond, @jdom The duplicate IDs could also be something like the pseudorandom generated being seeded again with the same seed (if some static was removed). In any event, it looks like there shouldn't be a logic failure. I might do some checking on this in a few hours.

sergeybykov:

sergeybykov:
We are wrapping up 1.5.0. Unless this is ready, we'll have to bump it to the next milestone.

sergeybykov:
@veikkoeeva Thank you for the details. We are done with 1.5.0. So this moves to 2.0.0

veikkoeeva:
Cross-referencing https://github.com/dotnet/orleans/issues/3019.

sergeybykov:
@dotnet-bot test bvt
@dotnet-bot test functional

sergeybykov:
@dotnet-bot test this please

sergeybykov:
@dotnet-bot test functional

liquidboy:
Sorry for the noise BUT i encountered this RelationalStorage deadlock issue (async) on a simple silo using a postgesql 10.x database.  I doubt any orleans project wanting to use a postgresql database will work without this fix, so would love to vote this up for inclusion in 2.0x release :)

veikkoeeva:
Cross-referencing https://github.com/dotnet/orleans/issues/3141 as this might be confusing without that one. At the time it appeared Npsql problem was fixed with that one, but @liquidboy sees otherwise. In any event, this should be included to the 2.0.0 series. I'll try to check how much effort there really is.

jdom:
@dotnet-bot test this please
Reason: try a few times to see if this is a regression or just a random failure

veikkoeeva:
@dotnet-bot test this please
Reason: It appears tests pass, hence no regression, but there is some other hard to why the build is marked as a failure.

sergeybykov:
@dotnet-bot test functional

veikkoeeva:
@sergeybykov The build fails on XUnit.NET serialiaization:
> System.Runtime.Serialization.SerializationException: Type 'Xunit.Sdk.EqualException' in Assembly 'xunit.assert, Version=2.3.1.3858, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c' is not marked as serializable.

Reuben noted from Japan it ought to be fixed. I updated the branch with `git pull --rebase upstream master`, so all the latest should be included. I also see this latest run fails on parallel reminders test due to deadlock. I could lower the stress level in the tests to help these arguably intermittent problems -- and noting the reminder implementation is probably as inefficient as it gets, cross referencing https://github.com/dotnet/orleans/issues/2429.

sergeybykov:
@dotnet-bot test functional

veikkoeeva:
@dotnet-bot test this please
Reason: It appears there was a rebase problem between the keyboard and the wall (no chair back then) that caused the described XUnit problem and now the tests hit a performance problem.

veikkoeeva:
Still some failures present, apparently also the XUnit one (using `( failure )` to find the troublesome parts). Seem to be unrelated to the changes.

ReubenBond:
@dotnet-bot test functional

veikkoeeva:
https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/920/testReport/junit/Tester.HeterogeneousSilosTests.UpgradeTests/AllVersionsCompatibleTests/DoNotUpgradeProxyCallNoPendingRequest/ Is something I have not seen before, the `UPSERT` tests overwhelmed the system and the XUnit thing seemed to lurk still (strange).

ReubenBond:
@dotnet-bot test functional

veikkoeeva:
Seems to be failing from taking a peek to the console. I'm off to bed now, but I wonder about that XUnit thing and if I should maybe squash the changes (would that be OK?) and reset the time of changes in the morning? I'll push the changed based on top of the latest master now anyhow.

ReubenBond:
oh, that test is actually failing. These failures are not an Orleans issue, they're a .NET (or maybe xUnit) issue

ReubenBond:
@dotnet-bot test functional

Reason: I wanted it to fail, but it passed

ReubenBond:
@dotnet-bot test functional

veikkoeeva:
For the reference, here is one such build: https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/921/.

ReubenBond:
I was looking for the source of the `EqualException`, @veikkoeeva - the one which couldn't traverse AppDomain boundaries (since no xUnit exceptions are marked as `[Serializable]`, apparently).

veikkoeeva:
@ReubenBond Is there something I could to help in local environment? The tests pass for me, but maybe I should look for first chance exceptions? I can do something over the weekend, all the better if there's some starting information I can use.

ReubenBond:
The suite seems to pass locally for me, too @veikkoeeva. Possibly each test class (or maybe the fixture) could add an AppDomain UnobservedException handler or FirstChanceException handler which then logs the exception so that we can find out where & why it's being thrown in CI

veikkoeeva:
@dotnet-bot test functional

Reason: The XUnit.NET serialization exception did not exist in the test logs, but a transient performance problem occurred.

veikkoeeva:
The problem seems to originate from `SqlStorageTests`. It looks to me the storage tests for MySQL and PostgreSQL are tried and skipped as those DBs aren't available, but for some reason SQL Server fails. Maybe adding more some sort of logging would help.

The problem looks like being described at https://github.com/dotnet/corefx/issues/13418. There's one fix proposed, I don't know if this is the right path to take.

veikkoeeva:
Looking at the logs, there are also failures such as 
> Initializing relational databases...
11:05:30 Dropping and recreating database 'OrleansStreamTest' with connectionstring 'Data Source = (localdb)\MSSQLLocalDB; Database = Master; Integrated Security = True; 
11:05:30                          Asynchronous Processing = True; Max Pool Size = 200; MultipleActiveResultSets = True'
11:05:30 Failed to initialize SQL Server for RelationalGeneralTests: System.Data.SqlClient.SqlException (0x80131904): Connection Timeout Expired.  The timeout period elapsed while attempting to consume the pre-login handshake acknowledgement.  This could be because the pre-login handshake failed or the server was unable to respond back in time.  The duration spent while attempting to connect to this server was - [Pre-Login] initialization=24847; handshake=32;  ---> System.ComponentModel.Win32Exception (0x80004005): The wait operation timed out

I don't have access to the server, but it could indicate a problem in the database instance running. Some reasons listed [here](https://stackoverflow.com/questions/15488922/connection-to-sql-server-works-sometimes) and [here]()https://social.msdn.microsoft.com/Forums/sqlserver/en-US/3bbcd9a3-f90b-4ef2-8244-f06ee8d9cfa3/timeout-error-while-attempting-to-consume-the-prelogin-handshake?forum=sqldataaccess. I.e. is the server alive and accepting connections and does it listen both IPv4 and IPv6 (and fails randomly due to that).

veikkoeeva:
@dotnet-bot test functional

Reason: Maybe less load, so no failures due to that.

veikkoeeva:
@dotnet-bot test functional

Reason: Too few resources (parallel insert failure).

veikkoeeva:
@dotnet-bot test functional

Reason: Too few resources (parallel insert failure).

ReubenBond:
> Reason: Too few resources (parallel insert failure).

Is there some way we can make the test less sensitive to resource availability?

veikkoeeva:
@ReubenBond 
> Is there some way we can make the test less sensitive to resource availability?

1. I have been thinking if I should implement a retry function, say a sigmoid function capped to some seconds. Since this has been discussed for all ADO.NET operations, I'll add for the benefit of other readers that in general the retries are somewhat application specific (i.e. how many retries, which kind of a function etc.), so I'm inclined to think in general in application code it's better to let everything through and let the developer use what is needed. Failures need to be handled in any case.

2. The test that gives trouble is reminders parallel insert that are nearly as inefficient as they get. It'd be greatly beneficial to improve them in any event and there's even an issue of it.

As of why I'm rerunning the tests, I want to be sure the tests really do fail only on XUnit serialization too.

veikkoeeva:
@dotnet-bot test functional

Reason: After making a modification to reduce a certain loop limit from `10` to `1` to eliminate some scenarios, there was 
> FirstChanceException in "Tester.SQLUtils: System.IO.FileNotFoundException: Could not load file or assembly 'TesterInternal.XmlSerializers, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.

which have not been there. I suspect this is some sort of a transient error.

veikkoeeva:
@ReubenBond It looks like if the CI machine changes from time to time. Should they be identical? It might be I unearthed a new set of problems with the latest changes or I have missed these ones, but now I got a assembly loading error during the few rounds.

I don't know if there's trouble in differences if the tests are run through Visual Studio or a console, taking into account bitness and things like that, which then cause all sorts of trouble.

veikkoeeva:
Oh, also, the line to look for in the logs is
> Dropping and recreating database 'OrleansStorageTests' with connectionstring 'Data Source = (localdb)\MSSQLLocalDB; Database = master; Integrated Security = True; Asynchronous Processing = True; Max Pool Size = 200;

There is plenty of output from exception handler logging for not being able to connect to MySQL or PostgreSQL databases. On that line forwards start the SQL Server storage tests which aren't skipped.

veikkoeeva:
@dotnet-bot test functional

Reason: Check CI effects.

veikkoeeva:
@dotnet-bot test functional

Reason: Check CI effects, without transient failures (maybe should add retry handler for tests).

veikkoeeva:
@dotnet-bot test functional

Reason: Hash placement failures certainly are a surprise. Really?

veikkoeeva:
@dotnet-bot test functional

Reason: Did the DB got cleaned up or why problems with duplicate inserts?

veikkoeeva:
@dotnet-bot test functional

Reason: To ensure the same problem manifests consistently as it hasn't been the case before the latest set of changes.

veikkoeeva:
@dotnet-bot test functional

Reason: To check if the membership tests still fail on primary key violation. A reason that has not been present before and it's difficult to see why that would be the case now related to these changes. The tests pass locally (as before). Maybe there is a difference when running from the command line, but using e.g. `Test.cmd`, the tests fail to other reasons.

veikkoeeva:
@dotnet-bot test functional

Reason: some placement and deactivation tests fail, likely bogus, so re-running functional tests.

veikkoeeva:
@dotnet-bot test functional

Reason: Failure on socket exceptions, unrelated to changes and seem to be transient CI failures.

veikkoeeva:
@dotnet-bot test functional

Reason: Failure due to async pipeline test that hasn't been there before and seem transient.

veikkoeeva:
@dotnet-bot test functional

Reason: Failure due to placement tests -- async pipeline test gone without changes to code. Seems transient.

veikkoeeva:
@dotnet-bot test

Reason: Transient socket errors.

veikkoeeva:
@dotnet-bot test please

Reason: Transient socket errors.

veikkoeeva:
@dotnet-bot test this please

Reason: Transient socket errors.

jdom:
Veikko, don't worry if there are some failures not related to your changes. The team is trying to reduce the flakiness of those. If you didn't make it less stable and you think the changes are ready for final review and merge, just let the team know. Thanks for your work

veikkoeeva:
@jdom I got the build to green and rebased and now just try to get it again green again just retrying enough many times to be sure everything is OK. :)

One thing I'd like to ask, if https://github.com/dotnet/orleans/pull/3163/commits/26c8c29cf1eb386e9bb6c6aec85c24971f37d197 is OK? Since this will fail the tests.

veikkoeeva:
I have some ideas to make the tests more robust and easier to manage. I've been thinking to overhaul the ADO.NET storage tests a bit by introducing a constructor injection system and make the DB setup, port reservations "process wide" and use them.

veikkoeeva:
@dotnet-bot test functional

Reason: Just making sure there aren't at least transient failures.

veikkoeeva:
@jdom, @ReubenBond Got the tests to function, now ready for review. Not the cleanest set of patches, but ready for review and naturally I can adjust as needed. I've got the tests green on two occasions today and the failures seem to have been transient ones, so in that respect I suppose this is OK.

ReubenBond:
@dotnet-bot test this please

src/AdoNet/Shared/Storage/RelationalStorage.cs
```diff
@@ -163,11 +163,7 @@ public Task<IEnumerable<TResult>> ReadAsync<TResult>(string query, Action<IDbCom
                 throw new ArgumentNullException("selector");
             }
 
-            //It is certain the result is already ready here and can be collected straight away. Having a truly asynchronous result collection
-            //IAsyncEnumerable<TResult> or equivalent method ought to be used. Taking the result here without async-await saves for generating
-            //the async state machine.
-            var ret = ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).GetAwaiter().GetResult().Item1;
-            return Task.FromResult(ret);
+            return (await ExecuteAsync(query, parameterProvider, ExecuteReaderAsync, selector, cancellationToken, commandBehavior).ConfigureAwait(false)).Item1;
         }
```
ReubenBond: Are the `ConfigureAwait(false)` calls necessary? If so, why?

test/TesterInternal/RemindersTest/ReminderTableTestsBase.cs
```diff
@@ -113,7 +119,12 @@ protected virtual string GetAdoInvariant()
             await Task.WhenAll(Enumerable.Range(1, iterations).Select(async i =>
             {
                 GrainReference grainRef = MakeTestGrainReference();
-                await remindersTable.UpsertRow(CreateReminder(grainRef, i.ToString()));
+
+                await RetryHelper.RetryOnExceptionAsync<Task>(10, RetryOperation.Sigmoid, async () =>
+                {
+                    await remindersTable.UpsertRow(CreateReminder(grainRef, i.ToString()));
+                    return Task.CompletedTask;
```
ReubenBond: This seems a little confusing to me: is the `operation` parameter to `RetryOnExceptionAsync` being inferred as `Func<Task<Task>>`? I think an overload is needed which allows `Func<Task>` instead. The inner task (`Task.CompletedTask`) is never awaited, from what I can tell. That's fine in this case, because it's a synchronously-completed, fixed value, but it could lead to a subtle bug in future.

Could instead add `.Unwrap()` to the `RetryOnExceptionAsync()` call, I think.

ReubenBond:
LGTM. If everyone else is happy and there are no more changes to be made (eg for the comments I left) then we can merge it.

I'm hoping that in future we will have AdoNet drivers which don't need us to jump through hoops

src/AdoNet/Shared/Storage/RelationalStorage.cs
jdom: From the callers perspective, this do make sense, as the context is not needed at all. Most of the times it is OK (although marginally less efficient) to not add these if all involved players do the right thing, nevertheless I remember that PostgreSQL driver was deadlocking with the Orleans execution context, and using `ConfigureAwait` in this class fixed it.

ReubenBond: I think that's going to be fixed in the next Postgres client release. Hopefully no other drivers cause trouble.

jdom: Hopefully, don't know. Still it makes sense to do the right thing just in case (the right thing is using ConfigureAwait)

test/TesterInternal/RemindersTest/ReminderTableTestsBase.cs
veikkoeeva: @ReubenBond Or call `await await`, might be entirely possible I just wrote noisy syntax. This `RetryHelper` is about five minutes of work and more like an internal, tempory resiliency measure in the CI (and slow machines) It's worth thinking if a robust one is needed and should be created or something else, such as Polly, be used. Either way, it occurred to me a set of precanned, parametrizable functions such as sigmoid, linear, exponential backoff etc. could be useful, together with diagrams.

src/AdoNet/Shared/Storage/RelationalStorage.cs
veikkoeeva: @ReubenBond, @jdom It's both "the right thing, easier to reader" but also to be surer to not to have deadlocks in the operations. It's not PostgreSQL only, also MySQL can cause trouble. I think even in the latest drivers their streaming operators will deadlock.

veikkoeeva:
@ReubenBond, @jdom Thanks. I think I'll do a separate pass to add `.this` and newer syntax and check these suggestions too.

Regarding the backoff thing, I might open a new issue contemplating resiliency measures for the CI first and foremost and additionally consider if there's a need for "retry functions". I should take a look what's on offer in this space.

It's related to other thing, which is asking the OS to allocate free ports from the ephemeral range to reduce port collisions, which seem to be a frequent source of transient test failures.

jdom:
Thanks a lot @veikkoeeva, I'm glad you were able to finish this. Cleanup is lower priority, but the bulk of the fixes are finally in! Great work, and you were very patient with our flaky test infrastructure :)

PR #3164, opened 2017-06-23
Flags PostgreSQL as synchronous to avoid locking in DB thread

veikkoeeva:
This allocates and extra task so as not to block the calling
thread and Orleans scheduler if the ADO.NET library has synchronous
functionality. This shouldn't have a major performance impact.

Fixes https://github.com/dotnet/orleans/issues/3141


jdom:

jdom:
Thanks @veikkoeeva, this seems like a safe approach :)

PR #3175, opened 2017-06-26
Avoid default port for localhost silo to conflict with Ants Profiler

### LINKED ISSUE #1663: Default port 22222 conflicts with RedGate Ants Profiler Service
Given the error for port conflict is the totally unintuitive "An attempt was made to access a socket in a way forbidden by its access permissions" and that it is likely that other developers will use Orleans and have RedGate's Ants Profiler installed, should the default ports in the project templates be changed to avoid conflict?

I should also add that I spent several hours figuring why TestingSiloHost was ignoring port for primary silo specified in xml file and using hardcoded port 22222 until I finally found that I had to explicitly pass a TestingSiloOptions to constructor overriding BasePort. It would be nice if other devs didn't have to go through that. I'm thinking that [unit testing documentation page](http://dotnet.github.io/orleans/Step-by-step-Tutorials/Unit-Testing-Grains) should explicitly say that xml configuration is ignored.

raol:
If I understand the problem correctly, all that needs to be done is
changing the siloPort from 22222 to any other. Decided to use 33333.
Fixes #1663

### DIFF
--- a/src/Orleans/Configuration/ClusterConfiguration.cs
+++ b/src/Orleans/Configuration/ClusterConfiguration.cs
@@ -569,7 +569,7 @@ namespace Orleans.Runtime.Configuration
         /// <param name="siloPort">TCP port for silo to silo communication</param>
         /// <param name="gatewayPort">Client gateway TCP port</param>
         /// <returns>ClusterConfiguration object that can be passed to Silo or SiloHost classes for initialization</returns>
-        public static ClusterConfiguration LocalhostPrimarySilo(int siloPort = 22222, int gatewayPort = 40000)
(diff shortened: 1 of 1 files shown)


sergeybykov:
@raol, thank you for your effort! 

After some deliberation, we think it's probably not a good idea to change the default port. It is trivial to override the default. While it does require an extra line of code, changing the default to 33333 could potentially impact existing users. And there's no guarantee that 33333 won't conflict with something else.

raol:
Closing then.
@sergeybykov I'm not sure, but maybe it is worth closing the referenced issue too.

sergeybykov:
@raol Thanks again!

PR #3176, opened 2017-06-26
Only allow InconsistentStateException to deactivate the current activation

ReubenBond:
As pointed out by @jdom, if a grain allows an `InconsistentStateException` to escape, then every grain in the call chain will be deactivated.

This fixes that by introducing a flag onto the exception itself. The flag is reset when the exception first passes through the invoker and the current activation is deactivated.


jdom:

jdom:

PR #3178, opened 2017-06-27
Service Fabric: make cluster change notifications determinsitic

ReubenBond:
There is some flakiness in the `HandlesSiloAdditionAndRemoval` test which was caused by non-deterministic execution order of the notifications since we started scheduling those on the task scheduler (see #3059). 

Tests use the default task scheduler (multi-threaded, no strict ordering), real deployments use single-threaded task scheduler. So this PR tightens up the ordering of notifications and fixes a few other issues:
* Starts an async worker using the captured task scheduler to process all notifications. The worker awaits an `AutoResetEvent` and pulls notifications from a queue and delivers them in order.
* `KillMyself` is always called to destroy the oracle, and it's always called last, so finalization logic has been moved from `Stop` and `Shutdown`.
* #3061 introduced an issue whereby subscribers would be re-notified of dead silos, this fixes that issue by ensuring that notifications for dead silos are only sent when the silo transitions from active to dead.

This code would be simplified by using an `ActionBlock<StatusChangeNotification>` from Dataflow, but I am averse to pulling that into this code. If we feel it's fine to add that dependency, then I can go ahead and use it.

### DIFF
--- a/src/OrleansServiceFabricUtils/FabricMembershipOracle.cs
+++ b/src/OrleansServiceFabricUtils/FabricMembershipOracle.cs
@@ -212,7 +212,7 @@ namespace Microsoft.Orleans.ServiceFabric
             this.fabricServiceSiloResolver.Subscribe(this);
             this.RefreshAsync().Ignore();
             this.UpdateStatus(SiloStatus.Joining);
-            return Task.FromResult(0);
+            return Task.CompletedTask;
         }
 
         /// <summary>
@@ -221,7 +221,7 @@ namespace Microsoft.Orleans.ServiceFabric
         public Task BecomeActive()
         {
             this.UpdateStatus(SiloStatus.Active);
-            return Task.FromResult(0);
+            return Task.CompletedTask;
         }
 
         /// <summary>
@@ -230,7 +230,7 @@ namespace Microsoft.Orleans.ServiceFabric
         public Task ShutDown()
         {
             this.UpdateStatus(SiloStatus.ShuttingDown);
-            return Task.FromResult(0);
+            return Task.CompletedTask;
         }
 
         /// <summary>
@@ -239,7 +239,7 @@ namespace Microsoft.Orleans.ServiceFabric
         public Task Stop()
         {
             this.UpdateStatus(SiloStatus.Stopping);
-            return Task.FromResult(0);
+            return Task.CompletedTask;
         }
 
         /// <summary>
@@ -249,7 +249,7 @@ namespace Microsoft.Orleans.ServiceFabric
         {
             this.UpdateStatus(SiloStatus.Dead);
             this.StopInternal();
-            return Task.FromResult(0);
+            return Task.CompletedTask;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


jdom:

jdom:
>             return Task.FromResult(0);

nit: Task.CompletedTask

---
Refers to: src/OrleansServiceFabricUtils/FabricMembershipOracle.cs:215 in 14ff254. [](commit_id = 14ff254b6dbe6fbe067d34361849bd725e0b1e61, deletion_comment = False)

jdom:
>             return Task.FromResult(0);

There's many of these in the file

---
Refers to: src/OrleansServiceFabricUtils/FabricMembershipOracle.cs:224 in 14ff254. [](commit_id = 14ff254b6dbe6fbe067d34361849bd725e0b1e61, deletion_comment = False)

src/OrleansServiceFabricUtils/Utilities/WaitHandleExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+            if (handle.WaitOne(0))
+                return Task.FromResult(true);
+
+            // If the timeout is zero, return synchronously.
+            if (timeout == TimeSpan.Zero)
+                return Task.FromResult(false);
+            
+            // Register a delegate on the thread pool to execute when the handle is set.
+            var tcs = new TaskCompletionSource<bool>();
+            ThreadPool.RegisterWaitForSingleObject(
```
jdom: What's the purpose of using this extension in conjunction with a WaitHandle, instead of just relying on a TCS? This way you skip the wait handle, the extra blocked thread, etc

jdom: Hmmm, maybe `RegisterWaitForSingleObject` is not very costly (I wasn't familiar with that API), probably doesn't create threads until the wait handle is triggered.
In that case this might be fine to keep as is.

ReubenBond: This approach is used in the AsyncEx library. Technically speaking we could replace the usage of this with `Task.Yield()` or `Task.Delay(100)` and it would be fine, but this seemed like the cleaner approach.

ReubenBond:
Addressed feedback, thanks

jdom:

src/OrleansServiceFabricUtils/FabricMembershipOracle.cs
```diff
@@ -366,25 +373,30 @@ private void ClearCaches()
...
         }
 
-        private void NotifySubscribers(SiloAddress address, SiloStatus newStatus, List<ISiloStatusListener> listeners)
+        private async Task ProcessNotifications()
         {
-            Task.Factory.StartNew(() =>
+            while (!this.notifications.IsAddingCompleted)
             {
-                foreach (var subscriber in listeners)
+                await this.notificationEvent.WaitAsync(TimeSpan.FromMinutes(1));
```
jdom: What's the purpose of setting the 1 minute timeout? Why not just infinite? Or are could there be work to do even if no one sets the event?

ReubenBond: Infinite might work, but the timeout is benign and it's more 'obviously correct' to me. So this task will wake up every minute to explicitly check if either:
a) some notification needs processing, or
b) the silo has shutdown and so it can terminate

Issue #3184, opened 2017-06-29
Allow reentrancy within a grain call chain


sergeybykov:
There've been multiple request for this narrow deadlock prevention feature - to allow reentrancy for calls to grains that are already part of the call chain. For example, if grain A calls grain B, and while executing the invoked method B calls back to A, today that B->A call will deadlock and eventually time out, unless A is marked as `[Reentrant]`. The proposal here is to allow the B->A call to execute (while A is awaiting its call to B) even if A is not marked as `[Reentrant]`.

The argument for this change is that reentrancy within a single call chain is relatively safe compared for reentrant parallel execution of multiple independent requests, and that the benefit of avoiding deadlocks within a call chain would save a significant amount of developer's cognitive effort of avoiding cycles in call chains, some of which may not be obvious, and finding the root causes of deadlock timeout happening at run time.

If we were to implement this, we could make this behavior either the default or an explicit opt-in via a configuration switch. The opt-in options appears safer because if doesn't change the default behavior. However, if we made this the default, the only cases that would be potentially impacted are those that would cause a deadlock today anyway. So it seems to me that making reentrancy within a call chain the default behavior is actually pretty safe, even though it would technically be a breaking change.

We would need to pass additional information with every message, which could have a performance impact. That's a potential argument for opt-in, unless we find no measurable performance difference.

gabikliot:
I think this is already implemented in our code base, just disabled.
We have deadlock detection phase one implemented and it does what u described above. Upon detecting a cyclic deadlock, it now throws deadlock exc. In your proposal, we will change it to allow concurrency.
Unless we deleted that code when we went open source. But I think it is still there.

:-)
Isn't that nice?

gabikliot:
https://github.com/dotnet/orleans/blob/2af52f7401b6f06b6c7f3d100349b889478591e0/src/OrleansRuntime/Core/InsideRuntimeClient.cs#L285
and we even have a test.

dVakulen:
@gabikliot  Just wanted to write that the part is that there's already tests for this.

sergeybykov:
@gabikliot Nice! I completely forgot about the deadlock detection feature that's been disabled for years, and didn't think we could reuse it. IIRC at the time there was a suspicious performance drop because of which we never enabled the feature. Need to retest now.

gabikliot:
[...]

PR #3185, opened 2017-06-30
Allow reentrancy within a grain call chain

dVakulen:
Addresses https://github.com/dotnet/orleans/issues/3184

Not configurable for now. 
In addition allows grain calls to itself. 
Forces creation of ```RequestContext``` dictionary + sometimes List object for message, so the performance drop should be visible, but as this only requires existence of two ```ActivationId``` properties in Message - there's big room for easy optimization. 

As usually - perf tests should be ran before making following decisions.


src/Orleans/Messaging/RequestInvocationInfoAccessor.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public void AddInvokationInfo(Message message)
+        {
+            if (message.TargetGrain.IsSystemTarget)
+            {
+                return;
+            }
+
+            // assumes deadlock information was already loaded into RequestContext from the message
+            var thisInvocation = Config.Globals.PerformDeadlockDetection ? new RequestInvocationHistory(message) : new RequestInvocationInfo(message);
```
dVakulen: Assumes knowledge of the callee, will rewrite

src/Orleans/Messaging/RequestInvocationInfoAccessor.cs
```diff
@@ -0,0 +1,75 @@
...
+            var list = obj as IList;
+            if (list != null)
+            {
+                invokationInfoList = list;
+                return true;
+            }
+
+            // only one info entry in request context, deadlock is impossible,
+            // as call to itself is now allowed
+            return false;
```
dVakulen: Another assupmtion, will rewrite

src/Orleans/Messaging/RequestInvocationHistory.cs
```diff
@@ -2,24 +2,34 @@
 
 namespace Orleans.Runtime
 {
+    [Serializable]
+    internal class RequestInvocationInfo
```
dVakulen: RequestInvocationHistory contains unnecessary ```GrainId``` and ```DebugContext``` properties, and the latter one may consume pretty large partof serialized message size.

dVakulen: Also optimizes common case of single grain - > grain call.

dVakulen:

dVakulen:
The performance tests has shown drops from 15% on NightlyLoadTest up to 50% on PingLoadTest_RandomReentrant_MultiSilos. They are due to creation of ```RequestContext``` dictionary. 
But, as this is meant to support only specific cases such as  A -> A, and A -> B -> A call chains - only two additional ActiovationIds needs to be present in message headers. So they can be stored as two new properties of ```HeadersContainer``` object, thus only additional cost will be on transporting of them.

ReubenBond:
@dVakulen a single ActivationId ought to be able to support call chains of arbitrary length / complexity as long as we temporarily record that ActivationId on each activation in a given call chain. We don't need a list, just a single id. It would be better if it wasn't the ActivationId, but some other correlation id (like the `ActivityId` of `RequestContext`, depending on how expensive that is to use). That way a single re-entrant grain doesn't make all grains it calls reentrant. Eg,
[Reentrant] GrainA calls regular GrainB, and while GrainB is processing that request, reentrant GrainA calls it again.

So we could use a single identifier, generated on the first activation which a call is destined for.

dVakulen:
@ReubenBond Current implementation doesn't make  grains called by re-entrant grain reentrant either. 

>temporarily record that ActivationId\CorrelationId on each activation in a given call chain. . We don't need a list, just a single id.

But what about fan-out calls case?

ReubenBond:
@dVakulen What I mean is that a solution involving a single id may be more performant without sacrificing correctness.

Any fan-out must originate from a single grain, so the id would be propagated to all subsequent calls. The id could be set when a grain first receives a message if that message does not already specify an id.

gabikliot:
Reuben, in that protocol with a single id, we will also need to locally record the IDs that passed via each activation along the chain, and obviously also delete them.
Was that your idea?

The msg can be tacked on the client. And on the silo, we already have a global dictionary of all current in flight msgs. This is callback data. But those are keyed by one hop id. So u will need another dictionary, by "chain I'd". But you can cleanup in the same time as cleaning callback datas.

Yes, seems like this can work. Nice idea!

gabikliot:
You obviously loose the details of the cycle, which was part of the goal of the original deadlock detection feature. But for reentrancy, we don't need the details.

gabikliot:
Not supporting compatibility is not an option in my opinion. Whatever solution is chosen, compatibility (mixed version deployment) should be supported. 
And it can be supported, with any of the options. In the original protocol it would mean not participating in the deadlock detection if the server is of an older version and making sure mixed version does not break anything else.
It might be easier to build backward compat version with the 2nd protocol, if you don't introduce a new header. In the 1st protocol the messaging stack should be ready to deal with a new, unrecognized header and pass it through.

sergeybykov:
>There is a benefit to keeping this information in the request context as opposed to moving it to the message headers directly: it wouldn't break compatibility with existing deployments. Maybe it's worth breaking compatibility, though

I'm not sure I understand the compatibility concerns here. My thought were similar - we check the header of the message we process for the call chain ID (instead of the activation ID or the whole call history), and generate a new one (GUID?) if we can't find it. We keep it in memory associated with the request we are processing and put a copy into any outgoing message. `RequestContext` is the obvious place, but the performance implications of using it make the header alternative attractive.

Where does compatibility get violated? The new header would be completely transparent for any code that isn't aware of it. So what's the problem?

ReubenBond:
@sergeybykov my bad: the serialization of the `HeadersContainer` isn't so rigid that it cannot be modified without breaking compat - so there's no issue here. There's room for a few more fields there.

dVakulen:
Implementation with single `CallChainId` in `Headers` shows the following performance results: 

 **This PR:**

| Test | TPS Moving Avg | 
|--|--|
| NightlyLoadTest | 182857.7 |
| ActivationCollectorStressTest | 212775.5 |
| PingLoadTest_LocalReentrant | 115425.3 |
| PingLoadTest_RandomReentrant_MultiSilos | 272430.3 |
Commit: 51c08fccaf7a162ef9f0cb94852ea92b6a32b9de

**MASTER:** 

| Test | TPS Moving Avg | 
|--|--|
| NightlyLoadTest | 188474.4 |
| ActivationCollectorStressTest | 207271.1 |
| PingLoadTest_LocalReentrant | 115394.7 |
| PingLoadTest_RandomReentrant_MultiSilos | 268381.9 |

Commit: 95cf545d0a86adebbfbc5810279d0f00d769dfcf

Some perf drop in NightlyLoadTest  may be just test fluctuation.

ReubenBond:
@dotnet-bot test functional

ReubenBond:

ReubenBond:
@dotnet-bot test functional

sergeybykov:
@dotnet-bot test functional

sergeybykov:
I see 3 test failures:
>UnitTests.StreamingTests.SMSStreamingTests.SMS_09_ManySame_ManyProducerGrainsManyConsumerGrains
UnitTests.StreamingTests.SMSStreamingTests.SMS_10_ManySame_ManyConsumerGrainsManyProducerGrains
UnitTests.StreamingTests.SMSStreamingTests.SMS_12_ManySame_ManyProducerClientsManyConsumerGrains

All caused by
>Orleans.Storage.InconsistentStateException : e-Tag mismatch in Memory Storage

which appears related to reentrancy. The question I have here is are these test issues or the Memory Storage provider needs to be adjusted/configured or there's an issue with this change?

dVakulen:
I'll take a look at these test failures tomorrow.

dVakulen:
The test fails were fixed. Implemented ~ zero cost [solution](https://github.com/dotnet/orleans/pull/3185/files#diff-05e8ff586dec9c6e8cc5310a9a07508dR339), which covers  A -> A, and A -> B -> A cases.

src/Orleans.Runtime/Catalog/ActivationData.cs
```diff
@@ -267,6 +267,8 @@ internal bool TryGetExtensionHandler(Type extensionType, out IGrainExtension res
 
         #endregion
 
+        public HashSet<GrainId> RunningRequestsSenders { get; } = new HashSet<GrainId>();
```
dVakulen: For simplicity duplicate senders tracking wasn't implemented. Can be added in case of need.

src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -325,14 +325,31 @@ private bool ActivationMayAcceptRequest(ActivationData targetActivation, Message
         public bool CanInterleave(ActivationData targetActivation, Message incoming)
         {
             bool canInterleave = 
-                   catalog.CanInterleave(targetActivation.ActivationId, incoming)
-                || incoming.IsAlwaysInterleave
+                   incoming.IsAlwaysInterleave
                 || targetActivation.Running == null
-                || (targetActivation.Running.IsReadOnly && incoming.IsReadOnly);
+                || (targetActivation.Running.IsReadOnly && incoming.IsReadOnly)
+                || targetActivation.ActivationId.Equals(incoming.SendingActivation)
```
dVakulen: For A -> A case. Also checks were rearranged accordingly to their costs.

src/Orleans.Runtime/Catalog/ActivationData.cs
```diff
@@ -477,6 +479,11 @@ public void RecordRunning(Message message)
             // Note: This method is always called while holding lock on this activation, so no need for additional locks here
 
             numRunning++;
+            if (message.Direction != Message.Directions.OneWay && message.SendingGrain != null)
```
ReubenBond: Should we add a check for `SendingGrain.IsClient`? We cannot reasonably expect that two calls from the same client are in the same call chain. We could support call chains from the client, but we would need to have some other kind of identifier (eg, client messages could stick a unique id into the request context) - but that adds complexity.

So for now we should consider just avoiding adding clients here.

ReubenBond:
LGTM after the last comment is addressed (`IsClient` check)

src/Orleans.Runtime/Catalog/ActivationData.cs
dVakulen: Agree, check added.

src/Orleans.Runtime/Catalog/ActivationData.cs
```diff
@@ -477,6 +479,13 @@ public void RecordRunning(Message message)
             // Note: This method is always called while holding lock on this activation, so no need for additional locks here
 
             numRunning++;
+            if (message.Direction != Message.Directions.OneWay 
+                && message.SendingGrain != null
+                && !message.SendingGrain.IsClient)
+            {
+                RunningRequestsSenders.Add(message.SendingGrain);
```
ReubenBond: I think this should be `SendingActivation`, instead. It's a corner case, but we could see multiple activations with the same `GrainId` (i.e, during partial failure scenarios).

ReubenBond:
Thinking more deeply about this, I'm concerned that we don't have sufficient test coverage.

A couple of specific scenarios which come to mind:
* Fan-out with cross-calls
  * We have non-reentrant grains `A`, `B`, & `C`.
  * `A` initiates a call to `B` & `C` simultaneously. `B` and `C` call each-other before returning to `A`.
  * These calls are in the same call chain (mutual reentrancy into `B` & `C` should be permitted here)
* `[Reentrant]` grain calls non-reentrant grain
  * `A` is reentrant and `B` is not
  * `A` receives multiple requests from external source and calls `B` which performs some async database work
  * These calls are not in the same call chain: reentrancy into `B` should not be permitted for multiple concurrent calls from `A` to `B`

Are those cases covered by this PR? It's understandable if the first case is not supported (deadlock), but the second one is a question of safety: `B` has no reason to expect reentrant calls since it's not making any grain calls.

The general strategy with the `RunningRequestsSenders` seems right to me, but I think the id stored in that value should be unique for the call chain rather than fixed & per-sender.

ReubenBond:
@dotnet-bot test bvt

ReubenBond:
@dotnet-bot test functional

sergeybykov:
@dotnet-bot test functional

sergeybykov:
@dotnet-bot test functional

dVakulen:
Added `AllowCallChainReentrancy` option to `GlobalConfiguration`.

Docs will be updated after merge.

ReubenBond:
@dVakulen I added two commits:
* To disable this by default in order to be conservative. We can enable it by default once everyone is back from their holidays
* To migrate to the IOptions pattern.

Does this sound fine to you?

ReubenBond:
@dotnet-bot test functional

dVakulen:
@ReubenBond  Yes, it does, thank you.

PR #3182, opened 2017-06-28
Make ReconstructClientIdPartitionTest_Request test more reliable

jdom:
Increase ProbeTimeout to reduce chances of failures in slower machines.
Speed up test setup by restarting client instead of entire cluster.

PR #3192, opened 2017-07-03
Commits cherry-picked for 1.5.0 final

sergeybykov:


jdom:
https://github.com/dotnet/orleans/pull/2856 should be included too. We finally reached consensus

jdom:
Please include #3140 and #3182. There're just test fixes, but they are more reliable in case we need to create patches for it too, not just the current release.

sergeybykov:
Added #3140 and #3182.
The consensus about #2856 is not to rush it to 1.5.0. :-)

PR #3194, opened 2017-07-04
Add compatibility director and version selector change in OrleansManager

benjaminpetit:
Only works with built-in director and selector. If the developer want to implement his own, he will need to build his own tool to support them

### DIFF
--- a/src/Orleans/Versions/Selector/IVersionSelector.cs
+++ b/src/Orleans/Versions/Selector/IVersionSelector.cs
@@ -16,5 +16,15 @@ namespace Orleans.Versions.Selector
     [Serializable]
     public abstract class VersionSelectorStrategy
     {
+        public static VersionSelectorStrategy Parse(string str)
+        {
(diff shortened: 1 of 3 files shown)


src/Orleans/Versions/Selector/IVersionSelector.cs
```diff
@@ -16,5 +16,15 @@ public interface IVersionSelector<TPolicy> : IVersionSelector where TPolicy : Ve
...
     {
+        public static VersionSelectorStrategy Parse(string str)
+        {
+            if (str.Equals(typeof(AllCompatibleVersions).Name))
+                return AllCompatibleVersions.Singleton;
+            if (str.Equals(typeof(LatestVersion).Name))
+                return LatestVersion.Singleton;
+            if (str.Equals(typeof(MinimumVersion).Name))
+                return MinimumVersion.Singleton;
+            return null;
```
jdom: Does this mean you can only change the strategy at runtime if you use one of the built-in ones?

PR #3196, opened 2017-07-05
ILeaseProvider interface

xiazen:
The ILeaseProvider interface. Which is optional to be included in 1.5

### DIFF
--- a/src/Orleans/Providers/ILeaseProvider.cs
+++ b/src/Orleans/Providers/ILeaseProvider.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Orleans.LeaseProviders
+{
+    /// <summary>
+    /// Aquired lease, which contains the resource key, duration of the lease and the token you can use to renew the lease with
+    /// </summary>
+    public class AcquiredLease
+    {
+        public string ResourceKey { get; }
+        public TimeSpan Duration { get; }
+        public string Token { get; }
+    }
+
+    /// <summary>
+    /// Lease request where you can specify ResourceKey, which you want to apply a lease on, and duration of your lease. 
+    /// </summary>
+    public class LeaseRequest
+    {
+        public string ResourceKey { get; set; }
+        public TimeSpan? Duration { get; set; }
+    }
+
+    /// <summary>
+    /// Lease provider interface 
+    /// </summary>
+    public interface ILeaseProvider
+    {
+        Task<AcquiredLease[]> Acquire(string category, LeaseRequest[] requestedLeases);
+        Task<AcquiredLease[]> Renew(string category, AcquiredLease[] aquiredLeases);
+        Task Release(string category, AcquiredLease[] aquiredLeases);
+    }
+
+}
--- a/src/Orleans/Providers/ILeaseProvider.cs
+++ b/src/Orleans/Providers/ILeaseProvider.cs
@@ -24,7 +24,7 @@ namespace Orleans.LeaseProviders
         /// </summary>
         public string Token { get; }
         /// <summary>
-        /// Start time for this lease, which is when the lease is acquired or renewed
+        /// Caller side start time for this lease, which is when the lease is acquired or renewed
         /// </summary>
         public DateTime StartTimeUtc { get; }
     }
@@ -39,15 +39,35 @@ namespace Orleans.LeaseProviders
         /// </summary>
         AcquiredLease AcquiredLease { get; }
         /// <summary>
-        /// Whether successfully acquire or renew the lease
+        /// Response status
         /// </summary>
-        public bool Success { get; }
+        public ResponseCode StatusCode { get; }
         /// <summary>
         /// If acquiring or renewing the lease failed, this is the exception which caused it. This field would be null if operation succeed.
         /// </summary>
         public Exception FailureException { get; }
     }
 
+    public enum ResponseCode
+    {
+        /// <summary>
+        /// Operation succeed
+        /// </summary>
+        OK,
+        /// <summary>
+        /// Lease is owned by other entity
+        /// </summary>
+        LeaseNotAvailable,
+        /// <summary>
+        /// The token in the AcquiredLease is invalid, which means the lease expired
+        /// </summary>
+        InvalidToken,
+        /// <summary>
+        /// TransientFailure, which should be retriable. 
+        /// </summary>
+        TransientFailure
+    }
+    
     /// <summary>
     /// Lease request where you can specify ResourceKey and duration of your lease. 
     /// </summary>
--- a/src/Orleans/Providers/ILeaseProvider.cs
+++ b/src/Orleans/Providers/ILeaseProvider.cs
@@ -9,7 +9,7 @@ namespace Orleans.LeaseProviders
     /// <summary>
     /// Acquired lease
     /// </summary>
-    public class AcquiredLease
+    public class AcquiredLease 
     {
         /// <summary>
         /// The resource key which the lease is attached to 
@@ -24,17 +24,28 @@ namespace Orleans.LeaseProviders
         /// </summary>
         public string Token { get; }
         /// <summary>
+        /// Start time for this lease, which is when the lease is acquired or renewed
+        /// </summary>
+        public DateTime StartTimeUtc { get; }
+    }
+
+    /// <summary>
+    /// AcquireLeaseResult class, which demonstrates result of acquiring or renewing lease operation
+    /// </summary>
+    public class AcquireLeaseResult
+    {
+        /// <summary>
+        /// Acquired lease, which will be null if acquire or renew operation failed.
+        /// </summary>
+        AcquiredLease AcquiredLease { get; }
+        /// <summary>
         /// Whether successfully acquire or renew the lease
         /// </summary>
         public bool Success { get; }
         /// <summary>
-        /// If acquiring or renewing the lease failed, this is the exception which caused it
+        /// If acquiring or renewing the lease failed, this is the exception which caused it. This field would be null if operation succeed.
         /// </summary>
         public Exception FailureException { get; }
-        /// <summary>
-        /// Start time for this lease, which is when the lease is acquired or renewed
-        /// </summary>
-        public DateTime StartTimeUtc { get; }
     }
 
     /// <summary>
@@ -49,7 +60,7 @@ namespace Orleans.LeaseProviders
         /// <summary>
         /// Duration of the lease
         /// </summary>
-        public TimeSpan? Duration { get; set; }
+        public TimeSpan Duration { get; set; }
     }
 
     /// <summary>
@@ -63,14 +74,14 @@ namespace Orleans.LeaseProviders
         /// <param name="category">resource category</param>
         /// <param name="leaseRequests"></param>
         /// <returns>Lease acquiring results array, whose order is the same with leaseRequstes</returns>
-        Task<AcquiredLease[]> Acquire(string category, LeaseRequest[] leaseRequests);
+        Task<AcquireLeaseResult[]> Acquire(string category, LeaseRequest[] leaseRequests);
         /// <summary>
         /// Batch renew lease operation
         /// </summary>
         /// <param name="category">resource category</param>
         /// <param name="aquiredLeases"></param>
         /// <returns>Lease renew results array, whose order is the same with acquiredLeases</returns>
-        Task<AcquiredLease[]> Renew(string category, AcquiredLease[] aquiredLeases);
+        Task<AcquireLeaseResult[]> Renew(string category, AcquiredLease[] aquiredLeases);
         /// <summary>
         /// Batch release lease operation
         /// </summary>
--- a/src/Orleans/Providers/ILeaseProvider.cs
+++ b/src/Orleans/Providers/ILeaseProvider.cs
@@ -7,21 +7,48 @@ using System.Threading.Tasks;
 namespace Orleans.LeaseProviders
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Orleans.LeaseProviders
+{
+    /// <summary>
+    /// Aquired lease, which contains the resource key, duration of the lease and the token you can use to renew the lease with
+    /// </summary>
+    public class AcquiredLease
+    {
+        public string ResourceKey { get; }
+        public TimeSpan Duration { get; }
+        public string Token { get; }
```
attilah: What do you think about adding a LeaseAcquired DateTime or DateTimeOffset (UTC) property to this class?

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string ResourceKey { get; set; }
+        public TimeSpan? Duration { get; set; }
+    }
+
+    /// <summary>
+    /// Lease provider interface 
+    /// </summary>
+    public interface ILeaseProvider
+    {
+        Task<AcquiredLease[]> Acquire(string category, LeaseRequest[] requestedLeases);
```
attilah: What about ```leaseRequests``` as parameter name to be conform with the type name?

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        Task<AcquiredLease[]> Renew(string category, AcquiredLease[] aquiredLeases);
```
ReubenBond: How is a failure represented in these return values? We should have comments describing the API, eg that the elements in the output must correspond to the elements in the input array.

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        Task Release(string category, AcquiredLease[] aquiredLeases);
```
attilah: How do we thinking of handling at API level, if one of the Acquire, Renew or Release operations failing when passing in multiple items? first throws, others are ignored?
Can that be useful to include a result field for these operations together with the results per item to make it more exception and multiple item friendly?

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Orleans.LeaseProviders
+{
+    /// <summary>
+    /// Aquired lease, which contains the resource key, duration of the lease and the token you can use to renew the lease with
```
attilah: Typo: 'c' is missing from Acquired.

src/Orleans/Providers/ILeaseProvider.cs
attilah: See my last comment, I just asked the same about the API design ;-)

xiazen: I was thinking only return successful leases. But if the API pattern you are suggesting is more common, yes lets do that

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+    public class LeaseRequest
+    {
+        public string ResourceKey { get; set; }
+        public TimeSpan? Duration { get; set; }
+    }
+
+    /// <summary>
+    /// Lease provider interface 
+    /// </summary>
+    public interface ILeaseProvider
```
jason-bragg: Do we want some sort of RequestFreeResources call that returns a list of resource keys that have expired or been released within a category?

src/Orleans/Providers/ILeaseProvider.cs
jason-bragg: In both the Acquire and Renew calls, returning a class called "AcquiredLease" gives the impression that only successfully owned leases will be returned.  This conforms to Xiazen's statement regarding the expected behavior.  If the results are to match, one to one, with the requests, then we'll need to rename AcquiredLease to something more indicative of the possibility of failure, or have it inherit from a base class, which AcquiredLease and some sort of RejectedLease class both derive from.

src/Orleans/Providers/ILeaseProvider.cs
xiazen: That would be good. I can see it  useful in some scenario.

xiazen: GetFreeResources(string category) maybe a better name

ReubenBond: That would but a significant extra burden on implementations - is it really needed? Should we have some kind of `IQueryableLeaseProvider` if it's needed?

xiazen: I can see it useful on debugging and monitoring, or deal with some edge case in queue balancing. 

There will be some extra burden on implementation. Why do you think it is significant ?

jason-bragg: As Reuben expressed (in other words), it can always be added as a natural extension to the current feature set.  So if debatable, we should exclude.  Just wanted to raise the question.

xiazen: And LeaseProvider manages leases,  so I think it is natural for it to have a query operation. So did I overthink ?

xiazen: I'm not against the natural extension of IQueryableLeaseProvider. Just want to voice my concern that ILeaseProvider and its interface implies it manages lease operation, so I think it is natural for it to have a query operation. But if majority of people think the other way, I will go with that

ReubenBond: It's a worthwhile question to raise. This lease manager does not currently need to know the total set resources it manages. If we had an API which queried which leases are free, it would need some way of enumerating them. Would that API return a consistent view of the leases or would it query one-by-one and hence return a torn view? Maybe the answer is the latter (which is fine), but questions like that would need to be answered.

I'm not strongly against it, but just wanted to make sure that we determined it was worth it and that we've considered the nuances.

sergeybykov: >And LeaseProvider manages leases, so I think it is natural for it to have a query operation. Or did I overthink ?

Not necessarily. I think querying functionality will inevitably harm or complicate scalability. And debugging isn't a good justification for that to me. The extension interface path seems like a better approach. Even there I would be cautious about querying.

xiazen: But scalability is a implementation detail, which shouldn't be affecting API design, right? API design is expressing the caller side's demand. 
But I do agree that API should only contain the minimum set of functionality which are needed . Extra functionality needed on minor cases should come with extension on top of the base API. And I don't see query api is a must have for leaseProvider. So I think we reached the concensus to not include query API. 

BTW, I didn't include it in this PR since we were still discussing it. Just so you know that I addressed all the comments, and the PR is ready for a new round of review.

xiazen:

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// Duration of the aquired lease
+        /// </summary>
+        public TimeSpan Duration { get; }
+        /// <summary>
+        /// Lease token, whcih will be null if acquiring or renewing the lease failed
+        /// </summary>
+        public string Token { get; }
+        /// <summary>
+        /// Start time for this lease, which is when the lease is acquired or renewed
```
jason-bragg: I wonder if we should note that this time is in the callers time, so implementers don't assign it with the server side time.

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        AcquiredLease AcquiredLease { get; }
+        /// <summary>
+        /// Whether successfully acquire or renew the lease
+        /// </summary>
+        public bool Success { get; }
+        /// <summary>
+        /// If acquiring or renewing the lease failed, this is the exception which caused it. This field would be null if operation succeed.
+        /// </summary>
+        public Exception FailureException { get; }
```
jason-bragg: Do we want to define exceptions as part of the public api.  I ask because a timeout exception on a renew does not mean that the lease has been lost, so renew can be retried.  Similarly, a timeout or storage failure on an acquire would also indicate it can be retried.  Seems like exceptions should fall into LeaseNotAvailableException meaning the lease is owned by another (should not retry), NotLeaseOwnerException which would indicate we're trying to renew or release a lease we no longer own (should not be retried), and LeaseActionFailedException which is a catch all for all other lease errors like timeout, storage, and the like (can be retried).
Without well defined exceptions, I don't think there is enough information in the results for uses to correctly handle errors.

xiazen: good point. With that being said, I think it may better to merge Exception and Success fields into one field, which is some sort of response state. Since exception and success are both a subset of response state. For example, RESTful API handle this situation with response code, 200 meaning succeed, 403 meaning forbidden.

xiazen: I guess there's two options to deal with this: one is more REST style, which it is what azure storage api did : https://docs.microsoft.com/en-us/dotnet/api/microsoft.windowsazure.storage.table.cloudtable.endexecutebatch?view=azurestorage-8.1.3
 ```

    public class AcquireLeaseResult
    {
        /// <summary>
        /// Acquired lease, which will be null if acquire or renew operation failed.
        /// </summary>
        AcquiredLease AcquiredLease { get; }
        /// <summary>
        /// Whether successfully acquire or renew the lease
        /// </summary>
        public int StatusCode { get; }
        /// <summary>
        /// If acquiring or renewing the lease failed, this is the exception which caused it. This field would be null if operation succeed.
        /// </summary>
        public Exception FailureException { get; }
    }
```

Another one is more OOP pattern which Jason suggested : 
```
    public class AcquireLeaseResult2
    {
        /// <summary>
        /// Acquired lease, which will be null if acquire or renew operation failed.
        /// </summary>
        AcquiredLease AcquiredLease { get; }

        /// <summary>
        /// If acquiring or renewing the lease failed, this is the exception which caused it. This field would be null if operation succeed.
        /// </summary>
        public LeaseException FailureException { get; }
    }

    public class LeaseException : Exception
    { }
    public class LeaseNotAvailableException : LeaseException
    { }
    public class LeaseNotOwnedException : LeaseException
    { }
    public class LeaseActionFailedException : LeaseException
    { }
```

Comment (xiazen) [inline src/Orleans/Providers/ILeaseProvider.cs]: I don't have a big preference, just want to throw options out here to see how people think.

Comment (xiazen) [inline src/Orleans/Providers/ILeaseProvider.cs]: After talking with Reuben, we are more in an agreement on this 

 ```
/// <summary>
    /// AcquireLeaseResult class, which demonstrates result of acquiring or renewing lease operation
    /// </summary>
    public class AcquireLeaseResult
    {
        /// <summary>
        /// Acquired lease, which will be null if acquire or renew operation failed.
        /// </summary>
        AcquiredLease AcquiredLease { get; }
        /// <summary>
        /// Response code
        /// </summary>
        public ResponseCode StatusCode { get; }
        /// <summary>
        /// If acquiring or renewing the lease failed, this is the exception which caused it. This field would be null if operation succeed.
        /// </summary>
        public Exception FailureException { get; }
    }

    public enum ResponseCode
    {
        /// <summary>
        /// Lease is owned by other entity
        /// </summary>
        LeaseNotAvailable,
        /// <summary>
        /// The token in the AcquiredLease is invalid, which means the lease expired
        /// </summary>
        InValidToken,
        /// <summary>
        /// Operation succeed
        /// </summary>
        Succeed,
        /// <summary>
        /// TransientFailure, which should be retriable. 
        /// </summary>
        TransientFailure
    }
```
StatusCode will be the main point which the caller side decide its flow control on. But still keeping the FailureException field, as an extensibility point to provide more information for diagnose. For example, if user build the LeaseProvider on top of azure blob, then we can pass in native azure blob exception in the FailureException field.

ReubenBond: LGTM, but prefer `Success` be first entry and change `InValid` to `Invalid`

jason-bragg: LG2M - I'd also suggest some sort of "ServiceFailure" code for unexpected errors like service bugs or misconfiguration.

xiazen: I included ServiceFailure or unexpected errors into TransientFailure. In the view of the caller, service failure should be categorized to be retriable and transient. Eg, fix the service bug or misconfiguration is the service side's responsibility, once it is fixed in the service, the caller can retry on the service again.

jason-bragg: @xiazen, I'll leave this to your discretion, but in my view transient errors, like storage throttling or timeouts, are those where we have a reason to believe that future calls can/should work.  For unforeseen errors like bugs or misconfigured storage accounts, we've no reason to believe that retrying the call would work, hence are not, in my view, transient.

xiazen: hmm, you are talking about caller side error, not service bug or service side mis config, which is out of the caller's reach to fix. I was confused by the word of "ServiceFailure"

Maybe we should call it ClientSideError?

jason-bragg: I'm talking about unexpected failures in the leasing service, not caller code.  If all unexpected failures are considered transient, and the user sees transient errors as errors that may work on retry, then there is no way for service implementers to communicate non-transient errors to the caller.  I'll merge anyway, as we can always add new error codes if we see them as needed.

PR #3204, opened 2017-07-07
Separate logic from base GrainReference

jdom:
In preparation to move `GrainReference` to an Abstractions assembly, but still keep the logic inside the Orleans implementation.
- `GrainReference` is now bound to `IGrainReferenceRuntime` instead of `IRuntimeClient`. The new runtime abstraction includes all the logic that was previously part of `GrainReference`, and it's a much more reduced surface area compared to `IRuntimeClient`, to expose to the public.
- Moved Orleans-specific serialization outside the base abstraction.

### DIFF
--- a/src/Orleans/Runtime/GrainReferenceRuntime.cs
+++ b/src/Orleans/Runtime/GrainReferenceRuntime.cs
@@ -1,4 +1,5 @@
 ﻿using Orleans.CodeGeneration;
+using Orleans.Serialization;
 using System;
 using System.Collections.Concurrent;
 using System.Text;
@@ -15,13 +16,19 @@ namespace Orleans.Runtime
         private readonly Action<Message, TaskCompletionSource<object>> responseCallbackDelegate;
         private readonly Logger logger;
         private readonly IInternalGrainFactory internalGrainFactory;
+        private readonly SerializationManager serializationManager;
 
-        public GrainReferenceRuntime(Factory<string, Logger> loggerFactory, IRuntimeClient runtimeClient, IInternalGrainFactory internalGrainFactory)
+        public GrainReferenceRuntime(
+            Factory<string, Logger> loggerFactory,
+            IRuntimeClient runtimeClient,
+            IInternalGrainFactory internalGrainFactory,
+            SerializationManager serializationManager)
         {
             this.responseCallbackDelegate = this.ResponseCallback;
             this.logger = loggerFactory.Invoke(nameof(GrainReferenceRuntime));
             this.RuntimeClient = runtimeClient;
             this.internalGrainFactory = internalGrainFactory;
+            this.serializationManager = serializationManager;
         }
 
         public IRuntimeClient RuntimeClient { get; private set; }
@@ -44,7 +51,7 @@ namespace Orleans.Runtime
             {
                 CheckForGrainArguments(arguments);
                 SetGrainCancellationTokensTarget(arguments, reference);
-                argsDeepCopy = (object[])this.RuntimeClient.SerializationManager.DeepCopy(arguments);
+                argsDeepCopy = (object[])this.serializationManager.DeepCopy(arguments);
             }
 
             var request = new InvokeMethodRequest(reference.InterfaceId, reference.InterfaceVersion, methodId, argsDeepCopy);
@@ -139,7 +146,7 @@ namespace Orleans.Runtime
             {
                 try
                 {
-                    response = (Response)message.GetDeserializedBody(this.RuntimeClient.SerializationManager);
+                    response = (Response)message.GetDeserializedBody(this.serializationManager);
                 }
                 catch (Exception exc)
                 {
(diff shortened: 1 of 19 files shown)


src/Orleans/Runtime/GrainReferenceRuntime.cs
```diff
@@ -0,0 +1,230 @@
...
+            if (arguments == null) return;
+            foreach (var argument in arguments)
+            {
+                (argument as GrainCancellationToken)?.AddGrainReference(target);
+            }
+        }
+
+        private bool IsUnordered(GrainReference reference)
+        {
+            return this.RuntimeClient.GrainTypeResolver?.IsUnordered(reference.GrainId.TypeCode) == true;
```
ReubenBond: This adds an additional dictionary lookup per call

src/Orleans/Runtime/IGrainReferenceRuntime.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <param name="options">Invocation options.</param>
+        /// <param name="silo">The target silo.</param>
+        /// <returns>Returns the response from the remote object.</returns>
+        Task<T> InvokeMethodAsync<T>(GrainReference reference, int methodId, object[] arguments, InvokeMethodOptions options, SiloAddress silo);
+
+        /// <summary>Casts the provided <paramref name="grain"/> to the specified interface.</summary>
+        /// <typeparam name="TGrainInterface">The target grain interface type.</typeparam>
+        /// <param name="grain">The grain reference being cast.</param>
+        /// <returns>A reference to <paramref name="grain"/> which implements <typeparamref name="TGrainInterface"/>.</returns>
+        TGrainInterface Cast<TGrainInterface>(IAddressable grain);
```
ReubenBond: Not something you've introduced here, but should we call this `Convert` instead of `Cast`? I've seen that people confuse casting, considering it a no-op type change.

src/Orleans/Runtime/GrainReferenceRuntime.cs
jdom: Yes, was planning to do a perf run. Although you think it could impact perf at all? It's O(1)

src/Orleans/Runtime/IGrainReferenceRuntime.cs
jdom: Sure

src/Orleans/Runtime/GrainReferenceRuntime.cs
ReubenBond: O(1) can still be a perf hit, but I think it's fine to leave this for now at least. We can see what perf runs say. Just wanted to call it out

src/OrleansRuntime/Scheduler/InvokeWorkItem.cs
```diff
@@ -43,7 +43,8 @@ public override void Execute()
             try
             {
                 var grain = activation.GrainInstance;
-                var runtimeClient = (ISiloRuntimeClient)grain.GrainReference.RuntimeClient;
+                var runtime = (GrainReferenceRuntime)grain.GrainReference.Runtime;
```
ReubenBond: This down-cast is a bit concerning, Should we have some kind of local 'InvokeOnActivation' method instead? Maybe it would require too big of a change for this PR - too many classes to expose or abstract away, what do you think?

src/OrleansRuntime/Core/InsideRuntimeClient.cs
```diff
@@ -98,6 +99,8 @@ private ILocalGrainDirectory Directory
 
         #region Implementation of IRuntimeClient
 
+        public IGrainReferenceRuntime GrainReferenceRuntime => this.grainReferenceRuntime ?? (this.grainReferenceRuntime = this.ServiceProvider.GetRequiredService<IGrainReferenceRuntime>());
```
ReubenBond: I take it that this lazy get is to break some dependency cycle?

src/OrleansRuntime/Scheduler/InvokeWorkItem.cs
jdom: Yes, this one annoyed me too. But since it was already downcasting before, I thought not to improve it in this PR. If you prefer, I could add the IRuntimeClient directly in ActivationData

src/OrleansRuntime/Core/InsideRuntimeClient.cs
jdom: Yeah, I copied the other members in this class. I'll do a quick check to see if I can break the dependency cycle, otherwise, we already know this runtimeclient acts as a glorified strongly typed service locator, that we should eventually get rid off (at least the properties that do this)

src/OrleansRuntime/Scheduler/InvokeWorkItem.cs
ReubenBond: I like your suggestion - let's move it to ActivationData

src/OrleansRuntime/Core/InsideRuntimeClient.cs
ReubenBond: LGTM either way

src/OrleansRuntime/Scheduler/InvokeWorkItem.cs
jdom: Moved to `Dispatcher` instead, which is a singleton

jdom:
Done with the fixes. Let me know if they look good. Kicking of perf tests just in case

PR #3206, opened 2017-07-07
Add documentation for GrainCancellationToken

dVakulen:


src/Documentation/Advanced-Concepts/Cancellation-Tokens.md
```diff
@@ -0,0 +1,49 @@
+---
+layout: page
+title: Grain cancellation tokens
+---
+
+# Grain cancellation tokens
+
+The Orleans runtime provides mechanism called grain cancellation token, that enable the developer to cancel asynchronous or long-running synchronous operations.
```
sergeybykov: I think if instead of "to cancel asynchronous or long-running synchronous operations" we say "to cancel an executing grain operation", it would be more clear.
Also a typo: "enable" -> "enables".

src/Documentation/Advanced-Concepts/Cancellation-Tokens.md
```diff
@@ -0,0 +1,49 @@
...
+* Instantiate a CancellationTokenSource object, which manages and sends cancellation notification to the individual cancellation tokens.
+``` csharp
+        var tcs = new GrainCancellationTokenSource();
+```
+* Pass the token returned by the GrainCancellationTokenSource.Token property to each grain method that listens for cancellation.
+
+``` csharp
+        var waitTask = grain.LongWait(tcs.Token, TimeSpan.FromSeconds(10));
+```
+* Use the underlying **CancellationToken** property of **GrainCancellationToken** from operations that receive the cancellation token.
```
sergeybykov: Would it be better to say something like "a cancellable grain operation needs to handle CancellationToken just like it would do in any other .NET code"?

src/Documentation/Advanced-Concepts/Cancellation-Tokens.md
```diff
@@ -0,0 +1,49 @@
...
+``` csharp
+        public async Task LongWait(GrainCancellationToken tc, TimeSpan delay)
```
sergeybykov: I think it would be more clear if we show a more real example of a long running operation here instead of a wait. For example, an async loop of IO operations.

src/Documentation/Advanced-Concepts/Cancellation-Tokens.md
```diff
@@ -0,0 +1,49 @@
...
+        {
+            await Task.Delay(delay, tc.CancellationToken);
+        }
+```
+* Call the ``GrainCancellationTokenSource.Cancel`` method to provide notification of cancellation.
```
sergeybykov: Nit: I'd say "to initiate cancellation" instead of "to provide notification of cancellation". Otherwise, some may read it as a notification about an external event or something.

dVakulen:
Comments were addressed.

sergeybykov:
Thank you, @dVakulen!

PR #3208, opened 2017-07-08
Preliminary Orleans Facet System implementation

jason-bragg:
This PR introduces the minimal infrastructure to support Orleans Facet System.

Improved support for activation services.
  - Added lifecycle managment for asynchronous grain state initialization.
  - Lifcycle and activation service testing.


src/Orleans/Lifecycle/ILifecycleObservable.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Orleans
+{
+    public interface ILifecycleObservable<in TRing>
```
jason-bragg: changed terms from 'ring' to 'stage', missed this interface.  Will fix

src/Orleans/Lifecycle/LifecyleExtensions.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Orleans
+{
+    public static class LifecyleExtensions
```
jason-bragg: LifecycleExtensions

src/Orleans/Runtime/IGrainLifeCycle.cs
```diff
@@ -0,0 +1,26 @@
+﻿
+namespace Orleans.Runtime
+{
+    /// <summary>
+    /// Stages of a grains lifecycle.
+    /// TODO: Add more later, see ActivationInitializationStage
+    /// Full grain lifecycle, including register, state setup, and 
+    ///   stream cleanup should all eventually be triggerd by the
```
jason-bragg: triggered

test/Tester/ActivationServices/IStorageActivationService.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            var parameter = ActivationServiceUtilities.BindToConstructorParameter(context, typeof(IStorageActivationService<TState>));
+
+            var attribute = parameter.GetCustomAttribute<StorageActivationServiceAttribute>();
+
+            this.storageActivationService = factory.Create(attribute?.StorageProviderName, attribute?.StateName ?? parameter.Name);
+            var observer = this.storageActivationService as ILifecycleObserver;
+            if (observer != null)
+            {
+                context.LifeCycle.Subscribe(GrainLifecyleStage.SetupState, observer);
```
jdom: why is this deciding to subscribe just to the SetupState stage?

src/OrleansRuntime/Catalog/ActivationData.cs
```diff
@@ -362,6 +348,10 @@ public GrainId Identity
 
         public IDictionary<object, object> Items { get; private set; }
 
+        public IGrainLifeCycle LifeCycle => FullLifeCycle;
+
+        internal GrainLifecycle FullLifeCycle { get; }
```
jdom: several places where Cycle is treated as a separate word. Should always be `Lifecycle`

src/Orleans/Lifecycle/ILifecycleObserver.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+
+namespace Orleans
+{
+    public interface ILifecycleObserver
+    {
+        Task OnStart();
+        Task OnStop();
```
jdom: Not sure I agree that there should be an OnStart and OnStop (or setup/teardown), instead of having the lifecycle enum have all the stages itself, including the ones when it's tearing down, and exceptional stages. Still undecided what's best (at least for grain lifecycle)

src/Orleans/ActivationService/ActivationServiceUtilities.cs
```diff
@@ -0,0 +1,53 @@
...
+            object item;
+            if (grainActivationContext.Items.TryGetValue(ConstructorParametersItemName, out item))
+            {
+                return (ParameterInfo[])item;
+            }
+
+            var constructor = grainActivationContext.GrainType.GetConstructors(BindingFlags.Public | BindingFlags.Instance).FirstOrDefault();
+            if (constructor == null) throw new InvalidOperationException("Could not find appropriate constructor");
+            ParameterInfo[] constructorParameters = constructor.GetParameters();
+            grainActivationContext.Items[ConstructorParametersItemName] = constructorParameters;
```
jdom: why is this info being stored in the grain activation context? this is static information that should be cached per grain type instead.

src/Orleans/Runtime/IGrainLifeCycle.cs
```diff
@@ -0,0 +1,26 @@
...
+    ///   stream cleanup should all eventually be triggerd by the 
+    ///   grian lifecycle.
+    /// </summary>
+    public enum GrainLifecyleStage
+    {
+        //None,
+        //Register,
+        SetupState,  // Setup grain state prior to activation
+        //InvokeActivate,
+        //Completed
```
jdom: Probably these well-known stages should have an explicit value, and leave values in between so that users can provide a value in between

jason-bragg:
Grain lifecycle management is critical to proper function of the virtual actor model, and the current logic is brittle, difficult to maintain, and not extensible.

I disagree that the lifecycle system is over engineered for grain lifecycle management.  Current grain lifecycle logic is a mess and very clumsy, in part due to the lack of good lifecycle management.  I do not address all of this in this PR because it's out of scope, but the introduction of a clean lifecycle management framework allows us to move towards cleanly addressing this in the future.  For more insight into this please look at how the enum ActivationInitializationStage is used, especially the fragility of the error handling, recovery around the grain activation stages, and the clumsy and poorly extensible way in which state setup is performed.

Is the lifecycle system strictly necessary for this change? No.  However, it (or something like it) is necessary for clean and maintainable grain lifecycle management, and supporting the needed features in a way that only adds more complexity and special cases to the current grain lifecycle logic would increase technical debt in exchange for short term expedience.

jdom:
Just to clarify, I think grain life cycle is important, that's not under questioning. I was merely commenting on the complexity of the infrastructure, and the tear up and down semantics, where it's probably more explicit for end users to say run this when activating, and this other thing when deactivating or when failed to activate (and that's it). This is of course a lot more flexible, although not sure how useful this flexibility is *yet*, at least when we talk about grain life cycle (silo is a completely different beast, where this flexibility will certainly be very useful, given the number of services and how long it takes to start up).
Still, overengineered or not, it will serve this other purpose, and hence I'm not pushing back, just commenting on it. If the goal is not reusing this, then I would push back and get rid of generics and levels of abstraction.

src/Orleans/ActivationService/ActivationServiceUtilities.cs
jason-bragg: Agreed.  Other prototypes kept this in the grain type info.  This was a minor perf improvement to keep from looking it up for every activation service, but I'll put it in the type info instead.

src/OrleansRuntime/Catalog/ActivationData.cs
jason-bragg: Cleaning this up in next iteration of PR.

src/Orleans/Lifecycle/ILifecycleObserver.cs
jason-bragg: It is done this way for error handling reasons.  Teardown should be opposite of startup and should only be done to the degree that startup succeeded.  If stages are started in order a, b, c , d, then they should be shutdown in d, c, b, a.  If during startup we fail at c, shutdown should start at c (not d).  Note that this is how grain activation lifecycle works now (most all lifecycles work this way), but current code has all of this hard coded in a way that is difficult to test, maintain, or extend.

src/Orleans/Runtime/IGrainLifeCycle.cs
jason-bragg: Agreed.  I also suspect there will need to be internal stages and application layer stages (for instance, application probably shouldn't have access to the registration stage), but at this point we're just setting up the basic framework.  More work is needed before grain lifecycle advantages will be fully demonstrable.

jason-bragg:
@dotnet-bot test this

jason-bragg:
@dotnet-bot test this please

src/Orleans/Lifecycle/ILifecycleObserver.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// Lifecycle observer used to handle start and stop notification.
+    /// </summary>
+    public interface ILifecycleObserver
+    {
+        /// <summary>
+        /// Handle start notifications
+        /// </summary>
+        /// <returns></returns>
+        Task OnStart();
```
jdom: I think it's probably easier to use if the stage is passed in (in case you want to subscribe to multiple stages)... Although that would require the interface to be generic. What do you say?

src/OrleansRuntime/Catalog/ActivationData.cs
jdom: still a few of these left

---
In reply to: [126530614](https://github.com/dotnet/orleans/pull/3208#discussion_r126530614) [](ancestors = 126530614)

test/Tester/ActivationServices/IStorageActivationService.cs
```diff
@@ -0,0 +1,78 @@
+﻿
+using Orleans.Runtime;
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+using Orleans;
+
+namespace Tester
+{
+    public interface IStorageActivationService<TState>
```
jdom: I created this for the prototype to have a more concrete picture of what it would look like in prod. Do you think we should add it with this name or have Test somewhere inside the name?

src/Orleans/Lifecycle/ILifecycleObserver.cs
jason-bragg: This is handled by the extension functions (See LifecycleExtensions), which allow for a user to pass in individual functions rather than a full observer.  For instance one could do something like:

Task MyConfig() {...}
Task MyInit() {...}
Task MyFinit() {...}
Task MyRun() {...}

lifcycle.Subscribe(Stages.Config, MyConfig):
lifcycle.Subscribe(Stages.Init, MyInit, MyFinit):
lifcycle.Subscribe(Stages.Run, MyRun):

Registering each callback at the appropriate stage.

src/Orleans/ActivationService/ActivationServiceUtilities.cs
jdom: putting it in grain type data would require changes and more allocations to it, when it will probably not be used that frequently. The reality is that this is just caching extra info, so I would just create a static CachedReadConcurrentDictionary<Type,. ParameterInfo[]> in here... it's all static information based on the the type anyway, so there is no risk for of this static dictionary

---
In reply to: [126529497](https://github.com/dotnet/orleans/pull/3208#discussion_r126529497) [](ancestors = 126529497)

test/Tester/ActivationServices/IStorageActivationService.cs
jason-bragg: I think this name is fine for tests.  I don't think actual activation services will have ActivationService" in their names.  They will likely be more closely related to the behavior they expose, like IPersistentState, IStreamConsumer, or ITransactionalState.

src/Orleans/ActivationService/ActivationServiceUtilities.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Orleans
+{
+    public static class ActivationServiceUtilities
+    {
+        private const string ActivationServiceNamesItemName = "ActivationServiceNamesItemName";
+        private const string ConstructorParametersItemName = "ConstructorParameters";
+
+        public static ParameterInfo BindToConstructorParameter(IGrainActivationContext grainActivationContext, Type activationServiceType)
+        {
```
xiazen: Minor. The method name confused me a little. I think the method is doing `BindActivationServiceTypeToGrainActivationContext`, or shorter `BindToActivationContext` in which the input params can come in and fill in the rest meaning. I don't have strong opinion on this. Just want to point out that it is a bit confusing and it may confuse other maintainer in the future.

src/Orleans/Lifecycle/ILifecycleObservable.cs
```diff
@@ -0,0 +1,22 @@
...
+    public interface ILifecycleObservable<in TStage>
+    {
+        /// <summary>
+        /// Subscribe for notification when a stage is reached while starting or stopping.
+        /// </summary>
+        /// <param name="stage">stage of interest</param>
+        /// <param name="observer">stage observer</param>
+        /// <returns>A disposable that can be disposed of to unsubscribe</returns>
+        IDisposable Subscribe(TStage stage, ILifecycleObserver observer);
+    }
```
xiazen: From the perspective of API design. unsubscribe through IDisposable is not clear for user to understand and remember IMO. 
Have you considered adding something like a  ISubscriptionHandle interface which has a method of Unsubscribe, and maybe some meta data of the Observable?

src/Orleans/Lifecycle/ILifecycleObserver.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        /// <summary>
+        /// Handle stop notifications
+        /// </summary>
+        /// <returns></returns>
+        Task OnStop();
+    }
```
xiazen: For the purpose of information, I think it maybe worthwhile to add metadata field in the interface level to contain information of Stage and Observable it subscribes to. Otherwise it will move the burden to the application layer to keep those info around.

xiazen: Although whether it is worthwhile to keep those info around is debatable. @jason-bragg probably has more say on this. For now I can see it may have value on debugging.

test/Tester/ActivationServices/IStorageActivationService.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    public class AttributedStorageActivationService<TState> : IStorageActivationService<TState>
+    {
+        private readonly IStorageActivationService<TState> storageActivationService;
+
+        public AttributedStorageActivationService(StorageActivationServiceFactory<TState> factory, IGrainActivationContext context)
+        {
+            var parameter = ActivationServiceUtilities.BindToConstructorParameter(context, typeof(IStorageActivationService<TState>));
```
xiazen: It seems that activation service must call `BindToConstructorParameters` to wire itself into the system. To achieve that, I wonder what's the thought behind using `ActivationServicesUtilities `vs `ActivationServiceBaseClass`. From the perspective of OOP, a base class make more sense to me. And also IMO it is slightly easier for user to remember and use.

src/Orleans/Lifecycle/ILifecycleObservable.cs
jason-bragg: I'm open to other approaches, but this pattern will be very familiar and comfortable to those familiar with IObservable which this is somewhat modeled on.
https://msdn.microsoft.com/en-us/library/dd990377(v=vs.110).aspx

The general observer pattern usually involves an attach and detach, which is a viable approach as well, but a bit more cumbersome to use.  We can always use extension functions to make it more usable though, so I'd not dismiss this out of hand, should anyone want to make an argument for it.
http://www.dofactory.com/net/observer-design-pattern

We use subscription handles in streams primarily because the unsubscribe operation is asynchronous so a disposable would not work, and since we had to deviate from the IObservable conventions, we added more stuff to the handle, but I'm not yet convinced that complex handles are a good idea, in general.

src/Orleans/Lifecycle/ILifecycleObserver.cs
jason-bragg: I don't foresee complex subscription management being a common practice in this space, but let's see what usage patterns evolve.  My preference would be to keep these interfaces minimalistic until more complexity is needed.

test/Tester/ActivationServices/IStorageActivationService.cs
jason-bragg: I agree more thought/work is needed here.  Will consider and attempt improvements in next pass.

sergeybykov:
I don't know if it's just me, but I think a high level description of the purpose and the design would help with reviewing.

test/Tester/ActivationServices/IStorageActivationService.cs
jdom: Just to clarify,  try to separate some of these concepts, as an activation service does not require to be bound to a constructor parameter. This particular approach with attributes is doing that, but in many cases it won't, so let's not couple the 2 concepts

xiazen: I see. if it is not shared by all activation services, then I withdraw my comments. 

And I second @sergeybykov 's comment, maybe more high description will help review. For example, what's the approach of using BindToConstructorParameters in an activation service. How does it wire up internally. And what's other approaches and what's their difference.

src/Orleans/Lifecycle/ILifecycleObservable.cs
xiazen: Sounds good to me :)

xiazen: I meant the IDisposable, now I see where it came from

src/Orleans/Lifecycle/ILifecycleObserver.cs
xiazen: Agreed. Let's keep it simple until needed. And I see this is also borrowing pattern from reactive programming. Just had a read on how to do testing and debugging in rx : https://msdn.microsoft.com/en-us/library/hh242967(v=vs.103).aspx. 
And Querying Observable Sequences using LINQ Operators: https://msdn.microsoft.com/en-us/library/hh242983(v=vs.103).aspx

Probably a overkill for this case. But It is a good read and maybe useful in the future for this or streaming.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
```diff
@@ -151,5 +154,11 @@ internal static PlacementStrategy GetPlacementStrategy(Type grainClass, Placemen
 
             return predicate;
         }
+
+        private ParameterInfo[] GetConstructorParameters()
+        {
+            var constructor = this.Type.GetConstructors(BindingFlags.Public | BindingFlags.Instance).FirstOrDefault();
+            return constructor?.GetParameters() ?? EmptyParameters;
+        }
```
jdom: I'd strongly suggest that we don't couple GrainTypeData to this. The reason is that this caching optimization (because it's nothing more than that) made it to something more core to Orleans, and IMO still needs more thought with regards to 3rd party containers. Other containers (or even other implementations of `IGrainActivator` might select a different constructor. I think if there's a need to couple it to something, I'd couple it to `DefaultGrainActivator`. Otherwise, probably even more flexible, make it completely standalone so that if you replace that other logic, you could replace this too.

jdom:
What's the purpose of the rename back to Facet? most of this is really activation services. The part that maps to your original design of facets is just the binding to the constructor parameter, right? Should we just call *that* part a facet (or maybe something else if the extent of that feature is now reduced)? I'd just like us to avoid introducing new concepts for standard stuff if possible.

jason-bragg:
This is still work in progress, but I'll be putting up an issue where we introduce facets, as a general Orleans concept, to the community.  That issue will reference this pr, which I will rename.  The change back to calling these facets will help the community, not familiar with the history of this feature, link the new concept to the test representations within this pr.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
jason-bragg: "Made it to something more core to Orleans"

As an extensibility pattern, it needs to be core to Orleans.  I'd argue we want the pattern to be uniform and formalized.  We don't want different facets using different patterns that only work with some containers.  Custom application specific facets can of course be generated, but those will be application specific and will not live in our tree.  Note that this change does not prevent those, only formalizes what we support.

"IMO still needs more thought with regards to 3rd party containers"
Agreed, and this is subject to change.  However, whatever we come up with should be the expected and supported pattern.  If application code deviates, that's an application problem.

"I'd couple it to DefaultGrainActivator"
I think we should think on this.  I'd originally suggested that capability be part of the activator, and there was push back because it broke the asp.net pattern that was being used.  The entire reason we're using this pattern to bind to parameters, rather than the original pattern of passing the constructed facets to the constructor was that we wanted to preserve the asp.net patterns, and now we're discussing deviating from this for what is being described as a caching optimization?

jdom: > As an extensibility pattern, it needs to be core to Orleans

Activation services, including lifetime, yes. Binding to constructor parameters so that we can use attributes in them, IMO no, it doesn't need to be core to Orleans, supporting it is a nice to have.

> > "IMO still needs more thought with regards to 3rd party containers"

> Agreed, and this is subject to change. However, whatever we come up with should be the expected and supported pattern.

But that's the point, you are coupling this to GrainTypeData, which has nothing to do with how we use DI to resolve stuff.

>> "I'd couple it to DefaultGrainActivator"

> I think we should think on this.

You were quoting me out of context... My recommendation is NOT to couple it to DefaultGrainActivator or to anything. If it has to be one, I was suggesting that's at least related to DI resolution, but I really think it should not be related to anything... if you resolve this activation service from the default configured container, then fine, it looks at the attribute information by using this workaround/hack. If you are using Autofac, you can probably change the service registration to something that is more straightforward that just looks at the metadata without having to use the BindToConstructorParameter stuff at all, which is just a workaround for how limited the default container implementation from Microsoft.Extensions.DependencyInjection is.

In summary, and being concrete with the IStorageFacet example, when I mentioned that 3rd party container integration needs more thought, I'm alluding at whether we want to also provide an alternative IStorageFacet implementation that doesn't do the binding to a constructor via the AttributedStorageFacet, but use a more formal approach for that container that is less hacky. The storage implementation doesn't change at all, just how it's wired up. Sure, Autofac can still use the hacky approach by if there is not an explicit override implementation for that container. After all, the reason you decide to use a different container is for no other reason that you prefer to configure things in a different idiomatic way.
That's also the reason why I don't think the storage interface should have the *Facet* suffix at all, because I don't want the resolution and binding to a constructor parameter to pollute the actual value of the real implementation of the storage feature. You could also configure your container in the ConfigureServices method to define the metadata the storage facet should use, people should not be forced to go through an attribute and define it statically in the Type.

jason-bragg: >Binding to constructor parameters so that we can use attributes in them, IMO no, it doesn't need to be core to Orleans.

Supported pattern for this feature does, IMO, need to be core.  As noted in previous comments, there is nothing stopping users from wiring up facets to constructors any way they want, but when in application logic, but we are not responsible for their patterns.  All of the facets we provide and maintain should, however, use the same pattern and that pattern should be the one we support for users.  I agree that the feature logic as a facet should not be tightly bound to how it's wired up, and we can work on that.

>you are coupling this to GrainTypeData, which has nothing to do with how we use DI to resolve stuff.

This data is static data related to the grain type and a grain specific behavior, so I don't see any issue caching it in the grain type data.  We need to have a deterministic means of constructor selection that is independent of DI container in any case, so I don't see how this is limiting or harmful in any way.

>I'm alluding at whether we want to also provide an alternative IStorageFacet implementation that doesn't do the binding to a constructor via the AttributedStorageFacet.

In general, and this needs more work/thought, the features we expose via facets should, IMO, be independent of how they are wired up.  A storage facet, for example, should be independent of the attribute logic and DI factories that wire it up.  The original design had this separation, and I see nothing about what we're doing so far that prevents this.

>I don't think the storage interface should have the Facet suffix

Real facets should not have the suffix, as I stated in other comments.  Facets in this PR are for test and use the name to associate them with the feature.

jason-bragg:

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
jdom: > We need to have a deterministic means of constructor selection that is independent of DI container in any case, so I don't see how this is limiting or harmful in any way.

This I very strongly disagree. This is not compatible with supporting a different IGrainActivator (remember that we will ship in the box an alternate ServiceBasedGrainActivator that will not use use ActivatorUtilities, but will use the container to resolve the transient/scoped instances that a user can configure in their container, similar to [ServiceBasedControllerActivator](https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/Controllers/ServiceBasedControllerActivator.cs) ), but also disallow custom activators or 3rd party containers that will choose the constructor in the way they want.
Absolutely defining which constructor to use is an unnecessary constraint (and fundamentally wrong IMO). After all, why do we even support switching containers if we put even more constraints to the abstraction, especially where user code is involved.
As I mentioned, the metadata for the facet con be provided by other means (not just attributes in a parameter, or working completely outside of DI), without putting this big constraint in our DI system, for no good reason.

If you want a 100% bullet-proof approach that will work regardless of how users decide to activate the grains (without any container integration code, even if that would be trivial to provide), then a good approach is to do property injection, similar to how the ControllerContext is being injected into controllers (which in their API, there is no need for an arbitrary constructor selection, nor even a base Controller class at all, which I hope we can achieve at some point with not requiring a base Grain class too). You can see how this is implemented by looking [DefaultControllerPropertyActivator](https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultControllerPropertyActivator.cs) and related classes, which runs after a controller instance was created in whichever way the user wants.

Also, looking at this decision from a different angle: Why put this restriction now and couple the constructor selection into GrainTypeData, when this adds absolutely zero value for it to be coupled so eagerly. You can pick this 1 constructor (and not make it configurable for now, even if some people don't like it) and cache it completely separate from GrainTypeData on first usage.

You mentioned you were going to write up an issue with an extended design. It would be nice to  see that, since this PR seems to be making a few assumptions that need to be validated and agreed first, and I'm not sure this thread is the right place for people to chime in.

jason-bragg: >disallow custom activators or 3rd party containers that will choose the constructor in the way they want.

We don't support an application code only container, so same container will be used for all grains, Orleans, extensions, and application.  Constructor selection needs be a well defined behavior because it must be the same for internal Orleans grains, and grains used in extensions like the providers.  This is not a new constraint introduced by this PR.

>This is not compatible with supporting a different IGrainActivator

Yes it does.  Grain activators simply can't select constructors differently for non-application grains.  If we want to specifically make constructor selection part of the grain activator, we can do that, but that can only be supported for application grains, which is of little benefit compared to the nuance and complexity it would introduce.  It can also be added at later time if/when needed.

>A good approach is to do property injection

Early prototypes took this approach and were abandoned because constructor injection is more common, intuitive, testable, and fits into standard OO designs.

>You mentioned you were going to write up an issue with an extended design

I stated I'd write up an issue introducing the capability to the community.  I did not mention an extended design.  This system is still in development and will be improved over time as we refine it.

jdom: I don't follow the logic, I don't think we will expose more than 1 constructor for system grains, so selection would always be trivial. Also, when using the DefaultGrainActivator, the selection is what we typically use. It's when using a different GrainActivator, including ServeBasedGrainActivator, that users will be able to specify how the grain is resolved, and hence can change the default. Of course if they don't override anything for system grains, they will be resolved as expected, so this more flexible activation doesn't go against how system grains are activated.
If people move away from defaults, they could of course shoot themselves in the foot, as with any extensibility feature, but I'd argue they would have to go way out of their way to mess the constructor selection part for system grains.
Again, I don't understand why constructor selection *needs* to be well defined. Can you explain the reason for it? because this is where we start deviating in the rest of the details. ASP.NET in particular does not put this constraint, and I don't understand why we should be different and deviate from that, and be a lot more restrictive than them. I don't see any advantage to it.
Also, what is the purpose of providing a different IGrainActivator if object activation is mostly fixed?

> Constructor selection needs be a well defined behavior because it must be the same for internal Orleans grains, and grains used in extensions like the providers. This is not a new constraint introduced by this PR.

This IS a new constraint for this PR. Where was that constraint enforced before? Sure, for providers we currently don't have an extensibility point to override the usage of ActivatorUtilities, so it currently only picks the same constructor every time, regarless of the container. But this is absolutely not a constraint today for grains, when defining your own IGrainActivator.

And to reiterate: what is the need to couple it to GrainTypeData in this PR? this was just for an optimization to not do reflection every time. This optimization does not warrant creating a new constraint, as it is absolutely not harmless, and there are trivial ways to avoid the coupling, as already stated.

jason-bragg: >I don't understand why constructor selection needs to be well defined.

Because without a uniform expected method of grain constructor selection, we've no means of having any confidence that the native Orleans grains nor grains in extension libraries will be created correctly with user provided containers.

>ASP.NET in particular does not put this constraint

Yes it does.  Conforming containers, which is all we support, have well defined constructor selection, which is what we should follow.
See DependencyInjectionSpecificationTests.ServiceContainerPicksConstructorWithLongestMatches
Current facet code does not conform to this as it requires a single public constructor, but we can address that as the system matures.  This is just the initial implementation.

>What is the need to couple it to GrainTypeData

Just a convenient place to cache grain type specific data.  This is an internal implementation detail.  If we find a better place we can move it.

jdom: >> ASP.NET in particular does not put this constraint

> Yes it does. Conforming containers, which is all we support, have well defined constructor selection, which is what we should follow.

Ah, I guess I now understand where the misconception is coming from. IServiceProvider will pick a specific constructor, but only if the registration just identified the type. But of course you can also even provide a factory method, so you don't even need an accessible constructor at all (not having a constructor is not supported in the default case, which uses ActivatorUtilities).

Also, the IControllerActivator extensibility point exists so you don't have to depend neither on ActivatorUtilities nor IServiceProvider at all. They even have an additional extensibility point so that if you are using the DefaultControllerActivator, you could use a different ITypeActivatorCache so that you can customize object creation (here is where the ActivatorUtilities is used by default actually).

I'm out, but if you can't find examples, on Wednesday I will be able to share some small samples of how to use different constructors or containers in ASP.NET. I'm 100% it is the case that ASP.NET doesn't put that constraint on which constructor to use from a controller. They have the same default as our default, but you can always change that.
See this article for some info on this extensibility point, as well as the ServiceBasedControllerActivator that ships in the box: https://andrewlock.net/controller-activation-and-dependency-injection-in-asp-net-core-mvc/

> Just a convenient place to cache grain type specific data. This is an internal implementation detail. If we find a better place we can move it.

Well, as I suggested, you can easily put it in a static field of `CachedReadConcurrentDictionary<Type, ParameterInfo[]>` on first use of that utility, and avoid adding the constraint and the coupling.

src/Orleans/Lifecycle/LifecycleObservable.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Threading.Tasks;
+using Orleans.Runtime;
+
+namespace Orleans
+{
+    public class LifecycleObservable<TStage> : ILifecycleObservable<TStage>, ILifecycleObserver
+    {
+        private readonly ConcurrentDictionary<object, OrderedObserver> subscribers;
```
ReubenBond: It's not immediately obvious what the key in this collection is for.

src/Orleans/Lifecycle/LifecycleObservable.cs
```diff
@@ -0,0 +1,115 @@
...
+                    .OrderBy(group => group.Key))
+                {
+                    highStage = observerGroup.Key;
+                    await Task.WhenAll(observerGroup.Select(orderedObserver => WrapExecution(orderedObserver.Observer.OnStart)));
+                }
+            }
+            catch (Exception ex)
+            {
+                logger?.Error(ErrorCode.LifecycleStartFailure, $"Lifecycle start canceled due to errors at stage {this.highStage}", ex);
+                throw new OperationCanceledException("Start canceled to failures.", ex);
```
ReubenBond: This could share an error message with the above log

src/Orleans/Lifecycle/LifecycleObservable.cs
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+        }
+
+        public IDisposable Subscribe(TStage stage, ILifecycleObserver observer)
+        {
+            if (observer == null) throw new ArgumentNullException(nameof(observer));
+
+            var key = new object();
+            subscribers.TryAdd(key, new OrderedObserver(stage, observer));
```
ReubenBond: The key here could be the OrderedObserver itself

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -1370,6 +1371,7 @@ private void RerouteAllQueuedMessages(ActivationData activation, ActivationAddre
                     {
                         RequestContext.Clear(); // Clear any previous RC, so it does not leak into this call by mistake. 
                         await activation.GrainInstance.OnDeactivateAsync();
+                        await activation.LifeCycle.OnStop();
```
ReubenBond: Perhaps it's acceptable, but if the grain throws in OnDeactivateAsync, then OnStop will never be called.

src/OrleansRuntime/Catalog/GrainLifecycle.cs
```diff
@@ -0,0 +1,10 @@
+﻿
+namespace Orleans.Runtime
+{
+    internal class GrainLifecycle : LifecycleObservable<GrainLifecyleStage>, IGrainLifeCycle
```
ReubenBond: Is LifecycleObservable a bit heavyweight for this purpose? It's built for multi-threaded access and the implementation is quite allocation happy. If it was just being used for silo startup, then that's fine, but for every single activation it seems a bit heavy

jason-bragg: For what it currently does, yes, this is a little heavy.  But the plan is to use the lifecycle to manage the full grain lifecycle, not just the facets.

src/OrleansRuntime/Catalog/Catalog.cs
jason-bragg: This will not be a problem once the lifecycle is used to trigger the OnDeactivateAsync call.

src/Orleans/Lifecycle/LifecycleObservable.cs
jason-bragg: It was just an opaque generic id.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
jason-bragg: >IServiceProvider will pick a specific constructor, but only if the registration just identified the type. But of course you can also even provide a factory method, so you don't even need an accessible constructor at all

We're adding features that we intend to expose via constructor injection, so if one is to follow the pattern we're setting up, well defined constructor selection is necessary.  The standards we've adopted for our supported DI containers, also have a well defined constructor selection method.  So where is the problem here?

> I'm 100% it is the case that ASP.NET doesn't put that constraint on which constructor to use from a controller. 

Conforming to ASP.NET's controller pattern is not a goal.  The feature set we wish to support is the goal.  If we can achieve this using ASP.NET's controller pattern, that would be great as it is more widely understood, but we're dealing with asynchronously constructed grains in a distributed system, there not the same as controllers.

jdom: Well, what I'm proposing still supports both aspnet DI story without crippling it and the facet story. I'll have to continue discussing on Wednesday though, but please at least take a look at the link. I strongly suggest we do not merge this with this unnecessary limitation

jason-bragg:

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
ReubenBond: It seems like there's no hard reason to keep the mapping from `grainType` to `ctor params` in GrainTypeData. Could we move it out?

jason-bragg: @ReubenBond, Of course we can move it out, but that seems arbitrary.  Can you provide a technical reason why it shouldn't be kept with all of the other grain type static data?  I'm not opposed to moving it elsewhere, but not on the basis of personal whims.  Please provide technical reasons for why this approach is bad/wrong and alternatives with their advantages.

What we need is a common way to provide constructed types wired up with specific metadata to a grains construction.  These types need to be able to take part in the grain lifecycle.  Originally this was done by passing the constructed objects as arguments during construction, but that was not supported by the asp.net controller pattern, so we moved to matching up items to parameters in the constructor (like in this PR), which necessitates the need for a well defined constructor selection process, which fortunately the DI specification we've adapted supports.  Where we cache the parameters is an internal detail and is not that important, but I am opposed to adding new static data sets without reason.

jdom: Just to chime in before I can devote more time: it's 100% not a personal whim. You seem to be dismissing actual technical reasons thinking it's a whim, without trying to understand it

ReubenBond: My reasoning for moving it out isn't firm, I just noted that `GrainTypeData` is `[Serializable]` and even though this particular field isn't serialized, it feels like we are overloading that class as an object with both data and logic. It might actually be that it isn't being sent over the wire (I thought it was). In that case, we can remove `[Serializable]`.

Julian's argument seems to boil down to "this is implementation specific" / "this makes too many assumptions about implementations which we do not control". I think those are valid. Your counter points seems to be "implementations must conform to the spec, making those assumptions requirements". There are two aspects to this:
1. Is the `IServiceProvider` conforming? Yes, we only support conforming service providers.
2. Is the `IGrainActivator` conforming? Well... what does conforming mean here?

Users are expected to be able to inject their own implementation for `IGrainActivator`, but we do not have any way for them to know how to create one correctly. It almost seems that we need to create a test kit for our own spec for `IGrainActivator` to check for 'conforming grain activators'.

If the user provided the following container implementation then it's (surprisingly?) not conforming:
```C#
public class ServiceProviderGrainActivator : IGrainActivator
{
    public object Create(IGrainActivationContext context)
        => context.ActivationServices.GetService(context.GrainType);

    public void Release(IGrainActivationContext context, object grain)
        => (grain as IDisposable)?.Dispose();
}
```

Even though a conforming `IServiceProvider` must be in use, we're not constructing the grain correctly and we do not know which constructor is being used: the user might have registered `MyGrain` with the conforming container via a lambda and that lambda can construct `MyGrain` in whatever whacky way it wants.

The fact that a user must be using a conforming container does not mean that we know for sure how a type is constructed. Maybe that's fine, but it needs to be acknowledged.

This is the issue I'm more interested in resolving (or becoming comfortable with): the higher-level question of how grain construction is performed and what is acceptable for an `IGrainActivator`.

jason-bragg: @ReubenBond It seems like, and please correct me if I'm mistaken, the concern is regarding how grain activator will custom construct grains, and in particular, how that will work with facets.  Is that correct?

The facet pattern in this PR is a pattern we support but do not enforce.  It does not limit user’s construction of their grain in any way.  The pattern, as a means of exposing Orleans features to grains, should be used to expose features, but the features should not depend on being exposed by attributes on arguments in a constructor, as the feature should not care how it gets wired up, only that it is.  With the introduction of DI, Orleans is extremely extensible, and by exposing (some parts of) the grain lifecycle to the application layer we allow for a much wider range of Orleans extensions.  However, we can't be responsible for supporting every possible way one might want to use this power, nor would it be desirable (even internally) to have several different patterns be used, due to the testability and maintainability problems we're already facing.

> how grain construction is performed and what is acceptable for an IGrainActivator

So the answer, imo, to the grain activator question is, they can use the facet pattern we support using grain constructors and attributes, or they can use the features directly wiring things up themselves.

The facet system design we came up with (and agreed on) was that features would be injected in constructors with metadata provided by attributes.  How is this to be done without constructor selection?  There are other ways we can wire this up.  We can modify the IgrainActivator interface to take arguments and pass the constructed facets to it.  We can construct the facets when creating the IGrainActivationContext and expose them to the grain activator via the context.  I’m sure there are other approaches as well, but all of which require constructor selection because the design requires access to the metadata stored in attributes on the parameters.  Am I missing something?

This PR is the first pass at the facet system.  More work is necessary, and the system will be fleshed out more as we use it.  In particular, the current constructor selection is incorrect, and the test facets do not have a strong separation between the feature logic and the attribution.  If you'd prefer I address these in this PR, I'm willing to do that, as the constructor selection logic will eventually need to be done anyway.  What I'm less willing to do is make the wiring of features to the grain an implementation detail of the facet rather than a common supported pattern.  A central goal of this system was to have a common supported pattern for exposing features to grains.  If there are alternatives with technical advantages, I'm happy to adopt them, but alternatives need to address the requirements and have technical advantages.

Regarding the GrainTypeData. It is serializable so grain type data can be sent to the client and other silos (heterogeneous deployments), but the constructor arguments are only ever required locally, so there is no reason to serialize them.  It already contains logic and data, it’s not simply a raw data type, and it’s where we keep static data about grain types, which is what the constructor parameters (imo) are.  Where the parameters cache is stored is entirely orthogonal to the question of constructor selection, which seems to be the main objection.  Again I note that this is internal logic that can be changed at any time without customer impact should we find a better place for this.  If we want to add some extensibility point for the application layer to provide it's own metadata to facets rather than getting the information from the constructors, making this cache unnecessary, I'm not opposed to that, but blocking this PR and the continued development of these systems on new requirements and advanced features no one has requested is, imo, poor prioritization at best, more likely speculative over engineering.

ReubenBond: > The facet system design we came up with (and agreed on) was that features would be injected in constructors with metadata provided by attributes. How is this to be done without constructor selection?

I need to be more clear: I like this feature *very much*, and I like the pattern with ctor parameter attributes for wiring it up. The question about `IGrainActivator` does need to be answered, though, because this system does place additional constraints on an otherwise very open-ended API. I don't think that we should allow this facets pattern to be optionally supported, if we are going down this road then it should be mandatory that `IGrainActivator` follows consistent ctor selection and doesn't allow custom registrations. It's enough that we tell implementers in doc comments that they *must* adhere to this and that the easiest way to do so is to construct their types using `ActivatorUtils.CreateFactory`. It should be mandatory, since the intention is that this is a central extensibility pattern.

> Regarding the GrainTypeData. It is serializable so grain type data can be sent to the client and other silos (heterogeneous deployments), but the constructor arguments are only ever required locally, so there is no reason to serialize them.

I won't push back, but note that we have had problems with this in the past.

attilah: > @jdom: This is not compatible with supporting a different IGrainActivator

> @jason-bragg: "Yes it does. Grain activators simply can't select constructors differently for non-application grains. If we want to specifically make constructor selection part of the grain activator, we can do that, but that can only be supported for application grains, which is of little benefit compared to the nuance and complexity it would introduce. It can also be added at later time if/when needed."

We have to construct app and system grains the *same* way IMO, since we support user specified container usage and user specified grain activators.

The problem here is that we like to have some information (the ctor selected by the configured DI container) which is not exposed through the DI abstraction since we don't have access to it, since the instance construction is internal to every DI container.

I like to have attributes on the constructor arguments, but since DI abstraction does not have any way to support this kind of metadata whatever we are adding here belongs to DI land. Maybe using attributes (however it is nice) the most unfortunate thing we can support in the initial implementation, due the lack of support.

What we can require from app developers that if they like to use facets, then go though some well defined registration methods if they're using facets, so during registration we can have ways to extract the metadata we need for our default implementation.

My 2 cents against ```GrainTypeData``` use of reflection is that we're trying to drive an effort where all of the reflection/grain metadata would be stored, accessed and populated in one place and adding such logic into classes going against this effort.

Since DI is deeply involved I'd like to see tests running the same code with Autofac for example to make us confident that other conforming containers can work nicely without providing extra code from the app developer who is using a 3rd party container.

jason-bragg: The more I think on this the more I favor the following:

**All grains that use facets must have a single public constructor.**
Acquiring constructed facets aligned with grain construction arguments is the responsibility of the grain context, and will align with arguments in the grain's single public constructor.
Default grain activator will get the arguments from the context, then construct the grain using activator utilities.
Custom grain activators, if they want to use facets, will need to have a single public constructor in their grains, acquire the arguments from the context, and construct the grain with the arguments.

If users want to deviate from this, we can't stop them, but this is the only pattern we should support.

Acquiring the facets for a given grain type can be an injectable behavior, but I don't think it should be.  It can also be added in the future if/when needed, and should not block this PR.

src/Orleans/Lifecycle/LifecycleObservable.cs
jason-bragg: TStage are the stages of the lifecycle.  The key in this dictionary is just an opaque id used to removed the item.  They are orthogonal.

src/Orleans/Lifecycle/LifecycleObservable.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            this.logger = logger?.GetLogger(GetType().Name);
+            subscribers = new ConcurrentDictionary<object, OrderedObserver>();
+        }
+
+        public async Task OnStart()
+        {
+            try
+            {
+                foreach (IGrouping<TStage, OrderedObserver> observerGroup in subscribers.Values
```
jason-bragg: I don't consider start/stop operations as part of hot path.  They are called once per lifecycle.  This sounds like a premature optimization at the cost of maintainability/readability.

As far as ordering guarantees, we make no commitments regarding subscription order.  There are no ordering guarantees at the stage level either, all operations at a stage should be run in parallel.  We only commit to ordering by stage.

src/Orleans/Lifecycle/LifecycleObservable.cs
```diff
@@ -0,0 +1,116 @@
...
+                    .GroupBy(orderedObserver => orderedObserver.Stage)
+                    .OrderBy(group => group.Key))
+                {
+                    highStage = observerGroup.Key;
+                    await Task.WhenAll(observerGroup.Select(orderedObserver => WrapExecution(orderedObserver.Observer.OnStart)));
```
jason-bragg: In your suggested whenall line, what happens if there are 10 observers in the group and the third one does this?

Task OnStop()
{
    throw new NotImplementedException():
}

The remaining 7 OnStop calls never get called, correct?  This is what the wrapper is for.  With the wrapper all OnStop calls are executed, regardless of errors.

src/Orleans/Lifecycle/LifecycleObservable.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+            catch (Exception ex)
+            {
+                string error = $"Lifecycle start canceled due to errors at stage {this.highStage}";
```
jason-bragg: OnError could be a good addition at a later point, but I don't see any need for it at this time.

src/Orleans/Lifecycle/LifecycleObservable.cs
```diff
@@ -0,0 +1,116 @@
...
+        public IDisposable Subscribe(TStage stage, ILifecycleObserver observer)
+        {
+            if (observer == null) throw new ArgumentNullException(nameof(observer));
+
+            var orderedObserver = new OrderedObserver(stage, observer);
+            subscribers.TryAdd(orderedObserver, orderedObserver);
+            return new Disposable(() => Remove(orderedObserver));
+        }
+
+        private void Remove(object key)
```
jason-bragg: >This method wraps TryRemove and this way swallows the result of the removal, which can fail.

TryRemove would only fail if the key was not found.  Why do we care if the remove failed or not?

>"misusing" IDisposable

How are we misusing IDisposable?

>if the registered Observer is an IDisposable then we should properly dispose it

The observer is not disposable, the subscription is.  Disposing the disposable returned by the subscribe call removes the subscription.

jason-bragg:
>it (lifecycle) lacks the support for CancellationToken support 

So are you suggesting:

    public interface ILifecycleObserver
    {
        Task OnStart(CancellationToken ct);
        Task OnStop(CancellationToken ct);
    }

or something else.  What would make these abstractions appropriate for silo lifecycle?

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
galvesribeiro: > If users want to deviate from this, we can't stop them, but this is the only pattern we should support.

I agree with it. However, we should add this to the (long) list of validations (like always return `Task/Task<T>`, don't use `lock`, etc.)  that we _should_ provide in an optional Roslyn analyzer package.

ReubenBond: IMO it's fine to push the requirement of consistent constructor selection onto `IGrainActivator` implementations. ctor injection is my preference, even if we also support property injection later. If we do support ctor injection of facets, I don't see how else we can do it.

Does anyone have suggestions?

src/Orleans/Runtime/IGrainActivationContext.cs
```diff
@@ -24,5 +25,9 @@ public interface IGrainActivationContext
 
         /// <summary>Gets a key/value collection that can be used to share data within the scope of the grain activation.</summary>
         IDictionary<object, object> Items { get; }
+
+        IGrainLifeCycle ObservableLifeCycle { get; }
+
+        ParameterInfo[] ConstructorParameters { get; }
```
ReubenBond: If the activator could somehow provide this info then they could choose their own ctor. Just an idea.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
jdom: Jason, I think I could stop by the office at around 11am on Friday to continue discussing in person.
I’m ready to meet in a middle ground as I mentioned in this PR since day 1, and still go with constructor injection.

Nevertheless, the more I think about it, the more it makes sense to support property injection with the exact same proven and extensible code here: [DefaultControllerPropertyActivator](https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultControllerPropertyActivator.cs)

When we built the unity container, we had lots of this religious wars that took weeks (since the container was our  deliverable after all). Back then I was on the side of taking a stand and not even supporting property and method injection, as most of times it was miss-used and led to user objects that required 2-step initialization (aka temporal coupling) before they are usable. This is not a limitation for us anyway, since we already have a 2-step initialization an users should not consider grains usable until OnActivateAsync is called.

Using property injection by copying the code from MVC means that we don’t put ANY constraint on how users decide to construct objects. It is also as intuitive, but at the same time we are standing on the decisions of the MVC team and their documented semantics (which we are not shoehorning, they are good IMO). It also is a proven approach that works and we can continue evolving (such as allowing to replace the IGrainPropertyActivator if we so decide to support).

Just trying to say to not dismiss a good and already well-thought approach just because “we think constructor injection is cooler than property injection”. We wanted to avoid those religious wars by standing on their decisions, but we started one just because we deviated and want to decide how grains should be resolved.

Just my 2 cents to think a little bit before we can continue the discussion

src/Orleans/Lifecycle/LifecycleObservable.cs
attilah: As we talked in person, hot-path is depending on how intense is the grain activation/deactivation, which is not a design, but a runtime thing. In IoT scenarios where you're working with lot of sensors this can be easily a hot-path. What we know that it is slow and knowing a better/faster way I am not considering premature.

src/Orleans/Runtime/IGrainActivationContext.cs
attilah: Since that data is internal to every container and the abstraction/default implementation does not give us a way to access this information it is "impossible" to get.

What we maybe can do is that since our default chosen container implementation is the Asp.Net Core DI container, we can use the internal ```ConstructorMatcher``` as source code to make sure we're selecting the same constructor what the container would use. BUT if the user is configuring a different container, then our default activator would not be usable or just to some extent with some constraint, but the serviceprovider based one would work, IF the configured conforming container supports retrieving metadata from ctor parameters.

attilah:
@jason-bragg yes, that is what I meant for cancellationtoken, but as I said adding these for grain lifecycle does not make any sense.

src/Orleans/Lifecycle/LifecycleObservable.cs
attilah: > The observer is not disposable, the subscription is. Disposing the disposable returned by the subscribe call removes the subscription.

```OrderedObserver```'s ```Observer``` is an ```ILifecycleObserver``` what is implemented by the application developer. If my class is also an ```IDisposable```, because I need to dispose stuff when my Observer's duty is over (on removal), then we've to honor that and call ```IDisposable.Dispose```.

```c#
public class MyLifecycleObserver : ILifecycleObserver, IDisposable
{
   ...
}
```

jason-bragg: It's not the responsibility of the lifecycle to cleanup subscribers.  I'm unclear where this view came from.  The observer pattern, which this is roughly modeled off of, doesn't do anything like that.  If observers need disposed at some point, the callers code needs to address this, it's not the responsibility of this class.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
jason-bragg: The IControllerPropertyActivator, like the IControllerActivator, is an abstraction which allows for applications to provide container specific versions of a behavior.  This is fine for MVC, as they do not need a common pattern, but we do.  That is the fundamental difference here.  We are introducing an extensibility point that requires, from the container, grain activator, or whatever other abstractions we introduce, that the pattern be respected.  By using property injection rather than constructor injection, we're just moving the problem to properties rather than constructors, and offering a inferior experience to our users in the process.

The idea that this is how MVC solved this problem indicates a misunderstanding of the problem.  MVC does not have this problem.  They have no need for a common pattern for either constructor injection or property injection.  They need only a means to allow applications to provide their own, container specific, versions of these features, which they do through these abstractions.

Whether we introduce facets via property injection or constructor injection, we still violate what MVC does.  In this, however, I do not care.  Using established patterns is only valuable if they solve the same problem, which these do not.  Given a choice between providing clean, intuitive, and maintainable capabilities to our users or arbitrarily confining our solution to existing common patterns, I will choose a better user experience every time.

src/Orleans/Lifecycle/LifecycleObservable.cs
jason-bragg: I agreed that it could be optimized, and that in extreme cases it could be costly.  I also agree that it may be valuable to revisit this in the future and introduce a grain specific lifecycle (it is an abstraction after all) which is more tuned to the performance needs of grain.  I do not agree that grain activation is part of the hot path.  Given all that grain activation does, the number of observers would need to be absurdly high before this ever even showed up as a blip in profiling.  Spending time optimizing this now, before we've any reason to believe it will effect performance at all, is, imo, premature optimization.

src/OrleansRuntime/GrainTypeManager/GrainTypeData.cs
jdom: Please review the implementation as this IS the exact same scenario that we have. They have a way to inject the mandatory (scoped) ControllerContext into the controller, regardless of the controller activator in use, and whether there are accessible constructors or not, as this is an extremely important *facet* of the controller.
If you still think it's not the same scenario, I'd like to understand why, because I've been reviewing it thoroughly for months (since I first created the scoped activation context PR), and the more I look at it, the more I believe is the exact same scenario.
BTW, I'm not suggesting to expose the IGrainPropertyActivator abstraction (in case you are worried they might replace it and make our system grains not work), just make the DefaultGrainPropertyActivator the only non-replaceable implementation for now. We can guarantee that the pattern is respected without any gray areas.

Moving this to property injection is not just moving the problem from one place to another. Is avoiding it entirely. No matter which container they choose, we can always know which property we are injecting one at a time, and read the metadata at the time we are building the facet and injecting it. There is no guessing which member is being constructed, nor limiting how grains are resolved.

> Given a choice between providing clean, intuitive, and maintainable capabilities to our users or arbitrarily confining our solution to existing common patterns, I will choose a better user experience every time.

I guarantee they have the same driving function... it's not because they are fans of less-popular practices that they chose this approach. Also, as I mentioned, in our case there is no additional temporal coupling which is the main reason property injection is less favorable. This approach is also not less intuitive (and far more documented given how MVC is). With regards to the maintainable or confining solution comment, the proven approach wins that by an order of magnitude, I thought there was no question there. What remains is (subjectively) preferring constructor injection regardless of the reasons (still haven't heard a reason why constructor injection would be better for this case, other than personal preference).

attilah:
I created a similar implementation which remove the need for the ```AttributedStorageFacet``` and having a custom factory class for each facet: https://gist.github.com/attilah/7efce01666a258378fac9df5431580cd

jason-bragg:
Update on the design

While the goals of the system remain the same, we’re changing how this capability will be exposed to users.  Rather than facets being injected via constructor injection, we’ve opted to use property injection instead.
So the grains from the example in "Orleans Facet System #3223" would instead look like:

    public class MyGrain : Grain, IMyGrain
    {
        [PersistentStateStorage("Blob", "MyState")]
        public IPersistentState<MyState> State { get;set; }
    }

And

    public class MyGrain : Grain, IMyGrain
    {
        [PersistentStateStorage("Blob", "StreamState")]
        public IPersistentState<StreamState> StreamState { get;set; }
        [PersistentStateStorage("Table", "UserState")]
        public IPersistentState<UserState> UserState { get;set; };
        [StreamProvider("AzureQueue")]
        public IStreamConsumer<UserData> StreamConsumer { get;set; };
    }

The facet system will perform property injection to wire up the facets after grain construction and prior to calling OnActivateAsync.

I'll update this PR to reflect this decision shortly.

galvesribeiro:
@jason-bragg I don't think it is a good Idea to have those properties... 

It will start making people confuse and expose internal grain implementation details (i.e. Grain State) to whoever has an instance or it.

I think the best way to do it, is to follow your previous design and use the ctor Injection. 

I'm seeing a big discussion here regarding this design and I think it would worth a skype call or your guys meet in person at office and decide how to move with the design but honestly, property injection is a good idea.

I wonder what are @gabikliot thoughts about it.

jason-bragg:
@galvesribeiro, The team discussed this Wednesday and today.  This was the general conclusion of the team.  I'm not asking that you withhold your objections, only making it clear that this was discussed extensively, and is not a snap decision that I've made on my  own.

galvesribeiro:
I see... Sad to hear that :(

jason-bragg:
Closing this PR.  Will submit property injected version on different PR.

PR #3209, opened 2017-07-09
Implemented Custom Serializer for GrainStateWithMetaDataAndETag

ibebbs:
Implemented Custom Serializer for GrainStateWithMetaDataAndETag so that the state could also provide a custom serializer rather than being stuck with the fallback serializer.
Consolidated GrainStateWtihMetaData into GrainStateWithMetaDataAndETag as former was not used anywhere and this made code cleaner.
Added unit tests for customer serializers

### DIFF
--- a/src/OrleansCounterControl/OrleansCounterControl.csproj
+++ b/src/OrleansCounterControl/OrleansCounterControl.csproj
@@ -67,6 +67,7 @@
     <None Include="Microsoft.Orleans.CounterControl.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="project.json" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\OrleansRuntime\OrleansRuntime.csproj">
--- a/src/OrleansCounterControl/project.json
+++ b/src/OrleansCounterControl/project.json
@@ -1,0 +1,11 @@
+﻿{
+  "dependencies": {
+  },
+  "frameworks": {
+    "net461": {}
+  },
+  "runtimes": {
+    "win-anycpu": {},
+    "win": {}
+  }
+}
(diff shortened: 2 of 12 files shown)


ibebbs:
Apparently the 'UnitTests.General.ElasticPlacementTests.ElasticityTest_StoppingSilos' test is failing with the error 'Expecting a value in the range between 1030 and 1630, but instead got 1674 outside the range'.

These numbers are pretty obtuse and, as the test passes locally, I'm unsure how to go about resolving this failure, especially as I feel it is highly unlikely my change-set has impacted this test.

Any advice?

sergeybykov:

sergeybykov:
@ibebbs Looks like most likely a timing issue in the test.

benjaminpetit:
`UnitTests.MembershipTests.ClientIdPartitionDataRebuildTests.ReconstructClientIdPartitionTest_Request` failed during the setup sequence, I wouldn't worry too much, it does not seems to be related to this PR.

I think Julian modified it a bit several days ago, if we see more failure on this test we ight want to fix something.

sergeybykov:

ReubenBond:
Thanks, @ibebbs :)

The changes are backwards compatible except for this one:
> Consolidated GrainStateWtihMetaData into GrainStateWithMetaDataAndETag as former was not used anywhere and this made code cleaner.

That change will cause existing state which is stored to break during deserialization. If the other changes are made without changing the existing members or inheritance hierarchy of the class, then this would be backwards compatible.

The root issue here is that there was no serializer being generated for `GrainStateWithMetaDataAndETag` and so it was using the fallback. It's fine to add a new serializer to a class which was previously using a fallback, that will not cause any issues. Modifying the class members & hierarchy will cause an issue, though.

sergeybykov:
I'll close this for now because there seems to have been no progress on https://github.com/ibebbs/orleans/pull/1 since July.

Issue #2057, opened 2016-08-14
Google Cloud Platform (GCP) integration


galvesribeiro:
This issue is part of #2005 by implementing a minimal set of providers to make Orleans to run well by integrating Google Cloud Platform (GCP).
- [ ] Cloud DataStore Grain State persistence provider 
- [ ] Cloud DataStore Membership Provider 
- [ ] Cloud DataStore Reminder Provider 
- [x] Cloud Pub/Sub Stream Provider #3210

galvesribeiro:
@gabikliot I didn't found any Stream/Queue service on GCP. The only thing usable for our case is the [Cloud DataStore](https://cloud.google.com/datastore/). Do you know anything else?

galvesribeiro:
Nvm, looks like Pub/Sub is the guy https://cloud.google.com/pubsub/

gabikliot:
Yes, exactly, pub sub. And datastore for storage.

galvesribeiro:
@gabikliot ok, btw, is there any trial that doesn't require a credit card on GCP? Brazilian issued cards (even international ones) are failing while creating the trial account with `Authorization failed: No reason provided by your financial institution` and when I call my bank they say everything is OK... Those are the same cards I use for either AWS and Azure :disappointed:

galvesribeiro:
We can't release nugets for it since all nugets from Google are on pre-release state :cry:

gabikliot:
We can have a pre-release Nuget for OrleansGoogleUtils - we already have it this way.
Whether we publish it on Nuget.org (just like datastore is https://www.nuget.org/packages/Google.Apis.Datastore.v1beta2) or not is a secondary question. Even if not, people could still build Orleans themselves and get that Nuget, or take the latest stable CI build.

galvesribeiro:
Sad thing. The Protobuf nuget that current project in main branch uses, now has a release nuget however, the others we will use in this issue isn't. Anyway, will follow your suggestion once I have everything from AWS done. Thanks

seniorquico:
@galvesribeiro It sounds like OrleansGoogleUtils will contain both the GCP integration providers and the Protobuf serializer? I have an interest in testing/using only the Protobuf serializer. If these are combined, wouldn't I have to pull all of the other GCP dependencies into my project? If that's the case, I think it would be great if these were split into separate packages. That would also mitigate the issue from your previous comment regarding the Google packages being pre-release except Protobuf.

sergeybykov:
[...]

PR #3210, opened 2017-07-10
Google PubSub Stream provider

galvesribeiro:
Part of #2057.

The code is complete. Tests are being coded and will update this PR along with details on how to run the PubSub emulator in order to run those tests.


src/Orleans/Properties/AssemblyInfo.cs
```diff
@@ -31,6 +31,7 @@
 [assembly: InternalsVisibleTo("OrleansHost")]
 [assembly: InternalsVisibleTo("OrleansAzureUtils")]
 [assembly: InternalsVisibleTo("OrleansAWSUtils")]
+[assembly: InternalsVisibleTo("OrleansGoogleUtils")]
```
xiazen: Ideally we should avoid using internal types in Orleans for extensions. Is there a way to avoid this?

xiazen: What internal type you are trying to access here?

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubAdapterConstants.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Orleans.Serialization.Providers.Streams
+{
+    public static class GooglePubSubAdapterConstants
+    {
+        public const string NUMBER_SUBSCRIPTIONS = "NumSubscriptions";
+        public const int NUMBER_SUBSCRIPTIONS_DEFAULT = 8;
+        internal const int CACHE_SIZE_DEFAULT = 4096;
```
xiazen: cache size param was removed in 1.5.

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubAdapterFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+        where TDataAdapter : IGooglePubSubDataAdapter
+    {
+        private string _projectId;
+        private string _topicId;
+        private string _deploymentId;
+        private string _providerName;
+        private int _cacheSize;
+        private int _numSubscriptions;
+        private TimeSpan? _deadline;
+        private HashRingBasedStreamQueueMapper _streamQueueMapper;
```
xiazen: This will limit GooglePubSubStreamProvider only be able to use HashRingBasedStreamQueueMapper as its queue mapper. Have you considered add a QueueMapperFactory here to make `IStreamQueueMapper `configurable?

galvesribeiro: @xiazen I'm trying to make it as simple as possible so 3 guys from Gitter who asked for this provider can be unblocked. This implementation is essentially the very basic queue-based Stream provider (like Azure Queue and AWS SQS). After it get in, we can refactor it to make more flexible. There are a bunch of Google's optimization that can be used. Specially in Serialization with Protobuf. But I think it is out of scope for an initial provider.

I'm just trying to get people using GCP unblocked.

src/Orleans/Properties/AssemblyInfo.cs
galvesribeiro: The `ErrorCode` is one case. There are others. As you can see, AWS and Azure utils do the same since they rely on it.

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubAdapterReceiver.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        public async Task<IList<IBatchContainer>> GetQueueMessagesAsync(int maxCount)
+        {
+            try
+            {
+                var pubSubRef = _pubSub; // store direct ref, in case we are somehow asked to shutdown while we are receiving.    
+                if (pubSubRef == null) return new List<IBatchContainer>();
+
+                int count = maxCount < 0 ? 10 : maxCount;
```
xiazen: why do we choose 10 here? seems like a small number for queueing messages. 
And keep in mind , maxCount can be QueueAdapterConstants.UNLIMITED_GET_QUEUE_MSG, which is a below-zero value "-1"

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubAdapterReceiver.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                var task = pubSubRef.GetMessages(count);
+                _outstandingTask = task;
+                IEnumerable<ReceivedMessage> messages = await task;
+
+                List<IBatchContainer> pubSubMessages = new List<IBatchContainer>();
+                foreach (var message in messages)
+                {
+                    IBatchContainer container = _dataAdapter.FromPullResponseMessage(message.Message, _lastReadMessage++);
+                    pubSubMessages.Add(container);
```
xiazen: I read through google pubsub. Found out they have two modes for subsribers, push and pull. Each mode is there for fulfilling different set of use cases. https://cloud.google.com/pubsub/docs/subscriber

Just curious, have you considered enable push model for GooglePubSubStreamProvider?

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubAdapterReceiver.cs
galvesribeiro: What you suggest here? Google has no suggestion on that. Unlike Azure which has a constant in `CloudQueueMessage`

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubAdapterReceiver.cs
galvesribeiro: @xiazen yes, I did. But the push model is not advised for high throughput scenarios as you can read on their doc. It works similar to what happens on Azure EventHubs. Even that we have `push` mode there, we sill pull messages so Streaming APIs can deal with backpressure, etc.

galvesribeiro: Also, push requires that each Silo has a public HTTPS endpoint open so Google can reach it to deliver the messages.

galvesribeiro:
BTW, @jason-bragg when you have some time, please ping me on Gitter... I'm with a failing test here which I have no idea why it is happening.

gabikliot:
I suggest to create a new folder named GCP or GoogleCloudPlatform and put all the Cloud related files there.
4 reasons:
1) There is a big difference between protobuf (and gRPC) versus GCP Datastore, PubSub, Spanner... The latter are part of the Cloud, while the former are unrelated to the Cloud, can be used totally unrelated from Google Cloud.
2) This will also make it easier for other to consume protobuf and not have to consume GCP in the same package. Just like we put Bond in a separate package from Azure.
3) naming by company is uncommon - we don't call it Microsoft Utils but rather Azure, and AWS and not Amazon.
4) I would personally really appreciate us sticking to and stressing the branding - the brand is GCP - Google Cloud Platform.

galvesribeiro:
Sure @gabikliot.

The next update I'll add a GCPUtils project and move everything there. I just followed the _pattern_ used everywhere else.

src/Orleans/Properties/AssemblyInfo.cs
jason-bragg: @xiazen is correct.  These should not rely on internals if at all possible.  The fact that we've done this in the past is not justification to continue the practice.

Enabling the extensibility of Orleans is one of our goals, so we'd rather fix the extensibility points than continue to allow internal access in this way.  For error codes, please look at how orleansservicebus handles this.  What other internals do you need?

galvesribeiro: Ok, will check on that. @jason-bragg I'm now on Gitter, when you can, plz ping me there.

test/GoogleUtils.Tests/GoogleUtils.Tests.csproj
```diff
@@ -128,6 +128,9 @@
     <Compile Include="ProtobufSerialization\Counter.cs" />
     <Compile Include="ProtobufSerialization\ProtobufSerializationTests.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Streaming\" />
```
xiazen: What's the plan for testing this stream provider?

galvesribeiro: As the PR description mention, I'm adding the tests :)

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubAdapterReceiver.cs
```diff
@@ -0,0 +1,150 @@
...
+                _pending.RemoveRange(0, finalizedDeliveries.Count);
+                // get the queue messages for all finalized deliveries that were delivered.
+                List<ReceivedMessage> deliveredMessages = finalizedDeliveries
+                    .Where(finalized => deliveredTokens.Contains(finalized.Token))
+                    .Select(finalized => finalized.Message)
+                    .ToList();
+                if (deliveredMessages.Count == 0) return;
+                // delete all delivered queue messages from the queue.  Anything finalized but not delivered will show back up later
+                _outstandingTask = Task.WhenAll(deliveredMessages.Select(m => pubSubRef.AcknowledgeMessages(new[] { m })));
+
```
xiazen: both google's native api and your `GooglePubSubDataManager` has a batch operation for AcknowledgeMessages. Why do you use it like a single operation? This would potentially bring performance penalty when you try to acknowledge a large amount of messages

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubAdapterReceiver.cs
```diff
@@ -0,0 +1,150 @@
...
+                // remove all finalized deliveries from pending, regardless of if it was delivered or not.
+                _pending.RemoveRange(0, finalizedDeliveries.Count);
+                // get the queue messages for all finalized deliveries that were delivered.
+                List<ReceivedMessage> deliveredMessages = finalizedDeliveries
+                    .Where(finalized => deliveredTokens.Contains(finalized.Token))
+                    .Select(finalized => finalized.Message)
+                    .ToList();
+                if (deliveredMessages.Count == 0) return;
+                // delete all delivered queue messages from the queue.  Anything finalized but not delivered will show back up later
+                _outstandingTask = Task.WhenAll(deliveredMessages.Select(m => pubSubRef.AcknowledgeMessages(new[] { m })));
```
xiazen: AcknowledgeMessages support a batch operation, why do you use it like a single operation, which would potentially bring performance penalty?

galvesribeiro: Ouch! My bad! Fixing.

src/OrleansGoogleUtils/Providers/Streams/GooglePubSubBatchContainer.cs
```diff
@@ -0,0 +1,83 @@
...
+        public bool ImportRequestContext()
+        {
+            if (requestContext != null)
+            {
+                RequestContext.Import(requestContext);
+                return true;
+            }
+            return false;
+        }
+
```
xiazen: I think GetEvents, ShouldDeliver and ImportRequestContext method are almost the same across most IBatchContainer implementations. I wonder can we have a utility class or a base class which every IBatchContainer implementation can benefit from?

galvesribeiro: There are tons of "same things" around in Streaming internal APIs. If you look at AWS SQS, Azure Queue and this implementation, they are almost the very same code. I've already mentioned it before that we should refactor out all those common code in a Streaming V2 framework. The same for Storage and Membership...

test/GoogleUtils.Tests/GoogleUtils.Tests.csproj
xiazen: oic. understood ! :)

galvesribeiro:
@dotnet-bot test this please

galvesribeiro:
Ok, added the missing projects/tests and all feedback was addressed.

src/OrleansGCPUtils/OrleansGCPUtils.csproj
```diff
@@ -0,0 +1,69 @@
...
+    <Compile Include="GoogleErrorCodes.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="Providers\Streams\PubSub\PubSubAdapter.cs" />
+    <Compile Include="Providers\Streams\PubSub\PubSubAdapterConstants.cs" />
+    <Compile Include="Providers\Streams\PubSub\PubSubAdapterFactory.cs" />
+    <Compile Include="Providers\Streams\PubSub\PubSubAdapterReceiver.cs" />
+    <Compile Include="Providers\Streams\PubSub\PubSubBatchContainer.cs" />
+    <Compile Include="Providers\Streams\PubSub\PubSubDataManager.cs" />
+    <Compile Include="Providers\Streams\PubSub\PubSubStreamProvider.cs" />
+    <Compile Include="Providers\Streams\PubSub\IPubSubDataAdapter.cs" />
```
xiazen: PubSub has another meaning in Orleans, which is the pubsub store. So I think it maybe confusing to have a separate folder named PubSub in GCP, even though it means GooglePubSub. I think "Providers\Streams\" is enough.

src/OrleansGCPUtils/Providers/Streams/PubSub/IPubSubDataAdapter.cs
```diff
@@ -7,12 +7,12 @@
 using System.Collections.Generic;
 using System.Linq;
 
-namespace Orleans.Providers.Streams
+namespace Orleans.Providers.GCP.Streams.PubSub
 {
     /// <summary>
     /// Converts event data to and from cloud queue message
     /// </summary>
-    public interface IGooglePubSubDataAdapter
```
xiazen: Why do we remove Google from the naming? It is easy for people to relate this with Orleans streaming pubsub store, even though it is not that at all. It can be confusing.

src/OrleansGCPUtils/OrleansGCPUtils.csproj
galvesribeiro: The reason it was put this way, is to follow the most complete stream implementation which is EventHubs and also follow @gabikliot advices regarding use the proper naming of the product. Also, keep in mind that it is one (the most basic) Stream provider for Google GCP. I plan to have later a more feature full one that uses Google PubSubIO, the same way we did with Azure Queue -> Event Hubs and that new provider would be under `Providers\Streams\PubSubIO`.

src/OrleansGCPUtils/Providers/Streams/PubSub/IPubSubDataAdapter.cs
gabikliot: Let's either call it GCPPubSub, or if it already lives under GCP namespace, just leave as PubSub. No need to duplicate namespace in the name.

Both of the above 2 sound good to me.

galvesribeiro: The current namespace is `Orleans.Providers.GCP.Streams.PubSub`. I think it dedup the name and its clear enough for users.

test/GoogleUtils.Tests/Streaming/GPSStreamTests.cs
```diff
@@ -0,0 +1,169 @@
...
+                {
+                    { "ProjectId",  GoogleTestUtils.ProjectId },
+                    { "TopicId",  GoogleTestUtils.TopicId },
+                    { "DeploymentId",  GoogleTestUtils.DeploymentId.ToString()},
+                };
+
+            options.ClientConfiguration.RegisterStreamProvider<GooglePubSubStreamProvider>("GPSProvider", providerSettings);
+
+            options.ClusterConfiguration.Globals.RegisterStreamProvider<GooglePubSubStreamProvider>("GPSProvider", providerSettings);
+            options.ClusterConfiguration.Globals.RegisterStreamProvider<GooglePubSubStreamProvider>("GPSProvider2", providerSettings);
```
xiazen: what's the reason to configure extra sms stream provider and second `GPSProvider2 `when only `GPSProvider `is used in the test?

galvesribeiro: Yay! Bad copy :)

Removing. Thanks

xiazen:
LGTM now. :)

xiazen:
@dotnet-bot test netfx-bvt

xiazen:

galvesribeiro:
@xiazen `Could not find a part of the path 'D:\j\workspace\netfx-bvt_prtest437995ca\Binaries\Debug\OrleansGCPUtils'` it ifail on netfx build at create nugets step... Do we have something "new" in there that I'm not aware of?

xiazen:
we probably do. Let me ask the expert here @attilah @jdom

galvesribeiro:
Ping

sergeybykov:
@attilah can you help here?

attilah:
@sergeybykov sure.

src/NuGet/Microsoft.Orleans.OrleansGCPUtils.nuspec
```diff
@@ -0,0 +1,31 @@
...
+    <dependencies>
+      <dependency id="Microsoft.Orleans.Core" version="$version$" />
+      <dependency id="Microsoft.Orleans.OrleansProviders" version="$version$" />
+      <dependency id="Google.Cloud.PubSub.V1" version="1.0.0-beta11" />
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="OrleansGCPUtils.dll" target="lib\net461" />
+    <file src="OrleansGCPUtils.pdb" target="lib\net461" />
+    <file src="OrleansGCPUtils\**\*.cs" target="src" />
```
attilah: This should be:
```<file src="$SRC_DIR$OrleansGCPUtils\**\*.cs" target="src" />```

galvesribeiro:

galvesribeiro:

xiazen:

src/NuGet/Microsoft.Orleans.OrleansGCPUtils.nuspec
galvesribeiro: Thanks @attilah! Bad copy/paste. That did the trick.

shlomiw:
@galvesribeiro - awesome! 👍

Issue #131, opened 2015-02-14
Provide a Powershell SnapIn


MovGP0:
There should be a PowerShell SnapIn with a set of CmdLet's for managing Orleans Silos. The SnapIn should provide the possibility to add and remove applications to the silo while the application is still running. 

Examples: 
- Start-Orleans
- Add-OrleansApplication
- Remove-OrleansApplication
- Stop-Orleans
- etc.

References: 
- [Code Example on Wikipedia](http://de.wikipedia.org/wiki/Windows_PowerShell#Snap-Ins)
- [MSDN: How to Create a Windows PowerShell Snap-in](https://msdn.microsoft.com/en-us/library/ms714450%28v=vs.85%29.aspx)

MovGP0:
It might needed to send commands to the currently running silo to register applications on the fly. 

This might be helpful with that: 
* [CodeProject: Single Instance Application, Passing Command Line Arguments](http://www.codeproject.com/Articles/15185/Single-Instance-Application-Passing-Command-Line-A)

sergeybykov:
PorwerShell cmdlets would be great no doubt. OrleansManager was meant to provide some similar functionality. Should the cmdlets be wrappers on top of OrleansManager.exe or re-implement its functionality? OrleansManager communicates via ManagementGrain leveraging Orleans messaging.

However, the add/remove application functionality isn't supported in the silo code yet.

MovGP0:
Reimplementing the functionality would duplicate functionality and therefore increase the amount of code to manage too much. 

However, a simple wrapper would not be too nice either. I would rather put the common functionality into a separate OrleansManagement.dll.

sergeybykov:
Is there any benefit in introducing OrleansManagement.dll vs. exposing the same refactored common functionality from the OrleansManager.exe assembly directly?

MovGP0:
it's more about the Separation of Concerns (SoC) principle than for technical reasons. Technically we could put everything in the same file or one UI library on top of the other. 

I believe it is preferable from the design perspective, to separate the (console) user interface(s) from the business logic into separate libraries. This way it is possible to introduce a change in one UI without effecting tho other UI.

sergeybykov:
Makes sense. Are you interested in crafting a PR for this?

MovGP0:
Interested yes, but already busy with an side project for an NGO, which currently has priority for me...

sergeybykov:
[...]

PR #3211, opened 2017-07-10
Rename Orleans.psd1 to OrleansPSUtils.psd1

sergeybykov:
Need to do this for publishing OrleansPSUtils.dll to the PS Gallery (#131).
I already published 1.5.0 version of it manually. This is to update the sources to match what I had to change by hand.


sergeybykov:
Done.

PR #3215, opened 2017-07-11
Rebuilt html docs

sergeybykov:


sergeybykov:
This is included in https://github.com/dotnet/orleans/pull/3218. So no need to merge it separately.

PR #3220, opened 2017-07-12
Update FAQ

sergeybykov:
This is only to fix outdated and not correct anymore statements. A bigger rewrite of the FAQ is needed, but that should be a separate effort.

### DIFF
--- a/src/Documentation/Frequently-Asked-Questions.md
+++ b/src/Documentation/Frequently-Asked-Questions.md
@@ -7,7 +7,8 @@ title: Frequently Asked Questions
 
 ## Does Microsoft support Orleans?
 
-Source code of Orleans has been released under an MIT license on [GitHub](https://github.com/dotnet/orleans). Microsoft continues to invest in Orleans and accepts community contributions to the codebase.
+Source code of Orleans has been released under an MIT license on [GitHub](https://github.com/dotnet/orleans).
+Microsoft continues to invest in Orleans and accepts community contributions to the codebase.
 
 ## Can I get a "Go Live" License?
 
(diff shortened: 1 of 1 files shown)


src/Documentation/Frequently-Asked-Questions.md
```diff
@@ -21,64 +22,88 @@ Orleans has been production ready and used in production for several years.
...
+Grains should return as soon as possible to let other grains execute for maximum throughput.
 
 ## In what cases can a split brain (same grain activated in multiple silos at the same time) happen?
 
-This can never happen during normal operations and each grain will have one and only one instance per ID.
-The only time this can occur is when a silo crashes or if it's killed without being allowed to properly shutdown.
-In that case there is a 30-60 seconds window (based on configuration) where a grain can exist in multiple silos before one is removed from the system.
+During normal operations the Orleans runtime guarantees that each grain will have at most one instance in the cluster.
+The only time this guarantee can be violated is when a silo crashes or gets killed without a proper shutdown.
+In that case, there is a ~30 second (based on configuration) window where a grain can potentially get temporarily instantiated in more than one silo.
```
ReubenBond: Is it worth mentioning that grains with persistent state will be immediately deactivated when they attempt to persist their state to storage in the event of a duplicate activation due to write conflicts?

sergeybykov: I don't think this fits the FAQ (at least in its current form). Plus, it won't necessarily get immediately deactivated because grain code may catch the exception.

In general, I think we need to rewrite the FAQ altogether. But that's not what I tried to do here. I tried only to update it and fix inaccuracies.

PR #3221, opened 2017-07-13
Update Prerequisites page

sergeybykov:
Updated some basic information and links.

### DIFF
--- a/src/Documentation/Installation/Prerequisites.md
+++ b/src/Documentation/Installation/Prerequisites.md
@@ -5,13 +5,20 @@ title: Prerequisites
 
 # Prerequisites
 
-Orleans is a set of .NET libraries. In order to use Orleans, you need [.NET Framework](http://www.microsoft.com/net) 4.5.1 or higher and a copy of [Visual Studio](https://www.visualstudio.com) 2015 or higher. Note that the Express versions of Visual Studio do not support extension packages, but you can use Orleans by adding references to the NuGet packages directly.
+Orleans is a set of .NET libraries delivered via [NuGet packages](NuGets.md).
+In order to use Orleans, you need [.NET Framework](http://dot.net) 4.6.1 (since 1.5.0, 4.5.1 for prior versions) or higher and a copy of [Visual Studio](https://www.visualstudio.com) 2015 or higher.
+Note that the Express versions of Visual Studio do not support extension packages, but you can use Orleans by adding references to the NuGet packages directly.
 
-In production, Orleans requires persistent storage. The following technologies are supported (only need one of those):
+In production, Orleans requires persistent storage for reliable cluster membership.
+The following storage technologies are supported for managing cluster membership state (only need one of those):
 
-* [Azure](https://azure.microsoft.com/en-us/pricing) - Tested with [Azure SDK](http://azure.microsoft.com/en-us/downloads) 2.4 - 2.8
+* [Azure Table Storage](https://azure.microsoft.com/en-us/services/storage/tables/) - Tested with [Azure SDK](http://azure.microsoft.com/en-us/downloads) 2.4 - 2.8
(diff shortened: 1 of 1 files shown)


Documentation/Installation/Prerequisites.html
```diff
@@ -74,16 +74,21 @@
...
+<p>In production, Orleans requires persistent storage for reliable cluster membership.
+The following storage technologies are supported for managing cluster membership state (only need one of those):</p>
 <ul>
-<li><a href="https://azure.microsoft.com/en-us/pricing">Azure</a> - Tested with <a href="http://azure.microsoft.com/en-us/downloads">Azure SDK</a> 2.4 - 2.8</li>
+<li><a href="https://azure.microsoft.com/en-us/services/storage/tables/">Azure Table Storage</a> - Tested with <a href="http://azure.microsoft.com/en-us/downloads">Azure SDK</a> 2.4 - 2.8</li>
 <li><a href="https://www.microsoft.com/en-us/server-cloud/products/sql-server">SQL Server</a> 2008 or higher</li>
 <li><a href="https://zookeeper.apache.org">ZooKeeper</a> 3.4.0 or higher</li>
 <li><a href="https://www.mysql.com">MySQL</a> 5.0 or higher</li>
 <li><a href="https://www.consul.io">Consul</a> 0.6.0 or higher</li>
 <li><a href="https://aws.amazon.com/dynamodb/">DynamoDB</a> - Tested with <a href="https://www.nuget.org/packages/AWSSDK.DynamoDBv2/3.1.5.3">AWSSDK - Amazon DynamoDB 3.1.5.3</a></li>
```
ReubenBond: Should we add Service Fabric to this list? We also support Postgres

src/Documentation/Installation/Prerequisites.md
```diff
@@ -5,13 +5,19 @@ title: Prerequisites
 
 # Prerequisites
 
-Orleans is a set of .NET libraries. In order to use Orleans, you need [.NET Framework](http://www.microsoft.com/net) 4.5.1 or higher and a copy of [Visual Studio](https://www.visualstudio.com) 2015 or higher. Note that the Express versions of Visual Studio do not support extension packages, but you can use Orleans by adding references to the NuGet packages directly.
+Orleans is a set of .NET libraries delivered via [NuGet packages](NuGets.md).
+In order to use Orleans, you need [.NET Framework](http://dot.net) 4.6.1 (since 1.5.0, 4.5.1 for prior versions) or higher and a copy of [Visual Studio](https://www.visualstudio.com) 2015 or higher.
+Note that the Express versions of Visual Studio do not support extension packages, but you can use Orleans by adding references to the NuGet packages directly.
 
-In production, Orleans requires persistent storage. The following technologies are supported (only need one of those):
+In production, Orleans requires persistent storage for reliable cluster membership.
```
ReubenBond: perhaps we should change the wording here so that we can have one consistent list instead of putting SF in a separate sentence.
eg "Production deployments of Orleans require reliable cluster membership. The membership system is pluggable and supported implementations are maintained for the following systems:"

sergeybykov: I struggled with this a bit. The way I see it is the list enumerates all membership table storage options while SF is a different case - a deployment solution. For example, one could deploy Orleans as a stateless SF service with Azure Table or SQL as the membership store. That's why I felt simply adding SF to the list would be inaccurate.

sergeybykov: Need to add Postgres, indeed, but what version do we support?

sergeybykov:
Added PosgreSQL. Removed other commits and html docs for simplicity.

src/Documentation/Installation/Prerequisites.md
ReubenBond: @veikkoeeva might know which version of Postgres is required.

The "Orleans requires persistent storage for reliable cluster membership" is inaccurate - it's not storage itself which is required, just some external system.

PR #3224, opened 2017-07-14
Add extension for one-way grain calls

dVakulen:
As per https://github.com/dotnet/orleans/pull/2993#issuecomment-298992424 there's still room for extension, and it will bring more flexibility to users - implemented https://github.com/dotnet/orleans/issues/2893#issuecomment-291806366


src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Orleans.CodeGeneration;
+using Orleans.Concurrency;
+
+namespace Orleans.Runtime.GrainMethodInvocationExtensions
+{
+    public static class GrainMethodInvocationExtensions
```
dVakulen: Still thinking on better name.

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -15,6 +15,12 @@ public class GrainReference : IAddressable, IEquatable<GrainReference>, ISeriali
         private readonly string genericArguments;
         private readonly GuidId observerId;
 
+        /// <summary>
+        /// Invoke method options specific to this grain reference instance
+        /// </summary>
+        [NonSerialized]
+        private readonly InvokeMethodOptions invokeMethodOptions;
```
dVakulen: Also can be used to provide more extension methods, such as  'InvokeUnordered', etc.

src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <typeparam name="T">Grain interface</typeparam>
+        /// <param name="grainReference">Grain reference which will be copied and then a call executed on it</param>
+        /// <param name="grainMethodInvocation">Function that should invoke grain method and return resulting task</param>
+        public static void InvokeOneWay<T>(this T grainReference, Func<T, Task> grainMethodInvocation) where T : class, IGrain
+        {
+            // double cast is safe, as it can be invoked by user only on grain references
+            var oneWayGrainReferenceCopy = new GrainReference((GrainReference) (object)grainReference, InvokeMethodOptions.OneWay).Cast<T>();
+
+            // Task always completed at this point. Should also help catching situation of mistakenly calling the method on original grain reference
+            if (!grainMethodInvocation(oneWayGrainReferenceCopy).IsCompleted)
```
ReubenBond: Very nitpicky, but could we move `grainMethodInvocation(oneWayGrainReferenceCopy)` into a variable named `result` or some such?

test/DefaultCluster.Tests/OneWayCallTests.cs
```diff
@@ -27,6 +29,27 @@ public class OneWayCallTests : HostedTestClusterEnsureDefaultStarted
...
+        [Fact]
+        public async Task MethodsInvokedThroughOneWayExtensionReturnSynchronously()
+        {
+            var grain = this.Client.GetGrain<ICanBeOneWayGrain>(Guid.NewGuid());
+
+            var observer = new SimpleGrainObserver();
+            var observerRef = await Client.CreateObjectReference<ISimpleGrainObserver>(observer);
+            grain.InvokeOneWay(g =>
+            {
+                Assert.True(g != grain, "One way call should be executed on copy of grain reference");
```
ReubenBond: should this be explicitly using reference equality, `ReferenceEquals`?

src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
ReubenBond: This could be moved to `GrainExtensions`, I think.

src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// Invokes a method of a grain interface is one-way fashion so that no response message will be sent to the caller.
+        /// </summary>
+        /// <typeparam name="T">Grain interface</typeparam>
+        /// <param name="grainReference">Grain reference which will be copied and then a call executed on it</param>
+        /// <param name="grainMethodInvocation">Function that should invoke grain method and return resulting task</param>
+        public static void InvokeOneWay<T>(this T grainReference, Func<T, Task> grainMethodInvocation) where T : class, IGrain
+        {
+            // double cast is safe, as it can be invoked by user only on grain references
+            var oneWayGrainReferenceCopy = new GrainReference((GrainReference) (object)grainReference, InvokeMethodOptions.OneWay).Cast<T>();
+
```
ReubenBond: `grainReference` here could be `grainReference.AsWeaklyTypedReference()` instead of casting, that would allow it to from inside grains like:
```C#
this.InvokeOneWay(self => self.SomeDeferedMethod());
```

src/Orleans/Runtime/GrainReference.cs
ReubenBond: Is that doc comment a little misleading, since the value is also copied via the copy constructor?
*Should* it be copied by the copy constructor? I think the answer is no, since the reference should be temporary and should never escape, but I'm not entirely certain yet.

src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
ReubenBond: We could also loosen the generic constraint to be `IAddressable` instead of grain - no big deal either way, but it might be useful for implementing some lower-level protocols using `GrainService`/`SystemTarget` at some point.

src/Orleans/Runtime/GrainReference.cs
dVakulen: If it woun't be copied by that constryctor - the will be no other way to set it besides making ```invokeMethodOptions``` internal protected property. Also, reference can escape even without that constructor by assingnin it to some outer campured reference. But there's always exists some ways of shooting in the leg, don't think can block them all. Also such escaping will allow reuse one-way enabled grain reference.

src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
dVakulen: The reason why it's not there is https://github.com/dotnet/orleans/issues/2893#issuecomment-292002821. If we don't actually worry about that - will move.

src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
dVakulen: Done

src/Orleans/Runtime/GrainReference.cs
ReubenBond: Oh, I understand - because `Cast<T>()` uses the copy constructor. That's fine, let's leave this.

src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
dVakulen: Right, will do

src/Orleans/Runtime/GrainMethodInvocationExtensions.cs
ReubenBond: Does anyone else want to weigh in here? My opinion is to let it be accessible, but it's not a firm opinion.

dVakulen: cc @jdom

dVakulen: Moved to ```GrainExtensions```

ReubenBond:
LGTM. When functionals complete I will merge, unless someone has objections?

dVakulen:

The test failures is unrelated to PR

dVakulen:
Thank you, @ReubenBond

PR #3229, opened 2017-07-17
OrleansAwsUtils: support implicit authentication to DynamoDB (e.g. via IAM Roles)

chrisbrook83:
OrleansAwsUtils currently requires explicit credentials to connect to DynamoDB. These are supplied via the Orleans SystemStore DataConnectionString in the silo config.

This change introduces support for implicit authentication via EC2 IAM Roles, which avoids the need to have explicit credentials in the config files.

If no 'AccessKey' and 'SecretKey' attributes are included in the connection string, the AmazonDynamoDBClient is created without any credentials supplied, which results in the AWS SDK checking the usual locations for credentials (including the EC2 instance profile service):
http://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/DynamoDBv2/MDynamoDBv2DynamoDBctorDynamoDBConfig.html

This is not a breaking change - the present behaviour will remain the same for any users already supplying explicit credentials.


sergeybykov:
Thank you, @chrisbrook83!

PR #3231, opened 2017-07-17
vNext Removal and merge with *.2017.csproj files

attilah:
Move common properties to Directory.Build.props, .targets.
Introduce BuildFlavor variable to control what to build:
- default to "Current" is not specified, which is always the "next" release, currently 2.0
- "Legacy" to build Orleans 1.x.
Merge vNext\src to 2017 csproj files.
Delete vNext folder.
Update CI build file to build 2017 sln with Current BuildFlavor.
Move NuGet PackageReference Versions to props file.


ReubenBond:
Reviewed up until "Change project factory guids in SLN file to CPS ones" (0e1cd75)

I think we just need to fix F# build/test before we merge.
We should also delete the "Full" sln file.
I encountered some build warnings about nuget packages (eg: `C:\dev\orleans\Tools\dotnetcli\sdk\2.0.0-preview3-006764\Microsoft.Common.CurrentVersion.targets(1984,5): warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed. [C:\dev\orleans\src\OrleansPSUtils\OrleansPSUtils.2017.csproj]`)

attilah:
@ReubenBond I think those warnings coming from the difference that Orleans is built against .Net Standard 1.5 and some projects referencing them against .Net 4.6.1, but makes no harm runtime.

F# will be fixed of course before the merge (as a last thing).

ReubenBond:
This results in NuGet pkgs being built in each project's subdirectory. Maybe that's desired, I'm not sure, but I wonder if we should set the `PackageOutputPath` MSBuild property to put them in one directory for the purpose of publishing / easier local consumption.

sergeybykov:
>I wonder if we should set the PackageOutputPath MSBuild property to put them in one directory for the purpose of publishing / easier local consumption.

I agree.

attilah:
Sure, it makes sense. I will add it, so local nuget consumption will be easier if needed for some scenarios.

Directory.Build.targets
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <!-- Begin Orleans: Without these lines the project won't build properly -->
+  <!-- Set path to ClientGenerator.exe -->
+  <PropertyGroup Condition=" '$(OrleansBuildTimeCodeGen)' == 'true' ">
+    <OrleansReferencesBase>$(ProjectDir)..\Bootstrap\</OrleansReferencesBase>
```
jdom: Probably all flavors do the same, not sure, but wouldn't it make sense to have BuildFlavor and Configuration part of the path for the Bootstrap.
Also, instead of ProjectDir, we probably want to use the path to this msbuild script, which will not change from project to project, even if they are nested differently.

attilah: We never had this separation, but it makes sense, thanks, I'm adding it!

src/Orleans.SDK.targets
```diff
@@ -30,15 +30,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <CodeGenToolExe>$(CodeGenToolExeDir)/ClientGenerator.exe</CodeGenToolExe>
       </PropertyGroup>
     </Otherwise>
   </Choose>
 
   <!-- This target is run just before Compile for an Orleans Grain Interface Project -->
   <Target Name="OrleansCodeGeneration"
-          AfterTargets="BeforeCompile"
-          BeforeTargets="CoreCompile"
+          BeforeTargets="AssignTargetPaths"
```
jdom: Is this a target that comes in all MSBuild projects or is it custom-defined? Because this has to run in a plain .NET project

src/Orleans.SDK.targets
```diff
@@ -9,9 +9,9 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
   <Choose>
     <!-- Using Orleans NuGet packages -->
-    <When Condition="Exists('$(MSBuildThisFileDirectory)..\tools\ClientGenerator.exe')">
+    <When Condition="Exists('$(MSBuildThisFileDirectory)../tools/ClientGenerator.exe')">
       <PropertyGroup>
-        <CodeGenToolExeDir>$(MSBuildThisFileDirectory)..\tools</CodeGenToolExeDir>
+        <CodeGenToolExeDir>$(MSBuildThisFileDirectory)../tools</CodeGenToolExeDir>
       </PropertyGroup>
     </When>
     <!-- Internal build within Orleans.sln -->
```
jdom: in the future we should just get rid of this logic from this file that we ship in the nuget package. We can just have this pre-set in the BuildDirectory.props and that's it

src/Orleans.SDK.targets
attilah: Yes, this is part of the normal flow. @ReubenBond suggested this change, because generated files were not picked up in the earlier configuration by NuGet pack, since ```Compile``` item collection happened earlier.

src/Orleans/Orleans.csproj
```diff
@@ -1,522 +1,92 @@
...
+  <Target Name="OrleansBootstrap" BeforeTargets="AssignTargetPaths" DependsOnTargets="$(OrleansBootstrapSteps)" Condition="'$(Bootstrap)' != 'true'" />
+
+  <Target Name="OrleansDllBootstrapUsingCodeGen" Inputs="@(ReferencePath)" Outputs="$(IntermediateOutputPath)$(TargetName).orleans.g.cs" Condition="'$(Bootstrap)' != 'true'">
+    <Message Text="[OrleansDllBootstrapUsingCodeGen] - Building ClientGenerator start" Importance="high" />
     <PropertyGroup>
-      <BootstrapOutputPath>$(OutDir)Bootstrap\</BootstrapOutputPath>
-      <BootstrapOutputPath Condition="'$(BuildingInsideVisualStudio)' == 'true'">$(ProjectDir)$(OutDir)Bootstrap\</BootstrapOutputPath>
-      <ExcludeCodeGen>$(DefineConstants);EXCLUDE_CODEGEN</ExcludeCodeGen>
+      <!-- Visual Studio or MsBuild .sln build -->
+      <BootstrapOutputPath>$(SourceRoot)/Bootstrap/$(BuildFlavor)/$(Configuration)/</BootstrapOutputPath>
```
jdom: wasn't this defined in the Build.Directory.props file already?

src/Orleans.PlatformServices/Orleans.PlatformServices.csproj
```diff
@@ -21,10 +25,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Runtime" Version="4.3.0" />
+    <PackageReference Include="System.Runtime" Version="$(SystemRuntimeVersion)" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.5'">
-    <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
+    <PackageReference Include="System.Runtime.Loader" Version="$(SystemRuntimeVersion)" />
```
jdom: SystemRuntimeVersion for System.Runtime.Loader. Is this by design or a mistake?

src/OrleansBondUtils/OrleansBondUtils.csproj
```diff
@@ -1,81 +1,35 @@
...
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <DocumentationFile>bin\Release\OrleansBondUtils.XML</DocumentationFile>
-  </PropertyGroup>
-  <PropertyGroup>
-    <SignAssembly>false</SignAssembly>
+    <PackageId>Microsoft.Orleans.Serialization.Bond</PackageId>
+    <Title>Microsoft Orleans Bond Serializer</Title>
+    <Description>Referencing this package allows Bond types to be serialized using the Bond Simple Binary serializer.</Description>
+    <PackageTags>$(PackageTags) Azure Bond</PackageTags>
```
jdom: Azure is not a related tag for this package

src/OrleansCodeGenerator/OrleansCodeGenerator.csproj
```diff
@@ -1,82 +1,30 @@
...
-    </ProjectReference>
+
+  <ItemGroup Condition=" '$(BuildFlavor)' == 'Current' ">
+    <ProjectReference Include="..\Orleans.PlatformServices\Orleans.PlatformServices.csproj" />
   </ItemGroup>
-  <ItemGroup>
-    <None Include="project.json" />
+
+  <ItemGroup Condition=" '$(BuildFlavor)' == 'Current' ">
+    <PackageReference Include="System.Reflection" Version="$(SystemRuntimeVersion)" />
```
jdom: wouldn't it be better to have a SystemReflectionVersion variable defined in the props file too? That one can also point (for now) to SystemRuntimeVersion if they do align, but I think it's better than then trying to find all places where you used SystemRuntimeVersion but you meant a different package

src/OrleansCounterControl/Microsoft.Orleans.CounterControl.targets
```diff
@@ -1,4 +1,4 @@
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="15.0" DefaultTargets="Build">
```
jdom: does it mean that this cannot be used by end users if they are still in Visual Studio 2015?

src/OrleansSQLUtils/OrleansSQLUtils.csproj
```diff
@@ -1,115 +1,41 @@
...
+
   <ItemGroup>
-    <Content Include="CreateOrleansTables_MySql.sql">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>
-    <Content Include="CreateOrleansTables_PostgreSql.sql">
+    <None Include="CreateOrleansTables_MySQL.sql">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Pack>true</Pack>
+      <PackagePath>lib\$(TargetFramework)\MySQL\</PackagePath>
```
jdom: Why in lib\$(TargetFramework)? Is this how it was done before and you are just preserving the location?
Otherwise it might make sense to put them in some other not-framework specific folder. But if it's preserving it as it was before, ignore and keep it like this for now

Directory.Build.props
```diff
@@ -15,32 +16,134 @@
...
+    <WindowsAzureServiceBusVersion>4.1.0</WindowsAzureServiceBusVersion>
+    <WindowsAzureStorageVersion>7.0.0</WindowsAzureStorageVersion>
+
+    <!-- 3rd party packages -->
+    <AWSSDKDynamoDBv2Version>3.1.5.2</AWSSDKDynamoDBv2Version>
+    <AWSSDKSQSVersion>3.1.0.9</AWSSDKSQSVersion>
+    <BondCoreCSharpVersion>4.0.1</BondCoreCSharpVersion>
+    <ConsulVersion>0.7.0.3</ConsulVersion>
+    <GoogleProtobufVersion>3.0.0</GoogleProtobufVersion>
+    <ZooKeeperNetExVersion>3.4.8.5</ZooKeeperNetExVersion>
```
jdom: IMO, these packages that are probably just used in a single project, I would just move the version there and not define a global msbuild variable.

Test.cmd
```diff
@@ -1,28 +1,37 @@
...
+if not exist %CMDHOME%\test\Tester\%_Directory%\TestVersionGrainsV2\ mkdir %CMDHOME%\test\Tester\%_Directory%\TestVersionGrainsV2
+
+copy %CMDHOME%\test\Versions\TestVersionGrains\%_Directory%\* %CMDHOME%\test\Tester\%_Directory%\TestVersionGrainsV1\
+copy %CMDHOME%\test\Versions\TestVersionGrains2\%_Directory%\* %CMDHOME%\test\Tester\%_Directory%\TestVersionGrainsV2\
+
+set TESTS=%CMDHOME%\test\AWSUtils.Tests\%_Directory%\AWSUtils.Tests.dll,%CMDHOME%\test\BondUtils.Tests\%_Directory%\BondUtils.Tests.dll,%CMDHOME%\test\Consul.Tests\%_Directory%\Consul.Tests.dll,%CMDHOME%\test\DefaultCluster.Tests\%_Directory%\DefaultCluster.Tests.dll,%CMDHOME%\test\GoogleUtils.Tests\%_Directory%\GoogleUtils.Tests.dll,%CMDHOME%\test\NonSilo.Tests\%_Directory%\NonSilo.Tests.dll,%CMDHOME%\test\ServiceBus.Tests\%_Directory%\ServiceBus.Tests.dll,%CMDHOME%\test\Tester\%_Directory%\Tester.dll,%CMDHOME%\test\TesterAzureUtils\%_Directory%\Tester.AzureUtils.dll,%CMDHOME%\test\TesterInternal\%_Directory%\TesterInternal.dll,%CMDHOME%\test\TesterSQLUtils\%_Directory%\Tester.SQLUtils.dll,%CMDHOME%\test\TesterZooKeeperUtils\%_Directory%\Tester.ZooKeeperUtils.dll
```
jdom: Need to include ServiceFabric tests also

src/Orleans.PlatformServices/Orleans.PlatformServices.csproj
attilah: It is intentional we reference every System* package is using the ```SystemRuntimeVersion``` variable. ```SystemVersion``` did not make sense, hence the name. Here it is "unfortunate" that the package starts with ```System.Runtime```.

src/Orleans/Orleans.csproj
attilah: This is where to build bootstrap there where to use it from. Different purpose, but pointing to the same directory now.

src/OrleansBondUtils/OrleansBondUtils.csproj
attilah: Agree, probably copy and paste error from AzureUtils.

src/OrleansCodeGenerator/OrleansCodeGenerator.csproj
attilah: We've a lot of System packages I don't like the idea of having versions separately. We can find a better name, but use the same var, except when we need exceptions for patch releases like ```System.Net.Http``` 4.3.1.

src/OrleansCounterControl/Microsoft.Orleans.CounterControl.targets
attilah: Can't tell, how it will behave, but we decided to not to support VS2015, hance the change there too.

src/OrleansSQLUtils/OrleansSQLUtils.csproj
attilah: Exactly, from nuspec:
```
<file src="CreateOrleansTables_MySql.sql" target="lib\net461\MySql" />
```

Directory.Build.props
attilah: Consistency is my reason to have them at a single place, we can change that, but...

Test.cmd
attilah: Nice catch!

src/OrleansCounterControl/Microsoft.Orleans.CounterControl.targets
jdom: but the thing is that we DO support our nugets in VS2015. We just stopped supporting opening the Orleans solution itself in VS2015. This file file ends up in the nuget package that users consume.

jdom: Talked in person. Until we ship that, let's at least remove which ToolsVersion is used, so that it implicitly uses the one where the solution is running

attilah: Removed from all the 3 targets files.

Directory.Build.props
jdom: Ok, I'm fine for now, we can revisit in the future if it proves unwieldy... for now it's just a few references

Test.cmd
```diff
@@ -1,28 +1,37 @@
...
+set CMDHOME=%CMDHOME:~0,-1%
+
+pushd "%CMDHOME%"
+@cd
+
+SET TestResultDir=%CMDHOME%\Binaries\%CONFIGURATION%\TestResults
+
+if not exist %TestResultDir% md %TestResultDir%
+
+SET _Directory=bin\%CONFIGURATION%\net461\win10-x64
```
jdom: is win10-x64 appropriate here? is there a less hardcody approach we could take?
if it works in both CI servers (which apparently does), I don't feel very strongly for now if it's not easy to generalize, since we will probably just replace this with a call to `dotnet test` in the near future

jdom:
Please don't merge yet, we are trying to troubleshoot a few issues when building in VS

ReubenBond:
builds locally in clean VS with F# tools installed

ReubenBond:

attilah:
@dotnet-bot test please

ReubenBond:

PR #3237, opened 2017-07-19
Lease based queue balancer

xiazen:
1. added LeaseBasedQueueBalancer, it will try to renew its leases and acquire or release leases to meet its responsibility, in initialization,  every 30 sec, when received siloStatusChangeNotification
2. added AzureBlobLeaseProvider, it will need to be configured through DI, see test as example
3. added tests for both

### DIFF
--- a/src/Orleans/Providers/ILeaseProvider.cs
+++ b/src/Orleans/Providers/ILeaseProvider.cs
@@ -103,6 +103,23 @@ namespace Orleans.LeaseProviders
         /// Duration of the lease
         /// </summary>
         public TimeSpan Duration { get; set; }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        public LeaseRequest()
+        { }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="resourceKey"></param>
+        /// <param name="duration"></param>
+        public LeaseRequest(string resourceKey, TimeSpan duration)
+        {
+            this.ResourceKey = resourceKey;
+            this.Duration = duration;
+        }
     }
 
     /// <summary>
--- a/src/Orleans/Providers/ILeaseProvider.cs
+++ b/src/Orleans/Providers/ILeaseProvider.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Orleans.Concurrency;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
@@ -9,6 +10,7 @@ namespace Orleans.LeaseProviders
     /// <summary>
     /// Acquired lease
     /// </summary>
+    [Immutable]
     public class AcquiredLease 
     {
         /// <summary>
@@ -47,6 +49,7 @@ namespace Orleans.LeaseProviders
     /// <summary>
     /// AcquireLeaseResult class, which demonstrates result of acquiring or renewing lease operation
     /// </summary>
+    [Immutable]
     public class AcquireLeaseResult
     {
         /// <summary>
@@ -93,6 +96,7 @@ namespace Orleans.LeaseProviders
     /// <summary>
     /// Lease request where you can specify ResourceKey and duration of your lease. 
     /// </summary>
+    [Immutable]
     public class LeaseRequest
     {
         /// <summary>
@@ -148,11 +152,5 @@ namespace Orleans.LeaseProviders
         /// <param name="aquiredLeases"></param>
         /// <returns></returns>
         Task Release(string category, AcquiredLease[] aquiredLeases);
-    }
-
-    public class ResourceCategory
-    {
-        public const string Streaming = "Streaming";
-        public const string Storage = "Storage";
     }
 }
--- a/src/Orleans/Providers/ILeaseProvider.cs
+++ b/src/Orleans/Providers/ILeaseProvider.cs
@@ -27,6 +27,21 @@ namespace Orleans.LeaseProviders
         /// Caller side start time for this lease, which is when the lease is acquired or renewed
         /// </summary>
         public DateTime StartTimeUtc { get; }
+
+        /// <summary>
+        /// Consructor
+        /// </summary>
+        /// <param name="resourceKey"></param>
+        /// <param name="duration"></param>
+        /// <param name="token"></param>
+        /// <param name="startTimeUtc"></param>
+        public AcquiredLease(string resourceKey, TimeSpan duration, string token, DateTime startTimeUtc)
+        {
+            this.ResourceKey = resourceKey;
+            this.Duration = duration;
+            this.Token = token;
+            this.StartTimeUtc = startTimeUtc;
+        }
     }
 
     /// <summary>
@@ -37,7 +52,7 @@ namespace Orleans.LeaseProviders
         /// <summary>
         /// Acquired lease, which will be null if acquire or renew operation failed.
         /// </summary>
-        AcquiredLease AcquiredLease { get; }
+        public AcquiredLease AcquiredLease { get; }
         /// <summary>
         /// Response status
         /// </summary>
@@ -46,6 +61,13 @@ namespace Orleans.LeaseProviders
         /// If acquiring or renewing the lease failed, this is the exception which caused it. This field would be null if operation succeed.
         /// </summary>
         public Exception FailureException { get; }
+
+        public AcquireLeaseResult(AcquiredLease acquiredLease, ResponseCode statusCode, Exception failureException)
+        {
+            this.AcquiredLease = acquiredLease;
+            this.StatusCode = statusCode;
+            this.FailureException = failureException;
+        }
     }
 
     public enum ResponseCode
@@ -111,4 +133,9 @@ namespace Orleans.LeaseProviders
         Task Release(string category, AcquiredLease[] aquiredLeases);
     }
 
+    public class ResourceCategory
+    {
+        public const string Streaming = "Streaming";
+        public const string Storage = "Storage";
+    }
 }
--- a/src/Orleans/Providers/ILeaseProvider.cs
+++ b/src/Orleans/Providers/ILeaseProvider.cs
@@ -96,7 +96,6 @@ namespace Orleans.LeaseProviders
     /// <summary>
     /// Lease request where you can specify ResourceKey and duration of your lease. 
     /// </summary>
-    [Immutable]
     public class LeaseRequest
     {
         /// <summary>
--- a/src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
+++ b/src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
@@ -15,28 +15,27 @@ namespace Orleans.LeaseProviders
     public class AzureBlobLeaseProviderConfig
     {
         public string DataConnectionString { get; set; }
+        public string BlobContainerName { get; set; }
     }
     public class AzureBlobLeaseProvider : ILeaseProvider
     {
         private CloudBlobContainer container;
-        private GlobalConfiguration globalConfig;
+        private AzureBlobLeaseProviderConfig providerConfig;
         private CloudBlobClient blobClient;
-        public AzureBlobLeaseProvider(AzureBlobLeaseProviderConfig config, GlobalConfiguration globalConfig )
+        public AzureBlobLeaseProvider(AzureBlobLeaseProviderConfig config)
         {
             var account = CloudStorageAccount.Parse(config.DataConnectionString);
             this.blobClient = account.CreateCloudBlobClient();
-            this.globalConfig = globalConfig;
+            this.providerConfig = config;
         }
 
-        private Task InitContainerIfNotExistsAsync()
+        private async Task InitContainerIfNotExistsAsync()
         {
             if (this.container == null)
             {
-                var containerName = $"Cluster-{globalConfig.DeploymentId}-{ResourceCategory.Streaming}-{typeof(AzureBlobLeaseProvider).Name}";
-                this.container = blobClient.GetContainerReference(containerName);
-                return this.container.CreateIfNotExistsAsync();
+                this.container = blobClient.GetContainerReference(this.providerConfig.BlobContainerName);
+                await this.container.CreateIfNotExistsAsync().ConfigureAwait(false); 
             }
-            return Task.CompletedTask;
         }
 
         public async Task<AcquireLeaseResult[]> Acquire(string category, LeaseRequest[] leaseRequests)
@@ -51,37 +50,35 @@ namespace Orleans.LeaseProviders
             return await Task.WhenAll(tasks);
         }
 
-        private ResponseCode MapHttpResponseCode(int httpResponseCode)
-        {
-            //This mapping is based on references : https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes
-            // https://docs.microsoft.com/en-us/rest/api/storageservices/Lease-Blob?redirectedfrom=MSDN
-            switch (httpResponseCode)
-            {
-                case 409: return ResponseCode.LeaseNotAvailable;
-                case 412: return ResponseCode.InvalidToken;
-                case 200:
-                case 201:
-                case 202: return ResponseCode.OK;
-                default: return ResponseCode.TransientFailure;
-            }
-        }
-
         private string GetBlobName(string category, string resourceKey)
         {
-            return $"{category}-{resourceKey}";
+            return $"{category.ToLower()}-{resourceKey.ToLower()}.json";
         }
 
         private async Task<AcquireLeaseResult> Acquire(string category, LeaseRequest leaseRequest)
         {
-            var blob = this.container.GetBlobReference(GetBlobName(category, leaseRequest.ResourceKey));
             try
             {
+                var blob = this.container.GetBlockBlobReference(GetBlobName(category, leaseRequest.ResourceKey));
+                blob.Properties.ContentType = "application/json";
+                //create this blob
+                await blob.UploadTextAsync("blob");
                 var leaseId = await blob.AcquireLeaseAsync(leaseRequest.Duration);
                 return new AcquireLeaseResult(new AcquiredLease(leaseRequest.ResourceKey, leaseRequest.Duration, leaseId, DateTime.UtcNow), ResponseCode.OK, null);
             }
             catch (StorageException e)
             {
-                return new AcquireLeaseResult(null, MapHttpResponseCode(e.RequestInformation.HttpStatusCode), e);
+                ResponseCode statusCode;
+                //This mapping is based on references : https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes
+                // https://docs.microsoft.com/en-us/rest/api/storageservices/Lease-Blob?redirectedfrom=MSDN
+                switch (e.RequestInformation.HttpStatusCode)
+                {
+                    case 404:
+                    case 409: 
+                    case 412: statusCode = ResponseCode.LeaseNotAvailable; break;
+                    default: statusCode = ResponseCode.TransientFailure; break;
+                }
+                return new AcquireLeaseResult(null, statusCode, e);
             }
         }
 
@@ -126,7 +123,17 @@ namespace Orleans.LeaseProviders
             }
             catch (StorageException e)
             {
-                return new AcquireLeaseResult(null, MapHttpResponseCode(e.RequestInformation.HttpStatusCode), e);
+                ResponseCode statusCode;
+                //This mapping is based on references : https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes
+                // https://docs.microsoft.com/en-us/rest/api/storageservices/Lease-Blob?redirectedfrom=MSDN
+                switch (e.RequestInformation.HttpStatusCode)
+                {
+                    case 404:
+                    case 409:
+                    case 412: statusCode = ResponseCode.InvalidToken; break;
+                    default: statusCode = ResponseCode.TransientFailure; break;
+                }
+                return new AcquireLeaseResult(null, statusCode, e);
             }
         }
     }
--- a/src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
+++ b/src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
@@ -33,15 +33,16 @@ namespace Orleans.LeaseProviders
         {
             if (this.container == null)
             {
-                this.container = blobClient.GetContainerReference(this.providerConfig.BlobContainerName);
-                await this.container.CreateIfNotExistsAsync().ConfigureAwait(false); 
+                var tmpContainer = blobClient.GetContainerReference(this.providerConfig.BlobContainerName);
+                await tmpContainer.CreateIfNotExistsAsync().ConfigureAwait(false);
+                this.container = tmpContainer;
             }
         }
 
         public async Task<AcquireLeaseResult[]> Acquire(string category, LeaseRequest[] leaseRequests)
         {
             await InitContainerIfNotExistsAsync();
-            var tasks = new List<Task<AcquireLeaseResult>>();
+            var tasks = new List<Task<AcquireLeaseResult>>(leaseRequests.Length);
             foreach (var leaseRequest in leaseRequests)
             {
                 tasks.Add(Acquire(category, leaseRequest));
@@ -82,11 +83,11 @@ namespace Orleans.LeaseProviders
             }
         }
 
-        public async Task Release(string category, AcquiredLease[] aquiredLeases)
+        public async Task Release(string category, AcquiredLease[] acquiredLeases)
         {
             await InitContainerIfNotExistsAsync();
-            var tasks = new List<Task>();
-            foreach (var acquiredLease in aquiredLeases)
+            var tasks = new List<Task>(acquiredLeases.Length);
+            foreach (var acquiredLease in acquiredLeases)
             {
                 tasks.Add(Release(category, acquiredLease));
             }
@@ -99,11 +100,11 @@ namespace Orleans.LeaseProviders
             return blob.ReleaseLeaseAsync(AccessCondition.GenerateLeaseCondition(acquiredLease.Token));
         }
 
-        public async Task<AcquireLeaseResult[]> Renew(string category, AcquiredLease[] aquiredLeases)
+        public async Task<AcquireLeaseResult[]> Renew(string category, AcquiredLease[] acquiredLeases)
         {
             await InitContainerIfNotExistsAsync();
-            var tasks = new List<Task<AcquireLeaseResult>>();
-            foreach (var acquiredLease in aquiredLeases)
+            var tasks = new List<Task<AcquireLeaseResult>>(acquiredLeases.Length);
+            foreach (var acquiredLease in acquiredLeases)
             {
                 tasks.Add(Renew(category, acquiredLease));
             }
--- a/src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
+++ b/src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
@@ -1,0 +1,133 @@
+﻿using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+using Orleans.Runtime.Configuration;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Orleans.LeaseProviders
+{
+    public class AzureBlobLeaseProviderConfig
+    {
+        public string DataConnectionString { get; set; }
+    }
+    public class AzureBlobLeaseProvider : ILeaseProvider
+    {
+        private CloudBlobContainer container;
+        private GlobalConfiguration globalConfig;
+        private CloudBlobClient blobClient;
+        public AzureBlobLeaseProvider(AzureBlobLeaseProviderConfig config, GlobalConfiguration globalConfig )
+        {
+            var account = CloudStorageAccount.Parse(config.DataConnectionString);
+            this.blobClient = account.CreateCloudBlobClient();
+            this.globalConfig = globalConfig;
+        }
+
+        private Task InitContainerIfNotExistsAsync()
+        {
+            if (this.container == null)
+            {
+                var containerName = $"Cluster-{globalConfig.DeploymentId}-{ResourceCategory.Streaming}-{typeof(AzureBlobLeaseProvider).Name}";
+                this.container = blobClient.GetContainerReference(containerName);
+                return this.container.CreateIfNotExistsAsync();
+            }
+            return Task.CompletedTask;
+        }
+
+        public async Task<AcquireLeaseResult[]> Acquire(string category, LeaseRequest[] leaseRequests)
+        {
+            await InitContainerIfNotExistsAsync();
+            var tasks = new List<Task<AcquireLeaseResult>>();
+            foreach (var leaseRequest in leaseRequests)
+            {
+                tasks.Add(Acquire(category, leaseRequest));
+            }
+            //Task.WhenAll will return results for each task in an array, in the same order of supplied tasks
+            return await Task.WhenAll(tasks);
+        }
+
+        private ResponseCode MapHttpResponseCode(int httpResponseCode)
+        {
+            //This mapping is based on references : https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes
+            // https://docs.microsoft.com/en-us/rest/api/storageservices/Lease-Blob?redirectedfrom=MSDN
+            switch (httpResponseCode)
+            {
+                case 409: return ResponseCode.LeaseNotAvailable;
+                case 412: return ResponseCode.InvalidToken;
+                case 200:
+                case 201:
+                case 202: return ResponseCode.OK;
+                default: return ResponseCode.TransientFailure;
+            }
+        }
+
+        private string GetBlobName(string category, string resourceKey)
+        {
+            return $"{category}-{resourceKey}";
+        }
+
+        private async Task<AcquireLeaseResult> Acquire(string category, LeaseRequest leaseRequest)
+        {
+            var blob = this.container.GetBlobReference(GetBlobName(category, leaseRequest.ResourceKey));
+            try
+            {
+                var leaseId = await blob.AcquireLeaseAsync(leaseRequest.Duration);
+                return new AcquireLeaseResult(new AcquiredLease(leaseRequest.ResourceKey, leaseRequest.Duration, leaseId, DateTime.UtcNow), ResponseCode.OK, null);
+            }
+            catch (StorageException e)
+            {
+                return new AcquireLeaseResult(null, MapHttpResponseCode(e.RequestInformation.HttpStatusCode), e);
+            }
+        }
+
+        public async Task Release(string category, AcquiredLease[] aquiredLeases)
+        {
+            await InitContainerIfNotExistsAsync();
+            var tasks = new List<Task>();
+            foreach (var acquiredLease in aquiredLeases)
+            {
+                tasks.Add(Release(category, acquiredLease));
+            }
+            await Task.WhenAll(tasks);
+        }
+
+        private Task Release(string category, AcquiredLease acquiredLease)
+        {
+            var blob = this.container.GetBlobReference(GetBlobName(category, acquiredLease.ResourceKey));
+            return blob.ReleaseLeaseAsync(AccessCondition.GenerateLeaseCondition(acquiredLease.Token));
+        }
+
+        public async Task<AcquireLeaseResult[]> Renew(string category, AcquiredLease[] aquiredLeases)
+        {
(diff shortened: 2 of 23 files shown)


src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -27,6 +27,21 @@ public class AcquiredLease
...
         public DateTime StartTimeUtc { get; }
+
+        /// <summary>
+        /// Consructor
+        /// </summary>
+        /// <param name="resourceKey"></param>
+        /// <param name="duration"></param>
+        /// <param name="token"></param>
+        /// <param name="startTimeUtc"></param>
+        public AcquiredLease(string resourceKey, TimeSpan duration, string token, DateTime startTimeUtc)
```
jason-bragg: This makes these classes immutable, which is good, imo.  Since they are immutable, they can be marked as immutable (attribute) which will make grain calls using them faster.

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -81,6 +103,23 @@ public class LeaseRequest
...
+        /// </summary>
+        public LeaseRequest()
+        { }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="resourceKey"></param>
+        /// <param name="duration"></param>
+        public LeaseRequest(string resourceKey, TimeSpan duration)
```
jason-bragg: If we can make this type immutable like the rest, that would be great :)

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -111,4 +150,9 @@ public interface ILeaseProvider
         Task Release(string category, AcquiredLease[] aquiredLeases);
     }
 
+    public class ResourceCategory
```
jason-bragg: I don't think we want the categories as part of the interface definition.  I suggest each category be defined by the system responsible for that category.  So the queue balancer would define StreamQueues category, for instance.

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingManager.cs
```diff
@@ -102,6 +102,7 @@ internal class PersistentStreamPullingManager : SystemTarget, IPersistentStreamP
             {
                 queuePrintTimer.Dispose();
             }
+            (this.queueBalancer as IDisposable)?.Dispose();
```
jason-bragg: nit: should probably set both disposables to null after disposing..?

src/Orleans/Providers/ILeaseProvider.cs
xiazen: good idea

src/Orleans/Providers/ILeaseProvider.cs
xiazen: Make sense, I will move this out of ILeaseProvider interface. Was hesitating then

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingManager.cs
xiazen: oopps, you are def correct . I forgot

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+                await this.ReleaseLeasesToMeetResponsibility();
+            }
+        }
+
+        private async Task ReleaseLeasesToMeetResponsibility()
+        {
+            var queueCountToRelease = this.myQueues.Count - this.maximumRespobsibility;
+            if (queueCountToRelease <= 0)
+                return;
+            var queuesToGiveUp = this.myQueues.GetRange(0, queueCountToRelease);
```
jason-bragg: Take(queueCountToRelease)?

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+                    }
+                }
+                //if reached expectedTotalLeaseCount
+                if (this.myQueues.Count >= expectedTotalLeaseCount)
+                {
+                    break;
+                }
+            }
+
+            this.logger.Info($"AcquireLeasesToMeetExpection: finished. Now own {this.myQueues.Count} queues. Used attemps : {attempts}, Current minimumReponsibility : {this.minimumResponsibilty}, current maximumResponsibility : {this.maximumRespobsibility}");
```
jason-bragg: AcquireLeasesToMeetExpectation?

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            this.myQueues = updatedQueues;
+            this.logger.Info($"RenewLeases: finished, currently own queues : {this.myQueues.Count}, current minimumResponsibilty : {this.minimumResponsibilty}, current maximunResponsibilty : {this.maximumRespobsibility}");
+        }
+
+        private void CalculateResponsibility()
+        {
+            int activeBuckets = 0;
+            if (isStarting)
+            {
+                activeBuckets = this.deploymentConfig.GetAllSiloNames().Count;
```
jason-bragg: Are we confident the list of silo names will properly reflect deployment in autoscale environment? I suspect buckets should be the number of active silos in the cluster, regardless of deployment configuration.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            this.queueSelector = new RoundRobinSelector<QueueId>(this.allQueues.ToList());
+            return MaintainAndBalanceQueues(null);
+        }
+
+        public override IEnumerable<QueueId> GetMyQueues()
+        {
+            return this.myQueues.Select(queue => queue.QueueId);
+        }
+
+        private async Task MaintainAndBalanceQueues(object state)
```
jason-bragg: I think we need to call NotifyListeners at the end of these timer calls any time myQueues changes.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            if (queueMapper == null)
+            {
+                throw new ArgumentNullException("queueMapper");
+            }
+            this.allQueues = new ReadOnlyCollection<QueueId>(queueMapper.GetAllQueues().ToList());
+            this.siloMaturityPeriod = siloMaturityPeriod;
+            NotifyAfterStart().Ignore();
+            //make lease renew frequency to be every half of lease time, to avoid renew failing due to timing issues, race condition or clock difference. 
+            this.renewLeaseTimer = new AsyncTaskSafeTimer(this.MaintainAndBalanceQueues, null, this.siloMaturityPeriod, this.leaseLength.Divide(2));
```
jason-bragg: Changes to myqueues were driven by silo status changes calls.  I don't think these timer callbacks will be done on the same thread, so we've a potential threading issue here, right?

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            {
+                this.lastSelection = (++this.lastSelection) % (this.resources.Count - 1);
+                if(!existingSelection.Contains(this.resources[this.lastSelection]))
+                    selection.Add(this.resources[this.lastSelection]);
+            }
+            return selection;
+        }
+    }
+
+    public class AzureDeploymentLeaseBasedBalancer : LeaseBasedQueueBalancer
```
jason-bragg: Should this live in azure utils?

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: it means acquire leases up to the expected count... is this a bad name? 

A balancer is expected to have a specific amount of leases. get it?

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: ?? what do you mean? 
queuesToGiveup is the first queueCountToRelease queues

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: nm, works as designed.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: I was suggesting using the Take(int) link function, as it would be more readable, but nm, this is more efficient anyway.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: I was unclear :/
I was trying to point out a typo.
AcquireLeasesToMeetExpection => AcquireLeasesToMeetExpectation

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: will fix this based on our discussion

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: I think you are right. will move it

xiazen: found out we cannot move this to azure utils. Since StreamQueueBalancerType referenced this type, but its csproj OrleansRuntime cannot take azure utils as dependency.

xiazen: With that being said, we can solve this problem by restructure our assembly, which I think it is part of the abstraction work.

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            return $"{category.ToLower()}-{resourceKey.ToLower()}.json";
+        }
+
+        private async Task<AcquireLeaseResult> Acquire(string category, LeaseRequest leaseRequest)
+        {
+            try
+            {
+                var blob = this.container.GetBlockBlobReference(GetBlobName(category, leaseRequest.ResourceKey));
+                blob.Properties.ContentType = "application/json";
```
ReubenBond: is this blob actually valid JSON? it doesn't seem like it. If you changed `"blob"` to `"true"` or `"0"` below, then that's valid JSON

ReubenBond: Should this call be something like ~CreateIfNotExist? It seems strange/erroneous to be always recreating the blob

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
```diff
@@ -0,0 +1,140 @@
...
+            this.blobClient = account.CreateCloudBlobClient();
+            this.providerConfig = config;
+        }
+
+        private async Task InitContainerIfNotExistsAsync()
+        {
+            if (this.container == null)
+            {
+                this.container = blobClient.GetContainerReference(this.providerConfig.BlobContainerName);
+                await this.container.CreateIfNotExistsAsync().ConfigureAwait(false);
```
ReubenBond: If this call fails then it can never be retried. Seems like we should only assign `this.container` after this has completed successfully

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
```diff
@@ -0,0 +1,140 @@
...
+                return new AcquireLeaseResult(new AcquiredLease(leaseRequest.ResourceKey, leaseRequest.Duration, leaseId, DateTime.UtcNow), ResponseCode.OK, null);
+            }
+            catch (StorageException e)
+            {
+                ResponseCode statusCode;
+                //This mapping is based on references : https://docs.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes
+                // https://docs.microsoft.com/en-us/rest/api/storageservices/Lease-Blob?redirectedfrom=MSDN
+                switch (e.RequestInformation.HttpStatusCode)
+                {
+                    case 404:
```
ReubenBond: Is there anywhere where these are defined as an enum which we could use here instead? If not, then that's fine.

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
```diff
@@ -0,0 +1,140 @@
...
+                    default: statusCode = ResponseCode.TransientFailure; break;
+                }
+                return new AcquireLeaseResult(null, statusCode, e);
+            }
+        }
+
+        public async Task Release(string category, AcquiredLease[] aquiredLeases)
+        {
+            await InitContainerIfNotExistsAsync();
+            var tasks = new List<Task>();
```
ReubenBond: initialize the capacity of the list to the length of `acquiredLeases`

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                tasks.Add(Renew(category, acquiredLease));
+            }
+            //Task.WhenAll will return results for each task in an array, in the same order of supplied tasks
+            return await Task.WhenAll(tasks);
+        }
+
+        private async Task<AcquireLeaseResult> Renew(string category, AcquiredLease acquiredLease)
+        {
+            var blob = this.container.GetBlobReference(GetBlobName(category, acquiredLease.ResourceKey));
```
ReubenBond: Maybe this is pre-mature optimization, but you could cache the blobName on a property in a prviate subclass of `AcquiredLease`. You could go all the way and cache the blob ref.

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Orleans.Streams
+{
+    /// <summary>
+    /// Base class for StreamQueueBalancer
+    /// </summary>
+    public abstract class QueueBalancerBaseClass : IStreamQueueBalancer
```
ReubenBond: Usually we leave out the `Class` suffix

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
```diff
@@ -0,0 +1,72 @@
...
+                throw new ArgumentNullException("observer");
+            }
+            lock (queueBalanceListeners)
+            {
+                return queueBalanceListeners.Contains(observer) && queueBalanceListeners.Remove(observer);
+            }
+        }
+
+
+        public static List<string> GetActiveSilos(ISiloStatusOracle siloStatusOracle, ConcurrentDictionary<SiloAddress, bool> immatureSilos)
```
ReubenBond: This method would benefit from some docs. It's not clear to me what the purpose of `immatureBit` is or why this API specifically requires a concurrent dictionary

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Threading.Tasks;
+
+namespace Orleans.Streams
+{
+    /// <summary>
+    /// Base class for StreamQueueBalancer
+    /// </summary>
+    public abstract class QueueBalancerBaseClass : IStreamQueueBalancer
+    {
+        protected readonly List<IStreamQueueBalanceListener> queueBalanceListeners;
```
ReubenBond: Does this actually need to be `protected`? Could it be `private` instead? If it's going to be exposed, please document the requirement that it's locked before access

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
```diff
@@ -0,0 +1,72 @@
...
+                }
+                queueBalanceListeners.Add(observer);
+                return true;
+            }
+        }
+        public virtual bool UnSubscribeFromQueueDistributionChangeEvents(IStreamQueueBalanceListener observer)
+        {
+            if (observer == null)
+            {
+                throw new ArgumentNullException("observer");
```
ReubenBond: nit: you can use `nameof(observer)` here

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+            lock (queueBalanceListeners)
+            {
+                return queueBalanceListeners.Contains(observer) && queueBalanceListeners.Remove(observer);
```
ReubenBond: The `Contains` call here is superfluous - or am I missing something?

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+
+
+        public static List<string> GetActiveSilos(ISiloStatusOracle siloStatusOracle, ConcurrentDictionary<SiloAddress, bool> immatureSilos)
+        {
```
ReubenBond: Does this method need to exist and does it need to be shared? If it does exist, does it need to be static? Usually, I'd expect `ISiloStatusOracle` to be injected and not directly passed to the method.

I realize this method was just moved from existing code, but the question is more pertinent now that it's becoming public API.

src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
```diff
@@ -76,6 +76,7 @@ internal class DeploymentBasedQueueBalancer : ISiloStatusListener, IStreamQueueB
             ISiloStatusOracle siloStatusOracle,
             IDeploymentConfiguration deploymentConfig,
             bool isFixed)
+            :base()
```
ReubenBond: nit: superfluous

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            this.lastSelection = -1;
+        }
+
+        /// <summary>
+        /// Try to select certain count of resources from resource list, which doesn't overlap with existing resources
+        /// </summary>
+        /// <param name="newSelectionCount"></param>
+        /// <param name="existingSelection"></param>
+        /// <returns></returns>
+        public IEnumerable<T> NextSelection(int newSelectionCount, IEnumerable<T> existingSelection)
```
ReubenBond: The constructor takes a `List<T>`, but this method takes/returns an `IEnumerable<T>` and then enumerates it many times (allocating an enumerator every time). Suggest you use `List<T>` instead of `IEnumerable<T>`. Alternatively, given this method works on contiguous collections, you only need to find the first and last item (first + count) in the existing selection and avoid all items in that range. So you could have a 'SelectionSpan` type which recirds that info.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            var selection = new List<T>();
+            while (selection.Count < newSelectionCount)
+            {
+                this.lastSelection = (++this.lastSelection) % (this.resources.Count - 1);
+                if(!existingSelection.Contains(this.resources[this.lastSelection]))
+                    selection.Add(this.resources[this.lastSelection]);
```
ReubenBond: If `newSelectionCount > resources.Count / 2` then the result will contain duplicates to fill the difference - is that correct?

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+
+        /// <summary>
+        /// Try to select certain count of resources from resource list, which doesn't overlap with existing resources
+        /// </summary>
+        /// <param name="newSelectionCount"></param>
+        /// <param name="existingSelection"></param>
+        /// <returns></returns>
+        public IEnumerable<T> NextSelection(int newSelectionCount, IEnumerable<T> existingSelection)
+        {
+            var selection = new List<T>();
```
ReubenBond: Given you know the size of the result, you can initialize this list to that size and avoid additional allocations as the list resizes itself.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            return selection;
+        }
+    }
+
+    public class AzureDeploymentLeaseBasedBalancer : LeaseBasedQueueBalancer
+    {
+        public AzureDeploymentLeaseBasedBalancer(ILeaseProvider leaseProvider, ISiloStatusOracle siloStatusOracle,
+            IServiceProvider serviceProvider)
+            : base(leaseProvider, siloStatusOracle, DeploymentBasedQueueBalancerUtils.CreateDeploymentConfigForAzure(serviceProvider),
+                  serviceProvider.GetRequiredService<Factory<string, Logger>>())
```
ReubenBond: `Factory<string, Logger>` can be injected directly into the ctor, instead of injecting `IServiceProvider` (which is usually considered an anti-pattern, even though we do it in some places out of necessity) and pulling it from there.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            }
+            return selection;
+        }
+    }
+
+    public class AzureDeploymentLeaseBasedBalancer : LeaseBasedQueueBalancer
+    {
+        public AzureDeploymentLeaseBasedBalancer(ILeaseProvider leaseProvider, ISiloStatusOracle siloStatusOracle,
+            IServiceProvider serviceProvider)
+            : base(leaseProvider, siloStatusOracle, DeploymentBasedQueueBalancerUtils.CreateDeploymentConfigForAzure(serviceProvider),
```
ReubenBond: This CreateDeploymentConfigForAzure method is so ugly. When this class is moved to az utils, please kill it and avoid injecting the service provider (inject `IDeploymentConfiguration` its return value instead)

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            }
+        }
+
+        public override Task Initialize(string strProviderName, IStreamQueueMapper queueMapper, TimeSpan siloMaturityPeriod)
+        {
+            if (queueMapper == null)
+            {
+                throw new ArgumentNullException("queueMapper");
+            }
+            this.allQueues = new ReadOnlyCollection<QueueId>(queueMapper.GetAllQueues().ToList());
```
ReubenBond: Maybe this should be a regular `List<QueueId>`, since you're converting it to a list below anyway, when creating the `RoundRobinSelector`

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            {
+                //select new queues to acquire
+                var expectedQueues = this.queueSelector.NextSelection(leasesToAquire, this.myQueues.Select(queue=>queue.QueueId)).ToList();
+                var leaseRequests = expectedQueues.Select(queue => new LeaseRequest() {
+                    ResourceKey = queue.ToString(),
+                    Duration = this.leaseLength
+                });
+                var results = await this.leaseProvider.Acquire(ResourceCategory.Streaming, leaseRequests.ToArray());
+                //add successfully acquired queue to myQueues list
+                for (int i = 0; i < results.Count(); i++)
```
ReubenBond: `.Length` instead of `.Count()`

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,345 @@
...
+            this.logger.Info($"AcquireLeasesToMeetExpection : Try to acquire {leasesToAquire} queues");
+            while (attempts ++ <= maxAttempts && leasesToAquire > 0)
+            {
+                //select new queues to acquire
+                var expectedQueues = this.queueSelector.NextSelection(leasesToAquire, this.myQueues.Select(queue=>queue.QueueId)).ToList();
+                var leaseRequests = expectedQueues.Select(queue => new LeaseRequest() {
+                    ResourceKey = queue.ToString(),
+                    Duration = this.leaseLength
+                });
+                var results = await this.leaseProvider.Acquire(ResourceCategory.Streaming, leaseRequests.ToArray());
```
ReubenBond: maybe I'm wrong, but I would have expected this to include the name of the provider, possibly prefixed with the streaming category.

Otherwise there will be clashes between different stream providers configured with the same lease provider. Is that true?

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
xiazen: Maybe I didn't look hard enough. But I don't think blob ref itself has a CreateIfNotExists API. Upload something to it is the way to create/overwrite it. Upload again won't recreate it, the blob will just be overwritten. 
But I think performance wise, CreateIfNotExists is not much better than Upload. Both of them cost one call to BlobService.

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
xiazen: yeah that can be better.

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
xiazen: We can use HttpStatusCode enum. But since it is an enum and `e.RequestInformation.HttpStatusCode` is an int , the code need to parse enum to int to use in switch block or using a "if else" block rather than switch, both of which are more performance heavier. 

And azure API e.RequestInformation.HttpStatusCode is an int. So we can say we are following azure standard. :)

src/OrleansAzureUtils/Providers/Lease/AzureBlobLeaseProvider.cs
xiazen: renew won't happen very frequent. I would suggest to use 30 - 60 sec renew frequency. So I won't worry about this.

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
xiazen: If you look at its child class, such as LeaseBasedQueueBalancer. You will have more clue.
Basically, whenever there's a SiloStatusNotification, code calls siloStatusOracle asking for active silo information. If there's new silo which was recorded in immatureSilos, then the code would wait `siloMaturityPeriod `to mark it as mature. If before it is marked as mature, GetActiveSilos is called, and siloStatusOracle report those new silo as active, we still think they are immature. 

I think it is a conservative way to deal with inconsistent view in a distributed cluster. Since this local siloStatusOracle thinks this silo is active doesn't mean other silo thinks it is active too. So we'd better wait for `siloMaturityPeriod`. 

It requires a concurrentDic because there could be under multi-thread scenario. For example, siloStatusOracle trigger SiloStatusChangeNotification will be running in one thread, and pullingAgentManager will be triggering GetMyQueues in another thread, both of which will need to accees immatureSIlos

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
xiazen: will add docs. and it need to be protected since its child classes are accessing it

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
xiazen: Yeah  you are right. I move this code here from somewhere else. Wasn't rechecking the quality of old code.

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
xiazen: it is like a utility method . but it is shared by some balancer classes, but not shared by all Balancer class. So this is why it is a static class.

xiazen: Yeah I think we should move this method to another less public place. Like a internal utility class.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: I'm with you. But this class cannot really be moved to azure utils. Because there's class in OrleansRuntime reference it, which is probably why it is done in this ugly way in the first place. But maybe it will be better after abstraction work is done.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: resourceKey is queueId, which contains providerName . So there won't be crashes :)

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: Sure.

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
ReubenBond: Some of these concerns are implementation specific (like the concurrency issue). Could this be moved into the class which uses it instead, or is this something which every implementation of this class will need?

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: why? it just round robin half a circle.

xiazen: And the caller (balancer) won't likely do so, since it can contain just a subset of the resources to a upper limit. So in short, taking a context this selector is used in, I'm no seeing a duplication being picked. And if this selector is used in other scenario for some reason, then the caller needs to respect what this class is doing. Since it is roundRobin :) it will have duplication eventually

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
ReubenBond: 😭

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
xiazen: Yes I will move it to a internal utility class. I mentioned that in one of your comments I think.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: There's no ReadOnlyList :( 

I will make the thing in RoundRobinSelector to be ReadOnlyCollection to avoid extra allocation,

xiazen:

xiazen:

src/Orleans/Providers/ILeaseProvider.cs
```diff
@@ -81,6 +107,23 @@ public class LeaseRequest
         /// Duration of the lease
         /// </summary>
         public TimeSpan Duration { get; set; }
```
jason-bragg: This is only immutable if their is no 'set' on any of the properties after construction.

xiazen: This is the LeaseRequest class which should be able to be created and modified by the caller. It is marked as immutable(attribute) so it is immutable through the wire

xiazen: Offline discussion agree to remove [Immutable] from the class. It is a trade-off between usability and minor perf cost

xiazen:
ping.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -0,0 +1,413 @@
...
+            }
+            this.minimumResponsibilty = this.allQueues.Count / activeBuckets;
+            //if allQueues count is divisible by active bukets, then every bucket should take the same count of queues, otherwise, there should be one bucket take 1 more queue
+            if (this.allQueues.Count % activeBuckets == 0)
+                this.maximumRespobsibility = this.minimumResponsibilty;
+            else this.maximumRespobsibility = this.minimumResponsibilty + 1;
+        }
+
+        /// <inheritdoc/>
+        public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
```
jason-bragg: I thought we'd agreed that we didn't need notifications from the status oracle any more?  And that they may introduce threading issues if they alter state..

src/OrleansRuntime/Streams/QueueBalancer/QueueBalancerBaseClass.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// Basically, whenever there's a SiloStatusNotification, balancer calls siloStatusOracle asking for active silo information. 
+        /// If there's new silo which was not recorded in immatureSilos, then balancer will first mark it as immature and then wait siloMaturityPeriod to mark it as mature. 
+        /// If GetActiveSilos is called before it is marked as mature, but siloStatusOracle report those new silo as active, we still think they are immature.
+        /// It is a conservative way to deal with inconsistent view in a distributed cluster.Since local siloStatusOracle thinks this silo is active 
+        /// doesn't mean other silo thinks it is active too. So we'd better wait for siloMaturityPeriod.
+        /// </summary>
+        /// <param name="siloStatusOracle"></param>
+        /// <param name="immatureSilos"></param>
+        /// <returns></returns>
+        public static List<string> GetActiveSilos(ISiloStatusOracle siloStatusOracle, ConcurrentDictionary<SiloAddress, bool> immatureSilos)
```
jason-bragg: Can/Should address this in different PR, but what this call does is wrong.  This is a pre-existing bug that I only noticed in this PR.  It is not related to your changes.

jason-bragg:
LG2M - A pre-existing bug was found that can/should be addressed in a separate PR, but this set of work looks solid.

src/OrleansRuntime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
xiazen: I kept this in order to only use the algorithm in GetActiveSilos, which I thought is the optimal/old way to do it. But you mentioned there's a bug there. So I think we can talk about that next week....

the treading issue we talked about is both timer thread and siloStatusOracleNotification thread can be calling MaintainAndBalanceQueues, hence the threading issue.... I changed siloStatusOracleNotification to not trigger `MaintainAndBalanceQueues` anymore... only timer will trigger `MaintainAndBalanceQueues`..... 

Anyway, I kept SiloStatusChangeNotification only because the old GetActiveSilo algorithm.... if that is a incorrect algorithm which has a bug... then no need to keep it.

xiazen:
I think we should merge this after the bug fix PR. so it is easier to cherry-pick the bug fix PR to 1.5.1

xiazen:
ping

PR #3241, opened 2017-07-20
Support custom silo names in Service Fabric integration

ReubenBond:
Just some minor thing which annoyed me during testing

### DIFF
--- a/src/OrleansServiceFabricUtils/OrleansCommunicationListener.cs
+++ b/src/OrleansServiceFabricUtils/OrleansCommunicationListener.cs
@@ -42,7 +42,8 @@ namespace Microsoft.Orleans.ServiceFabric
         /// </summary>
         /// <param name="context">The context.</param>
         /// <param name="configuration">The configuration.</param>
-        public OrleansCommunicationListener(ServiceContext context, ClusterConfiguration configuration)
+        /// <param name="siloName">The silo name, or <see langword="null"/> to have a name automatically selected.</param>
+        public OrleansCommunicationListener(ServiceContext context, ClusterConfiguration configuration, string siloName = null)
         {
             this.configuration = configuration;
             if (this.configuration == null)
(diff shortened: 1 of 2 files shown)


src/OrleansServiceFabricUtils/OrleansCommunicationListener.cs
```diff
@@ -42,7 +42,8 @@ public class OrleansCommunicationListener : ICommunicationListener
         /// </summary>
         /// <param name="context">The context.</param>
         /// <param name="configuration">The configuration.</param>
-        public OrleansCommunicationListener(ServiceContext context, ClusterConfiguration configuration)
+        /// <param name="siloName">The silo name, or <see langword="null"/> to have a name automatically selected.</param>
+        public OrleansCommunicationListener(ServiceContext context, ClusterConfiguration configuration, string siloName = null)
```
xiazen: Just curious. The siloName seems to be different than siloName defined in ClusterConfiguration. Why is that?

ReubenBond: Hm the `SiloName` you're talking about is in `ClusterConfiguration.Defaults`, I think. I believe that is not intended to be 'this' silo's name, it's just there because `Defaults` is also a `NodeConfiguration`, but `NodeConfiguration` also shouldn't need `SiloName`, since the `SiloName` is how you select the node configuration in the first place.

In the future, we can remove this notion of 'defaults' and 'overrides'. At that point, I think it does make sense to have the silo name in the config.

ReubenBond:
@dotnet-bot test this please

src/OrleansServiceFabricUtils/OrleansCommunicationListener.cs
xiazen: I see , yeah I meant ClusterConfiguration.Overrides.[...].SiloName. 
I was just confused since there seems to be two source of truth for siloName.

but the rest LGTM.

ReubenBond: I was confused, too. We will address this in the configuration replacement

PR #3245, opened 2017-07-24
Cherry-picked fixes for 1.5.1

sergeybykov:


ReubenBond:
Please also include:
#3249
#3250

sergeybykov:
Added #3249, #3250, and #3258.

ReubenBond:
In our meeting yesterday, @xiazen and @jason-bragg mentioned a bug in queue balancing logic which they wanted to fix in 1.5.1. Is #3267 that PR? i.e, can we ship once we merge that PR and include it here?

sergeybykov:
@xiazen thinks now that that bug is "compensated" in 1.5.0 (and before) with another logic. So there's no need to patch it in 1.5.1. She is double-checking with @gabikliot.

Issue #3223, opened 2017-07-13
Orleans Facet System


jason-bragg:
As Orleans has grown and new features have been introduced, it has become clear that formalizing patterns for adding new features to Orleans will improved the maintainability, stability, and testability of the framework.  Each new feature we’ve added has been introduced in slightly different ways with different advantages and disadvantages.  For instance, grain storage was introduced via a base class, limiting its composability, while streams were introduced using grain extensions making it nearly impossible for third party developers to build their own stream providers.  The variety of patterns, which mostly do similar things, have, over time, added significant complexity to maintaining, testing and improving Orleans.  To address these problems, we’re introducing an extensibility pattern we hope to use for most if not all new features added in the future.  Since all the good names for such generic extensibility patterns have been taken, we’re calling this Orleans Facet System.  This system allows ‘facets’ of the Orleans framework to be exposed to the application layer via grains.

Facets are Orleans features that can be injected into the grain via grain construction and take part in the grain’s lifecycle.  They are very similar to dependency injected services scoped to a grain, with the additional ability to asynchronously take part in stages of a grain’s lifecycle.

To illustrate how this is expected to look, let’s explore what grain state might look like as a facet rather than via inheritance, as it currently is.

Currently, to use grain state storage, one would inherit from a base class that might look something like:

    [StorageProvider("Blob")]
    public class MyGrain : Grain<MyState>, IMyGrain
    {
    }

Where the base class exposes the following properties/calls.

    MyState State { get; set; }
    Task ReadStateAsync();
    Task WriteStateAsync();
    Task ClearStateAsync();

Exposing similar state storage capabilities via a facet may look something like:

    public class MyGrain : Grain, IMyGrain
    {
        private readonly IPersistentState<MyState> state;
        public MyGrain(
            [PersistentStateStorage("Blob", "MyState")]
            IPersistentState<MyState> state,
        {
            this.state = state;
        }
    }

Where the IPersistentState interface exposes the following properties/calls.

    public interface IPersistentState<TState>
    {
        TState State { get; }
        Task ReadStateAsync();
        Task WriteStateAsync();
        Task ClearStateAsync();
    }

The infrastructure provided by Grain\<T\> currently wires the grain state to the storage associated with the named storage provider and loads the grain state, prior to grain activation.  The facet system would provide similar support for the IPersistentState facet, associating it to the storage identified by its attribution, and loading the state as part of the grain lifecycle.

[...]

PR #3247, opened 2017-07-25
Orleans facet property injection

jason-bragg:
Initial implementation of Orleans Facet System #3223  using property injection.


src/OrleansRuntime/Facet/GrainFacetPropertyInjector.cs
```diff
@@ -0,0 +1,70 @@
...
+            private readonly List<Action<IGrainActivationContext>> propertyInjectors;
+
+            public Injector(Type type)
+            {
+                this.propertyInjectors = CreatePropertyInjectors(type).ToList();
+            }
+
+            public void InjectProperties(IGrainActivationContext grainContext)
+            {
+                propertyInjectors.ForEach(injector => injector.Invoke(grainContext));
```
ReubenBond: Using `foreach` here would avoid these two per-activation allocations: one for the closure to capture `grainContext` and one for the conversion from method to `Func<Action<IGrainActivationContext>>`.

test/Tester/Facet/StorageFeatureAttribute.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using Orleans;
+using Orleans.Runtime;
+
+namespace Tester
+{
+    [AttributeUsage(AttributeTargets.Property)]
```
ReubenBond: If these attributes weren't restricted to properties, then they could one day be used on ctor params,

jason-bragg: Agreed, but we'd need some additional wiring.  Best to keep them constrained to the only supported pattern for now.  Do you disagree?

jason-bragg:
Returning to constructor injection.  Closing.

PR #3249, opened 2017-07-26
Dispose IServiceProvider during client shutdown.

ReubenBond:
`IServiceProvider` implementations do not necessarily implement `IDisposable`, but most do.

### DIFF
--- a/src/Orleans/Runtime/OutsideRuntimeClient.cs
+++ b/src/Orleans/Runtime/OutsideRuntimeClient.cs
@@ -886,6 +886,7 @@ namespace Orleans
             }
 
             (this.ServiceProvider as IDisposable)?.Dispose();
+            this.ServiceProvider = null;
             GC.SuppressFinalize(this);
         }
 
--- a/src/Orleans/Runtime/OutsideRuntimeClient.cs
+++ b/src/Orleans/Runtime/OutsideRuntimeClient.cs
@@ -884,6 +884,8 @@ namespace Orleans
                 ClientStatistics.Dispose();
                 ClientStatistics = null;
             }
+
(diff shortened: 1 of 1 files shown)


ReubenBond:

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -884,6 +884,8 @@ public void Dispose()
                 ClientStatistics.Dispose();
                 ClientStatistics = null;
             }
+
+            (this.ServiceProvider as IDisposable)?.Dispose();
```
xiazen: I think we should set `this.ServiceProvider = null `if it is disposable?

jdom:
Did you validate that our own services do not throw at least? When I added the Dispose on the silo service provider, I checked the logs (and debugged with first chance exceptions on) and found a few bugs in the dispose logic in some of our services.

jdom:
Did you see that the [ConstructorReset](https://github.com/ReubenBond/orleans/blob/bf2e20bf78d950dc8ed882b0fd0b6c62343aec9e/src/Orleans/Runtime/OutsideRuntimeClient.cs#L786) method also disposes the service provider. Do we need both or you missed that one?

ReubenBond:
If a user disposes the client, the container should also be disposed. Maybe ConstructorReset should be calling `this.Dispose()`

ReubenBond:
btw, the reason I came upon this was that I found that services weren't being disposed while performing reliability tests. The client would fail to connect a few times before the newly deployed cluster was ready to accept requests, but services which were created for that connection attempt were not being disposed and eventually there were multiple instances of the same 'singleton' service running.

PR #3250, opened 2017-07-26
Minor improvements to Service Fabric providers

ReubenBond:
In testing, I found that the method I had eagerly employed to ensure that we were not processing outdated Service Fabric partition change notifications was the cause of many lost updates during disaster recovery scenarios. Not blaming SF, it's likely because of how I was comparing partition equality. i.e, two `ResolvedServicePartition` instances which belong to a `Singleton` partition must belong to the same partition because logically there can only be one singleton partition. This PR removes that check, since it's unnecessary (any stale information will quickly be superseded by fresh information).

Fixed a `NullReferenceException` which was being thrown when a client attempts to resolve silos before the silo service has been successfully created.

Increased logging.

Reduced the eager refresh interval from 30s to 5s. If logging is verbose, then this will cause 5x more logs from that process. The `MaxStaleness` was also reduced to the refresh interval. This has the effect of blacklisted gateways being cleared much more quickly. I see no downside to this - the shortened polling interval is still not aggressive.

### DIFF
--- a/src/OrleansServiceFabricUtils/FabricGatewayProvider.cs
+++ b/src/OrleansServiceFabricUtils/FabricGatewayProvider.cs
@@ -67,14 +67,16 @@ namespace Microsoft.Orleans.ServiceFabric
         /// <inheritdoc />
         public bool SubscribeToGatewayNotificationEvents(IGatewayListListener subscriber)
         {
+            this.log.Verbose($"Unsubscribing {subscriber} to gateway notification events.");
             this.subscribers.TryAdd(subscriber, subscriber);
             return true;
         }
 
         /// <inheritdoc />
-        public bool UnSubscribeFromGatewayNotificationEvents(IGatewayListListener listener)
+        public bool UnSubscribeFromGatewayNotificationEvents(IGatewayListListener subscriber)
         {
-            this.subscribers.TryRemove(listener, out listener);
+            this.log.Verbose($"Unsubscribing {subscriber} from gateway notification events.");
+            this.subscribers.TryRemove(subscriber, out subscriber);
             return true;
         }
 
--- a/src/OrleansServiceFabricUtils/FabricGatewayProvider.cs
+++ b/src/OrleansServiceFabricUtils/FabricGatewayProvider.cs
@@ -41,8 +41,8 @@ namespace Microsoft.Orleans.ServiceFabric
         public FabricGatewayProvider(IFabricServiceSiloResolver siloResolver)
         {
             this.fabricServiceSiloResolver = siloResolver;
-            this.refreshPeriod = TimeSpan.FromSeconds(30);
-            this.MaxStaleness = TimeSpan.FromSeconds(this.refreshPeriod.TotalSeconds * 2);
+            this.refreshPeriod = TimeSpan.FromSeconds(5);
+            this.MaxStaleness = this.refreshPeriod;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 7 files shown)


src/OrleansServiceFabricUtils/FabricGatewayProvider.cs
```diff
@@ -67,14 +67,16 @@ public FabricGatewayProvider(IFabricServiceSiloResolver siloResolver)
         /// <inheritdoc />
         public bool SubscribeToGatewayNotificationEvents(IGatewayListListener subscriber)
         {
+            this.log.Verbose($"Unsubscribing {subscriber} to gateway notification events.");
```
xiazen: Subsribe * 
typo

src/OrleansServiceFabricUtils/FabricServiceSiloResolver.cs
```diff
@@ -132,14 +133,6 @@ private void OnPartitionChange(long handlerId, FabricPartitionResolutionChange a
                     if (!existing.IsSamePartitionAs(updated.Partition)) continue;
                     found = true;
 
-                    if (updated.Partition.IsOlderThan(existing))
-                    {
-                        this.log.Info($"Update for partition {updated} is superseded by existing version.");
-
-                        // Do not update the partition if the exiting one has a newer version than the update.
-                        break;
```
xiazen: I'm not an expert on service fabric. But would removing this post the risk of older partition overwrites newer partition? Yes the older partition would mostly be corrected by newer partition eventually. But this brought unnecessary handling on partitionChange, which can be avoided, right?  or is this pre-mature optimization?

src/OrleansServiceFabricUtils/Utilities/ServiceFabricExtensions.cs
```diff
@@ -7,19 +7,6 @@ namespace Microsoft.Orleans.ServiceFabric.Utilities
...
-        /// Returns a value indicating whether or not <paramref name="left"/> is older than <paramref name="right"/>.
-        /// </summary>
-        /// <param name="left">One resolved partition.</param>
-        /// <param name="right">The other resolved partition.</param>
-        /// <returns>
-        /// <see langword="true"/> if <paramref name="left"/> is older than <paramref name="right"/>, <see langword="false"/> otherwise.
-        /// </returns>
-        public static bool IsOlderThan(this ResolvedServicePartition left, ResolvedServicePartition right)
-        {
-            return left.Info.Id == right.Info.Id && left.CompareVersion(right) < 0;
```
xiazen: can this be just` left.CompareVersion(right) < 0; `?  so that we removed the unnecessary check and also compared the version, so that FabricServiceSiloResolver.OnPartitionChange won't be processing older partition.

ReubenBond: we can, but the check doesn't provide value

src/OrleansServiceFabricUtils/FabricGatewayProvider.cs
ReubenBond: ahhh thanks

ReubenBond: I added these logs because of the weird behavior which eventually led to #3249

src/OrleansServiceFabricUtils/FabricServiceSiloResolver.cs
ReubenBond: Not really, I was just finding that these checks really don't add value. In the worst case - that there's some race and updates are out of order, we have stale information for 1 polling cycle (5 seconds).

src/OrleansServiceFabricUtils/Utilities/ServiceFabricExtensions.cs
ReubenBond: Actually, that wouldn't work - because the version would be reset when the partition ids change (and anyhow the two versions wouldn't be related to each other since they're for different physical partitions)

xiazen:

PR #3254, opened 2017-07-26
Rewrite the Developing a Client doc

sergeybykov:


src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
```diff
@@ -3,71 +3,87 @@ layout: page
 title: Developing a Client
 ---
 
+### What Is Grain Client?
 
-Once we have our grain type implemented, we can write a client application that uses the type.
+The term "Client" or sometimes "Grain Client" is used for application code that interacts with grains but itself is not part of a grain logic.
+Client code runs outside of the cluster of [Orleans servers called "silos"](Silos.md) where grains are hosted.
+Hence, a client acts as a connector or conduit to the cluster and to all grains of the application.
```
ReubenBond: For future docs we should try to use simple, casual sentences to make the docs more accessible to readers who are less fluent in English. I need to keep reminding myself

sergeybykov: Can you give an example? I'm not sure I quite see it in this context.

ReubenBond: You're right, I think I was just tired.

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
```diff
@@ -3,71 +3,87 @@ layout: page
...
+Lastly, we need to call `Connect()` method on the constructed client object to make it connect to the Orleans cluster, and wait for connection to succeed.
+If the code runs withing an asynchronous method, then it is better to `await` the `Connect()` call, so that we do not block the executing thread:
+```csharp
+await client.Connect(); 
+```
 
-See the Key Concepts section for more details on the various ways to use `Task`s for execution scheduling and exception flow.
+Otherwise, if the running method cannot be made asynchronous, we have to block the thread with a `Wait()`:
+```csharp
+client.Connect().Wait();
```
ReubenBond: Not sure if we need to give the async Tasks 101 here - could it be assumed knowledge / a link? I suppose it's more newbie friendly this way

sergeybykov: Good point. I rewrote to be more concise and less about TPL. What do you think?

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
```diff
@@ -3,71 +3,81 @@ layout: page
...
+Hence, a client acts as a connector or conduit to the cluster and to all grains of the application.
 
-The following Orleans DLLs from either the `[SDK-ROOT]\Binaries\PresenceClient_ or _[SDK-ROOT]\Samples\References` directories need to be referenced in the client application project:
+![](Frontend-Cluster.png)
 
-* Orleans.dll
-* OrleansRuntimeInterfaces.dll
+Usually, clients are used on the frontend web servers to connect to an Orleans cluster that serves as a middle tier with grains executing business logic.
+In a typical setup, a frontend web server:
+* Receives a request over HTTP (or some other protocol)
```
xiazen: "Receives a web request" maybe more precise. Since mentioning protocol is not related in the context IMO

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
```diff
@@ -3,71 +3,81 @@ layout: page
...
+* Performs necessary authentication and authorization validation
+* Decides which grain(s) should process the request
```
xiazen: Maybe "Route the request to the controller which has the logic handling it" is more accurate 

"Decides" is not clear to me about how the server decide.

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
```diff
@@ -3,71 +3,81 @@ layout: page
...
+* Uses Grain Client to make one or more method call to the grain(s)
```
xiazen: The controller logic decides which grains take part in processing the request and communicate with them through GrainClient

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
```diff
@@ -3,71 +3,81 @@ layout: page
...
+* Handles successful completion or failures of the grain calls and any returned values
+* Sends a response for the HTTP request
```
xiazen: Sends a response for the web request

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
sergeybykov: Good suggestion. I incorporated it.

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
sergeybykov: I think it'll be to confusing to use "route" here because Orleans routes requests to the grains/silos under the covers. The job of the client is to decide which grain(s) to invoke. And that's application specific logic. How the app decides would be different for different apps. Maybe "Decides which grain(s) should process the request based on the content or metadata of the request" instead?

"Controller" has a different meaning in this context. I think it is better to stay away from creating an unnecessary confusion here and be specific, not overly general. Client makes a call to a grain, not to a controller, even if conceptually the grain plays a controller role in the app design.

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
sergeybykov: This would assume that there's a controller. I think we don't need to be that specific.

src/Documentation/Getting-Started-With-Orleans/Developing-a-Client.md
sergeybykov: Done.

xiazen:
LGTM

PR #3256, opened 2017-07-27
Execute TypeManager timers in the correct runtime context

ReubenBond:
In `Silo.DoStart()` we should be scheduling `TypeManager.Initialize()` on the TypeManager's context (it's a system target), since it starts a timer which potentially makes grain calls.

The timer started there also needs to invoke its callback on the TypeManager's context.

Otherwise errors (see stack):
image

### DIFF
--- a/src/OrleansRuntime/GrainTypeManager/TypeManager.cs
+++ b/src/OrleansRuntime/GrainTypeManager/TypeManager.cs
@@ -11,7 +11,7 @@ using Orleans.Versions.Selector;
 
 namespace Orleans.Runtime
 {
-    internal class TypeManager : SystemTarget, IClusterTypeManager, ISiloTypeManager, ISiloStatusListener
+    internal class TypeManager : SystemTarget, IClusterTypeManager, ISiloTypeManager, ISiloStatusListener, IDisposable
     {
         private readonly Logger logger = LogManager.GetLogger("TypeManager");
         private readonly GrainTypeManager grainTypeManager;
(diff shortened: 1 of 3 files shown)


benjaminpetit:
Makes sense, but I am not sure why so many tests failures...

ReubenBond:
Not sure why the tests are failing - I'll look into it.

ReubenBond:
In `ChangeDefaultVersionCompatibilityStrategy`, if I insert a delay between starting the v2 silo and getting a grain reference, then the test passes:
```C#
await StartSiloV2();

await Task.Delay(TimeSpan.FromMinutes(2));

var grainV2 = Client.GetGrain<IVersionUpgradeTestGrain>(2);
```

Any idea why putting a delay in that particular location might help?

benjaminpetit:
> Any idea why putting a delay in that particular location might help?

When we start a new silo in an existing cluster, we wait `TestCluster.GetLivenessStabilizationTime`. It doesn't seems to be enough here. On my machine I am not able to reproduce this race condition?

benjaminpetit:
@dotnet-bot test this please

benjaminpetit:
Could you try to add some delay after we started a new silo? It really seems that the value returned by `TestCluster.GetLivenessStabilizationTime` is not enough.

But I don't understand why it was more reliable before.

sergeybykov:
@dotnet-bot test netfx-bvt 

ReubenBond:
@benjaminpetit I've added a 90s wait there now, maybe those tests will be happier.

sergeybykov:
90s is an eternity. That shouldn't be required for running these tests correctly. Something must be wrong there.

sergeybykov:
I submitted https://github.com/ReubenBond/orleans/pull/7 for this PR, which I hope will fix the test failures.

sergeybykov:

sergeybykov:
Closing this one, as #3309 is a replacement for it.

PR #3258, opened 2017-07-27
Check that client is initialized in GrainClient.ClientInvokeCallback

### LINKED ISSUE #3251: Setting GrainClient.ClientInvokeCallback raises NullReferenceException [Orleans 1.5.0]
Today I updated my project from Orleans 1.4.2 to 1.5.0. Now I experience the following issue:

When setting `GrainClient.ClientInvokeCallback`, I get a `NullReferenceException`:

```csharp
GrainClient.ClientInvokeCallback = (request, grain) => { RequestContext.Set("IsExceptionConversionEnabled", true); };
```

This did not happen in Orleans 1.4.2.

After digging deeper I found a way to fix this: in my code I set the callback **before** calling `GrainClient.Initialize(...)`. I swapped the lines of code, so the callback is set **after** the grain client has been initialized. Now everything works properly.

I think if it is not allowed to set the callback before initialization, an `InvalidOperationException` with a hint in the exception message should be thrown instead of a `NullReferenceException`.

ReubenBond:
Better to throw an `InvalidOperationException` than a `NullReferenceException`.
Fixes #3251.

### DIFF
--- a/src/Orleans/Core/GrainClient.cs
+++ b/src/Orleans/Core/GrainClient.cs
@@ -3,12 +3,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Net;
-using System.Runtime.ExceptionServices;
-using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Orleans.CodeGeneration;
-using Orleans.Providers;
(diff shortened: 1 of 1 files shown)


ReubenBond:
@dotnet-bot test netfx-bvt

xiazen:
LGTM

PR #3262, opened 2017-07-30
Fixing MySQL assembly name for dotnet core and fixing SqliteFactory path

berdon:
Neither the MySQL or SQLite provider could be loaded for different reasons. The MySQL assembly for .netcore 1.1 is now Mysql.Data.Core. For SQLite, SqliteFactory's fully qualified path was wrong.

This is a quick and simple fix - a better fix, perhaps after, would maybe be to allow invariantName to include a fully qualified assembly?

### DIFF
--- a/src/OrleansSQLUtils/Storage/DbConnectionFactory.cs
+++ b/src/OrleansSQLUtils/Storage/DbConnectionFactory.cs
@@ -17,10 +17,10 @@ namespace Orleans.SqlUtils
             new Dictionary<string, Tuple<string, string>>
             {
                 { AdoNetInvariants.InvariantNameSqlServer, new Tuple<string, string>("System.Data.SqlClient", "System.Data.SqlClient.SqlClientFactory") },
-                { AdoNetInvariants.InvariantNameMySql, new Tuple<string, string>("MySql.Data", "MySql.Data.MySqlClient.MySqlClientFactory") },
+                { AdoNetInvariants.InvariantNameMySql, new Tuple<string, string>("MySql.Data.Core", "MySql.Data.MySqlClient.MySqlClientFactory") },
                 { AdoNetInvariants.InvariantNameOracleDatabase, new Tuple<string, string>("System.Data.SqlClient", "System.Data.SqlClient.SqlClientFactory") },
                 { AdoNetInvariants.InvariantNamePostgreSql, new Tuple<string, string>("Npgsql", "Npgsql.NpgsqlFactory") },
-                { AdoNetInvariants.InvariantNameSqlLite, new Tuple<string, string>("Microsoft.Data.SQLite", "Microsoft.Data.SQLite.SqliteFactory") },
+                { AdoNetInvariants.InvariantNameSqlLite, new Tuple<string, string>("Microsoft.Data.SQLite", "Microsoft.Data.Sqlite.SqliteFactory") },
             };
 
         private static CachedFactory GetFactory(string invariantName)


veikkoeeva:
@berdon That is a good suggestion and I remember it has come up before too. There are some backwards compatibility issues with that, though. I'm inclined to think your suggestion should be considered for the Orleans 2.0 version together with removing "SqlServer" from places where "AdoNet" is considered (this is due to historical reasons).

sergeybykov:

sergeybykov:
>There are some backwards compatibility issues with that, though.

@veikkoeeva can you elaborate?

veikkoeeva:
@sergeybykov Everywhere where there reads `SqlServer` when any ADO.NET compatible database is meant should be renamed to `AdoNet`, I think. Most prominently this is seen in `SystemStore=SqlServer` setting and elsewhere in configuration. Then there's renaming the class internally. A year ago this was a frequent cause of confusion if other DBs are supported.

Taking a quick search, looking at https://github.com/dotnet/orleans/search?p=1&q=SqlServer&type=& utf8=%E2%9C%93 gives some other things, such as `GlobalConfiguration.ReminderServiceProviderType.SqlServer` for which ADO.NET would make more sense. There are other too, though.

sergeybykov:
@veikkoeeva Thanks. This is about the future renaming. But this PR won't break backward compatibility, will it?

veikkoeeva:
@sergeybykov Now that you mention, indeed, this change would break most of MySQL clients when they uprgrade when looking at the Nuget statistics: https://www.nuget.org/stats/packages/MySql.Data?groupby=Version. Further, it appears https://www.nuget.org/stats/packages/MySql.Data.Core?groupby=Version is an unlisted pre-release version. @berdon Could you double-check if this is the case?

As for the SqlLite, it looks like the issue would be the factory class name. I didn't find the class source code and I don't want to delay this due to me, so if @berdon has tested that's the problem while loading the class, I'm good with that. But, @sergeybykov, we should note there's https://www.nuget.org/packages/Microsoft.Data.Sqlite.Core/ out there that seems to be legitimate. When the Orleans Core is out, it makes sense to check these and documentation, along with the other refactoring suggestions listed here.

berdon:
I haven't tested with SQLite outside of observing the discrepancy with the fully-qualified class name. I'll do some testing to make sure SQLite works with the change.

As far as backwards compatibility goes - I could the approach of simply using a tuple to something that knows of multiple package names and checks for the second in the event the first fails? This would support both packages at a minimal cost in performance. Someone more knowledgeable would need to comment on how often this factory gets called.

galvesribeiro:
That package name change happened with the `netstandard` new package. Is should work for .Net full framework as well.

If you don't want to break the user now, I would suggest use a `#if` to select the `netstandard` package on Orleans 2.0 build, and `#else` for current full CLR build.

However, users must be notified that as we move close to 2.0, we should stick only with the `netstandard` package.

sergeybykov:
>If you don't want to break the user now, I would suggest use a #if to select the netstandard package on Orleans 2.0 build, and #else for current full CLR build.

I think this is a reasonable compromise. It will unblock `netstandard` users (via Tech Preview) right away, and will become a no-op in 2.0. But if we suddenly need to release a 1.6, which we have no plans for at the moment, we won't break anybody.

sergeybykov:
@berdon Is this still relevant with .NET Core 2.0?

berdon:
MySql.Data.Core does seem incorrect. The SQLite change is probably still necessary though. I'm not at a point where I can test this though - so I wonder if this is a close until someone needs SQLite?

sergeybykov:
@berdon Thanks. I'll close it for now then. Can always reopen/redo later.

PR #3263, opened 2017-07-31
Migrate Presence sample to 1.5.0

sergeybykov:
Updated Orleans to 1.5.0.
Updated silo and client initialization code to make it simpler and aligned with 1.5.0.


Samples/Presence/PlayerWatcher/App.config
```diff
@@ -1,6 +1,70 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
     <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2"/>
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2" />
     </startup>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
```
xiazen: what's the reason for all these extra binding redirects? is there some other package needed to be upgraded too, so that we can avoid all these binding redirects?

Samples/Presence/LoadGenerator/App.config
```diff
@@ -1,6 +1,70 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
     <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2"/>
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2" />
     </startup>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
```
xiazen: what's the reason for all these extra binding redirects? is there some other package needed to be upgraded too, so that we can avoid some binding redirects?

sergeybykov: VS added them automatically, and I didn't want to fight the tooling.

Samples/Presence/PlayerWatcher/App.config
sergeybykov: VS added them automatically, and I didn't want to fight the tooling.

Samples/Presence/Host/Program.cs
```diff
@@ -10,46 +13,15 @@ public class Program
...
-            //       or initializing an HTTP front end for accepting incoming requests.
+            var siloConfig = ClusterConfiguration.LocalhostPrimarySilo();
+            siloConfig.Defaults.DefaultTraceLevel = Severity.Warning;
+            var silo = new SiloHost("Test Silo", siloConfig);
+            silo.InitializeOrleansSilo();
+            silo.StartOrleansSilo();
 
+            
             Console.WriteLine("Orleans Silo is running.\nPress Enter to terminate...");
             Console.ReadLine();
```
xiazen: should we do 
```
silo?.StopOrleansSilo();
silo?.UnInitializeOrleansSilo();
```
here to proper shutdown the silo at the end of the program?

Samples/Presence/PlayerWatcher/Program.cs
```diff
@@ -47,28 +58,27 @@ static void Main(string[] args)
...
-                // Block main thread so that the process doesn't exit. Updates arrive on thread pool threads.
                 Console.WriteLine("Subscribed successfully. Press <Enter> to stop.");
-                Console.ReadLine();
             }
             catch (Exception exc)
             {
                 Console.WriteLine("Unexpected Error: {0}", exc.GetBaseException());
             }
         }
+    }
```
xiazen: I think we need to proper close and dispose the client before program exit. maybe after `Console.Readline()`

xiazen: it will be an good example to user which reminds them to use `client.Close() ` and `client.Dispose()` to clear the resources client uses before their client app exit.

Samples/Presence/LoadGenerator/Program.cs
```diff
@@ -15,10 +15,20 @@ class Program
         /// </summary>
         static void Main(string[] args)
         {
+            RunLoadGenerator().Wait();
+            // Block main thread so that the process doesn't exit.
+            // Updates arrive on thread pool threads.
+            Console.ReadLine();
+        }
```
xiazen: here too, proper close and dispose the client before exit.

Samples/Presence/PlayerWatcher/Program.cs
sergeybykov: Good point. I added cleanup logic, but I had to return `client` from the call for that. Does this look right to you?

PR #3265, opened 2017-08-01
Update code sample in Client-Configuration.md to use ClientBuilder

ReubenBond:


src/Documentation/Orleans-Configuration-Guide/Client-Configuration.md
```diff
@@ -134,17 +134,42 @@ config.RegisterStreamProvider<SimpleMessageStreamProvider>("SimpleMessagingStrea
...
+        // maintain the connection, automatically reconnecting as necessary.
+        await client.Connect().ConfigureAwait(false);
+        break;
     }
-    catch (Exception exc)
+    catch (Exception exception)
     {
-        //Log "Exception when initializing Orleans Client"
+        // If the connection attempt fails, the client instance must be disposed.
+        client?.Dispose();
```
xiazen: client?.Dispose();
client = null;

src/Documentation/Orleans-Configuration-Guide/Client-Configuration.md
```diff
@@ -134,17 +134,42 @@ config.RegisterStreamProvider<SimpleMessageStreamProvider>("SimpleMessagingStrea
...
+        // Build a client and then connect it to the cluster.
+        client = new ClientBuilder()
+            .UseConfiguration(config)
+            .ConfigureServices(
+                services =>
+                {
+                    // Services can be provided to the client here. These services are made
+                    // available via dependency injection.
+                    // ConfigureServices can be called multiple times for a single
+                    // ClientBuilder instance.
```
xiazen: maybe add one example here would be better for user to understand, something like:
` services.AddSingleton(yourService);`

Because without looking into public API, it's not clear to me what type of delegate I should feed to `ConfigureServices`

src/Documentation/Orleans-Configuration-Guide/Client-Configuration.md
ReubenBond: fixed

src/Documentation/Orleans-Configuration-Guide/Client-Configuration.md
xiazen: I was just thinking adding an example about how user can inject their own service into DI using ClientBuilder. Literally just 
`services.AddSingleton<YourCustomService>()`, 
But maybe I was overthinking. And current example is clear enough. I will leave it to you to decide what you want to do.  I will merge it at the end of the day or tomorrow regardless.

ReubenBond: there's nowhere where the user would be able to use `YourCustomerService`, though, because there's nothing to inject it into - they can only meaningfully inject framework dependencies (eg, gateway providers)

xiazen: OIC. thanks for the explanation

PR #3267, opened 2017-08-01
Remove immature silo logic in DeploymentBasedQueueBalancer

xiazen:
Remove immature silo logic in DeploymentBasedQueueBalancer, which seems to brought extra complex but doesn't bring extra value. Can you sanity check this @gabikliot ?

### DIFF
--- a/src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
+++ b/src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
@@ -68,7 +68,6 @@ namespace Orleans.Streams
         private readonly IDeploymentConfiguration deploymentConfig;
         private ReadOnlyCollection<QueueId> allQueues;
         private readonly List<IStreamQueueBalanceListener> queueBalanceListeners;
-        private readonly ConcurrentDictionary<SiloAddress, bool> immatureSilos;
         private readonly bool isFixed;
         private bool isStarting;
 
@@ -89,19 +88,11 @@ namespace Orleans.Streams
             this.siloStatusOracle = siloStatusOracle;
             this.deploymentConfig = deploymentConfig;
             queueBalanceListeners = new List<IStreamQueueBalanceListener>();
-            immatureSilos = new ConcurrentDictionary<SiloAddress, bool>();
             this.isFixed = isFixed;
             isStarting = true;
 
             // register for notification of changes to silo status for any silo in the cluster
             this.siloStatusOracle.SubscribeToSiloStatusEvents(this);
-
-            // record all already active silos as already mature. 
-            // Even if they are not yet, they will be mature by the time I mature myself (after I become !isStarting).
-            foreach (var silo in siloStatusOracle.GetApproximateSiloStatuses(true).Keys.Where(s => !s.Equals(siloStatusOracle.SiloAddress)))
-            {
-                immatureSilos[silo] = false;     // record as mature
-            }
         }
 
         public Task Initialize(string strProviderName,
@@ -133,44 +124,11 @@ namespace Orleans.Streams
         /// <param name="status">new silo status</param>
         public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
         {
-            if (status == SiloStatus.Dead)
-            {
-                // just clean up garbage from immatureSilos.
-                bool ignore;
-                immatureSilos.TryRemove(updatedSilo, out ignore);
-            }
-            SiloStatusChangeNotification().Ignore();
-        }
-
-        private async Task SiloStatusChangeNotification()
-        {
-            List<Task> tasks = new List<Task>();
-            // look at all currently active silos not including myself
-            foreach (var silo in siloStatusOracle.GetApproximateSiloStatuses(true).Keys.Where(s => !s.Equals(siloStatusOracle.SiloAddress)))
-            {
-                bool ignore;
-                if (!immatureSilos.TryGetValue(silo, out ignore))
-                {
-                    tasks.Add(RecordImmatureSilo(silo));
-                }
-            }
             if (!isStarting)
             {
                 // notify, uncoditionaly, and deal with changes in GetMyQueues()
                 NotifyListeners().Ignore();
             }
-            if (tasks.Count > 0)
-            {
-                await Task.WhenAll(tasks);
-                await NotifyListeners(); // notify, uncoditionaly, and deal with changes it in GetMyQueues()
-            }
-        }
-
-        private async Task RecordImmatureSilo(SiloAddress updatedSilo)
-        {
-            immatureSilos[updatedSilo] = true;      // record as immature
-            await Task.Delay(siloMaturityPeriod);
-            immatureSilos[updatedSilo] = false;     // record as mature
         }
 
         public IEnumerable<QueueId> GetMyQueues()
@@ -179,17 +137,11 @@ namespace Orleans.Streams
             bool useIdealDistribution = isFixed || isStarting;
             Dictionary<string, List<QueueId>> distribution = useIdealDistribution
                 ? balancer.IdealDistribution
-                : balancer.GetDistribution(GetActiveSilos(siloStatusOracle, immatureSilos));
+                : balancer.GetDistribution(GetActiveSilos(siloStatusOracle));
 
             List<QueueId> myQueues;
             if (distribution.TryGetValue(siloStatusOracle.SiloName, out myQueues))
             {
-                if (!useIdealDistribution)
-                {
-                    HashSet<QueueId> queuesOfImmatureSilos = GetQueuesOfImmatureSilos(siloStatusOracle, immatureSilos, balancer.IdealDistribution);
-                    // filter queues that belong to immature silos
-                    myQueues.RemoveAll(queue => queuesOfImmatureSilos.Contains(queue));
-                }
                 return myQueues;
             }
             return Enumerable.Empty<QueueId>();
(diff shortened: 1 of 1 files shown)


xiazen:

gabikliot:
Is this still relevant? Do you still need clarification about that or not? The PR is closed.
I can try to recall why we had this "immature silos" if this is still relevant.

xiazen:
Yes it is still relevant. The PR is closed because it can somehow bring confusion to the community since we are not sure about this change....
But the PR is still relevant. I'd like to know why those immature silo logic is there.

gabikliot:
Roughly speaking, the idea was as following:

Image when the whole cluster starts first time, or when the cluster restarts, or even when a whole app domain is updated at once. You get a lot of silos joining together in a very short period of time. But they still join one at a time, due to how membership works. But even if mbr worked differently, there could still be a lot of cases when they join one at a time with very short period between each other.

Why is that a problem? It is a problem in the case of streaming from a Queue (or EH) and building the cache. Silo A is told he owns a range of queues, he starts pulling, reads the cache, and then a second later it is told he does not own that queue and needs to drop the cache. This is inefficient.
But even worse - once the silo connects to the queue it starts immediately sending  the msgs he got from the queue to subscribers, if there are any, of course. Lets assume there are. Depending on how the checkpointing works, they may not checkpoint in time (EH case), and then the other agent is getting the queue and starts delivery again. They will get duplicate msgs. 
There was also some very specific case in Halo - they had a case when some subscribers (I think they were implicit but I don't think it matters) were local placement. and it was very important for them to be local placement, on the same silo with the agent (we even fixed the local placement with count 1 to be precise and not to overshoot for this case). Now when the agent moves (the queue is moved to a new agent), the new local activation is activated. It now gets more eTag violations with the other local activation (they resolved conflicts via storage).

So - bottom line was that moving queues quickly from one silo to another, just during the cluster startup, created a lot of redundant cache filling and discarding, a lot of contention on the storage, etc..
And what makes it worse than usually a silo will take  a queue, start pulling, and very quickly drop it since it got re-assigned.

We wanted to solve it generally. The solution we came up with was a "grace period", or "maturity period". If silo joins and now owns a queue, don't give him this queue immediately. Let the silo "mature", wait a bit (like a minute or two), hopefully by this time all silos join and only now assign the queues. Hopefully by that time all silos joined and the assignment will be stable.

This needed to be done with extra care, to work correctly with a regular case and other balancers.

In your PR, if you remove the immature logic, everything will still work correctly. But you will loose this grace period logic and when a lot of silos join, you will get all those queues moving quickly between them.

Also, look here: 
http://dotnet.github.io/orleans/Presentations/Balancing%20Techniques%20in%20Orleans.pptx
slide 17, last 3 bullets.

@jason-bragg did not remember that? We came up together with all this. A lot of good debates about how exactly to do it.
Let me know if you have more questions. I hope now you will have enough context to trace down the actual logic in the implementation.

xiazen:
@jason-bragg remembers that. But we are still fuzzy on the implementation details. 

> Let the silo "mature", wait a bit (like a minute or two), hopefully by this time all silos join and only now assign the queues. Hopefully by that time all silos joined and the assignment will be stable.

To me, it is enough to just let the silo itself mature for a bit in startup, before it can take queues, which seems to me doesn't need the implementation to maintain a  `immatureSilos` dictionary. `immatureSilos` tracks immaturity of other silos, and greatly influence `GetActiveSilos` logic. I'm not sure how that logic plays in the picture.

src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
```diff
@@ -179,17 +137,11 @@ public IEnumerable<QueueId> GetMyQueues()
...
 
             List<QueueId> myQueues;
             if (distribution.TryGetValue(siloStatusOracle.SiloName, out myQueues))
             {
-                if (!useIdealDistribution)
-                {
-                    HashSet<QueueId> queuesOfImmatureSilos = GetQueuesOfImmatureSilos(siloStatusOracle, immatureSilos, balancer.IdealDistribution);
-                    // filter queues that belong to immature silos
-                    myQueues.RemoveAll(queue => queuesOfImmatureSilos.Contains(queue));
-                }
```
xiazen: This part also confuses me. 
In line 182, `GetActiveSilos` excluded immature silos from `activeSilos` list returned by `siloStatusOracle`, which would make current silo take some of immature silos' queues in the `distrubution`.
But in line 191, the code filtered out queues from immature silos in `myQueues`.

To me, it is like doing a + followed by a -, which cancels each other. it seems to give the same outcome if we do:  get `activeSilos` from `siloStatusOracle` ->distribution = balancer.GetDistribution(activeSilos)-> return current silo's queues from the distribution. 

And this seems to be the only part where `immatureSilos` (track immaturity for other silos) contribute to, which go back to my original confusion: why do we need to maintain that `immatureSilos` dic?

gabikliot:
>To me, it is enough to just let the silo itself mature for a bit in startup, before it can take queues

Ohh, no. This would not be correct. You need a distributed agreement on the set of immature silos, not just a local view.

Imagine the following scenario:
silo A owns queue 1. silo B joins. According to your suggestion , silo A will drop the queue since it falls into B's range, but B will be immature so it will not take it. It means that in the regular case of all silos OK and just one new joined, you have 2 min of queue not being served. Not awful, but sub optimal.
But this gets even worse - if multiple silos join quickly, A drops queue 1 for B, but then C joins, maybe D joins, maybe someone dies, the queue will be moving around between silos, some may take it, some not. This is exactly what we wanted to prevent - queue jumping around.

So instead we are saying - once silo joins, EVERYONE, him included, ignore it. Everyone ignore it for 2 min. Just ignore it, don't assign it anything. We assume that all silos learn about a new silo join at almost the same time. This is provided by our membership. So after 2 min, at approx. the same time, all already mature silos unfreeze him (set immature to false) and assign him the queues that belong to him. And since this is coupled with the ideal distribution balancer, this new silos gets his queues according to the ideal distribution. And now no-one will take them from him, since we don't take away queues in the ideal distribution from alive silos.

Does that make more sense now?

jason-bragg:
Hello @gabikliot, and thanks for chiming in on this.
Your explanation of the “queue shuffling at cluster start” problem match’s mine.  Roughly speeking we want to avoid rebalancing the queue every time a new silo is added or lost during a cluster start/restart.

My recollection, which may well be in error, of the behavior we wanted was for each silo to take responsibility for the queues it was expected to own in the ideal case (all silos active and healthy) immediately, but wait a little while (maturity time) before taking responsibility for queues intended for any missing silos.  This prevents the first silo from taking responsibility for all of the queues at start, simply because it was the first silo to start, then reshuffling them each time a new silo becomes active.

The thing about that solution is that, since each silo cares only about how long it has been active, because it’s waiting before taking on more responsibilities, the logic is independent of whether other silos are mature or not.  In fact, any silo that is active, despite its maturity time, will have taken ownership of its queues, so by waiting until it is mature to acknowledge this would mean keeping responsibility for its queues longer than necessary for no clear reason.

The behavior I described above, is in fact, what the algorithm currently does, it just does it in a way that seems unnecessarily complicated.  A silo now tracks the maturity of all other silos in the cluster, and ignores them when building its distribution, which causes those silos queues to be balanced across the mature silos, but then removes the queues from the immature silos from the distribution (after building it) thereby preventing others from taking responsibility for them (which is correct).  This is, in my view, effectively the same thing as not tracking their maturity at all, because the information is used, then the results of it’s use is filtered out.

The code to look at to see this is in GetMyQueues.  Upon becoming mature, and able to take on responsibility for queues which were intended for other silos, we build the distribution considering only the mature silos (GetActiveSilos), which distributes the queues from the immature silos across the active ones, but then filters their queues back out using 

     myQueues.RemoveAll(queue => queuesOfImmatureSilos.Contains(queue));

which is effectively the same as not filtering out the immature silos in the first place.

This PR simplifies all of this by just not tracking immature silos, as this information is not necessary for the expected behavior, and is having no effect on the distribution

@gabikliot , what are we missing?

gabikliot:
@jason-bragg , so just to be clear on what we are talking here:
this issue is NOT about why we need immature support, but rather the concrete protocol that we use, and whether it achieves what we need.
Because the description of the issue, the way it was opened says "Remove immature silo logic in DeploymentBasedQueueBalancer"  and everything will be fine, why do we even need it?

xiazen:
Yes, the issue is about clarifying the expected protocol, how immature silo logic is supposed to play in that picture, and whether the current immature algorithm fulfils its expected behavior. 
Since to me, the current immature algorithm is not doing what it is supposed to do. Maybe I/we missed something.

jason-bragg:
@gabikliot,

> This issue is NOT about why we need immature support, but rather the concrete protocol that we use, and whether it achieves what we need.

Yes.  What is the protocol?  Does it do what we need?  Does the implementation work as intended?  Does the tracking of other silo's maturity have some purpose we're not seeing, or can it be removed?

My understanding, which may be in error, is that the system works as intended, but the tracking of maturity of other silos is not necessary, so can/should be removed.  My concern is that their is some subtlety which we are missing, hence your input in sanity checking this is invaluable.

gabikliot:
What about the `bool useIdealDistribution = isFixed || isStarting;` case?
The code behaves differently for useIdealDistribution case. Why are you ignoring this?

gabikliot:
Ye... it is tricky.

The problem is the queues of **Dead silos**. 
Imagine silo A is active and mature, silo B is active immature, silo C is Dead.

In the current code, we don't consider immature silos in the balancer (`balancer.GetDistribution(GetActiveSilos(siloStatusOracle, immatureSilos))`). Therefore, silo C's queues will be taken by A (it will only be taken by mature silos). 

In your suggestion, C's queues might not be taken by anyone.
In your suggestion, A thinks B is active and mature (he does not track B's maturity) and therefore A does not take C's queue (since via balancer B has to take C).
BUT, since B is immature yet, he does not take anything not ideal, so he does not take C's queues.
So no one takes C queue.

Imagine we had B Dead. A was serving C's queue. Now B joins and in your suggestion is causes C's queue not be served at all, until B matures.

=================

Yes. It was tricky.
And yes, I thought a lot about it, then and now.
And yes, please do document it better.

jason-bragg:
Ah.  I think I see what you're saying.

Assume B's queues would be placed on A if B was dead.
Assume C's queues would be placed on B if C was dead.
Assume C's queues would be placed on D if C and B were both dead.
Assume B and C are dead
B's queues are on A, and C's queues are on D.

**Using current code:**

1. B becomes active, and takes ownership of it's queues.
2. A and D see B is active, but ignore it because it is immature.  B's queues continue to be processed by A, and C's queues continue to be processed by D.  (causing duplicate processing of B's queues)
3. B becomes mature, and takes responsibility for C's queues.
4. A and D see B is mature.
5. Final state: C is dead, A and D own their own queues, B owns it's queues as well as C's


**Using suggested change:**

1. B becomes active, and takes ownership of it's queues.
2. A and D see B is active, A stops processing B's queues, D stops processing C's queues. (causing C's queues to be unowned)
3. B becomes mature, and takes responsibility for C's queues.
4. Final state: C is dead, A and D own their own queues, B owns it's queues as well as C's

In the above scenarios, the difference is incurring (for a configurable period of time) duplicate processing of B, or unavailability of C.  I see why we'd favor availability, especially since duplicate processing needs be handled via deduplication logic at the application layer anyway.

Am I understanding this correctly?

gabikliot:
Sorry, I am not fully following your example. But if you convinced yourself with your example, than .. good. 

Approaching this without examples, here is the formal reasoning why the current approach works:
A is active mature
B is active immature
C id Dead.

Requirements:
1) ideal queues of A to be on A
2) ideal queues of B to be on B
3) **queues of C to be distributed among A and not among Bs**

The current code achives that:
1) we take As only (remove B's from all active)
2) calculate the balancer distribution (this will put B's queues and C queues on As). 
3) remove B's ideal queues from A, since B's need their ideal ones.

If we did not do that (your PR suggestion), C's queues will go to Bs, which is what we are trying to prevent.

Finally, recall one last and VERY important piece: the whole balancer logic is independant from the order. Given the same set of active and dead silos, it always calculates the same distribution.  This is at most importance for the whole approach to work correctly - history independance.

gabikliot:
@jason-bragg , are we OK on this?
Was this helpful?

jason-bragg:
@gabikliot Yes.  Thank you.
We're leaving the code as it is in cluster based queue balancers.  We'll add more comments to capture the subtleties of this system.

PR #3271, opened 2017-08-03
Switch to gcConcurrent=true

sergeybykov:
Tests on .NET 4.6.2 show that performance with concurrent GC enabled is the same or better than without it. Time to change the recommendation.


gabikliot:
Wow! I am impressed. 7 years delay,... but hey better later on then never.

gabikliot:
Wow! I am impressed. 7 years delay,... but hey better later on then never. You probably also want to change the doc.

sergeybykov:

PR #3279, opened 2017-08-08
Orleans Facet System initial infrastructure

jason-bragg:
This PR introduces the infrastructure for "Orleans Facet System #3223", and a test feature "ExampleStorage" to vet the intended pattern.

**Infrastructure**
The facet system infrastructure consists of a utility class (ConstructorArgumentFactory) which constructs facet arguments for grains, and an extensible pattern for mapping attributes to cacheable feature factories (IAttributeToFactoryMapper).

**ExampleStorage feature**
The test feature ExampleStorage explored one of our more commonly expected feature patterns.  It is a generic service, selectable by name, which requires grain specific configuration.  In this sample implementation we break the feature in to three areas.

- Abstractions - We expose the feature interface, expected configuration, factories, and attribution in the abstractions layer.  Grains should only need to access these abstractions.
- Infrastructure - The feature infrastructure deals with the feature factories and how the attribute is wired to the factories.  In the case of the ExampleStorage, we do this by calling the feature factory which leverages the existing names services capability to select a factory by name, and build the instance with the factory.
- Implementations - Each implementation of the ExampleStorage is independently registered as a named service, making it available to the ExampleStorage infrastructure.

**Developer Audiences**
The separation of the test ExampleStorage into abstractions, infrastructure, and implementations, illustrates the following prioritization:

1) Grain developer experience- Via the clean abstraction layer, grain developers need not concern themselves with implementation details.
2) Feature extensions - Extending the feature (adding support for new types of storage) is supported by the named service registration capability.  Developers adding new implementations of a feature need only implement the feature and it's factories, and register the factory as a names service.
3) New feature developers - Most of the work is placed on the feature developer to set all of this up.  While adding new features takes the most work, having this pattern still makes it easier and more straight forward to do, and exposes the features in a way that is very grain developer friendly.

This prioritization is intentional as it is expected that most developers will use features, many may extend features, and far fewer will add new features.

Edit: Removed comments about lifecycle code in PR, as it was removed.
Edit: Changed some names to reflect name changes requested in review.


src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Cached constructor arguement factorys by type
+        /// TODO: consider storing in grain type data and constructing at startup to avoid runtime errors. - jbragg
+        /// </summary>
+        private readonly ConcurrentDictionary<Type, ArguementsFactory> arguementsFactorys;
+        private readonly IServiceProvider services;
+
+        public GrainConstructorFacetArguementsFactory(IServiceProvider services)
+        {
+            this.services = services;
+            arguementsFactorys = new ConcurrentDictionary<Type, ArguementsFactory>();
```
ReubenBond: Typo on arguments

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
```diff
@@ -0,0 +1,95 @@
...
+                IEnumerable<ParameterInfo> parameters = type.GetConstructors()
+                                                            .FirstOrDefault()?
+                                                            .GetParameters() ?? Enumerable.Empty<ParameterInfo>();
+                foreach (ParameterInfo parameter in parameters)
+                {
+                    var attribute = parameter.GetCustomAttribute<FacetAttribute>();
+                    if (attribute == null) continue;
+                    IFactoryBridge bridge = Activator.CreateInstance(typeof(FactoryBridge<>).MakeGenericType(attribute.GetType())) as IFactoryBridge;
+                    if (bridge == null) continue;
+                    bridge.Factory = services.GetRequiredService(facetFactoryType.MakeGenericType(attribute.GetType()));
```
ReubenBond: I find it awkward to be specializing a factory based with the type of an attribute, but ok.

ReubenBond: The reason I find it awkward is that I thought we'd agreed that we should support users directly injecting and calling the factories, but it's very unusual for a user to new up an instance of some `Attribute` type.

jason-bragg: > I thought we'd agreed that we should support users directly injecting and calling the factories

We do.  Please see the factory abstractions defined in the feature, IStorageFeatureFactory and INamesStorageFeatureFactory, and their usages in the sample grains.  If users supply their own factories in DI, instead of our, the system will use them.

ReubenBond: It took a long time to understand what this code is doing, maybe you can check my understanding:
* Some attribute, `TAttribute : FacetAttribute` on the parameter holds the config.
* Get the `factoryFactory` from the container via the type `IParameterFacetFactory<TAttribute>`.
* Call `factoryFactory.Create(parameterInfo, attribute)` to get `factory`.
* On activation, call `factory(context, services)` to get the value to pass to the grain constructor.
* Under the hood, `factory` will pull the strongly typed factory interface (eg `INamedStorageFeatureFactory`) from the container and call it using some value captured in the closure (eg the config specified on the attribute).

src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
```diff
@@ -10,18 +10,27 @@ namespace Orleans.Runtime
...
+        private readonly GrainConstructorFacetArguementsFactory arguementsFactory;
+        private readonly ConcurrentDictionary<Type, ObjectFactory> typeActivatorCache;
 
-        /// <inheritdoc />
-        public virtual object Create(IGrainActivationContext context)
+        /// <summary>
+        /// Public constructor
+        /// </summary>
+        /// <param name="service"></param>
+        public DefaultGrainActivator(IServiceProvider service)
```
sergeybykov: service -> serviceProvider?

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.Runtime
+{
+    internal class GrainConstructorFacetArguementsFactory
```
sergeybykov: The general recommendation is to limit type names to three words. Maybe GrainConstructorFacetArguementsFactory -> FacetArguementsFactory or at least GrainFacetArguementsFactory ?

sergeybykov: Nit: Also, it's probably better to use "ArgumentFactory" than "ArgumentsFactory" in type and variable names.

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
```diff
@@ -0,0 +1,95 @@
...
+        public object[] CreateArguements(IGrainActivationContext grainActivationContext)
+        {
+            ArguementsFactory arguementsFactory = arguementsFactorys.GetOrAdd(grainActivationContext.GrainType, type => new ArguementsFactory(this.services, type));
+            return arguementsFactory.CreateArguements(grainActivationContext);
+        }
+
+        /// <summary>
+        /// Facet Arguement factory
+        /// </summary>
+        private class ArguementsFactory
```
sergeybykov: Here's another case of a plural in type name. Doesn't match the XML doc description.

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
```diff
@@ -0,0 +1,95 @@
...
+            public ArguementsFactory(IServiceProvider services, Type type)
+            {
+                this.arguementFactorys = new List<Factory<IGrainActivationContext, object>>();
+                List<Type> types = new List<Type>();
+                IEnumerable<ParameterInfo> parameters = type.GetConstructors()
+                                                            .FirstOrDefault()?
+                                                            .GetParameters() ?? Enumerable.Empty<ParameterInfo>();
+                foreach (ParameterInfo parameter in parameters)
+                {
+                    var attribute = parameter.GetCustomAttribute<FacetAttribute>();
```
sergeybykov: This will throw an `AmbiguousMatchException` if more than one attribute is defined for the argument. `GetCustomAttributes` is safer in that sense, as it allows to handle such a situation more gracefully.

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
```diff
@@ -0,0 +1,95 @@
...
+        private class ArguementsFactory
+        {
+            private readonly List<Factory<IGrainActivationContext, object>> arguementFactorys;
+
+            public ArguementsFactory(IServiceProvider services, Type type)
+            {
+                this.arguementFactorys = new List<Factory<IGrainActivationContext, object>>();
+                List<Type> types = new List<Type>();
+                IEnumerable<ParameterInfo> parameters = type.GetConstructors()
+                                                            .FirstOrDefault()?
```
sergeybykov: Is this a safe thing to do - to pick a first constructor?

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
sergeybykov: Didn't we discuss the reverse approach - that an attribute type would reference its respective factory type, not the other way around? That would make attributes technically optional.

test/Tester/StorageFacet/Feature.Abstractions/IStorageFeature.cs
```diff
@@ -0,0 +1,58 @@
...
+    ///  This is the actual functionality the users need.
+    /// </summary>
+    /// <typeparam name="TState"></typeparam>
+    public interface IStorageFeature<TState>
+    {
+        string Name { get; }
+
+        TState State { get; set; }
+
+        Task SaveAsync();
```
sergeybykov: I think we should stop using `Async` suffixes in method names. We've done it in the past, which I think was a mistake in the hindsight.

src/Orleans/Utils/IKeyedServiceCollection.cs
```diff
@@ -10,7 +12,40 @@ namespace Orleans.Runtime
...
+            where TInstance : TService
+    {
+        private readonly TKey key;
+
+        public KeyedService(TKey key)
+        {
+            this.key = key;
+        }
+
+        public TService GetService(IServiceProvider services) => services.GetService<TInstance>();
```
xiazen: Can you add more documentation on these KeyedService classes? 

Based on the code, this `KeyedService<TKey, TService, TInstance>`  can only be a transient keyed service, and cannot be a singleton keyed services, is this correct?

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
jason-bragg: Pattern is for grain using facets to have a single public constructor, so this should be safe.

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
jason-bragg: Do we want to support multiple attributes?  Seems safer to limit until we see a need for more.  Thoughts?

sergeybykov: I agree. I only suggest a safer (IMO) way of handling situations where that rule is violated by developer.

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
ReubenBond: @sergeybykov that was my understanding.

src/Orleans/Utils/IKeyedServiceCollection.cs
jason-bragg: I'll add more comments, but this keyed service is simply a mapping between a name and a type.  If the type is registered as being transient (as are the only cases currently supported by extension functions) the named type will be transient.  If the type is registered as a singleton or scoped, those properties will also apply.

In AddTransientKeyedService we add the instance type as transient but the mapping from name to type as a singleton.  So the named type will be transient, as would be expected from the name of the extension function.

    collection.TryAddTransient<TInstance>(); 
    collection.AddSingleton<IKeyedService<TKey, TService>>(sp => new KeyedService<TKey, TService, TInstance>(key)); 

However, we could have an AddScopedKeyedService extension function that does the following

    collection.TryAddScoped<TInstance>(); 
    collection.AddSingleton<IKeyedService<TKey, TService>>(sp => new KeyedService<TKey, TService, TInstance>(key)); 

Which would created a scoped service by name.  Same can be done for singletons.

src/OrleansRuntime/Facet/GrainConstructorFacetArguementsFactory.cs
jason-bragg: My understanding was:

- Attributes were to be treated as part of the abstraction layer, they should contain no logic.
- Feature factories were to be obtained from the container
- Attribute processing logic should use the same feature factories (obtained from the container) that the grains use.

All of these conditions were met.

I also argued:

- There is not likely a single common pattern to wire attributes for all features up to their factories.
- If we did develop a single common set of logic that would wire up attributes for all features to their factories it would likely be complicated and hard to maintain.
- If, after introducing a couple of features (3-6 months from now) and have more experience adding features, we conclude that there is a single (or a small set) of patterns that fit all of our needs, we can make them part of the framework and deprecate the IParameterFacetFactory (now called IAttributeToFactoryMapper), which is low risk.

The cost of this pattern is that for each new feature (new attribute really), a relatively small set of logic is needed (implementation of IAttributeToFactoryMapper) to wire up the feature factories (from data in the attribute) to the general factory pattern cached by the argument factory.  This, imo, is a relatively small price to pay to keep us from painting ourselves into a corner should we misjudge the conformity of features.

test/Tester/StorageFacet/Feature.Abstractions/IStorageFeature.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public string StateName { get; }
+    }
+
+    /// <summary>
+    /// Creates a storage feature from a configuration
+    /// </summary>
+    public interface IStorageFeatureFactory
+    {
+        IStorageFeature<TState> Create<TState>(IStorageFeatureConfig config);
```
jdom: Instead of the non-generic interface with a generic method, could we have a generic interface? 
That would make the factory more explicit of what it can create. Sure, we can by default register the factory as an open generic, but that would allow us to override it for a specific state (explicit dependency principle).
Also, that would avoid a lot of MakeGenericMethod calls an each grain activation at runtime, and incur in that reflection cost only once per type (nevertheless this is an optimization, main reason is being explicit and allowing a different factory for a specific state type).

test/Tester/StorageFacet/Feature.Implementations/TableStorageFeature.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public void Configure(IStorageFeatureConfig cfg)
+        {
+            this.config = cfg;
+        }
+    }
+
+    public class TableStorageFeatureFactory : StorageFeatureFactory
+    {
+        public TableStorageFeatureFactory(IGrainActivationContext context) : base(typeof(TableStorageFeature<>), context)
```
jdom: I understand this is just test code, but still, I would expect the specific factory implementation be responsible of newing up (using ActivationServices or with the `new` keyword), instead of just passing the type up to the base class and only allowing the base class to use the container to resolve the type. This is already a factory, so it makes sense that it is the responsibility of this derived class to decide how it instantiates, instead of having the `StorageFeatureFactory.Create<TState>` implementation.

Also, my preference would be that this derived class instead of just using the container to resolve, that it uses the `new` keyword or the `ActivatorUtilities` class, but pass the configuration to the constructor, instead of requiring a separate step to pass the configuration to the feature.  With that, you can remove the `IConfigurableStorageFeature` interface.
It will also allow different implementations of the feature to participate in different stages of the activation lifecycle if they so require.

src/OrleansRuntime/Facet/ConstructorArguementFactory.cs
```diff
@@ -0,0 +1,88 @@
...
+                IEnumerable<ParameterInfo> parameters = type.GetConstructors()
+                                                            .FirstOrDefault()?
+                                                            .GetParameters() ?? Enumerable.Empty<ParameterInfo>();
+                foreach (ParameterInfo parameter in parameters)
+                {
+                    var attribute = parameter.GetCustomAttribute<FacetAttribute>();
+                    if (attribute == null) continue;
+                    // Since the IAttributeToFactoryMapper is specific to the attribute specialization, we create a generic method to provide a attribute independent call pattern.
+                    MethodInfo getFactory = GetFactoryMethod.MakeGenericMethod(attribute.GetType());
+                    var argumentFactory = getFactory.Invoke(this, new object[] { services, parameter, attribute }) as Factory<IGrainActivationContext, object>;
```
jason-bragg: Probably not.  Will use hard cast

test/Tester/StorageFacet/Feature.Abstractions/IStorageFeature.cs
jason-bragg: I prototyped both patterns, and this one worked better with DI and end user experience.

Issues I encountered:

- Having a typed factory would prohibit (or significantly complicate) managing  IStorageFeatureFactorys as named services.
- By having typed factories, each specialization would necessitate the construction of a new factory.
- Users that wished to have the factory injected in their grain constructor directly would need to have a factory for each type of data they wished to construct.

Regarding the MakeGenericMethod, it is only called when creating the argument factory the first time, all subsequent calls simply invoke the method.

test/Tester/StorageFacet/Feature.Abstractions/StorageFeatureAttribute.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using Orleans;
+
+namespace Tester.StorageFacet.Abstractions
+{
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public class StorageFeatureAttribute : FacetAttribute, IStorageFeatureConfig
```
jason-bragg: I was using the term 'feature' to indicate what part of the code dealt with a set of functionality we wished to expose, and 'facet' to refer to the general facet system for which features are exposed.
I left 'facet' out of the name of the feature attribute because the feature's attribute is part of the feature's abstraction layer, not part of the facet system, though it is part of the glue code that bridges the two.  Probably another reason to get rid of the marker attribute...
I can rename storage feature as the use of the term 'feature' seems to be causing confusion.

test/Tester/StorageFacet/Feature.Abstractions/IStorageFeature.cs
jdom: > - Having a typed factory would prohibit (or significantly complicate) managing IStorageFeatureFactorys as named services.
> - By having typed factories, each specialization would necessitate the construction of a new factory.

Is this right? we can still register `IStorageFeatureFactory<>` in the container (as an open generic), so that by default `IStorageFeatureFactory<Foo>` can be resolved without any additional registration.

> - Users that wished to have the factory injected in their grain constructor directly would need to have a factory for each type of data they wished to construct.

This is actually a plus, and supports the explicit dependency principle. It's the same reason why injecting `IServiceProvider` to get a specific service is worst than directly injecting that service (in application level code, of course, which is what we are talking about here). This also allows `IStorageFeatureFactory<Foo>`  and `IStorageFeatureFactory<Bar>` to use different factory implementations if they require so.

> Regarding the MakeGenericMethod, it is only called when creating the argument factory the first time, all subsequent calls simply invoke the method.

Isn't `StorageFeatureFactory.Create<TState>` calling `MakeGenericMethod` on every argument for every activation? I may be confused with the code flow though.

test/Tester/StorageFacet/Feature.Implementations/TableStorageFeature.cs
jason-bragg: > This is already a factory, so it makes sense that it is the responsibility of this derived class to decide how it instantiates.

Yes, it is the responsibility of this class how it instantiates the concrete type.  For convenience and reuse, I introduced a common base class, rather than duplicating code.  This does not alter there responsibility, nor prevent other factories, from doing this any way a developer wishes.  The Table and Blob implementation of the factories could have been entirely different, one newing an a impl that takes the config in it's constructor, the other deferring to the container and configuring after construction..?  This is just how I wrote it to avoid code duplication.

> Also, my preference would be that this derived class instead of just using the container to resolve, that it uses the new keyword or the ActivatorUtilities class, but pass the configuration to the constructor, instead of requiring a separate step to pass the configuration to the feature.

A second step is necessary for any instance that needs be wired up to the grain lifecycle, so in a common pattern, which again is not enforced, just a convenience so I didn't have to write/maintain as much code, nothing was gained by passing the config in the constructor.

This code is here to demonstrate the general pattern of feature, config, and factory interfaces.  Each implementer can implement as they please in a way that suits their needs, so any user with your sensibilities can freely follow them, and nothing in the way this is lain out prevents that..

test/Tester/StorageFacet/Feature.Infrastructure/StorageFeatureParameterFacetFactory.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Reflection;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans;
+using Orleans.Runtime;
+using Tester.StorageFacet.Abstractions;
+
+namespace Tester.StorageFacet.Infrastructure
+{
+    public class StorageFeatureAttributeMapper : IAttributeToFactoryMapper<StorageFeatureAttribute>
```
jason-bragg: StorageFeatureAttributeMapper

src/Orleans/Facet/FacetAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿
+using System;
+
+namespace Orleans
+{
+    /// <summary>
+    /// Marker attribute for facets
+    /// </summary>
+    public abstract class FacetAttribute : Attribute
```
jason-bragg: We can do this different, of course.  Just a question of trade offs.

Marker attribute requires attributes to be inherited from it, requiring assemblies defining facet attributes to depend on Orleans.  It also may provide for better error detection since a facet attribute which does not have a mapper would indicate a misconfiguration.  Attributes on arguments in grain constructors is probably not a common enough occurrence for performance concerns from looking up a mapper for every property attribute, but the potential cost should be considered.  Considering all attributes also raises the question of multiple mapped attributes, and how to detect and handle that.

All in all, imo, having a marker interface make the system simpler and more strait forward, and costs very little, but I don't have a strong opinion on this.

@sergeybykov, @ReubenBond, ??, thoughts?

test/Tester/StorageFacet/Feature.Abstractions/IStorageFeature.cs
jason-bragg: You can register IStorageFeatureFactory<> in the container, correct, but you can't resolve it as a named service as that takes resolving a generic with an imbedded generic.
This is the technical difficulty with what generic factories, the rest of the issues were more subjective.  I don't think the tradeoffs favor generic factories, but unless we deal with the technical limitations of containers to resolve imbedded generics, other debatable points are moot.

Regarding the MakeGenericMethod in StorageFeatureFactory, that one is an implementation detail of the factory, so I wasn't considering that one, since it's trivial to work around without moving to a generic factory.  I was referring to the one used to call the factory, in StorageFeatureAttributeMapper.  The general pattern builds no generic types or methods per grain activation, only the implementation's factories do that, and I'm not terribly concerned about the implementation of test features.

test/Tester/StorageFacet/Feature.Implementations/TableStorageFeature.cs
jason-bragg: Killing StorageFeatureFactory.  Adds some code duplication, but make example more clear.

src/Orleans/Facet/FacetAttribute.cs
jdom: Regarding catching errors, it's not very different if we don't have a mapper registered, since you'll get a clear error if you try to resolve `IExampleStorage<T>` because there will be no service registered for it.

Regarding perf, it doesn't need to consider all attributes, just attributes that have a mapper associated with it. Also the (very minimal) cost is only paid once anyway, since reflection only happens the first time the grain type is encountered for activation. Also, even not having abstract attributes, and just have sealed ones is better for perf, so we "pay" for checking for a few attributes, but with faster look ups each. Having said that, I wouldn't put perf as a pro or con to keep the base attribute, as there is no difference after the type is cached.

I kind of like the idea of being able to use non-Orleans attributes, or even reuse `StorageProviderAttribute` for example, so you can use the old and the new approach with the same attribute, but varies whether you apply it to the `Grain<T>` class, or to a constructor parameter argument (or to an injected property in the future). After all it's just metadata, that can be interpreted by different logic in different contexts, to get the same behavior.

@sergeybykov, @ReubenBond, ??, thoughts?

test/Tester/StorageFacet/Feature.Abstractions/IStorageFeature.cs
jason-bragg: Removed StorageFeatureFactory.

jason-bragg:

src/OrleansRuntime/Catalog/DefaultGrainActivator.cs
```diff
@@ -10,18 +10,27 @@ namespace Orleans.Runtime
     /// </summary>
     public class DefaultGrainActivator : IGrainActivator
     {
-        private readonly Func<Type, ObjectFactory> createFactory = type => ActivatorUtilities.CreateFactory(type, Type.EmptyTypes);
-        private readonly ConcurrentDictionary<Type, ObjectFactory> typeActivatorCache = new ConcurrentDictionary<Type, ObjectFactory>();
+        private readonly ConstructorArguementFactory argumentFactory;
+        private readonly ConcurrentDictionary<Type, ObjectFactory> typeActivatorCache;
```
jdom: You probably missed this comment

src/Orleans/Facet/FacetAttribute.cs
jason-bragg: > Regarding catching errors, it's not very different 

Agreed, main difference is that by knowing a mapper was expected, we can provide a more specific error, and if we move the building of these mappings to startup time, rather than grain activation time, we can fail earlier.  Again, not much different, but slight advantage for marker attributes, IMO, slight..

> Regarding perf, it doesn't need to consider all attributes, just attributes that have a mapper associated with it.

I don't really understand what you're saying here, as the check cost I described was the lookup of the mapper in DI, which is what would be needed to consider "just attributes that have a mapper associated with it".

> I wouldn't put perf as a pro or con to keep the base attribute

Agreed, my intention was simply to prompt an examination of the potential differences, especially in services that use other containers which may provide their own attribution of parameters (I'd not suspect as being terribly common).  I don't see the perf differences as significant either.

Let's see if we can agree on what the Pros/Cons are:
Using marker attribute.
Pros:

- Allows more specific error handling.
- Allows for checking at silo start.
- Can't repurpose preexisting or third party attributes.

Cons:

- Attributes require dependency on Orleans.
- Can't repurpose preexisting or third party attributes.

jdom: Agree, yes. Although "Can't repurpose preexisting or third party attributes." is just a con, not both a pro and a con.
The more specific error checking is a potential improvement (not that it is already doing it), and might not be that useful anyway, since it's probably very trivial to see that you didn't configure any storage provider once it fails to resolve the storage object. But yes, it's a very minor con.

jason-bragg: I marked "Can't repurpose preexisting or third party attributes" as a pro and con because I think there is room for debate on whether that is strictly a good thing, especially with well known attributes of which people already expect specific meanings and interpretations.

I'm not suggesting that all of the pros/cons should be weighted equally, so feel free to weigh each as you feel appropriate (including 0 weight!).

src/Orleans/Utils/IKeyedServiceCollection.cs
jdom: The more I look at this approach (which is similar to what we talked about before we protoyped [Orleans hosting](https://github.com/jdom/Orleans.Hosting/blob/master/Orleans.Hosting/)), the less convinced I am that we should be putting the name mapping itself as singleton services in DI.
I think the previously proposed approach of separating the building of the individual services from the keyed service collection itself is useful. After all, there is no need to have the factory be a transient service, as the grain activation context could (and IMO should, regardless of this proposed change) be passed in when creating the concrete storage feature, and not be injected in the constructor.
If we make that minor change (having the service not be transient), then it opens up a much cleaner separation. An example of that potential separation can be seen here (notice that the key->service mapping is not part of the more general purpose ServiceCollection registrations): https://github.com/jdom/Orleans.Hosting/blob/master/Orleans.Hosting/KeyedServiceCollection.cs

If you want, I can submit a small PR adapting that old branch to your branch specifically, as there is some small logic that is not yet available in our current silo initialization, which is the step to configure the silo after the IServiceProvider has been built.

Otherwise, I'm fine to not do this full change now and potentially improve it later, once we add that mentioned capability (which we'll have at least when we incorporate the SiloBuilder). Nevertheless it would be nice to at least not have the `INamedExampleStorageFactory` implementations be transient services, but instead have the `INamedExampleStorageFactory` receive the `IGrainActivationContext` on every call to `Create`. I understand this factory is in test code, but the support for transient keyed services is not, as well as this is defining how the real implementations will use this feature.

jason-bragg: > Less convinced I am that we should be putting the name mapping itself as singleton services in DI.

Understood.  I'm not committed to this solution.  We discussed solving this different ways (the prototype you linked, earlier factory patterns, possibly using named options, ..., but none of those are in place now.  If/when we spend the time to improve our support for named services, I'd expect this code to be refactored/replaced/improved.

> There is no need to have the factory be a transient service

It's worse than that.  It's a scoped service.  It needs to be that way currently to obtain the context or the correct service provider.  I spoke with Sergey about this problem and will be opening an issue about this, I simply didn't want to complicate this PR with that problem, as it is something we can address in future changes.  To be clear, the requirement that the factories be scoped is orthogonal to the named mappings being in DI.  These are two separate (separable) problems.

> If you want, I can submit a small PR adapting that old branch to your branch specifically.

I would love to see improvements on this, after this PR is checked in and unblocks other work.  This has always been the plan.  I'm less open to introducing further changes at this stage that do not address specific problems in the general pattern.  The use of the keyed services is entirely within the ExampleStorage sample code and is not part of the general pattern.  Features can do this any way they like, as long as it works with the framework and respects our DI limitations.  I'd, however, object to each feature providing it's own keyed service implementation, as that would incur a high maintenance cost.  In short, we should solve the keyed service problem and use the same solution for all our feature factories, but that does not mean the solution in this pr **IS** that solution.

> have the INamedExampleStorageFactory receive the IGrainActivationContext on every call to Create.

I am not 100% against this, but the feature factories (INamedExampleStorageFactory, IExampleStorageFactory) are also intended to be injected directly into the grain and called by the grain logic, which does not have access to the IGrainActivationContext, unless it too is passed into the grain constructor.  There are ways around this, however, one of the requirements negotiated in the general pattern was that the attribute handling logic must use the same factories that the grain does to create the feature, so if we add the IGrainActivationContext to the factories, it must be passed by the grains, if we don't want to pass the grain context in the grains, the attribution logic can't pass it either (or we abandon that negotiated requirement).

jason-bragg:

src/Orleans/Utils/IKeyedServiceCollection.cs
jdom: Ah, you are absolutely right about the simplicity of usage of the factories if injecting the dependency as a scoped service. That's a good enough reason to keep that as is at least for now.
BTW, you mentioned that factories should be scoped, but you are registering them as transient. Is this a mistake or by design?

jason-bragg: < you mentioned that factories should be scoped, but you are registering them as transient. Is this a mistake or by design.

Mistake.  I'm corrected the scoping on the named and default factories, but missed it in the name mapping.

After in person discussion, moving all factories to transient to prevent potential GC hit.  Trading fewer long lasting allocations for more short term allocations.

jdom: BTW, just talked in person, but mentioning it here for others: I'm not suggesting switching to Scoped... Transient is probably better, otherwise we would hold a reference to the factory during the 2 hours that the grain is activated (or whatever the lifetime of the grain is) for no reason. It's better to use the factories when activating and just garbage collect them afterwards.

src/Orleans/Facet/FacetAttribute.cs
ReubenBond: We don't necessarily need a base attribute, but I'm not convinced that we shouldn't have some kind of static marker (attribute on attribute, interface on attribute, base class on attribute) - just to make things obvious. Attribute on attribute would require more work on this PR since it cannot be used as a generic type constraint and generally requires more dynamic checking.

If we went with an interface instead, we could more easily change the requirement later to open it up to any attribute type later. That way we're narrow on what we accept today but we have the option to change our decision later without breaking anyone.

How does that sound?

jason-bragg: >If we went with an interface instead, we could more easily change the requirement later to open it up to any attribute type

Does it make it more easy because we'd have to write attribute filtering logic now, so it would be easier to modify that logic later?
I don't see much difference between using an interface, base class, or base attribute marker, other than using GetCustomAttributes() + filtering the attributes ourselves instead of using GetCustomAttribute<T>().
I'm not opposed to some other marker, I'm just not seeing how it leads to "more easily change the requirement later", and it doesn't enable the scenarios Julian seems interested in...

ReubenBond: Just because modifying class hierarchy is a larger break than adding/removing interfaces.

jdom: Correct, an interface doesn't address all my concerns.
Still, if we decide to keep some marker contract, I agree that an interface is slightly better, since we could make StorageProviderAttribute implement that interface for example, instead of declaring that the attribute is only for facets, just that it is supported by facets. And in the future we can just obsolete the interface if we decide to support any 3rd party attribute.

jdom:
As I mentioned in a long series of discussions internally, **I don't like this syntactic sugar with attributes**, because:
- Most of this can be done with well-known factories, which can be as expressive as the attributes, and a lot more flexible too (just compare in the [test examples](https://github.com/dotnet/orleans/pull/3279/files#diff-96c3b8e919ff3da3f114fb2dd873711d) how you inject a grain with factories, and how you do it with attributes). So we are just giving an alternative option to do what can already be easily done (which could be argued one way or the other if it's easier to use).
- We are adding a lot of complexity (notice that this entire PR mostly deals with the attribute magic logic), 
- But most importantly we are adding a very big constraint (not in this PR, but with the comment added on `IGrainActivator`): If *system* (non-Application) grains rely on the attributes instead of the plain factories, it means that we add a non-trivial limitation to the `IGrainActivator` extensibility point. All future implementations of this interface (including `ServiceBasedGrainActivator`, which we alluded too in the past, which emulates [ServiceBasedControllerActivator](https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/Controllers/ServiceBasedControllerActivator.cs)) will have to white or black list grains that it can use to resolve, because these system grains will rely on the logic in `DefaultGrainActivator` only. If they use the factories, then all that system grains need to do is play nice with conforming containers, which they were already doing.

The last is my biggest concern, and I would be completely fine to have this syntactic sugar as an optional thing, even if it wouldn't be my personal preference.

What do we gain by using attributes in our system grains and unnecessarily adding this constraint on our end users who do want to use the `ServiceBasedGrainActivator` (or a custom one)? The ones who will use the different grain activators will not be a negligible amount of people (even if not the majority). On the other hand, system or reusable grains are probably 0.001% of all grain types ever built, and probably more like 0.000000001% of grains built by our end users (almost all will be included in Orleans, only a few in reusable Orleans components built outside of this repo). So who do we prefer to constrain when documenting this and what do we really gain from this cost?

We only started this very long discussion of how DI should work or not work in Orleans because we decided to deviate from the road laid out by the ASP.NET team, who has been purposedly designing and improving all the edge cases for extensibility through many years. So even if with some limitations, it would be safe to follow them. MVC Controllers are extremely similar to Grains in these regards, so we were not trying to shoehorn a solution that was not appropriate for our use case, so for that I can say that we are not strictly following this mantra: [De-inventing the wheel](https://blogs.msdn.microsoft.com/orleans/2017/04/12/de-inventing-the-wheel/)

**I'm hesitantly approving this PR** after giving up from discussing, but mainly due to the future of this feature as described in the `IGrainActivator` note, but I'm OK with this as optional syntactic sugar. This PR itself did not modify any system grains to make this mandatory just yet, and the implementation is OK (probably can improved in separate PRs, but there are no other major issues with it).

src/Orleans/Facet/FacetAttribute.cs
jason-bragg: I'll change the marker to an interface.  I considered (and prototyped) both at various times and settled on the abstract attribute base class simply because it did not require me to write a filter for the custom attributes (which is admittedly trivial).

sergeybykov:
We are making two bets here.

1. That the attributes are more declarative in nature than direct use of factories (although we'll support both APIs).
2. That non-default grain activators are a very narrow case that can be easily constrained to application grains.

These are educated guesses, but guesses nevertheless. If real life usage proves us wrong (unclear yet how we can quantify this), it should be fairly easy for us to adjust the runtime behavior in the future.

jdom:

jason-bragg:
All comments addressed.

jdom:

PR #3282, opened 2017-08-08
Added missing registration for 7-component tuple serialization

DixonDs:
While working on #3281, I've noticed that there is implemented support for 7-component tuples in `BuiltInTypes` (see `DeepCopyTuple7`/`SerializeTuple7`/`DeserializeTuple7`), but it is not registered in `SerializationManager`


sergeybykov:
Thank you, @DixonDs!

PR #3284, opened 2017-08-09
Legacy LogConsumer provider sink for Microsoft.Extensions.Logging

xiazen:
Added `OrleansLogger `and `OlreansLoggerProvider` to preserve legacy Orleans logging features and configuration style, including

- Message Bulking, including its configuration
- `ILogConsumer`, `IClosableLogConsumer`, `IFlushableLogConsumer`, including its configuration
- per category Severity Overrides, whose configuration style preserves old severity config style

Added OrleansLoggerExtension methods to preserve old Orleans logging methods. 

Wired asp.net logging abstraction into silo's DI container, with OrleansLoggerProvider as default logger provider, LogFileConsumer as default ILogConsumer. One can now configure his LogFactory into DI, one can also create ILogger<T> from DI now. See tests as example.

`OrleansLoggerProvider` supports
- Create `new OrleansLoggingDecorator(bulkingConfig, OrleansLogger logger)`
- Configure LogConsumers
- Configure Severity overrides
- Configure message bulking

`OrleansLoggingDecorator` 
- a decorator class which user can plug in their ILogger and then gain message bulking feature on top of their ILogger

`OrleansLogger` supports
- log messages to configured LogConsumers
- respects Severity overrides in `IsEnable(LogLevel level)` method
- Map LogLevel to Severity

`FileLogConsumer` is the replacement for FileTelemetryConsumer, which also preserves Orleans logging message format. Since general consensus I get is TelemetryConsumer should be replaced by asp.net built in log providers, or at least separate out from OrleansLogging. 

`ILogger `and `OlreansLoggerProvider` is supposed to replace `Logger`, `LogManager` eventually. OrleasnLogger is supposed to replace LoggerImpl. I haven't changed them  in this PR. So they work as is for now. 

About how `Microsoft.Extensions.Logging.LoggingFilter` and `OrleansLoggerSeverityOverrides` play together:
One can use either `Microsoft.Extensions.Logging.LoggingFilter` or `OrleansLoggerSeverityOverrides` to configure per logger Severity/LogLevel. Both config method works with `OrleansLogger`. If used together, think it as two layer of filters, need both filter being true for the end result to be true. 
See tests to get a clearer idea

EDIT:
- TelemetryConsumers are not wired into OrleansLogging eco system. How it would play into the picture is unclear, since there's different opinion about it. Need to get back to that discussion.  **EDIT:** I have received general consensus that Telemetry will evolve to have its own system, which doesn't share infrastructure with logging. So remove this TODO from this PR. This will be resolved in its own path with its own series of PRs. 
- support logging scope. **EDIT: ** Can be done in a separate PR. It wasn't part of Orleans legacy logging feature. So merge this PR without this piece won't break anyone. But need to be added before 2.0
- Move logging configuration to SiloBuilder. **EDIT:**  logging can be configured with DI. So I don't think there need to be extra config method on SiloBuilder for configuring logging. Although there can be that kind of config method on SiloBuilder, but I think it is out of scope of this PR. So remove this TODO from this PR.


src/Orleans/Logging/ILogConsumer.cs
```diff
@@ -47,19 +47,15 @@ public interface ILogConsumer
...
-        /// <param name="loggerType">The type of logger the message is being traced through.</param>
         /// <param name="caller">The name of the logger tracing the message.</param>
-        /// <param name="myIPEndPoint">The <see cref="IPEndPoint"/> of the Orleans client/server if known. May be null.</param>
         /// <param name="message">The message to log.</param>
         /// <param name="exception">The exception to log. May be null.</param>
         /// <param name="eventCode">Numeric event code for this log entry. May be zero, meaning 'Unspecified'. 
         /// In general, all log entries at severity=Error or greater should specify an explicit error code value.</param>
         void Log(
             Severity severity,
-            LoggerType loggerType,
```
xiazen: We agreed to remove LoggerType in that meeting. Is this correct? 

If LoggerType is removed from Orleans logging infra, its downstream ILogConsumer will not be able to know the LoggerType.

src/Orleans/Telemetry/Severity.cs
```diff
@@ -4,10 +4,11 @@
     public enum Severity
     {
         Off = 0,
-        Error = 1,
```
xiazen: it changed to have a better mapping to LogLevel. But yeah I think I should change it back. it's not worthwhile

src/Orleans/Logging/LoggerImpl.cs
```diff
@@ -240,13 +240,12 @@ private void WriteLogMessage(int errorCode, Severity sev, string format, object[
             {
                 try
                 {
-                    consumer.Log(sev, loggerType, Name, message, LogManager.MyIPEndPoint, exception, errorCode);
+                    consumer.Log(sev, Name, message, exception, errorCode);
```
xiazen: Yes that's correct. I Haven't made that change in this PR. Orleans.Logger will become a wrapper on top of ASP.Net ILogger . Is this what you are asking?

xiazen: The following PR will be replacing Orleans legacy infra with new ASP.net abstraction

jdom:
Haven't started reviewing, althogh from the description it looks like we are deprecating stuff and calling it `OrleansXXX` (ie OrleansLogger) and similar. I would call it `LegacyXXX` instead, otherwise it prompts users as the way to go for future code. Thoughts?

vNext/src/Orleans/Orleans.csproj
```diff
@@ -33,6 +33,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="1.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.0-preview2-final" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0-preview2-final" />
```
jdom: Logging 2.0.0 was released. Although it's a .NET Standard 2.0 library, so let's figure this out before we even try to merge

galvesribeiro: It works. But be aware that only people with latest VS/Nuget will be able to add `netstandard2.0` libraries on .Net Full framework projects.

xiazen: Yeah I need to update the PR with Logging 2.0.0. The PR is created before that was released.

xiazen:
I was thinking adding `[System.Obsolete] `attribute on those classes.  Do you think it is not enough?

vNext/src/Orleans/Orleans.csproj
xiazen: Yes this PR is not mergeable yet. There's some TODOs(see description) which need to be addressed before merging. I'd like to do some of there todos after the two sln merged into one sln.

vNext/src/Orleans/OrleansLogging/FileLogConsumer.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Net;
+using System.Text;
+using System.Threading;
+
+namespace Orleans.Extensions.Logging
+{
+    /// <summary>
+    /// FileLogConsumer, which logs message into a file in orleans logging message style
+    /// </summary>
+    public class FileLogConsumer : ICloseableLogConsumer, IFlushableLogConsumer
```
xiazen: `FileLogConsumer` is the replacement for `FileTelemetryConsumer`. FileTelemetryConsumer is the default component in Orleans to record tracing messages. But since we are separating logging and telemetry, so FileTelemetryConsumer won't be compatible with asp.net logging abstraction anymore. I created FileLogConsumer to replace its role in logging, just in case that people still want to log messages in Orleans logging message style.

xiazen:
ready for review

xiazen:

src/OrleansLoggerProvider/OrleansLoggerProvider.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
```
xiazen: ILogConsumer and all the deprecated logging features will be moved here as much as possible when internal runtime finished migrating off the legacy logging infra

src/OrleansLoggerProvider/OrleansLoggingDecorator.cs
```diff
@@ -74,9 +74,9 @@ public OrleansLoggingDecorator(MessageBulkingConfig config, ILogger thirdPartyLo
         /// <param name="eventId"></param>
         /// <param name="errorCode"></param>
         /// <returns></returns>
-        public static EventId CreateEventId(int eventId, int errorCode)
```
xiazen: this will be replaced by LoggingUtils.CreateEventId n #3333 after it merged

src/OrleansLoggerProvider/OrleansLogger.cs
```diff
@@ -0,0 +1,119 @@
...
+                    Console.WriteLine("Exception while passing a log message to log consumer. LogConsumer type:{0}, logger name:{1}, severity:{2}, message:{3}, error code:{4}, message exception:{5}, log consumer exception:{6}",
+                        consumer.GetType().FullName, this.name, sev, message, errorCode, exception, exc);
+                }
+            }
+
+            //flush flushable consumers
+            if ((DateTime.UtcNow - lastFlush) > flushInterval)
+            {
+                lastFlush = DateTime.UtcNow;
+                foreach (IFlushableLogConsumer consumer in this.logConsumers.OfType<IFlushableLogConsumer>())
```
ReubenBond: The `this.logConsumers.OfType<IFlushableLogConsumer>()` could be computed once and enumerated to a list in the ctor.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -886,6 +889,16 @@ private void HandleProcessExit(object sender, EventArgs e)
             }
         }
 
+        private void DisposeLogProviders(IServiceProvider svc)
+        {
+            var providers = svc.GetServices<ILoggerProvider>();
```
ReubenBond: This shouldn't be required - the container is disposed immediately after this call and it is responsible for disposing these.

src/OrleansLoggerProvider/OrleansLoggerProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        /// <summary>
+        /// Find severity level for the logger with categoryName
+        /// </summary>
+        /// <param name="categoryName"></param>
+        /// <returns></returns>
+        public Severity FindSeverityLevel(string categoryName)
+        {
+            if (this.loggerSeverityOverrides.LoggerSeverityOverrides.ContainsKey(categoryName))
```
ReubenBond: better to use `TryGetValue` here

src/OrleansRuntime/Silo/Silo.cs
xiazen: no it isn't disposing the providers. I tested it.

xiazen: nvm. you are right. I was doing things wrong

xiazen:
feedback I received online or offline require changes which is blocked by #3368 . Will go back after #3368 is merged

xiazen:
ready for review

xiazen:
please review this, and hopefully can merge this soon. Since this PR adds a new csproj, so the longer we delay its merging, the more I need to deal with changes from project restructuring #3353 .

jdom:
Anyone has thoughts on [this comment](https://github.com/dotnet/orleans/pull/3284#issuecomment-322617798)? It doesn't feel right that we are *adding* OrleansLogger and it's already obsolete. IMO it should be LegacyLogger or even LegacyOrleansLogger

test/Tester/OrleansLoggingTests.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Tester
+{
+    [TestCategory("BVT"), TestCategory("OrleansLogging")]
+    public class OrleansLoggingTests
+    {
+        [Fact]
+        [Obsolete]
```
jdom: Why are these tests marked obsolete?

test/Tester/OrleansLoggingTests.cs
```diff
@@ -0,0 +1,102 @@
...
+                new FileLogConsumer($"{this.GetType().Name}.log")
+            }, severityOverrides));
+            var serviceProvider = serviceCollection.BuildServiceProvider();
+            //get logger
+            var logger = serviceProvider.GetRequiredService<ILogger<OrleansLoggingTests>>();
+            Assert.True(logger.IsEnabled(LogLevel.Warning));
+            Assert.False(logger.IsEnabled(LogLevel.Information));
+
+            //dispose log providers
+            (serviceProvider as IDisposable)?.Dispose();
```
jdom: instead of this line, wrap the `serviceProvider` into a `using()` statement. Same with other facts.

src/OrleansLoggerProvider/OrleansLoggingUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// Format message in orleans logging style
+        /// </summary>
+        /// <param name="timestamp"></param>
+        /// <param name="severity"></param>
+        /// <param name="caller"></param>
+        /// <param name="message"></param>
+        /// <param name="myIPEndPoint"></param>
+        /// <param name="exception"></param>
+        /// <param name="errorCode"></param>
+        /// <param name="includeStackTrace"></param>
```
jdom: you copied this from the original code, why not include the XML comments?
```
        /// <summary>The method to call during logging to format the log info into a string ready for output.</summary>
        /// <param name="timestamp">Timestamp of the log message.</param>
        /// <param name="severity">The severity of the message being traced.</param>
        /// <param name="loggerType">The type of logger the message is being traced through.</param>
        /// <param name="caller">The name of the logger tracing the message.</param>
        /// <param name="myIPEndPoint">The <see cref="IPEndPoint"/> of the Orleans client/server if known. May be null.</param>
        /// <param name="message">The message to log.</param>
        /// <param name="exception">The exception to log. May be null.</param>
        /// <param name="errorCode">Numeric event code for this log entry. May be zero, meaning 'Unspecified'.</param>
```

galvesribeiro:
I'm with you @jdom.

src/OrleansLoggerProvider/OrleansLogger.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Linq;
+using System.Net;
+
+namespace Orleans.Extensions.Logging
+{
+    /// <summary>
+    /// OreansLogger supports legacy orleans logging features, including <see cref="ILogConsumer"/>, <see cref="ICloseableLogConsumer">,
+    /// <see cref="IFlushableLogConsumer">, <see cref="Severity">. 
+    /// </summary>
+    [Obsolete]
```
jdom: All obsolete attributes should have a description of what should be the replacement. You should probably define a constant with this string, as you will use it in multiple places.

src/OrleansLoggerProvider/OrleansLoggerProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// <summary>
+    /// Orleans severity overrides on a per logger base
+    /// </summary>
+    [Obsolete]
+    public class OrleansLoggerSeverityOverrides
+    {
+        /// <summary>
+        /// LoggerSeverityOverrides, which key being logger category name, value being its overrided severity
+        /// </summary>
+        public ConcurrentDictionary<string, Severity> LoggerSeverityOverrides { get; }
```
jdom: just do `public ConcurrentDictionary<string, Severity> LoggerSeverityOverrides { get; } = new ConcurrentDictionary<string, Severity>();` and get rid of the constructor.
Event better, don't use a concurrent dictionary, as it is only used for configuration, just do:

`public IDictionary<string, Severity> LoggerSeverityOverrides { get; set; } = new Dictionary<string, Severity>();` 

This way users can set to their own dictionary instead of adding to this one.

src/OrleansRuntime/Hosting/DefaultSiloServices.cs
```diff
@@ -55,6 +55,7 @@ internal static void AddDefaultServices(IServiceCollection services)
             services.TryAddFromExisting<IMessagingConfiguration, GlobalConfiguration>();
             services.TryAddFromExisting<ITraceConfiguration, NodeConfiguration>();
 
+            services.AddLogging();
```
jdom: technically this isn't needed nor used in this PR, right?

src/OrleansLoggerProvider/OrleansLoggingDecorator.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Default bulk message interval
+        /// </summary>
+        public static readonly TimeSpan DefaultBulkMessageInterval = TimeSpan.FromMinutes(1);
+    }
+
+    /// <summary>
+    /// OrleansLoggingDecorator class. User can plug in their own ILogger implementation into this decorator class to add message bulking feature on top of their logger. 
+    /// Message bulking feature will just log eventId count if the same eventId has appear more than BulkMessageLimit in a certain BulkMessageInterval.
+    /// </summary>
+    public class OrleansLoggingDecorator : ILogger
```
jdom: I'd rename to `MessageBulkingDecoratorLogger` or even (`EventBulkingDecoratorLogger`), as there is nothing inherent to Orleans nor the legacy logger. In theory this could be externalized as a separate package that non-Orleans users could use (not saying we should do the latter now though, just mentioning how decoupled this is).

src/OrleansLoggerProvider/OrleansLoggingDecorator.cs
```diff
@@ -0,0 +1,134 @@
...
+        private readonly ConcurrentDictionary<int, int> recentLogMessageCounts = new ConcurrentDictionary<int, int>();
+        private long lastBulkLogMessageFlushTicks = DateTime.MinValue.Ticks;
+        private readonly ILogger thirdPartyLogger;
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="config"></param>
+        /// <param name="thirdPartyLogger"></param>
+        public OrleansLoggingDecorator(MessageBulkingConfig config, ILogger thirdPartyLogger)
```
jdom: please rename thirdPartyLogger to `decoratedLogger` everywhere (or `wrappedLogger` if you prefer), as 3rd party isn't correct.

src/OrleansLoggerProvider/OrleansLoggingDecorator.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Threading;
+using Microsoft.Extensions.Logging;
+
+namespace Orleans.Runtime
+{
+
+    /// <summary>
+    /// Config for message bulking feature
+    /// </summary>
+    public class MessageBulkingConfig
```
jdom: how about `MessageBulkingOptions` (or `EventBulkingOptions` if you also rename the decorator that way)?

src/OrleansLoggerProvider/OrleansLoggingFactoryExtension.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <see cref="MessageBulkingLoggerProvider{TDecoratedLoggerProvider}"/> into <param name="builder"></param>.
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="provider"></param>
+        /// <param name="messageBulkingConfig"></param>
+        /// <returns></returns>
+        public static ILoggingBuilder AddMesageBulkingLoggerProvider<TDecoratedLoggerProvider>(this ILoggingBuilder builder, TDecoratedLoggerProvider provider, MessageBulkingConfig messageBulkingConfig)
+            where TDecoratedLoggerProvider : ILoggerProvider
+        {
+            builder.AddProvider(new MessageBulkingLoggerProvider<TDecoratedLoggerProvider>(provider, messageBulkingConfig));
```
jdom: this should probably be conditional on whether the bulking options are not null. Otherwise just add the other provider directly.

src/OrleansLoggerProvider/OrleansLoggingFactoryExtension.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="logConsumers">log consumers which user want to write log events to</param>
+        /// <param name="messageBulkingConfig">config for message bulking feature</param>
+        /// <returns></returns>
+        [Obsolete]
+        public static ILoggingBuilder AddOrleansLogging(
+            this ILoggingBuilder builder,
+            List<ILogConsumer> logConsumers,
+            IPEndPoint ipEndPoint = null,
```
jdom: hmmm, this ipEndPoint here might be tricky. We can probably address in a different PR, but maybe a solution is to add extension methods for ISiloBuilder and IClientBuilder also, so that we can resolve the IP endpoint first and then add the provider, but not sure about order of initialization.

test/Tester/OrleansLoggingTests.cs
xiazen: to avoid compile warnings

src/OrleansRuntime/Hosting/DefaultSiloServices.cs
xiazen: that's right. I can remove it

src/OrleansLoggerProvider/OrleansLoggingDecorator.cs
xiazen: I think `EventBulkingConfig` is good. Why do you prefer **options** over **config**, as in I think this is a config class?

src/OrleansLoggerProvider/OrleansLoggingDecorator.cs
xiazen: will do

src/OrleansLoggerProvider/OrleansLoggingDecorator.cs
xiazen: will do

src/OrleansLoggerProvider/OrleansLoggingFactoryExtension.cs
xiazen: we can discuss this in another PR. But IMO, this extensions should be decoupled with core interface as much as possible, since it is being deprecated. I see we provide convenience for users on how to resolve IpEndpoint using extenion method on ISiloBuilder, but at the mean time, we also couple those two closer than before, event it is just an extension method. I'm not concerned about resolving ipEndpoint, since it is pretty easy for user to resolve it as in they are the ones who configure that,.

src/OrleansLoggerProvider/OrleansLoggingDecorator.cs
jdom: with the new builders, we are moving to using strongly typed Options (that could be bound using Microsoft.Extensions.Options). We'll get rid of our legacy Config and Configuration classes, and morph them to Options (meaning we'll remove any parsing logic from them too, something similar to this message bulking options)

src/OrleansLoggerProvider/OrleansLoggingFactoryExtension.cs
jdom: yes, let's discuss afterwards. Although it might not be as easy to know the ipendpoint though, as it might be resolved and bound by the silo itself, not the user

src/OrleansLoggerProvider/OrleansLoggingFactoryExtension.cs
xiazen: I think the correct fix here is making bulkConfig optional, since it do have default config values. 
This extension method is designed to add event bulking feature onto the user's logger provider, so its behavior should stick with that IMO. 
If the user doesn't want event bulking feature, he/she can just use native builder.AddProvider(their logger provider).

xiazen:
addressed feedbacks. added legacy keyword to all obsolete classes. Also added Orleans.Extensions.Logging.Legacy namespace, which included all obsolete classes.

test/Tester/OrleansLoggingTests.cs
```diff
@@ -23,10 +24,10 @@ public void OrleansLoggingCanConfigurePerCategoryServeriyOverrides()
...
             var severityOverrides = new OrleansLoggerSeverityOverrides();
-            severityOverrides.LoggerSeverityOverrides.TryAdd(this.GetType().FullName, Severity.Warning);
-            serviceCollection.AddLogging(builder => builder.AddOrleansLogging(new List<ILogConsumer>()
+            severityOverrides.LoggerSeverityOverrides.Add(this.GetType().FullName, Severity.Warning);
+            serviceCollection.AddLogging(builder => builder.AddLegacyOrleansLogging(new List<ILogConsumer>()
             {
-                new FileLogConsumer($"{this.GetType().Name}.log")
+                new LegacyFileLogConsumer($"{this.GetType().Name}.log")
             }, severityOverrides));
             var serviceProvider = serviceCollection.BuildServiceProvider();
```
jdom: I meant that you do:
```
using (var serviceProvider = serviceCollection.BuildServiceProvider())
{
...
}
```
and then do not do not use Dispose explicitly. But on hindsight, BuildServiceProvider returns IServiceProvider which is not itself disposable, the concrete implementation is, so you might not be able to do it

src/OrleansLoggerProvider/OrleansLoggingFactoryExtension.cs
jdom: well, the point is that this is also being used by AddLegacyOrleansLogger... Then if not here, do the conditional there first and decide whether to call AddEventBulkingLoggerProvider whether that one is non-null

src/Orleans/Orleans.csproj
```diff
@@ -37,6 +37,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjection)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingVersion)" />
```
jdom: please remove this one for now if it's not needed

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.LegacyOrleansLoggerProvider</PackageId>
+    <Title>Microsoft Legacy Orleans Logger Provider</Title>
+    <Description>Library of LegacyOrleansLoggerProvider.</Description>
+    <PackageTags>$(PackageTags) Orleans</PackageTags>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AssemblyName>LegacyOrleansLoggerProvider</AssemblyName>
```
jdom: Please exclude the redundant Orleans from the assembly and package name. And I'd remove Provider from there too, since we are probably going to add the Logger stuff too in the very near future.
Just call it LegacyLogging (and package name Microsoft.Orleans.LegacyLogging)

src/Orleans/Orleans.csproj
xiazen: it is needed for LegacyLogger project, since it  have a dependency for Microsoft.Extensions.Logging and also Orleans. And Orleans need to have dependency on Microsoft.Extensions.Logging anyways, I'd like to make this change as soon as possible.

test/Tester/OrleansLoggingTests.cs
xiazen: no I cannot do that since serviceProvider is not disposable. you want me to revert to original line?

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.csproj
xiazen: at a second thought, I think we should just name it as Logging and Microsoft.Orleans.Logging. Since this assembly includes both legacy feature and non-legacy feature, such as event bulking. 

I added Orleans.Extensions.Logging.Legacy namespace to differentiate legacy feature and non-legacy ones

jdom: But bulking itself we are only including for legacy reasons. This should not be an Orleans concern, but an application concern. If we hadn't had that feature before, we should not be including this here.

src/Orleans/Orleans.csproj
jdom: As long as we don't have the, let's not pollute things here, as this is a high impact change

test/Tester/OrleansLoggingTests.cs
jdom: Yes

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Linq;
+using System.Net;
+
+namespace Orleans.Extensions.Logging.Legacy
+{
+    /// <summary>
+    /// LegacyOrleansLogger supports legacy orleans logging features, including <see cref="ILogConsumer"/>, <see cref="ICloseableLogConsumer">,
+    /// <see cref="IFlushableLogConsumer">, <see cref="Severity">. 
+    /// </summary>
+    [Obsolete("Use Microsoft.Extensions.Logging built-in loggers")]
```
attilah: I think we should rephrase it, because we encourage people to use ```ILogger``` calls to log everything but not to use builtin loggers, since they can use ANY conforming logger.

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
```diff
@@ -0,0 +1,149 @@
...
+        private readonly IPEndPoint ipEndPoint;
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="categoryName"></param>
+        /// <param name="logConsumers"></param>
+        /// <param name="maxSeverityLevel"></param>
+        public LegacyOrleansLogger(string categoryName, IList<ILogConsumer> logConsumers, IPEndPoint ipEndPoint)
+        {
+            this.logConsumers = logConsumers;
```
attilah: We only need an enumerable, but not modifiable version, I'd choose another type instead of ```IList```, same for ```flushableConsumers```.

Specially as we are calling it like this: ```this.LogConsumers.ToArray()```

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        private readonly IList<ILogConsumer> logConsumers;
+        private readonly string name;
+        private readonly IList<IFlushableLogConsumer> flushableConsumers;
+        private readonly LoggerType loggerType;
+        private readonly IPEndPoint ipEndPoint;
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="categoryName"></param>
```
attilah: No content for parameters.

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
```diff
@@ -0,0 +1,149 @@
...
+                throw new ArgumentNullException(nameof(state));
+            }
+            //TODO: support logging scope
+            return NullScope.Instance;
+        }
+
+        /// <inheritdoc/>
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return true;
```
attilah: Why no validating of log levels here?

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
```diff
@@ -0,0 +1,149 @@
...
+        public static Severity LogLevelToSeverity(LogLevel logLevel)
+        {
+            switch (logLevel)
+            {
+                case LogLevel.None: return Severity.Off;
+                case LogLevel.Critical: return Severity.Error;
+                case LogLevel.Error: return Severity.Error;
+                case LogLevel.Warning: return Severity.Warning;
+                case LogLevel.Information: return Severity.Info;
+                case LogLevel.Debug: return Severity.Verbose;
```
attilah: Since we cannot map ```Verbose2``` due to the lack of "enough" levels in the logging abstractions, IMHO we should replace all ```Verbose2``` calls with ```Verbose``` or ```Verbose3```.

For default I'd add at least a comment that it for ```Trace```, or explicitly make that a case (I prefer), and for default throw an exception (since that cannot happen, because we're mapping all values).

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        private LoggerType DetermineLoggerType(string category)
+        {
+            LoggerType type = LoggerType.Application;
+            if (category.Contains("Orleans"))
+            {
+                type = LoggerType.Runtime;
+                if (category.Contains("Provider"))
+                    type = LoggerType.Provider;
+                if (category.Contains("Grain"))
```
attilah: These can overwrite each other, no need for ```else``` ?

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// </summary>
+    [Obsolete("Use Microsoft.Extensions.Logging built-in logger providers")]
+    public class LegacyOrleansLoggerProvider : ILoggerProvider
+    {
+        /// <summary>
+        /// Default Severity for all loggers
+        /// </summary>
+        public static Severity DefaultSeverity = Severity.Info;
+
+        public ConcurrentBag<ILogConsumer> LogConsumers { get; private set; }
```
attilah: Why is it public? I see no external references to it.

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <see cref="ICloseableLogConsumer">, <see cref="IFlushableLogConsumer">, <see cref="Severity">. 
+    /// LegacyOrleansLoggerProvider also supports configuration on those legacy features.
+    /// </summary>
+    [Obsolete("Use Microsoft.Extensions.Logging built-in logger providers")]
+    public class LegacyOrleansLoggerProvider : ILoggerProvider
+    {
+        /// <summary>
+        /// Default Severity for all loggers
+        /// </summary>
+        public static Severity DefaultSeverity = Severity.Info;
```
attilah: Why should we make this public?

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.csproj
attilah: Agree, see my first comment in the review.

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <RootNamespace>Orleans.Extensions.Logging</RootNamespace>
+    <PackageTargetFallback Condition=" '$(BuildFlavor)' == 'Legacy' ">$(PackageTargetFallback);portable-net45+win8+wpa81+wp8</PackageTargetFallback>
```
attilah: I don't think we need ```PackageTargetFallback``` here, since we're not referencing any NuGets that would require it.

src/OrleansLoggerProvider/LegacyOrleansLoggingFactoryExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Add <see cref="LegacyOrleansLoggerProvider"> with event bulking feature to LoggerFactory
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="logConsumers">log consumers which user want to write log events to</param>
+        /// <param name="eventBulkingConfig">config for event bulking feature</param>
+        /// <returns></returns>
+        [Obsolete("Use Microsoft.Extensions.Logging built-in logger providers")]
+        public static ILoggingBuilder AddLegacyOrleansLogging(
+            this ILoggingBuilder builder,
+            List<ILogConsumer> logConsumers,
```
attilah: ```IEnumerable``` would be sufficent here.

src/OrleansLoggerProvider/LegacyOrleansLoggingFactoryExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        /// <summary>
+        /// Add <see cref="LegacyOrleansLoggerProvider"> with event bulking feature to LoggerFactory
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="logConsumers">log consumers which user want to write log events to</param>
+        /// <param name="eventBulkingConfig">config for event bulking feature</param>
+        /// <returns></returns>
+        [Obsolete("Use Microsoft.Extensions.Logging built-in logger providers")]
+        public static ILoggingBuilder AddLegacyOrleansLogging(
```
attilah: This is an ```ILoggingBuilder``` extension class and not ```ILoggingFactory```, please add ```ILoggingFactory``` implementation. SeriLog can be a good sample here perhaps: https://github.com/serilog/serilog-extensions-logging/tree/dev/src/Serilog.Extensions.Logging

src/OrleansLoggerProvider/LegacyOrleansLoggingFactoryExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            this ILoggingBuilder builder,
+            List<ILogConsumer> logConsumers,
+            IPEndPoint ipEndPoint = null,
+            EventBulkingConfig eventBulkingConfig = null)
+        {
+            if (builder == null)
+                throw new ArgumentNullException(nameof(builder));
+            builder.AddMesageBulkingLoggerProvider<LegacyOrleansLoggerProvider>(new LegacyOrleansLoggerProvider(logConsumers, ipEndPoint), eventBulkingConfig);
```
attilah: As a general rule of thumb, if extension method arguments differ, we should provide default values and forward the calls to the one method which does the work. From here it would be easy to not pass in ```severityOverrides``` and call the method below.

src/OrleansLoggerProvider/OrleansLoggingFactoryExtension.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Microsoft.Extensions.Logging;
+using Orleans.Runtime;
+
+namespace Orleans.Extensions.Logging
+{
+    public static class OrleansLoggingFactoryExtension
```
attilah: This is an ```ILoggingBuilder``` extension class and not ```ILoggingFactory```, please add ```ILoggingFactory``` implementation. SeriLog can be a good sample here perhaps: https://github.com/serilog/serilog-extensions-logging/tree/dev/src/Serilog.Extensions.Logging

test/Tester/OrleansLoggingTests.cs
attilah: Can't we pragma disable the warnings for the logging tests instead of putting on ```Obsolete``` on the test methods?

test/Tester/OrleansLoggingTests.cs
```diff
@@ -0,0 +1,104 @@
...
+            //configure logging with severity overrides
+            IServiceCollection serviceCollection = new ServiceCollection();
+            serviceCollection.AddLogging();
+            var severityOverrides = new OrleansLoggerSeverityOverrides();
+            severityOverrides.LoggerSeverityOverrides.Add(this.GetType().FullName, Severity.Warning);
+            serviceCollection.AddLogging(builder => builder.AddLegacyOrleansLogging(new List<ILogConsumer>()
+            {
+                new LegacyFileLogConsumer($"{this.GetType().Name}.log")
+            }, severityOverrides));
+            var serviceProvider = serviceCollection.BuildServiceProvider();
```
attilah: Why is DI involved in these tests? I see no need or value to it.

xiazen:
@dotnet-bot test netfx-bvt please

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
xiazen: it is intended this way. it basically means, if category contains Orleans and Provider than it is LoggerType.provider

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
xiazen: I believe this is not part of this PR's scope? and we are not mapping Verbose2 and Verbose3 to Trace, instead of "cannot map Verbose2"

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.cs
xiazen: so that user know what is the default Severity?

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.cs
xiazen: it is one way user can configure log consumers... there's no external reference because we don't have user code here..

Orleans.sln
```diff
@@ -118,7 +118,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OrleansGCPUtils", "src\Orle
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TestInternalGrainInterfaces", "test\TestInternalGrainInterfaces\TestInternalGrainInterfaces.csproj", "{EBD697E3-91BE-4844-B3F0-6997300A8C12}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Orleans.Core.Abstractions", "src\Orleans.Core.Abstractions\Orleans.Core.Abstractions.csproj", "{73514686-D25D-478B-9943-A86F6B0F3A37}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Orleans.Core.Abstractions", "src\Orleans.Core.Abstractions\Orleans.Core.Abstractions.csproj", "{73514686-D25D-478B-9943-A86F6B0F3A37}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "LegacyOrleansLoggerProvider", "src\OrleansLoggerProvider\LegacyOrleansLoggerProvider.csproj", "{0A5009BF-C7D2-463B-BF39-34E1F5EF2F23}"
```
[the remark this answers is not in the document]
jdom: I agree with `Microsoft.Olreans.Logging.Legacy` for root namespace, assembly name and package id, but for directory and project name I would call it `Logging.Legacy` only... The other is too verbose, as every single project in the solution will have the Microsoft.Olreans prefix.

src/OrleansLoggerProvider/LegacyOrleansLoggingFactoryExtensions.cs
xiazen: ILoggingBuilder is the 2.0 way of configuring logging. Serilog haven't migrated to it IMO because they haven't got to it. 

But take a look at Microsoft.Extensions.Logging.TraceSOurce code, they migrated to ILoggerBuilder. Let me @jdom here since he promoted that we migrated to ILoggerBuilder

src/OrleansLoggerProvider/OrleansLoggingFactoryExtension.cs
xiazen: see my comment above

test/Tester/OrleansLoggingTests.cs
xiazen: the runtime gets logger from DI. I want to emulate that

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
xiazen: What do you think the words need to be? please use ILogger? it is kind of confusing, since LegacyOrleansLogger is an Ilogger. 

please tell me the words you think fit here

Orleans.sln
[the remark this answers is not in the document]
xiazen: @jdom do you agree on the new namespace name?

src/OrleansLoggerProvider/EventBulkingDecoratorLogger.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Threading;
+using Microsoft.Extensions.Logging;
+
+namespace Orleans.Extensions.Logging
+{
+
+    /// <summary>
+    /// Config for event bulking feature
+    /// </summary>
+    public class EventBulkingConfig
```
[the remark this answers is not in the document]
xiazen: sure

src/OrleansLoggerProvider/EventBulkingDecoratorLogger.cs
```diff
@@ -0,0 +1,134 @@
...
+            // Increment recent message counts, if appropriate
+            if (isExcluded)
+            {
+                count = 1;
+                // and don't track counts
+            }
+            else
+            {
+                recentLogMessageCounts.AddOrUpdate(logCode, 1, (key, value) => ++value);
+                recentLogMessageCounts.TryGetValue(logCode, out count);
```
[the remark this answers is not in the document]
xiazen: will do

src/OrleansLoggerProvider/EventBulkingDecoratorLogger.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            var sinceIntervalTicks = now.Ticks - lastBulkLogMessageFlushTicks;
+            //if it is time to flush bulked messages
+            if (sinceIntervalTicks >= this.eventBulkingConfig.BulkEventInterval.Ticks)
+            {
+                // Take local copy of pending bulk message counts, now that this bulk message compaction period has finished
+                var bulkMessageCounts = recentLogMessageCounts.Where(keyPair => keyPair.Value >= this.eventBulkingConfig.BulkEventLimit).ToList();
+                recentLogMessageCounts.Clear();
```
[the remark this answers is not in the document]
xiazen: we are supposed to loosing items event when their value is below BulkEventLImit

src/OrleansLoggerProvider/EventBulkingDecoratorLogger.cs
```diff
@@ -0,0 +1,134 @@
...
+                //set lastBulkLogMessageFlushTicks to now
+                Interlocked.Exchange(ref this.lastBulkLogMessageFlushTicks, now.Ticks);
+          
+                // Output any pending bulk compaction messages
+                if (bulkMessageCounts != null)
```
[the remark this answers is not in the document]
xiazen: will refactor

src/OrleansLoggerProvider/LegacyFileLogConsumer.cs
```diff
@@ -0,0 +1,82 @@
...
+    [Obsolete]
+    public class LegacyFileLogConsumer : ICloseableLogConsumer, IFlushableLogConsumer
+    {
+        private StreamWriter logOutput;
+        private readonly object lockObj = new object();
+        private string logFileName;
+
+        public LegacyFileLogConsumer(string fileName)
+        {
+            logOutput = new StreamWriter(File.Open(fileName, FileMode.Append, FileAccess.Write, FileShare.Write));
```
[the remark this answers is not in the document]
xiazen: why reading? who is gonna read it? what should we do if multiple thread is trying to log ?

src/OrleansLoggerProvider/LegacyFileLogConsumer.cs
```diff
@@ -0,0 +1,82 @@
...
+                    this.logOutput.Flush();
+                    this.logOutput.Dispose();
+                    this.logOutput = null;
+                }
+            }
+            catch (Exception exc)
+            {
+                var msg = string.Format("Ignoring error closing log file {0} - {1}", this.logFileName,
+                    LogFormatter.PrintException(exc));
+                Console.WriteLine(msg);
```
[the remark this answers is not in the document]
xiazen: this is trying to preserve old behavior. But sure , I think we can remove this. not enssential

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
jdom: How about "The Microsoft.Orleans.Logging.Legacy namespace was kept to facilitate migration from Orleans 1.x but will be removed in the near future. It is recommended that you use the Microsoft.Extensions.Logging infrastructure and providers directly instead of Microsoft.Orleans.Logging.Legacy.Logger and Microsoft.Orleans.Logging.Legacy.ILogConsumer."

(I understand you haven't migrated Logger yet, but you can skip that part for now).
Again, I'd put this in a constant, and refer to it from every Obsolete attribute in this namespace.

Orleans.sln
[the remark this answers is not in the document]
jdom: yes

src/OrleansLoggerProvider/EventBulkingDecoratorLogger.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Time limit for bulk event output.
+        /// If the same event code is written more than <c>BulkEventLimit</c> times in the <c>BulkEventInterval</c> time period, 
+        /// then only the first <c>BulkEventLimit</c> individual events will be written, plus a count of how bulk event suppressed.
+        /// </summary>
+        public TimeSpan BulkEventInterval { get; set; } = DefaultBulkEventInterval;
+
+        /// <summary>
+        /// Default bulk event interval
+        /// </summary>
+        public static readonly TimeSpan DefaultBulkEventInterval = TimeSpan.FromMinutes(1);
```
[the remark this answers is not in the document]
xiazen: I think the user will want to know what's the default value , which is why they are public. And we are doing this with our other config classes

src/OrleansLoggerProvider/EventBulkingDecoratorLogger.cs
```diff
@@ -0,0 +1,134 @@
...
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            return this.decoratedLogger.BeginScope<TState>(state);
+        }
+
+        private bool CheckBulkMessageLimits(int logCode, LogLevel sev)
+        {
+            var now = DateTime.UtcNow;
+            int count;
+            bool isExcluded = excludedBulkLogCodes.Contains(logCode)
```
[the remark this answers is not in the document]
xiazen: sure

src/OrleansLoggerProvider/LegacyFileLogConsumer.cs
```diff
@@ -0,0 +1,82 @@
...
+            Severity severity,
+            LoggerType loggerType,
+            string caller,
+            string message,
+            IPEndPoint ipEndPoint,
+            Exception exception,
+            int eventCode = 0
+        )
+        {
+            var logMessage = OrleansLoggingUtils.FormatLogMessageToLegacyStyle(DateTime.UtcNow, severity, loggerType, caller, message, ipEndPoint, exception, eventCode, true);
```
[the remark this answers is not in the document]
xiazen: it is a util class user can also use to achieve legacy Orleans logging style, which is why it is public

src/OrleansLoggerProvider/LegacyFileLogConsumer.cs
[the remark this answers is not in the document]
xiazen: hmmm, ConsoleText is internal class, which I think is only designed to use internally? but this csproj are not supposed to have dependency on Orleans , including its internal class eventually. So I think we can address this in another PR to solve the unavailability in AzureIIRC issue

src/OrleansLoggerProvider/LegacyFileLogConsumer.cs
```diff
@@ -0,0 +1,82 @@
...
+                this.logOutput = null;
+                this.logFileName = null;
+            }
+        }
+
+        public void Flush()
+        {
+            lock (this.lockObj)
+            {
+                if (this.lockObj == null) return;
```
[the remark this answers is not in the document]
xiazen: when one thread closed this consumer , while another thread want to flush it.

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
xiazen: sounds good. except for it is kind of long lol

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
xiazen: no need, it is taken care by log filters

src/OrleansLoggerProvider/LegacyOrleansLoggerProvider.csproj
xiazen: cannot build without this line. may need more work here. can we address this in another PR. and get the bigger part out?

src/OrleansLoggerProvider/LegacyOrleansLoggingFactoryExtensions.cs
xiazen: which general rule of thumb, maybe refer me a link? And I think as a team we should have an agreement on which coding guideline we are following, it provider convenience for the coder to know what style of code he/she should write, instead of discussing it in PRs

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
xiazen: it map LogLevel.Trace to Verbose3 here so that both Verbose2 and Verbose3 will be logged if user set default llogLevel to Trace.

xiazen:
I address feedbacks

Directory.Build.props
```diff
@@ -59,12 +59,13 @@
...
-    <MicrosoftExtensionsConfigurationVersion>1.1.2</MicrosoftExtensionsConfigurationVersion>
-    <MicrosoftExtensionsOptionsVersion>1.1.2</MicrosoftExtensionsOptionsVersion>
-    <MicrosoftExtensionsOptionsConfigurationExtensionsVersion>1.1.2</MicrosoftExtensionsOptionsConfigurationExtensionsVersion>
+    <MicrosoftExtensionsConfigurationBinderVersion>2.0.0</MicrosoftExtensionsConfigurationBinderVersion>
+	<MicrosoftExtensionsLoggingVersion>2.0.0</MicrosoftExtensionsLoggingVersion>
+    <MicrosoftExtensionsConfigurationEnvironmentVariablesVersion>2.0.0</MicrosoftExtensionsConfigurationEnvironmentVariablesVersion>
+    <MicrosoftExtensionsConfigurationJsonVersion>2.0.0</MicrosoftExtensionsConfigurationJsonVersion>
+    <MicrosoftExtensionsConfigurationVersion>2.0.0</MicrosoftExtensionsConfigurationVersion>
+    <MicrosoftExtensionsOptionsVersion>2.0.0</MicrosoftExtensionsOptionsVersion>
+    <MicrosoftExtensionsOptionsConfigurationExtensionsVersion>2.0.0</MicrosoftExtensionsOptionsConfigurationExtensionsVersion>
```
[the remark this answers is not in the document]
xiazen: why is that? can you explain more

src/OrleansLoggerProvider/LegacyOrleansLogger.cs
jdom: Heh, but it's very explicit. BTW, you keep duplicating the string everywhere. Why are you not creating the constant and have the string defined just once?

xiazen: hmmm..... insert texts explaining I know what i'm doing but in fact not.

xiazen: fixed

jdom:
@dotnet-bot test this please

Directory.Build.props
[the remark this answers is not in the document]
attilah: Based on our discussions, leave these in the shared section, but make sure to remove the other 2.0s from the current, since they become shared versions and not overrides.

xiazen:
are we changing namespace  to Orleans.Logging.Legacy, based on our 2.0 discussion? 
I think we can change it to that to be consistent with the direction we are currently going. If we are to change the namespace naming principles in the future, we can change this one as well.

attilah:
I talked with @jdom about this, he suggested this, I'm not a fan of removing parts so save "screen estate", but we can do it if we decide to do so...

Folder: Logging.Legacy
Project filename: Logging.Legacy.csproj
Namespace: Microsoft.Orleans.Logging.Legacy
Assembly name: Microsoft.Orleans.Logging.Legacy
Package name: Microsoft.Orleans.Logging.Legacy

xiazen:
the current PR stick with your suggestion. so we are in an agreement

src/Logging.Legacy/OrleansLoggingUtils.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.Orleans.Logging.Legacy
+{
+    /// <summary>
+    /// Utility method for OrleansLogging
+    /// </summary>
+    public class OrleansLoggingUtils
+    {
+        internal const string ObsoleteMessageString =
+                "The Microsoft.Orleans.Logging.Legacy namespace was kept to facilitate migration from Orleans 1.x but will be removed in the near future. It is recommended that you use the Microsoft.Extensions.Logging infrastructure and providers directly instead of Microsoft.Orleans.Logging.Legacy.Logger and Microsoft.Orleans.Logging.Legacy.ILogConsumer";
```
jdom: nit: it's not `Microsoft.Orleans.Logging.Legacy.Logger` any more, it's `Microsoft.Orleans.Logging.Legacy.OrleansLegacyLogger`.

src/Logging.Legacy/Logging.Legacy.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.Logging.Legacy</PackageId>
+    <Title>Microsoft Orleans Legacy Logging</Title>
+    <Description>Library of Orleans Legacy Logging.</Description>
+    <PackageTags>$(PackageTags) Orleans</PackageTags>
```
jdom: replace with `<PackageTags>$(PackageTags) Logging Legacy</PackageTags>`

src/Logging.Legacy/OrleansLoggingUtils.cs
xiazen: ? Microsoft.Orleans.Logging.Legacy.Logger is the namespace, which is what we want to refer here right?

src/Logging.Legacy/Logging.Legacy.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.Logging.Legacy</PackageId>
+    <Title>Microsoft Orleans Legacy Logging</Title>
+    <Description>Library of Orleans Legacy Logging.</Description>
```
jdom: Please have something more descriptive, as this will go into the nuget package metadata. For example:
This package contains some legacy types used for logging to facilitate migration from Orleans 1.x but will be removed in the near future. It is recommended that you use the Microsoft.Extensions.Logging package and its infrastructure and associated providers directly if not upgrading any legacy log consumer.

src/Logging.Legacy/OrleansLoggingUtils.cs
xiazen: no I think we are referring to our old Orleans.Runtime.Logger abstraction here. and in the future, Orleans.Runtime.Logger will be changing to Microsoft.Orleans.Logging.Legacy.Logger

jdom:
LGTM other than those minor comments

src/Logging.Legacy/OrleansLoggingUtils.cs
jdom: ahh, ok, I remember now :)

PR #3289, opened 2017-08-09
Restructure documentation organization

sergeybykov:
I restructured the documentation TOC and moved some pages around. No change in the content.
You can preview the changes at http://sergeybykov.github.io/orleans/Documentation/Introduction.html.


src/Documentation/Getting-Started-With-Orleans/toc.yml
```diff
@@ -4,10 +4,6 @@
   href: Developing-a-Grain.md
 - name: Developing a Client
   href: Developing-a-Client.md
-- name: Grain Persistence
-  href: Grain-Persistence.md
-- name: Observers
-  href: Observers.md
 - name: Running the Application
   href: Running-the-Application.md
 - name: Silos
```
xiazen: I think the order here can be 
 - Grains
 - Grain lifecycle
 - Developping a grain
- Clients
- Developping a client
- Silos
- Running application
- Debugging

I suggested this to group strongly coupled documents as close to each other as possible. For example, Grain, Developing a grain, Grain lifecycle should be listed next to each other since they are strongly coupled. So user can go from Grain, to its development and its lifecycle. The same go with Client. 

And then after we finished explain concept about grain, silos and client, then we can go to topics about running the application and debugging.

sergeybykov: If you look at #3296, that's kind of how I laid out them, except I got rid of Clients and Silos as separate pages because I think they don't have enough content, and the concept is better explained within the bigger architectural picture of the remaining pages.

jdom:
> - name: Grain Life Cycle

nit: should be Grain *Lifecycle*

---
Refers to: src/Documentation/Getting-Started-With-Orleans/toc.yml:13 in c724776. [](commit_id = c724776ae5a23cb512e38c01af59bba0d7884253, deletion_comment = False)

jdom:
> # Test Framework Integration

Not related to this PR, but this is stale, and there is a section in the step-by-step guide that covers unit testing explicitly. We should probably combine the mocking of services with that.

---
Refers to: src/Documentation/Core-Features/Dependency-Injection.md:114 in c724776. [](commit_id = c724776ae5a23cb512e38c01af59bba0d7884253, deletion_comment = False)

jdom:
Should we include the Step-by-step guide (aka Tutorial) with this new TOC. It always throws me off, as I can't find stuff via the TOC, and I only remember there's a completely separate tutorial section once I use the search feature.

Issue #3285, opened 2017-08-09
InvalidOperationException thrown at regular intervals by OnRefreshClusterMapTimer


Liversage:
I am having some issues where `InvalidOperationException` is thrown in a silo with a message similar to this:

> Trying to send a message Request S127.0.0.1:11111:239965139-> #<span></span>2:  on a silo not from within grain and not from within system target (RuntimeContext is not set to SchedulingContext) RuntimeContext.Current=null TaskScheduler.Current=System.Threading.Tasks.ThreadPoolTaskScheduler

The exception is handled so perhaps I should just ignore it but I would still like to understand why this happens.

Using a debug version of Orleans (v1.5) I can inspect `debugContext` in [`InsideRuntimeClient.SendRequestMessage()`](https://github.com/dotnet/orleans/blob/v1.5.0/src/OrleansRuntime/Core/InsideRuntimeClient.cs#L135) and it has the following value:

    global::Orleans.Runtime.Versions.IVersionStoreGrain:GetCompatibilityStrategy()

I then set a breakpoint at `Orleans.Runtime.Versions.GrainVersionStore.GetCompatibilityStrategy()` and when the debugger breaks I see the following stack trace:

    OrleansRuntime.dll!Orleans.Runtime.Versions.GrainVersionStore.GetCompatibilityStrategy()
    OrleansRuntime.dll!Orleans.Runtime.TypeManager.GetAndSetDefaultCompatibilityStrategy()
    OrleansRuntime.dll!Orleans.Runtime.TypeManager.OnRefreshClusterMapTimer(object _)
    Orleans.dll!Orleans.Runtime.SafeTimerBase.HandleAsyncTaskTimerCallback(object state)
    Orleans.dll!Orleans.Runtime.SafeTimerBase.HandleTimerCallback(object state)
    mscorlib.dll!System.Threading.TimerQueueTimer.CallCallbackInContext(object state)

So apparently a timer fires to call [`TypeManager.OnRefreshClusterMapTimer()`](https://github.com/dotnet/orleans/blob/v1.5.0/src/OrleansRuntime/GrainTypeManager/TypeManager.cs#L141) which eventually tries to call [`IVersionStoreGrain:GetCompatibilityStrategy()`](https://github.com/dotnet/orleans/blob/v1.5.0/src/OrleansRuntime/Versions/GrainVersionStore.cs#L75) from the thread pool thread used by the timer leading to the exception being thrown by `InsideRuntimeClient`. At least that is my analysis.

[...]

PR #3290, opened 2017-08-09
Fix scheduling of notification interfaces

### LINKED ISSUE #3273: Implicit stream subscription dies on network partition
Hi,

We're experiencing a problem during a network partition.
Some (not all) implicit subscribed grains stop processing new messages. 

We're seeing this exception in the log, not sure if it's related..

> Trying to send a message Request S10.0.0.4:20015:239358855-> #196448: on a silo not from within grain and not from within system target (RuntimeContext is not set to SchedulingContext) RuntimeContext.Current=null TaskScheduler.Current=Orleans.Runtime.Scheduler.OrleansTaskScheduler

```
System.InvalidOperationException:
   at Orleans.Runtime.InsideRuntimeClient.SendRequestMessage (OrleansRuntime, Version=1.5.0.0, Culture=neutral, PublicKeyToken=null)
   at Orleans.Runtime.GrainReference.InvokeMethod_Impl (Orleans, Version=1.5.0.0, Culture=neutral, PublicKeyToken=null)
   at Orleans.Runtime.GrainReference.InvokeMethodAsync (Orleans, Version=1.5.0.0, Culture=neutral, PublicKeyToken=null)
   at Orleans.Streams.OrleansCodeGenStreamQueueBalanceListenerReference.QueueDistributionChangeNotification (Orleans, Version=1.5.0.0, Culture=neutral, PublicKeyToken=null)
   at Orleans.Streams.ConsistentRingQueueBalancer.RangeChangeNotification (OrleansRuntime, Version=1.5.0.0, Culture=neutral, PublicKeyToken=null)
   at Orleans.Runtime.ConsistentRing.EquallyDividedRangeRingProvider.RangeChangeNotification (OrleansRuntime, Version=1.5.0.0, Culture=neutral, PublicKeyToken=null)
```
When we restart the silo with the halted subscription, the queued messages are processed.

We're using orleans 1.5.1 on ServiceFabric and the Azure Queue stream provider

ReubenBond:
Fixes #3273

Implementations of `ISiloStatusListener` and `IRingRangeListener` must correctly marshall calls onto their own execution context in order to preserve correct scheduling semantics.

* Add `.ScheduleTask(...)` extension methods to `SystemTarget` so that implementations can easily schedule these callbacks.
* ~~Add `IAddressableContextScheduler` which provides methods for executing actions on an addressable scheduling context (so that grain calls can be made). It can be injected via `Factory<IAddressableContextScheduler>`.~~
* Remove `IAddressable` from `IStreamQueueBalanceListener` and have `PersistentStreamPullingManager` subscribe to ring range change notifications directly (instead of via a ref, `.AsReference<IStreamQueueBalanceListener>()`)
* Verified locally using Service Fabric Sample (modified to include streams)

### DIFF
--- a/src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
+++ b/src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
@@ -53,9 +53,8 @@ namespace Orleans.Streams
     /// indicating that the balance of queues has changed.
     /// It should be implemented by components interested in stream queue load balancing.
     /// When change notification is received, listener should request updated list of queues from the queue balancer.
-    /// This interface inherit from IAddressable for threading-safe concern
     /// </summary>
-    public interface IStreamQueueBalanceListener :IAddressable
+    public interface IStreamQueueBalanceListener
     {
         /// <summary>
         /// Receive notifications about adapter queue responsibility changes. 
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -27,7 +27,7 @@ using Orleans.Versions.Compatibility;
 
 namespace Orleans.Runtime
 {
-    internal class Catalog : SystemTarget, ICatalog, IPlacementRuntime, ISiloStatusListener
+    internal class Catalog : SystemTarget, ICatalog, IPlacementRuntime
     {
         [Serializable]
         internal class NonExistentActivationException : Exception
@@ -169,6 +169,7 @@ namespace Orleans.Runtime
             });
             maxWarningRequestProcessingTime = this.config.ResponseTimeout.Multiply(5);
             maxRequestProcessingTime = this.config.MaxRequestProcessingTime;
+            grainDirectory.SetSiloRemovedCatalogCallback(this.OnSiloStatusChange);
         }
 
         /// <summary>
@@ -1606,17 +1607,15 @@ namespace Orleans.Runtime
 
 #endregion
 
-#region Implementation of ISiloStatusListener
-
-        public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
-        {
+        private void OnSiloStatusChange(SiloAddress updatedSilo, SiloStatus status)
+        { 
             // ignore joining events and also events on myself.
             if (updatedSilo.Equals(LocalSilo)) return;
 
             // We deactivate those activations when silo goes either of ShuttingDown/Stopping/Dead states,
             // since this is what Directory is doing as well. Directory removes a silo based on all those 3 statuses,
             // thus it will only deliver a "remove" notification for a given silo once to us. Therefore, we need to react the fist time we are notified.
-            // We may review the directory behaiviour in the future and treat ShuttingDown differently ("drain only") and then this code will have to change a well.
+            // We may review the directory behavior in the future and treat ShuttingDown differently ("drain only") and then this code will have to change a well.
             if (!status.IsTerminating()) return;
             if (status == SiloStatus.Dead)
             {
@@ -1639,7 +1638,7 @@ namespace Orleans.Runtime
 
                             lock (activationData)
                             {
-                                // adapted from InsideGarinClient.DeactivateOnIdle().
+                                // adapted from InsideGrainClient.DeactivateOnIdle().
                                 activationData.ResetKeepAliveRequest();
                                 activationsToShutdown.Add(activationData);
                             }
@@ -1647,12 +1646,12 @@ namespace Orleans.Runtime
                         catch (Exception exc)
                         {
                             logger.Error(ErrorCode.Catalog_SiloStatusChangeNotification_Exception,
-                                String.Format("Catalog has thrown an exception while executing SiloStatusChangeNotification of silo {0}.", updatedSilo.ToStringWithHashCode()), exc);
+                                String.Format("Catalog has thrown an exception while executing OnSiloStatusChange of silo {0}.", updatedSilo.ToStringWithHashCode()), exc);
                         }
                     }
                 }
                 logger.Info(ErrorCode.Catalog_SiloStatusChangeNotification,
-                    String.Format("Catalog is deactivating {0} activations due to a failure of silo {1}, since it is a primary directory partiton to these grain ids.",
+                    String.Format("Catalog is deactivating {0} activations due to a failure of silo {1}, since it is a primary directory partition to these grain ids.",
                         activationsToShutdown.Count, updatedSilo.ToStringWithHashCode()));
             }
             finally
@@ -1664,7 +1663,5 @@ namespace Orleans.Runtime
                 }
             }
         }
-
-#endregion
     }
 }
--- a/src/OrleansRuntime/Placement/DeploymentLoadPublisher.cs
+++ b/src/OrleansRuntime/Placement/DeploymentLoadPublisher.cs
@@ -184,6 +184,11 @@ namespace Orleans.Runtime
 
         public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
         {
+            this.ScheduleTask(() => { Utils.SafeExecute(() => this.OnSiloStatusChange(updatedSilo, status), this.logger); }).Ignore();
+        }
+
+        private void OnSiloStatusChange(SiloAddress updatedSilo, SiloStatus status)
+        {
             if (!status.IsTerminating()) return;
 
             if (Equals(updatedSilo, this.Silo))
--- a/src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingManager.cs
+++ b/src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingManager.cs
@@ -89,8 +89,7 @@ namespace Orleans.Streams
             // Remove cast once we cleanup
             queueAdapter = qAdapter.Value;
             await this.queueBalancer.Initialize(this.streamProviderName, this.adapterFactory.GetStreamQueueMapper(), config.SiloMaturityPeriod, this.providerConfig);
-            var meAsQueueBalanceListener = this.AsReference<IStreamQueueBalanceListener>();
-            queueBalancer.SubscribeToQueueDistributionChangeEvents(meAsQueueBalanceListener);
+            queueBalancer.SubscribeToQueueDistributionChangeEvents(this);
 
             List<QueueId> myQueues = queueBalancer.GetMyQueues().ToList();
             Log(ErrorCode.PersistentStreamPullingManager_03, String.Format("Initialize: I am now responsible for {0} queues: {1}.", myQueues.Count, PrintQueues(myQueues)));
@@ -140,6 +139,11 @@ namespace Orleans.Streams
         /// and don't execute an older notification if a newer one was already delivered.
         /// </summary>
         public Task QueueDistributionChangeNotification()
+        {
+            return this.ScheduleTask(() => this.HandleQueueDistributionChangeNotification());
+        }
+
+        public Task HandleQueueDistributionChangeNotification()
         {
             latestRingNotificationSequenceNumber++;
             int notificationSeqNumber = latestRingNotificationSequenceNumber;
--- a/src/OrleansRuntime/Core/SystemTargetExtensions.cs
+++ b/src/OrleansRuntime/Core/SystemTargetExtensions.cs
@@ -1,0 +1,34 @@
+using System;
+using System.Threading.Tasks;
+using Orleans.Runtime.Scheduler;
+
+namespace Orleans.Runtime
+{
+    /// <summary>
+    /// Extensions for <see cref="SystemTarget"/>.
+    /// </summary>
+    public static class SystemTargetExtensions
+    {
+        /// <summary>
+        /// Schedules the provided <paramref name="action"/> on the <see cref="SystemTarget"/>'s <see cref="ISchedulingContext"/>.
+        /// </summary>
+        /// <param name="self">The <see cref="SystemTarget"/>.</param>
+        /// <param name="action">The action.</param>
+        /// <returns>A <see cref="Task"/> which completes when the <paramref name="action"/> has completed.</returns>
+        public static Task ScheduleTask(this SystemTarget self, Func<Task> action)
+        {
+            return self.RuntimeClient.Scheduler.RunOrQueueTask(action, self.SchedulingContext);
+        }
+
+        /// <summary>
+        /// Schedules the provided <paramref name="action"/> on the <see cref="SystemTarget"/>'s <see cref="ISchedulingContext"/>.
+        /// </summary>
+        /// <param name="self">The <see cref="SystemTarget"/>.</param>
+        /// <param name="action">The action.</param>
+        /// <returns>A <see cref="Task"/> which completes when the <paramref name="action"/> has completed.</returns>
+        public static Task ScheduleTask(this SystemTarget self, Action action)
+        {
+            return self.RuntimeClient.Scheduler.RunOrQueueAction(action, self.SchedulingContext);
+        }
+    }
+}
--- a/src/OrleansRuntime/GrainDirectory/ILocalGrainDirectory.cs
+++ b/src/OrleansRuntime/GrainDirectory/ILocalGrainDirectory.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Orleans.GrainDirectory;
@@ -122,5 +123,11 @@ namespace Orleans.Runtime.GrainDirectory
         /// The id of this cluster
         /// </summary>
         string ClusterId { get; }
+
+        /// <summary>
+        /// Sets the callback to <see cref="Catalog"/> which is called when a silo is removed from membership.
+        /// </summary>
+        /// <param name="catalogOnSiloRemoved">The callback.</param>
+        void SetSiloRemovedCatalogCallback(Action<SiloAddress, SiloStatus> catalogOnSiloRemoved);
     }
 }
--- a/src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
+++ b/src/OrleansRuntime/GrainDirectory/LocalGrainDirectory.cs
@@ -29,6 +29,7 @@ namespace Orleans.Runtime.GrainDirectory
         private readonly ISiloStatusOracle siloStatusOracle;
         private readonly IMultiClusterOracle multiClusterOracle;
         private readonly IInternalGrainFactory grainFactory;
+        private Action<SiloAddress, SiloStatus> catalogOnSiloRemoved;
 
         // Consider: move these constants into an apropriate place
         internal const int HOP_LIMIT = 3; // forward a remote request no more than two times
@@ -57,8 +58,6 @@ namespace Orleans.Runtime.GrainDirectory
         internal GrainDirectoryHandoffManager HandoffManager { get; private set; }
 
         public string ClusterId { get; }
-
-        internal ISiloStatusListener CatalogSiloStatusListener { get; set; }
 
         internal GlobalSingleInstanceActivationMaintainer GsiActivationMaintainer { get; private set; }
 
@@ -261,6 +260,16 @@ namespace Orleans.Runtime.GrainDirectory
             stopPreparationResolver.TrySetException(ex);
         }
 
+        /// <inheritdoc />
+        public void SetSiloRemovedCatalogCallback(Action<SiloAddress, SiloStatus> callback)
+        {
+            if (callback == null) throw new ArgumentNullException(nameof(callback));
+            lock (membershipCache)
+            {
+                this.catalogOnSiloRemoved = callback;
+            }
+        }
+
         #region Handling membership events
         protected void AddServer(SiloAddress silo)
         {
@@ -299,12 +308,12 @@ namespace Orleans.Runtime.GrainDirectory
                     return;
                 }
 
-                if (CatalogSiloStatusListener != null)
+                if (this.catalogOnSiloRemoved != null)
                 {
                     try
                     {
-                        // only call SiloStatusChangeNotification once on the catalog and the order is important: call BEFORE updating membershipRingList.
-                        CatalogSiloStatusListener.SiloStatusChangeNotification(silo, status);
+                        // Only notify the catalog once. Order is important: call BEFORE updating membershipRingList.
+                        this.catalogOnSiloRemoved(silo, status);
                     }
                     catch (Exception exc)
                     {
--- a/src/OrleansRuntime/MultiClusterNetwork/MultiClusterOracle.cs
+++ b/src/OrleansRuntime/MultiClusterNetwork/MultiClusterOracle.cs
@@ -114,7 +114,7 @@ namespace Orleans.Runtime.MultiClusterNetwork
         public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
         {
             // any status change can cause changes in gateway list
-            PublishChanges();
+            this.ScheduleTask(() => Utils.SafeExecute(() => this.PublishChanges())).Ignore();
         }
 
         public bool SubscribeToMultiClusterConfigurationEvents(GrainReference observer)
--- a/src/OrleansRuntime/Silo/Silo.cs
+++ b/src/OrleansRuntime/Silo/Silo.cs
@@ -136,7 +136,11 @@ namespace Orleans.Runtime
         /// Test hook connection for white-box testing of silo.
         /// </summary>
         internal TestHooksSystemTarget testHook;
-        
+
+        private SchedulingContext membershipOracleContext;
+        private SchedulingContext multiClusterOracleContext;
+        private SchedulingContext reminderServiceContext;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Silo"/> class.
         /// </summary>
@@ -490,11 +494,12 @@ namespace Orleans.Runtime
             // Load and init grain services before silo becomes active.
             CreateGrainServices(GlobalConfig.GrainServiceConfigurations);
 
-            ISchedulingContext statusOracleContext = (this.membershipOracle as SystemTarget)?.SchedulingContext;
-            scheduler.QueueTask(() => this.membershipOracle.Start(), statusOracleContext)
+            this.membershipOracleContext = (this.membershipOracle as SystemTarget)?.SchedulingContext ??
+                                       this.providerManagerSystemTarget.SchedulingContext;
+            scheduler.QueueTask(() => this.membershipOracle.Start(), this.membershipOracleContext)
                 .WaitWithThrow(initTimeout);
             if (logger.IsVerbose) { logger.Verbose("Local silo status oracle created successfully."); }
-            scheduler.QueueTask(this.membershipOracle.BecomeActive, statusOracleContext)
+            scheduler.QueueTask(this.membershipOracle.BecomeActive, this.membershipOracleContext)
                 .WaitWithThrow(initTimeout);
             if (logger.IsVerbose) { logger.Verbose("Local silo status oracle became active successfully."); }
             scheduler.QueueTask(() => this.typeManager.Initialize(versionStore), this.typeManager.SchedulingContext)
@@ -506,8 +511,9 @@ namespace Orleans.Runtime
                 logger.Info("Starting multicluster oracle with my ServiceId={0} and ClusterId={1}.",
                     GlobalConfig.ServiceId, GlobalConfig.ClusterId);
 
-                ISchedulingContext clusterStatusContext = (multiClusterOracle as SystemTarget)?.SchedulingContext;
-                scheduler.QueueTask(() => multiClusterOracle.Start(), clusterStatusContext)
+                this.multiClusterOracleContext = (multiClusterOracle as SystemTarget)?.SchedulingContext ??
+                                                          this.providerManagerSystemTarget.SchedulingContext;
+                scheduler.QueueTask(() => multiClusterOracle.Start(), multiClusterOracleContext)
                                     .WaitWithThrow(initTimeout);
                 if (logger.IsVerbose) { logger.Verbose("multicluster oracle created successfully."); }
             }
@@ -531,7 +537,9 @@ namespace Orleans.Runtime
                 if (this.reminderService != null)
                 {
                     // so, we have the view of the membership in the consistentRingProvider. We can start the reminder service
-                    this.scheduler.QueueTask(this.reminderService.Start, (this.reminderService as SystemTarget)?.SchedulingContext)
+                    this.reminderServiceContext = (this.reminderService as SystemTarget)?.SchedulingContext ??
+                                                           this.providerManagerSystemTarget.SchedulingContext;
+                    this.scheduler.QueueTask(this.reminderService.Start, this.reminderServiceContext)
                         .WaitWithThrow(this.initTimeout);
                     if (this.logger.IsVerbose)
                     {
@@ -725,14 +733,14 @@ namespace Orleans.Runtime
                     {
                         logger.Info(ErrorCode.SiloShuttingDown, "Silo starting to Shutdown()");
                         // 1: Write "ShutDown" state in the table + broadcast gossip msgs to re-read the table to everyone
-                        scheduler.QueueTask(this.membershipOracle.ShutDown, (this.membershipOracle as SystemTarget)?.SchedulingContext)
+                        scheduler.QueueTask(this.membershipOracle.ShutDown, this.membershipOracleContext)
                             .WaitWithThrow(stopTimeout);
                     }
                     else
                     {
                         logger.Info(ErrorCode.SiloStopping, "Silo starting to Stop()");
                         // 1: Write "Stopping" state in the table + broadcast gossip msgs to re-read the table to everyone
-                        scheduler.QueueTask(this.membershipOracle.Stop, (this.membershipOracle as SystemTarget)?.SchedulingContext)
+                        scheduler.QueueTask(this.membershipOracle.Stop, this.membershipOracleContext)
                             .WaitWithThrow(stopTimeout);
                     }
                 }
@@ -745,7 +753,7 @@ namespace Orleans.Runtime
                 if (reminderService != null)
                 {
                     // 2: Stop reminder service
-                    scheduler.QueueTask(reminderService.Stop, (reminderService as SystemTarget)?.SchedulingContext)
+                    scheduler.QueueTask(reminderService.Stop, this.reminderServiceContext)
                         .WaitWithThrow(stopTimeout);
                 }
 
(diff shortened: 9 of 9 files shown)


ReubenBond:
I rather try to implement this without `IAddressableContextScheduler`, since I dislike it. I think that's achievable by removing `IAddressable` from `IStreamQueueBalanceListener` and therefore not casting `PersistentStreamPullingManager` when it subscribes to ring change notifications.

ReubenBond:
@xiazen @jason-bragg please look at the updated PR. I removed the abstraction I had previously introduced, but more importantly: I changed `IStreamQueueBalanceListener` as mentioned in the above comment.

jason-bragg:
Looks technically correct, though I'd prefer we move to the shared state pattern (Shared state publisher prototype #3291), as the pattern in this PR relies on the implementer of a notification listener to 'do the right thing' whereas the shared state pattern prevents them from doing the wrong thing.

As an immediate fix, this looks fine.  As a longer term maintainable approach, more discussion may be needed.

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
```diff
@@ -53,9 +53,8 @@ public interface IStreamQueueBalancer
     /// indicating that the balance of queues has changed.
     /// It should be implemented by components interested in stream queue load balancing.
     /// When change notification is received, listener should request updated list of queues from the queue balancer.
-    /// This interface inherit from IAddressable for threading-safe concern
     /// </summary>
-    public interface IStreamQueueBalanceListener :IAddressable
+    public interface IStreamQueueBalanceListener
```
jason-bragg: Happy to remove IAddressable from the interface.  Never really liked it.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -1576,6 +1576,11 @@ private List<ActivationData> TryGetActivationDatas(List<ActivationAddress> addre
 
         public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
         {
+            this.ScheduleTask(() => this.OnSiloStatusChange(updatedSilo, status));
```
jason-bragg: This relies on implementers to know this is necessary.  Not obvious to maintainers or those writing new listeners.

src/OrleansRuntime/Placement/DeploymentLoadPublisher.cs
```diff
@@ -183,9 +183,13 @@ public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus sta
             if (Equals(updatedSilo, this.Silo))
                 this.publishTimer.Dispose();
 
-            SiloRuntimeStatistics ignore;
-            periodicStats.TryRemove(updatedSilo, out ignore);
-            NotifyAllStatisticsChangeEventsSubscribers(updatedSilo, null);
+            this.ScheduleTask(
```
jason-bragg: While I understand why the above code is outside the task, it would be more maintainable if all the logic was in another function which this call scheduled, like in other instances in this PR.  Especially the timer dispose, as that is state and the dispose call may not be thread safe.  Also having the dispose outside the task does not save us much since if the code has gotten that far, the task will still be created, unlike the earlier fast exit.

src/OrleansRuntime/Catalog/Catalog.cs
ReubenBond: Agreed, it's not great. Sergey and I settled on this as a 'good enough' fix for a patch release, but it's definitely not ideal.

sergeybykov: Task returned by `this.ScheduleTask()` is not awaited or `Ignore`'ed. I think it's best to wrap it in a `SafeExecute()`, so that potential exceptions are handled and logged. Alternatively, we could add another extension that would do that and return void instead of a `Task`.

xiazen:
LGTM

gabikliot:
Please wait merging, I might have comments.

src/OrleansRuntime/Placement/DeploymentLoadPublisher.cs
ReubenBond: Sounds good to me

src/OrleansRuntime/Catalog/Catalog.cs
ReubenBond: Good point, @sergeybykov. I updated to use `Ignore()` and `SafeExecute()`.

ReubenBond:
Sure thing, @gabikliot

gabikliot:
I am sorry, but I am not following.

I don't understand the problem in #3273. That bug refers to another bug #3256.
All are about scheduling context, I see that, but how they are all related?
How #3273 happens only during network partitions? How we escape the context during partitions only?
How TypeManager timers in #3256 are related?
And how that PR addresses the other ones?

The main concern I have is about ordering of silo status change notifications. There was a reason to make them on the caller context - to guarantee ordering. This PR changes multi silo, which I don't know much about, but also queue balancer.

sergeybykov:
We believe there are several issues here. No doubt it's confusing.

The root cause of #3285 we believe is the bug of using a regular, albeit safe, timer instead of registering a grain (system target) timer - https://github.com/dotnet/orleans/pull/3256/commits/98e39103711a59bd2f555352b7b98631dadfab1c#diff-9bd4c404a42cdafba3be5d76bf5e71fdR64.

Because the timer fires on a thread pool thread, that obviously violates the single-threading guarantee and makes the callback run with no scheduling context.

We believe we see the same thing happening with `FabricMembershipOracle` that, unlike `MembershipOracle`, is not a system target. SF invokes it with notifications on its own thread, and that causes the same issue of multi-threading and no scheduling context. I don't remember if it's #3273 or another one where SF is used for hosting.

The next level of realization was that even if the caller runs as a system target, e.g. `MembershipOracle`, with a scheduling context and all, in some places we make direct calls to other system targets, which violates their single-threading guarantee, and makes their grain calls to be made in a wrong (callers) context.

#3256 is attempting to patch the immediate issues without a significant refactoring of the code, so that the limited fix could be included in 1.5.1. At the same time, @ReubenBond is entertaining a more general solution to help ensure calls to and between system targets are enqueued on correct scheduling contexts. That change would be part of 2.0 release.

gabikliot:
Ok, that is more or less what I thought.
The 1st and 2nd issues - timer and SF - sure, fix them as you said. Those are just bugs. I would not call it "patch the immediate issues", but rather: fix bugs.

But re:
>in some places we make direct calls to other system targets, which violates their single-threading guarantee, and makes their grain calls to be made in a wrong (callers) context.

That is a bit more complicated. In some places, this was done by design.
So lets not jump into too quick realizations.

Fix the bugs, then lets open a new and clear issue about why MBR calls other system targets on its own context and not their context and lets discuss it there. We may decide at the end that this is wrong, or that maybe it is not. There was thinking and reasoning behind that.

sergeybykov:
>Fix the bugs, then lets open a new and clear issue about why MBR calls other system targets on its own context and not their context and lets discuss it there.

That's roughly the plan. Although, @ReubenBond is looking beyond membership, for a general safe invocation option (a proxy or something). Then we can see where it makes sense to use it and where we are fine violating the safety for well understood and documented reasons.

gabikliot:
Sounds good. The general safe invocation option is to use system target grain interface. We do it in most cases already. No need for any new mechanism.

ReubenBond:
I've updated the PR to address the comments. Thanks, @gabikliot. Having the `Catalog` implement `ISiloStatusListener` was confusign since it never subscribed to the `ISiloStatusOracle`, but was instead directly called via `LocalGrainDirectory`. I reworked the flow used to register the `Catalog` callback with the `LocalGrainDirectory` so that it registers itself in its constructor instead of it being [registered via a property setter in `Silo.InjectDependencies()`](https://github.com/dotnet/orleans/blob/e858a66de020f178dd5f7db6bf2fa72c9e2a5a83/src/OrleansRuntime/Silo/Silo.cs#L478).

I wonder if we could extract this 'dead silo activation cleanup' code into a separate class, since Catalog is so huge and the method only needs to call `Catalog.DeactivateActivation` once, outside of the lock on `ActivationDirectory`. `ActivationDirectory` itself is threadsafe for enumeration and is an injected singleton already shared between multiple classes.

gabikliot:
@ReubenBond, can we stick to the plan above?
1) fix all bugs in the timer in type manager
2) fix all bugs in Service Fabric integration
3) open a new issue and describe what is the problem we are founding with system targets calling to system targets directly, where and why is this a problem.
4) only after we discuss 3 lets implement (or not) what is decided.
It looks like you guys confusing 1+2 with 3 (at least this PR points to 1 as the problem it is trying to solve), which is totally unrelated and jumping into 4.
Its possible that you already discussed it all internally and for you it is not "jumping", but it is for me.

I thought that is already what Sergey wrote and we are in agreement that 1+2 are unrelated to 3 and we should first discuss 3 before jumping into 4.

ReubenBond:
What does that mean in concrete terms, @gabikliot? What I'm saying I've done in the most recent update is partially undo the original change, but remove the `ISiloStatusListener` from `Catalog` and (for cleanliness) change who is responsible for subscribing `Catalog` to `LocalGrainDirectory`. This is just to avoid future confusion.

The issue with `TypeManager` is addressed in PR #3256. They're the same core issue (scheduling in SystemTargets / system-level components), but there isn't any crossover between the fixes (there doesn't need to be).

What's the bug in Service Fabric integration?

ReubenBond:
It's not obvious that there should be ordering in `SiloStatusNotificationCalls` calls or that those calls must ensure that they handle the notification entirely before returning to the caller. With this subtle change, those subtle requirements are gone.

ReubenBond:

gabikliot:
I am not understanding what problem this PR addresses.

sergeybykov:

ReubenBond:
@gabikliot check out the stack trace in #3273, nevermind the text of that issue. This PR prevents that exception from occurring by ensuring that notifications which may result in a grain call are scheduled on an addressable runtime context (a SystemTarget's context).

Initially I wanted to use a SystemThread to fix this, but a SystemThread cannot make grain calls in its current state.

gabikliot:
Sorry, Reuben. If you are looking for my help, you will have to elaborate more.
I simply don't see how this is related or how your fix is doing the right thing to queue balancer. It used to make grain call, now it calls directly and then queues. Looks like the wrong pattern to me.

ReubenBond:
No worries, @gabikliot. I'll try to elaborate:
* Previously, the `IStreamQueueBalanceListener` was `IAddressable` for the purposes of having calls scheduled on its context - for single-threadedness - it was never a remote call. This is a divergence from the pattern already in use in the codebase when locally calling SystemTargets. That pattern is making a direct call, scheduled using `OrleansTaskScheduler`, on the SystemTarget's `ISchedulingContext`. Making this call required that the call was made from an addressable context already - the fabric oracle does not have an addressable context.
* `PersistentStreamPullingManager` makes remote calls to agents when the cluster changes - so notifications must be executed on an addressable context.
* The change makes the existing pattern (scheduling calls on the SystemTarget's context) simpler to use by adding an extension method to SystemTarget called `ScheduleTask`.
* The change implements the existing patterns in places where it was previously not being used << this is the bug fix.

The decision - for the time being - is that *consumers* of notifications are responsible for ensuring that they handle the notification on the correct scheduling context. Producers are merely responsible for notifying the consumer - with no guarantee of what scheduling context that notification comes from.

Is that clearer?

ReubenBond:

ReubenBond:
The conflict here is easy to resolve using the Resolve Conflicts button, so I won't push an update unless we want code changes.

ReubenBond:
Updated PR to reduce the impact of the changes - nothing superfluous is altered now.

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingManager.cs
```diff
@@ -89,8 +89,7 @@ internal class PersistentStreamPullingManager : SystemTarget, IPersistentStreamP
             // Remove cast once we cleanup
             queueAdapter = qAdapter.Value;
             await this.queueBalancer.Initialize(this.streamProviderName, this.adapterFactory.GetStreamQueueMapper(), config.SiloMaturityPeriod, this.providerConfig);
-            var meAsQueueBalanceListener = this.AsReference<IStreamQueueBalanceListener>();
-            queueBalancer.SubscribeToQueueDistributionChangeEvents(meAsQueueBalanceListener);
+            queueBalancer.SubscribeToQueueDistributionChangeEvents(this);
```
gabikliot: Why changing this? It used to be that the notification here were delivered as grain calls on this system target, which is the right way for system targets to communicate. There are some other examples when system targets call functions directly (breaking isolation), but this was not the case here. In this case it was done the right way.
So why doing it the wrong way now - call directly and then manually switch context?

src/OrleansRuntime/Placement/DeploymentLoadPublisher.cs
```diff
@@ -178,6 +178,11 @@ private void NotifyAllStatisticsChangeEventsSubscribers(SiloAddress silo, SiloRu
 
         public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
         {
+            this.ScheduleTask(() => { Utils.SafeExecute(() => this.OnSiloStatusChange(updatedSilo, status), this.logger); }).Ignore();
```
gabikliot: this is the other example: when we were actually breaking isolation (as opposite to below, where we were not breaking isolation).
I get it that in theory it was wrong to break isolation. But before I go and try to dig up why we did it this way, I want to ask:
why do you want to change it now? what did not work correctly? What was actually wrong with that?
The data structures here are concurrent and handled correctly I believe, so I don't yet see any reason to change it here, unless you can show me one.

src/Orleans/Streams/PersistentStreams/IStreamQueueBalancer.cs
gabikliot: > "This interface inherit from IAddressable for threading-safe concern"

It inherits from IAddressable so it can be implemented by system target and notification be called as grain call. The right and proper way to talk between components inside Orleans is via grain (system target) calls. Except for some small number of limited cases when we break isolation. But here is not such a case and thus I don't see a reason to avoid using grain calls for cross component communication and insist on breaking isolation (I call "invoking function directly, even if after that we enqueue an Task directly" as breaking isolation).

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingManager.cs
ReubenBond: How do you feel this general problem should be addressed?
* OS Thread notifies some Orleans component of some event
* That component propagates that notification to subscribers

Should the component marshal the notification onto an addressable context?
Alternatively, should the subscriber be responsible for ensuring that they execute on the correct context?

I am in favor of the latter approach, since the first is difficult without some publicly exposed mechanism for doing so (eg, SystemTarget - which the SF code does not currently have access to) and also because it's the subscriber which knows what its requirements are - not the publisher.

src/OrleansRuntime/Placement/DeploymentLoadPublisher.cs
ReubenBond: Is this breaking isolation? I'm not sure that SystemTargets ever had real isolation within a silo - isn't it more rare that we *don't* call using a mechanism similar to this one here? It seems that the vast majority of the time, calls to a system target follow this pattern:
```C#
scheduler.QueueTask(() => sysTarget.CallMethod(), sysTarget.SchedulingContext).WaitBlah();
```
Where `scheduler` = `OrleansTaskScheduler` and `sysTarget` is a direct .NET reference to the system target instance (not a proxy `GrainReference`).

Regardless, the reason I took this approach is what I mentioned in this comment above: https://github.com/dotnet/orleans/pull/3290#discussion_r133638582

ReubenBond: I do think that calling via some proxy is a better solution. The only problem is that `GrainReference` can only be used from an addressable context today (another `Grain` or `SystemTarget`) when running within the silo.

Maybe that restriction can be relaxed and we can convert all of the code to make proxied calls. The problem is that it's not relaxed today and I feel like that's a larger change than we could stomach in a patch release. That kind of thing requires more thought - I'm sure the exception which caused this issue was not added lightheartedly, it's likely there to make sure scheduling contexts are being used correctly.

ReubenBond:
@dotnet-bot test netfx-bvt

ReubenBond:
@dotnet-bot test netfx-bvt

ReubenBond:
@dotnet-bot test this please

sergeybykov:
@ReubenBond Please confirm if my understanding of this PR is correct, as it has evolved from the original, more ambitious version.

We are fixing two relatively narrow issues here, so that we can include this in 1.5.1:

1. Marshalling of cluster change notifications coming from the SF version of membership oracle that arrive on non-Orleans threads.
2. Marshalling of queue balancer notifications that may also arrive on arbitrary threads.

Ideally, we would like to accomplish such marshalling via regular grain calls on system target grain references. However, this is not possible for these two cases today because the objects that receive the original notifications are by themselves aren't system targets, and hence are not allowed to make grain calls. We should consider relaxing of that constrain for local system target calls, but that question is out of scope for this "patch" PR.

Having `IStreamQueueBalanceListener` extend `IAddressable` is misleading because in reality none of the existing queue balancers can make a grain call to it anyway because they aren't system targets. That's why passing `meAsQueueBalanceListener` is misleading, too.

We are making it a responsibility of the subscriber to free-threaded notification to enqueue execution of the notification to the right scheduling context because the notification producer, in general, cannot know how to do that properly.

Did I get it right?

ReubenBond:
@sergeybykov yes, exactly: even though it would be good if we could take the principled approach of using `GrainReference` everywhere, it's not yet feasible for these subscribable interfaces (`IStreamQueueBalancer`, `ISiloStatusOracle`, etc) which may not be implemented by a `SystemTarget`. `IStreamQueueBalancer` has 9 implementations, none of which are a `SystemTarget`.

gabikliot:
Can you wait with that please? I will reply.

src/OrleansRuntime/Placement/DeploymentLoadPublisher.cs
gabikliot: Ohh, now maybe I see.

Absolutely, definitely the original goal was to talk to system targets as grain - via grain ref, and not directly like here.
I am not sure about 1 vs 40. I am very doubtful it is indeed only 1. But it should be, at least according to the original vision, the vast majority.
There should be only 2 cases really:
1) init in silo (since system target or parts of the runtime may not exist yet, so we cant rely on regular messaging to work yet)
2) direct calls from membership, to ensure strict order under lock.

All the others can/should be replaced to grain calls to system targets. There may be some more exception, but I still think the rule is to use GR as much as possible.

src/OrleansRuntime/Streams/PersistentStream/PersistentStreamPullingManager.cs
gabikliot: The former of course.
And the reason is not that it is better, but to be consistent with all the rest that we do already.
For regular grain on the same silo: do wee call directly and ask the grain to ensure its own single threaded? WE call vei addressable context anywhere, always, and therefore we should simply be consistent and keep doing it.

ReubenBond:
@gabikliot sure thing

gabikliot:
There is only one impl. of IStreamQueueBalanceListener: it is PersistentStreamPullingManager.
It is ST (system target) and thus all notifications should arrive as Grain calls.

In all cases I think we should strive to preserve our existing abstraction. Not even since it is good or bad abstraction, but in order to be consistent. The abstraction is to make grain calls. We do it almost always, the only exceptions are: 
1) init in silo (since system target or parts of the runtime may not exist yet, so we cant rely on regular messaging to work yet)
2) direct calls from membership, to ensure strict order under lock.
If there are other exception, we should try to fix them, unless not possible.

In SF case it is clear for me what need to be done: either SF needs to use a Grain client, or if not possible (I thought it is now, with non static grain client) then it can do what silo does.

Re @ReubenBond  comments:
1) `ISiloStatusOracle ` - there is only one impl, the membership oracle and of course it is ST.
2) catalog to MBR is the 2nd of the above 2 exceptions. So it might be OK to do direct calls there, although most probably not really needed. Catalog is ST, so can be notified properly. I can take a look if there was any assumptions about ordering ad that required direct call.
3) Finally, I think you might be confusing IStreamQueueBalancer with IStreamQueueBalancerListener. Listener is implemented only once, by pulling agents, who is ST.

What is maybe misleading is:
Most/All maybe queue Balancers (not queue balancer listeners, but queue balancers) are just pass through. They are oracle listeners and receive cals from the MBR oracle. In theory we could have made them ST, but it was just redundant. So instead, if all they do are pass through to the next level of balancers, or to the final listener, we optimized ST out of them (it also allowed to have strict order or notifications).
Still, the caller, Oracle, calls them as if they were ST and they call their listeners as well. So if there is one of them that is ST, it will be invoked as grain ref.
And there is indeed one of them that is ST - the only one who is the real and not pass thought -  the pulling agent.

gabikliot:
Let me present it differently: we have 2 sides and 4 cases:
a) the caller - it maybe ST or not.
b) the callee is IAddressable - but it may be implemented by ST (or grain) or maybe just POCO.

1) *Caller ST, callee ST* - clearly, caller will do grain call.
2) *Caller ST, callee is POCO*. Caller has no way to know, and why would he care. He calls as if he did grain call (same way as in 1) and its up to callee to deal with it.
3) *Caller is non ST, callee is ST* - the caller has no way but to make grain call. Otherwise, Orleans throws. Thus is has to do scheduler.QueueTask
4) *Caller is not ST, callee is not ST*. And this is the case which are are debating (SF is not ST and you have pass through oracle listeners). I claim that from caller perspective case 3 and 4 are indistigushable (he does not really know, based on the interface, if the impl. is ST or not) and thus the caller has no other way as to call as grain ref, thus like case 3, via scheduler.QueueTask. Otherwise, it means the caller knows how callee is implemented ("client knows how server is implemented").

Is that convincing?

ReubenBond:
> In SF case it is clear for me what need to be done: either SF needs to use a Grain client, or if not possible (I thought it is now, with non static grain client) then it can do what silo does.

Are you suggesting that we expose `OrleansTaskScheduler` & `ISchedulingContext` + related bits publicly or some other approach? Does SF schedule these calls on its own context or on the caller's context? It doesn't have a context itself and you're not suggesting making it an ST (internal) so I assume the latter. Therefore we also need to expose some mechanism for obtaining the caller's context. What does it look like in code?

How would GrainClient work here? We would have to do something like make sure that the "subscribe" calls take a serialized reference and then deserialize them in the context of the grain client. Otherwise you're using silo's `IRuntimeClient` and therefore have today's issue. Doesn't that just seem hellish and error prone to you? What happens if the silo isn't a gateway? We don't currently have optimizations for in-mem client-silo coupling.

ReubenBond:
We could create a new type, `IAddressableScheduler` or somesuch, which was implemented by a SystemTarget (and registered on construction) and has methods like `ScheduleTask(object context, Func<object, Task> action)` which would schedule `action` in the context of `context` if `context` was a `SystemTarget` or `Grain` instance (grains can subscribe to `ISiloStatusOracle` directly, too) and otherwise used its own context (for passthrough scenarios).

gabikliot:
Re SF: I guess I am missing all the context about it. I thought so far it was like an MBR oracle, thus is het another runtime component, and has all the access to all the internals. I was not aware it is not.

ReubenBond:
If SF had access to runtime internals I would have just made it an ST, but on principle it does not have access. The intention was to ensure that such components could be made as outside contributions.

gabikliot:
Maybe we should have started this whole issue by saying that SF is special, very different from other runtime components. I think what you are saying is it is an internal runtime component, but with limited access, and you are trying to come ip with a model for that: allow others to write components but without full access.

We had a similar probelm in the past, with streaming: pulling agent and etc... We indeed finally came up with such a model: all the complex runtiem stuff is in pulling agent, and extensions are via the cache, balancer , ...
I think the juries are still out on the question if this was a good or bad model. I think @jason-bragg actually did not support it much. Now we have quite complex interactions between all those. Perhaps, perhaps, it would have been better to give them the full power of ST.
At this point, its up to you guys to decide how to make progress on those questions.

Specifically, with SF what you could do is: I guess now it is calls directly on all listeners. Instead you can pass it "listener adaptor" that will marshal his calls to the right context. the adapter will have full access to the runtime.

Or maybe indeed SF is yet another exception in the list above.

But I was under impression that you are pushing to a general pattern of callee marshaling to his context and caller calling on what ever. And that (this new general principle) is what I objected. If you are only looking to patch SF case with what ever hack or exception, then I go ahead.

What are your thought on my 4 cases explanation? Does it make sense, from the general standpoint of design principle for runtime components communication? Unrelated from SF issue.

ReubenBond:
> Maybe we should have started this whole issue by saying that SF is special, very different from other runtime components.

I mentioned that SF bits don't have access to SystemTarget earlier: https://github.com/dotnet/orleans/pull/3290#discussion_r133638582

> I think what you are saying is it is an internal runtime component, but with limited access, and you are trying to come ip with a model for that: allow others to write components but without full access.

What's an internal component? SF lives in the main tree, but it could live in a separate repo (it started in a separate repo). SF does not have `[InternalsVisibleTo]` it.

> But I was under impression that you are pushing to a general pattern of callee marshaling to his context and caller calling on what ever. And that (this new general principle) is what I objected.

Your impression is correct: I believe that the general pattern should be: if you subscribe to notifications/callbacks on a local (non-IAddressable) interface then you are responsible for marshalling calls onto your own context. You know what you are (poco, grain, ST) and the caller does not and should not. You also don't know what the caller is and you should not. This keeps implementation decisions local rather than having considerations for them throughout the code. It covers every one of your 4 cases above without the caller or callee knowing how the other is implemented.

> What are your thought on my 4 cases explanation? Does it make sense, from the general standpoint of design principle for runtime components communication? Unrelated from SF issue.

Point 4 assumes internal access. If it weren't for that I would still be against that pattern for the reason above (caller/callee shouldn't know each other's impl details).

> Instead you can pass it "listener adaptor" that will marshal his calls to the right context. the adapter will have full access to the runtime.

Maybe this is the best approach in the long run. We could create hand-crafted adaptors for each interface, based on the caller's impl (and thereby assume the callee is an oblivious poco as in my current impl).

gabikliot:
OK, I was not aware SF is in its own repo without Internal access.
Hmm....

so how this all worked with liveness notification going not on any context? yes, that is wrong.


>if you subscribe to notifications/callbacks on a local (non-IAddressable) interface then you are responsible for marshalling calls onto your own context

Yes, agree, IF that is what you do. But I was suggesting to keep balancerListener IAddressable, as now, and then my point 4 stands.  

Maybe indeed for this case you need to use GrainClient and come up with an abstraction of remotable subscriptions?
or use the adaptor idea. adaptor can be general and public.

ReubenBond:
> OK, I was not aware SF is in its own repo without Internal access.

It's in this repo now, but it started life in OrleansContrib. We want to allow for external implementations of various APIs, so we need a pattern which can be applied without internal access.

>> if you subscribe to notifications/callbacks on a local (non-IAddressable) interface then you are responsible for marshalling calls onto your own context
>
> Yes, agree, IF that is what you do. But I was suggesting to keep balancerListener IAddressable, as now, and then my point 4 stands.

I mean that the thing you're subscribing to is not IAddressable, eg: balancerListener subscribes to balancer: balancer is not IAddressable.

> Maybe indeed for this case you need to use GrainClient and come up with an abstraction of remotable subscriptions?

That can't be done in a patch release with a straight face. There is a strong desire for a 'silo-local client', eg for HTTPS case. GrainReferences can be re-targeted to a different `IRuntimeClient` using the `IGrainFactory.BindGrainReference(ref)` method - I was wrong that we would need to round-trip through serialization. So if we did have a silo-local client, then that could be injected into the interface implementation (eg, `FabricMembershipOracle`) and incoming subscriptions could be re-bound to it if they are `IAddressable`. This has a flaw, though, because what happens if the subscriber is not IAddressable but it proxies calls to another subscriber which is? This is the case of `IStreamQueueBalancer`, which is a POCO that `IStreamQueueBalanceListener` (`IAddressable`) subscribes to, and it subscribes to `ISiloStatusOracle`. So in that case do `IStreamQueueBalancer` impls need to inject `IClusterClient` so that it can rebind the `IStreamQueueBalancerListener` instances which subscribe to it?

We could probably also modify `InsideRuntimeClient` to support context-free calls like `OutsideRuntimeClient` does, then we don't need rebinding or context marshalling, but you have even less guarantees about scheduling of notifications and you lose the safety of the exception message catching unintended calls.

> or use the adaptor idea. adaptor can be general and public.

The first version of this PR had something like this:
```C#
/// <summary>
/// Provides services for executing actions on a single-threaded, addressable context.
/// </summary>
public interface IAddressableContextScheduler
{
    /// <summary>
    /// Schedules the provided <paramref name="action"/> on this context.
    /// </summary>
    /// <param name="action">The action.</param>
    /// <returns>A <see cref="Task"/> which completes when the <paramref name="action"/> has completed.</returns>
    Task Schedule(Action action);

    /// <summary>
    /// Schedules the provided <paramref name="action"/> on this context.
    /// </summary>
    /// <param name="action">The action.</param>
    /// <returns>A <see cref="Task"/> which completes when the <paramref name="action"/> has completed.</returns>
    Task Schedule(Func<Task> action);

    /// <summary>
    /// Schedules the provided <paramref name="action"/> on this context.
    /// </summary>
    /// <param name="action">The action.</param>
    /// <returns>A <see cref="Task"/> which completes when the <paramref name="action"/> has completed.</returns>
    Task<T> Schedule<T>(Func<Task<T>> action);
}

internal class AddressableSchedulingContext : SystemTarget, IAddressableContextScheduler
{
    private readonly OrleansTaskScheduler scheduler;

    /// <inheritdoc />
    public Task Schedule(Action action)
    {
        return this.scheduler.RunOrQueueAction(action, this.SchedulingContext);
    }

    /// <inheritdoc />
    public Task Schedule(Func<Task> action)
    {
        return this.scheduler.RunOrQueueTask(action, this.SchedulingContext);
    }

    /// <inheritdoc />
    public Task<T> Schedule<T>(Func<Task<T>> action)
    {
        return this.scheduler.RunOrQueueTask(action, this.SchedulingContext);
    }

    internal static Factory<IAddressableContextScheduler> GetFactory(IServiceProvider sp)
    {
        var siloProviderRuntime = sp.GetRequiredService<SiloProviderRuntime>();
        return GetFactory(sp, siloProviderRuntime);
    }

    internal static Factory<IAddressableContextScheduler> GetFactory(IServiceProvider sp, SiloProviderRuntime siloProviderRuntime)
    {
        var contextFactory = FactoryUtility.Create<AddressableSchedulingContext>(sp);
        return () =>
        {
            var result = contextFactory();
            siloProviderRuntime.RegisterSystemTarget(result);
            return result;
        };
    }

    public AddressableSchedulingContext(
        OrleansTaskScheduler scheduler,
        ILocalSiloDetails localSiloDetails) : base(GrainId.GetGrainId(UniqueKey.NewKey()), localSiloDetails.SiloAddress)
    {
        this.scheduler = scheduler;
    }
}
```

How do you feel about that approach? The `FabricMembershipOracle` would receive `Factory<IAddressableContextScheduler> schedulingContextFactory` in its constructor and create a context there which is can use to marshal notifications to its subscribers.

sergeybykov:
>That can't be done in a patch release with a straight face.

That's why I've been advocating separating a patch like fix from the bigger refactoring. Is there s reason the current state of the PR cannot be the patch? It fixes the two cases that've been reported, and is the last fix we are holding 1.5.1 for.

gabikliot:
I also advocated for patch only.
BUT: I see this PR not as a patch but rather as major refactoring.

The major refactoring is in one line:
removing IAddressable from IStreamQueueBalanceListener :IAddressable

It makes my case 4 from the comment above not existent: it will mean now that ANY MBR oracle (including our standard one with MBR Table) will not call Queue Balancer Listener as grain call, but rather directly, thus violating the basic principle of isolation between STs in Orleans.

This is done to help with patching SF Oracle, I get it, but with the bath water we are also throwing the baby away.

ReubenBond:
It doesn't matter to me how this issue is fixed in the short term as long as the fix follows sound reasoning and does not introduce too much entropy for a patch release.

I don not consider this an issue with the SF membership provider, it's an issue with consumers of synchronous notification interfaces. How would you fix this issue only making changes to the SF provider? If you can't do that cleanly, then maybe the issue actually lies elsewhere (which is what I'm arguing).

Calling local `SystemTargets` via a `GrainReference` is the *exception*, not the norm. I provided the data [here](https://github.com/dotnet/orleans/pull/3290#discussion_r133639432). You said you doubted it. Please search the code and see - maybe I'm wrong, show me how.

That previously `IAddressable` interface is subscribing to something which might not have an addressable context. How do you propose it should work? I already showed how the implied strategy from the 4 points is not feasible without larger changes in [this comment](https://github.com/dotnet/orleans/pull/3290#issuecomment-323555631) and suggested a more isolated solution in [this comment](https://github.com/dotnet/orleans/pull/3290#issuecomment-323556896). Maybe you have an opinion on that?

Let's say that interface *does* remain `IAddressable`. In that case, what about `ISiloStatusListener`? That currently is not `IAddressable`. Should it be? If it was, how about the interaction between `LocalGrainDirectory` and `Catalog`: are we breaking isolation when `LocalGrainDirectory` calls directly into `Catalog` using its own context? `Catalog` also has a scheduling context, it's also a `SystemTarget`. Should `LocalGrainDirectory` call `Catalog` via a `GrainReference`?

gabikliot:
LGTM for the changes.

Looks like you went with the provider runtime. LGTM.

ISiloStatusListener is a bit different from any others, since there was a need to send notifications directly as sync calls (neither grain call nor direct call plus queue will work).
But for others I thought we should keep using grain calls.

Sounds like if we want to take it further, we would need to talk over Skype to discuss it in person.

gabikliot:
Actually, how does this later commit addresses the original issue? How SF Oracle would send notifications on the scheduling context? It still wouldn't now, right?

sergeybykov:
I'm confused. How does the latest version solve the issue of marshalling SF notifications that originate on external threads to the subscribers? Just running `Start` or `BecomeActive` on the ProviderManager's context is not enough, is it?

I can see the fix in the https://github.com/dotnet/orleans/pull/3290/commits/aa351136eae030142e062996d6776e30c76f3ea3 version, but am struggling with this one.

ReubenBond:
Let's skype call so we can discuss it, Gabi.

I should have added a comment when I pushed that commit. It works because of a change made back in ~june where FabricMembershipOracle steals TaskScheduler.Current in its Start method and uses it to spin up a task which awaits notifications in ProcessNotifications. Then the OnUpdate method which is called by the OS thread from SF will push notifications onto a queue and pulse the AutoResetEvent which the former is awaiting. ProcessNotifications performs the actual SiloStatusChangeNotification call to subscribers, so it is called on that captured scheduler.

It works, I tested.

Honestly, though, I feel it's a dodgy fix and I definitely prefer the previous revision. I'm uncomfortable with it - it's a hack. I spoke with Sergey this morning, since he's going to be out for the next fortnight. He also prefers the previous revision.

Let me know when you've got time so we can speak.

ReubenBond:
@dotnet-bot test netfx-bvt

ReubenBond:
I'm testing it using this branch: https://github.com/ReubenBond/orleans/tree/upgrade-sd-sample

ReubenBond:
Gabi and I had a discussion about it and we're on the same page regarding the principles and ideals.

In the end, I feel we should essentially take both the 'new' and 'old' fix.
* FabricMembershipOracle will execute on the SchedulingContext of the ProviderManager SystemTarget. This also affects implementations of `IReminderService` and `IMultiClusterOracle` - if they aren't implemented by `SystemTarget` then they get a chance to capture the ProviderManager's TaskScheduler.
* Interfaces which are not IAddressable are responsible for marshalling onto their own context. In addition to this, we removed IAddressable from IStreamQueueBalanceListener because the classes which call it are not IAddressable. Eventually, we should implement either "silo local client" or IAddressableContextScheduler (see prev comments) so that we can handle these scenarios better (using GrainReferences).

Verified fix locally using the branch linked above (I build `netfx` version of the repo to make that work).

Assuming tests pass, could someone please give it a final review and merge?

gabikliot:
Great work @ReubenBond !

PR #3291, opened 2017-08-09
[WIP] Shared state publisher prototype

jason-bragg:
Prototype for discussion purposes.
NOT FOR CHECKIN

This PR explores using a shared state publisher pattern to notify other threads of immutable state in a thread safe manner.

In this prototype, the silo status change notifications are exposed as shared state to the queue balancer, so that no queue balancing logic is run on the silo status oracle's thread.

### DIFF
--- a/src/Orleans/Utils/SharedState/ISharedStatePublisher.cs
+++ b/src/Orleans/Utils/SharedState/ISharedStatePublisher.cs
@@ -1,0 +1,17 @@
+﻿
+namespace Orleans
+{
+    /// <summary>
+    /// Shared state publisher
+    /// NOTE: All implementations must be thread safe
+    /// NOTE: All state must be immutable
+    /// </summary>
+    public interface ISharedStatePublisher<in T>
+    {
+        /// <summary>
+        /// Publish state
+        /// </summary>
+        /// <param name="state">immutable state to publish</param>
+        void Publish(T state);
+    }
+}
--- a/src/OrleansRuntime/MembershipService/SharedSiloStatusListener.cs
+++ b/src/OrleansRuntime/MembershipService/SharedSiloStatusListener.cs
@@ -1,0 +1,40 @@
+﻿
+namespace Orleans.Runtime
+{
+    public struct SiloStatusChange
+    {
+        public SiloStatusChange(SiloAddress updatedSilo, SiloStatus status)
+        {
+            this.UpdatedSilo = updatedSilo;
+            this.Status = status;
+        }
+
+        public SiloAddress UpdatedSilo { get; }
+        public SiloStatus Status { get; }
+    }
+
+    public static class SiloStatusSharedState
+    {
+        public static ISharedState<SiloStatusChange> CreateSiloStatusSharedState(this ISiloStatusOracle oracle)
+        {
+            var currentStatus = new SiloStatusChange(oracle.SiloAddress, oracle.CurrentStatus);
+            var listener = new SharedSiloStatusListener(currentStatus);
+            oracle.SubscribeToSiloStatusEvents(listener);
+            return listener.State;
+        }
+
+        private class SharedSiloStatusListener : SharedStatePublisherBase<SiloStatusChange>, ISiloStatusListener
+        {
+            public SharedState<SiloStatusChange> State => base.currentState;
+
+            public SharedSiloStatusListener(SiloStatusChange startingState) : base(startingState)
+            {
+            }
+
+            public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
+            {
+                this.Publish(new SiloStatusChange(updatedSilo, status));
+            }
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


src/Orleans/Utils/SharedState/ISharedStatePublisher.cs
```diff
@@ -0,0 +1,17 @@
+﻿
+namespace Orleans
```
jason-bragg: classes in ../Utils/SharedState/.. are shared state infrastructure.  Common to all uses of the shared state.

src/OrleansRuntime/MembershipService/SharedSiloStatusListener.cs
```diff
@@ -0,0 +1,40 @@
+﻿
+namespace Orleans.Runtime
+{
```
jason-bragg: SiloStatusChange and SiloStatusSharedState expose silo status as shared state.

src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
```diff
@@ -87,6 +85,8 @@ public class DeploymentBasedQueueBalancer : QueueBalancerBase, ISiloStatusListen
             }
             this.allQueues = new ReadOnlyCollection<QueueId>(queueMapper.GetAllQueues().ToList());
             this.siloMaturityPeriod = siloMaturityPeriod;
```
jason-bragg: Queue balancer uses the shared state, rather than being  a listener itself, limiting it to only accessing the silo status changes from it's own thread.  In this case, the queue manager system target's thread.

src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
```diff
@@ -98,19 +98,33 @@ public class DeploymentBasedQueueBalancer : QueueBalancerBase, ISiloStatusListen
...
 
+        private async Task HandleSiloStatusChange()
+        {
+            while (!this.siloStatusOracle.CurrentStatus.IsTerminating())
+            {
+                try
+                {
+                    this.siloStatusSharedState = await this.siloStatusSharedState.NextAsync;
+                    HandleSiloStatusChange(this.siloStatusSharedState.State);
+                }
```
jason-bragg: As long as the silo is not terminating, the thread will wait for silo status changes, process them, then wait for the next change.  It can only access the change in the it's own thread.

src/OrleansRuntime/MembershipService/SharedSiloStatusListener.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            public SharedState<SiloStatusChange> State => base.currentState;
+
+            public SharedSiloStatusListener(SiloStatusChange startingState) : base(startingState)
+            {
+            }
+
+            public void SiloStatusChangeNotification(SiloAddress updatedSilo, SiloStatus status)
+            {
+                this.Publish(new SiloStatusChange(updatedSilo, status));
```
jason-bragg: When we get a SiloStatusChangeNotification (on the silo status oracle's thread) all we do is publish the change.  Other threads that want to handle the change must access it via the ISharedState from their own thread.

src/OrleansRuntime/Streams/QueueBalancer/DeploymentBasedQueueBalancer.cs
```diff
@@ -98,19 +98,33 @@ public class DeploymentBasedQueueBalancer : QueueBalancerBase, ISiloStatusListen
...
         }
 
+        private async Task HandleSiloStatusChange()
+        {
+            while (!this.siloStatusOracle.CurrentStatus.IsTerminating())
+            {
+                try
+                {
+                    this.siloStatusSharedState = await this.siloStatusSharedState.NextAsync;
+                    HandleSiloStatusChange(this.siloStatusSharedState.State);
```
xiazen: I think ideal behavior is only process new state when there's a change. But this code seems to process state regardless whether it is new or not. Since `NextAsync=> this.NextCompletion.Task;`  while NextCompletion.Result  only get updated when there's new state. So most of the time the code keep processing the old NextCompletion.Result. Am I right?

xiazen: And also keep a while loop here seems to bring more performance penalty, since there's always a thread occupied, keep running this while loop until `Terminating`. Am I right? 

So if this is the expected usage pattern, then we have a performance penalty here?

jason-bragg: > Think ideal behavior is only process new state when there's a change

A new state is only available for processing when the task on NextAsync returns the next shared state.  The task in the next shared state is not resolved until a new state is published, so upon the task of NextAsync being resolved, we have a new state to process, and a new task to await for the next state change.

jason-bragg: Regarding the performance, we'll always have a pending task on the caller thread (in this case, the queue manager's thread), but that does not take up cpu.

Issue #3212, opened 2017-07-11
NullReferenceException for multiple implicit stream subscriptions


charleszipp:
I am trying to create an actor that subscribes to multiple streams on the same stream namespace. This is done to facilitate both fire-and-forget style calls but, more importantly, to remove the temporal dependency between this actor and other actors in different solutions.

When i produce the CommandDispatched event, the actor is activated as expected. However, i want to also ensure a subscription is setup for other events (CommandStarted, CommandCompleted, etc). When it hits the second subscription for CommandStarted, a NullReferenceException is thrown.

At the time the exception is thrown, there are not actually any CommandStarted events on the stream. However, i need some way to subscribe to this stream before its events are produced. Do i need to do an explicit subscription here? Possibly from the CommandDispatched handler?

```csharp
[ImplicitStreamSubscription("Routes.Activities")]
public class ActivityMonitor : JournaledGrain<ActivityState, IEvent>, IActivityMonitor
{
	protected Task OnObserveCommandDispatched(CommandDispatched @event, StreamSequenceToken token = null)
    {
        //handle event
        return Task.CompletedTask;
    }

    protected Task OnObserveCommandStarted(CommandStarted @event, StreamSequenceToken token = null)
    {
        //handle event
        return Task.CompletedTask;
    }

    public override async Task OnActivateAsync()
    {
        await GetStreamProvider("activities").GetStream<CommandDispatched>(this.GetPrimaryKey(), "Routes.Activities").SubscribeAsync(OnObserveCommandDispatched);
        //this throws exception object null reference exception
        await GetStreamProvider("activities").GetStream<CommandStarted>(this.GetPrimaryKey(), "Routes.Activities").SubscribeAsync(OnObserveCommandStarted);
    }
}

public class Program
{
	private static IClientFactory _clients = new ServiceFabricClusterClientFactory();
	private static void Main(string[] args)
	{
		var client = _clients.Get<IActivityMonitor>().Result;
		var command = new CreateEngagementArgs(id, "Test 123", firmGuid);
        var @event = new EngagementCreated(id, "Test 123", firmGuid);

        client.GetStreamProvider("activities").GetStream<CommandDispatched>(activityId, "Routes.Activities").OnNextAsync(
            new CommandDispatched(activityId, "Routes.Activities", Newtonsoft.Json.JsonConvert.SerializeObject(command), DateTime.UtcNow)
            ).Wait();

        //todo: produce CommandStarted...
	}
}
```

```
[...]

PR #3301, opened 2017-08-14
Improve thrown NullReferenceException when calling GetStream<T>

xiazen:
Improve thrown NullReferenceException when calling `GetStream<T>`, by throwing a more understanable exception. 

Stream is typed, but stream id isn't. If application reference a stream with a different type, but using the same stream id, then `GetStream<T>` will throw a `NullReferenceException`. Because runtime is casting a `IAsyncStream<T1>` to a `IAsyncStream<T2>`

This is confusing to users. So throw a more understandable exception to help user fixing their application code.

### DIFF
--- a/src/Orleans/Streams/Internal/StreamDirectory.cs
+++ b/src/Orleans/Streams/Internal/StreamDirectory.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Threading.Tasks;
 
 namespace Orleans.Streams
@@ -24,7 +25,7 @@ namespace Orleans.Streams
             var streamOfT = stream as IAsyncStream<T>;
             if (streamOfT == null)
             {
-                throw new Runtime.OrleansException($"Stream type mismatch. A stream can only support a single type of data. The generic type of the stream requested ({typeof(T)}) does not match the previously requested type ({stream.GetType().GetGenericArguments().FirstOrDefault()}).");
+                throw new Runtime.OrleansException($"Stream type mismatch. A stream can only support a single type of data. The generic type of the stream requested ({typeof(T)}) does not match the previously requested type ({stream.GetType().GetTypeInfo().GetGenericArguments().FirstOrDefault()}).");
             }
 
             return streamOfT;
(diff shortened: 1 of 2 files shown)


xiazen:
Fix for #3212

src/Orleans/Streams/Internal/StreamDirectory.cs
```diff
@@ -20,7 +20,13 @@ internal StreamDirectory()
 
         internal IAsyncStream<T> GetOrAddStream<T>(StreamId streamId, Func<IAsyncStream<T>> streamCreator)
         {
-            return allStreams.GetOrAdd(streamId, _ => streamCreator()) as IAsyncStream<T>;
+            var stream = allStreams.GetOrAdd(streamId, _ => streamCreator()) as IAsyncStream<T>;
+            if (stream == null)
+            {
+                throw new Runtime.OrleansException($"Stream type mismatch. Please use GetStream<T>(Guid id, string streamNamespace) with the correct type, which is the type application first referenced the stream with.");
```
jason-bragg: Just my opinion here, but I'm not sure we should be making suggestions, rather than just telling the user what is wrong.  A stream can only support a single type of data.  The generic type of the stream requested does not match the previously requested type.  If we can get at the previously requested type to report the conflicting types, that would be great, but if that's hard to get, that's ok as it's not strictly necessary.

xiazen: Fixed. and added tests.
Sample exception is like:

Message: Orleans.Runtime.OrleansException : Stream type mismatch. A stream can only support a single type of data. The generic type of the stream requested does not match the previously requested type Orleans.Streams.StreamImpl`1[System.Int32].

xiazen:

xiazen:

src/Orleans/Streams/Internal/StreamDirectory.cs
```diff
@@ -20,7 +20,14 @@ internal StreamDirectory()
 
         internal IAsyncStream<T> GetOrAddStream<T>(StreamId streamId, Func<IAsyncStream<T>> streamCreator)
         {
-            return allStreams.GetOrAdd(streamId, _ => streamCreator()) as IAsyncStream<T>;
+            var stream = allStreams.GetOrAdd(streamId, _ => streamCreator());
+            var streamOfT = stream as IAsyncStream<T>;
+            if (streamOfT == null)
+            {
+                throw new Runtime.OrleansException($"Stream type mismatch. A stream can only support a single type of data. The generic type of the stream requested does not match the previously requested type {stream.GetType()}.");
+            }
```
jason-bragg: Looks good, but maybe:
The generic type of the stream requested ({typeof(T)}), does not match the previously requested type ({stream.GetType().GetGenericArguments().FirstOrDefault()}).

xiazen:
@dotnet-bot test this please

PR #3302, opened 2017-08-15
Minimal SiloBuilder

### LINKED ISSUE #3312: Avoid registering or resolving Func<T> with the container
Based on [this comment](https://github.com/dotnet/orleans/pull/3302#discussion_r133552953) by @davidfowl:
> Adding and resolving `Func<T>` will mess with providers that natively support `Func<T>`. I would just get rid of this.

We do plan to get rid of `Func<NodeConfiguration>` anyway in the short/medium term (due to our work with Silo Builder, where we will favor strongly typed POCO options, and resolve them via `IOptionsSnapshot<TOptions>`), but this is just 1 usage of `Func<T>`.

ReubenBond:
Fixes #3312

This is a minimal "silo builder". The purpose is to make it significantly easier to configure, create, and operate a silo than it is at present.

We hope to collaborate on and support the generic HostBuilder abstraction once it's ready: https://github.com/aspnet/Hosting/issues/1163, but this lets us continue making progress in the meantime.

Sample usage:
```C#
public static async Task MainAsync(string[] args)
{
    var silo = new SiloBuilder().ConfigureLocalHostPrimarySilo().Build();

    await silo.StartAsync();

    Task.Run(async () =>
    {
        // Wait an arbitrary amount of time and then kill the silo.
        // This could also be waiting for console input or some other condition.
        await Task.Delay(TimeSpan.FromMinutes(2));
        await silo.StopAsync();
    }).Ignore();

    await silo.Stopped;
}
```

* Moved default service implementations into `DefaultSiloServices` static class so that the implementation can be shared by SiloBuilder and Silo (for backwards compatibility)
* Broke cyclic dependency between `Silo` construction and: `MembershipOracle`, `MembershipOracleData`, and `DeploymentLoadPublisher`.
* Added `ISilo` interface for interacting with silo instances:
* Added `ISiloBuilder` interface and `SiloBuilder` implementation for incrementally configuring and eventually constructing an `ISilo` instance.
* Added an awaitable `siloTerminationTask` to `Silo` which is signalled on termination. Currently this sits alongside the similarly purposed (but not awaitable) `SiloTerminatedEvent`, which is a `WaitHandle`


src/Orleans/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -10,6 +10,45 @@ namespace Orleans.Configuration
...
+            var providedService = services.FirstOrDefault(service => service.ServiceType == typeof(TService));
+            if (providedService == null && implementation != null)
+            {
+                var newRegistration = new ServiceDescriptor(
+                    typeof(TService),
+                    sp => sp.GetRequiredService<TImplementation>(),
+                    implementation.Lifetime);
+                services.Add(newRegistration);
+            }
+        }
```
ReubenBond: I spotted a complaint from NuGet that it was skipping one of the files named ServiceCollectionExtensions.cs so I merged them. Also added this method.

src/OrleansRuntime/Hosting/SiloBuilderExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// IServiceProvider ConfigureServices(IServiceCollection);
+        /// </code>
+        /// </remarks>
+        public static ISiloBuilder UseLegacyStartup<TStartup>(this ISiloBuilder builder) where TStartup : class
+        {
+            var configureServicesMethod = StartupBuilder.FindConfigureServicesDelegate(typeof(TStartup));
+            if (configureServicesMethod == null) throw new ArgumentException($"Type {typeof(TStartup)} does not have a suitable method for configuring a service provider.");
+            if (configureServicesMethod.MethodInfo.IsStatic) throw new ArgumentException($"Method {configureServicesMethod} on type {typeof(TStartup)} must not be static.");
+            return builder.ConfigureServiceProvider(svc => StartupBuilder.ConfigureStartup(typeof(TStartup).AssemblyQualifiedName, svc));
+        }
```
ReubenBond: Should this method even be included? Legacy startup class can be configured as it is today - via `ClusterConfiguration`. I called it "legacy" to make it clear that this is not the best path forward but might be an easy migration option.

src/OrleansRuntime/MembershipService/MembershipOracle.cs
```diff
@@ -38,16 +38,16 @@ internal class MembershipOracle : SystemTarget, IMembershipOracle, IMembershipSe
         public SiloAddress SiloAddress { get { return membershipOracleData.MyAddress; } }
         private TimeSpan AllowedIAmAliveMissPeriod { get { return orleansConfig.Globals.IAmAliveTablePublishTimeout.Multiply(orleansConfig.Globals.NumMissedTableIAmAliveLimit); } }
 
-        public MembershipOracle(Silo silo, MembershipTableFactory membershipTableFactory, IInternalGrainFactory grainFactory)
-            : base(Constants.MembershipOracleId, silo.SiloAddress)
+        public MembershipOracle(ILocalSiloDetails siloDetails, ClusterConfiguration clusterConfiguration, NodeConfiguration nodeConfiguration, MembershipTableFactory membershipTableFactory, IInternalGrainFactory grainFactory)
+            : base(Constants.MembershipOracleId, siloDetails.SiloAddress)
```
ReubenBond: Intention here is to break the dependency on Silo, since Silo is now being constructed by DI when the builder is being used.

ReubenBond:

ReubenBond:

src/Orleans/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -10,6 +10,45 @@ namespace Orleans.Configuration
...
+        /// <summary>
+        /// Attempts to use an existing registration of <typeparamref name="TImplementation"/> to satisfy the service type <typeparamref name="TService"/>.
+        /// </summary>
+        /// <typeparam name="TService">The service type being provided.</typeparam>
+        /// <typeparam name="TImplementation">The implementation of <typeparamref name="TService"/>.</typeparam>
+        /// <param name="services">The service collection.</param>
+        internal static void TryAddFromExisting<TService, TImplementation>(this IServiceCollection services) where TImplementation : TService
+        {
+            var implementation = services.FirstOrDefault(service => service.ServiceType == typeof(TImplementation));
+            var providedService = services.FirstOrDefault(service => service.ServiceType == typeof(TService));
```
jdom: This is a little bit strange, are you sure it is doing the right thing?
Also, the XML docs for these 2 methods look the same, what's the intended difference?

src/OrleansRuntime/Hosting/ISiloBuilder.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <param name="configureServices">The service configuration delegate.</param>
+        /// <returns>The builder.</returns>
+        ISiloBuilder ConfigureServices(Action<IServiceCollection> configureServices);
+        
+        /// <summary>
+        /// Specified how the <see cref="IServiceProvider"/> for this silo is configured. 
+        /// </summary>
+        /// <param name="configureServiceProvider">The service provider configuration method.</param>
+        /// <returns>The builder.</returns>
+        ISiloBuilder ConfigureServiceProvider(Func<IServiceCollection, IServiceProvider> configureServiceProvider);
```
jdom: This one should probably be replaced with `ISiloBuilder UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory);` to align better with the generic host builder. Or is there a reason not to use the IServiceProviderFactory abstraction?

src/OrleansRuntime/Hosting/SiloBuilderExtensions.cs
jdom: Probably not, since I believe the plan is to deprecate it eventually. If the user can still configure the startup type via the cluster configuration, there's no need to add another API for it.
It is not very clear to me yet how do we execute/configure stuff after the service provider was created, but still not sure whether it will be needed once we have named options.

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -971,7 +848,8 @@ private void FastKill()
 
             // Setting the event should be the last thing we do.
             // Do nothijng after that!
-            siloTerminatedEvent.Set();  
+            siloTerminatedTask.SetResult(0);
+            siloTerminatedEvent.Set();
```
jdom: I believe you can just explicitly cast the siloTerminatedTask to a WaitHandle, instead of having 2 separate signals.

src/OrleansRuntime/Hosting/SiloWrapper.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <inheritdoc />
+        public IServiceProvider Services { get; }
+
+        /// <inheritdoc />
+        public Task Stopped { get; }
+
+        /// <inheritdoc />
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            this.silo.Start();
```
jdom: Should you check whether the token is canceled before starting?

src/OrleansRuntime/Hosting/SiloWrapper.cs
```diff
@@ -0,0 +1,46 @@
...
+            return Task.CompletedTask;
```
jdom: is the expected behavior that if the token was canceled before we return, that we shut down?

jdom:

src/Orleans/Configuration/ServiceCollectionExtensions.cs
ReubenBond: AddFromExisting = "if TImplementation exists, register it as a provider of TService"
TryAddFromExisting = "if TImplementation exists and there are currently no providers of TService, register TImplementation as a provider of TService"

I could simplify the Try method to call into AddFromExisting. I'll fix the doc comments.

src/OrleansRuntime/Hosting/ISiloBuilder.cs
ReubenBond: I'm trying to work out why that abstraction has those complexities and why it's useful.
Would it be called using reflection and stored as `object`? Otherwise, what's the type of `TContainerBuilder`?

ReubenBond: Maybe like this ;)
```C#
public ISiloBuilder UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory)
{
    this.containerBuilder = services => factory.CreateServiceProvider(factory.CreateBuilder(services));
    return this;
}
```

src/OrleansRuntime/Hosting/SiloWrapper.cs
ReubenBond: I think not, since that's very difficult to observe and test for. Cancellation must be occurring in parallel to the `StartAsync` call, so how can it distinguish between "about to return" and "just returned"?

src/OrleansRuntime/Hosting/SiloWrapper.cs
ReubenBond: I will put a little check there. It's not dangerous to ignore it, though.

src/OrleansRuntime/Silo/Silo.cs
ReubenBond: I can't see how. I thought about uniting them. I think we should obsolete/eliminate the WaitHandle instead, it was always annoying to use.

`SiloHost` can use the task in `WaitForOrleansSiloShutdown` (and we can hopefully obsolete `SiloHost`, too)

ReubenBond:
Updated with support for `IServiceProviderFactory<T>`, still not sure how I feel about it, though.

src/OrleansRuntime/Hosting/SiloWrapper.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        /// <inheritdoc />
+        public IServiceProvider Services { get; }
+
+        /// <inheritdoc />
+        public Task Stopped { get; }
+
+        /// <inheritdoc />
+        public Task StartAsync(CancellationToken cancellationToken)
```
jdom: I know that in several places we don't respect that, but since the Start api might be used in a very different place as the rest of Orleans, I would make sure to never throw an exception synchronously. Instead return a faulted task (sure, synchronously too). For that do a try/catch or mark the method async (although you used to get a warning with the latter, not sure if that's the case with the new c#).

src/OrleansRuntime/Hosting/SiloWrapper.cs
jdom: Yes, true, I thought about the same thing... it's not entirely clear in the spec in aspnet/Hosting#1163.
Perhaps @davidfowl can clarify what's the expected behavior when the token is cancelled before `StartAsync` finishes?
I believe that `StopAsync` case is clear: it should ungracefully stop when the token is canceled, as the process can be terminated immediately.

jdom:

src/OrleansRuntime/Hosting/IServiceProviderFactory.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.Runtime.Hosting
+{
+    /// <summary>
+    /// Abstraction for integrating <see cref="IServiceProviderFactory{TContainerBuilder}"/> implementations.
+    /// </summary>
+    internal interface IServiceProviderFactory
```
jdom: I understand why this is here, but it might be a little bit confusing to have this `IServiceProviderFactory` be different than the generic interface with the same name in the DI library. Typically interfaces of the same name that just vary in their generic argument arity are very similar to each other.
Perhaps `IServiceProviderFactoryAdapter` or `IServiceProviderFactoryBuilder` are not necessarily more precise on their own, but at least different enough from the other to not confuse them with each other. Thoughts? I'm fine if we decide to keep it though.

galvesribeiro:

src/OrleansRuntime/Hosting/DefaultSiloServices.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            services.AddOptions();
+
+            // Register system services.
+            services.TryAddSingleton<ISilo, SiloWrapper>();
+            services.TryAddFromExisting<ILocalSiloDetails, SiloInitializationParameters>();
+            services.TryAddSingleton(sp => sp.GetRequiredService<SiloInitializationParameters>().ClusterConfig);
+            services.TryAddSingleton(sp => sp.GetRequiredService<SiloInitializationParameters>().GlobalConfig);
+            services.TryAddTransient(sp => sp.GetRequiredService<SiloInitializationParameters>().NodeConfig);
+            services.TryAddSingleton<Func<NodeConfiguration>>(
```
davidfowl: Adding and resolving `Func<T>` will mess with providers that natively support `Func<T>`. I would just get rid of this.

jdom: Thanks, created https://github.com/dotnet/orleans/issues/3312 to track it, since this is just a move of the existing registrations, and there's several other occurrences. No need to address in this particular PR just yet.

ReubenBond:
Fixes #3312

jdom:

jdom:

Reason: unrelated failures that should be solved by now

Issue #3299, opened 2017-08-14
Memory leak when using ClusterClient


elqueffo:
I noticed our memory usage climbing to 6gb on the client side when calling into our very simple grains. I managed to get a dotMemory snapshot and saw that most of it is in ProxiedMessageCenter.

I set up a small test case where i create a client like so:

            var serviceName = new Uri("fabric:/StatelessCalculatorApp/StatelessCalculatorService");
            IClusterClient client = null;
            for (int i = 0; i < 100; i++)
            {
                client = new ClientBuilder()
                    .UseConfiguration(new ClientConfiguration())
                    .AddServiceFabric(serviceName)
                    .ConfigureServices(
                        services =>
                        {
                            // Some deployments require a custom FabricClient, eg so that cluster endpoints and certificates can be configured.
                            // A pre-configured FabricClient can be injected.
                            var fabricClient = new FabricClient();
                            services.AddSingleton(fabricClient);
                        })
                    .Build();
                await client.Connect();
                //await client.Close();
                client.Dispose();
            }
            Console.ReadKey();

Not calling client.Close() (note it is commented out above), will leak exactly 100 OutsideRuntimeClient, which retains ProxiedMessageCenter. ProxiedMessageCenter contains PendingInboundMessages, which in our case was the culprit of the enormous memory usage.

It seems to me that calling Dispose should clean up after itself, and not require a call to Close().

Changing the above code to call client.Close() will still cause there to be two OutsideRuntimeClient instances, even after forcing GC. As i understand it, there should only be one?

elqueffo:
So yesterday i added a Close() to our client code, but overnight i am again back at 2+gb of memory usage on the client.
![orleans-mem1](https://user-images.githubusercontent.com/8071240/29305661-270cce62-819a-11e7-8ad0-5e347ac94953.jpg)
Looks like ProxiedMessageCenter is using most of it, in PendingInboundMessages. Am I missing something obvious that i should be doing to stop this from growing seemingly unbounded?


![orleans-mem2](https://user-images.githubusercontent.com/8071240/29305743-873f0156-819a-11e7-81ca-a9881468f2d2.jpg)

ReubenBond:
@elqueffo how many clients are you creating in your application and what is their lifecycle? Clients are intended to live for the duration of your application and be shared between threads.

elqueffo:
@ReubenBond i create one client, but if it IsInitialized becomes false at any point after the initial setup, i recreate it.

[...]

PR #3306, opened 2017-08-16
ClusterClient: Dispose() is equivalent to Abort()

ReubenBond:
As per recommendation here https://github.com/dotnet/orleans/issues/3299#issuecomment-322528018

Note that:
* `Dispose()` is equivalent to `Abort()`
* `Abort()` is equivalent to a blocking call to `Stop(graceful: false)`
* `Close()` is equivalent to `Stop(graceful: true)`
(`Stop(bool)` is private)

### DIFF
--- a/src/Orleans/Core/ClusterClient.cs
+++ b/src/Orleans/Core/ClusterClient.cs
@@ -18,9 +18,11 @@ namespace Orleans
 
         private enum LifecycleState
         {
+            Invalid,
             Created,
             Starting,
             Started,
+            Disposing,
             Disposed,
         }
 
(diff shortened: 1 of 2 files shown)


ReubenBond:
@dotnet-bot please test this

ReubenBond:
@dotnet-bot test this please

sergeybykov:
@dotnet-bot test netfx-bvt 

ReubenBond:
Fixed an infinite recursion. ClusterClient now has a `Disposing` state which breaks the recursion.
Recursion occurred because `ClusterClient` is registered with the DI container but also transitively owns the container, so client.Dispose() results in container.Dispose() which results in client.Dispose() and so on.

ReubenBond:

ReubenBond:
Perhaps we should include this in 1.5.1

ReubenBond:
These test failures are ignorable.

PR #3307, opened 2017-08-16
Update Actor-Identity.md

Cronan:
The third argument must be passed to disambiguate between two overloads of  GetGrain() . Otherwise, the compiler thinks it's the  GetGrain(<int key>, <implementation class name prefix>) .


sergeybykov:
Thank you, @Cronan!

PR #3308, opened 2017-08-16
[WIP] Configure grain-level interceptors via DI

ReubenBond:
This PR allows grain-scoped injected services (eg, transactional storage, streams) to configure grain call filters. This is useful for features such as transactions where logic must be executed after a grain call has completed.

Hopefully something like this would allow us to integrate transactions without modifying core Orleans.

Note: the intention is *not* to merge this PR as-is, it's just a PoC.


sergeybykov:
Do we need this for transactions? If so, need to resolve the conflicts.

ReubenBond:
We don't need it, it just helps us separate the core transaction execution logic from the Invoke path. We can close it and maybe I'll implement it again some day.

PR #3313, opened 2017-08-17
SerializationManager: register additional generic collection interfaces

### LINKED ISSUE #3268: TypeAccessException when sending an immutable of ICollection between silos
We are on Orleans 1.5 now and we are running into an issue. The problem happens on a cluster of silos, not a single silo.

- Grain A calls Grain` B.MyMethod(Immutable<ICollection<IMyInterface>>)`
- Grain B has `[PreferLocalPlacement]` attribute.
- Even if we keep the `PreferLocalPlacement `on grain B, the error rarely happens, but it still happens.
- If we remove the local placement attribute, it happens quite a lot. This suggests that this error is happening when it is sending an immutable object to another silo in the cluster.
- If we do not send the ICollection<IMyInterface> as Immutable<>, we are not running into this issue at all but it is a significant performance hit.

The exception we are getting is something like this:
`"Orleans.Concurrency.Immutable1<ICollection1<MyNamespace.Entities.IMyInterface>>" is invalid: Type string "ICollection``1'1" cannot be resolved.`

In the Orleans code base, the exception is thrown in the `SerializationManager.ResolveTypeName(string) `method, as the string argument is passed in as `ICollection'1'1`. Therefore, `var i = typeName.IndexOf('<')` returns -1 and throws TypeAccessException in the same class.

As far as I can tell, we did not have this issue on 1.3 binaries.

Is this a known issue?

ReubenBond:
Fixes #3268

This is not blocking for 1.5.1, but it's small and low risk to include.

### DIFF
--- a/src/Orleans/Serialization/SerializationManager.cs
+++ b/src/Orleans/Serialization/SerializationManager.cs
@@ -324,6 +324,11 @@ namespace Orleans.Serialization
             Register(typeof(IList<>));
             Register(typeof(IDictionary<,>));
             Register(typeof(IEnumerable<>));
+            Register(typeof(ICollection<>));
+            Register(typeof(ISet<>));
+            Register(typeof(IReadOnlyDictionary<,>));
+            Register(typeof(IReadOnlyCollection<>));
+            Register(typeof(IReadOnlyList<>));
 
             // Enum names we need to recognize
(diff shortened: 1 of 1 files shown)


ReubenBond:
Ignore the test failure.

PR #3315, opened 2017-08-17
Directly propagate OnActivateAsync failures to callers

### LINKED ISSUE #3213: Retaining exceptions from grain OnActivateSync
Hi - wondering if anyone can help.

The scenario we are after here is the following:

- An Orleans Client calls a grain that is not currently active
- The OnActivateAsync method of the grain errors (i.e. SqlException, database is down for whatever reason).
- The exception of the error is bubbled up to the client's try-catch block as an "OrleansException".

At this point, the InnerException is null, leaving only the stack trace giving any indication of the error.

My question: is it possible for OrleansException to retain SqlException inside the InnerException?
Is there something I've missed or it just simply not possible?

Thanks in advance.

ReubenBond:
Fixes #3213

* Failures in `OnActivateAsync` will be propagated directly back to the caller.
* Calls will be retried up to `IMessagingConfiguration.MaxResendCount` times (default is 0) because the rejection is marked as `Transient`.
* Other message rejections which include an exception will also be propagated directly back to the caller, but this doesn't seem undesirable. The difference is that previously the exception would be discarded in favor of a generic exception (`OrleansMessageRejectionException`) which typically contained the original exception's `Message` in its `Message` property.

### DIFF
--- a/src/OrleansRuntime/Catalog/Catalog.cs
+++ b/src/OrleansRuntime/Catalog/Catalog.cs
@@ -680,7 +680,9 @@ namespace Orleans.Runtime
                                 SchedulingContext).Ignore();
                         }
 
-                        RerouteAllQueuedMessages(activation, null, "Failed InvokeActivate", exception);
+                        // Reject all of the messages queued for this activation.
+                        var activationFailedMsg = nameof(Grain.OnActivateAsync) + " failed";
+                        RejectAllQueuedMessages(activation, activationFailedMsg, exception);
                         break;
                 }
             }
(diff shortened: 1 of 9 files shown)


ReubenBond:
The test failure can be ignored.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -680,7 +680,9 @@ private enum ActivationInitializationStage
                                 SchedulingContext).Ignore();
                         }
 
-                        RerouteAllQueuedMessages(activation, null, "Failed InvokeActivate", exception);
+                        // Reject all of the messages queued for this activation.
+                        var activationFailedMsg = nameof(Grain.OnActivateAsync) + " failed";
+                        RejectAllQueuedMessages(activation, activationFailedMsg, exception);
```
sergeybykov: Isn't there a race here that will allow new requests to get enqueue right after we reject all messages that are already in the queue? Are we preventing enqueuing of new requests before we start the rejection loop?

ReubenBond: The activation is already marked as invalid and this is running inside of a lock on the activation. Is there room for a race?

sergeybykov: If it's already marked as invalid, then I think we are fine.

rvarna:
Is this change available in 1.5.1 or 1.5.2 (and 1.5.1)?

sergeybykov:
No, it's only in master. Will be in 2.0.0-beta.

PR #3318, opened 2017-08-17
More fixes cherry-picked for 1.5.1

sergeybykov:
I included some of the test fixes to reduce the noise while testing.
I think the last thing we want to bring is a fix to https://github.com/dotnet/orleans/issues/3273 before we can release 1.5.1.

### DIFF
--- a/src/Orleans/Core/ClusterClient.cs
+++ b/src/Orleans/Core/ClusterClient.cs
@@ -18,9 +18,11 @@ namespace Orleans
 
         private enum LifecycleState
         {
+            Invalid,
             Created,
             Starting,
             Started,
+            Disposing,
             Disposed,
         }
 
@@ -42,16 +44,6 @@ namespace Orleans
         public IGrainFactory GrainFactory => this.InternalGrainFactory;
 
         /// <inheritdoc />
-        internal IInternalGrainFactory InternalGrainFactory
-        {
-            get
-            {
-                this.ThrowIfDisposedOrNotInitialized();
-                return this.runtimeClient.InternalGrainFactory;
-            }
-        }
-
-        /// <inheritdoc />
         public Logger Logger
         {
             get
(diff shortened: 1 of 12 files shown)


src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -325,6 +325,11 @@ internal void RegisterBuiltInSerializers()
             Register(typeof(IList<>));
             Register(typeof(IDictionary<,>));
             Register(typeof(IEnumerable<>));
+            Register(typeof(ICollection<>));
+            Register(typeof(ISet<>));
+            Register(typeof(IReadOnlyDictionary<,>));
+            Register(typeof(IReadOnlyCollection<>));
+            Register(typeof(IReadOnlyList<>));
```
jdom: This is breaking if used for persistance or versioning. Are we OK including it in the patch release?

src/Orleans/Core/ClusterClient.cs
```diff
@@ -118,26 +126,31 @@ public void Abort()
...
         private async Task Stop(bool gracefully)
         {
-            if (this.state == LifecycleState.Disposed) return;
-
+            if (this.IsDisposing) return;
             using (await this.initLock.LockAsync().ConfigureAwait(false))
             {
                 if (this.state == LifecycleState.Disposed) return;
-                if (gracefully)
+                try
```
jdom: @ReubenBond shouldn't this check for IsDisposing after grabbing the lock?

src/Orleans/Core/ClusterClient.cs
```diff
@@ -118,26 +126,31 @@ public void Abort()
...
-                this.Dispose();
             }
         }
 
         /// <inheritdoc />
-        public void Dispose()
-        {
-            this.Dispose(true);
-        }
+        void IDisposable.Dispose() => this.Abort();
```
jdom: small breaking change here. Why was this turned into an explicit implementation of IDisposable? I think it's negligible, since the concrete class is internal, so the explicit implementation might not break anybody, but not sure I understand the reasoning

src/Orleans/Core/ClusterClient.cs
ReubenBond: `state` could never be `Disposing` here: only `Disposed` or `Invalid`. I'm not entirely sure if the state can ever actually transition to `Invalid` given `SafeExecute`, but maybe through some strange turn of events

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: shouldn't be, since this won't change how anything is actually serialized, only what's permissible during deserialization

src/Orleans/Core/ClusterClient.cs
ReubenBond: It's a small breaking change, hopefully it's not an issue. It was done to adhere to MS design guidelines

jdom:
Running tests here: https://ci.dot.net/job/dotnet_orleans/job/master/job/netfx-functional_prtest/722/

PR #3320, opened 2017-08-18
Update all xUnit related packages to the latest stable release

jdom:
Looks like test discovery was finding a lot less tests that it should, when using xunit.testrunner.console in our CI builds.
Specifically, non of the SkippableFact tests were being found


attilah:
This change is included in the source tree merge things. You want to try CI with these new version?

attilah:
Already addressed in https://github.com/dotnet/orleans/pull/3231.

PR #3327, opened 2017-08-20
Update A-Service-is-a-Collection-of-Communicating-Actors.md

siva-github:
It was mentioned " we're choosing to not rely on inheritance when defining the `Manager` class, even though a `Manager` is clearly also an `Employee`." but Manager is inherited from Employee later on.


sergeybykov:
Thank you, @siva-github!

PR #3333, opened 2017-08-23
Migrate to asp.net logging abstraction series 1

xiazen:
I'm working on [branch](https://github.com/xiazen/orleans/tree/asp-net-logging-move) to migrate Orleans internal runtime off Orleans legacy logger. That branch migrated about 80% of the code base off it. But to avoid big PR, I will start picking commits from that branch to form small PR. 

This is the first PR from that branch to start this migration. In this PR.

- Added LoggerWrapper to support legacy GetLogger method provided by Grain base class, ProvideRuntime classes. 
- Migrated SimpleMessageStreamProducerExtensions to use Microsoft.Extensions.Logging. ILogger
- Migrated IRuntimeClient and its implementations to use ILogger
- Added default Logging config to DefaultSiloServices 
- Mark methods related to Orleans.Runtime.Logger as obsolete.


src/Orleans/Core/ClientBuilder.cs
```diff
@@ -79,6 +80,8 @@ private static void AddBasicServices(IServiceCollection services, ClientConfigur
             services.AddFromExisting<IInternalGrainFactory, GrainFactory>();
             services.AddFromExisting<IGrainReferenceConverter, GrainFactory>();
             services.AddSingleton<ClientProviderRuntime>();
+            services.AddLogging();
+            services.AddSingleton<ILoggerFactory>(sp=> sp.GetRequiredService<ILoggerFactory>().AddConsole());
```
ReubenBond: does this mean we're always logging to the console?

src/Orleans/Core/ClientBuilder.cs
```diff
@@ -81,6 +83,8 @@ private static void AddBasicServices(IServiceCollection services, ClientConfigur
             services.AddFromExisting<IInternalGrainFactory, GrainFactory>();
             services.AddFromExisting<IGrainReferenceConverter, GrainFactory>();
             services.AddSingleton<ClientProviderRuntime>();
+
+            services.AddLogging();
```
xiazen: we can configure a fancier default logger when we upgrade to Microsoft.Extensions.Logging 2.0 . 
using 
`services.AddLogging(builder => builder.AddConsole())` or similar. currently LogBuilder is not available in 1.3

src/Orleans/Core/ClientBuilder.cs
xiazen: LogFactory is a singleton, I think if user configure their own logFacory, it overrides the default one. Is my understanding of DI correct?

jdom: hmmm, this doesn't seem right...

xiazen: this is outdated code, I changed it to just services.AddLogging() in latest code. 

in 1.3 configure logging is not as easy as 2.0. That's why this weird code is there.

xiazen:

xiazen:

src/Orleans/Core/ClientBuilder.cs
jdom: Yes, AddLogging is fine, what is not is AddConsole. That's the user that needs to configure it however he wants

xiazen:
ready for review

src/Orleans/Logging/Logger.cs
```diff
@@ -23,6 +24,7 @@ public abstract class Logger
         /// </summary>
         /// <param name="loggerName">Name of the Logger to find or create</param>
         /// <returns>Logger associated with the specified name</returns>
+       //TODO: Mark it as [Obsolete] after all runtime has migrated
```
jdom: no need to, since the entire type is obsolete

src/Orleans/Logging/LoggerWrapper.cs
```diff
@@ -0,0 +1,183 @@
...
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Orleans.Runtime
+{
+    //TODO: Mark it as [Obsolete] after all runtime has migrated
+    internal class LoggerWrapper : Logger
```
jdom: Shouldn't this be `LoggerImpl` as it was before? It would make following its history (and diff) easier

src/Orleans/Logging/ILoggerExtensions.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// <param name="logCode">The log code associated with this message.</param>
+        /// <param name="message">The log message.</param>
+        public static void Debug(this ILogger logger, int logCode, string message)
+        {
+            logger.LogDebug(LoggingUtils.CreateEventId(0, logCode), message);
+        }
+
+        internal static void Debug(this ILogger logger, ErrorCode logCode, string message)
+        {
+            logger.LogDebug(LoggingUtils.CreateEventId(0, logCode), message);
```
jdom: why are you passing `0` in all of these calls? That was the important piece of information, not just the string for the error code.
Shouldn't the ID be inferred from the ErrorCode instead?

src/Orleans/Runtime/GrainReferenceRuntime.cs
```diff
@@ -14,18 +15,18 @@ internal class GrainReferenceRuntime : IGrainReferenceRuntime
...
 
         private readonly Action<Message, TaskCompletionSource<object>> responseCallbackDelegate;
-        private readonly Logger logger;
+        private readonly ILogger logger;
         private readonly IInternalGrainFactory internalGrainFactory;
         private readonly SerializationManager serializationManager;
 
         public GrainReferenceRuntime(
-            Factory<string, Logger> loggerFactory,
+            ILoggerFactory loggerFactory,
```
jdom: I believe this is being created by the container. If that's the case, you can just inject `ILogger<GrainReferenceRuntime>` (and assign it to the `ILogger` field) so that you skip the step of using the factory to create the actual logger.

src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducer.cs
```diff
@@ -32,12 +33,14 @@ internal class SimpleMessageStreamProducer<T> : IInternalAsyncBatchObserver<T>
...
-
+        [NonSerialized]
+        private readonly ILoggerFactory loggerFactory;
         internal bool IsRewindable { get; private set; }
 
         internal SimpleMessageStreamProducer(StreamImpl<T> stream, string streamProviderName,
             IStreamProviderRuntime providerUtilities, bool fireAndForgetDelivery, bool optimizeForImmutableData,
-            IStreamPubSub pubSub, bool isRewindable, SerializationManager serializationManager)
+            IStreamPubSub pubSub, bool isRewindable, SerializationManager serializationManager,
+            ILoggerFactory loggerFac)
```
jdom: Please don't abbreviate in that way, just use `loggerFactory` as param name. There's many occurrences of this

src/OrleansRuntime/Hosting/DefaultSiloServices.cs
```diff
@@ -65,6 +68,8 @@ internal static void AddDefaultServices(IServiceCollection services)
             services.TryAddSingleton<LogConsistencyProviderManager>();
             services.TryAddSingleton<BootstrapProviderManager>();
             services.TryAddSingleton<LoadedProviderTypeLoaders>();
+            services.AddLogging();
+            services.TryAddSingleton<IPEndPoint>(sp => sp.GetService<NodeConfiguration>()?.Endpoint);
```
jdom: Can we avoid the `IPEndPoint` registration, since it's not a service? The object that needs this being injected can just inject `ILocalSiloDetails`, and then access it from `details.SiloAddress.Endpoint`

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
```diff
@@ -31,7 +31,7 @@ public OrleansPerfCounterTelemetryConsumer()
             this.isInitialized = new Lazy<bool>(this.Initialize, true);
             if (!AreWindowsPerfCountersAvailable())
             {
-                logger.Warn(ErrorCode.PerfCounterNotFound, "Windows perf counters not found -- defaulting to in-memory counters. " + ExplainHowToCreateOrleansPerfCounters);
+                LoggerExtensions.Warn(logger, ErrorCode.PerfCounterNotFound, "Windows perf counters not found -- defaulting to in-memory counters. " + ExplainHowToCreateOrleansPerfCounters);
```
jdom: Not sure I understand why the extension method can't still be used. Was this change by design?

src/Orleans/CodeGeneration/InvokeMethodRequest.cs
```diff
@@ -59,7 +59,7 @@ public enum InvokeMethodOptions
 
 
         /// <summary>Obsolete field.</summary>
-        [Obsolete]
+        //TODO: Mark it as [Obsolete] after all runtime has migrated
```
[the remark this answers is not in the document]
xiazen: will revert this. was a mistake

src/Orleans/Logging/ILoggerExtensions.cs
xiazen: I was under the impression that event id can be event sequence id or errorcode, which indicating an event type. So I left event id as 0, just in case we want to use it as sequence id in the future. But I maybe under a wrong impression. will change it.

src/Orleans/Logging/LoggerWrapper.cs
xiazen: LoggerWrapper is a wrapper on top of Microsoft.Extensions.Logging.ILogger, which doesn't share similarity with LoggerImpl. So I cannot see why it help its history. 
And I cannot make it as LoggerImpl as least now, since it is part of the old logging infra, which the remaining code base is still using. Until I completely migrate the runtime to new asp.net abstraction, LoggerImpl will stay as is. That is my current plan

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
xiazen: the ILoggerExtension.cs I introduced brought compiling ambitious to OrleansTelemetryConsumer proj, hence this change to call the extension method in a more explicit way. Maybe you know how to fix this?

src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducerExtension.cs
```diff
@@ -23,16 +25,16 @@ internal class SimpleMessageStreamProducerExtension : IStreamProducerExtension
         private readonly IStreamPubSub              streamPubSub;
         private readonly bool                       fireAndForgetDelivery;
         private readonly bool                       optimizeForImmutableData;
-        private readonly Logger                     logger;
+        private readonly ILogger                     logger;
 
-        internal SimpleMessageStreamProducerExtension(IStreamProviderRuntime providerRt, IStreamPubSub pubsub, bool fireAndForget, bool optimizeForImmutable)
+        internal SimpleMessageStreamProducerExtension(IStreamProviderRuntime providerRt, IStreamPubSub pubsub, ILoggerFactory loggerFac, bool fireAndForget, bool optimizeForImmutable)
         {
```
jdom: There's several occurrences of loggerFac sill

src/Orleans/Logging/LoggingUtils.cs
```diff
@@ -0,0 +1,25 @@
...
+        ///In order to user message bulking feature, Application needs to use this method to create EventId, 
+        /// and also build their loggerProvider on top of OrleansLoggerDecorator 
+        public static EventId CreateEventId(int errorCode)
+        {
+            return new EventId(errorCode, LogCodeString);
+        }
+
+        internal static EventId CreateEventId(ErrorCode errorCode)
+        {
+            return new EventId((int)errorCode, LogCodeString);
```
jdom: I would just drop passing "OrleansLogCode" to every EventId, as it adds no value. At most, I would pass `errorCode.ToString()` if we want to preserve the name (we were not doing that in the past, but it might be an OK improvement, not sure about the side effects).
In the future, we should probably just replace this ErrorCode enum with a class that has all the EventIds as static members.

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
jdom: Not sure I follow. I downloaded your branch to try this, and just reverting the changes to this file still compiles

src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducerExtension.cs
xiazen: will fix it. I thought that was the only occurrence

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
xiazen: hmm on my machine vs 17 cannot find the exact method. It cannot decide ILogger.Warn or Logger.Warn to use.... 

 maybe I try again,

xiazen:
since migrate logging to Microsoft.Extensions.Logging in different PRs will make logging much harder and confusing to use in the transitional period, so we decided to migrate it in one PR, but leave well separated commit to provide convenience for pr reviewers

xiazen:
current diff will be squashed into more defined commits.

src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducerExtension.cs
```diff
@@ -42,7 +44,7 @@ internal void AddStream(StreamId streamId)
             // so this call is only made once, when StreamProducer is created.
             if (remoteConsumers.TryGetValue(streamId, out obs)) return;
 
-            obs = new StreamConsumerExtensionCollection(streamPubSub, logger);
+            obs = new StreamConsumerExtensionCollection(streamPubSub, providerRuntime.ServiceProvider.GetRequiredService<ILoggerFactory>());
             remoteConsumers.Add(streamId, obs);
```
jdom: loggerFactory is being injected in the constructor already, so just pass that one

src/Orleans/Logging/LoggerWrapper.cs
```diff
@@ -0,0 +1,182 @@
...
+                case Severity.Verbose: return LogLevel.Debug;
+                default: return LogLevel.Trace;
+            }
+        }
+        private Severity FindSeverityForLogger(ILogger logger)
+        {
+            //traversal from the lowest LogLevel to the highest to find the Severity of current Logger
+            //If Trace is enabled, then minimun enabled LogLevel is Trace, which maps to Severity being Verbose2
+            if (logger.IsEnabled(LogLevel.Trace))
+                return Severity.Verbose2;
```
ReubenBond: Should this be Verbose3?

ReubenBond:
Reviewed up until this point. LGTM with 1 q: should we map Trace to Verbose3. I think yes, otherwise we may as well delete Verbose3 log lines - is that right?

We can squash + continue

src/Orleans/Logging/LoggerWrapper.cs
xiazen: yeah I think you are right, min LogLevel is Trace means enabling all loglevel, whose counterpart should be Verbose3

src/Orleans/Logging/ILoggerExtensions.cs
```diff
@@ -0,0 +1,274 @@
...
+
+            // Kill process
+            if (Debugger.IsAttached)
+            {
+                Debugger.Break();
+            }
+            else
+            {
+                logger.Error(ErrorCode.Logger_ProcessCrashing, "INTERNAL FAILURE! Process crashing!");
+                LogManager.Close();
```
ReubenBond: Do we need this?

src/Orleans/Streams/SimpleMessageStream/SimpleMessageStreamProducerExtension.cs
xiazen: I will be more careful next time.

src/Orleans/Logging/ILoggerExtensions.cs
xiazen: Nop ~

src/Orleans/Orleans.csproj
```diff
@@ -38,6 +38,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjection)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.1.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="1.1.2" />
```
jdom: Console is no longer necessary here, right?

xiazen: sorry for the bits and pieces of uncleanness. I did another round of clean up..

xiazen:
is current code good? Can I rebase those commits into one?

src/Orleans/Logging/LoggerWrapper.cs
```diff
@@ -0,0 +1,179 @@
...
+    {
+        public override Severity SeverityLevel => this.maxSeverityLevel;
+
+        public override string Name => this.name;
+        private string name;
+        private readonly Severity maxSeverityLevel;
+        private readonly ILogger logger;
+        private readonly IPEndPoint ipEndpoint;
+        private readonly ILoggerFactory loggerFactory;
+        public LoggerWrapper(string name, ILoggerFactory loggerFactory, IPEndPoint ipEndpoint)
```
jdom: IS IPEndPoint ever passed into this constructor? It can probably be removed, right?

xiazen: Yes, it will be passed in if there's one in DI. In current PR client side doesn't have ipEndPoint concept, that's why it is null. But ipEndPoint is needed there to achieve legacy logging message style

jdom: We synced in person. We decided that LoggerWrapper should not try to pre-format everything as the original LogConsumer sinks. It will just forward to ILogger and drop the IPEndPoint and manage thread id, and all that contextual info.
If the end user is also using the other back compat feature (added in the other PR) which keeps using `ILogConsumer`, then that contextual information will be added at that point. But if the end user configures the new `ILogProvider` sinks from MEL instead, then it's up to those sinks to enhance with whatever contextual information it wants.

xiazen:
don't review yet, I just want to run some tests

Igorbek:
That's awesome that Orleans is taking this abstraction.
I'd like to clarify a few things in regards to logging that I'm highly interested in.
Microsoft's logging abstraction allows to define scopes within the logger, where some additional data being added implicitly:
```cs
using (logger.BeginScope("Job {jobId}", jobId))
{
  logger.Log("Job starts"); // implicitly has jobId
}
```
That is extremely useful when these contexts are created on infrastructure level, to always have correlation with other events. For example I use it in the following ways in asp.net core applications:
- have request to have scope with request id, (and any additional request information)
- capture request correlation id from request header (all outgoing request also send this id, so all requested resources can be correlated)
- have scope with activated controller/action/parameters

So my questions regarding possibility to automatically have logs augmented with valuable information:
- logging within a silo could have: deployment id, server id
- logging within a grain activation could have: grain identity (type, id)
- logging with a grain method could have: action invoked, parameters, request id, client info (parent correlation id), was it remote call and so on

Ideally, should be a way to extend this functionality by being able to inject into lifetime stages and having access to logger at the same time (so all this functionality could be implemented).

Igorbek:
I probably should have read the docs before posting :) It looks like all this can be implemented already with [interceptors](http://dotnet.github.io/orleans/Documentation/Advanced-Concepts/Interceptors.html) and [`RequestContext`](http://dotnet.github.io/orleans/Documentation/Advanced-Concepts/Request-Context.html). However, I need to have the direct access to logger provider in order to inject contextual data (Serilog allows so).

jdom:
@Igorbek absolutely. One of the goals of moving to this established abstraction is so that we can get advantage of more and more features, and some correlation absolutely makes sense. Once we are fully migrated we can start adding those, and we will welcome your feedback for the design (and even code if you want to send a PR). Regarding deployment id and server ID, not entirely familiar, but I would think those would be something that can be configured in the logging providers, since they apply to the entire app, and are not really part of a small scoped context. Still, we will welcome feedback there, as you are probably a lot more familiar with it than me at least

src/Orleans/Messaging/IGatewayListProvider.cs
```diff
@@ -15,8 +15,7 @@ public interface IGatewayListProvider
         /// Initializes the provider, will be called before all other methods
         /// </summary>
         /// <param name="clientConfiguration">the given client configuration</param>
-        /// <param name="logger">the logger to be used by the provider</param>
-        Task InitializeGatewayListProvider(ClientConfiguration clientConfiguration, Logger logger);
+        Task InitializeGatewayListProvider(ClientConfiguration clientConfiguration);
```
xiazen: removed Logger from IGatewayListProvider.Init. now it gets logger from DI in constructor injection

src/Orleans/SystemTargetInterfaces/IReminderTable.cs
```diff
@@ -15,7 +14,7 @@ namespace Orleans
     /// </summary>  
     public interface IReminderTable
     {
-        Task Init(GlobalConfiguration config, Logger logger);
+        Task Init(GlobalConfiguration config);
```
xiazen: removed Logger from IReminderTable.Init . now it gets logger from DI in constructor injection

src/Orleans/SystemTargetInterfaces/IMembershipTable.cs
```diff
@@ -19,8 +20,7 @@ public interface IMembershipTable
         /// </summary>
         /// <param name="globalConfiguration">the give global configuration</param>
         /// <param name="tryInitTableVersion">whether an attempt will be made to init the underlying table</param>
-        /// <param name="logger">the logger used by the membership table</param>
-        Task InitializeMembershipTable(GlobalConfiguration globalConfiguration, bool tryInitTableVersion, Logger logger);
+        Task InitializeMembershipTable(GlobalConfiguration globalConfiguration, bool tryInitTableVersion);
```
xiazen: removed Logger from IMembershipTable.Init it now gets logger from constructor injection

src/OrleansTestingHost/DefaultSiloBuilderFactory.cs
```diff
@@ -8,7 +12,10 @@ public sealed class DefaultSiloBuilderFactory : ISiloBuilderFactory
         public ISiloBuilder CreateSiloBuilder(string siloName, ClusterConfiguration clusterConfiguration)
         {
             ISiloBuilder builder = new SiloBuilder();
-            return builder.ConfigureSiloName(siloName).UseConfiguration(clusterConfiguration);
+            return builder.ConfigureSiloName(siloName)
+                .UseConfiguration(clusterConfiguration)
+                .ConfigureLogging(loggingBuilder => TestingUtils.ConfigureDefaultLoggingBuilder(loggingBuilder,
```
jason-bragg: There are tests that don't use the default silo builder factory, but instead provide their own.  Can you search for all impls of ISiloBuilderFactory and ensure they all configure the logging.

src/ClientGenerator/ClientGenerator.cs
```diff
@@ -102,6 +103,28 @@ private static bool CreateGrainClientAssembly(CodeGenOptions options)
...
+        /// <summary>
+        /// Create the default logger factory, which would create <see cref="Microsoft.Extensions.Logging.ILogger"/>> that writes logs to <paramref name="filePath"/> and console.
+        /// This is to restore legacy default behavior of LogManager, which configure LogManager with a FileTelemetryConsumer and ConsoleTelemetryConsumer
+        /// by default;
+        /// </summary>
+        /// <param name="filePath">the logger file path</param>
+        /// <returns></returns>
+        private static ILoggerFactory CreateDefaultLoggerFactory(string filePath)
+        {
+            var factory = new LoggerFactory(new List<ILoggerProvider>());
```
benjaminpetit: Is the `new List<ILoggerProvider> ()` needed here?

xiazen: not needed.  a mistake when copied from other CreateDefaultLoggerFactory method when it can put in filters param

src/Orleans.Transactions/InClusterTM/ActiveTransactionsTracker.cs
```diff
@@ -20,11 +21,11 @@ internal class ActiveTransactionsTracker : IDisposable
         private long maxAllocatedTransactionId;
         private volatile bool disposed;
 
-        public ActiveTransactionsTracker(IOptions<TransactionsConfiguration> configOption, TransactionLog transactionLog, Factory<string, Logger> logFactory)
+        public ActiveTransactionsTracker(IOptions<TransactionsConfiguration> configOption, TransactionLog transactionLog, ILoggerFactory loggerFactory)
         {
             this.config = configOption.Value;
             this.transactionLog = transactionLog;
-            this.logger = logFactory(nameof(ActiveTransactionsTracker));
+            this.logger = new LoggerWrapper<ActiveTransactionsTracker>(loggerFactory);
```
Igorbek: As I know, DI supports registering open generic types and it is recommended to inject `ILogger<T>` instead of creating a typed logger manually.
Is there a reason why you prefer to have your own typed wrapper?
```cs
public ActiveTransactionsTracker(IOptions<TransactionsConfiguration> configOption, TransactionLog transactionLog, ILogger<ActiveTransactionsTracker> logger)
{
    ...
    this.logger = logger;
    ...
}
```

Igorbek: oh, I guess `LoggerWrapper` implements your previous logger abstraction, right?

xiazen: yes, this is the first step to migrate to ILogger: remove all usage of LogManager. Sorry for the confusion, I should change the description. So Orleans.Runtime.Logger is still used in most places in Orleans codebase, just underlying it will be logging to ILogger. After this step, all Orleans code base will be logging into ILogger technically, but through a proxy of LoggerWrapper

The next step will be remove usage of Logger from Orleans runtime. 

It is arranged this way mostly because the change it too big. Doing it in one PR is too painful for review.

xiazen: But Logger abstraction will still be kept for backward compatibility reasons, in the public Logger GetLogger(string) methods from runtime and grain base class.. Just the runtime will slowly move away from it.

src/ClientGenerator/ClientGenerator.cs
```diff
@@ -102,6 +105,28 @@ private static bool CreateGrainClientAssembly(CodeGenOptions options)
         }
 
         /// <summary>
+        /// Configure logger factory for your client generator
+        /// TODO: fix logging in ClientGenerator, I'm not sure its logging worked properly before, taken that LogManager's state doesn't serialize well across AppDomain
+        /// </summary>
+        public ILoggerFactory LoggerFactory { get; set; } = CreateDefaultLoggerFactory("ClientGenerator.log");
```
jdom: Is this property accessed somewhere else? otherwise just create the factory at the place you need it

src/ClientGenerator/ClientGenerator.cs
```diff
@@ -102,6 +105,28 @@ private static bool CreateGrainClientAssembly(CodeGenOptions options)
         }
 
         /// <summary>
+        /// Configure logger factory for your client generator
+        /// TODO: fix logging in ClientGenerator, I'm not sure its logging worked properly before, taken that LogManager's state doesn't serialize well across AppDomain
+        /// </summary>
```
jdom: What does this mean? it is no longer working?

src/Orleans/Configuration/ConfigUtilities.cs
```diff
@@ -79,85 +79,6 @@ internal static void ParseTelemetry(XmlElement root, TelemetryConfiguration tele
...
-                        }
-                    }
-                    catch (Exception exc)
-                    {
-                        throw new TypeLoadException("Cannot load LogConsumer class " + className + " from assembly " + assembly + " - Error=" + exc);
-                    }
-                }
-            }
-
             SetTraceFileName(config, nodeName, DateTime.UtcNow);
```
jdom: >SetTraceFileName(config, nodeName, DateTime.UtcNow); [](start = 12, length = 52)

If you aren't parsing the TraceFilePattern any more, why is this still here?

src/ClientGenerator/ClientGenerator.cs
xiazen: this is for user to configure their logger factory used for buildetime-codegen. ofc they can set it through constructor. But I was thinking having a setter might be better and more flexible

src/ClientGenerator/ClientGenerator.cs
xiazen: I don't think it worked before with LogManager. Since in CreateGrainClientAssembly call it created an app domain and created the GrainClientGenerator in that app domain. I don't think LogManager's state or LogFactory's state will serialize well across that. So I didn't break something, I just mark out that I don't think the logging worked in CientGenerator before

src/Orleans/Configuration/ConfigUtilities.cs
xiazen: this is for passing trace file pattern from xml config, which load tests is still using. So I'd like to conserve that in this pr so that I don't break load test. We can remove ITraceConfiguration and its infrastructure in my next pr/

src/Orleans/Logging/FileLogger.cs
```diff
@@ -0,0 +1,155 @@
...
+                var now = DateTime.UtcNow;
+                if (now - this.lastFlush > flushInterval)
+                {
+                    this.lastFlush = now;
+                    this.logOutput.Flush();
+                }
+            }
+        }
+
+        private static string FormetMessage(
```
jdom: There is a typo here. But wasn't this copy/pasted? The original code didn't have the typo, so I'm wondering if there are more changes...

src/Orleans/AssemblyLoader/AssemblyProcessor.cs
```diff
@@ -52,9 +54,10 @@ internal class AssemblyProcessor : IDisposable
         /// <param name="codeGeneratorManager">
         /// The code generator.
         /// </param>
-        public AssemblyProcessor(TypeMetadataCache typeCache, SerializationManager serializationManager, CodeGeneratorManager codeGeneratorManager)
+        /// <param name="loggerFactory"> the logger factory used to create loggers</param>
+        public AssemblyProcessor(TypeMetadataCache typeCache, SerializationManager serializationManager, CodeGeneratorManager codeGeneratorManager, ILoggerFactory loggerFactory)
         {
-            this.logger = LogManager.GetLogger("AssemblyProcessor");
+            this.logger = new LoggerWrapper<AssemblyProcessor>(loggerFactory);
```
jdom: This API looks clunky. Can we have an extension method on loggerFactory instead? Such as:
this.logger = loggerFactory.GetLegacyLogger<AssemblyProcessor>();

Yo can also have one that accepts as string name instead of a generic T.

src/Orleans/Logging/LoggerWrapper.cs
```diff
@@ -0,0 +1,111 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Threading;
+
+namespace Orleans.Runtime
+{
+    //TODO: Mark it as [Obsolete] and internal after all runtime has migrated
+    public class LoggerWrapper<T> : Logger
```
jdom: should this be `LegacyLogger<T>` ?

src/Orleans/Logging/LoggerWrapper.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        public override void Log(int errorCode, Severity sev, string format, object[] args, Exception exception)
+        {
+            this.logger.Log(errorCode, sev, format, args, exception);
+        }
+    }
+
+    //TODO: Mark it as [Obsolete] and internal after all runtime has migrated
+    public class LoggerWrapper : Logger
```
jdom: This could be internal today, if the onlt wat to create one of these is through an extension method from loggingFactory, or via DI using `LegacyLogger<T>` (currently `LegacyWrapper<T>`)

src/Orleans/Messaging/GatewayProviderFactory.cs
```diff
@@ -13,11 +14,11 @@ internal class GatewayProviderFactory
         private readonly IServiceProvider serviceProvider;
         private readonly Logger logger;
 
-        public GatewayProviderFactory(ClientConfiguration cfg, IServiceProvider serviceProvider)
+        public GatewayProviderFactory(ClientConfiguration cfg, IServiceProvider serviceProvider, ILoggerFactory loggerFactory)
```
jdom: I see that most classes inject ILoggerFactory and then get the logger from it, instead of injecting `LegacyLogger<T>`. Is it because they are not using DI? because injecting that one would be a lot more preferable where possible

src/Orleans.Transactions/InClusterTM/ActiveTransactionsTracker.cs
jdom: Still, as I also mentioned in another comment, we should register `LoggerWrapper<T>` (actually `LegacyLogger<T>` is my preference as a name) as an open generic and avoid using the factory all over the place.

src/ClientGenerator/ClientGenerator.cs
jdom: Not sure I follow the expected usage, since this is invoked as an `.exe` by MSBuild

src/Orleans/Logging/FileLogger.cs
xiazen: it was copied from Microsoft.Orleans.Logging.Legacy.OrleansLoggingUtils
u can do a diff checker online tool to findout the diff between them. Basically I removed "logging loggerType, ipEndpoint" from it.

src/Orleans/AssemblyLoader/AssemblyProcessor.cs
xiazen: it has to be a generic T since we want the category to be the class's fullName for logFiltering to work properly. I'd consider an extension method. But I'd rather not using it in our internal code, since that's a lot of changes to this pr. And since this pr is really transitional code anyways. Once the core moved away from Logger. then we won't be using LoggerWrapper in core anymore.

xiazen: I already have one that accepts string as name btw. LoggerWrapper<<'T'>> is built on top of that one LoggerWrapper. But again, I prefer to use oggerWrapper<<'T'>> here

src/Orleans/Logging/LoggerWrapper.cs
xiazen: why? application will not be interacting with this directly . they will be interacting with Logger. 
What's your concern about its name?

src/Orleans/Logging/LoggerWrapper.cs
xiazen: yes it can be internal eventually. But right now all extensions projects are using LoggerWrapper too. So this cannot be internal in this PR. If all extensions projects moved away from Logger and LoggerWrapper. then this can be marked as internal

src/Orleans/Messaging/GatewayProviderFactory.cs
xiazen: ideally we just want to inject loggerFactory into DI, and get ILogger or ILoggerFactory from DI right? why are we injecting LegacyLogger<T>?

src/Orleans/AssemblyLoader/AssemblyProcessor.cs
jdom: sure, it's fine to use the generic version. What I'm saying is have the extension method, since our usage will be only temporal (hopefully), but the API is not, and will live longer in the legacy package. So since the API is clunky for our internal usage, it will be clunky for our customers too.

src/Orleans/Logging/LoggerWrapper.cs
jdom: expecting users to use a Wrapper (when it's not entirely clear wrapper for what) feels bad. On the other hand, this is the Logger wrapper we keep for legacy reasons, so that's why I suggested that name.
Alternatively I propose that this is renamed to `Logger<T>`, and also that we follow suit and have an extension method `loggingFactory.CreateLegacyLogger<T>()` that returns a Logger. The user never has to know the implementation details of how we wrap things

src/Orleans/Logging/LoggerWrapper.cs
jdom: Right, but if we use the extension method approach (or `Logger<T>` in case of DI), then this wouldn't even need to exist today as a public API.

src/Orleans/Messaging/GatewayProviderFactory.cs
xiazen: we are moving away from LegacyLogger. why injectting it will be preferable?

jdom: because it's the expected usage even when we move to `ILogger<T>`. There we won't inject LoggingFactory just to get the ILogger instance, we rely on DI. Why not with our legacy logger too?

jdom:
Synced up in person with @xiazen regarding the LoggerWrapper. I was under the impression we would provide users with a way to create a legacy Logger from the logging factory in the back-compat package, but I guess we will only provide the extension method to get a logger from grain and provider runtime, and no other way to get a legacy logger. In that case, having a clunky API while we migrate our internals to use ILogger is acceptable. So please ignore comments related to that aspect.

src/Orleans/Runtime/CallbackData.cs
```diff
@@ -29,8 +30,8 @@ internal class CallbackData : ITimebound, IDisposable
         private TimeSpan timeout;
         private SafeTimer timer;
         private ITimeInterval timeSinceIssued;
-        private static readonly Logger logger = LogManager.GetLogger("CallbackData");
-
+        private readonly Logger logger;
+        private readonly ILoggerFactory loggerFactory;
```
jdom: Hmm, do we really want to allocate these 2 fields (1 of which it's not entirely trivial to initialize, and the other used to initialize the SafeTimer also in a non-trivial way) on each CallbackData? this may have a perf impact.
Thoughts @ReubenBond @sergeybykov ?

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -127,23 +128,20 @@ internal void ConsumeServices(IServiceProvider services)
...
-
             BufferPool.InitGlobalBufferPool(config);
 
 
             try
             {
                 LoadAdditionalAssemblies();
-                
+                //init logger for UnobservedExceptionsHandlerClass
+                UnobservedExceptionsHandlerClass.InitLogger(this.ServiceProvider.GetRequiredService<ILoggerFactory>());
```
jdom: why not this.loggerFactory?

src/Orleans/Runtime/CallbackData.cs
sergeybykov: I concur - we don't want to allocate a new logger for each `CallbackData`. That would be bad for performance on the hot path.

src/Orleans/Runtime/OutsideRuntimeClient.cs
xiazen: will do . I forget about that variable

src/ClientGenerator/ClientGenerator.cs
xiazen: I'm not entirely sure how it is invoked. let's sync up.

src/Orleans/Runtime/CallbackData.cs
xiazen: we can change them back to be static and set those two fields on silo start up through a public methods or some sort.

xiazen: similar to what I did with   UnobservedExceptionsHandlerClass.InitLogger

xiazen: synced with @jdom  we are taking the approach of pre-creating a Ilogger, cache it in the caller's scope, and feed it to CallBackData when constructing it.

src/Orleans/Async/UnobservedExceptionsHandlerClass.cs
```diff
@@ -26,6 +27,11 @@ static UnobservedExceptionsHandlerClass()
             }
         }
 
+        internal static void InitLogger(ILoggerFactory loggerFactory)
+        {
+            logger = loggerFactory.CreateLogger<UnobservedExceptionsHandlerClass>();
```
jdom: Now that we have different loggers for each instance of client and silos, this whole static class should go away I believe. And we also talked about not logging unobserved exceptions in the client. Not sure it needs to be addressed in this PR though

xiazen: I actually made it non-static and injected it into DI as singleton. But I reverted that change since I wasn't sure how that would affects perf. We can discuss that in another PR to give it more focus

src/OrleansRuntime/Scheduler/TaskWorkItem.cs
```diff
@@ -17,13 +18,15 @@ internal class TaskWorkItem : WorkItemBase
...
         /// <param name="t">Task to be performed</param>
         /// <param name="context">Execution context</param>
-        internal TaskWorkItem(ITaskScheduler sched, Task t, ISchedulingContext context)
+        /// <param name="loggerFactory">logger factory used to create loggers</param>
+        internal TaskWorkItem(ITaskScheduler sched, Task t, ISchedulingContext context, ILoggerFactory loggerFactory)
         {
             scheduler = sched;
             task = t;
             SchedulingContext = context;
+            this.logger = loggerFactory.CreateLogger<TaskWorkItem>();
```
jdom: TaskWorkItem and InvokeWorkItem need the same treatment as CallbackData and have a cached logger passed in, as they are also created VERY frequently (actually even more than the others you already addressed).
ActivationTaskScheduler is not created terribly often, but if it can get the same treatment without too many changes, do that one too.

xiazen: will do

src/OrleansRuntime/Services/GrainService.cs
```diff
@@ -40,16 +41,16 @@ protected GrainServiceStatus Status
...
             throw new Exception("This should not be constructed by client code.");
         }
 
         /// <summary>Constructor to use for grain services</summary>
-        protected GrainService(IGrainIdentity grainId, Silo silo, IGrainServiceConfiguration config) : base((GrainId)grainId, silo.SiloAddress, lowPriority: true)
+        protected GrainService(IGrainIdentity grainId, Silo silo, IGrainServiceConfiguration config, ILoggerFactory loggerFactory) : base((GrainId)grainId, silo.SiloAddress, lowPriority: true, loggerFactory:loggerFactory)
         {
             typeName = this.GetType().FullName;
-            Logger = LogManager.GetLogger(typeName);
+            Logger = new LoggerWrapper<GrainService>(loggerFactory);
```
jdom: This is a Logger that shares with the derived types, and it was using the actual type name before, and I think that should be preserved

src/OrleansRuntime/GrainDirectory/MultiClusterRegistration/GlobalSingleInstanceRegistrar.cs
```diff
@@ -33,14 +34,14 @@ internal class GlobalSingleInstanceRegistrar : IGrainRegistrar<GlobalSingleInsta
 
         public GlobalSingleInstanceRegistrar(
             LocalGrainDirectory localDirectory,
-            Factory<string, Logger> loggerFactory,
+            ILoggerFactory loggerFactory,
```
jdom: nit: can just inject the generic version, as the factory is never used after creating the logger

src/Orleans/Runtime/AsynchAgent.cs
```diff
@@ -161,8 +162,6 @@ private static void AgentThreadProc(Object obj)
             var agent = obj as AsynchAgent;
             if (agent == null)
             {
-                var log = LogManager.GetLogger("RuntimeCore.AsynchAgent");
-                log.Error(ErrorCode.Runtime_Error_100022, "Agent thread started with incorrect parameter type");
```
jdom: Should this throw an exception or something? I don't believe this path is even possible, that's why I would throw an InvalidOperationException if we break that invariant at some point.

src/OrleansRuntime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -363,8 +366,7 @@ private void ProcessAccept(SocketAsyncEventArgs e, bool completedSynchronously)
             }
             catch (Exception ex)
             {
-                var logger = ima != null ? ima.Log : LogManager.GetLogger("IncomingMessageAcceptor", LoggerType.Runtime);
-                logger.Error(ErrorCode.Messaging_IMA_ExceptionAccepting, "Unexpected exception in IncomingMessageAccepter.AcceptCallback", ex);
+                Log.Error(ErrorCode.Messaging_IMA_ExceptionAccepting, "Unexpected exception in IncomingMessageAccepter.AcceptCallback", ex);
```
jdom: I don't fully understand why they had that logic before, maybe because it was a static method long time ago? Regardless, the check for `ima != null` is all over the place. Maybe you should just in case preserve it closer to what it was before:
```
var logger = ima?.Log ?? this.Log;
```

src/OrleansRuntime/ReminderService/InMemoryRemindersTable.cs
```diff
@@ -18,11 +19,11 @@ internal class InMemoryRemindersTable
         // table version, as each read/insert/update should touch & depend on only one row at a time
         //internal TableVersion TableVersion;
 
-        [NonSerialized]
-        private readonly Logger logger = LogManager.GetLogger("InMemoryReminderTable", LoggerType.Runtime);
+        [NonSerialized] private readonly ILogger logger;
 
-        public InMemoryRemindersTable()
+        public InMemoryRemindersTable(ILoggerFactory loggerFactory)
```
jdom: I believe this new version with a non-default constructor can still be serialized without issues, but can @ReubenBond validate that there will be no issues?

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -54,8 +55,7 @@ public class EventHubAdapterFactory : IQueueAdapterFactory, IQueueAdapter, IQueu
         private ConcurrentDictionary<QueueId, EventHubAdapterReceiver> receivers;
         private EventHubClient client;
         private ITelemetryProducer telemetryProducer;
-
-        /// <summary>
+		private ILoggerFactory loggerFactory;        /// <summary>
```
jdom: merge conflict?

src/OrleansServiceFabricUtils/FabricMembershipOracle.cs
```diff
@@ -50,9 +51,9 @@ internal class FabricMembershipOracle : IMembershipOracle, IFabricServiceStatusL
             ILocalSiloDetails localSiloDetails,
             GlobalConfiguration globalConfig,
             IFabricServiceSiloResolver fabricServiceSiloResolver,
-            Factory<string, Logger> loggerFactory)
+            ILoggerFactory loggerFactory)
         {
-            this.log = loggerFactory("MembershipOracle");
+            this.log = new LoggerWrapper<FabricMembershipOracle>(loggerFactory);
```
jdom: nit: just inject the generic version

src/OrleansServiceFabricUtils/OrleansCommunicationListener.cs
```diff
@@ -62,7 +63,7 @@ public OrleansCommunicationListener(ServiceContext context, ClusterConfiguration
             var endpoints = activation.GetEndpoints();
             var siloEndpoint = GetEndpoint(endpoints, SiloEndpointName);
             var gatewayEndpoint = GetEndpoint(endpoints, GatewayEndpointName);
-
+            this.SiloHost = new SiloHostWrapper();
```
jdom: why this change?

src/Orleans/Runtime/AsynchAgent.cs
xiazen: I don't think this path is possible. That's why I removed the logging. But do you think it is worth to throw an exception here?

src/OrleansRuntime/Messaging/IncomingMessageAcceptor.cs
xiazen: sure. that's better

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
```diff
@@ -187,8 +189,7 @@ private bool Initialize()
                 var newPerfCounterData = new List<PerfCounterConfigData>(PerfCounterConfigData.StaticPerfCounters);
 
                 // TODO: get rid of this static access. Telemetry consumers now allow being injected with dependencies, so extract it as such
-                var grainTypes = LogManager.GrainTypes;
-                if (grainTypes != null)
+				var grainTypes = CrashUtils.GrainTypes;                if (grainTypes != null)
```
jdom: hmmm, CrashUtils? I know @ReubenBond is going to avoid this hack in the future, but crash utils? BTW, linefeed broken

src/OrleansServiceFabricUtils/FabricMembershipOracle.cs
xiazen: will do . I was cleaning those "should inject the generic version" stuff in previous commit. I guess I missed MemebershipTable, ReminderTable and GatewayList{Provider, which was implicit created from DI. Wil address these

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
```diff
@@ -61,9 +63,9 @@ public static bool AreWindowsPerfCountersAvailable()
             return false;
         }
 
-        private static PerformanceCounter CreatePerfCounter(string perfCounterName)
+        private static PerformanceCounter CreatePerfCounter(string perfCounterName, ILogger logger)
```
jdom: just make this method non-static

src/OrleansServiceFabricUtils/OrleansCommunicationListener.cs
xiazen: I think it is better to init those fields in the constructor, instead of init those directly from the field.

src/OrleansTelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs
xiazen: CrashUtils is named that because it contains CreateMiniDump method. CrashUtils is not kept there because @ReubenBond 's hack. @ReubenBond can remove his hack from it.

src/OrleansServiceBus/Providers/Streams/EventHub/EventHubAdapterFactory.cs
xiazen: maybe. I can fix those xml comments

src/OrleansRuntime/Services/GrainService.cs
xiazen: good catch

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -22,8 +22,6 @@ public class TestClusterOptions
         /// <summary>Extended options to be used as fallbacks in the case that explicit options are not provided by the user.</summary>
         public class FallbackOptions
         {
-            /// <summary>Gets or sets whether the cluster will output traces in the console by default</summary>
-            public bool? TraceToConsole { get; set; }
```
jdom: hmmm, I'm not so sure about getting rid of this, especially since the default was true... It's not completely straightforward to support though, but something we'll have to start doing for other properties too as we move to mainstream usage of SiloBuilder in TestCluster.
If not addressing yet, please add an issue to bring it back

src/OrleansTestingHost/TestClusterOptions.cs
```diff
@@ -268,16 +260,13 @@ public static ClientConfiguration BuildClientConfiguration(ClusterConfiguration
 
             config.DataConnectionString = clusterConfig.Globals.DataConnectionString;
             config.AdoInvariant = clusterConfig.Globals.AdoInvariant;
-            config.PropagateActivityId = clusterConfig.Defaults.PropagateActivityId;
```
jdom: This is not gone, is it?

xiazen: not gone. will made it back

src/OrleansTestingHost/TestClusterOptions.cs
xiazen: will add an issue

test/NonSilo.Tests/Async_TimingTests.cs
```diff
@@ -13,12 +15,12 @@ namespace UnitTests
...
-        private readonly Logger logger;
+        private readonly ILogger logger;
 
         public Async_TimingTests()
         {
-            LogManager.Initialize(ClientConfiguration.StandardLoad());
-            logger = LogManager.GetLogger("AC_TimingTests", LoggerType.Application);
+            var loggerFactory = TestingUtils.CreateDefaultLoggerFactory(ClientConfiguration.StandardLoad().TraceFileName);
+            logger = loggerFactory.CreateLogger<Async_TimingTests>();
             logger.Info("----------------------------- STARTING AC_TimingTests -------------------------------------");
```
jdom: just get rid of this constructor and logger field

test/NonSilo.Tests/AssemblyLoaderTests.cs
```diff
@@ -5,14 +5,15 @@
...
+using Orleans.TestingHost.Utils;
 using Xunit;
 
 namespace UnitTests
 {
     public class AssemblyLoaderTests 
     {
         const string ExpectedFileName = "OrleansProviders.dll";
-        private readonly Logger logger = LogManager.GetLogger("AssemblyLoaderTests", LoggerType.Application);
+        private readonly Logger logger = new LoggerWrapper<AssemblyLoaderTests>(TestingUtils.CreateDefaultLoggerFactory($"AssemblyLoaderTests.log"));
```
jdom: If you are newing up a LoggingFactory, please implement IDisposable and make sure to uninitialize it

test/NonSilo.Tests/ConfigTests.cs
```diff
@@ -347,58 +186,6 @@ public void ServerConfig_TraceFilePattern_Blank()
...
-
-            LogManager.Initialize(cfg.CreateNodeConfigurationForSilo("Primary"));
-
-            Assert.Collection(LogManager.LogConsumers,
-                lc => Assert.IsAssignableFrom<TelemetryLogConsumer>(lc),
-                lc => Assert.IsAssignableFrom<DummyLogConsumer>(lc));
-        }
-
-        [Fact, TestCategory("Functional"), TestCategory("Config"), TestCategory("Logger")]
-        public void ClientConfig_TelemetryConsumers()
```
jdom: most of the tests removed here (except for very few) are still relevant for the legacy package or for the telemetry consumer abstraction that we are still preserving. Please update how they are set up, but please bring them back, otherwise we loose all coverage

src/OrleansTestingHost/AppDomainSiloHost.cs
```diff
@@ -53,7 +55,8 @@ public class CodeGeneratorOptimizer : MarshalByRefObject
             /// <param name="cachedAssembly">The generated assembly.</param>
             public void AddCachedAssembly(string targetAssemblyName, GeneratedAssembly cachedAssembly)
             {
-                CodeGeneratorManager.AddGeneratedAssembly(targetAssemblyName, cachedAssembly);
+                //TODO: use NullLogger for now. 
+                CodeGeneratorManager.AddGeneratedAssembly(targetAssemblyName, cachedAssembly, NullLogger.Instance);
```
jdom: It's fine to use the null logger here. Remove the TODO

test/NonSilo.Tests/OrleansRuntime/AsyncSerialExecutorTests.cs
```diff
@@ -2,23 +2,25 @@
...
+        public ILogger logger;
         private SafeRandom random;
         public int operationsInProgress;
 
         public AsyncSerialExecutorTests()
         {
-            LogManager.Initialize(new NodeConfiguration());
-            logger = LogManager.GetLogger("AsyncSerialExecutorTests", LoggerType.Application);
+            var loggerFactory = TestingUtils.CreateDefaultLoggerFactory(new NodeConfiguration().TraceFileName);
+            logger = loggerFactory.CreateLogger<AsyncSerialExecutorTests>();
```
jdom: just use ITestOutputHelper instead of the logger, which is a lot more appropriate (this logger usage is probably from before we switched to xUnit). Otherwise you'd have to implement IDIsposable and get rid of the new factory, which is a lot more cumbersome.

test/NonSilo.Tests/SchedulerTests/OrleansTaskSchedulerAdvancedTests.cs
```diff
@@ -46,7 +49,6 @@ public void Dispose()
 
             this.runtimeStatisticsGroup.Dispose();
             this.performanceMetrics.Dispose();
-            LogManager.UnInitialize();
```
jdom: please dispose the logging factory

src/Orleans/Statistics/SchedulerStatisticsGroup.cs
```diff
@@ -36,7 +37,7 @@ internal class SchedulerStatisticsGroup
         private static object lockable;
         private static Logger logger;
 
-        internal static void Init()
+        internal static void Init(ILoggerFactory loggerFactory)
```
jdom: hmmm, this is very bad, as we'll leak the factory/logger into a static field and will survive event after the silo/client goes away. If not addressing right away, log it as an issue we need to fix, similar to the `UnobservedExceptionsHandlerClass` case

test/NonSilo.Tests/SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2.cs
```diff
@@ -25,23 +27,22 @@ public class OrleansTaskSchedulerAdvancedTests_Set2 : IDisposable
...
+            this.performanceMetrics = new SiloPerformanceMetrics(this.runtimeStatisticsGroup, this.loggerFactory);
+            masterScheduler = TestInternalHelper.InitializeSchedulerForTesting(context, this.performanceMetrics, loggerFactory);
         }
         
         public void Dispose()
         {
             masterScheduler.Stop();
             this.performanceMetrics.Dispose();
             this.runtimeStatisticsGroup.Dispose();
-            LogManager.UnInitialize();
```
jdom: Dispose logging factory

test/NonSilo.Tests/SchedulerTests/OrleansTaskSchedulerBasicTests.cs
```diff
@@ -657,23 +658,18 @@ private void LogContext(string what)
...
-            var traceLevels = new[]
-            {
-                Tuple.Create("Scheduler", Severity.Verbose3),
-                Tuple.Create("Scheduler.WorkerPoolThread", Severity.Verbose2),
-            };
-            LogManager.SetTraceLevelOverrides(new List<Tuple<string, Severity>>(traceLevels));
-
+            var filters = new LoggerFilterOptions();
+            filters.AddFilter("Scheduler", LogLevel.Trace);
+            filters.AddFilter("Scheduler.WorkerPoolThread", LogLevel.Trace);
```
jdom: Should it now be "Orleans.Runtime.Scheduler" which is the namespace where these classes live?

test/Tester/SiloInitializationTests.cs
```diff
@@ -72,7 +74,8 @@ private static AppDomain CreateAppDomain()
 
         private static SiloHost CreateSiloHost(AppDomain appDomain, ClusterConfiguration clusterConfig)
         {
-            var args = new object[] { nameof(SiloInitializationIsRetryableTest), clusterConfig };
+            var siloName = nameof(SiloInitializationIsRetryableTest);
+            var args = new object[] {siloName , clusterConfig};
```
jdom: merge conflict? Changes in this file don't look related to the PR

test/Tester/EventSourcingTests/EventSourcingClusterFixture.cs
```diff
@@ -30,15 +34,28 @@ protected override TestCluster CreateTestCluster()
...
+                    .ConfigureSiloName(siloName)
+                    .UseConfiguration(clusterConfiguration)
+                    .ConfigureLogging(builder => ConfigureLogging(builder, clusterConfiguration.GetOrCreateNodeConfigurationForSilo(siloName).TraceFileName));
             }
 
-            return new TestCluster(options);
+            private void ConfigureLogging(ILoggingBuilder builder, string filePath)
+            {
+                TestingUtils.ConfigureDefaultLoggingBuilder(builder, filePath);
+                builder.AddFilter("Storage.MemoryStorage", LogLevel.Debug);
```
jdom: this should now be nameof(MemoryStorage).FullName, right?

test/Tester/StorageFacet/StorageFacetTests.cs
```diff
@@ -31,7 +32,8 @@ public ISiloBuilder CreateSiloBuilder(string siloName, ClusterConfiguration clus
                 {
                     ISiloBuilder builder = new SiloBuilder()
                         .ConfigureSiloName(siloName)
-                        .UseConfiguration(clusterConfiguration);
+                        .UseConfiguration(clusterConfiguration)
+                        .ConfigureLogging(loggingBuilder => TestingUtils.ConfigureDefaultLoggingBuilder(loggingBuilder, clusterConfiguration.GetOrCreateNodeConfigurationForSilo(siloName).TraceFileName));
```
jdom: I think at this point we should introduce a base class or something. Although not sure @jason-bragg might have reservations, because he started there, and then moved away from it. Nevertheless the more stuff we move away from ClusterConfiguration, the more we will probably have to move to each test fixture, so it might make sense to have some defaults being defined somewhere else

test/TesterAzureUtils/AzureGossipTableTests.cs
```diff
@@ -26,7 +27,8 @@ public class AzureGossipTableTests : AzureStorageBasicTests
 
         public AzureGossipTableTests()
         {
-            logger = LogManager.GetLogger("AzureGossipTableTests", LoggerType.Application);
+            var loggerFactory = TestingUtils.CreateDefaultLoggerFactory($"{this.GetType().Name}.log");
```
jdom: implement IDisposable and dispose the factory after the test is done (or just pass the null factory instead).

test/TesterAzureUtils/AzureMembershipTableTests.cs
```diff
@@ -17,22 +19,28 @@ namespace Tester.AzureUtils
...
-            LogManager.AddTraceLevelOverride("OrleansSiloInstanceManager", Severity.Verbose3);
-            LogManager.AddTraceLevelOverride("Storage", Severity.Verbose3);
+        }
+
+        private static LoggerFilterOptions CreateFilters()
+        {
+            var filters = new LoggerFilterOptions();
+            filters.AddFilter("AzureTableDataManager", LogLevel.Trace);
+            filters.AddFilter("OrleansSiloInstanceManager", LogLevel.Trace);
+            filters.AddFilter("Storage", LogLevel.Trace);
```
jdom: are these correct now that you are using the type's full name as the category for the logger?

src/OrleansTestingHost/TestClusterOptions.cs
xiazen: but after this pr. traceToConsole will still be on by default. Since TestCluster is configured with a ConsoleLoggerProvider. But anyways I still created an issue #3429 for tracking,

test/TesterAzureUtils/AzureMembershipTableTests.cs
```diff
@@ -17,22 +19,28 @@ namespace Tester.AzureUtils
     [TestCategory("Membership"), TestCategory("Azure")]
     public class AzureMembershipTableTests : MembershipTableTestsBase, IClassFixture<AzureStorageBasicTests>
```
jdom: I can't find the base class (`MembershipTableTestsBase`) in this PR, but it should have changed I think, as you added a parameter to its constructor.
Also, I wanted to make sure the logging factory is being correctly disposed.

test/TesterAzureUtils/AzureMembershipTableTests.cs
```diff
@@ -17,22 +19,28 @@ namespace Tester.AzureUtils
...
         {
             TestUtils.CheckForAzureStorage();
-            return new AzureBasedMembershipTable();
+            return new AzureBasedMembershipTable(loggerFactory);
         }
 
         protected override IGatewayListProvider CreateGatewayListProvider(Logger logger)
         {
-            return new AzureGatewayListProvider();
+            return new AzureGatewayListProvider(loggerFactory);
```
jdom: You are being passed in the Logger. Maybe you can rework the base class to pass in ILoggerFactory instead if that's what's needed (but can't see the base class in this PR for some reason, so not sure if this is accurate guidance)

test/TesterAzureUtils/SiloInstanceTableManagerTests.cs
```diff
@@ -22,25 +24,23 @@ public class SiloInstanceTableManagerTests : IClassFixture<SiloInstanceTableMana
...
         private readonly ITestOutputHelper output;
 
         public SiloInstanceTableManagerTests(ITestOutputHelper output)
         {
             TestUtils.CheckForAzureStorage();
             this.output = output;
-            logger = LogManager.GetLogger("SiloInstanceTableManagerTests", LoggerType.Application);
+            var loggerFactory = TestingUtils.CreateDefaultLoggerFactory(new NodeConfiguration().TraceFileName);
+                
+            logger = loggerFactory.CreateLogger<SiloInstanceTableManagerTests>();
```
jdom: please create (and then dispose) the logger factory in the Fixture itself (and make it accessible as a property, so that it can be injected in the test class' constructor)

test/TesterAzureUtils/Deployment/AzureSiloTests.cs
```diff
@@ -24,7 +25,7 @@ public class AzureSiloTests
             var config = AzureSilo.DefaultConfiguration(serviceRuntime);
             config.AddMemoryStorageProvider();
 
-            AzureSilo orleansAzureSilo = new AzureSilo(serviceRuntime);
+            AzureSilo orleansAzureSilo = new AzureSilo(serviceRuntime, TestingUtils.CreateDefaultLoggerFactory(config.CreateNodeConfigurationForSilo(nameof(AzureSiloTests)).TraceFileName));
```
jdom: create  the loggerfactory in the constructor, and then implement IDisposable

src/OrleansTestingHost/TestClusterOptions.cs
jdom: And there is no way to turn it off? :(
We had it off in Orleans' tests... all the more reason to bring this feature back

src/Orleans/Statistics/SchedulerStatisticsGroup.cs
xiazen: issue created in #3429

test/Tester/StorageFacet/StorageFacetTests.cs
xiazen: added to the issue list in #3429

test/TesterAzureUtils/Streaming/AQStreamFilteringTests.cs
```diff
@@ -32,7 +34,7 @@ public override void Dispose()
             {
                 var deploymentId = this.HostedCluster?.DeploymentId;
                 base.Dispose();
-                AzureQueueStreamProviderUtils.DeleteAllUsedAzureQueues(StreamProvider, deploymentId, TestDefaultConfiguration.DataConnectionString)
+                AzureQueueStreamProviderUtils.DeleteAllUsedAzureQueues(this.HostedCluster.ServiceProvider.GetService<ILoggerFactory>(), StreamProvider, deploymentId, TestDefaultConfiguration.DataConnectionString)
```
jdom: hmmm, you are accessing the disposed hosted cluster, this should throw... just pass the null logger factory

test/TesterAzureUtils/Streaming/HaloStreamSubscribeTests.cs
```diff
@@ -62,20 +65,21 @@ public override void Dispose()
...
+            this.loggerFactory = fixture.HostedCluster.ServiceProvider.GetService<ILoggerFactory>();
         }
 
         public void Dispose()
         {
             var deploymentId = this.HostedCluster?.DeploymentId;
             if (deploymentId != null && _streamProvider != null && _streamProvider.Equals(AzureQueueStreamProviderName))
             {
-                AzureQueueStreamProviderUtils.ClearAllUsedAzureQueues(_streamProvider, deploymentId, TestDefaultConfiguration.DataConnectionString).Wait();
+                AzureQueueStreamProviderUtils.ClearAllUsedAzureQueues(this.loggerFactory, _streamProvider, deploymentId, TestDefaultConfiguration.DataConnectionString).Wait();
```
jdom: pass the null factory and get rid of the new field

test/TesterInternal/General/LoggerTest.cs
```diff
@@ -1,478 +0,0 @@
...
-
-            LogManager.AddTraceLevelOverride(fullName, Severity.Warning);
-            Assert.Equal(Severity.Warning,  logger.SeverityLevel);  //  "Logger level not reset after override added"
-
-            LogManager.RemoveTraceLevelOverride(fullName);
-            Assert.Equal(initialLevel,  logger.SeverityLevel);  //  "Logger level not reset after override removed"
-        }
-
-        [Fact, TestCategory("Functional"), TestCategory("Logger")]
-        public void Logger_BulkMessageLimit_DifferentLoggers()
```
jdom: none of these tests are useful for the legacy logging package?

test/TesterAzureUtils/AzureMembershipTableTests.cs
xiazen: will update them to the latest

xiazen: but I think they might still work in this way, since loggingFilters wil pick the most matching one to apply.

xiazen: And I think "Storage" is not a class but a namespace "Orleans.Storage"

test/TesterAzureUtils/AzureMembershipTableTests.cs
xiazen: It is odd. But I will make sure the factory is disposing

test/TesterAzureUtils/AzureMembershipTableTests.cs
xiazen: not sure I follow, why should I pass in the logger?

test/TesterAzureUtils/Deployment/AzureSiloTests.cs
xiazen: I switched to use output.WriteLine

test/TesterAzureUtils/SiloInstanceTableManagerTests.cs
xiazen: I switched to use output.WriteLine

test/TesterInternal/General/LoggerTest.cs
xiazen: legacy logging has its own tests. Maybe we can add some message bulking tests from this file to that. But most of these is not relevant I think

test/TesterAzureUtils/AzureMembershipTableTests.cs
xiazen: this loggerFactory is from base class. Since all those tests share similar logging config except the filters. So I only pass in filters in constructor. And base class created the loggerFactory, considering the filters

src/Orleans/Runtime/AsynchAgent.cs
xiazen: I threw an exception

src/OrleansTestingHost/TestClusterOptions.cs
xiazen: well there's. just configure your test without the console logger provider ;) I think we should discuss this. Since I don't know how much orleans wants to deal with trace config from this point further on. We can put it back as a obsolete feature for sure.

test/TesterAzureUtils/SiloInstanceTableManagerTests.cs
```diff
@@ -22,34 +24,29 @@ public class SiloInstanceTableManagerTests : IClassFixture<SiloInstanceTableMana
     {
         public class Fixture
         {
-            public Fixture()
-            {
-                LogManager.Initialize(new NodeConfiguration());
-            }
+            public ILoggerFactory LoggerFactory { get; set; } =
+                TestingUtils.CreateDefaultLoggerFactory(new NodeConfiguration().TraceFileName);
```
jdom: If this is created by the Fixture class, the Fixture class should also implement IDisposable and dispose the factory when the fixture is being disposed

xiazen:
The tests moved in last two commits were from `LoggerTests`.cs, they are all about message bulking. 
The logic changed in `EventBulkingDecoratorLogger` is trying to preserve the original logic of message bulking features, from  LoggImpl.CheckBulkMessageLimits`

test/Tester/Tester.csproj
```diff
@@ -30,4 +30,8 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
```
jdom: shouldn't be included, not sure why your environment is adding these

xiazen: good catch.

xiazen:
can we merge this PR ASAP if you don't have other comments? the longer it sits here the more merge conflicts I deal with everyday.

PR #3334, opened 2017-08-23
Fix tests in VS by upgrading xunit and fixing analyzer errors

ReubenBond:
I'm not certain about the `Build/Tools.csproj` change - can someone advise?

### DIFF
--- a/test/NonSilo.Tests/Serialization/BuiltInSerializerTests.cs
+++ b/test/NonSilo.Tests/Serialization/BuiltInSerializerTests.cs
@@ -338,7 +338,7 @@ namespace UnitTests.Serialization
             var deserialized = OrleansSerializationLoop(environment.SerializationManager, source1);
             ValidateDictionary<string, string>(source1, deserialized, "case-insensitive string/string");
             Dictionary<string, string> result1 = deserialized as Dictionary<string, string>;
-            Assert.Equal<string>(source1["Hello"], result1["hElLo"]); //Round trip for case insensitive string/string dictionary lost the custom comparer
+            Assert.Equal(source1["Hello"], result1["hElLo"]); //Round trip for case insensitive string/string dictionary lost the custom comparer
 
             Dictionary<int, DateTime> source2 = new Dictionary<int, DateTime>(new Mod5IntegerComparer());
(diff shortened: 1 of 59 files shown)


test/TesterInternal/OrleansRuntime/Streams/FixedSizeBufferTests.cs
```diff
@@ -35,7 +35,7 @@ public void EmptyBlockGetSegmentTooLargeBvt()
             Assert.False(buffer.TryGetSegment(TestBlockSize + 1, out segment), "Should not be able to get segement that is bigger than block.");
            Assert.Null(segment.Array);
             Assert.Equal(0, segment.Offset);
-            Assert.Equal(0, segment.Count);
+            Assert.Empty(segment);
```
jdom: Build Failure: Assert.Emtpy would enumerate, and it is not supported if the underlying array is empty. Please revert to previous assertion. Same with FillBlockTestBvt

test/NonSilo.Tests/Serialization/BuiltInSerializerTests.cs
```diff
@@ -415,9 +415,9 @@ public void Serialize_HashSetWithComparer(SerializerToUse serializerToUse)
             Assert.Equal(source1.Count, result.Count); //Count is wrong after round-trip of string hash set with comparer
             foreach (var key in source1)
             {
-                Assert.True(result.Contains(key)); //key is missing after round-trip of string hash set with comparer
+                Assert.Contains(key, result); //key is missing after round-trip of string hash set with comparer
             }
-            Assert.True(result.Contains("One")); //Comparer is wrong after round-trip of string hash set with comparer
+            Assert.Contains("One", result); //Comparer is wrong after round-trip of string hash set with comparer
```
jdom: Build failure: Assert.Contains is probably enumerating and using an equality comparer, where as the original assertion was relying on the HashSet to use CaseInsensitiveStringEquality. Same with Serialize_SortedSetWithComparer. I would just revert all of the assertions that use Contains in this class.

jdom (approved):
LGTM minus the 2 comments for test failures

PR #3337, opened 2017-08-23
Update Typical-Configurations.md

Cronan:
Added section on Reliable Production Deployment Using SQL Server (using programmatic configuration).

### DIFF
--- a/src/Documentation/Deployment-and-Operations/Configuration-Guide/Typical-Configurations.md
+++ b/src/Documentation/Deployment-and-Operations/Configuration-Guide/Typical-Configurations.md
@@ -109,7 +109,7 @@ else
 
 var config = new ClusterConfiguration{    
     Globals =    
-    {        
+    {        _siloName
         DataConnectionString = connectionString,
         DeploymentId = "<your deployment ID>",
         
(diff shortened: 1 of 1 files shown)


sergeybykov:
The storage provider part can be confusing here because it is orthogonal to cluster membership configuration, and is not mentioned in other cases described on the page. I think we better take it out.

sergeybykov:
Likewise, I think the whole logic of handling host name and address can be taken out, to leave the code to rely on the default behavior of silo when `HostNameOrIPAddress` isn't explicitly set. That will limit the scope of the example to what matters here.

Cronan:
Thanks, good points, I'll make the changes. How do I set ProxyGatewayEndpoint programmatically without having an address?

Cronan:
@sergeybykov Should I take out the logging stuff for the same reason as the storage provider part?

sergeybykov:
Sorry, I wasn't clear. I think in this specific doc page we don't need to show how the IP address is obtained. We can just use the `address` variable without assigning a value to it:
```
ProxyGatewayEndpoint = new IPEndPoint(address, 30000);
```

sergeybykov:
>Should I take out the logging stuff for the same reason as the storage provider part?

I think so.

sergeybykov:
Thank you, @Cronan!

PR #3338, opened 2017-08-23
Remove static state in ProviderTypeLoader 

xiazen:
Package static state in ProviderTypeLoader into a separate class LoadedProviderTypeLoader, and inject it into DI. 

Prep work for logging migration

### DIFF
--- a/src/Orleans/Providers/ProviderLoader.cs
+++ b/src/Orleans/Providers/ProviderLoader.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using System.Xml;
 using Orleans.Runtime;
 using Orleans.Runtime.Configuration;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Orleans.Providers
 {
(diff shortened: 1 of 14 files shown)


src/Orleans/Providers/ProviderLoader.cs
```diff
@@ -51,7 +52,7 @@ public void LoadProviders(IDictionary<string, IProviderConfiguration> configs, I
             }
 
             // Load providers
-            ProviderTypeLoader.AddProviderTypeManager(t => typeof(TProvider).IsAssignableFrom(t), RegisterProviderType);
+            ProviderTypeLoader.AddProviderTypeManager(t => typeof(TProvider).IsAssignableFrom(t), RegisterProviderType, serviceProvider.GetRequiredService<LoadedProviderTypeLoaders>());
```
ReubenBond: Is it possible to just inject `LoadedProviderTypeLoaders` into the constructor instead?

xiazen: to ProviderLoader's ctor? or to ProviderTypeLoader's ctor?

ReubenBond: Either way, as long as we're minimising the use of `GetService()` and friends

src/OrleansRuntime/Core/BootstrapProviderManager.cs
```diff
@@ -10,11 +10,11 @@ internal class BootstrapProviderManager : IProviderManager
         private readonly PluginManager<IBootstrapProvider> pluginManager;
         private readonly string configCategoryName;
 
-        internal BootstrapProviderManager()
+        internal BootstrapProviderManager(LoadedProviderTypeLoaders loadedProviderTypeLoaders)
```
ReubenBond: ctor needs to be public

xiazen:

PR #3339, opened 2017-08-23
Change build process to produce binary logs

attilah:
Update build to generate the new binary logs (new MSBuild 15.3 feature)
Update build to generate unique log files per build flavor per config per stage (restore, build, pack)
Modify CI script to archive the binlog files

Info on binary logs: http://msbuildlog.com

### DIFF
--- a/Build-Core.cmd
+++ b/Build-Core.cmd
@@ -38,15 +38,15 @@ call %_dotnet% restore "%CMDHOME%\Build\Tools.csproj" --packages %TOOLS_PACKAGES
 
 SET CURRENT_CONFIGURATION=Debug
 
-call %_dotnet% restore /bl:%LOGFILENAME_PREFIX%-%CURRENT_CONFIGURATION%-Restore.binlog;ProjectImports=Embed /p:Configuration=%CURRENT_CONFIGURATION% "%SOLUTION%"
+call %_dotnet% restore %BUILD_FLAGS% /bl:%LOGFILENAME_PREFIX%-%CURRENT_CONFIGURATION%-Restore.binlog /p:Configuration=%CURRENT_CONFIGURATION% "%SOLUTION%"
 @if ERRORLEVEL 1 GOTO :ErrorStop
 @echo RESTORE ok for %CURRENT_CONFIGURATION% %SOLUTION%
 
-call %_dotnet% build %BUILD_FLAGS% /bl:%LOGFILENAME_PREFIX%-%CURRENT_CONFIGURATION%-Build.binlog;ProjectImports=Embed /p:Configuration=%CURRENT_CONFIGURATION% "%SOLUTION%"
+call %_dotnet% build %BUILD_FLAGS% /bl:%LOGFILENAME_PREFIX%-%CURRENT_CONFIGURATION%-Build.binlog /p:Configuration=%CURRENT_CONFIGURATION% "%SOLUTION%"
 @if ERRORLEVEL 1 GOTO :ErrorStop
 @echo BUILD ok for %CURRENT_CONFIGURATION% %SOLUTION%
 
-call %_dotnet% pack --no-build %BUILD_FLAGS% /bl:%LOGFILENAME_PREFIX%-%CURRENT_CONFIGURATION%-Pack.binlog;ProjectImports=Embed /p:Configuration=%CURRENT_CONFIGURATION%;VersionDateSuffix=%DATE_SUFFIX% "%SOLUTION%"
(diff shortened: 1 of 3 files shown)


jdom:
I'm sold in how good these binary logs are, but it looks like this way we are not outputting anything to the console, so that it could be seen directly in Jenkins or VSO in the portal. It would be nice to have that back.

Build-Core.cmd
```diff
@@ -33,31 +38,31 @@ call %_dotnet% restore "%CMDHOME%\Build\Tools.csproj" --packages %TOOLS_PACKAGES
 
 SET CURRENT_CONFIGURATION=Debug
 
-call %_dotnet% restore /p:Configuration=%CURRENT_CONFIGURATION% "%SOLUTION%"
+call %_dotnet% restore /bl:%LOGFILENAME_PREFIX%-%CURRENT_CONFIGURATION%-Restore.binlog;ProjectImports=Embed /p:Configuration=%CURRENT_CONFIGURATION% "%SOLUTION%"
```
KirillOsenkov: `;ProjectImports=Embed` is not needed as its the default (unless it's on purpose for discoverability)

Build-Core.cmd
```diff
@@ -33,31 +38,31 @@ call %_dotnet% restore "%CMDHOME%\Build\Tools.csproj" --packages %TOOLS_PACKAGES
...
 @if ERRORLEVEL 1 GOTO :ErrorStop
 @echo RESTORE ok for %CURRENT_CONFIGURATION% %SOLUTION%
 
-call %_dotnet% build %BUILD_FLAGS% /p:Configuration=%CURRENT_CONFIGURATION% "%SOLUTION%"
+call %_dotnet% build %BUILD_FLAGS% /bl:%LOGFILENAME_PREFIX%-%CURRENT_CONFIGURATION%-Build.binlog;ProjectImports=Embed /p:Configuration=%CURRENT_CONFIGURATION% "%SOLUTION%"
```
KirillOsenkov: Keep in mind that this script won't work on MSBuild versions before 15.3. If that's OK with you then it's fine.

To preserve compat with all versions of MSBuild, use this instead:
https://www.nuget.org/packages/Microsoft.Build.Logging.StructuredLogger
`/logger:BinaryLogger,packages\Microsoft.Build.Logging.StructuredLogger.1.1.168\lib\net46\StructuredLogger.dll;%LOGFILENAME_PREFIX%-%CURRENT_CONFIGURATION%-Build.binlog`

Build-Core.cmd
```diff
@@ -2,7 +2,7 @@
 setlocal
 
 SET CMDHOME=%~dp0.
-SET BUILD_FLAGS=/m:1 /v:m /fl /flp:logfile=build.log;verbosity=detailed
```
KirillOsenkov: You can bring this back as @jdom mentions, just change the verbosity from `detailed` to `normal` so it's not as noisy.

KirillOsenkov: Unless I'm misunderstanding and the CI system consumes the log via a different channel...

galvesribeiro:
Wow! This is awesome! I wonder if we can contribute a MSBuild xplat log viewer. @KirillOsenkov Is it OSS and the file structure is documented? Maybe some GUI tool based on Electron could do the trick.

ReubenBond:
@galvesribeiro repo here: https://github.com/KirillOsenkov/MSBuildStructuredLog

Build-Core.cmd
attilah: That's ok, we're using .Net Core 2.0 Toolset to build, downloading ourselves.

Build-Core.cmd
attilah: I just checked and we can have /bl and /fl so binlog and normal log also generated.

@jdom would you prefer to add it back and write text log along the binary too with the new naming semantics?

jdom: As we talked in person, binary log and console output (the latter at `normal` verbosity) should be good for our needs

Build-Core.cmd
```diff
@@ -2,7 +2,7 @@
 setlocal
 
 SET CMDHOME=%~dp0.
-SET BUILD_FLAGS=/m:1 /v:m
+SET BUILD_FLAGS=/m:1 /v:m /v:n
```
KirillOsenkov: Did you mean to remove the `/v:m`?

attilah: Ahh yes :-S my bad

jdom:
As we discussed in person, we are moving back to minimal logging to the console, as it looks like that's what we were using before, and it was enough to troubleshoot common issues in our PRs

PR #3348, opened 2017-08-28
Add code generation to OrleansServiceBus

ReubenBond:
Port of #3344 from 1.5.1 to master
Fixes #3347

PR #3350, opened 2017-08-28
Update changelog for 1.5.1

jdom:
After this is merged, I'll update in master also

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,17 @@ All notable end-user facing changes are documented in this file.
 *Here are all the changes in `master` branch, and will be moved to the appropriate release once they are included in a published nuget package.
 The idea is to track end-user facing changes as they occur.*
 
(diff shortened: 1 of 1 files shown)


benjaminpetit:
LGTM

ReubenBond:
LGTM. Do we need a known issue for DI 2.0?

jdom:
I'm not sure we understand it exactly. If the issue is just with the BuildServiceProvider part, then there is a workaround, which is to provide your own Startup type, and explicitly use the 2.0.0 BuildServiceProvider method.

xiazen:
should #3348 be included also?

jdom:
Added (actually #3344 targets the 1.5.1 branch). Thanks.

I'll add a known issue regarding DI 2.0.0 in the release itself, not the changelog anyway

Issue #3353, opened 2017-08-28
Reorganize Orleans packages and project structure


jdom:
For several years we have been wanting to re-structure our assemblies and packages, as a lot of the decisions have been made before the NuGet rise when Orleans was still distributed in an MSI. We decided that Orleans 2.0 is finally the time for us to start that path.

TL;DR version: the biggest most important piece here is to get to a `Microsoft.Orleans.Core.Abstractions` package that we version very infrequently and try not to have breaking changes, and hopefully start applying SemVer practices at least on that one. The rest of the projects is less thought out, and we will improve with time, but the benefits of the Abstractions project heavily outweights any other comments in this issue.

Long version: This is a rough project structure that we came up with, although it might be too detailed for people not very familiar with the Orleans codebase itself:

![image](https://user-images.githubusercontent.com/1361686/29797414-c7bc0faa-8c0b-11e7-91c2-2dc323c3689c.png)

## Core.Abstractions 
  - Should not be revved frequently. Once we get to the point where we want it, it should not be modified frequently, as we should try to minimize breaking changes as much as possible. 
  - Users will mostly reference this package only from the application code, and hopefully only have to upgrade the host/client when we release new versions, but the Abstractions reference stays the same.
  - Should mostly have contracts, but no logic that might require updates/bug fixes.
  - When requiring to do changes to this assembly, extra scrutiny should be paid to the review to avoid breaking changes as much as possible (both compile time as well as runtime binding breaking changes).
  - Apply SemVer practices on this library as a first step (eventually also on other Orleans packages, but the biggest benefit for our users will be in this one).
  - Should have no configuration code. Configuration is the concern of clients and hosts.
  - This is not the place to put shared logic or small non-core utils (for example to share logic between Codegen and Core)
  - Currently requires some contracts for streaming and storage, but hopefully they can be eliminated with scoped activation services in the future (aka Facets).
      - For example, I'd like for `Grain` and `Grain<T>` to not link to storage/streaming stuff, and use scoped services to opt into that functionality (but this is too far out, as `Grain` is rooted strongly throughout the codebase)
  - We do not do build-time codegen on this assembly directly. The generated code for this assembly goes into the Core project.
    - This makes it so that updating codegen, or how we do scanning, etc, does not imply revving the abstractions nuget
    - This also avoids the need for the extremely annyoing bootstrap build step (when building Orleans itself), as the Codegen project should only reference Core.Abstractions and not Core.
  - Serialization considerations
[...]

PR #3355, opened 2017-08-29
Add Orleans.Core.Abstractions project

benjaminpetit:
First step for #3353 

I created the base project, moved `IAddressable` in it and added this project as a reference where needed.

I didn't prefixed with `Microsoft` but should be easily fixable.


src/Orleans.Core.Abstractions/Orleans.Core.Abstractions.csproj
```diff
@@ -0,0 +1,15 @@
...
+  <PropertyGroup>
+    <PackageId>Orleans.Core.Abstractions</PackageId>
+    <Title>Microsoft Orleans Core Abstractions</Title>
+    <Description>Core abstractions library of Microsoft Orleans</Description>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AssemblyName>Orleans.Core.Abstractions</AssemblyName>
+    <RootNamespace>Orleans</RootNamespace>
+    <TargetFramework>netstandard1.5</TargetFramework>
```
jdom: Doesn't it work if it's netstandard1.3? This is so that end users don't require new tooling to reference it, and since there will be no logic, it probably doesn't require anything from either 1.5 or 1.6, which are the biggest offenders in the new .NET Standard world.

src/Orleans.PlatformServices/Orleans.PlatformServices.csproj
```diff
@@ -31,4 +31,8 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.5'">
     <PackageReference Include="System.Runtime.Loader" Version="$(SystemRuntimeVersion)" />
   </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Orleans.Core.Abstractions\Orleans.Core.Abstractions.csproj" />
```
jdom: Not really sure about this one. PlatformServices should probably remain agnostic of anything related to Orleans. So I would remove it from here unless it's absolutely necessary in the future

PR #3356, opened 2017-08-29
Support custom DI container in ClientBuilder and unify with SiloBuilder

ReubenBond:
* ClientBuilder supports custom containers in the same fashion as SiloBuilder
* Extracted common logic from ClientBuilder and SiloBuilder into internal ServiceProviderBuilder
* ClientBuilder adds its default services after the user-configured services. Default services are only added if the user did not configure an implementation for a given service (using TryAdd*) - this unifies it with SiloBuilder

### DIFF
--- a/src/Orleans/Hosting/ServiceProviderBuilder.cs
+++ b/src/Orleans/Hosting/ServiceProviderBuilder.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Orleans.Hosting
@@ -29,8 +30,7 @@ namespace Orleans.Hosting
             // If no service provider factory has been specified, set a default.
             if (this.serviceProviderFactory == null)
             {
-                var factory = new DelegateServiceProviderFactory(svc => svc.BuildServiceProvider());
-                this.UseServiceProviderFactory(factory);
+                this.UseDefaultServiceProviderFactory();
             }
 
             // Create the service provider using the configured factory.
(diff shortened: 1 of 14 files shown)


src/Orleans/Hosting/ServiceProviderBuilder.cs
```diff
@@ -0,0 +1,81 @@
...
+            foreach (var configureServices in this.configureServicesDelegates)
+            {
+                configureServices(services);
+            }
+
+            // If no service provider factory has been specified, set a default.
+            if (this.serviceProviderFactory == null)
+            {
+                var factory = new DelegateServiceProviderFactory(svc => svc.BuildServiceProvider());
+                this.UseServiceProviderFactory(factory);
```
jdom: This default assignment should be done in a separate method that doesn't get inlined, so that it doesn't even attempt to load that type. Probably not a big deal, since the assembly will already be loaded, as we use some extension methods from the non-abstractions library elsewhere

src/OrleansRuntime/Hosting/SiloBuilderExtensions.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using Microsoft.Extensions.DependencyInjection;
+using Orleans.Hosting;
 using Orleans.Runtime.Configuration;
 
 namespace Orleans.Runtime.Hosting
```
jdom: Should you rename the entire namespace?

ReubenBond: yep, I will do that now

src/Orleans/Hosting/ServiceProviderBuilder.cs
ReubenBond: I moved it into a method marked as `NoInlining`

PR #3358, opened 2017-08-29
Orleans transaction core

jason-bragg:
- Exposed grain extensions for public use.
- Transaction system abstractions
- Some minimal implementations.
- Messaging and CodeGen support for transactions.


src/Orleans/Transactions/CommitRecord.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Orleans.Transactions
+{
+    [Serializable]
+    public class CommitRecord
```
ReubenBond: Will this cause build warnings for the lack of a doc comment?

src/Orleans/Transactions/Transaction.cs
```diff
@@ -0,0 +1,53 @@
...
+        // Transactions waiting on the result of this transaction.
+        public HashSet<Transaction> WaitingTransactions { get; private set; }
+
+        // Number of transactions this transaction is waiting for an outcome of.
+        public int PendingCount { get; set; }
+
+        public long HighestActiveTransactionIdAtCheckpoint { get; set; }
+
+        // Time the transaction was completed (i.e. either committed or aborted)
+        public long CompletionTime { get; set; }
```
ReubenBond: It's not clear how this long relates to a time here

src/Orleans/Transactions/TransactionContext.cs
```diff
@@ -0,0 +1,332 @@
...
+
+        public Dictionary<ITransactionalResource, TransactionalResourceVersion> ReadSet { get; }
+        public Dictionary<ITransactionalResource, int> WriteSet { get; }
+        public HashSet<long> DependentTransactions { get; }
+
+        public void Union(TransactionInfo other)
+        {
+            if (TransactionId != other.TransactionId)
+            {
+                // TODO: freak out
```
ReubenBond: Should we add an exception here?

jdom:
Builds fail because the unit test process crashes:
```
14:14:54 Process is terminated due to StackOverflowException.
14:14:54 Error when running tests
14:14:54     + CategoryInfo          : OperationStopped: (Error when running tests:Stri 
14:14:54    ng) [], RuntimeException
14:14:54     + FullyQualifiedErrorId : Error when running tests
14:14:54     + PSComputerName        : localhost
14:14:54  
14:14:54 Failed:  TesterInternal ( Failed )
```

ReubenBond:
StackOverflowException in tests usually indicates a dependency injection cycle

galvesribeiro:
Guys, I know this is a very much anticipated feature and is sort of crucial to other ones (i.e. indexing grains) and you guys had been discussion it with MSR for a while. However, could you please give a very high level overview of the design that ended up on this PR? I don't see an issue describing it so I'm kinda lost on transaction, transactionagent, transactionservice, transactionmanager, etc. etc. I would appreciate some very small topics on what each of those components are suppose to do and how they talk to each other. Also I see it uses some sort of log storage for the transaction log. Do we have a new provider abstraction being introduced here?

Sorry for what could looks like dumb questions but I would appreciate some description so we (from outside) can review and suggest anything.

Thanks!

galvesribeiro:
Ignore my last comment. 

For those interested in review this PR and/or want to know more about the rationale and design behind it, have a look on [this paper](https://www.microsoft.com/en-us/research/publication/transactions-distributed-actors-cloud-2/) from @philbe and Tamer Eldeeb.

jason-bragg:
@galvesribeiro, This set of work definitely deserves more upfront explanation.  I've added an issue, and will work with MSR to see if we can make their design docs public.  Thanks for reminding me that there is a world out here.. been heads down in code.. :)

jason-bragg:

PR #3359, opened 2017-08-30
[Abstractions] Remove dependency from Serialization in UniqueKey

benjaminpetit:
Since this is a potentially dangerous change, please review this CR carefully.

~~This PR will consist of 2 commits:~~

- the first will remove the dependency to serialization in `UniqueKey` by duplicating the existing logic from `BinaryTokenStreamWriter.Write(UniqueKey key)`. I added a test to check that the value returned by the new method is consistent with the previous method.
- ~~the second (to come after this first commit pass the build) will remove the code duplication in `BinaryTokenStreamWriter` by calling the method added in the first commit.~~

~~Basically this will "invert" the dependency.~~

After thoughts, I don't think we should to the 2nd commit, since they are not supposed to share the same logic after all. But I will keep the test to show that the value used for the hash is consistent with the previous method.

### DIFF
--- a/src/Orleans/IDs/UniqueKey.cs
+++ b/src/Orleans/IDs/UniqueKey.cs
@@ -2,7 +2,6 @@ using System;
 using System.Globalization;
 using System.IO;
 using System.Text;
-using Orleans.Serialization;
 
 namespace Orleans.Runtime
 {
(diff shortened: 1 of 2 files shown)


benjaminpetit:

benjaminpetit:

PR #1897, opened 2016-07-02
Fix bug with generic state parameter caused by inconsistent use of grainClassName / genericArgument / genericInterface

Maarten88:
After lots of debugging on an issue I have with a generic grain that uses storage, I think that this change fixes the issue. The genericArgument was used inconsistently, and sometimes contained the genericInterface value. Also the grainClass parameter did not include the genericArguments during setup of the state, causing the grainType parameter in the first ReadStateAsync to be incorrect.

I moved the location where the conversion was made from SetupActivationInstance() into catalog.GetOrCreateActivation(), passed genericArguments in some functions and fixed the naming.

I think this fixes several bugs, including #1579.

Please test thoroughly, this "works on my machine" but the change might do things that I don't really understand and this is my first pull request on orleans.

Issue #1579, opened 2016-03-18
Incomplete grainType when activating generic grains


jasonholloway:
Hello,

It seems that generic grains can never have their state loaded properly on activation, as the various `PlacementDirectors` that produce the `grainType` string to be used in retrieving the grain from storage are all without generic type information for the grain in question, and therefore produce `grainTypes` with generic args missing.

This contrasts with explicitly calling `ReadStateAsync()` and `WriteStateAsync()` from the grain itself, which, via `GrainStateStorageBridge`, produces a fully-specified `grainType` each time.

The upshot is, state is written to one storage key, but loaded from another in the case of activation.

There's a simple enough workaround in calling `ReadStateAsync()` from each generic grain's `OnActivateAsync()` - but this isn't fantastic, obviously.

Does this sound right?

sergeybykov:
Sounds like an embarrassing bug. Need to investigate. Thank you, @jasonholloway!

jasonholloway:
Well, I'm hoping the embarrassing mistake's not mine! But it looks right...

gabikliot:
Can you produce a test that reproduces that problem?

jasonholloway:
Yep, here you go: a fork with an added test method [here](/jasonholloway/orleans/tree/generic-activation-test).

The test method is this: `UnitTests.StorageTests.PersistenceGrainTests_Local.Persistence_Grain_Activate_StoredValue_Generic()`.

jasonholloway:
Just out of interest - would making a pull request with the above be the normal thing to do here?

gabikliot:
Yes, pull request. If the test fails, mark it with skip attribute.

Maarten88:
When testing with a custom storage provider, I see the value of the grainType parameter for a generic grain is different between the ReadStateAsync and the WriteStateAsync functions for the same grain:

    ReadState -> MyProject.Grains.GrainIndexGrain`1
    WriteState -> MyProject.Grains.GrainIndexGrain`1[[MyProject.GrainInterfaces.IApplicationGrain, MyProject.GrainInterfaces, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]

I have not found where this problem originates, but I think the inconsistency of the grainType parameter is what causes this bug.

sergeybykov:
Resolved via https://github.com/dotnet/orleans/pull/1897.

PR #3360, opened 2017-08-30
Reusable storage bridge

jason-bragg:
Refactored storage bridge and stateful grain handling to allow more flexable state management needed by transaction logic.
- Removed grain state managment from catalog and grain creator.
- Grain<> and LogConsistancy grain now setup their own state via the lifecycle.
- GrainStateStorageBridge has been replaced with StateStorageBridge which can be used for transaction and storage facet implementations which choose to store state using storage providers.

Note: With this change, only type managment, and statistics issues prevent event sourcing and stateful grain (grain<>) from being moved out of orleans core/runtime into orleans extension libraries.


src/OrleansRuntime/Catalog/GrainCreator.cs
```diff
@@ -47,56 +38,13 @@ public Grain CreateGrainInstance(IGrainActivationContext context)
...
-
-            if (statefulGrain == null)
-                return grain;
-
-            //Inject state and storage data into the grain
-            statefulGrain.GrainState.State = Activator.CreateInstance(stateType);
-            statefulGrain.SetStorage(storage);
+            // wire up to lifecycle
+            var participant = grain as ILifecycleParticipant<GrainLifecycleStage>;
+            participant?.Participate(context.ObservableLifecycle);
```
jason-bragg: Moving state management logic out of catalog and grain-creator into grain base classes that participate in the lifecycle is the bulk of this change.  While we should be moving away from base classes, this preserves the existing behaviors and moves this logic out of the catalog/grain-creator and provides a more extensible pattern for future features.

src/Orleans/Core/Grain.cs
```diff
@@ -332,5 +324,51 @@ protected virtual Task ReadStateAsync()
         {
             return storage.ReadStateAsync();
         }
+
+        public void Participate(ILifecycleObservable<GrainLifecycleStage> lifecycle)
+        {
+            lifecycle.Subscribe(GrainLifecycleStage.SetupState, this);
+        }
+
+        public async Task OnStart(CancellationToken ct)
```
ReubenBond: suggest we make this an explicit implementation so as not to pollute intellisense for Grain

src/OrleansRuntime/Hosting/DefaultSiloServices.cs
```diff
@@ -61,8 +63,15 @@ internal static void AddDefaultServices(IServiceCollection services)
             services.TryAddTransient(typeof(IStreamSubscriptionObserver<>), typeof(StreamSubscriptionObserverProxy<>));
 
             services.TryAddSingleton<StatisticsProviderManager>();
+
             services.TryAddSingleton<StorageProviderManager>();
+            services.TryAddSingleton<IKeyedServiceCollection<string, IStorageProvider>>(sp => sp.GetRequiredService<StorageProviderManager>());
```
ReubenBond: Doesn't need to change in this PR, but there's an extension method called `TryAddFromExisting` which essentially does this.

ReubenBond:
One change to consider is that when a lifecycle participant fails, propagate the original exception to the caller instead of an `OperationCanceledException`.

Highlighted in the failure of `Persistence_Grain_Activate_Error`:
`MESSAGE:
Assert.Throws() Failure\r\nExpected: typeof(UnitTests.StorageTests.StorageProviderInjectedError)\r\nActual: typeof(System.OperationCanceledException): Lifecycle start canceled due to errors at stage SetupState`

That change had no effect until this PR, but I believe it was already in place.

src/Orleans/Core/Grain.cs
```diff
@@ -332,5 +324,51 @@ protected virtual Task ReadStateAsync()
...
+            lifecycle.Subscribe(GrainLifecycleStage.SetupState, this);
+        }
+
+        async Task ILifecycleObserver.OnStart(CancellationToken ct)
+        {
+            if (ct.IsCancellationRequested)
+                return;
+            IStorageProvider storageProvider = GetStorageProvider();
+            string grainTypeName = this.GetType().FullName;
+            this.storage = new StateStorageBridge<TGrainState>(this.GetType().FullName, this.GrainReference, storageProvider);
```
jdom: you probably meant to use `grainTypeName` as a parameter here

src/Orleans/Core/Grain.cs
```diff
@@ -332,5 +324,51 @@ protected virtual Task ReadStateAsync()
         {
             return storage.ReadStateAsync();
         }
+
+        void ILifecycleParticipant<GrainLifecycleStage>.Participate(ILifecycleObservable<GrainLifecycleStage> lifecycle)
```
jdom: Implementing this explicitly would mean that app devs (or infrastructure extensions) will not be able to create a base class that is a participant and hook it up as expected, correct? Maybe it should just be protected virtual instead?

src/Orleans/Core/Grain.cs
```diff
@@ -332,5 +324,51 @@ protected virtual Task ReadStateAsync()
...
+        {
+            lifecycle.Subscribe(GrainLifecycleStage.SetupState, this);
+        }
+
+        async Task ILifecycleObserver.OnStart(CancellationToken ct)
```
jdom: maybe the `ILifecycleObserver` itself can be instantiated on the `Participate` phase, so that the grain itself doesn't have to implement `ILifecycleObserver` that could conflict with how the users implement their grains?
For example the `StateStorageBridge` itself could implement that interface, or an inner class of it, or even have it implicitly implemented with a lambda that can use closures.

src/Orleans/Core/Grain.cs
```diff
@@ -332,5 +324,51 @@ protected virtual Task ReadStateAsync()
...
+        {
+            if (ct.IsCancellationRequested)
+                return;
+            IStorageProvider storageProvider = GetStorageProvider();
+            string grainTypeName = this.GetType().FullName;
```
jdom: I think this is not backwards compatible. In the past we were using `$"{grainType}[{genericArguments}]` as the state name, where grain type is not even `type.FullName` or something like that.

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -1349,6 +1142,11 @@ private void RerouteAllQueuedMessages(ActivationData activation, ActivationAddre
             {
                 RequestContext.Import(requestContextData);
                 await activation.Lifecycle.OnStart();
+                // TODO: Consider pre and post activation stages in lifecycle for PreActivateProtocolParticipant and PostActivateProtocolParticipant like behaviors
+                if (activation.GrainInstance is ILogConsistencyProtocolParticipant)
+                {
+                    await ((ILogConsistencyProtocolParticipant)activation.GrainInstance).PreActivateProtocolParticipant();
```
jdom: question: shouldn't this happen before calling `activation.Lifecycle.OnStart();`

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -1349,6 +1142,11 @@ private void RerouteAllQueuedMessages(ActivationData activation, ActivationAddre
             {
                 RequestContext.Import(requestContextData);
                 await activation.Lifecycle.OnStart();
+                // TODO: Consider pre and post activation stages in lifecycle for PreActivateProtocolParticipant and PostActivateProtocolParticipant like behaviors
```
jdom: Yes, I was even considering that the `Grain.OnActivateAsync` could actually be implicitly hooked up by the `Grain` base class being a lifecycle participant, not as a special method that we handle separately in the catalog.
This way, we get closer to the goal of supporting POCO objects as grains in the future, if all that is in `Grain` can be implemented by scoped activation services and lifetime observers.

src/Orleans/Core/Grain.cs
jason-bragg: hmm.. looks like there was  bug..?  We load the initial sate using what you describe, but the storage bridge was setup with a different name, meaning the initial loaded state is different from the state the bridge uses.  Not sure how that got past tests, but that's a serious bug, unless I'm misunderstanding what I'm reading.

src/OrleansRuntime/Catalog/Catalog.cs
jason-bragg: The expected behavior is that state be loaded then pre-activate, activate, post-activate.  Since state load happens in lifecycle, this will need to be after onStart, for now.  When onActivateAsync is called by the lifecycle, this can be as well.

src/OrleansRuntime/Catalog/Catalog.cs
jason-bragg: That is the plan.  All of grain lifecycle should be handled within the lifecycle framework.  We've just not refactored it all yet.  A good next step would be to add pre-activate, activate, and post-activate stages to the lifecycle, and move the onactivate/deavtivate within the lifecycle (as you've describe) along with the pre/post activate logic.  Though, I'm not convinced we need the pre/post activate, so I'd like to understand why that was needed.

src/Orleans/Core/Grain.cs
jason-bragg: Change was introduced in July of last year, so it may be lower impact than I'd expected.   https://github.com/dotnet/orleans/commit/f634bde01f8d8fac696c0086986a1bbb61815cd8
Anyone familiar with what this was for?  @ReubenBond? @jdom?

jason-bragg: Ah.  I see.  This addressed a problem which only existed because the bridge was not being used to read the initial state.  Current code should be fine.

src/Orleans/LogConsistency/LogConsistentGrain.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// (in which case user MUST configure a consistency provider)
+        /// </summary>
+        protected abstract ILogViewAdaptorFactory DefaultAdaptorFactory { get; }
+
+        public virtual void Participate(ILifecycleObservable<GrainLifecycleStage> lifecycle)
+        {
+            lifecycle.Subscribe(GrainLifecycleStage.SetupState, OnSetupState);
+        }
+
+        Task OnSetupState(CancellationToken ct)
```
jdom: nit: add the private modifier

src/OrleansRuntime/Catalog/Catalog.cs
```diff
@@ -1376,13 +1178,16 @@ private void RerouteAllQueuedMessages(ActivationData activation, ActivationAddre
 
                 activationsFailedToActivate.Increment();
 
+                /// TODO: During lifecycle refactor discuss with team whether activation failure should have a well defined exception, or throw whatever
+                ///   exception caused activation to fail, with no indication that it occured durring activation
+                ///   rather than the grain call.
+                OrleansLifecycleCanceledException canceledException = exc as OrleansLifecycleCanceledException;
+                if(canceledException?.InnerException != null)
+                {
+                    throw canceledException.InnerException;
```
jdom: This would completely wipe out the original stack trace of the exception, which is very important for troubleshooting what went wrong. You might want to do the following instead:
```
ExceptionDispatchInfo.Capture(canceledException.InnerException).Throw();
```

test/TesterInternal/ErrorInjectionStorageProvider.cs
```diff
@@ -53,6 +48,14 @@ public override string Message
                 return "ErrorInjectionPoint=" + Enum.GetName(typeof(ErrorInjectionPoint), errorInjectionPoint);
             }
         }
+
+#if !NETSTANDARD
```
jdom: Please undo this change after rebasing on top of `master`, since we removed all NETSTANDARD conditional compilation statements in #3368 that was just merged

jason-bragg:

src/Orleans/Core/GrainExtensions.cs
```diff
@@ -223,6 +228,20 @@ public static string GetPrimaryKeyString(this IGrainWithStringKey grain)
                 throw new InvalidOperationException("Invoking of methods with one way flag must result in completed task");
             }
         }
+        
+        public static IStorageProvider GetStorageProvider(this Grain grain, IServiceProvider services)
```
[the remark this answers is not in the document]
jdom: At the very least I would constrain it to `Grain<T>`, not every `Grain`, but I think I prefer in something different than GrainExtensions that doesn't pollute the intellisense (or even not provide an extension method for this).
As you mentioned it's probably only going to be used by some infrastructure code (even just from us). And as we endorse facets, this is even less relevant.

jason-bragg:
Squashed rebased.
Addressed all comments.

src/Orleans/Core/GrainExtensions.cs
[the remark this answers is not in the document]
jdom: probably missed addressing this one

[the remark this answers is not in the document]
jason-bragg: Didn't fix because there was nothing broken.  Moved to namespace Orleans.Storage anyway, so it won't show up unless someone is already using storage features.

jason-bragg:
ping

jdom:
Oh, sorry, didn't merge as I thought you found a big issue with the name we use as state key. Was this resolved?

ReubenBond:
Oh, I thought that was the existing behavior which was just being carried forwards

jdom:
If that really existed, it would be a huge issue that needs to be investigated and potentially addressed asap (not just for 2.0). That's why I thought he was still investigating it, because it's crazy that no one encountered this before

jdom:
@jason-bragg can you confirm?

jason-bragg:
TL,DR: The original code was correct.  The new code is correct.

Full description of original problem: "Incomplete grainType when activating generic grains #1579"
Fix: "Fix bug with generic state parameter caused by inconsistent use of grainClassName / genericArgument / genericInterface #1897"

The difference in state name generation in the catalog vs. the bridge is due to the fact that the catalog used the type from the grain type data which did not have the generic value which is only known at runtime.  The bridge did not need to deal with this problem because it used the fully qualified runtime grain type.  By removing the (duplicate) load logic from the catalog, we no longer use the unqualified type from the grain type data, so the issue goes away.

PR #3368, opened 2017-08-30
Update Current to NetStandard 2.0

attilah:
- Update Current to NetStandard 2.0
- Test Projects targeting net461
- NETSTANDARD and NETSTANDARD_TODO is not defined anymore to reduce previous workarounds
- BUILDFLAVOR_CURRENT and BUILDFLAVOR_LEGACY defined constant introduced so it can be used to differentiate current and legacy code blocks
- Introduced USE_EVENTHUB  to control the switch between EventHub or ServiceBus to be more explicit than "NETSTANDARD"
- Removed unused code from PlatformServices


Directory.Build.props
```diff
@@ -39,12 +39,13 @@
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(BuildFlavor)' == 'Current' ">
-    <DefineConstants>$(DefineConstants);NETSTANDARD;NETSTANDARD_TODO</DefineConstants>
-    <TargetFramework>netstandard1.5</TargetFramework>
-    <TestProjectTargetFramework>net462</TestProjectTargetFramework>
+    <DefineConstants>$(DefineConstants);BUILDFLAVOR_CURRENT;USE_EVENTHUB</DefineConstants>
```
jdom: If we are already changing these, can we do the opposite? Let's have the Current build flavor not add any additional conditional compilation, and instead only add additional ones if we are building the Legacy flavor, which will eventually be killed

src/Orleans/Orleans.csproj
```diff
@@ -40,10 +39,10 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjection)" />
     <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)" />
     <PackageReference Include="System.Collections.Immutable" Version="$(SystemCollectionsImmutableVersion)" />
-    <PackageReference Include="System.ValueTuple" Version="$(SystemRuntimeVersion)" />
+    <PackageReference Include="System.ValueTuple" Version="$(SystemValueTupleVersion)" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(BuildFlavor)' == 'Legacy' ">
+  <ItemGroup>
```
jdom: super nit: these ItemGroups can nor be merged

src/Orleans/Runtime/GrainReference.cs
```diff
@@ -461,24 +461,10 @@ protected GrainReference(SerializationInfo info, StreamingContext context)
...
-            this.OnDeserialized(context);
-#endif
-        }
-
-#if !NETSTANDARD_TODO
-        /// <summary>
-        /// Called by BinaryFormatter after deserialization has completed.
-        /// </summary>
-        /// <param name="context">The context.</param>
-        [OnDeserialized]
```
jdom: are you sure getting rid of the method with `[OnDeserialized]` is correct? @ReubenBond can you advise?
BTW, there are more occurrences of this in other files

src/Orleans.PlatformServices/PlatformAssemblyLoader.cs
```diff
@@ -16,22 +12,7 @@ public static Assembly LoadFromBytes(byte[] assembly, byte[] debugSymbols = null
...
-                    }
-                }
-                else
-                {
-                    return AssemblyLoadContext.Default.LoadFromStream(assemblyStream);
-                }
-            }
-#elif NET46
+#if NETSTANDARD2_0
             return Assembly.Load(assembly, debugSymbols);
```
jdom: This doesn't look right... now the .NET version of this will not work correctly, right?

src/Orleans.PlatformServices/Orleans.PlatformServices.csproj
```diff
@@ -9,26 +9,7 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <TargetFramework />
-    <TargetFrameworks>net462;netstandard1.5</TargetFrameworks>
```
jdom: why was this removed? it kills the reason to have PlatformServices in the first place if we do not multi-target. Let's not remove it, and we can decide to remove the project later if it's not required

Directory.Build.props
attilah: We can do, my thinking was to minimize code changes. For the ServiceBus assembly we need both branches of the #ifs and we'll removed those parts later as you already mentioned.
The idea behind BUILDFLAVOR_CURRENT is to make the previously used NETSTANDARD, NETSTANDARD_TODO cleanup easier in the future, since now we're not depending on those.

@jdom, please advise on what would you like to see.

src/Orleans/Orleans.csproj
attilah: They're separated for better human readibility, and grouped by item type. This was not changed now, it was this way before. These cleanups should be done with the legacy removal IMHO. Changeset is minimized and focused on .Net Standard 2.0 support.

src/Orleans.PlatformServices/PlatformAssemblyLoader.cs
attilah: We're not multi targeting PlatformServices anymore so it is not an issue.

src/Orleans.PlatformServices/Orleans.PlatformServices.csproj
attilah: The functionality it provides is not needed anymore with NetStandard 2.0, that's why it was removed and yes the PlatformAssemblyLoader itself can be removed now, maybe we'll have need for different platforms when we are extending to *nix for example.

src/Orleans/Runtime/GrainReference.cs
attilah: We're using it at 2 places and this change was coordinated with @ReubenBond.

Directory.Build.props
jdom: but removing these would have the same amount of changes (although negating the conditional). So `#if NETSTANDARD` would be replaced with `#if !BUILD_FLAVOR_LEGACY` instead of `#if BUILD_FLAVOR_CURRENT`. So same amount of changes.
With event hub, not sure I understand why is that better than just using the build flavor conditional compilation, but at first look I think it would be preferred that it reads BUILD_FLAVOR_LEGACY

src/Orleans/Orleans.csproj
jdom: They were separated because this group had a condition, which doesn't have anymore, so they now belong together now even before we kill the legacy build flavor.
Still, I don't care enough

src/Orleans.PlatformServices/Orleans.PlatformServices.csproj
jdom: should we delete the entire project then? As it is, it has no reason to exist, as the ONLY reason it existed is so that it can multi-target.

ReubenBond: happy for us to remove it - we can add it back if it's needed

src/Orleans/Runtime/GrainReference.cs
ReubenBond: I was just being 'extra careful' with the OnDeserialized attribute. The ctor will be called anyway. We can remove it, we have tests which ensure that BinaryFormatter works with GrainReferences.

ReubenBond: `Serialize_GrainReference_ViaStandardSerializer` is the test

Directory.Build.props
xiazen: what does USE_EVENTHUB param do ?

galvesribeiro: @xiazen looks like it define to ServiceBus old package or new Eventhub netstandard one.

jdom:

PR #3370, opened 2017-08-30
[Abstractions] Move some "easily movable" classes to Orleans.Core.Abstractions

benjaminpetit:


jdom:
>     internal class GrainState<T> : IGrainState

Does this need to be moved to abstractions?

---
Refers to: src/Orleans/CodeGeneration/IGrainState.cs:20 in 39b918c. [](commit_id = 39b918c357d9df0cdb6846d96cb3bacf4bbc4c15, deletion_comment = True)

src/Orleans.Core.Abstractions/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,31 @@
...
+[assembly: InternalsVisibleTo("TestInternalGrains")]
+[assembly: InternalsVisibleTo("NonSilo.Tests")]
+[assembly: InternalsVisibleTo("TestExtensions")]
+[assembly: InternalsVisibleTo("Tester.AzureUtils")]
+[assembly: InternalsVisibleTo("Tester")]
+[assembly: InternalsVisibleTo("DefaultCluster.Tests")]
+[assembly: InternalsVisibleTo("Tester.SQLUtils")]
+[assembly: InternalsVisibleTo("Tester.ZooKeeperUtils")]
+[assembly: InternalsVisibleTo("AWSUtils.Tests")]
+[assembly: InternalsVisibleTo("GoogleUtils.Tests")]
```
jdom: Let's not have any InternalsVisibleTo in abstraction. If for some practical reason we temporarily need 1 or 2, let's add those explicitly, but not all by default. Ideally by the time we release 2.0 we'll have none

jdom:
>         internal static MinimumVersion Singleton { get; } = new MinimumVersion();

All of these internal statics make me cringe, as probable usages will need access to internals. Just make them public but avoid InternalsVisibleTo

---
Refers to: src/Orleans/Versions/Selector/MinimumVersion.cs:8 in 39b918c. [](commit_id = 39b918c357d9df0cdb6846d96cb3bacf4bbc4c15, deletion_comment = True)

src/Orleans.Core.Abstractions/Properties/AssemblyInfo.cs
benjaminpetit: .I agree.

But as we discussed weeks ago, not having internal to at least Orleans and OrleansRuntime will be almost impossible (GrainId...)

benjaminpetit:
Will split this PR in smaller one

PR #3373, opened 2017-08-31
[WIP] xplat build-time code generation via MSBuild Task

ReubenBond:
This is based upon @attilah NETStandard2.0 PR.

This PR enables build time code generation on linux via a new NuGet package: `Microsoft.Orleans.CodeGenerator.MSBuild`.

Sample project file:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Orleans.CodeGenerator.MSBuild" Version="2.0.0-preview3" PrivateAssets="All" />
  </ItemGroup>
</Project>
```

It includes a small hack to remove a build warning:
```xml
<_ResolveAssemblyReferenceResolvedFiles Include="$(InputAssembly)" />
```

That suppresses a warning during the [`IncrementalClean` target](https://github.com/Microsoft/msbuild/blob/e7019886b3fd520b547a7353d78d106d72a4b828/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L4463-L4520) caused by the lack of AppDomains in the codegen task: no AppDomain -> no unloading assembly once codegen completes -> file remains locked -> cannot delete file during `IncrementalClean`.

Note: this enables building Orleans apps on linux/mac with build time codegen - but it's not complete. We need to also enable building Orleans itself, which is an MSBuild change (at its core: instead of using Orleans.SDK.targets we use this package). We also need to consider NuGet changes so that the existing nuget package is replaced by this, for example.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -39,9 +39,10 @@
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(BuildFlavor)' == 'Current' ">
-    <DefineConstants>$(DefineConstants);NETSTANDARD;NETSTANDARD_TODO</DefineConstants>
-    <TargetFramework>netstandard1.5</TargetFramework>
-    <TestProjectTargetFramework>net462</TestProjectTargetFramework>
+<!--    <DefineConstants>$(DefineConstants);NETSTANDARD;NETSTANDARD_TODO</DefineConstants>-->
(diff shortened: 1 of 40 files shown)


ReubenBond:
So much has changed :) I'll open a new PR for xplat soon

PR #3375, opened 2017-09-01
Fix ClientGenerator nuget pack, add possibility to override build flags

attilah:
- Fix ClientGenerator nuget to include every dll from the output folder (netstandard 2.0 was left out after update)
- Add a possibility to override build flags if needed (for example to enable parallel build)

### DIFF
--- a/src/ClientGenerator/ClientGenerator.csproj
+++ b/src/ClientGenerator/ClientGenerator.csproj
@@ -36,7 +36,7 @@
       <PackagePath>build\$(PackageId).targets</PackagePath>
       <Visible>false</Visible>
     </None>
-    <None Include="$(OutDir)/ClientGenerator.exe.config;$(OutDir)/Orleans.Core.Abstractions.dll;$(OutDir)/Orleans.dll;$(OutDir)/OrleansCodeGenerator.dll;$(OutDir)/Microsoft.CodeAnalysis.dll;$(OutDir)/Microsoft.CodeAnalysis.CSharp.dll;$(OutDir)/System*.dll">
+    <None Include="$(OutDir)/ClientGenerator.exe.config;$(OutDir)/*.dll">
       <Pack>true</Pack>
       <PackagePath>tools\</PackagePath>
       <Visible>false</Visible>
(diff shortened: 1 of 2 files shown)


src/ClientGenerator/ClientGenerator.csproj
```diff
@@ -36,7 +36,7 @@
       <PackagePath>build\$(PackageId).targets</PackagePath>
       <Visible>false</Visible>
     </None>
-    <None Include="$(OutDir)/ClientGenerator.exe.config;$(OutDir)/Orleans.Core.Abstractions.dll;$(OutDir)/Orleans.dll;$(OutDir)/OrleansCodeGenerator.dll;$(OutDir)/Microsoft.CodeAnalysis.dll;$(OutDir)/Microsoft.CodeAnalysis.CSharp.dll;$(OutDir)/System*.dll">
```
jdom: Aren't you missing `ClientGenerator.exe` here?

attilah: Since that's the build output, that's included by default.

jdom: Or maybe that's included by default in the nuget package when packing this project?

ReubenBond: You could also use the method I used in #3373 https://github.com/dotnet/orleans/pull/3373/commits/af92d6ece204e7a33612820ff515720c2cc4394e#diff-ad120c34366a8c605281f3e94ff9176cR35
```xml
  <Target Name="PackTaskDependencies" BeforeTargets="GenerateNuspec">
   <!-- 
    The include needs to happen after output has been copied to build output folder
    but before NuGet generates a nuspec. See https://github.com/NuGet/Home/issues/4704.
  -->
    <ItemGroup>
      <_PackageFiles Include="bin\$(Configuration)\**\*" Exclude="bin\$(Configuration)\**\$(AssemblyName).*">
        <PackagePath>tasks\%(RecursiveDir)</PackagePath>
        <Visible>false</Visible>
        <BuildAction>Content</BuildAction>
      </_PackageFiles>
    </ItemGroup>
  </Target>
```

That way everything is captured (maybe that's too much)

ReubenBond: Although what you have looks cleaner. I guess it works because the `<None>` item is evaluated once for `Build` and again for the `Pack` target.

PR #3381, opened 2017-09-05
NuGet updates and Debug type change for test projects

attilah:
- Update nuget package versions for 2.0 to align with .NET Core 2.0 release and other minor upgrades based on NetStandard compatibility and/or features.
- Change Debug type to full for test projects, so Test Explorer source code navigation and CodeLens will work again in VS.


jdom:
@dotnet-bot test netfx-bvt

Issue #3369, opened 2017-08-30
Orleans Transactions


jason-bragg:
Over the course of the next few weeks we’ll be introducing a series of code changes to make a Microsoft Research (MSR) developed implementation of Orleans Transactions publicly available.  This initial version of Orleans Transactions is based off [this paper](https://www.microsoft.com/en-us/research/publication/transactions-distributed-actors-cloud-2/) from @philbe and Tamer Eldeeb, with the exception that it will use a timestamp based concurrency control instead of locking, and dependency injection (via the [Facet System]( https://github.com/dotnet/orleans/issues/3223)) instead of inheritance.

Related work:
[Facet System]( https://github.com/dotnet/orleans/issues/3223) - #3223 
[Orleans transaction core]( https://github.com/dotnet/orleans/pull/3358) - #3358 
[Transaction Orchestration]( https://github.com/dotnet/orleans/pull/3382) - #3382
[Transactional State]( https://github.com/dotnet/orleans/pull/3397) - #3397 
[Transaction manager tests ]( https://github.com/dotnet/orleans/pull/3473) - #3473 
[Reusable storage bridge]( https://github.com/dotnet/orleans/pull/3360) - #3360
[Orleans.Transactions.Azure]( https://github.com/dotnet/orleans/pull/3426) - #3426
[Simple account transfer transaction example](https://github.com/dotnet/orleans/pull/3653) - #3653 

Upcoming:
Design Documentation
ITransactionalJournaledState Facet + testing
User Documentation

galvesribeiro:
Thanks for the issue explaining the steps. 

> with the exception that it will use a timestamp based concurrency control instead of locking, and dependency injection (via the Facet System) instead of inheritance.

I understand, and agree that we should avoid inheritance and use DI/facets to make it work. I dream with some day we will have (almost) POCO grain classes. 

However, I don't understand the change from locking to timestamp. Are you saying that you are using optimistic concurrency on this implementation instead of the weak-lock we have first phase of 2PC? If that is the case, doesn't it break all the "new" strategy defended by @philbe paper? Can you elaborate more on that? Thanks

galvesribeiro:
Also:

> Grain Based Transaction Manager, backed by azure storage.

That get back to my question on #3358. Are we introducing a new provider interface here? Or will it just be an enhanced version of current state storage provider ones?

ReubenBond:
[...]

PR #3382, opened 2017-09-05
Minimal implementation of transaction orchestration

jason-bragg:
- Added orleans transaction assembly.
- Moved abstractions and implementations not necessary to orleans core into orleans transaction assembly.
- Added in-cluster grain-based transaction manager.
- Added in memory transaction log for test and/or development purposes.
- Added Minimal transaction orchestration testing.
- Added testing of expected behavior for disabled (not configured) transactions.


src/Orleans.Transactions/InClusterTM/TransactionManager.cs
```diff
@@ -0,0 +1,695 @@
...
+            checkpointedLSN = 0;
+
+            gcTimer = new Timer(GC, null, Timeout.Infinite, -1);
+        }
+
+        #region ITransactionManager
+
+        public async Task StartAsync()
+        {
+            await transactionLog.Initialize();
```
ReubenBond: If there's an exception during this process then we will be in an invalid state. I'm not sure what the best solution is to make this resilient, but maybe it ought to be retriable. Looking at the implementation of the Azure-backed tx log, I see that the `Initialize` method does not reset all state to a consistent point.

src/Orleans.Transactions/InClusterTM/TransactionManager.cs
```diff
@@ -0,0 +1,695 @@
...
+        private void GC(object args)
+        {
+            //
+            // Truncate log
+            //
+            if (checkpointedLSN > 0)
+            {
+                try
+                {
+                    transactionLog.TruncateLog(checkpointedLSN - 1).Wait();
```
ReubenBond: It's not 'good' that we're blocking a threadpool thread here. Preferably we would use an async timer and await this.

src/Orleans.Core.Abstractions/Orleans.Core.Abstractions.csproj
```diff
@@ -11,5 +11,9 @@
     <RootNamespace>Orleans</RootNamespace>
     <TargetFramework>netstandard1.3</TargetFramework>
   </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="1.1.1" />
```
xiazen: you can use the $(MicrosoftExtensionsDependencyInjection) for version number here. I think the whole sln is using that variable to keep the version number consistent

src/Orleans.Transactions/InClusterTM/TransactionManager.cs
jason-bragg: The TM needs a lot more love, but for now I'll make the TM and Memory log transient instead of singletons.  That way a failure in the initialization will cause the grain activation to fail and the next attempt will start with a clean state.

src/Orleans.Core.Abstractions/Orleans.Core.Abstractions.csproj
jason-bragg: Hmm..  Any idea how this variable is set?  It's not being used consistently, and I'm not sure using it at all is a good idea.  Changing a package version without going through the package manager can introduce many problems.  Will look into it more...

xiazen: it is set in Directory.Build.Props ... I see our csproj are using those. but I will @attilah here since he may know more about the suggested thing to do

src/Orleans/Transactions/TransactionContext.cs
```diff
@@ -90,13 +91,40 @@ public TransactionInfo(TransactionInfo other)
...
         public HashSet<long> DependentTransactions { get; }
 
-        public void Union(TransactionInfo other)
+        [NonSerialized]
+        public int PendingCalls;
+
+        [NonSerialized]
+        ConcurrentQueue<TransactionInfo> joined = new ConcurrentQueue<TransactionInfo>();
+        // protect from null ref in case join is accessed on serialized object.
+        private ConcurrentQueue<TransactionInfo> Joined => this.joined ?? (this.joined = new ConcurrentQueue<TransactionInfo>());
```
jdom: can this ever be null? you are assigning it in the field initializer... does it not get initialized too when deserializing?

jason-bragg: Unfortunately non-serialized fields are not initialized when deserializing, so yes, this can be null :(

src/Orleans.Core.Abstractions/Orleans.Core.Abstractions.csproj
```diff
@@ -11,5 +11,9 @@
...
+  </ItemGroup>
```
jdom: Why move `ISiloBuilder` to abstractions (which introduced this dependency too)?
At least in our current design (https://github.com/dotnet/orleans/issues/3353), the idea is not to have the building and configuration be part of the abstractions project. Otherwise this would make as rev this abstractions package project way more often.
We are potentially considering having a separate abstraction for hosting, but until then, everything related to hosting should go into Core.

jason-bragg: ISiloBuilder was moved so I could add transaction feature setup that was compatible with the builder pattern.
The general problem:  How should extension libraries register services with the silo builder?
Pls see SiloBuilderExtensions  and DevelopmentSiloBuilderExtensions.
I can remove these extension an provide only configuration services actions if that is preferred, but that seems a bit clumsy.

jdom: Well, the thing is that this project can reference Core (in addition to Core.Abstractions) and still define the extension methods (which you shouldn't remove).
At some point we can have some Microsoft.Orleans.Hosting.Abstractions (although I wouldn't go that way yet until we start using the generic hosting abstractions from the aspnet team). We could also (instead or in addition to) create a Microsoft.Orleans.Transactions.Abstractions so that grain developers would reference just that, instead of Transactions, which now brings a dependency to Core. Nevertheless, that's just going to add more issues than benefits for end users, at least for now until we know the abstractions are absolutely stable and we can guarantee no breaking changes.

src/Orleans.Transactions/Abstractions/TransactionalExtensionExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    public static class TransactionalExtensionExtensions
+    {
+        public static ITransactionalResource AsTransactionalResource(this ITransactionalExtension transactionalExtension, string resourceId)
+        {
+            return new TransactionalResourceExtensionWrapper(transactionalExtension, resourceId);
+        }
+
+        [Serializable]
+        [Immutable]
```
jdom: is this correct? the implementation doesn't look immutable, since it's mutating the wrapped extension, and looking at the implementation for it, it's neither immutable nor [Serializable].
If this is meant to never be serialized and used inside a single silo instance, then I'd recommend that you implement a class and apply the SerializerAttribute to it, and explicitly throw on Serialize, and just pass through when (deep)copying the object, to achieve a similar effect as you get with Immutable, but with the other false behaviors.
Then remove both Serializable and Immutable attributes from this class.
You can achieve something similar without an external class by using the CopierMethodAttribute (and SerializerMethod and DeserializerMethod attributes)

src/Orleans.Transactions/InClusterTM/ActiveTransactionsTracker.cs
```diff
@@ -0,0 +1,136 @@
...
+                        if (maxAllocatedTransactionId - highestActiveTransactionId <= config.AvailableTransactionIdThreshold)
+                        {
+                            var batchSize = config.TransactionIdAllocationBatchSize;
+                            transactionLog.UpdateStartRecord(maxAllocatedTransactionId + batchSize).GetAwaiter().GetResult();
+
+                            maxAllocatedTransactionId += batchSize;
+                        }
+                    }
+                }
+#if !NETSTANDARD
```
jdom: not needed any more, this is now supported in .NET Standard 2.0 (and the NETSTANDARD compilation constant is no longer defined in our solution anyway, only BUILD_FLAVOR_LEGACY could be used to differentiate between the 2 build flavors)

src/Orleans.Transactions/Utilities/MultiCompletionSource.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Orleans.Transactions
+{
+    /// <summary>
+    /// Utility class that waits for the completion of multiple other tasks.
+    /// Duplication of internal orleans utility MultiTaskCompletionSource.
+    /// TODO: Consider making orleans version public.
```
jdom: or link files, or something else, let's not make non-Orleans utilities public, as we couple runtime versions unnecessarily. See `Sharing non-core Code` in [https://github.com/dotnet/orleans/issues/3353](https://github.com/dotnet/orleans/issues/3353)

src/OrleansRuntime/Hosting/DefaultSiloServices.cs
```diff
@@ -184,13 +184,12 @@ internal static void AddDefaultServices(IServiceCollection services)
...
-            services.AddSingleton(typeof(IKeyedServiceCollection<,>), typeof(KeyedServiceCollection<,>));
+            services.TryAddSingleton(typeof(IKeyedServiceCollection<,>), typeof(KeyedServiceCollection<,>));
 
             // Transactions
-            services.AddSingleton<ITransactionAgent,TransactionAgent>();
-            services.AddSingleton(sp => new Lazy<ITransactionAgent>(() => sp.GetRequiredService<ITransactionAgent>()));
-            services.AddSingleton<TransactionServiceGrainFactory>();
-            services.AddSingleton(sp => sp.GetRequiredService<TransactionServiceGrainFactory>().CreateTransactionManagerService() );
+            services.TryAddSingleton<ITransactionAgent,TransactionAgent>();
+            services.TryAddSingleton(sp => new Lazy<ITransactionAgent>(() => sp.GetRequiredService<ITransactionAgent>()));
```
jdom: are you trying to register a Lazy? if that's the case, please use the Factory delegate instead, as Lazy is non-conforming, and some containers might resolve it differently

src/Orleans.Core.Abstractions/Orleans.Core.Abstractions.csproj
jason-bragg: Will attempt to use runtime dependency.  Should be ok I think.

src/Orleans.Transactions/Abstractions/TransactionalExtensionExtensions.cs
jason-bragg: Immutability:
In this case the ITransactionalExtension is a grain reference, so all of the state is immutable.  While this is not enforced in the interface, due to how extensions are installed, this can be misused, but as it's not a commonly needed behavior I don't think their is much risk in that.

Serializability:
I'm open to other approaches, but currently I'm forcing the generation of a serializer via GenerateSerializer in the assembly info.  This was my understanding of what our recommended approach was for types that are not explicitly references in a grain interface.  Am I mistaken, is this not the preferred approach?

src/Orleans.Transactions/Utilities/MultiCompletionSource.cs
jason-bragg: These are not public.  They are just duplicates of useful internal Orleans logic; logic these systems used when initially developed because they were part of Orleans core.  If Orleans core does not expose and support utilities like this publically, which I agree is not necessarily a good idea, then libraries need to provide their own solutions, which is what this library is doing, it's just doing so by copying the useful internal logic it originally used.

src/Orleans.Transactions/Abstractions/TransactionalExtensionExtensions.cs
jdom: You are not mistaken. It was my missunderstanding of ITransactionalExtension. I thought it was not immutable, but a grain reference is. So I thought you were not really serializing this, but just wanted to avoid deep copying.
Clarified now, ignore the comment, this approach is correct.

src/Orleans.Transactions/Utilities/MultiCompletionSource.cs
jdom: Correct, what I meant (as I also mentioned in person) is that instead of adding a TODO that says "consider making orleans version public", have something like `// TODO: replace with the approach that we'll implement for "Sharing non-core Code" in https://github.com/dotnet/orleans/issues/3353`

jason-bragg:
All feedback addressed.

src/Orleans/Transactions/TransactionContext.cs
ReubenBond: In this case, that's because there's no parameterless constructor, so the object must be created using `GetUninitializedObject` during deserialization.

jason-bragg: @ReubenBond Empty constructor is not a valid use case.  Added one for serialization purposes.

ReubenBond:
LGTM. @jdom @xiazen if this LGTYT ('looks good to you, too' 😉) then let's merge

xiazen:
LGTM

PR #3385, opened 2017-09-06
Add activate/deactivate to grain lifecycle

jason-bragg:


src/Orleans/Runtime/IGrainLifeCycle.cs
```diff
@@ -12,15 +15,30 @@ public enum GrainLifecycleStage
     {
         //None,
         //Register,
-        SetupState,  // Setup grain state prior to activation
-        //InvokeActivate,
+        SetupState = 1<<10,  // Setup grain state prior to activation
+        Activate   = SetupState + 1<<10, // Acivate grain
```
jdom: typo in comment.
Nevertheless, please turn these into XML comments so that it helps with intellisense

PR #3389, opened 2017-09-08
Quick fix for .NET Core PNSE on ReflectionOnlyType

ReubenBond:
This quick fix should allow loading on .NET Core 2.0. It's not ideal, but it ought to allow us to perform a 2.0 Technical Preview refresh

### DIFF
--- a/src/Orleans/AssemblyLoader/AssemblyLoader.cs
+++ b/src/Orleans/AssemblyLoader/AssemblyLoader.cs
@@ -217,7 +217,7 @@ namespace Orleans.Runtime
                         if (IsCompatibleWithCurrentProcess(j, out complaints))
                         {
                             if (logger.IsVerbose) logger.Verbose("Trying to pre-load {0} to reflection-only context.", j);
-                            Assembly.ReflectionOnlyLoadFrom(j);
+                            Assembly.LoadFrom(j);
                         }
                         else
                         {
@@ -344,7 +344,7 @@ namespace Orleans.Runtime
 
                 if (IsCompatibleWithCurrentProcess(pathName, out complaints))
                 {
-                    assembly = Assembly.ReflectionOnlyLoadFrom(pathName);
+                    assembly = Assembly.LoadFrom(pathName);
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


ashtonkj:
Hi Reuben. In my case I found one other change that needs to be made for this to work. Update: ReflectionOnlyLoadAssembly in AssemblyLoader.cs to use 

assembly = Assembly.LoadFrom(pathName); 

instead of 

assembly = Assembly.ReflectionOnlyLoadFrom(pathName);

ReubenBond:
Thanks, @ashtonkj :)

jdom:
Thanks @ashtonkj for the tip and confirming that it works. I'll try to publish a refresh to MyGet today

jdom:
Had to revert this, as tests started failing.
I could still create a signed build with this, but so many failures made me doubt whether it's a good fast workaround for the issue or would it make matters worse. Failures are probably due to the nature of our unit tests though, and probably won't surface in actual apps, but still

ashtonkj:
That sounds reasonable. It is a hack (but it works in my testing scenario). If you or @ReubenBond need any assistance with fixing things up properly for .net core 2 let me know.

jdom:
ok, thanks. 
BTW, what do you guys say if we ship the package without this hack? As a workaround you should be able to just pre-load all the assemblies, right? Or what are the conditions in which you get PNSE?
I understand that @galvesribeiro didn't get those in his .NET Core 2 app that runs in linux.

If that works for now, releasing a refresh can still be worth it, and we can release a new one as soon as we fix this issue for good

ashtonkj:
I get the PNSE on Silo startup (only on .net core 2) without the hack in. 

What do you mean by pre-loading the assemblies? How would you do this in practice?

galvesribeiro:
@ashtonkj you can add a reference to the assembly and then `typeof(SomeTypeFromTheAssenbly)` anywhere before start the silo.

And yes, as @jdom mentioned, I never had that issue here on OSX neither on a linux container.

ashtonkj:
I am going to create a sample with the issue. Maybe it will help diagnose the issue. For what its worth, I am basically just running the sample Silo that came with Orleans 1, upgraded to 2. I am not sure if something in the Silo initialization has changed since then (if so, we might want to look at updating the docs to show the breaking changes and the correct upgrade path).

ashtonkj:
A repro can be found here: https://github.com/ashtonkj/orleanspnse

The packages that are used have been built directly from master and are included in the packages directory. If there are any questions, feel free to ask and I'll assist as best as possible.

jdom:
Created a less disrupting PR here: https://github.com/dotnet/orleans/pull/3394

Issue #992, opened 2015-11-05
Orleans Telemetry APIs


galvesribeiro:
## Problem

Today is hard to measure performance live on Orleans due to the lack of metrics and KPIs being extracted at runtime from Orleans runtime, and neither from Grain code.

We have today [ILogConsumer](https://github.com/dotnet/orleans/blob/master/src/Orleans/Logging/ILogConsumer.cs) which is pretty handy when we talk about Logging and tracing but not when we talk about KPIs.

The idea here is to create a set of APIs that will be used to extract in real-time information from Grains, Requests, Silos, Cluster, Streams, and many other Orleans Runtime components in order to push it to a [APM](https://en.wikipedia.org/wiki/Application_performance_management) external tool, or any kind of persistent storage for further processing or real-time display pushed data on a dashboard like [PowerBI](http://powerbi.com).
## Proposed Solution

**Telemetry Provider**

Based on common APM frameworks, provide a plugable abstraction for a APM provider such as this interface:

```
public interface ITelemetryProvider
{
    Task TrackDependency(string dependencyName, string commandName, DateTimeOffset startTime, TimeSpan duration, bool success);
    Task TrackDependency(DependencyTelemetry telemetry);
    Task TrackEvent(EventTelemetry telemetry);
    Task TrackEvent(string eventName,IDictionary<string, string> properties = null, IDictionary<string, double> metrics = null);
    Task TrackException(ExceptionTelemetry telemetry);
    Task TrackException(Exception exception, IDictionary<string, string> properties = null, IDictionary<string, double> metrics = null);
    Task TrackMetric(MetricTelemetry telemetry);
    Task TrackMetric(string name, double value, IDictionary<string, string> properties = null);
    Task TrackRequest(RequestTelemetry request);
    Task TrackRequest(string name, DateTimeOffset startTime, TimeSpan duration, string responseCode, bool success);
    Task TrackTrace(TraceTelemetry telemetry);
    Task TrackTrace(string message);
    Task TrackTrace(string message, SeverityLevel severityLevel);
    Task TrackTrace(string message, SeverityLevel severityLevel, IDictionary<string, string> properties);
    Task TrackTrace(string message, IDictionary<string, string> properties);

}
```

[...]

PR #3390, opened 2017-09-08
Separate the APM methods from the tracing abstraction

jdom:
~Remove ITelemetryConsumer and derived interfaces except for IMetricTelemetryConsumer
Use IDisposable to dispose consumers~

We are abstracting away the metrics stuff from the tracing, as we will be migrating all of our tracing to Microsoft.Extensions.Logging, but that abstraction doesn't support metrics. So when we deprecate the current `Logger`, the metrics will continue to be supported.

Move all APM methods currently in the `Logger` abstraction to a new `ITelemetryClient` interface. Preserve all existing client as well as telemetry consumer methods, as an APM abstraction. Nevertheless, have Orleans still use the Logger abstraction (very soon to be replaced with ILogger) that does not send all the traces to the APM abstraction (the end user is welcome to hook them up explicitly so that it preserves that behavior).

The configuration is backwards compatible still with the old telemetry consumers. ~, the difference is that it is constrained to consumers that implement `IMetricTelemetryConsumer` only (all other TC interfaces were removed).~

~We could also get rid of `IMetricTelemetryConsumer`  and replace it with `IMetricsWriter` for consumers, similar to how `ILogger` from MEL works both as an interface for producing events, as well as a way for providers to implement it and act as a sink. This could be a separate PR, as it is not required to do that refactoring (albeit a nice to have) and also would be breaking compatibility in a bigger way with legacy telemetry consumers.~

We might want to make it configurable through ISiloBuilder and IClientBuilder, but that can be made as a separate pass.

In the future we might want to separate how Orleans publishes metrics so that they could be piped to the APM abstraction or to another place if needed, but that is completely outside the scope of this PR. The goal of this PR is just to decouple metrics from tracing, so that we can use MEL for tracing, but do not re-design any of the telemetry stuff yet.

~Do not merge this, just review for now, as we will figure out with @xiazen how to merge it, as it conflicts with #3333~

Edit: corrected description based on reduced scope, where we will not remove any of the telemetry abstractions


src/Orleans/Configuration/ClientConfiguration.cs
```diff
@@ -329,7 +331,7 @@ internal void LoadFromXml(XmlElement root)
                             }
                             break;
                         case "Telemetry":
-                            ConfigUtilities.ParseTelemetry(child);
+                            ConfigUtilities.ParseTelemetry(child, this.MetricsConfiguration);
```
xiazen: I thought we are moving away from xml config in 2.0. What's the thinking behind this ?

src/Orleans/Logging/LogManager.cs
```diff
@@ -149,36 +142,38 @@ public static void Initialize(ITraceConfiguration config, bool configChange = fa
...
-                }
 
-                if (Trace.Listeners.Count > 0)
-                {
-                    if (!TelemetryConsumers.OfType<TraceTelemetryConsumer>().Any())
-                    {
-                        TelemetryConsumers.Add(new TraceTelemetryConsumer());
-                    }
-                }
+                // TODO: need to re-connect this to the logging abstractions!!
```
xiazen: ? what's this re-connect? I think we can remove this , right?

src/Orleans/Telemetry/MetricsWriter.cs
```diff
@@ -0,0 +1,128 @@
...
+                if ((consumerConfig.Properties?.Count ?? 0) == 0)
+                {
+                    // first check whether it is registered in the container already
+                    consumer = (IMetricTelemetryConsumer)serviceProvider.GetService(consumerConfig.ConsumerType);
+                }
+                if (consumer == null)
+                {
+                    consumer = (IMetricTelemetryConsumer)ActivatorUtilities.CreateInstance(serviceProvider, consumerConfig.ConsumerType, consumerConfig.Properties?.Values?.ToArray() ?? new object[0]);
+                }
+                consumers.Add(consumer);
```
xiazen: Looks like using MetricTelemetryConfiguration in ClientConfig and NodeConfig is the suggested way to configure MetricsWriter. But using it means we need to use reflection to construct MetricTelemetryConsumer or ask application code to inject consumer type into DI. I think generally we should move away from using reflection and config.Properties to construct MetricTelemetryConsumer . For injecting consumer type into DI, it is not straightforward for user to figure this out, since neither MetricsWriter nor MetricTelemetryConfiguration has imply or enforced that programmatically.

xiazen: Why cannot we simply do 
  ```
  public class MetricsWriter : IMetricsWriter, IDisposable
    {
        public List<IMetricTelemetryConsumer> Consumers { get; set; }

        public MetricsWriter(IEnumerable<IMetricTelemetryConsumer> consumers)
        {
            this.Consumers = consumers.ToList();
        }
...remaining methods...
}
```

Making the configuration more programmatical? User can configure MetricWriter into DI using SiloBuilder

        private static IMetricsWriter CreateMetricWriter(IServiceProvider sp)
        {
            var consumer = new MetricConsumer(some param, sp.GetService<T2>);
            return new MetricsWriter(new List<IMetricTelemetryConsumer>{consumer});
        }

        public void ConfigureMetricWriter(IServiceCollection svc)
        {
            svc.AddSingleton<IMetricsWriter>(sp => CreateMetricWriter(sp));
            return svc;
        }

`siloBuilder.ConfigureServices(ConfigureMetricWriter)`

Comment (jdom) [reply src/Orleans/Telemetry/MetricsWriter.cs]: Agree, we can improve it, but as mentioned in the description, this first pass is still back-compat WRT config. If I break compatibility in the same PR, we could hold it off for very long, as we need to make sure we have a way to bridge the gap with our existing users, as well as make sure the design is correct. And even with that, `master` still doesn't have a reference to `Microsoft.Extensions.Configuration`, so no point introducing it here either.

For all that, I don't think it's worth it to try to tackle it in the same PR if your plan is to rev the Logging part in the short term, since this configuration is 100% back compat for metric telemetry consumers (so I didn't introduce any new concepts, just made the previous config still work).

Comment (xiazen) [reply src/Orleans/Telemetry/MetricsWriter.cs]: sorry I missed that description. Mostly just want to get to know the plan more. I see MetricTelemetryConfiguration  is here to maintain backward compatibility in the transitioning phase. Will it be part of 2.0? Is it in 2.0. We support two ways of configuring MetricTelemetryConsumer, one is through CLusterconfig and Nodeconfig, another one is through SiloBuilder and ClientBuilder?

Comment (jdom) [reply src/Orleans/Telemetry/MetricsWriter.cs]: Hopefully we can deprecate the ClusterConfiguration and ClientConfiguration object graphs soon. There is a small hint of what we could do with these god objects in https://github.com/dotnet/orleans/issues/3353 but there's no hard commitment on what to do with them for 2.0 (or if we will remove them after we ship 2.0).
That's why this PR is still separate from those decisions. For now the only way to configure telemetry is as it was before, through the ClusterConfiguration and ClientConfiguration. We can still enable the builder extension methods even if we don't deprecate these config objects yet (but again, separate PR, after logging settles).

File: src/Orleans/Telemetry/IMetricsWriter.cs [hunk 4]
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>Send a metric value to the registered telemetry consumers.</summary>
+        /// <param name="name">Metric name.</param>
+        /// <param name="value">Metric value.</param>
+        /// <param name="properties">Named string values you can use to classify and filter metrics.</param>
+        void TrackMetric(string name, TimeSpan value, IDictionary<string, string> properties = null);
+
+        /// <summary>Send a metric value of 1 to the registered telemetry consumers, to be displayed in an aggregated view.</summary>
+        /// <param name="name">Metric name.</param>
+        void IncrementMetric(string name);
```
ReubenBond: This and subsequent methods seem like they should be extension methods which call `TrackMetric`.

src/Orleans/Telemetry/IMetricsWriter.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <param name="name">Metric name.</param>
+        /// <param name="value">Metric value.</param>
+        /// <param name="properties">Named string values you can use to classify and filter metrics.</param>
+        void TrackMetric(string name, double value, IDictionary<string, string> properties = null);
+
+        /// <summary>Send a metric value to the registered telemetry consumers.</summary>
+        /// <param name="name">Metric name.</param>
+        /// <param name="value">Metric value.</param>
+        /// <param name="properties">Named string values you can use to classify and filter metrics.</param>
+        void TrackMetric(string name, TimeSpan value, IDictionary<string, string> properties = null);
```
ReubenBond: I feel this should be `long` instead of `TimeSpan`

xiazen: nah it is from old APM methods in Logger. it is designed to track TimeSpan values. 

The old APM methods design is not very flexible. It only tracks double and TimeSpan values. I think we should improve it in a separate PR. Maybe borrow some pattern from MEL, something like TrackMetric<T>(string name, IDictionary<string, string>) would be more flexible.

ReubenBond:
From an API-design perspective, I'd prefer something like `IMetric<T>` (or maybe just `IDoubleMetric`, `ILongMetric`) instead of passing a metric name string every time I record a metric (which implies a dictionary lookup). I'm not sure what the scope of the PR is, though. That approach is very similar to the internal APIs we use for recording metrics internally, where we have classes like `FloatValueStatistic : ICounter<float>`. When designing MDM, I only included support for `long` metrics, since they're very fast to aggregate, and the message to consumers was to use fixed decimal places within that (eg, multiply CPU usage by 1000 and make it a `long`). We're not including an aggregation framework here, so that doesn't matter too much.

I'm not convinced that we need to support passing dimensions to the `TrackMetric` method - that could also be handled at the factory level. What would common dimensions be? I imagine things like `HostName`, `GrainType`, `PrimaryKey`, & `TenantId` would be very common. Those can be determined at activation time. Dimensions which cannot be determined at activation time might be things which involve parameter values. Eg, maybe in a restaurant app we have a dimension per ⭐rating (name = "rating", Dimensions = { "stars", "5" }). That's still a limited range of dimensional values.

Cases where a dimension has a very wide range might be where a single shared grain is recording per-user metrics. In that case, I think going through a factory is a good way of dissuading users from using wide range dimensions. We could also cache in those cases.

src/Orleans/Telemetry/IMetricsWriter.cs
xiazen: I definitely +1 on redesign metric api in a separate PR. the current API design is not good.

jdom:
100% agree on redesigning so that the counter already includes the name, instead of passing it every time. We already discussed with @galvesribeiro as that being the next step for the Api. But we wanted to make that explicitly in a separate PR, this is just splitting metrics from tracing without any additional change

ReubenBond:
Ok, cool, got it - so this is purely an unblocking PR?

jdom:
Correct, to be able to move our tracing to Microsoft.Extensions.Logging and deprecate Logger

galvesribeiro:
@ReubenBond I'm not seeing it as an unblock PR. It is removing lots of code from the _old_ telemetry APIs, specially dropping all the interfaces... I'm not sure where is is "just unblocking" anything.

jason-bragg:
I'm confused about the goals of this PR.  The goal of the telemetry api's were to help support APM's as described in Orleans Telemetry APIs #992.  While it's arguable whether this should be part of Orleans core or an extension, and it's unclear how Orleans should expose metrics, answering those questions was, to my knowledge, not part of the 2.0 release.  So what's the goal of this PR?

In the comments this PR has been described as 'unblocking', which seems to indicate there is other work blocked which this PR would unblock, but what work that is, or what about the telemetry api's is causing this blockage is not clear from the description or the comments.

jdom:
The goal of this PR is to separate the metrics publishing from the tracing (which are now conflated under the "telemetry" umbrella), so that we can replace our custom tracing abstractions with Microsoft.Extensions.Logging (which do not support metrics). So it un blocks the move to those.
In the near future we can completely redesign the metrics API, but as mentioned, it wasn't the goal of *this* PR

galvesribeiro:
If that is the case, all you need to do is remove `ITraceTelemetryConsumer` and make changes on `Logger` static class. It has nothing todo with current telemetry code which can be left untouched until it is redesigned...

jdom:
@galvesribeiro I believe those are the minimal changes needed that I did. If you think there is even less we could do for the splitting, please suggest how, as I did not change at all how to publish metrics, nor did I change the metrics telemetry consumers. Just added a config object so that we could decouple from the static Log manager class, but that itself is still backcompat

galvesribeiro:
@jdom 
> Remove ITelemetryConsumer and derived interfaces except for IMetricTelemetryConsumer
Use IDisposable to dispose consumers

Unless I understood it wrong, and the changeset is wrong, all the `IXXXTelemetryConsumer` except `IMetricTelemetryConsumer` were removed. 

> We could also get rid of IMetricTelemetryConsumer and replace it with IMetricsWriter for consumers, similar to how ILogger from MEL works both as an interface for producing events, as well as a way for providers to implement it and act as a sink

I don't see why we should do that...

The only relation the telemetry interfaces has with the logging system, is that the publishing methods are all on Logger class, which was something I were against to in the first place, but was what the core team decided to move into.

My point is, why change the telemetry in this PR, if the purpose was just to move trace/logging out from the telemetry interfaces? If that is the case, we should just remove the `ITraceTelemetryConsumer` as I mentioned, and leave the rest untouched.

When/if the rest of the APIs are going to be redesigned later, then it will be changed. 

IMHO it is a completely different thing (now that you guys decided that it is, and logging will be handled differently - even if I'm strongly against it) which should not affect each other. I don't see how drop the telemetry interfaces and add `IMetricsWriter` is necessary.

jdom:
Ok, based on the feedback we decided to basically just split our logging from the telemetry API for now, and change nothing else. We will stop calling the trace events in this API in favor of the Microsoft.Extensions.Logging one, but our end users are free to hook them up back again (it would be trivial to do so, if that's what they want).
Eventually we may want to be out of the business of providing abstractions for APMs, but we are not doing anything in that regard for now.

jdom:
As mentioned, brought back the entire set of APM methods into `ITelemetryClient`. Also, as @xiazen is still doing the migration to MEL, I temporarily forwarded all of our traces coming from the Logger infrastructure into the ITelemetryClient (otherwise we would see no file nor console logs automatically). In the future this piece will be opt-in, and hooked up as an `ILoggerProvider` instead, but Orleans will not forward any traces to the telemetry client abstraction by default.

jdom:
BTW, this PR is now standalone and mergeable before @xiazen's migration to MEL.

jdom:
@dotnet-bot test netfx-bvt
Reason: unrelated failure

jdom:
All tests passing, ready to review, and is mergeable on its own

src/Orleans/Configuration/ConfigUtilities.cs
```diff
@@ -47,7 +47,7 @@ internal static void ParseAdditionalAssemblyDirectories(IDictionary<string, Sear
             }
         }
 
-        internal static void ParseTelemetry(XmlElement root)
+        internal static void ParseTelemetry(XmlElement root, TelemetryConfiguration metricsConfiguration)
```
xiazen: nit : param name should be telemetryConfiguration

src/Orleans/Statistics/AverageTimeSpanStatistic.cs
```diff
@@ -153,9 +153,9 @@ public override string ToString()
             return GetValueString();
         }
 
-        public void TrackMetric(Logger logger)
+        public void TrackMetric(ITelemetryClient telemetryClient)
         {
-            logger.TrackMetric(currentName, GetCurrentValue());
+            telemetryClient.TrackMetric(currentName, GetCurrentValue());
```
xiazen: I'm not sure ITelemetryClient is a good name. Maybe TelemetryProducer? Since this is a public interface. I'd like more people's opinion. 
What do you think ? @sergeybykov @jason-bragg .

src/Orleans/Telemetry/ITelemetryClient.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Orleans.Runtime
+{
+    /// <summary>Represents an object used for producing telemetry</summary>
+    public interface ITelemetryClient
```
xiazen: I'm not sure ITelemetryClient is a good name. Maybe ITelemetryProducer? Since this is a public interface. I'd like more people's opinion.
What do you think ? @sergeybykov @jason-bragg .

src/Orleans/Telemetry/TelemetryLogConsumer.cs
```diff
@@ -0,0 +1,45 @@
...
+using System;
+using System.Net;
+
+namespace Orleans.Extensions.Logging
+{
+    /// <summary>
+    /// Forward trace log calls to the telemetry abstractions. This will be replaced by an ILoggerProvider in the future.
+    /// </summary>
+    public class TelemetryLogConsumer : ILogConsumer, IFlushableLogConsumer, ICloseableLogConsumer
+    {
```
xiazen: this class can be internal, since it is going to be removed in the future?

src/Orleans/Telemetry/TelemetryManager.cs
```diff
@@ -0,0 +1,259 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Runtime.Configuration;
+
+namespace Orleans.Runtime
+{
+    public class TelemetryManager : ITelemetryClient, IDisposable
```
xiazen: naming: I'd like it to be named `TelemetryProducer`. Since it doesn't really manage anything, all the configuration part is in `TelemetryConfiguration`. This is really just the producing part.

src/Orleans/Telemetry/ITelemetryClient.cs
sergeybykov: ITelemetrySource or ITelemetryProducer?

jason-bragg: I don't have a strong opinion on this.  ITelemetryClient, ITelemetryLogger, ITelemetrySink, ... ?

jdom: I considered both, but from the user's perspective, the source or the producer can be seen as a thing you subscribe to, not the client you use to publish telemetry yourself... I thought that client was less ambiguous from the user's perspective. Still open for debate, but wanted to share my reasoning after considering all of those.
As a 2 cent, client is also the name that AppInsights uses for their client api.

src/Orleans/Telemetry/TelemetryManager.cs
```diff
@@ -0,0 +1,259 @@
...
+        private List<ITelemetryConsumer> consumers = new List<ITelemetryConsumer>();
+        private List<IMetricTelemetryConsumer> metricTelemetryConsumers = new List<IMetricTelemetryConsumer>();
+        private List<ITraceTelemetryConsumer> traceTelemetryConsumers = new List<ITraceTelemetryConsumer>();
+
+        public IEnumerable<ITelemetryConsumer> TelemetryConsumers => this.consumers;
+
+        public void AddConsumers(IEnumerable<ITelemetryConsumer> newConsumers)
+        {
+            this.consumers = new List<ITelemetryConsumer>(this.consumers.Union(newConsumers));
+            this.metricTelemetryConsumers = this.consumers.OfType<IMetricTelemetryConsumer>().ToList();
```
xiazen: can we avoid allocating a new List here? Can we do` this.consumers.AddRange(newConsumers);` ?

src/Orleans/Telemetry/ITelemetryClient.cs
xiazen: I didn't like Client mainly because it kind of imply that there's TelemtryService, which is the AppInsight case, but not our case. In our case, this is just a abstraction for people to use to pipe telemetry to their chosen telemetry service. 

But I do agree your concern on 'producer'. So actually ITelemetrySink sounds better to me

xiazen: ITelemetrySource is good with me too.

galvesribeiro: Guys, remember that it is a **very** temporary/transitory PR. I'm going to redo the telemetry APIs as discussed with @jdom before, right after this PR get merged so I can try have it merged in time for 2.0 release. So this interface will be dropped right away so I think the name shouldn't matter.

jdom: @galvesribeiro still, anything that goes in needs to be in shippable state.
All: I'll rename to ITelemetryProducer. I don't think it's better than ITelemetryClient, but at least is not as ambiguous as Source. Sink would be the entire opposite... the sinks in a logging provider system are the log consumers

src/Orleans/Telemetry/TelemetryManager.cs
jdom: In this one I disagree, this hooks up the producer with all the consumers, and allows adding more consumers... It is managing.

src/Orleans/Telemetry/TelemetryManager.cs
jdom: what's the purpose? I don't want this to ever get out of sync, so I thought this would be the easiest to future proof that. Why are you concerned with this allocation? This probably might happen once at startup, if it ever happens.

src/Orleans/Telemetry/TelemetryLogConsumer.cs
jdom: I was going to say yes, but the reality is that as as standalone PR, this makes sense to be public. Sure, it will o away very soon and before we release it, which is good, as there will be no breaking change, but logically if we were to release tomorrow with this new API, then this class needs to be public

src/Orleans/Telemetry/TelemetryManager.cs
xiazen: this function is AddConsumers, not SetConsumers. So It can be called multiple times. I image the call pattern can be:
var telemetryManager  = new TelemetryManager();
telemetryManager.AddConsumers(newConsumers1);
telemetryManager.AddConsumes(newConsumesr2);
//inject into DI
services.AddSingleton<ITelemetryClient>(telemetryManager). 

User can call it like this, right?

src/Orleans/Telemetry/ITelemetryClient.cs
xiazen: @galvesribeiro hopefully this will be a transitory PR for 2.0. I'd like we have something better for telemetry in 2.0. But no commitment was made there ASAIK. That's why I insisted on the naming, since there's a good chance that this is going to be part of 2.0

src/Orleans/Telemetry/TelemetryManager.cs
xiazen: sure

src/Orleans/Telemetry/TelemetryLogConsumer.cs
xiazen: make sense. I was on a more conservative thinking on preserving original logic

src/Orleans/Telemetry/ITelemetryClient.cs
```diff
@@ -0,0 +1,66 @@
...
+        void TrackTrace(string message);
+
+        [Obsolete(TelemetryManager.ObsoleteMessageTelemetry)]
+        void TrackTrace(string message, Severity severityLevel);
+
+        [Obsolete(TelemetryManager.ObsoleteMessageTelemetry)]
+        void TrackTrace(string message, Severity severityLevel, IDictionary<string, string> properties);
+
+        [Obsolete(TelemetryManager.ObsoleteMessageTelemetry)]
+        void TrackTrace(string message, IDictionary<string, string> properties);
```
xiazen: just one more thing. Isn't we should mark IEventTelemetryConsuemer, ITraceTelemetryConsumer, IExceptionTelemetryConsumer etc, as obsolete too?

xiazen: Since we are deprecating its producing methods. And I think we were in a concensus that we are going to deprecating infrastructure for those telemetry events as a whole.

src/Orleans/Telemetry/ITelemetryClient.cs
galvesribeiro: @xiazen yeah, I know there is no commitments on changing it from the team. I committed myself with @jdom and others that I will do it before 2.0, so unless people is using bleeding edge TP nugets, or building themselves from source, nobody will even notice it. I was waiting @jason-bragg PR with facets and this PR to be merged so I can work on it.

src/Orleans/Telemetry/TelemetryManager.cs
jdom: correct, but that's the point, we are nowhere near a hot or warm path, this is just a builder-like method. Even if you call it 100 times at startup and then never again, I wouldn't worry about those allocations. I prefer that if we add a RemoveConsumer, or the current dispose logic, that there is no chance we can screw up.
Also, now that you made me look closer, I did this mainly because I want an immutable copy of the list, so if there are any writes in progress while we update this, that we don't have race conditions that throw, and using a ConcurrentBag is not a good option for something that is not likely to ever change after startup.

src/Orleans/Telemetry/ITelemetryClient.cs
jdom: This part I'm not sure, but welcome feedback. I didn't deprecate all methods because if you were actually using other methods, it's because you wanted specifically the APM methods. But if you were using it for tracing just because it was previously conflating the 2 scenarios, I wanted to have a build time warning by obsoleting this. Although by that token I could also osbsolete `TrackException`. Do you have any more thoughts on this? or @sergeybykov ?
With regards to the consumer interfaces themselves, maybe it's a good idea to deprecate `ITraceTelemetryConsumer` at the very least, so if you have a custom implementation of it you know that you might not be getting trace events by default in the future.

galvesribeiro: It's important to note that both metrics and exceptions are correlated. You need exception info/count correlated with some metrics. That way you can have more context and track a whole request with it.

Removing it and leave only traces for exception, will completely loose the context on a given request. Remember, traces are very limited in terms of information/context.

xiazen:

src/Orleans/Telemetry/TelemetryManager.cs
xiazen: "this is not a hot or warm path " is not the reason to do additional allocation than needed. And  the reason of getting a immutable copy is not enough to handle race condition. You have potential to lose data when two are trying to AddConsumers at the same time. 

To handle race condition, you either make it a ConcurrentBag or remove the AddConsumer method entirely so that application can only set it through constructor. A half solution of using Union is not solving either problem of the two.

xiazen: We can address this in another PR or even in my logging PR since we want to merge this PR ASAP. but just state my opinion here.

galvesribeiro:
Great! Will start the new APIs. Thanks @jdom

PR #3394, opened 2017-09-12
Fall back to full loading on .NET Core

jdom:
If ReflectionOnly assembly loading is not supported by the platform (which is the case in .NET Core), then fall back to fully loading every assembly for scanning.
In the future we will avoid this loading entirely, and only scan assemblies that were explicitly passed in (and pre-loaded) by the application developer.
In the meantime, the old behavior is preserved (by using ReflectionOnly) if the application is running in .NET.
This is a follow up from the PR https://github.com/dotnet/orleans/pull/3389 that was reverted

### DIFF
--- a/src/Orleans/CodeGeneration/TypeUtils.cs
+++ b/src/Orleans/CodeGeneration/TypeUtils.cs
@@ -512,6 +512,26 @@ namespace Orleans.Runtime
             return CachedTypeResolver.Instance.TryResolveType(fullName, out type);
         }
 
+        private static Lazy<bool> canUseReflectionOnly = new Lazy<bool>(() =>
+        {
+            try
+            {
+                CachedReflectionOnlyTypeResolver.Instance.TryResolveType(typeof(TypeUtils).AssemblyQualifiedName, out _);
+                return true;
+            }
+            catch (PlatformNotSupportedException)
+            {
+                return false;
+            }
+            catch (Exception)
+            {
+                // if other exceptions not related to platform ocurr, assume that ReflectionOnly is supported
+                return true;
+            }
+        });
+
+        public static bool CanUseReflectionOnly => canUseReflectionOnly.Value;
+
         public static Type ResolveReflectionOnlyType(string assemblyQualifiedName)
         {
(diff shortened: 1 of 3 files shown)


jdom:
Actually, hold on one sec, I think there is still a case that it might fail. Reproing locally

jdom:
PR should now be ready for review. Previously I was including a change from the old PR, but didn't notice that I did not undo it.
I tried running the repro app from @ashtonkj and it does work in .NET Core, although I have to publish the app first to make sure all DLLs are in the same folder (I believe this is a change from our TP preview2, as I believe we were following all references even when not publishing the app).

src/Orleans/CodeGeneration/TypeUtils.cs
```diff
@@ -512,14 +512,41 @@ public static bool TryResolveType(string fullName, out Type type)
...
+                CachedReflectionOnlyTypeResolver.Instance.TryResolveType(typeof(TypeUtils).AssemblyQualifiedName, out _);
+                return true;
+            }
+            catch (PlatformNotSupportedException)
+            {
+                return false;
+            }
+            catch (Exception)
+            {
+                // if other exceptions not related to platform ocurr, assume that ReflectionOnly is supported
```
ReubenBond: wouldn't false be a safer option?

ReubenBond:
LGTM besides that one comment. I hope that we can completely ditch reflection-only loading asap

src/Orleans/CodeGeneration/TypeUtils.cs
jdom: well, I added that as I was troubleshooting the other issue (where I unknowingly included a change from the previous PR, and was getting some loaded assemblies when I was expecting them to be reflection only load).
The reality is that I don't expect that condition to ever be hit (and I never saw it being hit even while troubleshooting the other error). But still, in the absolutely unlikely situation that it does occur, then getting a PlatformNotSupportedException immediately is the only thing that should turn this feature off. Other non-related failures should not turn off reflection only. So even though unlikely, the logic IMO is correct.

jdom:

PR #3395, opened 2017-09-12
Do not restore packages when packing

jdom:


jdom:
It's odd, but I'm looking at the output and I see that it still restores before packing. I see the same running Build.cmd locally, nevertheless if I have a much simpler test app, I can see that `--no-restore` is correctly being honored when packing

sergeybykov:
Do we still need this one?

PR #3396, opened 2017-09-12
Updated TestHost to use ISiloBuilder.

jason-bragg:
Tests can provide their own configured silo builder by registering an ISiloBuilderFactory using UseSiloBuilderFactory in the TestOptions.


jason-bragg:
Example

Call UseSiloBuilderFactory on test cluster options to specify a silo builder factory.

    public class Fixture : BaseTestClusterFixture
    {
        protected override TestCluster CreateTestCluster()
        {
            var options = new TestClusterOptions(1);
            options.UseSiloBuilderFactory<TestSiloBuilderFactory>();
            return new TestCluster(options);
        }
    }

Silo builder factory should create a configured silo builder.

    public class TestSiloBuilderFactory : ISiloBuilderFactory
    {
        public ISiloBuilder CreateSiloBuilder(
            string siloName,
            ClusterConfiguration clusterConfiguration)
        {
            return new SiloBuilder()
                .ConfigureSiloName(siloName)
                .UseConfiguration(clusterConfiguration);
                .ConfigureServices(ConfigureServices);
        }

        private void ConfigureServices(IServiceCollection services)
        {
            services.AddSingleton<TestPlacementStrategyFixedSiloDirector>();
        }
    }

It can also inherit from the test DefaultSiloBuilderFactory and just register the new services needed by the test.

    public class TestSiloBuilderFactory : DefaultSiloBuilderFactory, ISiloBuilderFactory
    {
        ISiloBuilder ISiloBuilderFactory.CreateSiloBuilder(
            string siloName,
            ClusterConfiguration clusterConfiguration)
        {
            return base.CreateSiloBuilder(siloName, clusterConfiguration)
                .ConfigureServices(ConfigureServices);
        }

        private void ConfigureServices(IServiceCollection services)
        {
            services.AddSingleton<TestPlacementStrategyFixedSiloDirector>();
        }
    }

src/OrleansTestingHost/AppDomainSiloHandle.cs
```diff
@@ -58,7 +58,7 @@ public static SiloHandle Create(string siloName, Silo.SiloType type, ClusterConf
                     optimizer.AddCachedAssembly(assembly.Key, assembly.Value);
                 }
 
-                var args = new object[] { siloName, type, config };
+                var args = new object[] { siloName, siloBuilderFactory, config };
```
jdom: I'm not sure I understand now how the SiloType is being used... In the past that was what defined whether the Primary silo has the membership and reminder grains, but even though that approach was not very good, I'm not understanding how you are achieving that now. Can you clarify?

jason-bragg: My understanding is that primary is now keyed off of the silo name, so the 'type' is no longer necessary.
@ReubenBond, is this correct?

src/Orleans/Configuration/ConfigurationExtensions.cs
```diff
@@ -63,6 +63,7 @@ public static class ConfigurationExtensions
         /// Configures all cluster nodes to use the specified startup class for dependency injection.
         /// </summary>
         /// <typeparam name="TStartup">Startup type</typeparam>
+        [Obsolete("Use ISiloBuilder")]
```
xiazen: the message can be more specific. The counterpart in SiloBuild's `ConfigureServices(Action<IServiceCollection> configureServices);` or `UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory);`. So the message should be "Using ISiloBuilder.ConfigureServices(Action<IServiceCollection> configureServices) or UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory)"

src/OrleansTestingHost/AppDomainSiloHost.cs
```diff
@@ -92,13 +97,20 @@ public void UpdateGeneratedAssemblies(GeneratedAssemblies collection)
         /// <summary>Starts the silo</summary>
         public void Start()
         {
-            silo.Start();
+            this.silo.StartAsync().Wait();
```
jdom: use `.GetAwaiter().GetResult()` instead of `.Wait()` here and in the Shutdown method

src/OrleansTestingHost/AppDomainSiloHost.cs
```diff
@@ -12,31 +11,37 @@
...
+        /// <param name="siloBuilderFactoryType">Type of silo builder factory.</param>
         /// <param name="config">Silo config data to be used for this silo.</param>
-        public AppDomainSiloHost(string name, Silo.SiloType siloType, ClusterConfiguration config)
+        public AppDomainSiloHost(string name, Type siloBuilderFactoryType, ClusterConfiguration config)
         {
-            this.silo = new Silo(name, siloType, config);
-            this.silo.InitializeTestHooksSystemTarget();
+            var builderFactory = (ISiloBuilderFactory)Activator.CreateInstance(siloBuilderFactoryType);
+            ISiloBuilder builder = builderFactory.CreateSiloBuilder(name, config);
+            builder.ConfigureServices((services) => services.AddSingleton<TestHooksSystemTarget>());
```
jdom: nit: this is not needed in the container... you can just use ActivatorUtilities to resolve one and register it as a sys target when you are initializing it.

src/OrleansTestingHost/DefaultSiloBuilderFactory.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Orleans.Hosting;
+using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+
+namespace Orleans.TestingHost
+{
+    public class DefaultSiloBuilderFactory : ISiloBuilderFactory
+    {
+        public ISiloBuilder CreateSiloBuilder(string siloName, ClusterConfiguration clusterConfiguration)
```
jdom: please make this virtual, otherwise derived types have to explicitly implement the interface just to override it, which doesn't play nicely with several levels of inheritance (and it's actually discouraged to override in that way)

src/OrleansTestingHost/DefaultSiloBuilderFactory.cs
```diff
@@ -0,0 +1,15 @@
...
+        {
+            ISiloBuilder builder = new SiloBuilder();
+            return builder.ConfigureSiloName(siloName).UseConfiguration(clusterConfiguration);
+        }
```
jdom: you could even also provide a protected virtual method called ConfigureServices, so that derived classes don't need to override the main CreateSiloBuilder just to call their own ConfigureServices, as 99% of the times, that's what they'll do

src/OrleansTestingHost/AppDomainSiloHost.cs
jason-bragg: It is needed in other areas, not just when initialized.

src/OrleansTestingHost/DefaultSiloBuilderFactory.cs
jason-bragg: Upon reflection, I think I'll make this sealed and force custom factory builders to be in test code.  The reason for this is that it will be clearer in the test code how the factory is being used.  This hides too much.

jdom:
In the future please avoid squashing commits when addressing feedback, or after the PR was reviewed. We will still squash everything when merging, but in the meantime it makes the review harder.
Can you say what changed since the last time to avoid a full review? I was almost ready to merge then, and I assume it's still mergeable, but cannot know for sure since the commits were squashed

jdom:
Ok, validated in person: only the Default factory is now sealed and the derived classes implement the factory completely. And the Transactions test that were recently merged were updated with this new pattern

PR #3397, opened 2017-09-13
Transactional State

jason-bragg:
- Added simple transactional state facet to support transactional persistance of a single state object.
- Uses storage providers via the storage bridge to persiste state.  Future version will support other options.
- Minimal tests


src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -0,0 +1,516 @@
...
+            catch (Exception)
+            {
+                try
+                {
+                    await this.storage.ReadStateAsync();
+                    DoRecovery();
+                }
+                catch (Exception ex)
+                {
+                    throw new InconsistentStateException($"Transactional state {this.config.StateName} is in an inconsistent state.  Force reload.", ex);
```
ReubenBond: I'm not confident in the safety here. How are we ensuring that the `storage.State` mutations don't affect subsequent calls to this activation? I'm still reviewing, so maybe it will become clear soon.

src/Orleans.Transactions/Hosting/SiloBuilderExtensions.cs
```diff
@@ -29,6 +36,16 @@ public static void UseInClusterTransactionManager(this IServiceCollection servic
...
         }
 
+        public static void UseTransactionalState(this IServiceCollection services)
+        {
+            // TODO: Move configuration to container configuration phase, once we move to silo builder in tests.
+            services.TryAddSingleton(typeof(ITransactionDataCopier<>), typeof(DefaultTransactionDataCopier<>));
+            services.AddSingleton<IAttributeToFactoryMapper<TransactionalStateAttribute>, TransactionalStateAttributeMapper>();
+            services.TryAddTransient<ITransactionalStateFactory, TransactionalStateFactory>();
+            services.AddTransient(typeof(ITransactionalState<>), typeof(TransactionalState<>));
+            services.AddTransient(typeof(TransactionalState<>));
```
ReubenBond: Is this last registration needed?

src/Orleans.Transactions/State/TransactionalStateFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public TransactionalStateFactory(IGrainActivationContext context)
+        {
+            this.context = context;
+        }
+
+        public ITransactionalState<TState> Create<TState>(ITransactionalStateConfiguration config) where TState : class, new()
+        {
+            TransactionalState<TState> transactionalState = ActivatorUtilities.CreateInstance<TransactionalState<TState>>(this.context.ActivationServices, config, this.context);
+            transactionalState.Participate(context.ObservableLifecycle);
```
ReubenBond: Could `TransactionalState<>` subscribe to the lifecycle in its ctor instead? I might have asked a similar question before in another context. Apologies if so.

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -0,0 +1,516 @@
...
+            if (info.IsReadOnly && readVersion.TransactionId > this.stableVersion)
+            {
+                throw new OrleansTransactionUnstableVersionException(info.TransactionId);
+            }
+
+            info.RecordRead(transactionalResource, readVersion, this.storage.State.StableVersion);
+
+            writeLowerBound = Math.Max(writeLowerBound, info.TransactionId - 1);
+
+            TState copy = this.copier.DeepCopy(readState);
```
ReubenBond: No change needed here, but if we know that `readState` was created for a read-only tx then we could avoid the deep copy. Just an idea.

src/Orleans.Transactions/State/TransactionalStateFactory.cs
jason-bragg: It 'can', it's just unsafe.  If it registers itself with the lifecycle then something goes wrong with the construction (error thrown) an invalid object will be registered in the lifecycle.

src/Orleans.Transactions/State/TransactionalState.cs
jason-bragg: There is still some wiring here related to readonly, but readonly transactions had issues and is not fully supported yet.  Readonly optimization will be revisited after primary path is baked.  If we can't get it right it will be removed entirely.

src/Orleans.Transactions/State/TransactionalState.cs
jason-bragg: All modifications of storage.state should be serially executed via storageExecutor.  I welcome double checking of this of course, I am fallible. :/

src/Orleans.Transactions/Hosting/SiloBuilderExtensions.cs
jason-bragg: nope.  Left over junk. :/

src/Orleans.Transactions/State/TransactionalState.cs
ReubenBond: Jason & I had a discussion and we decided that to ensure safety we will wrap all calls to the `AsyncSerialExecutor` so that they ensure that `this.storage.State` is consistent before continuing.

In this context, consistent means that `State` represents data that was successfully persisted.

PR #3407, opened 2017-09-16
Remove usages of Silo class from the container

jdom:
Remove usages of Silo class that were being resolved from the container.
This reduces the exposure of this big class as a dependency to other components.

It enables silo builders that do not register `Silo` in the container, which is useful as we experiment with the generic HostBuilder in Microsoft.Extensions.Hosting (pre-release)

### DIFF
--- a/src/OrleansRuntime/Silo/SiloControl.cs
+++ b/src/OrleansRuntime/Silo/SiloControl.cs
@@ -177,7 +177,7 @@ namespace Orleans.Runtime
                 try
                 {
                     var candidate = providerManager.GetProvider(providerName);
-                    if (candidate?.GetType()?.FullName?.Equals(providerTypeFullName) == true)
+                    if (string.Equals(providerTypeFullName, candidate?.GetType()?.FullName))
                     {
                         provider = candidate;
                         break;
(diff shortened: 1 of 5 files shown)


src/OrleansRuntime/Silo/SiloControl.cs
```diff
@@ -155,28 +171,33 @@ public Task<int> GetActivationCount()
...
         {
-            IReadOnlyCollection<IProvider> allProviders = silo.AllSiloProviders;
-            IProvider provider = allProviders.FirstOrDefault(pr => pr.GetType().FullName.Equals(providerTypeFullName) && pr.Name.Equals(providerName));
+            IProvider provider = null;
+            foreach (var providerManager in this.providerManagers)
+            {
+                try
+                {
+                    var candidate = providerManager.GetProvider(providerName);
+                    if (candidate?.GetType()?.FullName?.Equals(providerTypeFullName) == true)
```
ReubenBond: better to use string.Equals here and avoid the need to compare an expression to `true`.

ReubenBond:
Looks good, just that one comment. I'm glad you've made this change 👍

PR #3411, opened 2017-09-18
Fix leak in registration of partition notifications

elqueffo:
We were seeing a steadily increasing CPU usage on our service-fabric hosted silos & orleans clients. We tracked this down to Fabric Data Collection Agent, which writes ETW trace logs to disk (among other things, i assume). 
cpu_graph

After 3 days of uptime on our cluster, we were seeing about 700,000 events with the id 55639 per minute in the ETW logs and even though these events contain basically no useful information, we tracked this down to RegisterServicePartitionResolutionChangeHandler. Every time FabricServiceSiloResolver.Refresh() was called, the ETW logs would be spammed with all the current handler IDs as well as the new one. 

I believe one should not re-register for the notifications, especially if not cleaning up the old handler IDs for those partitions.

I opted to change the code to not re-register for partitions that were previously registered, to avoid constant re-registrations, and at the same time not leak old handler ids for existing partitions.

### DIFF
--- a/src/OrleansServiceFabricUtils/FabricServiceSiloResolver.cs
+++ b/src/OrleansServiceFabricUtils/FabricServiceSiloResolver.cs
@@ -81,6 +81,13 @@ namespace Microsoft.Orleans.ServiceFabric
                 foreach (var partition in this.silos)
                 {
                     var partitionInfo = partition.Partition;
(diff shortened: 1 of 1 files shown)


ReubenBond:
I believe you're right, @elqueffo, thank you for tracking this down.
Looks Good To Me

sergeybykov:
Great job, @elqueffo! Thank you!

PR #3415, opened 2017-09-18
Remove Legacy build flavor from all over the tree

attilah:
Remove OrleansHost project and tests
Cleanup csproj, targets and props files to have only the current information
Fix missing or wrong xml comments
Bump version to 2.0.0 preview 4
Enable warning as error for the solution as default


Directory.Build.props
```diff
@@ -33,117 +34,69 @@
...
-  <PropertyGroup Condition=" '$(BuildFlavor)' == 'Current' ">
+  <PropertyGroup>
     <VersionPrefix Condition=" '$(VersionPrefix)'=='' ">2.0.0</VersionPrefix>
-    <VersionSuffix Condition=" '$(VersionSuffix)'=='' ">preview3</VersionSuffix>
-  </PropertyGroup>
-
-  <PropertyGroup Condition=" '$(BuildFlavor)' == 'Legacy' ">
-    <VersionPrefix Condition=" '$(VersionPrefix)'=='' ">1.6.0</VersionPrefix>
-    <VersionSuffix Condition=" '$(VersionSuffix)'=='' "></VersionSuffix>
+    <VersionSuffix Condition=" '$(VersionSuffix)'=='' ">preview4</VersionSuffix>
```
jdom: please don't update this, since we will probably not update the preview number, just the build number (the date), unless we have bigger breaking changes

src/Orleans/Configuration/ConfigUtilities.cs
```diff
@@ -614,12 +614,8 @@ public static string RuntimeVersionInfo()
         {
             var sb = new StringBuilder();
             sb.Append("   Orleans version: ").AppendLine(RuntimeVersion.Current);
-#if BUILD_FLAVOR_LEGACY
-            // TODO: could use Microsoft.Extensions.PlatformAbstractions package to get this info
             sb.Append("   .NET version: ").AppendLine(Environment.Version.ToString());
             sb.Append("   OS version: ").AppendLine(Environment.OSVersion.ToString());
-            sb.Append("   App config file: ").AppendLine(AppDomain.CurrentDomain.SetupInformation.ConfigurationFile);
```
jdom: I'd leave the conditional compilation (although just around this line) if we didn't solve this yet, as a reminder of stuff we need to do before release (either decide to kill this or have some other way of troubleshooting, since it's not always straightforward to know which config file Orleans chose in some weird deployment cases)

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
```diff
@@ -60,146 +56,6 @@ public EHPurgeLogicTests()
             this.logger = new NoOpTestLogger().GetLogger(this.GetType().Name);
         }
 
-        //Disable tests if in netstandard, because Eventhub framework doesn't provide proper hooks for tests to generate proper EventData in netstandard
-#if BUILD_FLAVOR_LEGACY
```
jdom: @xiazen @jason-bragg can you validate if removing these is the right call now that we are fully migrating to the new package? Maybe they provide the hooks now?

src/Orleans/Configuration/ConfigUtilities.cs
attilah: If the host runs full .net it has access to this class and property and can be logged when troubleshooting.
Since we're compiling against .Net Standard 2.0 we do not so I don't see the value of keeping the #if here, but of course if you insist I can add that back, please advise.

Directory.Build.props
attilah: Sure.

jdom:
@dotnet-bot test this please
Reason: after merging the groovy file, build only Current

jdom:
@dotnet-bot test this please
Reason: after merging the groovy file, build only Current

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
xiazen: please don't delete those tests. if they doesn't pass because netstandard, you can disable or comment them for now. I will come back and fix them.

PR #3416, opened 2017-09-19
Transactions tweaks

jason-bragg:
Tighten up transactional state recovery logic
Use proper service provider in transactional state

### DIFF
--- a/src/Orleans.Transactions/State/TransactionalState.cs
+++ b/src/Orleans.Transactions/State/TransactionalState.cs
@@ -1,14 +1,15 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
+using Orleans;
 using Orleans.Core;
 using Orleans.Providers;
 using Orleans.Runtime;
 using Orleans.Storage;
 using Orleans.Transactions.Abstractions;
-using System.Threading;
 
 namespace Orleans.Transactions
 {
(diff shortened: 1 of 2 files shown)


src/Orleans.Transactions/Abstractions/ITransactionalExtension.cs
```diff
@@ -1,8 +1,10 @@
 ﻿using System.Threading.Tasks;
-using Orleans.Concurrency;
+using Orleans;
 using Orleans.Runtime;
+using Orleans.Concurrency;
+using Orleans.Transactions;
```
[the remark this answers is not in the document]
jason-bragg: Yeah.  Orleans.dll still uses Orleans.* namespaces..  Not sure what that was supposed to be so left it.

sergeybykov:
I think we better stay with Orleans.* namespaces, as we are today. I don't see any real value in making them more verbose by adding "Microsoft." in front. Keeping the current namespaces will make migration easier.

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -176,7 +179,15 @@ Task ITransactionalResource.Abort(long transactionId)
...
             if (transactionId > this.stableVersion)
             {
-                bool success = await this.storageExecutor.AddNext(() => GuardState(() => PersistCommit(transactionId)));
+                try
+                {
+                    bool success = await this.storageExecutor.AddNext(() => GuardState(() => PersistCommit(transactionId)));
+                } catch(Exception)
+                {
+                    // On error, queue up a recovery action.  Will do nothing if state recovers before this is processed
+                    this.storageExecutor.AddNext(() => GuardState(() => Task.FromResult(true))).Ignore();
```
ReubenBond: I don't quite understand the reasoning behind this. Just to clarify:
1. Is this merely an optimization to attempt to refresh the state from storage before the next call begins?
2. Could this be moved into GuardState?

jason-bragg: Yes, this attempts to reread the state from storage prior to the next persist call (either from prepare or commit).  It does this because we don't know when the next persist call will come through and we don't want to leave the facet in a bad state any longer than is necessary.  On persist error we queue up a recovery action which will try once to recover the state, if it does so prior to the next commit, the next commit will not need to recover first, if it fails, the persist will recover before attempting the persist.

PR #3419, opened 2017-09-19
Fix locks in GrainDirectoryHandoffManager 1/2

Vhab:
This is the result of finding `System.InvalidOperationException`: `Collection was modified; enumeration operation may not execute.`  in our logs on ApplicationInsights.

The following is not the prettiest formatting but I don't know of a prettier way to get callstacks out of AI off hand:

```
[  
   {  
      "assembly":"mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
      "method":"System.ThrowHelper.ThrowInvalidOperationException",
      "level":0,
      "line":0
   },
   {  
      "assembly":"mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
      "method":"System.Collections.Generic.List`1+Enumerator.MoveNextRare",
      "level":1,
      "line":0
   },
   {  
      "assembly":"OrleansRuntime, Version=1.5.1.0, Culture=neutral, PublicKeyToken=null",
      "method":"Orleans.Runtime.GrainDirectory.GrainDirectoryHandoffManager+<HandoffMyPartitionUponStop>d__11.MoveNext",
      "level":2,
      "line":0
   },
   {  
      "assembly":"mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
      "method":"System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw",
      "level":3,
      "line":0
   },
   {  
      "assembly":"mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
      "method":"System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification",
      "level":4,
      "line":0
   },
   {  
      "assembly":"OrleansRuntime, Version=1.5.1.0, Culture=neutral, PublicKeyToken=null",
      "method":"Orleans.Runtime.GrainDirectory.GrainDirectoryHandoffManager+<ProcessSiloStoppingEvent_Impl>d__14.MoveNext",
      "level":5,
      "line":0
   }
]
```

We're running Orleans on top of Service Fabric and have regular deploys going throughout the day while running load tests, which is probably what uncovered this.
Unfortunately I don't have an exact reproduction case.

The issue itself is straight forward. `silosHoldingMyPartition` inside `HandoffMyPartitionUponStop` is enumerated while not holding a lock.

On the surface it looks like it's inside a lock due to being called from `ProcessSiloStoppingEvent_Impl` which is called from `ProcessSiloStoppingEvent`, which holds a lock.
However, `ProcessSiloStoppingEvent_Impl` is async and thus continues running after `ProcessSiloStoppingEvent` has already released the lock.

The solution wasn't entirely straight forward to us. As it didn't feel appropriate to try to fix the behavior `ProcessSiloStoppingEvent` lock was trying to accomplish but for async, with for example a semaphore.

Instead I opted to modify `silosHoldingMyPartition` safely inside a lock in `ProcessSiloStoppingEvent_Impl`, and then pass a copy of the modified list to `HandoffMyPartitionUponStop` which will do various async operations.

It does feel like this class and the entry paths toward it need some looking at in general, due to the transition of sync to async.
However, it felt beyond the scope of this fix to take a deep dive on that.

### DIFF
--- a/src/OrleansRuntime/GrainDirectory/GrainDirectoryHandoffManager.cs
+++ b/src/OrleansRuntime/GrainDirectory/GrainDirectoryHandoffManager.cs
@@ -53,16 +53,16 @@ namespace Orleans.Runtime.GrainDirectory
             return null;
         }
 
-        private async Task HandoffMyPartitionUponStop(Dictionary<GrainId, IGrainInfo> batchUpdate, bool isFullCopy)
+        private async Task HandoffMyPartitionUponStop(Dictionary<GrainId, IGrainInfo> batchUpdate, List<SiloAddress> silosHoldingMyPartitionCopy, bool isFullCopy)
         {
-            if (batchUpdate.Count == 0 || silosHoldingMyPartition.Count == 0)
+            if (batchUpdate.Count == 0 || silosHoldingMyPartitionCopy.Count == 0)
             {
                 if (logger.IsVerbose) logger.Verbose((isFullCopy ? "FULL" : "DELTA") + " handoff finished with empty delta (nothing to send)");
                 return;
             }
 
             if (logger.IsVerbose) logger.Verbose("Sending {0} items to my {1}: (ring status is {2})", 
-                batchUpdate.Count, silosHoldingMyPartition.ToStrings(), localDirectory.RingStatusToString());
+                batchUpdate.Count, silosHoldingMyPartitionCopy.ToStrings(), localDirectory.RingStatusToString());
 
             var tasks = new List<Task>();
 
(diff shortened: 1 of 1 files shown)


benjaminpetit:
Thanks for this PR!

I agree that this class need some looking, especially since we don't have test coverage for this area as it can be considered an optimization.

In the meantime, your fix makes sense, so I will merge it if no one has any objection.

sergeybykov:
Thank you, @Vhab!

PR #3420, opened 2017-09-19
Fix locks in GrainDirectoryHandoffManager 2/2

Vhab:
This pull request is something uncovered while trying to fix the issue outlined in #3419

While we have not experienced an issue or seen any exceptions regarding this, it felt like a good opportunity to tackle this one as well.

Inside `GrainDirectoryHandoffManager` there are modifications made to `directoryPartitionsMap` while not holding a lock.
The suggested fix is straight forward, wrap the function bodies in a lock.

I opted to wrap the entire body in a lock() instead of trying to minimize the time the lock is held. This is done to mirror the style the rest of the code in this class.

### DIFF
--- a/src/OrleansRuntime/GrainDirectory/GrainDirectoryHandoffManager.cs
+++ b/src/OrleansRuntime/GrainDirectory/GrainDirectoryHandoffManager.cs
@@ -273,37 +273,43 @@ namespace Orleans.Runtime.GrainDirectory
 
         internal void AcceptHandoffPartition(SiloAddress source, Dictionary<GrainId, IGrainInfo> partition, bool isFullCopy)
         {
-            if (logger.IsVerbose) logger.Verbose("Got request to register " + (isFullCopy ? "FULL" : "DELTA") + " directory partition with " + partition.Count + " elements from " + source);
-
-            if (!directoryPartitionsMap.ContainsKey(source))
+            lock (this)
             {
-                if (!isFullCopy)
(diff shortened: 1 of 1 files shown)


sergeybykov:

benjaminpetit:
`directoryPartitionsMap` is also modified in `ProcessSiloAddEvent` and `ProcessSiloRemoveEvent`, and  `GetHandedOffInfo` iterates on its values.

From what I understood reading the code, 

- `ProcessSiloAddEvent` and `ProcessSiloRemoveEvent` are only scheduled on the `LocalGrainDirectory.CacheValidator` context from the `LocalGrainDirectory`.
- `AcceptHandoffPartition` and `RemoveHandoffPartition` are called from the systemtarget `RemoteGrainDirectory`, so scheduled on `LocalGrainDirectory.RemoteGrainDirectory` context

`LocalGrainDirectory.CacheValidator` and `LocalGrainDirectory.RemoteGrainDirectory` are both of type `RemoteGrainDirectory` but since they are not the same instance, there is indeed concurrency issue here.

Plus when we iterate on the vaues of `directoryPartitionsMap` in `GetHandedOffInfo`.

TLDR: no concurrency issue between `AcceptHandoffPartition` and `RemoveHandoffPartition` since they are called on the same context 'A', but since `ProcessSiloAddEvent` and `ProcessSiloRemoveEvent` are called on a different context 'B', we can have concurrency issues in thie class.

Unless I am totally wrong, and that we have indeed an issue, the fix is far more complex...

sergeybykov:
Thank you, @Vhab!

PR #3424, opened 2017-09-21
Build-time code generation for .NET Core

ReubenBond:
Note: this does *not* yet support building Orleans.sln on Linux or other platforms but it does allow building projects which use Orleans on Linux.

The build fails the first time (and subsequently succeeds) *when building in VS*. We can fix that by setting Project Dependencies in VS so that `Orleans` relies on `ClientGenerator.Bootstrap`, but that breaks cmdline builds. It's a known issue: https://github.com/Microsoft/msbuild/issues/2274

Removed `BootstrapCodegen.proj` because it doesn't play nicely with nuget / msbuild. Instead, I added 3 `*.Bootstrap` projects under a folder called `BootstrapBuild`. Each one corresponds to a project in `src` and they link their cs files from there. I added those projects to `Orleans.sln` so that msbuild correctly restores dependencies and initializes its internal structures.

The linked sources in those projects are hidden using properties on the `<Compile>`  csproj item. The `<Link>` metadata is used to hide the directory structure. This is just to reduce visual clutter in VS: only the Dependencies node is visible in the image below.

image

To make builds work on on .NET Core in the absence of AppDomains, we spawn a new process for code generation. Therefore we need a way to find the correct `dotnet` executable. This is done using an MSBuild task called `GetDotNetHost`.

Hopefully we can eliminate bootstrapping in the near future :)


ReubenBond:
Ok, I've implemented a workaround for the MSBuild bug: now it should happily build in VS / cmdline without trouble.

Directory.Build.props
```diff
@@ -43,6 +44,9 @@
   <PropertyGroup>
     <!-- System packages -->
     <SystemRuntimeVersion>4.3.0</SystemRuntimeVersion>
+    
+    <MicrosoftBuildVersion>15.3.409</MicrosoftBuildVersion>
```
attilah: nit: It should go to Microsoft* section in props, alphabetical order ;-) Just to make if easier to locate these.

src/ClientGenerator/GenerateOrleansCode.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
```
attilah: nit: Filename should be GetDotNetHost.cs

src/OrleansSQLUtils/OrleansSQLUtils.csproj
```diff
@@ -23,17 +23,17 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="CreateOrleansTables_MySQL.sql">
+    <None Include="CreateOrleansTables_MySql.sql">
```
attilah: Why changing the casing?

src/ClientGenerator/buildMultiTargeting/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
```diff
@@ -0,0 +1,3 @@
+<Project>
```
attilah: Microsoft.Orleans.CodeGenerator.Build.targets should be a better naming here, since we're removing the duplicate "Orleans" from everywhere, project names, package names, assembly names, etc.

src/OrleansSQLUtils/OrleansSQLUtils.csproj
sergeybykov: I believe it fails on non-Windows builds otherwise.

attilah: Makes sense, but then we should correct the casing of the files itself, not pascal casing SQL in the csproj IMHO.

ReubenBond: Yes, because the casing here is incorrect.

src/ClientGenerator/buildMultiTargeting/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
ReubenBond: The name aligns to the NuGet package name

attilah: But not by our new standards, right?

ReubenBond:
Pushed a bunch of commits to address feedback.

The solution also is sort-of buildable on nix now, but that is a non-goal of this PR. To make it make it build cleanly on nix, we would need to remove the Windows-only projects (Service Fabric, etc) from the sln and switch `TestProjectTargetFramework` in `Directory.Build.props` to something nix compatible, like `netcoreapp2.0`.

src/ClientGenerator/buildMultiTargeting/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
ReubenBond: if the package name changes, we can change this. Otherwise it won't be automatically added to the project when it's installed.

ReubenBond: Unless we modify the name when we pack it, like:
`<PackagePath>build\$(PackageId).targets</PackagePath>`. I'm not sure that's better, though - it might be confusing.

PR #3426, opened 2017-09-22
Orleans.Transactions.Azure

jason-bragg:
Added support for transactions in azure.
- Placeholder transaction log (transaction log interface is being redesigned)
- Transaction tests against azure


src/Orleans.Transactions.Azure/Storage/AzureTransactionLogConfiguration.cs
```diff
@@ -0,0 +1,17 @@
+﻿
+namespace Orleans.Transactions.Azure
+{
+    public class AzureTransactionLogConfiguration
+    {
+        public string ConnectionString { get; set; } = "UseDevelopmentStorage=true";
```
ReubenBond: I'm not sure we use this default - or any.

src/Orleans.Transactions.Azure/Storage/AzureTransactionLogStorage.cs
```diff
@@ -0,0 +1,364 @@
...
+                }
+
+            } while (continuationToken != null);
+        }
+
+        public static Factory<Task<ITransactionLogStorage>> Create(IServiceProvider serviceProvider)
+        {
+            return async () =>
+            {
+                ObjectFactory factory = ActivatorUtilities.CreateFactory(typeof(AzureTransactionLogStorage), Type.EmptyTypes);
```
ReubenBond: In this case, a factory is being created, used, then discarded. Would `ActivatorUtilities.CreateInstance<AzureTransactionLogStorage>()` be a more appropriate alternative?

jason-bragg: yeah, this code is silly.  I refactored the pattern a couple of times and didn't really think through the final code well enough..

src/Orleans.Transactions/Abstractions/ITransactionLogStorage.cs
```diff
@@ -8,13 +8,6 @@ namespace Orleans.Transactions.Abstractions
     /// </summary>
     public interface ITransactionLogStorage
     {
-        /// <summary>
-        /// Initialize the log (in Recovery Mode). This method must be called before any other method
-        /// is called on the log.
-        /// </summary>
-        /// <returns></returns>
-        Task Initialize();
```
xiazen: why did we remove the async init method?

xiazen: oh I see, because we are going the factory pattern

jason-bragg: Yes.  Trying to keep the interface surface geared towards usage pattern, avoiding construction/initialization.

src/Orleans.Transactions/Development/DevelopmentSiloBuilderExtensions.cs
```diff
@@ -1,24 +1,22 @@
 ﻿using Microsoft.Extensions.DependencyInjection;
 using Orleans.Hosting;
-using Orleans.Transactions.Abstractions;
 
 namespace Orleans.Transactions.Development
 {
```
xiazen: you don't have to address this in this PR. I'm just pointing this out: I think all new nugget are going Microsoft.Orleans.XXX namespace, instead of just Orleans

sergeybykov: We agreed that we'll leave *namespaces* as Orleans.*.

xiazen: I missed that stand up and maybe I got incorrect information. I thought we agreed that all old assembly which use Orleans.* keeps using it for backward-compact reasons. But new assembly use Microsoft.Orleans.*. Is this not the case?

sergeybykov: >I thought we agreed that all old assembly which use Orleans.* keeps using it for backward-compact reasons. But new assembly use Microsoft.Orleans.*. Is this not the case?

No, that's not the case. We agreed to stay with Orleans.* namespaces, and revert newly added ones to that convention, too.

xiazen:
LGTM

src/Orleans.Transactions.Azure/Orleans.Transactions.Azure.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <PackageId>Microsoft.Orleans.Transactions.Azure</PackageId>
+    <Title>Microsoft Orleans Transactions on Azure</Title>
+    <Description>Azure Transaction library of Microsoft Orleans used on the server.</Description>
+    <PackageTags>$(PackageTags) Azure Transactions</PackageTags>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.Orleans.Transactions.Azure</AssemblyName>
+    <RootNamespace>Orleans.Transactions.Azure</RootNamespace>
+    <OrleansBuildTimeCodeGen>true</OrleansBuildTimeCodeGen>
```
attilah: This will cause a test fail once rebased with the updated codege, since we're not generating any code for this assembly, please remove it. Later we'll add a warning to the build if codegen is specified, but we did not find anything to generate code for.

src/Orleans.Transactions.Azure/Orleans.Transactions.Azure.csproj
```diff
@@ -0,0 +1,26 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="WindowsAzure.Storage" Version="$(WindowsAzureStorageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Orleans.Core.Abstractions\Orleans.Core.Abstractions.csproj" />
```
attilah: nit: P2P references are also transitive, no need to specify 4 projects here, Orleans.Transactions would be suffice. If you choose to leave it like this we can clean all of these up later.

jason-bragg:
Addressed feedback + rebase

PR #3427, opened 2017-09-22
Add minimal Orleans.runsettings for tests.

ReubenBond:
These minimal runsettings can be used for unit tests. We can consider adding more settings to it afterwards (eg, code coverage / exclusions).

A very similar file is used in LoadTests

### DIFF
--- a/Orleans.runsettings
+++ b/Orleans.runsettings
@@ -1,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+  <RunConfiguration>
+    <TargetPlatform>x64</TargetPlatform>
+    <TargetFrameworkVersion>Framework45</TargetFrameworkVersion>
+  </RunConfiguration>
+</RunSettings>


Orleans.runsettings
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+  <RunConfiguration>
+    <TargetPlatform>x64</TargetPlatform>
+    <TargetFrameworkVersion>Framework45</TargetFrameworkVersion>
```
sergeybykov: Why `Framework45`?

jdom:
@dotnet-bot test functional

Orleans.runsettings
jdom: I asked the same thing, and there is not much documentation on it. But can we just not specify it then? Supposedly all of these settings have defaults. You are not required to provide a value for a setting you don't care, which *I believe* is our case, right? or was it added by design?

attilah:
After vstest was open sourced, the docs is also on github:

https://github.com/Microsoft/vstest-docs/blob/master/docs/configure.md

TargetFrameworkVersion:	Configures the target runtime for a test run. Possible values: any valid FrameworkName

I think for us it should be: .NETFramework, Version=v4.6.1

But testing it, VS does not seems to pick this up, is it for you @ReubenBond ?

ReubenBond:
It doesn't pick it up automatically :( It's a real shame that there seems to be no way to have this automatically configured

jdom:
The question is? What do we need to be picked up? If we don't provide a value for that property, doesn't it pick up the right framework anyway?

ReubenBond:
What I really would like is just the two options we have to set in VS every time we open from a clean repo

jdom:
yup, I get that, and would like that too. But back to this property that we are setting to something we don't understand. What happens if we don't set it, will it at least set the runner as x64?

attilah:
Since settings cannot be picked up by VS, I don't think this PR has any useful for us unfortunately :-(

jdom:
Why closed? I was just waiting for the answer on that other unrelated property. Even if not selected by default by the solution, having it there can be useful for people to troubleshoot why tests aren't appearing... Or are these settings unusable entirely?

ReubenBond:
The way it currently is, there's no value add. In the case of the LoadTests sln, the runsettings file is used by VSO, but that's not the case here. So I don't see much reason to add it.

PR #3428, opened 2017-09-22
Update My-First-Orleans-Application.md

mattcartwright:
Updated so it matches the default code created by the Dev/Test silo solution when using visual studio 2017 & the Orleans Tools for VS version 1.5.0


sergeybykov:
Thank you, @mattcartwright!

PR #3431, opened 2017-09-22
Align our HostBuilder infrastructure with Microsoft.Extensions.Hosting

jdom:
Microsoft.Extensions.Hosting (from https://github.com/aspnet/Hosting/issues/1163) is still in preview and won't be released for a few months. Nevertheless, we want to align as much as possible with it, so then migrating to it is not a big breaking change for our users, as well as trying to use common and familiar patterns.

This change has something I'd like to validate more openly with you before merging.

* Renamed `SiloHostBuilder` to `HostBuilder` and made it more generic, similar to the one in `Microsoft.Extensions.Hosting.HostBuilder` (but kept it in `Orleans.Hosting` namespace).
* Renamed `ISilo` to `IHost`. In the near term, the code in `SiloWrapper` will be rewritten as a generic `IHostedService`, similar to how we expect the integration with Microsoft.Extensions.Hosting to work. Basically a single generic IHost should be be able to host an Orleans silo, ASP.NET, etc.
* Added `HostBuilderContext` for the `HostBuilder`, as well as a Properties property bag, similar to the generic version we are emulating.
* Introduced `ISiloBuilder` as a non-generic silo builder. **I'd like feedback here!**
  * Currently this builder is used when calling the `ConfigureSiloHost` extension method, like this:

```
return new HostBuilder()
    .ConfigureServices(ConfigureServices)
    .ConfigureSiloHost(ISiloBuilder silo => silo  // Made type explicit here in snippet for review
        .ConfigureSiloName(siloName)
        .UseConfiguration(clusterConfiguration));
```
   * An alternate approach I considered is to return the `ISiloBuilder` instead of nesting it in a delegate, such as this:

```
return new HostBuilder()
    .ConfigureServices(ConfigureServices)
    .ConfigureSiloHost() // The silo builder could be returned here, instead of the original IHostBuilder
        .ConfigureSiloName(siloName)
        .UseConfiguration(clusterConfiguration);
```

By the way, it is not strictly required to introduce a builder, but I found it more explicit for cases where you would configure ASP.NET as well as Orleans (and potentially others in the future) into the same HostBuilder. If I didn't add the ISiloBuilder, then tons of extension methods would hang from IHostBuilder or ServiceCollection, and it could be tricky to understand the scope of the method (whether it configures something in Orleans or in ASP.NET).

Do you have thoughts especially on this last design question?

I will continue to polish this PR, but I wanted to get it out to gather feedback before I'm done with everything and then decide to go into a different approach.

### DIFF
--- a/src/Orleans.Transactions.Azure/Hosting/SiloBuilderExtensions.cs
+++ b/src/Orleans.Transactions.Azure/Hosting/SiloBuilderExtensions.cs
@@ -10,7 +10,7 @@ namespace Orleans.Transactions.Azure
         /// <summary>
         /// Configure cluster to use azure transaction log.
         /// </summary>
-        public static ISiloBuilder UseAzureTransactionLog(this ISiloBuilder builder, AzureTransactionLogConfiguration config)
+        public static ISiloHostBuilder UseAzureTransactionLog(this ISiloHostBuilder builder, AzureTransactionLogConfiguration config)
         {
             return builder.ConfigureServices(UseAzureTransactionLog)
                           .Configure<AzureTransactionLogConfiguration>((cfg) => cfg.Copy(config));
--- a/src/Orleans.Transactions/Development/DevelopmentSiloBuilderExtensions.cs
+++ b/src/Orleans.Transactions/Development/DevelopmentSiloBuilderExtensions.cs
@@ -9,7 +9,7 @@ namespace Orleans.Transactions.Development
         /// Configure cluster to use an in-memory transaction log.
         /// For development and test purposes only
         /// </summary>
-        public static ISiloBuilder UseInMemoryTransactionLog(this ISiloBuilder builder)
+        public static ISiloHostBuilder UseInMemoryTransactionLog(this ISiloHostBuilder builder)
         {
             return builder.ConfigureServices(UseInMemoryTransactionLog);
         }
--- a/src/Orleans.Transactions/Hosting/SiloBuilderExtensions.cs
+++ b/src/Orleans.Transactions/Hosting/SiloBuilderExtensions.cs
@@ -11,7 +11,7 @@ namespace Orleans.Transactions
         /// <summary>
         /// Configure cluster to use an in-cluster transaction manager.
         /// </summary>
-        public static ISiloBuilder UseInClusterTransactionManager(this ISiloBuilder builder, TransactionsConfiguration config)
+        public static ISiloHostBuilder UseInClusterTransactionManager(this ISiloHostBuilder builder, TransactionsConfiguration config)
         {
             return builder.ConfigureServices(UseInClusterTransactionManager)
                           .Configure<TransactionsConfiguration>((cfg) => cfg.Copy(config));
@@ -20,7 +20,7 @@ namespace Orleans.Transactions
         /// <summary>
         /// Configure cluster to support the use of transactional state.
         /// </summary>
-        public static ISiloBuilder UseTransactionalState(this ISiloBuilder builder)
+        public static ISiloHostBuilder UseTransactionalState(this ISiloHostBuilder builder)
         {
             return builder.ConfigureServices(UseTransactionalState);
         }
--- a/src/Orleans/Configuration/ConfigurationExtensions.cs
+++ b/src/Orleans/Configuration/ConfigurationExtensions.cs
@@ -63,7 +63,7 @@ namespace Orleans.Runtime.Configuration
         /// Configures all cluster nodes to use the specified startup class for dependency injection.
         /// </summary>
         /// <typeparam name="TStartup">Startup type</typeparam>
-        [Obsolete("Using ISiloBuilder.ConfigureServices(Action configureServices) or ISiloBuilder.UseServiceProviderFactory(IServiceProviderFactory factory)")]
+        [Obsolete("Using ISiloHostBuilder.ConfigureServices(Action configureServices) or ISiloHostBuilder.UseServiceProviderFactory(IServiceProviderFactory factory)")]
         public static void UseStartupType<TStartup>(this ClusterConfiguration config) 
         {
             var startupName = typeof(TStartup).AssemblyQualifiedName;
--- a/src/Orleans/Core/ClientBuilder.cs
+++ b/src/Orleans/Core/ClientBuilder.cs
@@ -33,7 +33,7 @@ namespace Orleans
             
             // Configure default services and build the container.
             this.serviceProviderBuilder.ConfigureServices(
-                services =>
+                (context, services) =>
                 {
                     services.TryAddSingleton(this.clientConfiguration ?? ClientConfiguration.StandardLoad());
                     services.TryAddFromExisting<IMessagingConfiguration, ClientConfiguration>();
@@ -42,7 +42,8 @@ namespace Orleans
                     services.TryAddFromExisting<ITraceConfiguration, ClientConfiguration>();
                 });
             this.serviceProviderBuilder.ConfigureServices(AddDefaultServices);
-            var serviceProvider = this.serviceProviderBuilder.BuildServiceProvider();
+
+            var serviceProvider = this.serviceProviderBuilder.BuildServiceProvider(null);
 
             serviceProvider.GetRequiredService<OutsideRuntimeClient>().ConsumeServices(serviceProvider);
 
@@ -62,7 +63,8 @@ namespace Orleans
         /// <inheritdoc />
         public IClientBuilder ConfigureServices(Action<IServiceCollection> configureServices)
         {
-            this.serviceProviderBuilder.ConfigureServices(configureServices);
+            if (configureServices == null) throw new ArgumentNullException(nameof(configureServices));
+            this.serviceProviderBuilder.ConfigureServices((context, services) => configureServices(services));
             return this;
         }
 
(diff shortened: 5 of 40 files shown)


src/OrleansRuntime/Hosting/ISiloBuilder.cs
```diff
@@ -1,39 +1,17 @@
...
+        IHostBuilder HostBuilder { get; }
+    }
 
-        /// <summary>
-        /// Specifies how the <see cref="IServiceProvider"/> for this silo is configured. 
-        /// </summary>
-        /// <param name="factory">The service provider factory.</param>
-        /// <returns>The silo builder.</returns>
-        ISiloBuilder UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory);
+    internal class SiloBuilder : ISiloBuilder
```
jason-bragg: I would expect a silo to be a service within a host, so it's builder would be called via the host builder.

When we move to the asp host won't ISilo be an IHostedService?

The pattern in this PR seems backwards to me.  Don't we need a silo builder that can be called by a host builder?  If so, the placeholder silo builder should not be placed inside the silo builder..

jdom: I don't think I understand what you meant in this particular PR, even after we talked about some "related" stuff in person. Can you clarify so that it's clear to me and others?

> When we move to the asp host won't ISilo be an IHostedService?

Yes, this part I haven't touched yet, be indeed the IHost implementation (which is currently SiloWrapper as of this PR) will be completely agnostic of Orleans, and will just run all `IHostedService` implementations, and `SiloHostService` will be our implementation of that interface. I didn't address this here, as this PR aligns with their pattern, but doesn't add an IHostedService executor or anything like that, as it is not meant to replace the generic host implementation to be used to host other services, it's just a temporary shim that behaves like theirs for the parts that we need.

jason-bragg: My understanding of our intended direction was for silos (grain containers that handle virtual actors) to be logically distinct from their hosting environment.  This allows silos to be hosted in the generic host (Microsoft.Extensions.Hosting), service fabric hosts, custom service frameworks, as well as other unforeseen hosting models.

Our primary scenario (for a number of reasons I won’t go into now) is to be hosted in a generic host (Microsoft.Extensions.Hosting.IHost).  As a hosted service in a generic host, our silo would be an IHostedService, so I’d suspect that the configuration/creation patterns we should align with are those supported for hosted services.  In general, hosted services get things from their hosting environment, but do not configure their host unless the host is understood to be a dedicated host (for that single service), which I don’t think we should assume.

By having the silo builder know about the host builder and configure the host builder we’re, imo, going against the expectations of hosted services.  Consider a host that wants to configure an ‘error’ log filter and a silo that tries to do the same.  What should happen?  Silo logging should either use the host’s logging, or configure its own, but it should not configure the entire host’s logging.

What we have in the code now is a silo builder that builds a hosted silo, and we need to separate the silo from the hosting to integrate with Microsoft.Extensions.Hosting.IHostBuilder.  In this separation we will need to introduce a placeholder host builder for the hosting and an ISiloBuilder which creates an ISilo hosted service.  What we have in this PR, however, looks like an ISiloBuilder that configures both the host and the silo, rather than a placeholder hosting and an ISiloBuilder that creates a silo to run in a hosted environment.

IMO, the simplest path forward, at this time, is to have an ISiloHostBuilder, much like our current ISiloBuilder, which is intended to build a dedicated host for a silo (and the silo).  When Microsoft.Extensions.Hosting is released we can update ISiloHostBuilder (and ISiloHost) to inherit from the generic hosting and remove any duplicate capabilities.  At that time, we can break out the silo specific construction logic into an ISiloBuilder that plays well with IHostBuilder, making the ISiloHost and ISiloHostBuilder superfluous (but still available for legacy reasons).

jdom: It is true that the current ISiloBuilder knows how to play only with IHostBuilder, and it was built as a sibling to it, as it's just a way to centralize and light up all extension methods that configure the host but are relevant to an Orleans silo. This is instead of having tons of methods applying to the generic host (or ServiceCollection), where one would be confused if an `AddAuthentication()` call would be configuring ASP.NET or Orleans or framework Foo, for example. This sibling builder is not technically required, and we can go back to configuring the generic host directly if that's the preferred direction, that's the whole purpose of this design PR, so I'll appreciate more thoughts on this.

Stuff that is not Orleans specific (such as Logging) would not be configured in the ISiloBuilder, but on the IHostBuilder or ServiceCollection directly.

BTW, it is not the goal to be able to host many Orleans silos inside a single IHost, as it is not possible to host multiple ASP.NET apps in a single host (as far as I know). That's why the extension methods of ISiloBuilder freely configure services in the single ServiceCollection from IHostBuilder.

If we inherit from the generic host builder, then it means that we loose the ability to co-host with other application frameworks such as ASP.NET, correct? Or if everyone followed a similar design no other application framework could be co-hosted, except maybe the ones that don't have special purpose builders. While that's technically possible, why would we want to isolate ourselves like that? I would think that's a step backwards in the design. In this case I would 100% favor not having sibling builders and deal with user confusion by having something in each extension method that refers to Silo or Orleans (such as `UseOrleansAzureTableMemebership`, ugly but unambiguous when co-hosting).

> My understanding of our intended direction was for silos (grain containers that handle virtual actors) to be logically distinct from their hosting environment. This allows silos to be hosted in the generic host (Microsoft.Extensions.Hosting), service fabric hosts, custom service frameworks, as well as other unforeseen hosting models.

This comment I do not follow. Microsoft.Extensions.Hosting should integrate in all of those environments, it does not "replace" those hosting options. This is why they have integration points (such as IHostLifetime) to bind it to an executable, Console, Windows Service, and I assume ServiceFabric or any host that can control the IHost lifetime externally.
Also, configuring the ServiceCollection IS the extensibility mechanism we are using, so it's either that or having a separate container from the Host vs the one Orleans uses or configures, which brings a lot of extra complexity for not a lot of gain. And the complexity is not just ours, as it can add a lot of friction for people co-hosting different IHotedServices and expecting the services they used to be shared too, and not having to include their host services into our container, and manage lifetimes and so on for the multiple containers. As a data point, I believe @DamianEdwards said they regretted introducing the Startup type that required building a separate container from the host for this exact reason.

jason-bragg: > Stuff that is not Orleans specific (such as Logging) would not be configured in the ISiloBuilder, but on the IHostBuilder or ServiceCollection directly.

This was an example, please address the general problem rather than the specific example.  The general problem is that if specific hosted services can specify host behaviors that can also be configured in the host we have an ambiguous and hard to manage system.  For this reason, configuration of hosted services should not configure the host.

> It is not possible to host multiple ASP.NET apps in a single host (as far as I know). That's why the extension methods of ISiloBuilder freely configure services in the single ServiceCollection from IHostBuilder.

I am not of the opinion that we can/should assume that an Orleans silo is the hosted app, only a hosted service.  Virtual actors enables scenarios in an application service, it should not be assumed, imo, that the Orleans silo IS the application.  I can easily envision a web server as the application service which also hosts an Orleans silo as part of an application layer, or a streaming service which hosts and Orleans service to manage streams of data.

> If we inherit from the generic host builder, then it means that we loose the ability to co-host with other application frameworks such as ASP.NET, correct?

I felt I was clear on describing the ISiloHost suggestion as a temporary approach which allowed us to move from the pattern we currently have to full support of IHost once it's released.  From my post:

> At that time, we can break out the silo specific construction logic into an ISiloBuilder that plays well with IHostBuilder, making the ISiloHost and ISiloHostBuilder superfluous (but still available for legacy reasons).

So this would just be a transient limitation until IHost is available.  This introduces no new limitations, only an easier migration path which does not violate the expected behavior of a hosted service.

> Configuring the ServiceCollection IS the extensibility mechanism we are using, so it's either that or having a separate container from the Host vs the one Orleans uses or configures, which brings a lot of extra complexity for not a lot of gain.

I'm not (at this time) arguing whether a silo should or should not have it's own container, only that a hosted service (in this case a silo, via the silo builder) should not be configuring the host's container.  If the silo is using the host's container we should configure this on the host.

I am of the opinion that the confusion between the responsibilities of a silo and host currently exist due to our current silo builder building a host and silo together.  Rather than breaking these apart now, without an IHost to build on, I suggest we simply maintain (for a time) an ISiloHost pattern which embodies the logic to create a silo along with a dedicated host (Orleans silo as the hosted application), which will likely always be a useful scenario, and refactor to support ISilo as a hosted service once IHost is available.

ReubenBond: > I am not of the opinion that we can/should assume that an Orleans silo is the hosted app, only a hosted service. Virtual actors enables scenarios in an application service, it should not be assumed, imo, that the Orleans silo IS the application

I believe we all agree here. However, each host can have at most 1 Orleans silo and at most 1 ASP.NET service, etc. This is the consequence of the single container design which seems to be where the ASP.NET team is currently going with the Generic HostBuilder.

> The general problem is that if specific hosted services can specify host behaviors that can also be configured in the host we have an ambiguous and hard to manage system.

The host itself is supposed to be so simple that it only needs lifecycle hooks & the set of IHostedService implementations which it hosts. The concern is that `IHostedService` implementations all share a single container so how do we know that they're not accidentally sharing something which is logically a per-IHostedService concern? Can anyone think of an examples where this might come up which would not be classed as a cross-cutting concern (and therefore should arguably be shared)?

I'm not against a multi-container model where the host has a DI container and each hosted service has a container. In that case, the `ISiloBuilder` would register an `IHostedService` and maybe an `IClusterClient` with the generic `IHostBuilder`?

We should aim for consensus with the ASP.NET folks on this. It's possible that it either hasn't come up yet or that they've already got a direction in mind which works for us too.

jdom:
We decided that until the generic host is fully released, we'll just have our own versions of IHost and IHostBuilder (called ISiloHost and ISiloHostBuilder for now, so that it's clear it's not the generic versions, even if that will prompt a breaking change in the future if we decide to deprecate them).
Also, we decided to not split extension methods so that they act on an inner Orleans specific builder just yet (removed the last commit in the previous version of the PR).
Rebased as there were many conflicts with the churn in the last few days.

In an upcoming PR we'll fully align the host with regards to lifetime management, instead of the Stopped property we currently have, so that people can reuse existing code for IHostLifetime for example and integrate it in many environments.

jdom:
@dotnet-bot test this please

jdom:
@dotnet-bot test this please
Reason: bot not responding

PR #3432, opened 2017-09-23
Transactional state storage abstraction

jason-bragg:
Added a storage abstraction tuned towards the transaction state transactional resource.
If a storage specific or application specific implementation is not provided, the transactional state storage will default to using storage providers.

### DIFF
--- a/src/Orleans.Transactions/Abstractions/INamedTransactionalStateStorageFactory.cs
+++ b/src/Orleans.Transactions/Abstractions/INamedTransactionalStateStorageFactory.cs
@@ -1,0 +1,8 @@
+﻿
+namespace Orleans.Transactions.Abstractions
+{
+    public interface INamedTransactionalStateStorageFactory
+    {
+        ITransactionalStateStorage<TState> Create<TState>(string storageName) where TState : class, new();
+    }
+}
--- a/src/Orleans.Transactions/Abstractions/INamedTransactionalStateStorageFactory.cs
+++ b/src/Orleans.Transactions/Abstractions/INamedTransactionalStateStorageFactory.cs
@@ -1,8 +1,17 @@
 ﻿
 namespace Orleans.Transactions.Abstractions
 {
+    /// <summary>
+    /// Factory which creates an ITransactionalStateStorage by name.
+    /// </summary>
     public interface INamedTransactionalStateStorageFactory
     {
+        /// <summary>
+        /// Create an ITransactionalStateStorage by name.
+        /// </summary>
+        /// <typeparam name="TState"></typeparam>
+        /// <param name="storageName">Name of transaction state storage to create.</param>
+        /// <returns>ITransactionalStateStorage, null if not found.</returns>
         ITransactionalStateStorage<TState> Create<TState>(string storageName) where TState : class, new();
     }
 }
--- a/src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
+++ b/src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
@@ -1,0 +1,66 @@
+﻿using Orleans.Concurrency;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Orleans.Transactions.Abstractions
+{
+    public interface ITransactionalStateStorage<TState>
+        where TState : class, new()
+    {
+        Task<TransactionalStorageLoadResponse<TState>> Load(string stateName);
+
+        Task<string> Persist(
+            string stateName,
+            string expectedETag,
+            string metadata,
+            List<PendingTransactionState<TState>> statesToPrepare);
+
+        Task<string> Confirm(
+            string stateName,
+            string expectedETag,
+            string metadata,
+            string transactionIdToCommit);
+    }
+
+    [Serializable]
+    [Immutable]
+    public class PendingTransactionState<TState>
+        where TState : class, new()
+    {
+        public PendingTransactionState(string transactionId, long sequenceId, TState state)
+        {
+            this.TransactionId = transactionId;
+            this.SequenceId = sequenceId;
+            this.State = state;
+        }
+
+        public string TransactionId { get; }
+
+        public long SequenceId { get; }
+
+        public TState State { get; }
+    }
+
+    [Serializable]
+    [Immutable]
+    public class TransactionalStorageLoadResponse<TState>
+        where TState : class, new()
+    {
+        public TransactionalStorageLoadResponse(string etag, TState committedState, string metadata, List<PendingTransactionState<TState>> pendingStates)
+        {
+            this.ETag = etag;
+            this.CommittedState = committedState;
+            this.Metadata = metadata;
+            this.PendingStates = pendingStates;
+        }
+
+        public string ETag { get; }
+
+        public TState CommittedState { get; }
+
+        public string Metadata { get;  }
+
+        public IReadOnlyList<PendingTransactionState<TState>> PendingStates { get; }
+    }
+}
--- a/src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
+++ b/src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
@@ -47,7 +47,7 @@ namespace Orleans.Transactions.Abstractions
     public class TransactionalStorageLoadResponse<TState>
         where TState : class, new()
     {
-        public TransactionalStorageLoadResponse(string etag, TState committedState, string metadata, List<PendingTransactionState<TState>> pendingStates)
+        public TransactionalStorageLoadResponse(string etag, TState committedState, string metadata, IReadOnlyList<PendingTransactionState<TState>> pendingStates)
         {
             this.ETag = etag;
             this.CommittedState = committedState;
--- a/src/Orleans.Transactions/State/NamedTransactionalStateStorageFactory.cs
+++ b/src/Orleans.Transactions/State/NamedTransactionalStateStorageFactory.cs
@@ -28,8 +28,9 @@ namespace Orleans.Transactions
                 ? this.context.ActivationServices.GetService<IStorageProvider>()
                 : this.context.ActivationServices.GetServiceByName<IStorageProvider>(storageName);
             if (storageProvider != null) return new TransactionalStateStorageProviderWrapper<TState>(storageProvider, context);
-
-            throw new InvalidOperationException($"Transactional storage {storageName} was not found.");
+            throw (string.IsNullOrEmpty(storageName))
+                ? new InvalidOperationException($"No default {nameof(ITransactionalStateStorageFactory)} nor {nameof(IStorageProvider)} was found while attempting to create transactional state storage.")
+                : new InvalidOperationException($"No {nameof(ITransactionalStateStorageFactory)} nor {nameof(IStorageProvider)} with the name {storageName} was found while attempting to create transactional state storage.");
         }
     }
 }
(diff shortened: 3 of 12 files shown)


src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        Task<TransactionalStorageLoadResponse<TState>> Load(string stateName);
+
+        Task<string> Persist(
+            string stateName,
+            string expectedETag,
+            string metadata,
+            List<PendingTransactionState<TState>> statesToPrepare);
+
+        Task<string> Confirm(
```
ReubenBond: Would the traditional 2PC names of Prepare/Commit be more suitable here?

src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public TState State { get; }
+    }
+
+    [Serializable]
+    [Immutable]
+    public class TransactionalStorageLoadResponse<TState>
+        where TState : class, new()
+    {
+        public TransactionalStorageLoadResponse(string etag, TState committedState, string metadata, List<PendingTransactionState<TState>> pendingStates)
```
ReubenBond: Not a big deal, but this ctor could expose `pendingStates` as an `IReadOnlyList` like the corresponding property.

src/Orleans.Transactions/State/NamedTransactionalStateStorageFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+                : this.context.ActivationServices.GetServiceByName<ITransactionalStateStorageFactory>(storageName);
+            if (factory != null) return factory.Create<TState>();
+
+            // Else try to get storage provider and wrap it
+            IStorageProvider storageProvider = string.IsNullOrEmpty(storageName)
+                ? this.context.ActivationServices.GetService<IStorageProvider>()
+                : this.context.ActivationServices.GetServiceByName<IStorageProvider>(storageName);
+            if (storageProvider != null) return new TransactionalStateStorageProviderWrapper<TState>(storageProvider, context);
+
+            throw new InvalidOperationException($"Transactional storage {storageName} was not found.");
```
ReubenBond: We could flesh this error message out more, since it's an important one. Something like:
`No implementation of either {nameof(ITransactionalStateStorageFactory)} or {nameof(IStorageProvider)} was found while attempting to create transactional state storage.`
or `No implementation of either (a) or (b) with the name (name) was found...`
depending on whether or not `storageName` is set.

If might make sense to if/else the entire method based on a single check for `string.IsNullOrEmpty(storageName)` in that case.

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -30,20 +30,29 @@ public class TransactionalState<TState> : ITransactionalState<TState>, ITransact
...
         private readonly AsyncSerialExecutor<bool> storageExecutor;
 
-        private IStorage<TransactionalStateRecord<TState>> storage;
         private ITransactionalResource transactionalResource;
 
+        // storage
+        private ITransactionalStateStorage<TState> storage;
+
+        // only to be modified at save/load time
+        private MetaData metaData;
```
ReubenBond: Common convention is to capitalize it as `Metadata`

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -80,7 +89,7 @@ public void Save()
             // Validation
             //
 
-            if (this.version.TransactionId > info.TransactionId || this.writeLowerBound >= info.TransactionId)
+            if (this.version.TransactionId > info.TransactionId || this.HighestRead >= info.TransactionId)
```
ReubenBond: Maybe I'm thinking about it the wrong way (and it's kinda early in the morning & pre-coffee) but would `HighestPersistedTransactionId` or `HighestWrittenTransactionId` be a clearer name? Maybe it's too verbose

src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
jason-bragg: We intentionally deviated from the 2pc naming when dealing with the storage to make it easier to talk about the algorithm.  If we use the term 'prepare' to refer to a signal sent to the transactional resource and for how that resource writes to storage, nuance is lost or hard to communicate.  Now one may say on prepare the resource persists the change.

src/Orleans.Transactions/State/NamedTransactionalStateStorageFactory.cs
jason-bragg: Agree that the error could be more fleshed out.  Will do.

As for the storage name being unset, that is valid.  In those cases it should use the default, it will only reach this condition if the name is set and no storage of that name exists, or name is unset and no default exists for either the storage or storage providers.

src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
ReubenBond: Ok, sounds fine to me

jason-bragg:
Addressed PR feedback.

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -31,20 +31,28 @@ public class TransactionalState<TState> : ITransactionalState<TState>, ITransact
...
         private readonly AsyncSerialExecutor<bool> storageExecutor;
 
-        private IStorage<TransactionalStateRecord<TState>> storage;
         private ITransactionalResource transactionalResource;
 
+        // storage
+        private ITransactionalStateStorage<TState> storage;
+
+        // only to be modified at save/load time
+        private Metadata metaData;
```
jdom: nit: field name should be `metadata`

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -501,6 +482,38 @@ public override string ToString()
...
+            {
+                get { return this.StableVersion.ToString(); }
+                set { this.StableVersion = (TransactionalResourceVersion.TryParse(value, out TransactionalResourceVersion version)) ? version : default(TransactionalResourceVersion); }
+            }
+
+            public long HighestReadTransactionId { get; set; }
+
+            public override string ToString()
+            {
+                return JsonConvert.SerializeObject(this);
```
jdom: I'm a little scared of just using the default settings by using JsonConvert, since a user might modify the global defaults for JsonConvert. We should use `JsonSerializer.Create` and pass explicit settings (and then cache to a static field)

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -257,7 +272,11 @@ Task ITransactionalResource.Abort(long transactionId)
                 records.ForEach(kvp => this.log.Remove(kvp.Key));
             }
 
-            await Persist(transactionId, this.writeLowerBound);
+            this.metadata.StableVersion = stableversion;
+            this.metadata.HighestVersion = this.version;
+            this.metadata.HighestReadTransactionId = this.highestReadTransactionId;
```
jdom: Q: is it safe to mutate metadata in several methods?

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -501,30 +482,59 @@ public override string ToString()
...
-    }
 
-    [Serializable]
-    public class LogRecord<T>
-    {
-        public T NewVal { get; set; }
-        public TransactionalResourceVersion Version { get; set; }
-    }
+        [Serializable]
+        private class Metadata
```
jdom: Do you need to mark this class Serializable? I don't think this is relying on binary formatter nor Orleans serializers, right? or is there a path where those might be used?

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -501,30 +482,59 @@ public override string ToString()
...
+                MissingMemberHandling = MissingMemberHandling.Ignore,
+                NullValueHandling = NullValueHandling.Ignore,
+                ConstructorHandling = ConstructorHandling.AllowNonPublicDefaultConstructor,
+                TypeNameAssemblyFormatHandling = TypeNameAssemblyFormatHandling.Simple,
+                Formatting = Formatting.None
+            };
+
+            [JsonIgnore]
+            public TransactionalResourceVersion HighestVersion { get; set; }
+            public string HighestVersionString
```
jdom: This approach is a little hacky, with different properties for parsing, since Json.NET supports converters. Also, the JSON will contain methods suffixed with String, is that by design? I imagine we want something that looks nice but is still evolvable, and hence the reason to use Json.NET?
BTW, I'm fine addressing in a follow up PR after the beta (but before final)

src/Orleans.Transactions/State/TransactionalState.cs
jason-bragg: It is only modified in persistence related calls, which are only performed in serial execution by the serial executor.

src/Orleans.Transactions/State/TransactionalState.cs
jason-bragg: The purpose of this was to have some metadata that the storage layer needs only store, but has no reason to process in any other way.
There was a request that the data be formatted in json, so it is human readable for debugging purposes.
The TransactionalResourceVersion is convertible to string to have a simple, easy to store version value, and since this class has versions it was agreed that the format should be the same, hence the stream property.

My preferences deviate from what is here, and we'll continue to negotiate with users of this interface as the tech is proven out.

All of this is debatable, and still in flux.

src/Orleans.Transactions/State/TransactionalState.cs
jason-bragg: We've gone back and forth a couple of times.  It's not needed in this incarnation, but it does no harm either.

src/Orleans.Transactions/State/TransactionalStateStorageProviderWrapper.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            return this.stateStorages.GetOrAdd(stateName, name => new StateStorageBridge<TransactionalStateRecord<TState>>(name, this.context.GrainInstance.GrainReference, storageProvider));
+        }
+    }
+
+    [Serializable]
+    public class TransactionalStateRecord<TState>
+        where TState : class, new()
+    {
+        public TState CommittedState { get; set; } = new TState();
```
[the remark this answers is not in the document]
jason-bragg: > Be called every time we initialize this object?
This field 'should' only be initialized once on initial create, but the behavior is provider dependent.

> Most of the times will be overwritten, isn't it?
It will be overridden on any commit, and possibly on load, depending on provider implementation.

There is likely some unnecessary allocations due to this pattern, but state management with providers has been a source of bugs before and this is the pattern we've used to address this in the past (see PubSubGrainState and it's history).  Since this class is specific to how this feature wraps the existing storage providers, it's safest to follow the proven patterns.  If we formalize around a more efficient state management for storage providers, and retrofit all of the providers to follow that pattern, we can address it here at that time.

jason-bragg:
Rebased onto master.  Comments addressed.

src/Orleans.Transactions/Abstractions/INamedTransactionalStateStorageFactory.cs
```diff
@@ -0,0 +1,8 @@
+﻿
+namespace Orleans.Transactions.Abstractions
+{
+    public interface INamedTransactionalStateStorageFactory
```
ReubenBond: I'm not sure, but does this need a doc comment since it's public?

jdom:
heads up, a PR with folders renamed was merged, so be careful when rebasing not to add files in the wrong folder

jason-bragg:
rebased

PR #3439, opened 2017-09-25
[Abstractions] Update to .NET Standard 2.0 and move base Orleans exception

benjaminpetit:
.net standard < 2.0 does not have the exception constructor with serialization stuff, so I had to update the target framework


benjaminpetit:
Sent to early, oops

PR #3440, opened 2017-09-25
[Abstractions] Update abstractions to .NET Standard 2.0 and move OrleansException

benjaminpetit:
.NET Standard < 2.0 does not have the exception constructor with serialization stuff, so I had to update the target framework

A lot of exceptions are in `Exceptions.cs` in Orleans. As `OrleansException` is the base exception, it needs to be moved first.

I created a folder in Abstractions to move all exceptions from `Exceptions.cs`.

### DIFF
--- a/src/Orleans.Core.Abstractions/Exceptions/OrleansException.cs
+++ b/src/Orleans.Core.Abstractions/Exceptions/OrleansException.cs
@@ -1,0 +1,28 @@
+using System;
+using System.Runtime.Serialization;
+
+namespace Orleans.Runtime
+{
+    /// <summary>
+    /// An exception class used by the Orleans runtime for reporting errors.
+    /// </summary>
+    /// <remarks>
+    /// This is also the base class for any more specific exceptions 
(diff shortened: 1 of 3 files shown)


src/Orleans.Core.Abstractions/Exceptions/OrleansException.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// An exception class used by the Orleans runtime for reporting errors.
+    /// </summary>
+    /// <remarks>
+    /// This is also the base class for any more specific exceptions 
+    /// raised by the Orleans runtime.
+    /// </remarks>
+    [Serializable]
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1058:TypesShouldNotExtendCertainBaseTypes")]
```
jdom: I think this is not needed, as `Exception` is fine to derive from. Can be dealt with in a separate PR though

PR #3444, opened 2017-09-26
[Abstractions] Move placement strategies to Abstractions

benjaminpetit:
- I had to make some class/properties public, but I think they should be public for configuration anyway,
- I didn't move `SystemPlacement` since I think it should only be used by Orleans internally


src/Orleans.Core.Abstractions/Placement/HashBasedPlacement.cs
```diff
@@ -6,7 +6,7 @@ namespace Orleans.Runtime
     public  class HashBasedPlacement : PlacementStrategy
     {
 
-        internal static HashBasedPlacement Singleton { get; } = new HashBasedPlacement();
+        public static HashBasedPlacement Singleton { get; } = new HashBasedPlacement();
```
sergeybykov: No need to make all these internal methods/properties/classes public if Abstractions have InternalsVisibleTo for Orleans, OrleansRuntime, and TesterInternal, which we need anyway.

PR #3448, opened 2017-09-26
[Abstractions] Move grain codegen attributes to Abstractions

benjaminpetit:


ReubenBond:
There are other attributes which affect codegen, I assume they'll move in another PR

PR #3452, opened 2017-09-26
Fix memory leak in timers

vansha:
See https://github.com/dotnet/orleans/commit/b9bad19cbf58c0570f590cc2eaceace366183325#commitcomment-24577719 for details


yevhen:
Critical.

sergeybykov:
Great catch! Thank you, @vansha and @Yevhen!

yevhen:
Guys, when we can expect patch release with this fix?

sergeybykov:
We are looking into a couple of other reported issues to assess if they need fixes in 1.5.2. Should clear those this week. I don't see yet why we can't release 1.5.2 by end of the week, unless we get blocked with code signing.

PR #3454, opened 2017-09-26
Fixed the sample NuGet issue

ashkan-saeedi-mazdeh:
There was a local NuGet repo configured for samples which was causing issues when trying to upgrade, removed it


jdom:
Thanks @ashkan-saeedi-mazdeh

PR #3455, opened 2017-09-26
[Abstractions] Move IRemindable to Abstractions

benjaminpetit:


src/Orleans.Core.Abstractions/Timers/IRemindable.cs
```diff
@@ -55,7 +55,7 @@ public struct TickStatus
             /// </summary>
             public DateTime CurrentTickTime { get; private set; }
 
-            internal static TickStatus NewStruct(DateTime firstTickTime, TimeSpan period, DateTime timeStamp)
+            public static TickStatus NewStruct(DateTime firstTickTime, TimeSpan period, DateTime timeStamp)
```
sergeybykov: No need to change this if Abstractions have `InternalsVisibleTo` for Orleans, OrleansRuntime, and TesterInternal, which we need anyway.

jdom: I don't think that's a good idea. For practical reasons we might have a few internals, but the goal of the abstractions is that there shouldn't be anything that is only accessible internally where we might require changing and revving this assembly. Making this particular thing as public makes sense (or even add a constructor instead). Making construction public also helps with mocking or replacing the reminder service

sergeybykov: I don't think it should be a goal to not have internal members in Abstractions for 2.0. It causes unnecessary churn of code and slows down reviewing of meaningful changes.

We can always do that later if needed. But pretty much all of this types/members haven't changed for years.

PR #3458, opened 2017-09-27
[Abstractions] Move Constants to Abstractions

benjaminpetit:
`Constants` is used almost everywhere, so it is very hard to not include it in Abstractions.

I removed all the `GrainId` constants from it in a separate file, and moved other constants in abstractions.


jdom:
I don't think moving constant not related to abstractions is a good thing. What's the purpose of this move?

PR #3463, opened 2017-09-27
Make namespace in Logging.Legacy to be Orleans.*

xiazen:


xiazen:
I was under the impression that new nugget should be under namespace Microsoft.Orleans.* 
but it turns out not be the case. hence this change

PR #3465, opened 2017-09-28
Move configuration system to an IOptions based one

attilah:
This PR is first of the n PRs which will add IOptions based configuration system for Orleans 2.0, replacing the old one.

The first class that was added is the divided version of IMessagingConfiguration. It is separated into its Client and Silo counterparts, SerializationProviderOptions was extracted into a separate class.

For easier transition LegacyConfigurationExtensions adds 1-1 extension method to support the mapping of old configuration objects to their IOptions counterpart for client and silo.

That is not a goal of the OR to replace every reference to IMessagingConfiguration and MessagingConfiguration, those changes will be applied incrementally to the codebase.

After the full migration of the old configuration system, the configuration classes and legacy support extension methods will be moved to the Microsoft.Orleans.Configuration.Legacy NuGet package.

- Add MessagingOptions classes
- Add IOption support for MessagingOptions classes
- Update SerializationManager to consume IOptions<SerializationProviderOptions>
- Update BufferPool to consume IOptions<MessagingOptions>
- Add Version suffix to MicrosoftExtensionsDependencyInjection
- Remove unused properties from MessagingConfiguration and their usages
- Remove MessagingConfiguration dependency from Message class
- Remove unused MessagingConfiguration private field from AsynchQueueAgent class
- Update OutsideRuntimeClient.ConsumeServices to use the ServiceProvider property consistently
- Remove redundant Globals property from SiloInitializatonParameters since it is accessible through the ClusterConfig property
- Add extension methods to simplify the support of legacy configuration objects

### DIFF
--- a/src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
+++ b/src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
@@ -7,7 +7,7 @@ namespace Orleans.Configuration
 {
     public static class LegacyConfigurationExtensions
     {
-        public static IServiceCollection AddLegacySiloConfigurationSupport(this IServiceCollection services)
+        public static IServiceCollection AddLegacyClusterConfigurationSupport(this IServiceCollection services)
         {
             // Global configuration
 
--- a/src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
+++ b/src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
@@ -1,0 +1,76 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Orleans.Runtime.Configuration;
+using System;
+
+namespace Orleans.Configuration
+{
+    public static class LegacyConfigurationExtensions
+    {
+        public static IServiceCollection AddLegacySiloConfigurationSupport(this IServiceCollection services)
+        {
+            // Global configuration
+
+            services.Configure<ClusterConfiguration, SiloMessagingOptions>((configuration, options) =>
+            {
+                CopyCommonMessagingOptions(configuration.Globals, options);
+
+                options.SiloSenderQueues = configuration.Globals.SiloSenderQueues;
+                options.GatewaySenderQueues = configuration.Globals.GatewaySenderQueues;
+                options.MaxForwardCount = configuration.Globals.MaxForwardCount;
+                options.ClientDropTimeout = configuration.Globals.ClientDropTimeout;
+            });
+
+            services.Configure<ClusterConfiguration, SerializationProviderOptions>((configuration, options) =>
+            {
+                options.SerializationProviders = configuration.Globals.SerializationProviders;
+                options.FallbackSerializationProvider = configuration.Globals.FallbackSerializationProvider;
+            });
+
+            return services;
+        }
+
+        public static IServiceCollection AddLegacyClientConfigurationSupport(this IServiceCollection services)
+        {
+            // Global configuration
+
+            services.Configure<ClientConfiguration, ClientMessagingOptions>((configuration, options) =>
+            {
+                CopyCommonMessagingOptions(configuration, options);
+
+                options.ClientSenderBuckets = configuration.ClientSenderBuckets;
+            });
+
+            services.Configure<ClientConfiguration, SerializationProviderOptions>((configuration, options) =>
+            {
+                options.SerializationProviders = configuration.SerializationProviders;
+                options.FallbackSerializationProvider = configuration.FallbackSerializationProvider;
+            });
+
+            return services;
+        }
+
+        private static void CopyCommonMessagingOptions(IMessagingConfiguration configuration, MessagingOptions options)
+        {
+            options.OpenConnectionTimeout = configuration.OpenConnectionTimeout;
+            options.ResponseTimeout = configuration.ResponseTimeout;
+            options.MaxResendCount = configuration.MaxResendCount;
+            options.ResendOnTimeout = configuration.ResendOnTimeout;
+            options.MaxSocketAge = configuration.MaxSocketAge;
+            options.DropExpiredMessages = configuration.DropExpiredMessages;
+            options.BufferPoolBufferSize = configuration.BufferPoolBufferSize;
+            options.BufferPoolMaxSize = configuration.BufferPoolMaxSize;
+            options.BufferPoolPreallocationSize = configuration.BufferPoolPreallocationSize;
+        }
+
+        private static void Configure<TConfiguration, TOptions>(this IServiceCollection services, Action<TConfiguration, TOptions> configureOptions) where TOptions : class
+        {
+            services.AddSingleton<IConfigureOptions<TOptions>>(serviceProvider =>
+            {
+                var configuration = serviceProvider.GetRequiredService<TConfiguration>();
+
+                return new ConfigureOptions<TOptions>(options => configureOptions(configuration, options));
+            });
+        }
+    }
+}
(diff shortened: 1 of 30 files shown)


src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
```diff
@@ -0,0 +1,76 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Orleans.Runtime.Configuration;
+using System;
+
+namespace Orleans.Configuration
+{
+    public static class LegacyConfigurationExtensions
+    {
+        public static IServiceCollection AddLegacySiloConfigurationSupport(this IServiceCollection services)
```
jdom: >AddLegacySiloConfigurationSupport [](start = 41, length = 33)

nit: ClusterConfiguration instead of SiloConfiguration

src/Orleans/Configuration/Options/SiloMessagingOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// Specifies global messaging options that are silo related.
+    /// </summary>
+    public class SiloMessagingOptions : MessagingOptions
+    {
+        /// <summary>
+        /// The SiloSenderQueues attribute specifies the number of parallel queues and attendant threads used by the silo to send outbound
+        /// messages (requests, responses, and notifications) to other silos.
+        /// If this attribute is not specified, then System.Environment.ProcessorCount is used.
+        /// </summary>
+        public int SiloSenderQueues { get; set; } = Environment.ProcessorCount;
```
jdom: I don't know if having an environment based default is the right approach for Options. Maybe this should be constant, and use the IPostConfigure infrastructure to configure it if it remains unset.

src/Orleans/Core/ClientBuilder.cs
```diff
@@ -37,6 +37,8 @@ public IClusterClient Build()
                 {
                     services.TryAddSingleton(this.clientConfiguration ?? ClientConfiguration.StandardLoad());
                     services.TryAddFromExisting<IMessagingConfiguration, ClientConfiguration>();
+                    // register legacy logging to new options mapping for Client options
```
jdom: legacy logging?

src/Orleans/Messaging/BufferPool.cs
```diff
@@ -38,9 +39,9 @@ public string Name
             private set;
         }
 
-        internal static void InitGlobalBufferPool(IMessagingConfiguration config)
+        internal static void InitGlobalBufferPool(IOptions<MessagingOptions> messagingConfigurationOptions)
         {
-            GlobalPool = new BufferPool(config.BufferPoolBufferSize, config.BufferPoolMaxSize, config.BufferPoolPreallocationSize, "Global");
+            GlobalPool = new BufferPool(messagingConfigurationOptions.Value.BufferPoolBufferSize, messagingConfigurationOptions.Value.BufferPoolMaxSize, messagingConfigurationOptions.Value.BufferPoolPreallocationSize, "Global");
         }
```
jdom: since this is setting the pattern, please access .Value just once first, and then use the options from it

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -27,6 +29,7 @@ internal class OutsideRuntimeClient : IRuntimeClient, IDisposable, IClusterConne
         private Logger callBackDataLogger;
         private ILogger timerLogger;
         private ClientConfiguration config;
+        private IOptions<ClientMessagingOptions> clientMessagingOptions;
```
jdom: just keep the ClientMessagingOptions, no need for the `IOptions<T>` wrapper, right?

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -130,19 +132,17 @@ internal IRuntimeClient RuntimeClient
...
-            var fallbackType = config.FallbackSerializationProvider;
             this.LargeObjectSizeThreshold = Constants.LARGE_OBJECT_HEAP_THRESHOLD;
             this.serializationContext = new ThreadLocal<SerializationContext>(() => new SerializationContext(this));
             this.deserializationContext = new ThreadLocal<DeserializationContext>(() => new DeserializationContext(this));
 
             logger = new LoggerWrapper<SerializationManager>(loggerFactory);
             this.serviceProvider = serviceProvider;
             RegisterBuiltInSerializers();
-            fallbackSerializer = GetFallbackSerializer(serviceProvider, fallbackType);
+            fallbackSerializer = GetFallbackSerializer(serviceProvider, serializatonProviderOptions.Value.FallbackSerializationProvider);
```
jdom: similar, just access the .Value once and then use that unwrapped object as many times as you want

src/Orleans/Configuration/Options/SiloMessagingOptions.cs
jdom: I realize after talking to Attila that this comment wasn't clear enough. What I meant is that the options POCO class should be very basic. Reading data from the environment is non-trivial (easy to do, but open to unforeseen side effects). For example, what if you send options to a remote server with different environment, or have a different default in a different platform.
Defaults for the strongly typed options should be constants/statics that do not change due to environment. That said, the Silo runtime itself does want to set it's own defaults based on the environment. That's where the PostConfigure method comes in, so by default the host builder can be configured to do the following to normalize the options:
```
services.PostConfigure<SiloMessagingOptions>(options => 
{
    if (options.SiloSenderQueues == 0) { options.SiloSenderQueues = Environment.ProcessorCount;
    if (options.GatewaySenderQueues == 0) { options.GatewaySenderQueues = Environment.ProcessorCount;
});

File: src/Orleans/Messaging/BufferPool.cs [hunk 7]
```diff
@@ -38,9 +39,11 @@ public string Name
             private set;
         }
 
-        internal static void InitGlobalBufferPool(IMessagingConfiguration config)
+        internal static void InitGlobalBufferPool(IOptions<MessagingOptions> messagingOptions)
         {
-            GlobalPool = new BufferPool(config.BufferPoolBufferSize, config.BufferPoolMaxSize, config.BufferPoolPreallocationSize, "Global");
+            var messagingOptionsValue = messagingOptions.Value;
```
jdom: nit for future reference: I would just call this `options`, not `messagingOptionsValue`

src/Orleans/Runtime/OutsideRuntimeClient.cs
```diff
@@ -27,6 +29,7 @@ internal class OutsideRuntimeClient : IRuntimeClient, IDisposable, IClusterConne
         private Logger callBackDataLogger;
         private ILogger timerLogger;
         private ClientConfiguration config;
+        private ClientMessagingOptions clientMessagingOptions;
```
jdom: I don't see any usages of this other than the assignment. What is it for?

jdom:
Looks great, thanks for addressing comments. I'll merge after I get home and the build finishes (unless someone has more comments)

src/Orleans/Messaging/BufferPool.cs
attilah: In extension methods I followed this, to make the body of those look like the same, but not locally. Once we set a standard here I agree to rename the ones where we're using only 1 type of options.

src/Orleans/Runtime/OutsideRuntimeClient.cs
attilah: we've references for "config" other places, if more usages will be migrated we'll could that variable, if not, we can remove it later of course, like I removed the unused config members from one of the classes.

PR #3467, opened 2017-09-28
[Abstractions] Move GrainId to Abstractions

benjaminpetit:
To break the dependency on `Interner` in `GrainId`, I introduced a `GrainIdFactory` that will live in `Orleans`.

The only small issue with this PR, is that now in the `GrainId.ToString()` we lost the capability of displaying the name of a system target id. We can find a proper way to fix it later, for now I added some conditional compilation to skip the dependency on `Constants`


xiazen:
LGTM. As of the `GrainId.ToString` issue. I think we can solve it by copy 

```
 private static readonly Dictionary<GrainId, string> systemGrainNames = new Dictionary<GrainId, string>
        {
            {SystemMembershipTableId, "MembershipTableGrain"},
            {SiloDirectConnectionId, "SiloDirectConnectionId"}
        };
```
to GrainIdFactory. I think that is all what GrainId.ToString() is using.

benjaminpetit:
If I copy that, then `GrainId` will have a dependency on `GrainIdFactory`, which lives in Orleans, so I wouldn't be able to move `GrainId` in abstractions :)

src/Orleans/IDs/GrainIdFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            return FindOrCreateGrainId(UniqueKey.NewKey(Guid.NewGuid(), UniqueKey.Category.Grain));
+        }
+
+        public static GrainId NewClientId(string clusterId = null)
+        {
+            return NewClientId(Guid.NewGuid(), clusterId);
+        }
+
+        internal static GrainId NewClientId(Guid id, string clusterId = null)
```
jdom: quick question (not necessarily prompting a change): why are these members internal, if the factory type itself is internal?

benjaminpetit: Because I just used resharper refactoring utilities. Yes I guess we could mark these members public in the future

xiazen:
not necessary, if you store the key of `systemGrainNames` as a string or guid, or something similar so that we can get systemTargetName based on the information just from `GrainId`. But yeah I'm not concerned by it. I don't think users are using `toString()` in serious business logic, or they shouldn't. You can create an issue recording it and we fix it later.

src/Orleans/IDs/GrainIdFactory.cs
jdom: yes, that's what I assumed. It is fine though for now anyway, I don't really mind, and the minimal diff is probably better at this point

jdom:
>         [JsonProperty]

can we get the same behavior without attributing this with newtsonsoft json? I think this is a bad dependency to bring into Abstractions, as it changes so often (mostly without breaking changes, but it does require binding redirects all over the place)

---
Refers to: src/Orleans.Core.Abstractions/IDs/UniqueIdentifier.cs:10 in 4b94dfc. [](commit_id = 4b94dfcc7fb3e2353e04fb584571642a9390aaa1, deletion_comment = False)

src/Orleans.Core.Abstractions/IDs/GrainId.cs
```diff
@@ -218,10 +126,12 @@ internal string ToDetailedString()
         private string ToStringImpl(bool detailed)
         {
             string name = string.Empty;
+#if ABSTRACTIONS_TODO
             if (Constants.TryGetSystemGrainName(this, out name))
```
jdom: if this info is important, you could be passing in the name of the system target in the factory method, and just store it here. We need to consider potential serialization issues with backwards compatibility though

benjaminpetit:
> can we get the same behavior without attributing this with newtsonsoft json? I think this is a bad dependency to bring into Abstractions, as it changes so often (mostly without breaking changes, but it does require binding redirects all over the place)

Not sure about that... I tried and I think some tests broke. Invoking @ReubenBond !

src/Orleans/IDs/GrainIdFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            return FindOrCreateGrainId(key);
+        }
+
+        internal static GrainId GetSystemGrainId(Guid guid)
+        {
+            return FindOrCreateGrainId(UniqueKey.NewKey(guid, UniqueKey.Category.SystemGrain));
+        }
+
+        internal static GrainId GetGrainIdForTesting(Guid guid)
```
jdom: If we were to have this method (and/or others) as a public utility in the TestingSiloHost project, would it avoid the issue of having way too many InternalsVisibleTo in Abstractions? Again, due to time, we could do that separately, but if it's possible, it might be good to keep in the list of things to change before 2.0

jdom:
BTW, talked with Ben in person. My recommendation to make further moves easier is to make Interner self-contained and just move it inside Abstractions (not public though). That would require almost not more changes (this PR would be trivial), as well as enable moving GrainReference whose serialization constructor requires using the factory methods for GrainId and SiloAddress.
I'm the same one that said the Abstractions shouldn't have logic, but the Interner is not really related to silo runtime, just a memory & cpu optimization for some of these data-types that are used over and over. And also we'll keep the diff during the move to a minimum to release 2.0-beta

sergeybykov:
@dotnet-bot test functional

PR #3468, opened 2017-09-28
[Abstractions] Move SiloAddress to Abstractions

benjaminpetit:
Following the same pattern as in #3467 

The only issue with this PR is that we cache the value returned by `GetConsistentHashCode()` and `GetUniformHashCodes` in this class, and these methods have dependencies that we want to avoid pulling in Abstractions.

As a temporary work-around, I marked the "cache members" as internal and set them in the `SiloAddressFactory` that is responsible for computing these values.

We could also compute these value when constructing the object; I don't think there will be a lot of performance hit, since, except in the client, we need theses values anyway.

### DIFF
--- a/src/Orleans/Core/Grain.cs
+++ b/src/Orleans/Core/Grain.cs
@@ -82,7 +82,7 @@ namespace Orleans
 
         internal SiloAddress SiloAddress
         {
-            get { return Runtime?.SiloAddress ?? SiloAddress.Zero; }
+            get { return Runtime?.SiloAddress ?? SiloAddressFactory.Zero; }
         }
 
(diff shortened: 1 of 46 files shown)


benjaminpetit:
@dotnet-bot test bvt
@dotnet-bot test functional

benjaminpetit:
@dotnet-bot test bvt
@dotnet-bot test functional

benjaminpetit:
@dotnet-bot test this please

sergeybykov:
@dotnet-bot test bvt

benjaminpetit:
Replaced by #3496

PR #3469, opened 2017-09-29
Remove ITraceConfiguration

xiazen:
- Remove trace related config from Orleans core and TestCluster. Now users are supposed to configure tracing through DI,  using Microsoft.Extensions.Logging. 
- And also enable TestCluster to use IClientBuilder, so that TestCluster doesn't lose the ability to configure logging in client side. 

These two are in two different commit 

EDIT: I reverted the commit of "enable TestCluster to use IClientBuilder to configure client", since I did that in #3538. We can merge #3538 first and I will rebase on top of that if there's any merge conflicts


src/OrleansTestingHost/DefaultSiloBuilderFactory.cs
```diff
@@ -15,7 +18,46 @@ public ISiloHostBuilder CreateSiloBuilder(string siloName, ClusterConfiguration
             return builder.ConfigureSiloName(siloName)
                 .UseConfiguration(clusterConfiguration)
                 .ConfigureLogging(loggingBuilder => TestingUtils.ConfigureDefaultLoggingBuilder(loggingBuilder,
-                    clusterConfiguration.GetOrCreateNodeConfigurationForSilo(siloName).TraceFileName));
+                    GetLegacyTraceFileName(siloName, DateTime.UtcNow)));
+        }
+
+        internal static string GetLegacyTraceFileName(string nodeName, DateTime timestamp, string traceFileFolder = "logs", string traceFilePattern = "{0}-{1}.log")
+        {
```
jdom: Hmmm, this is an odd name for a non-legacy feature. Also it's in an odd class.
Also, there is a ton of logic here that doesn't make sense. Probably the easiest trace file name is to concatenate `$"logs\\{DeploymentId}_{nodeName}"`
The deployment ID should already have the timestamp by default

xiazen: Why is this not a legacy feature? I marked it as internal since I think we are moving away from supporting trace file name configuration, which is part of ITraceConfiguration. User should configure their own file name from now on. 

and it is an internal method. Actually it maybe should be a private method.

jdom: What I mean, having a log file per node & client is not legacy, we want that moving forward, and not just as an internal way of doing things. Our end users would need this working easily by default. It should be an accessible utility. Could even be a public static utility in `TestClusterOptions` if there's no better place to put it

xiazen: OK it make sense. Will make it an public utility method. Part of the purpose of this PR is also testing water on what we should continue supporting as legacy or what not. Since I'm definitely in the other end of more breaking change and less supporting legacy

xiazen: BTW I reread your comment. Yeah there's a lot of logic here which doesn't make sense but just for context that it is copied from old setTraceFileName function. 
But I think $"logs\\{DeploymentId}_{nodeName}" is good enough and much clearer here.

jdom:
As we talked about, the previous PR that partially removed ITraceConfiguration also mistakenly removed the parsing of PropagateActivityId.
We should bring that back. Other than that, this PR looks OK. We'll have to create a default silo/client builder for our tests that we build upon, as it's becoming a little unwieldy to duplicate the logic everywhere

xiazen:
yeah. agree. The thing is those duplicate logic are not really the same, otherwise we will be just using DefaultCLientBuilder or DefaultSiloHostBuilder. They try to configure stuff on top of the default client and silo builder. Maybe we can create some abstraction to make that easier.

jdom:
BTW, I think it's appropriate that in the future we move PropagateActivityId to the MessagingOptions class. Still in the legacy config it makes sense that we still parse it from the same place in the XML

xiazen:
yeah agree. it should be part of MessagingOptions class , where it is most related

src/Orleans.Runtime/Silo/SiloHost.cs
```diff
@@ -29,19 +29,6 @@ public class SiloHost :
...
-        /// </summary>
-        /// <remarks>
-        /// <para>
-        /// The values of <c>null</c> or <c>"None"</c> mean no log file will be written by Orleans Logger manager.
-        /// </para>
-        /// <para>
-        /// When deciding The values of <c>null</c> or <c>"None"</c> mean no log file will be written by Orleans Logger manager.
-        /// </para>
-        /// </remarks>
-        public string TraceFilePath { get; set; }
```
jdom: Looking at this made me realize that SiloHost cannot be configured to use a new filename every time.
I think that we don't need this to be configurable, but at least default to using a filename like $SiloHost-{siloName}_{DateTime.UtcNow.ToString(dateFormat)}.log".
Also, I would skip the ILoogerFactory entirely, and just create the FileLoggerProvider. Also remember to dispose the provider (or factory if you don't do the latter suggestion) when disposing this class

jdom:
@dotnet-bot test functional

src/Orleans.Runtime/Silo/SiloHost.cs
enyim: There should be a way disabling the creation of log files by the silo. E.g. the location the silo is running from is read-only, or the adminstrator wants to control where and when are random files created.

jdom: @enyim SiloHost is still there for legacy reasons. The recommendation is to move to SiloHostBuilder, and there you can fully configure logging using Microsoft.Extensions.Logging. Would that address your concern?

PR #3473, opened 2017-09-30
Transaction manager tests and minor bug fixes

jason-bragg:


jason-bragg:
@dotnet-bot test functional

src/Orleans.Transactions/InClusterTM/TransactionManager.cs
```diff
@@ -41,15 +42,15 @@ internal class TransactionManager : ITransactionManager
...
-
         private bool IsRunning;
         private Task transactionLogMaintenanceTask;
 
-        public TransactionManager(TransactionLog transactionLog, IOptions<TransactionsConfiguration> configOption, ILoggerFactory loggerFactory)
+        public TransactionManager(TransactionLog transactionLog, IOptions<TransactionsConfiguration> configOption, ILoggerFactory loggerFactory, TimeSpan? logMaintenanceInterval = null)
         {
             this.transactionLog = transactionLog;
             this.config = configOption.Value;
             this.Logger = new LoggerWrapper<TransactionManager>(loggerFactory);
```
xiazen: consider migrate off Logger and use ILogger  if you have time :)

test/Orleans.Transactions.Azure.Test/GoldenPathTransactionManagerTest.cs
```diff
@@ -0,0 +1,47 @@
...
+        private static readonly TimeSpan StorageDelay = TimeSpan.FromSeconds(30);
+
+        public GoldenPathTransactionManagerTest(ITestOutputHelper output)
+            : base(MakeTransactionManager(), LogMaintenanceInterval, StorageDelay, output)
+        {
+        }
+
+        private static ITransactionManager MakeTransactionManager()
+        {
+            ILoggerFactory loggerFactory = new LoggerFactory();
```
xiazen: You can use NullLoggerFactory.Instance if you want want your logger to log anything

test/Orleans.Transactions.Tests/Memory/GoldenPathTransactionManagerMemoryTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        private static readonly TimeSpan StorageDelay = TimeSpan.FromMilliseconds(1);
+
+        public GoldenPathTransactionManagerMemoryTests(ITestOutputHelper output)
+            : base(MakeTransactionManager(), LogMaintenanceInterval, StorageDelay, output)
+        {
+        }
+
+        private static ITransactionManager MakeTransactionManager()
+        {
+            ILoggerFactory loggerFactory = new LoggerFactory();
```
xiazen: same here

test/Orleans.Transactions.Tests/Runners/GoldenPathTransactionManagerTestRunner.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        protected GoldenPathTransactionManagerTestRunner(ITransactionManager transactionManager, TimeSpan logMaintenanceInterval, TimeSpan storageDelay, ITestOutputHelper output)
+        {
+            this.transactionManager = transactionManager;
+            this.logMaintenanceInterval = logMaintenanceInterval;
+            this.storageDelay = storageDelay;
+            this.output = output;
+        }
+
+        [SkippableFact]
```
xiazen: if it is `SkippableFact`, then you should have a method in your constructor to determine whether the precondition are met and whether to skip the test. For example, in `AzureStorageBasicTests`, it calls   `TestUtils.CheckForAzureStorage();` to check if azure storage is availale, if not, then throw a `SkipException` which would cause the tests to skip. 
I think you need to do that for at least the `GoldenPathTransactionManagerTest ` since it is using Azure storage?

xiazen:
the test failure should be fixed by #3482 , which is merged into master.

jason-bragg:
Rebased + Addressed comments

src/Orleans.Transactions/InClusterTM/TransactionManager.cs
```diff
@@ -41,15 +41,15 @@ public class TransactionManager : ITransactionManager
...
+        protected readonly ILogger logger;
         private bool IsRunning;
         private Task transactionLogMaintenanceTask;
 
         public TransactionManager(TransactionLog transactionLog, IOptions<TransactionsConfiguration> configOption, ILoggerFactory loggerFactory, TimeSpan? logMaintenanceInterval = null)
         {
             this.transactionLog = transactionLog;
             this.config = configOption.Value;
-            this.Logger = new LoggerWrapper<TransactionManager>(loggerFactory);
+            this.logger = loggerFactory.CreateLogger(nameof(TransactionManager));
```
xiazen: loggerFactory.CreateLogger<TransationManager>().

src/OrleansTestingHost/Utils/StorageEmulator.cs
```diff
@@ -141,6 +141,32 @@ public static bool Stop()
             }
         }
 
+        public static bool TryCheckForAzureStorage(string dataConnectionString, out string error)
```
xiazen: I'm not sure this method should be brought to the public package, since what it is doing is not directly related to `StorageEmulator` and its behavior is not well defined. It did many things, first check `dataConnectionString` empty or not, then check if the user intend to use the emulator. If not then return true. But this method didn't check if the user really have a remote azure storage service configured. So in that path, the behavior is uncertain. At last if set to use local emulator, then start the emulator. This part is related but the `StorageEmulator` api already covers that. 

I prefer not to put this method in the public API. we can put it in `TestExtensions.csproj ` which is a common dependency for our test projects.

@jdom what do you think ?

jdom: I agree, this isn't a general utility, so it shouldn't be part of the testing nuget package that we ship.
Having said that, StorageEmulator itself is currently part of this package, and it itself needs to be removed entirely. Adding this method temporarily is not a huge concern to me, as we have a pending task to remove the whole class. Nevertheless, it might be bad to add more stuff if we plan to deprecate it.

jdom: BTW, I agree that the TestExtensions project is where this should go (and eventually StorageEmulator too)

jason-bragg: I think the proper home of this behavior depends on what pattern we want to support for extension developers.  We should, ourselves, try to develop extension libraries as we expect others to.  While we're not strictly following this ATM, I am of the opinion this should be our goal.

In general, I think this behavior makes sense as part of the azure emulator test support, so it's in the correct class.  If that class is in the wrong assembly, we should address that once we decide where it belongs, rather than peppering it across multiple projects.

jdom: Well, the distinction is that this method just uses the StorageEmulator as a fallback, but if the connection string is not with the emulator's connection string, then it will not try to start it. So in theory it is separable from the StorageEmulator class anyway.

xiazen: I'm of the opinion that azure emulator test support should be azure team's responsibility. Since it is purely relate to them, not us. It make sense that we have it internally for our internal use. But I wouldn't agree more than that. 
BTW I didn't know StorageEmulator is going to be removed. If it is going to be removed in 2.0 beta, then sure let's merge this PR and remove it later. If not, we shouldn't add it in our public package and had it released in 2.0 beta

test/Orleans.Transactions.Azure.Test/TestFixture.cs
```diff
@@ -33,5 +34,14 @@ public ISiloHostBuilder CreateSiloBuilder(string siloName, ClusterConfiguration
                     .UseTransactionalState();
             }
         }
+
+        public static void CheckForAzureStorage()
```
jdom: I highly recommend that you use BaseAzureTestClusterFixture (or a similar approach). The way you are using it here will only skip the test after it attempted to start the test cluster (and will possible throw an exception and fail the test, instead of marking it as skipped).
The gist of it is that the fixture should not try to create a cluster if it cannot use azure storage (but without throwing), and then the test class itself will just call `fixture.EnsurePreconditionsMet()` throw that it's correctly skipped

jason-bragg: > I highly recommend that you use BaseAzureTestClusterFixture 

BaseAzureTestClusterFixture is in tester which pulls in an unreasonable number of dependencies.  I shouldn't need to use tester to write test for extension libraries.

> The way you are using it here will only skip the test after it attempted to start the test cluster

I don't mind making it part of fixture, but the pattern I'm using throws during the test class construction, which is complete before the cluster is created, correct?

> So in theory it is separable from the StorageEmulator class anyway.

Logically separable, yes, but dependent on it and related to our storage emulator test support, which is in src/OrleansTestingHost/Utils/StorageEmulator.cs.

I don't think this adds (relatively) any meaningful risk to test host, but I'll remove the call all the same.

jdom:
LGTM.
Nevertheless, as we talked about, the test classes that do still use `TestFixture`, should call `fixture.EnsurePreconditionsMet()` in their constructor.
I'll merge this PR, as there are no new usages of that fixture here.

PR #3475, opened 2017-09-30
HelloWorld.NetCore Sample: upgrade to dotnet core 2.0 and use latest orleans packages

claylaut:
Upgrade to .NET Core 2 (fix solution since it wasn't building with .NET CORE 1 and Orleans Preview 3)
Used latest packages for Client and Silo (`Microsoft.Orleans.Server` and `Microsoft.Orleans.Client`)
Updated Orleans `Client` to start using `IClusterClient` instead of `GrainClient.GrainFactory`


jdom:
Thank you, @claylaut!

claylaut:
@jdom welcome 👍

PR #3485, opened 2017-10-02
AWSUtils SQS allow localhost connection

asterius1:
You can now use connection strings like: `Service=http://localhost:20003`.
It allows usage of elasticmq (SQS drop in replacement) which is useful for local development.

Code for it I took from DynamoDBStorage with minimal changes.
Didn't know if I should make a test for it.
Works for me.


jdom:
Thanks @asterius1 !

Issue #3429, opened 2017-09-22
Orleans logging migrate to MEL remaining issues


xiazen:
This is the issue list keeping tracking breaking changes in logging after pr #3333. Some breaking changes we want to bring back in subsequent PRs, some needs more discussion and decision making. This is the tracking list to remind ourselves. 
- TestClusterOptions TraceToConsole was removed, so that user cannot configure TraceToConsole through TestClusterOptions anymore. One need to configure that in a Microsoft.Extensions.Logging fashion, through ISiloBuilder and ISiloBuilderFactory. 
- UnobservedExceptionsHandlerClass's logging is initialized through a static method and its logger is static too. This won't work in many cases, for example when client and silo share the same AppDomain, the logging config in client side and silo side will be mixed up in UnobservedExceptionsHandlerClass
- Same with SchedulerStatisticsGroup.cs Init(ILoggerFactory) method
- We removed most trace related config in ClientConfiguration and NodeConfiguration. Now one need to configure their logging in a Microsoft.Extensions.Logging. So in Orleans test projects, should we add a base test class with default logging config so that we don't need to configure logging repeatly with different tests? taking that every tests' logging config should be almost the same

xiazen:
close this since the issue are fixed

PR #3486, opened 2017-10-02
(WIP) Bring TraceToConsole back to TestClusterOptions 

xiazen:
Brought `TraceToConsole `back to `TestClusterOptions `as a property,  instead as an FallBackOptions(which is the logic before). 

TraceToConsole is set to false by default. if set to true, then TestCluster will configure ConsoleLoggerProvider to both client and silo's DI container. It is a shortcut to configure console logging, compare to configure logging through ISiloBuilderFactory or IClientBuilder. But the two methods use the same infrastructure, which is through DI. 

Effort to address #3429


xiazen:
do not merge, I want to see if this design is good or not. But if people think the design is good. Then

attilah:

xiazen:
error CS1061: 'ILoggingBuilder' does not contain a definition for 'AddConsole' and no extension method 'AddConsole' accepting a first argument of type 'ILoggingBuilder' could be found (are you missing a using directive or an assembly reference?) [D:\j\workspace\functional_pr---83d31c72\src\OrleansTestingHost\OrleansTestingHost.csproj]
--


Functional failed on this. But why BVT doesn't fail on this? I think Jenkins env need to refresh or something, since apparently it's holding to old data

attilah:
@dotnet-bot test bvt please

My guess that it will fail now.

jdom:
Let's postpone the decision of whether to include this functionality for after beta, unless you think this must go in

attilah:
As I expected it fails, needs to be rebased on master and fix the compilation error.

the error happens because a missing package reference in OrleansTestingHost project:
```xml
<PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsLoggingVersion)" />
```

xiazen:
but this branch doesn't have that reference. why ci need to load that package? Why do I have to rebase with master for build to build? I don't get it.

xiazen:
ping. Are you OK with the approach? If yes I will rebase.

jdom:
Let's put it on the back-burner for the short term. We'll have to revamp the setup of the test cluster very soon, so not sure how it would affect this

xiazen:
sounds good

PR #3490, opened 2017-10-03
Migrate Membership to use Options for their configuration

xiazen:
- migrate IMembershipTable to use Options as configuration object. 
- MembershipTableFactory now use DI to create IMembershiptable, instead of reflection
- split IGatewayListProvider implementation from IMembershupTable implementation if these two share the same class. Have to do this for correct initialization of GatewayListProvider in the client side
-migrate tests to use the new way of configure membership tables. Some tests still use the legacy ways just to test out the legacy config still works


src/OrleansAWSUtils/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+            services.AddSingleton<IMembershipTable, DynamoDBMembershipTable>();
+            return services;
+        }
+
+        /// <summary>
+        /// Configure DI container with DynamoDBMemebershipTable and get its configuration from <see cref="GlobalConfiguration"/>
+        /// </summary>
+        /// <param name="services"></param>
+        /// <returns></returns>
+        public static IServiceCollection UseDynamoDBMemebershipTableFromLegacyConfigurationSupport(
```
jdom: @ReubenBond I assume this is a candidate to use the ApplicationPartManager instead of the user being explicit with both configuring the XML and then have to call this method?

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Orleans;
+using Orleans.Runtime.Configuration;
+using Orleans.Runtime.MembershipService;
+using OrleansAWSUtils.Configuration;
+
+namespace OrleansAWSUtils
```
jdom: I'd use Microsoft.Orleans.Hosting as the namespace for all these extension methods. Thoughts? /cc @ReubenBond @attilah

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    public static class IServiceCollectionExtensions
```
jdom: whatever namespace we decide, all extensions will probably be in the same namespace, so make the name of this class be more specific, such as DynamoDbServiceCollectionExtensions

src/OrleansAzureUtils/ISiloHostBuilderExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public static class ISiloHostBuilderExtensions
+    {
+
+        /// <summary>
+        /// Configure DI to use AzureBasedMemebershipTable
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="configureOptions"></param>
+        public static ISiloHostBuilder UseAzureMemebershipTable(this ISiloHostBuilder builder,
```
jdom: to be honest, I would drop the suffix Table from Membership. The end user doesn't care if this is implemented via a membership table, or via a different approach, such as in ServiceFabric.
And in particular for this implementation, I would be explicit that this membership uses AzureTable. So the name would end up being `UseAzureTableMembership()`. Do not confuse `Table` as being moved around, it's just that this implementation of membership uses Azure Table, and not Azure Blob nor CosmosDb, which the latter we'll likely implement at some point, so just Azure as the name would be ambiguous; AzureTable is not

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
xiazen: ? mind explain a bit more on the context>? I'm not quite follow

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
xiazen: It make sense since it is extension method for Microsoft.Orleans.Hostting.

src/OrleansAzureUtils/ISiloHostBuilderExtensions.cs
xiazen: sounds good to me

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
attilah: @jdom yes, putting them where they make more sense:
- Either where they're adding functionality: Hosting
- Or where extended interface/class besides: ```IServiceCollection```

src/OrleansAWSUtils/Options/DynamoDBMembershipTableOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Text;
+
+namespace OrleansAWSUtils.Configuration
+{
+    public class DynamoDBMembershipTableOptions
+    {
+        /// <summary>
+        /// Deployment Id.
+        /// </summary>
+        public string DeploymentId { get; set; }
```
attilah: DeploymentId is a Global option, I'd not copy those around into different options, but give accept it from DI. For now it must be GlobalConfiguration, but having GlobalConfiguration as a dependency is better then replicating the same property into different Options classes.

What about extracting DeploymentId into some foreseeable class, like DeploymentOptions and pass it in as a dependency? Then later we can couple the other deployment related config properties there.

I guess based on @jdom's comment about "Table" we'll loose Table everywhere, so it will become DynamoDBMembershipOptions?

src/OrleansAWSUtils/Options/DynamoDBMembershipTableOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+        /// <summary>
+        /// Connection string for DynamoDB Storage
+        /// </summary>
+        public string DataConnectionString { get; set; }
```
attilah: naming: ConnectionsString, I see no need for Data, since there is only 1 connection string for this provider, no need to differentiate.

src/OrleansAzureUtils/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+            return services;
+        }
+
+
+        /// <summary>
+        /// Configure DI to use AzureBasedMemebershipTable, and get its configuration from <see cref="GlobalConfiguration"/>.
+        /// </summary>
+        /// <param name="services"></param>
+        /// <returns></returns>
+        public static IServiceCollection UseAzureMembershipTableFromLegacyConfigurationSupport(
```
attilah: @jdom, @xiazen: What is the plan with the legacy configuration extension methods long term? We did not think about those for these extensions IIRC. Moving them to .Legacy packages once we moved legacy config classes too?

src/OrleansAzureUtils/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                options.DataConnectionString = configuration.DataConnectionString;
+                options.DeploymentId = configuration.DeploymentId;
+                options.MaxStorageBusyRetries = configuration.MaxStorageBusyRetries;
+            });
+            services.AddSingleton<IMembershipTable, AzureBasedMembershipTable>();
+            return services;
+        }
+
+        private static void Configure<TConfiguration, TOptions>(this IServiceCollection services,
```
attilah: I think we should make this method in Orleans.dll public or internal and use that from every extension, no need to copy, maybe a name change there to TransformLegacyConfigurationToOptions or something like that would be better?

src/OrleansConsulUtils/ConsulBasedMembershipTable.cs
```diff
@@ -6,23 +6,21 @@
 using Orleans.Messaging;
```
attilah: Please split to 1 class per file, so when separating the membership, gateway packages, no need to modify files, just move them between projects.

src/OrleansConsulUtils/ConsulBasedMembershipTable.cs
```diff
@@ -33,46 +31,70 @@ public Boolean IsUpdatable
...
         /// </summary>
-        /// <param name="config">The configuration for this instance.</param>
         /// <param name="tryInitTableVersion">Will be ignored: Consul does not support the extended Membership Protocol TableVersion</param>
         /// <returns></returns>
         /// <remarks>
         /// Consul Membership Provider does not support the extended Membership Protocol,
         /// therefore there is no MembershipTable to Initialise
         /// </remarks>
-        public Task InitializeMembershipTable(GlobalConfiguration config, Boolean tryInitTableVersion)
+        public Task InitializeMembershipTable(Boolean tryInitTableVersion)
```
attilah: Boolean vs. bool

src/OrleansConsulUtils/OrleansConsulUtils.csproj
```diff
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Orleans.Core.Abstractions\Orleans.Core.Abstractions.csproj" />
```
attilah: Please leverage transitive P2P references, need to reference only OrleansRuntime.

src/OrleansRuntime/MembershipService/MembershipTableFactory.cs
```diff
@@ -28,44 +28,31 @@ public MembershipTableFactory(IServiceProvider serviceProvider, LoggerWrapper<Me
...
-                switch (globalConfig.LivenessType)
+                
+                // get membership through DI
+                var result = this.serviceProvider.GetService<IMembershipTable>();
+                //if empty, then try to check if user configured using GranBasedMembership
+                if (result == null)
                 {
-                    case LivenessProviderType.MembershipTableGrain:
+                    //if configured through UseGrainBasedMembershipTable method on ISiloHostBuilder, then this won't be null
+                    var options = this.serviceProvider.GetService<GrainBasedMembershipTableOptions>();
```
attilah: MembershipTableGrain should be a treated like the others and instead of using dummy GrainBasedMembershipTableOptions class for check, we can register that by default during the setup phase of the silo like MVC is registering default implementations. The check based on the GlobalConfiguration liveness provider type value should not have this logic in MembershipTableFactory, the Legacy Config transform extension method would be a better place if we really need to keep it.
Long term SiloBuilder should be able to check before building if an IMembershipTable is configured for the silo or not.

@jdom, thoughts?

src/OrleansSQLUtils/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        /// <summary>
+        /// Configure DI container with SqlMemebershipTable
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="configuration"></param>
+        /// <returns></returns>
+        public static IServiceCollection UseSqlMembershipTable(this IServiceCollection services,
+            IConfiguration configuration)
```
attilah: Why is it IConfiguration, what am I missing?

src/OrleansRuntime/MembershipService/MembershipTableFactory.cs
jdom: Agree, all of this initialization should happen as part of the `IMembershipTable.InitializeMembershipTable()` of the concrete implementation. It does require several changes, so I'd be OK if we postpone that refactoring in a separate PR.
What is confusing is how come you get the `GrainBasedMembershipTableOptions` directly from the container. What's in the container is typically the `IOptions<GrainBasedMembershipTableOptions>`

---
In reply to: [142585221](https://github.com/dotnet/orleans/pull/3490#discussion_r142585221) [](ancestors = 142585221)

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
jdom: Hehe, I was summoning @ReubenBond , as this is related to stuff not yet checked in, but will be part of the type discovery part (instead of the full assembly load and scanning we are currently doing).
What I mean is that once that's in, we could have the method in the host builder that when you pass in the legacy configuration object, then it uses the same discovery process to find these types that convert the legacy configuration into strongly typed options, even if they reside on different provider specific assemblies. Not to worry about it for this PR, only after Reuben introduces that functionality.

src/OrleansAzureUtils/IServiceCollectionExtensions.cs
jdom: See https://github.com/dotnet/orleans/pull/3490#discussion_r142747242

src/OrleansAWSUtils/Options/DynamoDBMembershipTableOptions.cs
jdom: I think this configuration should be in the Options for this provider, nevertheless I would add a default IPostConfigure to read it from the place were is globally defined, if it wasn't defined by the end user.
BTW, in this new configuration, we should drop the notion of DeploymentId. Probably call it ClusterId. What do you think @sergeybykov ?

src/OrleansAzureUtils/IServiceCollectionExtensions.cs
xiazen: I think we can make it public. It is a simple function to support. 
I don't think make it internal make sense. Since other csproj which uses it will need to set InternalVisiable to, which is what we want to avoid.

src/OrleansConsulUtils/ConsulBasedMembershipTable.cs
xiazen: Boolean come from the public IMembershipTable interface . Are you suggesting we change the pubic IMembershipTable interface?

src/OrleansRuntime/MembershipService/MembershipTableFactory.cs
xiazen: I vote to refactor it in another PR. There's enough to review in this PR. 
And yes I think you are right about the IOptions<GrainBasedMembershipTableOptions>

src/OrleansSQLUtils/IServiceCollectionExtensions.cs
xiazen: IConfiguration is another main pattern Option team support to configure options. So we provide both choices for  user to configure their provider. What is confusing you?

src/OrleansConsulUtils/ConsulBasedMembershipTable.cs
jdom: `bool` is a C# alias for the `System.Boolean` type. They are equivalent, it's just that by convention we use `bool` everywhere. Not sure why we got `Boolean` there in the first place

test/AWSUtils.Tests/MembershipTests/DynamoDBMembershipTableTest.cs
```diff
@@ -36,8 +38,12 @@ protected override IMembershipTable CreateMembershipTable(Logger logger)
...
             if (!AWSTestConstants.IsDynamoDbAvailable)
                 throw new SkipException("Unable to connect to AWS DynamoDB simulator");
-
-            return new DynamoDBMembershipTable(this.loggerFactory);
+            var options = new DynamoDBMembershipTableOptions()
+            {
+                DataConnectionString = this.connectionString,
+                DeploymentId = this.deploymentId
+            };
+            return new DynamoDBMembershipTable(this.loggerFactory, new OptionsWrapper<DynamoDBMembershipTableOptions>(options));
```
jdom: use `Options.Create(options)` instead of the OptionsWrapper

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
xiazen: I thin it should be like AWSUtilsServiceCollectionExtensions. So future extension methods can belong here too .

src/OrleansAWSUtils/Options/DynamoDBMembershipTableOptions.cs
xiazen: why this provider is different than other provide, that we don't pass deploymentId from cluster config?

jdom: it's not different, I believe this comment extrapolates to all providers

---
In reply to: [142803303](https://github.com/dotnet/orleans/pull/3490#discussion_r142803303) [](ancestors = 142803303)

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
jdom: BTW, as we talked about in person, we do want to do something to avoid users having to call this method from each provider implementation explicitly, whether that's using the upcoming AssemblyPartsManager, or by using and approach similar to the one in MembershipTableFactory today (where you call `AssemblyLoader.LoadAndCreateInstance<ILegacyMembershipConfigurator>` or something like that)

src/OrleansAWSUtils/IServiceCollectionExtensions.cs
jdom: it might be that the current assembly for AWS is not correctly split, but we intend to split some of them, so don't take that as a pattern for everything, even if it currently applies for this one.
In theory, I would expect a nuget with the minimum dependencies to use DynamoDb, which should be separate from the stuff related to SQS.
On a separate note, design guidelines for naming say that you should treat acronyms like 1 pascal cased word, so if you do want to go with that name, it would be `AwsUtilsServiceCollectionExtensions`. We probably don't follow this everywhere, but this is likely one of the few public types with that incorrect casing that the user sees anyway.

xiazen: hmm. we can rename this file when we did split them. No need to over-complicate the naming now. I will take AwsUtilsServiceCollectionExtensions

src/OrleansAWSUtils/Options/DynamoDBMembershipTableOptions.cs
xiazen: @jdom, just to confirm, you think DeploymentId should be in the Option for all MembershipTables? What's the reason of that? It seems to me DeploymentId is not a membershipTable specific concept.

jdom: oh, sorry for the confusion, I think what @attilah suggested is correct too, and probably more appropriate than the PostConfigure suggestion, as I can't think of a scenario where you would want to configure a different ClusterId for the membership provider than the actual silo's ClusterId.
In this PR I'd start with injecting GlobalConfiguration into each provider, but eventually we can get it from a different typed option in a separate PR, once we extract that bit of configuration into a ClusterId property (maybe in the existing SiloIdentityOptions or some other)

src/OrleansConsulUtils/OrleansConsulUtils.csproj
xiazen: hmm transitive P2P references somehow doesn't work

src/OrleansAzureUtils/AzureUtilsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.Orleans.Hosting
+{
+    public static class AzureUtilsServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Configure DI container to use AzureTableBasedMemebership
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="configureOptions"></param>
+        public static IServiceCollection UseAzureTableMemebership(this IServiceCollection services,
```
jdom: typo: several occurrences of `Memeber` instead of `Member`

src/OrleansAWSUtils/Options/DynamoDBMembershipTableOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace OrleansAWSUtils.Configuration
+{
+    public class DynamoDBMembershipTableOptions
```
jdom: still occurrences of `MembershipTable` instead of just `Membership`

src/OrleansConsulUtils/ConsulBasedGatewayListProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Consul;
+using Orleans.Messaging;
+using Orleans.Runtime.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Orleans.Runtime.Host
+{
+    public class ConsulBasedGatewayListProvider : IGatewayListProvider
```
xiazen: @jdom which namespace should those GatewaylistProvider live in? and the same question to membershiptables

src/OrleansAWSUtils/AWSUtilsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Orleans;
+using Orleans.Runtime.Configuration;
+using Orleans.Runtime.MembershipService;
+using OrleansAWSUtils.Configuration;
+
+namespace Microsoft.Orleans.Hosting
```
jdom: I think we were using `Orleans.Runtime.Hosting` as the namespace for the extension methods and the Options, don't recall exactly, but I don't think we were using Microsoft.

src/OrleansConsulUtils/ConsulBasedGatewayListProvider.cs
jdom: I think it should be `Orleans.Runtime.Membership`, but don't know for sure. The one defining this and tracking it was @attilah

src/OrleansAWSUtils/AWSUtilsServiceCollectionExtensions.cs
xiazen: I'd use Microsoft.Orleans.Hosting as the namespace for all these extension methods. Thoughts? /cc @ReubenBond @attilah

quote from you ;) wink wink. But  yeah I should know anything start with Microsoft.Orleans doesn't sound right.

jdom: hehe, my mistake then, but I really meant `Orleans.Hosting` (where the `ISiloHostBuilder` is, which I'm now checking, it doesn't include Runtime either) instead of `OrleansAWSUtils` or some non easily  discoverable namespace when configuring the builder

src/OrleansRuntime/MembershipService/MembershipTableFactory.cs
```diff
@@ -22,50 +23,74 @@ public MembershipTableFactory(IServiceProvider serviceProvider, LoggerWrapper<Me
...
+        private async Task<IMembershipTable> GetMembershipTableLegacy(GlobalConfiguration globalConfiguration)
+        {
+            IMembershipTable result = null;
+            switch (globalConfiguration.LivenessType)
+            {
+                case LivenessProviderType.MembershipTableGrain:
+                    result = await this.GetMembershipTableGrain();
+                    break;
+                case LivenessProviderType.SqlServer:
+                    result = AssemblyLoader.LoadAndCreateInstance<IMembershipTable>(Constants.ORLEANS_SQL_UTILS_DLL, this.logger, this.serviceProvider);
```
jdom: Will this approach work? you are not configuring the strongly typed options for the particular providers... that's why I was suggesting having a configurator instead that gets loaded by the fact of using ClusterConfiguration in the builder

xiazen: You are absolutely right.. my bad. But I wonder how those LivenessTests in AzureUtils pass. Those tests are using the legacy way to create membershiptable. Maybe I need to add more tests.

xiazen: Ah. I think Jenkins wouldn't run those tests....

src/OrleansAWSUtils/AWSUtilsSiloHostBuilderExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using Microsoft.Extensions.Configuration;
+using Orleans.Hosting;
+using OrleansAWSUtils.Configuration;
+
+namespace Orleans.Runtime.Hosting
+{
+    public static class AwsUtilsSiloHostBuilderExtensions
```
jdom: I don't see a lot of value in separating the identical methods where 1 targets ISiloHostBuilder and the other IServiceCollection into 2 different types. This just pollutes the Orleans.Hosting namespace unnecessarily. Just merge the 2 types.

src/OrleansAWSUtils/AWSUtilsSiloHostBuilderExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using Microsoft.Extensions.Configuration;
+using Orleans.Hosting;
+using OrleansAWSUtils.Configuration;
+
+namespace Orleans.Runtime.Hosting
```
jdom: BTW, I believe we are sometimes adding to `Orleans.Runtime.Hosting` and sometimes to `Orleans.Hosting`. Let's be consistent. My opinion is that we go with where `ISiloHostBuilder` is, which is `Orleans.Hosting` as of today.

src/OrleansAWSUtils/AWSUtilsSiloHostBuilderExtensions.cs
xiazen: Hmm I disagree. We have that convention in Orleans core too. And it is clearer that way. One is for IServiceCollection and the other is for ISiloHostBuilder.

jason-bragg: As a general pattern, we tend to have extension functions grouped by what is being extended, so a separate class for IServiceCollection extensions and ISiloHostBuilder extensions makes sense.  I don't think it's a big deal either way in this case, but neither do I see any reason to change this.

src/OrleansRuntime/Hosting/SiloHostBuilderExtensions.cs
```diff
@@ -79,5 +82,15 @@ public static ISiloHostBuilder ConfigureLogging(this ISiloHostBuilder builder, A
         {
             return builder.ConfigureServices(collection => collection.AddLogging(loggingBuilder => configureLogging(loggingBuilder)));
         }
+
+        /// <summary>
+        /// Configure silo to use GrainBasedMembership
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <returns></returns>
+        public static ISiloHostBuilder UseGrainBasedMembership(this ISiloHostBuilder builder)
```
jdom: As we talked about, it's currently confusing, but this class is for extension methods that also exist in the generic host builder, and are not Orleans specific. There's another one for Orleans specific configuration

src/OrleansRuntime/MembershipService/MembershipTableFactory.cs
```diff
@@ -22,50 +35,73 @@ public MembershipTableFactory(IServiceProvider serviceProvider, LoggerWrapper<Me
             this.logger = logger;
         }
 
+        private async Task<IMembershipTable> GetMembershipTableLegacy(GlobalConfiguration globalConfiguration)
```
jdom: IMO this should be done a little bit differently. Basically configure with the legacy configuration in DI as soon as you pass in a legacy configuration to load.
By the time we ask for the membership table at runtime there should be no fallback, and this should have already been configured.
I can submit a PR to your branch if it's not clear

xiazen: The challenge there is you don't know the user configure with legacy way or the new way. And the IServiceCollection interface doesn't provide you much flexibility to help there.
I'm about to apply the same to Gatewaylist provider. let's discuss what you mean here

attilah:

xiazen:
please review this ASAP. I need this to be merged to submit a PR about gatewayListProvider migration.

jdom:
Added https://github.com/xiazen/orleans/pull/8 as a suggestion, so it's more clear what I mean by converting the legacy configuration into DI registrations

test/TesterAzureUtils/Streaming/StreamReliabilityTests.cs
```diff
@@ -47,7 +47,6 @@ public override TestCluster CreateTestCluster()
 
             this.numExpectedSilos = 2;
             var options = new TestClusterOptions(initialSilosCount: (short)this.numExpectedSilos);
-            options.ClusterConfiguration.Globals.LivenessType = GlobalConfiguration.LivenessProviderType.AzureTable;
```
jdom: tests are failing due to this. Since you are removing the LivenessType setting from the GlobalConfiguration, then when the ClientConfiguration is created in the TestClusterOptions.BuildClientConfiguration() method,  the client is not configured to use GatewayProviderType.AzureTable setting. You'll have to set this setting yourself (do it after ClusterConfiguration has been fully configured, because order matters):
```
options.ClientConfiguration.GatewayProvider = ClientConfiguration.GatewayProviderType.AzureTable;
```

Please do this on all tests where you moved off the `Globals.LivenessType` to set the membership provider

xiazen: Why is liveness provider type been copied to client config side? What's the reason for TestCluster doing this behavior?

jdom: Convenience. So that if you configure a non-default liveness type on the test cluster, it is automatically converted to it's par on the client side.
I understand we loose that ability once we switch to extension methods, but to be fair this is mostly a convenience for Orleans' internal testing, as most of the times end users will not change the liveness for their in-memory tests...

xiazen: I'm just saying this is pretty confusing, and probably confusing for end users too. I didn't know why the decision was made that way since I wasn't there. So I was just asking for the reason then.

jdom: yes, agree it's confusing. Nevertheless it's pretty low impact, since most of our users just use the `TestCluster` but don't change the membership. The only reason to change the membership is when you are actually testing a custom membership or restarting all silos (and hence the GrainBased one miss-behaves), which is *mostly* just us, and a few rare exceptions.
Anyway, once we move away from the legacy configuration also in `TestCluster,` this *convenience transformation* will likely go away too

src/OrleansSQLUtils/Messaging/SqlMembershipTable.cs
```diff
@@ -1,36 +1,35 @@
...
         private ILogger logger;
         private RelationalOrleansQueries orleansQueries;
-
-        public SqlMembershipTable(IGrainReferenceConverter grainReferenceConverter, ILogger<SqlMembershipTable> logger)
+        private readonly SqlMembershipOptions membershipTableoptions;
+        public SqlMembershipTable(IGrainReferenceConverter grainReferenceConverter, GlobalConfiguration globalConfig, IOptions<SqlMembershipOptions> membershipTableoptions, ILogger<SqlMembershipTable> logger)
         {
             this.grainReferenceConverter = grainReferenceConverter;
             this.logger = logger;
+            this.membershipTableoptions = membershipTableoptions.Value;
```
jdom: typo: lowercase *options* in a few places

src/OrleansZooKeeperUtils/ZooKeeperHostingExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    public static class ZooKeeperHostingExtensions
+    {
+        /// <summary>
+        /// Configure siloHostBuilder with ZooKeeperMembershipTable
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="configureOptions"></param>
+        /// <returns></returns>
+        public static ISiloHostBuilder UseZooKeeperMembershipTable(this ISiloHostBuilder builder,
```
jdom: these methods suffixed with `Table` but the overloads for IServiceCollection are not. I assume the should be the same?

jdom:
@dotnet-bot test functional

jdom:
@dotnet-bot test functional

test/TesterAzureUtils/Persistence/PersistenceGrainTests_AzureTableStore.cs
```diff
@@ -41,7 +41,6 @@ protected override TestCluster CreateTestCluster()
                 Guid serviceId = Guid.NewGuid();
                 var options = new TestClusterOptions(initialSilosCount: 4);
                 options.ClusterConfiguration.Globals.DataConnectionString = TestDefaultConfiguration.DataConnectionString;
-                options.ClusterConfiguration.Globals.LivenessType = GlobalConfiguration.LivenessProviderType.AzureTable;
```
jdom: I see that you dropped this, but I don't see that you are using the SiloBuilder extension method either. Is this being set somewhere else?
I see the same thing in `PersistenceGrainTests_AzureBlobStore`

test/Tester/CollectionFixtures.cs
```diff
@@ -17,5 +23,21 @@ protected override void CheckPreconditionsOrThrow()
...
         }
+
+        public class SiloBuilderFactory : ISiloBuilderFactory
+        {
+            public ISiloHostBuilder CreateSiloBuilder(string siloName, ClusterConfiguration clusterConfiguration)
+            {
+                return new SiloHostBuilder()
+                    .ConfigureSiloName(siloName)
+                    .UseConfiguration(clusterConfiguration)
+                    .UseAzureTableMembership(options =>
```
jdom: hmmm, not sure about defining this at this level. The `BaseAzureTestClusterFixture` is mostly used for tests that require Azure for storage and/or membership, but most often than not they don't require membership. I would define this silo builder factory for the few specific tests that do require changing membership. At the very least, define this factory as a standalone thing outside the `BaseAzureTestClusterFixture` type, and with a clear name of what is doing, so that you can reuse it in the tests that actually do require a different membership

src/OrleansSQLUtils/OrleansSQLUtils.csproj
```diff
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Orleans.Core.Abstractions\Orleans.Core.Abstractions.csproj" />
+    <ProjectReference Include="..\OrleansRuntime\OrleansRuntime.csproj" />
```
jdom: Hopefully once we split the packages, we can have client one that doesn't require a reference to the silo package, right @attilah?

attilah: It should be possible, I'd really like to explode Orleans.Core after abstractions into Orleans.Client and Core, that would be really useful if client would not be only a meta package.

PR #3492, opened 2017-10-04
Migrate StatisticsConfiguration to StatisticsOptions

attilah:


src/Orleans/Configuration/Options/StatisticsOptions.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using System;
+
+namespace Orleans.Configuration
+{
+    public class StatisticsOptions
+    {
+        /// <summary>
+        /// The MetricsTableWriteInterval attribute specifies the frequency of updating the metrics in Azure table.
```
ReubenBond: I realize these are copied from existing code, but they should probably all be rewritten. For one, none of these are attributes in the C# sense.

src/Orleans/Configuration/Options/StatisticsOptions.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using System;
+
+namespace Orleans.Configuration
+{
+    public class StatisticsOptions
```
xiazen: I think there should be an extension method on ISiloHostBuilder to configure Statistics, for better user experience. 
Based on this PR, the two way to configure statictics. One is using `AddLegacyNodeConfigurationSupport`, which is the legacy way. Second is configure a `StatisticOptions `into DI directly, which is not very straight for user .

src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
```diff
@@ -30,10 +28,18 @@ public static IServiceCollection AddLegacyClusterConfigurationSupport(this IServ
             return services;
         }
 
-        public static IServiceCollection AddLegacyClientConfigurationSupport(this IServiceCollection services)
+        public static IServiceCollection AddLegacyNodeConfigurationSupport(this IServiceCollection services)
```
jdom: BTW, these legacy transformation methods should be called automatically when loading the legacy configuration into SiloBuilder and ClientBuilder.

src/Orleans/Configuration/Options/StatisticsOptions.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using System;
+
+namespace Orleans.Configuration
```
jdom: Should be `Orleans.Hosting` I think

src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
jdom: nevermind, I now saw that at least in DefaultSiloServices you are calling this one. I assume it's the same for others, so please ignore

test/TesterInternal/StatsTests.cs
```diff
@@ -57,7 +59,8 @@ public StatsInitTests(ITestOutputHelper output, Fixture fixture)
             var clientStatisticsManager = this.HostedCluster.ServiceProvider.GetService<ClientStatisticsManager>();
             Assert.NotNull(clientStatisticsManager); // Client Statistics Manager is setup
 
-            Assert.Equal("MockStats",  config.StatisticsProviderName);  // "Client.StatisticsProviderName"
+            var statisticsOptions = this.HostedCluster.ServiceProvider.GetService<IOptions<StatisticsOptions>>();
```
jdom: nit: whenever you get IOptions from the container, call .Value before assigning it.
No big deal here, but for consistency... we should almost never explicitly assign a variable to IOptions, it's only a tool for resolving via DI.

src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
```diff
@@ -30,10 +28,18 @@ public static IServiceCollection AddLegacyClusterConfigurationSupport(this IServ
...
-            // Global configuration
+            services.Configure<NodeConfiguration, StatisticsOptions>((configuration, options) =>
+            {
+                CopyStatisticsOptions(configuration, options);
+            });
 
+            return services;
+        }
+
+        public static IServiceCollection AddLegacyClientConfigurationSupport(this IServiceCollection services)
```
jdom: the namespace for this class should be moved to `Orleans.Hosting`

src/Orleans/Configuration/Options/StatisticsOptions.cs
attilah: Good catch thanks, I also remove the Statistics prefix from the prop names.

src/Orleans/Configuration/Options/StatisticsOptions.cs
attilah: If we decide to put all options to hosting, sure, move them there!

src/Orleans/Configuration/Options/StatisticsOptions.cs
jdom: This is just my opinion, and I don't feel that strongly about it, but I don't believe we should add an extension method for each typed Options. It would make it more discoverable, for sure, but still I don't know, as long as the Options types are in the `Orleans.Hosting` namespace, it might be enough.
We can always add it later for convenience.

src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
jason-bragg: As we don't really want legacy to be the primary usage, I'm wondering if we want to put legacy hosting options into it's own namespace, so users need to explicitly opt into it to use it.  This would make the new and expected patterns more clear and not confuse users of the new experience with legacy support.
Maybe Orleans.Hosting.Legacy?
We can also make this decision at a later time, once the new hosting patterns have baked a bit more.

src/Orleans/Configuration/Options/StatisticsOptions.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+using System;
+
+namespace Orleans.Hosting
```
jason-bragg: By having component configuration poco classes in the hosting namespace, as opposed to the namespace of the component being configured, we are unnecessarily associating all components in the system with how these components are hosted.  This is architecturally backwards.  Technically this works, but IMO is a bad practice.  For discoverability it should be sufficient that the extension functions be in the hosting namespace.  To configure a specific component, say an AzureTableStorageProvider, it is not unreasonable, in fact it's more clear and explicit, to include the namespace of the feature being configured.

attilah: @jdom ok, what to do then I thought we agree on it during standup...

jdom: I'm not convinced yet with either approach. If we were to have an OptionsBuilder like suggested in https://github.com/aspnet/Options/issues/237, then perhaps the decision would be easier and similar to what @jason-bragg is saying, because we could have extension methods for configuring each specific option we would like to expose (and most of the times you would add the provider instance at the same time you configure).
If we don't have that, then the decision is more difficult, as we want the `.Configure<TOptions>(...)` be easily discoverable.
IMO, as of today without OptionsBuilder or similar functionality, I would prefer that everything needed to configure the silo or client be in the default Hosting namespace.

src/Orleans/Configuration/Legacy/LegacyConfigurationExtensions.cs
attilah: I suggest that for Beta have everything in one place, after the FULL options revamp we should move types around (Orleans.Core, Orleans.Runtime), namespaces, etc.
I don't think we've Legacy Hosting ;-) so that's maybe not the best. My plan is that after the full Options work we'll be able to move all legacy hosting options along with the types out of Orleans assemblies to some Legacy ones.

jdom:
This needs to be rebased now

PR #3496, opened 2017-10-04
[Abstractions] Move SiloAddress to Abstractions

benjaminpetit:
I had to duplicate some codes to break some dependencies, but I think they are ok


test/TestGrains/SiloAddressUtils.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Linq;
+using System.Net;
+using System.Net.NetworkInformation;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace UnitTests.Grains
+{
+    public static class SiloAddressUtils
```
jdom: TestExtensions is probably a better place for this, isn't it?
Also, do you really need to use a real IP address with all the complex logic it brings? can't you just use the loopback or another hardcoded address for the tests?

benjaminpetit: The issue is that TestExtensions depends on OrleansTestingHost, not sure we want to add that dependencies on test grain interfaces project.

I already simplified a bit the logic, but you are probably right I could just return the loopback for tests

jdom: ah, I didn't realize this was being used by grain themselves. Well, if not needing a real IP Address is valid, then you won't need this utility at all

---
In reply to: [142800798](https://github.com/dotnet/orleans/pull/3496#discussion_r142800798) [](ancestors = 142800798)

src/Orleans.Core.Abstractions/IDs/SiloAddress.cs
```diff
@@ -177,7 +167,25 @@ public int GetConsistentHashCode()
 
             // Note that Port cannot be used because Port==0 matches any non-zero Port value for .Equals
             string siloAddressInfoToHash = Endpoint + Generation.ToString(CultureInfo.InvariantCulture);
-            hashCode = Utils.CalculateIdHash(siloAddressInfoToHash);
+            // TODO: the following is the equivalent to:
+            // hashCode = Utils.CalculateIdHash(siloAddressInfoToHash);
```
jdom: don't just inline the code. At least create a separate method for it with the same name as the original, so that the method itself is a duplicate of what's in Utils.CalculateIdHash.
Otherwise it's very hard to keep them in sync

test/TestGrains/SiloAddressUtils.cs
jdom: ah, I didn't realize this was being used by grain themselves. Well, if not needing a real IP Address is valid, then you won't need this utility at all

PR #3497, opened 2017-10-04
[Abstractions] Remove static usage of SerializationManager in generated code

benjaminpetit:
Since we don't want to move `SerializationManager` in abstractions, I replaced the call to static methods from this class in the generated code by directly  using method in the serialization contexts.

It should be easier for you to review this PR commit by commit.


src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -25,41 +25,41 @@
...
-        /// </summary>
-        /// <param name="original">Original object to be deep copied.</param>
-        /// <param name="context">The serialization context.</param>
-        /// <returns>Deep copy of the original object.</returns>
-        public delegate object DeepCopier(object original, ICopyContext context);
+    /// <typeparam name="TDeclaring">The declaring type of the field.</typeparam>
+    /// <typeparam name="TField">The field type.</typeparam>
+    /// <param name="instance">The instance having its field set.</param>
+    /// <param name="value">The value being set.</param>
+    public delegate void ValueTypeSetter<TDeclaring, in TField>(ref TDeclaring instance, TField value);
```
ReubenBond: Move to FieldUtils?

src/Orleans/Serialization/SerializationContext.cs
```diff
@@ -52,6 +55,16 @@ public interface ISerializationContext : ISerializerContext
         int CheckObjectWhileSerializing(object raw);
 
         int CurrentOffset { get; }
+
+        void SerializeInner(object obj, Type expected);
+    }
+
+    public static class ISerializationContextExtensions
```
ReubenBond: I get why, but maybe this should be just `SerializationContextExtensions` to avoid IDE squiggles.

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -25,41 +25,41 @@
...
-        /// <param name="instance">The instance having its field set.</param>
-        /// <param name="value">The value being set.</param>
-        public delegate void ValueTypeSetter<TDeclaring, in TField>(ref TDeclaring instance, TField value);
+    /// <summary>
+    /// Deep copier function.
+    /// </summary>
+    /// <param name="original">Original object to be deep copied.</param>
+    /// <param name="context">The serialization context.</param>
+    /// <returns>Deep copy of the original object.</returns>
+    public delegate object DeepCopier(object original, ICopyContext context);
```
ReubenBond: If we can move these to Abstractions, that would be useful, too

src/Orleans/Serialization/FieldUtils.cs
```diff
@@ -0,0 +1,110 @@
+using System;
+using System.Reflection;
+using System.Reflection.Emit;
+
+namespace Orleans.Serialization
+{
+    public sealed class FieldUtils
```
jdom: This class will eventually move to Abstractions? If that's the case, and I'm not suggesting this for the beta but for longer term, can we avoid having it in Abstractions and instead have the codegen add this method to the generated code if it's required?

src/Orleans/Serialization/SerializationManager.cs
benjaminpetit: You mean inside the class `FieldUtils`? Should I move only this one or all the delegate?

src/Orleans/Serialization/SerializationManager.cs
benjaminpetit: Yes they will move with the ISerializerContext interfaces

src/Orleans/Serialization/FieldUtils.cs
benjaminpetit: We can keep that in mind

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: Just this one, since it's specific to FieldUtils

src/Orleans/Serialization/FieldUtils.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        /// <summary>
+        /// The delegate used to set fields in value types.
+        /// </summary>
+        /// <typeparam name="TDeclaring">The declaring type of the field.</typeparam>
+        /// <typeparam name="TField">The field type.</typeparam>
+        /// <param name="instance">The instance having its field set.</param>
+        /// <param name="value">The value being set.</param>
+        public delegate void ValueTypeSetter<TDeclaring, in TField>(ref TDeclaring instance, TField value);
```
jdom: I believe @ReubenBond meant to move it to this code file, not necessarily as a nested type. I'm fine merging it regardless, but I also feel it makes more sense to move it outside the class, inside the same codefile.

ReubenBond: This is fine by me

PR #3499, opened 2017-10-04
Enable EHPurgeLogicTests

xiazen:
It was disabled under non-legacy build condition.  So re-enable it

### DIFF
--- a/test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
+++ b/test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
@@ -223,8 +223,10 @@ namespace ServiceBus.Tests.EvictionStrategyTests
             }
             return itemCount;
         }
-        private Task AddDataIntoCache(EventHubQueueCacheForTesting cache, int count)
+        private async Task AddDataIntoCache(EventHubQueueCacheForTesting cache, int count)
         {
+            await Task.Delay(10);
+            int sequenceNumber = 0;
             while (count > 0)
             {
                 count--;
@@ -235,12 +237,11 @@ namespace ServiceBus.Tests.EvictionStrategyTests
                 var offSet = Guid.NewGuid().ToString() + now.ToString();
                 eventData.SetOffset(offSet);
                 //set sequence number
-                eventData.SetSequenceNumber(count);
+                eventData.SetSequenceNumber(sequenceNumber++);
                 //set enqueue time
                 eventData.SetEnqueuedTimeUtc(now);
                 cache.Add(eventData, DateTime.UtcNow);
             }
-            return Task.CompletedTask;
         }
 
         private NodeConfiguration GetNodeConfiguration()
--- a/test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
+++ b/test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
@@ -59,8 +59,6 @@ namespace ServiceBus.Tests.EvictionStrategyTests
             this.telemetryProducer = new NullTelemetryProducer();
         }
 
-        //Disable tests if in netstandard, because Eventhub framework doesn't provide proper hooks for tests to generate proper EventData in netstandard
-#if BUILD_FLAVOR_LEGACY
         [Fact, TestCategory("BVT")]
         public async Task EventhubQueueCache_WontPurge_WhenUnderPressure()
         {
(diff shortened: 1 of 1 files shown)


test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
```diff
@@ -233,7 +230,15 @@ private Task AddDataIntoCache(EventHubQueueCacheForTesting cache, int count)
                 count--;
```
jason-bragg: HE sequence numbers are ascending.  Why are they descending in this test?

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
```diff
@@ -233,7 +230,15 @@ private Task AddDataIntoCache(EventHubQueueCacheForTesting cache, int count)
...
                 //just to make compiler happy
                 byte[] ignore = { 12, 23 };
-                cache.Add(new EventData(ignore), DateTime.UtcNow);
+                var eventData = new EventData(ignore);
+                DateTime now = DateTime.UtcNow;
```
jason-bragg: 'now' is not guaranteed to be unique, does it need to be?

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
```diff
@@ -233,7 +230,15 @@ private Task AddDataIntoCache(EventHubQueueCacheForTesting cache, int count)
...
+                var offSet = Guid.NewGuid().ToString() + now.ToString();
+                eventData.SetOffset(offSet);
+                //set sequence number
+                eventData.SetSequenceNumber(count);
+                //set enqueue time
+                eventData.SetEnqueuedTimeUtc(now);
+                cache.Add(eventData, DateTime.UtcNow);
             }
             return Task.CompletedTask;
```
jason-bragg: Why is this asynchronous?  Doesn't seem to need to be.

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
xiazen: no. I added those to avoid NullReferenceException when runtime try to serialize those missing fields, EnqueueTime and such. 
In full dotnet, those fields would be filled out by default by just calling new EventData(ignore). But in netstandard, it is not the case anymore. 

So I need to manually assign the values. But how much is the value doesn't affect tests

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
xiazen: the tests doesn't care about sequenceToken and such. I added those to fill in the fields so that runtime won't throw NullReferenceException when try to serialize eventData. Those fields are automatically filled by calling `new EventData()`in in full dotnet, but not anymore in netstandard. Hence those new lines

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
xiazen: I think I wanted to mimic 'concurrently adding items into caches' situation in the tests. In tests, items would be added to a list of cache in parallel. And the caches are sharing the same block pool . So I wanted to test at least in a minimum level that block pool handles 'concurrently adding items to cache' situation well.

jason-bragg: Understood, but c# is a bit to smart sometimes.  Because their is no real async work being done here, the logic will be run sequentially, and no parallelism will be generated.  Suggest adding a Task.Delay(1) to the start of this call, which will allow all tasks to be spun up prior to execution of core logic.

xiazen: That's good to know ! I didn't know that.  I will added it.

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
jason-bragg: Fair enough.  Since our purge logic was time based, I suspected the timestamps may have specific requirements.  NM :)

test/ServiceBus.Tests/EvictionStrategyTests/EHPurgeLogicTests.cs
jason-bragg: Sequence numbers do effect cache behavior, so having them in reverse order in tests seems dangerous.  Basically we're generating test data that violates the expectations of the cache.  The fact that it works in this case is, imo, not a good reason to test against invalid data.

xiazen: I will fix this

xiazen:
This mergeable . test failure are not related.

jdom:
@dotnet-bot test functional 
Reason: unrelated failure

PR #3503, opened 2017-10-05
[Abstractions] Move ISerializerContext & Co. in Abstraction

benjaminpetit:
Generated code uses the interface for serializations/deserialization context, so we need to move them.

The issue is that `ISerializerContext` directly references `SerializationManager`, and we don't want to pull it into Abstraction.

I removed the `SerializationManager` from `ISerializerContext` and instead, when needed, we use the `ServiceProvider` to get it.

To avoid performance issue, I created a nested custom `IServiceProvider` implementation that will directly return the `SerializationManager` if asked, and fallback to the traditional implementation of the `IServiceProvider` if needed.

This pattern will also be used to break the dependency of `GrainReference` in `IClientRuntime`.


src/Orleans/Serialization/SerializationContextBase.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+
+namespace Orleans.Serialization
+{
+    public static class SerializerContextExtensions
+    {
+        public static SerializationManager GetSerializationManager(this ISerializerContext context)
```
ReubenBond: Could we instead do the trick we do with GrainFactory and have the classes implement some `IInternalSerializerContext`? Perf would suffer less

ReubenBond:
Essentially LGTM. I'm a little concerned about perf with all the extra virtual method dispatches for the reader/writer methods. Also not 100% how this will play with some of the implementation/optimizations in `ILBasedSerializer` (eg [this expects a concrete `BinaryTokenStreamWriter`](https://github.com/dotnet/orleans/blob/db5acaab74026adff99403c8458607b2c8b4a2d8/src/Orleans/Serialization/ILSerializerGenerator.cs#L361)), but we'll see what tests say and fix any issues.

benjaminpetit:
@dotnet-bot test functional

PR #3505, opened 2017-10-06
Re-enable intellisense for generated code

ReubenBond:
When `CodeGeneratorEnabled` is `true`, we also include `OutputFile` as a `<Compile>` item, so this just makes sure that it's always included as long as it exists.

Result is that users can navigate to generated code in Visual Studio and presumably other IDEs.


attilah:
What do you think about adding the generated file to the FileWrites itemgroup also?

jdom:
Tried it an it works for me locally

ReubenBond:
@attilah maybe, but I'm not sure if it's needed. It shouldn't hurt

PR #3508, opened 2017-10-06
[Abstractions] Move Logger to Abstractions

benjaminpetit:
`Logger` is used by somne public types like `Grain`


attilah:
We'll remove/move this class from Abstractions once we have updated all Logger calls, right? It is just to not block the abstractions itself?

benjaminpetit:
`Grain` has a `Logger` member, so removing it would be a breaking change. And it is also used by some other components that need to move in Abstraction, so I am not sure of that

xiazen:
There's discussion about moving that Grain Logger method to a legacy package as an extension method on Grain. But that is ongoing discussion. Neitherless, this need to be moved to Abstraction temporarily or permanently.

PR #3517, opened 2017-10-09
Added constructors to set GrainIdentity and GrainRuntime

akosradler:
Added constructors to set GrainIdentity and GrainRuntime on LogConsistentGrain and JournaledGrain for allowing JournaledGrain implementations to be unit tested.


jdom:
Thanks @akosradler, we were swamped with PRs for the upcoming beta so we didn't get back to this one. Looks good, but would need to be rebased, as the `Orleans` folder was renamed.
Do you mind doing that so we can merge it? We are targeting a candidate build today or early tomorrow, so if you'd like this in 2.0.0-beta1, please do it asap, otherwise it'll going in for the final release.

akosradler:
Sorry @jdom that I didn't get back to you earlier. I'll to the rebase. It's not an issue if it doesn't go into the beta, for us this is more like a nice-to-have anyway :)

sergeybykov:
Thank you, @akosradler!

PR #3518, opened 2017-10-09
Generate type metadata during codegen

ReubenBond:
This PR implements a feature similar to ASP.NET's ApplicationParts.

* Users are required to provide a set of assemblies (application parts) when configuring the silo or the client.
* Extension methods are provided for adding all assemblies in the current directory or adding all assemblies in the current `AppDomain`.
* The process of calling `AssemblyLoader` process has been removed from the startup path.
* Instead of scanning every assembly and every type which is loaded for grain interfaces/classes, and serialization information, that occurs at code generation time and the results are loaded on startup.
* Runtime Code Generation is removed. It's *possible* that we could enable runtime code generation which runs before the client/silo is created, but that's out of scope.
* `CachedTypeResolver` performs a more thorough search when resolving types.
* `ITypeResolver` was introduced to allow users to decide how types are resolved and potentially blacklist/whitelist allowed types. `CachedTypeResolver` is the default implementation.

The application parts is like so:
* An `ApplicationPartManager` holds a collection of *Parts* and *FeatureProviders*.
* `ApplicationPartManager` has a method, `void Populate<TFeature>(TFeature feature)`, which can be passed a POCO "feature" class. It gives each FeatureProvider a chance to mutate the feature, given the list of Parts.
* The feature is typically an object containing collections of simple metadata about the application (list of grain interfaces, for example).

Code generation support is like so:
* Code generator defines an assembly level `FeaturePopulatorAttribute` attribute that references generated POCO class which implement an interface `IFeaturePopulator<TFeature>` for each feature (currently: `SerializerFeature`, `GrainInterfaceFeature`, `GrainClassFeature`).
* each `IFeaturePopulator<>` implementation has a `Populate(TFeature)` method which contains code to add metadata to the feature.
* A FeatureProvider called `AssemblyAttributeFeatureProvider<TFeature>` scans each of the provided parts for assemblies which have the assembly-level attribute `FeaturePopulatorAttribute` and then executes the `Populate(feature)` method of each of the references, augmenting the metadata on the feature.

Sample generated code (snippets):
```C#
[assembly: global::Orleans.Metadata.FeaturePopulatorAttribute(typeof (OrleansGeneratedCode.OrleansCodeGenf2f4f9be9eFeaturePopulator))]
namespace OrleansGeneratedCode
{
    internal sealed class OrleansCodeGenf2f4f9be9eFeaturePopulator : global::Orleans.IFeaturePopulator<global::Orleans.GrainInterfaceFeature>, global::Orleans.IFeaturePopulator<global::Orleans.GrainClassFeature>, global::Orleans.IFeaturePopulator<global::Orleans.Serialization.SerializerFeature>
    {
        public void Populate(global::Orleans.GrainInterfaceFeature feature)
        {
            feature.Interfaces.Add(new global::Orleans.GrainInterfaceMetadata(typeof (global::Orleans.IGrainWithStringKey), typeof (global::Orleans.OrleansCodeGenGrainWithStringKeyReference), typeof (global::Orleans.OrleansCodeGenGrainWithStringKeyMethodInvoker), -1277021679));
            feature.Interfaces.Add(new global::Orleans.GrainInterfaceMetadata(typeof (global::Orleans.IRemindable), typeof (global::Orleans.OrleansCodeGenRemindableReference), typeof (global::Orleans.OrleansCodeGenRemindableMethodInvoker), -831689659));
            feature.Interfaces.Add(new global::Orleans.GrainInterfaceMetadata(typeof (global::Orleans.IMembershipTableGrain), typeof (global::Orleans.OrleansCodeGenMembershipTableGrainReference), typeof (global::Orleans.OrleansCodeGenMembershipTableGrainMethodInvoker), -1256503757));
            feature.Interfaces.Add(new global::Orleans.GrainInterfaceMetadata(typeof (global::Orleans.IReminderService), typeof (global::Orleans.OrleansCodeGenReminderServiceReference), typeof (global::Orleans.OrleansCodeGenReminderServiceMethodInvoker), 1450843817));
           // etc
        }

        public void Populate(global::Orleans.Serialization.SerializerFeature feature)
        {
            feature.AddSerializerType(typeof (global::Orleans.Versions.Selector.AllCompatibleVersions), typeof (OrleansGeneratedCodeA059C520.OrleansCodeGenOrleans_Versions_Selector_AllCompatibleVersionsSerializer));
            feature.AddSerializerType(typeof (global::Orleans.Versions.Selector.LatestVersion), typeof (OrleansGeneratedCodeA059C520.OrleansCodeGenOrleans_Versions_Selector_LatestVersionSerializer));
            feature.AddKnownType("Orleans.GrainCallFilterDelegate,Microsoft.Orleans.Core.Abstractions", "Orleans.GrainCallFilterDelegate");
            feature.AddKnownType("Orleans.TransactionOption,Microsoft.Orleans.Core.Abstractions", "Orleans.TransactionOption");
            feature.AddKnownType("Orleans.Versions.Selector.AllCompatibleVersions,Microsoft.Orleans.Core.Abstractions", "Orleans.Versions.Selector.AllCompatibleVersions");
            feature.AddKnownType("Orleans.Versions.Selector.VersionSelectorStrategy,Microsoft.Orleans.Core.Abstractions", "Orleans.Versions.Selector.VersionSelectorStrategy");
            feature.AddKnownType("Orleans.Versions.Selector.LatestVersion,Microsoft.Orleans.Core.Abstractions", "Orleans.Versions.Selector.LatestVersion");
```

This lays the groundwork for implementing 'type forwarding' where a persisted type is forwarded to a different type with the same layout at runtime (eg different assembly, different namespace, different name).

TODO: address `ProviderTypeLoader` which still scans the world


src/Orleans/ApplicationParts/ApplicationPart.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    /// <summary>
+    /// A part of an Orleans application.
+    /// </summary>
+    public abstract class ApplicationPart
+    {
+        /// <summary>
+        /// Gets the <see cref="ApplicationPart"/> name.
+        /// </summary>
+        public abstract string Name { get; }
```
ReubenBond: I'm not sure parts need names. I'd be happy to remove this and make it an interface

src/Orleans/CodeGeneration/TypeUtils.cs
```diff
@@ -1149,10 +1139,21 @@ private static bool DoReferencesContain(IReadOnlyCollection<AssemblyName> refere
 
         private static string EscapeIdentifier(string identifier)
         {
+            if (IsCSharpKeyword(identifier)) return "@" + identifier;
+            return identifier;
+        }
+
+        internal static bool IsCSharpKeyword(string identifier)
```
ReubenBond: This could be moved to another PR if desired. This method is used when escaping C# identifiers. Current we have @ symbols in front of every single identifier in generated code, this PR tries to add them only when needed. It's not perfect: some keywords are only keywords in certain contexts but we don't account for that and just always escape them.

src/OrleansCodeGenerator/GrainInterfaceDescription.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Orleans.CodeGenerator
+{
+    internal class GrainInterfaceDescription
```
ReubenBond: Each "feature" has a corresponding class used during code generation.

test/DefaultCluster.Tests/ApplicationParts/ApplicationPartTests.cs
```diff
@@ -0,0 +1,38 @@
...
+using TestGrainInterfaces;
+using UnitTests.GrainInterfaces;
+using Xunit;
+
+namespace DefaultCluster.Tests.ApplicationParts
+{
+    public class ApplicationPartTests
+    {
+        [Fact]
+        public void T1()
```
ReubenBond: More tests needed as well as appropriate test names

test/NonSilo.Tests/General/InterfaceRules.cs
```diff
@@ -13,7 +13,7 @@ namespace UnitTests.General
 {
     #region simple interfaces
 
-    public interface ITestGrain_VoidMethod : IAddressable
+/*    public interface ITestGrain_VoidMethod : IAddressable
```
ReubenBond: Need to work out how to support these tests given we're performing code generation on this project - probably by moving these types to a non-codegen'd project or similar.

src/Orleans/AssemblyLoader/CachedTypeResolver.cs
```diff
@@ -2,46 +2,38 @@
...
-        public static CachedTypeResolver Instance { get; private set; }
-
-        protected CachedTypeResolver()
-        {
-            cache = new ConcurrentDictionary<string, Type>();
-        }
+        private readonly ConcurrentDictionary<string, Type> typeCache = new ConcurrentDictionary<string, Type>();
+        private readonly CachedReadConcurrentDictionary<string, Assembly> assemblyCache = new CachedReadConcurrentDictionary<string, Assembly>();
+      
+        public static ITypeResolver Instance { get; } = new CachedTypeResolver();
```
jdom: can this eventually go as a singleton in the container instead of static property?

src/Orleans/Core/ClientBuilder.cs
```diff
@@ -115,6 +120,16 @@ private static void AddDefaultServices(HostBuilderContext context, IServiceColle
...
             services.TryAddFromExisting<IClusterClient, IInternalClusterClient>();
+            services.TryAddSingleton<ITypeResolver, CachedTypeResolver>();
+
+            // Application parts
+            var parts = context.GetApplicationPartManager();
+            services.TryAddSingleton<ApplicationPartManager>(parts);
+            parts.AddApplicationPart(typeof(RuntimeVersion).Assembly);
+            parts.AddFeatureProvider(new BuiltInTypesSerializationFeaturePopulator());
+            parts.AddFeatureProvider(new AssemblyAttributeFeatureProvider<GrainInterfaceFeature>());
+            parts.AddFeatureProvider(new AssemblyAttributeFeatureProvider<GrainClassFeature>());
```
jdom: Does the client builder care about concrete grain implementations?

src/Orleans/AssemblyLoader/CachedTypeResolver.cs
ReubenBond: Yes, `ITypeResolver` is usually used as a singleton. This is currently used in 2 places:
* `ApplicationConfiguration.ValidateConfiguration(logger)`
*  `FilterPredicateWrapperData.RehydrateStaticFuncion`

We could new up an instance each time in the 1st case and in the 2nd case we could have a static instance on the class.

I'm not so sure about the 2nd case, and I'm not a fan of how stream predicates are implemented.

src/Orleans/Core/ClientBuilder.cs
ReubenBond: Seems like it doesn't. I wanted everything symmetrical, Maybe using a common "AddDefaultFeatureProviders" method, but I can remove it.

src/Orleans.Core.Abstractions/CodeGeneration/ActivationAttribute.cs
```diff
@@ -72,4 +72,10 @@ public SerializerAttribute(Type targetType)
         {
         }
     }
+
+    /// <summary>Identifies generated code.</summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class OrleansGeneratedCodeAttribute : Attribute
```
jdom: Can't we just leverage [System.CodeDom.Compiler.GeneratedCodeAttribute](https://msdn.microsoft.com/en-us/library/system.codedom.compiler.generatedcodeattribute(v=vs.110).aspx) ?

src/Orleans.Core.Abstractions/GrainFeature.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Orleans
+{
+    public class GrainInterfaceMetadata
```
jdom: hmm, strange that you don't get warnings for the missing XML comments

src/Orleans.Core.Abstractions/GrainFeature.cs
```diff
@@ -0,0 +1,50 @@
...
+            this.ReferenceType = referenceType;
+            this.InvokerType = invokerType;
+            this.InterfaceId = interfaceId;
+        }
+
+        public Type InterfaceType { get; }
+
+        public Type ReferenceType { get; }
+
+        public Type InvokerType { get; }
```
jdom: Maybe I confused GrainClassXXX with Invoker stuff. This InvokerType is not necessarily available on the client, right? But the rest of this information must be available on the client, correct?
If that's the case, shouldn't InvokerType be part of `GrainClassMetadata`?
Or what is this type? Some XML comments on all of these would help a lot

src/Orleans.Core.Abstractions/GrainFeature.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public Type ClassType { get; }
+    }
+
+    public class GrainClassFeature
+    {
+        public IList<GrainClassMetadata> Classes { get; } = new List<GrainClassMetadata>();
+    }
+
+    public interface IFeaturePopulator<in TFeature>
```
jdom: this should go on its own file, right?

src/Orleans.Core.Abstractions/CodeGeneration/ActivationAttribute.cs
ReubenBond: I think we should stop leveraging that, since it's common for all code generators, but I can change it.

src/Orleans.Core.Abstractions/GrainFeature.cs
ReubenBond: Ah yes, I need to fix that. And it is strange... I wonder why? Don't we have WarningAsError enabled, too?

src/Orleans.Core.Abstractions/GrainFeature.cs
ReubenBond: Invoker *is* available on the client, since it's associated with the interface and not the class. We don't generate any code for the class (until now, with metadata)

src/Orleans.Core.Abstractions/GrainFeature.cs
ReubenBond: Yes, thanks. I'll split everything up

src/Orleans/ApplicationParts/ApplicationPart.cs
jdom: I assume it was for easier troubleshooting. I don't care that much either way, but since it's not bad in itself, I think I would prefer to keep it in the same shape as their infrastructure.

src/Orleans/ApplicationParts/ApplicationPartManagerExtensions.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                if (processedAssemblies.Add(assembly) && loadReferencedAssemblies)
+                {
+                    AddReferencedAssemblies(assembly, processedAssemblies);
+                }
+
+                applicationPartManager.AddApplicationPart(assembly);
+            }
+
+            void AddReferencedAssemblies(Assembly asm, HashSet < Assembly > loadedAssemblies)
```
jdom: nit: weird spacing 😛

src/Orleans/ApplicationParts/ApplicationPartManagerExtensions.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+
+        private static readonly string CoreAssemblyName = typeof(RuntimeVersion).Assembly.GetName().Name;
+        private static readonly string AbstractionsAssemblyName = typeof(IGrain).Assembly.GetName().Name;
+        private static readonly IEnumerable<string> NoReferenceComplaint = new[] {$"Assembly does not reference {CoreAssemblyName} or {AbstractionsAssemblyName}"};
+
+        public static void AddApplicationPartsFromDefaultDirectories(this ApplicationPartManager applicationPartManager)
+        {
+            var exeRoot = Path.GetDirectoryName(typeof(ApplicationPartManagerExtensions).GetTypeInfo().Assembly.Location);
```
jdom: Is this correct? exeRoot will not be a folder, but a DLL path, so combining it with `Applications` will not be very nice.
Also, does this work in .NET Core? We've seen weird issues in our preview3 assembly loading that now requires to `publish` the build first, otherwise the assemblies run from weird locations. If we are taking this approach forward, then we should probably fix it (in case it wasn't, I didn't validate if it's the same approach as before).

src/Orleans.Core.Abstractions/CodeGeneration/ActivationAttribute.cs
jdom: But are we actually depending on this attribute or are we adding it for other tools? (like static analyzers, which skip code that has this attribute).
Also the attribute in CodeDom takes the tool name, so we can pass "Orleans" there to differentiate.

src/Orleans/ApplicationParts/AssemblyAttributeFeatureProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            return typeof(AssemblyAttributeFeatureProvider<TFeature>).GetHashCode();
+        }
+
+        public void PopulateFeature(IEnumerable<ApplicationPart> parts, TFeature feature)
+        {
+            foreach (var part in parts.OfType<AssemblyPart>())
+            {
+                var attributes = part.Assembly.GetCustomAttributes(true);
+                foreach (var attribute in attributes.OfType<IFeaturePopulator<TFeature>>())
```
jdom: I don't follow this logic. The assembly attribute itself implements `IFeaturePopulator<T>`?

src/Orleans/ApplicationParts/ApplicationPartManagerExtensions.cs
ReubenBond: I believe it's correct: we're calling `Path.GetDirectoryName` on the result. This is used during tests and it seems to be working there.
As far as I know, publishing a build is the only valid way to deploy it. I'm not sure the best approach.

Should I perhaps add an extra parameter to specify paths?

src/Orleans/ApplicationParts/ApplicationPartManagerExtensions.cs
```diff
@@ -0,0 +1,147 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Extensions.Logging.Abstractions;
+using Orleans.Hosting;
+using Orleans.Runtime;
+
+namespace Orleans.ApplicationParts
```
ReubenBond: Should these be in Orleans.Hosting?

src/Orleans/ApplicationParts/AssemblyAttributeFeatureProvider.cs
ReubenBond: yes, that's correct

ReubenBond: So we have an assembly level attribute like this:
```C#
[assembly: OrleansGeneratedCode.OrleansCodeGenb8790f1e97FeaturePopulator()]
```

jdom: Yeah, I downloaded the code and saw that. It does work, although it's a little odd to have the populators be defined as attributes themselves, instead of just POCOs that can easily be used outside of the assembly populator (granted, it *can* be used, but it's odd that it is an attribute).
Maybe it's more appropriate to have a normal POCO and then have a well known assembly level attributes defined that points to it?

```
[assembly: FeaturePopulator(typeof(OrleansGeneratedCode.OrleansCodeGenb8790f1e97FeaturePopulator))]
```

ReubenBond: How could it be used outside of the assembly populator given it has a randomly generated name? Users can follow the same logic with the `attributes.OfType<IFeaturePopulator<T>>`. I'm not seeing the downside of this approach but I can change it if you prefer. It just means we need to use `Activator.CreateInstance()` to create it

jdom: That actually I found a little odd. What's the reason for not using the AssemblyName as the differentiator instead of a random string?

ReubenBond: Can do that. No one should be referencing generated code directly, though

ReubenBond: Back in the olden days, when per-interface GrainFactory instances were being generated, that was normal, but that was years ago.

ReubenBond: Oh, the reason for not using AssemblyName was that I wasn't sure what characters were valid in AssemblyName (eg whitespace), so I would have to sanitize it into what's available in a type name. Anyway, I'll convert to some well-known attribute with the randomly generated name for now. Can't use a fixed name because `[InternalsVisibleTo]` might make it ambiguous

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -396,10 +303,7 @@ private void Register(Type t, DeepCopier cop, Serializer ser, Deserializer deser
...
                     registeredTypes.Add(t);
                     string name = t.OrleansTypeKeyString();
-                    lock (types)
-                    {
-                        types[name] = t;
-                    }
+                    types[name] = t;
                     if (cop != null)
                     {
                         lock (copiers)
```
jdom: are all of these locks still required? Maybe it makes sense to use the `CachedReadConcurrentDictionary` if you expect some minor concurrency during warmup?

jdom: BTW, fine if we postpone all the locks removals from this class into a separate PR that we can test in isolation. It's probably a lot more appropriate to do it that way

---
In reply to: [143850025](https://github.com/dotnet/orleans/pull/3518#discussion_r143850025) [](ancestors = 143850025)

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -494,234 +395,85 @@ private void Register(Type t)
         /// <param name="serializerType">The type containing serialization methods for <paramref name="type"/>.</param>
         private void Register(Type type, Type serializerType)
         {
-            try
+            GetSerializationMethods(serializerType, out var copier, out var serializer, out var deserializer);
+
+            if (serializer == null && deserializer == null && copier == null)
```
jdom: IS this a change in behavior? In the past, at least in this method we allowed to specify some of the delegates, but not necessarily all. (for example you could provide the copier, but not the serializer/deserializer pair, and the other way around). Then additional calls to this could enhance this and provide the missing delegate (or override just a few of them).
Not sure if that was a feature or just how it was implemented, but it's worth checking whether we are changing *expected* functionality

src/Orleans/Serialization/SerializationManager.cs
```diff
@@ -1984,11 +1733,16 @@ internal Type ResolveTypeName(string typeName)
...
+            
             throw new TypeAccessException("Type string \"" + typeName + "\" cannot be resolved.");
         }
 
 #endregion
+        [AttributeUsage(AttributeTargets.ReturnValue)]
+        public class StreamedAttribute : Attribute
+        {
+        }
+        [return: Streamed]
```
jdom: what is this for?

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: That's still supported. These two guards throw if either:
* All three are null
* Serializer is null or Deserializer is null but not both

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: Let's postpone it until afterwards

src/Orleans/Serialization/SerializationManager.cs
ReubenBond: oops I was just playing with attributes 😱

src/Orleans/CodeGeneration/TypeUtils.cs
attilah: Roslyn has code for this: https://github.com/dotnet/roslyn/blob/master/src/Tools/Source/CompilerGeneratorTools/Source/BoundTreeGenerator/BoundNodeClassWriter.cs#L1600 maybe it's worth just to take and maintain that code as-is?

src/ClientGenerator/CodeGenerator.cs
```diff
@@ -65,19 +67,28 @@ public static bool GenerateCode(CodeGenOptions options)
...
                     });
-                var codeGenerator = new RoslynCodeGenerator(new SerializationManager(null, serializationProviderOptions, null, loggerFactory), loggerFactory);
+                var applicationPartManager = new ApplicationPartManager();
+                applicationPartManager.AddFeatureProvider(new BuiltInTypesSerializationFeaturePopulator());
+                applicationPartManager.AddFeatureProvider(new AssemblyAttributeFeatureProvider<GrainInterfaceFeature>());
+                applicationPartManager.AddFeatureProvider(new AssemblyAttributeFeatureProvider<GrainClassFeature>());
+                applicationPartManager.AddFeatureProvider(new AssemblyAttributeFeatureProvider<SerializerFeature>());
+                applicationPartManager.AddApplicationPart(grainAssembly);
+                applicationPartManager.AddApplicationPart(typeof(RuntimeVersion).Assembly);
+                applicationPartManager.AddApplicationPartsFromAppDomain();
```
attilah: Why add assemblies from the ```AppDomain``` when generating code? During codegen we rely more on the passed in assembly then on what is loaded, not?

src/Orleans.Core.Abstractions/Metadata/FeaturePopulatorAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+    public sealed class FeaturePopulatorAttribute : Attribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FeaturePopulatorAttribute"/> class.
+        /// </summary>
+        /// <param name="populatorType">The feature populator type.</param>
+        public FeaturePopulatorAttribute(Type populatorType)
+        {
+            this.PopulatorType = populatorType;
```
attilah: A check for ```IFeaturePopulator``` would make sense here IMO, just to be foolproof.

src/ClientGenerator/CodeGenerator.cs
ReubenBond: Adds the code generator assemblies, like Orleans.dll. Instead we could use the AddApplicationPartsFromReferences extension.

src/Orleans.Core.Abstractions/Metadata/FeaturePopulatorAttribute.cs
ReubenBond: essentially: `populatorType.GetInterfaces().Any(iface => iface.IsConstructedGenericType && typeof(IFeaturePopulator<>).IsAssignableFrom(iface.GetGenericTypeDefinition()))`;?

src/Orleans/ApplicationParts/AssemblyAttributeFeatureProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <inheritdoc />
+        public void PopulateFeature(IEnumerable<IApplicationPart> parts, TFeature feature)
+        {
+            foreach (var part in parts.OfType<AssemblyPart>())
+            {
+                var attributes = part.Assembly.GetCustomAttributes(true);
+                foreach (var attribute in attributes.OfType<FeaturePopulatorAttribute>())
+                {
+                    if (!typeof(IFeaturePopulator<TFeature>).IsAssignableFrom(attribute.PopulatorType)) continue;
```
attilah: Why not fail? Silent skipping can be misleading for developers who adds their own features, not?

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -64,19 +64,20 @@ internal class AssemblyLoader
...
             List<string> discoveredAssemblyLocations = loader.DiscoverAssemblies();
             foreach (var pathName in discoveredAssemblyLocations)
             {
                 loader.logger.Info("Loading assembly {0}...", pathName);
+
                 // It is okay to use LoadFrom here because we are loading application assemblies deployed to the specific directory.
                 // Such application assemblies should not be deployed somewhere else, e.g. GAC, so this is safe.
-                Assembly.LoadFrom(pathName);
-                ++count;
+                loadedAssemblies.Add(Assembly.LoadFrom(pathName));
```
attilah: We should put it around try/catch to not to fail the whole loading loop if something fails to load.

src/Orleans/AssemblyLoader/AssemblyLoaderCriteria.cs
```diff
@@ -40,10 +40,17 @@ internal static AssemblyLoaderReflectionCriterion LoadTypesAssignableFrom(Type[]
                 AssemblyLoaderReflectionCriterion.NewCriterion(
                     (Type type, out IEnumerable<string> ignored) =>
                     {
+                        var reflectionOnlyType = TypeUtils.ToReflectionOnlyType(type);
+                        if (reflectionOnlyType.Name == "MemoryStorage")
+                        {
+                            Console.WriteLine();
```
attilah: Is this a leftover or intentional?

src/Orleans/AssemblyLoader/CachedReflectionOnlyTypeResolver.cs
```diff
@@ -11,7 +11,7 @@ static CachedReflectionOnlyTypeResolver()
             Instance = new CachedReflectionOnlyTypeResolver();
         }
 
-        public static new CachedReflectionOnlyTypeResolver Instance { get; private set; }
+        public new static CachedReflectionOnlyTypeResolver Instance { get; private set; }
```
attilah: can't we remove the requirement of this static?

src/Orleans/ApplicationParts/AssemblyAttributeFeatureProvider.cs
ReubenBond: Because not all attributes in an assembly are feature populators and the ones that are might not be feature populators for this particular type.

ReubenBond:
The latest two commits won't affect functionals, which passed in the commit prior to that.

src/Orleans/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -64,50 +64,29 @@ internal class AssemblyLoader
...
-                var foundType =
-                    TypeUtils.GetTypes(
-                        assembly,
-                        type => typeof(T).IsAssignableFrom(type) && !type.GetTypeInfo().IsInterface,
-                        logger).FirstOrDefault();
-                if (foundType == null)
+                try
                 {
-                    return null;
+                    loadedAssemblies.Add(Assembly.LoadFrom(pathName));
```
jdom: hmmm, I'm a little scared of still using `LoadFrom` now that we are getting rid of the handling of AssemblyResolve (are we?). Not sure if this can go away. At least I'm glad that you are returning the assembly itself instead of the string name.
But didn't follow the entire approach, it's merely a heads up. I can take a closer look tomorrow at the office

ReubenBond: This should be correct - it's used only when adding assembly parts. We still have the assembly resolve hook during codegen (this occurs after assembly resolution is intercepted)

ReubenBond: Or... do we need to try a direct `Load` instead?

src/OrleansRuntime/Hosting/DefaultSiloServices.cs
```diff
@@ -202,13 +201,27 @@ internal static void AddDefaultServices(HostBuilderContext context, IServiceColl
...
+            services.TryAddSingleton<SerializationManager>();
+            services.TryAddSingleton<ITypeResolver, CachedTypeResolver>();
+            
             // Transactions
             services.TryAddSingleton<ITransactionAgent, TransactionAgent>();
             services.TryAddSingleton<Factory<ITransactionAgent>>(sp => () => sp.GetRequiredService<ITransactionAgent>());
             services.TryAddSingleton<ITransactionManagerService, DisabledTransactionManagerService>();
+
+            // Application Parts
+            var parts = context.GetApplicationPartManager();
```
attilah: variable name should be ```applicationPartManager```

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -179,7 +179,13 @@ internal Silo(SiloInitializationParameters initializationParams, IServiceProvide
                 serviceCollection.AddSingleton<Silo>(this);
                 serviceCollection.AddSingleton(initializationParams);
                 serviceCollection.AddLegacyClusterConfigurationSupport(config);
-                DefaultSiloServices.AddDefaultServices(null, serviceCollection);
+                var hostContext = new HostBuilderContext(new Dictionary<object, object>());
+                DefaultSiloServices.AddDefaultServices(hostContext, serviceCollection);
+
+                var parts = hostContext.GetApplicationPartManager();
+                parts.AddApplicationPartsFromDefaultDirectories();
```
attilah: We discussed that we're not preserving the "old behavior" by processing assemblies from file system and appdomain, so this should be removed, not?

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -179,7 +179,13 @@ internal Silo(SiloInitializationParameters initializationParams, IServiceProvide
                 serviceCollection.AddSingleton<Silo>(this);
                 serviceCollection.AddSingleton(initializationParams);
                 serviceCollection.AddLegacyClusterConfigurationSupport(config);
-                DefaultSiloServices.AddDefaultServices(null, serviceCollection);
+                var hostContext = new HostBuilderContext(new Dictionary<object, object>());
+                DefaultSiloServices.AddDefaultServices(hostContext, serviceCollection);
+
+                var parts = hostContext.GetApplicationPartManager();
```
attilah: variable name: ```applicationPartManager```

src/OrleansTestingHost/AppDomainSiloHost.cs
```diff
@@ -35,8 +32,11 @@ public class AppDomainSiloHost : MarshalByRefObject
         public AppDomainSiloHost(string name, Type siloBuilderFactoryType, ClusterConfiguration config)
         {
             var builderFactory = (ISiloBuilderFactory)Activator.CreateInstance(siloBuilderFactoryType);
-            ISiloHostBuilder builder = builderFactory.CreateSiloBuilder(name, config);
-            builder.ConfigureServices((services) => services.AddSingleton<TestHooksSystemTarget>());
+            ISiloHostBuilder builder = builderFactory
+                .CreateSiloBuilder(name, config)
+                .ConfigureServices(services => services.AddSingleton<TestHooksSystemTarget>())
+                .AddApplicationPartsFromAppDomain()
```
attilah: Same as above, loading "old" defaults.

src/OrleansTestingHost/DefaultSiloBuilderFactory.cs
```diff
@@ -12,7 +13,9 @@ public sealed class DefaultSiloBuilderFactory : ISiloBuilderFactory
         public ISiloHostBuilder CreateSiloBuilder(string siloName, ClusterConfiguration clusterConfiguration)
         {
             var builder = new SiloHostBuilder();
+
             return builder.ConfigureSiloName(siloName)
+                .AddApplicationPartsFromAppDomain()
```
attilah: Same as above, loading "old" defaults.

src/OrleansTestingHost/TestCluster.cs
```diff
@@ -460,6 +461,8 @@ public void InitializeClient()
 
             this.InternalClient = (IInternalClusterClient)new ClientBuilder()
                 .UseConfiguration(clientConfig)
+                .AddApplicationPartsFromDefaultDirectories()
+                .AddApplicationPartsFromAppDomain()
```
attilah: Same as above for client, loading "old" defaults.

src/Orleans/ApplicationParts/AssemblyAttributeFeatureProvider.cs
attilah: We are only iterating on ```FeaturePopulatorAttribute```s, so first part of your reply is is covered,  what other types are we supporting on the ```PopulatorType``` property which ones are not ```IFeaturePopulator<TFeature>```?

src/OrleansTestingHost/TestCluster.cs
```diff
@@ -94,6 +93,8 @@ public class TestCluster
         /// </summary>
         public SerializationManager SerializationManager { get; private set; }
 
+        public HashSet<AssemblyName> AssembliesToLoad { get; } = new HashSet<AssemblyName>();
```
attilah: What about we utilize the new ```ApplicationParts``` infrastructure instead of this construct and populate it with parts via an extension method, instead of adding assemblies to a list what we're loading, then processing from appdomain?

src/OrleansRuntime/Hosting/DefaultSiloServices.cs
ReubenBond: that's a bit verbose, but alright

src/OrleansRuntime/Silo/Silo.cs
ReubenBond: how could it possibly work if we're not doing this here? This code is executed when a user constructs a Silo instance directly (no builder), eg when using the Azure host. This is a legacy support mechanism.

Whether or not we stop supporting that and remove this block is a matter for another PR

src/OrleansTestingHost/TestCluster.cs
ReubenBond: I can concat these together with `;` and set that string on the remote app domain and then explicitly load them and add them as parts but this mechanism works just as well with less code.

jdom:
We should update the changelog ASAP, while we remember exactly all the breaking changes

src/Orleans.Core/ApplicationParts/AssemblyAttributeFeatureProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            return typeof(AssemblyAttributeFeatureProvider<TFeature>).GetHashCode();
+        }
+
+        /// <inheritdoc />
+        public void PopulateFeature(IEnumerable<IApplicationPart> parts, TFeature feature)
+        {
+            foreach (var part in parts.OfType<AssemblyPart>())
+            {
+                var attributes = part.Assembly.GetCustomAttributes(true);
```
jdom: since we are using a well known attribute now, you can get this attribute directly instead of getting all and then trying to cast to `FeaturePopulatorAttribute`. I believe that approach is also safer against assembly load exception if some attribute is in another assembly that can't be loaded

test/DefaultCluster.Tests/CollectionFixtures.cs
```diff
@@ -1,11 +1,12 @@
...
+﻿using DefaultCluster.Tests.General;
+using TestExtensions;
 using Xunit;
 
 namespace DefaultCluster.Tests
 {
     // Assembly collections must be defined once in each assembly
     [CollectionDefinition("DefaultCluster")]
-    public class DefaultClusterTestCollection : ICollectionFixture<DefaultClusterFixture> { }
+    public class DefaultClusterTestCollection : ICollectionFixture<DefaultClusterFixture>, ICollectionFixture<DefaultClusterTestsFixture> { }
```
jdom: hmmm, not sure I follow, these should create both fixtures, doesn't it? essentially starting 2 different clusters?

test/NonSilo.Tests/RuntimeTypeNameFormatterTests.cs
```diff
@@ -42,15 +45,16 @@ public void FormattedTypeNamesAreRecoverable()
                     .MakePointerType()
                     .MakePointerType()
                     .MakeArrayType(10)
-                    .MakeByRefType()
+                    .MakeByRefType(),
+                typeof(NameValueCollection)
```
jdom: this is added twice by design?

jdom:
LGTM (well, not just good, but *awesome*!)
As we expected, this makes our tests **much faster**, because the cluster startup is much faster too :) 
Excited that we are finally doing this thing we started discussing almost a year ago, to avoid a massive type scanning at startup. We can still optimize it more, but it's a great improvement already.
I ran these changes through some of our load test apps as a smoke test, and it works well there too

PR #3522, opened 2017-10-09
Fixed duplicate stream message cache monitoring bug.

jason-bragg:
Problems:
- Monitoring for message caches used timers which were not being cleaned up when caches were no longer in use.  This lead to duplicate monitoring calls on disposed caches.
- Caches not disposed, but not active, continue to periodically log state metrics.

Fix
- Clean up caches correctly on receiver shutdown and reinitialize (due to failure).
- No longer rely on timers to log periodically, instead allow activity to drive periotic metrics logging.  This removes the need to dispose of timers, and eliminates state metrics from inactive caches.

### DIFF
--- a/src/OrleansProviders/Streams/Common/PooledCache/ChronologicalEvictionStrategy.cs
+++ b/src/OrleansProviders/Streams/Common/PooledCache/ChronologicalEvictionStrategy.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using Orleans.Runtime;
-using System.Threading;
 
 namespace Orleans.Providers.Streams.Common
 {
@@ -20,8 +19,8 @@ namespace Orleans.Providers.Streams.Common
         /// </summary>
         protected readonly Queue<FixedSizeBuffer> inUseBuffers;
         private FixedSizeBuffer currentBuffer;
-        private ICacheMonitor cacheMonitor;
-        private Timer timer;
+        private readonly ICacheMonitor cacheMonitor;
+        private readonly PeriodicAction periodicMonitoring;
         private long cacheSizeInByte;
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


src/OrleansProviders/Streams/Common/PooledCache/ChronologicalEvictionStrategy.cs
```diff
@@ -96,9 +99,15 @@ public void OnBlockAllocated(FixedSizeBuffer newBlock)
...
         }
 
-        /// <inheritdoc />
         public void PerformPurge(DateTime nowUtc)
         {
+            PerformPurgeInternal(nowUtc);
+            this.periodicMonitoring?.TryAction(nowUtc);
+        }
+
+        /// <inheritdoc />
```
xiazen: move this line back to PerformPurge

test/TesterInternal/OrleansRuntime/Streams/PooledQueueCacheTests.cs
```diff
@@ -236,15 +236,16 @@ private int RunGoldenPath(PooledQueueCache<TestQueueMessage, TestCachedMessage>
...
-                cache.Add(new TestQueueMessage
+            List<TestQueueMessage> messages = Enumerable.Range(0, MessagesPerBuffer * PooledBufferCount)
+                .Select(i => new TestQueueMessage
                 {
                     StreamGuid = i % 2 == 0 ? stream1.Guid : stream2.Guid,
                     StreamNamespace = TestStreamNamespace,
-                    SequenceNumber = sequenceNumber++,
-                }, DateTime.UtcNow);
-            }
+                    SequenceNumber = sequenceNumber + i
```
xiazen: why it this `SequenceNumber = sequenceNumber + i `, instead of the original logic of  `sequenceNumber++ `? seems like you changed the logic here. And btw this test is failing.

xiazen:
remaining code LGTM as a fix for 1.5.2.

test/TesterInternal/OrleansRuntime/Streams/PooledQueueCacheTests.cs
jason-bragg: I was trying to avoid modifying state in an enumeration, can be hard to maintain.
Introduced cut+paste error, reusing 'messages' rather than the new 'moremessages'

PR #3523, opened 2017-10-10
Service Fabric: Async shutdown support for CloseAsync

elqueffo:
This is my suggestion to fix two things

1) Service Fabric CloseAsync expects us to wait async for shutdown so we don't block. I added async shutdown code paths, using a thread to do the actual shutdown, and waiting for the terminated wait event to fire. I picked thread here because that's what the other, existing event triggered background shutdown seemed to be doing as well.

2) Service Fabric integration should in my opinion call shutdown instead of stop on the silo, so that OnDeactivateAsync for grains gets called when SF moves/scales/messes with the silo service in general.

### DIFF
--- a/src/Orleans.Membership.ServiceFabric/OrleansCommunicationListener.cs
+++ b/src/Orleans.Membership.ServiceFabric/OrleansCommunicationListener.cs
@@ -118,8 +118,7 @@ namespace Microsoft.Orleans.ServiceFabric
         /// <returns>A <see cref="Task"/> representing the work performed.</returns>
         public Task CloseAsync(CancellationToken cancellationToken)
         {
-            this.SiloHost.Stop();
-            return Task.CompletedTask;
+            return this.SiloHost.ShutdownAsync(cancellationToken);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/OrleansRuntime/Silo/SiloHost.cs
```diff
@@ -249,6 +250,39 @@ public void ShutdownOrleansSilo()
...
+        /// <param name="cancellationToken">Token that cancels waiting for shutdown.</param>
+        /// <returns></returns>
+        public Task ShutdownOrleansSiloAsync(int millisecondsTimeout, CancellationToken cancellationToken)
+        {
+            if (orleans == null || !IsStarted)
+                return Task.CompletedTask;
+
+            IsStarted = false;
+
+            var shutdownThread = new Thread(o =>
```
sergeybykov: Why not Task.Run() instead of creating a thread? Not a big deal, just curious.

sergeybykov:
@elqueffo Could you please rebase.

src/OrleansRuntime/Silo/SiloHost.cs
elqueffo: Only to follow the style of the other shutdown code. We can change it if you want.

elqueffo:
@sergeybykov rebased

sergeybykov:
@dotnet-bot test functional

sergeybykov:
Thank you, @elqueffo!

PR #3526, opened 2017-10-10
Implement IKeyedServiceCollection in StreamProviderManager and remove it from GrainRuntime

benjaminpetit:
`IStreamProviderManager` is depending on too many elements, so we don't want to pull it into Abstraction.

The modification proposed here is to use `IKeyedServiceCollection , like it is done for the `StorageProviderManager`.

### DIFF
--- a/src/OrleansRuntime/Silo/Silo.cs
+++ b/src/OrleansRuntime/Silo/Silo.cs
@@ -771,7 +771,7 @@ namespace Orleans.Runtime
                 // 9:
                 SafeExecute(() =>
                 {                
-                    var siloStreamProviderManager = (StreamProviderManager) runtimeClient.CurrentStreamProviderManager;
+                    var siloStreamProviderManager = (StreamProviderManager) StreamProviderManager;
                     scheduler.QueueTask(() => siloStreamProviderManager.CloseProviders(), providerManagerSystemTarget.SchedulingContext)
                             .WaitWithThrow(initTimeout);
                 });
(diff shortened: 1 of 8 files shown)


src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -244,6 +244,7 @@ internal Silo(SiloInitializationParameters initializationParams, IServiceProvide
 
             // GrainRuntime can be created only here, after messageCenter was created.
             grainRuntime = Services.GetRequiredService<IGrainRuntime>();
+            StreamProviderManager = Services.GetRequiredService<IStreamProviderManager>();
```
jason-bragg: I'm not convinced we need StreamProviderManager as member in silo any more, but removing it can easily be considered out of scope for this PR.  Up 2 u. :)

src/OrleansRuntime/Silo/Silo.cs
```diff
@@ -770,7 +771,7 @@ private void Terminate(bool gracefully)
                 // 9:
                 SafeExecute(() =>
                 {                
-                    var siloStreamProviderManager = (StreamProviderManager)grainRuntime.StreamProviderManager;
+                    var siloStreamProviderManager = (StreamProviderManager) runtimeClient.CurrentStreamProviderManager;
```
jason-bragg: why switch to runtimeclient?
why not just pass in local siloStreamProviderManager we already have?

src/OrleansRuntime/Silo/Silo.cs
benjaminpetit: Yes it is something I was considering... But I prefer keeping changes minimal for now, I will remove that after beta if it is ok for you

jason-bragg:

PR #3529, opened 2017-10-10
[Abstractions] Remove dependency of IStorageProvider in Grain<T>

benjaminpetit:
Move logic to construct the storage bridge in the `GrainRuntime` to remove some dependencies in `Grain<T>`.

I chose to use `GrainRuntime` since it was already there. But if the opposition is strong, I could introduce a new factory for that.

### DIFF
--- a/src/Orleans/Runtime/IGrainRuntime.cs
+++ b/src/Orleans/Runtime/IGrainRuntime.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Orleans.Core;
 using Orleans.Streams;
 using Orleans.Timers;
 
@@ -41,5 +42,7 @@ namespace Orleans.Runtime
         void DeactivateOnIdle(Grain grain);
 
         void DelayDeactivation(Grain grain, TimeSpan timeSpan);
+
+        IStorage<TGrainState> GetStateStorageBridge<TGrainState>(Grain<TGrainState> grain) where TGrainState : new();
     }
 }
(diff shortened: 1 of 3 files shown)


src/Orleans/Runtime/IGrainRuntime.cs
```diff
@@ -41,5 +42,7 @@ public interface IGrainRuntime
         void DeactivateOnIdle(Grain grain);
 
         void DelayDeactivation(Grain grain, TimeSpan timeSpan);
+
+        IStorage<TGrainState> GetStateStorageBridge<TGrainState>(Grain<TGrainState> grain) where TGrainState : new();
```
jason-bragg: Storage providers and this behavior are not limited to Grain<T>.
Suggest:
IStorage<TGrainState> GetStateStorageBridge<TGrainState>(Grain grain)

benjaminpetit: Not sure to follow you?

src/Orleans/Runtime/IGrainRuntime.cs
```diff
@@ -41,5 +42,7 @@ public interface IGrainRuntime
         void DeactivateOnIdle(Grain grain);
 
         void DelayDeactivation(Grain grain, TimeSpan timeSpan);
+
+        IStorage<TGrainState> GetStateStorageBridge<TGrainState>(Grain grain) where TGrainState : new();
```
jdom: naming-wise this should be GetStorage, right? The bridge is the implementation

benjaminpetit: I guess it should...

PR #3530, opened 2017-10-10
CachedMessageBlock disposed while in use bug fix.

jason-bragg:
Problem:
Cached message blocks are eligible for purge when full.  Cached message pool was holding onto a block, even if full, until new messages showed up.  This meant it could be holding onto a block that had already been purged.
Fix:
Check block capacity after adding message, and get new block if full.


jdom:
@dotnet-bot test functional
Reason: NullReferenceException in EHSlowConsumingTests (which I believe is unrelated): https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/240/

PR #3533, opened 2017-10-11
Rename sources to fit new naming scheme for 2.0

attilah:
- Rename all source projects, assemblies ~and packages~ to conform with the new planned naming for 2.0
- Sort InternalsVisibleTo for easier maintainability
- Fix Orleans, OrleansCodeGenerator and OrleansRuntime literal references to assembly names to match the new ones.


attilah:

src/Orleans.CodeGeneration.Build/Orleans.CodeGeneration.Build.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Build;PostBuildPublish">
 
   <PropertyGroup>
-    <PackageId>Microsoft.Orleans.OrleansCodeGenerator.Build</PackageId>
+    <PackageId>Microsoft.Orleans.CodeGeneration.Build</PackageId>
```
jdom: I'm fine with renaming DLLs, but changing package IDs should be done into its own targeted PR imo. I'm very worried that will end up with dozens of stale package ids in nuget.org just because we like a new name better. I'm not saying we shouldn't do it if it obviously makes sense, but I'm hesitant of jumping the gun quickly so that you don't have to constantly rebase this huge PR that moves every file.
Another similar risk is that we may release beta in a few days by trying to be quick to merge this, and then realize that we need to change it again, and suddenly have several more package IDs that only exist for a beta version

attilah: I only did what we agreed upon.

Please walk through the changes and try to point out individual packages where you suspect that can "become stale" after preview, because it was not a good rename.

jdom: Well, that would be 1 of the 2 risks. What about the first one? that we'll have both `Microsoft.Orleans.OrleansCodeGenerator.Build` and `Microsoft.Orleans.CodeGeneration.Build` as listed packages in nuget, and not a completely straightforward way for end users to know which one to choose, especially once we start versioning some nuget packages with different numbers (such as Core.Abstractions). Again, not saying we shouldn't fix a few of them, but not duplicate the number of package IDs just for the sake of refactoring the code.
Splitting this change into its own PR should not be a big deal, so I don't understand the push-back. Unless you are OK with rebasing it until we fully settle on how to move forward, I will definitely not single-handedly merge the PR without full consensus.

src/Orleans.CodeGeneration/Orleans.CodeGeneration.csproj
```diff
@@ -1,12 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <PackageId>Microsoft.Orleans.OrleansCodeGenerator</PackageId>
+    <PackageId>Microsoft.Orleans.OrleansCodeGeneration</PackageId>
```
jdom: I think there's a chance we'll drop runtime codegen, or to drastically change its shape, so this is a candidate to become stale

src/Orleans.Core/IDs/ActivationId.cs
```diff
@@ -47,6 +46,19 @@ public static ActivationId GetSystemActivation(GrainId grain, SiloAddress locati
...
+                throw new ArgumentException("ClientGW activation IDs can only be created for client grains");
+
+            // Construct a unique and deterministic ActivationId based on GrainId and SiloAddress.
+            string stringToHash = grain.ToParsableString() + location.Endpoint + location.Generation.ToString(System.Globalization.CultureInfo.InvariantCulture);
+            Guid hash = Utils.CalculateGuidHash(stringToHash);
+            UniqueKey key = UniqueKey.NewKey(hash);
+
+            return FindOrCreate(key);
+        }
+
```
jdom: Bad merge, this method was recently removed from this class

src/Orleans.Runtime/GrainDirectory/ClientObserverRegistrar.cs
```diff
@@ -134,10 +134,7 @@ private ActivationAddress GetClientActivationAddress(GrainId clientId)
             // Need to pick a unique deterministic ActivationId for this client.
             // We store it in the grain directory and there for every GrainId we use ActivationId as a key
             // so every GW needs to behave as a different "activation" with a different ActivationId (its not enough that they have different SiloAddress)
-            string stringToHash = clientId.ToParsableString() + myAddress.Endpoint + myAddress.Generation.ToString(System.Globalization.CultureInfo.InvariantCulture);
-            Guid hash = Utils.CalculateGuidHash(stringToHash);
-            UniqueKey key = UniqueKey.NewKey(hash);
-            return ActivationAddress.GetAddress(myAddress, clientId, ActivationId.GetActivationId(key));
```
jdom: bad merge of the same PR I mentioned earlier

src/Orleans.Core/IDs/ActivationId.cs
attilah: Damn, fixing.

src/Orleans.Runtime/GrainDirectory/ClientObserverRegistrar.cs
attilah: Damn, fixing.

PR #3537, opened 2017-10-11
[Abstractions] Move IKeyedServiceCollection to Abstractions

benjaminpetit:


src/Orleans.Core.Abstractions/Utils/IKeyedService.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    /// Collection of services that can be disambiguated by key
+    /// </summary>
+    public interface IKeyedServiceCollection<in TKey, out TService>
+        where TService : class
+    {
+        TService GetService(IServiceProvider services, TKey key);
+    }
+
+    public interface IKeyedService<TKey, out TService> : IEquatable<TKey>
```
jdom: what's the purpose of moving `IKeyedService`? I think it's only required by the `KeyedServiceCollection` concrete implementation, no?

benjaminpetit: Yes I think you are right. I don't see it used anywhere else

PR #3538, opened 2017-10-11
Migrate GatewayListProvider to use options and Enable TestCluster to use IClientBuilder to configure client 

xiazen:
First Commit: enable TestCluster to use IClientBuilder to configure client

Second commit : Migrate GatewayListProvider to use options
- migrated GatewayListProvider to be configured through options
- legacy way to configure it through client configuration is still supported 
-  added UseXXXGatewayProvider extension methods on IClientBuilder and IServiceCollection
- naming: XXXGatewayProvider. Instead of XXXGatewayListProvider, or XXXGatewayList. But open to better suggestion


src/OrleansTestingHost/TestCluster.cs
```diff
@@ -102,6 +102,16 @@ public TestCluster UseSiloBuilderFactory<TSiloBuilderFactory>() where TSiloBuild
             return this;
         }
 
+        private Func<ClientConfiguration, IClientBuilder> clientBuilderFactory;
+        /// <summary>
+        /// Set client builder factory, which would create a client builder to build the <see cref="TestCluster"/> client.
+        /// </summary>
+        public TestCluster UseClientBuilderFactory(Func<ClientConfiguration, IClientBuilder> clientBuilderFactory)
```
jdom: I plan to revamp this and the silo builder approach very soon, but I think this is OK in the meantime, as it's similar to how currently the silo builder is configured.

xiazen: yeah feel free to revamp it., I just hook it up in the silobuilder fashion in current PR so I can test the new way to configure gateway

xiazen:
ping. needs review

src/OrleansAWSUtils/Options/DynamoDBGatewayProviderOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Text;
+
+namespace OrleansAWSUtils.Options
+{
+    public class DynamoDBGatewayProviderOptions
+    {
+        /// <summary>
+        /// Connection string for DynamoDB Storage
+        /// </summary>
+        public string ConnectionString { get; set; }
```
attilah: As we are using strongly typed options now, we should keep the connectionstring parsing only for legacy and accept DynamoDB native options here like: ```AmazonDynamoDBConfig```, separate ```AccessKey``` and ```Secret``` properties. having a Func<AmazonDynamoDBClient> as part of options is also a good solution, similarly to how other "Provider" configurations are done in AspNetCore.

src/Orleans.Core/AssemblyLoader/AssemblyLoader.cs
```diff
@@ -124,6 +124,20 @@ internal class AssemblyLoader
             }
         }
 
+        /// <summary>
+        /// Create instance for type T in certain assembly, using its parameterless constructor
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="assemblyName"></param>
+        /// <returns></returns>
+        internal static T CreateInstance<T>(string assemblyName)
```
jdom: It's odd to put this method here with this name, because it is very similar to LoadAndCreateInstance, except that it uses Activator instead of ActivatorUtilities. So if you are going to generalize this and move it to this class, it shouldn't be ambiguous compared to the others

xiazen: hmm it is more like CreateInstanceWithoutDIContainerUsingParametrerlessConstructor... But this is a confusing name. I think I will just put this method in the LegacyMembershipConfigurator.cs file so that other people won't use it

src/Orleans.Core/Messaging/IGatewayListProvider.cs
```diff
@@ -14,8 +12,7 @@ public interface IGatewayListProvider
         /// <summary>
         /// Initializes the provider, will be called before all other methods
         /// </summary>
-        /// <param name="clientConfiguration">the given client configuration</param>
-        Task InitializeGatewayListProvider(ClientConfiguration clientConfiguration);
+        Task InitializeGatewayListProvider();
```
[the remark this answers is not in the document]
xiazen: I think we still need this async init method, since User may need to do some async action to initialize their gatewayListProvider. 
Although there's discussion on change from async init method to factory pattern. But I think that is a major pattern change which is out of scope of this PR

src/Orleans.Core/Messaging/ILegacyGatewayListProviderConfigurator.cs
```diff
@@ -0,0 +1,72 @@
...
+using Orleans.Configuration.Options;
+using Orleans.Runtime.Configuration;
+using Orleans.Runtime;
+
+namespace Orleans.Messaging
+{
+    /// <summary>
+    /// LegacyGatewayProviderConfigurator configure GatewayListProvider in the legacy way, which is from ClientConfiguration
+    /// </summary>
+    public interface ILegacyGatewayListProviderConfigurator
```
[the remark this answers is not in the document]
xiazen: It need to be public so it can be accessible in the extension projects, where this interface is implemented. 
As for the doubts on making it public. I worry a little about it too. But since we will be announcing that we are supporting the legacy way to configure gateway and membership, so it make sense for users to understand the existence of this class if they discovered this class. And it is not very discoverable either.

src/Orleans.Core/Messaging/ILegacyGatewayListProviderConfigurator.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        void ConfigureServices(ClientConfiguration configuration, IServiceCollection services);
```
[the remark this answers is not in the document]
xiazen: This is not an extension method on IServiceCollection. So I don't think that rule applies here.

src/Orleans.Core/Messaging/ILegacyGatewayListProviderConfigurator.cs
```diff
@@ -0,0 +1,72 @@
...
+                    break;
+
+                case ClientConfiguration.GatewayProviderType.Custom:
+                    configurator = AssemblyLoader.CreateInstance<ILegacyGatewayListProviderConfigurator>(
+                            clientConfiguration.CustomGatewayProviderAssemblyName);
+                    break;
+
+                case ClientConfiguration.GatewayProviderType.ZooKeeper:
+                    configurator = AssemblyLoader.CreateInstance<ILegacyGatewayListProviderConfigurator>(
+                            Constants.ORLEANS_ZOOKEEPER_UTILS_DLL);
```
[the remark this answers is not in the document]
xiazen: hmmm sounds like that's something you should fix in that naming change PR. but sure I will fix it here....

src/OrleansAWSUtils/AWSUtilsHostingExtensions.cs
```diff
@@ -29,6 +32,30 @@ public static class AwsUtilsHostingExtensions
         }
 
         /// <summary>
+        /// Configure ClientBuilder with DynamoDBGatewayProvider
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="configureOptions"></param>
+        /// <returns></returns>
+        public static IClientBuilder UseDynamoDBGatewayProvider(this IClientBuilder builder,
```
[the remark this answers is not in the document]
xiazen: I used GatewayProvider consistently through the code. I used that one because GatewayProvider is the word got mentioned a lot in our legacy ClientConfiguration. So I think user are more used to GatewayProvider instead of GatewayListProvider. But this is debatable. Any opinion @jdom ?

src/OrleansAWSUtils/Options/DynamoDBGatewayProviderOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace OrleansAWSUtils.Options
+{
+    public class DynamoDBGatewayProviderOptions
```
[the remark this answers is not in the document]
xiazen: I agree that is valuable. But our current implementation doesn't provide the option to configure table name. So I suggest to improve that in another PR. That is not in the scope of migrate gatewaylistprovider to use options anymore, that is improving DynamoDBGatewayProvider configurability .

src/Orleans.Membership.Consul/ConsulBasedGatewayListProvider.cs
```diff
@@ -6,37 +6,39 @@
...
-        public Task InitializeGatewayListProvider(ClientConfiguration configuration)
+        public Task InitializeGatewayListProvider()
         {
-            _maxStaleness = configuration.GatewayListRefreshPeriod;
-
-            this.deploymentId = configuration.DeploymentId;
-
             consulClient =
-                new ConsulClient(config => config.Address = new Uri(configuration.DataConnectionString));
+                new ConsulClient(config => config.Address = new Uri(options.ConnectionString));
```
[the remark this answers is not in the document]
xiazen: I think a more customized option for different gatewayProvider is a good direction to go. That's one of the reasons we moved to options in the first place. But in the other hand, it deviates more from how ConsulGatewayProvider is configured currently. I'm worrying that the more we deviates, the more friction users get when we try to migrate. 

@jdom what do you think?

src/Orleans.Core/Runtime/Constants.cs
```diff
@@ -27,7 +27,6 @@ internal class Constants
         public const string INVARIANT_NAME_SQL_SERVER = "System.Data.SqlClient";
 
         public const string ORLEANS_ZOOKEEPER_UTILS_DLL = "OrleansZooKeeperUtils";
```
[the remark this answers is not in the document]
xiazen: I think Orleans.Membership.ZooKeeper is the new value. I checked the new assembly name for ZooKeeper, it is Orleans.Membership.ZooKeeper

src/Orleans.Core/Configuration/Options/StaticGatewayProviderOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Options for Configure ConfigBasedStaticGateway
+    /// </summary>
+    public class StaticGatewayProviderOptions
+    {
+        /// <summary>
+        /// Static gateways to use
+        /// </summary>
+        public IList<Uri> Gateways { get; set; }
```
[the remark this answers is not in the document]
jdom: Agree, define it as: `public IList<Uri> Gateways { get; set; } = new List<Uri>();`

---
In reply to: [144888645](https://github.com/dotnet/orleans/pull/3538#discussion_r144888645) [](ancestors = 144888645)

src/Orleans.Membership.Consul/ConsulBasedGatewayListProvider.cs
jdom: I think that if we are configuring an Uri and not a connection string, the options should have that. The legacy configurator can still do the same conversion as it was doing, but the options should be correct from the get go. I wouldn't add anything related to certificates or stuff we don't currently support, but as Attila mentioned, strongly typed options open the door for that in the future if someone cares to add it

---
In reply to: [144927825](https://github.com/dotnet/orleans/pull/3538#discussion_r144927825) [](ancestors = 144927825)

src/Orleans.Membership.Consul/LegacyConsulGatewayProviderConfigurator.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Hosting;
+using Orleans.Messaging;
+using Orleans.Runtime.Configuration;
+
+namespace OrleansConsulUtils
+{
+    /// <inheritdoc cref="ILegacyGatewayListProviderConfigurator"/>
```
jdom: nit: just do `<inheritdoc />`
Being explicit makes it tougher for refactoring

src/OrleansAWSUtils/AWSUtilsHostingExtensions.cs
jdom: no strong opinion. Agree with being consistent throughout the configuration. I don't care if it's GatewayListProvider or GatewayProvider, but whatever we choose needs to apply to both the extension methods and the Option classes.

---
In reply to: [144923186](https://github.com/dotnet/orleans/pull/3538#discussion_r144923186) [](ancestors = 144923186)

src/OrleansAWSUtils/Membership/DynamoDBGatewayListProvider.cs
```diff
@@ -17,23 +19,24 @@ internal class DynamoDBGatewayListProvider : IGatewayListProvider
...
         private readonly ILoggerFactory loggerFactory;
-        public DynamoDBGatewayListProvider(ILoggerFactory loggerFactory)
+        private readonly DynamoDBGatewayProviderOptions options;
+        private readonly TimeSpan maxStaleness;
+        public DynamoDBGatewayListProvider(ILoggerFactory loggerFactory, ClientConfiguration clientConfiguration, IOptions<DynamoDBGatewayProviderOptions> options)
         {
             this.loggerFactory = loggerFactory;
+            this.options = options.Value;
+            this.deploymentId = clientConfiguration.DeploymentId;
+            this.maxStaleness = clientConfiguration.GatewayListRefreshPeriod;
```
jdom: This property seems like it should be in all gateway provider Option classes where it applies, doesn't it? Why wasn't it migrated from ClientConfiguration yet? or are there other uses for it?

src/Orleans.Membership.Consul/ConsulBasedGatewayListProvider.cs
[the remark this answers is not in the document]
jdom: I think that if we are configuring an Uri and not a connection string, the options should have that. The legacy configurator can still do the same conversion as it was doing, but the options should be correct from the get go. I wouldn't add anything related to certificates or stuff we don't currently support, but as Attila mentioned, strongly typed options open the door for that in the future if someone cares to add it

src/OrleansAWSUtils/Membership/DynamoDBGatewayListProvider.cs
xiazen: which property? GatwayListRefreshPeriod? 
GatewayListRefreshPeriod should be part of client config since client is the one which refreshes gateway, and client is where this parameter is used mainly. 

MaxStaleness isn't equal to GatewayRefreshPeriod in my understanding of IGatewayListProvider interface. The value of maxStaleness is specific to each gatewayListProvider's implementation.

src/OrleansAWSUtils/AWSUtilsHostingExtensions.cs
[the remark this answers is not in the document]
jdom: no strong opinion. Agree with being consistent throughout the configuration. I don't care if it's GatewayListProvider or GatewayProvider, but whatever we choose needs to apply to both the extension methods and the Option classes.

src/OrleansAWSUtils/Membership/DynamoDBGatewayListProvider.cs
xiazen: It is just happen to be the case that maxStaleness are all set to GatewatListRefreshPeriod in all our gatewayListProvider

jdom: I see. Well, can't dig into it, but I'll trust your judgement, we can always move it later if we need, but you might be right that we can move it to a different Option class that is not gateway impl-specific

xiazen: I'm pretty sure it is the client refresh gateways through one of its component GatewayManager. 
Hence this setting shouldn't be part of GatewayListProviderOptions. GatewayListProvider doesn't refresh itself. 
As to in which way ClientConfiguration will be split to how many options, that is unknown to me. But I think this setting will belong to one of those options.

src/OrleansAWSUtils/AWSUtilsHostingExtensions.cs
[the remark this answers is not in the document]
xiazen: Offline discussion agreed on GatewayListProvider

src/Orleans.Membership.Consul/ConsulBasedGatewayListProvider.cs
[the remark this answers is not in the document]
xiazen: sounds good. I changed ConsuleMembershipOptions to use Uri too in this PR.
 we can change DynamoDBMemebershipOptions to use AccessKey and Secret key etc in another PR. I didn't fix that in this PR. I just fixed ConsulMembership/GatewayListProviderOptions, and DYnamoDBGatewayListProvider

src/Orleans.Core/Messaging/ILegacyGatewayListProviderConfigurator.cs
```diff
@@ -57,13 +56,28 @@ internal class LegacyGatewayProviderConfigurator
...
+        /// <summary>
+        /// Create instance for type T in certain assembly, using its parameterless constructor
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="assemblyName"></param>
+        /// <returns></returns>
+        internal static T CreateInstanceWithParameterlessConstructor<T>(string assemblyName)
+        {
+            var assembly = Assembly.Load(new AssemblyName(assemblyName));
+            var foundType = TypeUtils.GetTypes(assembly, type => typeof(T).IsAssignableFrom(type), null).FirstOrDefault();
```
jdom: You removed the `First()`, but you must now check for null explicitly and throw a specific InvalidOperationException if it is. This is particularly useful in case the users use `Custom`.

test/Tester/ClientConnectionTests/GatewayConnectionTests.cs
```diff
@@ -48,16 +49,24 @@ public override TestCluster CreateTestCluster()
         {
             var options = new TestClusterOptions(1)
             {
-                ClientConfiguration =
-                {
-                    GatewayProvider = ClientConfiguration.GatewayProviderType.Custom,
-                    CustomGatewayProviderAssemblyName = "Tester",
-                    GatewayListRefreshPeriod = TimeSpan.FromMilliseconds(100)
```
jdom: you lost this configuration in the transformation, is it not needed?

jdom:
LGTM other than those minor comments I added

src/Orleans.Core/Messaging/ILegacyGatewayListProviderConfigurator.cs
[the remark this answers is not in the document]
attilah: I know it is not an extension method, and not that I'm arguing, but logically when configuring something, that something should be the first which shows that here is what I am configuring and then the parameters coming for that configuration...but as you like then.

src/Orleans.Core/Messaging/IGatewayListProvider.cs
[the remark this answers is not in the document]
attilah: Ok

test/Tester/ClientConnectionTests/GatewayConnectionTests.cs
xiazen: no I didn't. take a look on the right :) ClientConfiguration = {GatewayListRefreshPeriod = TimeSpan.FromMilliseconds(100)}

jdom:
@dotnet-bot test bvt
Reason: unrelated failure

PR #3539, opened 2017-10-11
[Abstractions] Move some forgotten types

benjaminpetit:
Not many changes, just moving some interfaces/classes to Abstractions


jdom:
>     public enum SiloStatus

why move SiloStatus? Doesn't seem to be a general purpose abstraction, but I could be convinced

---
Refers to: src/Orleans.Core.Abstractions/Runtime/SiloStatus.cs:6 in cff3f44. [](commit_id = cff3f4400fe0d497cc54d836f9f42d31b724e6f4, deletion_comment = False)

jdom:
>     public interface IPlacementContext

Are these placement related abstractions needed for normal grain development? Because if it's just for extending Orleans, I wouldn't move them to Abstractions

---
Refers to: src/Orleans/Placement/IPlacementContext.cs:8 in cff3f44. [](commit_id = cff3f4400fe0d497cc54d836f9f42d31b724e6f4, deletion_comment = True)

jdom:
>     public interface IGrainExtension : IAddressable

The comment seems to indicate this is only used for internal runtime extensions... Do we need to move it?

---
Refers to: src/Orleans.Core.Abstractions/Runtime/IGrainExtension.cs:6 in cff3f44. [](commit_id = cff3f4400fe0d497cc54d836f9f42d31b724e6f4, deletion_comment = False)

benjaminpetit:
> why move SiloStatus? Doesn't seem to be a general purpose abstraction, but I could be convinced otherwise
Needed by `IPlacementContext` but...
> Are the placement related abstractions needed for normal grain development? Because if it's just for extending Orleans, I wouldn't move them to Abstractions
Not sure what we can call "Normal" but yes I guess we can leave them in Orleans...

ReubenBond:
`IGrainExtension` is public now, so we should move it and maybe update the comment.

... eventually we should document it.

jdom:
Thanks @ReubenBond , yeah, we looked at it closer with Benjamin, and it looks safe to move. It's a marker interface anyway, and we do codegen based on it too, so makes sense to move

jdom:
Regarding placement context and SiloStatus, I would leave them out in that case. Domain logic / grain developers would not require it. Only devs extending the runtime, which is not the audience of this Core.Abstractions package. It is fine that they reference Orleans.Core for that extension

sergeybykov:
I think it's an interesting question - whether or not placement related interfaces (`IPlacementDirector`, `IPlacementContext`, `PlacementStrategy`, `PlacementTarget`) should be part of Abstractions or not.

My feeling, not strong though, is that they can be viewed as part of the grain programming model (for custom placement policies). But I can also see the argument that they are runtime extensions.

PR #3542, opened 2017-10-12
[Abstractions] Move GrainCancellationTokenSource and GrainCancellationToken in Abstractions

benjaminpetit:
To remove some runtime code from `GrainCancellationToken` I had to introduce a new interface to isolate some code (like we did for `Grain` with `GrainRuntime`)


src/Orleans.Core/Cancellation/GrainCancellationToken.cs
```diff
@@ -69,43 +64,22 @@ internal GrainCancellationToken(Guid id, bool canceled) : this(id)
...
         {
-            // propagate the exception from the _cancellationTokenSource.Cancel back to the caller
-            // but also cancel _targetGrainReferences. 
-            Task task = OrleansTaskExtentions.WrapInTask(_cancellationTokenSource.Cancel);
-
-            if (_targetGrainReferences.IsEmpty)
+            if (_cancellationTokenRuntime == null)
             {
-                return task;
+                _cancellationTokenSource.Cancel();
```
jdom: this can now throw synchronously, where as before we were returning a faulted task if that was the case

src/Orleans.Core/Cancellation/GrainCancellationToken.cs
```diff
@@ -32,6 +24,8 @@ public sealed class GrainCancellationToken : IDisposable
         [NonSerialized]
         private readonly ConcurrentDictionary<GrainId, GrainReference> _targetGrainReferences;
 
+        private IGrainCancellationTokenRuntime _cancellationTokenRuntime;
```
jdom: This should be marked with NonSerialized I believe, right?

src/Orleans.Core/Runtime/GrainReferenceRuntime.cs
```diff
@@ -18,6 +18,7 @@ internal class GrainReferenceRuntime : IGrainReferenceRuntime
         private readonly ILogger logger;
         private readonly IInternalGrainFactory internalGrainFactory;
         private readonly SerializationManager serializationManager;
+        private readonly IGrainCancellationTokenRuntime cancellationTokenRuntime = new GrainCancellationTokenRuntime();
```
jdom: you could register a singleton in the container and reduce the number of these runtimes by just using DI

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -104,6 +104,8 @@ internal static void AddDefaultServices(HostBuilderContext context, IServiceColl
             services.AddFromExisting<IProviderManager, IStreamProviderManager>();
             services.TryAddSingleton<GrainRuntime>();
             services.TryAddSingleton<IGrainRuntime, GrainRuntime>();
+            services.TryAddSingleton<GrainCancellationTokenRuntime>();
```
jdom: why register as a concrete class? Also, these 2 registration will end up as 2 different singletons by the by the way they were registered

src/Orleans.Core/Runtime/GrainReferenceRuntime.cs
jdom: BTW, I now saw that you registered the singleton on the silo side. You need to register in the client side, and then you should be able to resolve from this class' constructor

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
benjaminpetit: I just did that same thing as the `GrainRuntime`

src/Orleans.Core/Cancellation/GrainCancellationToken.cs
benjaminpetit: Forgot to remove this, as it is done in the runtime part

src/Orleans.Core.Abstractions/Cancellation/GrainCancellationToken.cs
```diff
@@ -69,49 +65,43 @@ internal GrainCancellationToken(Guid id, bool canceled) : this(id)
...
-            // propagate the exception from the _cancellationTokenSource.Cancel back to the caller
-            // but also cancel _targetGrainReferences. 
-            Task task = OrleansTaskExtentions.WrapInTask(_cancellationTokenSource.Cancel);
-
-            if (_targetGrainReferences.IsEmpty)
+            if (_cancellationTokenRuntime == null)
             {
-                return task;
+                WrapInTask(_cancellationTokenSource.Cancel);
+                return Task.CompletedTask;
```
jdom: you need to return the result of `WrapInTask`.
BTW, since this is not a general purpose method, just inline the wrapping here and do the try/catch in the same method, no need to create a separate WrapInTask method

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
jdom: You uncovered a (non-impacting) bug with GrainRuntime. We shouldn't do that. I checked and there's no usages of resolving the concrete `GrainRuntime` class, so that's why it's not impacting, but we should remove the registration. If we wanted consumers to resolve either of those, we should use the AddFromExisting extension method (but here we should remove the concrete class registration anyway)

jdom:
LGTM after you address the 2 remaining comments

PR #3546, opened 2017-10-13
Limit request context to messaging layer

jason-bragg:
We use AsyncLocal for RequestContext, but that mechanism allows the request context to be passed to other scheduled contexts, when it should be confined to a grain call.  This change clears the request context for any scheduled actions.  If the action is a grain or system target call, the invoker will get the request context from the message, so request context via messaging should function correctly.


jason-bragg:
Related note:  This is the cause of threading contention on the TransactionInfo storage in the request context.

veikkoeeva:
One note `ShedulerExtensions.cs`, has a typo I see.

jason-bragg:
@veikkoeeva, I almost asked what the typo was.. *face palm*

src/Orleans.Runtime/Scheduler/SchedulerExtensions.cs
```diff
@@ -13,6 +13,7 @@ internal static Task<T> QueueTask<T>(this OrleansTaskScheduler scheduler, Func<T
                 {
                     try
                     {
+                        RequestContext.Clear();
```
ReubenBond: Is the request context valid after awaiting `QueueTask`?
```C#
RequestContext.Set("val", 5);
await scheduler.QueueTask(() => directory.LookupAsync(xxx), directory.SchedulingContext);
// Is RequestContext valid here?
Assert.Equal(RequestContext.Get("val"), 5); 
```

rikbosch:
Do these changes have any effect on stream processing? Currently the requestcontext travels with a published message from producer to consumer (at least in the azure queue provider).

sebastianburckhardt:
> but that mechanism allows the request context to be passed to other scheduled contexts, when it should be confined to a grain call

I am assuming that the request context will still be around for all async resumptions (i.e. code that resumes execution after awaiting remote grain calls or other IO operations), as long as those resumptions happen before the original grain call returns?

src/Orleans.Runtime/Scheduler/SchedulerExtensions.cs
jason-bragg: Yes.  If the action is actually asynchronous, the logic you wrote should work.  However if it is not actually asynchronous the callers context may be cleared.  For instance:

     RequestContext.Set("val", 5);
     await Blarg();
     Assert.Equal(RequestContext.Get("val"), 5);   // will pass, because Blarg is asynchronous.
     await Blarg2();
     Assert.Equal(RequestContext.Get("val"), 5);  // will fail, because Blarg2 is not asynchronous.

     async Task Blarg()
     {
         RequestContext.Clear();
         await Task.Delay(1);
     }

     Task Blarg2()
     {
         RequestContext.Clear();
         return Task.CompletedTask;
     }

I'll add some testing to confirm.  If anyone knows how to force this behavior and have it not depend on whether the function is actually asynchronous or not, I'd love to here about it.  Best I could find is adding an await Task.Delay(Timespan.Zero) to the call. but I wasn't comfortable with that.

jason-bragg:
@rikbosch  For persistent streams (non-sms), the caller's context is stored with the event, and the context is reset prior to delivering the event to consumers.  That behavior 'should' not be altered by this change.

jason-bragg:
@sebastianburckhardt That is my understanding, yes.  The risk here is when the awaited operation is not actually asynchronous, in that case, if we clear the RequestContext, without an actual asynchronous operation to force the new asynchronous frame, we'll be clearing the original RequestContext.  Plz see response to Reuben's related question.

src/Orleans.Runtime/Scheduler/SchedulerExtensions.cs
sebastianburckhardt: I think you can test for a "pseudo-async" by examining the returned task object
```
var task = Blarg();
if (task.IsCompleted)
{
    // Blarg was not async but executed synchronously
}
else
{
   // Blarg is really async
}
await task;

```

jason-bragg: Understood, but I'm unclear how that test would help us.

sebastianburckhardt: perhaps you could use this to detect if you overzealously deleted the callers requestcontext (i.e. if it is cleared before a synchronous call), and restore it.

jason-bragg: Hmm, something about the QueueTask is ensuring that it's always being handled correctly.  Added tests to confirm.  Will update PR with them.

ReubenBond: Ah ok, I believe this works because of copy-on-write semantics of the backing store for AsyncLocal. I was unaware before. I wonder if there's a need to clear the context in any other cases or if there's a way to move this clearing into the scheduling layer and have it more centralized (eg, clear when switching to a different scheduling context - not sure).

PR #3547, opened 2017-10-13
Fixed intermitent transaction orchestration test failures

jason-bragg:
Two fixes:
- Transaction calls return when transaction is complete, but the transactional resource itself may not have been told that the transaction completed yet, so this change waits a little while for that to occur.
- Transactional resources will (eventually) be told about the highest committed transaction, but will not necessarily be told about every committed transaction, so we test only against the highest.

### DIFF
--- a/test/Orleans.Transactions.Tests/Runners/OrchestrationsTransactionsTestRunner.cs
+++ b/test/Orleans.Transactions.Tests/Runners/OrchestrationsTransactionsTestRunner.cs
@@ -9,8 +9,13 @@ namespace Orleans.Transactions.Tests
 {
     public abstract class OrchestrationsTransactionsTestRunner : TransactionTestRunnerBase
     {
-        protected OrchestrationsTransactionsTestRunner(IGrainFactory grainFactory, ITestOutputHelper output)
-        : base(grainFactory, output) { }
+        private readonly TimeSpan DefaultWaitTime = TimeSpan.FromSeconds(30);
+        private readonly TimeSpan waitTime;
+        protected OrchestrationsTransactionsTestRunner(IGrainFactory grainFactory, ITestOutputHelper output, TimeSpan? waitTime = null)
+        : base(grainFactory, output)
+        {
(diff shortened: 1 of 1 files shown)


jdom:
@dotnet-bot test bvt

test/Orleans.Transactions.Tests/Runners/OrchestrationsTransactionsTestRunner.cs
```diff
@@ -108,18 +113,40 @@ protected OrchestrationsTransactionsTestRunner(IGrainFactory grainFactory, ITest
...
             TransactionOrchestrationResult results = await resultGrain.GetResults();
 
-            Assert.Equal(perpareCount, results.Prepared.Count);
-            Assert.Equal(abortCount, results.Aborted.Count);
-            Assert.Equal(commitCount, results.Committed.Count);
+            if(assert)
+            {
+                Assert.Equal(perpareCount, results.Prepared.Count);
+                Assert.Equal(abortCount, results.Aborted.Count);
+                Assert.Equal(results.Prepared.Max(), results.Committed.Aggregate((long)0, (t1, t2) => Math.Max(t1, t2)));
```
jdom: Not entirely sure I follow this logic, as I'm not familiar with this structure we are using. But it kind of looks like you should be using `Sum` instead of `Max` if intermediate commits are batched, or am I misunderstanding the assert?

ReubenBond: Shouldn't the check be "max prepared = max committed" in which case, `.Max()` on both collections?

jason-bragg: The prepare is called for every transaction, so the list should have all the transaction ID's in it.  The commit is called for at least the highest transaction.  So the max of the prepared should match the max of the committed transaction id's.  I use an aggregate function rather than a link max, because link max will throw an exception when there are 0 items, and aggregate will not.

PR #3548, opened 2017-10-13
[Abstractions] Enable codegen on project that reference only Abstractions

benjaminpetit:
~~Please do not review the first 3 commits, since they are in PR #3542. The most relevant commit to review here is the last one (c97a25f)~~

To enable codegen on project that only reference Abstractions I had to move some files and modify some check in codegen.

I also had to add a nuget reference to `System.Reflection.Emit.Lightweigh` for now. We will work to remove it in 2.0 final


jdom:
@dotnet-bot test this please
Reason: Several codegen changes were merged in master

src/Orleans.Core.Abstractions/Async/PublicOrleansTaskExtensions.cs
```diff
@@ -0,0 +1,168 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Orleans
+{
+    /// <summary>
+    /// Utility functions for dealing with Tasks.
+    /// </summary>
+    public static class PublicOrleansTaskExtensions
```
jdom: Let's not move these to Abstractions. They are not Orleans related, they are just helpers, and we have been guilty of changing these and breaking people that use binding redirects in the past already.
Also we have some potential plans for codegen to avoid these much boxing that we do.

If some method is required by generated code, we can discuss ways to avoid it. It might be fine to let it go for beta, but not for the final release.

src/Orleans.Core.Abstractions/Orleans.Core.Abstractions.csproj
```diff
@@ -13,6 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.Reflection.Emit.Lightweight" Version="$(SystemRuntimeVersion)" />
```
jdom: Similarly for other reviewers, we should eventually get rid of this dependency before we release 2.0 final.

jdom:
Please rebase this when you have a chance, so that we can review without the extra stuff

benjaminpetit:
Rebased

src/Orleans.CodeGeneration.Build/CodeGenerator.cs
```diff
@@ -20,7 +20,7 @@ namespace Orleans.CodeGeneration
 {
     public class CodeGenerator : MarshalByRefObject
     {
-        public static readonly string OrleansAssemblyFileName = Path.GetFileName(typeof(RuntimeVersion).Assembly.Location);
+        public static readonly string OrleansAssemblyFileName = Path.GetFileName(typeof(IGrain).Assembly.Location);
```
jdom: Rename to `OrleansCoreAbstractionsAssemblyFileName`

jdom:
>     public static class TaskUtility

this also shouldn't be moved by the final release

---
Refers to: src/Orleans.Core.Abstractions/Async/TaskUtility.cs:9 in 1a449a1. [](commit_id = 1a449a1811159a132bc009bfba1e8f1f79dbdf91, deletion_comment = False)

jdom:
>         internal static GrainReference AsWeaklyTypedReference(this IAddressable grain)

do we need this method in Abstractions?

---
Refers to: src/Orleans.Core.Abstractions/Core/GrainExtensions.cs:16 in 1a449a1. [](commit_id = 1a449a1811159a132bc009bfba1e8f1f79dbdf91, deletion_comment = False)

jdom:
>     internal interface ISystemTargetBase

is this needed in abstractions?

---
Refers to: src/Orleans.Core.Abstractions/SystemTargetInterfaces/ISystemTargetBase.cs:8 in 1a449a1. [](commit_id = 1a449a1811159a132bc009bfba1e8f1f79dbdf91, deletion_comment = False)

benjaminpetit:
- About `TaskUtility`: used in generated code
- About `ISystemTargetBase`: used by `GrainExtensions.AsWeaklyTypedReference`
- About `GrainExtensions.AsWeaklyTypedReference`: used by the public `AsReference`, that is used by generated code

jdom:
Let's add all of those to the list of candidates to remove for final. In particular TaskUtility is a very bad utility, and also is not very different in nature than `PublicOrleansTaskExtensions`

jdom:
@dotnet-bot test bvt
Reason: merged another PR, making sure the moved files are unaffected for the build

PR #3552, opened 2017-10-16
[Abstractions] Move RequestContext, public Exceptions and grain attributes

benjaminpetit:


src/Orleans.Core.Abstractions/Runtime/RequestContext.cs
```diff
@@ -142,7 +142,7 @@ public static void Import(Dictionary<string, object> contextData)
             }
         }
 
-        public static Dictionary<string, object> Export(SerializationManager serializationManager)
+        public static Dictionary<string, object> Export(ISerializationManager serializationManager)
```
jdom: If we extracted the serialization manager interface only for this, I believe it's the wrong approach. Here we were leveraging the DeepCopy method only, but what's most important is that the Import, Export and Clear are all APIs relevant to the runtime, but not to end-users, so they could be split from the abstractions move

src/Orleans.Core/Core/DeadlockException.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// Deadlock detection is not enabled by default in Orleans silos, 
+    /// because it introduces some extra overhead in call handling.
+    /// </para>
+    /// <para>
+    /// There are some constraints on the types of deadlock that can currently be detected 
+    /// by Orleans silos.
+    /// </para>
+    /// </remarks>
+    [Serializable]
+    public class DeadlockException : OrleansException
```
jdom: why not move this one to the Exceptions folder with the rest?

benjaminpetit: Forgot to remove it

jdom:
@dotnet-bot test functional

src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -357,8 +357,14 @@ private void CheckDeadlock(Message message)
...
-                throw new DeadlockException(newChain);
+
+                throw new DeadlockException(
+                    String.Format(
+                        "Deadlock Exception for grain call chain {0}.",
+                        Utils.EnumerableToString(
+                            newChain,
+                            elem => String.Format("{0}.{1}", elem.GrainId, elem.DebugContext)),
+                    newChain.Select(req => new Tuple<GrainId, string>(req.GrainId, req.DebugContext)).ToList()));
             }
```
ReubenBond: There's a `)` and `,` in the wrong place here, my PR told me 😜

PR #3553, opened 2017-10-16
All_Small non-deterministic test failure fix

attilah:
All_Small test is time sensitive, timeout raised for the conflict resolution part to make it pass on VSO


jdom:
@dotnet-bot test bvt

PR #3556, opened 2017-10-17
Update 1.5.2 changelog with #3522

sergeybykov:


jdom:
We'll need to update the changelog in master too

Issue #3557, opened 2017-10-17
Cleaning Orleans.Core.Abstractions


benjaminpetit:
For 2.0 beta, we moved some internals types to Abstractions, that we want to remove it for 2.0 final.

- [x] Remove `FieldUtils` to remove dependency on `System.Reflection.Emit.Lightweight`
- [x] Embbed in generated code used code from `PublicOrleansTaskExtensions`, `TaskUtils` and `GrainExtensions`
- [x] Remove the member `Silo.StreamProviderManager` and uses DI instead
- [x] Add missing XML comments
- [x] Move logger to a "legacy" project

ReubenBond:
I assume we're planning to keep `PublicOrleansTaskExtensions.Ignore()` exposed. Perhaps it should be in Core and not Abstractions, though

jason-bragg:
Orleans has a set of 'utilities' that are not central to the actor model.  We've discussed how to expose these without adding risk to the core framework and abstractions.  I suspect these should be considered in that discussion.

sergeybykov:
`Ignore()` is a good example of the tradeoff. We want (I think) it available to grain developers. So it can't be in Core, even though it has some logic, and in general we don't want to have logic in Abstractions.

benjaminpetit:
It seems that everything is ticked now. We can close this one

PR #3559, opened 2017-10-18
Move Task Box method to Core & remove TaskUtility

ReubenBond:
Part of #3557

In generated invokers, we will use `return await <method call>` instead of `return <method call>.Box()`.


src/Orleans.Core/Async/TaskExtensions.cs
```diff
@@ -1,13 +1,135 @@
...
+            return default(object);
+        }
+
+        private static async Task<object> BoxAwait<T>(Task<T> task)
+        {
+            return await task;
+        }
+
+        private static Task<T> UnboxContinuation<T>(Task<object> task)
+        {
```
ReubenBond: I believe there's room for improvement here and in related methods, but we can do that in another PR.

ReubenBond:
The failures here are fixed in #3560

jdom:
@dotnet-bot test functional

src/Orleans.Core/Async/TaskExtensions.cs
jdom: Absolutely, we can also update codegen to avoid so much boxing, at least separating the paths for Task and TaskOfT

PR #3560, opened 2017-10-18
Fix deadlock exception + tweak tests

ReubenBond:
The exception didn't have its `CallChain` property populated, causing some functionals to fail.

Changed test code to use `ThrowsAsync<T>` since we want to ensure that the test does throw.


gabikliot:
:-)

Good to see that finally this is being used. It was on a shelf for probably ... 4 years.

PR #3564, opened 2017-10-18
Add SourceLink support for easier debugging of nuget packages

attilah:
This PR enables the [SourceLink](https://github.com/ctaggart/SourceLink) feature or Portable PDBs, which enables the debugging of Orleans code from consumed NuGet packages.

A blog about source link debugging configuration for AspNetCore 2.0 can be found [here](https://www.stevejgordon.co.uk/debugging-asp-net-core-2-source).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -17,7 +17,7 @@
     <RepositoryUrl>https://github.com/dotnet/Orleans</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
     <IncludeSymbols>true</IncludeSymbols>
-    <IncludeSource>true</IncludeSource>
+    <IncludeSource>false</IncludeSource>
   </PropertyGroup>
 
   <!-- Common compile parameters -->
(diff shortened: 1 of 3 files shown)


Directory.Build.props
```diff
@@ -17,7 +17,7 @@
     <RepositoryUrl>https://github.com/dotnet/Orleans</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
     <IncludeSymbols>true</IncludeSymbols>
-    <IncludeSource>true</IncludeSource>
+    <IncludeSource>false</IncludeSource>
```
jdom: This would affect the *.symbols.nupkg, right? I thought the plan was to include the sourcelink metadata, especially for the Microsoft Symbols server, but I don't think we should modify the symbols nuget packages. Whether we decide to not publish them later is another question, but those are still useful for people forking the repo I believe.

attilah: Yes, that is right, ok, let me revert this.

PR #3568, opened 2017-10-19
Service Fabric: ensure all silos reach a terminal state.

ReubenBond:
Unlike other membership providers, the Service Fabric provider does not record a total set of all silos which ever attempted to join a cluster. Because of this, there are cases where a silo might query the membership oracle for silos which were never before seen by the oracle.

This PR adds an `UnknownSiloMonitor` which attempts to alleviate issues which arise because of this. It records queries for the status of unknown silos and tracks those silos until:
* The silo is determined to be alive by a membership update
* The silos is determined to be dead because:
  * A silo with the same IP:Port exists with a later generation
  * A silo with the same IP (but different port) exists with a later generation in a singleton partition (max 1 silo per IP) **Note: this won't work on development clusters since all nodes share an IP**
  * A period of time passes, defaulting to 1 minute and configurable using the Options subsystem via `ServiceFabricMembershipOptions`.

Additionally, this PR improves the logic around querying Service Fabric for cluster membership changes. Previously it's possible that membership queries were returning cached results. By caching the results of previous queries and passing those cached results to subsequent queries we tell SF to avoid returning a cached result. If this performs poorly then we should consider increasing the polling interval from 5 seconds.

### DIFF
--- a/src/Orleans.Membership.ServiceFabric/FabricMembershipOracle.cs
+++ b/src/Orleans.Membership.ServiceFabric/FabricMembershipOracle.cs
@@ -19,7 +19,6 @@ namespace Microsoft.Orleans.ServiceFabric
         private readonly Dictionary<SiloAddress, SiloEntry> silos = new Dictionary<SiloAddress, SiloEntry>();
         private readonly ConcurrentDictionary<ISiloStatusListener, ISiloStatusListener> subscribers =
             new ConcurrentDictionary<ISiloStatusListener, ISiloStatusListener>();
-
         private readonly AutoResetEvent notificationEvent = new AutoResetEvent(false);
         private readonly BlockingCollection<StatusChangeNotification> notifications = new BlockingCollection<StatusChangeNotification>();
         private readonly TimeSpan refreshPeriod = TimeSpan.FromSeconds(5);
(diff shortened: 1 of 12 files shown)


src/Orleans.Membership.ServiceFabric/FabricMembershipOracle.cs
```diff
@@ -170,40 +152,32 @@ public SiloStatus GetApproximateSiloStatus(SiloAddress siloAddress)
...
-        /// </summary>
-        /// <returns>The silo so ask about.</returns>
+        /// <inheritdoc />
         public bool IsFunctionalDirectory(SiloAddress siloAddress)
         {
             if (siloAddress.Equals(this.SiloAddress)) return true;
 
             var status = this.GetApproximateSiloStatus(siloAddress);
-            return !status.IsTerminating();
+            return !status.IsUnavailable();
```
sergeybykov: Why not `!(status.IsTerminating() || status.IsUnavailable())`?

ReubenBond: `IsUnavailable` is defined as `siloStatus.IsTerminating() || siloStatus == SiloStatus.None;`, so the additional call to `IsTerminating()` would be redundant.

ReubenBond:
I pushed a second commit which removes the Singleton partition check since it wouldn't work on dev clusters.

PR #3573, opened 2017-10-19
Move PublishDir from Bootstrap folder

jdom:
As it's not related to Bootstrap, and it's impacting signing


jdom:
Verified that this works in VSO for our signing. Please merge ASAP.
I tried to use: `<PublishRoot>$(BaseOutputPath)\$(Configuration)\$(PublishDirName)\</PublishRoot>` instead of `<PublishRoot>bin\$(Configuration)\publish\</PublishRoot>` but neither `BaseOutputPath` nor `PublishDirName` seems to be defined at that point it time, so I had to hardcode those, which should be safe anyway.

PR #3578, opened 2017-10-21
AddApplicationPartsFromReferences includes the specified assembly

jdom:
- AddApplicationPartsFromReferences was not adding the specified assembly as a part, but it should, so it's doing that now.
- Fix extension method to target ISiloHostBuilder


jdom:
These fixes would make the experience easier, as can be seen in #3579

PR #3585, opened 2017-10-23
CodeGen: Set Publish=true on PackageReference elements to fix deps.json

ReubenBond:


claylaut:
@jdom,
Can you please take a look at this? Since its quite small.

Cause currently there are two issues:
- Grain not implemented exception at runtime.
- Could not load file or assembly 'Microsoft.Extensions.DependencyModel

Thanks in advance.
Clayton

jdom:
Absolutely, I'm taking a look. BTW, as I mentioned in twitter, installing codegen only in grains and interface projects is correct. The reason for the runtime exception not finding the implementations is that you need to add the relevant application parts in the client and silo, so assuming codegen does indeed work on those projects, you shouldn't be blocked. See in the HelloWorld.NetCore sample for how it is done: https://github.com/dotnet/orleans/blob/master/Samples/HelloWorld.NetCore/src/SiloHost/Program.cs#L44 (Same on client)

src/Orleans.CodeGeneration.Build/Orleans.CodeGeneration.Build.csproj
```diff
@@ -36,16 +37,19 @@
...
+    <PackageReference Include="System.Runtime.Loader" Version="$(SystemRuntimeVersion)" />
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsLoggingVersion)" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildVersion)" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildVersion)" />
+  </ItemGroup>
+  <ItemGroup>
+    <!-- This must come after all other PackageReference elements -->
+    <PackageReference Update="@(PackageReference)" PrivateAssets="All" Publish="true" />
```
jdom: I couldn't find much info on this. Is it an undocumented feature?

I did validate that it works though, both in command line and VS (in Windows, both .NET Framework and .NET Core)

jdom: BTW, what I couldn't find information is on the Publish=true thing. I do understand PrivateAssets though

src/Orleans.CodeGeneration.Build/Orleans.CodeGeneration.Build.csproj
```diff
@@ -20,6 +20,7 @@
     <PublishDir>$(PublishRoot)$(TargetFramework)</PublishDir>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
+    <DevelopmentDependency>true</DevelopmentDependency>
```
jdom: Unfortunately NuGet tooling does not yet honor this on restore, but eventually they will, so it makes sense to add it anyway. Tracking issue: https://github.com/NuGet/Home/issues/4125

src/Orleans.CodeGeneration.Build/Orleans.CodeGeneration.Build.csproj
ReubenBond: I didn't find any docs on `Publish=true`, but I traced it back through to the targets file. See here: https://github.com/dotnet/sdk/blob/5d4aa5c712512861c5f0987f6695ca0868192aca/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Publish.targets#L508

claylaut:
@jdom Thanks for updating the sample it was super helpful, now its working perfectly :)

I didnt know about this ` .AddApplicationPartsFromReferences(typeof(HelloGrain).Assembly)`

Issue #3293, opened 2017-08-10
Make interception of unhandled exceptions optional and non-static


sergeybykov:
Today, `UnobservedExceptionsHandlerClass` unconditionally intercept unhandled exceptions by adding itself to `TaskScheduler.UnobservedTaskException`. This interferes with application code if it intercepts those exceptions already and pollutes the logs. Need to make our interception optional (still on by default).

For logging unhandled exceptions, `UnobservedExceptionsHandlerClass` gets a logger instance via a static `LogManager.GetLogger("UnobservedExceptionHandler")` call. That likely has to change too, to align with the logging infra getting revamped.

xiazen:
The way it gets its logger will be changing from LogManager.GetLogger to get a ILogger<UnobservedExceptionsHandlerClass > from DI, after we completely migrate to asp.net logging abstraction. 

To make it optional, the class will need to at least remove its static constructor. But I would suggest the following: 
- remove its static properties and methods. Make it a singleton in DI if multiple classes in the same runtime need to access it. But based on my observation in code, that is not the case. So mostly it will just become a normal class . 
- add a config in ClusterConfiguration for user to turn it on or off. By default turn it off. 

what do you think?

xiazen:
can close after #3829 is merged

PR #3592, opened 2017-10-23
Remove static UnobservedExceptionsHandlerClass and detach from client library

xiazen:
PR for issue #3293 . 
This PR makes interception of unhandled exceptions non-static, as well as it doesn't listen to unobserved task exceptions in the client library (as the client is not a hosting runtime).
Still may need another PR to make interception of unhandled exceptions completely optional on the Silo too, if that's desired.

### DIFF
--- a/src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
+++ b/src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
@@ -6,17 +6,18 @@ namespace Orleans
 {
     using Orleans.Runtime;
 
-    internal class UnobservedExceptionsHandlerClass
+    internal class UnobservedExceptionsHandler
     {
-        private static readonly Object lockObject = new Object();
-        private static ILogger logger;
-        private static UnobservedExceptionDelegate unobservedExceptionHandler;
-        private static readonly bool alreadySubscribedToTplEvent = false;
+        private readonly Object lockObject = new Object();
+        private ILogger logger;
+        private UnobservedExceptionDelegate unobservedExceptionHandler;
+        private readonly bool alreadySubscribedToTplEvent = false;
 
         internal delegate void UnobservedExceptionDelegate(ISchedulingContext context, Exception exception);
         
-        static UnobservedExceptionsHandlerClass()
+        public UnobservedExceptionsHandler(ILogger<UnobservedExceptionsHandler> logger)
         {
+            this.logger = logger;
             lock (lockObject)
             {
                 if (!alreadySubscribedToTplEvent)
@@ -27,12 +28,8 @@ namespace Orleans
             }
         }
 
-        internal static void InitLogger(ILoggerFactory loggerFactory)
-        {
-            logger = loggerFactory.CreateLogger<UnobservedExceptionsHandlerClass>();
-        }
 
-        internal static bool TrySetUnobservedExceptionHandler(UnobservedExceptionDelegate handler)
+        internal bool TrySetUnobservedExceptionHandler(UnobservedExceptionDelegate handler)
         {
             if (handler == null) throw new ArgumentNullException(nameof(handler));
             lock (lockObject)
@@ -48,7 +45,7 @@ namespace Orleans
             return true;
         }
 
-        internal static void ResetUnobservedExceptionHandler()
+        internal void ResetUnobservedExceptionHandler()
         {
             lock (lockObject)
             {
@@ -56,7 +53,7 @@ namespace Orleans
             }
         }
 
-        private static void InternalUnobservedTaskExceptionHandler(object sender, UnobservedTaskExceptionEventArgs e)
+        private void InternalUnobservedTaskExceptionHandler(object sender, UnobservedTaskExceptionEventArgs e)
         {
             var aggrException = e.Exception;
             var baseException = aggrException.GetBaseException();
--- a/src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
+++ b/src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
@@ -6,12 +6,12 @@ namespace Orleans
 {
     using Orleans.Runtime;
 
-    internal class UnobservedExceptionsHandler
+    internal class UnobservedExceptionsHandler : IDisposable
     {
         private readonly Object lockObject = new Object();
-        private ILogger logger;
+        private readonly ILogger logger;
         private UnobservedExceptionDelegate unobservedExceptionHandler;
-        private readonly bool alreadySubscribedToTplEvent = false;
+        private bool alreadySubscribedToTplEvent = false;
 
         internal delegate void UnobservedExceptionDelegate(ISchedulingContext context, Exception exception);
         
@@ -45,14 +45,6 @@ namespace Orleans
             return true;
         }
 
-        internal void ResetUnobservedExceptionHandler()
-        {
-            lock (lockObject)
-            {
-                unobservedExceptionHandler = null;
-            }
-        }
-
         private void InternalUnobservedTaskExceptionHandler(object sender, UnobservedTaskExceptionEventArgs e)
         {
             var aggrException = e.Exception;
@@ -82,6 +74,16 @@ namespace Orleans
                     logger.Error(ErrorCode.Runtime_Error_100005, errorStr);
                     logger.Error(ErrorCode.Runtime_Error_100006, "Exception remained UnObserved!!! The subsequent behavior depends on the ThrowUnobservedTaskExceptions setting in app config and .NET version.");
                 }
+            }
+        }
+
+        public void Dispose()
+        {
+            lock (lockObject)
+            {
+                TaskScheduler.UnobservedTaskException -= InternalUnobservedTaskExceptionHandler;
+                unobservedExceptionHandler = null;
+                alreadySubscribedToTplEvent = false;
             }
         }
     }
--- a/src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
+++ b/src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
@@ -8,39 +8,26 @@ namespace Orleans
 
     internal class UnobservedExceptionsHandler : IDisposable
     {
-        private readonly Object lockObject = new Object();
         private readonly ILogger logger;
         private UnobservedExceptionDelegate unobservedExceptionHandler;
-        private bool alreadySubscribedToTplEvent = false;
 
         internal delegate void UnobservedExceptionDelegate(ISchedulingContext context, Exception exception);
         
         public UnobservedExceptionsHandler(ILogger<UnobservedExceptionsHandler> logger)
         {
             this.logger = logger;
-            lock (lockObject)
-            {
-                if (!alreadySubscribedToTplEvent)
-                {
-                    TaskScheduler.UnobservedTaskException += InternalUnobservedTaskExceptionHandler;
-                    alreadySubscribedToTplEvent = true;
-                }
-            }
+            TaskScheduler.UnobservedTaskException += InternalUnobservedTaskExceptionHandler;
         }
 
 
         internal bool TrySetUnobservedExceptionHandler(UnobservedExceptionDelegate handler)
         {
             if (handler == null) throw new ArgumentNullException(nameof(handler));
-            lock (lockObject)
+            if (unobservedExceptionHandler != null)
             {
-                if (unobservedExceptionHandler != null)
-                {
-                    return false;
-                }
-
-                unobservedExceptionHandler = handler;
+                return false;
             }
+            unobservedExceptionHandler = handler;
 
             return true;
         }
@@ -79,12 +66,8 @@ namespace Orleans
 
         public void Dispose()
         {
-            lock (lockObject)
-            {
-                TaskScheduler.UnobservedTaskException -= InternalUnobservedTaskExceptionHandler;
-                unobservedExceptionHandler = null;
-                alreadySubscribedToTplEvent = false;
-            }
+            TaskScheduler.UnobservedTaskException -= InternalUnobservedTaskExceptionHandler;
+            unobservedExceptionHandler = null;
         }
     }
 }
--- a/src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
+++ b/src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
@@ -69,7 +69,6 @@ namespace Orleans
         private IPAddress localAddress;
         private IGatewayListProvider gatewayListProvider;
         private readonly ILoggerFactory loggerFactory;
-        private readonly UnobservedExceptionsHandler unobservedExceptionsHandler;
         public SerializationManager SerializationManager { get; set; }
 
         public ActivationAddress CurrentActivationAddress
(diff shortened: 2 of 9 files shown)


src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -145,9 +147,7 @@ internal void ConsumeServices(IServiceProvider services)
 
             try
             {
-                //init logger for UnobservedExceptionsHandlerClass
-                UnobservedExceptionsHandlerClass.InitLogger(this.loggerFactory);
-                if (!UnobservedExceptionsHandlerClass.TrySetUnobservedExceptionHandler(UnhandledException))
+                if (!unobservedExceptionsHandler.TrySetUnobservedExceptionHandler(UnhandledException))
```
jdom: I meant to remove this for a long time. The client should not subscribe to unobserved exceptions, as it is not the owner of the host, it's just a client sdk running in a frontend. Several people already complained about Orleans logging all unrelated unobserved exceptions with our logger.

sergeybykov: +1 The original rational here IIRC was that this would help us investigate customer issue by logging unhandled exceptions. I agree with @jdom that we should remove it. It doesn't add much of real value at the cost of confusing people.

xiazen: sounds good

src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
```diff
@@ -48,15 +45,15 @@ internal static bool TrySetUnobservedExceptionHandler(UnobservedExceptionDelegat
             return true;
         }
 
-        internal static void ResetUnobservedExceptionHandler()
+        internal void ResetUnobservedExceptionHandler()
```
jdom: class should be IDisposable and this should be done on dispose

jdom:
>                     TaskScheduler.UnobservedTaskException += InternalUnobservedTaskExceptionHandler;

Since this class is no longer static, Dispose should unsubscribe to this static method too

---
Refers to: src/Orleans.Core/Async/UnobservedExceptionsHandler.cs:25 in 0fd85d8. [](commit_id = 0fd85d8766e335b73520d4330a398c6ca8ae7977, deletion_comment = False)

src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
```diff
@@ -6,17 +6,18 @@ namespace Orleans
...
+    internal class UnobservedExceptionsHandler : IDisposable
     {
-        private static readonly Object lockObject = new Object();
-        private static ILogger logger;
-        private static UnobservedExceptionDelegate unobservedExceptionHandler;
-        private static readonly bool alreadySubscribedToTplEvent = false;
+        private readonly Object lockObject = new Object();
+        private readonly ILogger logger;
+        private UnobservedExceptionDelegate unobservedExceptionHandler;
+        private bool alreadySubscribedToTplEvent = false;
```
jdom: no reason for this field any more, since it's only subscribed during construction anyway

src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
```diff
@@ -6,17 +6,18 @@ namespace Orleans
...
 
         internal delegate void UnobservedExceptionDelegate(ISchedulingContext context, Exception exception);
         
-        static UnobservedExceptionsHandlerClass()
+        public UnobservedExceptionsHandler(ILogger<UnobservedExceptionsHandler> logger)
         {
+            this.logger = logger;
             lock (lockObject)
```
jdom: no need for locks in the constructor

src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
```diff
@@ -87,5 +76,15 @@ private static void InternalUnobservedTaskExceptionHandler(object sender, Unobse
                 }
             }
         }
+
+        public void Dispose()
+        {
+            lock (lockObject)
```
jdom: just get rid of the lock field entirely, as it's not really protecting anything anymore

jdom:
>                     unobservedExceptionHandler(context, baseException);

Avoid race conditions by avoiding the 2-step access to `unobservedExceptionHandler`.
Just do: `unobservedExceptionHandler?.Invoke(context, baseException);` (without the if preceding it)

---
Refers to: src/Orleans.Core/Async/UnobservedExceptionsHandler.cs:60 in b7aae46. [](commit_id = b7aae4689e92a20345ca67fc2624d6c37cff3191, deletion_comment = False)

src/Orleans.Runtime/Silo/SiloHost.cs
```diff
@@ -114,7 +114,9 @@ public void InitializeOrleansSilo()
         /// </summary>
         public void UnInitializeOrleansSilo()
         {
-            Utils.SafeExecute(UnobservedExceptionsHandlerClass.ResetUnobservedExceptionHandler);
+            var unObservedExceptionHandler = this.orleans?.Services?.GetService<UnobservedExceptionsHandler>();
+            if(unObservedExceptionHandler != null)
+                Utils.SafeExecute(unObservedExceptionHandler.Dispose);
```
jdom: This is not needed, since the Silo itself would dispose it if it was initialized by it.
I guess we can remove this `UnInitializeOrleansSilo` method entirely, but maybe we can preserve in case we need it in the future. TBH, we are probably killing SiloHost soon anyway, so that might not be a compelling argument. I don't really care either way.

xiazen: lol. You comments rhymed. I will keep the UnInitializeOrleansSilo method for backward-compatible reason and maybe we want it in the future.

src/Orleans.Core/Async/UnobservedExceptionsHandler.cs
```diff
@@ -6,57 +6,33 @@ namespace Orleans
...
+            TaskScheduler.UnobservedTaskException += InternalUnobservedTaskExceptionHandler;
         }
 
-        internal static void InitLogger(ILoggerFactory loggerFactory)
-        {
-            logger = loggerFactory.CreateLogger<UnobservedExceptionsHandlerClass>();
-        }
 
-        internal static bool TrySetUnobservedExceptionHandler(UnobservedExceptionDelegate handler)
+        internal bool TrySetUnobservedExceptionHandler(UnobservedExceptionDelegate handler)
```
jdom: Since this class is now only usable by the silo, please avoid this 2-step initialization, and just have the handler log directly, instead of requiring the silo to pass it how to log.
Also, can now move to Orleans.Runtime project

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -66,7 +66,7 @@ public enum SiloType
         private readonly IncomingMessageAgent incomingAgent;
         private readonly IncomingMessageAgent incomingSystemAgent;
         private readonly IncomingMessageAgent incomingPingAgent;
-        private readonly Logger logger;
+        private readonly ILogger logger;
```
xiazen: as a general rule of thumb, also migrated silo.cs off Logger, since we should all try to migrate in our PRs! It's a team effort

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -882,7 +874,7 @@ private void DomainUnobservedExceptionHandler(object context, UnhandledException
         {
             var exception = (Exception)args.ExceptionObject;
             if (context is ISchedulingContext)
-                UnobservedExceptionHandler(context as ISchedulingContext, exception);
+                UnobservedExceptionHandler(context as ISchedulingContext, exception, this.logger);
```
xiazen: Also have to migrate Silo off Logger because I want to make `UnObservedExceptionHandler` static.

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -489,7 +483,7 @@ private void DoStart()
                     providerManagerSystemTarget.SchedulingContext)
                         .WaitWithThrow(initTimeout);
             runtimeClient.CurrentStreamProviderManager = siloStreamProviderManager;
-            if (logger.IsVerbose) { logger.Verbose("Stream provider manager created successfully."); }
+            if (logger.IsEnabled(LogLevel.Debug)) { logger.Debug("Stream provider manager created successfully."); }
```
jdom: I see several of these, which were not introduced by you, but still: if not doing any string interpolation or something like that, then there is no need to check `IsEnabled`, it just adds a lot of noise

xiazen: Those are there also for perf reasons. If logging not set to debug level, then no need to create those logging strings at all.

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -76,7 +76,7 @@ internal static void AddDefaultServices(HostBuilderContext context, IServiceColl
             services.TryAddTransient(typeof(IStreamSubscriptionObserver<>), typeof(StreamSubscriptionObserverProxy<>));
 
             services.TryAddSingleton<ProviderManagerSystemTarget>();
-
+            services.TryAddSingleton<UnobservedExceptionsHandler>(sp => ActivatorUtilities.CreateInstance<UnobservedExceptionsHandler>(sp, Silo.UnobservedExceptionDelegate));
             services.TryAddSingleton<StatisticsProviderManager>();
```
jdom: If nobody resolves this, then it will never be instantiated

src/Orleans.Runtime/Silo/Silo.cs
jdom: That's why I mentioned `if not doing any string interpolation or something like that` :) For constant strings, that optimization trick doesn't apply

xiazen: Ok. Anything else in this PR?

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -862,7 +854,7 @@ private void HandleProcessExit(object sender, EventArgs e)
             FastKill();
         }
 
-        private void UnobservedExceptionHandler(ISchedulingContext context, Exception exception)
+        private static void UnobservedExceptionHandler(ISchedulingContext context, Exception exception, ILogger logger)
```
jdom: does this have to be passed into the handler? can't the handler just own this? and possible just rename the other to SiloUnobservedExceptionsHandler? I would move `DomainUnobservedExceptionHandler` related stuff there too

xiazen: I want to make this method static, so constructing a UnobservedExceptionsHandler class won't rely on constructing a Silo class first. 

As to "move DomainUnobservedExceptionHandler related stuff there too", this PR is about make UnobservedExceptionsHanlder class non static, so I don't think that is inside the scope of that. Let's make the scope stick to the PR's title. We can always refactor silo.cs in another PR.

jdom: I wasn't implying we need to construct the silo first. I was suggesting not to pass any delegates, just have this utility log directly without any reference to Silo. I suggested renaming only because this is no longer used by the client, so the error message that is logged is only relevant when running inside a silo (not a reference to `Silo` class).
Regarding the Domain stuff... sure, it can be separate, but they belong in the same place, so that's why I'm saying that if the handler owns the log messages for unobserved task exceptions, it should also own the logging for the `AppDomain.UnhandledException` event. I find that refactoring very relevant for this PR, but sure, we can postpone fixing this inconsistency.

src/Orleans.Runtime/Utilities/UnobservedExceptionsHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            try
+            {
+                UnobservedExceptionHandler(context, baseException);
+            }
+            finally
+            {
+                if (e.Observed)
+                {
+                    logger.Info(ErrorCode.Runtime_Error_100311, "UnobservedExceptionsHandlerClass caught an UnobservedTaskException which was successfully observed and recovered from. BaseException = {0}. Exception = {1}",
```
jdom: `UnobservedExceptionsHandlerClass` does no longer exist. Just say `Silo` anyway, no need to be explicit about this internal type (even though it will be part of the logger category anyway)

src/Orleans.Runtime/Utilities/UnobservedExceptionsHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+                            baseException.Message, LogFormatter.PrintException(aggrException));
+                }
+                else
+                {
+                    var errorStr = String.Format("UnobservedExceptionsHandlerClass Caught an UnobservedTaskException event sent by {0}. Exception = {1}",
+                            OrleansTaskExtentions.ToString((Task)sender), LogFormatter.PrintException(aggrException));
+                    logger.Error(ErrorCode.Runtime_Error_100005, errorStr);
+                    logger.Error(ErrorCode.Runtime_Error_100006, "Exception remained UnObserved!!! The subsequent behavior depends on the ThrowUnobservedTaskExceptions setting in app config and .NET version.");
```
jdom: @sergeybykov, should we just remove this alarming error altogether? This might have been important before, but the default now is safe, and it would only crash if the developer explicitly turned that knob on because he knew what he was doing

src/Orleans.Runtime/Utilities/UnobservedExceptionsHandler.cs
```diff
@@ -0,0 +1,90 @@
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Orleans.Runtime.Scheduler;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.Runtime
+{
+    internal static class ServiceProviderExtensions
```
benjaminpetit: Maybe you could find a name related to SiloUnobservedExceptionsHandler, because this name is very generic and we might want to use it later in a broader context.

But it is mostly nitpicking here, no need to fix that if you don't want to

src/Orleans.Runtime/Utilities/UnobservedExceptionsHandler.cs
xiazen: I don't know the context on that. But @sergey do want to make configuring UnObservedExceptionHandler optional. I guess we can achieve that by adding an extension method on IServiceCollection or IServiceProvider to op in UnobservedExceptionsHandler. And silo shouldn't turn it on by default. 

regardless, can we do that in another PR? this PR is really just about making UnObservedExceptionHandler non-static. Not making NonObservedExceptionHandler perfect.

xiazen:
hmm two people both approved. someone mind hit the merge button?

src/Orleans.Runtime/Utilities/UnobservedExceptionsHandler.cs
jdom: Sure, we can address later. BTW, I didn't mean remove the entire thing, just the last very alarming error, which is not useful at all.
BTW, let's sync up later, as I don't think it's that critical to make it optional now... it was mainly because it was noise especially in the client, but not in the silo.

PR #3594, opened 2017-10-24
CodeGen: use .NET Core asm for Core & Standard targets regardless of host

ReubenBond:
* If target project is .NET Core or .NET Standard, launch the .dll version of the code generator
  * If MSBuild is a .NET Core dll, use .NET Core assembly to load `GetDotNetHost` MSBuild Task & determine `dotnet` executable location.
  * If MSBuild is a .NET Framework exe, assume `dotnet` is on the `PATH`.
* If target is .NET Framework, use the .exe version of the code generator regardless of MSBuild

### DIFF
--- a/src/Orleans.CodeGeneration.Build/build/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
+++ b/src/Orleans.CodeGeneration.Build/build/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
@@ -1,4 +1,4 @@
-<Project TreatAsLocalProperty="CodeGenDirectory;IsCore;TaskAssembly;OutputFileName;CoreAssembly;FullAssembly;GeneratorAssembly">
+<Project TreatAsLocalProperty="CodeGenDirectory;IsCore;MSBuildIsCore;TargetIsCore;TaskAssembly;OutputFileName;CoreAssembly;FullAssembly;GeneratorAssembly">
   
   <PropertyGroup Condition="'$(OrleansCodeGeneratorAssembly)' != ''">
     <!-- OrleansCodeGeneratorAssembly is used here to override the MSBuildIsCore value during Orleans.sln builds -->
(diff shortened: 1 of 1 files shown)


src/Orleans.CodeGeneration.Build/build/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
```diff
@@ -1,16 +1,39 @@
...
     <CodeGenDirectory Condition="'$(CodeGenDirectory)' == ''">$(ProjectDir)$(IntermediateOutputPath)</CodeGenDirectory>
     <OutputFileName>$(CodeGenDirectory)$(TargetName).orleans.g.cs</OutputFileName>
     <CodeGeneratorEnabled Condition="'$(OrleansCodeGenPrecompile)'!='true' and '$(DesignTimeBuild)' != 'true'">true</CodeGeneratorEnabled>
   </PropertyGroup>
 
-  <UsingTask TaskName="Orleans.CodeGeneration.GetDotNetHost" AssemblyFile="$(TaskAssembly)" Condition="'$(CodeGeneratorEnabled)' == 'true' and '$(IsCore)' == 'true'" />
+  <UsingTask TaskName="Orleans.CodeGeneration.GetDotNetHost" AssemblyFile="$(TaskAssembly)" Condition="'$(CodeGeneratorEnabled)' == 'true' and '$(DotNetHost)' == '' and '$(MSBuildIsCore)' == 'true'" />
 
   <!-- This target is run just before Compile for an Orleans Grain Interface Project -->
   <Target Name="GenerateOrleansCode"
```
attilah: Before the target is executed and we'd use the binaries defined in the variables set above for easier troubleshooting I'd add Exist checks for them. When we're depending on dotnet.exe being in the path, for the exist check it should be treated differently, so we can give better error message if dotnet.exe execution fails. Since we don't know the full path we can defer the error message after execution I think.

attilah:

src/Orleans.CodeGeneration.Build/build/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
attilah: With fresh head I looked into the possible solutions here and since the worst case scenario we can have here is that we're using ```dotnet``` from the path, and that will not be found. But in that case ```Exec``` task will fail anyway, in the other cases we're covered, since either we're resolving ```dotnet``` location or using the freshly built code generator assembly - when building Orleans itself - so I think we're good, **no update needed from my part LGTM**

PR #3595, opened 2017-10-24
Support for non-static serializer

benjaminpetit:
With this PR:

- Orleans generated serializers are no longer statics 
- Developers can implement their own non static serializers. Orleans will create instances with `ActivatorUtilities` that relies on DI. Static serialization methods are still supported.
- Orleans.Core.Abstractions does not depends on `System.Reflection.Emit.Lightweight` anymore
 
A shortcut that I took (and that we may want to fix in the future) is that if `SerializationManager` is instanciated without any `IServiceProvider` (such as in `CodeGenerator`), the serializers will *not* be instanciated. Instead the `SerializationManager` will only populate the data structure needed for `SerializationManager.HasSerializer<T>()` (needed for codegen).


jdom:
>         private static MemberDeclarationSyntax[] GenerateStaticFields(List<FieldInfoMember> fields)

This and the documentation are no longer true, right? BTW, do we still need these fields? Can't they be used in the constructor as variables and then drop them or are they really needed as fields?

---
Refers to: src/Orleans.CodeGeneration/SerializerGenerator.cs:367 in f4b5d73. [](commit_id = f4b5d7315028f6026035f3aeea969b9b420024e9, deletion_comment = False)

src/Orleans.Core/Serialization/SerializationManager.cs
```diff
@@ -428,11 +444,12 @@ object DeserializeGeneric(Type expected, IDeserializationContext context)
                 }
                 else
                 {
+                    var target = this.InitializeSerializer(serializerType);
```
jdom: target is missleading, as `target` usually refers to the target type being serialized, and in this case it is the serializer object itself

src/Orleans.Core/Serialization/SerializationManager.cs
```diff
@@ -222,6 +222,22 @@ public void Register(Type t, DeepCopier cop, Serializer ser, Deserializer deser)
             Register(t, cop, ser, deser, false);
         }
 
+        private object InitializeSerializer(Type serializerType)
+        {
+            if (this.ServiceProvider == null)
+                return null;
```
jdom: Hmmm, this shortcut can be error prone, as there are cases where it is valid to return `null` at runtime...

src/Orleans.Core/Serialization/SerializationManager.cs
```diff
@@ -222,6 +222,22 @@ public void Register(Type t, DeepCopier cop, Serializer ser, Deserializer deser)
...
+
+            if (!serializerType.GetCustomAttributes(typeof(SerializerAttribute), true).Any())
+                return null;
+
+            var constructors = serializerType.GetConstructors();
+            if (constructors == null || constructors.Length < 1)
```
jdom: would default constructors show up in this list even if they aren't explicitly defined in code?

jdom:
As I mentioned in person, I'm not that fond of that shortcut, unless the alternative is too costly. I believe we are using SerializationManager *mainly* as a collection of registered serializers with just a tiny little bit of extra logic in the HasSerializer method. If that's the case, we could remove the dependency on it and just use a HashSet and share that small logic on how to know if there is a serializer defined already.

src/Orleans.Core/Serialization/SerializationManager.cs
```diff
@@ -575,6 +594,19 @@ internal static void GetSerializationMethods(Type type, out MethodInfo copier, o
             }
         }
 
+        private Delegate CreateDelegate(MethodInfo methodInfo, Type delegateType, object target)
```
jdom: delegateType could be a generic type parameter, to avoid all the unnecessary casting on the caller side

benjaminpetit: I previously tried to do that, it wasn't possible to use a generic parameter and cast it in the return statement. It seems that if I use the `as` operator it will do the trick.

https://stackoverflow.com/questions/35321586/cannot-cast-delegate-to-a-generic-type-t

src/Orleans.Core/Serialization/SerializationManager.cs
benjaminpetit: Yes. This will return an empty list only if the constructor is not accessible or the class is static.

src/Orleans.Core/Serialization/SerializationManager.cs
benjaminpetit: I could throw in the GetSerializer/Copier/Deserializer methods if the `ServiceProvider` wasn't set?

src/Orleans.Core/Serialization/SerializationManager.cs
```diff
@@ -222,6 +222,22 @@ public void Register(Type t, DeepCopier cop, Serializer ser, Deserializer deser)
...
+                return null;
+
+            if (!serializerType.GetCustomAttributes(typeof(SerializerAttribute), true).Any())
+                return null;
+
+            var constructors = serializerType.GetConstructors();
+            if (constructors == null || constructors.Length < 1)
+                return null;
+
+            var serializer = ActivatorUtilities.CreateInstance(this.ServiceProvider, serializerType);
```
ReubenBond: Maybe it's out of scope, but does it make sense to use GetServiceOrCreateInstance here instead since we expect 1 instance of each serializer?

src/Orleans.Core/Serialization/SerializationManager.cs
```diff
@@ -558,7 +577,7 @@ internal static void GetSerializationMethods(Type type, out MethodInfo copier, o
             copier = null;
             serializer = null;
             deserializer = null;
-            foreach (var method in type.GetMethods(BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic))
+            foreach (var method in type.GetMethods(BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic))
```
ReubenBond: Probably a case of "who cares", but will this fail at runtime if we have a mix of static & instance methods for copy/serialize/deserialize on a given type?

src/Orleans.Core/Serialization/SerializationManager.cs
```diff
@@ -222,6 +222,22 @@ public void Register(Type t, DeepCopier cop, Serializer ser, Deserializer deser)
             Register(t, cop, ser, deser, false);
         }
 
+        private object InitializeSerializer(Type serializerType)
+        {
+            if (this.ServiceProvider == null)
+                return null;
+
+            if (!serializerType.GetCustomAttributes(typeof(SerializerAttribute), true).Any())
```
ReubenBond: Worth adding a comment here, like:
`// If the type lacks a Serializer attribute then it's a self-serializing type and all serialization methods must be static.`

src/Orleans.CodeGeneration/SerializerGenerator.cs
```diff
@@ -362,49 +424,33 @@ private static MemberDeclarationSyntax[] GenerateStaticFields(List<FieldInfoMemb
...
-                                            getReferenceSetter.Invoke()
-                                                .AddArgumentListArguments(SF.Argument(fieldInfoField)))));
+                        var fieldSetterVariable = SF.VariableDeclarator(field.SetterFieldName);
 
                         result.Add(
                             SF.FieldDeclaration(SF.VariableDeclaration(setterType).AddVariables(fieldSetterVariable))
                                 .AddModifiers(
-                                    SF.Token(SyntaxKind.PrivateKeyword),
-                                    SF.Token(SyntaxKind.StaticKeyword),
-                                    SF.Token(SyntaxKind.ReadOnlyKeyword)));
```
ReubenBond: not readonly?

ReubenBond:
It's worth adding a comment to include what the new generated code looks like, since it's hard to determine without building it.

Left some comments but otherwise LGTM.

src/Orleans.Core/Serialization/SerializationManager.cs
```diff
@@ -173,6 +166,15 @@ internal IRuntimeClient RuntimeClient
             RegisterSerializationProviders(serializatonProviderOptionsValue.SerializationProviders);
         }
 
+        public void SetApplicationPartManager(ApplicationPartManager applicationPartManager)
```
ReubenBond: How does this break the dependency cycle? Is it possible to break it in another way so that SerializationManager doesn't need an initialization step?

jdom: To be honest, logically it does make sense that there is an explicit point in time that we populate the serializer feature from the parts and feature providers, instead of relying on luck in when the container happens to resolve the SM

ReubenBond: It's deterministic: you cannot access the serialization manager until it's created and the registrations are populated on creation. This change makes it less certain. It's not like SM is populating some state which might already be in use by some other class

benjaminpetit: The issue is that, if a serializer ask for a `SerializationManager`in its constructor, and that we instantiate these serializer in the `SerializationManager`.... boom.

src/Orleans.Core/Serialization/SerializationManager.cs
benjaminpetit: To uses that, we would need to register the serializers in the `IServiceProvider`, correct?

I am not sure we want to register all of them, since you may have A LOT of serializers in your app

src/Orleans.CodeGeneration/SerializerGenerator.cs
```diff
@@ -20,6 +20,32 @@ namespace Orleans.CodeGenerator
...
+    //    public void Serializer(global::System.Object untypedInput, global::Orleans.Serialization.ISerializationContext context, global::System.Type expected)
+    //    {
+    //            [...]
+    //    }
+    //    [global::Orleans.CodeGeneration.DeserializerMethodAttribute]
+    //    public global::System.Object Deserializer(global::System.Type expected, global::Orleans.Serialization.IDeserializationContext context)
+    //    {
+    //            [...]
+    //    }
+    //}
```
jdom: You can use XML comments for these comments

src/Orleans.Core/Serialization/SerializationManager.cs
jdom: No, this means that *if* it is registered as a service, it will use that, otherwise falls back to CreateInstance.

src/Orleans.Core/Serialization/SerializationManager.cs
jdom: would `RegisterSerializers` be a more appropriate name? this is not "setting the app manager for later use", it's commanding the SM to register the serializers (from the app manager) *now*

ReubenBond: @benjaminpetit oh right :( Alternatively we could call the method `Initialize` and store the `ApplicationPartManager` in the ctor so that it's more obvious that it needs initialization before use.

jdom: My preference is for calling this method `RegisterSerializers` instead and pass the `ApplicationPartManager` in at this point in time, and is never stored by the SM.

In theory in the future we could call this multiple times (not that we will, nor that we should support it now, but would be applicable logically)

ReubenBond: My intention was that registration occurs exactly once, before usage. As a result, not all of the data structures are protected by locks / concurrent collections.

jdom:
@dotnet-bot test functional

src/Orleans.Core/Core/DefaultClientServices.cs
```diff
@@ -34,6 +34,7 @@ public static void AddDefaultServices(IClientBuilder builder, IServiceCollection
             services.TryAddFromExisting<IInternalGrainFactory, GrainFactory>();
             services.TryAddFromExisting<IGrainReferenceConverter, GrainFactory>();
             services.TryAddSingleton<ClientProviderRuntime>();
+            services.TryAddSingleton<IFieldUtils>(new FieldUtils());
```
jdom: any reason why this is passing an instance instead of just the concrete type, as you did in `DefaultSiloServices`?

src/Orleans.Core/Serialization/SerializationManager.cs
jdom: BTW, are you doing this change or decided not to?

src/Orleans.Core/Serialization/SerializationManager.cs
jdom: It's fine, since we should be removing this statement entirely in https://github.com/dotnet/orleans/pull/3669 now that we'll stop using the SM for the codegen path. /cc @ReubenBond

src/Orleans.Core/Core/DefaultClientServices.cs
benjaminpetit: Nope, will fix that

src/Orleans.Core/Serialization/SerializationManager.cs
jdom: BTW, @ReubenBond, remember to remove this and not allow a null service provider once you rebase your other PR to not use `SerializationManager` for codegen.

PR #3599, opened 2017-10-25
Migrate service fabric extension off Logger

xiazen:


src/Orleans.Membership.ServiceFabric/FabricServiceSiloResolver.cs
```diff
@@ -41,7 +41,7 @@ internal class FabricServiceSiloResolver : IFabricServiceSiloResolver
         {
             this.serviceName = serviceName;
             this.queryManager = queryManager;
-            this.log = new LoggerWrapper<FabricServiceSiloResolver>(loggerFactory);
+            this.log = loggerFactory.CreateLogger<FabricServiceSiloResolver>();
```
jdom: Please don't inject `ILoggerFactory`, and inject `ILogger<FabricServiceSiloResolver>` instead.
You can replace the service registration (in OrleansServiceFabricExtensions) to do the following:
```
services.AddSingleton<IFabricServiceSiloResolver>(sp => ActivatorUtilities.CreateInstance<FabricServiceSiloResolver>(sp, context.ServiceName));
```

This way all DI services are automatically injected without being explicit about every parameter, just the serviceName which is not in the container

PR #3600, opened 2017-10-25
Fix used task name in targets file

attilah:
This change fixes an issue when AspNet.Core Mvc is used in a project together with Orleans Build time code generation.

GetDotNetHost task name is conflicted with their GetDotNetHost task.

[Here](https://github.com/aspnet/MvcPrecompilation/blob/311c52b12efa92f46e19529f0335b8150a28ede3/src/Microsoft.AspNetCore.Mvc.Razor.ViewCompilation/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.ViewCompilation.targets#L33) is the conflicting line.

### DIFF
--- a/src/Orleans.CodeGeneration.Build/build/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
+++ b/src/Orleans.CodeGeneration.Build/build/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
@@ -47,9 +47,9 @@
       <InputAssembly>$(IntermediateOutputPath)$(TargetName)$(TargetExt)</InputAssembly>
       <ArgsFile>$(IntermediateOutputPath)$(TargetName).orleans.g.args.txt</ArgsFile>
     </PropertyGroup>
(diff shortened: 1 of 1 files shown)


ReubenBond:
Should we rename the task itself? Or will this never cause issues in ASP.NET apps?

attilah:
Nope it is good, we just need to use the fully qualified name, that's it. It is a really hard to see error and no one can protect themselves against other developer's task naming it seems!

jdom:
That would solve *our* ambiguous reference to the task, but could it result in ambiguity when the MVC target is trying to use the `GetDotNetHost` task?

ReubenBond:
But what about them, do they use a FQ name everywhere? i.e, can we cause ambiguity in their targets?

jdom:
Right, that's what I meant too. Can we cause ambiguity for them, since we copied the initial code from them?

attilah:
They're not FQ-ing anywhere, that's causing the problem, but it is a MSBuild "hidden feature" that they optimize the taskname to not to be verbose, but this "feature" can cause anyone some unexpected bugs.

We are safe since we're using FQ with the fix and we're in our own namespace.

PR #3601, opened 2017-10-25
Remove internal usage of Logger in AzureUtils

xiazen:
Remove internal usage of Logger in AzureUtils.
There's still usage of Logger in public methods, which is non-trivia. We talked about marking them as obsolete or move them to a legacy package. This can be addressed in another PR.

### DIFF
--- a/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
+++ b/src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
 using Newtonsoft.Json;
@@ -11,6 +12,7 @@ using Orleans.Providers.Azure;
 using Orleans.Runtime;
 using Orleans.Runtime.Configuration;
 using Orleans.Serialization;
+using LogLevel = Microsoft.Extensions.Logging.LogLevel;
 
 namespace Orleans.Storage
 {
@@ -56,6 +58,7 @@ namespace Orleans.Storage
         /// <see cref="IStorageProvider.Log"/>
         public Logger Log { get; private set; }
 
+        private ILogger logger;
         /// <summary> Name of this storage provider instance. </summary>
         /// <see cref="IProvider.Name"/>
         public string Name { get; private set; }
@@ -64,7 +67,9 @@ namespace Orleans.Storage
         /// <see cref="IProvider.Init"/>
         public async Task Init(string name, IProviderRuntime providerRuntime, IProviderConfiguration config)
         {
-            Log = providerRuntime.GetLogger("Storage.AzureBlobStorage");
+            var loggerFactory = providerRuntime.ServiceProvider.GetRequiredService<ILoggerFactory>();
+            this.logger = loggerFactory.CreateLogger<AzureBlobStorage>();
+            Log = new LoggerWrapper(this.GetType().FullName, loggerFactory);
 
             try
             {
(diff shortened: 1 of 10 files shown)


src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
```diff
@@ -64,7 +67,9 @@ public class AzureBlobStorage : IStorageProvider
         /// <see cref="IProvider.Init"/>
         public async Task Init(string name, IProviderRuntime providerRuntime, IProviderConfiguration config)
         {
-            Log = providerRuntime.GetLogger("Storage.AzureBlobStorage");
+            var loggerFactory = providerRuntime.ServiceProvider.GetRequiredService<ILoggerFactory>();
+            this.logger = loggerFactory.CreateLogger<AzureBlobStorage>();
+            Log = new LoggerWrapper(this.GetType().FullName, loggerFactory);
```
jdom: not sure we should preserve this property in the interface, as we'll never be able to move it to Legacy.
If we do preserve it (even if temporarily), just add another constructor to LoggerWrapper that takes an ILogger, because in this particular case you already created an ILogger for it, no need to create a new one.

src/OrleansAzureUtils/Providers/Storage/AzureTableStorage.cs
```diff
@@ -107,9 +109,10 @@ public Task Init(string name, IProviderRuntime providerRuntime, IProviderConfigu
 
             isDeleteStateOnClear = config.Properties.ContainsKey(DeleteOnClearPropertyName) &&
                 "true".Equals(config.Properties[DeleteOnClearPropertyName], StringComparison.OrdinalIgnoreCase);
-
-            Log = providerRuntime.GetLogger("Storage.AzureTableStorage." + id);
-
+            var loggerFactory = providerRuntime.ServiceProvider.GetRequiredService<ILoggerFactory>();
+            var loggerName = this.GetType().FullName + id;
```
jdom: Maybe just suffix it with `-{Name}`. This `id` thing doesn't make much sense. Also, I would do a similar thing in AzureBlobStorage. Regardless of the approach, they should probably do the same thing

jdom:
LGTM other than those 2 comments

src/OrleansAzureUtils/Providers/Storage/AzureBlobStorage.cs
xiazen: I think the Logger in the interface also serves as a LoggerFactory, because of its GetLogger method. LoggerWrapper takes a loggerFactory for that. This PR is "remove internal usage of Logger in AzureUtils", which is meant to be a brain dead logging migration. Whether we should preserve the Logger property in the interface is not in the scope of this PR. we can discuss it in another PR, something like "remove/obsolete Logger in IStoerageProvider" interface.

jdom: Ok, sure, but you can still have an overload that takes an ILogger AND ILoggerFactory. What I was trying to avoid there is to create yet a new instead of ILogger inside the wrapper, when it's not needed.

jdom:
>         private readonly int id;

No need to keep this ID any more I believe

---
Refers to: src/OrleansAzureUtils/Providers/Storage/AzureTableStorage.cs:61 in 932a51e. [](commit_id = 932a51ec246f7419c9fa012154ccdc65c449a21b, deletion_comment = False)

src/OrleansAzureUtils/Providers/Storage/AzureTableStorage.cs
jdom: BTW, you partially addressed this. We would still need to do the same thing for AzureBlobStorage

xiazen: I did it for AzureBlobStorage

jdom: You are right, a full refresh in CodeFlow revealed it, sorry 'bout that :P

PR #3602, opened 2017-10-25
Add cross platform Solution file without the netfx and test project

attilah:
With the removal of the .Net Framework and Test projects this solution is buildable on non-windows systems by executing: ```dotnet build OrleansCrossPlatform.sln``` and if NuGet Packages are required: ```dotnet pack OrleansCrossPlatform.sln``` can be used.


jdom:
Are you going to add this to the groovy file so that it gets built on CI in linux?

galvesribeiro:
I think that is a good idea @jdom. Even without tests, I believe at least build it on *NIX systems would be good.

attilah:
@dotnet-bot test ci please

attilah:
@dotnet-bot test ci please

attilah:
@dotnet-bot test ci please

attilah:
@dotnet-bot test ci please

run-build.sh
```diff
@@ -0,0 +1,183 @@
...
+[ -z "$DOTNET_INSTALL_DIR_PJ" ] && export DOTNET_INSTALL_DIR_PJ=$REPOROOT/.dotnet_stage0PJ/$ARCHITECTURE
+[ -d "$DOTNET_INSTALL_DIR_PJ" ] || mkdir -p $DOTNET_INSTALL_DIR_PJ
+
+export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
+
+# Don't resolve shared frameworks from user or global locations
+export DOTNET_MULTILEVEL_LOOKUP=0
+
+# Install a stage 0
+(set -x ; "$REPOROOT/scripts/obtain/dotnet-install.sh" --channel "release/2.0.2" --install-dir "$DOTNET_INSTALL_DIR" --architecture "$ARCHITECTURE" $LINUX_PORTABLE_INSTALL_ARGS)
```
jdom: I understand you copied all this from the dotnet cli repo, and if it *just works*, then it's OK as a starting point (EDIT: now looking at the amount of scripts you added, I'm not that comfortable either, I thought it was just this script). The reality is that they need to be able to build in several different environments, that's why the build scripts and dockerfiles are so complex and install everything from scratch.
Our complexity is orders of magnitude less, our build could just be to run a container with an already set up image such as this one https://hub.docker.com/r/microsoft/dotnet/ and call `dotnet build` from it, reducing these amount of scripts to less than 10 lines (probable less than 3 lines)

jdom: And of course that has the added benefit that we talked about, which is that the first time pulling the image might take long, but subsequent would be immediate as they use the cached image. Where as these scripts install everything from scratch every single time, regardless of cache

attilah:
@dotnet-bot test ci please

attilah:
@dotnet-bot test ci please

attilah:
@dotnet-bot test ci please

attilah:
@dotnet-bot test ci please

attilah:
@dotnet-bot test ci please

jdom:
@dotnet-bot test ci
@dotnet-bot test functional

netci.groovy
```diff
@@ -5,36 +5,60 @@ import jobs.generation.Utilities;
...
-        if (testCategory == 'functional') { testScript = "TestAll.cmd" }
+def platformList = ['Linux', 'Windows_NT']
 
-        def newJob = job(Utilities.getFullJobName(project, newJobName, isPR)) {
+platformList.each { platform ->
+    if (platform == 'Linux') {
+        def os = 'Linux'
+        def architecture = 'x64'
+        def configuration = 'Release'
+        def isPR = true
```
jdom: we should probably add the Linux job for both PR and non-PR

src/Orleans.CodeGeneration.Build/Orleans.CodeGeneration.Build.csproj
```diff
@@ -9,6 +9,7 @@
   <PropertyGroup>
     <TargetFramework />
     <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.0</TargetFrameworks>
```
jdom: will this `OS` property be set on local builds or is this a CI server-specific value?

src/Orleans.CodeGeneration.Build/build/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
```diff
@@ -1,31 +1,34 @@
 <Project TreatAsLocalProperty="CodeGenDirectory;IsCore;MSBuildIsCore;TargetIsCore;TaskAssembly;OutputFileName;CoreAssembly;FullAssembly;GeneratorAssembly">
-  
+
   <PropertyGroup Condition="'$(OrleansCodeGeneratorAssembly)' != ''">
     <!-- OrleansCodeGeneratorAssembly is used here to override the MSBuildIsCore value during Orleans.sln builds -->
     <MSBuildIsCore></MSBuildIsCore>
     <TargetIsCore></TargetIsCore>
+    <!-- For non-windows OS we force .Net Core  -->
+    <MSBuildIsCore Condition="'$(OS)' != 'Windows_NT'">true</MSBuildIsCore>
+    <TargetIsCore Condition="'$(OS)' != 'Windows_NT'">true</TargetIsCore>
```
jdom: These default values seem to contradict what @ReubenBond intended on the lines immediately above them... Is it correct?

src/Orleans.CodeGeneration.Build/Orleans.CodeGeneration.Build.csproj
attilah: I think that's coming from MSBuild itself, default SDK targets/props also using this property to check for OS.

src/Orleans.CodeGeneration.Build/build/Microsoft.Orleans.OrleansCodeGenerator.Build.targets
attilah: In case of non-windows we've no options so we set them explicitly, so later checks see these as non-empty values.

netci.groovy
attilah: Updated.

netci.groovy
```diff
@@ -9,22 +9,23 @@ def platformList = ['Linux', 'Windows_NT']
...
+                }
             }
-        }
 
-        Utilities.setMachineAffinity(newJob, osUsedForMachineAffinity, 'latest-or-auto')
-        Utilities.standardJobSetup(newJob, project, isPR, "*/${branch}")
-        Utilities.addGithubPRTriggerForBranch(newJob, branch, "${os} ${architecture} ${configuration} Build")
+            Utilities.setMachineAffinity(newJob, osUsedForMachineAffinity, 'latest-or-auto')
+            Utilities.standardJobSetup(newJob, project, isPR, "*/${branch}")
+            Utilities.addGithubPRTriggerForBranch(newJob, branch, "${os} ${architecture} ${configuration} Build")
```
jdom: hmmm, shouldn't this be added only if isPR is true?

attilah:
@dotnet-bot test ci

jdom:
@dotnet-bot test bvt

PR #3611, opened 2017-10-27
Remove internal usage of Logger from AWsUtils

xiazen:
braindead PR to remove internal usage of Orleans.Runtime.Logger from AwsUtils, except for usage of Logger in `IStorageProvider` interface, which can be addressed in another PR.

### DIFF
--- a/src/OrleansAWSUtils/Storage/Provider/DynamoDBStorageProvider.cs
+++ b/src/OrleansAWSUtils/Storage/Provider/DynamoDBStorageProvider.cs
@@ -14,6 +14,7 @@ using System.IO;
 using Microsoft.Extensions.DependencyInjection;
 using OrleansAWSUtils.Storage;
 using Microsoft.Extensions.Logging;
+using Orleans.Runtime.Configuration;
 
 namespace Orleans.Storage
 {
@@ -61,6 +62,7 @@ namespace Orleans.Storage
         /// </summary>
         public Logger Log { get; private set; }
 
+        private ILogger logger;
         private DynamoDBStorage storage;
         private SerializationManager serializationManager;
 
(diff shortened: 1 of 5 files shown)


src/OrleansAWSUtils/Storage/Provider/DynamoDBStorageProvider.cs
```diff
@@ -86,9 +88,10 @@ public Task Init(string name, IProviderRuntime providerRuntime, IProviderConfigu
 
             isDeleteStateOnClear = config.Properties.ContainsKey(DELETE_ON_CLEAR_PROPERTY_NAME) &&
                 "true".Equals(config.Properties[DELETE_ON_CLEAR_PROPERTY_NAME], StringComparison.OrdinalIgnoreCase);
-
-            Log = providerRuntime.GetLogger("Storage.AWSDynamoDBStorage." + id);
-
+            var loggerFactory = providerRuntime.ServiceProvider.GetRequiredService<ILoggerFactory>();
+            var loggerName = $"{this.GetType().FullName}.{name}";
+            Log = new LoggerWrapper(loggerName, loggerFactory);
```
jdom: similar comment that remained unaddressed before. Let's pass the logger (and loggerFactory) in the LoggerWrapper. No need to create an extra logger for it, as the infrastructure will optimize a few things for us

jdom:
>         private readonly int id;

Similarly, this no longer used. Let's remove it.
Interestingly, I noticed that it was incorrectly used, as it was not even a static field, so the id would always be 1. The same bug was in AzureTableStorage

---
Refers to: src/OrleansAWSUtils/Storage/Provider/DynamoDBStorageProvider.cs:49 in 4494811. [](commit_id = 4494811878e451e5908eb3332cd477996b7d9ac7, deletion_comment = False)

test/AWSUtils.Tests/StorageTests/PersistenceGrainTests_AWSDynamoDBStore.cs
```diff
@@ -211,5 +214,15 @@ public void AWSDynamoDBStore_ConvertToFromStorageFormat_GrainReference_List()
...
             Assert.Equal(initialState.Grain, convertedState.Grain);  // "Grain"
         }
+
+        private async Task<DynamoDBStorageProvider> InitDynamoDBTableStorageProvider(IProviderRuntime runtime, string storageName)
+        {
+            Dictionary<string, string> providerCfgProps = new Dictionary<string, string>();
+            var store = new DynamoDBStorageProvider();
+            providerCfgProps["DataConnectionString"] = TestDefaultConfiguration.DataConnectionString;
+            var cfg = new ProviderConfiguration(providerCfgProps, null);
+            await store.Init(storageName, runtime, cfg);
```
jdom: have you tested this change? because we don't run this in CI

xiazen: Yes! tested against real dynamoDB

PR #3623, opened 2017-10-30
Add simple docker sample that will be used in the documentation

benjaminpetit:


Samples/Docker-Simple/Build.cmd
```diff
@@ -0,0 +1,51 @@
+@echo off
+
+echo [7mTesting prerequisites[0m
```
jdom: I learned something, and looks cool with color, but at the same time, I think it would be preferable to just have PowerShell scripts (easier to maintain, easier to port to bash if needed (although PowerShell will be cross plat soon)).
Fine to ignore comment if you are doing something that would take some time to port to PowerShell

Samples/Docker-Simple/README.md
```diff
@@ -0,0 +1,28 @@
+# Prerequisites
+
+You need to have installed:
+ - dotnet core sdk
+ - docker engine and docker-compose
```
jdom: Links for these would be useful

Samples/Docker-Simple/Client/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.IO;
+using System.Threading;
+
+namespace Client
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            var connectionString = File.ReadAllText("connection-string.txt");
```
jdom: I would recommend leveraging the Microsoft.Extensions.Configuration system to read either from an INI or json file. We are going to be endorsing this more and more as we get rid of the legacy XML configuration.

Samples/Docker-Simple/Build.cmd
ReubenBond: image
Are these color markers or some weird characters?

Samples/Docker-Simple/Client/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+                GatewayProvider = ClientConfiguration.GatewayProviderType.AzureTable,
+                DataConnectionString = connectionString
+            };
+
+            var client = new ClientBuilder()
+                .AddApplicationPartsFromBasePath()
+                .UseConfiguration(config)
+                .Build();
+
+            client.Connect().Wait();
```
jdom: just move all of this to an async method and have Main just call the async method by doing:
```
ExecuteAsync().GetAwaiter().GetResult();
```
Then don't use `.Wait()` nor `.Result` to block. This way exceptions and so on are displayed correctly

Samples/Docker-Simple/Build.cmd
benjaminpetit: Yes I tried to be a bit fancy here. This will only work with Windows 10. https://gist.github.com/mlocati/fdabcaeb8071d5c75a2d51712db24011#file-win10colors-cmd

Not sure why I didn't wrote this using powershell...

Samples/Docker-Simple/Silo/Program.cs
```diff
@@ -0,0 +1,60 @@
...
+                .UseConfiguration(config)
+                .ConfigureLogging(builder => builder.SetMinimumLevel(LogLevel.Warning).AddConsole())
+                .Build();
+
+            Task.Run(StartSilo);
+
+            AppDomain.CurrentDomain.ProcessExit += (object sender, EventArgs e) =>
+            {
+                Console.WriteLine("ProcessExit fired");
+                Task.Run(StopSilo);
```
jdom: Is this correct? If you don't block ProcessExit, I believe it will close immediately, right?

benjaminpetit: You may be right. I will just use a sync call here. 

And use `AssemblyLoadContext.Default.Unloading`, I put previsouly ProcessExit earlier for test purpose

Samples/Docker-Simple/Client/Program.cs
benjaminpetit: Yes I know, but this is a simple sample, I find it easier to just drop the content of your connection string in a file than replacing a string in a previously existing json file.

Samples/Docker-Simple/Silo/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Runtime.Loader;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Silo
+{
+    class Program
+    {
+        private static ISiloHost silo;
+        private static readonly AutoResetEvent closing = new AutoResetEvent(false);
```
jdom: Since you are waiting on several places for this to be signaled, you probably want to use a `ManualResetEvent` instead of Auto...
Also, probably a more intuitive name would be `siloStopped`, as this is signaled after the silo is stopped. Whether as closing is ambiguous: is when the silo is closing, when the program is closing, etc

PR #3627, opened 2017-11-01
Remove internal usage of Logger in extension projects and remove Initialize(Logger) from IExternalSerializer

xiazen:
- Remove internal usage of Logger in extension projects
- Remove `Initialize(Logger logger)` method from `IExternalSerializer` interface

### DIFF
--- a/src/Orleans.Core/Serialization/IExternalSerializer.cs
+++ b/src/Orleans.Core/Serialization/IExternalSerializer.cs
@@ -1,5 +1,4 @@
 using System;
-using Orleans.Runtime;
 
 namespace Orleans.Serialization
 {
@@ -12,12 +11,6 @@ namespace Orleans.Serialization
     /// </summary>
     public interface IExternalSerializer
     {
-        /// <summary>
-        /// Initializes the external serializer. Called once when the serialization manager creates 
-        /// an instance of this type
-        /// </summary>
-        void Initialize(Logger logger);
-
         /// <summary>
         /// Informs the serialization manager whether this serializer supports the type for serialization.
         /// </summary>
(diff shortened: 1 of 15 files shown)


src/Orleans.Core/Serialization/IExternalSerializer.cs
```diff
@@ -13,12 +12,6 @@ namespace Orleans.Serialization
     public interface IExternalSerializer
```
ReubenBond: This is a breaking change, we need to make sure that it's called out in the release notes.

xiazen: @jdom to notify

src/OrleansSQLUtils/Storage/Provider/AdoNetStorageProvider.cs
```diff
@@ -200,10 +202,12 @@ public class AdoNetStorageProvider: IStorageProvider
                 queries.Single(i => i.Item1 == "ReadFromStorageKey").Item2,
                 queries.Single(i => i.Item1 == "ClearStorageKey").Item2);
 
-            Log = providerRuntime.GetLogger(GetType().FullName);
+            var loggerFactory = providerRuntime.ServiceProvider.GetService<ILoggerFactory>();
+            this.logger = loggerFactory.CreateLogger<AdoNetStorageProvider>();
+            Log = new LoggerWrapper(this.logger, this.GetType().FullName, loggerFactory);
```
ReubenBond: Do we need the wrapper here?

You should be able to get `ILogger<AdoNetStorageProvider>` from the container directly, with no factory.

xiazen: that is for IStorageProvider.Logger method. There's discussion about moving that to a legacy package, or even remove it. But I don't want to involve that in this PR yet.

xiazen: This is the same with what I did in #3611

PR #3628, opened 2017-11-01
Remove internal usage of Logger in non-core projects

xiazen:


src/Orleans.Core/CodeGeneration/TypeUtils.cs
```diff
@@ -548,7 +549,34 @@ public static IEnumerable<Type> GetTypes(Assembly assembly, Predicate<Type> wher
...
+                    return typeLoadException.Types?.Where(type => type != null).Select(type => type.GetTypeInfo()) ??
+                           Enumerable.Empty<TypeInfo>();
+                }
+
+                return Enumerable.Empty<TypeInfo>();
+            }
+        }
+
+        //TODO: delete this one after runtime migrate off Logger
+        public static IEnumerable<TypeInfo> GetDefinedTypes(Assembly assembly, Logger logger = null)
```
jdom: why have this duplicate code, even if temporarily, just for the logger? Convert from one to the other and that's it, but we are still doing changes in these methods for 2.0.0 final, and it's easy to miss this unnecessary duplication

src/Orleans.EventSourcing/LogStorage/LogConsistencyProvider.cs
```diff
@@ -38,15 +41,6 @@ public bool UsesStorageProvider
             }
         }
 
-        private static int counter; // used for constructing a unique id
-        private int id;
-
-        /// <inheritdoc/>
-        protected virtual string GetLoggerName()
-        {
-            return string.Format("LogViews.{0}.{1}", GetType().Name, id);
```
jdom: If you are changing this, there were some tests filtering on the LogViews prefix when configuring logging, that should be updated

src/Orleans.Core/CodeGeneration/TypeUtils.cs
xiazen: that might bring much more changes , and I don't know how much more, since I need to migrate classes using GetDefinedTypes too and so on and so forth. I think the better approach is migrate classes using GetDefinedTypes (Logger) to GetDefinedTypes(ILogger). Then I can just delete GetDefinedTypes(Logger) at the end. 

I have gone on the route of what you suggested before, and that just end up with a spider web spread from the utility method using Logger to so much more. I don't want to go into the route again.

jdom:
@dotnet-bot test functional

PR #3634, opened 2017-11-02
Align better with the generic Microsoft.Extensions.Hosting.HostBuilder

jdom:
* Separate Orleans-specific concerns from the pseudo-generic SiloHostBuilder:
  * Configure Host separate from Application: allows configuring host information that can drive the rest of the application configuration. Will be very useful without the legacy configuration object.
  * Separate Orleans-specific service registration from the generic builder
* Add tests for SiloHostBuilder: The tests were copied from the Microsoft.Extensions.Hosting solution, to make sure our temporary implementation aligns with theirs.
* Some small behavior in the builder was tweaked to align with the generic one.


jdom:
I was forced to add Host configuration separate from the Application one, in preparation to be able to configure the host builder in our tests without coupling to the legacy configuration

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
```diff
@@ -16,6 +18,32 @@ public static class CoreHostingExtensions
         /// <param name="builder">The host builder.</param>
         /// <param name="siloName">The silo name.</param>
         /// <returns>The silo builder.</returns>
+        public static ISiloHostBuilder ConfigureOrleans(this ISiloHostBuilder builder)
+        {
+            // Configure the container to use Orleans, including the default silo name & services.
+            builder.ConfigureServices((context, services) =>
+            {
+                if (!context.Properties.ContainsKey("OrleansServicesAdded"))
```
ReubenBond: Is this pattern also used in ASP.NET?

jdom: Not really, this is because until we move to the generic HostBuilder, I wanted to have a way to opt-out from adding the services, so that I can test the SiloHostBuilder in isolation from Orleans, to verify it aligns with the other one.
In the generic host builder, if the user doesn't explicitly call `ConfigureWebHost()`, the host builder doesn't include anything ASP.NET related

PR #3638, opened 2017-11-06
Service Fabric: update to use SiloHostBuilder

ReubenBond:
Brings a much needed update to Service Fabric support, primarily to support `SiloHostBuilder` and `ClientBuilder` and separate hosting from membership.

* Splits the `Microsoft.Orleans.ServiceFabric` into `Microsoft.Orleans.Hosting.ServiceFabric` and `Microsoft.Orleans.Membership.ServiceFabric`. Membership depends on hosting.
* The `Microsoft.Orleans.ServiceFabric` package will no longer be created.

Updated samples are in a separate branch, will open PR when we bump the version.
Example usage:
```C#
protected override IEnumerable<ServiceInstanceListener> CreateServiceInstanceListeners()
{
    // Listeners can be opened and closed multiple times over the lifetime of a service instance.
    // A new Orleans silo will be both created and initialized each time the listener is opened and will be shutdown 
    // when the listener is closed.
    var listener = OrleansServiceListener.CreateStateless(
        (serviceContext, builder) =>
        {
            // Use Service Fabric for cluster membership.
            builder.AddServiceFabricMembership(serviceContext);

            // Optional: configure logging.
            builder.ConfigureLogging(logging => logging.AddDebug());

            config.Globals.RegisterBootstrapProvider<BootstrapProvider>("poke_grains");
            config.Globals.ReminderServiceType = GlobalConfiguration.ReminderServiceProviderType.ReminderTableGrain;

            // Service Fabric manages port allocations, so update the configuration using those ports.
            config.Defaults.ConfigureServiceFabricSiloEndpoints(serviceContext);

            // Tell Orleans to use this configuration.
            builder.UseConfiguration(config);

            // Add your application assemblies.
            builder.AddApplicationPart(typeof(ICalculatorGrain).Assembly);

            // Alternative: add all loadable assemblies in the current base path (see AppDomain.BaseDirectory).
            builder.AddApplicationPartsFromBasePath();
        });

    return new[] { listener };
}
```
xref: #1059


src/Orleans.Hosting.ServiceFabric/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.ObjectModel;
+using System.Fabric;
+using System.Fabric.Description;
+using System.Net;
+using Orleans.ServiceFabric;
+using NodeConfiguration = Orleans.Runtime.Configuration.NodeConfiguration;
+
+namespace Orleans.Hosting
+{
+    public static class ConfigurationExtensions
```
jdom: Please use a class name that has `ServiceFabric` in it. Especially since this goes to the default Orleans.Hosting namespace where collisions are likely to occur

ReubenBond: good call, will fix

src/Orleans.Hosting.ServiceFabric/FabricSiloInfo.cs
```diff
@@ -11,24 +12,6 @@
     public class FabricSiloInfo
```
jdom: I see that this Json-serializable class is also use with IOptions. Should they really be the same thing? Haven't reviewed the whole thing to understand what's the JSON representation used for

ReubenBond: The JSON representation is used for Service Fabric's naming service.

The IOptions portion is mostly so that external developers could publish their own endpoints via the `OtherEndpoints` property.

src/Orleans.Hosting.ServiceFabric/OrleansCommunicationListener.cs
```diff
@@ -0,0 +1,117 @@
...
+        public OrleansCommunicationListener(Action<ISiloHostBuilder> configure)
+        {
+            this.configure = configure ?? throw new ArgumentNullException(nameof(configure));
+        }
+
+        /// <summary>
+        /// Gets or sets the underlying <see cref="ISiloHost"/>.
+        /// </summary>
+        /// <remarks>Only valid after <see cref="OpenAsync"/> has been invoked. Exposed for testability.</remarks>
+        public ISiloHost SiloHost { get; private set; }
```
jdom: I would just call this property Host, so that we can hopefully then switch the type to use the generic IHost

src/Orleans.Hosting.ServiceFabric/FabricSiloInfo.cs
ReubenBond: BTW, originally I thought of having two, but I figured it was unnecessary/over-engineering

jdom: I'm not really convinced, but don't feel that strongly either, especially since we haven't full migrated off the legacy configuration and into Options. I think we can re-asses afterwards

ReubenBond: What's your preference for now? Change the PR or leave it?

jdom: nah, leaving it for now it's OK. I was just waiting for the build to finish

src/Orleans.Membership.ServiceFabric/Orleans.Membership.ServiceFabric.csproj
```diff
@@ -1,15 +1,15 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <PackageId>Microsoft.Orleans.ServiceFabric</PackageId>
```
jdom: We should probably create the `Microsoft.Orleans.ServiceFabric` meta-package. Do we do it now or in a different PR? I suppose it's better to do it in the same PR where we split the package

ReubenBond: Oh originally I had the metapackage, but I figured it was just noise. I can add it again.

jdom: Yeah, I think we should do that

jdom:
@dotnet-bot test functional

PR #3639, opened 2017-11-06
Fixing codegen when a type's name contains comma

sunnycase:
While Roslyn may generate some types that their names may contain ','.
For example, when implement an interface which has more than one generic parameters and the method is async.
When orleans codegen generating code, it produces some code like:
```c#
feature.AddKnownType("UnitTests.Grains.AFactory+<UnitTests-Grains-IFactory<System-Int32\,System-Double>-Product>d__0,TestGrains", "UnitTests.Grains.<UnitTests-Grains-IFactory<System-Int32\,System-Double>-Product>d__0");
```
The `\,` in the string literals will cause compiler errors. (unrecognized escape sequence)
Convert them to verbatim string literals will solve the issue.
```c#
feature.AddKnownType(@"UnitTests.Grains.AFactory+<UnitTests-Grains-IFactory<System-Int32\,System-Double>-Product>d__0,TestGrains", @"UnitTests.Grains.<UnitTests-Grains-IFactory<System-Int32\,System-Double>-Product>d__0");
```

### DIFF
--- a/src/Orleans.CodeGeneration/Utilities/StringExtensions.cs
+++ b/src/Orleans.CodeGeneration/Utilities/StringExtensions.cs
@@ -6,6 +6,7 @@ namespace Orleans.CodeGenerator.Utilities
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using System.IO;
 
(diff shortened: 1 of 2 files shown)


ReubenBond:
Thank you, @sunnycase! I'll create a follow-up PR to additionally stop the code generator from writing those unnecessary lines.

PR #3643, opened 2017-11-07
CodeGen: do not generate KnownTypes for nested private or compiler-generated classes

ReubenBond:
Follow-on from #3639.

These classes either can't or shouldn't be referenced, so we should not generate code referring to them.


ReubenBond:
@dotnet-bot test functional

PR #3644, opened 2017-11-07
Throw exception during Build() when user does not provide application assemblies

ReubenBond:
* Adds `IConfigurationValidator` interface which is used during client/silo build to validate the system before returning from `Build()`.
* Add `ApplicationPartValidator` implementation which checks that at least one non-framework assembly has been added and that codegen has run on at least one non-framework assembly.
* Tests for client/silo builders


src/Orleans.Core/Core/ClientBuilder.cs
```diff
@@ -28,6 +19,12 @@ public class ClientBuilder : IClientBuilder
         /// <inheritdoc />
         public IDictionary<object, object> Properties { get; } = new Dictionary<object, object>();
 
+        /// <summary>
+        /// Returns a new default client builder.
+        /// </summary>
+        /// <returns>A new default client builder.</returns>
+        public static IClientBuilder CreateDefault() => new ClientBuilder().ConfigureDefaults();
```
jdom: Not sure I understand the value of this static method here and in SiloHostBuilder

ReubenBond: The reason for it is that we don't have some kind of PostConfigure method (that I can see?), so the container is empty when the user's configure delegates execute.

Aside from that, I think it's better that we have a separate default builder to add some friendly defaults, just like ASP.NET's `WebHost.CreateDefaultBuilder(args)` method. The idea is that we can add configuration which users *probably* want, but not necessarily.

After that, it might be better to make `new ClientBuilder()` and `new SiloHostBuilder()` give a bare-bones implementation which doesn't add those defaults.

`ClientBuilder.CreateDefault()` -> happy path
`new ClientBuilder()` -> bare bones

jdom: Ah, ok, that would explain it, it's just the the current implementation still calls `ConfigureOrleans()` within `Build()`, so that didn't make much sense.

src/Orleans.Runtime/Hosting/Generic/SiloHostBuilder.cs
```diff
@@ -33,14 +44,16 @@ public ISiloHost Build()
...
-            
+
             BuildHostConfiguration();
             CreateHostingEnvironment();
             CreateHostBuilderContext();
             BuildAppConfiguration();
 
             var serviceProvider = CreateServiceProvider();
 
+            ValidateSystemConfiguration(serviceProvider);
```
jdom: When using the generic HostBuilder, would this just be an `IHostedService` that throws in its constructor (instead of StartAsync) or how would it map to an exception during Build?

src/Orleans.Core/Core/ClientBuilder.cs
ReubenBond: I didn't want to break the current implementation, but maybe we should? Need to see what ASP.NET Core does

Possibly we should create a different static type and call the method `DefaultClienBuilder.Create()`. On the other hand, that would be less discoverable.

src/Orleans.Runtime/Hosting/Generic/SiloHostBuilder.cs
ReubenBond: I'm not sure what the best approach is. We need some kind of lifecycle hook that we can put this and similar procedures in.

PR #3645, opened 2017-11-07
CodeGen: Remove warning for System types.

ReubenBond:


ReubenBond:
@dotnet-bot test functional

src/Orleans.CodeGeneration/SerializerGenerationManager.cs
```diff
@@ -101,9 +101,6 @@ internal bool RecordTypeToGenerate(Type t, Assembly targetAssembly)
 
             if (typeInfo.Namespace != null && (typeInfo.Namespace.Equals("System") || typeInfo.Namespace.StartsWith("System.")))
             {
-                var message = "System type " + t.Name + " may require a custom serializer for optimal performance. "
-                              + "If you use arguments of this type a lot, consider submitting a pull request to https://github.com/dotnet/orleans/ to add a custom serializer for it.";
-                log.Warn(ErrorCode.CodeGenSystemTypeRequiresSerializer, message);
```
jdom: Is this warning completely worthless or should we keep it at least as Debug level, which is not typically output to by end-users?

ReubenBond: I think it's useless because it's not actionable. What would be better is just introspection so users can ask 'what types have serializers and which types use fallback', or 'what serializer will this type use'

jdom: Well, that's why I meant Debug, because it is mostly actionable for us. Still users could in theory create custom serializers for system types in their assemblies, no?

ReubenBond: No worries, I've changed it to Verbose

PR #3653, opened 2017-11-08
Simple account transfer transaction example

jason-bragg:


jason-bragg:
I like addding random extra letters :)
Will fix.

PR #3656, opened 2017-11-09
Add logging to grain in HelloWorld.NetCore

jokin:
I would have find helpful if the new way of logging in the 2.0 version was included in some of the samples.

### DIFF
--- a/Samples/HelloWorld.NetCore/src/HelloWorld.Grains/HelloGrain.cs
+++ b/Samples/HelloWorld.NetCore/src/HelloWorld.Grains/HelloGrain.cs
@@ -1,5 +1,6 @@
 using HelloWorld.Interfaces;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 
 namespace HelloWorld.Grains
 {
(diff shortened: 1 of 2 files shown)


jokin:
The Ci build are timeout erros, is there a way to run again without pushing another change?

jdom:
@dotnet-bot test this please
Reason: unrelated test failure

Samples/HelloWorld.NetCore/src/HelloWorld.Grains/HelloGrain.cs
```diff
@@ -8,8 +9,16 @@ namespace HelloWorld.Grains
...
+        private readonly ILogger logger;
+
+        public HelloGrain(ILogger<HelloGrain> logger)
+        {
+            this.logger = logger;
+        }  
+
         Task<string> IHello.SayHello(string greeting)
         {
+            logger.LogWarning($"SayHello message received: greeting = '{greeting}'");
```
jdom: I know it's nitpicking, but for a new-comer, receiving a warning might be surprising, they might think something went wrong. Could you change to informational?

jokin: Done, I started with Info level but changed to warning to standout between all the messages, but it definitely should be and information level message. Thanks.

jdom:
Thank you @jokin !

PR #3662, opened 2017-11-10
Add more debug level logging to transaction abort and prepare failure cases

xiazen:


src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -60,7 +60,7 @@ public TransactionalState(ITransactionalStateConfiguration transactionalStateCon
             this.copier = copier;
             this.transactionAgent = transactionAgent;
             this.runtime = runtime;
-            this.logger = loggerFactory.CreateLogger($"{this.context.GrainIdentity}+{this.config.StateName}");
+            this.logger = loggerFactory.CreateLogger($"{this.GetType().FullName}.{this.context.GrainIdentity}.{this.config.StateName}");
```
jason-bragg: not sure if we want to prepend this by the facet type name, for filtering I think the grain type name, which I believe is in the grian Id is better.

xiazen: hmm what if the grain has multiple facet?

jason-bragg: each will have a unique statename

xiazen: ohh I see what you mean. I misunderstood for a sec. 
I don't have strong opinion on this. Just throw my idea here. Which category to filter on depends on what information user want to investigate on. For example, if user want to investigate on transaction stuff, then he will set a filter on transaction namespace Orleans.Transaction, in which case the facet type name will be useful here. If user want to investigate on the specific grain, then he will filter on the grain's namespace. 

Since this class is transaction related. So I think filtering on this class will mostly be related to transaction investigation. Hence I think we should add the facet type name. If you think it is too long, I can make it just its namespace .

jason-bragg: Fair enough.  I suspected that users would be more interested in filtering on their grains, but yeah, it's not clear.

Issue #3498, opened 2017-10-04
Enhancing threading model


dVakulen:
Based on research done in https://github.com/dotnet/orleans/pull/2060 which is currently in some unreadable state, but have shown performance  improvements of [up to 80%](https://github.com/dotnet/orleans/pull/2060#issuecomment-297853999).

**Short overview  of current design:**
 - Orleans uses SEDA pattern with `BlockingCollection`s as work items storage. 
 - Main contention and CPU burning points are located in descendants of `AsynchQueueAgent` and in `WorkQueue`. They caused by usage of `BlockingCollection`'s, specifically the behavior of spinning for a bit on dequeue when there's no items, which happens quite often under low to medium load. This also is source of increased thread switching with related overhead. 
 - Descendants of `AsynchQueueAgent`  encapsulate in themselves threading, queuing and 'business' logic at the same time, thus having a few orthogonal responsibilities. 

**Theoretical solution:** 
 - Use of single work-stealing queue and single thread pool should cut latency of end to end request-response cycle by ~50% due to increased data cache locality (tests have shown that with such implementation more than half of work items are being retrieved from thread local storage). It also completely removes CPU burning, which can lead to increased operational costs while running in cloud and on premises, and reduces number of thread context switches, thus improving latency and throughput at the same time. 

**Suggested solution:**
- Abstract threading and queuing logic behind  `ExecutorService` with API somewhat similar to .Net ThreadPool, this service will have switchable implementations. 
One of them will be just moved current logic, and another - the new one implemented as described in theoretical solution part. After move of current implementation descendants of `AsynchQueueAgent`  will be responsible only for the relevant to each of them algorithms and queuing of tasks to `ExecutorService`,  while `WorkerThread`s will be executing not only grain call path related items, but also those from  `AsynchQueueAgent`s. It's fairly easy to make such stage logic move without losing of any of current contracts, stages will still be existent, but behind the abstraction. 

**Details of additional threading implementation**:
 - In https://github.com/dotnet/orleans/pull/2060 it was implemented by adopting .Net ThreadPool code with necessary changes to make it work outside .Net library and support some features of existing implementation. Such approach gives flexibility, but also brings high costs of reviewing and supporting. 
[...]

PR #3663, opened 2017-11-10
Make execution service AsynchAgent's dependency

dVakulen:
First part of work needed for https://github.com/dotnet/orleans/issues/3498 

The main purpose of this PR is to make threading logic injectable dependency of `AsynchAgent`.

### New entities introduced: 

- `ThreadPerTaskScheduler` - contains extracted from `AsynchAgent`s threading logic. 
- `AsynchAgentTask` - task in scope of which now  `AsynchAgent` logic is being executed. Currently needed only for  passing of agent's name to executing thread.  
- `ExecutorService` - For now only dispatches work to the scheduler, but will be responsible for selecting executor based on configuration. 

### Planned further steps:
 - Unify `WorkQueue` and `AsynchQueueAgent` by moving queuing logic into executor. 
 - Remove the need for `BlockingCollection` in `InboundMessageQueue` by using executor's queuing abilities.
- Add ThreadPool based executor.


dVakulen:
@dotnet-bot test this please

dVakulen:
cc @gabikliot

src/Orleans.Core/Runtime/AsynchAgent.cs
```diff
@@ -130,30 +129,7 @@ public virtual void Stop()
             }
             Log.Verbose("Stopped agent");
         }
-
-        public void Abort(object stateInfo)
```
dVakulen: Was unused, as well as `Join` method.

src/Orleans.Core/Runtime/ThreadPerTaskScheduler.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Threading.Tasks;
+
+namespace Orleans.Runtime
+{
+    /// <summary>Provides a task scheduler that dedicates a thread per task.</summary>
+    public class ThreadPerTaskScheduler : TaskScheduler
+    {
+        private readonly Func<Task, string> threadNameProvider;
+
+        public ThreadPerTaskScheduler(Func<Task, string> threadProvider)
```
ReubenBond: nit: naming on this parameter is a little misleading

src/Orleans.Core/Runtime/ThreadPerTaskScheduler.cs
```diff
@@ -0,0 +1,46 @@
...
+        protected override IEnumerable<Task> GetScheduledTasks()
+        {
+            return Enumerable.Empty<Task>();
+        }
+
+        /// <summary>Starts a new thread to process the provided task.</summary>
+        /// <param name="task">The task to be executed.</param>
+        protected override void QueueTask(Task task)
+        {
+            new Thread(() => TryExecuteTask(task))
```
ReubenBond: Don't we need some method for joining/aborting/disposing these threads during shutdown? Eg by making this & ExecutorService disposable, so that ExecutorService disposes this. That way, both will be disposed when the container is disposed.

dVakulen: I tried to introduce as little additional changes as possible. The `AsynchAgent`s threads weren't being joined on dispose before, and I I didn't see the need in it here.

ReubenBond: After thinking about it more, I agree with you. Current code uses a cooperative cancellation approach and it's best if we continue doing that.

jdom:
Load test numbers were unaffected (as expected by looking at the code), so this should be good for merging. Any concerns @gabikliot or @jason-bragg ? This will unblock future perf-related PRs, but still allow users to make the changes opt-in until we are confident of lifting up a different default

PR #3668, opened 2017-11-13
Extract Microsoft.Orleans.Clustering.AzureStorage package from Microsoft.Orleans.OrleansAzureUtils package

attilah:
Separate Azure membership into a separate library and move shared pieces to Azure.Core


jdom:
Do we really need a shared component? I'd strongly suggest that if we can avoid it, let's do so. Even if that means linking some files to share logic, but avoid dependency on a shared binary.

attilah:
This is how the code is written. First we need to separate, then we can optimize. The same what we did with PlatformServices.

Why are you against the shared component **exactly**?

jdom:
Not sure I understand the comparison with platform services. It was really required with netstandard 1.5. It wasn't a temporary component. Fortunately netstandard 2.0 changed a lot and we did not require it any more in that platform.

I did not review yet why we need this shared component, that's why I'm asking you about it, since you are looking at that code. We will get rid of AzureSilo and AzureClient in favor of extension methods on Silo and Client builders, in case that changes anything. Would waiting for that change first make the shared component not needed?

attilah:
Streaming, EventHub and other azure parts are also sharing ErrorCode enum and Table Storage, Reminders sharing the Azure Table Storage helpers.

ReubenBond:
If there are no types which are communicated between the different split-out modules (there shouldn't be), then we can hopefully avoid a shared module. I'd personally prefer splitting the ErrorCodes enum and duplicating any utility classes which are actually shared (either c/p code or <Link>).

jdom:
Exactly, if it's just helper or constants, they can be split and logic linked. We already started doing some of that in several places

src/Orleans.Azure.Core/Orleans.Azure.Core.csproj
```diff
@@ -0,0 +1,24 @@
...
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.Azure.Core</PackageId>
+    <Title>Microsoft Orleans Azure Core</Title>
+    <Description>Library for shared Azure related classes of Microsoft Orleans.</Description>
+    <PackageTags>$(PackageTags) Azure</PackageTags>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AssemblyName>Orleans.Azure.Core</AssemblyName>
+    <RootNamespace>Orleans.AzureUtils</RootNamespace>
```
xiazen: Orleans.AzureUtils doesn't sound like a good namespace. Why don't use Orleans.Azure.Core.

galvesribeiro:
Will we do that on other projects like GCP, AWS and SQL?

Orleans.sln
```diff
@@ -140,13 +140,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Orleans.Transactions.Azure.
...
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Orleans.Membership.Azure.TableStorage", "src\Orleans.Membership.Azure.TableStorage\Orleans.Membership.Azure.TableStorage.csproj", "{E9A8A1F9-1943-45E2-9B9C-1E51A9154D76}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Orleans.Azure.Core", "src\Orleans.Azure.Core\Orleans.Azure.Core.csproj", "{6B6A56A7-1994-4A04-9D29-D5A0764FC9CA}"
-EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Orleans.Hosting.ServiceFabric", "src\Orleans.Hosting.ServiceFabric\Orleans.Hosting.ServiceFabric.csproj", "{1AF48B10-B2CE-4DE9-9A3F-0A656FBF1C02}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ServiceFabric", "ServiceFabric", "{A15BD3B2-513B-449C-A2E9-81DA2DE2FF92}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Orleans.ServiceFabric", "src\Orleans.ServiceFabric\Orleans.ServiceFabric.csproj", "{814FF01E-1466-4A71-A48F-F40143EF49AA}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Orleans.ServiceFabric", "src\Orleans.ServiceFabric\Orleans.ServiceFabric.csproj", "{814FF01E-1466-4A71-A48F-F40143EF49AA}"
```
ReubenBond: ah thanks

src/Shared/Azure/Utilities/ErrorCode.cs
```diff
@@ -1,6 +1,10 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 
+#if MEMBERSHIP_AZURE_TABLE_STORAGE
+namespace Orleans.Membership.Azure.TableStorage.Utilities
+#else
```
ReubenBond: These seem a little unnecessary, but it's fine.

src/Orleans.Core/Properties/AssemblyInfo.cs
```diff
@@ -2,8 +2,10 @@
 using Orleans.CodeGeneration;
 using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Orleans.Azure.Core")]
```
jdom: this no longer exists, right?
How about `OrleansAzureUtils` below, is it gone too?

src/Orleans.Membership.Azure.TableStorage/AzureBasedMembershipTable.cs
```diff
@@ -9,9 +9,9 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Orleans.AzureUtils.Configuration;
-using Orleans.Configuration;
 using LogLevel = Microsoft.Extensions.Logging.LogLevel;
 using Orleans.Runtime.Configuration;
+using Orleans.Membership.Azure.TableStorage;
```
jdom: perhaps also define `using ErrorCode = Orleans.Membership.Azure.TableStorage.Utulities.ErrorCode; ` to avoid such verbosity thought out the file (and undo all changes here, other than the namespace)?

src/Orleans.Membership.Azure.TableStorage/AzureMembershipTableStorageHostingExtensions.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Orleans.Hosting
 {
-    public static class AzureUtilsHostingExtensions
+    public static class AzureMembershipTableStorageHostingExtensions
```
jdom: order of words is incorrect, should be `AzureTableMembershipHostingOptions`

src/Orleans.Membership.Azure.TableStorage/Orleans.Membership.Azure.TableStorage.csproj
```diff
@@ -0,0 +1,37 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.Membership.Azure.TableStorage</PackageId>
+    <Title>Microsoft Orleans Membership Azure Table Storage</Title>
+    <Description>Membership provider for Azure Table Storage of Microsoft Orleans.</Description>
+    <PackageTags>$(PackageTags) Azure</PackageTags>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AssemblyName>Orleans.Membership.Azure.TableStorage</AssemblyName>
```
jdom: Was this the final decision on naming? I thought it was `Orleans.Membership.AzureTable`, and then we can have `Orleans.Membership.CosmosDb` and what not.
I'm not pushing back, and fine with whatever you guys decide, is just that I remember differently. /cc @sergeybykov

jdom:
LGTM. Most comments are minor, but I'd like Sergey's configurmation on the new package ID before we merge it.
BTW, we should make it standard to be explicit about additions and removals (and renames) of nuget packages in the PR description itself, as this is a greatly impacting change.

src/Orleans.Membership.Azure.TableStorage/AzureBasedMembershipTable.cs
attilah: That doesn't work because of the common Orleans namespace and it is ambiguous, but to reduce verbosity I'm adding a ```TableStorageErrorCode``` alias.

src/Orleans.Core/Properties/AssemblyInfo.cs
attilah: ```OrleansAzureUtils``` still needed for some types, but should be gone when that assembly is gone too.

src/Orleans.Membership.Azure.TableStorage/Orleans.Membership.Azure.TableStorage.csproj
sergeybykov: I like `Microsoft.Orleans.Membership.AzureTable` better than `Microsoft.Orleans.Membership.Azure.TableStorage`. Although Microsoft.Orleans.ClusterMembership.AzureTable` seems more clear to me.

ReubenBond: During the meeting, we decided that `Clustering` is a clearer name than `Membership` for most users

ReubenBond: We also decided we would call packages `AzureStorage`. instead of `AzureTables`, etc

ReubenBond: For package names, we decided to use `Persistence` instead of `Storage`, like `Microsoft.Orleans.Persistence.AzureStorage`

src/OrleansAzureUtils/OrleansAzureUtils.csproj
```diff
@@ -21,7 +28,12 @@
   <ItemGroup>
     <ProjectReference Include="..\Orleans.Core.Abstractions\Orleans.Core.Abstractions.csproj" />
     <ProjectReference Include="..\Orleans.Core\Orleans.Core.csproj" />
+    <ProjectReference Include="..\Orleans.Clustering.AzureStorage\Orleans.Clustering.AzureStorage.csproj" />
     <ProjectReference Include="..\Orleans.Runtime\Orleans.Runtime.csproj" />
     <ProjectReference Include="..\OrleansProviders\OrleansProviders.csproj" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Utilities\" />
```
jdom: This is probably a remnant of some deleted code?

src/Orleans.Clustering.AzureStorage/Orleans.Clustering.AzureStorage.csproj
```diff
@@ -0,0 +1,37 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.Clustering.AzureStorage</PackageId>
+    <Title>Microsoft Orleans Membership Azure Table Storage</Title>
+    <Description>Membership provider for Azure Table Storage of Microsoft Orleans.</Description>
+    <PackageTags>$(PackageTags) Azure</PackageTags>
```
jdom: might want to include both AzureTable and Table as tags

src/OrleansAzureUtils/OrleansAzureUtils.csproj
attilah: It is because of the linked file which resides in the Utilities folder, VS added it.

attilah:

src/Orleans.Clustering.AzureStorage/Orleans.Clustering.AzureStorage.csproj
attilah: AzureTable as a term is not a thing as of search: https://www.nuget.org/packages?q=Tags%3A%22AzureTable%22, Table is too common IMO.

PR #3669, opened 2017-11-14
Re-introduce runtime code generation

ReubenBond:


jdom:
This will conflict with https://github.com/dotnet/orleans/pull/3595. Which one should go in first? I assume that @benjaminpetit's is ready to be merged

test/TesterInternal/RuntimeCodeGenTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using Orleans.Serialization;
+using Xunit;
+
+namespace UnitTests
+{
+    [TestCategory("BVT"), TestCategory("CodeGen")]
+    public class RuntimeCodeGenTests
+    {
+        [Fact]
+        public void RuntimeCodeGenAddsSupportClasses()
```
jdom: Is it worth adding a functional that starts a silo and verifies the grain and serialization works?

ReubenBond: Sure thing, will do

jdom:
Although what you are solving by not requiring a SerializationManager for codegen is what @benjaminpetit wanted to do eventually, but couldn't do yet, so he hacked around it... maybe it's better to merge this one first and have him rebase on this one? I'm fine with whatever you guys decide. Although from what I see about the extension methods discussion this PR might take longer?

src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+            }
+
+            tempPartManager.AddApplicationPart(new AssemblyPart(typeof(RuntimeVersion).Assembly));
+            tempPartManager.AddFeatureProvider(new AssemblyAttributeFeatureProvider<GrainInterfaceFeature>());
+            tempPartManager.AddFeatureProvider(new AssemblyAttributeFeatureProvider<GrainClassFeature>());
+            tempPartManager.AddFeatureProvider(new AssemblyAttributeFeatureProvider<SerializerFeature>());
+            tempPartManager.AddFeatureProvider(new BuiltInTypesSerializationFeaturePopulator());
+            
+            var codeGenerator = new RoslynCodeGenerator(tempPartManager, new NullLoggerFactory());
+            var generatedAssembly = codeGenerator.GenerateAndLoadForAssemblies(manager.Assemblies);
```
jdom: Does it mean that codegen is run immediately? Shouldn't it wait until a specific point in time just before building the container, so that all the different codegen tasks occur together? Because if I use the WithCodeGeneration extension method multiple times, it will run multiple times. Or did I misunderstand this?

ReubenBond: That's correct - code gen will run immediately, but not necessarily for one assembly at a time. The user has a degree of control over it.

In the following code, codegen is executed twice: once for IMyGrain's asm and MyGrain's asm and then again for whatever asms are in the app base dir. The second execution will see the code generated by the first, so it won't double-generate.
```C#
.ConfigureApplicationPartManager(parts =>
{
    parts.AddApplicationPart(typeof(IMyGrain).Assembly)
         .AddApplicationPart(typeof(MyGrain).Assembly)
         .WithCodeGeneration();
    parts.AddFromApplicationBaseDirectory().WithCodeGeneration();
});
```

Does that seem alright or would you prefer that all actions are queued up and executed only during Build()?

jdom: I guess it's fine if there is a way to chain everything and then execute once. And if they don't double-generate even when executing twice, then that's harmless anyway.

ReubenBond:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test functional

src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// Extensions for <see cref="IApplicationPartManagerWithAssemblies"/> for invoking code generation.
+    /// </summary>
+    public static class ApplicationPartManagerCodeGenExtensions
+    {
+        /// <summary>
+        /// Generates support code for the the provided assembly and adds it to the builder.
+        /// </summary>
+        /// <param name="manager">The builder.</param>
+        /// <returns>A builder with support parts added.</returns>
+        public static IApplicationPartManagerWithAssemblies WithCodeGeneration(this IApplicationPartManagerWithAssemblies manager)
```
jdom: should the return be `void`. Otherwise it would be confusing what would happen if you keep chaining, wouldn't it?

src/Orleans.Core/Serialization/TypeUtilities.cs
```diff
@@ -14,10 +14,14 @@ namespace Orleans.Serialization
...
+        internal static bool IsOrleansPrimitive(this Type t)
         {
-            var t = typeInfo.AsType();
-            return typeInfo.IsPrimitive || typeInfo.IsEnum || t == typeof(string) || t == typeof(DateTime) || t == typeof(Decimal) || (typeInfo.IsArray && typeInfo.GetElementType().GetTypeInfo().IsOrleansPrimitive());
+            return t.IsPrimitive ||
+                   t.IsEnum ||
+                   t == typeof(string) ||
+                   t == typeof(DateTime) ||
+                   t == typeof(Decimal) ||
+                   (t.IsArray && t.GetElementType().GetTypeInfo().IsOrleansPrimitive());
```
jdom: You can probably avoid calling GetTypeInfo() since the extension method is now on `Type`

src/Orleans.Core/Serialization/TypeUtilities.cs
```diff
@@ -231,12 +242,12 @@ public static bool IsAccessibleFromAssembly(Type type, Assembly assembly)
...
             else if (typeInfo.IsGenericTypeDefinition)
             {
-                // Guard against invalid type constraints, which appear when generating code for some languages, such as F#.
+                // Guard against unrepresentable type constraints, which appear when generating code for some languages, such as F#.
                 foreach (var parameter in typeInfo.GenericTypeParameters)
                 {
                     foreach (var constraint in parameter.GetTypeInfo().GetGenericParameterConstraints())
                     {
-                        if (IsSpecialClass(constraint)) return false;
+                        if (constraint == typeof(Array) || constraint == typeof(Delegate) || constraint == typeof(Enum)) return false;
```
jdom: You dropped `ValueType` when inlining. Was it by design?

src/Orleans.Core.Abstractions/CodeGeneration/KnownAssemblyAttribute.cs
```diff
@@ -34,6 +34,7 @@ public KnownAssemblyAttribute(string assemblyName)
         /// serializable.
         /// </summary>
         /// <remarks>This is equivalent to specifying <see cref="KnownTypeAttribute"/> for all types.</remarks>
+        [Obsolete("This property is no longer supported. Serializers will be generated for all accessible types.")]
```
jdom: What's the reason for this change?

src/Orleans.Core/Core/ClientBuilderExtensions.cs
```diff
@@ -173,5 +177,46 @@ public static IClientBuilder UseStaticGatewayListProvider(this IClientBuilder bu
...
+            return builder;
+        }
+
+        /// <summary>
+        /// Configures the <see cref="ApplicationPartManager"/> for this builder.
+        /// </summary>
+        /// <param name="builder">The builder.</param>
+        /// <param name="configure">The configuration delegate.</param>
+        /// <returns>The builder.</returns>
+        public static IClientBuilder ConfigureApplicationPartManager(this IClientBuilder builder, Action<IApplicationPartManager> configure)
```
jdom: How about dropping *Manager* and just have `ConfigureApplicationParts(...)`? It is using the manager under the hoods, but the user is configuring which parts the app will use (and of course it's less verbose)

src/Orleans.Runtime/Hosting/Generic/SiloHostBuilderExtensions.cs
```diff
@@ -108,5 +110,46 @@ public static ISiloHostBuilder ConfigureLogging(this ISiloHostBuilder builder, A
...
+        /// <returns>The <see cref="ApplicationPartManager"/> for this instance.</returns>
+        public static ApplicationPartManager GetApplicationPartManager(this ISiloHostBuilder builder) => ApplicationPartManagerExtensions.GetApplicationPartManager(builder.Properties);
+
+        /// <summary>
+        /// Adds an <see cref="IApplicationPart"/> to the <see cref="ApplicationPartManager"/>.
+        /// </summary>
+        /// <param name="builder">The builder.</param>
+        /// <param name="assembly">The assembly.</param>
+        /// <returns>The builder.</returns>
+        public static ISiloHostBuilder AddApplicationPart(this ISiloHostBuilder builder, Assembly assembly)
```
jdom: I'm on the fence with this one. Do you envision this to be the most common case? Otherwise to make the jump to add references too for example, then it's a lot different, and maybe the best is to get them to use the fluent interface in the first place.

src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
ReubenBond: you can call WithCodeGeneration().WithReferences(), for example - but we could make it void and force the order. What do you think?

src/Orleans.Core/Serialization/TypeUtilities.cs
ReubenBond: It is: it was supposed to tell us what types cannot be used as generic constraints in C#, but `ValueType` can (`T : struct`).

src/Orleans.Core.Abstractions/CodeGeneration/KnownAssemblyAttribute.cs
ReubenBond: It's simpler and less error prone to just consider all types in the few assemblies we're processing

src/Orleans.Runtime/Hosting/Generic/SiloHostBuilderExtensions.cs
ReubenBond: I'm on the fence, too - should we just axe it? Shortcuts like this can be added later if we want, so maybe it's better to be conservative.

jdom: My vote is yes: let's axe it for now

src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
jdom: But what would that do? it's not clear to me

src/Orleans.Core.Abstractions/CodeGeneration/KnownAssemblyAttribute.cs
jdom: But could be a lot more expensive now that runtime codegen is back, right? I'm not pushing back, just trying to understand

src/Orleans.Runtime/Hosting/Generic/SiloHostBuilderExtensions.cs
jdom: Should I wait for this change or you think we should keep it then?

src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
ReubenBond: It should include those assemblies, generate code for them, and also include their references

src/Orleans.Core.Abstractions/CodeGeneration/KnownAssemblyAttribute.cs
ReubenBond: It could be, since we need to perform accessibility checks for more types now. We cache those checks, though, so it shouldn't be an issue.

PR #3670, opened 2017-11-14
Minimal silo lifecycle

jason-bragg:
Added staged startup of silo via Lifecycle.
This work is still primitive and minimal.
Stages are preliminary, and will need to be further refined.
We still need to modify components to interact direct with the lifecycle rather than requiring silo to know about and start all components.  Only component modified so far, for proof of concept purposes, is the InsideRuntimeClient.  These changes can be made in future passes.

### DIFF
--- a/src/Orleans.Runtime/Core/InsideRuntimeClient.cs
+++ b/src/Orleans.Runtime/Core/InsideRuntimeClient.cs
@@ -20,13 +20,14 @@ using System.Diagnostics;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Orleans.Configuration;
+using System.Threading;
 
 namespace Orleans.Runtime
 {
     /// <summary>
     /// Internal class for system grains to get access to runtime object
     /// </summary>
-    internal class InsideRuntimeClient : ISiloRuntimeClient
+    internal class InsideRuntimeClient : ISiloRuntimeClient, ILifecycleParticipant<ISiloLifecycle>
     {
         private readonly ILogger logger;
         private readonly Logger callbackDataLogger;
@@ -688,7 +689,7 @@ namespace Orleans.Runtime
 
         #endregion
 
-        internal void Stop()
+        private Task OnRuntimeInitializeStop(CancellationToken tc)
         {
             lock (disposables)
             {
@@ -704,11 +705,14 @@ namespace Orleans.Runtime
                     }
                 }
             }
+            return Task.CompletedTask;
         }
 
-        internal void Start()
+        private Task OnRuntimeInitializeStart(CancellationToken tc)
         {
+            typeManager.Start();
             GrainTypeResolver = typeManager.GetTypeCodeMap();
+            return Task.CompletedTask;
         }
 
         public IGrainTypeResolver GrainTypeResolver { get; private set; }
@@ -801,6 +805,11 @@ namespace Orleans.Runtime
             throw new ArgumentException(
                 $"Provider extension handler type {handlerType} was not found in the type manager",
                 nameof(handlerType));
+        }
+
+        public void Participate(ISiloLifecycle lifecycle)
+        {
+            lifecycle.Subscribe(SiloLifecycleStage.RuntimeInitialize, OnRuntimeInitializeStart, OnRuntimeInitializeStop);
         }
     }
 }
--- a/src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
+++ b/src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
@@ -9,21 +9,21 @@ namespace Orleans.Runtime
         /// <summary>
         /// Initialize silo runtime
         /// </summary>
-        RuntimeInitialize = (1 << 10),
+        RuntimeInitialize = 1000,
 
         /// <summary>
         /// Start runtime services
         /// </summary>
-        RuntimeServices = RuntimeInitialize + (1 << 10),
+        RuntimeServices = 2000,
 
         /// <summary>
         /// Start application layer services
         /// </summary>
-        ApplicationServices = RuntimeServices + (1 << 10),
+        ApplicationServices = 3000,
 
         /// <summary>
         /// Silo is active and available to service requests
         /// </summary>
-        SiloActive = ApplicationServices + (1 << 10),
+        SiloActive = 4000,
     }
 }
--- a/src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
+++ b/src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
@@ -9,21 +9,21 @@ namespace Orleans.Runtime
         /// <summary>
         /// Initialize silo runtime
         /// </summary>
-        RuntimeInitialize = 1 << 10,
+        RuntimeInitialize = (1 << 10),
 
         /// <summary>
         /// Start runtime services
         /// </summary>
-        RuntimeServices = RuntimeInitialize + 1 << 10,
+        RuntimeServices = RuntimeInitialize + (1 << 10),
 
         /// <summary>
         /// Start application layer services
         /// </summary>
-        ApplicationServices = RuntimeServices + 1 << 10,
+        ApplicationServices = RuntimeServices + (1 << 10),
 
         /// <summary>
         /// Silo is active and available to service requests
         /// </summary>
-        SiloActive = ApplicationServices + 1 << 10,
+        SiloActive = ApplicationServices + (1 << 10),
     }
 }
--- a/src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
+++ b/src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
@@ -1,0 +1,29 @@
+﻿
+namespace Orleans.Runtime
+{
+    /// <summary>
+    /// Stages of a silo's lifecycle.
+    /// </summary>
+    public enum SiloLifecycleStage
+    {
+        /// <summary>
+        /// Initialize silo runtime
+        /// </summary>
+        RuntimeInitialize = 1 << 10,
+
+        /// <summary>
+        /// Start runtime services
+        /// </summary>
+        RuntimeServices = RuntimeInitialize + 1 << 10,
+
+        /// <summary>
+        /// Start application layer services
+        /// </summary>
+        ApplicationServices = RuntimeServices + 1 << 10,
+
+        /// <summary>
+        /// Silo is active and available to service requests
+        /// </summary>
+        SiloActive = ApplicationServices + 1 << 10,
+    }
+}
--- a/src/Orleans.Runtime/Silo/Silo.cs
+++ b/src/Orleans.Runtime/Silo/Silo.cs
@@ -131,7 +131,6 @@ namespace Orleans.Runtime
         private SchedulingContext membershipOracleContext;
         private SchedulingContext multiClusterOracleContext;
         private SchedulingContext reminderServiceContext;
-        private bool gracefully = true;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Silo"/> class.
@@ -371,7 +370,7 @@ namespace Orleans.Runtime
             logger.Debug("Finished creating System Targets for this silo.");
         }
 
-        private void InjectDependencies()
+        private async Task InjectDependencies()
         {
             healthCheckParticipants.Add(membershipOracle);
 
@@ -398,8 +397,8 @@ namespace Orleans.Runtime
             }
 
             RegisterSystemTarget(catalog);
-            scheduler.QueueAction(catalog.Start, catalog.SchedulingContext)
-                .WaitWithThrow(initTimeout);
+            await scheduler.QueueAction(catalog.Start, catalog.SchedulingContext)
+                .WithTimeout(initTimeout);
 
             // SystemTarget for provider init calls
             providerManagerSystemTarget = Services.GetRequiredService<ProviderManagerSystemTarget>();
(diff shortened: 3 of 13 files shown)


src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -704,11 +705,14 @@ internal void Stop()
                     }
                 }
             }
+            return Task.CompletedTask;
         }
 
-        internal void Start()
+        private Task OnRuntimeInitializeStart(CancellationToken tc)
         {
+            typeManager.Start();
```
jason-bragg: May want a 'load resources' stage prior to runtime initialization stage for assembly loading and related.

src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        /// <summary>
+        /// Start runtime services
+        /// </summary>
+        RuntimeServices = RuntimeInitialize + 1 << 10,
+
+        /// <summary>
+        /// Start application layer services
+        /// </summary>
+        ApplicationServices = RuntimeServices + 1 << 10,
```
jason-bragg: Consider preventing users from registering actions earlier than this stage.

src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        /// <summary>
+        /// Silo is active and available to service requests
+        /// </summary>
+        SiloActive = ApplicationServices + 1 << 10,
```
jdom: Would you mind using parenthesis? I believe the operator precedence is not what you intend, but even if it is, can you make it explicit to avoid confusion?

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -644,7 +660,8 @@ private void ConfigureThreadPoolAndServicePointSettings()
         /// </summary>
         public void Stop()
         {
-            Terminate(false);
+            this.gracefully = false;
+            StopAsync(CancellationToken.None).GetAwaiter().GetResult();
```
jdom: Shouldn't you pass a canceled token?
BTW, the `gracefully` field is immediately overwritten after calling `StopAsync`

jason-bragg: good catch.  I change the behavior and forgot about the other calls.

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -802,5 +806,10 @@ internal static IGrainExtensionMethodInvoker TryGetExtensionInvoker(GrainTypeMan
                 $"Provider extension handler type {handlerType} was not found in the type manager",
                 nameof(handlerType));
         }
+
+        public void Participate(ISiloLifecycle lifecycle)
+        {
+            lifecycle.Subscribe(SiloLifecycleStage.RuntimeInitialize, OnRuntimeInitializeStart, OnRuntimeInitializeStop);
```
jdom: Using the same stage would mean that it will run in parallel with the InsideRuntimeClient, correct? Couldn't that cause issues?

jason-bragg: This means that the Inside runtime client receives its OnRuntimeInitializeStart and OnRuntimeInitializeStop calls in  parallel with everything else registered at the RuntimeInitialize stage, which, at this time, is only the silo.

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -654,21 +671,22 @@ public void Stop()
         /// </summary>
         public void Shutdown()
         {
-            Terminate(true);
+            this.gracefully = true;
```
ReubenBond: I'm not certain about this, but does it make sense to use a cancelled CTS instead of having a `gracefully` field to communicate ungraceful termination? On one hand, it means that ungraceful termination is communicated to all lifecycle subscribers, but on the other they might not know how to deal with that and might throw `OperationCancelledException`. It's just a thought, you probably already considered it. In the ASP.NET Generic Host case, they use a cancelled token to signify ungraceful termination.

jason-bragg: I'm not sure, unfortunately. :/  We should think about this more.

In the lifecycle, a canceled token means cancel the operation.  So on Start it means cancel the start operation and on Stop it means cancel the stop operation.  This means that, when canceled, the lifecycle observers never even get called, so they never get a chance to even perform an ungraceful shutdown.  I think this is the desired behavior, but I've not fully thought through the implications of this pattern in the context of existing shutdown logic.  The goal of this PR was to wrap the existing behaviors in the lifecycle pattern (which I think it does..?) allowing us to refactor over time to improve the system.

In general, I think a canceled token should indicate ungraceful shutdown, but this should mean that lifecycle observers are simply not called to shutdown, not that they are called in a way that let's them know they are being ungracefully shutdown.  IMO, we should consider how to tell components to ungracefully shutdown only if the simpler pattern demonstrates itself as insufficient.  This means removing the 'gracefully' flag and related logic in a future refactor.

src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Stages of a silo's lifecycle.
+    /// </summary>
+    public enum SiloLifecycleStage
+    {
+        /// <summary>
+        /// Initialize silo runtime
+        /// </summary>
+        RuntimeInitialize = (1 << 10),
```
jdom: I would be explicit about the bucket size using round (base 10) numbers, which are easier to troubleshoot at runtime. Say `1000` instead of `1024` (which is what `1 <<10` is)

ReubenBond: To this end, we could use `1e10` instead. My first impression was the same, and we have precedence with ErrorCodes to use base 10

jdom:
>             this.subscribers.TryAdd(orderedObserver, orderedObserver);

Since it might now be common to subscribe after the lifecycle already started, we should probably make sure that the new subscriber catches up and gets started somehow too, right? I believe with the current approach it might not start even if the subscription is for a stage that is in the future, and for sure it won't start if the stage is already in the past.

---
Refers to: src/Orleans.Core/Lifecycle/LifecycleObservable.cs:76 in 56ea196. [](commit_id = 56ea196df8421f47da699369d8f90c93e0c72a50, deletion_comment = False)

src/Orleans.Runtime/Silo/Silo.cs
ReubenBond: I'm fine with this, let's keep the bool for now

jason-bragg:
"Since it might now be common to subscribe after the lifecycle already started'.

I don't think we want to support or encourage that pattern.

jdom:
> I don't think we want to support or encourage that pattern.

I'm not sure yet that we don't want to support it. But I'm OK postponing that decision now that at least we enforce that it doesn't fail silently. If we ever find a reason to allow it, we can enable it then

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -424,10 +439,11 @@ private void DoStart()
 
             // Set up an execution context for this thread so that the target creation steps can use asynch values.
             RuntimeContext.InitializeMainThread();
-            
+
             // Initialize the implicit stream subscribers table.
             var implicitStreamSubscriberTable = Services.GetRequiredService<ImplicitStreamSubscriberTable>();
-            implicitStreamSubscriberTable.InitImplicitStreamSubscribers(this.grainTypeManager.GrainClassTypeData.Select(t => t.Value.Type).ToArray());
+            var grainTypeManager = Services.GetRequiredService<GrainTypeManager>();
```
jdom: Sorry to keep pushing on this, but it doesn't feel completely safe (at least by just browsing the code here) to use the grain type manager when it might still be starting in parallel. In the past we were sure that by the time we try to use it, it was already started. Here we are relying on luck a little bit, aren't we?

jason-bragg: I don't see how?  This logic runs in RuntimeServices stage, which comes after RuntimeInitialize sage, which is where the grainTypeManager is started.  Each stage is performed in parallel, but the stages are run in order.  I don't see how we could get here without the grainTypeManager being started.  Am I missing something?

jason-bragg:
dotnet-bot test this please

jason-bragg:
@dotnet-bot test this please

src/Orleans.Runtime/Silo/Silo.cs
jdom: ah, true, I missed that detail, I thought this was running in the same stage as the other service, and hence potentially in parallel. I missed the fact that the silo participates in 2 stages at this point

---
In reply to: [151573273](https://github.com/dotnet/orleans/pull/3670#discussion_r151573273) [](ancestors = 151573273)

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -702,106 +720,102 @@ private void Terminate(bool gracefully)
...
-                }
-                catch (Exception exc)
+                if (this.gracefully)
                 {
-                    logger.Error(ErrorCode.SiloFailedToStopMembership, String.Format("Failed to {0} membership oracle. About to FastKill this silo.", operation), exc);
-                    return; // will go to finally
+                    logger.Info(ErrorCode.SiloShuttingDown, "Silo starting to Shutdown()");
+                    // 1: Write "ShutDown" state in the table + broadcast gossip msgs to re-read the table to everyone
+                    scheduler.QueueTask(this.membershipOracle.ShutDown, this.membershipOracleContext)
+                        .WaitWithThrow(stopTimeout);
```
jdom: Use `await xxx.WithTimeout` everywhere and turn the method to be fully async. Otherwise the `LifecycleObservable` will not parallelize as expected.
The same with the `OnRuntimeInitializeStop` method.

src/Orleans.Runtime/Silo/Silo.cs
jdom: I don't think the current approach is good, as it only works because we temporarily made the `Silo` itself participate in the lifecycle, and it is the only one that knows whether it's graceful or ungraceful shutdown.
I think we should allow every single participant know whether the shutdown is graceful or not and let *it* decide to no-op if it's not important. Unfortunately that means querying the cancelation token, which is a little unintuitive, but can be documented and eventually will be familiar to people creating `IHostedService` implementations too (and that's a good thing, as Orleans extensibility will be similar to the extensibility of the generic Host).
Canceling Start IMO should preclude more participating services to start, and the currently starting services should receive the cancellation request. Then (externally the Silo will) initiate the reverse stopping from that stage on, with a canceled token or not, depending whether the shutdown is graceful.
Also, the host might initiate a graceful stop, but a few seconds in might decide to cancel the token and continue as ungraceful (this is to note that the "Start" and "Stop" part of the lifecycle use effectively different cancellation tokens).

For more info about the semantics of IHostedService, please see the comments in there: [IHostedService](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs)
BTW, looking at [how they are executed](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.Extensions.Hosting/Internal/Host.cs#L51), it looks like even if the Start operation is canceled, they still call Start on all the IHostServices implementations, but passing a canceled token. I'm fine with doing that or not even starting the remainder of participants, since I can't think of anything that a lifecycle participant could ever do (other than no-op) on Start if the token is canceled. But I'm fine either way.

jason-bragg: What you suggest makes sense and is more inline with asp.net.  On further reflection, 'canceling' the stop operation does not make as much sense as canceling the start operation, so using the token to communicate graceful/ungraceful shutdown has more utility.  If we find we need to actually cancel a stop operation, we can address that when needed.

jdom: On hindsight, even the Shutdown operation can pass a token that automatically cancels after 5 minutes, in case there is something stalling the shutdown

src/Orleans.Runtime/Silo/Silo.cs
jason-bragg: I'll change what I can, but a lot of our startup logic is synchronous and it's out of scope for this PR to significantly refactor all of that.  For instance, there are many cases where async actions are wrapped in SafeExecute, which assumes synchronous actions.

jdom: Sure, address the trivial ones. As long as the first currently blocking operation just yields, then we already achieve parallelism, we can address others later

src/Orleans.Runtime/Silo/Silo.cs
jason-bragg: Let's note these concerns and resolve them in design discussions, not in PR.  If this PR preserves the existing behaviors and enables future improvements, then it has achieved it's purpose.

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -413,11 +424,15 @@ private void DoStart()
...
             ConfigureThreadPoolAndServicePointSettings();
 
-            // This has to start first so that the directory system target factory gets loaded before we start the router.
-            grainTypeManager.Start();
-            runtimeClient.Start();
+            return Task.CompletedTask;
+        }
+
+        private Task OnRuntimeServicesStart(CancellationToken ct)
+        {
```
jdom: BTW, you can do an `await Task.Yield();` to make sure the method releases immediately instead of blocking synchronously and can run in parallel with others.
You can also address the trivial WaitWithThrow calls and replace with await WithTimeout, but as we mentioned in the other comment, I wouldn't convert everything to async at this point. It will be a little bit cleaner (in terms of how exceptions propagate), but the reality is that making them truly async is not a boost in perf or anything like that (as long as at least the method yields fast when starting the lifecycle method)

PR #3671, opened 2017-11-14
Transaction manager benchmarks

jason-bragg:
Benchmarks of transction manager for development purposes.  Usefull for targeted optimization.
Currently seeing:
In Memory ~ 200k/sec
Agains Azure ~ 4k/sec

### DIFF
--- a/test/Benchmarks/Benchmarks/TransactionManager/TransactionManagerBentchmarks.cs
+++ b/test/Benchmarks/Benchmarks/TransactionManager/TransactionManagerBentchmarks.cs
@@ -16,7 +16,7 @@ using TestExtensions;
 
 namespace Benchmarks.TransactionManager
 {
-    public class TransactionManagerBentchmarks
+    public class TransactionManagerBenchmarks
     {
         private const int TransactionsPerRun = 500000;
         private const int ConcurrentTransactionsTransactions = TransactionsPerRun/10;
--- a/test/Benchmarks/Benchmarks/TransactionManager/TransactionManagerBentchmarks.cs
+++ b/test/Benchmarks/Benchmarks/TransactionManager/TransactionManagerBentchmarks.cs
@@ -1,0 +1,163 @@
+﻿
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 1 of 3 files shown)


test/Benchmarks/Benchmarks/TransactionManager/TransactionManagerBentchmarks.cs
```diff
@@ -0,0 +1,163 @@
...
+using Orleans.Runtime.Configuration;
+using Orleans.Transactions.Abstractions;
+using Orleans.Transactions;
+using Orleans.Transactions.Development;
+using Orleans.Transactions.Azure;
+using TestExtensions;
+
+namespace Benchmarks.TransactionManager
+{
+    public class TransactionManagerBentchmarks
```
ReubenBond: typo on class/file name

dVakulen:
I'd suggest to consider making such kind of benchmarks (which exists in source repository, and could be ran on bvt basis) to be executed alongside already existing load performance test as additional round. That could be inner / outer test circles with inner focusing on process wide case, while outer handing distributed one. (Existing way of ensuring performance stability is quite heavy and focused on multi-node case, thus leaving single-host improvements harder to measure, and leaving performance assurance of covered by such benchmarks cases to manual observation on heterogeneous instances)

jason-bragg:
@dVakulen, I don't disagree with what you're suggesting, but it's a question of work vs. benefit.  Nothing is free; building and maintaining performance testing and related infrastructure must demonstrate more value than other things that time/effort could be spent on.

I developed these bench marks, not to actually measure the performance, but to have a means of rapidly iterating while working on optimizations.  That is, these benchmarks were created as  development tools.  For instance, I was able to improve the TM's performance against azure storage from ~4k to ~14k while prototyping optimizations.  I could not have achieved this without a micro-benchmark, so having developed one, I figured I should check it in.

At this time, I advocate we develop and leverage micro-benchmarks, and related infrastructure, primarily as development tools for rapid iteration.

We have load tests (and a load test framework) which are far better at measuring performance.  Unfortunately these are in a private repo.  I would love to see this moved into the public repo, but time has not allowed for this.  I'd definitely advocate time be spent on making the load tests public over a set of official micro-benchmarks.

PR #3672, opened 2017-11-15
Add metrics to monitor transaction agent batch calls

xiazen:
Add metrics to monitor transaction agent batch TM calls
Metric monitor configured through NodeConfig.StatisticsMetricsTableWriteInterval as reporting frequency

### DIFF
--- a/src/Orleans.Runtime/Transactions/TransactionAgent.cs
+++ b/src/Orleans.Runtime/Transactions/TransactionAgent.cs
@@ -1,13 +1,99 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Orleans.Runtime;
 using Orleans.Concurrency;
+using Orleans.Runtime.Configuration;
+using DateTime = System.DateTime;
 
 namespace Orleans.Transactions
 {
+    internal class TransactionAgentMetrics
+    {
+        //TPS in current monitor window
+        private const string BatchStartTransanctionsDeltaTPS = "TransactionAgent.BatchStartTransactions.Delta.TPS";
+        //avg latency in current monitor window
+        private const string AvgBatchStartTransactionsDeltaLatency = "TransactionAgent.BatchStartTransactions.Delta.AvgLatency";
+        private const string BatchCommitTransactionsDeltaTPS = "TransactionAgent.BatchCommitTransactions.Delta.TPS";
+        private const string AvgBatchCommitTransactionsDeltaLatency = "TransactionAgent.BatchCommitTransactions.Delta.AvgLatency";
+
+        internal int BatchStartTransactionsRequestCounter { get; set; }
+
+        internal int BatchCommitTransactionsRequestsCounter { get; set; }
+        private TimeSpan batchStartTransactionsRequestLatencyCounter = TimeSpan.Zero;
+        internal TimeSpan BatchStartTransactionsRequestLatencyCounter
+        {
+            get { return batchStartTransactionsRequestLatencyCounter; }
+            set
+            {
+                batchStartTransactionsRequestLatencyCounter = value;
+                this.periodicMonitor.TryAction(DateTime.UtcNow);
+            }
+        }
+        private TimeSpan batchCommitTransactionsRequestLatencyCounter = TimeSpan.Zero;
+
+        internal TimeSpan BatchCommitTransactionsRequestLatencyCounter
+        {
+            get { return batchCommitTransactionsRequestLatencyCounter; }
+            set
+            {
+                batchCommitTransactionsRequestLatencyCounter = value;
+                this.periodicMonitor.TryAction(DateTime.UtcNow);
+            }
+        }
+        private DateTime lastReportTime = DateTime.UtcNow;
+        private ITelemetryProducer telemetryProducer;
+        private PeriodicAction periodicMonitor;
+
+        public TransactionAgentMetrics(ITelemetryProducer producer, TimeSpan interval)
+        {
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Transactions/TransactionAgent.cs
```diff
@@ -1,13 +1,99 @@
...
+            this.batchStartTransactionsRequestLatencyCounter = TimeSpan.Zero;
+            this.batchCommitTransactionsRequestLatencyCounter = TimeSpan.Zero;
+        }
+
+        private void ReportMetrics()
+        {
+            var now = DateTime.UtcNow;
+           
+            //batch start metrics
+            var batchStartTransactionTPS = BatchStartTransactionsRequestCounter / (now - lastReportTime).TotalSeconds;
```
jason-bragg: Plz take max of total seconds and 1 to ensure no div by zero error

src/Orleans.Runtime/Transactions/TransactionAgent.cs
```diff
@@ -274,11 +302,87 @@ public bool IsAborted(long transactionId)
...
+        private async Task CommitTransactions(List<TransactionInfo> committingTransactions,
+            HashSet<long> outstandingCommits)
+        {
+            var stopWatch = Stopwatch.StartNew();
+            try
+            {
+                metrics.BatchCommitTransactionsRequestsCounter++;
+                var commitResponse = await this.tmService.CommitTransactions(committingTransactions, outstandingCommits);
+                metrics.BatchCommitTransactionsRequestLatencyCounter += stopWatch.Elapsed;
+                stopWatch.Stop();
```
jason-bragg: Probably doesn't matter, but usually call stop prior to getting the elapsed time.

src/Orleans.Runtime/Transactions/TransactionAgent.cs
```diff
@@ -274,11 +302,87 @@ public bool IsAborted(long transactionId)
...
+                    commitResponse.ReadOnlyTransactionId);
+                this.abortLowerBound = Math.Max(this.abortLowerBound, commitResponse.AbortLowerBound);
+                logger.Debug(ErrorCode.Transactions_ReceivedTMResponse,
+                    "{0} transactions committed. readOnlyTransactionId {1}, abortLowerBound {2}",
+                    committingTransactions.Count, ReadOnlyTransactionId, abortLowerBound);
+            }
+            catch (Exception e)
+            {
+                metrics.BatchCommitTransactionsRequestLatencyCounter += stopWatch.Elapsed;
+                stopWatch.Stop();
```
jason-bragg: If we're trying to handle timing of commit call regardless of error, should probably wrap the this.tmService.CommitTransactions(..) call in a try finally, with the metrics done in the finally, and leave this section to error handling.  The way this is written we could double count the metrics if an error occurs after the CommitTransactions call.

src/Orleans.Runtime/Transactions/TransactionAgent.cs
```diff
@@ -274,11 +302,87 @@ public bool IsAborted(long transactionId)
...
         }
 
+        private async Task CommitTransactions(List<TransactionInfo> committingTransactions,
+            HashSet<long> outstandingCommits)
+        {
+            var stopWatch = Stopwatch.StartNew();
+            try
+            {
+                metrics.BatchCommitTransactionsRequestsCounter++;
+                var commitResponse = await this.tmService.CommitTransactions(committingTransactions, outstandingCommits);
```
jason-bragg: To get a handle on the batching behavior, which is the point of all of this complexity, it would be good to track the average committingTransactions size per request.
Same for the start transaction operation.

src/Orleans.Runtime/Transactions/TransactionAgent.cs
```diff
@@ -1,13 +1,99 @@
...
+
+        internal int BatchCommitTransactionsRequestsCounter { get; set; }
+        private TimeSpan batchStartTransactionsRequestLatencyCounter = TimeSpan.Zero;
+        internal TimeSpan BatchStartTransactionsRequestLatencyCounter
+        {
+            get { return batchStartTransactionsRequestLatencyCounter; }
+            set
+            {
+                batchStartTransactionsRequestLatencyCounter = value;
+                this.periodicMonitor.TryAction(DateTime.UtcNow);
```
jason-bragg: May want to expose some sort of periotic trigger in the public surface and have it called in the TransactionAgents  ProcessRequests loop, rather than in property setters, which is definitely an unexpected behavior.

xiazen: I wanted to encapsulate metric related logic in this Metric class. So in that sense, metric reporting logic should be embedded in the class itself, without TransactionAgent/its monitoring target to worry about it. 
And setters are called during ProcessRequests loop, so functionality wise it is the same. 
But it is a bit weird to do it in setters, I will update this

src/Orleans.Runtime/Transactions/TransactionAgent.cs
xiazen: why max? I can see have a zero check here , and skip if it is zero secs.
But take a max between total secs and 1 make the math wrong, isn't it?

jason-bragg: On the odd chance someone configures this wrong and it fires in less than a second, it will either report all the transactions in the window as TPS (which is incorrect for a window less than a second) or it will report nothing.
Either way works but yes, what I suggested will give wrong math, but only a little (in theory), whereas the alternative would produce no info a all.
Maybe we should ensure that the report interval is >= 1second?

jason-bragg:
@dotnet-bot test functional

xiazen:

ReubenBond:
@dotnet test functional

PR #3678, opened 2017-11-17
Update PubSubRendezvousGrain.cs

bob-dawson:
GetAllSubscriptions should compare consumerReference to consumerReference, not to consumer

### DIFF
--- a/src/Orleans.Runtime/Streams/PubSub/PubSubRendezvousGrain.cs
+++ b/src/Orleans.Runtime/Streams/PubSub/PubSubRendezvousGrain.cs
@@ -316,7 +316,7 @@ namespace Orleans.Streams
             if (grainRef != null)
             {
                 List<StreamSubscription> subscriptions =
-                    State.Consumers.Where(c => !c.IsFaulted && c.Consumer.Equals(streamConsumer))
+                    State.Consumers.Where(c => !c.IsFaulted && c.consumerReference.Equals(streamConsumer))
                         .Select(
                             c =>
                                 new StreamSubscription(c.SubscriptionId.Guid, streamId.ProviderName, streamId,


src/Orleans.Runtime/Streams/PubSub/PubSubRendezvousGrain.cs
```diff
@@ -316,7 +316,7 @@ public Task<List<StreamSubscription>> GetAllSubscriptions(StreamId streamId, ISt
             if (grainRef != null)
             {
                 List<StreamSubscription> subscriptions =
-                    State.Consumers.Where(c => !c.IsFaulted && c.Consumer.Equals(streamConsumer))
```
xiazen: Correct me if I'm wrong.` c.Consumer` is of type `IStreamConsumerExtension`, and  so is `streamConsumer`. Why change it to consumerReference?

bob-dawson: No, streamConsumer is of type stream reference.  And because Consumer is marked json ignore, when reactivated, it's value ia null.

bob-dawson: I made a mistake. You're correct.

bob-dawson: This is the problem, I'm doing a custom mongodb store provider, when I run into this line, the State.Consumers Consumer property is null, but consumerReference is not.
for the code:
 public IStreamConsumerExtension Consumer { get { return consumerReference as IStreamConsumerExtension; } }
which means consumerReference is just a plain grainReference. I find the implementation of IStreamConsumerExtension,  only find the autogenerated OrleansCodeGenStreamConsumerExtensionReference and StreamConsumerExtension class.  Does it mean that consumerReference should be OrleansCodeGenStreamConsumerExtensionReference?  But when I debug, it's not.  Is it the codegen problem, or my serializsation problem?

bob-dawson: This is what i found in the mongodb, for the value of a PubSubRendezvousGrain
{
    "Producers": [{
        "Stream": {
            "Guid": "00000000-0000-0000-0000-000000000000",
            "ProviderName": "SimpleStreamProvider",
            "Namespace": "StateEventStream"
        },
        "producerReference": {
            "GrainId": "00000000000000000000000000000000060000000b30fff1+Master",
            "GenericArguments": ""
        }
    }, {
        "Stream": {
            "Guid": "00000000-0000-0000-0000-000000000000",
            "ProviderName": "SimpleStreamProvider",
            "Namespace": "StateEventStream"
        },
        "producerReference": {
            "GrainId": "00000000000000000000000000000000060000000b30fff1+Master",
            "GenericArguments": ""
        }
    }],
    "Consumers": [{
        "SubscriptionId": {
            "Guid": "de1b4bdc-ef21-44cf-9bd9-a1b2d273497d"
        },
        "Stream": {
            "Guid": "00000000-0000-0000-0000-000000000000",
            "ProviderName": "SimpleStreamProvider",
            "Namespace": "StateEventStream"
        },
        "consumerReference": {
            "GrainId": "0000000000000000000000000000000003ffffffddace0dd",
            "GenericArguments": ""
        },
        "filterWrapper": null,
        "state": 0
    }, {
        "SubscriptionId": {
            "Guid": "9a9577ed-6109-417e-b46d-92b853e64b27"
        },
        "Stream": {
            "Guid": "00000000-0000-0000-0000-000000000000",
            "ProviderName": "SimpleStreamProvider",
            "Namespace": "StateEventStream"
        },
        "consumerReference": {
            "GrainId": "00000000000000000000000000000000060000000b30fff1+Master",
            "GenericArguments": ""
        },
        "filterWrapper": null,
        "state": 0
    }]
}

bob-dawson:
I cleared the old data, and build from source, now it's OK.

PR #3679, opened 2017-11-17
Make OnSetupState overrideable.

SebastianStehle:
I would like to catch errors in the initial setup of the state.

I have not found an easy way in the current implementation. Instead I had to make a V2 Grain:

https://gist.github.com/SebastianStehle/21cf042385ed61fd8cdcda8dd8b6682f

### DIFF
--- a/src/Orleans.Core.Abstractions/Core/Grain.cs
+++ b/src/Orleans.Core.Abstractions/Core/Grain.cs
@@ -325,7 +325,7 @@ namespace Orleans
             Stopwatch sw = Stopwatch.StartNew();
             try
             {
-                await this.ReadStateAsync();
+                await this.OnSetupStateAsync();
                 sw.Stop();
                 // TODO: find a way to reenable StorageStatisticsGroup here
                 //StorageStatisticsGroup.OnStorageActivate(grainTypeName, sw.Elapsed);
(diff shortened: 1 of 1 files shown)


benjaminpetit:
For your use case, overriding `ReadStateAsync()` is not enough?

SebastianStehle:
No, because when it fails you cannot react on it. The grain cannot be activated and thats it.

benjaminpetit:
Not sure to follow you here?

Are you worried that the call to `Grain<TGrainState>.OnSetupState` might fail or just the first call to the storage with ``Grain<TGrainState>ReadStateAsync`?

SebastianStehle:
My UseCase: The state structure has changed but I know how to fix it (because I have a CQRS App and have the original events). When I activate the grain the OnSetup method is called and the deserialization fails. Tzerefore my grain cannot be activated. 

One solution would be to rename the grain, e.g. suffix it with V2 or so, bu then I have a lot of old states in my system. My solution would be to set to state to null and let my process handle it.

benjaminpetit:
Maybe that's not ideal, but with the current implementation in Orleans, you could just do that in your grain:

```csharp
protected override Task ReadStateAsync()
{
    try
    {
        await base.ReadStateAsync();
    }
    catch
    {
        // Your error handling code here
    }
}
```

Or am I missing something?

SebastianStehle:
Sure, soooo stupid :D

benjaminpetit:
Ahh I thought I was missing something :) Glad that you have a solution!

PR #3682, opened 2017-11-17
CodeGen: Allow per-project log level configuration

ReubenBond:
Adds a property which can be set in user's .csproj files to configure the log level of the build-time code generator. Log levels are defined in the `LogLevel` enum from `Microsoft.Extensions.Logging`. If the user specifies an invalid log level, they'll see a message like this in the build log: `error : "Trakjjkjkce" is not a valid log level. Valid values are Trace, Debug, Information, Warning, Error, Critical, None` and the build will fail. Default level is `Warning`.

Example, to set log to `Trace`.
```xml
<PropertyGroup>
    <OrleansCodeGenLogLevel>Trace</OrleansCodeGenLogLevel>
</PropertyGroup>
```

### DIFF
--- a/src/Orleans.CodeGeneration.Build/CodeGenOptions.cs
+++ b/src/Orleans.CodeGeneration.Build/CodeGenOptions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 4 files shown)


xiazen:
LGTM

PR #3683, opened 2017-11-18
Move queuing logic from AsynchQueueAgent into Executor

dVakulen:
Continuation of https://github.com/dotnet/orleans/pull/3663

### Changes description:
 - Introduced `IExecutor` interface with work queuing API compatible to .NET `ThreadPool`.
-  `ExecutorService` now returns suitable executors based on stage type. In future  `ExecutionPlan` could be introduced, allowing fine grained control over executor - stage assignment.
 - Moved queuing logic (blocking collection, its processing loop and statistics gathering) from `AsynchQueueAgent` into newly created `QueuedExecutor`. It resulted in separation of descendants of `AsynchAgent` and `AsynchQueueAgent`, as the former one schedules single long running work on start, while the latter - multiple short running tasks during execution. Stages with single task were made inheritors of  new `SingleTaskAsynchAgent` which contains extracted from `AsynchAgent` thread processing logic.
 - Due to move of statistics gathering into executors - `Message` is no longer implements `ITimeInterval` which was used for calculation of time spent in queues. Instead this measure is now being tracked inside `QueuedExecutor`.

Unification with `WorkQueue ` is matter for a separate PR. 

Suggestions on better names are welcomed, such as `AsynchAgent` -> `StageDefinition` (this one could go into global rename / refactor PR after), `QueuedExecutor` -> `ThreadPoolExecutor`, etc.

### DIFF
--- a/src/Orleans.Core/Messaging/GatewayConnection.cs
+++ b/src/Orleans.Core/Messaging/GatewayConnection.cs
@@ -28,6 +28,7 @@ namespace Orleans.Messaging
                 Silo = addr.ToSiloAddress();
             }
         }
+
         internal SiloAddress Silo { get; private set; }
 
         private readonly GatewayClientReceiver receiver;
@@ -72,7 +73,6 @@ namespace Orleans.Messaging
             IsLive = false;
             receiver.Stop();
             base.Stop();
-            DrainQueue(RerouteMessage);
             MsgCenter.RuntimeClient.BreakOutstandingMessagesToDeadSilo(Silo);
             Socket s;
             lock (Lockable)
@@ -83,6 +83,19 @@ namespace Orleans.Messaging
             if (s == null) return;
 
             CloseSocket(s);
+        }
+
+        protected override void Process(Message msg)
+        {
+            // not threadsafe check, but behavior is preserved
+            if (!IsLive)
+            {
+                RerouteMessage(msg);
+            }
+            else
+            {
+                base.Process(msg);
+            }
         }
 
         // passed the exact same socket on which it got SocketException. This way we prevent races between connect and disconnect.
@@ -98,7 +111,7 @@ namespace Orleans.Messaging
                     s = Socket;
                     Socket = null;
                     Log.Warn(ErrorCode.ProxyClient_MarkGatewayDisconnected, String.Format("Marking gateway at address {0} as Disconnected", Address));
-                    if ( MsgCenter != null && MsgCenter.GatewayManager != null)
+                    if (MsgCenter != null && MsgCenter.GatewayManager != null)
                         // We need a refresh...
                         MsgCenter.GatewayManager.ExpediteUpdateLiveGatewaysSnapshot();
                 }
--- a/src/Orleans.Core/Messaging/GatewayConnection.cs
+++ b/src/Orleans.Core/Messaging/GatewayConnection.cs
@@ -87,7 +87,6 @@ namespace Orleans.Messaging
 
         protected override void Process(Message msg)
         {
-            // not threadsafe check, but behavior is preserved
             if (!IsLive)
             {
                 RerouteMessage(msg);
--- a/src/Orleans.Core/Messaging/GatewayConnection.cs
+++ b/src/Orleans.Core/Messaging/GatewayConnection.cs
@@ -12,7 +12,7 @@ namespace Orleans.Messaging
     /// 
     /// Note that both sends and receives are synchronous.
     /// </summary>
-    internal class GatewayConnection : OutgoingMessageSender
+    internal class GatewayConnection : OutgoingMessageSender, IQueueDrainable
     {
         private readonly MessageFactory messageFactory;
         internal bool IsLive { get; private set; }
@@ -87,6 +87,7 @@ namespace Orleans.Messaging
 
         protected override void Process(Message msg)
         {
+            // After stop GatewayConnection needs to reroute not yet sent messages to another gateway 
             if (!IsLive)
             {
                 RerouteMessage(msg);
--- a/src/Orleans.Core/Runtime/ExecutorService.cs
+++ b/src/Orleans.Core/Runtime/ExecutorService.cs
@@ -1,15 +1,41 @@
-﻿using System.Threading.Tasks;
-using Orleans.Runtime.Scheduler;
+﻿using System;
+using System.Threading;
 
 namespace Orleans.Runtime
 {
-    internal class ExecutorService : ITaskScheduler
+    internal interface IExecutor
     {
-        private readonly TaskScheduler taskScheduler = new ThreadPerTaskScheduler(task => (task as AsynchAgentTask)?.Name);
+        void QueueWorkItem(WaitCallback callBack, object state = null);
 
-        public void RunTask(Task task)
+        int WorkQueueLength { get; }
+    }
+
+    internal class ExecutorService
+    {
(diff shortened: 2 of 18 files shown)


src/Orleans.Runtime/Messaging/IncomingMessageAgent.cs
```diff
@@ -42,48 +42,30 @@ public override void Start()
 
         protected override void Run()
         {
-            try
```
dVakulen: Shifted due to to part of statistics tracking with try .. finally removal.

src/Orleans.Core/Runtime/ExecutorService.cs
```diff
@@ -1,15 +1,54 @@
...
 
-        public void RunTask(Task task)
+        int WorkQueueCount { get; }
+    }
+
+    internal class ExecutorService
+    {
+        public IExecutor GetExecutor(GetExecutorRequest request)
+        {
+            if (typeof(GatewayConnection).IsAssignableFrom(request.StageType))
```
ReubenBond: I find this a bit strange, as though `GetExecutor` has to know too much about its caller. Could there be a different way to identify the required kind of `IExecutor` which still separates policy & mechanism?

src/Orleans.Core/Runtime/ExecutorService.cs
```diff
@@ -1,15 +1,54 @@
...
+            }
+
+            if (typeof(SingleTaskAsynchAgent).IsAssignableFrom(request.StageType))
+            {
+                return new ThreadPerTaskExecutor(request.StageName);
+            }
+
+            return ConstructQueuedExecutor()(false);
+
+            Func<bool, QueuedExecutor> ConstructQueuedExecutor()
```
ReubenBond: This method might be clearer if it just returns the executor immediately, like this:
```C#
QueuedExecutor ConstructQueuedExecutor(bool drainAfterCancel)
{
  return new QueuedExecutor(request.StageName, request.CancellationTokenSource, drainAfterCancel);
}
```

And then call it specifying the argument name, `drainAfterCancel`.

Alternatively, DrainAfterCancel might be a good candidate to move into the request, since it seems like it's important for the caller to know what the behavior is.

src/Orleans.Core/Messaging/GatewayConnection.cs
```diff
@@ -85,6 +85,18 @@ public override void Stop()
...
+        {
+            if (!IsLive)
+            {
+                RerouteMessage(msg);
+            }
+            else
+            {
+                base.Process(msg);
+            }
+        }
```
ReubenBond: Note: this logic is in lieu of `DrainQueue`

src/Orleans.Core/Runtime/QueuedExecutor.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                if (!drainAfterCancel && cancellationTokenSource.IsCancellationRequested)
+                {
+                    return;
+                }
+
+                QueueWorkItemCallback workItem;
+                try
+                {
+                    workItem = workQueue.Take();
```
ReubenBond: The old `DrainQueue` logic used `TryTake`  whereas this uses `Take`. I suppose the idea is to make the queue throw `InvalidOperationException` when `Take` is called on the empty (and completed) queue. In Orleans today, I often see a bunch of those exceptions when a silo is being killed. Ultimately I think there's little that could be done about it except for maybe injecting one final no-op work item after the CTS has been canceled so threads get a chance to exit gracefully.

I'm ok with no-action here

src/Orleans.Core/Runtime/QueuedExecutor.cs
```diff
@@ -0,0 +1,157 @@
...
+            cancellationTokenSource = cts;
+            cancellationTokenSource.Token.Register(() => workQueue.CompleteAdding());
+            new ThreadPerTaskExecutor(name).QueueWorkItem(_ => ProcessQueue());
+        }
+
+        public int WorkQueueCount => workQueue.Count;
+
+        public void QueueWorkItem(WaitCallback callback, object state = null)
+        {
+            var workItemCallback = new QueueWorkItemCallback(callback, state);
```
ReubenBond: (out of scope) could these benefit from pooling?

src/Orleans.Core/Runtime/SingleTaskAsynchAgent.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+        }
+
+        public override void OnStart()
+        {
+            executor.QueueWorkItem(_ => AgentThreadProc(this));
+        }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
+        private void AgentThreadProc(Object obj)
```
ReubenBond: This could be made static, since `obj` = `this`, otherwise no need to use `obj` at all. If using `obj`, it should be passed as the context parameter to `QueueWorkItem` rather than closed over.

None of this matters from a perf perspective since this is called so infrequently.

src/Orleans.Core/Runtime/ExecutorService.cs
dVakulen: Indeed, the `drainAfterCancel` usage here is just leaking knowledge about `GatewayConnection` implementation.

src/Orleans.Core/Runtime/QueuedExecutor.cs
dVakulen: It was modeled after `QueueUserWorkItemCallback`  from https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/ThreadPool.cs#L897, which isn't being pooled most likely to more complicated structure and costs of contention on thread safe pool. For our case  - access will be in many cases single threaded (except part with WorkQueue), so it could be pooled easily.

src/Orleans.Core/Runtime/SingleTaskAsynchAgent.cs
dVakulen: Fixed

src/Orleans.Core/Runtime/QueuedExecutor.cs
dVakulen: Added no-op action on cancel.

src/Orleans.Core/Messaging/GatewayConnection.cs
dVakulen: Added comment.

src/Orleans.Core/Runtime/ExecutorService.cs
dVakulen: If to make completely reusable component - that could be a problem. But I think for this case, where list of stages is known beforehand and almost constant - this knowledge will allow greater flexibility in assigning executors to stages. Something alike was implemented in ElasticSearch - 
https://github.com/elastic/elasticsearch/blob/8caf7d4ff8738131ae65cca86869c904d0e5f19b/core/src/main/java/org/elasticsearch/threadpool/ThreadPool.java#L123

ReubenBond:
There's still a change we're waiting on: to allow graceful shutdown of queued agents.

jdom:
Any concerns @gabikliot / @sergeybykov ? Is the new naming acceptable?

sergeybykov:
Thank you, @dVakulen!

PR #3684, opened 2017-11-18
fix null reference exception when unit testing grains that use streams

timbarker:
When unit testing a grain that uses streams a NullReferenceException will occur as the IActivationData on the grain will be null. This appears to be introduced with 2.0 beta 1. 

I have put a null check on the Grain IActivationData so that under unit testing scenarios it will defer to the injected IGrainRuntime to get a ServiceProvider.


sergeybykov:
Thank you, @timbarker!

PR #3688, opened 2017-11-20
Allow for multiple fallback serializers in a backwards compatible manner

### LINKED ISSUE #3640: Exceptions serialization error 2.0 beta1
After the update to .NET Core 2 with Orleans, 2.0 beta-1, we are getting some issues with the exceptions, is there a way to mark them as serializable?

This is the error we are getting:

```cs
Orleans.Runtime.OrleansException: No copier found for the object of type System.ArgumentNullException. 
Perhaps you need to mark it [Serializable] or define a custom serializer for it?
```

For our own exceptions, we could modify the code to get them working, however for the ones we don't have control over I'm not sure what to do. Any suggestions?

ReubenBond:
Fixes #3640

When moving to .NET Core 2.0, we removed the `#if` guards which set the default fallback serializer to `ILBasedSerializer` on .NET Core. This meant that the default fallback serializer for all frameworks would be `BinaryFormatterSerializer`.

I was fine with this decision (I may have even suggested it), but it turns out that `BinaryFormatter` is unable to serialize many important types (such as many *exceedingly common* exception types) in .NET Core 2.0. The result is that the current system is unusable on .NET Core 2.0 without a workaround (in linked issue).

This PR adds a new kind of serializer called `IKeyedSerializer` which is an internal subtype of `IExternalSerializer`, adding a `SerializerId` field. When the serializer is used, its `SerializerId` is serialized first in order to later identify which serializer to use during deserialization. This is in contrast to `SerializationTokenType.Fallback` which necessarily points to the `FallbackSerializerType` configuration value.

The result is that we can keep the current default fallback serializer and can safely add additional `IKeyedSerializer` implementations with well-defined ids when necessary. This helps with backwards compatibility because users can still deserialize values which were serialized using the fallback serializer.

PR also adds initial .NET Core test project with a single test for this case. In future we can port all tests to multi-target (somehow) and remove that project.

* Set `CurrentObjectOffset`  based on parent offset when creating `NestedDeserializationContext`
* Pass correct context in calls to `SerializeInner`/`DeserializeInner` in nested contexts
* Fix `ILBasedSerializer` test which was checking for incorrect behavior.
* Use `RuntimeTypeFormatter` to serialize types instead of `AssemblyQualifiedName`


src/OrleansProviders/Storage/HierarchicalKeyStore.cs
```diff
@@ -7,9 +7,7 @@
 
 namespace Orleans.Storage
 {
-    internal class HierarchicalKeyStore :
-        MarshalByRefObject,
-        ILocalDataStore
+    internal class HierarchicalKeyStore : ILocalDataStore
```
jdom: If I'm not mistaken, this had MarshalByRefObject because it was being used by a test via the app domain test hook... Was the test re-written? Good, it was my intention to do it at some point, but just wanted to be sure you searched for it

test/NetCore.Test/ExceptionTests.cs
```diff
@@ -0,0 +1,33 @@
...
+    [Trait("Category", "BVT")]
+    public class ExceptionTests
+    {
+        private readonly ISiloHost silo;
+        private readonly IClusterClient client;
+
+        public ExceptionTests()
+        {
+            this.silo = SiloHostBuilder.CreateDefault().AddApplicationPartsFromAppDomain().ConfigureLocalHostPrimarySilo().Build();
+            this.silo.StartAsync().GetAwaiter().GetResult();
```
jdom: This is leaking silos and clients. Be sure to make the class IDisposable and always stop them

test/TestLibrary/ExceptionGrain.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+using System.Threading.Tasks;
+using Orleans;
+
+namespace TestLibrary
```
jdom: What is this loose file without any project?

jdom:
If I understand correctly, this will change to use `ILBasedSerializer` as the default "fallback" serializer when there is no serializer for the type, correct? I understand there is still a subsequent chance to use the configure fallback serializer, but that path would mostly be hit when deserializing old data, but new data will always serialize with the ILBased one...

If that's a correct assumption, follow up question: is it always recommended to use the ILBasedSerializer instead of BinaryFormatter, even for types that were custom crafted to work well with the binary formatter? If I remember correctly, you would honor the serialization constructor if it's there, but want to double check as a more general question

src/OrleansProviders/Storage/HierarchicalKeyStore.cs
ReubenBond: Yep, I searched for it and couldn't find it - but of course it's possible I missed it.
The reason for this change, btw, was the disparity between `MarshalByRefObject` on CoreFx and FullFx. Because OrleansProviders was being built using the FullFx bootstrap code generator, it had the FullFx fields in codegen. This then caused an NRE when trying to load initialize the serializer on CoreFx

test/TestLibrary/ExceptionGrain.cs
ReubenBond: oops, meant to be deleted, thanks

ReubenBond:
@jdom the assumption is correct. For the follow up, we don't respect the constructor. The plan to tackle types which are `ISerializable` was to add explicit support for it using another keyed serializer: https://github.com/ReubenBond/orleans/commit/3018a4d24de5cc000d117f21fa7ce896d62ec3e4

There's also the case where a type is built to work with serialization callbacks but is not `ISerializable`. This is the case for some F# types (`FSharpSet<T>` for example). In those cases I'm thinking of adding support for those callbacks to `ILBasedSerializer`, taking it from `DotNetSerializableSerializer` in the linked commit

src/OrleansProviders/Storage/HierarchicalKeyStore.cs
jdom: I'm fine removing this here if it is no longer needed, but regarding the NRE, is there something we should do there? Something like lazily loading the serializers, similar to what we do with known types? Because I'm pretty sure we just generated a serializer for this, but the user didn't have any intention of using it (and in fact, we used this for remoting to an AppDomain, completely outside of Orleans SM).

test/NetCore.Test/ExceptionTests.cs
```diff
@@ -29,5 +30,16 @@ public ExceptionTests()
             var exception = await Assert.ThrowsAsync<InvalidOperationException>(() => grain.ThrowsInvalidOperationException());
             Assert.Equal("Test exception", exception.Message);
         }
+
+        public void Dispose()
+        {
+            var cancel = new CancellationTokenSource();
+            cancel.Cancel();
+            this.silo.StopAsync(cancel.Token).GetAwaiter().GetResult();
+            this.silo?.Dispose();
```
jdom: It's odd that you are using the null propagator here, but not above. Shouldn't it use it everywhere? Same with the `client.Abort()` call

test/NetCore.Test/NetCore.Test.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
jdom: We probably want to run this in CI. That means adding it to `Test.cmd`
We might also want to include this in the linux build (at least building, and eventually running the tests too)

test/NetCore.Test/ExceptionTests.cs
ReubenBond: ReSharper did it when generating the method. I'll remove them

ReubenBond: Or I'll add them to the other locations. It's possible that this class would initialize a silo and fail to initialize a client. In that case, Dispose won't be called and the silo will live on in memory.

jdom: I meant you probably want to add them everywhere, since it might fail to start the silo or the client and never assign those fields

test/NetCore.Test/NetCore.Test.csproj
ReubenBond: Alright, I gave it a go so let's see what happens.

PR #3695, opened 2017-11-21
Add ArgumentNullException in Grain

seralexeev:
I guess `ArgumentNullException` is more suitable exception instead of `NullReferenceException` for case when developer does mistake and call Grain methods by passing null


sergeybykov:
Thank you, @seralexeev!

PR #3696, opened 2017-11-21
Use OptionsBuilder to allow chaining Configure calls for providers

jdom:
I temporarily copied the implementation of `OptionsBuilder` into the Orleans codebase, but the plan is to delete it once `Microsoft.Extensions.Options` v2.1 is released, and includes these. I put them in an Orleans namespace to avoid collisions if they co-exist for a short while.

For an explanation why `OptionsBuilder` is better than having extension method overloads for every variation of how to configure, see the PR itself: https://github.com/aspnet/Options/pull/238

This will even be a lot more useful once we start using named options for our named service providers configuration.

I migrated our current extension methods for the gateway list and membership providers to use this. I did preserve the overload that takes `Action<TOptions>`, and the reasoning is explained and shown in the second commit in this PR (let me know if you are OK with this or prefer not to have that overload, which is less boilerplate code for us, but potentially a harder to use API for the common case).

In the future, we can also port this PR that was merged today and builds on top of OptionsBuilder: https://github.com/aspnet/Options/pull/239 We already do a very similar thing [here](https://github.com/dotnet/orleans/blob/8f2181a31264e85d91223e492a1c7247dfcdd48b/src/Orleans.Runtime/Hosting/LegacyClusterConfigurationExtensions.cs#L84))

### DIFF
--- a/src/Orleans.Core/Configuration/ServiceCollectionExtensions.cs
+++ b/src/Orleans.Core/Configuration/ServiceCollectionExtensions.cs
@@ -90,6 +90,18 @@ namespace Orleans.Configuration
         /// <param name="configureOptions"></param>
         /// <returns></returns>
         public static IServiceCollection UseStaticGatewayListProvider(this IServiceCollection collection,
+            Action<StaticGatewayListProviderOptions> configureOptions)
+        {
+            return collection.UseStaticGatewayListProvider(ob => ob.Configure(configureOptions));
+        }
+
+        /// <summary>
+        /// Add an <see cref="StaticGatewayListProvider"/> and configure it using <paramref name="configureOptions"/>
+        /// </summary>
+        /// <param name="collection"></param>
+        /// <param name="configureOptions"></param>
+        /// <returns></returns>
+        public static IServiceCollection UseStaticGatewayListProvider(this IServiceCollection collection,
             Action<OptionsBuilder<StaticGatewayListProviderOptions>> configureOptions)
         {
             configureOptions?.Invoke(collection.AddOptions<StaticGatewayListProviderOptions>());
--- a/src/Orleans.Core/Configuration/ServiceCollectionExtensions.cs
+++ b/src/Orleans.Core/Configuration/ServiceCollectionExtensions.cs
@@ -2,7 +2,6 @@
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using System;
-using Microsoft.Extensions.Configuration;
 using Orleans.Configuration.Options;
 using Orleans.Hosting;
 using Orleans.Messaging;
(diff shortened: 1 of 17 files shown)


src/Orleans.Core/Hosting/Options/OptionsBuilder.cs
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
ReubenBond: Review note: See https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.Options/OptionsBuilder.cs

src/Orleans.Core/Hosting/Options/OptionsBuilderConfigurationExtensions.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
ReubenBond: From https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.Options.ConfigurationExtensions/OptionsBuilderConfigurationExtensions.cs

src/Orleans.Membership.Consul/ConsulUtilsHostingExtensions.cs
```diff
@@ -49,67 +39,49 @@ public static class ConsulUtilsHostingExtensions
...
         public static IClientBuilder UseConsulGatewayListProvider(this IClientBuilder builder,
-            IConfiguration configuration)
+            Action<OptionsBuilder<ConsulGatewayListProviderOptions>> configureOptions)
         {
-            return builder.ConfigureServices(services => services.UseConsulGatewayListProvider(configuration));
+            return builder.ConfigureServices(services => services.UseConsulGatewayListProvider(configureOptions));
         }
 
         /// <summary>
         /// Configure DI container with ConsuleBasedMemebership
```
ReubenBond: typo on Consul

src/Orleans.Membership.ZooKeeper/ZooKeeperHostingExtensions.cs
```diff
@@ -25,21 +22,15 @@ public static class ZooKeeperHostingExtensions
         /// <summary>
         /// Configure siloHostBuilder with ZooKeeperMembership
```
ReubenBond: "Configures the silo to use ZooKeeper for cluster membership" - but it's perhaps out of scope for this pr to change the comment

src/Orleans.Core/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -91,23 +92,20 @@ public static IServiceCollection AddGrainCallFilter(this IServiceCollection coll
...
-                .Configure(configureOptions)
-                .AddSingleton<IGatewayListProvider, StaticGatewayListProvider>();
+            return collection.UseStaticGatewayListProvider(ob => ob.Configure(configureOptions));
         }
 
         /// <summary>
-        /// Add an <see cref="StaticGatewayListProvider"/> and configure it using <paramref name="configuration"/>
+        /// Add an <see cref="StaticGatewayListProvider"/> and configure it using <paramref name="configureOptions"/>
         /// </summary>
         /// <param name="collection"></param>
```
ReubenBond: In other files, empty params were removed. Could do the same here

src/Orleans.Core/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -91,23 +92,20 @@ public static IServiceCollection AddGrainCallFilter(this IServiceCollection coll
...
+        /// <param name="configureOptions"></param>
         /// <returns></returns>
         public static IServiceCollection UseStaticGatewayListProvider(this IServiceCollection collection,
-            IConfiguration configuration)
+            Action<OptionsBuilder<StaticGatewayListProviderOptions>> configureOptions)
         {
-            return collection
-                .Configure<StaticGatewayListProviderOptions>(configuration)
-                .AddSingleton<IGatewayListProvider, StaticGatewayListProvider>();
+            configureOptions?.Invoke(collection.AddOptions<StaticGatewayListProviderOptions>());
```
ReubenBond: given this can be null, should the param have a default value of `null`?

ReubenBond: Actually, scrap that - a default value would make overload resolution ugly.

jdom: I'm not really making that call in this PR, just porting it. Although responding to this particular thing, it might be that we don't really want null, and potentially throw an argument exception... I don't really know at this point which behavior to choose, so I'll leave it undefined for now, unless you already thought through it and can recommend one

ReubenBond:
@dotnet-bot test bvt

jdom:
@dotnet-bot test bvt

Issue #3507, opened 2017-10-06
What is the purpose of specifying multiple seed nodes?


Kimserey:
In my development scenario with 2 silos, I am using MembershipTableGrain as Livenesstype and have only 1 seed node pointing to my primary node with the second silo being secondary node.

I can see in Globals that seed node is an array of addresses therefore **what is the purpose of specifying multiple seed nodes?**

sergeybykov:
Support for multiple seed nodes is there for historical reasons. In early prototypes of Orleans we had that. I'm almost sure it doesn't work anymore with more than one seed (a.k.a. primary) node configured. Should probably remove that in 2.0.

xiazen:
It is confusing indeed. So We are thinking about reducing the confusion there by changing `List<SeedNode> SeedNodes` to be `IpEndpoint SeedNode` in 2.0. 

But we are also migrating off GlobalConfiguration in 2.0, so that GlobalConfiguration will become obsolete after 2.0. So there maybe not much value in 2.0 to fix it. 

Or there's a third option: update the documentation on SeedNodes and warn user to not use more than one seed nodes. 

Which one do you like personally?

Kimserey:
> Support for multiple seed nodes is there for historical reasons. In early prototypes of Orleans we had that. 

I got confused because I thought that if the seed node stated as primary would go down, the membership state together with the primary status would be passed to the next seed node alive.

> Or there's a third option: update the documentation on SeedNodes and warn user to not use more than one seed node.

I think this would be enough since GlobalConfiguration will be obsolete.

gabikliot:
What Sergey said: we should remove seed nodes and move to a single primary node.

xiazen:
But `GlobalConfiguration` are going to be deprecated in 2.0, its settings will be break down into a collection of smaller options (Microsoft.Extensions.Options) .  Do you still think it is worthy to fix it?

sergeybykov:
Since we are deprecating `GlobalConfiguration` in 2.0, it doesn't seem like a worthwhile effort to change it with regards to seed/primary nodes.

PR #3703, opened 2017-11-22
Enable grain based membership without the need for ClusterConfiguration

jdom:
Create strongly typed options for `DevelopmentMembershipOptions`
Also split networking-related options from `MessagingOptions`


src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -484,9 +485,10 @@ public SiloAddress CalculateTargetSilo(GrainId grainId, bool excludeThisSiloIfSt
                     string grainName;
                     if (!Constants.TryGetSystemGrainName(grainId, out grainName))
                         grainName = "MembershipTableGrain";
-                    
-                    var errorMsg = grainName + " cannot run without Seed node - please check your silo configuration file and make sure it specifies a SeedNode element. " +
-                        " Alternatively, you may want to use AzureTable for LivenessType.";
+
+                    var errorMsg = $"{grainName} cannot run without Seed node. Please check your silo configuration make sure it specifies a SeedNode element. " +
```
ReubenBond: Seed node here is really Primary node, right? And there's at most one? The concept of having multiple was never implemented afaik

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
```diff
@@ -92,17 +93,38 @@ public static ISiloHostBuilder ConfigureLocalHostPrimarySilo(this ISiloHostBuild
...
+        public static ISiloHostBuilder UseGrainBasedMembership(this ISiloHostBuilder builder, Action<GrainBasedMembershipOptions> configureOptions)
         {
-            return builder.ConfigureServices(services => services.UseGrainBasedMembership());
+            return builder.ConfigureServices(services => services.UseGrainBasedMembership(configureOptions));
+        }
+
+        /// <summary>
+        /// Configure silo to use GrainBasedMembership
+        /// </summary>
+        public static ISiloHostBuilder UseGrainBasedMembership(this ISiloHostBuilder builder, Action<OptionsBuilder<GrainBasedMembershipOptions>> configureOptions)
```
ReubenBond: UseGrainBasedClustering? Maybe there's a friendlier name which also indicates that it shouldn't really be used in production when resilience is important, like UseDevelopmentClustering?

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
jdom: Yes, I validated that there current be more than 1 seed node, and it's indeed the same as the primary. Right now it's really messed up, but we are getting rid of this concept. See https://github.com/dotnet/orleans/issues/3507

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
jdom: Yes, I think I would prefer something more obvious than grain based. Nevertheless, I understand we are changing the nuget packages for membership and gateway list providers to be in a "Clustering" package, but I though this change didn't affect the implementations?
If I use Clustering for what we currently have for membership, how do we disambiguate with gateway list providers?

src/Orleans.Runtime/MembershipService/ILegacyMembershipConfigurator.cs
```diff
@@ -56,7 +56,15 @@ private class LegacyGrainBasedMembershipConfigurator : ILegacyMembershipConfigur
...
             {
-                services.UseGrainBasedMembership();
+                services.UseGrainBasedMembership(options => CopyGlobalGrainBasedMembershipOptions(configuration, options));
+            }
+
+            private static void CopyGlobalGrainBasedMembershipOptions(GlobalConfiguration configuration, GrainBasedMembershipOptions options)
+            {
+                if (configuration.SeedNodes?.Count > 0)
+                {
+                    options.SeedNode = configuration.SeedNodes?.FirstOrDefault()?.ToGatewayUri();
```
ReubenBond: ToGatewayUri here seems a bit odd to me. I think the `SeedNode` property should be a `SiloAddress` instead of a `Uri`

test/TestExtensions/DefaultClusterFixture.cs
```diff
@@ -22,7 +22,6 @@ public ISiloHostBuilder CreateSiloBuilder(string siloName, ClusterConfiguration
                 return new SiloHostBuilder()
                     .ConfigureSiloName(siloName)
                     .UseConfiguration(clusterConfiguration)
-                    .UseGrainBasedMembership()
```
xiazen: if we are removing this, then we should add a test to configure GrainBasedMemebership in the new way. Otherwise there's no test using testing `UseGrainBasedMembership`

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
jdom: I'd say that whatever we decide regarding Clustering vs Membership, we should do it in a different pass.
Still, I'd like to address the other part of this feedback in this PR, should I call it `UseDevelopmentMembership`? `UseNonReliableMembership`?
I think for transactions we already started calling some things Development, but the extension method was kept as `UseInMemoryTransactionLog()`.
What do you guys say we should do here?

test/TestExtensions/DefaultClusterFixture.cs
jdom: Yes, I'm aware of that, and agree, but the soon to come PR that revamps test cluster will be using it extensively

src/Orleans.Runtime/MembershipService/ILegacyMembershipConfigurator.cs
jdom: yeah, I'm going back and forth, the thing is that it is not serializable as a string, so you cannot set it from external config...

ReubenBond: can we somehow fix that? Implicit constructor?

ReubenBond: Given we will never know the `generation`, we could use IPEndpoint

jdom: with `IPEndpoint` we can't use implicit constructor for sure though. With SiloAddress it *might* work, but it's true that there is no generation info.
I could use a `string` as well, but it's less clear what is the expected format

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
xiazen: In transaction, there's a namespace `Orleans.Transactions.Development` contains stuff we recommended to use in Development, and also added in xml comments of "Recommended for Development only"

It is a bit weird to add intended use to the method name. It differs from our naming convention on membership too. All of our membership is named with the technology used behind it.

ReubenBond: Whatever we choose, I hope is implies this to developers that this is inherently not fault tolerant and is primarily intended for dev/test scenarios.

If we say `UseInMemoryClustering()`, they might assume it's in-memory but replicated (which is valid and relatively fault-tolerant). If we say `UseDevelopmentClustering()` there's the chance they will be scared to use it even in cases where they have a fixed set of on-prem servers. I'm *ok* with that.

Maybe someone else has a suggestion?

jdom: Yes, I agree with Reuben, this is not similar to the InMemory stuff we have even. So I'll go with `UseDevelopmentMembership` for now (that we can then rename to `UseDevelopmentClustering` if that's what we decide for all providers)

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -101,7 +101,7 @@ internal class LocalGrainDirectory :
             Factory<GrainDirectoryPartition> grainDirectoryPartitionFactory,
             RegistrarManager registrarManager,
             ExecutorService executorService,
-            IOptions<GrainBasedMembershipOptions> grainMembershipOptions,
+            IOptions<DevelopmentMembershipOptions> developmentOptions,
```
xiazen: Just to point out that DevelopmentMembership is a new name, which users are not familiar with. If we are going this way, should call it out in the docs or release note.

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -135,8 +135,8 @@ internal class LocalGrainDirectory :
                     loggerFactory);
             GsiActivationMaintainer = new GlobalSingleInstanceActivationMaintainer(this, this.Logger, globalConfig, grainFactory, multiClusterOracle, executorService, loggerFactory);
 
-            var seedNode = grainMembershipOptions.Value.SeedNode?.ToSiloAddress();
-            this.seed = this.MyAddress.Endpoint.Equals(seedNode?.Endpoint) ? this.MyAddress : seedNode;
+            var primarySiloEndPoint = developmentOptions.Value.PrimarySiloEndPoint;
```
xiazen: developmentOptions name is ambiguous, since it sounds like it includes anything development related,  can we call it developmentMembershipOptions?

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
```diff
@@ -93,33 +93,33 @@ public static ISiloHostBuilder ConfigureLocalHostPrimarySilo(this ISiloHostBuild
         /// <summary>
         /// Configure silo to use GrainBasedMembership
```
xiazen: Rename the xml too?

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
```diff
@@ -93,33 +93,33 @@ public static ISiloHostBuilder ConfigureLocalHostPrimarySilo(this ISiloHostBuild
...
         /// </summary>
-        public static IServiceCollection UseGrainBasedMembership(this IServiceCollection services, Action<GrainBasedMembershipOptions> configureOptions)
+        public static IServiceCollection UseDevelopmentMembership(this IServiceCollection services, Action<DevelopmentMembershipOptions> configureOptions)
         {
-            return services.UseGrainBasedMembership(ob => ob.Configure(configureOptions));
+            return services.UseDevelopmentMembership(ob => ob.Configure(configureOptions));
         }
 
         /// <summary>
         /// Configure silo to use Grain-based membership
```
xiazen: here too

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
jdom: but that one still uses the legacy-based configuration... it wasn't updated to use the new config

jdom: nevermind, I thought it was pointing to the ConfigureLocalHostPrimarySilo method... github confused me. I'll update

src/Orleans.Runtime/MembershipService/DevelopmentMembershipOptions.cs
```diff
@@ -0,0 +1,13 @@
+using System.Net;
+
+namespace Orleans.Hosting
+{
+    /// <summary>Configures development clustering options</summary>
+    public class DevelopmentMembershipOptions
+    {
+        /// <summary>
+        /// Gets or sets the seed node to find the membership system grain.
```
xiazen: "seed node" is still in xml comments. I think it should be primary silo node?

jdom: I wanted to keep this explicitly as a different way of referring to the same thing... otherwise if I just say primary again, it doesn't add anything extra

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
jdom: I honestly preferred developmentOptions, but I'll change it if you think the longest is preferable

src/Orleans.Runtime/MembershipService/DevelopmentMembershipOptions.cs
xiazen: that's fair

PR #3708, opened 2017-11-25
Improve samples: remove misleading code to configure ServicePointManager

lhotari:
- remove misleading code in samples to configure ServicePointManager
- [ServicePointManager.DefaultConnectionLimit gets set in Silo initialization](https://github.com/dotnet/orleans/blob/1591fd13a2868336bd111d01251f20aba503dbf3/src/Orleans.Runtime/Silo/Silo.cs#L657) with value from
  ClusterConfiguration.Defaults.DefaultConnectionLimit (defaults to 200)


sigmadon:
Should definitely remove some potential confusion. 

Would it be a good thing to add something to the docs describing the default values? Or are we too deep into details?

ReubenBond:
Thank you, @lhotari!

PR #3710, opened 2017-11-26
Enable TCP FastPath support

eisendle:
This pull request enables TCP Fastpath support to Orleans. Fastpath is a technique which speeds up loopback tcp connections on windows server 2012 / Windows 8 upwards. For more information see [here](https://blogs.technet.microsoft.com/wincat/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path/).
I did some benchmarks and enabling fastpath adds some minor response time & throughput improvements: 

| Fastpath enabled | Average latency (ms) | Throughput msg/sec |
|------------------|----------------------|--------------------|
| Yes              | 1.574                | 635                |
| No               | 1.671                | 598                |

This PR improves response time & throughput for co hosting scenarios until support for local silo clients is added (#3362).


benjaminpetit:
Always nice to improve response time and throughput, even for local-only scenario. I would feel more comfortable to add a config flag no enable this, but it might be overkill for this change...

I read the documentation, and they do not say if it can have an impact on non-loopback scenario?

eisendle:
The documentation [here](https://msdn.microsoft.com/en-us/library/windows/desktop/jj841212(v=vs.85).aspx) (see Remarks) states that setting SIO_LOOPBACK_FAST_PATH on non loopback sockets has no effect. Also Fastpath is enabled in Kestrel [#416](https://github.com/aspnet/KestrelHttpServer/pull/416) too. So the risk to break something should be minimal.

benjaminpetit:
Thanks @eisendle !

Issue #3715, opened 2017-11-27
Hardening Client-to-Silo and Silo-to-Silo connections


benjaminpetit:
Currently `IncomingMessageAcceptor` (used both in the Gateway and for Silo-to-Silo) is processing the "receive socket preamble" synchronously, without any enforced timeout. If someone open a connection without sending anything to a given silo, no other connection to this silo can be opened.

We should;

- [X] Set a timeout value for the synchronous socket read operations (#3716)
- [X] Run the (at least) `ReceiveSocketPreample` in another thread/task so Silo can process other connections (#3729)

PR #3716, opened 2017-11-27
Set a timeout value for the synchronous socket read operations

benjaminpetit:
First fix for #3715 

Basically reusing the connection timeout value from the configuration in the Socket configuration for synchronous call to `Socket.Receive` (when waiting connections preamble).

This do not interfere with the asynchronous receive that we do thereafter : https://msdn.microsoft.com/en-us/library/system.net.sockets.socket.receivetimeout(v=vs.110).aspx

> **Remarks**
> This option applies to synchronous Receive calls only

### DIFF
--- a/test/Tester/ClientConnectionTests/StallConnectionTests.cs
+++ b/test/Tester/ClientConnectionTests/StallConnectionTests.cs
@@ -1,0 +1,68 @@
+﻿using Orleans.Runtime;
+using Orleans.TestingHost;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading.Tasks;
+using TestExtensions;
+using Xunit;
+
+namespace Tester.ClientConnectionTests
+{
+    public class StallConnectionTests : TestClusterPerTest
+    {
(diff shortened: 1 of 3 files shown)


test/Tester/ClientConnectionTests/StallConnectionTests.cs
```diff
@@ -0,0 +1,68 @@
...
+
+                var mgmtGrain = this.Client.GetGrain<IManagementGrain>(0);
+                var hosts = await mgmtGrain.GetHosts();
+
+                Assert.Equal(2, hosts.Count);
+
+                stalledSocket.Disconnect(true);
+            }
+        }
+    }
```
ReubenBond: Will these tests break when you implement the second fix for that issue, where receiving the connection preamble is offloaded onto a background thread? Either way, that could be addressed later and maybe it's good if they break.

benjaminpetit: No it will not break this test. But we may modify these tests to check that we are no longer waiting the connection timeout before the second connection works

jdom:
@dotnet-bot test functional

Issue #3681, opened 2017-11-17
Consecutive transactions on grains can cause rollbacks


jsteinich:
Setup:

1. Start with the Account Transfer sample
2. Put the atm transfer (optionally the balance retrieval as well) inside a small for loop
3. Run

Observations:

- The ending balances don't always match what they should be.
- This is more apparent if adding `await Task.Delay(100)` at the end of the for loop
- [exampleOutput.txt](https://github.com/dotnet/orleans/files/1483434/exampleOutput.txt)

Theories:

It appears as though the `TransactionAgent`'s aborted transaction checks are too aggressive. When `TransactionalState` goes to get or save the current state, it goes through `Restore()` which ends up rolling back transactions that are still in the log but older than the abort threshold that `TransactionAgent` is using. Adding a delay that is longer than `TransactionAgent`'s processing interval causing more issues seems to back up the aggressive aborting since it ensures that the abort threshold is updated between transactions.

jason-bragg:
@jsteinich, can you provide your DoClientWork logic?
I have:

        private static async Task DoClientWork(IClusterClient client)
        {
            IATMGrain atm = client.GetGrain<IATMGrain>(0);
            Guid from = Guid.NewGuid();
            Guid to = Guid.NewGuid();
            for(int i=0; i<5; i++)
            {
                await atm.Transfer(from, to, 100);
                uint fromBalance = await client.GetGrain<IAccountGrain>(from).GetBalance();
                uint toBalance = await client.GetGrain<IAccountGrain>(to).GetBalance();
                Console.WriteLine($"\n\nWe transfered 100 credits from {from} to {to}.\n{from} balance: {fromBalance}\n{to} balance: {toBalance}\n\n");
            }
        }

and am not seeing what you are..?
Tnx

jason-bragg:
NM, I upped it to 10 loops and was able to repro.
Looking into it.
Thanks!

jason-bragg:
@jsteinich,
Plz see fix "Fixed intermittent transaction test failures. #3725"

Transactions is still in alpha, so if you want to pay with the latest, I suggest using the transaction benchmarks in Orleans/tests/Benchmarks.  

Thanks for helping us find this.

jsteinich:
Thanks for the fix.

I'm planning to use transactions quite a bit, so I'll probably look into running from master to continue testing.

jason-bragg:
@jsteinich, Great!  Please let us know if you see anything else unexpected.

[...]

PR #3725, opened 2017-11-30
Fixed intermittent transaction test failures.

jason-bragg:
Fixed intermittent transaction test failures.
Added new transaction test to reproduce community reported issue.

see Consecutive transactions on grains can cause rollbacks #3681

### DIFF
--- a/src/Orleans.Transactions/State/TransactionalState.cs
+++ b/src/Orleans.Transactions/State/TransactionalState.cs
@@ -82,7 +82,7 @@ namespace Orleans.Transactions
                 throw new OrleansReadOnlyViolatedException(info.TransactionId);
             }
 
-            Restore();
+            Rollback();
 
             var copiedValue = this.transactionCopy[info.TransactionId];
 
@@ -201,7 +201,7 @@ namespace Orleans.Transactions
                     bool success = await this.storageExecutor.AddNext(() => GuardState(() => PersistCommit(transactionId)));
                 } catch(Exception ex)
                 {
-                    this.logger.Error(OrleansTransactionsErrorCode.Transactions_CommitFailed, $"Commit {transactionId} failed.  Recovering.", ex);
+                    this.logger.Info(OrleansTransactionsErrorCode.Transactions_CommitFailed, "Commit {0} failed.  Recovering.  Exception: {1}", transactionId, ex);
                     // On error, queue up a recovery action.  Will do nothing if state recovers before this is processed
                     this.storageExecutor.AddNext(() => GuardState(() => Task.FromResult(true))).Ignore();
                     throw;
@@ -251,8 +251,6 @@ namespace Orleans.Transactions
                 HighestVersion = this.version,
                 HighestReadTransactionId = wlb
             };
-            this.metadata.HighestVersion = this.version;
-            this.metadata.HighestReadTransactionId = wlb;
             this.eTag = await this.storage.Persist(StateName, this.eTag, metadata.ToString(), pending);
             this.metadata = metadata;
 
@@ -305,7 +303,7 @@ namespace Orleans.Transactions
             this.eTag = await this.storage.Confirm(StateName, this.eTag, newMetadata.ToString(), stableversion.ToString());
             this.metadata = newMetadata;
             this.commitedState = stableState;
-            DoRecovery();
+            UpdateActiveState();
 
             return true;
         }
@@ -349,14 +347,14 @@ namespace Orleans.Transactions
 
             var info = TransactionContext.GetTransactionInfo();
 
-            Restore();
+            Rollback();
 
             if (this.transactionCopy.TryGetValue(info.TransactionId, out TState state))
             {
                 return state;
             }
 
-            this.logger.Debug("GetState {0}", info);
+            if (this.logger.IsEnabled(LogLevel.Debug)) this.logger.Debug("GetState {0}", info);
 
             if (!TryGetVersion(info.TransactionId, out TState readState, out TransactionalResourceVersion readVersion))
             {
@@ -416,7 +414,7 @@ namespace Orleans.Transactions
         /// <summary>
         /// Check with the transaction agent and rollback any aborted transaction.
         /// </summary>
-        private void Restore()
+        private void Rollback()
         {
             foreach (var transactionId in this.log.Keys)
             {
@@ -436,7 +434,7 @@ namespace Orleans.Transactions
             // Validate that we still have all of the transaction's writes.
             var validate  = this.log.TryGetValue(writeVersion.Value.TransactionId, out LogRecord<TState> logRecord) && logRecord.Version == writeVersion.Value;
             if(!validate && this.logger.IsEnabled(LogLevel.Debug))
-                this.logger.LogDebug($"ValidateWrite failed, because version is not the same as recorded in state log record of the same transaction, write version in the log record is {logRecord.Version}, version to be validated is {writeVersion.Value}");
+                this.logger.Debug($"ValidateWrite failed, because version is not the same as recorded in state log record of the same transaction, write version in the log record is {logRecord.Version}, version to be validated is {writeVersion.Value}");
             return validate;
         }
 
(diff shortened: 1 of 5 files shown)


src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -187,15 +190,18 @@ Task ITransactionalResource.Abort(long transactionId)
...
             if (transactionId > this.metadata.StableVersion.TransactionId)
             {
                 this.highCommitTransactionId = Math.Max(this.highCommitTransactionId, transactionId);
                 try
                 {
                     bool success = await this.storageExecutor.AddNext(() => GuardState(() => PersistCommit(transactionId)));
-                } catch(Exception)
+                } catch(Exception ex)
                 {
+                    this.logger.Error(OrleansTransactionsErrorCode.Transactions_CommitFailed, $"Commit {transactionId} failed.  Recovering.", ex);
```
ReubenBond: I guess it's a matter of opinion, but are failed commits errors? I think we can leave this on Error level for now, but failed commits are somewhat expected and normal.

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -207,7 +213,8 @@ Task ITransactionalResource.Abort(long transactionId)
         {
             if (!this.validState)
             {
-                await DoRecovery();
+                this.logger.Debug("Invalid state found. Recovering.");
```
ReubenBond: Maybe `IsEnabled` this

jason-bragg: We tend to do that when the string building is heavy due to complex arguments.  The logger should drop the message if it's disabled, so what we avoid with the pre check is ths cost of building the string, right?  or is there a cost here I'm not understanding.

src/Orleans.Transactions/State/TransactionalState.cs
jason-bragg: Probably should be info level, since system will retry.

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -231,14 +240,23 @@ Task ITransactionalResource.Abort(long transactionId)
...
+            {
+                StableVersion = this.metadata.StableVersion,
+                HighestVersion = this.version,
+                HighestReadTransactionId = wlb
+            };
             this.metadata.HighestVersion = this.version;
             this.metadata.HighestReadTransactionId = wlb;
-            this.eTag = await this.storage.Persist(StateName, this.eTag, this.metadata.ToString(), pending);
+            this.eTag = await this.storage.Persist(StateName, this.eTag, metadata.ToString(), pending);
+            this.metadata = metadata;
```
ReubenBond: I'm a little confused by this section. Is the issue that `this.metadata` is mutated in the background somehow? Otherwise, why create a new instance of it if we're mutating both instances in the same manner.

The answer is probably obvious, but I'm missing something.

src/Orleans.Transactions/State/TransactionalState.cs
xiazen: I think it should be Info level. Failed commit is expected and handled by tx system.

src/Orleans.Transactions/State/TransactionalState.cs
jason-bragg: nope, my mistake.  Intended to not change this.metadata unless the operation complete successfully.  Will fix.

src/Orleans.Transactions/State/TransactionalState.cs
ReubenBond: that's true, good point. Another way of looking at it is to always do the same thing for consistency, but I'm not firm on it either way.

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -321,15 +349,15 @@ private TState GetState()
...
-            this.logger.Debug("Read {0}", info);
-
             Restore();
 
             if (this.transactionCopy.TryGetValue(info.TransactionId, out TState state))
             {
                 return state;
             }
 
+            this.logger.Debug("GetState {0}", info);
```
ReubenBond: this one isn't free, so we can guard this one

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -441,33 +480,47 @@ private bool ValidateRead(long transactionId, TransactionalResourceVersion? read
...
                 return false;
             }
 
             return true;
         }
 
-        private async Task DoRecovery()
+        private async Task DoLoad()
         {
+            this.logger.LogDebug("DoLoad");
```
ReubenBond: nit: prefer `Debug` over `LogDebug`, which is just there for ease of porting

test/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
```diff
@@ -80,10 +80,44 @@ protected GoldenPathTransactionTestRunner(IGrainFactory grainFactory, ITestOutpu
...
+                if (expected != actual) this.output.WriteLine($"{grain} - failed");
                 Assert.Equal(expected, actual);
             }
         }
 
+        [SkippableTheory]
+        [InlineData(TransactionTestConstants.SingleStateTransactionalGrain)]
+        public virtual async Task RepeatGrainReadWriteTransaction(string transactionTestGrainClassName)
+        {
+            const int repeat = 10;
```
xiazen: maybe repeat 20? I know you reproduced this bug by just repeat 10 times, but this kind of behavior is hard to control, especially this will be running in CI, instead of your local machine. Running it 20 should be safer, more assuring.

test/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
```diff
@@ -80,10 +80,44 @@ protected GoldenPathTransactionTestRunner(IGrainFactory grainFactory, ITestOutpu
...
+            {
+                await coordinator.MultiGrainDouble(grains);
+
+                int expected = delta * (int)Math.Pow(2,i+1);
+                foreach (var grain in grains)
+                {
+                    int actual = await grain.Get();
+                    if (expected != actual) this.output.WriteLine($"{grain} - failed");
+                    Assert.Equal(expected, actual);
+                }
```
xiazen: > This is more apparent if adding await Task.Delay(100) at the end of the for loop

the user in the issue mentioned above , it seems adding that delay make the bug more apparent, I think we should add that in this loop too, ensuring what we did actually fix the bug .

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -484,7 +537,7 @@ private bool ValidateRead(long transactionId, TransactionalResourceVersion? read
             this.storage = storageFactory.Create<TState>(this.config.StorageName);
 
             // recover state
```
xiazen: fix the xml comment

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -441,33 +480,47 @@ private bool ValidateRead(long transactionId, TransactionalResourceVersion? read
...
+        private void DoRecovery()
+        {
+            this.logger.LogDebug("DoRecovery");
+            // load inital state
+            this.version = this.metadata.HighestVersion;
+            this.value = this.log.TryGetValue(this.version.TransactionId, out LogRecord<TState> record)
+                ? record.NewVal
+                : this.commitedState;
+            this.log.Clear();
+
```
ReubenBond: So when a grain is activated, we can set the current state to an unconfirmed value, which gives the transaction a chance to commit still. Is that the purpose of this change or is it something else?

ReubenBond:
@dotnet-bot test functional

test/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
jason-bragg: 10 has been sufficient and is still fast test, even against azure.  We can push higher if/when needed.

src/Orleans.Transactions/State/TransactionalState.cs
jason-bragg: This has two purposes.
1) Use common logic to set current state for both save and load.
2) In case of initial load, set the active state to the correct state (it was previously always being set to stable state which is only correct if there are no pending states.
- will cleanup up comments, as they are no longer correct.
- will consider procedure name change, as this is not really recovering.

ReubenBond: I think w.r.t naming, calling this "Recover" is fine, since it's common in fault tolerant systems to have a recovery process which runs on startup to bring the system to a consistent state.

PR #3727, opened 2017-11-30
[WIP] Optimize grain call with strongly typed arguments

sunnycase:
The pr is intended to minimize objects allocation and box / unbox when making grain calls by using strongly typed arguments.
This will also make arguments serialization / deserialization faster by providing expected types.


sunnycase:
@dotnet-bot test this please

sunnycase:
@ReubenBond  It seems the ci doesn't get triggered when there are new commits.

ReubenBond:
@sunnycase it might just be a temporary issue with the CI, eg it might be overloaded.

Thank you for this contribution! I've been thinking about this area for a while now and I'm still unsure what the best solution is. For example, what if we completely deferred deserialization of the arguments until the call site. In that case, we could deserialize them as `ValueTask<ArgOneType, ArgTwoType, etc>` and avoid allocations for methods with multiple parameters, also. That has the added benefit of serializing types more efficiently, too: in many cases it would allow the type to be serialized as `ExpectedType` (fastest) instead of `NamedType` (which includes the type name and requires a dictionary lookup).

Thank you again for this PR and please don't hesitate to message me on Gitter.

sunnycase:
@ReubenBond Maybe you mean `ValueTuple<ArgOneType, ArgTwoType, etc>` ?
I think it's a good idea, but copying struct is not free. We probably need a benchmark.

ReubenBond:
`ValueTuple<>`, my mistake - yes we need a benchmark. I'm running performance tests on your PR now.

ReubenBond:
@dotnet-bot test functional

Kingwl:
it's very helpful to me!

sunnycase:
@dotnet-bot test functional

ReubenBond:
@sunnycase I'm investigating the CI failure

sunnycase:
I have run some benchmarks on my old i3-3220 PC.
The ping test shows the pr will improve a little performance.

## before
`
Warm Run.
495493 calls in 30734ms.
16121.9821695842 calls per second.
`

## after
`
Warm Run.
502389 calls in 30635ms.
16399.183939938 calls per second.
`

sunnycase:
@ReubenBond 
I'm trying to implement your thought with strongly typed arguments. But there is an obstacle in implementing deffered deserialization:
In order to find the arguments' types, we should find the `interfaceId` and `methodId`, but now they are in `InvokeMethodRequest` which isn't deserialized yet. What do you think of it? Move them to the message headers?

I'll push current works has been done.

dVakulen:
The strongly typed grain call path was done as part of larger rewrite in https://github.com/dotnet/orleans/pull/2060 :  [GrainReference](https://github.com/dotnet/orleans/pull/2060/files#diff-64f4df81b54d326e17440c81957e905cR365),  [InsideRuntimeAgent](https://github.com/dotnet/orleans/pull/2060/files#diff-faf1fffc73efc52f0cc10c13764b9283R97), and it has shown up to 7% throughput increase.

sunnycase:
> The strongly typed grain call path was implemented as part of larger rewrite in #2060

@dVakulen But i still see `Task<T> InvokeMethodAsync<T>(int methodId, object[] arguments, InvokeMethodOptions options, SiloAddress silo)` signature in GrainReference.cs in that pr.

dVakulen:
@sunnycase That change allowed to avoid  `await` and task boxing - unboxing on grain method invocation. Removing `object []` parameters array was not the priority.

sunnycase:
@dVakulen So that means your pr avoid boxing on task result. But this pr is for avoiding boxing and serializing `NamedType` for arguments. They are different approaches.

dVakulen:
They are indeed. With different amounts of required effort and expected benefits.

ReubenBond:
@sunnycase I think we should close this PR, but I incorporated the main ideas into a branch here: https://github.com/ReubenBond/orleans/tree/poc-new-rpc which uses the existing code generator instead of `.tt` files. The .tt files are my primary concern with this change.

So far the performance improvement is quite good:
image

My hope is that we can include that change in a 3.0 release later this year.

Another similar optimization for 3.0 is to enable serializer sessions where there is two sessions for each socket: one for sending and one for receiving. Those can let us optimize how type serialization works by letting us refer to previously serialized types. Each type name would only be serialized once per session and thereafter referred to using a 'reference id'. That helps even with the above branch and this PR because derived types can be more efficiently serialized.

Thank you for the help and the initiative, @sunnycase!

sunnycase:
@ReubenBond  Sorry for not having much time for continually working on this.
Thank you for your excellent work. 👍

ReubenBond:
Your work here was useful in influencing the design of the new RPC

Issue #3694, opened 2017-11-21
Documentation on client configuration does not explain what 'target deployment ID' is


urig:
The [documentation page](https://github.com/dotnet/orleans/blob/gh-pages/src/Documentation/Deployment-and-Operations/Configuration-Guide/Client-Configuration.md/#L1) for Client Configuration makes references to the term 'target deployment ID'. However, the page does not explain what this term means or how to obtain the value for it. It might be a good idea to add an explanation of the term and instructions on how a user can find out the value to put in their client configuration.

Kimserey:
From what I understand the deployment ID on the client configuration targets the deployment ID of a cluster set in the `server configuration` in `GlobalConfiguration`.

There is slightly more explanation on the `server configuration` section of the doc with the following:
> DeploymentId is a unique string that defines a particular deployment. When deploying an Orleans based service to Azure it makes most sense to use the Azure deployment ID of the worker role.

I am quite confuse with the purpose of it.  

**When is it necessary to set a deployment ID?**

For example, when using a localhost primary silo configuration, the deployment ID is not set.

sergeybykov:
In the hindsight, `DeploymentId` should have been called `ClusterId` because it identifies the cluster that a set of silos belong to. That's how silos find each other to form a cluster. It is also used by clients to connect to a cluster.

@jdom is working on a PR that should improve the API for 2.0.0. We'll need to update the docs when 2.0.0 is out.

sergeybykov:
@urig We generally moved away from DeploymentId in 2.0 and settled on ClusterId instead. Do you agree that this is resolved now?

urig:
Yes. Thank you @sergeybykov and team.

PR #3728, opened 2017-11-30
Collapse DeploymentId with ClusterId

jdom:
Related to #3694 
Rename most usages of `DeploymentId` to `ClusterId`. This pass should be mostly back-compat.
I did not modify the providers that write a `DeploymentId` column to storage, as this would break in non-trivial ways, nor did I change the legacy configuration objects that parse the `DeploymentId` value (for both the main cluster id, as well as provider specific configuration that uses `DeploymentId` today). We plan to keep those object as legacy anyway, the new config system would use ClusterId as the new name for their strongly typed options.

Note to reviewers, I did split into a few commits that are probably easier to review in isolation, as they are doing similar things.

### DIFF
--- a/src/Orleans.Core/Configuration/GlobalConfiguration.cs
+++ b/src/Orleans.Core/Configuration/GlobalConfiguration.cs
@@ -222,7 +222,7 @@ namespace Orleans.Runtime.Configuration
         public string ClusterId { get; set; }
 
         /// <summary>
-        /// Deployment Id.
+        /// Deployment Id. This is the same as ClusterId and has been deprecated in favor of it.
         /// </summary>
         [Obsolete("DeploymentId is the same as ClusterId.")]
         public string DeploymentId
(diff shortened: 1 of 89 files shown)


src/Orleans.Core/Configuration/GlobalConfiguration.cs
```diff
@@ -840,9 +827,9 @@ internal override void Load(XmlElement root)
                             ServiceId = ConfigUtilities.ParseGuid(child.GetAttribute("ServiceId"),
                                 "Invalid Guid value for the ServiceId attribute on the Azure element");
                         }
-                        if (child.HasAttribute("DeploymentId"))
+                        if (child.HasAttribute("ClusterId"))
                         {
-                            DeploymentId = child.GetAttribute("DeploymentId");
+                            this.ClusterId = child.GetAttribute("ClusterId");
```
xiazen: This is a breaking change, xml config now need to change its "DeploymentId" to "ClusterId", is this intentional? @jdom

`ClientConfiguration` parse DeploymentId as ClusterId, but silo doesn't. This is inconsistent. Is this intentional?

jdom: oops, no... I explicitly told resharper not to change this one :/ I guess it insisted until he won. Nice catch. My intention is that XML parsing stays the same for legacy

jdom: Thanks, addressed in #3736

PR #3729, opened 2017-11-30
Wrap the connection preamble read check in a task

benjaminpetit:
Second fix for #3715

Basically, start a new ignored task for reading and validating the connection just accepted on the silo.

### DIFF
--- a/src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs
+++ b/src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs
@@ -5,6 +5,7 @@ using System.Net.Sockets;
 using Microsoft.Extensions.Logging;
 using Orleans.Messaging;
 using Orleans.Serialization;
+using System.Threading.Tasks;
 
 namespace Orleans.Runtime.Messaging
 {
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -340,19 +341,22 @@ private void ProcessAccept(SocketAsyncEventArgs e, bool completedSynchronously)
...
-                    }
-                    else
-                    {
-                        ima.SafeCloseSocket(sock);
-                    }
+                            StartReceiveAsync(sock, readEventArgs, ima);
+                        }
+                        else
+                        {
+                            ima.SafeCloseSocket(sock);
```
ReubenBond: I guess it's out-of-scope for this PR, but we're moving a blocking call onto the default TaskScheduler.
`SafeCloseSocket` calls `SocketManager.CloseSocket` which calls `Socket.Disconnect`, which ideally be async.

This might not affect Orleans code too much since most of our code uses either its own Threads or a custom TaskScheduler, but it can have a detrimental effect in co-hosting scenarios.

Unless I'm mistaken about `Disconnect` blocking.

PR #3730, opened 2017-12-01
Clear transaction context in top level transaction

jason-bragg:
Clear transaction context in top level transaction to prevent context from being passed back to the caller.

Context returned to the caller created a dependency on the transaction assembly for Orleans clients.  That assembly should be silo only.

### DIFF
--- a/src/Orleans.Runtime/Core/InsideRuntimeClient.cs
+++ b/src/Orleans.Runtime/Core/InsideRuntimeClient.cs
@@ -468,14 +468,12 @@ namespace Orleans.Runtime
                         this.transactionAgent.Value.Abort(TransactionContext.GetTransactionInfo(), abortException);
                     }
                 }
-                catch (Exception) { }
                 finally
                 {
                     TransactionContext.Clear();
+                    if (message.Direction != Message.Directions.OneWay)
+                        SafeSendExceptionResponse(message, exc2);
                 }
-
-                if (message.Direction != Message.Directions.OneWay)
-                    SafeSendExceptionResponse(message, exc2);
             }
             finally
             {
--- a/src/Orleans.Runtime/Core/InsideRuntimeClient.cs
+++ b/src/Orleans.Runtime/Core/InsideRuntimeClient.cs
@@ -411,6 +411,7 @@ namespace Orleans.Runtime
 
                     if (message.Direction != Message.Directions.OneWay)
                     {
+                        TransactionContext.Clear();
                         SafeSendExceptionResponse(message, exc1);
                     }
                     return;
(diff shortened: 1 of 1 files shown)


src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -449,21 +452,30 @@ public void SniffIncomingMessage(Message message)
...
+                    {
+                        // Must abort the transaction on exceptions
+                        TransactionContext.GetTransactionInfo().IsAborted = true;
+                        var abortException = (exc2 as OrleansTransactionAbortedException) ??
+                            new OrleansTransactionAbortedException(TransactionContext.GetTransactionInfo().TransactionId, exc2);
+                        this.transactionAgent.Value.Abort(TransactionContext.GetTransactionInfo(), abortException);
+                    }
                 }
-                else if (TransactionContext.GetTransactionInfo() != null)
+                catch (Exception) { }
```
xiazen: I don't quite understand why we are ignoring exception here

xiazen: if it is not important, should we at least log it?

jason-bragg: Squeshed because we're already in error handling logic and there is nothing that can be done, but yeah, probably best not to squash these.  Fixed

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -411,6 +411,7 @@ public void SniffIncomingMessage(Message message)
 
                     if (message.Direction != Message.Directions.OneWay)
                     {
+                        TransactionContext.Clear();
```
jdom: shouldn't this be cleared regardless of the direction?

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -430,6 +431,7 @@ public void SniffIncomingMessage(Message message)
...
```
jdom: Same here... at first thought at least seems that it should clear before returning, regardless of direction... I might be wrong, as I didn't follow it very thoroughly

jdom: ah, nevermind, I see now that this whole block is wrapped with a try/finally where you clear in the finally statement

---
In reply to: [154475381](https://github.com/dotnet/orleans/pull/3730#discussion_r154475381) [](ancestors = 154475381)

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
jason-bragg: It is, but it probably doesn't matter.  What's critical here is whether the context is still set when building the response message.  For calls that created the context, or when exceptions are thrown, the transaction context should be cleared prior to building a response.

PR #3731, opened 2017-12-01
Conform transaction configurations to options pattern

jason-bragg:


src/Orleans.Transactions.Azure/Storage/AzureTransactionLogStorage.cs
```diff
@@ -43,29 +43,29 @@ public class AzureTransactionLogStorage : ITransactionLogStorage
...
+            this.options = configurationOptions.Value;
         }
 
         public async Task Initialize()
         {
-            if (string.IsNullOrWhiteSpace(configuration.ConnectionString))
+            if (string.IsNullOrWhiteSpace(this.options.ConnectionString))
             {
-                throw new ArgumentNullException(nameof(configuration.ConnectionString));
+                throw new ArgumentNullException(nameof(this.options.ConnectionString));
```
ReubenBond: In addition to this, we could add an `IConfigurationValidator` to DI so that we throw on `Build()`. It might help to surface errors even sooner.

src/Orleans.Transactions.Azure/Hosting/SiloBuilderExtensions.cs
```diff
@@ -1,7 +1,6 @@
-﻿using Microsoft.Extensions.DependencyInjection;
+﻿using System;
+using Microsoft.Extensions.DependencyInjection;
 using Orleans.Hosting;
-using Orleans.Transactions.Abstractions;
-using System.Threading.Tasks;
 
 namespace Orleans.Transactions.Azure
```
ReubenBond: I wonder if we should put these extension methods in a more common namespace to improve discovery. No strong opinion.

test/Orleans.Transactions.Tests/Memory/MemoryTransactionsFixture.cs
```diff
@@ -27,7 +25,7 @@ public ISiloHostBuilder CreateSiloBuilder(string siloName, ClusterConfiguration
                     .ConfigureSiloName(siloName)
                     .UseConfiguration(clusterConfiguration)
                     .ConfigureLogging(builder => TestingUtils.ConfigureDefaultLoggingBuilder(builder, TestingUtils.CreateTraceFileName(siloName, clusterConfiguration.Globals.ClusterId)))
-                    .UseInClusterTransactionManager(new TransactionsConfiguration())
+                    .UseInClusterTransactionManager(options => { })
```
jdom: You seem to be using defaults very often... If this is the most likely scenario for end-users, consider and overload that doesn't take any parameters (or an optional parameter that by default is null). If this would be common just in tests, then ignore the suggestion

src/Orleans.Transactions.Azure/Storage/AzureTransactionLogOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿
+namespace Orleans.Transactions.Azure
+{
+    public class AzureTransactionLogOptions
+    {
+        public string ConnectionString { get; set; }
+
+        public string TableName { get; set; } = "TransactionLog";
```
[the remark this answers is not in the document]
jason-bragg: Pretty much all of our table storage related features have default table names.  It's a question of ease of use vs. safety, I think.

test/Orleans.Transactions.Tests/Memory/MemoryTransactionsFixture.cs
jason-bragg: Defaults should be common for this.  Will had overload or something.

src/Orleans.Transactions/Hosting/SiloBuilderExtensions.cs
```diff
@@ -8,38 +9,63 @@ namespace Orleans.Transactions
...
+        public static ISiloHostBuilder UseInClusterTransactionManager(this ISiloHostBuilder builder)
+        {
+            return builder.UseInClusterTransactionManager(options => { });
+        }
+
         /// <summary>
         /// Configure cluster to use an in-cluster transaction manager.
         /// </summary>
-        public static ISiloHostBuilder UseInClusterTransactionManager(this ISiloHostBuilder builder, TransactionsConfiguration config)
+        public static ISiloHostBuilder UseInClusterTransactionManager(this ISiloHostBuilder builder, Action<TransactionsOptions> configureOptions)
```
jdom: You are missing the overload that takes both `ISiloHostBuilder` and `Action<OptionsBuilder<TransactionsOptions>>`

src/Orleans.Transactions/Hosting/SiloBuilderExtensions.cs
```diff
@@ -8,38 +9,63 @@ namespace Orleans.Transactions
 {
     public static class SiloBuilderExtensions
     {
+        /// <summary>
+        /// Configure cluster to use an in-cluster transaction manager using defaults.
+        /// </summary>
+        public static ISiloHostBuilder UseInClusterTransactionManager(this ISiloHostBuilder builder)
+        {
+            return builder.UseInClusterTransactionManager(options => { });
```
jdom: nit: a better way would be `builder.UseInClusterTransactionManager(default(Action<OptionsBuilder<TransactionsOptions>>));` so that it doesn't even attempt to create the options builder, as the delegate is null.
Not that I care that much, just for future work, or if you are going to address the other comment too

src/Orleans.Transactions.Azure/Storage/AzureTransactionLogStorage.cs
jason-bragg: added validator, but keeping these checks in place.  Component should not assume configuration is valid.

src/Orleans.Transactions/Hosting/SiloBuilderExtensions.cs
jdom: btw, this is not what I meant. What I meant is passing a `null` delegate entirely, not a cached no-op delegate (no reason to cache this single use delegate and keep it forever in memory). This way `configureOptions?.Invoke(services.AddOptions<TransactionsOptions>());` will not even attempt to create an options builder, nor it would register dummy objects in the container. That whole part will be skipped and the container stay clean.

Not a big deal, but if we were keeping a dummy object, I'd prefer not caching it

---
In reply to: [154483640](https://github.com/dotnet/orleans/pull/3731#discussion_r154483640) [](ancestors = 154483640)

src/Orleans.Transactions.Azure/Storage/AzureTransactionLogOptions.cs
```diff
@@ -7,4 +10,26 @@ public class AzureTransactionLogOptions
...
+        public AzureTransactionLogOptionsValidator(IOptions<AzureTransactionLogOptions> configurationOptions)
+        {
+            this.options = configurationOptions.Value;
+        }
+
+        public void ValidateConfiguration()
+        {
+            if (string.IsNullOrWhiteSpace(this.options.ConnectionString))
+            {
+                throw new OrleansConfigurationException($"The azure transaction log was configured with an invalid connection string");
```
ReubenBond: nit: Azure should be capitalized. Should we point out how to configure the ConnectionString/TableName values in these errors or is it too obvious?

jason-bragg: Made error message a bit more explicit.

PR #3735, opened 2017-12-01
Skipping transaction test due to intermittent failures.

jason-bragg:


xiazen:
nvm, you want to skip the second test. I see now

Issue #3737, opened 2017-12-02
Provider model refactor


jason-bragg:
In an effort to clean up our provider model to support more advanced scenarios required by features like streaming, we'll be making changes to our provider model in the following areas.

_Areas of focus:_
_**Lifecycle**_ - In general, the silo lifecycle should be sufficient to address any provider startup/shutdown needs, even for features like streaming.  
_**Naming**_ – Many providers, such as storage and streams, are named and may have multiple instances.  For this we plan to use named services and named options.  This will require some prototyping.
_**Dynamic providers**_ – Stream providers are currently the only providers that support dynamically adding/removing and stopping/starting outside the initial silo lifecycle.  For providers of this nature we’ll need to either generalize the logic used for this for stream providers or work out how to do this safely.

Legacy support is planned for third party providers.

To vet the approach in phases we're starting with the simplest provider model, bootstrap providers.  The main goal of the bootstrap provider change is to vet that the silo lifecycle is sufficient for such purposes.

Next we'll be looking at storage providers to work out the use of names services and named options (in conjunction with the lifecycle, vetted in the bootstrap refactor) to support extensions to Orleans core actor model.

Last we’ll look at stream providers, using the lifecycle to improve their startup/shutdown logic and working out how to generalize dynamically adding/removing providers at runtime.

jason-bragg:
Boostrap refactor can be found in "Added support for legacy bootstrap providers. #3738"

ReubenBond:
Do we need dynamic providers - could we drop support for it?

jason-bragg:
@ReubenBond, Good question.  Let's discuss in planning meeting.

jason-bragg:
Update:

- We're dropping dynamic providers for 2.0 release.  May revisit in future.

- Named provider prototype is mostly working.  Will be under review soon, but will provide some details here now.

_**Existing**_
To start, it may be helpful to revisit some of the existing Orleans capabilities this work relies on.

_Keyed services_
Since we've chosen to limit our dependency injection usage to that provided by the Microsoft.Extensions.DependencyInjection abstraction layer, we do not have named services at the container level.  However, named services are quite useful, so we' built some minimal support for them on top of the DI abstraction.
For instance, if we needed to access different implementations of IFruit by name we could do this by registering the instance as name services in the container, then accessing them as such.

    // register in service collection
    serviceCollection.AddTransientNamedService<IFruit, Apple>("apple");
[...]

PR #3738, opened 2017-12-02
Remove provider management while maintaining support for legacy providers

jason-bragg:
This change is part of "Provider model refactor #3737"

This PR removes provider managers and loaders.  The extensibility which provider model supported is now native to the structure of Orleans core and requires no special management logic.  Legacy providers are now registered with the silo lifecycle.

- Original description - changed after team discussions
With the introduction of the silo lifecycle, bootstrap providers should not be necessary, as any system can now take part in the silo lifecycle.
This PR mostly pulls out the bootstrap provider from core, into a legacy provider system, and vets that the lifecycle is sufficient for this purpose.


src/Orleans.Providers.Legacy.Runtime/Hosting/LegecyProviderSiloBuilderExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <summary>
+        /// Configures the silo to use legacy Bootstrap providers
+        /// </summary>
+        public static IServiceCollection UseBootstrapProviders(this IServiceCollection services)
+        {
+            services.AddSingleton<BootstrapProviderLifecycle>();
+            services.AddFromExisting<ILifecycleParticipant<ISiloLifecycle>, BootstrapProviderLifecycle>();
+            services.AddSingleton<BootstrapProviderManager>();
+            services.AddFromExisting<IProviderManager, BootstrapProviderManager>();
+            return services;
```
jdom: we should attempt to make this method idempotent I believe

src/Orleans.Providers.Legacy.Core/Configuration/BootstrapProviderConfiguration.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        public const string BOOTSTRAP_PROVIDER_CATEGORY_NAME = "Bootstrap";
+
+        /// <summary>
+        /// Registers a given type of <typeparamref name="T"/> where <typeparamref name="T"/> is bootstrap provider
+        /// </summary>
+        /// <typeparam name="T">Non-abstract type which implements <see cref="IBootstrapProvider"/> interface</typeparam>
+        /// <param name="providerName">Name of the bootstrap provider</param>
+        /// <param name="properties">Properties that will be passed to bootstrap provider upon initialization</param>
+        public static void RegisterBootstrapProvider<T>(this GlobalConfiguration config, string providerName, IDictionary<string, string> properties = null) where T : IBootstrapProvider
```
jdom: Why do you have this in the legacy *Core* library, since it's only usable in a silo?

jason-bragg: As a general approach every thing I moved from abstractions or core was moved to core.  So this is in core because original code was in core or abstractions.  Since bootstrap providers are silo only, I can probably move this to runtime, but not all providers are silo only, so some core logic will still be needed :/

src/Orleans.Providers.Legacy.Runtime/Hosting/LegecyProviderSiloBuilderExtensions.cs
jason-bragg: Suggestions?
Use TryXXX calls?  Some other means?

jason-bragg: Will address, though I feel compelled to note we don't seem to be adding such protections for many core and more critical systems, so it's unclear why this is deemed relevant here.

jdom: It is not strictly done everywhere, although we are still in the process of adding more and more of these extension methods that we expect *end users* to call, where they might call it multiple times, or even infrastructure (such as the Orleans Dashboard, not sure if it's a great example) might want to make sure bootstrap providers are enabled, and call it just in case.

In a lot of our registrations we use the `TryAddXXX` methods, which makes them idempotent, and I do flag code that doesn't do it if I happen to see it. Nevertheless, even if we miss a few, the conforming container behavior says that resolving a service makes the last one registered be resolved, basically making it less important which actual registration worked if they were similar.

But, and why here is very important to do it correctly, when resolving all instances of a service, then it's no longer "the last registration wins and gets resolved". All of them are. And `ILifecycleParticipant<ISiloLifecycle>` is an important example of why adding it twice would break things, as all instances will be resolved and started (even if naively we thought of them as singletons when writing the code).

test/Orleans.Providers.Legacy.Test/Orleans.Providers.Legacy.Test.csproj
```diff
@@ -0,0 +1,33 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.Providers.Legacy.Tests</PackageId>
```
jdom: Should not have PackageId and have IsPackable false.
Also, name of the folder and the project should be `.Tests` (not `.Test`)

src/Orleans.Providers.Legacy.Core/Configuration/BootstrapProviderConfiguration.cs
jdom: I don't know... Do we really need to be that fine-grained? After all, this is just a legacy package that we hope users do not have to use. For the ones who do, just having a single legacy package, even if it brings up a few more dependencies than what they need, it might be OK. Not sure we need to optimize for that. On the other side, having an explosion of Legacy packages that we don't intend many will use will just be noise.
I *think* I would even prefer that we have a single or *very* few legacy packages, and not even 1 for just the `Providers` package, such as `Microsoft.Orleans.Legacy`, where we put this, we put the legacy configuration abstractions, legacy logger, etc

/cc @attilah @xiazen @sergeybykov

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -130,7 +128,7 @@ internal static void AddDefaultServices(HostBuilderContext context, IServiceColl
             services.TryAddSingleton<InsideRuntimeClient>();
             services.TryAddFromExisting<IRuntimeClient, InsideRuntimeClient>();
             services.TryAddFromExisting<ISiloRuntimeClient, InsideRuntimeClient>();
-            services.TryAddFromExisting<ILifecycleParticipant<ISiloLifecycle>, InsideRuntimeClient>();
+            services.AddFromExisting<ILifecycleParticipant<ISiloLifecycle>, InsideRuntimeClient>();
```
xiazen: Why is this AddFromExisting, not TryAddFromExisting. We tend to follow that pattern in this file

test/Orleans.Providers.Legacy.Test/Bootstrap/BootstrapProviderTests.cs
```diff
@@ -154,5 +168,30 @@ public BootstrapProvidersTests(ITestOutputHelper output, Fixture fixture)
             }
             return -1;
         }
+
+        private async Task<int> GetAllSiloProviderNames(string providerName)
+        {
+            var mgmt = this.fixture.GrainFactory.GetGrain<IManagementGrain>(0);
+            // request provider InitCount on all silos in this cluster
+            object[] results = await mgmt.SendControlCommandToProvider(typeof(GrainCallBootstrapper).FullName, GrainCallBootstrapperName, (int)MockBootstrapProvider.Commands.InitCount, null);
```
xiazen: Function name isn't expressing the actual behavior. This is more like `GetGrainCallBootstrapperProviderCountOnAllSilos`.

xiazen: And is this method still in use? It seems being replaced by the method below

src/Orleans.Runtime/Silo/TestHooks/TestHooksSystemTarget.cs
```diff
@@ -63,18 +63,6 @@ public Task<bool> HasStreamProvider(string providerName)
             }
         }
 
-        public Task<bool> HasBoostraperProvider(string providerName)
```
xiazen: hmm should we support this in the legacy package too? If we support the legacy feature,   I think we should support the ability to test it too right?

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
jdom: Because Try here will find other registrations of `ILifecycleParticipant<ISiloLifecycle>` and not add this one. See https://github.com/dotnet/orleans/pull/3738#discussion_r155381597

src/Orleans.Runtime/Silo/TestHooks/TestHooksSystemTarget.cs
jdom: The reality is that these test hooks are mostly useful to test the internals of Orleans (aka: by us in Orleans.sln). Not really that important to provide this test hook.

src/Orleans.Providers.Legacy.Core/Configuration/BootstrapProviderConfiguration.cs
xiazen: I agree. Or we have two, one for silo, one for client. 
I suggested it because, most legacy features aren't removed, they just live in our new core code in a better form. In other words, their core logic is till living in orleans core code. What we are maintaining in these legacy packages are really just legacy ways to configuring them. Hence we really just need one package for the legacy configuring methods.

If we have a package called `Legacy.Core`  for example, it sounds like it contains some core business logic. 

but Legacy.Logger package is a bit different, because it actually have lots of business logic in it. It contains legacy features which we want to discard in the future, and it also contains feature like bulk we wanted to support continuely. I'm not against to put it into the big legacy package, just mention my thoughts on this.

jason-bragg: I'm ok with legacy packages not being per feature, but due to the pre-existing dependencies we need at least a core and a runtime version.
What do you think of merging ..Providers.Legacy.Core, Providers.Legacy.Runtime and Legacy.Logger into Orleans.Legacy.Core and Orleans.Legacy.Runtime?

jdom: @jason-bragg I think it should be fine to have 2, although not sure about which dependencies differences there are that concern you.

BTW, I believe it might be more clear if `Legacy` is the suffix, and not in the middle of the name, as it is extending Core with legacy features.
This way they packages would be:
* `Microsoft.Orleans.Core.Legacy`
* `Microsoft.Orleans.Runtime.Legacy`

Still, it's just a little misleading, as these packages will have a little more dependencies than just Core or Runtime (for example Providers), but I think I can live with that.

@xiazen regarding the Logger functionality, I think we just have bulking there because of legacy reasons, it will not be our recommendation to tell users to use the legacy logging abstraction just to have bulking, and instead do bulking on `ILogger` natively. So I would not consider Logger to be a special case, and instead would bundle it with the rest of the legacy stuff that we intend to deprecate and eventually stop supporting altogether.

xiazen: what do you mean by bulking on ILogger natively? They support bulking feature natively?

jdom: not built-in (that I know of), what I mean is to have a similar functionality on top of ILogger, without having to go through our `ILogConsumer` abstraction.
If they are using a 3rd party logger to write to files or external stores, they can benefit from the bulking in those logging frameworks. Regardless of current availability, the only reason we still have that is because of legacy... otherwise it shouldn't be in a legacy package (but I'll push back if we want to make bulking 1st class citizen of Orleans, it's not our concern, it's the concern of logging frameworks)

xiazen: but our bulking is built on top of ILogger, not ILogConsumer. Even it is not Orleans core related, but because of the historical reason, we are supporting it as a feature. Either you call it legacy or not, I'm just saying it is not a deprecated feature. I'm not against to put it into the Legacy.Runtime, Legacy.Core package. I was just saying , to me , those Legacy.Runtime, Legacy.Core contains legacy configuration methods which are to be deprecated in the future, but bulking is not. So when we remove all the deprecated pieces, bulking will still be in the legacy package. At that time, we might discuss different options to deal with it, either move it to Orleans Contrib, or keep it in Orleans tree but rename the package name to be Logging.Bulking or something. 
Or we can keep logging legacy features in a separate package now.

jason-bragg:
Last push is still work in progress.  Mostly working but still needs rebase, and full test.
Getting it out early to get eyes on it, as it's obnoxiously large.

src/Orleans.Core/Configuration/GlobalConfiguration.cs
```diff
@@ -1036,35 +1036,6 @@ internal override void Load(XmlElement root)
             }
         }
 
-        /// <summary>
-        /// Registers a given type of <typeparamref name="T"/> where <typeparamref name="T"/> is bootstrap provider
-        /// </summary>
-        /// <typeparam name="T">Non-abstract type which implements <see cref="IBootstrapProvider"/> interface</typeparam>
-        /// <param name="providerName">Name of the bootstrap provider</param>
-        /// <param name="properties">Properties that will be passed to bootstrap provider upon initialization</param>
-        public void RegisterBootstrapProvider<T>(string providerName, IDictionary<string, string> properties = null) where T : IBootstrapProvider
```
xiazen: Legacy configuration class should stay unchanged, right? What's the reason of removing this two methods from it ?

src/Orleans.Core/Configuration/Legacy/LegacyConfigurationExtensions.cs
```diff
@@ -6,11 +6,15 @@
...
 using Orleans.Hosting;
 using Orleans.Runtime;
+using Orleans.Providers;
 
 namespace Orleans.Configuration
 {
     internal static class LegacyConfigurationExtensions
     {
+        private const int ClusterClientDefaultProviderInitStage = 1000;
+        private const int ClusterClientDefaultProviderStartStage = 2000;
```
xiazen: avoid use hard coded stage number. Isn't you suppose to have a `ClientLifeCycleStage` class, use it here?

jason-bragg: Currently, the only thing in the lifecycle is the providers, so these are just default stages to wire them up with.  

We need to think through if and how we want to use lifecycle in the client.  If we do fully utilized the lifecycle to control client startup/shutdown, then we'll need to merge these values with the existing client lifecycle logic.

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -252,8 +234,6 @@ await this.gatewayListProvider.InitializeGatewayListProvider()
                 ct).Ignore();
             grainInterfaceMap = await transport.GetTypeCodeMap(this.InternalGrainFactory);
             
-            await ClientStatistics.Start(statisticsProviderManager, transport, clientId)
```
xiazen: please point me to the logic which handles `ClientStatistics.Start` now. `ClientStatisticsManager.Start` init statistic publishers and also start the statistic runtime, which keep a list of internal metrics and also statistic tool classes to calculate the metrics and report them periodically.

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
```diff
@@ -35,17 +34,15 @@ public ClientStatisticsManager(ClientConfiguration config, SerializationManager
...
             runtimeStats.Start();
 
             // Configure Metrics
-            IProvider statsProvider = null;
             if (!string.IsNullOrEmpty(statisticsOptions.ProviderName))
             {
                 var extType = statisticsOptions.ProviderName;
-                statsProvider = statsManager.GetProvider(extType);
-                var metricsDataPublisher = statsProvider as IClientMetricsDataPublisher;
+                IClientMetricsDataPublisher metricsDataPublisher = this.serviceProvider.GetServiceByName<IClientMetricsDataPublisher>(extType);
```
xiazen: hmmm where is the code which register dataPublisher as `IClientMetricDataPublisher` named service into DI? I didn't see such code in `LegacyProviderConfigurator`. If it is not registered, then we cannot get service here right?

src/Orleans.Core/Streams/Providers/IStreamProviderManager.cs
```diff
@@ -1,10 +0,0 @@
-using System.Collections.Generic;
-using Orleans.Providers;
-
-namespace Orleans.Streams
-{
-    public interface IStreamProviderManager : IProviderManager
-    {
-        IEnumerable<IStreamProvider> GetStreamProviders();
```
xiazen: why remove this public interface? or remove the ability to get all stream providers? we need to introduce as least as possible breaking change to legacy system, right?

xiazen: We can have a new implementation which wraps a `IServiceProvider`, which can easily give us all stream providers.

src/Orleans.Core/Streams/Providers/StreamProviderManagerExtensions.cs
```diff
@@ -1,14 +0,0 @@
-using System;
-
-namespace Orleans.Streams
-{
-    public static class StreamProviderManagerExtensions
-    {
-        public static IStreamProvider GetStreamProvider(this IStreamProviderManager streamProviderManager, string name)
```
xiazen: same here

src/Orleans.Core/SystemTargetInterfaces/ISiloControl.cs
```diff
@@ -21,9 +20,6 @@ internal interface ISiloControl : ISystemTarget, IVersionManager
         Task<DetailedGrainReport> GetDetailedGrainReport(GrainId grainId);
 
         Task UpdateConfiguration(string configuration);
-
-        /// <summary>Load and initialize newly added stream providers. Remove providers that are not on the list that's being passed in.</summary>
-        Task UpdateStreamProviders(IDictionary<string, ProviderCategoryConfiguration> streamProviderConfigurations);
```
xiazen: I think dynamic adding stream provider feature depends on this method, are we removing that feature ? 
No objection on removing it, just want to confirm that's the intension

src/Orleans.Runtime.Legacy/Configuration/BootstrapProviderConfiguration.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        public const string BOOTSTRAP_PROVIDER_CATEGORY_NAME = "Bootstrap";
+
+        /// <summary>
+        /// Registers a given type of <typeparamref name="T"/> where <typeparamref name="T"/> is bootstrap provider
+        /// </summary>
+        /// <typeparam name="T">Non-abstract type which implements <see cref="IBootstrapProvider"/> interface</typeparam>
+        /// <param name="providerName">Name of the bootstrap provider</param>
+        /// <param name="properties">Properties that will be passed to bootstrap provider upon initialization</param>
+        public static void RegisterBootstrapProvider<T>(this GlobalConfiguration config, string providerName, IDictionary<string, string> properties = null) where T : IBootstrapProvider
```
xiazen: ohhhh ignore my earlier comment on `GlobalConfiguration`, you moved those methods here.

src/OrleansProviders/Storage/ShardedStorageProvider.cs
```diff
@@ -1,141 +0,0 @@
-using System;
-using System.Collections.Generic;
-using System.Threading;
-using System.Threading.Tasks;
```
xiazen: are we removing `SharedStorageProvider`? and why?

test/AWSUtils.Tests/StorageTests/PersistenceGrainTests_AWSDynamoDBStore.cs
```diff
@@ -47,22 +47,7 @@ protected override TestCluster CreateTestCluster()
                     options.ClusterConfiguration.Globals.RegisterStorageProvider<DynamoDBStorageProvider>("DDBStore1", new Dictionary<string, string> { { "DataConnectionString", dataConnectionString } });
```
xiazen: this tests are not included in vso or Jenkins. make sure you run it locally and it passes after this change.

src/Orleans.Core/Configuration/ProviderConfiguration.cs
```diff
@@ -167,29 +157,6 @@ public override string ToString()
             var propsStr = properties == null ? "Null" : Utils.EnumerableToString(properties.Keys);
             return string.Format("Name={0}, Type={1}, Properties={2}", Name, Type, propsStr);
         }
-
-        /// <summary>
-        /// Children providers of this provider. Used by hierarchical providers.
-        /// </summary>
-        public IList<IProvider> Children
```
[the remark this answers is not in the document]
jason-bragg: You are correct.  I raised this as a concern when I was advocating moving the provider managers rather then removing them.  It was agreed that we'd drop sharded providers and child providers.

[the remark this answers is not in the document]
xiazen: Ok I guess I missed that meeting. This looks good to me then. and ignore all my further comments regarding child provider and shared storage provider . thanks for explaining it

src/Orleans.Core/Core/ClusterClient.cs
```diff
@@ -72,13 +94,6 @@ private IInternalGrainFactory InternalGrainFactory
         private bool IsDisposing => this.state == LifecycleState.Disposed ||
                                     this.state == LifecycleState.Disposing;
 
-        /// <inheritdoc />
-        public IEnumerable<IStreamProvider> GetStreamProviders()
-        {
```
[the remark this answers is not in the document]
jason-bragg: Providers in general are now just natural extensions to Orleans.  No special support is needed for the behaviors,  The legacy provider support introduced in this PR is there only to allow externally developed providers to continue to work.

These functions no longer fit how Orleans works.  We can continue to support them in the legacy packages via extension functions, but I'm not convinced that is necessary or good.  Further, in these cases, I don't think these calls are commonly used, so we'd get little by maintaining them.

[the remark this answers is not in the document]
jason-bragg: I'd note that their is a meaningful difference between providers and logging.  Most systems need logging and those that use it will use the related calls a lot.  Providers have far less application layer surface and their api's will likely be used far less often.  This means the impact of breaking change is dramatically less with providers than logging.

src/Orleans.Runtime.Legacy/Configuration/BootstrapProviderConfiguration.cs
xiazen: but on a side note. if `GlobalConfiguration` is going to be moved to `Core.Legacy` or `Runtime.Legacy` package, is there still value to keep those as extension methods? Might just keep them as `GlobalConfiguration` methods as before.

Or is this a intermedia step for moving boostrap provider to legacy project

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,162 @@
...
+                if (providerGroup.Key == ProviderCategoryConfiguration.STREAM_PROVIDER_CATEGORY_NAME)
+                {
+                    foreach (IProviderConfiguration providerConfig in providerGroup.SelectMany(kvp => kvp.Value.Providers.Values))
+                    {
+                        RegisterProvider<IStreamProvider>(providerConfig, services, defaultInitStage, defaultStartStage);
+                    }
+                }
+                else if (providerGroup.Key == ProviderCategoryConfiguration.STORAGE_PROVIDER_CATEGORY_NAME)
+                {
+                    services.AddSingleton<IStorageProvider>(sp => sp.GetServiceByName<IStorageProvider>(ProviderConstants.DEFAULT_STORAGE_PROVIDER_NAME));
```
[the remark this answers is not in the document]
jason-bragg: This is just creating a mapping in DI to the default provider, so that if someone resolves an IStrorageProvider (not by name) they will get the default provider..  The default provider must still be in the list of configured providers, and be registered with RegisterProvider.

src/Orleans.Core/Runtime/Constants.cs
```diff
@@ -37,7 +37,7 @@ internal class Constants
         public static readonly GrainId CatalogId = GrainId.GetSystemTargetGrainId(14);
         public static readonly GrainId MembershipOracleId = GrainId.GetSystemTargetGrainId(15);
         public static readonly GrainId TypeManagerId = GrainId.GetSystemTargetGrainId(17);
-        public static readonly GrainId ProviderManagerSystemTargetId = GrainId.GetSystemTargetGrainId(19);
+        public static readonly GrainId FallbackSystemTargetId = GrainId.GetSystemTargetGrainId(19);
```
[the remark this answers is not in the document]
jason-bragg: Unclear about original code or goals, but it looked like there was a system target setup specifically to provide a scheduling context for providers.  Later others started to use it as a default scheduling context if various components didn't have their own.. which is not really what it was for.

I would have removed (or moved) this context if it was, as originally intended, only for the providers, but since it was now used as some sort of fallback, I chose to rename it to what it was actually being used for, then use that in the legacy provider support, as it was before.

I'm not claiming this is how this logic should be, only that it's an improvement over what it was.  I think we should recognize that some of our components expect to be run in a scheduling context, and address that problem what to do if they do not have one, but that's out of scope for this PR.

src/Orleans.Core/Runtime/IRuntimeClient.cs
```diff
@@ -59,8 +59,6 @@ internal interface IRuntimeClient
 
         void DeleteObjectReference(IAddressable obj);
         
-        Streams.IStreamProviderManager CurrentStreamProviderManager { get; }
-
         Streams.IStreamProviderRuntime CurrentStreamProviderRuntime { get; }
```
jason-bragg: merge issues

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
jason-bragg: I suspect you're correct.  Will look into.

src/Orleans.Core/Streams/Providers/IStreamProviderManager.cs
jason-bragg: We're removing the entire provider model, which is a breaking change, but 'should' not be bad as most users never see or use these surfaces.  The main thing we're keeping is the ability to load the custom providers which application developers have already developed.

src/Orleans.Core/SystemTargetInterfaces/ISiloControl.cs
jason-bragg: Yup.  Runtime addition/removal of providers is being dropped.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    internal static class LegacyProviderConfigurator<TLifeCycle>
+        where TLifeCycle : ILifecycleObservable
+    {
+        /// <summary>
+        /// Legacy way to configure providers. Will need to move to a legacy package in the future
+        /// </summary>
+        /// <returns></returns>
+        internal static void ConfigureServices(IDictionary<string, ProviderCategoryConfiguration> providerConfigurations, IServiceCollection services, int defaultInitStage, int defaultStartStage)
```
jason-bragg: This is the heart of this change.  This is where we wrap provider construction and initialization into lifecycle participants for initialization and register them as named services.

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
jason-bragg: I likely broke this.  Will look.

src/Orleans.Runtime.Legacy/Configuration/BootstrapProviderConfiguration.cs
jason-bragg: This depends on how we choose to move the legacy configs, if we do it all at once then this is probably redundant.  I wrote this when the plan was to move things in chunks.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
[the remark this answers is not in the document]
xiazen: >This is just creating a mapping in DI to the default provider, so that if someone resolves an IStrorageProvider (not by name) they will get the default provider.

why do we support this? if someone resolve IStorageProvider(not by name) then orleans return the default one, this seems arbitrary. Is it a replacement for some legacy feature we support before or?

src/Orleans.Core/Streams/Providers/IStreamProviderManager.cs
xiazen: >We're removing the entire provider model

I guess I missed more than one meeting.. Do we still keep `IProvider` interface ,or is that removed too? can you share a big summary here about what removing entire provider model means?

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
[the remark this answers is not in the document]
jason-bragg: This is convention only, and is based off of the rules around storage provider attribution.  If no provider name is indicated in the storage provider attribute, the default storage provider is assumed.  Applying this convention to DI and Names services, it means that if one asks for a storage provider by name (using getServiceByName) they get the specific service, but if they just resolve a storage provider they should get the default storage provider.  Without this, callers who just want to use the default provider would have to know the name assigned to the default provider.

src/Orleans.Core/Streams/Providers/IStreamProviderManager.cs
jason-bragg: Yes, IProvider should eventually be removed (or moved to legacy).

To support extensibility comparable to the provider model, we need only the following:
*  Names services support (existing)
*  Generic silo lifecycle which services can take part in (existing)
*  Way for named services to take part in silo lifecycle (added in this PR), see usages of IKeyedServiceCollection<string, ILifecycleParticipant<IXXXClientLifecycle>>.

With these capabilities, orleans has no need for the provider model and related framework, including provider managers, and related abstractions (IProvider, IProviderConfiguration).  However, as framework extensions, users have developed custom providers, so we want to retain the ability to load those, which this PR also enables.

jason-bragg:
got past testing, and address statistics issues.  Removing WIP label.  Ready for full review.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        private static void RegisterProvider<TService>(IProviderConfiguration config, IServiceCollection services, int defaultInitStage, int defaultStartStage)
+            where TService : class
+        {
+            services.AddSingletonNamedService<TService>(config.Name, (s, n) => {
+                Type providerType = s.GetRequiredService<ProviderTypeLookup>().GetType(config.Type);
+                return Activator.CreateInstance(providerType) as TService;
+            });
+            services.AddSingletonNamedService<ILifecycleParticipant<TLifeCycle>>(config.Name, (s, n) => new ProviderLifecycleParticipant<TService>(config, s, s.GetRequiredService<ILoggerFactory>(), defaultInitStage, defaultStartStage));
+            services.AddSingletonNamedService<IControllable>(config.Name, (s, n) => s.GetServiceByName<TService>(n) as IControllable);
```
jason-bragg: All providers are registered as controllable even if they aren't, because we can't check the concrete type.  IControllable pattern always checked if the provider was controllable anyway, so this is safe, though a bit hacky.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+                {
+                    foreach (IProviderConfiguration providerConfig in providerGroup.SelectMany(kvp => kvp.Value.Providers.Values))
+                    {
+                        RegisterProvider<IProvider>(providerConfig, services, defaultInitStage, defaultStartStage);
+                    }
+                }
+            }
+        }
+
+        private class ProviderLifecycleParticipant<TService> : ILifecycleParticipant<TLifeCycle>
```
jason-bragg: This is the wrapper class that allows providers to take part in the silo lifecycle.  It also adds the ability for users to specify the stage they wish the provider to initialize and start in, should they choose.  (not required)

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+            where TService : class
+        {
+            private readonly ILogger logger;
+            private readonly IProviderConfiguration config;
+            private readonly IServiceProvider services;
+            private readonly LegacyProviderConfigurator.QueueTask schedualer;
```
jason-bragg: Since provider initialization logic in a silo needs be run in an Orleans scheduling context, but client does not, to avoid maintaining two classes that are mostly the same, I opted to make the scheduling plagable. and get it from the container.  Open to other suggestions.

src/Orleans.Core.Abstractions/Core/Grain.cs
```diff
@@ -185,7 +185,9 @@ protected virtual IStreamProvider GetStreamProvider(string name)
                 throw new ArgumentNullException(nameof(name));
                 
             EnsureRuntime();
-            return this.ServiceProvider.GetServiceByName<IStreamProvider>(name);
+            IStreamProvider streamProvider = this.ServiceProvider.GetServiceByName<IStreamProvider>(name);
+            if (streamProvider == null) throw new KeyNotFoundException(name);
```
xiazen: I think we need to throw a more descriptive exception message here. Before the message is something like "stream provider with name blablba is not configured". If we suddenly change the exception message and also changed the whole provider infra. people may get confused.

src/Orleans.Core/Configuration/ProviderConfiguration.cs
```diff
@@ -18,12 +18,9 @@ public class ProviderConfiguration : IProviderConfiguration
         private readonly IList<ProviderConfiguration> childConfigurations = new List<ProviderConfiguration>();
```
xiazen: remove this too , since we are removing child config related stuff

src/Orleans.Core/Core/ClusterClient.cs
```diff
@@ -14,6 +17,7 @@ namespace Orleans
     internal class ClusterClient : IInternalClusterClient
     {
         private readonly OutsideRuntimeClient runtimeClient;
+        private readonly ClusterClientLifecycle clusterClientLifecycle;
```
xiazen: use `IClusterClientLifecycle` here? instead of the concrete class

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -76,32 +74,12 @@ internal static void AddDefaultServices(HostBuilderContext context, IServiceColl
...
-            services.TryAddSingleton<LogConsistencyProviderManager>();
-            services.AddFromExisting<IProviderManager, LogConsistencyProviderManager>();
-            services.TryAddFromExisting<IKeyedServiceCollection<string, ILogConsistencyProvider>, LogConsistencyProviderManager>(); // as named services
-            services.TryAddSingleton<ILogConsistencyProvider>(sp => sp.GetRequiredService<LogConsistencyProviderManager>().GetDefaultProvider()); // default
-
-            services.TryAddSingleton<BootstrapProviderManager>();
-            services.AddFromExisting<IProviderManager, BootstrapProviderManager>();
-            services.TryAddSingleton<LoadedProviderTypeLoaders>();
+            services.TryAddSingleton<FallbackSystemTarget>();
+            services.TryAddSingleton<SiloUnobservedExceptionsHandler>();
```
xiazen: why are we adding `SiloUnobservedExceptionsHandler` back? it was made optional , turned off by default in #3829

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -257,11 +243,22 @@ public Silo(ILocalSiloDetails siloDetails, IServiceProvider services)
...
             this.siloLifecycle = fullSiloLifecycle;
-            IEnumerable<ILifecycleParticipant<ISiloLifecycle>> lifecyleParticipants = this.Services.GetServices<ILifecycleParticipant<ISiloLifecycle>>();
-            foreach(ILifecycleParticipant<ISiloLifecycle> participant in lifecyleParticipants)
+            // register all lifecycles
+            IEnumerable<ILifecycleParticipant<ISiloLifecycle>> lifecycleParticipants = this.Services.GetServices<ILifecycleParticipant<ISiloLifecycle>>();
+            foreach(ILifecycleParticipant<ISiloLifecycle> participant in lifecycleParticipants)
             {
                 participant.Participate(fullSiloLifecycle);
             }
+            // register all named lifecycles
```
xiazen: lifecycle participant

src/Orleans.Runtime/Silo/TestHooks/TestHooksSystemTarget.cs
```diff
@@ -40,60 +39,51 @@ public Task<string> GetConsistentRingProviderDiagnosticInfo()
...
-            allProviders.AddRange(storageProviderManager.GetProviderNames());
-
-            var streamProviderManager = this.host.Services.GetRequiredService<IStreamProviderManager>();
-            allProviders.AddRange(streamProviderManager.GetStreamProviders().Select(p => p.Name));
+            var storageProviderCollection = this.host.Services.GetRequiredService<IKeyedServiceCollection<string,IStorageProvider>>();
+            allProviders.AddRange(storageProviderCollection.GetServices(this.host.Services).Select(keyedService => keyedService.Key));
 
-            var statisticsProviderManager = this.host.Services.GetRequiredService<StatisticsProviderManager>();
-            allProviders.AddRange(statisticsProviderManager.GetProviders().Select(p => p.Name));
+            var streamProviderCollection = this.host.Services.GetRequiredService<IKeyedServiceCollection<string, IStreamProvider>>(); ;
```
xiazen: why not reuse `GetStreamProviderNames`

src/Orleans.Runtime/Silo/TestHooks/TestHooksSystemTarget.cs
```diff
@@ -40,60 +39,51 @@ public Task<string> GetConsistentRingProviderDiagnosticInfo()
...
         public Task<ICollection<string>> GetAllSiloProviderNames()
         {
             List<string> allProviders = new List<string>();
 
-            var storageProviderManager = this.host.Services.GetRequiredService<StorageProviderManager>();
-            allProviders.AddRange(storageProviderManager.GetProviderNames());
-
-            var streamProviderManager = this.host.Services.GetRequiredService<IStreamProviderManager>();
-            allProviders.AddRange(streamProviderManager.GetStreamProviders().Select(p => p.Name));
+            var storageProviderCollection = this.host.Services.GetRequiredService<IKeyedServiceCollection<string,IStorageProvider>>();
```
xiazen: why not reuse `GetStorageProviderNames`

src/Orleans.Runtime/Versions/GrainVersionStore.cs
```diff
@@ -14,21 +17,13 @@ internal class GrainVersionStore : IVersionStore
...
-            this.IsEnabled = false;
-        }
-
-        public void SetStorageManager(IStorageProviderManager storageProviderManager)
-        {
-            IStorageProvider unused;
-            IsEnabled = storageProviderManager.TryGetProvider(
-                ProviderConstants.DEFAULT_STORAGE_PROVIDER_NAME,
-                out unused);
+            this.IsEnabled = services.GetServices<IKeyedService<string,IStorageProvider>>().Any(s => s.Equals(ProviderConstants.DEFAULT_STORAGE_PROVIDER_NAME));
```
xiazen: I think we can use `services.GetService<IStorageProvider>` here to see if default storage provider is configured

src/Orleans.TestingHost/AppDomainSiloHost.cs
```diff
@@ -79,15 +78,8 @@ public AppDomainTestHooks(ISiloHost host)
...
-        internal IBootstrapProvider GetBootstrapProvider(string name)
-        {
-            var bootstrapProviderManager = this.host.Services.GetRequiredService<BootstrapProviderManager>();
-            IBootstrapProvider provider = (IBootstrapProvider)bootstrapProviderManager.GetProvider(name);
-            return CheckReturnBoundaryReference("bootstrap provider", provider);
-        }
-
         /// <summary>Find the named storage provider loaded in this silo. </summary>
-        internal IStorageProvider GetStorageProvider(string name) => CheckReturnBoundaryReference("storage provider", (IStorageProvider)this.host.Services.GetRequiredService<StorageProviderManager>().GetProvider(name));
+        internal IStorageProvider GetStorageProvider(string name) => CheckReturnBoundaryReference("storage provider", this.host.Services.GetServiceByName<IStorageProvider>(name));
```
xiazen: throw `KeyNotFOundException` to preserve original logic, which was maintained by `StorageProviderManager`

src/Orleans.Core/Core/ClusterClient.cs
jason-bragg: The IClusterClientLifecycle interface is observable portion only, used by application layer to register with the lifecycle.  This logic uses both the observable and observer calls, so full impl is used.  For further abstraction we can store two interfaces (like we do with the grain lifecycle), but I'm unclear how valuable that is here.  It's done that way for the grain lifecycle because there is separate logic using the separate interfaces.

src/Orleans.Core/Configuration/ProviderConfiguration.cs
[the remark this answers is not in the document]
jason-bragg: It seems fairly tied with the legacy configuration.  I may be able to move it, but I think it's better to do that in a future pass.

src/Orleans.Core/Providers/IOrleansProvider.cs
```diff
@@ -11,7 +11,6 @@ namespace Orleans.Providers
     /// <summary>
     /// Base interface for all type-specific provider interfaces in Orleans
     /// </summary>
```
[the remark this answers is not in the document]
jason-bragg: Unclear, but I think it will be removed something like:
1) Add legacy support for providers (this PR)
2) Port all supported providers directly to names lifecycle participants that do not inherit from IProvider.
3) Move IProvider to legacy packages (Core.Legacy)
Somewhere along the way, we'll need to move cluster and client configuratiosn to legacy as well.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Linq;
+using Orleans.Runtime;
+
+namespace Orleans.Providers
+{
+    public class LegacyProviderConfigurator
+    {
+        public const string InitStageName = "ProviderInitStage";
+        public const string StartStageName = "ProviderStartStage";
+        internal const string SchedulingContextName = "Fallback";
```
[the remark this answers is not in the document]
jason-bragg: left over junk from a different approach I tried while prototyping.  will remove.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+            private readonly IServiceProvider services;
+            private readonly LegacyProviderConfigurator.QueueTask schedualer;
+            private readonly int defaultInitStage;
+            private int initStage;
+            private readonly int defaultStartStage;
+            private int startStage;
+
+            public ProviderLifecycleParticipant(IProviderConfiguration config, IServiceProvider services, ILoggerFactory loggerFactory, int defaultInitStage, int defaultStartStage)
+            {
+                this.logger = loggerFactory.CreateLogger($"{config.Type}-{config.Name}");
```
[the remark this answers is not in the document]
jason-bragg: config.Type is not an actual type.  It's the configured type string, usually the full type name.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            private async Task Start(CancellationToken ct)
+            {
+                var stopWatch = Stopwatch.StartNew();
+                IStreamProviderImpl provider = this.services.GetServiceByName<TService>(this.config.Name) as IStreamProviderImpl;
+                if (provider == null) return;
+                await schedual(() => provider.Start());
+                stopWatch.Stop();
+                this.logger.Info(ErrorCode.SiloStartPerfMeasure, $"Starting in stage {this.startStage} took {stopWatch.ElapsedMilliseconds} Milliseconds");
```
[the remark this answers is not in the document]
jason-bragg: Provider name is already part of the logger name.

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
```diff
@@ -35,23 +35,13 @@ public ClientStatisticsManager(ClientConfiguration config, SerializationManager
...
-        internal async Task Start(StatisticsProviderManager statsManager, IMessageCenter transport, GrainId clientId)
+        internal async Task Start(IMessageCenter transport, GrainId clientId)
         {
             runtimeStats.Start();
 
-            // Configure Metrics
-            IProvider statsProvider = null;
-            if (!string.IsNullOrEmpty(statisticsOptions.ProviderName))
+            IClientMetricsDataPublisher metricsDataPublisher = this.serviceProvider.GetService<IClientMetricsDataPublisher>();
+            if (metricsDataPublisher != null)
```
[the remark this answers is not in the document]
jason-bragg: Looks valid, but this may be due to the fact that I pulled in recent changes.  If it's wrong after next push let me know.

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
jason-bragg: Likely I missed this in rebase/merge.  Will remove.  Thanks, good catch.

src/Orleans.Runtime/Versions/GrainVersionStore.cs
jason-bragg: those calls were unnecessarily asynchrony, but don't think that matters really.  Will use them, it's better than code duplication.

src/Orleans.Runtime/Silo/TestHooks/TestHooksSystemTarget.cs
jason-bragg: those calls were unnecessarily asynchrony, but don't think that matters really.  Will use them, it's better than code duplication.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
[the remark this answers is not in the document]
xiazen: sorry I wasn't clear, I referred logger name as category name. Since not all ILogger implementation prints out category/logger name, so it maybe a good idea to print the provider name since it is useful in debugging.

src/Orleans.Core.Abstractions/Core/Grain.cs
```diff
@@ -185,7 +185,8 @@ protected virtual IStreamProvider GetStreamProvider(string name)
                 throw new ArgumentNullException(nameof(name));
                 
             EnsureRuntime();
-            return this.ServiceProvider.GetServiceByName<IStreamProvider>(name);
+            return this.ServiceProvider.GetServiceByName<IStreamProvider>(name)
```
ReubenBond: Maybe we should add a `GetRequiredServiceByName` extension method

src/Orleans.Core/Configuration/Legacy/LegacyConfigurationExtensions.cs
ReubenBond: I do want lifecycle on the client. I'm not sure how lifecycle stages should be shared in code. Eg, stream providers should be able to subscribe to the same lifecycle on both client and silo, I would imagine. Maybe there'd be a separate, shared lifecycle which those providers can subscribe to?

src/Orleans.Core/Configuration/ProviderConfiguration.cs
```diff
@@ -115,7 +103,7 @@ internal void Load(XmlElement child, IDictionary<string, IProviderConfiguration>
                 }
             }
 
-            LoadProviderConfigurations(child, nsManager, alreadyLoaded, c => childConfigurations.Add(c));
+            LoadProviderConfigurations(child, nsManager, alreadyLoaded, c => { });
```
ReubenBond: Should we drop the last parameter from the method?

src/Orleans.Core/Core/ClusterClient.cs
```diff
@@ -37,8 +41,26 @@ public ClusterClient(OutsideRuntimeClient runtimeClient, ClientConfiguration con
...
+            {
+                participant.Participate(clusterClientLifecycle);
+            }
+
+            // register all named lifecycle participants
+            IEnumerable<IKeyedServiceCollection<string, ILifecycleParticipant<IClusterClientLifecycle>>> namedLifecycleParticipantCollections = this.ServiceProvider.GetServices<IKeyedServiceCollection<string, ILifecycleParticipant<IClusterClientLifecycle>>>();
+            foreach (ILifecycleParticipant<IClusterClientLifecycle> participant in namedLifecycleParticipantCollections
+                .SelectMany(c => c.GetServices(this.ServiceProvider)
+                .Select(s => s.GetService(this.ServiceProvider))))
+            {
```
ReubenBond: The spacing here is a little confusing: the Select call is inside the SelectMany call, so it should be indented further.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+                if (providerGroup.Key == ProviderCategoryConfiguration.STREAM_PROVIDER_CATEGORY_NAME)
+                {
+                    foreach (IProviderConfiguration providerConfig in providerGroup.SelectMany(kvp => kvp.Value.Providers.Values))
+                    {
+                        RegisterProvider<IStreamProvider>(providerConfig, services, defaultInitStage, defaultStartStage);
+                    }
+                }
+                else if (providerGroup.Key == ProviderCategoryConfiguration.STORAGE_PROVIDER_CATEGORY_NAME)
+                {
+                    services.AddSingleton<IStorageProvider>(sp => sp.GetServiceByName<IStorageProvider>(ProviderConstants.DEFAULT_STORAGE_PROVIDER_NAME));
```
ReubenBond: Can provider groups be specified multiple times? I'm not sure, but we should add this exactly once regardless of whether or not a storage provider with `DEFAULT_STORAGE_PROVIDER_NAME` has been registered.

i.e, I believe we should move it out of the loop.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+                        RegisterProvider<ILogConsistencyProvider>(providerConfig, services, defaultInitStage, defaultStartStage);
+                    }
+                }
+                else if (providerGroup.Key == ProviderCategoryConfiguration.STATISTICS_PROVIDER_CATEGORY_NAME)
+                {
+                    IProviderConfiguration providerConfig = providerGroup.SelectMany(kvp => kvp.Value.Providers.Values).FirstOrDefault();
+                    if(providerConfig != null)
+                    {
+                        // Looks like we only support a single statistics provider that can be any of the publisher interfaces. fml
+                        // TODO: Kill our statistics system.. please!?
```
ReubenBond: +1

src/Orleans.Core/Configuration/Legacy/LegacyConfigurationExtensions.cs
jason-bragg: For clarification these are just placeholders until we've time to think through how to implement a client lifecycle.  Having said that:

Do you mean subscribe to the same lifecycle or same lifecycle stage?

I don't think they can have the same lifecycle as the silo and client are, in theory, entirely separate, and in most cases even different processes.  While I've considered some sort of distributes startup logic for cluster start, that's a distant consideration for possible future enhancements, and even in that model I'm unconvinced clients can reliably take part in the cluster lifecycle unless we assume clients are part of the cluster.

As far as stages, as we flesh out the client lifecycle, I do think it would be good to mirror the stages where applicable, but I'm not yet familiar enough with the client initialization/shutdown logic to be confident of any design decisions without further investigation.  If you've more familiarity and can point out which stages of the silo lifecycle should be part of the client lifecycle can attempt a more accurate set of stages.

src/Orleans.Core/Configuration/ProviderConfiguration.cs
jason-bragg: wanted to.. it still has legit uses in other parts of the code :/

src/Orleans.Core/Configuration/Legacy/LegacyConfigurationExtensions.cs
ReubenBond: perhaps they could use a separate lifecycle which is shared, like `IStreamProviderLifecycle` or `IProviderLifecycle` or something more broad, like `ISystemLifecycle`?

Separate client and silo lifecycles might not be needed: it seems fine to me if a particular lifecycle stage isn't mapped to any actual event on the client.

src/Orleans.Core/Configuration/ProviderConfiguration.cs
ReubenBond: 🤮

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+                {
+                    foreach (IProviderConfiguration providerConfig in providerGroup.SelectMany(kvp => kvp.Value.Providers.Values))
+                    {
+                        RegisterProvider<IProvider>(providerConfig, services, defaultInitStage, defaultStartStage);
+                    }
+                }
+            }
+        }
+
+        private class ProviderLifecycleParticipant<TService> : ILifecycleParticipant<TLifeCycle>
```
ReubenBond: nit: Lifecycle doesn't require capitalization on the c

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -0,0 +1,168 @@
...
+            private readonly IServiceProvider services;
+            private readonly LegacyProviderConfigurator.ScheduleTask schedule;
+            private readonly int defaultInitStage;
+            private int initStage;
+            private readonly int defaultStartStage;
+            private int startStage;
+
+            public ProviderLifecycleParticipant(IProviderConfiguration config, IServiceProvider services, ILoggerFactory loggerFactory, int defaultInitStage, int defaultStartStage)
+            {
+                this.logger = loggerFactory.CreateLogger(config.Type);
```
ReubenBond: This Type string is free-form, so it might look ugly and non-uniform in the logs - possibly `namespace.type, assembly`? Maybe including the strong name of the assembly?

jason-bragg: All we have at this point is the "Type" string provided by the applications configuration.  We can't even convert it to an actual type at this point.  Open to other suggestions but our choices are somewhat limited..

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
jason-bragg: Not currently in scope.  Suggest at planning meeting?

src/Orleans.Core/Streams/Internal/StreamConsumerExtension.cs
```diff
@@ -125,12 +125,16 @@ public Task<StreamHandshakeToken> DeliverImmutable(GuidId subscriptionId, Stream
             }
             else if(this.streamSubscriptionObserver != null)
             {
-                var subscriptionHandlerFactory = new StreamSubscriptionHandlerFactory(this.providerManager, streamId, streamId.ProviderName, subscriptionId);
-                await this.streamSubscriptionObserver.OnSubscribed(subscriptionHandlerFactory);
-                // check if an observer were attached after handling the new subscription, deliver on it if attached
-                if (allStreamObservers.TryGetValue(subscriptionId, out observer))
+                var streamProvider = this.providerRuntime.ServiceProvider.GetServiceByName<IStreamProvider>(streamId.ProviderName);
```
ReubenBond: should the results of this be cached somewhere?

ReubenBond: It also seems like 'GetObserver' logic could be split into a new method and shared between the two

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
jason-bragg: We perform a groupby, so this should only happen once.

src/Orleans.Core/Streams/Internal/StreamImpl.cs
```diff
@@ -160,7 +160,9 @@ internal IInternalAsyncObservable<T> GetConsumerInterface()
 
         private IInternalStreamProvider GetStreamProvider()
         {
-            return this.runtimeClient.CurrentStreamProviderManager.GetProvider(streamId.ProviderName) as IInternalStreamProvider;
+            var streamProvider = this.runtimeClient.ServiceProvider.GetServiceByName<IStreamProvider>(streamId.ProviderName)
+                ?? throw new KeyNotFoundException(streamId.ProviderName);
+            return streamProvider as IInternalStreamProvider;
```
ReubenBond: (out of scope) this down-cast seems to limit the ability for users to implement `IStreamProvider` on their own, without using one our ordained base classes like `PersistentStreamProvider`

src/Orleans.Core/Utils/KeyedService.cs
```diff
@@ -6,26 +6,53 @@
...
-            return Equals(key, other);
+            return Equals(Key, other);
+        }
+    }
+
+    public class KeyedSingletonService<TKey, TService> : IKeyedService<TKey, TService>
+    where TService : class
+    {
+        Func<IServiceProvider, TKey, TService> factory;
+        private TService instance;
```
ReubenBond: Is there scope for concurrency issues? Maybe we need a lock in `GetService`?

src/Orleans.Core/Streams/Internal/StreamConsumerExtension.cs
jason-bragg: This should only be done on streams that do not already have an observer attached, and only for the first time that happens per stream.  While it's most likely that all streams on a grain will be on the same provider, multiple stream providers are supported, so we can't assume the same stream provider.  With this, I suspect a local cache would be of limited value.  The minimal possible perf gain would, imo, not justify any additional complexity.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
ReubenBond: Is it possible that we can resolve the string to a real type earlier on?

src/Orleans.Core/Streams/Internal/StreamConsumerExtension.cs
ReubenBond: Fair enough, this code is still fairly foreign to me.

For example, I don't see how `allStreamObservers.TryGetValue(subscriptionId, out observer)` will ever evaluate to `false` above and then `true` below, but I suppose the idea is that the `IStreamSubscriptionObserver` has a chance to subscribe below.

src/Orleans.Runtime/Lifecycle/SiloLifecycleStage.cs
```diff
@@ -16,14 +16,24 @@ public enum SiloLifecycleStage
         /// </summary>
         RuntimeServices = 2000,
 
+        /// <summary>
+        /// Start runtime services
```
ReubenBond: nit: the comment here and below are dupes

src/Orleans.Core/Configuration/Legacy/LegacyConfigurationExtensions.cs
jason-bragg: Ah, I see.  I'm not 100% opposed to this, but I think we should delay such a decision.  My reasoning is that most provider are silo only, and those that can also run on the client will require different initialization logic when running on client vs silo.  This has been a lingering issue in the stream providers which run on both.  I think having two distinct lifecycles is actually the right thing to do as it supports distinct initialization for the distinct environments.  In any case, I am of the opinion that we'll be in a better position to judge this after we've the experience of porting some of our providers to use lifecycles.

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -773,7 +706,33 @@ public Task StopAsync(CancellationToken cancellationToken)
             return this.siloLifecycle.OnStop(cancellationToken);
         }
 
-        private async Task OnRuntimeServicesStop(CancellationToken cancellationToken)
+        private Task OnRuntimeServicesStop(CancellationToken cancellationToken)
+        {
+            if (!cancellationToken.IsCancellationRequested)
+            {
+                // 3: Deactivate all grains
```
ReubenBond: nit: these numbers are inconsistent

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
jason-bragg: maybe.  Let me look into that.  Initially I thought we needed to wait until the assembly loading (parts) had occurred, so I delayed it as late as possible, but I'm less convinced of that now.  Will take a look.  Resolving the actual types here would make this code cleaner.

src/Orleans.Core/Streams/Internal/StreamImpl.cs
jason-bragg: This logic is shared code between our sms and persistent stream providers.  It's not associated with an extensibility point.  I agree that this is not 'good', but I doesn't really limit extensibility as the supported extensibility points are at the provider level or adapter level.

src/Orleans.Core/Utils/KeyedService.cs
jason-bragg: Will look into.

src/Orleans.Core/Core/ClusterClient.cs
```diff
@@ -37,8 +41,26 @@ public ClusterClient(OutsideRuntimeClient runtimeClient, ClientConfiguration con
...
+            foreach (ILifecycleParticipant<IClusterClientLifecycle> participant in lifecycleParticipants)
+            {
+                participant.Participate(clusterClientLifecycle);
+            }
+
+            // register all named lifecycle participants
+            IEnumerable<IKeyedServiceCollection<string, ILifecycleParticipant<IClusterClientLifecycle>>> namedLifecycleParticipantCollections = this.ServiceProvider.GetServices<IKeyedServiceCollection<string, ILifecycleParticipant<IClusterClientLifecycle>>>();
+            foreach (ILifecycleParticipant<IClusterClientLifecycle> participant in namedLifecycleParticipantCollections
+                .SelectMany(c => c.GetServices(this.ServiceProvider)
+                .Select(s => s.GetService(this.ServiceProvider))))
```
ReubenBond: I think it's valid for a service provider factory to return `null` (you're relying on it in the change to `VersionStoreGrain`), so maybe this should be `s?.GetService(...)` and have a `.Where(s => s != null)` clause added

src/Orleans.Runtime/Silo/Silo.cs
jason-bragg: cut'n paste code.. didn't really look at too hard.. will fix though :)

src/Orleans.Runtime/Versions/GrainVersionStore.cs
```diff
@@ -14,21 +17,13 @@ internal class GrainVersionStore : IVersionStore
...
-            this.IsEnabled = false;
-        }
-
-        public void SetStorageManager(IStorageProviderManager storageProviderManager)
-        {
-            IStorageProvider unused;
-            IsEnabled = storageProviderManager.TryGetProvider(
-                ProviderConstants.DEFAULT_STORAGE_PROVIDER_NAME,
-                out unused);
+            this.IsEnabled = services.GetService<IStorageProvider>() != null;
```
ReubenBond: I'm not actually sure if it's fine for a factory to return `null` - I'm not seeing it in the DI specification tests https://github.com/aspnet/DependencyInjection/blob/dev/src/DI.Specification.Tests/DependencyInjectionSpecificationTests.cs

It might be worth quickly clarifying, or alternatively changing the code which registers this default provider to only register it if a keyed service with that name exists.

ReubenBond:
Should we have an `IConfigurationValidator` to ensure that only one provider of each interface (`IStorageProvider`, etc) is registered for each name? i.e, only one storage provider registered for the name "users"

src/Orleans.Runtime/Versions/GrainVersionStore.cs
jason-bragg: Other places in our code make this same assumption, and we use GetRequiredService when null is not valid.  If this is not safe we'll need to open an issue and address all instances of this.  I suggest we leave this logic as is, and loop back on the possible improper use of get service in a future work item.

ReubenBond: Sounds fine to me, it's not a major concern: any issues here will blow up the silo on startup anyway.

jason-bragg:
Regarding IConfigurationValidator to prevent provider duplication, provider configurations in the config object are kept in dictionaries, so duplicates should not be possible.  As for general validation, as this is legacy support, I don't think we should spend time validating the configuration beyond what the existing configuration objects already offer.  This is not the primary path, and it's unclear that further enhancements will provide sufficient value to justify their efforts.  If such enhancements become necessary, they can be added when demand is demonstrated.

ReubenBond:
LGTM. FYI this PR has conflicts

PR #3748, opened 2017-12-06
[1.5.x] CodeGen: support builds which use reference assemblies

ReubenBond:
Backported codegen fixes to 1.5.x branch.
Supports resolving assemblies using `*.deps.json` file.
This includes some divergences from the current master, notably:
* We ensure the deps is generated via `AfterTargets="GenerateBuildDependencyFile"`
* We copy the generated deps file to sit next to the intermediate assembly, since I couldn't see any other way to make MSBuild emit the deps file to that location (not without triggering the target manually).


sergeybykov:
Please resubmit against 1.5.3 branch instead.

PR #3755, opened 2017-12-06
Make client stream test run more deterministic

xiazen:
Make it pass more often by increase waiting time


jason-bragg:
This is a general test harness used against many stream providers.  I don't think it's safe to assume that all stream providers have the same characteristics in regards to producer/consumer behaviors.  For instance sms uses (unless configured for fire&forget) a direct call from producer to consumer, hence it's produce call will fail if consumer has not received messages, whereas the memory provider will write it to a queue grain and return without the consumer necessarily getting the message.  The safe way to test this, imo, is to ensure that when we successfully produce events, that the consumers receive them.

The non-deterministic part of this test should be related to delivery time, not behavior, unless we consider delivery time a requirement for test failure.  We do eventually need to give up on the test and move on, so some timeout is needed, but I suggest we simply raise the timeout rather than removing coverage that the messages actually reached the consumer.

PR #3758, opened 2017-12-07
Expedite gateway retries when gateway list is exhausted

ReubenBond:
When a client runs out of gateways (all are marked as dead), we can force-retry all gateways which were previously marked as dead.

Hopefully this would improve reconnect times in development/on-prem scenarios in cases where a cluster reboot is not coordinated, especially in single silo scenarios.

Low pri


sergeybykov:
@dotnet-bot test this

sergeybykov:
@dotnet-bot test this please

jdom:
@dotnet-bot test functional

amccool:
Is this something that is going into the 1.5x branch, or will remain in master destined for 2.x?

ReubenBond:
2.0 only, since it's low impact and not critical. For 1.5 you can still improve the experience by reducing the GatewayRefreshPeriod setting on the client. Setting it to something as low as 5 or 10 seconds should be suitable

ReubenBond:
@amccool regarding your question above, this is included in 1.5.4

PR #3767, opened 2017-12-08
Docs: Code Generation in Orleans 2.0

ReubenBond:
Simple docs on codegen in 2.0, mostly about how it differs from 1.5 and the user's options.

### DIFF
--- a/src/Documentation/2.0/Codegen.md
+++ b/src/Documentation/2.0/Codegen.md
@@ -1,6 +1,14 @@
 ---
 layout: page
-title: Code generation in Orleans 2.0
+title: Code Generation in Orleans 2.0
 ---
 
-# Placeholder for Code Generation documentation
+# Code Generation in Orleans 2.0
+
+Code generation has been improved in Orleans 2.0, improving startup times and providing a more deterministic, debuggable experience. As with earlier versions, Orleans provides both build-time and run-time code generation.
+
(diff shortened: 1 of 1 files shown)


src/Documentation/2.0/Codegen.md
```diff
@@ -1,6 +1,14 @@
...
+title: Code Generation in Orleans 2.0
 ---
 
-# Placeholder for Code Generation documentation
\ No newline at end of file
+# Code Generation in Orleans 2.0
+
+Code generation has been improved in Orleans 2.0, improving startup times and providing a more deterministic, debuggable experience. As with earlier versions, Orleans provides both build-time and run-time code generation.
+
+1.	**During Build** - This is the recommended option and only supports C# projects. In this mode, code generation will run every time your project is compiled. A build task is injected into your project's build pipeline and the code is generated in the project's intermediate output directory. To activate this mode, add the package `Microsoft.Orleans.OrleansCodeGenerator.Build` to all projects which contain **Grains**, **Grain Interfaces**, or **Serializers**. Additional diagnostics can be emitted at build-time by specifying value for `OrleansCodeGenLogLevel` in the target project's *csproj* file. For example, `<OrleansCodeGenLogLevel>Trace</OrleansCodeGenLogLevel>`.
```
jdom: More than project that contain serializers, I would say that contain types that will be used for messaging and would require serializers to be generated for them

src/Documentation/2.0/Codegen.md
```diff
@@ -1,6 +1,14 @@
...
+
+2.	**During Configuration** - This is the only supported option for F#, Visual Basic, and other non-C# projects. This mode generates code during the configuration phase. To access this, see the [Configuration](Configuration2.0.md) documentation.
```
jdom: Well, alternatively you can have a c# project reference the f# project and still do build time codegen, correct? Wouldn't that still be preferable?

src/Documentation/2.0/Codegen.md
ReubenBond: will fix

src/Documentation/2.0/Codegen.md
ReubenBond: That's correct. I didn't want to suggest something complicated like that, though. I figure they can work it out or we can tell anyone who has an issue with using runtime codegen

PR #3773, opened 2017-12-08
Revise eventsourcing api

sebastianburckhardt:
One commit covers several small revisions to the event sourcing API.
 - use the shortest name `RaiseEvent` for the most common expected case, i.e. conditional events that are not retried on races.
 - throw a descriptively named exception (`LostRaceException`) if a raised event loses a race to the shared log
 - use a descriptive name 'EnqueueEvent' for events that get queued, retried, and do not care about races

The other commit covers fixes and improvements to the retry logic and error diagnostics  
 - by default (if user does not add handler code to JournaledGrain), create warnings in Orleans log for all connection errors
 - fix bug that caused retry delay to not be set correctly
 - allow users to customize retry timing, suspension, resumption
 - change default policy: pause retries for six hours if there is no application activity for 20 minutes

Once this is done I will need to update the documentation accordingly as well.


lmagyar:
@sebastianburckhardt Shouldn't this be merged in v2.0? This is even small, but a breaking change in the API.

sebastianburckhardt:
@Imagyar I agree with that suggestion.

I am still waiting for @jason-bragg to take a look, he has been very busy with other work.

jason-bragg:
We were hoping to fit this in for 2.0, but it doesn't look like it's going to make it.  Post 2.0 we'll revisit prioritization and see where this fits in.  I fully understand that this PR and other Event Sourcing changes have been pushed off for months now.  I can't commit to a timeframe at this point, only that we'll revisit the prioritization post 2.0 and should be in a better position to set expectations then.

src/Orleans.EventSourcing/JournaledGrain.cs
```diff
@@ -56,60 +56,73 @@ protected JournaledGrain(IGrainIdentity identity, IGrainRuntime runtime)
...
         /// <summary>
-        /// Raise an event.
+        /// Raises an event, and either confirms it, or throws <see cref="LostRaceException"/> 
+        /// and discards it (if another instance raised an event in the meantime).
         /// </summary>
         /// <param name="event">Event to raise</param>
-        /// <returns></returns>
-        protected virtual void RaiseEvent<TEvent>(TEvent @event) 
+        /// <returns>a task that completes when the event has been confirmed</returns>
+        protected async Task RaiseEvent<TEvent>(TEvent @event)
```
rikbosch: Migrations of old code will be error-prone when this method is not awaited. Maybe it is better to rename the method, so code breaks at compile time instead of runtime.

sergeybykov:
Looking at this PR again after a **very long** time. There's a number of breaking changes here. I don't see a way to take this without breaking all existing users of `JournaledGrain`. A parallel implementation of it could be introduced instead. But I'm not sure if that would be worthwhile.

sebastianburckhardt:
As far as I remember, the only breaking changes are name changes that do not affect functionality. Those were suggested by @jdom and @ReubenBond when we had a meeting to discuss this, a long time ago. I guess at that time, introducing a breaking change seemed not a major issue, and we were trying to improve the names slightly.

At this point of time, I would agree that changing the names seems neither necessary nor a good idea. I think it makes sense to keep the current names and not change the API. I can revise this accordingly.

jdom:
It's been too long to remember, but I think I recall concluding that JournaledGrain wasn't really usable as is, so there were probably no consumers of it. Perhaps we should ask. Every time I heard people doing ES in Orleans, they had their own infrastructure for it

sebastianburckhardt:
I don't think "unusable" is the right word, as people are apparently using it. It just appears they implemented their own LogConsistencyProvider since the included providers were not quite right for their purposes. We knew that all along.

@jkonecki and I implemented a better provider (for EventStore) in a PR but it all got stuck for some reason, other things being considered more important by the Orleans team at the time (e.g. the whole 2.0 overhaul). If there is interest in continuing this line of work it can still happen.

jkonecki:
I would be happy to work on event sourcing support - I have a library that
implements event store on top of CosmosDB which will be easy to integrate
(currently using it directly from grains, OnActivate)

jdom:
Good to hear that you could take it forward. I assume that with the revamp we did in 2.0 and with GrainActivationContext, all of this can actually be implemented without any changes to the core runtime, correct? The Orleans team is no longer modifying the core when it can be avoided. So for example in the Transactions work -which requires lots of hooks into the grain activation lifecycle- they were able to implement it on top of the current extension points, as would any 3rd party customer would be able to do so without forking the code.

jkonecki:
I don't think I'll be able to do everything on my own: I'm not familiar
with core changes as I'm currently using 1.5.x and I haven't looked how
transactions are implemented.

sergeybykov:
We just talked with @sebastianburckhardt and agreed to revamp this PR and add `EnqueueEvent*` methods added in parallel with the current `RaiseConditionalEvent` that will get marked as `[Obsolete]`. This will make the change non-breaking for people that use `JournaledGrain` today.

@jkonecki Your contribution will be very welcome. We can help with putting it in the right context in the 2.0/3.0 world. It doesn't have to be confined by `JournaledGrain`. We are for introducing alternative APIs in this space because Event Sourcing appears to be one of the most opinionated areas in software, and it's very hard to get everyone agree to a single API. 🙂

src/Orleans.EventSourcing/JournaledGrain.cs
sebastianburckhardt: Good point. I think for this particular method, we can just use `RaiseEventAsync`, which is a better name anyway.

sergeybykov:
Closing as this got much outdated.

PR #3780, opened 2017-12-11
Move test running to dotnet xunit from xunit console runner

attilah:


Test.cmd
```diff
@@ -7,6 +7,10 @@ SET CMDHOME=%~dp0
 @REM Remove trailing backslash \
 set CMDHOME=%CMDHOME:~0,-1%
 
+call Ensure-DotNetSdk.cmd
+
+set /p _dotnet=< "%~dp0build\dotnetexe.txt"
```
jdom: Why do you have to do this here but not in Build.cmd?

attilah: the ```set``` statement is a leftover, I've removed that in the commit below. ```Ensure-DotNetSdk.cmd```s responsibility when running after ```Build.cmd``` is just to set ```_dotnet``` variable to the correct (and same) ```dotnet.exe``` to be used by test runner PowerShell.

PR #3785, opened 2017-12-12
Refactor transaction abstractions to enable injection of alternate protocols

sebastianburckhardt:
This PR is a starting point for letting us run the next version of the transaction algorithm.

It creates extension points, by abstracting
 (1) the TransactionInfo (using an interface ITransactionInfo)
 (2) the (prepare/abort/commit) messages on the ITransactionalResource interface

Therefore, an injected TransactionAgent can use alternative implementations of TransactionInfo, and a different commit protocol for transactional resources.


src/Orleans.Core/Transactions/ITransactionalResource.cs
```diff
@@ -9,40 +9,43 @@ namespace Orleans.Transactions
...
 
         /// <summary>
-        /// Notification of a transaction abort.
+        /// Handle a remote request that returns a response or an exception.
         /// </summary>
-        /// <param name="transactionId">Id of the aborted transaction</param>
-        Task Abort(long transactionId);
+        /// <typeparam name="T">the type of the response</typeparam>
+        /// <returns>A task that completes when the response is received</returns>
+        Task<T> Handle<T>(IRequestWithResponse<T> call);
```
jason-bragg: Why move from well defined calls to generic handlers.  This introduces more runtime risk and maintenance cost related to future documenting, implementing, refactoring and testing.

sebastianburckhardt: The point is to allow extensionality, i.e. avoid hard-coding a specific facet interface that all facets have to use (which goes against the idea of using facets for a wide range of purposes, and means we cannot have multiple transaction algorithms available at the same time). The current facet transactional interface (Prepare/Commit/Abort) does not exist the same way in the new transaction algorithm, rather the new facet transactional interface will be (Prepare/Commit/Prepared/Abort/Confirm/Cancel/Collect/Ping), some of which are requests with responses and some of which are one-way messages. 

The mechanism introduced here allows facets to define their own "interfaces" (currently the facet system does not permit that), and therefore allows the new transaction algorithm to be injected. This helps maintenance and keeps `Orleans.Core` free from implementation-specifics.

Runtime risk and maintenance cost are mitigated as follows:
 - the exact same information as in the interface definition, including all the same comments, still exist. They are just expressed using classes instead of an interface, and have moved to the file `TransactionalResourceCommunication.cs` inside `Orleans.Runtime/Transactions`
 (note: I just noticed that I accidentally deleted one of the comments, will put it back)
 - all requests and responses are typed, thus type errors are still caught by the compiler

sebastianburckhardt:
Based on our discussion yesterday, I backed out the generalization of the `ITransactionalResource` interface, and instead created a new `ITransactionalParticipant` interface.

I used the same pattern throughout, and copied all the code needed for the plumbing.

src/Orleans.Core/Transactions/ITransactionParticipant.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <param name="participant">The participant sending the message</param>
+        Task Ping(Guid transactionId, DateTime timeStamp, ITransactionParticipant participant);
+    }
+
+
+    /// <summary>
+    /// Used to propagate information about the status of a transaction. Used for transaction orchestration, for diagnostics, 
+    /// and for generating informative user exceptions
+    /// </summary>
+    public enum Status
```
jason-bragg: Can we name this something more specific?  TransactionStatus? ..?

src/Orleans.Runtime/Transactions/TransactionAgent.cs
```diff
@@ -203,8 +205,10 @@ public TransactionAgent(ILocalSiloDetails siloDetails, ITransactionManagerServic
             await completion.Task;
         }
 
-        public void Abort(TransactionInfo transactionInfo, OrleansTransactionAbortedException reason)
+        public void Abort(ITransactionInfo info, OrleansTransactionAbortedException reason)
         {
+            var transactionInfo = (TransactionInfo)info;
```
jason-bragg: This violates strong behavioral subtyping, which we do in some cases, but never without reason.
Under what conditions would info not be a TransactionInfo?  Can we throw a more meaningful error in this case?

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -72,7 +72,7 @@ public TransactionalState(ITransactionalStateConfiguration transactionalStateCon
         /// </summary>
         public void Save()
         {
-            var info = TransactionContext.GetTransactionInfo();
+            var info = (TransactionInfo) TransactionContext.GetTransactionInfo();
```
jason-bragg: As with other areas where we hard cast, avoid if we can, and if we can't avoid, we should throw meaningful exceptions.

src/Orleans.Core/Transactions/TransactionContext.cs
```diff
@@ -13,18 +11,18 @@ public static class TransactionContext
         internal const string TransactionInfoHeader = "#TC_TI";
         internal const string Orleans_TransactionContext_Key = "#ORL_TC";
 
-        public static TransactionInfo GetTransactionInfo()
+        public static ITransactionInfo GetTransactionInfo()
```
jason-bragg: Upon further reflection, since we're allowing ITransactionInfo to be a general interface, but specific systems will need aspecific implementation (which seems to be the current pattern) maybe the context can have a GetTransactionInfo<T>() where T : ITransactionInfo call which returns null if the specific type of ITransactionInfo has not been set.  Then all of the related systems need only perform null checks (which they 'should' be doing anyway, but are not).

sebastianburckhardt: I have added an overload GetRequiredTransactionInfo<T>() which throws an informative exception if the info is null or the wrong type.

src/Orleans.Core/Transactions/ITransactionParticipant.cs
sebastianburckhardt: Sure. Currently fighting with VS to let me rename.

src/Orleans.Runtime/Transactions/TransactionAgent.cs
sebastianburckhardt: I don't see any circumstance in which this could be the wrong type.

src/Orleans.Core/Transactions/ITransactionParticipant.cs
sebastianburckhardt: Ok, done.

jason-bragg:
@dotnet-bot test functional

PR #3788, opened 2017-12-13
Add processing time measure on silo start up sequence and code gen

xiazen:
Add processing time measure 
- silo start up
- built time code gen
- runtime code gen

AFAIK, there's only two participants on silo lifecycle now, which is silo itself, and InsideRuntimeClient. So I added processing time measure to their OnRuntimeInitializeStart and OnRuntimeServicesStart methods. 
Let me know if I missed anything on silo start up sequence

### DIFF
--- a/src/Orleans.CodeGeneration.Build/Program.cs
+++ b/src/Orleans.CodeGeneration.Build/Program.cs
@@ -120,6 +120,7 @@ namespace Orleans.CodeGeneration
                 }
 
                 // STEP 5 : Finally call code generation
+                var stopWatch = Stopwatch.StartNew();
                 if (referencesOrleans)
                 {
                     if (!CodeGenerator.GenerateCode(options))
@@ -132,7 +133,8 @@ namespace Orleans.CodeGeneration
                     Console.WriteLine("ERROR: Orleans-CodeGen - the input assembly does not reference Orleans and therefore code can not be generated.");
                     return -2;
                 }
-
+                stopWatch.Stop();
+                Console.WriteLine($"Built time code generation for assembly {options.InputAssembly} took {stopWatch.ElapsedMilliseconds} miliseconds");
                 // DONE!
                 return 0;
             }
--- a/src/Orleans.CodeGeneration.Build/Program.cs
+++ b/src/Orleans.CodeGeneration.Build/Program.cs
@@ -134,7 +134,7 @@ namespace Orleans.CodeGeneration
                     return -2;
                 }
                 stopWatch.Stop();
-                Console.WriteLine($"Built time code generation for assembly {options.InputAssembly} took {stopWatch.ElapsedMilliseconds} miliseconds");
+                Console.WriteLine($"Build time code generation for assembly {options.InputAssembly} took {stopWatch.ElapsedMilliseconds} milliseconds");
                 // DONE!
                 return 0;
             }
--- a/src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
+++ b/src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
@@ -1,4 +1,7 @@
-﻿using Microsoft.Extensions.Logging.Abstractions;
+﻿using System;
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Orleans.ApplicationParts;
 using Orleans.CodeGenerator;
 using Orleans.Metadata;
@@ -16,9 +19,11 @@ namespace Orleans.Hosting
         /// Generates support code for the the provided assembly and adds it to the builder.
         /// </summary>
         /// <param name="manager">The builder.</param>
+        /// <param name="logger">optional logger</param>
         /// <returns>A builder with support parts added.</returns>
-        public static IApplicationPartManagerWithAssemblies WithCodeGeneration(this IApplicationPartManagerWithAssemblies manager)
+        public static IApplicationPartManagerWithAssemblies WithCodeGeneration(this IApplicationPartManagerWithAssemblies manager, ILogger logger = null)
         {
+            var stopWatch = Stopwatch.StartNew();
             var tempPartManager = new ApplicationPartManager();
             foreach (var provider in manager.FeatureProviders)
             {
@@ -38,7 +43,8 @@ namespace Orleans.Hosting
             
             var codeGenerator = new RoslynCodeGenerator(tempPartManager, new NullLoggerFactory());
             var generatedAssembly = codeGenerator.GenerateAndLoadForAssemblies(manager.Assemblies);
-
+            stopWatch.Stop();
+            logger?.LogInformation(0, $"Runtime code generation for assemblies {String.Join(",", manager.Assemblies.ToStrings())} took {stopWatch.ElapsedMilliseconds} milliseconds");
             return manager.AddApplicationPart(generatedAssembly);
         }
     }
--- a/src/Orleans.Runtime/Core/InsideRuntimeClient.cs
+++ b/src/Orleans.Runtime/Core/InsideRuntimeClient.cs
@@ -722,8 +722,11 @@ namespace Orleans.Runtime
 
         private Task OnRuntimeInitializeStart(CancellationToken tc)
         {
+            var stopWatch = Stopwatch.StartNew();
             typeManager.Start();
             GrainTypeResolver = typeManager.GetTypeCodeMap();
+            stopWatch.Stop();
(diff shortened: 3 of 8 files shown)


src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
```diff
@@ -16,9 +19,11 @@ public static class ApplicationPartManagerCodeGenExtensions
         /// Generates support code for the the provided assembly and adds it to the builder.
         /// </summary>
         /// <param name="manager">The builder.</param>
+        /// <param name="logger">optional logger</param>
         /// <returns>A builder with support parts added.</returns>
-        public static IApplicationPartManagerWithAssemblies WithCodeGeneration(this IApplicationPartManagerWithAssemblies manager)
+        public static IApplicationPartManagerWithAssemblies WithCodeGeneration(this IApplicationPartManagerWithAssemblies manager, ILogger logger = null)
```
xiazen: if user want to enable processing time measure on runtime code gen, they need to provider a logger

src/Orleans.CodeGeneration.Build/Program.cs
```diff
@@ -132,7 +133,8 @@ public static int Main(string[] args)
                     Console.WriteLine("ERROR: Orleans-CodeGen - the input assembly does not reference Orleans and therefore code can not be generated.");
                     return -2;
                 }
-
+                stopWatch.Stop();
+                Console.WriteLine($"Built time code generation for assembly {options.InputAssembly} took {stopWatch.ElapsedMilliseconds} miliseconds");
```
ReubenBond: s/Built/Build
s/miliseconds/milliseconds
I think `ms` as a suffix would be fine, too

src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
ReubenBond: Ah thanks for adding this, it can be useful. The doc comment should start with a capital.

src/Orleans.CodeGeneration.Build/Program.cs
xiazen: My English teacher is crying in the corner...

src/Orleans.Runtime/Lifecycle/SiloLifecycle.cs
```diff
@@ -1,12 +1,28 @@
...
     internal class SiloLifecycle : LifecycleObservable, ISiloLifecycle
     {
-        public SiloLifecycle(ILoggerFactory loggerFactory) : base(loggerFactory)
+        private SiloLifecyclePerfReporter perfReporter;
+        public SiloLifecycle(ILoggerFactory loggerFactory, SiloLifecyclePerfReporter reporter) : base(loggerFactory)
         {
+            this.perfReporter = reporter;
+        }
+
+        public override async Task OnStart(CancellationToken ct)
```
xiazen: @jason-bragg let me know if you are ok with this, marking OnStart and OnStop as `virtual `on LifeCycleObservable. I think it will be useful for people who want to use LifecycleObservable implementation but also add some custom logic

Orleans.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27004.2010
+VisualStudioVersion = 15.0.27004.2009
```
attilah: We should not submit file changes in PRs like this. Please exclude this change from the PR.

src/Orleans.CodeGeneration/ApplicationPartManagerCodeGenExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static class ApplicationPartManagerCodeGenExtensions
         /// Generates support code for the the provided assembly and adds it to the builder.
         /// </summary>
         /// <param name="manager">The builder.</param>
-        /// <param name="logger">optional logger</param>
+        /// <param name="logger"> Optional logger</param>
```
attilah: There is an extra space char in the text.

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -725,7 +725,13 @@ private Task OnRuntimeInitializeStart(CancellationToken tc)
             typeManager.Start();
             GrainTypeResolver = typeManager.GetTypeCodeMap();
             stopWatch.Stop();
-            this.logger.Info(ErrorCode.SiloStartPerfMeasure, $"Start InsideRuntimeClient took {stopWatch.ElapsedMilliseconds} Milliseconds");
+            var report = new LifeCyclePerfReport()
```
attilah: What about storing the ```TimeSpan``` value in the report instead of the formatted string, since that string can be generated from the ```TaskName``` and the ```Duration``` (if we call it that).

src/Orleans.Runtime/Lifecycle/SiloLifecycle.cs
```diff
@@ -1,12 +1,28 @@
...
 
 namespace Orleans.Runtime
 {
     internal class SiloLifecycle : LifecycleObservable, ISiloLifecycle
     {
-        public SiloLifecycle(ILoggerFactory loggerFactory) : base(loggerFactory)
+        private SiloLifecyclePerfReporter perfReporter;
+        public SiloLifecycle(ILoggerFactory loggerFactory, SiloLifecyclePerfReporter reporter) : base(loggerFactory)
         {
+            this.perfReporter = reporter;
```
attilah: incoming ctor param should also called ```perfReporter```

src/Orleans.Runtime/Silo/SiloLifecyclePerfReporter.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Microsoft.Extensions.Logging;
+
+namespace Orleans.Runtime
+{
+    internal class SiloLifecyclePerfReporter
```
attilah: Don't we want to have a public interface for this? I think it could be beneficial for others as well, our implementation could stay internal.

The public interface should contain the ```AddReport``` method and a property or method to get the added reports, but NOT the ```FlushReports```, which is our implementation.

FlushReports could be an extension method or something similar, but the ```SiloLifecyclePerfReporter``` class should be "as dump as possible" IMHO.

src/Orleans.Runtime/Silo/SiloLifecyclePerfReporter.cs
```diff
@@ -0,0 +1,49 @@
...
+                this.logger?.Info(ErrorCode.SiloStartPerfMeasure, $"===========Start reporting on ring {stageReports.Key}===========");
+                foreach (var report in stageReports.Value)
+                {
+                    this.logger?.Info(ErrorCode.SiloStartPerfMeasure, report.Report);
+                }
+            }
+        }
+    }
+
+    internal class LifeCyclePerfReport
```
attilah: ```LifeCycle``` casing. "C" is not capital.

All properties should be immutable.

src/Orleans.Runtime/Lifecycle/SiloLifecycle.cs
jason-bragg: I'm fine with the calls being virtual, but the reporter mechanism seems clumsy.  If let's talk about the design in person.

xiazen: I'd call it minimalist :). The requirement of this feature wasn't clear to me, so In my mind, I just want a internal feature, which provide a place to cache all the reports for tasks run in silo lifecycle rings, and flush those reports when silo finish onStart or finish onStop. But I agree it has space to improve

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
xiazen: I think it is more flexible to store strings, it give user more flexibility to report metrics more than just time.

src/Orleans.Runtime/Silo/SiloLifecyclePerfReporter.cs
xiazen: It is unclear to me on the requirements of this feature, so in my mind, I wanted this feature as an internal feature first, and then we can work out the patterns in this feature by testing it in our internal silo lifecycle participants. If we feel the feature is mature, we can then release it as public.

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
attilah: But then we loose the most important part of perf, the time itself.

xiazen:
reverted to previous version and fixed typo

ReubenBond:
@dotnet-bot test functional

PR #3791, opened 2017-12-13
CodeGen: avoid potential confusion between overridden properties in a type hierarchy

### LINKED ISSUE #3786: Codegen is generating code for AspNetCore types
After update to beta2 today, one of my projects fail to build because codegen is generating code for some AspNet types. 

image

I tried to remove obj/ and bin/ and rebuild but got the same problem.

It may be related to #3771

ReubenBond:
Fixes #3786

Without this PR, a serializer for `SubClassOverridingAutoProp` will try to use `SubClassOverridingAutoProp.AutoProp` to get/set the backing field of the base class' `AutoProp`, which will result in an `InvalidCastException`.

Rather than try to disambiguate between the properties, this PR will avoid any properties which have been overridden and instead will access the backing field using our `IFieldUtils` getter/setter delegates.

```c#
public class BaseClassWithAutoProp
{
    public int AutoProp { get; set; }
}

public class SubClassOverridingAutoProp : BaseClassWithAutoProp
{
    public new string AutoProp { get => base.AutoProp.ToString(); set => base.AutoProp = int.Parse(value); }
}
```

### DIFF
--- a/src/Orleans.CodeGeneration/SerializerGenerator.cs
+++ b/src/Orleans.CodeGeneration/SerializerGenerator.cs
@@ -87,7 +87,7 @@ namespace Orleans.CodeGenerator
         /// </returns>
         internal static TypeDeclarationSyntax GenerateClass(string className, Type type, Action<Type> onEncounteredType)
         {
-            var typeInfo = type.GetTypeInfo();
+            var typeInfo = type;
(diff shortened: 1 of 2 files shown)


sergeybykov (approved):

jdom:
@dotnet-bot test functional
Reason: unrelated failure

PR #3793, opened 2017-12-13
Split OrleansSQLUtils into separate packages

attilah:
Split OrleansSQLUtils into separate packages:
- Orleans.Clustering.AdoNet
- Orleans.Persistence.AdoNet
- Orleans.Reminders.AdoNet
- Orleans.Statistics.AdoNet

Keep OrleansSQLUtils as Meta Package

Breaking change: Internalize shared types (like for Azure and AWS)


attilah:
@veikkoeeva please review it.

src/AdoNet/Orleans.Persistence.AdoNet/Orleans.Persistence.AdoNet.csproj
```diff
@@ -0,0 +1,87 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.Persistence.AdoNet</PackageId>
+    <Title>Microsoft Orleans Persistence AdoNet Providers</Title>
+    <Description>Microsoft Orleans persistence providers for AdoNet Providers</Description>
```
veikkoeeva: Should these "AdoNet"s be _ADO.NET_ here? Are these used as search words in Nuget? ADO.NET is the official term, I think.

src/AdoNet/Orleans.Reminders.AdoNet/Orleans.Reminders.AdoNet.csproj
```diff
@@ -0,0 +1,86 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.Reminders.AdoNet</PackageId>
+    <Title>Microsoft Orleans Reminders AdoNet Providers</Title>
+    <Description>Microsoft Orleans reminders provider for AdoNet Providers</Description>
```
veikkoeeva: As previously mentioned about "AdoNet" -> "ADO.NET", here too. I suppose in the other packages also.

src/AdoNet/Shared/Storage/DbExtensions.cs
```diff
@@ -7,12 +7,24 @@
...
+namespace Orleans.Tests.SqlUtils
+#else
+// No default namespace intentionally to cause compile errors if something is not defined
+#endif
 {
     /// <summary>
     /// Contains some convenience methods to use in conjunction with <see cref="IRelationalStorage">IRelationalStorage</see> and <see cref="RelationalStorage">GenericRelationalStorage</see>.
     /// </summary>
-    public static class DbExtensions
+    internal static class DbExtensions
```
veikkoeeva: Just a note about this and other changes to visibility, these can be used as a micro-ORM like Dapper, for instance. If someone has done that, this is a breaking change. If someone wish to do so, it's not possible anymore after this change.

src/AdoNet/OrleansSQLUtils/OrleansSQLUtils.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Microsoft.Orleans.OrleansSQLUtils</PackageId>
+    <Title>Microsoft Orleans SQL Utilities</Title>
+    <Description>Library of utility types for relational storage of Microsoft Orleans.</Description>
```
veikkoeeva: The other descriptions do not have '.' here, should this one? I don't know how it has been.

src/AdoNet/Shared/Storage/DbConnectionFactory.cs
```diff
@@ -5,7 +5,19 @@
 using System.Linq;
 using System.Reflection;
 
-namespace Orleans.SqlUtils
```
veikkoeeva: Could there be one main package that have these common parts to every other functionality which the persistence, reminders, statistics and membership implementations then reference? This way, it seems by reading this part, one could avoid some of the conditional namespaces.

src/AdoNet/Shared/Storage/DbExtensions.cs
attilah: Yes we're aware of that, but this is the solution we chose to share files between pieces.

src/AdoNet/Shared/Storage/DbConnectionFactory.cs
attilah: That was one another idea we had, having a "Core" package, but decided to share files instead.

veikkoeeva:
I don't think I have further to add. I would be intriqued to know about why sharing the files and change namespace instead of defining a core package since it seems a bit more complicated. Maybe that is also the reason for the change of some of the parts to internal in addition to, perhaps, reducing API surface.

PR #3794, opened 2017-12-14
Fix Exception serialization on .NET Core

ReubenBond:
Very small fix + some QoL test changes. I was unable to `dotnet xunit` the NetCore.Tests project without specifying `<RuntimeFrameworkVersion>`. It seems related to [these](https://github.com/dotnet/cli/issues/7901) [three](https://github.com/dotnet/core-setup/issues/3462) [issues](https://github.com/dotnet/core-setup/issues/3469).

Need to get NetCore.Tests running in CI - I'll experiment

### DIFF
--- a/src/Orleans.Core/Serialization/BinaryFormatterISerializableSerializer.cs
+++ b/src/Orleans.Core/Serialization/BinaryFormatterISerializableSerializer.cs
@@ -20,7 +20,7 @@ namespace Orleans.Serialization
         /// <inheritdoc />
         public bool IsSupportedType(Type itemType)
         {
-            return SerializableType.IsAssignableFrom(itemType) && DotNetSerializableUtilities.HasSerializationConstructor(itemType);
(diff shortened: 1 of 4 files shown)


test/NetCore.Tests/NetCore.Tests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
+   <RuntimeFrameworkVersion>2.0.3</RuntimeFrameworkVersion>
```
jdom: hmmm, @attilah do you have a recommendation to avoid this? Because we are installing 2.0.3 if nothing is installed already, but if it is, then not sure whether this would break...

ReubenBond: I'd love to use a different solution if anyone has an idea of where to look.

jdom:
@dotnet-bot test functional
Reason: unrelated failure

PR #3795, opened 2017-12-14
Remove legacy initialization from Silo class

jdom:
- Remove legacy initialization from Silo class.
- Remove most usages of `SiloInitializationParameters` concrete type.

This is in preparation for getting rid of the legacy configuration objects as a Core dependency

### DIFF
--- a/src/Orleans.Runtime/Hosting/LegacyClusterConfigurationExtensions.cs
+++ b/src/Orleans.Runtime/Hosting/LegacyClusterConfigurationExtensions.cs
@@ -58,7 +58,7 @@ namespace Orleans.Hosting
                 {
                     options.HasMultiClusterNetwork = true;
                     options.BackgroundGossipInterval = globals.BackgroundGossipInterval;
-                    options.DefaultMultiCluster = globals.DefaultMultiCluster?.ToList() ?? new List<string>();
+                    options.DefaultMultiCluster = globals.DefaultMultiCluster?.ToList();
                     options.GlobalSingleInstanceNumberRetries = globals.GlobalSingleInstanceNumberRetries;
                     options.GlobalSingleInstanceRetryInterval = globals.GlobalSingleInstanceRetryInterval;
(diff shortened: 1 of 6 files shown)


jdom:
BTW, tests passed even if they show as pending (click details to see the results). I also ran a smoke tests in our distributed tests environment, which uses SiloHost, and seems to be working fine.
Feel free to merge any time

src/Orleans.Runtime/Hosting/LegacyClusterConfigurationExtensions.cs
```diff
@@ -58,7 +58,7 @@ public static IServiceCollection AddLegacyClusterConfigurationSupport(this IServ
                 {
                     options.HasMultiClusterNetwork = true;
                     options.BackgroundGossipInterval = globals.BackgroundGossipInterval;
-                    options.DefaultMultiCluster = globals.DefaultMultiCluster?.ToList() ?? new List<string>();
+                    options.DefaultMultiCluster = globals.DefaultMultiCluster?.ToList();
```
ReubenBond: Is this a change in behavior?

samcov:
Does that mean ClusterConfiguration.LocalhostPrimarySilo(); is gone?

sergeybykov:
>Does that mean ClusterConfiguration.LocalhostPrimarySilo(); is gone? And if so, what is the new way to do that?

It got moved to Orleans.Core.Legacy.

samcov:
Thanks!!!

jdom:
Still, there will be a straightforward replacement that doesn't use the legacy configuration at all, but accomplishes the same. @ReubenBond can probably comment on this

samcov:
I hope so, I'm using core now and want to configure the way I did in the full framework with xml. 

To be honest, I'm just using the local stuff for now to get the code migrated, so I'm not in a rush, but it wasn't clear how to do that and I didn't want to use the stuff I knew was going to be deprecated.

PR #3796, opened 2017-12-15
Fix shutdown sequence in linuxcontainer

benjaminpetit:
Related to #3621 

Two fixes here:

- First, do not try to restart the listening socket when the silo is stopping
- Second, for some reason, in Linux calling `Socket.Disconnect` on a Socket that is not connected completely blow up the process instead of just throwing an exception. So check if the socket is connected before calling this method.


ReubenBond:
@dotnet-bot test bvt

PR #3807, opened 2017-12-19
Add transaction exceptions for distributed TM

sebastianburckhardt:
in preparation for the distributed TM implementation, some straightforward modifications of the application-visible exceptions:

 - use string instead of long for transaction ids (since distributed TM algorithm uses Guid, not long)
 - add some constructor overloads to some exceptions, so we can specify message and inner exception
 - add several new exception that describe failures specific to the distributed TM algorithm

### DIFF
--- a/src/Orleans.Core/Transactions/OrleansTransactionException.cs
+++ b/src/Orleans.Core/Transactions/OrleansTransactionException.cs
@@ -70,6 +70,16 @@ namespace Orleans.Transactions
             this.TransactionId = transactionId;
         }
 
+        public OrleansTransactionInDoubtException(string transactionId, Exception exc) : base(string.Format("Transaction {0} is InDoubt", transactionId), exc)
+        {
+            this.TransactionId = transactionId;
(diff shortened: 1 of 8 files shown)


jason-bragg:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test functional

src/Orleans.Core/Transactions/OrleansTransactionException.cs
```diff
@@ -300,4 +331,73 @@ public OrleansTransactionServiceNotAvailableException(SerializationInfo info, St
...
+        }
+
+        public OrleansTransactionLockUpgradeException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
+        {
+        }
+    }
+
+    /// <summary>
+    /// Signifies that the executing transaction has aborted because its execution lock was broken
```
ReubenBond: Summary here needs updating to reflect the exception.

PR #3808, opened 2017-12-20
Move legacy logging methods to legacy package

xiazen:
- remove Logger related method from IStorageProvider and ILogConsistencyProvider
- move legacy Logger and related method on Grain, IGrainRuntime, IProviderRuntime, IClusterClient to legacy package as extension method

note: because the nature of extension method, the call pattern on those GetLogger, Log{get} method changed slightly. 
- For Grain, IGrainRuntime, IProviderRuntime, you need to explicitly use `this`
- For IClusterClient, `this.Logger `-> `this.Logger()`
- For legacy GrainClient,` GrainClient.Logger `-> `GrainClient.Instance.Logger()`


src/Orleans.Logging.Legacy/LoggerExtensionMethods.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <summary>
+        /// Returns a logger object that this grain's code can use for tracing.
+        /// </summary>
+        /// <returns>Name of the logger to use.</returns>
+        public static Logger GetLogger(this Grain grain, string loggerName)
+        {
+            if(grain.Runtime == null)
+                throw new InvalidOperationException("Grain was created outside of the Orleans creation process and no runtime was specified.");
+            var loggerFactory = grain.Runtime.ServiceProvider.GetRequiredService<ILoggerFactory>();
+            return new LoggerWrapper(loggerName, loggerFactory);
```
jason-bragg: I think the cost of doing a service provider lookup + allocation of wrapper is too high grains that do not cache their loggers.

The IGrainActivationContext has a generic container (Items) to store instance specific stuff in.  It is not currently exposed by the grain (which I thought it was) but we can expose it internally and cache a grains loggers in it.

ReubenBond:
looks like a legitimate failure: https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/861/testReport/junit/UnitTests.SchedulerTests/OrleansTaskSchedulerAdvancedTests_Set2/OrleansSched_Test1_Bounce/

src/Orleans.Logging.Legacy/LoggerExtensionMethods.cs
xiazen: ?  I did a cache of Logger in next method. This one ask for a specific loggerName, which can be new or old. For new ones, LoggerFactory will create one, for old ones, it will look it up. The loggerName->logger instance map is already cached in LoggerFacotory. I don't see why we should cache it again.

xiazen:
yeah I send this PR to vso. it is possible that I missed something which would cause test failures

src/Orleans.Logging.Legacy/LoggerExtensionMethods.cs
jason-bragg: This assumes the ILoggerFactory will cache the logger, which is not specified by the contract, hence is an implementation detail that we shouldn't assume.  Even with that being the case, we still incur a lookup of the logger factory in the container, and an allocation of a wrapper object.

Additionally if we cache the loggers in the IGrainActivationContext as suggested, we don't need to have an object GetLogger on the grain for no other reason than to cache the legacy logger, leaving the grain cleaner.  The grain context already has a field for instance specific stuff we can use, we just need to make it available (internally for now) to the grain.

xiazen: I think the goal of the whole moving to asp.net logging is move to a logging infrastructure which we don't need to maintain, and also not worse than our original logging infrastructure. These two goals are both met by asp.net logging infra. I don't see why we should build a thin layer of logger cache here, and I don't see why LoggerFactory would revert their current implementation and move to a version with worse performance. Even they did, I think the work here is to go suggest them not to , and fix it in their logging infra, instead of create a cache here as a patch of theirs. 
For the additional look up on DI container, we can cache LoggerFactory in GrainActivationContext to avoid it , if you are OK with it.

jason-bragg: > I think the goal of the whole moving to asp.net logging is ..

We didn't move to asp.net logging, we moved to their abstraction layer.  By moving to the abstraction we can use asp.net's logging for default logging, but support any solution that conforms to the abstraction.

> I don't see why LoggerFactory would revert their current implementation and move to a version with worse performance.

Agreed, my point was that we're running on top of an abstraction and should avoid making assumptions about the implementation.  The point of running against an abstraction is to allow different implementations.  When running on service fabric, the logger implementation may be one that uses that framework's logging.  When running in an environment that uses APM's, the implementation may wrap the APM apis.  We just don't know and should avoid assumptions.

> For the additional look up on DI container, we can cache LoggerFactory in GrainActivationContext to avoid it , if you are OK with it.

If we're going through the work to cache things in the GrainActivationContext why cache the logger factory instead of the actual logging wrappers we are creating?

xiazen: We didn't just move to their abstraction, we moved to their whole infrastructure, which includes their abstraction, default implementation, and other implementation which conforms to it. Just want to clarify that. The idea there is we move away from logging support. 

If we're going through the work to cache things in the GrainActivationContext why cache the logger factory instead of the actual logging wrappers we are creating?

The call pattern of GetLogger(loggerName) mostly conforms to user keeping a cache of logger on their application code, which is unlike GetLogger(). And even they don't, then we have one extra allocation, is that worth the effort to keep a cache of logger wrappers, which is extra support on logging?

xiazen:
poke. tests passed . pr feedback addressed

src/Orleans.Logging.Legacy/LoggerExtensionMethods.cs
xiazen: discussed offline. Agreed on cache the Grain's logger in IGrainActivationContext

src/Orleans.Logging.Legacy/LoggerExtensionMethods.cs
```diff
@@ -23,17 +23,28 @@ public static Logger GetLogger(this Grain grain, string loggerName)
             var loggerFactory = grain.Runtime.ServiceProvider.GetRequiredService<ILoggerFactory>();
             return new LoggerWrapper(loggerName, loggerFactory);
         }
-
+        private const string GrainLoggerKey = "GrainLogger";
```
jason-bragg: suggest GrainLoggerKey be a static object.

private static object GrainLoggerKey = new object(); 

This ensures uniqueness in bag.  No chance someone else may also try to cache items with the same name because they don't know we've already reserved the name "GrainLogger".

src/Orleans.Logging.Legacy/LoggerExtensionMethods.cs
```diff
@@ -23,17 +23,28 @@ public static Logger GetLogger(this Grain grain, string loggerName)
...
+            if (grainActivatinContext == null)
+            {
+                return grain.GetLogger(grain.GetType().FullName);
+            }
+            else
+            {
+                if (!grainActivatinContext.Items.TryGetValue(GrainLoggerKey, out var grainLogger))
+                {
+                    grainLogger = grain.GetLogger(grain.GetType().FullName);
+                    grainActivatinContext.Items.Add(GrainLoggerKey, grainLogger);
```
jason-bragg: suggest using set

grainActivatinContext.Items[GrainLoggerKey] = grainLogger

While this logic is not intended to be thread safe, should someone call this from multiple threads Add will fail while set will be fine.

src/Orleans.Logging.Legacy/LoggerExtensionMethods.cs
jason-bragg: > We didn't just move to their abstraction, we moved to their whole infrastructure.

Just checked, you're correct, but this was not the plan to my understanding.  We should address this in future PRs

jason-bragg:

Issue #2463, opened 2016-11-30
Security scan of orleans


tsibelman:
I encountered this service build by Coverty to opensource community https://scan.coverity.com
It scans your code for known vulnerabilities and give a detailed report of issue. 

What do you think about starting this process for Orleans ?

veikkoeeva:
I'm greatly in favour of all activity like this. In the current climate, which seem to tighten still, this would signal clearly a mature approach.

sergeybykov:
I don't see right away how they handle false positives.

tsibelman:
@sergeybykov can you elobarate,  what do you mean ?  I understand that they going to have some degree of false positives during analysis, but what are you looking for exactly are you looking for a way to mark such issuese as not real issues ?

veikkoeeva:
I'll provide a bit more details, currently in the project I am in, we take Nuget packages offline and do some checks already. This is not the only project and customers are wondering aloud how much of an impediment would it be to require some sort of an analysis report or assurance security issues have been considered. I'm not exactly sure of the exact reason for this, but at least in Europe [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) is coming and could result in hefty fines (maybe even someone going to jail) if there's a breach due to neglicence.

Maybe this is the reason there has reasonably been interest to create Roslyn analyzers and other tools to provide quick security analysis.

sergeybykov:
@tsibelman When @centur run static analysis with PVS-Studio, there was a certain level of noise in the results - http://shcherbak.me/post/139216516264/checking-microsoft-orleans-with-pvs-studio-for-fun. That's why I wonder 1) how much noise Coverity will produce, and 2) if there are ways to mark false positives for exclusion, so that the report will not scream "56 security issues!" when there are actually zero. For managed code the signal-to-noise ratio is often not great.

I guess we can try and see how useful and flexible the report is.

@veikkoeeva By quickly glancing at GDPR, isn't it about data protection and privacy?

veikkoeeva:
@sergeybykov It is. It mandates breaches should be reported and fines are hefty and someone might be held accountable. Part of the strategy to prevent breaches is to use pathed software, patched libraries. Which results in making sure libraries are appropriately patched and that is checked somehow so it looks good one isn't resting on the laurels. I'm likely stating the obvious here, but I want to be sure and perhaps this is useful for future readers. :)

sergeybykov:
[...]

PR #3818, opened 2017-12-22
Make buffer size bounded when reading connection preamble

lhotari:
Currently it's possible to DoS Orleans by sending bytes that cause the current connection preamble handling to allocate a large buffer. This PR sets an upper bound for the buffer size.

### DIFF
--- a/src/Orleans.Core/Logging/ErrorCodes.cs
+++ b/src/Orleans.Core/Logging/ErrorCodes.cs
@@ -690,6 +690,7 @@ namespace Orleans
         ClientRegistrarTimerFailed              = GatewayBase + 19,
         GatewayAcceptor_WrongClusterId          = GatewayBase + 20,
         GatewayManager_AllGatewaysDead          = GatewayBase + 21,
+        GatewayAcceptor_InvalidSize             = GatewayBase + 22,
 
         TimerBase                               = Runtime + 1400,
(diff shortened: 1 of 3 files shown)


jason-bragg:
Not at all opposed to this change, just adding some context.

Orleans silos are expected to be hosted in a protected environment, i.e. not public facing.
It is expected that the communication between the Orleans client and silo are on a protected network.
A common pattern is to have public facing services (web hosts usually) which use Orleans clients to talk to the cluster over a protected network.

There is value in hardening these communication pipes, as this PR does, but it should be clear that neither our networking layer nor protocol is remotly secure enough to be used over an unprotected network.

benjaminpetit:
I agree with Jason, Orleans expect to be in a "protected" network, but a bug can happen...

@lhotari , your fix is really simple (in a good way!), and add hardens a bit the connection so I will merge it.

Thanks a lot for your contribution!

veikkoeeva:
For the interested about security, there is some discussion at [Expire messages based on arrival time instead of send time ](https://github.com/dotnet/orleans/issues/1524) (i.e. one used attack vector is timer adjustments, see the last messages in the thread), [Hardening Client-to-Silo and Silo-to-Silo connections](https://github.com/dotnet/orleans/issues/3715) and [Security scan of orleans](https://github.com/dotnet/orleans/issues/2463).

PR #3820, opened 2017-12-22
draft of distributed TM implementation

sebastianburckhardt:
First draft of state-based per-grain-TM implementation.
(ported from previous branch where this was developed and tested).

No tests yet; so that will be the next step.


sebastianburckhardt:
Now with tests.

src/Orleans.Transactions/DistributedTM/Implementation/TransactionParticipant.cs
```diff
@@ -0,0 +1,904 @@
...
+                            if (logger.IsEnabled(LogLevel.Trace))
+                                logger.Trace($"{commitQueue.Count} waiting on: RemoteCommitEntry {bottom.Timestamp:o} IsReadOnly={bottom.IsReadOnly} PrepareIsPersisted={bottom.PrepareIsPersisted} LastSent={bottom.LastSent}");
+
+                            break;
+                        }
+
+                    default:
+                        {
+                            logger.Error(777, $"internal error: impossible case");
+                            throw new NotImplementedException();
```
jason-bragg: suggest throwing not supported exception with bottom.Role value used in message body?  I understand this should not happen, but if it does it would be useful to know which case was not handled.  May also be valuable in the log line.

src/Orleans.Transactions/DistributedTM/Implementation/TransactionParticipant.cs
```diff
@@ -0,0 +1,904 @@
...
+                                entry.PromiseForTA.TrySetResult(TransactionalStatus.Ok);
+                            });
+
+                            break;
+                        }
+
+                    default:
+                        {
+                            logger.Error(777, $"internal error: impossible case");
+                            throw new NotImplementedException();
```
jason-bragg: Same as other case statement, i'll stop noting these as comment applies to all case statements.

src/Orleans.Transactions/DistributedTM/Factories/NamedTransactionalStateStorageFactory.cs
```diff
@@ -0,0 +1,38 @@
...
+            where TState : class, new()
+        {
+            // Try to get ITransactionalStateStorage from factory
+            ITransactionalStateStorageFactory factory = string.IsNullOrEmpty(storageName)
+                ? this.context.ActivationServices.GetService<ITransactionalStateStorageFactory>()
+                : this.context.ActivationServices.GetServiceByName<ITransactionalStateStorageFactory>(storageName);
+            if (factory != null) return factory.Create<TState>();
+
+            // Else try to get storage provider and wrap it
+            IStorageProvider storageProvider = string.IsNullOrEmpty(storageName)
```
xiazen: I think this should be updated to use IGrainStorage, `TransactionalStateStorageProviderWrapper` doesn't take IStorageProvider anymore

PR #3829, opened 2018-01-04
Make SiloUnObservedExceptionHandler optional

xiazen:
Make SiloUnObservedExceptionHandler optional, turn it off for silo by default. User can op it in by call `UseSiloUnobservedExceptionsHandler` on `ISiloHostBuilder` 

related issue #3293

### DIFF
--- a/src/Orleans.Runtime/Utilities/SiloUnobservedExceptionsHandler.cs
+++ b/src/Orleans.Runtime/Utilities/SiloUnobservedExceptionsHandler.cs
@@ -3,15 +3,26 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Orleans.Runtime.Scheduler;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Orleans.Hosting;
 
 namespace Orleans.Runtime
 {
-    internal static class SiloUnobservedExceptionsHandlerServiceProviderExtensions
+    public static class SiloUnobservedExceptionsHandlerServiceProviderExtensions
     {
-        public static void InitializeSiloUnobservedExceptionsHandler(this IServiceProvider services)
+        internal static void InitializeSiloUnobservedExceptionsHandler(this IServiceProvider services)
         {
             //resolve handler from DI to initialize it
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Utilities/SiloUnobservedExceptionsHandler.cs
```diff
@@ -3,16 +3,27 @@
...
 namespace Orleans.Runtime
 {
-    internal static class SiloUnobservedExceptionsHandlerServiceProviderExtensions
+    public static class SiloUnobservedExceptionsHandlerServiceProviderExtensions
     {
-        public static void InitializeSiloUnobservedExceptionsHandler(this IServiceProvider services)
+        internal static void InitializeSiloUnobservedExceptionsHandler(this IServiceProvider services)
         {
             //resolve handler from DI to initialize it
             var ignore = services.GetService<SiloUnobservedExceptionsHandler>();
```
benjaminpetit: `InitializeSiloUnobservedExceptionsHandler` is still called in the `Silo` ctor, no? So it should fail if the user didn't registered the `SiloUnobservedExceptionsHandler`, or am I missing something?

xiazen: no it won't fail because it is just try to get UnObservedHandler from DI. if there's not one then it returns null, which is fine

benjaminpetit: Ah yes, you're right, it returns null here

xiazen:

lmagyar:
@xiazen Shouldn't `UseSiloUnobservedExceptionsHandler` return `ISiloHostBuilder`?

xiazen:
why is that?

lmagyar:
All extensions working on `ISiloHostBuilder` returns `siloBuilder` to be able to chain the fluent config methods, this is the only one that differs.

xiazen:
yeah you are right. I misread your question somehow , which is why I have the confusion. Will submit a PR to fix this.

xiazen:
Thanks for reporting this :)

xiazen:
fixed in https://github.com/dotnet/orleans/pull/4132

Issue #198, opened 2015-03-05
OrleansAzureUtils has a static dependnecy on OrleansRuntime


sergeybykov:
This brings OrleansRuntime to any client that uses AzureClient. There seem to be two options.
1. Split OrleansAzureUtils into OrleansClientAzureUtils & OrleansRuntimeAzureUtils.
2. Remove the static dependency and make OrleansAzureUtils dynamically load OrleansRuntime only when AzureSilo is instantiated.

gabikliot:
Seems like given [the new re-structuring of Orleans NuGets](http://dotnet.github.io/orleans/NuGets) we should go with option 1. We already have a separate NuGet for client and server hosting, so if we split the OrleansAzureUtils.dll into OrleansClientAzureUtils.dll & OrleansRuntimeAzureUtils.dll and include the appropriate one in the appropriate NuGet package, we won't introduce any new complications or dependencies or NuGets for users, while avoiding the potential complications of dynamic loading of OrleansRuntime.

gabikliot:
Why did u close that ? I thought it is still an open and valid concern?

sergeybykov:
My mistake. Need more sleep.

sergeybykov:
As we split AzureUtils into separate modules, this will get resolved.

sergeybykov:
Resolved via #3833.

PR #3833, opened 2018-01-05
Remove dependency on Orleans.Runtime from most extension projects

sergeybykov:
This address #198, although a bit indirectly.

With this change most extensions don't need to reference `Orleans.Runtime` anymore. That is particularly important for `.Clustering` and `.Streaming` extensions that are used on the client side. Only .Hosting extensions now reference `Orleans.Runtime`, which makes perfect sense for them to do.

Being a meta-package now, `AzureUtils` will still bring `Orleans.Runtime` as a dependency because `AzureUtils` includes `Orleans.Hosting.AzureCloudServices`. But by referencing instead individual packages, such as `Orleans.Clustering.AzureStorage`, `Orleans.Streaming.AzureStorage`, and `Orleans.Streaming.EventHubs`, one can now avoid bringing `Orleans.Runtime` to the client.

`OrleansAWSUtils` and `OrleansSQLUtils` don't bring `Orleans.Runtime` anymore because they don't have hosting packages. Neither do `Orleans.Clustering.Consul` and `Orleans.Clustering.ZooKeeper`.

`Orleans.Runtime.Abstractions` is probably not the best name for this assembly/package, but I couldn't come up with a better one.


ReubenBond:
Rebased to fix merge conflict

PR #3843, opened 2018-01-09
Fix programatic subscribe bugs

xiazen:
Fix bugs 
- #3843 
- use `StreamSubscriptionManager.RemoveSubscription` with custom `IStreamIdentity` implementation throws casting exception

### DIFF
--- a/src/Orleans.Core/Streams/PubSub/ImplicitStreamPubSub.cs
+++ b/src/Orleans.Core/Streams/PubSub/ImplicitStreamPubSub.cs
@@ -97,6 +97,11 @@ namespace Orleans.Streams
             return implicitTable.IsImplicitSubscriber(GrainExtensions.GetGrainId(addressable), streamId);
         }
 
+        internal bool IsImplicitSubscribeEligibleStreamNameSpace(StreamId streamId)
+        {
+            return !string.IsNullOrEmpty(streamId.Namespace);
+        }
+
         internal bool IsImplicitSubscriber(GuidId subscriptionId, StreamId streamId)
         {
             return SubscriptionMarker.IsImplicitSubscription(subscriptionId.Guid);
--- a/src/Orleans.Core/Streams/PubSub/ImplicitStreamPubSub.cs
+++ b/src/Orleans.Core/Streams/PubSub/ImplicitStreamPubSub.cs
@@ -26,7 +26,7 @@ namespace Orleans.Streams
         public Task<ISet<PubSubSubscriptionState>> RegisterProducer(StreamId streamId, string streamProvider, IStreamProducerExtension streamProducer)
         {
             ISet<PubSubSubscriptionState> result = new HashSet<PubSubSubscriptionState>();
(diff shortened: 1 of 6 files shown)


test/TestGrains/ProgrammaticSubscribe/SubscribeGrain.cs
```diff
@@ -64,10 +64,10 @@ public Task<IEnumerable<StreamSubscription>> GetSubscriptions(FullStreamIdentity
             return subManager.GetSubscriptions(streamIdentity.ProviderName, streamIdentity);
         }
 
-        public async Task RemoveSubscription(StreamSubscription subscription)
+        public async Task RemoveSubscription(FullStreamIdentity streamId, Guid subscriptionId)
```
xiazen: there's tests testing `SubscrptionManager.RemoveSubscription` before, it is just not using custom implemented `IStreamIdentity`, this change would make that test testing this scenario.

xiazen: because `SubscrptionManager.RemoveSubscription` is using `FullStreamIdentity ` which is a custom implementation for `IStreamIdentity`, directly.

src/Orleans.Core/Streams/PubSub/StreamPubSubImpl.cs
```diff
@@ -75,7 +75,14 @@ public Task<int> ConsumerCount(Guid streamId, string streamProvider, string stre
             }
             else
             {
-                var implicitSubs = await implicitPubSub.GetAllSubscriptions(streamId);
+                List<StreamSubscription> implicitSubs = new List<StreamSubscription>();
+                //implicit pubsub requires a non-empty nor null namespace, so if the namespace is null or empty, there won't be any 
+                // subscription for that stream in implicitPubSub. Skip it to avoid unnecessary exception thrown
+                if (implicitPubSub.IsImplicitSubscribeEligibleStreamNameSpace(streamId))
```
jason-bragg: Implicit subscriptions are the domain of the implicitPubSub. it seems odd for the caller of GetAllSubscriptions to check this rather than for the GetAllSubscriptions to simply return no subscriptions if a namespace is invalid, similar to how RegisterProducer does.

src/Orleans.Core/Streams/PubSub/ImplicitStreamPubSub.cs
```diff
@@ -97,6 +97,11 @@ internal bool IsImplicitSubscriber(IAddressable addressable, StreamId streamId)
             return implicitTable.IsImplicitSubscriber(GrainExtensions.GetGrainId(addressable), streamId);
         }
 
+        internal bool IsImplicitSubscribeEligibleStreamNameSpace(StreamId streamId)
```
jason-bragg: I'm concerned this constraint is duplicated in multiple places.  Here, ImplicitStreamPubSub.RegisterProducer, and ImplicitStreamSubscriberTable.GetImplicitSubscribers.  I think we need a single authoritative place which defines this constraint (probably the ImplicitStreamSubscriberTable) and all other places reference to that logic.

Issue #3844, opened 2018-01-09
Resolve code omitted by ABSTRACTIONS_TODO


ReubenBond:
https://github.com/dotnet/orleans/blob/a132a5acaf74ceb233fdf08ba497f384ce1c95ff/src/Orleans.Core.Abstractions/IDs/GrainId.cs#L219

https://github.com/dotnet/orleans/blob/a132a5acaf74ceb233fdf08ba497f384ce1c95ff/src/Orleans.Core.Abstractions/IDs/GrainId.cs#L253

SystemTargets are showing up in logs with with `???` in place of their type, like `Orleans.Streams.???/ea97e894.Router]	Created PersistentStreamPullingAgent ???/459873b2bfab8e5b6bbeab2bea97e89401000000000000ff-0x6AE30F88`

PR #3849, opened 2018-01-10
Display GrainId/type of SystemGrain and SystemTarget in GrainId.ToStringImpl

benjaminpetit:
Related to #3844

### DIFF
--- a/src/Orleans.Core.Abstractions/IDs/GrainId.cs
+++ b/src/Orleans.Core.Abstractions/IDs/GrainId.cs
@@ -244,10 +244,10 @@ namespace Orleans.Runtime
                     fullString = $"*gcl/{Key.KeyExt}/{idString}";
                     break;
                 case UniqueKey.Category.SystemGrain:
-                    fullString = $"*/sgn/{Key.PrimaryKeyToGuid()}/{idString}";
+                    fullString = $"*sgn/{Key.PrimaryKeyToGuid()}/{idString}";
                     break;
                 case UniqueKey.Category.SystemTarget:
-                    fullString = $"*/stg/{Key.N1}/{idString}";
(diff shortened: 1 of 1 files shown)


src/Orleans.Core.Abstractions/IDs/GrainId.cs
```diff
@@ -241,25 +235,20 @@ private string ToStringImpl(bool detailed)
...
                 case UniqueKey.Category.Client:
-                    fullString = "*cli/" + idString;
+                    fullString = $"*cli/{idString}";
                     break;
                 case UniqueKey.Category.GeoClient:
-                    fullString = string.Format("*gcl/{0}/{1}", Key.KeyExt, idString);
+                    fullString = $"*gcl/{Key.KeyExt}/{idString}";
+                    break;
+                case UniqueKey.Category.SystemGrain:
+                    fullString = $"*/sgn/{Key.PrimaryKeyToGuid()}/{idString}";
```
ReubenBond: why do these and the have a `/` at the start?

benjaminpetit: I probably need new eyes

ReubenBond: I mean, `*/blah` instead of just `*blah` like the others

benjaminpetit: > I mean, */blah instead of just *blah like the others

Yes I saw it, my response was for the "why": I was certain that all cases had a `/` after the `*` :p

ReubenBond: got it ;)

PR #3851, opened 2018-01-10
Add StreamSubscriptionHandleFactory to subscribe onbehalf feature

xiazen:
subscribe on behalf feature has a bug in it which makes it unusable. The cause of the bug is, in streaming infra, stream type is not parsed correctly to `IStreamSubscriptionObserver.OnSubscribe`. This PR introduces `IStreamSubscriptionHandleFactory` which requires user to give that type information, as a way to fix that bug. 

There were other solutions discussed for this bug , but they all require more engineer hours. Based on the team's current capacity, this seems to be the best way ATM to fix it.


src/Orleans.Core.Abstractions/Streams/Core/IStreamSubscriptionHandleFactory.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Orleans.Runtime;
+
+namespace Orleans.Streams.Core
+{
+    public interface IStreamSubscriptionHandleFactory
+    {
+        StreamSubscriptionHandle<T> Create<T>(GuidId subscriptionId, IStreamIdentity streamId, string streamProviderName);
```
jason-bragg: Only thing we need from the application is the type (T).
And the application needs know only the stream identity and provider to make the decision if it should subscribe or not.

So Is suggest:

    public interface IStreamSubscriptionHandleFactory 
    {
        IStreamIdentity StreamId { get; }
        string ProviderName { get; }
        StreamSubscriptionHandle<T> Create<T>();
    }

This would make
 
    public interface IStreamSubscriptionObserver
    { 
        Task OnSubscribed(IStreamSubscriptionHandleFactory handleFactory);
    }

The advantage of this is that it provides less room for error.  In the previous pattern there is nothing prohibiting the user from passing in incorrect streamed, subscriptionid or provider name.

test/TesterAzureUtils/Streaming/ProgrammaticSubscribeTestAQStreamProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+using Tester.StreamingTests;
+using TestExtensions;
+using UnitTests.StreamingTests;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Tester.AzureUtils.Streaming
+{
+    [TestCategory("BVT"), TestCategory("Streaming"), TestCategory("Functional"), TestCategory("AQStreaming")]
+    public class ProgrammaticSubscribeTestAQStreamProvider : ProgrammaticSubcribeTestsRunner, IClassFixture<ProgrammaticSubscribeTestAQStreamProvider.Fixture>
```
jason-bragg: not terribly important, but we tend to prepend the specialization to the runner name.  For instance EHClientStreamTests, AQClientStreamTests, SMSClientStreamTests, and so on.

xiazen:
@dotnet-bot test bvt please

xiazen:
poke poke. ready to ship ~

PR #3856, opened 2018-01-11
Make FileLogger, LegacyFileLogConsumer and FileTelemetryConusmer all use UTF8 encoding.

xiazen:
- Make FileLogger, FileLogConsumer and FileTelemetryConusmer all use UTF8 encoding
- Remove read lock on log file so that file can be read when messages is writing to the file.

### DIFF
--- a/src/Orleans.Core/Logging/FileLogger.cs
+++ b/src/Orleans.Core/Logging/FileLogger.cs
@@ -27,7 +27,7 @@ namespace Orleans.Logging
         /// <param name="fileName"></param>
         public FileLoggingOutput(string fileName)
         {
-            logOutput = new StreamWriter(File.Open(fileName, FileMode.Append, FileAccess.Write, FileShare.Write));
+            logOutput = new StreamWriter(File.Open(fileName, FileMode.Append, FileAccess.Write, FileShare.ReadWrite), Encoding.UTF8);
             this.logFileName = fileName;
         }
 
--- a/src/Orleans.Core/Telemetry/Consumers/FileTelemetryConsumer.cs
+++ b/src/Orleans.Core/Telemetry/Consumers/FileTelemetryConsumer.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Text;
 
 namespace Orleans.Runtime
 {
@@ -17,9 +18,7 @@ namespace Orleans.Runtime
         public FileTelemetryConsumer(FileInfo file)
         {
             _logFileName = file.FullName;
-            var fileExists = File.Exists(_logFileName);
-            _logOutput = fileExists ? file.AppendText() : file.CreateText();
-            file.Refresh();
+            _logOutput = new StreamWriter(File.Open(_logFileName, FileMode.Append, FileAccess.Write, FileShare.ReadWrite), Encoding.UTF8);
         }
 
         public void TrackTrace(string message)
--- a/src/Orleans.Logging.Legacy/LegacyFileLogConsumer.cs
+++ b/src/Orleans.Logging.Legacy/LegacyFileLogConsumer.cs
@@ -2,6 +2,7 @@
 using System;
 using System.IO;
 using System.Net;
+using System.Text;
 
 namespace Orleans.Logging.Legacy
 {
@@ -17,7 +18,8 @@ namespace Orleans.Logging.Legacy
 
         public LegacyFileLogConsumer(string fileName)
         {
-            logOutput = new StreamWriter(File.Open(fileName, FileMode.Append, FileAccess.Write, FileShare.Write));
+            this.logFileName = fileName;
+            logOutput = new StreamWriter(File.Open(fileName, FileMode.Append, FileAccess.Write, FileShare.ReadWrite), Encoding.UTF8);
         }
 
         public void Log(


src/Orleans.Core/Logging/FileLogger.cs
```diff
@@ -27,8 +27,11 @@ public class FileLoggingOutput
         /// <param name="fileName"></param>
         public FileLoggingOutput(string fileName)
         {
-            logOutput = new StreamWriter(File.Open(fileName, FileMode.Append, FileAccess.Write, FileShare.Write));
-            this.logFileName = fileName;
+            var file = new FileInfo(fileName);
+            logFileName = file.FullName;
+            var fileExists = File.Exists(logFileName);
+            logOutput = fileExists ? file.AppendText() : file.CreateText();
+            file.Refresh();
```
ReubenBond: Is there any point refreshing the object just before it falls out of scope?

src/Orleans.Core/Logging/FileLogger.cs
```diff
@@ -27,8 +27,11 @@ public class FileLoggingOutput
         /// <param name="fileName"></param>
         public FileLoggingOutput(string fileName)
         {
-            logOutput = new StreamWriter(File.Open(fileName, FileMode.Append, FileAccess.Write, FileShare.Write));
-            this.logFileName = fileName;
+            var file = new FileInfo(fileName);
+            logFileName = file.FullName;
+            var fileExists = File.Exists(logFileName);
+            logOutput = fileExists ? file.AppendText() : file.CreateText();
```
ReubenBond: I would have thought we want to explicitly use UTF8 encoding,
It seems like we should use the `StreamWriter(Stream, Encoding)` ctor where the stream is `File.Open(fileName, FileMode.Append, FileAccess.Write, FileShare.ReadWrite)`. Possibly `FileShare.Read` is sufficient, but probably some apps open the file for read + write (eg, maybe if the user wants to open the file in a text editor - not sure).

src/Orleans.TestingHost/Utils/TestingUtils.cs
```diff
@@ -40,8 +41,9 @@ public static string CreateTraceFileName(string nodeName, string clusterId)
             {
                 Directory.CreateDirectory(traceFileFolder);
             }
-
-            var traceFileName = $"{traceFileFolder}\\{clusterId}_{nodeName}.log";
+            DateTime now = DateTime.UtcNow;
+            string DateTimeFormat = @"yyyy-MM-dd\tHH-mm-ss-fff";
+            var traceFileName = $"{traceFileFolder}\\{clusterId}_{nodeName}_{now.ToString(DateTimeFormat, CultureInfo.InvariantCulture)}.log";
```
ReubenBond: Should use Path.Combine instead of explicit `\`.

Maybe add `Z` to the end to signify that it's a UTC time

galvesribeiro: `DateTime.UtcNow.ToString("o")` should be better...

galvesribeiro:
Having a second thought after one comment @ReubenBond made in Gitter and thinking about the direction Orleans took on use MEL abstractions and have no logging infrastructure, I think you should just drop that `AddFile()` method. For backcompat sake, you could add it to the Legacy package, but I'm not sure we should have it anymore. 

When I found and reported that issue, I was in a hurry to get file logging, but thinking straight, it is not supposed to be on Orleans anyway. I got it working after I added Serilog, which is OK.

We should only add the `ConsoleLog` provider to the `ILoggerFactory` by default and only if the developer doesn't do that, because they may want their own console implementation (i.e. colored console from Serilog or nLog or whatever) instead of use the default one.

So if it were on me, I would just drop it but thanks for fixing it anyway. :)

xiazen:
I think I should explain more on the intention behind `FileLogger`, `FileLoggerProvider` and `AddFile`. I realize that part of information isn't clear to the public users. 

- we support those two NOT out of legacy or backcompact reason, we support them because ASP.NET doesn't supports a `ILogger` implementation which logs to a file, but the ability to logging to a file is important to orleans and orleans users. For example, orleans use it heavily in our unit test and load test logging and I think this is also a solid case for orleans users.  That's why we support this `FileLogger` and it is not in the legacy package. Because this is something we are willing to support going forward

- for backcompact reason, we have a `LegacyFileLogConsumer` in Legacy logging package, which conforms to `ILogConsumer` interface. This is the one exists only for back compact reason. Avoid to confuse it with `FileLogger`

- silo is configured with 0 `ILoggerProvider` by default. We don't want to put any assumption on user's application. It's their responsibility to configure any provider they want. 

- `TestCluster` is configured with a `FileLogger` by default, since file loggers are the most common logger user want to configure with their Test cluster. But user can totally configure it other ways. And also `TestCluster` is under revamping by @ReubenBond and @jdom . So I'm not sure if this is still true. We will see after their PR I guess. 

Hope this make sense. :)

galvesribeiro:
I got it.

> we support those two NOT out of legacy or backcompact reason, we support them because ASP.NET doesn't supports a ILogger implementation which logs to a file, but the ability to logging to a file is important to orleans and orleans users. For example, orleans use it heavily in our unit test and load test logging and I think this is also a solid case for orleans users. That's why we support this FileLogger and it is not in the legacy package. Because this is something we are willing to support going forward

I still think that if people want to use the `ILogger`, it should be from a proper `ILoggerProvider` (like I did by adding Serilog to write to the files). But anyway, that was just what I _would_ do. Its ok if Orleans still have that.

Again, thanks for the fix.

src/Orleans.TestingHost/Utils/TestingUtils.cs
jdom: This is a utility for the TestCluster (which BTW we are updating in the TestCluster revamp branch if I'm not mistaken), and for these, clusterId already includes the timestamp in the name... This would have the timestamp 2 times in our tests, correct?

xiazen:
I reverted the approach I was taking before and took @ReubenBond 's advice to create `StreamWriter` with UTF8 encoding, and apply this rule to `FileLogger`, `LegacyFileLogConsumer` and `FileTelemetryConsumer`.

xiazen:
poke?

xiazen:
any feedback?

ReubenBond:
apologies for the delay, @xiazen :(

xiazen:
np. don't worry about it :)

PR #3858, opened 2018-01-12
Move some streaming related helpers in abstractions

benjaminpetit:
While they are not really abstractions, they are used by Grain project, and it would be a pity to add a dependency on Core just for this helper functions...


benjaminpetit:
@dotnet-bot test bvt please

PR #3867, opened 2018-01-16
Fixed Dead Link to XSD Source

normanhh3:
Found the XSD source was dead linking to a 404 page.  Found the correct link and corrected it.

Any chance there is a [linting tool for docs](https://github.com/markdownlint/markdownlint) that could be run as part of the build process to ensure links still work?  (The linked tool does some other cool stuff with Markdown verification, but does not appear to check links for validity!)

Anyway, this link is fixed for now. ;-)


ReubenBond:
Thanks! It seems that DocFx could take care of this kind of thing - I wonder if they've considered it.

Issue #3865, opened 2018-01-15
Improve and fix ADO.NET configuration documentation


veikkoeeva:
The ADO.NET documentation should be both fixed and improved.

- There is a problem described at https://github.com/dotnet/orleans/issues/3862.
- The documentation should also be clear that one ought to consider schemas and filegroups in a production application.

Cross-referenging https://github.com/dotnet/orleans/issues/3864.

Issue #3864, opened 2018-01-15
Improve ADO.NET script finding


veikkoeeva:
It looks like the ADO.NET scripts are quite difficult to find and apply since they're split and located in Nuget cache directory that in the newer versions is located "far away" from the project. Using `nuget locals all -list`, reveals the location is something such as `C:\Users\<username>\.nuget\packages\microsoft.orleans.*.adonet\2.0.0-beta3\lib\netstandard2.0`. An improvement would be to use [Nuget ContentFile](https://blog.nuget.org/20160126/nuget-contentFiles-demystified.html) definitions and move the scripts either to the project referencing them or to solution root.

Further, the documentation -- and exception messages, should be clear that the associated ADO.NET libraries should be loaded too. This is traditionally a question people have been asking when they've used something else than SQL Server, a connector library is needed to load and configured separately.

Cross-referencing https://github.com/dotnet/orleans/issues/3862.

Issue #3862, opened 2018-01-13
CreateOrleansTables sql script?


dazinator:
I am attempting to use sql server for storage.

The link to the sql script to create orleans tabes on this page 404's: http://dotnet.github.io/orleans/Documentation/Advanced-Concepts/Configuring-SQL-Tables 

Any ideas where I can find it?

larsfjerm:
Here is what you're looking for - run script to your DB:
https://pastebin.com/XK5v063w

veikkoeeva:
@dazinator 

1. It appears the link in the documentation has gone stale (should be fixed, naturally).
2. The scripts are co-located with the Nuget binaries, seeing you're in .NET Standard and if you run `nuget locals all -list`, I'd imagine it's something like `C:\Users\<username>\.nuget\packages\microsoft.orleans.*.adonet\2.0.0-beta3\lib\netstandard2.0`. There was recently a refactoring to split the script into multiple parts. You need to run separately the ones you need (e.g. in your case don't forget `C:\Users\Veikko\.nuget\packages\microsoft.orleans.clustering.adonet\2.0.0-beta3\lib\netstandard2.0\SQLServer`). The problem is further exarcarbarated in that the Nuget packages are places "far away from the project", i.e. in the Nuget cache. This should be remedied somehow, like bringing the scripts to project directory or something.
3. It should be explicitly mentioned in the documentation (and in the exception message) that one needs to reference also an appropriate ADO.NET library. The reason for this is that there are tens of providers for ADO.NET and currently at least there isn't a separate Nuget package that would add a helper function (e.g. `.ConfigureSqlServer`) which would then also bring in SQL Server packages.

A further note, if you plan to run database in production, do note you need to think a bit of adding a schema and appropriate filegroups. I.e. modify the script and queries a bit. I'm working [on a sample](https://github.com/veikkoeeva/orleans/tree/adonetstample/Samples/AdoNetSample/src/AdotNetSample.Database) which would use [Dacpac](https://docs.microsoft.com/en-us/sql/relational-databases/data-tier-applications/data-tier-applications).

Does this help you?

dazinator:
Thank you, I will take a look in the packages - i thought I had done that already - perhaps missed them.

I understand that there are many ado.net providers, but the dialect of sql used by a provider is usually specific for that provider. For example a postgress sql script is obviously different from an ms sql script. Due to that, I think I was expecting for the scripts to live with a concrete provider package. I was a bit off in my assumption :-)

[...]

PR #3869, opened 2018-01-16
Fixed links to the ADO.NET Scripts

rvlieshout:


veikkoeeva:
Cross-referencing https://github.com/dotnet/orleans/issues/3862 (this issue is noted there). Also relevant on arranging so that the document and practice support each other: https://github.com/dotnet/orleans/issues/3864 and https://github.com/dotnet/orleans/issues/3865.

PR #3870, opened 2018-01-16
Allow for null access/secret for EC2 provisioned credentials.

berdon:
These need to be nullable so the parse routine can properly setup the AmazonDynamoDBClient with implicit auth.


sergeybykov:
Thank you, @berdon!

PR #3871, opened 2018-01-16
Move dispose method inside fixture class

xiazen:
`Dispose` method was in the wrong place. 
It should be called after all the tests finished, so it should be in its `Fixture`'s `Dispose()`, instead of in the testclass level Dispose(), which is called whenever per test finished.


xiazen:
UnitTests.MembershipTests.SqlServerMembershipTableTests.MembershipTable_SqlServer_UpdateRowInParallel 
test failure are not related

xiazen:

PR #3872, opened 2018-01-16
Add .editorconfig

### LINKED ISSUE #2332: Adopt some codified code style standards
Standards for code style are useful for reducing cognitive load when reading code. Currently we don't have a consistent standard

I'm not particularly fussed about _what_ the standards are, but my preference is for:
- Prefix `this.` on instance members
- No leading underscore for instance variables
- Braces around statement blocks (after `if`, `else`, etc)

Pretty much the default, standard Microsoft [StyleCop](https://github.com/StyleCop/StyleCop) ruleset.
Happy to invert the top two as long as we apply them consistently.

It would be nice if we could check in a StyleCop/ReSharper config file. Is there another suitable formatting tool?

ReubenBond:
Fixes #2332

We can change the style as we see fit, but at least this is a start and I believe it's consistent with our current style.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,200 @@
+# http://EditorConfig.org
+
+# This file is the top-most EditorConfig file
+root = true
+
+# All Files
+[*]
+charset = utf-8
+end_of_line = crlf
+indent_style = space
+indent_size = 4
+insert_final_newline = false
+trim_trailing_whitespace = true
+
+# Solution Files
+[*.sln]
+indent_style = tab
+
+# XML Project Files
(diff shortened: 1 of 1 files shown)


galvesribeiro:
Nice! Finally! 

Btw, it worth change the README.md and the docs whenever it mention contributions so welcomers understand it. 

Also, there is an extension for VS that helps a lot when you have the `.editorconfig` file in place [here](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.EditorConfig).

galvesribeiro:
Btw, for whatever reason the some of the globbing patterns are not working as expected:

image

PR #3878, opened 2018-01-17
Orleans.TestingHost support test clusters without legacy configuration

ReubenBond:
This should allow us to more easily migrate configuration over to non-legacy classes bit-by-bit as well as hopefully allow us to run our tests on .NET Core (i.e, by removing the reliance on creating AppDomains)

cc @jdom, who performed the initial work


test/Tester/GrainActivatorTests.cs
```diff
@@ -25,27 +22,19 @@ public class Fixture : BaseTestClusterFixture
...
-                var options = new TestClusterOptions(1);
-                options.UseSiloBuilderFactory<TestSiloBuilderFactory>();
-                return new TestCluster(options);
+                var builder = new TestClusterBuilder(1);
+                builder.AddSiloBuilderConfigurator<TestSiloBuilderFactory>();
+                return builder.Build();
             }
 
-            private class TestSiloBuilderFactory : ISiloBuilderFactory
+            private class TestSiloBuilderFactory : ISiloBuilderConfigurator
```
jdom: nit: you could rename to `TestSiloBuilderConfigurator`

ReubenBond: ah, yeah, I renamed some, but I'm sure I missed many

jdom:
>         }

Is this fine or should it use `new TestClusterBuilder().Build()` instead? Just asking, because I don't remember if the defaults are useable

---
Refers to: test/TestExtensions/HostedTestClusterBase.cs:62 in 2db2741. [](commit_id = 2db27411cdf8c8770e2b93c389598b3386a1b7a8, deletion_comment = False)

ReubenBond:
The latter, will fix. BTW, your comment isn't anchored to the code.

src/Orleans.TestingHost/Grains/ITestClusterManagementGrain.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.TestingHost.Grains
+{
+    public interface ITestClusterManagementGrain : IGrainWithGuidKey
+    {
+        Task<T> GetService<T>();
```
jdom: I wouldn't add this grain here (in the package that we ship) since almost no service would be serializable. The only one I can think of is the legacy config objects, so wouldn't pollute the surface with this.

ReubenBond: Alright, should I make a more specific legacy-themed grain to get config objects or try to retrieve them some other way?

test/Tester/ClientConnectionTests/InvalidPreambleConnectionTests.cs
```diff
@@ -8,9 +12,10 @@ namespace Tester.ClientConnectionTests
     public class InvalidPreambleConnectionTests : TestClusterPerTest
     {
         [Fact, TestCategory("Functional")]
-        public void ShouldCloseConnectionWhenClientSendsInvalidPreambleSize()
+        public async Task ShouldCloseConnectionWhenClientSendsInvalidPreambleSize()
         {
-            var gwEndpoint = this.HostedCluster.Primary.NodeConfiguration.ProxyGatewayEndpoint;
+            var clusterConfig = await this.Client.GetGrain<ITestClusterManagementGrain>(Guid.Empty).GetService<ClusterConfiguration>();
+            var gwEndpoint = clusterConfig.Overrides[Silo.PrimarySiloName].ProxyGatewayEndpoint;
```
jdom: There are other ways to get the proxy endpoint without the need to resort to serializing the config back from the silo (and in fact that would break if you stop using the legacy config).
If you hold on to the TestClusterOptions, you can peek in there. Otherwise there's also the opportunity to get the gateway list provider options from the client itself without asking the silo for the config

src/Orleans.TestingHost/SiloHandle.cs
```diff
@@ -10,24 +8,24 @@ namespace Orleans.TestingHost
...
         /// <summary> Get or set the name of the silo </summary>
         public string Name { get; set; }
 
         /// <summary>Get or set the address of the silo</summary>
         public SiloAddress SiloAddress { get; set; }
 
-        /// <summary>Get the proxy address of the silo</summary>
-        public SiloAddress ProxyAddress => SiloAddress.New(this.NodeConfiguration.ProxyGatewayEndpoint, 0);
+        ///// <summary>Get the proxy address of the silo</summary>
+        public SiloAddress GatewayAddress { get; set; }
```
jdom: the default implementation *I think* could just look for this information in the TestClusterOptions if it uses test cluster membership.

This way other subclasses do not have to rely on extended RPC for it to work (such as MarshalByRefObject as we have in the AppDomainSiloHandle implementation)

src/Orleans.TestingHost/TestClusterHostFactory.cs
```diff
@@ -0,0 +1,261 @@
...
+                TryConfigureTestClusterMembership(configuration, services);
+                TryConfigureFileLogging(configuration, services, hostName);
+
+                if (Debugger.IsAttached)
+                {
+                    // TODO: configure this without requiring the legacy configuration, when that's available
+                    var clientConfiguration = GetOrCreateClientConfiguration(services, configuration);
+
+                    // Test is running inside debugger - Make timeout ~= infinite
+                    clientConfiguration.ResponseTimeout = TimeSpan.FromMilliseconds(1000000);
```
jdom: Note: Once we move this to strongly typed options, we can get rid of the dependency to the legacy config objects in the TestClusterHostFactory

src/Orleans.TestingHost/Grains/ITestClusterManagementGrain.cs
jdom: If it was just to get the gateway endpoint, you probably don't even need that, you can derive it from the options, or even ask the client

ReubenBond:
I'm thinking of modifying `TestClusterPerTest.CreateTestCluster` (and similar base class methods) so that instead of returning `TestCluster`, it returns `TestClusterBuilder`. That way, the base class has a chance to modify the configuration and, eg, insert the defaults loaded from `OrleansTestSecrets.json` / env vars - the ones loaded in `TestDefaultConfiguration`.

Another alternative is to change the method to something like ~`void ConfigureTestCluster(TestClusterBuilder builder)` and to modify `LegacyTestClusterBuilder` so that it's no longer a subclass of `TestClusterBuilder`, but instead offers an extension method to let the consumer configure the legacy configuration, eg:

```C#
void ConfigureTestCluster(TestClusterBuilder builder)
{
  // probably this would be done in the base class.
  builder.AddTestEnvironmentConfiguration();

  // Set some legacy configuration options.
  builder.ConfigureLegacyConfiguration(cfg =>
  {
    cfg.ClusterConfiguration.SomeProperty = 3;

    cfg.ClientConfiguration.SomeProperty = 3
  });
}
```

What do you think?

jdom:
I do like the change to use extension methods on the 1 and and only test cluster builder instead of inheritance :+1: :+1: in fact now that I remember, I thought of avoiding inheritance too, but never got to it.

src/Orleans.TestingHost/AppDomainSiloHost.cs
```diff
@@ -56,13 +50,6 @@ public void Shutdown()
         {
             this.host.StopAsync().GetAwaiter().GetResult();
         }
-
-        private void InitializeTestHooksSystemTarget()
```
xiazen: where is this part of logic moved to ?

ReubenBond: Moved to [TestClusterHostFactory](https://github.com/dotnet/orleans/pull/3878/files#diff-238a7057e8b8da8eeefdc924dd6abc7cR254)

src/Orleans.TestingHost/LegacyTestClusterConfiguration.cs
```diff
@@ -82,20 +82,28 @@ public static ClientConfiguration BuildClientConfiguration(ClusterConfiguration
...
+                configuration.Bind(options);
 
-            if (this.builder.Properties.TryGetValue(ClientConfigurationKey, out object clientConfig))
-            {
-                configSource[ClientConfigurationKey] = Serialize(serializationManager, clientConfig);
+                for (var i = 0; i < this.builder.InitialSilosCount; i++)
+                {
+                    var siloName = options.GetSiloSpecificOptions(i).SiloName;
+
+                    this.ClusterConfiguration.GetOrCreateNodeConfigurationForSilo(siloName);
```
jdom: Why go back to defining these? can't you avoid it and only do so specifically if a particular test was incorrectly relying on it?
I'd like to prevent re-adding this at the infrastructure level if it was just for an edge case test

ReubenBond: I was hoping to avoid it, too, but it's used in non-test code in a non-trivial way which was making tests fail. I had a fix for the one test usage I sent to you, but I reverted it and went with this approach.

https://github.com/dotnet/orleans/blob/9bd486bb7368a83a37b7565e2d267aed959e4df7/src/Orleans.Runtime/Streams/QueueBalancer/StaticClusterDeploymentConfiguration.cs#L26

It will [have to] be addressed when we move away from ClusterConfiguration internally

test/TestInternalGrains/PersistenceTestGrains.cs
```diff
@@ -290,6 +291,14 @@ public Task DoSomething()
         }
     }
 
+    public class ServiceIdGrain : Grain, IServiceIdGrain
+    {
+        public Task<Guid> GetServiceId()
+        {
+            return Task.FromResult(this.ServiceProvider.GetRequiredService<GlobalConfiguration>().ServiceId);
```
jdom: Why not get this from the strongly typed option (instead of the legacy config) which is the authoritative source of truth?

ReubenBond: Will fix

ReubenBond:
@dotnet-bot test bvt

src/Orleans.TestingHost/SiloHandle.cs
ReubenBond: I wanted this so that I could query it for some tests. It seems right that if SiloAddress is there, then GatewayAddress should also be there.

On the other hand, I'm not confident about the overall class structure for this change (i.e, could things be separated more cleanly?)

src/Orleans.TestingHost/TestClusterHostFactory.cs
ReubenBond: Moved this to LegacyTestClusterConfiguration

xiazen:
can we have some write up about new interfaces and major changes, so it is easier to review?

src/Orleans.TestingHost/TestClusterHostFactory.cs
jdom: The reality is that this is useful in the non-legacy case too going forward. I would keep this by default, it's just that when you eventually move this setting out to a strongly typed option, then you can stop relying on the legacy configuration entirely

ReubenBond:
@xiazen the gist of it is this:
* `TestCluster` creation now has a builder pattern via `TestClusterBuilder`
* Legacy configuration was removed from `TestCluster` and into separate classes which can be added to the `TestClusterBuilder` by using the `ConfigureLegacyConfiguration` extension method
* Instead of serializing configuration objects over an AppDomain boundary, everything is configured via the `IConfiguration` system
* The hope is to be able to swap the AppDomainSiloHandle implementation for different implementations (eg, in-proc, new proc, container, whatever) later, since it's easier to serialize the `IConfigurationSource` objects which are used to create configuration.

src/Orleans.TestingHost/TestClusterHostFactory.cs
```diff
@@ -205,7 +196,9 @@ private static void ConfigureAppServices(IConfiguration configuration, IClientBu
         private static void TryConfigureTestClusterMembership(HostBuilderContext context, IServiceCollection services)
         {
             bool.TryParse(context.Configuration["UseTestClusterMemebership"], out bool useTestClusterMemebership);
-            if (useTestClusterMemebership)
+            if (useTestClusterMemebership
+                && services.All(svc => svc.ServiceType != typeof(IMembershipTable))
+                && services.All(svc => svc.ServiceType != typeof(IMembershipOracle)))
```
jdom: nit: could avoid double enumeration

ReubenBond: I'll remove this and make sure `UseTestClusterMembership` is correctly set instead

jdom:
@dotnet-bot test this please

ReubenBond:
https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/1013/
```
14:12:09 No executable found matching command "dotnet-xunit"
14:12:09     + CategoryInfo          : NotSpecified: (No executable f... "dotnet-xunit" 
14:12:09    :String) [], RemoteException
14:12:09     + FullyQualifiedErrorId : NativeCommandError
14:12:09     + PSComputerName        : localhost
```

jdom:
@dotnet-bot test this please

ReubenBond:
Please review but don't merge (I'm sure there will be changes requested anyway) - I need to ensure we have a green run in VSO first.

jdom:
yup, sure, pay special attention to Azure tests, since they don't run here, as using different membership providers are directly impacted with how this new test infrastructure works. Should be trivial to resolve if there are issues though

src/Orleans.TestingHost/SiloHandle.cs
jdom: Sure, it's fine, I don't really mind if we keep this way for now. The reality is that I initially was retrieving it from the container in that way before I did all the work to move to strongly typed config. Today you can just as well get it directly from the configuration.
But as I said, it's fine... Both AppDomain and in-domain strategies can RPC from the TestCluster... If we have other approaches (like out of process or containers) we can refactor it so that the silo handle infers it from the config instead of asking the remote host.

src/Orleans.TestingHost/TestCluster.cs
```diff
@@ -284,42 +237,39 @@ public static TimeSpan GetLivenessStabilizationTime(GlobalConfiguration global,
         /// <summary>
         /// Start an additional silo, so that it joins the existing cluster.
         /// </summary>
-        /// <returns>SiloHandle for the newly started silo.</returns>
+        /// <returns>SiloHandle2 for the newly started silo.</returns>
```
jdom: Search for old usages of SiloHandle2 and replace with SiloHandle

src/Orleans.TestingHost/TestCluster.cs
```diff
@@ -436,14 +395,15 @@ public SiloHandle RestartSilo(SiloHandle instance)
...
         }
 
-        /// <summary>
-        /// Restart a previously stopped.
-        /// </summary>
-        /// <param name="siloName">Silo to be restarted.</param>
+        ///// <summary>
+        ///// Restart a previously stopped.
+        ///// </summary>
+        ///// <param name="siloName">Silo to be restarted.</param>
```
jdom: uncomment XML comments

jdom:
LGTM. Ship it! :)

src/Orleans.Core/Configuration/Legacy/LegacyConfigurationExtensions.cs
```diff
@@ -15,13 +16,25 @@ internal static class LegacyConfigurationExtensions
...
 
+            if (configuration == null)
+            {
+                try
+                {
+                    configuration = ClientConfiguration.StandardLoad();
+                }
+                catch (FileNotFoundException)
+                {
+                    configuration = new ClientConfiguration();
```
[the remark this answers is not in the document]
jdom: Not really, this is for the path we're users didn't explicitly pass a config object. If they do want, then they'll get the exception in their code when they try to load the config.
Btw, this code path will be removed as soon as the runtime itself stops requiring the legacy config in the first place

src/Orleans.TestingHost/ClientBuilderConfiguratorBase.cs
```diff
@@ -0,0 +1,27 @@
...
+        void IClientBuilderConfigurator.Configure(IConfiguration configuration, IClientBuilder clientBuilder)
+        {
+            this.TestClusterOptions = configuration.GetTestClusterOptions();
+            this.Configure(configuration, clientBuilder);
+        }
+
+        /// <summary>
+        /// Configures the client builder.
+        /// </summary>
+        public abstract void Configure(IConfiguration configuration, IClientBuilder clientBuilder);
```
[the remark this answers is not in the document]
ReubenBond: The explicit `IClientBuilderConfigurator` implementation does some work before calling the abstract method. I could collapse the class hierarchy instead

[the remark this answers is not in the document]
xiazen: I might be dumb here, but why cannot we merge two Configure method?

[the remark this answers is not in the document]
ReubenBond: we can, it just means moving this code into any consumer. The idea was that this class would be reusable, but ... YAGNI

test/TesterAzureUtils/Persistence/PersistenceGrainTests_AzureTableStore.cs
```diff
@@ -41,7 +41,7 @@ protected override void ConfigureTestCluster(TestClusterBuilder builder)
             {
                 Guid serviceId = Guid.NewGuid();
                 builder.Options.InitialSilosCount = 4;
-
+                builder.Options.UseTestClusterMembership = false;
```
jdom: Aren't other external membership tests (non azure) affected by the same issue. They wouldn't show up as we don't run them in CI

src/Orleans.TestingHost/Orleans.TestingHost.csproj
```diff
@@ -12,6 +12,12 @@
     <OrleansBuildTimeCodeGen>true</OrleansBuildTimeCodeGen>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="Grains\**" />
+    <EmbeddedResource Remove="Grains\**" />
+    <None Remove="Grains\**" />
```
jdom: What's all this?

ReubenBond: Oh, I thought I fixed that - my bad.

jdom:
There are conflicts now. If you resolve them and tests are passing, I'll merge it asap

ReubenBond:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test bvt

ReubenBond:
I'm happy for us to merge this now

jdom:
I was about to on Friday, but some multi cluster tests in vso are failing because of this. Should I merge anyway or wait until they are fixed?

sergeybykov:
There were other issues with the multi-cluster tests. I think there's no need to block this waiting for them to be investigated.

jdom:
Oh, OK then. I'll merge as soon as the conflicts are resokved

ReubenBond:
resolved

jdom:
oh, too slow :(
Was waiting for tests to validate the rebase

xiazen:
:) yeah. you are late for the party ;) . I verified that it passed in vso with Reuben. So I merged it

ReubenBond:
The tests passed in .NET CI, so it's all good

jdom:
Yeah, sorry for the delay, I was thinking that you were planning to fix the multi-cluster tests... I was running tests in VSO for the entire week last week, waiting for the moment they passed so I could merge :)

PR #3887, opened 2018-01-19
Replaced life cycle stage enums with constants

jason-bragg:
Lifecycle stage enums gave the impression that the stages were limited to only the defined enumes, which was not the intention of the design.  They also necessitated the development and maintenance of extensions to support each enum.


jason-bragg:
Tagging @dVakulen, as he aided in bringing attention to this.

Issue #3586, opened 2017-10-23
Performance Counter alternatives in 2.0.0


attilah:
In current Orleans we are using Performance counters for multiple things:
- Expose [Orleans system metrics](https://github.com/dotnet/orleans/blob/master/src/Orleans.Core/Statistics/RuntimeStatisticsGroup.cs) as Performance counters
- Read CPU utilization for the load shedding [streaming](https://github.com/dotnet/orleans/blob/bfcc00f4fcd001334431bbf2511915e75747c192/src/Orleans.Runtime/Streams/LoadShedQueueFlowController.cs) feature
- Collect [Thread context switch](https://github.com/dotnet/orleans/blob/bfcc00f4fcd001334431bbf2511915e75747c192/src/Orleans.Core/Statistics/ThreadTrackingStatistic.cs#L209) statistics
- Make it possible to [register performance counters](https://github.com/dotnet/orleans/blob/bfcc00f4fcd001334431bbf2511915e75747c192/src/Orleans.TelemetryConsumers.Counters/OrleansPerfCounterTelemetryConsumer.cs) for grain by grain type, so grain count per type can be monitored

As Orleans 2.0 will support multiple OS platforms, Performance Counters are not a viable way ahead, so we're looking for alternate solutions:
- Keep the performance counter feature, but separate it into a Windows only package
- Move away from performance counters at all and implement a similar functionality on top of [System.Diagnostics.Tracing.EventSource](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource?view=netcore-2.0) and associated classes, since they're part of .Net Standard 2.0
- Since reading the processor utilization without elevated privileges through ETW is not possible, it's probable, that load shedding will be abstracted out from the system and be a Windows only feature in Orleans 2.0

Building on top of the Tracing infrastructure it enables us to provide richer statistics and insights - with the support of event payloads - into a running system which can be consumed with standard ETW (Event Tracing for Windows) tools, when running under Windows, like [perfview](https://github.com/Microsoft/perfview).

xiazen:
I thought performance counters is back in netstandard, according to @jdom 's notes. Is this not true? 

btw just linking related issue so we can tackle them together. #2609

Issue #3567, opened 2017-10-19
Bring CPU and Memory statistics back 


xiazen:
When Orleans master move to netstandard, we temporarily disabled CPU and Memory statistics reporting , because these two statistics rely on performance counters. There was complication on CPU and memory statistics reporting at the time
- performance counters wasn't supported in netstandard in early versions
- performance counters only work in windows. So when we move to netstandard, we need to find a way to report CPU and memory statistics in other platforms too.  

Now the situation is more clearer with netstandad 2.0. We know that per counters are coming back to netstandard, although only on windows But regardless, we need to find a way to bring CPU and Memory statistic back, on windows, and other platforms.

xiazen:
related issue: #2609 #3586 .

xiazen:
assign it to ben since he is solving this problem now . Feel free to close this one if there's duplicated issue on your bucket

PR #3891, opened 2018-01-22
Bring back PerfCounterEnvironmentStatistics 

benjaminpetit:
Related issues: #3567 #3586

I added abstractions to get the total physical memory and the cpu/memory usage (`IAppEnvironmentStatistics` and `IHostEnvironmentStatistics`), and refactored the old `PerfCounterEnvironmentStatistics` to implement `IAppEnvironmentStatistics`, as a Windows-only solution. This class also now follow the lifecycle pattern.

The built-in implementation of  `IAppEnvironmentStatistics` should be cross-platform, but the developer can make a custom implementation if needed.

A Linux-specific implementation of `IHostEnvironmentStatistics` could be done in the future,

### DIFF
--- a/src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
+++ b/src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
@@ -1,11 +1,11 @@
-﻿namespace Orleans.Statistics
+namespace Orleans.Statistics
 {
     public interface IHostEnvironmentStatistics
     {
-        long TotalPhysicalMemory { get; }
+        long? TotalPhysicalMemory { get; }
 
-        float CpuUsage { get; }
+        float? CpuUsage { get; }
 
-        long AvailableMemory { get; }
+        long? AvailableMemory { get; }
     }
 }
--- a/src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
+++ b/src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
@@ -1,0 +1,11 @@
+﻿namespace Orleans.Statistics
+{
+    public interface IHostEnvironmentStatistics
+    {
+        long TotalPhysicalMemory { get; }
+
+        float CpuUsage { get; }
+
+        long AvailableMemory { get; }
+    }
+}
--- a/src/Orleans.Core/Core/DefaultClientServices.cs
+++ b/src/Orleans.Core/Core/DefaultClientServices.cs
@@ -6,6 +6,7 @@ using Orleans.Metadata;
 using Orleans.Providers;
 using Orleans.Runtime;
 using Orleans.Serialization;
+using Orleans.Statistics;
 using Orleans.Streams;
 using Orleans.Streams.Core;
 
@@ -17,6 +18,7 @@ namespace Orleans
         {
             services.TryAddSingleton<TelemetryManager>();
             services.TryAddFromExisting<ITelemetryProducer, TelemetryManager>();
+            services.TryAddSingleton<IAppEnvironmentStatistics, AppEnvironmentStatistics>();
             services.AddLogging();
             services.TryAddSingleton<ExecutorService>();
             services.TryAddSingleton<TypeMetadataCache>();
--- a/src/Orleans.Core/Core/DefaultClientServices.cs
+++ b/src/Orleans.Core/Core/DefaultClientServices.cs
@@ -18,6 +18,8 @@ namespace Orleans
         {
             services.TryAddSingleton<TelemetryManager>();
             services.TryAddFromExisting<ITelemetryProducer, TelemetryManager>();
+            // if no IHostEnvironmentStatistics registered, use the dummy one
+            services.TryAddSingleton<IHostEnvironmentStatistics, DummyHostEnvironmentStatistics>();
             services.TryAddSingleton<IAppEnvironmentStatistics, AppEnvironmentStatistics>();
             services.AddLogging();
             services.TryAddSingleton<ExecutorService>();
--- a/src/Orleans.Core/Core/DefaultClientServices.cs
+++ b/src/Orleans.Core/Core/DefaultClientServices.cs
@@ -18,8 +18,7 @@ namespace Orleans
         {
             services.TryAddSingleton<TelemetryManager>();
             services.TryAddFromExisting<ITelemetryProducer, TelemetryManager>();
-            // if no IHostEnvironmentStatistics registered, use the dummy one
-            services.TryAddSingleton<IHostEnvironmentStatistics, DummyHostEnvironmentStatistics>();
+            services.TryAddSingleton<IHostEnvironmentStatistics, NoOpHostEnvironmentStatistics>();
             services.TryAddSingleton<IAppEnvironmentStatistics, AppEnvironmentStatistics>();
             services.AddLogging();
             services.TryAddSingleton<ExecutorService>();
--- a/src/Orleans.Core/Shims/RuntimeStatisticsGroup.cs
+++ b/src/Orleans.Core/Shims/RuntimeStatisticsGroup.cs
@@ -4,7 +4,7 @@ using Orleans.Statistics;
 
 namespace Orleans.Runtime
 {
-    internal class DummyHostEnvironmentStatistics : IHostEnvironmentStatistics
+    internal class NoOpHostEnvironmentStatistics : IHostEnvironmentStatistics
     {
         public long? TotalPhysicalMemory => null;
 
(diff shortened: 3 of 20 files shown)


src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Orleans.Statistics
+{
+    public interface IHostEnvironmentStatistics
+    {
+        long TotalPhysicalMemory { get; }
+
+        float CpuUsage { get; }
```
ReubenBond: Prefer if these names were suffixed with their units, like Bytes/Percentage/Proportion. Is CpuUsage 0-100 or 0-1?

src/Orleans.Core/Core/DefaultClientServices.cs
```diff
@@ -17,6 +18,9 @@ public static void AddDefaultServices(IClientBuilder builder, IServiceCollection
         {
             services.TryAddSingleton<TelemetryManager>();
             services.TryAddFromExisting<ITelemetryProducer, TelemetryManager>();
+            // if no IHostEnvironmentStatistics registered, use the dummy one
+            services.TryAddSingleton<IHostEnvironmentStatistics, DummyHostEnvironmentStatistics>();
```
xiazen: I don't think we should inject a dummy one y default. If user missed the warning line of   "No implementation of IHostEnvironmentStatistics was found. Load shedding will not work yet", user may actually take those dummy statistic reading seriously.

xiazen: I'm fine with injecting a NoOp one, to make dependency injection easier. But you may need to change `IHostEnvironmentStatistics ` interface to have a no-op implementation.

src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
benjaminpetit: I agree with you. I wanted to not change too much previously existing naming, since they are used elsewhere. But if everybody is ok, I can update it everywhere

src/Orleans.TelemetryConsumers.Counters/Statistics/PerfCounterEnvironmentStatistics.cs
```diff
@@ -0,0 +1,274 @@
...
+                    new PerformanceCounter(".NET CLR Memory", "Gen 2 heap size", thisProcess, true)
+                };
+                allocatedBytesPerSecPF = new PerformanceCounter(".NET CLR Memory", "Allocated Bytes/sec", thisProcess, true);
+                promotedMemoryFromGen1PF = new PerformanceCounter(".NET CLR Memory", "Promoted Memory from Gen 1", thisProcess, true);
+                numberOfInducedGCsPF = new PerformanceCounter(".NET CLR Memory", "# Induced GC", thisProcess, true);
+                largeObjectHeapSizePF = new PerformanceCounter(".NET CLR Memory", "Large Object Heap size", thisProcess, true);
+                promotedFinalizationMemoryFromGen0PF = new PerformanceCounter(".NET CLR Memory", "Promoted Finalization-Memory from Gen 0", thisProcess, true);
+#endif
+
+#if !(NETSTANDARD || __MonoCS__)
```
ReubenBond: We don't build this package for .NET Standard or Mono, do we? That's fine, it just means some code will be untested.

src/Orleans.TelemetryConsumers.Counters/Statistics/PerfCounterEnvironmentStatistics.cs
```diff
@@ -0,0 +1,274 @@
...
+                // GetAvailableThreads Retrieves the difference between the maximum number of thread pool threads
+                // and the number currently active.
+                // So max-Available is the actual number in use. If it goes beyond min, it means we are stressing the thread pool.
+                ThreadPool.GetAvailableThreads(out workerThreads, out completionPortThreads);
+                return maXworkerThreads - workerThreads;
+            });
+            IntValueStatistic.FindOrCreate(StatisticNames.RUNTIME_DOT_NET_THREADPOOL_INUSE_COMPLETIONPORTTHREADS, () =>
+            {
+                int maXworkerThreads;
+                int maXcompletionPortThreads;
```
ReubenBond: nit: capitalization on X

src/Orleans.Core/Shims/RuntimeStatisticsGroup.cs
```diff
@@ -1,34 +1,22 @@
...
         public float CpuUsage => 0;
 
-        public RuntimeStatisticsGroup(ILoggerFactory loggerFactory)
-        {
-            this.logger = loggerFactory.CreateLogger<RuntimeStatisticsGroup>();
-        }
+        public long AvailableMemory => TotalPhysicalMemory;
 
-        internal void Start()
+        public DummyHostEnvironmentStatistics(ILoggerFactory loggerFactory)
```
xiazen: you can just inject `ILogger<DummyHostEnvironmentStatistics>` instead of `ILoggerFactory`

galvesribeiro:
Nice @benjaminpetit! I'll add the *NIX implementation once this PR is merged.

src/Orleans.Core/Core/DefaultClientServices.cs
jdom: Unless the user is doing something very custom, *I believe* the default way of using the silo builder is that `AddDefaultServices` will be called before all app-level services are configured. You might want to double check, but that's what I remember, so the comment would not be accurate.
IIRC, the reason many of these use the `TryAddXXX` methods is to allow infrastructure to be tweaked, but not necessarily for end users to provide their own by default.

src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
ReubenBond: I'm ok either way

benjaminpetit:
@galvesribeiro that would be awesome!

src/Orleans.Core/Shims/RuntimeStatisticsGroup.cs
```diff
@@ -1,34 +1,22 @@
-﻿using System;
+using System;
 using Microsoft.Extensions.Logging;
+using Orleans.Statistics;
 
 namespace Orleans.Runtime
 {
-    internal class RuntimeStatisticsGroup : IDisposable
+    internal class DummyHostEnvironmentStatistics : IHostEnvironmentStatistics
```
xiazen: naming Dummy is not clear of what it is doing. Maybe `NoOpHostEnvironmentStatistics`?

src/Orleans.TelemetryConsumers.Counters/Statistics/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿#define LOG_MEMORY_PERF_COUNTERS
+
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Hosting;
+
+namespace Orleans.Statistics
+{
+    public static class SiloHostBuilderExtensions
+    {
+        public static ISiloHostBuilder UsePerfCounterEnvironmentStatistics(this ISiloHostBuilder builder)
```
xiazen: miss xml comment

src/Orleans.TelemetryConsumers.Counters/Statistics/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
+            return builder.ConfigureServices(services => services.AddSingleton<IHostEnvironmentStatistics, PerfCounterEnvironmentStatistics>());
+        }
+    }
+
+    public static class ClientBuilderExtensions
+    {
+        public static IClientBuilder UsePerfCounterEnvironmentStatistics(this IClientBuilder builder)
```
xiazen: same here

src/Orleans.Core/Shims/RuntimeStatisticsGroup.cs
```diff
@@ -1,34 +1,22 @@
...
-            this.logger = loggerFactory.CreateLogger<RuntimeStatisticsGroup>();
-        }
-
-        internal void Start()
-        {
-
-            logger.Warn(ErrorCode.PerfCounterNotRegistered,
-                "CPU & Memory perf counters are not available in .NET Standard. Load shedding will not work yet");
-        }
+        public float? CpuUsage => null;
```
xiazen: I think they are null by default , you don't need to assign them

benjaminpetit: - I think it is bad to rely on "default" value
- Here I defined a getter, I have to return something

src/Orleans.Core/Shims/RuntimeStatisticsGroup.cs
```diff
@@ -4,17 +4,17 @@
...
 
         public float? CpuUsage => null;
 
         public long? AvailableMemory => null;
 
-        public DummyHostEnvironmentStatistics(ILoggerFactory loggerFactory)
+        public NoOpHostEnvironmentStatistics(ILoggerFactory loggerFactory)
         {
-            var logger = loggerFactory.CreateLogger<DummyHostEnvironmentStatistics>();
+            var logger = loggerFactory.CreateLogger<NoOpHostEnvironmentStatistics>();
```
jdom: should just inject `ILogger<NoOpHostEnvironmentStatistics>`, no?

PR #3893, opened 2018-01-22
increase timeout on ClientStreamTestRunner

xiazen:
Test fail intermittently in vso. Increase the timeout to make it pass more consistently


xiazen:
timeout is already pretty high, not sure increasing it will make the situation  better. close this solution for now

PR #3897, opened 2018-01-24
Add TelemetryOptions to configure Telemetry consumers

benjaminpetit:
I tested locally, but unfortunately I don't have unit tests for this, since it is a Windows specific features...

We should be able to add some testing in the future I think.

### DIFF
--- a/src/Orleans.Core/Configuration/Options/TelemetryOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/TelemetryOptions.cs
@@ -1,0 +1,20 @@
+using Orleans.Runtime;
+using System;
+using System.Collections.Generic;
+
+namespace Orleans.Configuration.Options
+{
+    public class TelemetryOptions
+    {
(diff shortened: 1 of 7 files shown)


src/Orleans.Core/Configuration/Options/TelemetryOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Orleans.Configuration.Options
+{
+    public class TelemetryOptions
+    {
+        internal IList<Type> Consumers { get; set; } = new List<Type>();
+    }
+
+    public static class TelemetryOptionsExtensions
+    {
+        public static TelemetryOptions AddConsumer<T>(this TelemetryOptions options) where T : ITelemetryConsumer
```
ReubenBond: I thought we had build warnings for lack of doc comments on public types

jason-bragg: Don't think that is active on core assembly.

PR #3901, opened 2018-01-24
Move ClusterConfiguration to legacy

jason-bragg:
Moved Cluster configuration to Orleans.Core.Legacy.
All references to it and related configuration classes have been removed from core Orleans packages.
Added a number of options to account for configuration data that had not been moved to options.
External packages and test libraries still rely on ClusterConfiguration and now depend on Orleans.Core.Legacy.  This can be cleaned up in future passes.


jason-bragg:
Areas of special interest for review:

New options
All new classes in Orleans.Core/Configuration/Options
(note: Statistics and Messaging options need be shared and are currently broke, wip)

Data+Logic
Configuration was mixed with logic in many cases.  Options deal only with data now, so some of the logic needed to be duplicated or refactored.  Most of this was in relation to ApplicationConfiguration (which was not moved yet, as it's also used by the client configuration, and the LimitsManager.
Strategies

Placement, Compatibility, and Versioning strategies all followed a pattern that was not well suited to component configuration and DI.  Wiring of these classes have been refactored to use keyed services lookup.

Legacy configurators
Plan for how to handle legacy configurators (like ILegacyMembershipConfigurator) didn't work.  Went back to original suggestion.  Open to further discussion, but as original plan didn't pan out and this PR is already a large change, I'd rather revisit alternatives in a future pass.

jason-bragg:
Follow-up - this is a list of configuration related task we need to follow up on after this PR goes in
* Duplicate defaults.  Many of the previously moved configuration duplicated the defaults in the options, rather than having the options have the authoritative default and referencing that from the legacy configurations.  All (or most) options I added in this PR use consistent defaults, but we need to go back and clean up all of the previously duplicated defaults.
* Xml docs for all new options
* Move ClientConfiguration and remaining config classes to legacy.
* Move legacy provider interfaces to legacy
* Remove use of legacy configuration from TestHost, introducing a TestHost.Legacy if needed.
* Update extension libraries to use options rather than legacy configuration (and remove depenancies on Legacy packages).

jason-bragg:
Broke changes into smaller changes sets to make it easier for review.
- Core (and abstractions)
- Runtime (and it's abstractions)
- Legacy assemblies (core and runtime)
- Remaining core packages (providers, transactions, even sourcing, ..)
- Everything else

src/Orleans.Core/Configuration/Options/MessagingOptions.cs
```diff
@@ -1,20 +1,19 @@
...
+    public abstract class MessagingOptions
     {
         /// <summary>
         /// The ResponseTimeout attribute specifies the default timeout before a request is assumed to have failed.
         /// </summary>
         public TimeSpan ResponseTimeout { get; set; } = Constants.DEFAULT_RESPONSE_TIMEOUT;
 
-        /// <summary>
-        /// The MaxResendCount attribute specifies the maximal number of resends of the same message.
+        /// <summaryResponseTimeout number of resends of the same message.
```
jason-bragg: looks like I messed up this summary somehow.. will fix

src/Orleans.Core/Configuration/Options/MultiClusterOptions.cs
```diff
@@ -42,6 +51,9 @@ public class MultiClusterOptions
         /// <summary>
         /// The time between the slow retries for DOUBTFUL activations.
         /// </summary>
-        public TimeSpan GlobalSingleInstanceRetryInterval { get; set; } = TimeSpan.FromSeconds(30);
+        public TimeSpan GlobalSingleInstanceRetryInterval { get; set; } = DEFAULT_GLOBAL_SINGLE_INSTANCE_RETRY_INTERVAL;
+        public static readonly TimeSpan DEFAULT_GLOBAL_SINGLE_INSTANCE_RETRY_INTERVAL = TimeSpan.FromSeconds(30);
+
+        public Dictionary<string, string> GossipChannels { get; set; } = new Dictionary<string, string>();
```
xiazen: GossipeChannels are used in MultiClusterGossipeFactory to create `IGossipChannel` by using `AssemblyLoader.LoadAndCreateInstance<IGossipChannel>`. Should we make GossipChannels to be Dictionary<Type,String> , and use DI to configure `IGossipChannel` in `MultiClusterGossipFactory`?

src/Orleans.Core/Configuration/Options/ReminderOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+            public const string Disabled = "Disabled";
+
+            /// <summary>Use custom Reminder Service from third-party assembly</summary>
+            public const string Custom = "Custom";
+        }
+
+        /// <summary>
+        /// The ReminderService attribute controls the type of the reminder service implementation used by silos.
+        /// </summary>
+        public string ReminderService { get; set; }
```
xiazen: same here, string vs Type

veikkoeeva:
@jason-bragg Regarding the comment about XML documents and other documentation, could it be possible to have them in this PR? Or will there be beta 4 with comments? It would make migrating configuration heaps smoother. :)

src/AWS/Orleans.Clustering.DynamoDB/Membership/LegacyDynamoDBMembershipConfigurator.cs
```diff
@@ -7,9 +7,10 @@ namespace Orleans.Runtime.MembershipService
     /// <inheritdoc />
     public class LegacyDynamoDBMembershipConfigurator : ILegacyMembershipConfigurator
     {
-        public void ConfigureServices(GlobalConfiguration configuration, IServiceCollection services)
+        public void ConfigureServices(object configuration, IServiceCollection services)
         {
-            services.UseDynamoDBMembership(options => options.ConnectionString = configuration.DataConnectionString);
+            var reader = new GlobalConfigurationReader(configuration);
```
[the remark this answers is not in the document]
jason-bragg: I touched on this in the PR comments.  The pattern we'd planned on didn't work.  The issue was that either way something needs to know about something it has no knowledge of, either the options or the clusterConfiguration.  As you can see from how this pattern turned out, it's not a bad approach.  Probably the least bad of the possible approaches.

src/AWS/Orleans.Clustering.DynamoDB/Orleans.Clustering.DynamoDB.csproj
```diff
@@ -22,6 +22,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\Orleans.Core.Legacy\Orleans.Core.Legacy.csproj" />
```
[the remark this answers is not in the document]
jason-bragg: I've pulled all references of Cluster config out of the core packages, but not all of the extension libraries.  So any library that still references cluster config needs depend on core legacy.  This can (and IMHO should) be handled in smaller, more targeted change sets rather than adding to this one.

src/AdoNet/Orleans.Clustering.AdoNet/Orleans.Clustering.AdoNet.csproj
```diff
@@ -34,7 +34,9 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\Orleans.Core.Legacy\Orleans.Core.Legacy.csproj" />
```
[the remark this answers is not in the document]
jason-bragg: same answer. :/

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -272,29 +274,27 @@ protected virtual void InitEventHubClient()
...
 
-            var receiverMonitorDimensions = new EventHubReceiverMonitorDimensions();
-            receiverMonitorDimensions.EventHubPartition = config.Partition;
-            receiverMonitorDimensions.EventHubPath = config.Hub.Path;
-            receiverMonitorDimensions.NodeConfig = this.serviceProvider.GetRequiredService<NodeConfiguration>();
-            receiverMonitorDimensions.GlobalConfig = this.serviceProvider.GetRequiredService<GlobalConfiguration>();
+            var receiverMonitorDimensions = new EventHubReceiverMonitorDimensions
+            {
+                EventHubPartition = config.Partition,
+                EventHubPath = config.Hub.Path,
```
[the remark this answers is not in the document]
jason-bragg: Yes, I removed these.  If we can clearly identify what configuration information is needed, I don't mind adding it into the dimensions, but I'm not clear what is needed.  Only thing we were using was the ip, which I don't think really has much value.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -272,29 +274,27 @@ protected virtual void InitEventHubClient()
         /// <returns></returns>
         protected virtual IEventHubQueueCacheFactory CreateCacheFactory(EventHubStreamProviderSettings providerSettings)
         {
-            var globalConfig = this.serviceProvider.GetService<GlobalConfiguration>();
-            var nodeConfig = this.serviceProvider.GetService<NodeConfiguration>();
-            var eventHubPath = hubSettings.Path;
-            var sharedDimensions = new EventHubMonitorAggregationDimensions(globalConfig, nodeConfig, eventHubPath);
-            return new EventHubQueueCacheFactory(providerSettings, SerializationManager, sharedDimensions, this.loggerFactory);
+            var eventHubPath = this.hubSettings.Path;
+            var sharedDimensions = new EventHubMonitorAggregationDimensions(eventHubPath);
```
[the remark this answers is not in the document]
jason-bragg: yeah removed cluster and node configs from the dimensions.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/StatisticMonitors/MonitorAggregationDimentions.cs
```diff
@@ -22,11 +15,8 @@ public class EventHubMonitorAggregationDimensions : MonitorAggregationDimensions
         /// <summary>
         /// Constructor
         /// </summary>
-        /// <param name="globalConfig"></param>
-        /// <param name="nodeConfig"></param>
         /// <param name="ehHubPath"></param>
-        public EventHubMonitorAggregationDimensions(GlobalConfiguration globalConfig, NodeConfiguration nodeConfig, string ehHubPath)
-            :base(globalConfig, nodeConfig)
+        public EventHubMonitorAggregationDimensions(string ehHubPath)
```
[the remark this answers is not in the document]
jason-bragg: yup.. all gone.

src/Orleans.Core.Abstractions/Placement/ActivationCountBasedPlacement.cs
```diff
@@ -6,18 +6,5 @@ namespace Orleans.Runtime
...
         internal static ActivationCountBasedPlacement Singleton { get; } = new ActivationCountBasedPlacement();
-
-        private ActivationCountBasedPlacement()
-        {}
-        
-        public override bool Equals(object obj)
-        {
-            return obj is ActivationCountBasedPlacement;
-        }
-
```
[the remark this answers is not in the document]
jason-bragg: Yeah, this was just useless noise.  Ran across it when trying to understand how strategies where used.  Killed it because it just created confusion and was pointless.

src/Orleans.Core.Abstractions/Placement/StatelessWorkerPlacement.cs
```diff
@@ -5,32 +5,20 @@ namespace Orleans.Runtime
...
-            MaxLocal = maxLocal > 0 ? maxLocal : defaultMaxStatelessWorkers;
+            this.MaxLocal = maxLocal > 0 ? maxLocal : DefaultMaxStatelessWorkers;
         }
 
         public override string ToString()
         {
-            return String.Format("StatelessWorkerPlacement(max={0})", MaxLocal);
-        }
-
-        public override bool Equals(object obj)
```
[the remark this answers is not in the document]
jason-bragg: more noise

src/Orleans.Core.Abstractions/Versions/Selector/AllCompatibleVersions.cs
```diff
@@ -6,18 +6,5 @@ namespace Orleans.Versions.Selector
     public class AllCompatibleVersions : VersionSelectorStrategy
     {
         public static AllCompatibleVersions Singleton { get; } = new AllCompatibleVersions();
-
-        private AllCompatibleVersions()
```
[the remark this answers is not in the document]
jason-bragg: It doesn't matter.  These classes are just marker.

src/Orleans.Core.Abstractions/Versions/Selector/IVersionSelector.cs
```diff
@@ -9,22 +9,8 @@ public interface IVersionSelector
         IReadOnlyList<ushort> GetSuitableVersion(ushort requestedVersion, IReadOnlyList<ushort> availableVersions, ICompatibilityDirector compatibilityDirector);
     }
 
-    public interface IVersionSelector<TPolicy> : IVersionSelector where TPolicy : VersionSelectorStrategy
```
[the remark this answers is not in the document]
jason-bragg: These interfaces existed only to provide mappings from strategy types to other implementations, which requires reflection and is somewhat clumsy.  Replaced this with typed services using type as key.

src/Orleans.Core.Legacy/Configuration/GlobalConfiguration.cs
```diff
@@ -85,6 +86,28 @@ public enum ReminderServiceProviderType
             Custom
         }
 
+        public static string Remap(ReminderServiceProviderType type)
+        {
+            switch(type)
+            {
+                case ReminderServiceProviderType.NotSpecified:
+                    return ReminderOptions.BuiltIn.NotSpecified;
```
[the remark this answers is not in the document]
jason-bragg: Yeah.  I already had this working this way when we made that decision.  This uses names to perform type lookups.  I think we should vet that reading Types from configuration works as expected before moving off of this.  Though I don't have a strong opinion.  In either case, this mapping logic is necessary, it's just a question of what we map to.

src/Orleans.Core.Legacy/Configuration/GlobalConfiguration.cs
```diff
@@ -474,140 +539,112 @@ internal bool UseZooKeeperSystemStore
...
 
-        internal bool RunsInAzure { get { return UseAzureSystemStore && !String.IsNullOrWhiteSpace(this.ClusterId); } }
-
-        private static readonly TimeSpan DEFAULT_LIVENESS_PROBE_TIMEOUT = TimeSpan.FromSeconds(10);
-        private static readonly TimeSpan DEFAULT_LIVENESS_TABLE_REFRESH_TIMEOUT = TimeSpan.FromSeconds(60);
-        private static readonly TimeSpan DEFAULT_LIVENESS_DEATH_VOTE_EXPIRATION_TIMEOUT = TimeSpan.FromSeconds(120);
-        private static readonly TimeSpan DEFAULT_LIVENESS_I_AM_ALIVE_TABLE_PUBLISH_TIMEOUT = TimeSpan.FromMinutes(5);
-        private static readonly TimeSpan DEFAULT_LIVENESS_MAX_JOIN_ATTEMPT_TIME = TimeSpan.FromMinutes(5); // 5 min
-        private static readonly TimeSpan DEFAULT_REFRESH_CLUSTER_INTERFACEMAP_TIME = TimeSpan.FromMinutes(1);
-        private const int DEFAULT_LIVENESS_NUM_MISSED_PROBES_LIMIT = 3;
```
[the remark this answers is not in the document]
jason-bragg: yes.  imo defaults should live in their associated options, not in the legacy configurations.  Moves defaults for all options I added, or properties I moved, still more work to do here for pre-existing options, which duplicate the defaults (bad in my view).  Those changes are noted in future work comments on this issue.

src/Orleans.Core.Legacy/Configuration/NodeConfiguration.cs
```diff
@@ -190,131 +191,122 @@ public IPEndPoint Endpoint
         public string SiloShutdownEventName { get; set; }
 
         internal const string DEFAULT_NODE_NAME = "default";
-        private static readonly TimeSpan DEFAULT_STATS_METRICS_TABLE_WRITE_PERIOD = TimeSpan.FromSeconds(30);
-        private static readonly TimeSpan DEFAULT_STATS_PERF_COUNTERS_WRITE_PERIOD = TimeSpan.FromSeconds(30);
-        private static readonly TimeSpan DEFAULT_STATS_LOG_WRITE_PERIOD = TimeSpan.FromMinutes(5);
-        internal static readonly StatisticsLevel DEFAULT_STATS_COLLECTION_LEVEL = StatisticsLevel.Info;
-        private static readonly int DEFAULT_MAX_ACTIVE_THREADS = Math.Max(4, System.Environment.ProcessorCount);
-        private const int DEFAULT_MIN_DOT_NET_THREAD_POOL_SIZE = 200;
-        private static readonly int DEFAULT_MIN_DOT_NET_CONNECTION_LIMIT = DEFAULT_MIN_DOT_NET_THREAD_POOL_SIZE;
```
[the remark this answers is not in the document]
jason-bragg: each was moved to it's associated options class.  Most of the names were preserved, so they can be found by name.

src/Orleans.Core/Configuration/Options/MultiClusterOptions.cs
jason-bragg: I'm of the opinion that the gossip channels are poorly structured.  I agree we should refactor them.  IMO, such refactoring is out of scope for this PR.  I added gossip channel support in the simplest way, as I don't have the bandwidth to think through how these "should" be configured.  Basically, since their currently configured poorly, I've opted to continue to support them poorly :/

src/Orleans.Core/Configuration/Options/ReminderOptions.cs
jason-bragg: This is a good example of why types in config can be a problem.   We support several reminder types, but we can't case of them an load the related assemblies by type, because if the assembly is not loaded we can't have the type.  Please take a look at ReminderTableFactory's usage of this value and consider how that logic can be performed with Type.

jason-bragg:
@veikkoeeva, I may have been unclear, I wasn't referring to documentation, only noting that, in this port of properties from configuration to options, I've not taken the time to add comments for settings that did not already have comments.  This is not a regression, just a noting of more work that needs be done.  For instance, consider the setting UseVirtualBucketsConsistentRing which is now configured via the ConsistentRingOptions.  The original setting had no comments, so neither does the new setting.  Given the scope of this PR it would be challenging to investigate every setting and provide an adequate comment.  Though, it is important for this to be done, hence I'm calling it out as future work which should be done prior to 2.0 RC.

src/AWS/Orleans.Clustering.DynamoDB/Membership/LegacyDynamoDBMembershipConfigurator.cs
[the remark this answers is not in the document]
xiazen: sorry my bad, I went read it and deleted some noisy comments

src/Orleans.Core/Configuration/Options/MultiClusterOptions.cs
xiazen: can we add a todo item for those components which use assembly loading to create instance, since this pattern contradicts the direction we are going ? And then let's discuss the priority of those todos?

src/Orleans.Core/Configuration/Options/ReminderOptions.cs
xiazen: sorry I wasn't clear on what my concerns are. The usage in `ReminderTableFactory` is the reason I raised this concern. Using assembly loading to create reminder table isn't compile time safe and more error prone. The pattern used in other places to avoid this, say GatewayListProvider, is we have extension methods like `UseAzureGatewayListProvider` to configure DI with the proper gateway list provider. This avoided using string to map to the related provider type, and use assembly loader to load the assembly and then create the provider. 

Whether this is out of the scope of this PR is debatable. Since for gatewaylist provider, creating those extension methods and refactor the infra is part of the work for move gatewatlist provider to use options. But if we decide it is out of scope, at least we add a todo to go back and refactor this.

jason-bragg: I agree that refactoring systems to fit the new model needs be done, and in some cases  (as with the strategies) I've done this, but in other cases I've not.  So minimally I've been inconsistent on this effort.  Given the size and complexity of this PR I suggest we push to future iterations any changes that we can.  As you've also touched on, we can (should) note these and revisit them in future passes (prior to 2.0 RC).

xiazen: agreed

src/Orleans.Core/Configuration/Options/ConsistentRingOptions.cs
```diff
@@ -0,0 +1,12 @@
+
+namespace Orleans.Hosting
+{
+    public class ConsistentRingOptions
```
xiazen: I think this class should be in `orleans.runtime`, where it is used and where `VirtualBucketsRingProvider` lives

src/Orleans.Core/Configuration/Options/GrainCollectionOptions.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+
+namespace Orleans.Hosting
+{
+    /// <summary>
+    /// Silo options for grain garbage collection.
+    /// </summary>
+    public class GrainCollectionOptions
```
xiazen: I think this class should be in Orleans.Runtime, since the component it configures lives there.

src/Orleans.Core/Configuration/Options/GrainDirectoryOptions.cs
```diff
@@ -0,0 +1,66 @@
+
+using System;
+
+namespace Orleans.Hosting
+{
+    public class GrainDirectoryOptions
```
xiazen: I noticed all the option are in orleans.core/configurations. is this by design or ? 

maybe making some exception for some edge case, but as as general thumb of rules, IMO the option class should be in the same project where its component lives.

I will skip the same comment on the option class where I think lived in the wrong place

src/Orleans.Core/Configuration/Options/ReminderOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        /// <summary>
+        /// Assembly to use for custom ReminderTable implementation
+        /// </summary>
+        public string ReminderTableAssembly { get; set; }
+
+        /// <summary>
+        /// For TEST - Not for use in production environments.
```
xiazen: group these three property in a region or add the same comment to them. since I think they are all for test purpose. 
adding comment on just one of them isn't clear enought IMO .

Since if the user hover on one of them, the xml comment will only show up for the first one.

src/Orleans.Core/Configuration/Options/ServicePointOptions.cs
```diff
@@ -0,0 +1,13 @@
+
+namespace Orleans.Hosting
+{
+    public class ServicePointOptions
```
xiazen: if `ServicePointOptions` is only azure storage related , maybe this is not the place it belongs. and also line 637 - 643 in silo.cs should be moved to azure extensions maybe

src/Orleans.Core/Configuration/Options/StorageOptions.cs
```diff
@@ -0,0 +1,13 @@
+
+namespace Orleans.Hosting
+{
+    public class StorageOptions
```
xiazen: I'm not comfortable with the approach where different storage provider uses the same storage options. The point of moving to ASP.NET is to move to the direction where different component has its own options, which fits it most. Having one option for all storage component is against that principle. 

For example, you used this option in DynamoDB reminder table. But in dynamo Db storage , there's no concept of data connection string. Instead of that, it has access key and secret key. While azure storage has the concept of data connection string. So a one for all solution is not good.

src/Orleans.Core/Configuration/Options/StorageOptions.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        public string DataConnectionString { get; set; }
+
+        /// <summary>
+        /// Set this property to override <see cref="DataConnectionString"/> for reminders.
+        /// </summary>
+        public string DataConnectionStringForReminders { get; set; }
```
xiazen: isn't think should belong to reminder option, instead of storage option? I'm so confused here.

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -132,10 +132,10 @@ internal void ConsumeServices(IServiceProvider services)
 
             this.GrainReferenceRuntime = this.ServiceProvider.GetRequiredService<IGrainReferenceRuntime>();
 
-            var statisticsOptions = this.ServiceProvider.GetRequiredService<IOptions<StatisticsOptions>>();
+            var statisticsOptions = this.ServiceProvider.GetRequiredService<IOptions<ClientStatisticsOptions>>();
```
xiazen: I think the general rule of thumb is to store `IOption<TOptions>`.Value, instead of `IOption<TOPtions>`

src/Orleans.Core/Configuration/Options/GrainDirectoryOptions.cs
jason-bragg: Where options should live is definitely a subject we should consider.  My understanding was that the options that are considered part of our framework (all these options) should be kept in core, and be in the Orleans.Hosting namespace.  Options for extensions and other components should be with those components.  For this PR, this is what I've tried to do.

The reason for the namespace was, if I recall, discoverability.
The reason for keeping them all in core, was that core was all that was needed to configure Orleans before, so moving these around could introduce new dependencies.

What should be mostly clear from the state of the code before this PR is that the pattern I've followed here conformed to the pre-existing pattern and revisiting that pattern was not part of this work.

src/Orleans.Core/Configuration/Options/ServicePointOptions.cs
jason-bragg: Service point manager seems to be part of .net, not azure.  I think it's only used by azure storage, but it's likely used by other things.  IMO, this should not be part of Orleans at all, but let's look at removing it in a future pass, after we've examined the implications.

src/Orleans.Core/Configuration/Options/SiloStatisticsOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+        public bool LoadSheddingEnabled { get; set; }
+
+        /// <summary>
+        /// The LoadLimit attribute specifies the system load, in CPU%, at which load begins to be shed.
+        /// Note that this value is in %, so valid values range from 1 to 100, and a reasonable value is
+        /// typically between 80 and 95.
+        /// This value is ignored if load shedding is disabled, which is the default.
+        /// If load shedding is enabled and this attribute does not appear, then the default limit is 95%.
+        /// </summary>
+        public int LoadSheddingLimit { get; set; } = DEFAULT_LOAD_SHEDDING_LIMIT;
```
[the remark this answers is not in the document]
jason-bragg: Agreed, but I was trying to avoid making too many tiny options objects, so I grouped things 'mostly' by the systems responsible for them.  The statistics system is currently responsible for this, even if it shouldn't be.  My preference would be to keep this here until we refactor the statistics system, but if you've a strong objection, I can move it to it's own options.

It should be noted that moving this to it's own options also means wiring up that new option in every place that needs this, most of which already had access to the statistics options.

src/Orleans.Core/Configuration/Options/StorageOptions.cs
jason-bragg: I understand and agree, but this was how the previous configuration worked.  The host storage information was used for everything unless overridden by a component.  I do not agree with this pattern, but designing a new pattern was not part of this work.  Further, most of the existing providers make this same assumption, so we need to clean up all of that before we can move away from this pattern.  I have done some thought on how to address this, but am not addressing it in this PR.  I'll add it to the notes to be addressed in future passes.

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
jason-bragg: We don't do this remotely consistently.  Why does this matter?

jason-bragg:
Added note a for adding formatters for all the new options to the list of follow-on work.  Won't be addressing it in this PR.  In general I think we should add formatters with the options as we add them, but the scope of this PR makes this unwise.  Anything that can be pushed to follow-up work, should be.

src/Orleans.Core/Messaging/BufferPool.cs
```diff
@@ -39,11 +39,9 @@ public string Name
             private set;
         }
 
-        internal static void InitGlobalBufferPool(IOptions<MessagingOptions> messagingOptions)
+        internal static void InitGlobalBufferPool(MessagingOptions messagingOptions)
         {
-            var messagingOptionsValue = messagingOptions.Value;
-
-            GlobalPool = new BufferPool(messagingOptionsValue.BufferPoolBufferSize, messagingOptionsValue.BufferPoolMaxSize, messagingOptionsValue.BufferPoolPreallocationSize, "Global");
+            GlobalPool = new BufferPool(messagingOptions.BufferPoolBufferSize, messagingOptions.BufferPoolMaxSize, messagingOptions.BufferPoolPreallocationSize, "Global");
```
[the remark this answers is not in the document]
jason-bragg: That should be addressed in the work slated for removing statics.

src/Orleans.Core/Configuration/Options/StorageOptions.cs
jason-bragg: I don't know the history, or why this is the way it is, but it looks like there was a default account for storage, then an override was added for reminder storage, and each provider used it's own overrides for whichever default is used.  I didn't make this pattern, just ported it :/

Please keep in mind that refactoring these patterns are non-trivial, as there are expectations here used by various extensions, and we need to be able to map from the cluster configuration onto whatever new pattern we come up with.

src/Orleans.Core.Abstractions/Placement/PreferLocalPlacement.cs
```diff
@@ -6,18 +6,5 @@ namespace Orleans.Runtime
     internal class PreferLocalPlacement : PlacementStrategy
     {
         internal static PreferLocalPlacement Singleton { get; } = new PreferLocalPlacement();
-        
-        private PreferLocalPlacement()
```
ReubenBond: why these changes?

src/Orleans.Core/Configuration/Options/GrainDirectoryOptions.cs
ReubenBond: I think either `Orleans` or `Orleans.Configuration`, but I'm not sure. What matter most is discoverability

src/Orleans.Core/Configuration/Options/ServicePointOptions.cs
ReubenBond: >  IMO, this should not be part of Orleans at all, but let's look at removing it in a future pass, after we've examined the implications.

👍

PR #3914, opened 2018-01-26
Remove locks in LoadedProviderTypeLoaders and ProviderTypeLoader

sergeybykov:
Remove locks in LoadedProviderTypeLoaders and ProviderTypeLoader that appear unnecessary and caused occasional deadlocks in tests.

### DIFF
--- a/src/Orleans.Core/Providers/ProviderTypeLoader.cs
+++ b/src/Orleans.Core/Providers/ProviderTypeLoader.cs
@@ -28,21 +28,16 @@ namespace Orleans.Providers
                 return;
             }
 
-            // We do this under the lock to avoid race conditions when an assembly is added 
-            // while a type manager is initializing.
-            lock (this.Managers)
+            // We assume that it's better to fetch and iterate through the list of types once,
+            // and the list of TypeManagers many times, rather than the other way around.
+            // Certainly it can't be *less* efficient to do it this way.
+            foreach (var type in TypeUtils.GetDefinedTypes(args.LoadedAssembly, logger))
             {
-                // We assume that it's better to fetch and iterate through the list of types once,
-                // and the list of TypeManagers many times, rather than the other way around.
-                // Certainly it can't be *less* efficient to do it this way.
-                foreach (var type in TypeUtils.GetDefinedTypes(args.LoadedAssembly, logger))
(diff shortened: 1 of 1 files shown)


jason-bragg:
This looks safe given the current usage patterns.  The question this PR raises in my mind is whether the container guarantees single construction of singletons under race conditions.  I would think it should, but I'm unsure.

If the issue persists after this, it may be worth looking into the singleton guarantees declared in the abstraction.  If none are provided, we may want to resolve a Lazy<ProviderTypeLookup>.

While I understand the reasoning for this change, it does seem speculative, as the threads seemed to be blocked in type loading rather than where the locks actually existed, which, to me, would point to a deeper problem, possibly in the runtime.  I'd normally be hesitant to speculate in this direction, but the frequency of recent updates and instability in our tooling have left more room for this suspicion.

sergeybykov:
>While I understand the reasoning for this change, it does seem speculative, as the threads seemed to be blocked in type loading rather than where the locks actually existed, which, to me, would point to a deeper problem, possibly in the runtime.

I don't think it's too speculative. The deadlock I see in the stuck tests I believe is between https://github.com/dotnet/orleans/blob/master/src/Orleans.Core/Providers/ProviderTypeLoader.cs#L74 and https://github.com/dotnet/orleans/blob/master/src/Orleans.Core/Providers/ProviderTypeLoader.cs#L33 that is triggered on a different thread. I've looked at 3 memory dumps with exact same picture.

xiazen:
looks fine to me, the lock seems unnecessary.

xiazen:
> If the issue persists after this, it may be worth looking into the singleton guarantees declared in the abstraction. If none are provided, we may want to resolve a Lazy.

I think jason's comment says he is ok with the change , but was just speculative on how DI handle race conditions. And suggesting to check the code in first, and observe. If we still observe race condition after this change, then we may need to look into how our DI handle race conditions when resolving a singleton.

xiazen:
So I will assume there's a agreement here and merge this pr

jason-bragg:
Yes.  Change looks good, I was simply commenting on my uncertenty that this change will adders the issue.  I'd be happy to be wrong :)
Only reason I didn't merge was that it was still marked WIP, so I wasn't sure if there were more changes on the way..

Issue #3909, opened 2018-01-25
Microsoft.Extensions.Logging issue


tinpl:
Seems that from v2 release of Microsoft.Extensions.Logging the way of providing messages templates to logger.Log* functions became more suitable for structured/semantic logging.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?tabs=aspnetcore2x#log-message-template
However, Orleans still uses the 'string.Format' style of generating messages. In comparison to M.E.L proposed way of naming variables, it just hides the possibility to implement custom log providers to have more intuitive/straightforward extraction of key/value pairs of important variables.
Moreover, in case of line https://github.com/dotnet/orleans/blob/master/src/Orleans.Core/Serialization/SerializationManager.cs#L747  , the old code leads to a bug. Calling this ( case of serializing non-immutable 1dim arrays of primitive types ) throws a `
System.AggregateException: 'An error occurred while writing to logger(s). (Index (zero based) must be greater than or equal to zero and less than the size of the argument list.)'
InnerException	{System.FormatException: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
at System.String.Format(IFormatProvider provider, String format, Object[] args)
at NLog.Extensions.Logging.NLogLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func 3 formatter)
   at Microsoft.Extensions.Logging.Logger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func 3 formatter)}	System.Exception {System.FormatException}`

do not mind the NLog here, had this exceptiong on office pc using regular M.E.L.Console

veikkoeeva:
This could also be  larger issue. One particular thread to dig deeper into actual issues is at https://github.com/aspnet/Logging/issues/612 (and then maybe something like https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage is suitable in some cases). For instance, I'd be interested about how to flow around correlation IDs (and other data).

xiazen:
@tinpl I'm not sure I fully follow you. 
SerializationManager line747 uses `ILogger.Info` which is an extension methods we built on top of ASP.NET logging extension methods:
[...]

PR #3916, opened 2018-01-26
Fix logger exceptions

tinpl:
PR fixing the issue https://github.com/dotnet/orleans/issues/3909
Removed  some logging code seems to be left from testing
Test for big 1-dimensional array of primitive type serialization, showing off the issue

### DIFF
--- a/src/Orleans.Core/Logging/ILoggerExtensions.cs
+++ b/src/Orleans.Core/Logging/ILoggerExtensions.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging;
 using System;
 using System.Diagnostics;
 
@@ -13,7 +13,9 @@ namespace Orleans.Runtime
         /// Writes a log entry at the Debug severity level.
         /// </summary>
         /// <param name="logger">The logger</param>
-        /// <param name="format">A standard format string, suitable for String.Format.</param>
+        /// <param name="format">Format string of the log message with named parameters
+        /// <remarks>Not always suitable for <c>String.Format</c>. See Microsoft.Extensions.Logging MessageTemplate section for more information. Suggest to use their pattern over this extension method</remarks>
+        /// </param>
         /// <param name="args">Any arguments to the format string.</param>
         public static void Debug(this ILogger logger, string format, params object[] args)
         {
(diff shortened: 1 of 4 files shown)


xiazen:
this is great. thanks for finding those string format errors in our code.

src/Orleans.Core/Logging/ILoggerExtensions.cs
```diff
@@ -13,7 +13,9 @@ public static class OrleansLoggerExtension
         /// Writes a log entry at the Debug severity level.
         /// </summary>
         /// <param name="logger">The logger</param>
-        /// <param name="format">A standard format string, suitable for String.Format.</param>
+        /// <param name="format">Format string of the log message with named parameters
+        /// <remarks>Not always suitable for <c>String.Format</c>. See Microsoft.Extensions.Logging usage</remarks>
```
xiazen: can we do: See Microsoft.Extensions.Logging Message Template section. 
just `Microsoft.Extensions.Logging` is too broad, not clear on which documents reader should read

src/Orleans.Runtime/ReminderService/InMemoryRemindersTable.cs
```diff
@@ -116,28 +116,24 @@ public bool RemoveRow(GrainReference grainRef, string reminderName, string eTag)
             // assuming the calling grain executes one call at a time, so no need to lock
             if (!reminderTable.TryGetValue(grainRef, out data))
             {
-                logger.Info("1");
```
xiazen: is the reason removing those that logging messaging is garbage?

tinpl: Seems that it is garbage, it just wast left there after initial commit and is not used anywhere, neither informative. Level of these (removed) log messages (Info) is too quite high, everywhere else in this class Trace is used, so I considered it as just debug info which was left by somebody. Actually, seems that all this class InMemoryRemindersTable, and GrainBasedReminderTable ( the only one which instantiates inmemory reminders ) - are not used anywhere in the codebase, so nobody cares about these 1,2,3,4 messages

src/Orleans.Core/Logging/ILoggerExtensions.cs
tinpl: Thought the same way, however was not capable to come up with something better. Indeed, the option you proposed is more informative.

src/Orleans.Runtime/ReminderService/InMemoryRemindersTable.cs
xiazen: yeah we should remove it .

src/Orleans.Core/Logging/ILoggerExtensions.cs
xiazen: we can merge this after you update the xml. 
Let's do "Not always suitable for String.Format, See Microsoft.Extensions.Logging MessageTemplate sectionfor more information. Suggest to use their pattern over this extension method"

Since these extension method exists really is just to help runtime migrate to MEL. If runtime uses MEL patterns, we can remove these extension methods.

xiazen:

xiazen:

sergeybykov:
Thank you, @tinpl!

PR #3927, opened 2018-01-29
Replaced AzureTableStorage storage provider with AzureTableGrainStorage.

jason-bragg:
Moved the first provider  to the new extension pattern.

Split storage calls and initialization from IStorageProvider so usage patterns are independent of initialization.

IGrainStorage is the new grain storage abstraction.

IStorageProvider is supported for legacy storage providers and inherits from IGrainStorage and IProvider.

All grain storage logic now talks to IGrainStorage.

AzureTableStorage was replaced with AzureTableGrainStorage which no longer uses the provider pattern, but instead uses DI and the silo lifecycle directly, and is configured using options.


ReubenBond:
Glad there are lots of reviewers
image

xiazen:
if you tag a person three time, her productivity will triple

src/Orleans.Core/Providers/IGrainStorage.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// <param name="getExtendedErrors">Whether or not to extract REST error code</param>
+        /// <returns></returns>
+        bool DecodeException(Exception e, out HttpStatusCode httpStatusCode, out string restStatus, bool getExtendedErrors = false);
+    }
+
+    /// <summary>
+    /// Exception thrown whenever a grain call is attempted with a bad / missing storage configuration settings for that grain.
+    /// </summary>
+    [Serializable]
+    public class BadStorageConfigException : OrleansException
```
xiazen: maybe name it `BadGrainStorageConfigurationException` ? storage is too broad, that's why you named `IGrainStorage` to be grain specific I think .

xiazen: And avoid using short names like config, instead of configuration, in public class names

test/ServiceBus.Tests/Streaming/EHStreamProviderCheckpointTests.cs
```diff
@@ -171,7 +171,7 @@ private static void AdjustConfig(ClusterConfiguration config)
         {
             // register stream provider
             config.Globals.RegisterStreamProvider<EventHubStreamProvider>(StreamProviderName, BuildProviderSettings());
-            config.AddAzureTableStorageProvider(ImplicitSubscription_RecoverableStream_CollectorGrain.StorageProviderName);
+            config.AddAzureBlobStorageProvider(ImplicitSubscription_RecoverableStream_CollectorGrain.StorageProviderName);
```
xiazen: you just changed to use Blob storage..

src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorage.cs
```diff
@@ -522,5 +451,54 @@ public bool DecodeException(Exception e, out HttpStatusCode httpStatusCode, out
         {
             return AzureStorageUtils.EvaluateException(e, out httpStatusCode, out restStatus, getRESTErrors);
         }
+
+        private async Task Init(CancellationToken ct)
+        {
+            this.logger.Info((int)AzureProviderErrorCode.AzureTableProvider_InitProvider, $"AzureTableStorage initializing: {this.options.ToString()}");
```
xiazen: suggest to use `ILogger.LogInformation` instead. that's their native method. `ILogger.Info` extension exists mainly to make the migration to MEL easier.

src/Azure/Orleans.Persistence.AzureStorage/Options/AzureTableStorageOptionsFormatterResolver.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        private class AzureTableStorageOptionsFormatter : IOptionFormatter<AzureTableStorageOptions>
+        {
+            public string Category { get; } = "Storage";
+
+            public string Name => nameof(AzureTableStorageOptions);
+
+            private AzureTableStorageOptions options;
+
+            public AzureTableStorageOptionsFormatter(AzureTableStorageOptions options)
```
xiazen: add `name` to this formatter's constructor, and fill in its `Name` property to include the `name`. Since this is a named formatter, we need some identifier to tell user which named option it is printing.

xiazen:
why do you moved `SiloLifecycleStage` from runtime to abstraction? It should be in runtime right, since it is only silo side code

src/Orleans.Core/Providers/IGrainStorage.cs
jason-bragg: This was moved from IStorageProvider.  It's a preexisting exception.  If we change the name it could break application code's error handling.  Not 100% against changing the name, but don't see much advantage in it.

src/Azure/Orleans.Persistence.AzureStorage/Hosting/SiloBuilderExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Orleans.Hosting
+{
+    public static class SiloBuilderExtensions
+    {
+        /// <summary>
+        /// Configure silo to use an azure table storage for grain storage.
+        /// </summary>
+        public static ISiloHostBuilder AddAzureTableGrainStorage(this ISiloHostBuilder builder, Action<AzureTableStorageOptions> configureOptions)
```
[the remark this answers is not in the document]
jason-bragg: AddAzureTableGrainStorageAsDefault ?

jason-bragg:
On a side note.. I didn't add everyone 3 times, I added each once, and GitHub seemed to think that wasn't enough :)

jason-bragg:
Regarding SiloLifecycleStage, providers can wire themselves up to lifecycle without knowing the stages, as those can be set at configuration time, where the stages can be known, but if we want providers to be able to set reasonable defaults, they'll need to know about the stages.

jason-bragg:
Broke json support.  Fixing, Marked wip until this is addressed.

jason-bragg:
Fixed json serialization, removed 'work in progress' marker.  Ready for review.

src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorageOptions.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        /// Table name where grain stage is stored
+        /// </summary>
+        public string TableName { get; set; } = DEFAULT_TABLE_NAME;
+        public const string DEFAULT_TABLE_NAME = "OrleansGrainState";
+
+        /// <summary>
+        /// Indicates if grain data should be deleted or reset to defaults when a grain clears it's state.
+        /// </summary>
+        public bool DeleteStateOnClear { get; set; } = false;
```
xiazen: I think we should document the default value. something like "off by default"

src/Orleans.Core/Core/StateStorageBridge.cs
```diff
@@ -27,12 +28,14 @@ public string Etag
             get { return grainState.ETag; }
         }
 
-        public StateStorageBridge(string name, GrainReference grainRef, IStorageProvider store, ILogger logger)
+        public StateStorageBridge(string name, GrainReference grainRef, IGrainStorage store, ILoggerFactory loggerFactory)
```
xiazen: do we need a `LoggerFactory` or just a logger?

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -106,31 +108,43 @@ public ProviderLifecycleParticipant(IProviderConfiguration config, IServiceProvi
...
-                lifecycle.Subscribe(this.startStage, this.Start);
+                lifecycle.Subscribe(this.startStage, this.Start, this.StreamProviderClose);
 
             }
 
             private async Task Init(CancellationToken ct)
             {
                 var stopWatch = Stopwatch.StartNew();
-                IProvider provider = this.services.GetServiceByName<TService>(this.config.Name) as IProvider;
+                IProvider provider = this.provider.Value;
```
xiazen: what's the reason of lazy initialization?

src/Orleans.Core/Providers/IStorageProvider.cs
```diff
@@ -1,60 +1,21 @@
...
-        /// <returns></returns>
-        bool DecodeException(Exception e, out HttpStatusCode httpStatusCode, out string restStatus, bool getExtendedErrors = false);
     }
 
     /// <summary>
     /// Exception thrown whenever a grain call is attempted with a bad / missing storage provider configuration settings for that grain.
     /// </summary>
     [Serializable]
-    public class BadProviderConfigException : OrleansException
+    public class BadProviderConfigException : BadStorageConfigException
```
ReubenBond: This hierarchy seems inverted

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
jason-bragg: Can't resolve concrete types at construction time because parts loader may not have loaded assemblies yet.

src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorageOptions.cs
jason-bragg: As a general pattern, we don't tend to document defaults if they are the same as the default for the type.  The default for a bool is false, so assumed default for the type, given no stated default, is correct.

Not opposed to this suggestion, but we should be consistent.

src/Orleans.Core/Providers/IGrainStorage.cs
jason-bragg: My mistake, this was added to communicate grain storage issues, I got confused. :/
Good catch, tnx! :)

test/ServiceBus.Tests/Streaming/EHStreamProviderCheckpointTests.cs
jason-bragg: Yep.  I did this because these tests don't really care about what storage they use, and it was easier than adding a builder just to use azure table storage.  We'll eventually need to add a builder when we move other providers to the new model, but I didn't see the need to address that in this PR.

Anyway, I think we need more tests to use blob storage anyway.  Most of our tests use table storage.

src/Orleans.Core/Core/StateStorageBridge.cs
jason-bragg: Oh, we need a logger factory because we want the logging associated with the provider type, which we can't resolve in the constructor.

ReubenBond:
@dotnet-bot test bvt

Issue #3933, opened 2018-01-30
FastKillOnCancelKeyPress waits forever


jms69:
When FastKillOnCancelKeyPress=true, SystemStatus is set to stopping but nothing signals the LifeCycle to stop.

The Stop() method then wait for the SystemStatus to Terminate but it never does.

Working on a local fix.

jms69:
Pull request https://github.com/dotnet/orleans/pull/3935.

benjaminpetit:
Resolved with #3935

PR #3935, opened 2018-01-30
Fix FastKillOnCancelKeyPress not stopping the process.

jms69:
When FastKillOnCancelKeyPress=true State is set to Closing but Stop process never starts.


jms69:
Issue https://github.com/dotnet/orleans/issues/3933

src/Azure/Orleans.Clustering.AzureStorage/LegacyAzureMembershipConfigurator.cs
```diff
@@ -16,6 +17,9 @@ public void ConfigureServices(object configuration, IServiceCollection services)
                     options.MaxStorageBusyRetries = reader.GetPropertyValue<int>("MaxStorageBusyRetries");
                     options.ConnectionString = reader.GetPropertyValue<string>("DataConnectionString");
                 });
+
+            services.AddTransient<IConfigurationValidator>(sp => new AzureTableMembershipConfigurationValidator(configuration));
```
ReubenBond: This and the above IConfigurationValidator change seem orthogonal to this PR - maybe they should be moved into a separate PR

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -811,15 +811,22 @@ private void HandleProcessExit(object sender, EventArgs e)
...
             lock (lockable)
             {
                 if (!this.SystemStatus.Equals(SystemStatus.Running)) return;
-                    
+
                 this.SystemStatus = SystemStatus.Stopping;
+
+                // force a non-graceful stop
+                cancellationSource.Cancel();
+                this.siloLifecycle.OnStop(cancellationSource.Token);  // don't wait for it to stop
```
ReubenBond: Why this `OnStop` call here when `OnStop` will be called via `StopAsync` below?

ReubenBond:
It seems to me that the correct fix here is to *not* set `SystemStatus` directly at all and just call `Stop()` or `StopAsync(cancellation)`, possibly with a cancelled token to abort it - but my leaning is towards graceful shutdown on Ctrl+C

galvesribeiro:
I'm having the same issue you described @ReubenBond

src/Azure/Orleans.Clustering.AzureStorage/LegacyAzureMembershipConfigurator.cs
jms69: Sorry I am new to this.  I thought checking this into my fork would not affect this pull request.  This issue is something else.  When ClusterId is not set you get a Azure Storage Error "Bad Request".  I was had to figure out what was wrong.  This check-in was meant for my branch.

src/Orleans.Runtime/Silo/Silo.cs
jms69: Because the OnStop after call will get into the mode that just waits for State to Terminate, but nothing signals the lifecycle to stop.

jms69:
The way I worked around the issue in my console app is to turn FastKillOnCancelKeyPress=false and handle the cancel myself.

Console.CancelKeyPress += Console_CancelKeyPress;


        private void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
        {
            Console.WriteLine("*****************Shutdown Started********************");
            cancelToken.Cancel();
            silo.StopAsync(cancelToken.Token);            
        }

ReubenBond:
@jms69 that's how I think it should be done. We should remove the option and allow end-users to configure it themselves.

Particularly since `Microsoft.Extension.Hosting` will have an abstraction for controlling lifetime of a collection of hosted services.

src/Azure/Orleans.Clustering.AzureStorage/LegacyAzureMembershipConfigurator.cs
ReubenBond: No need to apologize :)
You can revert those changes in this branch and they'll disappear from here.

If you'd like assistance, let me know.

jms69:
It's still not working for me in Kubernetes.  When the container is killed its still waiting until the process is forcibly killed (my timeout was set to 2 min).  I believe this line in Orleans is short nullifying my handler.

AppDomain.CurrentDomain.ProcessExit += HandleProcessExit;

ReubenBond:
@jms69 I submitted a PR against your fork: https://github.com/jms69/orleans/pull/1

If it works for you then you can merge it and *this* PR will automatically update

jms69:
I merged the pull request and do some more  testing.  I am trying to get scale down to work in AKS Kubernetes cluster.

As you mentioned earlier it would be cool if we can provide our own cancel action in case we wanted to do some extra shutdown code:

builder.ConfigurationCancelAction(Action<ISiloHost> action);

jms69:
Scale up/down in Kubernetes now works perfectly.

ReubenBond:
@jms69 
>it would be cool if we can provide our own cancel action in case we wanted to do some extra shutdown code:
builder.ConfigurationCancelAction(Action action);

That's the idea behind the [Generic Host Builder](https://github.com/aspnet/Hosting/issues/1163) which will be released alongside .NET Core 2.1. Once that's available we will work to strip away this functionality and rely on that.

ReubenBond:
I'm hoping we can have one more set of eyes on this. I'm happy with it. Please merge if you're happy with it.

sergeybykov:
Thank you, @jms69!

benjaminpetit:
From what I understand, here you are doing a clean shutdown, not a fast kill like before (and as the name in the config suggest). I think by default, we should not attempt to do a clean shutdown.

Or at the very least, rename the current config flag

ReubenBond:
Good point, the config flag name implies abrupt termination - which I would imagine looks like `Environment.FaskKill()`. Either way, I'm happy for it to be changed to `this.Stop()` instead `this.Shutdown()`.

Longer term, Ctrl+C should at least attempt a clean shutdown the first time it's pressed, but this is a concern which we should shift to the generic host.

PR #3943, opened 2018-02-01
Fix tests. Remove ServiceId and ClusterId getter, turn on file logging

xiazen:
- fix some tests
- remove `ServiceId` and `ClusterId` getter from `TestCluster`, since they are misleading. They only reflect values you set using `TestClusterOptions`, may not reflect values if you configure cluster using legacy `ClusterConfiguration`. 
- turn on file logger for TestCluster by default

### DIFF
--- a/src/Orleans.TestingHost/TestCluster.cs
+++ b/src/Orleans.TestingHost/TestCluster.cs
@@ -28,21 +28,26 @@ namespace Orleans.TestingHost
     /// </remarks>
     public class TestCluster
     {
+        private readonly List<SiloHandle> additionalSilos = new List<SiloHandle>();
+        private readonly TestClusterOptions options;
+        private readonly StringBuilder log = new StringBuilder();
+
         private int startedInstances;
 
         /// <summary>
-        /// Primary silo handle
+        /// Primary silo handle, if applicable.
         /// </summary>
-        /// <remarks>This only provide valid data if your cluster are using GrainBasedMemebership</remarks>
+        /// <remarks>This handle is valid only when using Grain-based membership.</remarks>
         public SiloHandle Primary { get; private set; }
 
         /// <summary>
-        /// List of handles to the secondary silos
+        /// List of handles to the secondary silos.
         /// </summary>
         public IReadOnlyList<SiloHandle> SecondarySilos => this.additionalSilos;
 
-        private readonly List<SiloHandle> additionalSilos = new List<SiloHandle>();
-
+        /// <summary>
+        /// Collection of all known silos.
+        /// </summary>
         public ReadOnlyCollection<SiloHandle> Silos
         {
             get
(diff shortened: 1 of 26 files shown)


src/Orleans.TestingHost/TestClusterBuilder.cs
```diff
@@ -101,7 +101,7 @@ public TestCluster Build()
             return testCluster;
         }
 
-        private static string CreateClusterId()
+        public static string CreateClusterId()
```
xiazen: make this public in case user want to configure test cluster id explicitly, through `ClusterConfiguration`

src/Orleans.TestingHost/TestClusterHostFactory.cs
```diff
@@ -227,12 +227,8 @@ private static void TryConfigureTestClusterMembership(IConfiguration configurati
 
         private static void TryConfigureFileLogging(IConfiguration configuration, IServiceCollection services, string name)
         {
-            bool.TryParse(configuration["UseTestClusterMemebership"], out bool configureFileLogging);
```
xiazen: not sure why turn on file logging has to depend on `UseTestClusterMemebership`. Seems unrelated. so turn it on by default now.

xiazen:
merged commits from @ReubenBond 's branch for typo fix and use `nameof`

src/Orleans.TestingHost/TestClusterHostFactory.cs
jdom: I think it was probably a bad copy-paste. this should do `TryParse(configuration[nameof(TestClusterOptions.ConfigureFileLogging)], ...`

---
In reply to: [165226223](https://github.com/dotnet/orleans/pull/3943#discussion_r165226223) [](ancestors = 165226223)

xiazen:
Do not merge yet. want to run it in vso

src/Orleans.TestingHost/TestCluster.cs
```diff
@@ -33,6 +33,7 @@ public class TestCluster
         /// <summary>
         /// Primary silo handle
         /// </summary>
+        /// <remarks>This only provide valid data if your cluster are using GrainBasedMemebership</remarks>
```
ReubenBond: `This handle is valid only when using Grain-based membership`

ReubenBond:
Will merge when VSO CI completes

jdom:
@ReubenBond FYI, Xiao is going to address the ConfigureFileLogging comment

ReubenBond:
ok, we can leave this until tomorrow then

ReubenBond:
@dotnet-bot test functional

xiazen:
run a vso build on this : 
https://orleans.visualstudio.com/Orleans/_build/index?buildId=7848&_a=summary&tab=ms.vss-test-web.test-result-details

there's still remaining test failure related to 

>  Lifecycle start canceled due to errors at stage 4000\r\n---- System.TimeoutException : WithTimeout has timed out after 00:05:00.

We can fix those bugs in another pr. this pr fixed some tests failures and did improvement on TestCluster. It can be merged by itself.

jdom:
I agree, they can be fixed separately. BTW, we should make that exception clearer, not just fix the tests so that it doesn't happen.
That exception is extremely cryptic, and you can't know what went wrong during the lifecycle calls.

jdom:
BTW, these are new and related failures (to the changes in this PR), right? https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/1115/

PR #3946, opened 2018-02-01
Remove reflection usage from hot path

dVakulen:
`IsStatelessWorker` check was iterating through type attributes.


sergeybykov:
Many thanks, @dVakulen!

PR #3947, opened 2018-02-02
More options formatting, and related cleanup

jason-bragg:
Added formatters for grain directory options, grain placement options, schedualling options, thread pool options, type managment options, versioning options and silo options.
Fixed formatting of messaging options.
Fixed some options defaults.

### DIFF
--- a/src/Orleans.Core/Configuration/Options/MessagingOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/MessagingOptions.cs
@@ -63,7 +63,7 @@ namespace Orleans.Hosting
             this.options = options;
         }
 
-        protected List<string> FormatStatisticsOptions()
+        protected List<string> FormatSharedOptions()
         {
             return new List<string>()
             {
--- a/src/Orleans.Core/Configuration/Options/MessagingOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/MessagingOptions.cs
@@ -1,7 +1,6 @@
 using Orleans.Runtime;
 using System;
 using System.Collections.Generic;
-using Microsoft.Extensions.Options;
 
 namespace Orleans.Hosting
 {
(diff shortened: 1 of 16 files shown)


src/Orleans.Core/Configuration/Options/MessagingOptions.cs
```diff
@@ -54,4 +53,30 @@ public abstract class MessagingOptions
...
+    public abstract class MessagingOptionsFormatter
+    {
+        private MessagingOptions options;
+
+        protected MessagingOptionsFormatter(MessagingOptions options)
+        {
+            this.options = options;
+        }
+
+        protected List<string> FormatStatisticsOptions()
```
xiazen: naming: FormatMessagingOptions

src/Orleans.Runtime.Abstractions/Silo/SiloOptions.cs
```diff
@@ -25,4 +27,28 @@ public class SiloOptions
         public bool FastKillOnCancelKeyPress { get; set; } = DEFAULT_FAST_KILL_ON_CANCEL;
```
jdom: not sure when this was added, but I wouldn't put this in Orleans' specific SiloOptions. We didn't copy everything related to IHost but the IHostLifetime already has things such as these when you use the ConsoleHostLifetime way of hosting... Not suggesting you copy everything related to that (or maybe you do want to), but at least keep it isolated, since it will go away when using Microsoft.Extensions.Hosting

ReubenBond: @jdom it should go away, but we're not using Microsoft.Extensions.Hosting yet. I see that we were both against having any kind of Ctrl+C handling in the first place: https://github.com/dotnet/orleans/pull/3109

Hopefully we can remove it soon.

jason-bragg: Added to https://github.com/dotnet/orleans/issues/3957

PR #3948, opened 2018-02-02
Move Silo-specific options from Orleans.Core to Orleans.Runtime

sergeybykov:
There no need for silo-specific options to pollute `Orleans.Core`.

While it is somewhat ugly to duplicate some default values in the legacy config classes, I think that's an acceptable price to pay for cleanness of the forward looking APIs. The legacy classes are ... legacy anyway, and their defaults are unlikely to change anyway.

Moved the following options classes to Orleans.Runtime:

* SiloMessagingOptions
* SiloStatisticsOptions
* ConsistentRingOptions
* GrainCollectionOptions
* GrainDirectoryOptions
* GrainPlacementOptions
* GrainServiceOptions
* ReminderOptions
* SchedulingOptions

Also moved OrleansConfiguration.xsd to Orleans.Core.Legacy.


src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -629,6 +629,13 @@ private void ResendMessageImpl(Message message, ActivationAddress forwardingAddr
             }
         }
 
+        // Forwardings are used by the receiver, usualy when it cannot process the message and forwars it to another silo to perform the processing
```
ReubenBond: nit: typo here

jason-bragg:
https://github.com/dotnet/orleans/issues/3956

xiazen:
there is a discussion on where should options stay. One idea is options are part of the framework, so it should stay in core.  And there's another idea that option should stay with its component. That was one of the discussion carried away from the cluster configuration pr, and added to todos in #3956 . Suggest we discuss which pattern we should follow first and then move the option to its proper places.

sergeybykov:
Orleans.Core is meant to be shared by client and silo. Hence, I don't think we want silo specific types there.

jason-bragg:
In general, I'm of the view that a component's options should live in the library where the implementation of the component that uses it lives.  So If the serialization manager implementation lives in Core, so should any options related to it, whereas if the placement manager implementation lives in Runtime, so should it's options.  For pluggable components like the Azure reminder table, it's options should live in the extension assembly which implements that reminder system.

Having said that, it may be non-trivial to get there from here, but I'd prefer we get as close as is practicable.

xiazen:
lol how many option you plan to move? I think you should update the pr description , in case people just merged this pr, without realizing new commit messages which haven't been reviewed yet. 
If you are still working on it, maybe mark it as work in progress

sergeybykov:
Updated description. This can be reviewed/merged now.

sergeybykov:
@dotnet-bot test this please

src/OrleansProviders/OrleansProviders.csproj
```diff
@@ -14,5 +14,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Orleans.Core.Abstractions\Orleans.Core.Abstractions.csproj" />
     <ProjectReference Include="..\Orleans.Core\Orleans.Core.csproj" />
+    <ProjectReference Include="..\Orleans.Runtime\Orleans.Runtime.csproj" />
```
xiazen: I thought OrleansProvider shouldn't have an dependency on Orleans runtime.

src/Orleans.Core.Legacy/Configuration/NodeConfiguration.cs
```diff
@@ -204,22 +204,22 @@ public NodeConfiguration()
...
 
-            this.MaxActiveThreads = SchedulingOptions.DEFAULT_MAX_ACTIVE_THREADS;
-            this.DelayWarningThreshold = SchedulingOptions.DEFAULT_DELAY_WARNING_THRESHOLD;
-            this.ActivationSchedulingQuantum = SchedulingOptions.DEFAULT_ACTIVATION_SCHEDULING_QUANTUM;
-            this.TurnWarningLengthThreshold = SchedulingOptions.DEFAULT_TURN_WARNING_THRESHOLD;
-            this.EnableWorkerThreadInjection = SchedulingOptions.DEFAULT_ENABLE_WORKER_THREAD_INJECTION;
+            this.MaxActiveThreads = Math.Max(4, Environment.ProcessorCount);
+            this.DelayWarningThreshold = TimeSpan.FromMilliseconds(10000);
+            this.ActivationSchedulingQuantum = TimeSpan.FromMilliseconds(100);
+            this.TurnWarningLengthThreshold = TimeSpan.FromMilliseconds(200);
```
xiazen: generally speaking we don't use random numbers. Can you add those numbers as DEFAULXXX in this file so the code will be more readable?

src/OrleansProviders/OrleansProviders.csproj
xiazen: It would make Orleans.Client has a transient dependency on Orleans.Runtime.

sergeybykov: Good catch. Looks like we need a separate change to avoid this dependency that would have to go in before this one.

sergeybykov:
Addressed feedback.

PR #3962, opened 2018-02-02
Add exception msg to WithTimeout method. Refactor MembershipTableFactory

xiazen:
- Add exception msg to WithTimeout method
- refactor MembershipTableFac


src/Orleans.Runtime/MembershipService/MembershipTableFactory.cs
```diff
@@ -7,38 +7,18 @@ namespace Orleans.Runtime.MembershipService
...
         {
-            this.serviceProvider = serviceProvider;
+            this.membershipTable = membershipTable;
             this.logger = logger;
         }
 
-        internal async Task<IMembershipTable> GetMembershipTable()
+        internal IMembershipTable GetMembershipTable()
         {
-            if (membershipTable != null) return membershipTable;
```
xiazen: please take a look of this code. To me the lock and all those null handling seems unnecessary since membership is not injected from DI and should only be initialized in `MembershipOracle.Start` method. 

ManagementGrain calls `GetMembershipTable` for hosts information. But it shouldn't call it before it is initialized nor initialize it before `MembershipOracle.Start`. So I think this change is safe. But do take a look in case I missed anything.

jason-bragg: It seems that the previous pattern always relied on the factory to create and initialize the membership table, whereas now the factory can create and return an uninitialized table.  While this may work in the current call patterns it seems fragile.  Am I misunderstanding what I'm seeing?

xiazen: fragile as in caller for GetMembershipTable may get one instance which hasn't been initialized yet? yes it is true. 
Or am I missing other fragile points?

xiazen: But I think the previous pattern is fragile too. So any caller calls GetMembershipTable will initialize the table. But to initialize the table,  it need certain dependency to be initialized too. Previous pattern doesn't handle this well. 

New pattern doesn't initialize the table in GetMembershipTable() call, it holds the risk of returning a uninitialized table. If this uninitialized table get called , then the table implementation should throw UnInitializedException or some sort. I think this should be the correct behavior. Do you disagree?

jason-bragg: There are many things wrong with the current pattern.
Factory
* A factory should create instances in a usable form, initializing it if it's necessary
* In general factories should not cache instances of what they create.  Caches or singletons should do that.

Basically this is backwards, the factory does not create the membership table, it gets it from the container and initializes it, so it's not a factory at all..?

I don't think we need or should use a factory in this case.  I think membership tables should register in the container as an IMembershipTable and as a lifecycle participant and initialize themselves in the startup process.  If they are called prior to initialization, they should throw an error.

If anything calls the membership service prior to it's initialization, it would be due to misconfiguration, and users could configure the membership to start at an earlyer stage, or start dependent systems at later stages.

Open to other patterns, but this PR seems to just exchange a questionable approach for a different questionable approach, which admitadly is far less work than what I'd suggested above.

xiazen: offline discussion agreed to remove the factory since it is useless at this point.

ReubenBond: sounds good to me, the factory was a crutch. Just make sure that everything still works when not using *any* `IMembershipTable` implementation and replacing the `IMembershipOracle` implementation instead (Service Fabric case)

src/Orleans.Core/Async/TaskExtensions.cs
```diff
@@ -198,7 +198,7 @@ internal static T WaitForResultWithThrow<T>(this Task<T> task, TimeSpan timeout)
         /// <param name="timeout">Amount of time to wait before timing out</param>
         /// <exception cref="TimeoutException">If we time out we will get this exception</exception>
         /// <returns>The completed task</returns>
-        internal static async Task WithTimeout(this Task taskToComplete, TimeSpan timeout)
+        public static async Task WithTimeout(this Task taskToComplete, TimeSpan timeout, string exceptionMessage = null)
```
ReubenBond: Does this need to be public?

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -483,22 +483,22 @@ await scheduler.QueueTask(transactionAgent.Start, transactionAgentContext)
...
-                    .WithTimeout(initTimeout, $"MembershipOracle.BecomeAction() failed due to timeout {initTimeout}");
+                    .WithTimeout(initTimeout, $"MembershipOracle activating failed due to timeout {initTimeout}");
                 logger.Debug("Local silo status oracle became active successfully.");
             }
 
             var versionStore = Services.GetService<IVersionStore>();
             await StartAsyncTaskWithPerfAnalysis("Init type manager", () => scheduler
                 .QueueTask(() => this.typeManager.Initialize(versionStore), this.typeManager.SchedulingContext)
-                .WithTimeout(this.initTimeout, $"TypeManager.Initialize failed due to timeout{initTimeout}"), stopWatch);
+                .WithTimeout(this.initTimeout, $"TypeManager Initializing failed due to timeout{initTimeout}"), stopWatch);
```
ReubenBond: nit: missing space after "timeout"

src/Orleans.Core/Async/TaskExtensions.cs
xiazen: yeah. There's an identical one in Orleans.TestingHost. to make function less ambiguous, I made this public and removed that one. And `WithTimeout<T>` is already public, so I think it is fine to make this public,

src/Orleans.Runtime/Silo/Silo.cs
xiazen: :< Xiao's typo is everywhere

src/Orleans.Core/Async/TaskExtensions.cs
ReubenBond: Ok, seems reasonable. Maybe they should both be internal, but at least this is more consistent now.

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -233,7 +233,7 @@ await this.gatewayListProvider.InitializeGatewayListProvider()
             grainTypeResolver = await transport.GetGrainTypeResolver(this.InternalGrainFactory);
 
             await ClientStatistics.Start(transport, clientId)
-                .WithTimeout(initTimeout);
+                .WithTimeout(initTimeout, $"clientStatistics.Start failed due to timeout {initTimeout}");
```
ReubenBond: This one is still unchanged

PR #3963, opened 2018-02-03
Move ClientConfiguration to legacy 

xiazen:
- move `ClientConfiguration` to legacy
- move extension method which depends on `ClientConfiguration` to legacy
- created `GatewayOptions` for `GatewayManager`
- created `MonitoringStorageOptions` for silo and client metrics and statistic table


xiazen:
@ReubenBond  sorry I removed the coding style file. We can added it back after 2.0

ReubenBond:
> @ReubenBond sorry I removed the coding style file. We can added it back after 2.0

src/Azure/Orleans.Statistics.AzureStorage/Storage/ClientMetricsTableDataManager.cs
```diff
@@ -66,21 +67,22 @@ internal class ClientMetricsTableDataManager : IClientMetricsDataPublisher
         private readonly ILogger logger;
         private readonly ILoggerFactory loggerFactory;
         private static readonly TimeSpan initTimeout = AzureTableDefaultPolicies.TableCreationTimeout;
-
-        public ClientMetricsTableDataManager(ILoggerFactory loggerFactory, IOptions<ClusterClientOptions> clusterClientOptions)
+        private StorageOptions storageOptions;
```
jason-bragg: Storage options have only been used on the cluster side previously.  I don't think they are being configured from the client config on the client side.

xiazen:
added `MonitoringStorageOptions` for metric and statistic table. for following reason 

- grain storage and monitoring storage are two different feature, so should use different options
- in client side, we cannot use `StorageOption` , since `ClientConfiguration` only have one `DataConnectionString`, it doesn't have a concept for `DataConnectionstringForReminders`. so it doesn't really make sense for client to configure a StorageOptions

src/AdoNet/Orleans.Statistics.AdoNet/Storage/SqlStatisticsPublisher.cs
```diff
@@ -103,9 +104,9 @@ public void AddConfiguration(string clusterId, bool silo, string siloId, SiloAdd
             }
         }
 
-        async Task IClientMetricsDataPublisher.Init(ClientConfiguration config, IPAddress address, string clientId)
+        Task IClientMetricsDataPublisher.Init(IPAddress address, string clientId)
         {
-            orleansQueries = await RelationalOrleansQueries.CreateInstance(config.AdoInvariant, config.DataConnectionString, this.grainReferenceConverter);
+            return Task.CompletedTask;
```
ReubenBond: This still works?

src/Orleans.Core/Configuration/Options/ClientMessagingOptions.cs
```diff
@@ -37,6 +46,8 @@ public IEnumerable<string> Format()
             format.AddRange(new List<string>
             {
                 OptionFormattingUtilities.Format(nameof(options.ClientSenderBuckets), options.ClientSenderBuckets),
+                OptionFormattingUtilities.Format(nameof(options.PreferredFamily), options.PreferredFamily),
+                OptionFormattingUtilities.Format(nameof(options.NetInterface), options.NetInterface),
```
ReubenBond: Maybe this should be `InterfaceName` or `NetworkInterfaceName`

src/Orleans.Core/Configuration/Options/GatewayOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// </summary>
+        public static readonly TimeSpan DEFAULT_GATEWAY_LIST_REFRESH_PERIOD = TimeSpan.FromMinutes(1);
+        /// <summary>
+        /// Gateway list refresh period
+        /// </summary>
+        public TimeSpan GatewayListRefreshPeriod { get; set; } = DEFAULT_GATEWAY_LIST_REFRESH_PERIOD;
+        /// <summary>
+        /// Prefered gateway index
+        /// </summary>
+        public int PreferedGatewayIndex { get; set; } = -1;
```
ReubenBond: The special value of -1 should be documented, like: `Preferred gateway index or -1 to prefer no gateway`.

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -125,34 +125,28 @@ internal void ConsumeServices(IServiceProvider services)
...
-                responseTimeout = Debugger.IsAttached ? Constants.DEFAULT_RESPONSE_TIMEOUT : config.ResponseTimeout;
-                this.localAddress = ConfigUtilities.GetLocalIPAddress(config.PreferredFamily, config.NetInterface);
+                responseTimeout = Debugger.IsAttached ? Constants.DEFAULT_RESPONSE_TIMEOUT : this.clientMessagingOptions.ResponseTimeout;
+                this.localAddress = ConfigUtilities.GetLocalIPAddress(this.clientMessagingOptions.PreferredFamily, this.clientMessagingOptions.NetInterface);
 
                 // Client init / sign-on message
                 logger.Info(ErrorCode.ClientInitializing, string.Format(
                     "{0} Initializing OutsideRuntimeClient on {1} at {2} Client Id = {3} {0}",
-                    BARS, config.DNSHostName, localAddress, handshakeClientId));
+                    BARS, Dns.GetHostName(), localAddress, handshakeClientId));
```
ReubenBond: Is there a reason why we might want hostname to be configurable on the client? I'm not sure, but this is ok for now

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
```diff
@@ -56,18 +59,18 @@ internal class ClientStatisticsManager : IDisposable
...
-                        this.clusterClientOptions.ClusterId, config.DNSHostName, clientId.ToString(), transport.MyAddress.Endpoint.Address);
+                        this.clusterClientOptions.ClusterId, dnsHostName, clientId.ToString(), transport.MyAddress.Endpoint.Address);
                 }
                 tableStatistics = new ClientTableStatistics(transport, metricsDataPublisher, this.hostEnvironmentStatistics, this.appEnvironmentStatistics, this.loggerFactory)
                 {
                     MetricsTableWriteInterval = statisticsOptions.MetricsTableWriteInterval
                 };
             }
-            else if (config.UseAzureSystemStore)
+            else if (CanUseAzureTable())
```
ReubenBond: Can we avoid this altogether?

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
```diff
@@ -83,17 +86,27 @@ internal class ClientStatisticsManager : IDisposable
                     logStatistics.StatsTablePublisher = statsProvider;
                     // Note: Provider has already been Init-ialized above.
```
ReubenBond: I couldn't spot where it's initialized above

src/AdoNet/Orleans.Statistics.AdoNet/Storage/SqlStatisticsPublisher.cs
xiazen: it will be initialized as a `IProvider`. in client you will need to configure SqlStatisticsPublisher as a `IProvider`

xiazen: see its `IProvider.Init` method

ReubenBond: So previously it was being initialized twice?

xiazen: Yeah I think you are partly right. In the new way, user will need to configure `SqlStatisticsPublisher` as a `IProvider` and register it with client lifecycle 

But in legacy way, we need to add extra logic in `LegacyProviderConfigurator.ConfigureServices` to register `Statistic` provider into client life cycle.

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
xiazen: the xml on `ClientConfiguration.DNSHostName` is 

/// <summary>Gets the true host name, no IP address. It equals Dns.GetHostName()</summary>

So I think this change is safe.

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
xiazen: I'd love to. But I think refactor statistic system is out of the scope of this PR. We can refactor it in another pr , with refactoring silo statistics too.

src/Orleans.Core/Configuration/Options/ClientMessagingOptions.cs
xiazen: This is the name it was originally called, which is the name user are mostly familiar with. This is why I kept it this way. 
But I don't have strong opinion on both. @jason-bragg what do you think ?

src/AdoNet/Orleans.Statistics.AdoNet/Storage/SqlStatisticsPublisher.cs
xiazen: wait , I think in the legacy way, we are registering provider into client lifecycle as well correctly. So I think we are good here. @jason-bragg  can you take a look? I am not that familiar with the provider code .

xiazen: yes I think it was being initialized twice before.

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
xiazen: it is not above anymore, IStatisticsPublisher are now being initialized as a `IProvider`. Take a look at `LegacyProviderConfigurator.COnfigureServices` to see how it is configured using the legacy way. 

In the new way, user will need to configure it with DI as  `IProvider` and then register it into lifecycle.

xiazen: but the xml is probably wrong at this point.

src/Orleans.Core/Configuration/Options/ClientMessagingOptions.cs
xiazen: at a second thought, I think you are right. This is the new option class, should have a name which better represents its intention.

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
ReubenBond: This feels hacky, but as long as there's an issue opened for refactoring it before 2.0

xiazen: yeah. I will open one if there isn't one.

xiazen: it was hacky before ;) this pr didn't made it worse. yeah But we should refactor it.

xiazen: opened an issue in #3976

ReubenBond:
Need to fix the 4 failing BVTs and then we can merge

sergeybykov:
@dotnet-bot test functional

xiazen:
oh thanks for submitting the pr to add it back! we just forgot to added it back after 2.0 .

PR #3964, opened 2018-02-03
Sort list of silos in HashBasedPlacementDirector

sergeybykov:
Sort list of available compatible silos before picking a target one based on the grain ID hash.
Fixes non-deterministic behavior of HashBasedPlacementDirector and related tests.

### DIFF
--- a/src/Orleans.Runtime/Placement/HashBasedPlacementDirector.cs
+++ b/src/Orleans.Runtime/Placement/HashBasedPlacementDirector.cs
@@ -8,7 +8,7 @@ namespace Orleans.Runtime.Placement
         public virtual Task<SiloAddress> OnAddActivation(
             PlacementStrategy strategy, PlacementTarget target, IPlacementContext context)
         {
-            var allSilos = context.GetCompatibleSilos(target).OrderBy(s => s).ToArray();
+            var allSilos = context.GetCompatibleSilos(target).OrderBy(s => s).ToArray(); // need to sort the list, so that the outcome is deterministic
             int hash = (int) (target.GrainIdentity.GetUniformHashCode() & 0x7fffffff); // reset highest order bit to avoid negative ints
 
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Placement/HashBasedPlacementDirector.cs
```diff
@@ -8,11 +8,10 @@ internal class HashBasedPlacementDirector : IPlacementDirector
         public virtual Task<SiloAddress> OnAddActivation(
             PlacementStrategy strategy, PlacementTarget target, IPlacementContext context)
         {
-            var allSilos = context.GetCompatibleSilos(target);
-
+            var allSilos = context.GetCompatibleSilos(target).OrderBy(s => s).ToArray();
```
veikkoeeva: Would a comment such as "Silos are ordered to make logging of information mor deterministic and subsequently troubleshooting easier". If there is a functional dependency, such as tests failing underministically, mention that also?

It often happens there are "hidden dependencies" like this could be that are not visible on the API surface, but a comment might help subsequent readers to see them and perhaps even think ways to refactor and expose dependencies. Also there is often a tendency for some people to order "just in case" and other people to remove orderings as useless. :)

sergeybykov: It is tempting to presort the list of silos returned by `GetCompatibleSilos`, but that would be incur a cost on all call to it, whether sorting is needed or not. So it seems to me, as fragile as it is, it should be a responsibility of the callers to sort when they care.

benjaminpetit: I don't think @veikkoeeva wanted to move this sort to `GetCompatibleSilos`? But I agree with him, you should have an explicit comment here. It's free! :)

sergeybykov:
Sorry, I missed the comment part. Added it now.

Issue #3929, opened 2018-01-30
TestMultiClusterConf_3_3 fails intermittently


sergeybykov:
TestMultiClusterConf_3_3 has been failing intermittently for some time now. I submitted #3928 to skip it for now. This issue is to remember to investigate it.

https://orleans.visualstudio.com/Orleans/Orleans%20Team/_TestManagement/Runs#_a=resultSummary&runId=3414&resultId=101084
>Assert.Equal() Failure\r\n ↓ (pos 10)\r\nExpected: 21100,21101\r\nActual: 21100,21102\r\n ↑ (pos 10)

https://orleans.visualstudio.com/Orleans/Orleans%20Team/_TestManagement/Runs?_a=resultSummary&runId=3397&resultId=101169
>Assert.Equal() Failure\r\n ↓ (pos 10)\r\nExpected: 21100,21101\r\nActual: 21100,21102\r\n ↑ (pos 10)

https://orleans.visualstudio.com/Orleans/Orleans%20Team/_TestManagement/Runs#_a=resultSummary&runId=3341&resultId=101222
>Assert.Equal() Failure\r\n ↓ (pos 10)\r\nExpected: 21200,21201\r\nActual: 21200,21202\r\n ↑ (pos 10)

sebastianburckhardt:
Yes, I am investigating. I am seeing failures in other tests as well.

sebastianburckhardt:
I have a fix for this test (see PR #3970)

It turned out to be a simple oversight in the test construction - need to sort the silo addresses in the test so they match the order used by the multi-cluster gateway selection.

I have found some other issues in other tests but will track them separately.

sergeybykov:
The test is still failing intermittently but quite often.

rafikiassumani-msft:
@benjaminpetit I am closing this issue as I could not find the test that is failing.

PR #3970, opened 2018-02-05
Fix multicluster test (#3929)

sebastianburckhardt:
It turned out to be a simple oversight in the test construction - need to sort the silo addresses in the test so they match the order used by the multi-cluster gateway selection.

There are three commits here:
- an easy fix for the test
- while trying to find the root cause, I came across a small bug
- added tracing  that I needed to track this down


veikkoeeva:
Is this related to https://github.com/dotnet/orleans/pull/3964 ?

sebastianburckhardt:
@veikkoeeva: It looks like the same mistake, though made independently, in a different place.

There could be more "copies" of this mistake. Perhaps we should _always_ sort the lists of silos that are returned. For example, TestCluster.GetActiveSilos() does not sort them currently.

sergeybykov:
@dotnet-bot test functional

Issue #3955, opened 2018-02-02
Configuration Work - Formatters


jason-bragg:
Add options formatters for all the new options. (many already have formatters)

jason-bragg:
Closing - We've a default generic formatter which should cover all options registered for logging.

Issue #3950, opened 2018-02-02
Configuration Work - XML docs


jason-bragg:
While moving from monolithic Cluster and Client configurations to options, settings without xml comments were copied without comment, leaving it unclear what these settings are for.

We need to go through all of the new options and ensure that the options themselves as well as all of it's properties have xml comments.

sergeybykov:
#4183 re-enables building of XML docs. Now need to fix the errors.

PR #3972, opened 2018-02-06
Added formatters and xml comments for remaining options.

jason-bragg:
Added formatters for all remaining options except StorageOptions, which will be refactored anyway.
Added xml comments for all remaining options and properties except StorageOptions and ServicePointOptions.

Formatters have only been registered on silo, will need to register them on the client once we've an idea of what options are needed client side (Out of scope).

Related to:
"Configuration Work - Formatters" #3955 
"Configuration Work - XML docs" #3950

### DIFF
--- a/src/Orleans.Core/Configuration/Options/AdoNetOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/AdoNetOptions.cs
@@ -1,6 +1,12 @@
 
+using System.Collections.Generic;
+using Microsoft.Extensions.Options;
+
 namespace Orleans.Hosting
 {
+    /// <summary>
+    /// AdoNet settings
+    /// </summary>
     public class AdoNetOptions
     {
         /// <summary>
@@ -15,5 +21,26 @@ namespace Orleans.Hosting
         /// Set this property to override <see cref="AdoInvariant"/> for reminders.
         /// </summary>
         public string InvariantForReminders { get; set; }
+    }
+
+    public class AdoNetOptionsFormatter : IOptionFormatter<AdoNetOptions>
+    {
+        public string Category { get; }
+
+        public string Name => nameof(AdoNetOptions);
+        private AdoNetOptions options;
+        public AdoNetOptionsFormatter(IOptions<AdoNetOptions> options)
+        {
+            this.options = options.Value;
+        }
+
+        public IEnumerable<string> Format()
+        {
+            return new List<string>()
+            {
+                OptionFormattingUtilities.Format(nameof(this.options.Invariant), this.options.Invariant),
+                OptionFormattingUtilities.Format(nameof(this.options.InvariantForReminders), this.options.InvariantForReminders),
+            };
+        }
     }
 }
--- a/src/Orleans.Core/Configuration/Options/GrainCollectionOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/GrainCollectionOptions.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Microsoft.Extensions.Options;
 
 namespace Orleans.Hosting
 {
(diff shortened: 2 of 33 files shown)


src/Orleans.Core/Configuration/Options/AdoNetOptions.cs
```diff
@@ -16,4 +22,25 @@ public class AdoNetOptions
         /// </summary>
         public string InvariantForReminders { get; set; }
     }
+
+    public class AdoNetOptionsFormatter : IOptionFormatter<AdoNetOptions>
+    {
+        public string Category { get; }
```
ReubenBond: Is this supposed to be set somewhere?

src/Orleans.Core/Configuration/Options/GrainCollectionOptions.cs
```diff
@@ -8,12 +9,47 @@ namespace Orleans.Hosting
...
+        }
+
+        public IEnumerable<string> Format()
+        {
+            var formated = new List<string>()
+            {
+                OptionFormattingUtilities.Format(nameof(this.options.CollectionQuantum), this.options.CollectionQuantum),
+                OptionFormattingUtilities.Format(nameof(this.options.CollectionAge), this.options.CollectionAge),
+            };
+            foreach(KeyValuePair<string, TimeSpan> classCollectionAge in this.options.ClassSpecificCollectionAge)
```
ReubenBond: Do we need a null check before enumerating this?

src/Orleans.Core/Configuration/Options/GrainServiceOptions.cs
```diff
@@ -1,10 +1,34 @@
...
 {
+    /// <summary>
+    /// Configuration of silo grain services.
+    /// </summary>
     public class GrainServiceOptions
     {
+        /// <summary>
+        /// List of grain services to initialize at startup.  List of full type name (string) and service id (short).
+        /// </summary>
         public List<KeyValuePair<string, short>> GrainServices { get; set; } = new List<KeyValuePair<string, short>>();
```
ReubenBond: Not for this PR, but it seems like these should be registered as a collection of services, maybe keyed (I don't know) in future. Sound right to you?

src/Orleans.Core/Configuration/Options/MultiClusterOptions.cs
```diff
@@ -54,6 +55,40 @@ public static class BuiltIn
...
+        {
+            this.options = options.Value;
+        }
+
+        public IEnumerable<string> Format()
+        {
+            var formated = new List<string>()
+            {
+                OptionFormattingUtilities.Format(nameof(this.options.HasMultiClusterNetwork), this.options.HasMultiClusterNetwork),
+                OptionFormattingUtilities.Format(nameof(this.options.DefaultMultiCluster), string.Join(";", this.options.HasMultiClusterNetwork)),
```
ReubenBond: Wrong property here

src/Orleans.Core/Configuration/Options/ThreadPoolOptions.cs
```diff
@@ -1,11 +1,18 @@
...
 using System.Collections.Generic;
+using Microsoft.Extensions.Options;
 
 namespace Orleans.Hosting
 {
+    /// <summary>
+    /// Thread pool settings
+    /// NOTE: Should this be in orleans core? - jbragg
+    /// </summary>
     public class ThreadPoolOptions
```
ReubenBond: Probably belongs in runtime

src/Orleans.Core/Configuration/Options/VersioningOptions.cs
```diff
@@ -1,16 +1,25 @@
...
 using Orleans.Versions.Compatibility;
 using Orleans.Versions.Selector;
-using System.Collections.Generic;
 
 namespace Orleans.Hosting
 {
+    /// <summary>
+    /// Versioning options govern grain implementation selection in heterogeneous deployments.
+    /// </summary>
     public class VersioningOptions
```
ReubenBond: We should probably rename this GrainVersioningOptions since it's not relevant to model versioning or silo/client versioning

ReubenBond:
Needs just a couple of fixes before we can merge it

src/Orleans.Core/Configuration/Options/ThreadPoolOptions.cs
jason-bragg: Should be in Orleans at all?  Shouldn't this be the application's responsibility?

src/Orleans.Core/Configuration/Options/GrainServiceOptions.cs
jason-bragg: Agreed.  Grain services pattern needs work.  I refactored it some when moving from ClusterConfig, but it needs more love.

src/Orleans.Core/Configuration/Options/AdoNetOptions.cs
jason-bragg: I don't recall what category is for.  I'll check with Xiao.

src/Orleans.Core/Configuration/Options/GrainCollectionOptions.cs
jason-bragg: Null ClassSpecificCollectionAge is not valid.  Think we should probably handle that in a validator if/when necessary.

src/Orleans.Core/Configuration/Options/ThreadPoolOptions.cs
ReubenBond: You're right, we should try to remove it at some point

ReubenBond:
Only remaining comment is about the `Category` property

jason-bragg:
Category is for classifying named options.  Not necessary if not a named option.

src/Orleans.Core/Configuration/Options/AdoNetOptions.cs
jason-bragg: Category is for classifying named options. Not necessary if not a named option.

ReubenBond:
Sure thing - it seems that it shouldn't be null, still, and maybe the name category is misleading if it's only for named options. If it's never used then perhaps it should be removed until it's needed.

PR #3974, opened 2018-02-06
Fixes for Multi-Cluster Support

sebastianburckhardt:
It's been a while since we added multi-cluster support. Running all the tests, I found some bugs:

- when running with at least three clusters, and activating a grain concurrently, messages could get forwarded excessively, causing activation to fail (due to the forward count limit). I fixed this by forwarding only to destinations that are known to be the owner.
- Message.cs contained a bug that made the IsReturnedFromRemoteCluster flag not work correctly. This could also cause excessive forwarding, and failure to remove stale directory cache entries.
- in two places, NullReferenceExceptions were thrown when multi-clusters had no gateways reporting. I put in the proper handling of these situations.
- some tests were failing because the gsi-directory-maintenance was happening too late. To fix this I changed the periodic maintenance loop so it  runs the maintenance immediately after each multi-cluster-configuration change.

I also fixed some tracing issues and added more tests to the Functional category, so we can cover these types of errors in regression.

### DIFF
--- a/src/Orleans.Runtime/GrainDirectory/MultiClusterRegistration/GlobalSingleInstanceActivationMaintainer.cs
+++ b/src/Orleans.Runtime/GrainDirectory/MultiClusterRegistration/GlobalSingleInstanceActivationMaintainer.cs
@@ -10,10 +10,12 @@ using Orleans.SystemTargetInterfaces;
 using Orleans.Runtime.Scheduler;
 using OutcomeState = Orleans.Runtime.GrainDirectory.GlobalSingleInstanceResponseOutcome.OutcomeState;
 using Orleans.Runtime.MultiClusterNetwork;
+using Orleans.MultiCluster;
+using System.Threading;
 
 namespace Orleans.Runtime.GrainDirectory
 {
-    internal class GlobalSingleInstanceActivationMaintainer : AsynchAgent
+    internal class GlobalSingleInstanceActivationMaintainer : AsynchAgent, IMultiClusterConfigurationListener
     {
         private readonly object lockable = new object();
         private readonly LocalGrainDirectory router;
@@ -28,6 +30,9 @@ namespace Orleans.Runtime.GrainDirectory
         // therefore, we maintain a list of potentially doubtful activations on the side.
         // maintainer periodically takes and processes this list.
         private List<GrainId> doubtfulGrains = new List<GrainId>();
+
+        // used to cut short the waiting time before next run
+        private ManualResetEvent runNow = new ManualResetEvent(false);
 
         public GlobalSingleInstanceActivationMaintainer(
             LocalGrainDirectory router,
@@ -47,6 +52,7 @@ namespace Orleans.Runtime.GrainDirectory
             this.siloDetails = siloDetails;
             this.multiClusterOptions = multiClusterOptions.Value;
             this.period = multiClusterOptions.Value.GlobalSingleInstanceRetryInterval;
+            multiClusterOracle.SubscribeToMultiClusterConfigurationEvents(this);
             logger.Debug("GSIP:M GlobalSingleInstanceActivationMaintainer Started, Period = {0}", period);
         }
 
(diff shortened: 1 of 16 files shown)


src/Orleans.Runtime/GrainDirectory/MultiClusterRegistration/GlobalSingleInstanceActivationMaintainer.cs
```diff
@@ -95,10 +101,21 @@ public void TrackDoubtfulGrains(Dictionary<GrainId, IGrainInfo> newstuff)
             {
                 try
                 {
```
ReubenBond: The `AsynchAgent.Run()` method should not be `async void`, it's supposed to be synchronous and run in its own thread.

This will manifest as:
  * Tasks scheduled here will be continued on `TaskScheduler.Default`, not the agent's thread.
  * The logs will print out a message like "Starting AsyncAgent X on managed thread Y" immediately followed by "Stopping AsyncAgent X that runs on managed thread Y".
  * The agent's fault handling policies don't get a chance to take effect.

That behavior is all probably fine here, it seems to defeat the purpose of using `AsynchAgent`.

Since this isn't new to this PR, we don't need to change it here, but it's worth noting.

src/Orleans.Runtime/GrainDirectory/MultiClusterRegistration/GlobalSingleInstanceActivationMaintainer.cs
```diff
@@ -95,10 +101,21 @@ public void TrackDoubtfulGrains(Dictionary<GrainId, IGrainInfo> newstuff)
...
+                        await Task.Delay(period, cts.Token);
 
-                    logger.Debug("GSIP:M running periodic check (having waited {0})", period);
+                        logger.Debug("GSIP:M running periodic check (having waited {0})", period);
+                    }
+                    catch (TaskCanceledException)
+                    {
+                        logger.Debug("GSIP:M running check");
+
+                        cts = new CancellationTokenSource(); // reset for next time
```
ReubenBond: There's also a protected `Cts` member which is in scope here. The two might be misleading. `Cts` will be cancelled when the agent is supposed to terminate.

test/TesterInternal/GeoClusterTests/LogConsistencyTestFixture.cs
```diff
@@ -203,6 +207,16 @@ public virtual void Dispose()
 
         private const int Xyz = 333;
 
+        private void AssertEqual<T>(T expected, T actual, string grainidentity)
```
ReubenBond: Generally we use lower camel case for instance fields, variables, and parameters, so `grainidentity` would be `grainIdentity`.

sebastianburckhardt: yep.

src/Orleans.Runtime/GrainDirectory/MultiClusterRegistration/GlobalSingleInstanceActivationMaintainer.cs
sebastianburckhardt: that's definitely confusing. I will choose some better name.

src/Orleans.Runtime/GrainDirectory/MultiClusterRegistration/GlobalSingleInstanceActivationMaintainer.cs
sebastianburckhardt: Makes sense. May as well change it in this PR.

ReubenBond:
@dotnet-bot test functional

sebastianburckhardt:
Don't merge this yet. I am still seeing some failures in the tests.

ReubenBond:
@sebastianburckhardt there's an issue (not caused by this) which is causing functionals to run slowly. I squashed and rebased your branch to fix a merge conflict, so you will need to pull it to make any changes

sebastianburckhardt:
I found the root cause for the failures I saw... it is a race condition in the state-based log view adaptor. I have added a (3-line) commit to fix this issue.

Independently from this, I now see failures since the last merge caused by configuration issues. Probably related to the use of legacy configuration.

ReubenBond:
@sebastianburckhardt could you rebase this on top of master to fix the conflicts? That might also help with the tests timing out.

sebastianburckhardt:
I am rebasing but now I see new failures.

sebastianburckhardt:
I rebased and fixed 2 minor configuration and build issues. Everything looks like it is passing now.

PR #3977, opened 2018-02-07
Move load shedding settings to LoadSheddingOptions

sergeybykov:
This is a preparation step for #3948 and a useful cleanup step in general.

### DIFF
--- a/src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
+++ b/src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
@@ -294,7 +294,7 @@ namespace Orleans.ServiceBus.Providers
             };
 
             return new EventHubAdapterReceiver(config, this.CacheFactory, this.CheckpointerFactory, this.loggerFactory, this.ReceiverMonitorFactory(receiverMonitorDimensions, this.loggerFactory, this.telemetryProducer), 
-                this.serviceProvider.GetRequiredService<IOptions<SiloStatisticsOptions>>().Value,
+                this.serviceProvider.GetRequiredService<IOptions<LoadSheddingOptions>>().Value,
                 this.telemetryProducer,
                 this.EventHubReceiverFactory);
         }
--- a/src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
@@ -1,0 +1,48 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Options;
+
+namespace Orleans.Hosting
+{
+    public class LoadSheddingOptions
+    {
+        /// <summary>
+        /// The LoadShedding element specifies the gateway load shedding configuration for the node.
+        /// If it does not appear, gateway load shedding is disabled.
+        /// </summary>
+        public bool LoadSheddingEnabled { get; set; }
+
+        /// <summary>
+        /// The LoadLimit attribute specifies the system load, in CPU%, at which load begins to be shed.
+        /// Note that this value is in %, so valid values range from 1 to 100, and a reasonable value is
+        /// typically between 80 and 95.
+        /// This value is ignored if load shedding is disabled, which is the default.
+        /// If load shedding is enabled and this attribute does not appear, then the default limit is 95%.
+        /// </summary>
+        public int LoadSheddingLimit { get; set; } = DEFAULT_LOAD_SHEDDING_LIMIT;
+        public const int DEFAULT_LOAD_SHEDDING_LIMIT = 95;
+    }
+
+    public class LoadSheddingOptionsFormatter : IOptionFormatter<LoadSheddingOptions>
+    {
+        public string Category { get; }
+
+        public string Name => nameof(LoadSheddingOptions);
+
+        private readonly LoadSheddingOptions options;
+
+        public LoadSheddingOptionsFormatter(IOptions<LoadSheddingOptions> options)
+        {
+            this.options = options.Value;
+        }
+
+        public IEnumerable<string> Format()
+        {
+            return new List<string>()
+            {
+                OptionFormattingUtilities.Format(nameof(this.options.LoadSheddingEnabled), this.options.LoadSheddingEnabled),
+                OptionFormattingUtilities.Format(nameof(this.options.LoadSheddingLimit), this.options.LoadSheddingLimit),
+            };
+        }
+    }
+}
--- a/src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
@@ -7,13 +7,13 @@ namespace Orleans.Hosting
     public class LoadSheddingOptions
     {
         /// <summary>
-        /// The LoadShedding element specifies the gateway load shedding configuration for the node.
-        /// If it does not appear, gateway load shedding is disabled.
+        /// Specifies whether or not load shedding in the client gateway and stream providers is enabled.
+        /// The default value is false, meaning that load shedding is disabled.
         /// </summary>
         public bool LoadSheddingEnabled { get; set; }
 
         /// <summary>
-        /// The LoadLimit attribute specifies the system load, in CPU%, at which load begins to be shed.
+        /// Specifies the system load, in CPU%, at which load begins to be shed.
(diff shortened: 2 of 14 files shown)


src/OrleansProviders/Streams/LoadShedQueueFlowController.cs
```diff
@@ -24,7 +24,7 @@ public class LoadShedQueueFlowController : IQueueFlowController
         /// <param name="options">The silo satistics options.</param>
         /// <param name="percentOfSiloSheddingLimit">Percentage of load shed limit which triggers a reduction of queue read rate.</param>
         /// <returns></returns>
-        public static IQueueFlowController CreateAsPercentOfLoadSheddingLimit(SiloStatisticsOptions options, int percentOfSiloSheddingLimit = SiloStatisticsOptions.DEFAULT_LOAD_SHEDDING_LIMIT)
+        public static IQueueFlowController CreateAsPercentOfLoadSheddingLimit(LoadSheddingOptions options, int percentOfSiloSheddingLimit = LoadSheddingOptions.DEFAULT_LOAD_SHEDDING_LIMIT)
         {
             if (percentOfSiloSheddingLimit < 0.0 || percentOfSiloSheddingLimit > 100.0) throw new ArgumentOutOfRangeException(nameof(percentOfSiloSheddingLimit), "Percent value must be between 0-100");
```
xiazen: isn't the logic here is a bug and we are going to remove it ?

src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Collections.Generic;
+using Microsoft.Extensions.Options;
+
+namespace Orleans.Hosting
+{
+    public class LoadSheddingOptions
+    {
+        /// <summary>
+        /// The LoadShedding element specifies the gateway load shedding configuration for the node.
+        /// If it does not appear, gateway load shedding is disabled.
```
ReubenBond: This should read something like, `Specified whether or not load gateway load shedding is enabled. The default value is false, meaning that gateway load shedding is disabled.`

Given this is about gateway load shedding, I wonder if we should name the class/props as such? Fine either way, we can change it in future.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -294,7 +294,7 @@ private EventHubAdapterReceiver MakeReceiver(QueueId queueId)
             };
 
             return new EventHubAdapterReceiver(config, this.CacheFactory, this.CheckpointerFactory, this.loggerFactory, this.ReceiverMonitorFactory(receiverMonitorDimensions, this.loggerFactory, this.telemetryProducer), 
-                this.serviceProvider.GetRequiredService<IOptions<SiloStatisticsOptions>>().Value,
```
xiazen: client side can be configured with EHStreamProvider too. But I didn't see code which configure a `LoadSheddingOptions` in client side. If no option is configured in DI in the client side , this call will fail right? Or am I missing anything. 
If load shedding is a silo only feature, client side still need to configure a no-op LoadSheddingOptions to just make DI happy.

src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        public int LoadSheddingLimit { get; set; } = DEFAULT_LOAD_SHEDDING_LIMIT;
+        public const int DEFAULT_LOAD_SHEDDING_LIMIT = 95;
+    }
+
+    public class LoadSheddingOptionsFormatter : IOptionFormatter<LoadSheddingOptions>
+    {
+        public string Category { get; }
+
+        public string Name => nameof(LoadSheddingOptions);
```
ReubenBond: I feel that we've swapped `Category` and `Name` in almost every IOptionFormatter instance.

`Category` = the option class name
`Name` = the named options instance (i.e, for named options)

@jason-bragg @xiazen am I wrong about this?

src/Orleans.Core/Configuration/Options/SiloStatisticsOptions.cs
```diff
@@ -60,8 +43,7 @@ public IEnumerable<string> Format()
             format.AddRange(new List<string>
             {
                 OptionFormattingUtilities.Format(nameof(this.options.DeploymentLoadPublisherRefreshTime), this.options.DeploymentLoadPublisherRefreshTime),
-                OptionFormattingUtilities.Format(nameof(this.options.LoadSheddingEnabled), this.options.LoadSheddingEnabled),
-                OptionFormattingUtilities.Format(nameof(this.options.LoadSheddingLimit), this.options.LoadSheddingLimit),
+                OptionFormattingUtilities.Format(nameof(this.options.DeploymentLoadPublisherRefreshTime), this.options.DeploymentLoadPublisherRefreshTime)
```
ReubenBond: This line is doubled up

src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
xiazen: Category means the category this component is in. For example, AzureTableStorageOption should have category name Storage, and name AzureTableStorageOption-name

If Category is too confusing, even for us, maybe we should just remove it.

src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
ReubenBond: I mean that we don't have attributes or elements anymore, because this isn't XML config, and the notion of a value appearing or not is no longer applicable.

src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
ReubenBond: @xiazen I agree with removing it. We can use `Name` values like `Storage.AzureTableStorageOptions.Default` and `Storage.AzureTableStorageOptions.MyHappyLittleGrainStore` to deal with named options and categories of options. That way, we can sort all formatters lexicographically when we display them and they'll be grouped nicely.

src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
sergeybykov: Fixed.

src/OrleansProviders/Streams/LoadShedQueueFlowController.cs
sergeybykov: I discussed it with Jason. It's not a bug. It's a protection from a corner case of stream pulling agents potentially overloading silo and preventing grain calls from getting processed.

src/Orleans.Core/Configuration/Options/SiloStatisticsOptions.cs
sergeybykov: Fixed.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
sergeybykov: I simply mimicked how we handle SiloStatisticsOptions. I don't see where we configure it in DI other than via legacy config.

jason-bragg: Options behavior in DI is such that the default will be returned if not option is configured.  So if LoadSheddingOptions is not configured on the client, this call will not file, it will just get the default LoadSheddingOptions.

jason-bragg: Having said that, this logic will never be called on the client because clients never create receivers.

src/Orleans.Core/Configuration/Options/LoadSheddingOptions.cs
xiazen: sounds good. I will have a PR out for removing it. or @jason-bragg do you want to do it? since you are touching formatter a lot. I don't want to produce merge conflicts for you

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
xiazen: that's right, I forgot that

Issue #3203, opened 2017-07-06
Subnet configuration is not usable with programmatic configuration


benjaminpetit:
In a "multi-homed" silo, the silo will choose the IP address to use by filtering with `NodeConfiguration.Subnet`. The issue is that this property is internal only, and this valus is set only if we are loading a xml configuration file: https://github.com/dotnet/orleans/blob/master/src/Orleans/Configuration/ClusterConfiguration.cs#L89

I think we should either 

1. Make `NodeConfiguration.Subnet` public: user can directly set the default value. It will also allow to customize this value per node using `ClusterConfiguration.Overrides`. We could then deprecate/remove `GlobalConfiguration.Subnet`, but this is a breaking change.
2. When setting `GlobalConfiguration.Subnet`, propagate it to `NodeConfiguration.Subnet`: we need to be sure that there is no issue with the current configuration parsing.

I am more leaning to solution 1. We could also rename the field since I find it a bit misleading: it is not really a subnet, more a "IpAddress.StartWith(xx)". Or implement a real subnet check.

xiazen:
+1  for first option. It is a small breaking change, and it brought more flexibility and ability to programmatic config. 

And I think we should do a real subset check, instead of change subnet to "IpAddressPrefix", for the purpose of less breaking change.

sergeybykov:
I'm in favor of option 2. The thinking here was that settings such as "subnet" (or IP address prefix) would be the same across all silos in the cluster. Hence, it is set as part of `GlobalConfiguiration`. Is there really a case for overriding "subnet" on a silo by silo basis?

benjaminpetit:
You could deploy a new set of silos that lives on a different subnet, but that can reach the "original" subnet. I agree that it is an edge case, but it is not harder for us to implement and it will give total flexibility if needed.

And I think we really should implement a real subnet network option (IP Address + netmask)

sergeybykov:
Is this needed now that #4005 has been merged?

benjaminpetit:
Not needed anymore

Issue #2054, opened 2016-08-13
How to configure multi-homed servers?


NielsHoogeveen:
At work, we are trying to set up an Orleans cluster on multi-homed servers, Each server is attached to two networks. while all of our client machine run in either one or the other network. The goal is to make it possible to connect as a client from any client machine in the network to the silo-cluster.

Right now, we create two silos per server where each server registers an IP-address for a corresponding network. The question arises if it is possible to register a silo to listen to multiple IP-addresses at the same time. 

Each client machine connects to the cluster by explicitly listing every silo reachable within the network. When using the system store, all silos (even the ones that can't be reached) are returned to the client. Leaving the question whether it is possible to add a filter to the client such that it only tries to connect to silos with a specific sub-net mask.

veikkoeeva:
Cross-referencing https://github.com/dotnet/orleans/issues/1812 (andhttps://github.com/dotnet/orleans/issues/1136#issuecomment-164364293). Maybe @amccool has pointers?

NielsHoogeveen:
The setup we're talking about is actually pretty simple. There are a number of servers  (at the moment 7), all of which have at least two NICs and all of which are connected to two networks.

One network is for computers connected to manufacturing machinery and one network is for all other computers. Only the latter has internet access. 

The two networks are only connected to each other through the servers, otherwise, they are galvanically isolated. 

Two servers are configured as domain controllers, two servers are configured as web servers with network load balancing for HA purposes, and two servers are configured as database server. The seventh server is mostly used for test purposes and software development.

All of those servers need to be accessible from both networks. The computers connected to the manufacturing machinery need access to the databases to log machine status. They need access to the web servers to retrieve work order information, and they need access to the domain controllers for naming purposes and access control. 

The computers, attached to the other network, need to access the databases for ad-hoc queries, excel import etc. They need to access the web servers to create work orders, to maintain bills-of-material, to maintain recipes etc. They also need access to the domain controllers for the obvious reasons. 

This setup works decent, but we do want to add a notification system. Now users have to pull data from the database or from our web application, but we want to notify them as well when certain events take place (eg. a new work order has been created, triggering a refresh of the list of work orders to process on the machines that control our manufacturing machines). Another application are real time dashboards with manufacturing information. These notifications need to become available in both networks.

To achieve this, we have installed Zookeeper on each of the seven server, creating a pretty robust store for the silo configuration. 

[...]

PR #3980, opened 2018-02-07
Add networking mask options in EndpointOptions

benjaminpetit:
Related issue: #3203 

Right now, using the legacy configuration, in the case of a "multi-homed" server, to force the selection of a specific IP address for the membership/gateway table, users have to use `GlobalConfiguration.SubNet`.  This solution is not suitable if the networks available on the server are not using network sliced in 8/16/24.

With this implementation in `EndpointOptions`, the user can use a real network address, with a real network mask.

### DIFF
--- a/src/Orleans.Runtime/Hosting/EndpointOptions.cs
+++ b/src/Orleans.Runtime/Hosting/EndpointOptions.cs
@@ -1,4 +1,7 @@
-﻿using System.Net;
+using Orleans.Runtime;
+using System;
+using System.Net;
+using System.Net.Sockets;
 
 namespace Orleans.Hosting
 {
@@ -6,16 +9,23 @@ namespace Orleans.Hosting
     /// Configures the Silo endpoint options
     /// </summary>
     public class EndpointOptions
-    { 
+    {
         /// <summary>
-        /// The external IP address or host name used for clustering.
+        /// The external IP address or host name used for clustering. IP address will be inferred from <see cref="NetworkAddress"/> if this is not
+        /// directly specified.
         /// </summary>
         public string HostNameOrIPAddress { get; set; }
 
         /// <summary>
-        /// The IP address used for clustering. Will be inferred from <see cref="HostNameOrIPAddress"/> if this is not directly specified.
+        /// The IP address used for clustering. Will be inferred from <see cref="HostNameOrIPAddress"/> or <see cref="NetworkAddress"/> if this is not directly specified.
         /// </summary>
         public IPAddress IPAddress { get; set; }
+
+        /// <summary>
+        /// Specify the subnet to determine to IP address used for clustering.
+        /// If <see cref="IPAddress"/> is specified, this property will be ignored.
+        /// </summary>
+        public NetworkAddress NetworkAddress { get; set; } = new NetworkAddress();
 
         /// <summary>
         /// The port this silo uses for silo-to-silo communication.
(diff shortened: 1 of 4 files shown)


src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -29,4 +39,65 @@ public class EndpointOptions
...
+        public NetworkAddress()
+            : this(IPAddress.Any, IPAddress.Any)
+        { }
+
+        /// <summary>
+        /// Create a new NetworkAddress
+        /// </summary>
+        public NetworkAddress(IPAddress address, IPAddress mask)
+        {
+            if (address.AddressFamily != mask.AddressFamily)
```
ReubenBond: This invariant can be broken by using the getters and setters. Maybe we should make the setters private

src/Orleans.Runtime/Hosting/EndpointOptionsExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            var candidates = new List<IPAddress>();
+
+            // Get IP address from DNS. If addrOrHost is localhost will 
+            // return loopback IPv4 address (or IPv4 and IPv6 addresses if OS is supported IPv6)
+            var nodeIps = Dns.GetHostAddresses(options.HostNameOrIPAddress);
+            foreach (var nodeIp in nodeIps.Where(x => x.AddressFamily == family))
+            {
+                // If the subnet does not match - we can't resolve this address.
+                // If subnet is not specified - pick smallest address deterministically.
```
ReubenBond: Is nodeIps sorted? Otherwise, how is this comment implemented?

src/Orleans.Runtime/Hosting/EndpointOptionsExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (IPAddress.TryParse(options.HostNameOrIPAddress, out address))
+            {
+                return address;
+            }
+
+            var candidates = new List<IPAddress>();
+
+            // Get IP address from DNS. If addrOrHost is localhost will 
+            // return loopback IPv4 address (or IPv4 and IPv6 addresses if OS is supported IPv6)
+            var nodeIps = Dns.GetHostAddresses(options.HostNameOrIPAddress);
```
ReubenBond: This is a blocking call. Maybe it's permissible here? I think it would be better to make this method async and push the blocking all the way up the stack to (I think) `LocalSiloDetails`

src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -29,4 +39,65 @@ public class EndpointOptions
...
+            var otherAddressBytes = otherAddress.GetAddressBytes();
+            var maskBytes = this.Mask.GetAddressBytes();
+
+            for (var i = 0; i < otherAddressBytes.Length; i++)
+            {
+                otherAddressBytes[i] = (byte)(otherAddressBytes[i] & maskBytes[i]);
+            }
+
+            return new IPAddress(otherAddressBytes).Equals(this.Address);
+        }
```
ReubenBond: This method only works if all bits which are 0 in `Mask` are also 0 in `Address`. i.e this will fail:
```C#
var netAddress = new NetworkAddress(IPAddress.Parse("10.0.0.1"), IPAddress.Parse("255.0.0.0"));
Assert.True(netAddress.Contains(IPAddress.Parse("10.0.0.2")));
```

So if I understand correctly, we should consider verifying that's true in the ctor.

src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -1,22 +1,32 @@
...
     /// </summary>
     public class EndpointOptions
-    { 
+    {
         /// <summary>
-        /// The external IP address or host name used for clustering.
+        /// The external IP address or host name used for clustering. IP address will be inferred from <see cref="NetworkAddress"/> if this is not
+        /// directly specified.
         /// </summary>
         public string HostNameOrIPAddress { get; set; }
```
ReubenBond: Could we maybe remove this property? Having a precedence order for properties is cumbersome/error prone.

We can have an async helper method for resolving an IPAddress from a string instead and let users consume that.

src/Orleans.Runtime/Hosting/EndpointOptions.cs
benjaminpetit: I agree, but I was worried about deserialization of those fields. You don't think that can be an issue with private setter?

src/Orleans.Runtime/Hosting/EndpointOptionsExtensions.cs
benjaminpetit: It's just an old comment that I forgot to remove. I don't think we care here., I will remove this comment

src/Orleans.Runtime/Hosting/EndpointOptionsExtensions.cs
benjaminpetit: Yes I removed the async from the original method, since I didnd't see the point since it is used in `LocalSiloDetails`... And I don't see why we would use it elsewhere?

veikkoeeva:
Cross-referencing https://github.com/dotnet/orleans/issues/2054, which in turn cross-references a bunch of other issues discussing this. :)

benjaminpetit:
Replaced by #4005

PR #3981, opened 2018-02-08
Options formating improvements.

jason-bragg:
Removed catalog, as it was deemed unnecessary.
Order logging by formatter name.  Helps with grouping of related options when using default formatters.
Allow user override of formatters before or after defaults.
Fixed issues with named options not reporting correctly.
Updated tests to verify by confirming expected log output.

### DIFF
--- a/src/Orleans.Core/Configuration/OptionLogger/IOptionsLogger.cs
+++ b/src/Orleans.Core/Configuration/OptionLogger/IOptionsLogger.cs
@@ -40,14 +40,18 @@ namespace Orleans
         }
         public void LogOptions()
         {
-            var optionFormatters = services.GetServices<IOptionFormatter>();
-            foreach (var optionFormatter in optionFormatters)
+            this.LogOptions(services.GetServices<IOptionFormatter>());
+        }
+
+        public void LogOptions(IEnumerable<IOptionFormatter> formatters)
+        {
+            foreach (var optionFormatter in formatters.OrderBy(f => f.Name))
             {
-                LogOption(optionFormatter);
+                this.LogOption(optionFormatter);
             }
         }
 
-        private void LogOption(IOptionFormatter formatter)
+        public void LogOption(IOptionFormatter formatter)
         {
             var stringBuiler = new StringBuilder();
             stringBuiler.AppendLine($"Configuration {formatter.Name}: ");
--- a/src/Orleans.Core/Configuration/OptionLogger/IOptionsLogger.cs
+++ b/src/Orleans.Core/Configuration/OptionLogger/IOptionsLogger.cs
@@ -29,6 +29,9 @@ namespace Orleans
         }
     }
 
+    /// <summary>
+    /// Base class for client and silo default options loggers.
+    /// </summary>
     public abstract class OptionsLogger 
     {
         private ILogger logger;
@@ -38,11 +41,19 @@ namespace Orleans
             this.logger = logger;
             this.services = services;
         }
+
+        /// <summary>
+        /// Log all options with registered formatters
+        /// </summary>
         public void LogOptions()
         {
             this.LogOptions(services.GetServices<IOptionFormatter>());
         }
 
+        /// <summary>
+        /// Log options using a set of formatters
+        /// </summary>
+        /// <param name="formatters"></param>
         public void LogOptions(IEnumerable<IOptionFormatter> formatters)
         {
             foreach (var optionFormatter in formatters.OrderBy(f => f.Name))
@@ -51,6 +62,10 @@ namespace Orleans
             }
         }
 
+        /// <summary>
+        /// Log an options using a formatter
+        /// </summary>
+        /// <param name="formatter"></param>
         public void LogOption(IOptionFormatter formatter)
         {
             var stringBuiler = new StringBuilder();
--- a/src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
+++ b/src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
@@ -42,35 +42,41 @@ namespace Orleans
         {
             var registration = services.FirstOrDefault(service => service.ServiceType == typeof(IOptionFormatter<TOptions>));
             if (registration == null)
-                services.AddSingleton<IOptionFormatter<TOptions>, TOptionFormatter>()
-                        .AddFromExisting<IOptionFormatter, IOptionFormatter<TOptions>>();
+                services.ConfigureFormatter<TOptions, TOptionFormatter>();
             return services;
         }
 
         /// <summary>
         /// Configure option formatter resolver for named option TOptions
         /// </summary>
-        public static IServiceCollection ConfigureFormatter<TOptions, TOptionFormatterResolver>(this IServiceCollection services, string name)
+        public static IServiceCollection ConfigureFormatterResolver<TOptions, TOptionFormatterResolver>(this IServiceCollection services)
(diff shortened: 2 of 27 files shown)


test/Tester/LogFomatterTests.cs
```diff
@@ -14,157 +16,447 @@ public class LogFomatterTests
...
+            services.ConfigureFormatter<TestOptions, TestOptionsFormatter.Resolver>("3");
+            var servicesProvider = services.BuildServiceProvider();
+            servicesProvider.GetRequiredService<OptionsLogger>().LogOptions();
+
+            var logFormatters = servicesProvider.GetServices<IOptionFormatter>();
+            Assert.Equal(3, logFormatters.Count());
+            Assert.True(logFormatters.First() is TestOptionsFormatter);
+            Assert.True(logFormatters.ElementAt(1) is TestOptionsFormatter);
+            Assert.True(logFormatters.ElementAt(2) is TestOptionsFormatter);
+            var logFormatter = servicesProvider.GetService<IOptionFormatterResolver<TestOptions>>();
```
xiazen: So named option 1 and 2 are configured and TestOptionsFormatter2, and named option 3 are configured with TestOptionsFormatter. But then they all use TestOptionsFormatter. Is this expected behavior?

src/Orleans.Core/Configuration/OptionLogger/IOptionsLogger.cs
```diff
@@ -40,14 +40,18 @@ protected OptionsLogger(ILogger logger, IServiceProvider services)
         }
         public void LogOptions()
         {
-            var optionFormatters = services.GetServices<IOptionFormatter>();
-            foreach (var optionFormatter in optionFormatters)
+            this.LogOptions(services.GetServices<IOptionFormatter>());
+        }
+
+        public void LogOptions(IEnumerable<IOptionFormatter> formatters)
```
xiazen: xml comments

src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
```diff
@@ -51,7 +55,7 @@ public static class OptionConfigureExtensionMethods
             where TOptionFormatterResolver : class, IOptionFormatterResolver<TOptions>
         {
             services.AddSingleton<IOptionFormatterResolver<TOptions>, TOptionFormatterResolver>()
-                .AddSingleton<IOptionFormatter>(sp => sp.GetService<IOptionFormatterResolver<TOptions>>().Resolve(name));
+                    .AddSingleton<IOptionFormatter>(sp => sp.GetService<IOptionFormatterResolver<TOptions>>().Resolve(name));
```
xiazen: if some user called 
```
ConfigureFormatter<TOptions, TOptionFormatterResolver1>(name1);
ConfigureFormatter<TOptions, TOptionFormatterResolver2>(name1);
```
Will this result in duplicated IOptionFormatter for name1 option in DI? I didn't see a test case covering this case

src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
```diff
@@ -64,7 +68,8 @@ public static class OptionConfigureExtensionMethods
         {
             var registration = services.FirstOrDefault(service => service.ServiceType == typeof(IOptionFormatterResolver<TOptions>));
             if (registration == null)
-                services.ConfigureFormatter<TOptions, TOptionFormatterResolver>(name);
+                services.AddSingleton<IOptionFormatterResolver<TOptions>, TOptionFormatterResolver>();
+            services.AddSingleton<IOptionFormatter>(sp => sp.GetService<IOptionFormatterResolver<TOptions>>().Resolve(name));
```
xiazen: same here. If call patter as follows:
```
TryConfigureFormatter<TOptions, TOptionFormatterResolver1>(name1);
ConfigureFormatter<TOptions, TOptionFormatterResolver2>(name1);
```
or 
```
ConfigureFormatter<TOptions, TOptionFormatterResolver2>(name1);
TryConfigureFormatter<TOptions, TOptionFormatterResolver1>(name1);
```

What's the expected behavior?

jason-bragg:
KK fixed :)

src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
```diff
@@ -42,35 +42,41 @@ public static class OptionConfigureExtensionMethods
...
-                services.AddSingleton<IOptionFormatterResolver<TOptions>, TOptionFormatterResolver>();
-            services.AddSingleton<IOptionFormatter>(sp => sp.GetService<IOptionFormatterResolver<TOptions>>().Resolve(name));
+                return services.ConfigureFormatterResolver<TOptions, TOptionFormatterResolver>();
             return services;
         }
+
+        /// <summary>
+        /// Configure a named option to be logged
+        /// </summary>
+        public static IServiceCollection ConfigureNamedOptionForLogging<TOptions>(this IServiceCollection services, string name)
```
xiazen: I'm not entirely sure what's the pattern here to use this one. So I think the idea is this function can only be called once. But how do we enforce it ?

xiazen: nvm.  I think I understand it

jason-bragg: We don't.  If users call it twice, the get information logged twice.  :/ User error.

xiazen:
I had  a different use pattern in mind before. But after some thinking, I think your pattern is better. 
Just can we add a test for named options. showing if a user configure the same named option twice, only one of them would be printed out ?

src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
xiazen: I thought I understood, But I was wrong.
So consider this scenario. In one of our in house provider, `AddAzureTableGrainStorage` for example, we would call `ConfigureNamedOptionForLogging` in the method to configure formatter by default. If user want to configure a different one, they would call `ConfigureNamedOptionForLogging` again. So really user cannot avoid calling this twice.

jason-bragg: Why would a user call ConfigureNamedOptionForLogging?  There is never a need unless they've added a new named option.  If they want to change the formatter for a type of named option they'd call ConfiguredConfigureFormatterResolver.

jason-bragg:
@dotnet-bot please test functional

xiazen:

src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
xiazen: Sorry I asked mainly because the use pattern of `ConfigureNamedOptionForLogging` isn't clear to me. But we cleared that out in person.... I wasn't trying to push back or anything, because after all this pr improves the feature. We can merge this after tests passing

jason-bragg: all good.  Pattern still has some holes we can/should plug.

jason-bragg:

jason-bragg:
Test failures look unrelated and pass locally.. thoughts?

PR #3986, opened 2018-02-08
Gateway should honor address settings and allow localhost calls from client.

jms69:
```
config.Defaults.ProxyGatewayEndpoint = new IPEndPoint(IPAddress.Any, 40000);
```
This setting does not work because the Gateway endpoint will always be set to the IP address of the silo.

I sidecar scenarios we want to be able to call the gateway as localhost:40000 by setting the service to IPAddress.Any or IPAddress.Loopback.

This fix allows the Gateway address setting to be honored.  If the IPAddress is set to None, then we function same as today by copying Silo address to Gateway address.


jms69:
Cancelling.  Did not do this properly.

Issue #3976, opened 2018-02-06
Refactor statistic and metric system


xiazen:
Currently orleans metric and statistic system has some problems which  made it confusing to use, and error prone 
- [ ]  use assembly loading to configure default `IMetricsDataPublisher` and `IStatisticsPublisher`, which is error prone and also against the new way of configuring providers 
- [ ] use questionable way to determine to use default metric publisher and statistic publisher or not. See `CanUseAzureTable` in `ClientStatisticsManager` as an example. 
- [x] has multiple interfaces, such as `IStatisticsPublisher`, `IMetricDataPublisher`, whose difference isn't clear. For example, Which metrics belongs to statistics, and which belongs to metrics? 
- [ ] overlap with telemetry. So basically now we have three system reporting metrics in orleans, and their responsibility and infrastructure overlap with each other. This made metrics harder to maintain and harder to use. 

Considering refactor it.

sergeybykov:
The plan here is to remove the notion of metrics altogether, as they have outlived their utility, and to get out of assembly loading for statistics. Refactoring and reconciliation of statistics and telemetry will be considered post 2.0.0.

ReubenBond:
`SiloPerformanceMetrics` is used in a few places still, after #3988:
* For [gateway load shedding](https://github.com/dotnet/orleans/blob/d88fa34b68c9100423bc43ecb608e4caa48693e2/src/Orleans.Runtime/Messaging/GatewayAcceptor.cs#L101) `IsOverloaded` + the related test hooks to force the value)
* For thread injection in [`WorkerPoolThread`](https://github.com/dotnet/orleans/blob/d88fa34b68c9100423bc43ecb608e4caa48693e2/src/Orleans.Runtime/Scheduler/WorkerPoolThread.cs#L332)
* As the basis for `SiloRuntimeStatistics` which is pushed to other silos in `DeploymentLoadPublisher`.

Should we keep that or try to remove it? Either way, I can try to slim it down

xiazen:
I think we can remove it. 
Most of its metrics is covered by `IAppEnvironmentStatistics` and `IHostEnvironmentStatistics`. So for `SiloRuntimeStatistics` and `WorkerPoolThread`, we can use those two , instead of SiloPerformanceMetrics. Right? 

[...]

PR #3988, opened 2018-02-08
Remove metrics publishers

ReubenBond:
This partially addresses #3976

Snapshots of current system behavior are published using various interfaces. This removes that publishing mechanism.


sergeybykov:
@dotnet-bot test functional

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
```diff
@@ -3,47 +3,36 @@
...
-        private readonly ILoggerFactory loggerFactory;
-        private MonitoringStorageOptions storageOptions;
-        private string dnsHostName;
+        private readonly MonitoringStorageOptions storageOptions;
+        private readonly string dnsHostName;
         public ClientStatisticsManager(
             IOptions<MonitoringStorageOptions> storageOptions,
             SerializationManager serializationManager, 
             IServiceProvider serviceProvider,
-            IHostEnvironmentStatistics hostEnvironmentStatistics,
```
xiazen: why removing this two. Doesn't statistics still reports cpu, memory and all that ?

xiazen: Oh I guess these two are here only for constructing the default metric publisher.
Can you make sure statistics still reports `IHostEnvironmentStatistics ` and `IAppEnvironmentStatistics ` ? since those are pretty important statistics

src/Orleans.Core/Statistics/ClientTableStatistics.cs
```diff
@@ -1,142 +0,0 @@
-#define LOG_MEMORY_PERF_COUNTERS 
-using System;
-using System.Threading.Tasks;
-using Microsoft.Extensions.Logging;
-using Orleans.Statistics;
-
-namespace Orleans.Runtime
-{
-    internal class ClientTableStatistics : IClientPerformanceMetrics, IDisposable
```
xiazen: why is this removed? isn't this statistics, not metrics?

src/Orleans.Core/Statistics/ClientStatisticsManager.cs
ReubenBond: These values are only published to the metrics table, so we can remove them since they're no longer used.

src/Orleans.Core/Statistics/ClientTableStatistics.cs
ReubenBond: The class name is statistics, but the interface is metrics. This is only used for snapshot metrics

ReubenBond:
Fixed test: metrics & stats values were mixed up in the test grain.

ReubenBond:
Simplified some classes and refactored a little, including removing something which has bothered me for a long time:
`Silo.cs:`
```diff
-            siloStatistics.MetricsTable.ActivationDirectory = activationDirectory;
-            siloStatistics.MetricsTable.ActivationCollector = catalog.ActivationCollector;
-            siloStatistics.MetricsTable.MessageCenter = messageCenter;
```

src/Orleans.Core/Statistics/IPerformanceMetrics.cs
```diff
@@ -29,60 +28,13 @@ public interface ICorePerformanceMetrics
...
-        /// are used for different grain types).
-        /// </summary>
-        int RecentlyUsedActivationCount { get; }
-
-        /// <summary>
-        /// Number of currently connected clients
-        /// </summary>
-        long ClientCount { get; }
-
+    {
```
xiazen: why do we remove IClientPerformanceMetrics and keep `ISiloPerformanceMetrics ` and `ICorePerformanceMetrics ` ? 

Especially `ICorePerformanceMetrics`, all of whose metrics are covered by `SiloRuntimeStatistics` ? 

Is it because other component is using it , such as Scheduler and WorkerPool? If yes, can they use `IAppEnvironmentStatistics` and `IHostEnvironmentStatistics` instead ?

ReubenBond: My bad, I explained it here: https://github.com/dotnet/orleans/issues/3976#issuecomment-364274529

ReubenBond:
@dotnet-bot test functional

ReubenBond:
@xiazen I removed the class entirely now

sergeybykov:
@dotnet-bot test functional

PR #3990, opened 2018-02-08
Rename "SQL" to "AdoNet" everywhere

### LINKED ISSUE #3915: ADO.NET configuration methods naming (Sql -> AdoNet?)
It appears the new Orleans ADO.NET related configuration methods use the term `Sql` (e.g. `.UseSqlMembership`), whereas `AdoNet` is used where support for older public interface hasn't been issue. Then the older pieces use `SqlServer`.

A more consistent experience would be picking one and considering how to best promote it. Since the library builds on ADO.NET and Nuget packages use that, the most natural, neutral choice, could be `AdoNet`. It's perhaps worth entertaining an idea if in the 2.0.0 release timeframe the `SqlServer` ones, such as `GlobalConfiguration.LivenessProviderType.SqlServer`, could have an alias such as `GlobalConfiguration.LivenessProviderType.AdoNet`.

sergeybykov:
Fixes #3915.


src/AdoNet/Orleans.Clustering.AdoNet/AdoNetHostingExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+using Orleans.Runtime.MembershipService;
+using Orleans.AdoNet.Configuration;
+using Orleans.AdoNet.Options;
+
+namespace Orleans.Hosting
+{
+    public static class AdoNetHostingExtensions
+    {
+        /// <summary>
+        /// Configure SiloHostBuilder with SqlMembership
```
veikkoeeva: Worth noting there is a string "sql" used rather copiously too. Maybe something like
> Configure SiloHostBuilder using ADO.NET membership.

I wonder also, since it seems to be in .NET API docs that whole sentenced starting with a capital letter and full stop is used if these should be fixed too? If not rejected, I could volunteer a PR for that for ADO.NET too as I was thinking to do something over the weekend. :)

sergeybykov: Good catch. Fixed.

sergeybykov:
@dotnet-bot test bvt

ReubenBond:
The enum changes break backwards compatibility, particularly with xml config. I added an update for the xsd.

PR #3998, opened 2018-02-08
Removed Core.Legacy dependency some azure extension pacakges

jason-bragg:
Removed Orleans.Core.Legacy dependency from:
Orleans.Cluster.AzureStorage
Orleans.Reminders.AzureStorage
Orleans.Streaming.Eventhubs

### DIFF
--- a/src/Orleans.Clustering.Consul/LegacyConsulGatewayListProviderConfigurator.cs
+++ b/src/Orleans.Clustering.Consul/LegacyConsulGatewayListProviderConfigurator.cs
@@ -1,10 +1,7 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Text;
 using Microsoft.Extensions.DependencyInjection;
 using Orleans.Hosting;
 using Orleans.Messaging;
-using Orleans.Runtime.Configuration;
 
 namespace Orleans.ConsulUtils
 {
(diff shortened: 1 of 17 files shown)


src/Orleans.Clustering.Consul/LegacyConsulGatewayListProviderConfigurator.cs
```diff
@@ -1,22 +1,20 @@
...
     {
         /// <inheritdoc/>
-        public void ConfigureServices(ClientConfiguration configuration, IServiceCollection services)
+        public void ConfigureServices(object configuration, IServiceCollection services)
         {
             services.UseConsulGatewayListProvider(options =>
             {
-                options.Address = new Uri(configuration.DataConnectionString);
+                var reader = new ClientConfigurationReader(configuration);
+                options.Address = new Uri(reader.GetPropertyValue<string>("DataConnectionString"));
```
xiazen: not necessary related to this pr. But since @jason-bragg is doing something to make diagnosing failure easier. Maybe it is good  to consider adding logging here to report reading failures. 
It can be 
- object is not a ClientConfiguration
- DataConnectionString on client config is not set. 

Stuff like that.

xiazen: Since this is using reflection to do hacky stuff, I suspect the exception trace will be friendly to read

xiazen: just a suggestion, can totally be done in another pr

Issue #3581, opened 2017-10-21
Observer invocation exceptions are swallowed


yevhen:
When client observer is disconnected/crashed invoking it (via reference) will not throw any exception and instead a warning is logged:

```log
[2017-10-20 18:14:57.257 GMT 22  WARNING 101002  Dispatcher  10.71.1.73:0]   
Discarding OneWay rejection for message OneWay 
S10.71.1.73:100:246193670*grn/A5EDFCAD/00000000+notifications@c5aef7ab
->*cli/ad24b798bf0261b8-df60-4510-8178-33598b710d08 #13027512: 
global::Orleankka.Core.IClientEndpoint:Receive(). 
Exc = No activation for client *cli/ad24b798
```

This leads to dead observer references being hold and accumulated by the grains since they can't reckon the situation and execute a clean up, which is a potential memory leak. Plus, resources are wasted by making redundant invocations. It means that built-in `ObserverSubscriptionManager` is also broken. 

This is a new bug in 1.5.x.

sergeybykov:
How come this is new? This is by design behavior of observers from the beginning, unfortunate as it is. I don't believe anything changed in 1.5.x. 

Because observer methods are `void` and they return immediately, there is no opportunity to throw an exception. This behavior can lead to a memory leak and wasting of resources of sending messages that get dropped on the client side.

That's why the recommended pattern is to periodically reconfirm observers, and expire subscriptions otherwise. Maybe we should update http://dotnet.github.io/orleans/Documentation/Core-Features/Observers.html to be more explicit about that.

In the hindsight, I think it was a mistake to include `ObserverSubscriptionManager` in its current form. I'm not sure if it's worth updating it with expiration logic.

yevhen:
> How come this is new?

You're right. I haven't used observers since streams were introduced and completely forgot how they do work and being confused by code in `ObserverSubscriptionManager`. It might be better to remove this code to avoid confusion (https://github.com/dotnet/orleans/blob/bfcc00f4fcd001334431bbf2511915e75747c192/src/Orleans.Core/Async/ObserverSubscriptionManager.cs#L93-L99).

> That's why the recommended pattern is to periodically reconfirm observers, and expire subscriptions otherwise

That's what I do.

sergeybykov:
>It might be better to remove this code to avoid confusion

You might be right. I think that code was written with an assumption that such exceptions would indicate failed subscriptions. In reality though, they can only indicate corner case immediate errors.

[...]

PR #3999, opened 2018-02-09
Move ObserverSubscriptionManager to legacy

xiazen:
#3581 
ObserverSubscriptionManager is a helper class for observer feature, which has bugs in it. It is not essential to observer feature. It is something user can totally handle using their way , specific useful for their application. We shouldn't stand in there , suggesting a solution which has bugs in it. 

Move this to legacy package. We only support it for legacy reasons.


MarcinJuraszek:
If it's legacy, should it be removed from the documentation? 

http://dotnet.github.io/orleans/Documentation/grains/observers.html

kentcb:
Arrived here in confusion as per the comment by @MarcinJuraszek, the documentation still references the class.

ReubenBond:
@kentcb we need to update that... In the mean time, ObserverSubscriptionManager is just a helper class. Here's an alternative: https://gist.github.com/ReubenBond/0bb430a52eedce50e6976f2f89154f5a
Or here: https://github.com/dotnet/orleans/blob/10af0f4af588cd4aa45cb3e250dfbffa389d59c7/test/Grains/TestGrains/ObserverManager.cs

PR #4003, opened 2018-02-09
Improved error logging of legacy systems

jason-bragg:


src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -121,41 +121,76 @@ public virtual void Participate(TLifecycle lifecycle)
...
+                {
+                    IProvider provider = this.provider.Value;
+                    IProviderRuntime runtime = this.services.GetRequiredService<IProviderRuntime>();
+                    await Schedule(() => provider.Init(this.config.Name, runtime, this.config));
+                    stopWatch.Stop();
+                    this.logger.Info(ErrorCode.SiloStartPerfMeasure, $"Initializing provider {this.config.Name} of type {this.config.Type} in stage {this.initStage} took {stopWatch.ElapsedMilliseconds} Milliseconds.");
+                } catch(Exception ex)
+                {
+                    stopWatch.Stop();
+                    this.logger.Error(ErrorCode.SiloStartPerfMeasure, $"Initialization failed for provider {this.config.Name} of type {this.config.Type} in stage {this.initStage} in {stopWatch.ElapsedMilliseconds} Milliseconds.", ex);
```
sergeybykov: Strange error code. Plus, as a general rule, we don't reuse error codes, so that there's no ambiguity which line logged it.

Issue #3984, opened 2018-02-08
Gateway cannot be called using localhost when running in non-local silo mode.  Kubernetes pod scenerio


jms69:
This is the scenario that breaks:

- Orleans silo running in container listening on port 22222 and gateway 40000 with IP.Any using Azure table membership
- Web app in container listening to port 80 using Orleans client calling Gateway client endpoint 127.0.0.1 port 40000
- Both containers are in side one Kubernetes Statefulset pod with external port 80 exposed.
- Silo to silo communications works well.

It is expected that the web app can communicate with the Orleans silo at localhost:40000 inside the pod.  This is not the case.  It seems that the gateway will always listen on the same address as the silo address.

```
this.gatewayAddressLazy = new Lazy<SiloAddress>(() => endpointOptions.ProxyPort != 0 ? SiloAddress.New(new IPEndPoint(this.SiloAddress.Endpoint.Address, endpointOptions.ProxyPort), 0) : null);
```
The config setting below where IP address is set to ANY is not being honored:
```
config.Defaults.ProxyGatewayEndpoint = new IPEndPoint(IPAddress.Any, 40000);
```

This breaks any sidecar scenarios in containerized applications.

jms69:
Proposed fix: https://github.com/dotnet/orleans/pull/3986

jms69:
Seems like the same issue as https://github.com/dotnet/orleans/issues/3676

I was trying to add an envoy proxy side car to the gateway  port when I discovered this problem.  I was experimenting on possibly securing silo communication with the envoy proxy.

https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/ssl.html?highlight=tls

galvesribeiro:
@jms69 looks like a good idea. Was discussing reverse proxy options last week. Will have a look on Envoy.

sergeybykov:
Resolved via #4005.

PR #4005, opened 2018-02-10
Refactor EndpointOptions to allow listening on an address that is different from the externally reachable address 

### LINKED ISSUE #3676: Allow listening on an address that is different from the externally reachable address
It would be nice if Orleans supports listening in an address that is not necessarily the same one that is externally addressable.
A very common example of this is when running in containers, where the Silo could be configured to bind to localhost and a fixed port, but then this is mapped into a virtual network and the host acts like a NAT.

So basically allow all the silos in their containers to listen on `localhost:22222` and `localhost:30000` but in their membership register with a distinct IP address and/or a different port that is passed in by the host (it should support changing ports, not just IP, since there might be many silos/containers running in the same host, and to avoid port conflicts with the host).

This might be a good candidate for doing when we replace the legacy `ClusterConfiguration` with the new `IOptions`-based configuration.

benjaminpetit:
Replace #3980

This will fix #3676 and will #3202 irrelevant. This should also address #3984 

We could add in the future more helper function to configure given a subnet, ect.


jms69:
Should we call it "AdvertisedIPAddress" instead of just IPAddress?
With this code I can put a silo behind a NAT and advertise the external IP address.

Is there a way to support AdvertiseGateway address?  In a load-balanced deployment where the Gateway is behind a load balancer can I advertise the external VIP of the gateway?

src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -8,25 +10,97 @@ namespace Orleans.Hosting
...
         /// </summary>
-        public int Port { get; set; }
+        public int SiloPort { get; set; }
 
         /// <summary>
         /// The port this silo uses for silo-to-client (gateway) communication. Specify 0 to disable gateway functionality.
         /// </summary>
         public int ProxyPort { get; set; }
 
-        //public bool BindToAny { get; set; }
```
jdom: wouldn't a boolean like this be enough?

benjaminpetit: No, with the proposed solution you can also bind to another port, different from the published port, if you want

benjaminpetit:
> Should we call it "AdvertisedIPAddress" instead of just IPAddress?
> With this code I can put a silo behind a NAT and advertise the external IP address.
Yes I think it would be a better name. With this solution you could put the silo behind a NAT with some port-forwarding. I'm not sure that this is a good idea regarding the performance, but it should work.

> Is there a way to support AdvertiseGateway address?
Not yet, but we plan to add this functionality later.

> In a load-balanced deployment where the Gateway is behind a load balancer can I advertise the external VIP of the gateway?
Currently the client expects direct connection to the gateways, for reliability and performance reasons. There is a blacklist mechanism on the client if it detects connection issue with a gateway, be careful with that if you "publish" only one endpoint.
Maybe we could add a specific provider for this case?

src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -8,25 +10,97 @@ namespace Orleans.Hosting
...
         /// <summary>
         /// The port this silo uses for silo-to-silo communication.
         /// </summary>
-        public int Port { get; set; }
+        public int SiloPort { get; set; }
 
         /// <summary>
         /// The port this silo uses for silo-to-client (gateway) communication. Specify 0 to disable gateway functionality.
         /// </summary>
         public int ProxyPort { get; set; }
```
sergeybykov: `ProxyPort` is a confusing name that's been used for historical reason that don't matter anymore. I think `GatewayPort` is the proper name here and everywhere else "proxy" is used.

src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -8,25 +10,97 @@ namespace Orleans.Hosting
...
+                options.IPAddress = ip;
+                options.ProxyPort = proxyPort;
+                options.SiloPort = siloPort;
+            });
+            return builder;
+        }
+
+        public static ISiloHostBuilder ConfigureEndpoints(
+            this ISiloHostBuilder builder, 
+            string addrOrHost,
```
sergeybykov: Since we'll have the overload above that takes an `IPAddress`, I think it'll be less confusing if we name this one `hostName`, even if we leave the logic that allows to pass string from of an IP address to it as well.

src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -8,25 +10,97 @@ namespace Orleans.Hosting
...
-        /// <summary>
-        /// The IP address used for clustering. Will be inferred from <see cref="HostNameOrIPAddress"/> if this is not directly specified.
+        /// The IP address used for clustering.
         /// </summary>
         public IPAddress IPAddress { get; set; }
 
         /// <summary>
         /// The port this silo uses for silo-to-silo communication.
         /// </summary>
-        public int Port { get; set; }
```
sergeybykov: For ports, do we want to have defaults? We've been documenting that 11111/30000 are the default ports we use, even though I don't see that we actually do that. I see three options here:
* Set them to 11111/30000 be default
* Set tthem to randomly picked available ports
* Do nothing, and fail to start a silo if the ports aren't specified

benjaminpetit: - If we fail we should do it with a proper exception message. We could add the required components to configure in the SiloHostBuilder
- I like the idea of random ports, if people want to run multiple silos on the same machine, it can really help them, but...
- Since it is not the only required core components for Orleans to run (we always need membership for example), could we have a default builder/methods that configure all required components with default options? This way all default will be in one place

src/Orleans.Runtime/Hosting/EndpointOptions.cs
jdom: ah, ok, I was under the old assumption that we didn't want to support port re-mapping.

src/Orleans.Runtime/Hosting/EndpointOptions.cs
jdom: I'd down vote the random port suggestion: it's more important to be deterministic, than supporting the multi-silo deployment without extra configuration. I'd say it would be more common to use containers for multi-silo deployments, where in fact random ports can be an issue (as it is in many locked-down-network deployments where the firewall needs to be explicitly configured, which is the majority of cases).

I'd prefer fixed port, but it would be not terrible if we throw a clear exception when unassigned, the question is what do we gain? (except that we silently bind even if there is another silo listening on the same port, behavior that I disagree with and shouldn't be the default).

sergeybykov: My vote is for defaulting to 11111/30000.

benjaminpetit: Anyway I don't think we should fix this in the PR?

src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -8,25 +10,97 @@ namespace Orleans.Hosting
     public class EndpointOptions
     { 
         /// <summary>
-        /// The external IP address or host name used for clustering.
+        /// The IP address used for clustering.
         /// </summary>
-        public string HostNameOrIPAddress { get; set; }
```
jdom: My concern with dropping this and only parsing when using the extension methods is that we drop support for declaratively configuring this (in appSettings.json, or XML or environment variables) so that it just works. Nevertheless of course the end user can explicitly parse this via the configuration subsystem and then pass it to the extension method, but there is a lot of extra juggling there...

jdom: BTW, even though this wasn't "pure", this parsing was being done from the string in the EndpointOptions class instead of like the new proposed way, by design. I'm OK if you decide to change it consciously, but wanted to make sure the issue is understood before moving on with a more pure but less practical approach

---
In reply to: [167736599](https://github.com/dotnet/orleans/pull/4005#discussion_r167736599) [](ancestors = 167736599)

PR #4007, opened 2018-02-10
Update Docker-Deployment.md

galvesribeiro:
Including reference to OrleansContrib\Orleans.Clustering.Kubernetes

### DIFF
--- a/src/Documentation/Deployment-and-Operations/Docker-Deployment.md
+++ b/src/Documentation/Deployment-and-Operations/Docker-Deployment.md
@@ -529,8 +529,7 @@ To run this article in a `Swarm` cluster, you don't have any extra work. When yo
 
 # Google Kubernetes (K8s)
 
-Coming soon...
-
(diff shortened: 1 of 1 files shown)


benjaminpetit:
Could you remote/update the part regarding the gateway also?

galvesribeiro:
@benjaminpetit what you mean? Sorry, not following.

benjaminpetit:
In the Gateway configuration section:

> You may have noticed that we are not using the membership provider in our client configuration. The > reason for that (and I figured out after days suffering and talking with Docker folks) is because the way our client -> membership -> gateway(s) -> silo works, is not compatible with the way Docker overlay network is designed.

Which is not true, it is misleading some people. I know I had opened a PR weeks ago about this, but still not merged.

But it would be great if you could fix this in this PR!

galvesribeiro:
Well, for Swarm it still sort of valid. But ok, let me remove that and update the PR.

benjaminpetit:
Awesome, thanks you!

PR #4018, opened 2018-02-13
More cleanup of types

sergeybykov:
Move `SystemStatuc` enum to `Orleans.Runtime`.
Remove `MetricsTableWriteInterval` from `StatisticsOptions` and all downstream dependencies.
Move `TransactionsOptions` to Orleans.Runtime to use it for transactions' `DefaultMetricWritePeriod`.


src/Orleans.Runtime/Orleans.Runtime.csproj
```diff
@@ -12,15 +12,17 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Options" Version="$(MicrosoftExtensionsOptionsVersion)" />
+    <Compile Remove="Providers\**" />
+    <EmbeddedResource Remove="Providers\**" />
+    <None Remove="Providers\**" />
```
ReubenBond: Are these elements needed? I think we can we just remove them since the folder is empty.

ReubenBond: I'll fix it in a subsequent PR

Issue #3969, opened 2018-02-05
Using system given free ports in tests to remove transient test failures in CI


veikkoeeva:
Anedoctally I experienced multiple errors in opening sockets in CI server, which appeared to be due to some socket being already reserved. 

As a first measure, I suggest adding a functionality to ask the system to give a free port from the ephemeral range and use that. The code could be like:

```
/// <summary>
/// Finds a free port from an ephemeral range.
/// </summary>
/// <returns>The found free port.</returns>
public static int GetFreePortFromEphemeralRange()
{
    var listener = new TcpListener(IPAddress.Loopback, 0);
    listener.Start();
    int port = ((IPEndPoint)listener.LocalEndpoint).Port;
    listener.Stop();

    return port;
}
```

When this change is implemented, it needs to be ensured all places are modified and the current code does not assume _well known_ ports. Assuming all places are changed to use system given free ports, failing tests should reveal places that assume _well known_ ports.

One option to refactor the tests could be to implement a cache, or a DI-like container to (lazily) generate and hold parameters that are then injected in all test class constructors. This could would allow sharing assets where feasible and perhaps make the testing infrastructure more uniform (and more approachable). There are some examples on how to do this at https://github.com/xunit/xunit/issues/687.

benjaminpetit:
A bit late but we already have something to take random unused port in `TestClusterBuilder.GetAvailableConsecutiveServerPortsPair`.

Feel free to reopen if you think it is not sufficient!

veikkoeeva:
@benjaminpetit This has been open for a while, so probably time to close.

Though the logic looks like asking if a semi-random port is open and if not, retrying instead of just asking an open port. :)

PR #4022, opened 2018-02-13
Migrate AdoNetGrainStorage to use options and lifecycle

xiazen:
- Migrate AdoNetGrainStorageto use options and lifecycle
- Add GrainPersistenceTestRunner 
- Add persistence tests to AdoNetGrainStorage


xiazen:
@veikkoeeva do you mind take a look of this? You are one of the people who are most familiar with AdoNetStorageProviders. Would love to get some feedback from you. Thanks you in advance

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Text;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using Orleans.Persistence.AdoNet.Storage;
+using Orleans.Runtime;
+using Orleans.Runtime.Configuration;
+
+namespace Orleans.Hosting
+{
+    public class AdoNetGrainStorageOptions
```
veikkoeeva: It would look nicer to comment all public types. :)

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// Default init stage in silo lifecycle
+        /// </summary>
+        public const int DEFAULT_INIT_STAGE = SiloLifecycleStage.ApplicationServices;
+
+        /// <summary>
+        /// The default ADO.NET invariant used for storage if none is given. 
+        /// </summary>
+        public const string DEFAULT_ADONET_INVARIANTS = AdoNetInvariants.InvariantNameSqlServer;
+        /// <summary>
+        /// The invariant name for storage
```
veikkoeeva: Missing dot.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+        public int InitStage { get; set; } = DEFAULT_INIT_STAGE;
+        /// <summary>
+        /// Default init stage in silo lifecycle
+        /// </summary>
+        public const int DEFAULT_INIT_STAGE = SiloLifecycleStage.ApplicationServices;
+
+        /// <summary>
+        /// The default ADO.NET invariant used for storage if none is given. 
+        /// </summary>
+        public const string DEFAULT_ADONET_INVARIANTS = AdoNetInvariants.InvariantNameSqlServer;
```
veikkoeeva: It looks like this should be in singular, i.e. `public const string DEFAULT_ADONET_INVARIANTS`. It looks like this will in public, user facing API, so a .NET naming convention `DefaultAdoNetInvariant` would seem less extravagant.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+        public TypeNameHandling? TypeNameHandling { get; set; }
+        #endregion
+        /// <summary>
+        /// Whether storage string payload should be formatted in Xml.
+        /// </summary>
+        public bool UseXmlFormat { get; set; }
+    }
+    public class AdoNetGrainStorageOptionsValidator : IConfigurationValidator
+    {
+        private readonly AdoNetGrainStorageOptions options;
```
veikkoeeva: Probably a matter of style more, but writing `private AdoNetGrainStorageOptions Options { get; }` would have the same effect and additional VS would show references immediately in CodeLens.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+        public bool UseFullAssemblyNames { get; set; }
+        public bool IndentJson { get; set; }
+        public TypeNameHandling? TypeNameHandling { get; set; }
+        #endregion
+        /// <summary>
+        /// Whether storage string payload should be formatted in Xml.
+        /// </summary>
+        public bool UseXmlFormat { get; set; }
+    }
+    public class AdoNetGrainStorageOptionsValidator : IConfigurationValidator
```
veikkoeeva: I'll add here too that having description of public types would be nice.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        private readonly AdoNetGrainStorageOptions options;
+        private readonly string name;
+        public AdoNetGrainStorageOptionsValidator(AdoNetGrainStorageOptions configurationOptions, string name)
```
veikkoeeva: Comments on public fields would be nice.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        private IOptionsSnapshot<AdoNetGrainStorageOptions> optionsSnapshot;
+        public AdoNetStorageOptionsFormatterResolver(IOptionsSnapshot<AdoNetGrainStorageOptions> optionsSnapshot)
+        {
+            this.optionsSnapshot = optionsSnapshot;
+        }
+
+        public IOptionFormatter<AdoNetGrainStorageOptions> Resolve(string name)
+        {
+            return new AzureTableStorageOptionsFormatter(name, optionsSnapshot.Get(name));
```
veikkoeeva: What does `AzureTableStorageOptionsFormatter` do here?

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </summary>
+        public bool UseXmlFormat { get; set; }
+    }
+    public class AdoNetGrainStorageOptionsValidator : IConfigurationValidator
+    {
+        private readonly AdoNetGrainStorageOptions options;
+        private readonly string name;
+        public AdoNetGrainStorageOptionsValidator(AdoNetGrainStorageOptions configurationOptions, string name)
+        {
+            this.options = configurationOptions;
```
veikkoeeva: It would definitely be cooler to check for parameters in the constructor. Even if it were "duplicate effort" that user is instructured optionally to use `ValidateConfiguration`.

test/TesterAdoNet/Persistence/PersistenceGrainTests_AdoNetSql.cs
```diff
@@ -0,0 +1,85 @@
...
+using TestExtensions.Runners;
+using UnitTests.General;
+using UnitTests.StorageTests.Relational;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Tester.AdoNet.Persistence
+{
+    /// <summary>
+    /// PersistenceGrainTests using AzureGrainStorage - Requires access to external Azure table storage
```
veikkoeeva: AzureGrainStorage?

veikkoeeva:
@xiazen Did Orleans binary (de)serialization get removed? If so, we should there might need to be an option for the user to move from binary to some other format or else they have potentially swathes of business critical information stranded. There is an example in the tests that switches from XML to JSON (or something like that, can't remember which now), but I'm unsure how to expose this on the public API.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: it format the option settings to an array and would be later logged out when silo start up.

xiazen:
@veikkoeeva binary (de)serialization didn't get removed. If you didn't configure either `UseXmlFormat`or `UseJsonFormat`, then the system will wire up with a binary (de)serializer. This is the original logic. I just removed `UseBinaryFormat` from the option, be consistent with other grain storage options. For example, in AzureTableGrainStorageOptions, it is doing the same thing. If `UseJsonFormat` is not set true, a binary serializer would be wired up. 

Does this sounds good ?

test/TesterAdoNet/Persistence/PersistenceGrainTests_AdoNetSql.cs
xiazen: yeah typo. I copied this tests from azure tests

xiazen:
fix pr feedback, rebased with master  and tests passed!

veikkoeeva:
> If you didn't configure either UseXmlFormator UseJsonFormat, then the system will wire up with a binary (de)serializer. This is the original logic. I

@xiazen Ah, I didn't realize! Now that you're on the hoods sorthing this out already, should the code comments have a `remark` section making this explicit?

Another question that can be addressed later, if you look at the storage provider code, you see the (de)serializers are wrapped into canonical interfaces. Would there be a way to expose these on the surface? Since doing that would also mean any other (de)serializer could be used. The naming and and some details might need to change, but I also wonder if this ought to be thought about before 2.0 release.

veikkoeeva:
Otherwise this looks OK to me (I haven't checked out the branch). :)

xiazen:
@veikkoeeva will add the remarks ! good idea. 

And to the question of (de)serializers, I think it is important to revamp that and fit it into 2.0, since it will touch public interfaces. I will bring it to the team and try to make it into 2.0.

But maybe not rc. RC release is to present major interface change to users. And this de(serializer) thing just belong to AdoNet. Does this sound good? 

Also, I have some question about this (de)serializer feature. Those (de)serializer are wrapped into their own canonical interfaces, and further put into a `DefaultRelationalStoragePicker`. This confuses me because it seems like only one (de)serializer is allowed. So why do we have a picker? And also there's a ` public IStorageSerializationPicker StorageSerializationPicker { get; set; }` on `AdoNetStorageProvider`, which seems to provide a way to modify the picker dynamically in runtime. I guess I need more information on how this feature are supposed to work.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,138 @@
...
+    {
+        /// <summary>
+        /// Connection string for AdoNet storage.
+        /// </summary>
+        public string ConnectionString { get; set; }
+
+        /// <summary>
+        /// Stage of silo lifecycle where storage should be initialized.  Storage must be initialzed prior to use.
+        /// </summary>
+        public int InitStage { get; set; } = DEFAULT_INIT_STAGE;
```
ReubenBond: I think this shouldn't be configurable

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,138 @@
...
+        public const int DEFAULT_INIT_STAGE = SiloLifecycleStage.ApplicationServices;
+
+        /// <summary>
+        /// The default ADO.NET invariant used for storage if none is given. 
+        /// </summary>
+        public const string DEFAULT_ADONET_INVARIANT = AdoNetInvariants.InvariantNameSqlServer;
+        /// <summary>
+        /// The invariant name for storage.
+        /// </summary>
+        public string AdoInvariant { get; set; } = DEFAULT_ADONET_INVARIANT;
```
ReubenBond: In other options, I've just been calling it `Invariant`, which might be bad, but I figure if `AdoNet` is in the class name then it doesn't need to be in property names

ReubenBond: Either way, we should be consistent

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,138 @@
...
+        public bool UseJsonFormat { get; set; }
+        public bool UseFullAssemblyNames { get; set; }
+        public bool IndentJson { get; set; }
+        public TypeNameHandling? TypeNameHandling { get; set; }
+        #endregion
+        /// <summary>
+        /// Whether storage string payload should be formatted in Xml.
+        /// <remarks>If neither <see cref="UseJsonFormat"/> nor <see cref="UseXmlFormat"/> is set to true, then BinaryFormatSerializer will be configured to format storage string payload.</remarks>
+        /// </summary>
+        public bool UseXmlFormat { get; set; }
```
ReubenBond: Should these be an enum? I'd prefer if we could somehow remove them, but maybe not for 2.0

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// </summary>
+        public string AdoInvariant { get; set; } = DEFAULT_ADONET_INVARIANT;
+
+        #region json serialization related settings
+        /// <summary>
+        /// Whether storage string payload should be formatted in JSON.
+        /// <remarks>If neither <see cref="UseJsonFormat"/> nor <see cref="UseXmlFormat"/> is set to true, then BinaryFormatSerializer will be configured to format storage string payload.</remarks>
+        /// </summary>
+        public bool UseJsonFormat { get; set; }
+        public bool UseFullAssemblyNames { get; set; }
```
ReubenBond: Is this a JSON-only option?

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: Yeah probably should be Invariant

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: We can revamp the serializer configuration after 2.0 rc, including improve configuration, make it possible to use other serializer. Just as @veikkoeeva mentioned 

But without doing all that, just make this an enum doesn't bring much benefits IMO.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
ReubenBond: My reasoning:
* Other lifecycle participants must be able to rely on certain things happening at certain stages - that's the point of having named stages
* There's no good reason for users to change this value

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
ReubenBond: I'm ok with leaving it for now

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: it is called that in AzureTableGrainStorageOptions too. we need to keep it consistent. And also it was called that before.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: Yeah this is definitely an advanced feature. Maybe we shouldn't make it this discoverable. @jason-bragg on what's his thought on this. He added this when he is doing the AzureTableGrainStorage. He may have his reasoning. 

But either way, I think we can merge this first. It is accumulating merge conflicts everyday:) 

If we decided that this is not a good idea, we can submit another pr to fix it from all GrainStorageProviders.

jason-bragg: The reason for allowing users to configure when grain storage starts is because some systems, like versioning, start very early in the lifecycle, and can be configured to use storage, while some storage systems themselves use grains which may trigger systems that use storage (leasing or transactions for instance.  Dependencies between systems are not always clear.  Allowing this to be configurable, along with the wide ranges between the predefined stages, give application developers a way out should they use a component that has unexpected dependency chains.  As with many of our settings, the default should be used unless their is reasons to change, so most users will never need to concern themselves with this, but for the cases where they need to change this, they can.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
jason-bragg: The json serialization settings we support at somewhat of a train wreck :/.  We should definitely take a pass at cleaning up/refactoring them, but I'm of the view that should be out of scope for this PR.

src/AdoNet/Orleans.Persistence.AdoNet/Storage/Provider/AdoNetGrainStorage.cs
```diff
@@ -136,83 +128,22 @@ public class AdoNetStorageProvider: IStorageProvider
...
-            var loggerFactory = providerRuntime.ServiceProvider.GetService<ILoggerFactory>();
-            this.logger = loggerFactory.CreateLogger<AdoNetStorageProvider>();
-            Name = name;
-
-            logger.Info((int)RelationalStorageProviderCodes.RelationalProviderInitProvider, $"Initialized storage provider: ServiceId={ServiceId} ProviderName={Name} Invariant={Storage.InvariantName} ConnectionString={Storage.ConnectionString}.");
+            this.options = options.Value;
+            this.providerRuntime = providerRuntime;
+            this.name = name;
+            this.logger = logger;
+            this.serviceId = providerRuntime.ServiceId.ToString();
```
ReubenBond: Pull this from `SiloOptions` instead

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
ReubenBond: I see your point. I'm concerned that breaking basic assumptions about initialization order introduces too much complexity. Debugging issues surrounding it would be difficult.

If it's fine for storage providers to sometimes start earlier, then we should always start them earlier.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// <inheritdoc cref="IConfigurationValidator"/>
+        public void ValidateConfiguration()
+        {
+            if (string.IsNullOrWhiteSpace(this.options.ConnectionString))
+            {
+                throw new OrleansConfigurationException($"Invalid AdoNetGrainStorageOptions for AdoNetGrainStorage {name}. ConnectionString is required.");
+            }
+            if (options.UseXmlFormat&&options.UseJsonFormat)
+            {
+                throw new OrleansConfigurationException($"Invalid AdoNetGrainStorageOptions for AdoNetGrainStorage {name}. Only one serializer and deserializer should be given.");
```
ReubenBond: These error messages should be more explicit, eg: `Invalid {nameof(AdoNetStorageOptions)} values for {nameof(AdoNetGrainStorage)} \"{name}\". {nameof(options.UseXmlFormat)} and {nameof(options.UseJsonFormat)} cannot both be set to true.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                this.options = options;
+                this.Name = OptionFormattingUtilities.Name<AdoNetGrainStorageOptions>(name);
+            }
+
+            public IEnumerable<string> Format()
+            {
+                return new List<string>()
+                {
+                    OptionFormattingUtilities.Format(nameof(this.options.ConnectionString),this.options.ConnectionString),
```
ReubenBond: This value must be redacted. Use `ConfigUtilities.RedactConnectionStringInfo`

veikkoeeva: We should take steps towards security...

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
jason-bragg: Since we're porting these from providers, and the original logic expected initialization at startup, I've advocated preserving the pattern of initializing on startup.  Since we're initializing on startup, we're taking part in the lifecycle and should probably afford users the ability to specify the stage for dependency issues.

Having said all that, in the future I'd prefer to use the lifecycle primarily for starting systems that are services (like the stream providers) or have heavy initialization (like transaction manager) rather than logic which just needs asynchronous initialization (like storage provider).  For things that simply require asynchronous initialization, I'd prefer we perform a lazy initialization, on demand, rather than plugging them into the lifecycle.  Barring circular dependencies (which can happen) lazy initialization is a simpler patter.  Basically if a system has no other dependencies, it probably shouldn't be initialized via the lifecycle, it can, but probably shouldn't.

These initial ports away from the provider model open the door for many improvements to our storage patterns, and I don't suspect that most should really need to take part in the lifecycle once fully fitted to the new patterns.  I suspect we can/will clean this up once we move grain storage to facets.  Since the public IGrainStorage interface is now independent of the initialization, this all becomes implementation details which only effect how a provider is configured.

In short, we can do better, this is just once step.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
jason-bragg: Defintely should be redacted, we can -never- log secrets.  However, I'm unsure ConfigUtilities.RedactConnectionStringInfo will redact from an ado connection string. 

@veikkoeeva, are you confident that the connection string format is consistent?

test/TesterInternal/TestRunners/GrainPersistenceTestRunner.cs
```diff
@@ -0,0 +1,296 @@
...
+using Orleans;
+using Orleans.TestingHost;
+using Tester;
+using UnitTests.GrainInterfaces;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace TestExtensions.Runners
+{
+    public class GrainPersistenceTestsRunner : OrleansTestingBase
```
jason-bragg: nice :)

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
veikkoeeva: Ideally these wouldn't be needed. As mentioned elsewhere, the idea is to allow any format to the database through the canonical interfaces and then just note which "type slot" to fit them in the database (binary, XML and JSON) where binary really swallos anything, but JSON or XML could be preferred for other reasons (e.g. in-storage processing).

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: we should do a pass on redacting all the connection string which were logged.... I might just do it in this pr

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: @veikkoeeva you haven't answered my question above. How is serializer picker feature suppose to work? I copied my original question below: 

Also, I have some question about this (de)serializer feature. Those (de)serializer are wrapped into their own canonical interfaces, and further put into a `DefaultRelationalStoragePicker`. This confuses me because it seems like only one (de)serializer is allowed. So why do we have a picker? And also there's a ` public IStorageSerializationPicker StorageSerializationPicker { get; set; }` on `AdoNetStorageProvider`, which seems to provide a way to modify the picker dynamically in runtime. I guess I need more information on how this feature are supposed to work.

xiazen: You are the last person touched the code, if you know background on this, much appreciation on sharing it.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: I will edit ConfigUtilities.RedactConnectionStringInfo, adding a default logic, which basic replace the connection string with "<--SNIP--> " if none of the secrectKeys are in the conneciton string. 
This should be enough to handle connection string format not consistent problem.

xiazen: And I'm of the opinion on we should just use the default logic, instead of creating complex logic on finding the secret key and just replace the secret key. I don't see much benefits of us doing that.

xiazen:
feedback addressed

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
veikkoeeva: Sorry, @xiazen I just missed it. It seems to be I need to be excessively pinged occassionally. :)

My original idea was that:
- Any (de)serializer can be wrapped into a canonical interface and loaded and handed over from elsewhere. That's the wrapping part. One complication is that there might be business or technical reasons to favour one kind of a storage layout over other (i.e. binary, JSON or XML), so it's indicated by tagging.

- The name `IStorageSerializationPicker` is not particulary good one. The idea is that this contains a rule to choose a (de)serializer with the parameters you see there. Since it's not possible to write a general rule to choose a (de)serializer, the idea is that by default one (de)serializer is instantiated with a rule to pick it. But it would allow one to use multiple (de)serializers with quite many parameters if the users could implement `IStorageSerializationPicker` by self.

There is one test exercising this idea by storing data (serializing) in one format and then reading (deserializing) it in other format. Practically switching the format. This could be used to:
1. Switch from Orleans binary to JSON, for instance.
2. But also implement custom logic to otherwise reshape data being read from the database (or being inserted to the database). This should make refactoring easier, or fixing "data bugs" easier.

In real deployment this could work so that in "deployment number n1" one deploys some customer transformation logic that checks the data to determine if it's in some format (say, some fields are present, indicating old format, so reshaped into new format consistent with new state shape) and once all is converted, during the next round of deployments just remove the customer refactoring code. The user has, naturally, also the option to do this as in-storage processing too.

But I'm not entirely sure how this should be exposed, what should be the naming of the classes be, the "surface" could be easier too (e.g. separate the reshaping logic into a lambda or some other interface from the picker logic).

Thinking in more general terms, keeping things consistent throughout Orleans, this is basically pipes-and-filters, VET(R)O, interceptors etc. and might be something to consider applied later elsewhere too (e.g. when deploying new grains, if not every can be handled by deactivating grains in old format and routing calls to grains accepting new format).

veikkoeeva: Originally I had an idea to inject these to these to the storage provider, but there were problems. I also tried a bit to "detach" the tests from the quite tight coupling the then system and work a bit on logic on how to ensure the system is ready for testing (initialize database etc.) and then actually to run the tests (so that they look OK also in the explorer).

I have ideas that should clean up the DB initialization logic considerable and introduce also construction injection with XUnit.NET to the tests making it cleaner still (I mentioned this in https://github.com/dotnet/orleans/issues/3969).

Maybe these help. If not, do ping me. I believe I'm off to bed soon, but I'll be reading again 8-9 hours before leaving to work. It's also end of sprint tomorrow too, so I'm likely more around in chat etc. too.

veikkoeeva:
@xiazen I'm not sure if GH pings on replies to code comments, so just to say I tried to address your question.

src/AdoNet/Orleans.Persistence.AdoNet/Storage/Provider/SiloBuilderExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// </summary>
+        public static ISiloHostBuilder AddAdoNetGrainStorage(this ISiloHostBuilder builder, string name, Action<OptionsBuilder<AdoNetGrainStorageOptions>> configureOptions = null)
+        {
+            return builder.ConfigureServices(services => services.AddAdoNetGrainStorage(name, configureOptions));
+        }
+
+        /// <summary>
+        /// Configure silo to use  AdoNet grain storage as the default grain storage.
+        /// </summary>
+        public static IServiceCollection AddAdoNetGrainStorage(this IServiceCollection services, Action<AdoNetGrainStorageOptions> configureOptions)
```
ReubenBond: Could we make these internal/private?

xiazen: I think this is a decision we should make as a team. whether to remove extension method on IServiceColleciton or not . Either way I think this is out of the scope of this PR.

src/AdoNet/Orleans.Persistence.AdoNet/Options/AdoNetGrainStorageOptions.cs
xiazen: Saw it :) in case you worry I didn't see this. Thanks for providing the details on how the feature was supposed to work. I meant to bring those to the issue I created too. But I kind of forgot to do it in the process. RC is a bit crazy. Will do it now.

PR #4023, opened 2018-02-13
Remove statistics table publishers

### LINKED ISSUE #3976: Refactor statistic and metric system

ReubenBond:
Resolves #3976 

It was decided in our meeting today that the `IStatisticsPublisher` interface is not adding value any longer. This change now simplifies the internal metrics tracking to textual logging + tracking via the telemetry system, which has support for NewRelic, ApplicationInsights, PerfCounters, File, Console, and Trace output.


src/AdoNet/Orleans.Statistics.AdoNet/MySQL-Statistics.sql
```diff
@@ -1,40 +0,0 @@
-CREATE TABLE OrleansStatisticsTable
```
xiazen: those are needed for other AdoNet tests :( I will add them back and rename them .

xiazen: sorry edit: those are needed for other AdoNet providers such as membership and persistence, Since liveness tests and persistence tests started failing and crying for this sql.

xiazen: or maybe just their tests,  I don't know. I will figure it out. no action item here. I'm just rambling on.

ReubenBond: These shouldn't be needed for anything... unless I removed something non-statistics related?

xiazen: sorry it was just one line you forget to remove in SqlServerStorageForTesting. and that caused tests failure because the test infrastructure is looking for it .

ReubenBond: Oh, thanks, Xiao

xiazen:
```
protected override string[] SetupSqlScriptFileNames
        {
            get { return new[] {
                    "SQLServer-Main.sql",
                    "SQLServer-Clustering.sql",
                    "SQLServer-Persistence.sql",
                    "SQLServer-Reminders.sql",
                    "SQLServer-Statistics.sql"
                };
            }
        }
```
I removed it in my pr. sorry for the exaggeration. Since from first sight, it looks like the removal of those files caused test failure. While it really just caused the test infra to failed to start

PR #4024, opened 2018-02-13
Add a configuration validator and options formatter for EndpointOptions

ReubenBond:
@benjaminpetit this might also need a default value for `AdvertisedIPAddress`, but I'm not sure if the default should be `Any`, `Localhost`, or something else.

### DIFF
--- a/src/Orleans.Runtime/Hosting/EndpointOptions.cs
+++ b/src/Orleans.Runtime/Hosting/EndpointOptions.cs
@@ -104,14 +104,12 @@ namespace Orleans.Hosting
 
         internal static IPEndPoint GetPublicSiloEndpoint(this EndpointOptions options)
         {
-            return options?.AdvertisedIPAddress != null
-                ? new IPEndPoint(options.AdvertisedIPAddress, options.SiloPort)
-                : null;
+            return new IPEndPoint(options.AdvertisedIPAddress, options.SiloPort);
         }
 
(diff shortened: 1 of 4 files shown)


src/Orleans.Runtime/Hosting/EndpointOptions.cs
```diff
@@ -76,31 +104,29 @@ public static class EndpointOptionsExtensions
 
         internal static IPEndPoint GetPublicSiloEndpoint(this EndpointOptions options)
         {
-            return new IPEndPoint(options.AdvertisedIPAddress, options.SiloPort);
+            return options?.AdvertisedIPAddress != null
```
benjaminpetit: I don't think this check is needed since the validator should throw before that.

We could also throw an exception here just in case, currently the silo cannot work without setting the `AdvertisedIPAddress`. Same comment for `GetPublicProxyEndpoint`

benjaminpetit:
> this might also need a default value for AdvertisedIPAddress, but I'm not sure if the default should be Any, Localhost, or something else.

I think we should do like in 1.5, by default take one non-localhost ip address that we found, with `ConfigUtilities.ResolveIPAddress`

ReubenBond:
@benjaminpetit sounds good to me. This is just a drive-by PR, since I hit the issue of not having a default endpoint value in another PR

sergeybykov:
@ReubenBond Please rebase.

PR #4025, opened 2018-02-13
Add a configuration validator for cluster membership providers

ReubenBond:
Validates that a clustering provider has been configured on the client and silo.

### DIFF
--- a/src/Orleans.Core/Configuration/Validators/ClientClusteringValidator.cs
+++ b/src/Orleans.Core/Configuration/Validators/ClientClusteringValidator.cs
@@ -1,0 +1,36 @@
+﻿using System;
+
+using Microsoft.Extensions.DependencyInjection;
+
+using Orleans.Messaging;
+using Orleans.Runtime;
+
+namespace Orleans.Configuration.Validators
+{
+    internal class ClientClusteringValidator : IConfigurationValidator
+    {
+        internal const string ClusteringNotConfigured =
+            "Clustering has not been configured. Configure clustering using one of the clustering packages, such as: "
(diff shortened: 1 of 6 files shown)


src/Orleans.Core/Configuration/Validators/ClientClusteringValidator.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Orleans.Configuration.Validators
+{
+    internal class ClientClusteringValidator : IConfigurationValidator
+    {
+        internal const string ClusteringNotConfigured =
+            "Clustering has not been configured. Configure clustering using one of the clustering packages, such as: "
+            + "\n  * Microsoft.Orleans.Clustering.AzureStorage for Microsoft Azure Storage"
+            + "\n  * Microsoft.Orleans.Clustering.AdoNet for ADO.NET systems such as SQL Server, MySQL, PostgreSQL, and Oracle"
+            + "\n  * Microsoft.Orleans.Clustering.DynamoDB"
+            + "\n  * Microsoft.Orleans.Clustering.ServiceFabric"
```
sergeybykov: ZooKeeper and Consul too?

ReubenBond: I can add them, I thought the list might get too long

sergeybykov: Never mind then.

ReubenBond: Well, there's no real reason to be compact here. I'll update it

ReubenBond: It's more important that users see something they can use

PR #4027, opened 2018-02-13
Remove builder CreateDefault methods

ReubenBond:
As discussed in our meeting yesterday. We can add this sort of thing back using an extension method. That will better support the 'generic host builder' model.


ReubenBond:
Rebased

ReubenBond:
Extension methods can be added in a later PR when we work out what exactly we want them to do

PR #4029, opened 2018-02-13
Rename Use*Membership extensions to Use*Clustering

ReubenBond:
Removes extensions from `IServiceCollection`
Renames client and silo methods so that they are symmetric (`UseAzureStorageClustering()`);


sergeybykov:
Rebase please.

PR #4032, opened 2018-02-13
Add more logging to AzureTableGrainStorage init, Add ConfiguratorValidator

xiazen:
- Add more logging to AzureTableGrainStorage init to improve debugging experience
- Add Configuration validator to  `AzureTableGrainStorageOptions`
- rename `DataConnectionString` to be `ConnectionString`. To be consistent with the `ConnectionString` in AzureGatewaylistprovider and AzureMembershipProvier

### DIFF
--- a/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorage.cs
+++ b/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorage.cs
@@ -462,9 +462,9 @@ namespace Orleans.Storage
             try
             {
                 this.logger.LogInformation((int)AzureProviderErrorCode.AzureTableProvider_InitProvider, $"AzureTableGrainStorage {name} initializing: {this.options.ToString()}");
-                this.logger.LogInformation((int)AzureProviderErrorCode.AzureTableProvider_ParamConnectionString, $"AzureTableGrainStorage {name} is using DataConnectionString: {ConfigUtilities.RedactConnectionStringInfo(this.options.DataConnectionString)}");
+                this.logger.LogInformation((int)AzureProviderErrorCode.AzureTableProvider_ParamConnectionString, $"AzureTableGrainStorage {name} is using DataConnectionString: {ConfigUtilities.RedactConnectionStringInfo(this.options.ConnectionString)}");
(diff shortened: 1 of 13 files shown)


src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorage.cs
```diff
@@ -44,16 +45,18 @@ public class AzureTableGrainStorage : IGrainStorage, IRestExceptionDecoder, ILif
 
         private const string BINARY_DATA_PROPERTY_NAME = "Data";
         private const string STRING_DATA_PROPERTY_NAME = "StringData";
-
+        private string name;
         /// <summary> Default constructor </summary>
-        public AzureTableGrainStorage(AzureTableStorageOptions options, SerializationManager serializationManager, IGrainFactory grainFactory, ITypeResolver typeResolver, ILoggerFactory loggerFactory)
+        public AzureTableGrainStorage(AzureTableStorageOptions options, SerializationManager serializationManager, IGrainFactory grainFactory, ITypeResolver typeResolver, ILoggerFactory loggerFactory, string name)
```
benjaminpetit: NIT:  I think the name should be the first parameter, it's more clear
NIT²: Could you let a space between the latest member and the ctor? You are doing the same in the validator below
NIT³: When there is a lot of param like here, I think it is better to break them in multiple lines. It's more readable, and avoid merge conflicts sometimes

src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorageOptions.cs
```diff
@@ -45,6 +45,27 @@ public class AzureTableStorageOptions
...
+        private readonly string name;
+        public AzureTableGrainStorageOptionsValidator(AzureTableStorageOptions options, string name)
+        {
+            this.options = options;
+            this.name = name;
+        }
+
+        public void ValidateConfiguration()
+        {
+            if (string.IsNullOrEmpty(this.options.ConnectionString))
```
benjaminpetit: Should we parse it to check that it is in fact an azure connection string? A simple call to `CloudStorageAccount.TryParse`maybe?

xiazen: good idea

benjaminpetit: I think we should also validate that the `TableName` is valid too

xiazen:
tests passed. Pr feedback addressed

Issue #4015, opened 2018-02-12
Logger is not used currently in Orleans.Runtime.Dispatcher


BarakHirsch:
There is a bug in the usage of the logger in Orleans.Runtime.Dispatcher and possibly in other places.
For example, the LoggerExtensions.Info method signature is:
`internal static void Info(this Logger logger, ErrorCode errorCode, string format, params object[] args)`
But it is used as follows
`logger.Info(ErrorCode.Dispatcher_Intermediate_GetOrCreateActivation, String.Format("Intermediate NonExistentActivation for message {0}", message), ex);`
So the params are as follows:
ErrorCode errorCode =  ErrorCode.Dispatcher_Intermediate_GetOrCreateActivation
string format = String.Format("Intermediate NonExistentActivation for message {0}", message)
params object[] args = ex

This means that we try to call string.Format on the result of this string format and the exception as args.
In most cases this is not a major problem as it just means that the exception is never logged.
It is a big issue if the grain key contains curly braces(maybe not the best idea, but it's the current implementation), the call fails over System.FormatException.
This is the call stack:

`System.FormatException: Input string was not in a correct format.    at System.Text.StringBuilder.AppendFormat(IFormatProvider provider, String format, Object[] args)    at System.String.Format(IFormatProvider provider, String format, Object[] args)    at Orleans.Runtime.LoggerImpl.WriteLogMessage(Int32 errorCode, Severity sev, String format, Object[] args, Exception exception)    at Orleans.Runtime.LoggerExtensions.Info(Logger logger, ErrorCode errorCode, String format, Object[] args)    at Orleans.Runtime.Dispatcher.ReceiveMessage(Message message)`

xiazen:
sorry about this. I will try to do a pass to fix this. Those should use String tempalte, instead of String.Format

xiazen:
submitted a PR #4040 to fix it

xiazen:
pr got merged. issue fixed. gonna close the issue. feel free to reopen it if any further issue occurs

PR #4040, opened 2018-02-14
Migrate to use string template, instead of String.Format

xiazen:
Migrate part of our logging message to use string template, instead of String.Format 
Reasons:
- String.Format is error prone. You may miss one arg or give extra arg. While string template doesn't have that problem
- Easy to use the logger extension method in a wrong way. See #4015 for more info.


xiazen:

PR #4041, opened 2018-02-15
BUG: Some MembershipOptions default values were not being set causing errors.

jms69:
How Found:
Set silo to startup via code without using legacy mode.  Exception thrown due to TimeSpan being set to 0.

SafeRandom.NextTimeSpan timeSpan must be a positive number.
Parameter name: timeSpan
Actual value was 00:00:00.

Code that fails on MembershipOracle.Start()
```
                // randomly delay the startup, so not all silos write to the table at once.
                // Use random time not larger than MaxJoinAttemptTime, one minute and 0.5sec*ExpectedClusterSize;
                var random = new SafeRandom();
                var maxDelay = TimeSpan.FromMilliseconds(500).Multiply(this.membershipOptions.ExpectedClusterSize);
                maxDelay = StandardExtensions.Min(maxDelay, StandardExtensions.Min(this.membershipOptions.MaxJoinAttemptTime, TimeSpan.FromMinutes(1)));
                var randomDelay = random.NextTimeSpan(maxDelay);
                await Task.Delay(randomDelay);
```

ExpectedClusterSize and MaxJoinAttempTime all had the value of zero before the fix.

### DIFF
--- a/src/Orleans.Core/Configuration/Options/MembershipOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/MembershipOptions.cs
@@ -30,38 +30,39 @@ namespace Orleans.Configuration
         /// <summary>
         /// The number of seconds to periodically probe other silos for their liveness or for the silo to send "I am alive" heartbeat  messages about itself.
         /// </summary>
-        public TimeSpan ProbeTimeout { get; set; }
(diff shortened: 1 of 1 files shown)


ReubenBond:
Thank you, @jms69!

Issue #634, opened 2015-07-22
Add Streams on top of relational database


veikkoeeva:
Implementing an [Orleans Stream Provider](http://dotnet.github.io/orleans/Orleans-Streams/) on top of a relational backend has come up as a subject a few times in Gitter, in https://github.com/dotnet/orleans/issues/623 and is connected to https://github.com/dotnet/orleans/issues/255. We could open an official issue about it as queuing, and streams, is a common scenario and especially on-premises development happens largely on top of relational backends. The following ideas are applicable to all relational databases, but to keep this simple, let’s assume SQL Server for the time being.

As mentioned in https://github.com/dotnet/orleans/issues/623 already, there’s an interesting article [Using tables as Queues](http://rusanu.com/2010/03/26/using-tables-as-queues/) by _Remus Rusanu_. The ideas presented about FIFO queues look like sound plan to implement [Orleans Streams](http://dotnet.github.io/orleans/Orleans-Streams/) of interest are the ordering guarantees. However, it might be prudent to explore the use of [ANSI SQL 2003 SEQUENCE](https://www.simple-talk.com/sql/learn-sql-server/sql-server-sequence-basics/). This feature is widely available in various relational databases and also on [SQL Server 2012](https://msdn.microsoft.com/en-us/library/ff878091.aspx) and later.

As of myself, I don't have a more concrete plan or schedule on starting this (though I've expressed interest), but a rough idea for the `SEQUENCE` part could be to mimick a ring buffer on a table. The table could something like following (might be inaccurate with regard to the needs of Orleans Streams, but the gist of it)

```
CREATE TABLE StreamRingBuffer
(
    [SequenceToken] BIGINT NOT NULL,
    [Slot] BIGINT NOT NULL,
    [StreamId] NVARCHAR(256),
    [EnqueuedOn] DATETIME2(7) NOT NULL,
    [ReferenceCount] TINYINT NOT NULL,
    [Payload] VARBINARY(MAX)
);
```

Here sequence token would be generated like 

```
CREATE SEQUENCE EnqueueSequence AS BIGINT START WITH 1 INCREMENT BY 1 CACHE 500000;
SET @sequenceToken = NEXT VALUE FOR EnqueueSequence;
SET @slot = @sequenceToken % 500000;
```

[...]

PR #4042, opened 2018-02-15
Persistent stream provider to lifecycle

jason-bragg:
Pushing this PR out to get eyes on it, but it's not passing testing yet.  Still WIP.

Since persistent stream provider is the base for most of our steam providers, we couldn't refactor these one at a time.

Following stream providers now use options and lifecycle.
* Azure Queue
* EventHub
* Sqs
* PubSub
* Memory
* Generator
* EventDataGenerator
* Various test providers

TBD:
Fix tests.
Move options to Orleans.Configuration namespace.
Replace DataConnectionStream with ConnectionString in options to conform to other options.
Possibly validators, but would like to push that to separate PR, given the size of this change.

NOTE:  Persistent stream providers and various adapters are complex and patterns can be improved now that we can inject dependencies, but current patterns are best effort port.  Please note where further refactoring in the future is warranted.

Future Refactor:
* Data Adapters
* Inject checkpointing
* Inject generators (for generated streams)
* Inject failure handlers
* Possible adapter or provider builders


ReubenBond:
The last big PR before we tag an RC 🎉

jason-bragg:
Side note:  There is lots of crazy in all this, but the updated pattern is much cleaner and gives us a good foundation for future improvements.

@galvesribeiro, can you spare a glance at the external stream providers you've added.  Maybe plan in some time to help me test them latter in the week?  It would be much appreciated! :)

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        public static IServiceCollection AddClusterClientSqsStreams(this IServiceCollection services, string name, Action<SqsStreamOptions> configureOptions)
+        {
+            return services.AddClusterClientSqsStreams(name, ob => ob.Configure(configureOptions));
+        }
+
+        /// <summary>
+        /// Configure cluster client to use SQS persistent streams.
+        /// </summary>
+        public static IServiceCollection AddClusterClientSqsStreams(this IServiceCollection services, string name,
```
ReubenBond: I think we should *not* add any extension methods which hang from `IServiceCollection` and we should endeavor to remove the existing ones.

galvesribeiro:
@jason-bragg absolutely. Will grab your branch here and check it. Meanwhile, feel free to ping me on skype of gitter to talk about it.

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System.Collections.Generic;
+using Microsoft.Extensions.Options;
+
+namespace Orleans.Hosting
+{
+    public class SqsStreamOptions : PersistentStreamOptions
```
ReubenBond: Move options to the `Orleans.Configuration` namespace

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+                this.parentFormatter = new PersistentStreamOptionsFormatter(options);
+            }
+
+            public IEnumerable<string> Format()
+            {
+                List<string> formatted = this.parentFormatter.FormatSharedOptions();
+                formatted.AddRange(new[]
+                {
+                    OptionFormattingUtilities.Format(nameof(this.options.DeploymentId), this.options.DeploymentId),
+                    OptionFormattingUtilities.Format(nameof(this.options.DeploymentId), "<-Redacted->"),
```
ReubenBond: Error here, duplicated line. I think the latter is supposed to be `DataConnectionString`. We can use the existing `ConfigUtils.RedactConnectionStringInfo` method as in other providers - unless you feel it isn't adequate?

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System.Collections.Generic;
+using Microsoft.Extensions.Options;
+
+namespace Orleans.Hosting
+{
+    public class SqsStreamOptions : PersistentStreamOptions
+    {
+        public string DeploymentId { get; set; }
+
+        public string DataConnectionString { get; set; }
```
ReubenBond: Since this is provider-specific, could we call this `ConnectionString` instead?

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System.Collections.Generic;
+using Microsoft.Extensions.Options;
+
+namespace Orleans.Hosting
+{
+    public class SqsStreamOptions : PersistentStreamOptions
+    {
+        public string DeploymentId { get; set; }
```
ReubenBond: Should this value be configured here or should it come from some other, more central place? My understanding is that this should be replaced with `SiloOptions.ClusterId`, since we smooshed `DeploymentId` and `ClusterId` together here: https://github.com/dotnet/orleans/pull/3728

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
galvesribeiro: I think it should not be standardazied as connection string on all providers... others don't have it but instead, they ask for something like endpoint and API key... for those cases, our options should reflect that with 2 properties just like I did with the CosmosDB providers.

This `DataConnectionString` there was a hack back in the day just to "follow the standard". 

I'll submit a PR to @jason-bragg branch changing that as well

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
galvesribeiro: Yes, this value should be dropped from there and inject the `SiloOptions`.  The newer providers are already doing that. Will do that on a PR.

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
ReubenBond: Because otherwise we will have methods which are ambiguous between client & silo & SignalR & ASP.NET. I'm thinking about the future, but the issue is present here today: we need to name the methods differently for client and silo instead of having a nice simple "AddSqsStreams" method.

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
ReubenBond: Yes, specific names per-provider is fine. Whatever is the typical lingo for that system is what we should use.

jason-bragg:
@galvesribeiro, I've done very little testing with this so far.  I should have it stable enough for you to try by mid-day tomorrow.  If you want to play with it before then, you're welcome to do so, but I'd expect there to be issues outside the providers I requested your help on.  Just don't want to use up your patience on issues I should be solving. :)

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
jason-bragg: Good catch on the dupe, tnx.
The redact utility, by my understanding, expects an azure connection string, and knows how to pull the secrets out of that.  I don't know if we can trust it to redact secrets from other service connection strings.  It would be valuable to be able to strip out secrets for all of these, but I don't have the time to investigate each format, so I've been just placing the generic <-redact-> line in there.

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
galvesribeiro: I see... Well, other framework (including asp.net) does that IIRC. However, I think we should have diff assemblies for client and silo for the providers. For example, in Membership, makes no sense to bring the Membership table to the client, but instead it should be only have the gateway. So `Orleans.Clustering.XXXX` and `Orleans.Clustering.XXXX.Client`(or `Gateway`). I know, extra nuget package to maintain, but the benefits are:

1. Avoid this confusion with extensions
2. Avoid having 2 methods for client and silo
3. In the case of clustering, remove the dependency on the runtime/silo assemblies on teh client.

galvesribeiro:
@jason-bragg Ok. Take your time. Once you feel it is ready let me know and I'll amend a PR to your branch. I will run all tests as I've essentially ported all providers to .net core back in day and not just the ones I introduced and then will report back.

veikkoeeva:
I'll page @joeybrown, since he was working on #634.

xiazen:
woke up today and a BIG present is awaiting on GitHub 
and .. how to insert emoji here ?! lol 
good work @jason-bragg !

jason-bragg:
That was my valentines day present to the team/community.. I'm so romantic, as you can all see.

src/AWS/Orleans.Streaming.SQS/Hosting/SiloBuilderExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        /// <summary>
+        /// Configure silo to use SQS persistent streams.
+        /// </summary>
+        public static IServiceCollection AddSiloSqsStreams(this IServiceCollection services, string name,
+            Action<OptionsBuilder<SqsStreamOptions>> configureOptions = null)
+        {
+            return services.TryConfigureFormatterResolver<SqsStreamOptions, SqsStreamOptionsFormatterResolver>()
+                           .AddSiloPersistentStreams<SqsStreamOptions>(name, SQSAdapterFactory.Create, configureOptions);
```
xiazen: missing `ConfigureNamedOptionForLogging`

jason-bragg:
Heads up, for anyone tinkering with this, PR is hilariously broken, as provider initialization needs be run in Orleans scheduling context, which is still under development.  I'll update the thread when there is a faint chance this works :/

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingManager.cs
```diff
@@ -30,7 +30,7 @@ internal class PersistentStreamPullingManager : SystemTarget, IPersistentStreamP
         private readonly IQueueAdapterCache queueAdapterCache;
         private IStreamQueueBalancer queueBalancer;
         private readonly IQueueAdapterFactory adapterFactory;
-        private PersistentStreamProviderState managerState;
+        private RunState managerState;
```
xiazen: RunState? should just state be sufficient here ? What does run imply here ?

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -190,15 +133,18 @@ public LeaseBasedQueueBalancer(IServiceProvider serviceProvider, ISiloStatusOrac
...
+        public override Task Initialize(string strProviderName, IStreamQueueMapper queueMapper, TimeSpan siloMaturityPeriod)
         {
             if (queueMapper == null)
             {
                 throw new ArgumentNullException("queueMapper");
             }
-            var balancerConfig = new LeaseBasedQueueBalancerConfig(providerConfig);
-            this.leaseProvider = this.serviceProvider.GetRequiredService(balancerConfig.LeaseProviderType) as ILeaseProvider;
-            this.leaseLength = balancerConfig.LeaseLength;
+            var options = this.serviceProvider.GetServiceByName<LeaseBasedQueueBalancerOptions>(strProviderName)
```
xiazen: we should have providerBuilder and provider lifecycle. :) 

There's too many components in streams. The complexity can benefits from lifecycle and builder pattern, in both perspective of configuring and start and shutdown sequence. 

Some of those are managed by Silo(Client)ProviderRuntime, some are managed by the PersistentStreamProvider. Those unnecessary coupling partially result in bad consequence, such as streaming infra has to live in core. 

Just throwing ideas out there, not necessary the scope of this PR.

src/OrleansProviders/Streams/Generator/GeneratorAdapterFactory.cs
```diff
@@ -77,36 +83,33 @@ public class GeneratorAdapterFactory : IQueueAdapterFactory, IQueueAdapter, IQue
...
-            this.telemetryProducer = svcProvider.GetService<ITelemetryProducer>();
+            this.receivers = new ConcurrentDictionary<QueueId, Receiver>();
             if (CacheMonitorFactory == null)
                 this.CacheMonitorFactory = (dimensions, telemetryProducer) => new DefaultCacheMonitor(dimensions, telemetryProducer);
             if (this.BlockPoolMonitorFactory == null)
                 this.BlockPoolMonitorFactory = (dimensions, telemetryProducer) => new DefaultBlockPoolMonitor(dimensions, telemetryProducer);
             if (this.ReceiverMonitorFactory == null)
                 this.ReceiverMonitorFactory = (dimensions, telemetryProducer) => new DefaultQueueAdapterReceiverMonitor(dimensions, telemetryProducer);
-            if (adapterConfig.GeneratorConfigType != null)
+            generatorConfig = this.serviceProvider.GetServiceByName<IStreamGeneratorConfig>(this.Name);
```
xiazen: constructor injection instead of this.

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
ReubenBond: That's quite fair - I simply don't know how to ensure that we can redact SQS keys correctly, or if we even should. Redacting the entire thing is safer

jason-bragg:
@galvesribeiro I think core is solid enough for you to test against in the latest PR.  Still has test failures, and hacks i'm resolving, but it's mostly their I think..

galvesribeiro:
@jason-bragg sure, will do my changes in a PR early tomorrow. Probably when you come to office it should be there. Thanks!

xiazen:
@jason-bragg  Submitted a [PR ](https://github.com/jason-bragg/orleans/pull/3) to your branch regarding migrate SMS stream provider to use option and lifecycle. 
Nothing urgent there, just wanted to get that out before I left tomorrow. That pr will likely be resubmitted as a pr against master after your branch has been merged .

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingManager.cs
jason-bragg: it replaces PersistentStreamProviderState and tracks if the provider is running or not.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: I'm not convinced of the lifecycle aspect, but it's definitely worth consideration.  In the description I've started a list of further refactoring we may want to do in the future.  Please feel free to edit the description to add more suggestions.

src/OrleansProviders/Streams/Generator/GeneratorAdapterFactory.cs
jason-bragg: generator config is not required, it can be provided at runtime, hence constructor injection will not work :/

src/Orleans.Core/Streams/PersistentStreams/PersistentStreamProvider.cs
```diff
@@ -36,82 +30,66 @@ public enum PersistentStreamProviderCommand
...
+
+            this.Name = name;
+            this.options = options;
+            this.runtime = runtime.ServiceProvider.GetRequiredService<IStreamProviderRuntime>();
+            this.runtimeClient = runtime.ServiceProvider.GetRequiredService<IRuntimeClient>();
+            this.serializationManager = serializationManager;
+            this.logger = logger;
+        }
+
+        private async Task Init(CancellationToken token)
```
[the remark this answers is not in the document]
jason-bragg: Most of the adapter interfaces have not changed.  The main difference is that they'd need to move to constructor injection as nothing is passed in during the init, and add an options class for their adapter.  These are breaking changes, but not dramatic ones (imo).

src/Azure/Orleans.Streaming.AzureStorage/Hosting/SiloBuilderExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        /// <summary>
+        /// Configure silo to use azure queue persistent streams.
+        /// </summary>
+        public static IServiceCollection AddSiloAzureQueueStreams<TDataAdapter>(this IServiceCollection services, string name,
+            Action<OptionsBuilder<AzureQueueStreamOptions>> configureOptions = null)
+           where TDataAdapter : IAzureQueueDataAdapter
+        {
+            return services.TryConfigureFormatterResolver<AzureQueueStreamOptions, AzureQueueStreamOptionsFormatterResolver>()
```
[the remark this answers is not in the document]
jason-bragg: it was being done in AddPersistentStreams, but that's not the right place.  They should be there now.

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            this.optionsSnapshot = optionsSnapshot;
+        }
+
+        public IOptionFormatter<SqsStreamOptions> Resolve(string name)
+        {
+            return new Formatter(name, optionsSnapshot.Get(name));
+        }
+
+        public class Formatter : IOptionFormatter<SqsStreamOptions>
```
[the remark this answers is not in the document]
jason-bragg: Subclass, so scoped with proper name.

src/AWS/Orleans.Streaming.SQS/Streams/SQSAdapterFactory.cs
```diff
@@ -90,5 +74,13 @@ public Task<IStreamFailureHandler> GetDeliveryFailureHandler(QueueId queueId)
         {
             return StreamFailureHandlerFactory(queueId);
         }
+
+        public static SQSAdapterFactory Create(IServiceProvider services, string name)
+        {
+            IOptionsSnapshot<SqsStreamOptions> streamOptionsSnapshot = services.GetRequiredService<IOptionsSnapshot<SqsStreamOptions>>();
+            var factory = ActivatorUtilities.CreateInstance<SQSAdapterFactory>(services, name, streamOptionsSnapshot.Get(name));
+            factory.Init();
```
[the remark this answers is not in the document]
jason-bragg: Init is not async, so no reason I can see to link adapters to lifecycle, they are just pluggable components.

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
jason-bragg: Will discuss then address in separate PR based on consensus.

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
jason-bragg: Looking at the RedactConnectionStringInfo logic, I'm not confident it's safe.  If a connection string has one secret in the list, and one not in the list, it's secrets will be logged.  I'm going to stick with the <--SNIP--> for now.  Developers/maintainers of a component can develop component specific redactors if it's important.  IMO it's better to log less info than risk logging users secrets.

galvesribeiro:
Sorry for my delay. The PR with the AWS updates were made to your branch @jason-bragg (https://github.com/jason-bragg/orleans/pull/4).

I've also ran other provider tests and they looks fine so far.

jason-bragg:
Ready (assuming tests pass)

galvesribeiro:
@jason-bragg did you saw the PR on your branch?

xiazen:

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿
+namespace Orleans.Configuration
+{
+    public class SqsStreamOptions : PersistentStreamOptions
+    {
+        public string ClusterId { get; set; }
+
+        [Redact]
+        public string ConnectionString { get; set; }
```
ReubenBond: Should this be `[RedactConnectionString]`? Xiao made a change to redact the entire thing by default if no redactions were made.

BTW, @xiazen, that change also means that `UseDevelopmentStorage=true` would be redacted, but better to be safe than sorry.

src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueAdapterFactory.cs
```diff
@@ -5,85 +5,54 @@
...
         /// Gets the serialization manager.
         /// </summary>
-        public SerializationManager SerializationManager { get; private set; }
+
+        protected SerializationManager SerializationManager { get; }
 
         /// <summary>
         /// Application level failure handler override.
         /// </summary>
         protected Func<QueueId, Task<IStreamFailureHandler>> StreamFailureHandlerFactory { private get; set; }
```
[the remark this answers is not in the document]
jason-bragg: Same as previous version, must subclass adapter factory.  Getting it from DI would be better, but refactoring these patterns to take full advantage of DI is beyond the scope of this PR.  Main goal here is to move all settings to options and initialize/start provider via the lifecycle.  Further refactoring can be performed in future iterations.

src/Azure/Orleans.Streaming.EventHubs/Providers/EventDataGeneratorStreamProvider/EventDataGeneratorAdapterFactory.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        public EventDataGeneratorAdapterFactory(string name, EventDataGeneratorStreamOptions options, IServiceProvider serviceProvider, SerializationManager serializationManager, ITelemetryProducer telemetryProducer, ILoggerFactory loggerFactory)
+            : base(name, options, serviceProvider, serializationManager, telemetryProducer, loggerFactory)
+        {
+            this.ehGeneratorOptions = options;
+        }
+
+        public override void Init()
+        {
+            this.CheckpointerFactory = partition => Task.FromResult<IStreamQueueCheckpointer<string>>(NoOpCheckpointer.Instance);
```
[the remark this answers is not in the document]
jason-bragg: Agreed.  Builder pattern or injection would be preferred.  Such refactors are out of scope for this PR.

src/Orleans.Core/Providers/LegacyProviderConfigurator.cs
```diff
@@ -102,8 +98,6 @@ public virtual void Participate(TLifecycle lifecycle)
             {
                 this.initStage = this.config.GetIntProperty(LegacyProviderConfigurator.InitStageName, this.defaultInitStage);
                 lifecycle.Subscribe(this.initStage, this.Init, this.ProviderClose);
-                this.startStage = this.config.GetIntProperty(LegacyProviderConfigurator.StartStageName, this.defaultStartStage);
-                lifecycle.Subscribe(this.startStage, this.Start, this.StreamProviderClose);
```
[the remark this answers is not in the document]
jason-bragg: The "Start" call was only supported by the persistent stream provider, which is no longer an IProvider.  All custom providers that used this were likely specializations of the Persistent stream provider with custom queue adapters.

src/Orleans.Core/Streams/PersistentStreams/PersistentStreamProvider.cs
```diff
@@ -36,82 +30,66 @@ public enum PersistentStreamProviderCommand
...
-                || this.myConfig.PubSubType == StreamPubSubType.ExplicitGrainBasedOnly)
+            if (String.IsNullOrEmpty(name)) throw new ArgumentNullException(nameof(name));
+            if (options == null) throw new ArgumentNullException(nameof(options));
+            if (runtime == null) throw new ArgumentNullException(nameof(runtime));
+            if (serializationManager == null) throw new ArgumentNullException(nameof(serializationManager));
+            if (logger == null) throw new ArgumentNullException(nameof(logger));
+
+            this.Name = name;
+            this.options = options;
+            this.runtime = runtime.ServiceProvider.GetRequiredService<IStreamProviderRuntime>();
```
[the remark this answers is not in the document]
jason-bragg: Can't, IStreamProviderRuntime is internal.

src/AWS/Orleans.Streaming.SQS/Streams/SqsStreamOptions.cs
jason-bragg: I don't think so.  I'm unconvinced RedactConnectionString is safe for unknown formats.  By the logic of the call, if any of the patterns match the string is logged with the recognized patterns redacted.  This means if a connection string has a matched, and an unmatched secret, the unmatched secret will logged. which is not acceptable.

ReubenBond: ok, no worries. In the future we can create a version which only includes known-good portions of the connection string, instead. We can wait until someone complains, if they ever do.

jason-bragg:
Functionals seem to be failing again due to test timeout.  Pushing a possible optimization so lifecycle to see if it helps.  If it doesn't I'll role it back.

xiazen:
I kicked out a test build in vso to run azure tests, which were skipped here. Also hoping the time out in vso is longer, so the build won't fail due to test timeout

jason-bragg:
Functionals timed out again. Trying something different.  This time I'm running silo lifecycle actions outside the lifecycle scheduling context.

jason-bragg:
hmm.. functionals seem say they took 30~ish minutes..  I don't trust this.

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
rikbosch: Maybe the extension methods should at least live in their own namespace? One for the clientbuilder and one for the silo? Better for discoverability.

jason-bragg:
Removed lifecycle perf fix, and rebased.  Should be good to merge after tests run.

PR #4049, opened 2018-02-15
Add a generic implementation of IOptionFormatter and IOptionFormatterResolver using reflection

benjaminpetit:
Instead of writing all the boilerplate code, we could use almost all the time a simple implementation that relies on reflection.

If everyone agree with this implementation, I will start to add missing options and DELETE the existing implementation (if they produce the same output as this one of course)


benjaminpetit:
@dotnet-bot test bvt

benjaminpetit:
@dotnet-bot test this

benjaminpetit:
@dotnet-bot test bvt functional

benjaminpetit:
@dotnet-bot test bvt

benjaminpetit:
@dotnet-bot test functional

src/Orleans.Core/Configuration/OptionLogger/GenericOptionFormatter.cs
```diff
@@ -0,0 +1,65 @@
...
+        internal GenericOptionsFormatter(string name, T options)
+        {
+            this.options = options;
+            this.Name = OptionFormattingUtilities.Name<T>(name);
+        }
+
+        public IEnumerable<string> Format()
+        {
+            return typeof(T)
+                .GetProperties()
```
jason-bragg: I think we only want properties which are public, and have public getters and setters.  Something like

.GetProperties(BindingFlags.Public).where(prop => prop.GetGetMethod() != null && prop.GetSetMethod() != null)

src/Orleans.Core/Configuration/OptionLogger/GenericOptionFormatter.cs
```diff
@@ -0,0 +1,65 @@
...
+        private string FormatProperty(PropertyInfo property)
+        {
+            var name = property.Name;
+            var value = property.GetValue(this.options);
+            var redactAttribute = property.GetCustomAttribute<RedactAttribute>(inherit: true);
+
+            return OptionFormattingUtilities.Format(
+                name, 
+                redactAttribute != null 
+                    ? redactAttribute.Redact((string) value)
```
jason-bragg: I think we need to ToString() the value rather than cast it.

src/Orleans.Core/Configuration/OptionLogger/GenericOptionFormatter.cs
```diff
@@ -0,0 +1,65 @@
...
+        private IOptionsSnapshot<T> optionsSnapshot;
+
+        public GenericOptionsFormatterResolver(IOptionsSnapshot<T> optionsSnapshot)
+        {
+            this.optionsSnapshot = optionsSnapshot;
+        }
+
+        public IOptionFormatter<T> Resolve(string name)
+        {
+            return new GenericOptionsFormatter<T>(name, Options.Create(optionsSnapshot.Get(name)).Value);
```
jason-bragg: y are we creating an option only to get it's value?

src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
```diff
@@ -27,6 +27,15 @@ public static class OptionConfigureExtensionMethods
             return services;
         }
 
+        /// <summary>
+        /// configure option formatter for <typeparam name="TOptions"/>
+        /// </summary>
+        public static IServiceCollection ConfigureFormatter<TOptions>(this IServiceCollection services)
+            where TOptions : class, new()
+        {
+            return services.ConfigureFormatter<TOptions, GenericOptionsFormatter<TOptions>>();
```
jason-bragg: If this is our default, I don't think we need ass it to these extension functions, only to our default services.

services.AddSingleton(typeof(IOptionFormatter<>), typeof(GenericOptionsFormatter<>));
services.AddSingleton(typeof(IOptionFormatterResolver<>), typeof(GenericOptionsFormatterResolver<>));

This allows any IOptionsFormater mapped to an IOptionsFormatter<O> to be automatically resolved to the generic formatter unless an impl is specifically provided.

test/Tester/LogFomatterTests.cs
```diff
@@ -284,6 +372,17 @@ private TestLoggerFactory BuildNamedOptionsExpectedResult()
             return expected;
         }
 
+        private class TestOptionsWithSecrets
```
jason-bragg: suggest including a wider variety of property types, including things like ints and lists to see what happens.

src/Orleans.Core/Configuration/OptionLogger/GenericOptionFormatter.cs
xiazen: lol you did this thing also before . optionsSnapshot.Get(name) is enough

jason-bragg:
@xiazen, Yeah, I'll follow up, assuming @ReubenBond doesn't get there first..

test/Tester/LogFomatterTests.cs
benjaminpetit: string are already tested, not sure if it is really necessary to test other types, as long as they have a string representation...

For enumerable type that's an interesting question. For now I left them as is, since in our codebase sometimes we concatenate a full list to a string, sometimes we do something else...

src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
```diff
@@ -44,6 +55,20 @@ public static class OptionConfigureExtensionMethods
...
         }
 
+        /// <summary>
+        /// Configure a option formatter for option <typeparam name="TOptions"/> if none is configured
+        /// </summary>
+        /// <typeparam name="TOptions"></typeparam>
+        /// <typeparam name="TOptionFormatter"></typeparam>
+        /// <param name="services"></param>
+        /// <returns></returns>
+        public static IServiceCollection TryConfigureFormatter<TOptions>(this IServiceCollection services)
```
jason-bragg: This function doesn't really do anything, as there will almost always be multiple IOptionFormatter in the system.  Should probably remove, as it will only confuse users.

benjaminpetit: Are you talking about this specific Try method or both `TryConfigureFormatter` and `ConfigureFormatter`?

jason-bragg:

benjaminpetit:
Mhhh Ado test fails..... @xiazen 

@dotnet-bot test functional

src/Orleans.Core/Configuration/OptionLogger/OptionFormatterExtensionMethods.cs
jason-bragg: Only this Try extension.  TryConfigureFormatter<TOptions>(..)

ConfigureFormatter<TOptions>(..) is fine

and the existing TryConfigureFormatter<TOptions<TFOptionsFormatter>
(..) is fine.

jason-bragg:
Jenkins seems Janky today.  Can't get Reuben's to get pass the build, now this one..

veikkoeeva:
How do the ADO.NET tests fail? Something I could do?

PR #4057, opened 2018-02-20
Remove redundant IOptionFormatter/IOptionFormatterResolver implementations

benjaminpetit:
- Added handling of `IDictionary` and `ICollection` in ` DefaultOptionsFormatter`
- Removed redundant implementations of IOptionFormatter and IOptionFormatterResolver


src/Orleans.Core/Configuration/OptionLogger/DefaultOptionsFormatter.cs
```diff
@@ -30,21 +32,60 @@ public IEnumerable<string> Format()
...
+            else if (typeof(ICollection).IsInstanceOfType(value))
+            {
+                var coll = (ICollection)value;
+                if (coll.Count > 0)
+                {
+                    StringBuilder sb = null;
+                    foreach (var item in coll)
+                    {
+                        if (sb == null)
+                            sb = new StringBuilder(RedactIfNeeded(redactAttribute, item).ToString());
```
xiazen: if move the string builder construction out of foreach loop can save one "if" check every looping.

src/Orleans.Core/Configuration/OptionLogger/DefaultOptionsFormatter.cs
```diff
@@ -30,21 +32,60 @@ public IEnumerable<string> Format()
...
+                        else
+                            sb.Append($", {RedactIfNeeded(redactAttribute, item).ToString()}");
```
xiazen: I think ToString is redundant here

src/Orleans.Core/Configuration/OptionLogger/DefaultOptionsFormatter.cs
```diff
@@ -30,21 +32,60 @@ public IEnumerable<string> Format()
...
-                redactAttribute != null 
-                    ? redactAttribute.Redact(value)
-                    : value);
+            // If it is a dictionary -> one line per item
+            if (typeof(IDictionary).IsInstanceOfType(value))
+            {
+                var dict = (IDictionary)value;
+                foreach (DictionaryEntry kvp in dict)
+                {
+                    result.Add(OptionFormattingUtilities.Format($"{nameof(dict)}.{kvp.Key}", RedactIfNeeded(redactAttribute, kvp.Value)));
```
ReubenBond: The fixed string "dict" is a little weird here. Maybe `$"{name}[{kvp.Key}]"` instead?
Similarly for ICollection, we could keep each value on a separate line and format it as `$"{name}[{index}]"` -- I'm either way about this

src/Orleans.Core/Configuration/OptionLogger/DefaultOptionsFormatter.cs
```diff
@@ -30,21 +32,60 @@ public IEnumerable<string> Format()
...
             var value = property.GetValue(this.options);
             var redactAttribute = property.GetCustomAttribute<RedactAttribute>(inherit: true);
 
-            return OptionFormattingUtilities.Format(
-                name, 
-                redactAttribute != null 
-                    ? redactAttribute.Redact(value)
-                    : value);
+            // If it is a dictionary -> one line per item
+            if (typeof(IDictionary).IsInstanceOfType(value))
```
ReubenBond: this could be `value is IDictionary dict` instead

src/AWS/Orleans.Reminders.DynamoDB/Reminders/DynamoDbReminderServiceOptions.cs
```diff
@@ -14,29 +14,7 @@ public class DynamoDBReminderTableOptions
         /// <summary>
         /// Gets or sets the connection string.
         /// </summary>
+        [RedactConnectionString]
```
jason-bragg: Are we confident RedactConnectionString is safe with DynamoDb connection string?

src/AWS/Orleans.Reminders.DynamoDB/SiloHostBuilderReminderExtensions.cs
```diff
@@ -63,7 +63,7 @@ public static IServiceCollection UseDynamoDBReminderService(this IServiceCollect
         {
             services.AddSingleton<IReminderTable, DynamoDBReminderTable>();
             services.Configure(configure);
-            services.TryConfigureFormatter<DynamoDBReminderTableOptions, DynamoDBReminderTableOptionsFormatter>();
+            services.TryConfigureFormatter<DynamoDBReminderTableOptions>();
```
jason-bragg: TryConfigureFormatter<TOption> won't work and should be removed.
This call should always call ConfigureFormatter.

src/AdoNet/Orleans.Reminders.AdoNet/ReminderService/AdoNetReminderTableOptions.cs
```diff
@@ -17,30 +17,7 @@ public class AdoNetReminderTableOptions
         /// <summary>
         /// Gets or sets the connection string.
         /// </summary>
+        [RedactConnectionString]
```
jason-bragg: Same as other RedactConnectionString comment. I think we need use Redact, for all formats we've not verified the format of.

src/AWS/Orleans.Reminders.DynamoDB/SiloHostBuilderReminderExtensions.cs
jason-bragg: Same applies to all users of TryConfigureFormatter< TOptions >

src/Orleans.Core/Configuration/OptionLogger/DefaultOptionsFormatter.cs
jason-bragg: I think we need to check to see if the property has a redact action, and if so, redact it, and only check specific data types if not redacted.  The reason is that the redact action takes and object so it's intuitive that the redact would be called on the entire property, and not on parts of the property we select by some internal logic.

src/Orleans.Core/Configuration/OptionLogger/DefaultOptionsFormatter.cs
benjaminpetit: I use this trick to now when to add or not a comma. I am playing with ICollection here (and not the generic one) so I don't have a lot of options

src/AWS/Orleans.Reminders.DynamoDB/Reminders/DynamoDbReminderServiceOptions.cs
benjaminpetit: It is explicitly handled in `ConfigUtilities.RedactConnectionStringInfo`

src/Orleans.Core/Configuration/OptionLogger/DefaultOptionsFormatter.cs
benjaminpetit: Yep wrong copy paste, should be `name`. I like your proposal for `ICollection`

src/AdoNet/Orleans.Reminders.AdoNet/ReminderService/AdoNetReminderTableOptions.cs
benjaminpetit: We explicitly handle SQL connection string in the implementation, but it is true that ADO.NET is generic so I will change it to the plain Redact

src/Orleans.Core/Configuration/OptionLogger/DefaultOptionsFormatter.cs
benjaminpetit: I agree

Issue #4065, opened 2018-02-22
Duplicate cluster ID on the client.


jason-bragg:
ClusterId is defined in the SiloOptions, which is available on the client (per my understanding), and in the ClusterClientOptions.  This ambiguity introduces a number of maintenance and configuration questions. 

Client and silo should each have only one source of cluster information, if it's the same or different option is less important than each having only a single source.

sergeybykov:
Resolved via #4070.

PR #4070, opened 2018-02-22
Refactoring SiloOptions, ClientClusterOptions to ClusterOptions

benjaminpetit:
Right now provider that run on the client and the silo (like streaming) are broken on the client, because they expect `SiloOptions` which is obviously not set, because the client expect to use `ClusterClientOptions`.

The proposal is to configure ClusterId/ServiceId in a common option for clients and silos.

In this PR, I:

- Renamed `ClusterClientOptions` to a more generic term `ClusterOptions`
- Rremoved `ClusterId` and `ServiceId` from `SiloOption`, and added `ServiceId` to `ClusterOptions`
- Used `ClusterOptions` instead of `SiloOption` when `ClusterId` was needed `ServiceId`


benjaminpetit:
Related issue: #4065

ReubenBond:
@dotnet-bot test bvt
It seems that it isn't running BVTs

Issue #4075, opened 2018-02-23
IVersionStore should not be used before the silo is ready


benjaminpetit:
Right now it is used in `ServiceLifecycleStage.RuntimeGrainServices`, and sometimes when the first silo starts you can get these errors in the logs:

> [2018-02-23 01:23:05.417 GMT 45	Error	102202	Orleans.Storage.AzureTableGrainStorage]	!!!!!!!!!! Error from storage provider AzureTableGrainStorage.Orleans.Runtime.Versions.VersionStoreGrain during ReadState for grain Type=Orleans.Runtime.Versions.VersionStoreGrain Pk=*grn/13BAB4D8/000000000000000000000000000000000600000013bab4d8+loadtests636549168964888958-0xD87DF39E Id=GrainReference:*grn/13BAB4D8/00000000+loadtests636549168964888958 Error=

I think we should move its initialization in  `ServiceLifecycleStage.ApplicationServices`

sergeybykov:
Resolved via #4077.

PR #4077, opened 2018-02-26
Start GrainVersionStore later at runtime

benjaminpetit:
Related issue: #4075

### DIFF
--- a/src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
+++ b/src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
@@ -169,7 +169,9 @@ namespace Orleans.Hosting
             services.TryAddSingleton<VersionSelectorManager>();
             services.TryAddSingleton<CachedVersionSelectorManager>();
             // Version selector strategy
-            services.TryAddSingleton<IVersionStore, GrainVersionStore>();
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -169,7 +169,9 @@ internal static void AddDefaultServices(HostBuilderContext context, IServiceColl
             services.TryAddSingleton<VersionSelectorManager>();
             services.TryAddSingleton<CachedVersionSelectorManager>();
             // Version selector strategy
-            services.TryAddSingleton<IVersionStore, GrainVersionStore>();
+            services.TryAddSingleton<GrainVersionStore>();
+            services.AddFromExisting<IVersionStore, GrainVersionStore>();
```
jason-bragg: Suggest
services.TryAddSingleton<IVersionStore>(sp => sp.GetRequiredService<GrainVersionStore>());
As this would allow users to provide their own impl of IVersionStore prior to registering the default services.

jason-bragg:
This addresses the issue, but I remain concerned about running application logic (application services and servicing application requests) before the configured versioning and compatibility strategies have been loaded.

PR #4078, opened 2018-02-26
Add basic sample using docker and asp.net core

benjaminpetit:
The idea is to have a basic sample combining Docker, Orleans and ASP.NET Core.

This sample is pretty simple: a single grain identified by a integer, that can hold a string value. A basic webpage allow user to get and set these grain values.

It uses ASP.NET Docker tooling, so you can build and debug using Docker containers, which is nice. But it seems that if the containers are started with VS, the silo doesn't shutdown gracefully. I validated that if started from Docker CLI the silo shutdowns properly.

I am not an expert in ASP.NET Core, so I may have made mistake here and there.

### DIFF
--- a/Samples/docker-aspnet-core/API/Startup.cs
+++ b/Samples/docker-aspnet-core/API/Startup.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Threading.Tasks;
+using GrainInterfaces;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans;
+using Orleans.Hosting;
+
+namespace API
+{
+    public class Startup
+    {
+        public Startup(IConfiguration configuration)
+        {
+            Configuration = configuration;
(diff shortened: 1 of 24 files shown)


Samples/docker-aspnet-core/Silo/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+WORKDIR /src/Silo
+RUN dotnet build -c Release -o /app
+
+FROM build AS publish
+RUN dotnet publish -c Release -o /app
+
+FROM base AS final
+WORKDIR /app
+COPY --from=publish /app .
+COPY connection_string.txt .
```
ReubenBond: does this file exist and is it being used?

Samples/docker-aspnet-core/API/Startup.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            for (var i=0; i<5; i++)
+            {
+                try
+                {
+                    await client.Connect();
+                    return;
+                }
+                catch(Exception)
+                { }
```
SebastianStehle: there is a `if (i ==4) throw;` missing

PR #4079, opened 2018-02-26
Align IClientBuilder APIs with ISiloHostBuilder

ReubenBond:


ReubenBond:
Coincidentally, this makes ClientBuilder and SiloHostBuilder almost identical.
cc @jdom

src/Orleans.Core/Core/ClientBuilder.cs
```diff
@@ -41,10 +60,24 @@ public IClusterClient Build()
         }
 
         /// <inheritdoc />
-        public IClientBuilder ConfigureServices(Action<IServiceCollection> configureServices)
+        public IClientBuilder ConfigureHostConfiguration(Action<IConfigurationBuilder> configureDelegate)
```
jason-bragg: Why change this name?  ConfigureHostConfiguration seems to indicate that this is for a host, which is not the case, correct?  Can't multiple clients can be configured on the same host?  Isn't this only for a specific client?

src/Orleans.Core/Core/ClientBuilderExtensions.cs
```diff
@@ -19,7 +22,7 @@ public static class ClientBuilderExtensions
         /// <param name="builder">The host builder.</param>
         /// <param name="siloName">The silo name.</param>
         /// <returns>The silo builder.</returns>
```
jason-bragg: The silo builder => The client builder

jason-bragg: and related references to silo in comments

src/Orleans.Core/Core/ClientBuilderExtensions.cs
```diff
@@ -19,7 +22,7 @@ public static class ClientBuilderExtensions
...
-        public static IClientBuilder ConfigureDefaults(this IClientBuilder builder)
+        public static IClientBuilder Configure(this IClientBuilder builder)
```
jason-bragg: Isn't this a convenience function to setup common defaults for the client?  If so, didn't the previous name communicate that better?

ReubenBond: The SiloHostBuilder has a method called `Configure` which does the same thing as this. In that case, the method was previously called `ConfigureOrleans`.

I'm happy to revert these changes

src/Orleans.Core/Core/ClientBuilder.cs
ReubenBond: It's just bad diffing, I'm not *changing* the name.
re: ~"multiple clients configured on the same host", there's not really any notion of a host here

This method configures `IConfiguration` for use in other builder calls such as `ConfigureServices` and `ConfigureAppConfiguration`. Maybe the method should be renamed `ConfigureBuilderConfiguration`?

src/Orleans.Core/Core/ClientBuilderExtensions.cs
xiazen: no matter what we choose, the naming need to be consistent with silo and client. +1 for configureDefaults

Issue #3951, opened 2018-02-02
Configuration Work - TestingHost


jason-bragg:
Remove use of legacy configuration from Orleans.TestingHost, introducing a Orleans.TestingHost.Legacy if necessary.

xiazen:
finished in #4081

xiazen:
we haven't talked about what do we want to do with Orleans.TestingHost.Legacy package. Do we release that package, or just keep it internal?

sergeybykov:
I don't think we want to encourage usage of `Orleans.TestingHost.Legacy` by releasing a package with it. I hope we can keep for internal testing of the legacy configuration only.

xiazen:
there's opinion in both sides. let's keep this issue and talk about I t

xiazen:
I'm with the side of keeping it internal. @jason-bragg here since I remember he has a different opinion. And also @benjaminpetit @ReubenBond in case they have opinion

jason-bragg:
I'm ok with keeping it internal, unless the community rebels.  I mainly wanted to make sure this was an informed choice, not an oversight.

ReubenBond:
I'm also ok with keeping it internal. We can always publish it later on if we wish

sergeybykov:
Closing this then.

Issue #3949, opened 2018-02-02
Configuration Work - Defaults


jason-bragg:
While moving from monolithic Cluster and Client configurations to options we've duplicated defaults. Many of the options introduced duplicated the defaults in the options, rather than having the options have the authoritative default and referencing these defaults from the legacy configurations.

We need to go through all of the added options and ensure the defaults in the options are correct and authoritative.

sergeybykov:
Resolved via #4182.

PR #4081, opened 2018-02-27
Move legacy configuration methods to Orleans.TestingHost.Legacy

xiazen:
- Remove Orleans.TestingHost dependency on legacy packges
- add Orleans.TestingHost.Legacy to configure TestingHost using legacy configuration objects 

Fix issue #3951


src/Orleans.TestingHost/TestClusterHostFactory.cs
```diff
@@ -59,13 +59,10 @@ public static ISiloHost CreateSiloHost(string hostName, IEnumerable<IConfigurati
                 TryConfigureTestClusterMembership(context, services);
                 TryConfigureFileLogging(configuration, services, siloName);
 
-                // TODO: make SiloHostBuilder work when not using the legacy configuration, similar to what we did with ClientBuilder.
-                // All The important information has been migrated to strongly typed options (everything should be migrated, but the minimum required set is already there).
-                var clusterConfiguration = GetOrCreateClusterConfiguration(services);
```
xiazen: if all the defaults are indeed set in options now, then this step can be removed and tests should all be passing. Otherwise this PR may depends on #3949 or we can finish #3949 in this PR.

xiazen:

xiazen:
this PR needs some review ~~. and tests are passing

PR #4085, opened 2018-02-27
Update changelog for 2.0.0-rc1

sergeybykov:


changelog.md
```diff
@@ -7,6 +7,57 @@ All notable end-user facing changes are documented in this file.
 *Here are all the changes in `master` branch, and will be moved to the appropriate release once they are included in a published nuget package.
 The idea is to track end-user facing changes as they occur.*
 
+
+### [2.0.0-rc1]
+
+- Major changes
```
ReubenBond: Should we be clear that these are changes since -beta3 (and not since the last stable)?

ReubenBond: Maybe that's obvious, I don't know. At least for 2.0 final we should be clear on the distinction

sergeybykov:
Add a clarification "changes since 2.0.0-beta3".

PR #4088, opened 2018-02-28
Update HelloWorld.NetCore sample to use 2.0 rc1

xiazen:
Update HelloWorld.NetCore sample to use 2.0 rc0 as a preparation step to update the doc. Since docs use this sample as main example.


ReubenBond:
rc1?

xiazen:
we released rc0 AFAIK. I didn't see a rc1 @sergeybykov am I right?

xiazen:
hmm you are right, we released rc1.

Samples/HelloWorld.NetCore/src/OrleansClient/Program.cs
```diff
@@ -45,9 +48,11 @@ static int Main(string[] args)
             {
                 try
                 {
-                    var config = ClientConfiguration.LocalhostSilo();
+                    var siloAddress = IPAddress.Loopback;
+                    var gatewayPort = 30000;
                     client = new ClientBuilder()
-                        .UseConfiguration(config)
+                        .ConfigureCluster(options => options.ClusterId = "helloworldcluster")
+                        .UseStaticClustering(options => options.Gateways = new List<Uri>(){ (new IPEndPoint(siloAddress, gatewayPort)).ToGatewayUri() })
```
benjaminpetit: NIT: 
`UseStaticClustering(options => options.Gateways.Add(new IPEndPoint(siloAddress, gatewayPort)).ToGatewayUri()) `

We are good people. We instantiate collections.

Samples/HelloWorld.NetCore/NuGet.Config
```diff
@@ -1,7 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
benjaminpetit: I think we could and should remove this file. This sample do not uses the myget feed, and if for any reason we delete this feed from myget, it will break the build.

xiazen: hmm maybe sometimes it does, like when we want to update the sample with things published on myget? lol I'm fine with both, just assume myget is configured for a reason. But I don't know much as myget. @sergeybykov  is only nightly build published in myget? or it is the one we are going to remove?

sergeybykov: I think sample should point to nuget.org. MyGet is only for nightly builds and bleeding edge enthusiasts.

Samples/HelloWorld.NetCore/src/OrleansClient/OrleansClient.csproj
```diff
@@ -11,7 +11,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Orleans.Core" Version="$(OrleansPackageVersion)" />
+    <PackageReference Include="Microsoft.Orleans.Core" Version="2.0.0-rc1" />
```
ReubenBond: In the past we were just updating the value of OrleansPackageVersion in Directory.Build.props

ReubenBond: I think we should continue doing that for now

xiazen: but that would update every sample to use that version, which indicates we update all samples at once? or am I misunderstanding anything

sergeybykov:
I suggest we move this sample to Samples\2.0\ folder, and keep using '$(OrleansPackageVersion)` for now for all 2.0 samples there.

benjaminpetit:
> I suggest we move this sample to Samples\2.0\ folder, and keep using '$(OrleansPackageVersion)` for now for all 2.0 samples there.

Let's avoid merge conflicts, for now, @xiazen, could you move this sample to a 2.0 subfolder? I will do the same in my PR

Samples/HelloWorld.NetCore/src/OrleansClient/OrleansClient.csproj
benjaminpetit: The issue is that the value of `OrleansPackageVersion` is currently `2.0.0-dev`, so it wouldn't work, correct?

I think we should have a Directory.Build.props dedicated for samples

xiazen:
e. if we are to avoid merge conflics, isn't you should do it in your pr, and then we merge your pr and I rebase with master, and vice versa. 

If we both do the same thing, there's more merge conflicts

benjaminpetit:
Na, I will not touch this folder, so you can move it right now

Samples/HelloWorld.NetCore/src/OrleansClient/OrleansClient.csproj
benjaminpetit: Clarification: I think we should have one Directory.Build.props per sample. Otherwise we will break all samples when updating it

xiazen:
rebased and addressed feedback

Issue #4086, opened 2018-02-27
SiloBuilder extensions/options usability issues


benjaminpetit:
Let's track here all change that we should do in SiloBuilder extensions methods and options before release.

- [ ] `UseDevelopmentClustering` is not very user friendly; user should be able to just call `siloBuilder.UseDevelopmentClustering()`, not `siloBuilder.UseDevelopmentClustering((Action<DevelopmentMembershipOptions>) null)`
- [ ] `ConfigureEndpoints` extensions are in namespace `Orleans.Configuration` instead of `Orleans.Hosting`

ReubenBond:
We will add new methods to client and silo:
`siloBuilder.UseLocalhostClustering(int siloPort = ..., int gatewayPort = ...)` and `clientBuilder.UseLocalhostClustering(int gatewayPort = ...)`

These will configure the system for having a single silo on the local host.

To configure a single silo/client on localhost:
```C#
// on silo
  siloBuilder.UseLocalhostClustering();

// on client
  clientBuilder.UseLocalhostClustering();
```

To configure multiple silos/clients on localhost:
```C#
// on primary
   siloBuilder.UseLocalhostClustering(siloPort, gatewayPort);

// on other silos
  siloBuilder.UseLocalhostClustering(siloPort, gatewayPort, primarySiloEndpoint);

// on client
  clientBuilder.UseLocalhostClustering(gatewayPort1, gatewayPort2, gatewayPort3)
```

PR #4090, opened 2018-02-28
Improve usability of dev cluster

### LINKED ISSUE #4084: Failure loading MemoryStorageGrain
On latest RC, calling `.AddMemoryGrainStorageAsDefault()` fail to register the provider if we are adding any type to application parts manager (i.e. the auto-discovery is disable since I'm adding assemblies manually).

This method is supposed to add the `IMemoryStorageGrain` to the App parts internally.

Logs attached.

Thanks
[logs-from-kubesilo-in-kubesilo-57bf49b4b7-zzs9f (1).txt](https://github.com/dotnet/orleans/files/1764074/logs-from-kubesilo-in-kubesilo-57bf49b4b7-zzs9f.1.txt)

ReubenBond:
Fixes most of #4086
Fixes #4084

### DIFF
--- a/src/Orleans.Core/Core/ClientBuilderExtensions.cs
+++ b/src/Orleans.Core/Core/ClientBuilderExtensions.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Linq;
+using System.Net;
 using System.Collections.Generic;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -8,6 +10,7 @@ using Orleans.Hosting;
 using Orleans.ApplicationParts;
 using Orleans.CodeGeneration;
 using Orleans.Messaging;
+using Orleans.Runtime;
 
 namespace Orleans
 {
@@ -143,6 +146,41 @@ namespace Orleans
         public static IClientBuilder ConfigureLogging(this IClientBuilder builder, Action<ILoggingBuilder> configureLogging)
         {
             return builder.ConfigureServices(collection => collection.AddLogging(loggingBuilder => configureLogging(loggingBuilder)));
+        }
+
+        /// <summary>
+        /// Configures the client to connect to a silo on the localhost.
+        /// </summary>
+        /// <param name="gatewayPort">The local silo's gateway port.</param>
+        public static IClientBuilder UseLocalhostClustering(this IClientBuilder builder, int gatewayPort = 30000, string clusterId = null)
+        {
+            return builder.UseStaticClustering(new IPEndPoint(IPAddress.Loopback, gatewayPort))
+                .ConfigureCluster(options =>
+                {
+                    if (!string.IsNullOrWhiteSpace(clusterId)) options.ClusterId = clusterId;
+                });
+        }
+
+        /// <summary>
+        /// Configures the client to connect to a silo on the localhost.
+        /// </summary>
+        /// <param name="gatewayPorts">The local silo gateway port.</param>
+        public static IClientBuilder UseLocalhostClustering(this IClientBuilder builder, params int[] gatewayPorts)
+        {
+            return builder.UseStaticClustering(gatewayPorts.Select(p => new IPEndPoint(IPAddress.Loopback, p)).ToArray())
+                .ConfigureCluster(options =>
+                {
+                    options.ClusterId = ClusterOptions.DevelopmentClusterId;
+                });
+        }
+
+        /// <summary>
+        /// Configures the client to use static clustering.
+        /// </summary>
+        /// <param name="endpoints">The gateway endpoints.</param>
+        public static IClientBuilder UseStaticClustering(this IClientBuilder builder, params IPEndPoint[] endpoints)
+        {
+            return builder.UseStaticClustering(options => options.Gateways = endpoints.Select(ep => ep.ToGatewayUri()).ToList());
         }
 
         /// <summary>
--- a/src/Orleans.Core/Core/ClientBuilderExtensions.cs
+++ b/src/Orleans.Core/Core/ClientBuilderExtensions.cs
@@ -152,7 +152,10 @@ namespace Orleans
         /// Configures the client to connect to a silo on the localhost.
         /// </summary>
         /// <param name="gatewayPort">The local silo's gateway port.</param>
-        public static IClientBuilder UseLocalhostClustering(this IClientBuilder builder, int gatewayPort = 30000, string clusterId = null)
+        public static IClientBuilder UseLocalhostClustering(
+            this IClientBuilder builder,
+            int gatewayPort = 30000,
+            string clusterId = ClusterOptions.DevelopmentClusterId)
         {
             return builder.UseStaticClustering(new IPEndPoint(IPAddress.Loopback, gatewayPort))
                 .ConfigureCluster(options =>
--- a/src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
+++ b/src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
@@ -134,11 +134,7 @@ namespace Orleans.Hosting
             });
 
             builder.UseDevelopmentClustering(primarySiloEndpoint ?? new IPEndPoint(IPAddress.Loopback, siloPort));
-
-            if (!string.IsNullOrWhiteSpace(clusterId))
-            {
-                builder.Configure(options => options.ClusterId = clusterId);
-            }
+            builder.Configure(options => options.ClusterId = clusterId);
 
             return builder;
         }
--- a/src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
+++ b/src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Net;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Orleans.Configuration;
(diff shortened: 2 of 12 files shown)


ReubenBond:
@dotnet-bot test bvt

src/Orleans.Runtime/MembershipService/GrainBasedMembershipTable.cs
```diff
@@ -31,7 +31,7 @@ public GrainBasedMembershipTable(IServiceProvider serviceProvider, ILogger<Grain
         {
             var options = this.serviceProvider.GetRequiredService<IOptions<DevelopmentMembershipOptions>>().Value;
             var siloDetails = this.serviceProvider.GetService<ILocalSiloDetails>();
-            bool isPrimarySilo = siloDetails.SiloAddress.Endpoint.Equals(options.PrimarySiloEndpoint);
+            bool isPrimarySilo = siloDetails.SiloAddress.Endpoint.Equals(options.PrimarySiloEndpoint) || options.PrimarySiloEndpoint == null;
```
benjaminpetit: I think it should throw an exception if `PrimarySiloEndpoint` isn't set, maybe in a validator.

The reason is that if someone do not set any value for it and start two silos, it will be hard for this person to understand what's going on: the two silos will not be in the same cluster.

src/Orleans.Core/Core/ClientBuilderExtensions.cs
```diff
@@ -90,6 +93,41 @@ public static IClientBuilder ConfigureLogging(this IClientBuilder builder, Actio
             return builder.ConfigureServices(collection => collection.AddLogging(loggingBuilder => configureLogging(loggingBuilder)));
         }
 
+        /// <summary>
+        /// Configures the client to connect to a silo on the localhost.
+        /// </summary>
+        /// <param name="gatewayPort">The local silo's gateway port.</param>
+        public static IClientBuilder UseLocalhostClustering(this IClientBuilder builder, int gatewayPort = 30000, string clusterId = null)
```
xiazen: why isn't clusterId to set to `ClusterOptions .DevelopmentClusterId` so we can save those `IsNullOrWhiteSpace` dance?

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
```diff
@@ -49,26 +75,15 @@ public static ISiloHostBuilder Configure(this ISiloHostBuilder builder, string s
...
-                    services.PostConfigure<SiloOptions>(options => options.SiloName = options.SiloName
-                                           ?? context.HostingEnvironment.ApplicationName
-                                           ?? $"Silo_{Guid.NewGuid().ToString("N").Substring(0, 5)}");
-
-                    services.TryAddSingleton<Silo>();
-                    DefaultSiloServices.AddDefaultServices(context, services);
-
-                    context.Properties.Add("OrleansServicesAdded", true);
-                }
+                    if (!string.IsNullOrEmpty(siloName))
```
jason-bragg: shouldn't we throw if silo name isn't valid?
Or if null is valid, make it an optional argument to make this explicit?

src/Orleans.Core/Core/ClientBuilderExtensions.cs
xiazen: generally speaking, the optional param should be set to its default value, instead of null and not being used later, right?

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
```diff
@@ -95,11 +110,62 @@ public static ISiloHostBuilder ConfigureSiloName(this ISiloHostBuilder builder,
...
         /// </summary>
-        public static ISiloHostBuilder UseDevelopmentClustering(this ISiloHostBuilder builder, Action<DevelopmentMembershipOptions> configureOptions)
+        /// <param name="builder">The silo builder.</param>
+        /// <param name="siloPort">The silo port.</param>
+        /// <param name="gatewayPort">The gateway port.</param>
+        /// <param name="primarySiloEndpoint">
+        /// The endpoint of the primary silo, or <see langword="null"/> to use this silo as the primary.
+        /// </param>
+        /// <returns>The silo builder.</returns>
+        public static ISiloHostBuilder UseLocalhostClustering(
```
jason-bragg: Can we keep development extensions in a different namespace.  Maybe Orleans.Development?  This would hide them unless the user specifically opts into them for development purposes, and would prevent users from being confused by them when writing their production logic.

test/TesterInternal/TesterInternal.csproj
```diff
@@ -27,6 +27,7 @@
     <ProjectReference Include="..\..\src\Azure\Orleans.Persistence.AzureStorage\Orleans.Persistence.AzureStorage.csproj" />
     <ProjectReference Include="..\..\src\Orleans.CodeGeneration\Orleans.CodeGeneration.csproj" />
     <ProjectReference Include="..\..\src\Orleans.Core.Legacy\Orleans.Core.Legacy.csproj" />
+    <ProjectReference Include="..\..\src\Orleans.TestingHost.Legacy\Orleans.TestingHost.Legacy.csproj" />
```
xiazen: why are we fall back to use TestingHost.Legacy? we shouldn't be adding tests using this package. The end goal is to move away from it and also testing the new `TestingHost` patterns

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
```diff
@@ -95,11 +110,62 @@ public static ISiloHostBuilder ConfigureSiloName(this ISiloHostBuilder builder,
...
+            this ISiloHostBuilder builder,
+            int siloPort = EndpointOptions.DEFAULT_SILO_PORT,
+            int gatewayPort = EndpointOptions.DEFAULT_GATEWAY_PORT,
+            IPEndPoint primarySiloEndpoint = null,
```
xiazen: NIT: can we use its real default here, instead of null? Can make code more cleaner

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
```diff
@@ -95,11 +110,62 @@ public static ISiloHostBuilder ConfigureSiloName(this ISiloHostBuilder builder,
...
+                options.AdvertisedIPAddress = IPAddress.Loopback;
+                options.SiloPort = siloPort;
+                options.GatewayPort = gatewayPort;
+            });
+
+            builder.UseDevelopmentClustering(primarySiloEndpoint ?? new IPEndPoint(IPAddress.Loopback, siloPort));
+
+            if (!string.IsNullOrWhiteSpace(clusterId))
+            {
+                builder.Configure(options => options.ClusterId = clusterId);
```
xiazen: hmm if clusterId is null, which means user explicitly set it to null, I don't think we should just ignore their choice without telling them, leaving them the impression that clusterId is not needed. We should do a ConfigurationValidaiton instead

src/Orleans.Runtime/MembershipService/GrainBasedMembershipTable.cs
xiazen: should add a configuration validator to detect that earlier

src/Orleans.Core/Core/ClientBuilderExtensions.cs
ReubenBond: Good point.
Yes, optional param should be the default value. The exception is for values which cannot be represented as compile-time constants, but that doesn't apply here.

src/Orleans.Runtime/MembershipService/GrainBasedMembershipTable.cs
ReubenBond: Hmm, the alternative is that when a use calls `UseLocalhostClustering` it automatically sets the primary silo endpoint to the current silo endpoint (equivalent to `ClusterConfiguration.LocalhostPrimarySilo()`), which doesn't really buy us anything.

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
ReubenBond: Other extensions intentionally pass null for this value to use the default. I rather not make it an optional argument because with all these overloads of a method simply called `Configure`, we're bound to hit ambiguity

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
ReubenBond: Users wont find them if they're not in the right namespace. Is `UseLocalhostClustering` not obvious enough that it's not a scale-out solution?

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
ReubenBond: We can't, only compile-time constants can be used as defaults

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
ReubenBond: Perhaps we set a default value for ClusterId if it's not set by the user, similar to how we handle SiloName

test/TesterInternal/TesterInternal.csproj
ReubenBond: `ClientInitTests` uses it, that's unchanged but I moved the class from DefaultCluster.Tests to here, since it doesn't use the default cluster fixture.

This PR moved almost every test in DefaultCluster.Tests away from legacy.

src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs
jason-bragg: To be clear, I'm ok with the existing namespace, just suggesting we use a different one.

As far as users not finding them, that, imo, was the goal.  I think it's good to have developer only options less discoverable than the expected production behaviors.  Should one include the development namespace, they will be equally discoverable.

As far as the name being clear enough, I was suggesting this as a general pattern for all developer features, not just these calls.  For instance, extensions setting up the grain based membership service, the memory storage provider, the in-memory transaction log, or any other developer tooling would all be exposed under a development namespace, communicating clearly and in a consistent way what features we've provided for development purposes vs. production use.

This is a suggestion, entirely optional and not something we should block this PR for.

ReubenBond: Ok, got it.
I think we should do something to make sure that users explicitly opt-in to enabling features which we deem unsuitable for production use.

We could use `HostBuilderContext` (eg `Properties` or `HostingEnvironment`) to signal that the developer has enabled "dev/test only" features.

xiazen:
LGTM, but It has merge conflicts now. Can merge after address that

ReubenBond:
Rebased

PR #4091, opened 2018-02-28
Extensions should add their own application parts

ReubenBond:
Hopefully I didn't miss any

This addresses #4084


src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
```diff
@@ -14,32 +14,36 @@ public static class ClientBuilderExtensions
...
+            return builder
+                .ConfigureApplicationParts(parts => parts.AddFrameworkPart(typeof(SQSAdapterFactory).Assembly))
+                .ConfigureServices(services => services.AddClusterClientSqsStreams(name, configureOptions));
         }
 
         /// <summary>
         /// Configure cluster client to use SQS persistent streams.
         /// </summary>
-        public static IServiceCollection AddClusterClientSqsStreams(this IServiceCollection services, string name, Action<SqsStreamOptions> configureOptions)
+        private static void AddClusterClientSqsStreams(this IServiceCollection services, string name, Action<SqsStreamOptions> configureOptions)
```
xiazen: why are we making those methods private?

src/AWS/Orleans.Streaming.SQS/Hosting/SiloBuilderExtensions.cs
```diff
@@ -13,33 +13,39 @@ public static class SiloBuilderExtensions
...
+            return builder
+                .ConfigureApplicationParts(parts => parts.AddFrameworkPart(typeof(SQSAdapterFactory).Assembly))
+                .ConfigureServices(services => services.AddSiloSqsStreams(name, configureOptions));
         }
 
         /// <summary>
         /// Configure silo to use SQS persistent streams.
         /// </summary>
-        public static IServiceCollection AddSiloSqsStreams(this IServiceCollection services, string name, Action<SqsStreamOptions> configureOptions)
+        private static void AddSiloSqsStreams(this IServiceCollection services, string name, Action<SqsStreamOptions> configureOptions)
```
xiazen: same here and many other occurrence

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
ReubenBond: From the linked issue: https://github.com/dotnet/orleans/issues/4084#issuecomment-369024506

In other discussions (I started an email thread on it) I mentioned that IServiceCollection is an ambiguous place for most silo-specific or client-specific extension methods

xiazen: I see

src/Orleans.Core/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -124,7 +124,7 @@ public static IServiceCollection AddIncomingGrainCallFilter(this IServiceCollect
         /// <param name="services">The service collection.</param>
         /// <param name="filter">The filter.</param>
         /// <returns>The service collection.</returns>
-        public static IServiceCollection AddOutgoingGrainCallFilter(this IServiceCollection services, IOutgoingGrainCallFilter filter)
+        internal static IServiceCollection AddOutgoingGrainCallFilter(this IServiceCollection services, IOutgoingGrainCallFilter filter)
```
xiazen: why make those methods internal? as I understand, we make provider extension methods on IServiceCollection because they need to configure ApplicationPart and IServiceCollection doesn't have access to ApplicationPart. But this is not true with GrainCallFilters , right?

xiazen:
Maybe I should @jason-bragg here since he participated most in the issue

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
```diff
@@ -14,32 +14,36 @@ public static class ClientBuilderExtensions
         /// </summary>
         public static IClientBuilder AddSqsStreams(this IClientBuilder builder, string name, Action<SqsStreamOptions> configureOptions)
         {
-            return builder.ConfigureServices(services => services.AddClusterClientSqsStreams(name, configureOptions));
+            return builder
+                .ConfigureApplicationParts(parts => parts.AddFrameworkPart(typeof(SQSAdapterFactory).Assembly))
```
jason-bragg: Due to the previous pattern we've many overrides for adding things to the service collection that are not necessary if we're only supporting the builders (which I agree), but the advantage of the previous pattern was that everything eventually ended in one place with a single authoritative set of logic that did the work.

In changing these, I advocate that the options action override call through to the options action builder override, which has the single authoritative set of logic to run.  It should add the parts, and setup the services.

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
jason-bragg: Making these private is sufficient, but my preference would be to delete them entirely and do the work in one of the builder calls.  At this point, these service collection overrides are just excess cruft now, imo.

src/Orleans.Core/ApplicationParts/ApplicationPartManagerExtensions.cs
```diff
@@ -59,7 +59,31 @@ public static TFeature CreateAndPopulateFeature<TFeature>(this IApplicationPartM
...
+            }
+
+            if (assembly == null)
+            {
+                throw new ArgumentNullException(nameof(assembly));
+            }
+
+            return new ApplicationPartManagerWithAssemblies(
+                manager.AddApplicationPart(
+                    new AssemblyPart(assembly) {IsFrameworkAssembly = true}),
```
jason-bragg: What's the difference between the way framework assemblies are treated, vs. non-framework assemblies?

ReubenBond: Previously we would throw if no non-framework assemblies were added, but now will load default assemblies (everything in the current folder + deps.json). If even one non-framework assembly is added, we will *not* do that (and it's up to the user to specify all of their application parts).

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
ReubenBond: Sure thing, fixed in latest

ReubenBond:
The pattern should be cleaner now

jason-bragg:
@dotnet-bot test bvt please

src/Orleans.Core/Configuration/ServiceCollectionExtensions.cs
ReubenBond: We haven't released this code as stable yet, so I rather make them internal while there are no consumers to break.

ReubenBond:
Seems like this could be merged

PR #4097, opened 2018-02-28
Improve configuration validators for ADO.NET configuration

ReubenBond:
* Consolidate naming of ADO.NET invariant option: now it's `Invariant` in all places, instead of being `AdoInvariant` in clustering and `Invariant` everywhere else.
* Add configuration validators for clustering and reminders and improve grain storage validator to check that `Invariant` is set.
* Propagate invariant value from legacy configuration to reminder configuration.
* Add `invariant` parameter on reminder config extension method which takes a connection string.

### DIFF
--- a/src/AdoNet/Orleans.Reminders.AdoNet/SiloHostBuilderReminderExtensions.cs
+++ b/src/AdoNet/Orleans.Reminders.AdoNet/SiloHostBuilderReminderExtensions.cs
@@ -23,10 +23,11 @@ namespace Orleans.Hosting
         /// <returns>
         /// The provided <see cref="ISiloHostBuilder"/>, for chaining.
         /// </returns>
-        public static ISiloHostBuilder UseAdoNetReminderService(this ISiloHostBuilder builder, Action<AdoNetReminderTableOptions> configure)
+        public static ISiloHostBuilder UseAdoNetReminderService(
+            this ISiloHostBuilder builder,
+            Action<AdoNetReminderTableOptions> configureOptions)
         {
-            builder.ConfigureServices(services => services.UseAdoNetReminderService(configure));
-            return builder;
+            return builder.UseAdoNetReminderService(ob => ob.Configure(configureOptions));
         }
 
         /// <summary>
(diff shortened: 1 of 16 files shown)


src/AdoNet/Orleans.Reminders.AdoNet/SiloHostBuilderReminderExtensions.cs
```diff
@@ -59,12 +66,12 @@ public static ISiloHostBuilder UseAdoNetReminderService(this ISiloHostBuilder bu
         /// <returns>
         /// The provided <see cref="ISiloHostBuilder"/>, for chaining.
         /// </returns>
-        public static IServiceCollection UseAdoNetReminderService(this IServiceCollection services, Action<AdoNetReminderTableOptions> configure)
+        private static void UseAdoNetReminderService(this IServiceCollection services, Action<AdoNetReminderTableOptions> configure)
```
xiazen: I believe this is out of the scope of this PR. And we haven't decide to hiding those IServiceCollection extension entirely or not as a team, at least I don't remember we had that discussion. 
Even if we decided to make those extension methods private, then we should do it to all extension methods, so that it is consistent through the team.

ReubenBond: I'll leave it as public now, but we're going to remove it eventually.

xiazen:
@dotnet-bot test this please

src/AdoNet/Orleans.Reminders.AdoNet/SiloHostBuilderReminderExtensions.cs
```diff
@@ -59,30 +61,12 @@ public static ISiloHostBuilder UseAdoNetReminderService(this ISiloHostBuilder bu
         /// <returns>
         /// The provided <see cref="ISiloHostBuilder"/>, for chaining.
         /// </returns>
-        public static IServiceCollection UseAdoNetReminderService(this IServiceCollection services, Action<AdoNetReminderTableOptions> configure)
+        public static void UseAdoNetReminderService(this IServiceCollection services, Action<OptionsBuilder<AdoNetReminderTableOptions>> configureOptions)
```
xiazen: return type should be IServiceCollection, no?

xiazen: no need to change public interface now

ReubenBond: The public interface needs to change anyway, since it was wrong. The overload which takes a connection string simply cannot work without a second call to also configure the Invariant.

The whole method should be removed

xiazen: ? didn't you already removed that wrong method, which is just below this one. This one is fine , right? I was just wondering why you changed the return type.

xiazen:
LGTM.

xiazen:
Failed tests : 
- Orleans.Transactions.Tests.GrainFaultTransactionMemoryTests.AbortTransactionOnExceptions(transactionTestGrainClassName: \"Orleans.Transactions.Tests.SingleStateTransactiona\"...)
- Orleans.Transactions.Tests.GrainFaultTransactionMemoryTests.MultiGrainAbortTransactionOnExceptions(transactionTestGrainClassName: \"Orleans.Transactions.Tests.SingleStateTransactiona\"...)


test failure : MESSAGE:
Assert.Throws() Failure\r\nExpected: typeof(Orleans.Transactions.OrleansTransactionAbortedException)\r\nActual: typeof(System.TypeAccessException): Named type \"Orleans.Transactions.Tests.OrleansCodeGenTransactionTestGrainReference\" is invalid: Type string \"Orleans.Transactions.Tests.OrleansCodeGenTransactionTestGrainReference\" cannot be resolved.

doesn't seem intermittent. But also not related to the PR. So I don't know why them failed

xiazen:
@dotnet-bot test this please

PR #4102, opened 2018-03-01
Improve usability of custom grain placement configuration

### LINKED ISSUE #4098: Custom placement strategy usability improvements
In 2.0.0-rc1, custom placement strategy registration isn't as obvious as it should be:
```C#
builder.ConfigureServices(services =>
{
    services
      .AddSingletonNamedService<PlacementStrategy, VersionAwarePlacementStrategy>(
        nameof(VersionAwarePlacementStrategy))
      .AddSingletonKeyedService<Type, IPlacementDirector, VersionAwarePlacementDirector>(
        typeof(VersionAwarePlacementStrategy));
});
```

The director registration should be obvious and discoverable. The strategy registration should not be needed.

cc @yzhao174

ReubenBond:
Fixes #4098


ReubenBond:
@dotnet-bot test functional

src/Orleans.Core.Abstractions/Placement/ActivationCountBasedPlacement.cs
```diff
@@ -3,7 +3,7 @@
 namespace Orleans.Runtime
 {
     [Serializable]
-    internal class ActivationCountBasedPlacement : PlacementStrategy
```
sergeybykov: What's the reason for making all these built-in strategies public? Seems like it's not something that application code will need to reference.

sergeybykov:
To expand on my comment. These are built-in placement policies, and I think it's a bad idea to let application code override their behavior because, for example, it would be very difficult to figure out in the logs why a well known placement strategy causes very different placement decisions. It is better to be explicit and introduce new strategies for new placement directors, so that there is no confusion. Hence, I think the built-in strategies should stay immutable.

ReubenBond:
All fixed :)

PR #4133, opened 2018-03-05
Support treating all descendants of a base class as [Serializable]

### LINKED ISSUE #3910: [Proposal] New codegen hint
The functional way (style) for doing communication is via messages, which in C# are usually implemented by classes. Currently, Orleans codegen can only generate serializers for class if it's marked with `Serializable` attribute, which means that each and every message class need to be marked with it. That creates a lot of unnecessary noise and make definitions more verbose.

It would be great to have new hint for codegen (ie done via config) to generate serializers for all types which could be assigned from specified type. For example, Orleankka provides 2 base interfaces `Message` and `Result` (for requests and responses respectively) which are used for strongly typed actors (`ActorMessage<TActor>``, ActorMessage<TActor, TResult>`). Having this ability will save 1 line of code per each message definition and will make them more compact.

ReubenBond:
Resolves #3910

Adding `[KnownBaseType]` to a base class will cause the code generator to consider that class and all subclasses as `[Serializable]`.

Note: only base *classes* are supported, not interfaces. Supporting interfaces is a simple addition which we can add later but it might add a perf hit during build (branch-out search is more costly). We generally prefer "pay for what you use" but this could have an impact on all users.

### DIFF
--- a/src/Orleans.CodeGeneration/RoslynCodeGenerator.cs
+++ b/src/Orleans.CodeGeneration/RoslynCodeGenerator.cs
@@ -20,7 +20,6 @@ using Orleans.ApplicationParts;
 using Orleans.Metadata;
 using GrainInterfaceUtils = Orleans.CodeGeneration.GrainInterfaceUtils;
 using SF = Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
-using TypeInfo = System.Reflection.TypeInfo;
 
 namespace Orleans.CodeGenerator
 {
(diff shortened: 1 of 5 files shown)


yevhen:
I would prefer to have support for interfaces more than for base classes. But if you think this may be costly - it’s ok. I won’t be able to mark Message and Result interfaces in Orleankka but they can create similar classes and mark them.

ReubenBond:
Are Message and Result interfaces? We can implement that instead if you prefer and see if it has a significant effect on codegen perf.

yevhen:
Yes, all meta messages are interfaces https://github.com/OrleansContrib/Orleankka/tree/master/Source/Orleankka/Meta

yevhen:
I think that price will only be paid if there interfaces marked with KnownBaseType, no?

sergeybykov:
@dotnet-bot test functional

PR #4135, opened 2018-03-06
Fix silo startup

benjaminpetit:
Currently in rc1, grain can be activated on a silo not really ready. This PR tries to fix this behavior

### DIFF
--- a/src/Orleans.Runtime/Silo/Silo.cs
+++ b/src/Orleans.Runtime/Silo/Silo.cs
@@ -321,7 +321,7 @@ namespace Orleans.Runtime
                 logger.Debug("Creating {0} System Target", "MultiClusterOracle");
                 RegisterSystemTarget((SystemTarget)multiClusterOracle);
             }
-
+            
             var transactionAgent = this.Services.GetRequiredService<ITransactionAgent>() as SystemTarget;
             if (transactionAgent != null)
             {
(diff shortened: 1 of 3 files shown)


src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -557,8 +531,57 @@ void StartGateway()
...
 
+        private async Task OnBecomeActiveStart(CancellationToken ct)
+        {
+            var stopWatch = Stopwatch.StartNew();
+            StartTaskWithPerfAnalysis("Start gateway", StartGateway, stopWatch);
+            void StartGateway()
+            {
+                // Now that we're active, we can start the gateway
+                var mc = this.messageCenter as MessageCenter;
+                mc?.StartGateway(this.Services.GetRequiredService<ClientObserverRegistrar>());
```
ReubenBond: We could stick this into ClientObserverRegistrar instead and make it responsible for adding itself to the MessageCenter by subscribing to `BecomeActive`.

The comment on `BecomeActive` suggests that only `MembershipOracle` should subscribe to it, but we're doing other things here.

ReubenBond: It's fine to leave this, though - we can untangle gateway initialization later. Maybe just update the `BecomeActive` comment to also include the gateway.

ReubenBond:
@dotnet-bot test bvt
@dotnet-bot test functional

src/Orleans.Runtime/Silo/Silo.cs
benjaminpetit: For membership and gateway, we want to do it explicitly in Silo.cs to ensure that they are started in this order. I agree that I can add this in the comment

test/TesterInternal/ActivationsLifeCycleTests/DeactivateOnIdleTests.cs
```diff
@@ -192,6 +192,12 @@ public void Dispose()
                 legacy.ClusterConfiguration.Globals.MaxForwardCount = forwardCount;
                 // For this test we only want to talk to the primary
                 legacy.ClientConfiguration.Gateways.RemoveAt(1);
+                if (forwardCount == 0)
+                    // Disable reminder service for this test: when the secondary silo starts it may
```
ReubenBond: Not blocking, but this could use braces since it's a multi-line body

PR #4136, opened 2018-03-06
Normalize cluster config & simplify binding IConfiguration to TOptions

### LINKED ISSUE #4086: SiloBuilder extensions/options usability issues

ReubenBond:
Fixes #4086.

Unsure if we should also add some sugar methods like `ConfigureClusterId(string)` and `ConfigureSiloName(string)`. `ClusterId` is much more important than `SiloName`.

We could also consider setting `ClusterId` to `"default"` if it's not otherwise set.

### DIFF
--- a/src/Orleans.Runtime.Legacy/Configuration/LegacyClusterConfigurationExtensions.cs
+++ b/src/Orleans.Runtime.Legacy/Configuration/LegacyClusterConfigurationExtensions.cs
@@ -7,7 +7,6 @@ using Orleans.Configuration;
 using Orleans.Runtime;
 using Orleans.Runtime.Configuration;
 using Orleans.Runtime.MembershipService;
-using Orleans.Runtime.Scheduler;
 using Orleans.Providers;
 using System.Collections.Generic;
 
@@ -54,7 +53,6 @@ namespace Orleans.Hosting
         {
             string siloName = Silo.PrimarySiloName;
             builder.Configure<SiloOptions>(options => options.SiloName = siloName);
-            ISiloHostBuilder temp = builder;
             return builder.UseConfiguration(ClusterConfiguration.LocalhostPrimarySilo(siloPort, gatewayPort));
         }
 
--- a/src/Orleans.Runtime.Legacy/Hosting/SiloHost.cs
+++ b/src/Orleans.Runtime.Legacy/Hosting/SiloHost.cs
@@ -102,10 +102,8 @@ namespace Orleans.Runtime.Host
             try
             {
                 if (!this.ConfigLoaded) LoadOrleansConfig();
-                string siloName = this.Name;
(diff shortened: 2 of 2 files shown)


sergeybykov:
>We could also consider setting ClusterId to "default" if it's not otherwise set.

I'm concerned this could lead to error prone usage with people deploying multiple services that will end up joining a single cluster. It seems to be it is better to be explicit with `ClusterId`, unless their a natural unique ID to automatically pick up like deployment ID in Azure Cloud Services.

ReubenBond:
I'm alright with being explicit - in that case we should consider having a `ConfigureClusterId(string)` method, since everyone will need it.

I'll close the other PR. What about this one, though?

src/Orleans.Runtime.Legacy/Configuration/LegacyClusterConfigurationExtensions.cs
```diff
@@ -52,7 +52,9 @@ public static ISiloHostBuilder LoadClusterConfiguration(this ISiloHostBuilder bu
         /// <returns>The silo builder.</returns>
         public static ISiloHostBuilder ConfigureLocalHostPrimarySilo(this ISiloHostBuilder builder, int siloPort = 22222, int gatewayPort = 40000)
         {
-            builder.ConfigureSiloName(Silo.PrimarySiloName);
+            string siloName = Silo.PrimarySiloName;
+            builder.Configure<SiloOptions>(options => options.SiloName = siloName);
+            ISiloHostBuilder temp = builder;
```
sergeybykov: What's this for?

src/Orleans.Runtime.Legacy/Hosting/SiloHost.cs
```diff
@@ -101,8 +102,10 @@ public void InitializeOrleansSilo()
             try
             {
                 if (!this.ConfigLoaded) LoadOrleansConfig();
-                var builder = new SiloHostBuilder()
-                    .ConfigureSiloName(this.Name)
+                string siloName = this.Name;
+                SiloHostBuilder tempQualifier = new SiloHostBuilder();
```
sergeybykov: What's "tempQualifier"?

ReubenBond: uhoh, ReSharper inlining issue. How embarrassing - I'll push a fix

ReubenBond:
All fixed, thanks

PR #4140, opened 2018-03-06
ADO.NET: Fix formatting of generic class names in storage provider

### LINKED ISSUE #3418: Using grain type with nested template arguments crashes the AdoNetStorageProvider
Using a grain interface with nested template arguments crashes the AdoNetStorageProvider. I think the logic inside the `ExtractBaseClass` method assumes only one level of template arguments is present.

This works as expected: `ITemplatedGrain<SomeClass>`

This crashes with the error below: `ITemplatedGrain<OtherTemplatedClass<SomeClass>>`

```
Length cannot be less than zero.
Parameter name: length
Exception = System.ArgumentOutOfRangeException: Length cannot be less than zero.
Parameter name: length
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Orleans.Storage.AdoNetStorageProvider.<>c.<ExtractBaseClass>b__47_1(String i)
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.String.Join(String separator, IEnumerable`1 values)
   at Orleans.Storage.AdoNetStorageProvider.ExtractBaseClass(String typeName)
   at Orleans.Storage.AdoNetStorageProvider.<ReadStateAsync>d__42.MoveNext()
   at Orleans.Runtime.Scheduler.SchedulerExtensions.<>c__DisplayClass1_0.<<QueueTask>b__0>d.MoveNext()
   at Orleans.Runtime.Catalog.<SetupActivationState>d__73.MoveNext()
   at Orleans.Runtime.Catalog.<InitActivation>d__66.MoveNext()
```

ReubenBond:
Fixes #3418

The ADO.NET storage provider was unable to correctly handle nested generic types.

This PR adds support for nested generics.

An ad-hoc test to compare the modified method with the existing method indicates that the modified method is backwards-compatible for simple generics. The string I tested was:
`UnitTests.Grains.GrainStorageGenericGrain`1[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]`

I'd prefer if we didn't have to go down this route of converting a Type into a string and then parsing it and reformatting it into a slightly prettier string.

Note that both the existing method and the modified method are unable to correctly distinguish arrays. I don't think we should add array support in this PR - it means fundamentally changing how the parsing works (no more `string.Split`).

### DIFF
--- a/src/AdoNet/Orleans.Persistence.AdoNet/Storage/Provider/AdoNetGrainStorage.cs
+++ b/src/AdoNet/Orleans.Persistence.AdoNet/Storage/Provider/AdoNetGrainStorage.cs
@@ -3,6 +3,7 @@ using Orleans.Providers;
 using Orleans.Runtime;
 using Orleans.Serialization;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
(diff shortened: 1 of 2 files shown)


veikkoeeva:
@ReubenBond Good catch, thanks. This could be an external helper function too. Or just have something that strips out assembly version etc.

PR #4145, opened 2018-03-06
Rename MembershipOptions to ClusteringOptions

sergeybykov:
Rename `MembershipOptions` to `ClusteringOptions` and `DevelopmentMembershipOptions` to `DevelopmentClusteringOptions`.


ReubenBond:
This may lead to confusion between this and the existing `ClusterOptions`.
I feel that `ClusterMembershipOptions` is an appropriate name to avoid any ambiguity.

PR #4146, opened 2018-03-06
Specify default ClusterId & simplify related extension methods

ReubenBond:
The rationale is that ClusterId only needs to be set when there's more than one cluster using the same storage (membership, reminders, persistence, multi-cluster channel).


sergeybykov:
See my comment in https://github.com/dotnet/orleans/pull/4136#issuecomment-371220515. I think it's fine and helpful to auto-set cluster ID for development case. I'm afraid that's too error prone to do for production deployments.

ReubenBond:
Sounds good to me :)

PR #4160, opened 2018-03-08
Validate that a ClusterId has been specified

ReubenBond:


ReubenBond:
@dotnet-bot test functional

(Intermitted xunit error with deps.json)

PR #4161, opened 2018-03-08
Support retries in ClusterClient.Connect(...)

ReubenBond:
For review only.

I didn't propagate the retry delegate all the way up to the API surface, but this is roughly what it might look like.

### DIFF
--- a/src/Orleans.Core/Core/ClusterClient.cs
+++ b/src/Orleans.Core/Core/ClusterClient.cs
@@ -104,7 +104,7 @@ namespace Orleans
         }
 
         /// <inheritdoc />
-        public async Task Connect()
+        public async Task Connect(Func<Exception, Task<bool>> retryFilter = null)
         {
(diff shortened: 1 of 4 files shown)


sergeybykov:
Don't we want the delegate to be
```
Func<Exception, Task<bool>>
```
instead, so that the delegate can introduce a non-blocking delay before retrying?

onionhammer:
@ReubenBond  I think there might be a race condition here; if there is a retry while the silo is still initializing.

I have this logic
```
            client.Connect(exception => Task.FromResult(DateTimeOffset.Now < end))
                  .Wait(cancellationToken);
``

where end = `DatetTime.Now.AddSeconds(5)`, If I start the silo at about the 3rd second the client comes back as `IsInitialized == true`, but I get an object reference error whenever I try to use it. If I start the silo while the client is not initializing the client works fine.

lmagyar:
@onionhammer See #4427.

Issue #4080, opened 2018-02-26
Improve persistent stream provider configuration pattern 


xiazen:
Persistent stream providers are very complex, it has many components which user can customize. Now the configuration of those components has scattered in three places, one is through inheritance, another one is through giant god  stream provider options which is intended to configure everything, third one is through the component's own option. All three of them has pain points, listed below. 

- Use inheritance,  instead of DI to configure some core component, such as `CheckpointerFactory`, `QueueMapperFactory`, `StreamFailureHandlerFactory` setters on `TAdapterFactory`. This make configuring harder, since if user want to configure those, they need to actually have a reasonable understanding of how Orleans streaming is wired up internally, to know which class to inherit and when to set those factories. Although this is true before, so one may argue it might be fine now. But the fact is, after provider been integrated with lifecycle, its configuration process has been changed entirely. For users who configured those factory through inheritance, they cannot use our extension methods to configure their custom stream provider, and the configuration their are familiar with is gone, and they need to get used to a new one which is almost as bad as before. So I think we should kill this pain point because we provided a new configuration method, it will suck if it is as bad as before. 

- The disadvantage of god stream provider options is its poor flexibility, extensibility and complexity. I will skipped listing its cons here since I think there's a consensus on Why god configuration is bad. 

- Configure one component through its own option is the way to go forward. The problem with it right now is poor discoverability, especially when the configuration is in three places. For example, to configure a queue balancer now, you need to first set the queue balancer type right in the stream provider options, then you need to configure the specific type of queue balancer options through DI, which there's no extra infrastructure to guide you on how to do that. 

I would argue some of those need to be improved in 2.0 scope, because we create a fuss to provide a new way of configuration, it will be bad if this new configuration method is as bad as before, and user need to learn about this new bad way and adapt to it, instead of happily adapt a better way. 

General direction is to uniform these three ways of configuration and develop one consistent and better way to configure stream providers, which also has good discoverability, extensibility and flexibility.

jason-bragg:
For a number of reasons I won't go into, most of the complexity in the PersistentStreamProvider is due to it's queue adapter and queue adapter factory, which, now that we're using DI and names services, can probably mostly be replaced by verious named services.

Some of these are:
- IQueueAdapter (which should probably be IQueueAdapterSender)
- IQueueAdapterReceiver
- IQueueAdapterCache (possibly just IQueueCache)
- IStreamQueueMapper
- IStreamFailureHandler

Each of the above components can be independantly configured.
[...]

PR #4164, opened 2018-03-08
Persistent stream configurators

xiazen:
Attempt to solve issue #4080 

Major interface and configurator class and its inheritance hierarchy
Silo side: 
```
//main interface and implementation
interface ISiloPersistentStreamConfigurator
interface IRecoverableStreamConfigurator:  ISiloPersistentStreamConfigurator
class SiloPersistentStreamConfigurator : ISiloPersistentStreamConfigurator
class SiloRecoverableStreamConfigurator : SiloPersistentStreamConfigurator 

//specific stream configurator
class SiloAzureQueueStreamConfigurator : SiloPersistentStreamConfigurator
class SiloEventHubStreamConfigurator : SiloRecoverableStreamConfigurator
..and other stream providers are similar 
```
Client side:
```
//main interface and implementation
interface IClusterClientPersistentStreamConfigurator
class ClusterClientPersistentStreamConfigurator : IClusterClientPersistentStreamConfigurator

//specific stream configurator
class ClusterClientAzureQueueStreamConfigurator : ClusterClientPersistentStreamConfigurator
class ClusetrClientEventHubStreamConfigurator : ClusterClientPersistentStreamConfigurator

..and other stream providers are similar 
```
The main idea is to use configurator/builder pattern to achieve a cleaner configuring pattern. Stream providers are complex which has many components, this PR try to use configurator/builder pattern to make configuring more modular, aka per component configuring. 

Stream providers are complex , where named service and named options can be utilized to make stream provider cleaner. But Rome isn't built in one day, we cannot make stream provider great in one day. So the strategy I propose here is to hide stream provider configuration under a builder pattern. So from the public surface, stream provider is already very modular. Under the public surface, we can make stream providers more modular without introducing much breaking changes. 

Sample call pattern:
```
  hostBuilder
  .AddAzureQueueStreams<AzureQueueDataAdapterV2>(StreamProviderName)
   .ConfigureAzureQueue(ob => ob.Configure(
                            options =>
                            {
                                options.ConnectionString = TestDefaultConfiguration.DataConnectionString;
                            }));
```

TODOs:
- check ConfigureNamedOptionForLogging called for each options correctly or not.
- should we configure default queue balancer?
- configure configurationValidators for options which needs it but haven't , such as sqsOptions and pubsubOptions


veikkoeeva:
Hmm, I remember at least @yevhen had ideas with regards this, I'll page. :)

xiazen:
@veikkoeeva thanks. I didn't know he had ideas on this. But welcome to page :) . We definitely need more eyes on this.

src/Orleans.Core/Streams/PersistentStreams/Options/PersistentStreamProviderOptions.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Orleans.Configuration
 {
-    public class PersistentStreamOptions
+    public class StreamInitializationOptions
```
jason-bragg: Suggest StreamStartupOptions or StreamLifecycleOptions, as these are more related to how the provider is started/stopped than it's initialization.

src/Orleans.Runtime.Abstractions/Streams/ISiloPersistentStreamConfigurator.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Text;
+
+namespace Orleans.Streams
+{
+    public interface ISiloPersistentStreamConfigurator
+    {
+        ISiloPersistentStreamConfigurator Configure<TOptions>(Action<OptionsBuilder<TOptions>> configureOptions)
+          where TOptions : class, new();
+
+        ISiloPersistentStreamConfigurator ConfigureComponent<TOptions, TComponent>(Action<OptionsBuilder<TOptions>> configureOptions, Func<IServiceProvider, string, TComponent> factory)
```
jason-bragg: nit:  As configuring options in the ConfigureComponent is optional, seems like listing the factory first then the "optional" options would make the call pattern more consistent.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -98,14 +100,19 @@ public class EventHubAdapterFactory : IQueueAdapterFactory, IQueueAdapter, IQueu
         internal ConcurrentDictionary<QueueId, EventHubAdapterReceiver> EventHubReceivers { get { return this.receivers; } }
         internal IEventHubQueueMapper EventHubQueueMapper { get { return this.streamQueueMapper; } }
 
-        public EventHubAdapterFactory(string name, EventHubStreamOptions options, IServiceProvider serviceProvider, SerializationManager serializationManager, ITelemetryProducer telemetryProducer, ILoggerFactory loggerFactory)
+        public EventHubAdapterFactory(string name, EventHubOptions ehOptions, EventHubReceiverOptions receiverOptions, EventHubStreamCacheOptions cacheOptions, StreamStatisticOptions statisticOptions,
```
benjaminpetit: Line overflow exception

src/AWS/Orleans.Streaming.SQS/Streams/SQSStreamBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+                .AddPersistentStreams(name, SQSAdapterFactory.Create)
+                .ConfigureServices(services =>
+                {
+                    services.ConfigureNamedOptionForLogging<SqsOptions>(name)
+                        .ConfigureNamedOptionForLogging<SimpleQueueCacheOptions>(name)
+                        .ConfigureNamedOptionForLogging<HashRingStreamQueueMapperOptions>(name);
+                });
+        }
+
+        public SiloSqsStreamConfigurator ConfigureAzureQueue(Action<OptionsBuilder<SqsOptions>> configureOptions)
```
benjaminpetit: Azure queue on AWS?

benjaminpetit: Same on the client

src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
```diff
@@ -0,0 +1,100 @@
...
+                configureOptions?.Invoke(services.AddOptions<TOptions>(this.name));
+            });
+            return this;
+        }
+    }
+    public class SiloPersistentStreamConfigurator : ISiloPersistentStreamConfigurator
+    {
+        protected readonly string name;
+        protected readonly ISiloHostBuilder siloBuilder;
+        public SiloPersistentStreamConfigurator(string name, ISiloHostBuilder siloBuilder)
```
benjaminpetit: NIT: new line...

src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
```diff
@@ -0,0 +1,100 @@
...
+        public ISiloPersistentStreamConfigurator ConfigureComponent<TComponent>(Func<IServiceProvider, string, TComponent> factory)
+           where TComponent : class
+        {
+            siloBuilder.ConfigureServices(services =>
+            {
+                services.AddSingletonNamedService<TComponent>(name, factory);
+            });
+            return this;
+        }
+        //try configure defaults if required is not configured
```
benjaminpetit: NIT: new line...

src/OrleansProviders/Streams/Common/RecoverableStreamOptions.cs
```diff
@@ -1,11 +1,15 @@
...
+using Orleans.Streams;
+using System;
 
 namespace Orleans.Configuration
 {
     /// <summary>
     /// Common stream provider settings shared by EventHubStreamProvider, MemoryStreamProvider and GeneratorStreamProvider
     /// </summary>
-    public class RecoverableStreamOptions : PersistentStreamOptions
+   /* public class RecoverableStreamOptions : PersistentStreamOptions
```
benjaminpetit: Why is it commented?

xiazen: should be removed

xiazen:
This PR is only 90% complete. So I will attach the review guide here. please Review in high-level public interface first and call pattern first. As long as we agreed on the public interfaces, then we can work on the details. 

I'm still work on migrate tests, and some small internal refactor.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubStreamOptions.cs
```diff
@@ -30,6 +34,27 @@ public class EventHubStreamOptions : RecoverableStreamOptions
         /// </summary>
         public bool StartFromNow { get; set; } = DEFAULT_START_FROM_NOW;
         public const bool DEFAULT_START_FROM_NOW = true;
+    }
+
+    public class EventHubStreamCacheOptions
```
jason-bragg: I understand these are the options for the eventhub cache, but these options are not really eventhub specific, and they seem to be more tightly related to a purge strategy which can be used with many other stream providers.  I'm wondering if something like DataAgeCachePurgeOptions might be more appropriate and allow the components using these settings to be used across a wider range of providers.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubCheckpointerOptions.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Orleans.Configuration
+{
+    public class EventHubCheckpointerOptions
```
jason-bragg: These options are more specific to using azure table storage for checkpointing than they are to eventhub.  They are only related to EH because Azure table is the default checkpoint storage for the EH stream provider.  I suggests AzureTableStreamCheckpointerOptions, or something of the sort.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubCheckpointerOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// Azure table storage connections string.
+        /// </summary>
+        [RedactConnectionString]
+        public string CheckpointConnectionString { get; set; }
```
jason-bragg: Since these are in their own options now (a good thing) we should remove the "Checkpoint" prefix from all the settings.  Those were only necessary for specificity in a large flat config object.

src/Orleans.Runtime.Abstractions/Streams/ISiloPersistentStreamConfigurator.cs
```diff
@@ -0,0 +1,66 @@
+﻿using Orleans.Configuration;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Orleans.Streams
+{
+    public interface ISiloPersistentStreamConfigurator
```
jason-bragg: By not having some sort of final 'build' step this abstraction lacks a means of checking the completeness of the various required configuration steps for a specific provider.  This may be addressable with validators, or some other means, so I'm not pushing back on this pattern, just noting a characteristic of this abstraction.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubStreamOptions.cs
xiazen: yes that's what I originally did. But then I think not all cache has purge logic. So how do I prevent users to configure that purge option with stream providers which doesn't support purging.

I'm still unsure which route to take. Just want to bring my thinking here

jason-bragg: Only exposing the configuration of this purge strategy in providers that support it (which is currently event hub atm, correct?) should be sufficiently limiting for now, imo.

The distinction here is the difference between what is being configured (a purge strategy) and the provider which utilizes it.  In theory, this purge strategy can be used by many other stream providers.  The fact that it is currently only being used by the event hub stream provider does not make it an event hub stream provider purge strategy.

jason-bragg: I may have misunderstood your question.. For eh stream provider, i'd suggest a ConfigureCachePurgeStrategy (or something of the sort) rather than the current ConfigureCache.  Sure, we'll want a ConfigureCache at some point to allow customization, but for configuring the defaults, since all we can configure is the purge strategy, that's all I'd suggest we expose at this time.

src/Orleans.Runtime.Abstractions/Streams/ISiloPersistentStreamConfigurator.cs
xiazen: agreed. And I'm thinking move this method to extension . Since this method are really just calling the other two methods

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubCheckpointerOptions.cs
xiazen: sounds good

src/Orleans.Runtime.Abstractions/Streams/ISiloPersistentStreamConfigurator.cs
xiazen: as we discussed offline, I'm hesitating to add that Validate or Builder method on the interface, since we have no way to force user to call that, and it is really the user's responsibility to call that. since we call validate in runtime. In short, it is not required for user to call it, then we shouldn't add it to the interface.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubStreamOptions.cs
xiazen: but we are not  just configuring purge here, we are also configure cache pressure monitors. 

The cache purge settings can be reused in MemoryStreamProvider. So does this mean we have two methods, one is configure purge, which both eh and memory stream provider has it, and another is cache monitor?

jason-bragg: Good point.  Yes, I think options should be associated with the components they configure, so options related to backpressure monitoring should be associated with the backpressure monitor.
We may need to change the default behavior here to better fit the pattern as well as future expectations.  Having no back pressure monitoring by default, expecting users to opt into the monitoring they expect, seems appropriate at this time.  So maybe an AddAveragingCachePressureMonitoring call with the related options?

xiazen:
@dotnet-bot test bvt please

src/AWS/Orleans.Streaming.SQS/Streams/SQSStreamBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+            this.Configure<SqsOptions>(configureOptions);
+            return this;
+        }
+        public SiloSqsStreamConfigurator ConfigureCacheSize(int cacheSize = SimpleQueueCacheOptions.DEFAULT_CACHE_SIZE)
+        {
+            this.Configure<SimpleQueueCacheOptions>(ob => ob.Configure(options => options.CacheSize = cacheSize));
+            return this;
+        }
+
+        public SiloSqsStreamConfigurator ConfigureQueueMapper(int numOfQueues = HashRingStreamQueueMapperOptions.DEFAULT_NUM_QUEUES)
```
jason-bragg: ConfigureQueueCount?  Sure, this is configuring the mapper, but users probably don't know what that is.

src/AWS/Orleans.Streaming.SQS/Streams/SQSStreamBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            this.Configure<HashRingStreamQueueMapperOptions>(ob => ob.Configure(options => options.NumQueues = numOfQueues));
+            return this;
+        }
+    }
+
+    public class ClusterClientSqsStreamConfigurator : ClusterClientPersistentStreamConfigurator
+    {
+        public ClusterClientSqsStreamConfigurator(string name, IClientBuilder builder)
```
jason-bragg: Client needs to know about the queue count as well (mapper settings), as it uses the mapper to determine which queues to send messages too.  This is probably the case with all stream providers that have configurable numbers of queues (azure, memory, ..)

jason-bragg: If we're only using the default queue count in all of our tests, we should probably configure a non-default in at least a few tests.  Don't think we need custom tests for this, just not use the defaults in all tests.

src/AWS/Orleans.Streaming.SQS/Streams/SQSStreamBuilder.cs
xiazen: agreed on name QueueMapping

src/Azure/Orleans.Streaming.EventHubs/Hosting/ClientBuilderExtensions.cs
```diff
@@ -8,13 +8,24 @@ namespace Orleans.Hosting
     public static class ClientBuilderExtensions
     {
         /// <summary>
-        /// Configure cluster client to use event hub persistent streams.
+        /// Configure cluster client to use event hub persistent streams. This return a configurator which allows further configuration
         /// </summary>
-        public static ClusterClientEventHubStreamConfigurator AddEventHubStreams(
+        public static ClusterClientEventHubStreamConfigurator ConfigureEventHubStreams(
```
jason-bragg: This works, but I worry that it won't really be clear if one is adding a new provider or configuring an existing one.

Suggest an override of AddEventHubStreams something like:

    public static ClusterClientEventHubStreamConfigurator AddEventHubStreams( 
            this IClientBuilder builder, 
            string name) 

or

    public static IClientBuilder AddEventHubStreams( 
            this IClientBuilder builder, 
            string name,
            Action<ClusterClientEventHubStreamConfigurator> configure) 

ConfigureEventHubStreams works though, if you're not fond of the suggested patterns.   Just suggestions.

xiazen: this pattern doesn't solve the problem of telling user what config is required to configure , which is mainly why I added AddXXXStreams and ConfigureXXXStreams.

xiazen: nvm. I see your point. Will add those overrides

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubStreamOptions.cs
xiazen: I'd love to go in that direction, like make all those component injectable. But with this overrlap with EventHUbQueueCacheFactory.AddCachePressureMonitor, ,and how eh streaming currently is. It is rather big change, with some source of truth problems.  I think we can do it in another pr after rc2 ?

src/AWS/Orleans.Streaming.SQS/Streams/SQSStreamBuilder.cs
xiazen: good point

sergeybykov:
@dotnet-bot test functional
Reason: to see if failure of `Tester.StreamingTests.PlugableQueueBalancerTests.PluggableQueueBalancerTestsWithMemoryStreamProvider.PluggableQueueBalancerTest_ShouldUseInjectedQueueBalancerAndBalanceCorrectly` will reproduce.

PR #4170, opened 2018-03-08
protobuf-net serializer

martinothamar:
Hey, considering using [protobuf-net](https://github.com/mgravell/protobuf-net) since it's so easy to use as opposed to the code-generation approach (imo). Thought this might be a nice addition here.


sergeybykov:
Thanks for submitting the PR!

Can you help me understand the benefits over https://github.com/dotnet/orleans/tree/master/src/Orleans.Serialization.Protobuf?

martinothamar:
I'd just prefer to modify existing POCO classes adding attributes rather than generating classes from .proto files, otherwise their basically the same thing afaik.

sergeybykov:
@dotnet-bot test functional

test/Orleans.ProtobufNet.Tests/ProtobufNetTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using Orleans.Runtime.Configuration;
+using Xunit;
+using System.IO;
+using FluentAssertions;
+
+namespace Orleans.ProtobufNet.Tests
+{
+    public class ProtobufNetTests
+    {
+        private readonly SerializationTestEnvironment environment;
```
xiazen: suggest turn `GoogleUtils.Tests.Serialization.ProtobufSerializationTests` tests into a test runner. So that both ProtobugSerializer and ProtobufNetSerializer can run the same set of tests, by using the same test runner. 

Also with the test runner, we can slowly adding tests to the test runner to increase coverage and all external serializer will benefit at the same time. 

But bottom line, the tests here only copied two of the tests from  `GoogleUtils.Tests.Serialization.ProtobufSerializationTests`, we should at least have them all?

martinothamar: What do you mean by a test-runner? They use different classes and different libraries/methods, so I'm having a hard time seeing how we can make the tests cover both Google Protobuf and protobuf-net. Any ideas? I can copy over all the other tests as well no problem, but I understand that ideally they would be the same tests

martinothamar: Or did you mean to simply have the protobuf-net tests copied into the same tests as the GoogleUtils one?
As in

```csharp
public void ProtobufSerializationTest_3_ProtoSerialization()
{
       var input = CreateAddressBook();
       var input2 = CreatePerson();
       ..... // Test both types and libraries
}
```

src/Orleans.Serialization.ProtobufNet/ProtobufNetSerializer.cs
```diff
@@ -0,0 +1,105 @@
...
+                // Special handling for null value. 
+                // Since in this ProtobufSerializer we are usually writing the data lengh as 4 bytes
+                // we also have to write the Null object as 4 bytes lengh of zero.
+                context.StreamWriter.Write(0);
+                return;
+            }
+            
+            using (var stream = new MemoryStream())
+            {
+                ProtoBuf.Serializer.Serialize(stream, item);
```
ReubenBond: I wonder if users would ever want to use a `RuntimeTypeModel` other than `RuntimeTypeModel.Default` and in that case whether or not it makes sense to pass that in via some options. I don't know for sure, since I'm not very experienced with protobuf-net.

ReubenBond:
Generally LGTM, good stuff :)

Is it possible we could avoid allocating that `MemoryStream`? Maybe that would require us to come up with a `Stream` adapter around `BinaryTokenStreamWriter`/`BinaryTokenStreamReader`.

I'm not too sure about the method used for deep copying objects, eg: is that a deep copy? It wouldn't handle immutable values.

This won't handle grain references, from what I can tell. That's ok - users just have to be aware.

test/Orleans.ProtobufNet.Tests/ProtobufNetTests.cs
xiazen: a test runner is a set of tests which can be applied to different implementation of the same component/interface. For example, `GrainPersistenceTestsRunner` , is used in AdoNet persistence tests and Azure persistence tests. 

Yes I understand those two serializer uses different libraries, but they are both `IExternalSerializer`, is it not? the test runner should be `ExternalSerializerTestRunner` which can be applied to all IExternalSerializer implementations. Does this make sense? 

I suggested copy tests over basically to increase the test coverage for this serializer when you don't have time to make a test runner.  Since they are both IExternalSerializer so they should pass the same set of tests. Is it not the case?

martinothamar: Ahh, thanks! I'll give it a shot later today

martinothamar:
Implemented the test-runner, proper DeepCopy and a benchmark:

``` ini

BenchmarkDotNet=v0.10.9, OS=Windows 10.0.16299
Processor=Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), ProcessorCount=8
Frequency=2742187 Hz, Resolution=364.6724 ns, Timer=TSC
  [Host]     : .NET Framework 4.7 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2633.0
  DefaultJob : .NET Framework 4.7 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2633.0


```
 |                Method |                Serializer |     Mean |     Error |    StdDev |  Gen 0 | Allocated |
 |---------------------- |-------------------------- |---------:|----------:|----------:|-------:|----------:|
 |   **SerializerBenchmark** |                   **Default** | **45.71 us** | **0.6701 us** | **0.5940 us** | **2.6855** |   **8.38 KB** |
 | DeserializerBenchmark |                   Default | 52.94 us | 1.2299 us | 2.5942 us | 2.8687 |   8.98 KB |
 |      FallbackFullLoop |                   Default | 19.25 us | 0.3779 us | 0.3350 us | 1.4343 |   4.44 KB |
 |   **SerializerBenchmark** | **IlBasedFallbackSerializer** | **46.42 us** | **0.4936 us** | **0.4617 us** | **2.6855** |   **8.38 KB** |
 | DeserializerBenchmark | IlBasedFallbackSerializer | 49.30 us | 0.5005 us | 0.4681 us | 2.8687 |   8.98 KB |
 |      FallbackFullLoop | IlBasedFallbackSerializer | 19.39 us | 0.2153 us | 0.2014 us | 1.4343 |   4.44 KB |
 |   **SerializerBenchmark** |               **ProtoBufNet** | **44.71 us** | **0.1574 us** | **0.1314 us** | **2.6855** |   **8.38 KB** |
 | DeserializerBenchmark |               ProtoBufNet | 50.96 us | 0.9806 us | 1.0070 us | 2.8687 |   8.98 KB |
 |      FallbackFullLoop |               ProtoBufNet | 20.44 us | 0.1032 us | 0.0915 us | 1.4343 |   4.44 KB |

Probably could make this a little faster by avoiding that MemoryStream allocation, but not sure what the best way to do that would be

ReubenBond:
@xiazen if this looks good to you, too, and we're happy to introduce a new package this late in 2.0 cycle, then please squash+merge

xiazen:
Sorry for the late reply. was oof yesterday. yep looks good to me.

Issue #4158, opened 2018-03-07
Dns.GetHostAddressesAsync failure preventing Silo start


jsteinich:
I'm getting the following stack trace when attempting to start a silo. I believe the cause is the default value for `AdvertisedIPAddress` in `EndpointOptions`. I'm trying to run the silo (rc1)  in an AWS Fargate (managed container service) task. I can try to troubleshoot the underlying error more, but I'm pretty sure I'll need to set the `AdvertisedIPAddress` manually anyways, so having the default specified in a different manner (as suggested in https://github.com/dotnet/orleans/issues/4115) may make more sense.

```
Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.AggregateException: One or more errors occurred. (No such device or address) ---> System.Net.Internals.SocketExceptionFactory+ExtendedSocketException: No such device or address
at System.Net.Dns.InternalGetHostByName(String hostName, Boolean includeIPv6)
at System.Net.Dns.ResolveCallback(Object context)
--- End of stack trace from previous location where exception was thrown ---
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
at System.Net.Dns.HostResolutionEndHelper(IAsyncResult asyncResult)
at System.Net.Dns.EndGetHostAddresses(IAsyncResult asyncResult)
at System.Net.Dns.<>c.<GetHostAddressesAsync>b__25_1(IAsyncResult asyncResult)
at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
--- End of stack trace from previous location where exception was thrown ---
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
at Orleans.Runtime.Configuration.ConfigUtilities.<ResolveIPAddress>d__18.MoveNext()
--- End of inner exception stack trace ---
at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
at Orleans.Configuration.EndpointOptions..ctor()
--- End of inner exception stack trace ---
[...]

PR #4171, opened 2018-03-08
Do not call ResolveIPAddress in EndpointOptions constructor

benjaminpetit:
Related issue #4158 

* Removed call to `ConfigUtilities.ResolveIPAddress` in `EndpointOptions` constructor
* `ConfigUtilities.ResolveIPAddress` no longer use DNS if no hostname provided


test/NonSilo.Tests/SiloHostBuilderTests.cs
```diff
@@ -168,6 +168,8 @@ private static void RemoveConfigValidators(IServiceCollection services)
         {
             var validators = services.Where(descriptor => descriptor.ServiceType == typeof(IConfigurationValidator)).ToList();
             foreach (var validator in validators) services.Remove(validator);
+            // Configure endpoints because validator is set just before Build()
+            services.Configure<EndpointOptions>(options => options.AdvertisedIPAddress = IPAddress.Loopback);
```
ReubenBond: This is a slightly confusing place to put this code, since the method is called `RemoveValidators`.

You could rename the method or put the code somewhere else

ReubenBond:
If `AdvertisedIPAddress` is not set, the result is a `NullReferenceException` during startup.
image

benjaminpetit:
The validator should check that this property was set. Is it possible that this test removes it?

Issue #4174, opened 2018-03-08
When Silo shutdown, Grain deactivation is happening too late in the lifecycle


benjaminpetit:
Silo will accept new activation in the `Active` state, but will deactivate its grain activation only at stage `RuntimeServices` which is happening after the `ApplicationServices` state.

It will cause issue if the grain implementation rely on a component that is dependant on the `ApplicationServices` state (like storage providers...)

sergeybykov:
Resolved via #4177.

PR #4177, opened 2018-03-08
When the silo shutdown, deactivate grain activations at an earlier stage

benjaminpetit:
Proposed fix for #4174


ReubenBond:
This seems like the right move.

Additionally, though, what is supposed to happen during non-graceful shutdown - we don't need to gracefully deactivate the grains, but we do need to stop them from processing messages/tasks. Is that already taken care of?

sergeybykov:
>What is supposed to happen during non-graceful shutdown - we don't need to gracefully deactivate the grains, but we do need to stop them from processing messages/tasks. Is that already taken care of?

We prevent placement of new grain by switching silo to the `ShuttingDown` state. I don't think we want to stop them from processing, as they might still call each other as part of their deactivation logic.

PR #4180, opened 2018-03-09
Set GrainService.Status to Started in the base implementation of StartInBackground()

### LINKED ISSUE #4128: How to let timers survive a (graceful) silo shutdown
I have a Game grain, which manages the logic for a turn-based game between two players. Inside this grain, a timer is used to measure the time for each turn and forcefully end a player's turn if they fail to play in a timely manner.

If this grain gets deactivated (even gracefully), the timer will cease to exist. The creation of a new activation of this grain will depend on it receiving external calls, which means it may not happen until one of the players makes a move, which defeats the purpose of the timer. I can save the game's state and let it continue off from where the old silo was shut down, but I currently have no way of bringing the timer back online.

One way is to constantly poll the grain from another grain. This fails because the other grain is also subject to the same limitations as the original (namely, it may get deactivated with no way to restore its timers).

Another way is to have a second grain (let's call it pinger). One could have the game ask the pinger (on deactivation) to call it back after a second or so. This is likely to succeed most of the time, but if the pinger happens to be on the same silo and get deactivated after the game, it will still fail.

Yet another may be to have the game call itself. However, I'm not entirely sure this is a stable solution as I don't know how the silos handle the case of a grain receiving a call while it's already deactivating but not yet fully deactivated. If the grain is removed from the global directory *before* the OnDeactivateAsync call this may succeed. I just thought of this while writing this question, so I have yet to see if this actually works.

A reminder is not appropriate in this scenario because the minimum interval for a reminder is 60 seconds IIRC. An entire turn takes about half a minute.

Any other ideas on how to handle this (and other similar cases)?

sergeybykov:
Fixes #4128.
This should help prevent mistakes of not setting the status to started, which leads to not receiving cluster repartitioning notifications.

### DIFF
--- a/src/Orleans.Runtime/Services/GrainService.cs
+++ b/src/Orleans.Runtime/Services/GrainService.cs
@@ -81,8 +81,15 @@ namespace Orleans.Runtime
             return Task.CompletedTask;
         }
 
-        /// <summary>Deferred part of initialization that executes after the service is already started (to speed up startup)</summary>
-        protected abstract Task StartInBackground();
+        /// <summary>
+        /// Deferred part of initialization that executes after the service is already started (to speed up startup).
+        /// Sets Status to Started.
+        /// </summary>
+        protected virtual Task StartInBackground()
+        {
+            Status = GrainServiceStatus.Started;
+            return Task.CompletedTask;
+        }
 
         /// <summary>Invoked when service is being stopped</summary>
         public virtual Task Stop()


sergeybykov:
@dotnet-bot test functional
Reason: unrelated test failure: Tester.HeterogeneousSilosTests.HeterogeneousTests.MergeGrainResolverTest

PR #4185, opened 2018-03-09
[WIP] Shutdown forward workaround

benjaminpetit:
Related issue: #2287 

It's not a fix, it's a workaround. Not pretty, but I believe it has some value.

The issue is pretty "simple" when the silo shutdown, it will

1. Deactivate all grain activations 
2. Then forward the pending requests.

The issue is that we do not await step 2. And it is not possible without some heavy code change right now. The proposed workaround is just to add a delay after deactivating grains, so the system has some time to send the pending requests.

Note that it is happening after we set the silo status to "stopping", so other silos should not try to place new activation on it.

I did not enabled `SiloGracefulShutdown_ForwardPendingRequest` since it is still failing sometimes... but at least it pass most of the time now.

### DIFF
--- a/src/Orleans.Runtime/Silo/Silo.cs
+++ b/src/Orleans.Runtime/Silo/Silo.cs
@@ -774,8 +774,14 @@ namespace Orleans.Runtime
                     // Write "ShutDown" state in the table + broadcast gossip msgs to re-read the table to everyone
                     await scheduler.QueueTask(this.membershipOracle.ShutDown, this.membershipOracleContext)
                         .WithTimeout(stopTimeout, $"MembershipOracle Shutting down failed due to timeout {stopTimeout}");
+                    // Awful, but some components queue a task when receiving silo status notification,
+                    // add a delay here to give them a chance to execute this task
+                    await Task.Delay(1000);
                     // Deactivate all grains
                     SafeExecute(() => catalog.DeactivateAllActivations().WaitWithThrow(stopTimeout));
+                    // Awful again, but since we don't track the forwarding of the activation pending messages
+                    // add a small delay here so silo has some time to send them
+                    await Task.Delay(1000);
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -774,8 +774,14 @@ private Task OnRuntimeInitializeStop(CancellationToken ct)
                     // Write "ShutDown" state in the table + broadcast gossip msgs to re-read the table to everyone
                     await scheduler.QueueTask(this.membershipOracle.ShutDown, this.membershipOracleContext)
                         .WithTimeout(stopTimeout, $"MembershipOracle Shutting down failed due to timeout {stopTimeout}");
+                    // Awful, but some components queue a task when receiving silo status notification,
+                    // add a delay here to give them a chance to execute this task
```
xiazen: does this call for including those components shutdown logic into lifecycle so we don't need to do this hack?

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -774,8 +774,14 @@ private Task OnRuntimeInitializeStop(CancellationToken ct)
...
+                    await Task.Delay(1000);
                     // Deactivate all grains
                     SafeExecute(() => catalog.DeactivateAllActivations().WaitWithThrow(stopTimeout));
+                    // Awful again, but since we don't track the forwarding of the activation pending messages
+                    // add a small delay here so silo has some time to send them
+                    await Task.Delay(1000);
```
jdom: Now that we have a proper silo lifecycle, can't you just perform these interlocked shutdown steps steps by using different stages, even when they are performed by the same component?
Graceful shutdown (and faster startup) were the main reasons we wanted a proper lifecycle in the first place

benjaminpetit:
Not relevant anymore

PR #4192, opened 2018-03-12
Fix EHStreamProviderCheckpointTests

xiazen:
Its dispose method was failing because wrong connection string parsed.


xiazen:
we should have an issue to include those eh streaming tests into nightly. Currently it is excluded because tests depends on timing very much so fail intermittently.

Issue #3830, opened 2018-01-05
Calling grain method using GrainClient hangs after running for some time


chris-eaton:
I am using ASP.NET and I have a controller method uses `GrainClient` within   If I run ASP.NET then instantly call this controller, the call to the grain works correctly.  

At some point in the future this stops working, I havent tracked down exactly how long it takes but certainly if I leave it running over night it fails in the morning.

If I step through the code I get a grain reference of  `{GrainReference:*grn/A97D1470/00000000}`. 

If I turn on all exceptions the call to the grain throws this exception

> System.Net.Sockets.SocketException: 'An existing connection was forcibly closed by the remote host'

The NativeErrorCode is `10054`

Oddly enough this exception doesnt seem to bubble up through ASP.NET and make the call fail.

sergeybykov:
I'm not sure I understand. What do you mean by:
>If I turn on all exceptions 

Are you awaiting the grain calls? If you do, then every failed call should throw an exception. Is this not what you are seeing?

chris-eaton:
Sorry, what I meant was no exceptions were being thrown so I turned on "Catch all exceptions" in the exceptions options of Visual Studio. But still no exceptions were thrown.

One thing I had noticed in another issue was that timeouts for Grain calls are quite large when ran under Debug? Is this true?

Here is my code for full transparency

```      
public async Task SaveRouteTemplate(RouteTemplate routeTemplate)
{
    if (this.bifrostClient.GrainClient == null)
    {
        throw new Exception("Bifrost client not started");
    }

    var geocodeGrain = this.bifrostClient.GrainClient.GetGrain<IGeocodeGrain>(0);

    foreach (var stop in routeTemplate.RouteStops)
    {
        try
        {
            stop.GeocodedAddress = await geocodeGrain.GetGeocode(stop.CenterPoint.Latitude, stop.CenterPoint.Longitude);
            stop.GeocodedAddressId = stop.GeocodedAddress.Id;
        }
        catch (Exception e)
        {
            Console.WriteLine(e);
        }
    }

    await this.tservice.AddRouteTemplate(routeTemplate);
}
```

chris-eaton:
After probably 15-30 minutes the call finally timed out with this message

[...]

PR #4193, opened 2018-03-12
Add config option to override MessagingOptions.ResponseTimeout when the debugger is attached

benjaminpetit:
Fix for #3830 

Please let me know if what I did is not "option" compatible, but I think this is the cleanest way to implement such flag.

### DIFF
--- a/src/Orleans.Core/Configuration/Options/MessagingOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/MessagingOptions.cs
@@ -10,20 +10,21 @@ namespace Orleans.Configuration
     {
         /// <summary>
         /// The ResponseTimeout attribute specifies the default timeout before a request is assumed to have failed.
+        ///<seealso cref="ResponseTimeoutWithDebugger"/>
         /// </summary>
         public TimeSpan ResponseTimeout
         {
-            get { return (DebuggerOverridesTimeout && Debugger.IsAttached) ? RESPONSE_TIMEOUT_WITH_DEBUGGER : responseTimeout; }
+            get { return Debugger.IsAttached ? RESPONSE_TIMEOUT_WITH_DEBUGGER : responseTimeout; }
             set { this.responseTimeout = value; }
         }
         public static readonly TimeSpan DEFAULT_RESPONSE_TIMEOUT = TimeSpan.FromSeconds(30);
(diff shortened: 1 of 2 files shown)


src/Orleans.Core/Configuration/Options/MessagingOptions.cs
```diff
@@ -11,8 +11,20 @@ public abstract class MessagingOptions
...
+        }
+        public static readonly TimeSpan DEFAULT_RESPONSE_TIMEOUT = TimeSpan.FromSeconds(30);
+        private TimeSpan responseTimeout = DEFAULT_RESPONSE_TIMEOUT;
+
+        /// <summary>
+        /// By default, the value from<see cref="ResponseTimeout"/> will be ignored and set to 30 minutes
+        /// if the debugger is attached. Set this flag to false to always use <see cref="ResponseTimeout"/>
+        /// </summary>
+        public bool DebuggerOverridesTimeout { get; set; } = true;
+        public static readonly TimeSpan RESPONSE_TIMEOUT_WITH_DEBUGGER = TimeSpan.FromMinutes(30);
```
ReubenBond: It seems that we should just lift the whole debug override code out to the top level so we can set it in our tests and others in theirs

src/Orleans.Core/Configuration/Options/MessagingOptions.cs
```diff
@@ -10,9 +10,22 @@ public abstract class MessagingOptions
     {
         /// <summary>
         /// The ResponseTimeout attribute specifies the default timeout before a request is assumed to have failed.
+        ///<seealso cref="ResponseTimeoutWithDebugger"/>
         /// </summary>
-        public TimeSpan ResponseTimeout { get; set; } = DEFAULT_RESPONSE_TIMEOUT;
-        public static readonly TimeSpan DEFAULT_RESPONSE_TIMEOUT = Debugger.IsAttached ? TimeSpan.FromMinutes(30) : TimeSpan.FromSeconds(30);
+        public TimeSpan ResponseTimeout
+        {
+            get { return Debugger.IsAttached ? RESPONSE_TIMEOUT_WITH_DEBUGGER : responseTimeout; }
```
lmagyar: @benjaminpetit Shouldn't you use the `ResponseTimeoutWithDebugger` **property** instead of the `RESPONSE_TIMEOUT_WITH_DEBUGGER` **constant**?

PR #4201, opened 2018-03-13
Bboyle1234 shutdowndocs

bboyle1234:


src/Documentation/2.0/SiloShutdown.md
```diff
@@ -0,0 +1,131 @@
+---
+layout: page
+title: Shutting down Orleans
+---
+
+This document explains how to gracefully shutdown an Orleans silo before application exit, first as a Console app, and then as a Docker container app.
+
+# Graceful shutdown - Console app
+The following code shows how to gracefully shutdown an orleans silo console app in response to the user pressing Ctrl+C, which generates the `Console.CancelkeyPress` event.
```
sergeybykov: Capitalize "Orleans".

src/Documentation/2.0/SiloShutdown.md
```diff
@@ -0,0 +1,131 @@
...
+
+Normally when that event handler returns, the application will exit immediately, causing a catastrophic Orleans silo crash and loss of data. But in the sample code below, we set `a.Cancel = true;` to prevent the application closing before the Orleans silo has completed its graceful shutdown.
```
sergeybykov: To be precise, in an ungraceful shutdown you lose in-memory state. "Loss of data" has a worse connotation to me that somebody might misinterpret as a loss of persistent state.

src/Documentation/2.0/SiloShutdown.md
```diff
@@ -0,0 +1,131 @@
...
+                    }
+                }
+                /// Event handler execution exits immediately, leaving the silo shutdown running on a background thread,
+                /// but the app doesn't crash because a.Cancel has been set = true
+            };
+        }
+
+        static ISiloHost CreateSilo() {
+            return new SiloHostBuilder()
+                .Configure(options => options.ClusterId = "mysilo")
```
sergeybykov: "mysilo" is a bit misleading here when used as a cluster ID. Easy to misread it as a different cluster ID per silo. Maybe "mytestcluster" instead?

src/Documentation/2.0/SiloShutdown.md
```diff
@@ -0,0 +1,131 @@
...
+        static void SetupApplicationShutdown() {
+            /// Capture the user pressing Ctrl+C
+            Console.CancelKeyPress += (s, a) => {
+                /// Prevent the application from crashing ungracefully.
+                a.Cancel = true;
+                /// Don't allow the following code to repeat if the user presses Ctrl+C repeatedly.
+                lock (syncLock) {
+                    if (!siloStopping) {
+                        siloStopping = true;
+                        Task.Run(StopSilo);
```
sergeybykov: Since the `Task` returned by `Task.Run()` isn't awaited, it's better to add `.Ignore()` to it, to make sure any exception it might throw will not get to the finalizer thread.

src/Documentation/2.0/SiloShutdown.md
```diff
@@ -0,0 +1,131 @@
...
+        static ISiloHost silo;
+        static bool siloStopping = false;
+        static readonly object syncLock = new object();
+
+        static void Main(string[] args) {
+
+            SetupApplicationShutdown();
+
+            silo = CreateSilo();
+            Task.Run(StartSilo);
```
sergeybykov: Why not simply `Task.Run(StartSilo).Wait();`? Won't need `_siloStarted ` at all then.

src/Documentation/2.0/SiloShutdown.md
sergeybykov: As a general rule, we try to put every sense on a new line. That helps with diffing future edits.

bboyle1234:
Those changes look good @sergeybykov , thanks for commenting them so I know where your head's at. Is there anything for me to do here?

sergeybykov:
If you could address these comments and update the PR (with another commit), we'll merge it. Thanks!

sergeybykov:
If you don't think you'll have time to address the comments, we can try to push a commit with those changes to your branch. Let us know. Thanks.

sergeybykov:
Thank you, @bboyle1234!

PR #4214, opened 2018-03-13
Fix ADO.NET Reminder configuration & re-enable tests

### LINKED ISSUE #4207: [2.0 RC2] Ado.Net reminders provider is broken in RC2
After upgrade to RC2 I have exception when using Ado.Net reminders provider. I configure it as:

```
        .UseAdoNetReminderService(options =>
         {
                 options.ConnectionString = ConnectionString;
                 options.Invariant = "System.Data.SqlClient";
          })
```

and on `_siloHost = builder.Build();` I got: 
```
Orleans.Runtime.OrleansConfigurationException
  HResult=0x80131500
  Message=Invalid AdoNetReminderTableOptions values for AdoNetReminderTable. Invariant is required.
  Source=Orleans.Reminders.AdoNet
  StackTrace:
   at Orleans.Configuration.AdoNetReminderTableOptionsValidator.ValidateConfiguration()
   at Orleans.Hosting.SiloHostBuilder.ValidateSystemConfiguration(IServiceProvider serviceProvider)
   at Orleans.Hosting.SiloHostBuilder.Build()
   at <....>.Orleans.Start() in <....>\Orelans.cs:line 53
   at <....>.Program.Service.Start() in <....>\Program.cs:line 28
   at <....>.Program.<>c.<Main>b__1_3(Service tc) in <....>\Program.cs:line 53
   at Topshelf.ServiceConfiguratorExtensions.<>c__DisplayClass2_0`1.<WhenStarted>b__0(T service, HostControl control)
   at Topshelf.Builders.DelegateServiceBuilder`1.DelegateServiceHandle.Start(HostControl hostControl)
   at Topshelf.Hosts.ConsoleRunHost.Run()

```

ReubenBond:
Fixes #4207 

If this issue affects you, there's a workaround in #4207

### DIFF
--- a/src/AWS/Orleans.Reminders.DynamoDB/SiloHostBuilderReminderExtensions.cs
+++ b/src/AWS/Orleans.Reminders.DynamoDB/SiloHostBuilderReminderExtensions.cs
@@ -43,7 +43,7 @@ namespace Orleans.Hosting
         public static IServiceCollection UseDynamoDBReminderService(this IServiceCollection services, Action<DynamoDBReminderStorageOptions> configure)
         {
             services.AddSingleton<IReminderTable, DynamoDBReminderTable>();
-            services.Configure(configure);
+            services.Configure<DynamoDBReminderStorageOptions>(configure);
             services.ConfigureFormatter<DynamoDBReminderStorageOptions>();
             return services;
         }
--- a/src/AdoNet/Orleans.Reminders.AdoNet/SiloHostBuilderReminderExtensions.cs
+++ b/src/AdoNet/Orleans.Reminders.AdoNet/SiloHostBuilderReminderExtensions.cs
@@ -42,7 +42,7 @@ namespace Orleans.Hosting
             services.AddSingleton<IReminderTable, AdoNetReminderTable>();
             services.ConfigureFormatter<AdoNetReminderTableOptions>();
             services.AddSingleton<IConfigurationValidator, AdoNetReminderTableOptionsValidator>();
-            services.Configure(configureOptions);
+            configureOptions(services.AddOptions<AdoNetReminderTableOptions>());
             return services;
         }
     }
--- a/src/Azure/Orleans.Reminders.AzureStorage/SiloHostBuilderReminderExtensions.cs
+++ b/src/Azure/Orleans.Reminders.AzureStorage/SiloHostBuilderReminderExtensions.cs
@@ -62,7 +62,7 @@ namespace Orleans.Hosting
         public static IServiceCollection UseAzureTableReminderService(this IServiceCollection services, Action<AzureTableReminderStorageOptions> configure)
         {
             services.AddSingleton<IReminderTable, AzureBasedReminderTable>();
-            services.Configure(configure);
+            services.Configure<AzureTableReminderStorageOptions>(configure);
             services.ConfigureFormatter<AzureTableReminderStorageOptions>();
             return services;
         }
--- a/test/TesterAdoNet/ReminderTests_AdoNet.cs
+++ b/test/TesterAdoNet/ReminderTests_AdoNet.cs
@@ -1,8 +1,12 @@
 //#define USE_SQL_SERVER
 
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.Memory;
 using Orleans;
+using Orleans.Hosting;
 using Orleans.Runtime.Configuration;
 using Orleans.Tests.SqlUtils;
 using Orleans.TestingHost;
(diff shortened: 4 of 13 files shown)


ReubenBond:
@dotnet-bot test this please

sergeybykov:
The tests are failing in VSO.

ReubenBond:
Thanks, @sergeybykov - I'll take a look. They succeeded locally

src/AdoNet/Orleans.Reminders.AdoNet/SiloHostBuilderReminderExtensions.cs
```diff
@@ -42,7 +42,7 @@ public static IServiceCollection UseAdoNetReminderService(this IServiceCollectio
             services.AddSingleton<IReminderTable, AdoNetReminderTable>();
             services.ConfigureFormatter<AdoNetReminderTableOptions>();
             services.AddSingleton<IConfigurationValidator, AdoNetReminderTableOptionsValidator>();
-            services.Configure(configureOptions);
+            configureOptions(services.AddOptions<AdoNetReminderTableOptions>());
```
xiazen: wow this is dangerous mistake which really easy to miss, since it doesn't have compile time validation. I wonder do we make the same mistake in our other part of code base..

xiazen: how do we find a way to solve the ambiguous between `Action<TOptions>` and `Action<OptionsBuilder<TOptions>>`. If we are making this mistake, our users would probably make the same mistake somewhere else.. @sergeybykov @BenjaminGibbs @jason-bragg @jdom

xiazen: The PR LGTM, but just want to bring up this issue.

sergeybykov: I suspect you meant @benjaminpetit, not @BenjaminGibbs? :-)

xiazen: yep :) thanks for pointing this out. And apologize to the other Ben. But welcome to leave feedback if you want :)

jdom: I'm trying to find places where end-users would use the wrong signature too, but can'd think of any. It's not common to define a method that takes and OptionsBuilder, only infrastructure code like ours or 3rd party extensions would do this, to provide syntactic sugar for the end user. End users would just either use our extension methods or call `.Configure(Action<TOptions>)` and not use the builder itself.  Still, this is the API for Options, not Orleans, not much we should do if it was an issue (which I still don't believe it is).

src/AWS/Orleans.Reminders.DynamoDB/SiloHostBuilderReminderExtensions.cs
```diff
@@ -43,7 +43,7 @@ public static ISiloHostBuilder UseDynamoDBReminderService(this ISiloHostBuilder
         public static IServiceCollection UseDynamoDBReminderService(this IServiceCollection services, Action<DynamoDBReminderStorageOptions> configure)
         {
             services.AddSingleton<IReminderTable, DynamoDBReminderTable>();
-            services.Configure(configure);
+            services.Configure<DynamoDBReminderStorageOptions>(configure);
```
jdom: This is not required, is it? Is there a reason why being explicit is better here? Not pushing back, just commenting, as I don't necessarily find it better (unless it was required to compile to the right thing)

src/AdoNet/Orleans.Reminders.AdoNet/SiloHostBuilderReminderExtensions.cs
xiazen: Yeah the only thing which bothered me as an user, is when both has defaults to null . For example : 
```
UseXXX(Action<TOptions> configure = null);
UseXXX(Action<OptionBuilder<TOptions>> configure = null)
```

As a user, if I want to call `UseXXX()` I will have to cast it somehow to determine which one. I definitly encounterred this before when writting tests, not sure if there's still occurence like this in our code base

ReubenBond: Yes, we should be consistent with our APIs - no ambiguities. Only the `Action<T>` overload should have a default value in my opinion. Someone can change those things in another PR

xiazen: yeah totally. I didn't intend to push those in this PR. it just triggered my thought. This pr is good.

xiazen: I can merge it if tests are passing in vso. @sergeybykov mentioned they weren't

jdom: oh, sorry, just read this last comment. Nevertheless I did check that a newer VSO build was passing before merging

ReubenBond: I should have updated the thread to say that the tests were passing - my bad.

Issue #3480, opened 2017-10-02
Extending AddGrainCallFilter to define interface to intercept.


ronMilne:
Would it be possible to extend AddGrainCallFilter so that along with the IGrainCallFilter one could also specify the interface to intercept. Something akin to.

            public IServiceProvider ConfigureServices(IServiceCollection services)
            {
                services.AddGrainCallFilter<InterfaceToIntercept, IGrainFilterImplementation>();
                return services.BuildServiceProvider();
            }

Ron Milne.

ReubenBond:
We could support by adding a helper method which added a silo-wide interceptor that filtered down to only certain interfaces, but it would execute that check for every single call. Another option is to overhaul how grain call filtering works on a per-grain level so that only grains which implement that interface would have such a filter. I'm a little reluctant to dive into it at the moment, though, until I understand the problem better.

Could you help me to understand the use case?

Would the step after this be to add an overload which also specifies the method?

bytelabsco:
@ReubenBond and I talked briefly about this tonight, so I wanted to provide a use-case I have.

I'm using event sourcing with JournaledGrains and a custom log consistency provider which stores each set of events as it's own commit.  This commit is tracked  both on a grain level and global level.  For every successful commit, I'd like to place the commit information onto a Queue that can then be processed by a process outside of orleans to update a read model that I use for easy querying.  I did not want to bake this functionality into the log consistency provider, so being able to add a filter specific to the commit grain would allow me to place the commit information onto a queue.

ronMilne:
Hi Reuben,

My initial thoughts where for AOP semantics to be applied to a grain interface (Telemetry, caching, security, etc). My original need was to augment test data with real time data for testing purposes (for particular methods), rather than using dependency injection.

ReubenBond:
I've been considering how to tackle this.

My thought is that we should make this feature easy for users to implement without implementing it in the core unless it's really necessary.

Currently, this is not easy to implement for users, because `IGrainCallContext.Method` returns the *implementation* method when it's used in `IIncomingGrainCallFilter` and the interface method only when it's used in `IOutgoingGrainCallFilter`. This asymmetry is not desirable.

[...]

PR #4216, opened 2018-03-14
Grain Call Filtering: discriminate InterfaceMethod and ImplementationMethod

ReubenBond:
Supports #3480 

See this comment for context: https://github.com/dotnet/orleans/issues/3480#issuecomment-372865472


src/Orleans.Core/Core/GrainCallFilterExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+            throw new NotSupportedException($"{nameof(AddGrainCallFilter)} is no longer supported. Use ISiloHostBuilder.AddIncomingGrainCallFilter(...) instead.");
+        }
+
+        /// <summary>
+        /// Adds an <see cref="IIncomingGrainCallFilter"/> to the filter pipeline.
+        /// </summary>
+        /// <param name="services">The service collection.</param>
+        /// <param name="filter">The filter.</param>
+        /// <returns>The service collection.</returns>
+        internal static IServiceCollection AddIncomingGrainCallFilter(this IServiceCollection services, IIncomingGrainCallFilter filter)
```
stephenlautier: please can  you change these to `public` as they are currently not usable (same for the below)

ReubenBond: @stephenlautier, I'll consider it since this seems to be a popular request. Can you use the methods on ISiloHostBuilder instead?

stephenlautier: Ahh oki, wasn't aware there was another way of doing it. My bad misread the Obsolete message and tried to use the internal instead. So its good enough, thanks!

Issue #4129, opened 2018-03-05
Bug: PerfCounterEnvironmentStatistics never reports CpuUsage


SebastianStehle:
I think the registration is wrong, because OnStart is never called. 

cpuCounterPF != null, but cpuUsageTimer == null

benjaminpetit:
You are right, it seems we are missing something like `services..AddFromExisting<ILifecycleParticipant<ISiloLifecycle>, PerfCounterEnvironmentStatistics>()` in the extension methods

sergeybykov:
Resolved via #4148.

SebastianStehle:
Still does not work, I think the PR should fix it: https://github.com/dotnet/orleans/pull/4219

benjaminpetit:
Tested manually and it seems that now it works on silo and clients. Thanks!

PR #4219, opened 2018-03-14
Fix for #4129

SebastianStehle:


xiazen:
LGTM but it has compile errors : 
error CS0023: Operator '.' cannot be applied to operand of type 'void' [D:\j\workspace\functional_pr---83d31c72\src\Orleans.TelemetryConsumers.Counters\Orleans.TelemetryConsumers.Counters.csproj]

PR #4234, opened 2018-03-16
Remove Service Fabric configuration helpers & update sample

### LINKED ISSUE #4226: ServiceFabric :  A cluster id has not been configured, BUG
what's error with me?
how to do it?

```
var listener = OrleansServiceListener.CreateStateless(
                (serviceContext, builder) =>
                {
                    // Optional: use Service Fabric for cluster membership.
                     builder.UseServiceFabricClustering(serviceContext);

                    config.Globals.ReminderServiceType = ReminderServiceProviderType.Disabled;

                    // Service Fabric manages port allocations, so update the configuration using those ports.
                    builder.Configure<EndpointOptions>(options => {
                        options.AdvertisedIPAddress = IPAddress.Loopback;
                        options.ConfigureFromServiceContext(serviceContext);
                    });

                    builder.AddIncomingGrainCallFilter<ExceptionConversionFilter>();

                    // Tell Orleans to use this configuration.
                    builder.UseConfiguration(config);

                    builder.ConfigureServices(ConfigureDelegate);

                    // Add your application assemblies.
                    builder.ConfigureApplicationParts(parts =>
                    {
                        // Alternative: add all loadable assemblies in the current base path (see AppDomain.BaseDirectory).
                        parts.AddFromApplicationBaseDirectory().WithCodeGeneration();
                    });
                });
```

ReubenBond:
Fixes #4226

The configuration pattern used in the current release oversimplifies endpoint configuration. Instead, I believe we should remove the configuration helper and update the sample to show how the user can configure endpoints on Service Fabric in a less magic manner.

I added an alternative observer manager to the sample because the original one moved to legacy and this version is superior. I'd be happy to remove it entirely and simplify the sample so that it does not have any observers.


Samples/2.0/ServiceFabric/Stateless/StatelessCalculatorService/StatelessCalculatorService.cs
```diff
@@ -37,21 +41,33 @@ protected override IEnumerable<ServiceInstanceListener> CreateServiceInstanceLis
...
 
-                    var config = new ClusterConfiguration();
-                    config.Globals.RegisterBootstrapProvider<BootstrapProvider>("poke_grains");
-                    config.Globals.ReminderServiceType = GlobalConfiguration.ReminderServiceProviderType.ReminderTableGrain;
+                    builder.AddStartupTask<StartupTask>();
+                    builder.UseInMemoryReminderService();
 
                     // Service Fabric manages port allocations, so update the configuration using those ports.
-                    config.Defaults.ConfigureServiceFabricSiloEndpoints(serviceContext);
+                    builder.Configure<EndpointOptions>(options =>
```
Vhab: Considering `builder.ConfigureEndpoints` is used, using `builder.Configure<EndpointOptions>` might be unnecessary?

ReubenBond: Yes, you're right, thank you

PR #4235, opened 2018-03-16
Make providers use Cluster ID and Service ID from ClusterOptions

sergeybykov:
Make providers use Cluster ID and Service ID from ClusterOptions.
Also fix Azure Queue, SQS, and GCP pubsub providers to use Service ID instead of Cluster ID.

### DIFF
--- a/test/GoogleUtils.Tests/Streaming/PubSubStreamTests.cs
+++ b/test/GoogleUtils.Tests/Streaming/PubSubStreamTests.cs
@@ -44,7 +44,6 @@ namespace GoogleUtils.Tests.Streaming
                     {
                         options.ProjectId = GoogleTestUtils.ProjectId;
                         options.TopicId = GoogleTestUtils.TopicId;
-                        options.ClusterId = GoogleTestUtils.DeploymentId.ToString();
                         options.Deadline = TimeSpan.FromSeconds(600);
                     }));
             }
@@ -61,7 +60,6 @@ namespace GoogleUtils.Tests.Streaming
                     {
                         options.ProjectId = GoogleTestUtils.ProjectId;
                         options.TopicId = GoogleTestUtils.TopicId;
-                        options.ClusterId = GoogleTestUtils.DeploymentId.ToString();
                         options.Deadline = TimeSpan.FromSeconds(600);
                     }));
             }
--- a/test/TesterAzureUtils/Persistence/PersistenceGrainTests_AzureTableGrainStorage.cs
+++ b/test/TesterAzureUtils/Persistence/PersistenceGrainTests_AzureTableGrainStorage.cs
@@ -72,7 +72,8 @@ namespace Tester.AzureUtils.Persistence
             }
         }
 
-        public PersistenceGrainTests_AzureTableGrainStorage(ITestOutputHelper output, Fixture fixture) : base(output, fixture, ServiceId)
+        public PersistenceGrainTests_AzureTableGrainStorage(ITestOutputHelper output, Fixture fixture) : 
+            base(output, fixture)
         {
             fixture.EnsurePreconditionsMet();
         }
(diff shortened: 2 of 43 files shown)


sergeybykov:
A bunch of functional tests failed due to `ClusterOptions` missing in DI. Need to fix them before this can be merged.

ReubenBond:
I can do that if you'd like, @sergeybykov

sergeybykov:
I won't mind. 😀

ReubenBond:
@dotnet-bot test functional

> An attempt was made to access a socket in a way forbidden by its access permissions

test/GoogleUtils.Tests/Streaming/PubSubStreamTests.cs
```diff
@@ -61,7 +60,6 @@ public void Configure(IConfiguration configuration, IClientBuilder clientBuilder
                     {
                         options.ProjectId = GoogleTestUtils.ProjectId;
                         options.TopicId = GoogleTestUtils.TopicId;
-                        options.ClusterId = GoogleTestUtils.DeploymentId.ToString();
```
jason-bragg: Does GoogleTestUtils still have a DeploymentId?

test/TesterAzureUtils/Streaming/HaloStreamSubscribeTests.cs
```diff
@@ -42,15 +42,13 @@ public void Configure(ISiloHostBuilder hostBuilder)
                         .AddMemoryGrainStorage("MemoryStore", options => options.NumStorageGrains = 1)
                         .AddAzureTableGrainStorage("AzureStore", builder => builder.Configure<IOptions<ClusterOptions>>((options, silo) =>
```
jason-bragg: if we're not reading from ClusterOptions we don't need to use the builder in all these cases. 
call should be something like:

    .AddAzureTableGrainStorage("AzureStore", options => 
    { 
        options.ConnectionString = TestDefaultConfiguration.DataConnectionString; 
         options.DeleteStateOnClear = true; 
    })

This applies to all of the AddAzureTableGrainStorage and AddAzureTableGrainStorageAsDefault calls which use the options builder to get information from the ClusterOptions

jason-bragg:
This looks to address issues in the providers, but the work to move from DeploymentId to ClusterId was , as near as I can tell, never completed.  Deployment id is still peppered through the system; silo host, consul, zookeeper, ado, ..
If this PR isn't intended to address that, we should open in issue to track the remaining work.

ReubenBond:
> the work to move from DeploymentId to ClusterId was , as near as I can tell, never completed.

Note that we had to leave the property as `DeploymentId` in some cases since the property name is also the name of the table column, etc.

test/TesterAzureUtils/Persistence/PersistenceGrainTests_AzureTableGrainStorage.cs
```diff
@@ -29,8 +28,6 @@ protected override void ConfigureTestCluster(TestClusterBuilder builder)
                 {
                     legacy.ClusterConfiguration.Globals.DataConnectionString = TestDefaultConfiguration.DataConnectionString;
 
-                    legacy.ClusterConfiguration.Globals.ServiceId = ServiceId;
```
benjaminpetit: Why id you remove this?

sergeybykov: Technically, the tests work without it - with the default empty GUID `ServiceId`. I pushed a couple of commits to set it to a new GUID to avoid that.

ReubenBond:
I'm happy to merge this if there are no outstanding concerns

sergeybykov:
`SendAndReceiveFromAzureQueue` is the only test left that fails periodically due to an empty GUID `ServiceId`.

ReubenBond:
I thought I had committed a fix for that - might be on another branch. I'll submit a fix if you'd like.

ReubenBond:
@dotnet-bot test functional

PR #4240, opened 2018-03-16
Add MRI community company blurb

DarkCow:
MRI has a deployment of around 17vm's (4-8 core, 4-16GB of ram) and 3 instances of IIS.

### DIFF
--- a/src/Community/Who-Is-Using-Orleans.md
+++ b/src/Community/Who-Is-Using-Orleans.md
@@ -48,6 +48,9 @@ Argentinian-based electronic brokerage firm
 Nomnio IoT Platform and industry projects where reliability and performance is required
 * <a href="https://www.realartists.com/index.html" target="_blank">__Real Artists Ship 2.0__</a>
 Fast, native, comprehensive issue tracking for GitHub 
+* <a href="https://mri-inc.net/" target="_blank">__Manufacturing Resources International__</a>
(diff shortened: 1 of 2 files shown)


jdom:
Great! Thanks for adding it to the list, your domain looks really exciting, and a place where Orleans could indeed help a lot :)
But can you modify the markdown file instead? The HTML gets overwritten from the source markdown. This is the correct file to modify: https://github.com/dotnet/orleans/blob/gh-pages/src/Community/Who-Is-Using-Orleans.md

DarkCow:
@jdom fixed, let me know if you want me squash

src/Community/Who-Is-Using-Orleans.md
```diff
@@ -48,6 +48,9 @@ Argentinian-based electronic brokerage firm
 Nomnio IoT Platform and industry projects where reliability and performance is required
 * <a href="https://www.realartists.com/index.html" target="_blank">__Real Artists Ship 2.0__</a>
 Fast, native, comprehensive issue tracking for GitHub 
+* <a href="https://mri-inc.net/" target="_blank">__Manufacturing Resources International__</a>
```
jdom: odd, I wonder why these are using html tags and not just markdown for the links. Just a comment for @sergeybykov . Nothing for you to do @DarkCow as this is consistent with the rest.

jdom:
Thanks @DarkCow! BTW, no need to squash, I can do it when "merging" the PR.

PR #4242, opened 2018-03-16
Dedup option formatters and add more logging to streaming

xiazen:
- Dedup option formatters if ConfigureNamedOptions called twice or ConfigureFormatter called twice.
- Add more logging to streaming.

### DIFF
--- a/src/Orleans.Core/Configuration/OptionLogger/IOptionFormatter.cs
+++ b/src/Orleans.Core/Configuration/OptionLogger/IOptionFormatter.cs
@@ -32,6 +32,28 @@ namespace Orleans
         IOptionFormatter<T> Resolve(string name);
     }
 
+    public class OptionsFormatterComparer : IEqualityComparer<IOptionFormatter>
+    {
+        public static OptionsFormatterComparer Instance  = new OptionsFormatterComparer();
+        public bool Equals(IOptionFormatter x, IOptionFormatter y)
+        {
+            //if name equal and formatted strings are equal, then equal.
+            //perf wise, most formatters will fail at the first condition, which is relately cheap. 
+            //So don't switch order of the two condition
+            if (x.Name.Equals(y.Name)&& x.Format().IEnumerableEquals(y.Format()))
+            {
+                return true;
+            }
+            return false;
+
+        }
+
+        public int GetHashCode(IOptionFormatter obj)
+        {
+            var hashCode = $"{obj.Name}{obj.Format()}";
+            return hashCode.GetHashCode();
+        }
+    }
     /// <summary>
     /// Utility class for option formatting
     /// </summary>
--- a/src/Orleans.Core/Configuration/OptionLogger/IOptionsLogger.cs
+++ b/src/Orleans.Core/Configuration/OptionLogger/IOptionsLogger.cs
@@ -28,7 +28,7 @@ namespace Orleans
             return Task.CompletedTask;
         }
     }
-
+    
     /// <summary>
     /// Base class for client and silo default options loggers.
     /// </summary>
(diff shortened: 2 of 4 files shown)


src/Orleans.Core/Configuration/OptionLogger/IOptionFormatter.cs
```diff
@@ -32,6 +32,28 @@ public interface IOptionFormatterResolver<T>
         IOptionFormatter<T> Resolve(string name);
     }
 
+    public class OptionsFormatterComparer : IEqualityComparer<IOptionFormatter>
```
jason-bragg: Worried this will mask junk in our system.  A duplicate formatter indicates a bug right?

xiazen: current option formatter behavior doesn't guarantee just one formatter per option. If a user call ConfigureNamedOptions twice, it will print two identical ones. AFAIK, this is accepted by three people touched this piece of code, hence not a bug?

ReubenBond: In other areas we use `Properties` on either `ISiloHostBuilder` or `HostBuilderContext`.
Eg here: https://github.com/dotnet/orleans/blob/51fea8aedbec0ef7a33ce9a76692cf392ff507b1/src/Orleans.Runtime/Hosting/CoreHostingExtensions.cs#L25
or here: https://github.com/dotnet/orleans/blob/51fea8aedbec0ef7a33ce9a76692cf392ff507b1/src/Orleans.TestingHost.Legacy/LegacySiloBuilderConfigurator.cs#L27

ReubenBond: If you're worried that `HostBuilderContext` wont be accessible because you've only got `IServiceCollection` at the point where this is used, we can modify SiloHostBuilder to ensure that `HostBuilderContext`, etc, are always added to `IServiceCollection` before any of the other `ConfigureServices` delegates are run.

xiazen: I'm not sure I follow. This dedup is to remove duplicated OptionsFormatter if user configure the same OptionFormatter twice. Can you explain more on why is this related to the `Properties`?

jason-bragg: > Current option formatter behavior doesn't guarantee just one formatter per option.

Understood.  The system does not inforce this.  If we can figure out how to enforce this, I'd be for doing so.  If not, my preference would be to leave the misconfiguration observable rather than masking it.

I suspect we may be able to enforce the desired behavior using what Reuben suggested, but i've not looked into it.

ReubenBond: @xiazen, Jason was saying above "Worried this will mask junk in our system" because the PR merely filters duplicates after the fact rather than preventing them from being added in the first place.

What I'm suggesting is to prevent duplicates rather than filter them out when you're reading the formatters.

Ultimately it doesn't affect too much, it's just a matter of cleanliness.

xiazen: Oh I see. I can do a pass using `Properties`. I introduced here in this PR mainly because I run into the duplication problem in streaming. If I add ConfigureNamedOptions in IPersistentStreamProviderConfigurator.Configure<TOptions>, then it is really hard for user to figure out when to call ConfigureNamedOptions and when not, and the whole ConfigureNamedOptions and ConfigureFormatter is kind of fragile. 

But thanks @ReubenBond  I will take a pass on that.

xiazen:
close this since it is low priority now and I don't have a bandwidth to finish this in the near future

PR #4243, opened 2018-03-17
Improves ADO.NET script finding by moving them to project directory

### LINKED ISSUE #3864: Improve ADO.NET script finding

veikkoeeva:
Currently since they're split and located in Nuget cache directory "far away"
from the project. Using nuget locals all -list, reveals the location is
something such as
C:\Users\<username>\.nuget\packages\microsoft.orleans.*.adonet\2.0.0-beta3\lib\netstandard2.0.

This improves the situation such that upon Nuget restore the files re copied to
project directory to \OrleansAdoNetContent where each supported ADO.NET provider
has its own directory.

Further, this makes it easier later to implement helper methods, such as automatic
database deployment, schema naming, filegroup additions and layout updates even with
large amounts of data.

Fixes #3864.

### DIFF
--- a/src/AdoNet/Orleans.Clustering.AdoNet/Microsoft.Orleans.Clustering.AdoNet.targets
+++ b/src/AdoNet/Orleans.Clustering.AdoNet/Microsoft.Orleans.Clustering.AdoNet.targets
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="MicrosoftOrleansClusteringAdoNetCopySqlFiles" BeforeTargets="PrepareForBuild">
+    <!-- These are the source content files needed in the project. Note that $(TargetFramework) cannot be used here since it can be netcoreapp2.0 or somesuch. -->
+    <ItemGroup>
+      <MicrosoftOrleansClusteringAdoNetItems Include="$(MSBuildThisFileDirectory)..\..\lib\netstandard2.0\**\*.sql" />
+    </ItemGroup>
+
+    <!-- The user may have modified the *.sql files fit for deployment. So they shouldn't be overwritten when Nuget restore is made. -->
+    <Copy
+      SourceFiles="@(MicrosoftOrleansClusteringAdoNetItems)"
+      DestinationFolder="$(MSBuildProjectDirectory)/OrleansAdoNetContent/%(RecursiveDir)"
+      Condition="!(Exists(@(Packages->'$(TargetDir)%(Filename)%(Extension)')))"
+      OverwriteReadOnlyFiles="false"
+      Retries="2"
+      RetryDelayMilliseconds="400"/>
+  </Target>
+</Project>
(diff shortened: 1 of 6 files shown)


veikkoeeva:
When all the ADO.NET libraries are added, the result looks like this in a project folder:
all ADO.NET nuget packages added

src/AdoNet/Orleans.Clustering.AdoNet/Microsoft.Orleans.Clustering.AdoNet.targets
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="MicrosoftOrleansClusteringAdoNetCopySqlFiles" BeforeTargets="PrepareForBuild">
```
veikkoeeva: I don't know if there is a `AfterTargets` for Nuget restore, which might be better than this. On the other than, this restores files automatically when potentially needed.

veikkoeeva:
@jdom, @xiazen, @sergeybykov, @ReubenBond Would this be OK and if so, make it to the 2.0 release? Apologies for tagging everyone, just trying to get a handle of someone. 

Also, @galvesribeiro, this might be something if you're also using files in CosmosDB for potential user customization.

galvesribeiro:
Makes total sense. Thanks for ping me @veikkoeeva.

Lets see what others think. I'll bring it in on CosmosDB providers later on.

Thanks

src/AdoNet/Orleans.Clustering.AdoNet/Microsoft.Orleans.Clustering.AdoNet.targets
```diff
@@ -0,0 +1,18 @@
...
+    <!-- These are the source content files needed in the project. Note that $(TargetFramework) cannot be used here since it can be netcoreapp2.0 or somesuch. -->
+    <ItemGroup>
+      <MicrosoftOrleansClusteringAdoNetItems Include="$(MSBuildThisFileDirectory)..\..\lib\netstandard2.0\**\*.sql" />
+    </ItemGroup>
+
+    <!-- The user may have modified the *.sql files fit for deployment. So they shouldn't be overwritten when Nuget restore is made. -->
+    <Copy
```
veikkoeeva: These files are only copied to the project directory, but not added to the project content. The purpose is to make them more findable after Nuget restore. It's not clear what should be done after the files have been brought to the project:

1. The user may want to use some tool to create the database explicitly.
2. The user may want to copy the files to output and distribute with the silo host so the files are available if the database is deployed (or checked if it's deployed) upon start-up.

Including the files as content and marking them as build output is supposedly harmless, albeit might distribute files needlessly. The user may forget/doesn't think to distribute the files if they aren't added automatically and have a slightly worse deployment experience (this is an issue supposedly when one is "kicking the tyres", since running a production deployment one needs to do some QA anyhow). This can be changed also when there's more configuration surface available in the .NET libraries.

sergeybykov:
Thank you, @veikkoeeva!

PR #4245, opened 2018-03-18
Improve startup time for localhost cluster.

SebastianStehle:
The startup time is heavily increased by the delay in the membership oracle. For development we can set the clustersize to 1 and also do not wait. Reduced my startup time from 15 to 3 seconds.


sergeybykov:
Thank you, @SebastianStehle!

Issue #4241, opened 2018-03-16
Parameter binding problem on ODP .NET Managed Core


fduman:
I have struggled silo cluster and storage configuration on Oracle DB. I'm using Orleans 2.0.0 rc2 on .NET Core also using ODP .NET Managed Core 2.12.0-beta2.

My regional settings are set to Turkish on Windows.

I configured my host silo using ADO with ODP .Net Managed. I prepared my database using the Oracle DB scripts in your repository. I came across parameter binding exceptions on my silo. I realized that all exceptions point a parameter which contains "i" or "I" character. So I corrected the SQL statements in ORLEANSQUERY tables like:
:DEPLOYMENTID => :DeploymentId
:IAMALIVETIME => :IAmAliveTime
etc.. 

Then I was able to start my silo with cluster configuration.

If you look at https://msdn.microsoft.com/en-us/library/xk2wykcz(VS.71).aspx, you will see the Turkish "i" character casing problem in detail. This is a common problem in developer world which bothers Turkish users many times.

I think it is a bug of ODP .NET Managed Core but if you agree to change the statements like I changed, I will be appreciated.

veikkoeeva:
@fduman Could you open a PR? If what you did fixed the problem, it doesn't look like there's a reason to not to change the literals. As an added benefit, should be more in line too with the others also.

sergeybykov:
Resolved via #4246.

PR #4246, opened 2018-03-19
Change Oracle SQL parameter casing for Turkish language binding support on ODP.NET Managed

fduman:
for resolving issue #4241


sergeybykov:
@veikkoeeva Do you approve?

veikkoeeva:
@sergeybykov Pondering: Looks like nothing else has changed but casing. Though thinking now, and looking at the casing, the `I` letters are still capitalized. Should the same problem persist in that case? I think should, but... If the problem is gone, the change is benign and though less idiomatic Oracle naming, more on line with other scripts.

With these I don't see there is a problem accepting. Caveat emptor, I don't have Oracle DB installed to test.

sergeybykov:
Thank you, @fduman and @veikkoeeva!

Issue #1069, opened 2015-11-25
Orleans clients using PersistentStreamProviders must have configured gateways for every silo in the cluster.


jason-bragg:
Orleans clients communicate with the cluster through configured gateways.
http://dotnet.github.io/orleans/Orleans-Configuration-Guide/Client-Configuration.
If a grain activation is on a silo that is not on one of the configured gateways, Orleans will route communication with that grain through one of the configured gateways.  This is possible because each gateway can obtain the address of the silo responsible for any activation.

System targets, however, are not routed, as their address cannot be obtained.  Any call to a system target must be made directly to the silo responsible for that system target's activation.  This means that an Orleans client that needs to send a message to a system target must have the silo responsible for the system target's activation in it's list of configured gateways.

Since the Persistent Stream Providers used system targets to read from the queue, for Orleans Clients to use Persistent Stream Providers they must have gateway connections to every silo in the cluster.

Orleans clients should require only a single gateway.  Since this breaks that expectation, this is a bug.

gabikliot:
Hmmm, interesting.
It sounds like there is indeed a bug around that area. But may I propose a different fix from what you suggested.
You suggested as a fix that a client has to be connected to every GW to be able to send request to any system target. I think we can achieve that differently/better.

When client sends a msg to a system target, it has the system target address, including silo address (system target address in the streaming case comes from the producer Grain Reference returned by the pub sub). So what you wrote: "System targets, however, are not routed, as their address cannot be obtained. " **is not correct**. GW doesn't need to use the directory to route to system target, obviously, but it has the system target address in the msg header. So it can route the msg to the correct silo.

The bug is that most probably the system target silo address is not copied somewhere/gets lost. We just need to make sure it is not lost.

Regardless of how we fix it, requiring the client to be connected to all GWs is wrong - the system was designed, specifically, to allow the client to be connected to a subset, even one, GW and everything should work correctly.

If you agree with all said above, I can help search through the code t find out where we are loosing that system target address.

gabikliot:
OK, so the problem is on that line:
https://github.com/dotnet/orleans/blob/master/src/Orleans/Messaging/ProxiedMessageCenter.cs#L178

A msg to a system target where TargetSilo is specified can go only to this GW. Otherwise, we just drop it. The bug is that we just drop it.
The fix I recommend is:
[...]

PR #4254, opened 2018-03-20
Enable SystemTarget routing in the Gateway

benjaminpetit:
Fix for #1069 

Currently, gateways override the value of `Message.SendingSilo` by their own `GatewayAddress` when sending message to clients, so they can always trust this value when responding to messages.

This is an issue when clients need to send a response to a SystemTarget activation when this activation is not on a silo that has a gateway (or when the client is not connected to that gateway).

This is a workaround, and more work should be done on client messaging routing in the future, but the idea is:

- Gateways will not override `Message.SendingSilo` value if they are proxying a message from a SystemTarget activated on another silo
- Client will no longer trust blindly the value received from `Message.SendingSilo`, so they will check that this is a valid gateway address. If this is not a valid gateway address, they will send it to any connected gateway
- Gateway will do some routing for SystemTarget since the value from `Message.TargetSilo` can be a valid Silo-to-Silo address

### DIFF
--- a/src/Orleans.Runtime/Messaging/Gateway.cs
+++ b/src/Orleans.Runtime/Messaging/Gateway.cs
@@ -180,7 +180,7 @@ namespace Orleans.Runtime.Messaging
             // it to this address...
             // EXCEPT if the value is equal to the current GatewayAdress: in this case we will return
             // null and the local dispatcher will forward the Message to a local SystemTarget activation
-            if (msg.TargetGrain.IsSystemTarget && !msg.TargetSilo.Matches(this.gatewayAddress))
+            if (msg.TargetGrain.IsSystemTarget && !this.gatewayAddress.Matches(msg.TargetSilo))
                 return msg.TargetSilo;
 
             // for responses from ClientAddressableObject to ClientGrain try to use clientsReplyRoutingCache for sending replies directly back.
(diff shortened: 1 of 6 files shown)


src/Orleans.Runtime/Messaging/Gateway.cs
```diff
@@ -170,6 +170,19 @@ internal void RecordClosedSocket(Socket sock)
...
+            // When a client make a request/response to/from a SystemTarget, the TargetSilo can be set to either 
+            //  - the GatewayAddress of the target silo (for example, when the client want get the cluster typemap)
+            //  - the "internal" Silo-to-Silo address, if the client want to send a message to a specific SystemTarget
+            //    activation that is on a silo on which there is no gateway available (or if the client is not
+            //    connected to that gateway)
+            // So, if the TargetGrain is a SystemTarget we always trust the value from Message.TargetSilo and forward
+            // it to this address...
+            // EXCEPT if the value is equal to the current GatewayAdress: in this case we will return
+            // null and the local dispatcher will forward the Message to a local SystemTarget activation
+            if (msg.TargetGrain.IsSystemTarget && !msg.TargetSilo.Matches(this.gatewayAddress))
```
ReubenBond: Can either of these values ever be null here?

benjaminpetit: It doesn't make sense to have a message without a `TargetGrain`. `TargetSilo` is always set on the client, but I think it is cheap enough to add a null check here to be more defensive, I will add it

Issue #4096, opened 2018-02-28
TypeMap are not dynamically refreshed on clients


benjaminpetit:
Related to #4028 with in-place upgrades.

The issue is that if you try to update both silo and client at the same time, new clients are more likely to start before the new silos, so they get an outdated `TypeMap` (clients get it when they connect to their first silo).

Right now the workaround is to create a new `ClusterClient` that will pick-up a new `TypeMap`, when getting an exception saying `Cannot find an implementation class for grain `

Here are our options:

1. Keep the current behavior (no `TypeMap` refresh on clients). Developer will have to handle the case explicitely.
2. Refresh periodically the `TypeMap`:
a) With a timer on the client
b) Silo will send a push notification to the client telling him to update the `TypeMap`when needed (but since clients are connected to multiple silos, but only one of them should send such notifications?)

benjaminpetit:
Fixed with #4257

Issue #4028, opened 2018-02-13
New grain Implementation is not loaded after its first upgrade


yzhao174:
We are running Orleans 2.0 on ServiceFabric cluster. When we try to add a new interface/implementation of a grain or change the name of existing interface/implementation, it will fail with error message:

> {"error":{"innererror":"System.ArgumentException: Cannot find an implementation class for grain interface: Fog.GrainInterfaces.IFogFailureManagementGrain. Make sure the grain assembly was correctly deployed and loaded in the silo.
  at Orleans.GrainFactory.GetGrainClassData(Type interfaceType, String grainClassNamePrefix)
  at Orleans.GrainFactory.GetGrain[TGrainInterface](Guid primaryKey, String grainClassNamePrefix)
  at Orleans.ClusterClient.GetGrain[TGrainInterface](Guid primaryKey, String grainClassNamePrefix)
  at Fog.GrainInterfaces.GrainFactoryExtension.GetFogFailureManagementGrainReference(IGrainFactory grainFactory, Guid key) in C:\\BA\\185\\s\\src\\ControlPlane\\Fog.GrainInterfaces\\GrainFactoryExtension.cs:line 37
  at Fog.Frontend.Controllers.BatchController.ThrowIfFailureOn(AzureRegion regionId, String accountName, String poolId, String methodName) in C:\\BA\\185\\s\\src\\ControlPlane\\Fog.Frontend\\Controllers\\BatchController.cs:line 49
  at Fog.Frontend.Controllers.BatchController.<CreatePool>d__6.MoveNext() in C:\\BA\\185\\s\\src\\ControlPlane\\Fog.Frontend\\Controllers\\BatchController.cs:line 186\r\n--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
  at System.Threading.Tasks.TaskHelpersExtensions.<CastToObject>d__3`1.MoveNext()\r\n--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #4257, opened 2018-03-20
Client typemap refresh

benjaminpetit:
Fix for #4096 and #4028 

Right now, clients get the grain type map only once when they first connect to the cluster. If there is any change in this map during the lifetime of the client, it will not be propagated.

With this fix, client will get an updated typemap every `TypeManagementOptions.TypeMapRefreshInterval` (by default 1 minute). It should help developer who are using in-place updgrade that cannot ensure that new clients do not start before new silos.

### DIFF
--- a/src/Orleans.Core/Logging/ErrorCodes.cs
+++ b/src/Orleans.Core/Logging/ErrorCodes.cs
@@ -1024,6 +1024,7 @@ namespace Orleans
 
         TypeManagerBase = Runtime + 4200,
         TypeManager_GetSiloGrainInterfaceMapError = TypeManagerBase + 1,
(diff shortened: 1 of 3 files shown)


ReubenBond:
@benjaminpetit does this mean that the documentation is out of date?

A question was raised on Gitter about this today: https://gitter.im/dotnet/orleans?at=5bed97b562866f7473a322bc

Docs say that this is a limitation of heterogeneous clusters: https://dotnet.github.io/orleans/Documentation/clusters_and_clients/heterogeneous_silos.html

> Connected clients will not be notified if the set of supported Grain Types changed.

benjaminpetit:
That's correct, we should update the doc

PR #4264, opened 2018-03-21
Remove duplicate references to AdoNet.targets

sergeybykov:
This is a follow-up to #4243.

There are currently build warnings while building `Orleans.Persistence.AdoNet/Orleans.Persistence.AdoNet.csproj` and `Orleans.Reminders.AdoNet/Orleans.Reminders.AdoNet.csproj` complaining about duplicate references to `Microsoft.Orleans.Persistence.AdoNet.targets`. This change seems to make build happy, and the content files still get copied to the project folder.

@veikkoeeva Could you double-check please.


veikkoeeva:
The `.target` files seem to be still there, looking at the diff, and one time only this time. Also @sergeybykov mentions files are copied. The right thing to do, hopefully all good now. 👍

PR #4267, opened 2018-03-21
Lease based queue balancer fixes

jason-bragg:
- Timers were not on orleans thread so couldn't make grain calls.
- Aqure logic could be too greedy
- logging and spelling cleanup (can you believe -I- fixed spelling?)
- prevented possible devide by 0 issues.

### DIFF
--- a/src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
+++ b/src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
@@ -115,8 +115,6 @@ namespace Orleans.Streams
         public static IStreamQueueBalancer Create(IServiceProvider services, string name, IDeploymentConfiguration deploymentConfiguration)
         {
             var options = services.GetRequiredService<IOptionsSnapshot<LeaseBasedQueueBalancerOptions>>().Get(name);
-            if (options == null)
-                throw new KeyNotFoundException($"No lease base queue balancer options was configured for provider {name}, nor was a default configured.");
             return ActivatorUtilities.CreateInstance<LeaseBasedQueueBalancer>(services, name, options, deploymentConfiguration);
         }
         /// <inheritdoc/>
@@ -259,7 +257,7 @@ namespace Orleans.Streams
                 activeBuckets = GetActiveSiloCount(this.siloStatusOracle);
             }
             activeBuckets = Math.Max(1, activeBuckets);
-            this.minimumResponsibility = Math.Max(1, this.allQueues.Count / activeBuckets);
+            this.minimumResponsibility = this.allQueues.Count / activeBuckets;
             //if allQueues count is divisible by active bukets, then every bucket should take the same count of queues, otherwise, there should be one bucket take 1 more queue
             if (this.allQueues.Count % activeBuckets == 0)
                 this.maximumResponsibility = this.minimumResponsibility;
--- a/src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
+++ b/src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
@@ -8,8 +8,8 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Orleans.LeaseProviders;
 using Orleans.Runtime;
-using Orleans.Hosting;
 using Orleans.Configuration;
+using Orleans.Timers;
 
 namespace Orleans.Streams
 {
@@ -51,9 +51,10 @@ namespace Orleans.Streams
         public List<T> NextSelection(int newSelectionCount, List<T> existingSelection)
         {
             var selection = new List<T>(newSelectionCount);
-            while (selection.Count < newSelectionCount)
+            int tries = 0;
+            while (selection.Count < newSelectionCount && tries++ < this.resources.Count)
             {
-                this.lastSelection = (++this.lastSelection) % (this.resources.Count - 1);
+                this.lastSelection = (++this.lastSelection) % (this.resources.Count);
                 if(!existingSelection.Contains(this.resources[this.lastSelection]))
                     selection.Add(this.resources[this.lastSelection]);
             }
@@ -87,11 +88,11 @@ namespace Orleans.Streams
         private ReadOnlyCollection<QueueId> allQueues;
         private List<AcquiredQueue> myQueues;
         private bool isStarting;
-        private AsyncTaskSafeTimer renewLeaseTimer;
-        private AsyncTaskSafeTimer tryAcquireMaximumLeaseTimer;
+        private IDisposable renewLeaseTimer;
+        private IDisposable tryAcquireMaximumLeaseTimer;
         private IResourceSelector<QueueId> queueSelector;
-        private int minimumResponsibilty;
-        private int maximumRespobsibility;
+        private int minimumResponsibility;
+        private int maximumResponsibility;
         private IServiceProvider serviceProvider;
         private ILogger logger;
         private ILoggerFactory loggerFactory;
@@ -99,12 +100,7 @@ namespace Orleans.Streams
         /// <summary>
         /// Constructor
         /// </summary>
-        /// <param name="options"></param>
-        /// <param name="serviceProvider"></param>
-        /// <param name="siloStatusOracle"></param>
-        /// <param name="deploymentConfig"></param>
-        /// <param name="loggerFactory"></param>
-        public LeaseBasedQueueBalancer(LeaseBasedQueueBalancerOptions options, IServiceProvider serviceProvider, ISiloStatusOracle siloStatusOracle, IDeploymentConfiguration deploymentConfig, ILoggerFactory loggerFactory)
+        public LeaseBasedQueueBalancer(string name, LeaseBasedQueueBalancerOptions options, IServiceProvider serviceProvider, ISiloStatusOracle siloStatusOracle, IDeploymentConfiguration deploymentConfig, ILoggerFactory loggerFactory)
         {
             this.serviceProvider = serviceProvider;
             this.deploymentConfig = deploymentConfig;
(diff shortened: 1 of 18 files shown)


src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -113,32 +109,33 @@ public LeaseBasedQueueBalancer(LeaseBasedQueueBalancerOptions options, IServiceP
...
-            this.leaseProvider = this.serviceProvider.GetRequiredService(options.LeaseProviderType) as ILeaseProvider;
             this.allQueues = new ReadOnlyCollection<QueueId>(queueMapper.GetAllQueues().ToList());
+            if (this.allQueues.Count == 0)
+                return Task.CompletedTask;
+            this.leaseProvider = this.serviceProvider.GetRequiredService(options.LeaseProviderType) as ILeaseProvider;
             NotifyAfterStart().Ignore();
             //make lease renew frequency to be every half of lease time, to avoid renew failing due to timing issues, race condition or clock difference. 
-            var timerLogger = this.loggerFactory.CreateLogger<AsyncTaskSafeTimer>();
-            this.renewLeaseTimer = new AsyncTaskSafeTimer(timerLogger, this.MaintainAndBalanceQueues, null, this.options.SiloMaturityPeriod, this.options.LeaseLength.Divide(2));
+            ITimerRegistry timerRegistry = this.serviceProvider.GetRequiredService<ITimerRegistry>();
```
jason-bragg: Is this a good way to create a timer callback that is scheduled on an Orleans thread?

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -174,41 +171,44 @@ public override IEnumerable<QueueId> GetMyQueues()
...
             int maxAttempts = 5;
             int attempts = 0;
             int leasesToAquire = expectedTotalLeaseCount - this.myQueues.Count;
-            this.logger.Info($"AcquireLeasesToMeetExpectation : Try to acquire {leasesToAquire} queues");
-            while (attempts ++ <= maxAttempts && leasesToAquire > 0)
+            if (leasesToAquire <= 0) return;
+            while (attempts ++ < maxAttempts && leasesToAquire > 0)
             {
+                this.logger.LogDebug($"I have {this.myQueues.Count} queues.  Trying to acquire {leasesToAquire} queues to reach {expectedTotalLeaseCount}");
+                leasesToAquire = expectedTotalLeaseCount - this.myQueues.Count;
```
jason-bragg: Need to recalculate leases to acquire every loop, or we could over acquire.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
ReubenBond: Only if you're using it from a grain, but `null` is being passed, which will cause an NRE

ReubenBond: Oh, I see, TimerRegistry is being modified

jason-bragg: Yes.  The behavior is that if it's called from an Orleans thread, the callback will be scheduled on an Orleans thread.  If not it won't be.  The grain reference is only necessary for cleanup.  I'm not 100% confident in this change because the api surface doesn't make it clear that a null grain is valid, so other implementations may not account for that.

If we agree that this is a valid behavior (debatable) we may want to modify the interface to make this more clear.

If this behavior is unacceptable, then we need to figure out how to schedule timers on Orleans threads from Orleans threads.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -113,32 +109,33 @@ public LeaseBasedQueueBalancer(LeaseBasedQueueBalancerOptions options, IServiceP
...
-            this.logger = loggerFactory.CreateLogger<LeaseBasedQueueBalancer>();
+            this.logger = loggerFactory.CreateLogger($"{typeof(LeaseBasedQueueBalancer).FullName}-{name}");
         }
 
         public static IStreamQueueBalancer Create(IServiceProvider services, string name, IDeploymentConfiguration deploymentConfiguration)
         {
-            var options = services.GetService<IOptionsSnapshot<LeaseBasedQueueBalancerOptions>>().Get(name);
-            return ActivatorUtilities.CreateInstance<LeaseBasedQueueBalancer>(services, options, deploymentConfiguration);
+            var options = services.GetRequiredService<IOptionsSnapshot<LeaseBasedQueueBalancerOptions>>().Get(name);
+            if (options == null)
```
xiazen: if there's no `LeaseBasedQueueBalancerOptions` with this name registered. would IOptionSnapshot return a default one or null? if return default one, then this null check isn't effective, right?

jason-bragg: yup.. I just moved this logic from Initialize.. didn't really look too hard at it.. bleh..
Will fix.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -174,41 +171,44 @@ public override IEnumerable<QueueId> GetMyQueues()
...
                 //select new queues to acquire
-                var expectedQueues = this.queueSelector.NextSelection(leasesToAquire, this.myQueues.Select(queue=>queue.QueueId).ToList()).ToList();
-                var leaseRequests = expectedQueues.Select(queue => new LeaseRequest() {
+                List<QueueId> expectedQueues = this.queueSelector.NextSelection(leasesToAquire, this.myQueues.Select(queue=>queue.QueueId).ToList()).ToList();
+                IEnumerable<LeaseRequest> leaseRequests = expectedQueues.Select(queue => new LeaseRequest() {
                     ResourceKey = queue.ToString(),
                     Duration = this.options.LeaseLength
                 });
-                var results = await this.leaseProvider.Acquire(LeaseCategory, leaseRequests.ToArray());
+                AcquireLeaseResult[] results = await this.leaseProvider.Acquire(LeaseCategory, leaseRequests.ToArray());
```
xiazen: why are we so explicit about the type. I thought using var is the suggested way

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -174,41 +171,44 @@ public override IEnumerable<QueueId> GetMyQueues()
...
         }
 
         private async Task AcquireLeasesToMeetExpectation(int expectedTotalLeaseCount)
         {
             int maxAttempts = 5;
             int attempts = 0;
             int leasesToAquire = expectedTotalLeaseCount - this.myQueues.Count;
-            this.logger.Info($"AcquireLeasesToMeetExpectation : Try to acquire {leasesToAquire} queues");
-            while (attempts ++ <= maxAttempts && leasesToAquire > 0)
+            if (leasesToAquire <= 0) return;
```
xiazen: you don't need this line, the while loop already check this

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: Alternatively, it could be argued that it's not the responsibility of the LeaseBasedQueueBalancer to schedule these in Orleans threads, instead making it the responsibility of the lease provider to solve this if it's implementation uses grains.  I solved it this way mainly because this logic _was_ called in an Orleans thread, so making grain calls should have been safe, but was not because of the way the timers were setup.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -256,11 +258,12 @@ private void CalculateResponsibility()
             {
                 activeBuckets = GetActiveSiloCount(this.siloStatusOracle);
             }
-            this.minimumResponsibilty = this.allQueues.Count / activeBuckets;
+            activeBuckets = Math.Max(1, activeBuckets);
+            this.minimumResponsibility = Math.Max(1, this.allQueues.Count / activeBuckets);
```
xiazen: I think we should remove this Math.Max here. If this.allQueues.Count/activeBUckets <1, then minResponsibility should be 0, but we still have maxResponsibility to be 1. so each silo will try to get 1 if they can. 
if min is 1, then max is 2, then each silo will do their best to get that 1, while they can really just settle for 0. It is just a matter of aggressiveness. does this make sense?

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: It's a trade off, maintainability vs readability.  Historically, the convention on the team has been to use explicit type if it's not obvious from the call.  This is a carryover from 343 coding conventions.
So something like:

    var foo = services.GetService<IFoo>()

is fine because it's obvious what the type is, but:

    var thingys = foo.GetAllThingys()

is less desirable, because it's not clear to a reviewer what thingy is.  We've never strongly enforced this, and its not clear if this convention should carry forward with the new team.  I still follow the convention mostly out of habit.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: This prevents log spam when nothing is changing.  Otherwise the log line at the end of this function is fired when nothing is done.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: yup

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -51,9 +51,10 @@ public RoundRobinSelector(ReadOnlyCollection<T> resources)
         public List<T> NextSelection(int newSelectionCount, List<T> existingSelection)
         {
             var selection = new List<T>(newSelectionCount);
-            while (selection.Count < newSelectionCount)
+            int tries = 0;
+            while (selection.Count < newSelectionCount && tries++ < this.resources.Count)
             {
-                this.lastSelection = (++this.lastSelection) % (this.resources.Count - 1);
+                this.lastSelection = (++this.lastSelection) % (this.resources.Count);
```
jason-bragg: Please sanity check this change.  I believe the previous logic would never allow for the last resource to be allocated.
It also allowed infinite loop if more resource were requested than were available.

xiazen: You are probably right. But can we add some tests testing this. I forgot why I added that minus 1, but I remember I do spent sometime on this. 
Add tests : 
- NextSelectionWillGoThroughEveryResourceIfExistingSelectionEmpty
- NextSelectionWontGoInfinitely
- NextSelectionWontReSelectExisitngSelections

jason-bragg:
Test failure was UnitTests.TimerTests.ReminderTests_TableGrain.Rem_Grain_1J_MultiGrainMultiReminders
Looks unrelated.
Will re-run functionals.  If passes, we should look at above test, as it's probably an intermittent failure we should fix.

jason-bragg:

PR #4271, opened 2018-03-21
Clean up EH checkpointing

jason-bragg:
- checkpoints new uses serviceid for partitioning
- table name has default

### DIFF
--- a/src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubCheckpointer.cs
+++ b/src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubCheckpointer.cs
@@ -5,24 +5,28 @@ using Orleans.Streams;
 using Orleans.Streaming.EventHubs;
 using Orleans.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 
 namespace Orleans.ServiceBus.Providers
 {
     public class EventHubCheckpointerFactory : IStreamQueueCheckpointerFactory
     {
-        private ILoggerFactory loggerFactory;
-        private string providerName;
-        private AzureTableStreamCheckpointerOptions options;
-        public EventHubCheckpointerFactory(string providerName, AzureTableStreamCheckpointerOptions options, ILoggerFactory loggerFactory)
+        private readonly ILoggerFactory loggerFactory;
+        private readonly string providerName;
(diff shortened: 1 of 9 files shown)


xiazen:
why are we doing this ? for consistency with storage provider using serviceId as partition or? 
I'm concerned about this because checkpointerNamespace wasn't related to serviceId in anyway before. Now we change it to be serviceId, first is a breaking change, second how do users migrate, third, why?

jason-bragg:
@dot test functional please

jason-bragg:

jason-bragg:
The CheckpointNamespace is the partitioning mechanism, which most providers either used serviceId or deployment for.  Namespace was a poor name, and it didn't default to anything previously.  This conforms the checkpointer to the decisions we've made for other persistent mechanisms, and has the same draw backs.  The workaround for this is to provide a custom check pointer that partitions the data as done previously.

Is this a breaking change - yes
How to migrate - custom checkpointer that partitions the old way
why - consistency with the rest of orleans.

xiazen:
> How to migrate - custom checkpointer that partitions the old way

That's a expensive way, which I'm not convinced that every users are willing to take, meaning there's certain risks involved. I suggest we make this Overridable just like other providers.

xiazen:

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubCheckpointer.cs
```diff
@@ -80,7 +85,7 @@ private EventHubCheckpointer(AzureTableStreamCheckpointerOptions options, string
             }
             persistInterval = options.PersistInterval;
             dataManager = new AzureTableDataManager<EventHubPartitionCheckpointEntity>(options.TableName, options.ConnectionString, loggerFactory);
```
xiazen: add a log line here to say: created a checkpointer with streamProviderName xxx, partition xxxx, serviceId xxx, so if users configured their serviceId wrong, or isn't aware of that serviceId is part of partition now, we can ask them to send us their logs and see what exactly is used as partition

xiazen:
minor comments, otherwise LGTM

sergeybykov:
Test failures.

jason-bragg:
Look unrelated, but being addressed in Lease based queue balancer fixes #4267 
Will rebase after that passes and is merged.

Issue #3484, opened 2017-10-02
Stream resubscription issues


yevhen:
Every time we redeploy silos clients cannot resubscribe to stream and so we're forced to redeploy them as well. We do resubscribe once we received ClusterConnectionLost event (we retry infinitely). The weird thing that client actually reconnects and can communicate with actors, it's only streams which are broken.

This is what we see in a logs:

```
System.TimeoutException: Response did not arrive on time in 00:00:10 for message: Request *cli/e1fa0799@64d32ea2->*grn/716E8E94/00000000+udp_theme.task.notifications #6691: global::Orleans.Streams.IPubSubRendezvousGrain:RegisterConsumer(). Target History is: <*grn/716E8E94/00000000+udp_theme.task.notifications:>.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Orleans.Streams.StreamConsumer`1.<SubscribeAsync>d__14.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Orleankka.StreamRef.<Subscribe>d__9.MoveNext() in C:\Work\OSS\Orleankka\Source\Orleankka\StreamRef.cs:line 70
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Youscan.Web.Api.HubClient.<DoSubscribe>d__11.MoveNext() in D:\TeamCity\buildAgent02\work\58c475d9a27a23a9\Source\Web.Api\HubClient.cs:line 143
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #4272, opened 2018-03-21
Added test to exersize queue rebalancing while actively streaming

jason-bragg:
This tests is a follow-up on stream rebalancing issues raised by @ilyalukyanov in Stream resubscription issues #3484.
It will not compile or pass tests until Lease based queue balancer fixes #4267 is merged.


jason-bragg:
replaced with #4294.

Issue #4196, opened 2018-03-12
Silo should validate if a PubSubStore is configured or not before starting


attilah:
Issue:
When you setup the streaming, but not adding a storage with the baked in name of "PubSubStore", the Silo starts without errors and then later, when it needs to access the store to persist a subscription it fails during runtime.

Solution:
Add additional validation if a "PubSubStore" is registered in DI or not. If not, fail early with configuration exception.

Nice to have:
Make "PubSubStore" provider name configurable

galvesribeiro:
> Make "PubSubStore" provider name configurable

Already spoke with @jason-bragg about that. Unfortunatelly it is not simple right now. Probably when the revamp happens on Stream APIs we get it outsourced.

xiazen:
> Probably when the revamp happens on Stream APIs we get it outsourced.

not sure what triggered this conclusion. the name "PubSubStore" is always hard coded to be the pubsubgrain state store name. This doesn't change after streaming api revamp. How easy to make it configurable doesn't have much to do with streaming, it is how to make a grain state storage name configurable. So I don't quite understand what @galvesribeiro 's points. Mind explain more ?

xiazen:
@attilah  I think we can add a validation to see if a "PubSubStore" grain storage is registered into DI. should be trivia, and also bring a lot benefits to users. I personally tripped on this from time to time.

jason-bragg:
@xiazen, @galvesribeiro, I'll leave the clarification of galvesribeiro's post to him, but the requirement that a user needs be aware that a particular implementation of the pubsub system uses grains which use a specific grain state storage named "pubsubstore" (which needs be configured) is very cryptic and unintuitive.  This implementation detail is unrelated to the public stream surfaces, but is the type of shortcoming that should be addressed if we revamp streams.

Recent (and future) refactoring of components allow more options.  For instance, if the pubsub grains used storage facets, the name of the grain storage could be configured, rather than being required to always be "pubsubstore".  This would be much more intuitive and flexible.

galvesribeiro:
@xiazen 
> the name "PubSubStore" is always hard coded to be the pubsubgrain state store name

I know, and that is the problem.

@jason-bragg 

> I'll leave the clarification of galvesribeiro's post to him

Don't need to. As always, you did that for me! 💃

xiazen:
submitted a pr #4273 to add the validation.

xiazen:
fixed in #4273

PR #4273, opened 2018-03-21
Add validation for pubsubstore when using persistent streams

xiazen:
Persistent streams which has Explicit pubsub configured requires a grain storage with name `PubSubStore` configured with silo. But this is not obvious to users. So in this PR, adding a validation for this if user forget to configure one. 

Fix for #4196

### DIFF
--- a/src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
+++ b/src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
@@ -8,9 +8,35 @@ using Orleans.Runtime;
 using System;
 using System.Collections.Generic;
 using System.Text;
+using Orleans.Storage;
 
 namespace Orleans.Streams
 {
+    public class PersistentStreamStorageConfigurationValidator : IConfigurationValidator
+    {
+        private IServiceProvider services;
+        private const string pubsubStoreName = "PubSubStore";
+        private string streamProviderName;
+        public PersistentStreamStorageConfigurationValidator(IServiceProvider services, string streamProviderName)
+        {
+            this.services = services;
+            this.streamProviderName = streamProviderName;
+        }
+
+        public void ValidateConfiguration()
+        {
+            var pubsubOptions = services.GetOptionsByName<StreamPubSubOptions>(this.streamProviderName);
+            if (pubsubOptions.PubSubType != StreamPubSubType.ImplicitOnly)
+            {
+                var pubsubStore = services.GetServiceByName<IGrainStorage>(pubsubStoreName);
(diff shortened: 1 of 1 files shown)


src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
```diff
@@ -8,9 +8,35 @@
...
+        public PersistentStreamStorageConfigurationValidator(IServiceProvider services, string streamProviderName)
+        {
+            this.services = services;
+            this.streamProviderName = streamProviderName;
+        }
+
+        public void ValidateConfiguration()
+        {
+            var pubsubOptions = services.GetOptionsByName<StreamPubSubOptions>(this.streamProviderName);
+            if (pubsubOptions.PubSubType != StreamPubSubType.ImplicitOnly)
```
jason-bragg: It would be better to explicitly check the pubsub systems this check applies to, rather than those it does not.  That would be more extensible if/when more implementations are added.

src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
```diff
@@ -8,9 +8,35 @@
...
+using Orleans.Storage;
 
 namespace Orleans.Streams
 {
+    public class PersistentStreamStorageConfigurationValidator : IConfigurationValidator
+    {
+        private IServiceProvider services;
+        private const string pubsubStoreName = "PubSubStore";
+        private string streamProviderName;
+        public PersistentStreamStorageConfigurationValidator(IServiceProvider services, string streamProviderName)
```
jason-bragg: It makes sense to have a PersistentStreamStorageConfigurationValidator, but I'm wondering if this specific logic would be better as a per pubsub system validator, rather than a single validator expected to know about all the possible pubsub systems.

src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
xiazen: sounds good

src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
xiazen: if pubsub system is injectable, then I think we should have a per pubsub validator. But that is not the case yet. So, I suggest let's worry about that when we have other pubsub systems in the future.

ReubenBond:
@dotnet-bot test functional

xiazen:
can we merge this now?

PR #4276, opened 2018-03-22
[WIP] Cluster settings as service, to allow overrides

jason-bragg:
Prototype to explore a means of overriding cluster options by named component (mostly providers).

Instead of accessing the cluster options directly, we access an IClusterSettings service which is populated via the cluster options.  Since we get the settings from a service, components can try to get the service by name or fall back to the default if a named service is not found.  This allows application developers to register names services that can be used by a component to override the IClusterSettings for that component.

### DIFF
--- a/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorage.cs
+++ b/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorage.cs
@@ -28,7 +28,7 @@ namespace Orleans.Storage
     public class AzureTableGrainStorage : IGrainStorage, IRestExceptionDecoder, ILifecycleParticipant<ISiloLifecycle>
     {
         private readonly AzureTableStorageOptions options;
-        private readonly ClusterOptions clusterOptions;
+        private readonly IClusterSettings clusterSettings;
         private readonly SerializationManager serializationManager;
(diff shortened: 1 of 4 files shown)


src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorage.cs
```diff
@@ -498,7 +498,8 @@ public static class AzureTableGrainStorageFactory
         public static IGrainStorage Create(IServiceProvider services, string name)
         {
             IOptionsSnapshot<AzureTableStorageOptions> optionsSnapshot = services.GetRequiredService<IOptionsSnapshot<AzureTableStorageOptions>>();
-            return ActivatorUtilities.CreateInstance<AzureTableGrainStorage>(services, optionsSnapshot.Get(name), name);
+            IClusterSettings clusterSettings = services.GetServiceByNameOrDefault<IClusterSettings>(name);
```
jason-bragg: If the application layer registers a populated ClusterOptions by name as an IClusterSettings service, it will be picked up here.  If they don't the clusteroptions defined for the service will be used.  Any provider that does not do this can access the cluster options via IOptions, as is currently done, or via the default IClusterSettings interface.

sergeybykov:
Closing this in favor of #4277.

PR #4277, opened 2018-03-22
Overridable options

jason-bragg:
Another prototype, like #4276, which allows override of options, by registering names services to override an option.

### DIFF
--- a/src/Orleans.Core/Utils/KeyedService.cs
+++ b/src/Orleans.Core/Utils/KeyedService.cs
@@ -91,7 +91,7 @@ namespace Orleans.Runtime
         public static IOptions<TOptions> GetOverridableOption<TOptions>(this IServiceProvider services, string key)
             where TOptions : class, new()
         {
-            TOptions option = services.GetRequiredServiceByName<TOptions>(key);
+            TOptions option = services.GetServiceByName<TOptions>(key);
             return option != null
                 ? Options.Create(option)
                 : services.GetRequiredService<IOptions<TOptions>>();
--- a/src/Orleans.Core/Utils/KeyedService.cs
+++ b/src/Orleans.Core/Utils/KeyedService.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
 
 namespace Orleans.Runtime
 {
@@ -84,6 +85,18 @@ namespace Orleans.Runtime
 
     public static class KeyedServiceExtensions
     {
+        /// <summary>
+        /// Gets option that can be overriden by named service.
+        /// </summary>
+        public static IOptions<TOptions> GetOverridableOption<TOptions>(this IServiceProvider services, string key)
+            where TOptions : class, new()
+        {
+            TOptions option = services.GetRequiredServiceByName<TOptions>(key);
+            return option != null
+                ? Options.Create(option)
+                : services.GetRequiredService<IOptions<TOptions>>();
+        }
+
         /// <summary>
         /// Register a transient keyed service
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Orleans.Core/Utils/KeyedService.cs
```diff
@@ -84,6 +85,18 @@ public TService GetService(IServiceProvider services, TKey key)
 
     public static class KeyedServiceExtensions
     {
+        /// <summary>
+        /// Gets option that can be overriden by named service.
+        /// </summary>
+        public static IOptions<TOptions> GetOverridableOption<TOptions>(this IServiceProvider services, string key)
+            where TOptions : class, new()
+        {
+            TOptions option = services.GetRequiredServiceByName<TOptions>(key);
```
jason-bragg: this should be GetServiceByName not GetRequiredServiceByName

src/Orleans.Core/Utils/KeyedService.cs
```diff
@@ -84,6 +85,18 @@ public TService GetService(IServiceProvider services, TKey key)
...
+        /// Gets option that can be overriden by named service.
+        /// </summary>
+        public static IOptions<TOptions> GetOverridableOption<TOptions>(this IServiceProvider services, string key)
+            where TOptions : class, new()
+        {
+            TOptions option = services.GetServiceByName<TOptions>(key);
+            return option != null
+                ? Options.Create(option)
+                : services.GetRequiredService<IOptions<TOptions>>();
+        }
```
ReubenBond: If we do it this way, we're circumventing the `IOptions<T>` system for the named ones - is that the intention? It means that we need to inject overrides as `TOptions` while under the hood we're usually injecting `IConfigureOptions<TOptions>`

Additionally, the delineation between this and _named options_ is not obvious

ReubenBond: I'm _okay_ with this. We could also introduce a special type to make the intention more clear, like `IOptionsOverride<T>` or `NotUnlessYouReallyHaveTo<T>`.

> We just can't use named options without some way of detecting if an override is set or not.

We could work it out - by checking for registrations of `IConfigureOptions<TOptions>`, or whatever the named options variant is.

Anyhow, looks tenable for me as long as we're clear that we're limiting the scope of this.

jason-bragg: We could add an extension function over the builder (or service collection) which registers an override as a named option.

    ISiloHostBuilder OverridableOptions<T>(this ISiloHostBuilder, sting name, OptionsBuilder<T> builder = null)

That would probably make the pattern cleaner.

    builder
        .AddSomeProvider("Blarg", options => options.Something = something)
        .OverrideOptions<ClusterOptions>("Blarg", builder => builder.Configure(options => options.ServiceId = "oldStuff"));

jason-bragg: I'd considered IOptionsOverride or something like it.  I'd not ruled it out, but it seemed like an extra allocation that was not necessary in the current pattern.  It is more uniform with IOptionsSnapshot though, as we could make the interfaces lineup.

I'd also considered registering a specific instance of IOptionsSnapshot<T> for a concrete type, which would override the default behavior, but I couldn't see a good way to get at the default behavior when needed, like if a user wanted to use named options for the overrides.  Not saying there is no way of doing this.. just not figured one out..

xiazen: between the two pr. I like this one more, since conceptually we are overriding option, keyed service is just implementation details, especially if we add the `ISiloHostBuilder.OverrideOptions<T>` method. The other one emphasize too much on keyed service. 

It is still sneaky, as in it is not clear which option, used by which component,  would be overridden . But I guess this is by design. As in this is an advanced feature, we don't expect user to use it much. But then, as a sneaky feature, it lives on iSiloHostBuilder, which is a very obvious place. Can we add something here to prevent user misuse it . or limit its visibility. Does this even make sense ?

ReubenBond:
If the change looks fine, we can merge

ReubenBond:
Scratch that. If this doesn't look fine, we can revert ;)

PR #13, opened 2015-01-24
Fix warning CA2002: Do not lock on objects with weak identity

zerhacken:
This commit fixes warning CA2002. HierarchicalKeyStore extends
MarshalByRefObject which allows code code running in one AppDomain
to lock on an object in another AppDomain. Fix is to lock on a private
object which only HierarchicalKeyStore is knowledgeable about.

PR #4280, opened 2018-03-22
Validate IReminderTable during startup if any grains implement IRemindable

### LINKED ISSUE #4278: Obscure error message when using Reminders without registering a reminders provider
```C#
  public class Info : Grain, IInfo, IRemindable
    {
        public Task ReceiveReminder(string reminderName, TickStatus status)
        {
            Console.WriteLine("Thanks for reminding me-- I almost forgot!");
            return Task.CompletedTask;
        }
        public async override Task OnActivateAsync()
        {
            Console.WriteLine("OnActivateAsync..");
            return;
        }
        public async Task<string> Get()
        {
             Console.WriteLine("get..");
            TimeSpan t1 = TimeSpan.FromMinutes(2);
            TimeSpan t2 = TimeSpan.FromMinutes(3);
            try
            {
                IGrainReminder reminder = await RegisterOrUpdateReminder("testtime", t1, t2);
            }
            catch (Exception ex)
            {
                throw ex;
            }
            //base.RegisterTimer(OnClientRefreshTimer, null, t1 - TimeSpan.FromSeconds(2), t1);

            return "info";
        }
    }
```
```
SystemTarget *stg/0/00000000 not active on this silo. Msg=Request S127.0.0.1:11112:259398066*grn/54700641/00000000@e5d40db1->S127.0.0.1:11112:259398066*stg/0/00000000@S00000000 #23: 
```
```C#
  client = StartClientWithRetries().Result;

            var info = client.GetGrain<IInfo>(0);
            var result = info.Get();
```
I want to implement ReceiveReminder at ten o'clock every day

ReubenBond:
Fixes #4278 

This assumes that if the user adds some grain which implements `IRemindable`, then that means they intend to use reminders. That assumption might not be true, but maybe we'll decide it's better to err on the side of caution.

An alternative is late checking, like so:
image

I'm happy with either and looking for input.

### DIFF
--- a/src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
+++ b/src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
@@ -38,6 +38,7 @@ using Microsoft.Extensions.Options;
 
 using Orleans.Configuration.Validators;
 using Orleans.Runtime.Configuration;
+using System.Linq;
 
 namespace Orleans.Hosting
 {
@@ -92,7 +93,6 @@ namespace Orleans.Hosting
 
             services.AddLogging();
             services.TryAddSingleton<ITimerRegistry, TimerRegistry>();
-            services.TryAddSingleton<IReminderRegistry, ReminderRegistry>();
             services.TryAddSingleton<GrainRuntime>();
             services.TryAddSingleton<IGrainRuntime, GrainRuntime>();
             services.TryAddSingleton<IGrainCancellationTokenRuntime, GrainCancellationTokenRuntime>();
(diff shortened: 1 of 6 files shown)


ReubenBond:
@dotnet-bot test functional

> UnitTests.ActivationsLifeCycleTests.DeactivateOnIdleTests.DeactivateOnIdle_NonExistentActivation_1 
>  MESSAGE:
Orleans.Runtime.OrleansLifecycleCanceledException : Lifecycle start canceled due to errors at stage 20000\r\n---- Orleans.Runtime.OrleansMessageRejectionException : Forwarding failed: tried to forward message NewPlacement Request S127.0.0.1:27756:259412658*stg/0/00000000@S00000000->S127.0.0.1:27756:259412658*grn/FCB3F509/00003039@a067ac8a #13: for 0 times after Duplicate activation to invalid activation. Rejecting now.

ReubenBond:
Maybe we should table this approach and go for the lazy approach in the diff image above.

sergeybykov:
Let's close this one for now. The concern is that this might be too aggressive, and lead to unintended consequences in corner cases.

PR #4287, opened 2018-03-22
Reminders: Improve error message when using reminders without a reminder table

### LINKED ISSUE #4278: Obscure error message when using Reminders without registering a reminders provider

ReubenBond:
Alternative implementation of #4280

Fixes #4278

### DIFF
--- a/src/Orleans.Runtime/ReminderService/ReminderRegistry.cs
+++ b/src/Orleans.Runtime/ReminderService/ReminderRegistry.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+
+using Microsoft.Extensions.DependencyInjection;
 using Orleans.Runtime.Services;
 using Orleans.Timers;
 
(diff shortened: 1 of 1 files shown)


src/Orleans.Runtime/ReminderService/ReminderRegistry.cs
```diff
@@ -1,21 +1,36 @@
...
+using Microsoft.Extensions.DependencyInjection;
 using Orleans.Runtime.Services;
 using Orleans.Timers;
 
 namespace Orleans.Runtime.ReminderService
 {
     internal class ReminderRegistry : GrainServiceClient<IReminderService>, IReminderRegistry
     {
+        private const string ReminderServiceNotConfigured =
+           "The reminder service has not been configured. Reminders can be configured using extension methods from the following packages:"
```
jason-bragg: Not convinced we should be listing all of these in code.
No good way to keep this list accurate.
A reference to docs should be sufficient.

PR #4301, opened 2018-03-24
Add Dns name look up clustering for container frameworks

jsukhabut:
Docker/Kubernetes can communicate with other containers using DNS name lookup.  When the silo containers are scaled and there are multiple instances, we can use Dns.GetHostEntry to look up the multiple addresses and pass it to UseStaticClustering().

The problem is that this list is static so as I scale up and down, or as containers shutdown and restart, the client only has a static list of gateways.  We should do a continuous DNS resolution so the client will have the most recent list of IPs.
```
builder.UseDnsNameLookupClustering("mysilocontainer", 40000);
```
I had created a web UI Orleans client in a container and the silo in another container.  After scaling the silo container up and down and failing out some of the original containers, the client could no longer connect with the silo because it has a stale IP list.

### DIFF
--- a/src/Orleans.Core/Configuration/Options/DnsNameGatewayListProviderOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/DnsNameGatewayListProviderOptions.cs
@@ -1,0 +1,21 @@
+using System;
+using System.Collections.Generic;
+
+namespace Orleans.Configuration
+{
+    /// <summary>
+    /// Options for Configure DnsNameGatewayListProviderOptions
+    /// </summary>
+    public class DnsNameGatewayListProviderOptions
+    {
+        /// <summary>
+        /// Dns name to use
+        /// </summary>
+        public string DnsName { get; set; }
+
+        /// <summary>
+        /// Port to use
+        /// </summary>
+        public int Port { get; set; }
+    }
+}
--- a/src/Orleans.Core/Messaging/DnsNameGatewayListProvider.cs
+++ b/src/Orleans.Core/Messaging/DnsNameGatewayListProvider.cs
@@ -18,13 +18,11 @@ namespace Orleans.Messaging
     public class DnsNameGatewayListProvider : IGatewayListProvider
     {
         private readonly DnsNameGatewayListProviderOptions options;
-        private readonly TimeSpan maxStaleness;
-        private readonly ILoggerFactory loggerFactory;
+        private readonly TimeSpan maxStaleness;        
         private readonly ILogger logger;
 
         public DnsNameGatewayListProvider(ILoggerFactory loggerFactory, IOptions<DnsNameGatewayListProviderOptions> options, IOptions<GatewayOptions> gatewayOptions)
-        {
-            this.loggerFactory = loggerFactory;
+        {            
             this.logger = loggerFactory.CreateLogger<DnsNameGatewayListProvider>();
             this.options = options.Value;
             this.maxStaleness = gatewayOptions.Value.GatewayListRefreshPeriod;
@@ -35,24 +33,32 @@ namespace Orleans.Messaging
         public Task InitializeGatewayListProvider() => Task.CompletedTask;
 
 
-        public Task<IList<Uri>> GetGateways()
+        public async Task<IList<Uri>> GetGateways()
         {
             try
             {
-                var endpointUris = Dns.GetHostEntry(this.options.DnsName)
+                var addresses = (await Dns.GetHostEntryAsync(this.options.DnsName))
                                                 .AddressList
-                                                .Select(a => new IPEndPoint(a, this.options.Port).ToGatewayUri())
-                                                .OrderBy(a => a.AbsoluteUri)
+                                                .OrderBy(a => a.ToString())
                                                 .ToList();
 
-                return Task.FromResult<IList<Uri>>(endpointUris);
+                var endpointUris = addresses.Select(a => new IPEndPoint(a, this.options.Port).ToGatewayUri())
+                                            .ToList();
+                
+                if (this.logger.IsEnabled(LogLevel.Debug))
+                {
+                    var addressesStr = string.Join("\n", addresses.Select(s => s.ToString()));
+                    this.logger.Debug($"DNS Gateway {this.options.DnsName} resolved to: \n {addressesStr}");
+                }
+
+                return endpointUris;
             }
             catch (System.Net.Sockets.SocketException se)
             {
                 if (se.Message == "No such host is known")
                 {
-                    logger.Warn(ErrorCode.ProxyClient_GetGateways, $"No addresses found for silo gateways with DNS hostname: {this.options.DnsName}");
-                    return Task.FromResult<IList<Uri>>(new List<Uri>());
+                    logger.Warn(123, $"No addresses found for silo gateways with DNS hostname: {this.options.DnsName}");
+                    return new List<Uri>();
                 }
                 throw;
             }
--- a/src/Orleans.Core/Messaging/DnsNameGatewayListProvider.cs
+++ b/src/Orleans.Core/Messaging/DnsNameGatewayListProvider.cs
@@ -1,0 +1,55 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
+using Orleans.Configuration;
+using Orleans.Runtime;
+
+namespace Orleans.Messaging
+{
+    /// <summary>
+    /// This Gateway list provider looks up ip addresses based on a dns
+    /// name.  This is ideal for container environments (kubernetes, swarm)
+    /// as well as Service Fabric using the dns feature.
+    /// </summary>
+    public class DnsNameGatewayListProvider : IGatewayListProvider
+    {
+        private readonly DnsNameGatewayListProviderOptions options;
+        private readonly TimeSpan maxStaleness;
+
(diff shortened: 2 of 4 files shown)


jms69:
@dotnet-bot test functional
@dotnet-bot test bvt

jms69:
@dotnet-bot test bvt

src/Orleans.Core/Messaging/DnsNameGatewayListProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+        #region Implementation of IGatewayListProvider
+
+        public Task InitializeGatewayListProvider() => Task.CompletedTask;
+
+
+        public Task<IList<Uri>> GetGateways()
+        {
+            try
+            {
+                var endpointUris = Dns.GetHostEntry(this.options.DnsName)
```
benjaminpetit: Why not use `Dns.GetHostEntryAsync` here?

src/Orleans.Core/Messaging/DnsNameGatewayListProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// <summary>
+    /// This Gateway list provider looks up ip addresses based on a dns
+    /// name.  This is ideal for container environments (kubernetes, swarm)
+    /// as well as Service Fabric using the dns feature.
+    /// </summary>
+    public class DnsNameGatewayListProvider : IGatewayListProvider
+    {
+        private readonly DnsNameGatewayListProviderOptions options;
+        private readonly TimeSpan maxStaleness;
+        private readonly ILoggerFactory loggerFactory;
```
benjaminpetit: Not very important: You don't use the loggerFactory in the class at all, you could remove this field

src/Orleans.Core/Messaging/DnsNameGatewayListProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    /// <summary>
+    /// This Gateway list provider looks up ip addresses based on a dns
+    /// name.  This is ideal for container environments (kubernetes, swarm)
+    /// as well as Service Fabric using the dns feature.
+    /// </summary>
+    public class DnsNameGatewayListProvider : IGatewayListProvider
+    {
+        private readonly DnsNameGatewayListProviderOptions options;
+        private readonly TimeSpan maxStaleness;
```
benjaminpetit: Not very important: You keep a reference on the `DnsNameGatewayListProviderOptions`, you could just read the value from this option?

src/Orleans.Core/Configuration/Options/DnsNameGatewayListProviderOptions.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using System.Collections.Generic;
+
+namespace Orleans.Configuration
+{
+    /// <summary>
+    /// Options for Configure DnsNameGatewayListProviderOptions
+    /// </summary>
+    public class DnsNameGatewayListProviderOptions
```
benjaminpetit: Could we have a validator for this option, for example check if the port is not equal to `0` or check if the `DnsName` is resolvable?

benjaminpetit:
Thanks for your contribution!

Even if this seems low risk, I think we should wait to merge this and not include it in the first 2.0 release. If people wants to use this provider, they can copy it and inject it themself in the `SiloHostBuilder`.

The main risk I see with such provider is on how often the DNS results are updated in the infrastructure and how Orleans client will react when a gateway disappear from the list if it is still connected.

Also you will still need a membership provider for the silos (azure, consul, ...), the same connection string can be used for the client.

I want to also point that **silo endpoints are not supposed to be public**, be sure to keep them on private/internal endpoint since we don't have any security or validation on connection.

jms69:
I am working on some AKS [samples](https://github.com/jms69/orleans/tree/akssample/Samples/2.0/Kubernetes).

 The built-in discovery mechanism of Kubernetes is using DNS name resolution.  In one deployment scenario, I am deploying the silo containers as a service with no ClusterIP (internal vip).  See [diagram](https://github.com/jms69/orleans/blob/akssample/Samples/2.0/Kubernetes/deployment/AKS/twoservices.png).  Each pod of the service is internal and not exposed.  The service can be located by DNS name and the IPs of all instances will be returned when resolved.  I am using DnsNameGatewayListProvider locally within the sample code.  I prefer this method because UseAzureStorageClustering()requires me to inject the azure storage connection string in the client container so the container will have write credentials to the storage and membership table.

One other consideration is that the silo can easily switch to other membership providers, for example using consul internally in the Kubernetes cluster.  The client will be unaffected by this change when using DNS name resolution.

I totally understand the decision to make this a core code or not.  I will make the fix to DnsNameGatewayListProvider anyways then the decision can be made later if this should just be included or not.

benjaminpetit:
Hey, could you fix the comments on this PR? I think we can merge it for a future 2.1 release

jms69:
Ok.  I will do.  Sorry for the delay day job got in the way 🙂

jdom:
Ping @galvesribeiro as he was working on some deep integration with Kubernetes. Maybe it's unrelated, but just a heads up in case it makes sense to join forces.

benjaminpetit:
Not sure if I remembered correctly what you proposed previously @galvesribeiro , but the call to `Dns.GetHostEntry`` is supposed to return the IP of the running silos. So Orleans will connect directly to individual silos.

If it is not the case and instead return the IP of K8 nodes or Swarm worker, then yes this will not work properly. Orleans client expect to have a direct connections to the cluster, since it is doing some logic based on the ip address of the silos.

galvesribeiro:
There are 2 levels of DNS used by KubeProxy and Kubeletes.

1. Pod DNS -> Resolve to the single Pod IP in the cluster.
2. Service DNS -> It resolve to (1) the Service IP or (2) the active Pods IPs selected by the deployment selector. In case of Service IP, it will use a balancing algoh like Round Robin for example.

This PR assume the settings on Kube Deployment is using a Service DNS resolving to Pods IPs which is not _always_ true as the user can configure it to resolve to a load balancer.

Regardless of how it was configured, there still the OS' DNS cache TTL. I think we will eventually hit problems and more questions with it.

Both Kubernetes and Swarm deserves a proper membership provider crafted to its reality and not just _trust_ on DNS names and that was the motivation to create that provider.

jms69:
@galvesribeiro I do not disagree that we need a membership provider for Kubernetes and Docker especially for silo-to-silo discovery.  It would be grate if reminders and state storage can be added as a single solutions in Kubernetes.

This DNS clustering name lookup is just an extension of the static gateway one.  I am deploying to silo's as a statefulset set with a specific number of instance, and I want to inject the IPs of these silos to my client why not a simple DNS lookup that returns multi-ip response of all the silos that exists?  Yes there is a small refresh delay if I scale down but I think that will exist with any provider.  This was the scenario I was building in my example (note that in the example still uses Azure table for silo-to-silo membership) https://github.com/jms69/orleans/tree/akssample/Samples/2.0/Kubernetes/deployment/AKS

Deploying ZooKeeper/Consul in the cluster would also achieve a complete solution.  There are many options here, DNS resolution is just one of them.

galvesribeiro:
@jms69 

>  It would be grate if reminders and state storage can be added as a single solutions in Kubernetes.

Reminders can easily be integrated. I'll do it later on in another package. Storage... I need to think more about that... Kubernetes objects are meant to store metadata (like I did in the provider) so Clustering info and reminders are ok. However, persist full grain state is something that would heavily impact the performance of Kubernetes internal ectd, so I would not do that. In other hand, there are distributed storage mechanisms on Kubernetes that can be useful to store grain state. I just need to have a look on it later on.

Well, if your intent is just to have Orleans _running_ on Kubernetes feel free to extent the gateway provider to have the option to resolve thru DNS names. However, I'm against this PR be merged on Orleans main source code but instead, it should be its own package and probably OrleansContrib is the best place to keep it. 

IMHO, have this on Orleans core would bring more confusion to users than benefits and if the intent is to properly support Kubernetes on Orleans main code, I would bring in the provider. 

Again, this is my own opinion and the core team can think different and merge it on their own.

jms69:
DNS gateway lookup is just as practical as the Static Gateway.  In some scenarios I want a static gateway list that has IP resolvable at runtime and not config time.  This scenario is not specific to Kubernetes.  I am currently building a Service Fabric app and am considering using the DNS look up feature as well because it seems Microsoft.Orleans.Clustering.ServiceFabric seems to have disappeared from the nuget.

The class is so small with no dependencies so it can either be in the main code or separate nuget.

src/Orleans.Core/Configuration/Options/DnsNameGatewayListProviderOptions.cs
jms69: Validator for port < 0 completed.

I put in the DnsName resolvable validator but took it out.  The component already warns if DnsName is not resolvable and the silo handles the empty list accordingly.  The nice part is that it recovers very nicely once the DNS name is registered.

There are scenarios where the client component/frontend may complete deployment before the backend silo have not completed deployment and the DNS entries have not been registered.  If we throw in validation the client may crash on Init.

galvesribeiro: The client should throw if there is no gateways available because the DNS isn't there. No problem with that. The client has an optional parameter to pass a delegate for retries so it can retry again. It is common to have the client initialize faster than silos.

benjaminpetit:
Closing because it appears that it is not needed anymore. Fell free to update this thread if you disagree

GersonDias:
What is the recommend way for hosting silos on kubernetes? There are some guidance on it?

PR #4309, opened 2018-03-26
Fix DelayedQueueRebalancingTests test set up

xiazen:
Somewhere between 1.5 and 2.0, the set up for `DelayedQueueRebalancingTests` changed. 
In 1.5, the test cluster size is 4, but only 2 silo started at the beginning. There's another 2 started later in `DelayedQueueRebalancingTests_2` . Hence the name -- delayed... I guess it means delayed silo starting. 

Before I thought there's a bug in `DynamicClusterConfigDeploymentBalancer`. But now I think the failure before is caused by `StartAdditionalSilos` behave differently between 1.5 and 2.0. and also the test set up was different.

### DIFF
--- a/test/TesterAzureUtils/Streaming/DelayedQueueRebalancingTests.cs
+++ b/test/TesterAzureUtils/Streaming/DelayedQueueRebalancingTests.cs
@@ -1,15 +1,19 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Orleans.Configuration;
 using Orleans.Hosting;
+using Orleans.Logging;
 using Orleans.Providers.Streams.AzureQueue;
 using Orleans.Providers.Streams.Common;
 using Orleans.Runtime;
 using Orleans.Runtime.Configuration;
 using Orleans.Streams;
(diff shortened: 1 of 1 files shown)


xiazen:
ServiceBus.Tests.PluggableQueueBalancerTestsWithEHStreamProvider.PluggableQueueBalancerTest_ShouldUseInjectedQueueBalancerAndBalanceCorrectly test failure not related

test/TesterAzureUtils/Streaming/DelayedQueueRebalancingTests.cs
```diff
@@ -72,13 +75,12 @@ public DelayedQueueRebalancingTests()
             await ValidateAgentsState(2, 4, "2");
         }
 
-        [SkippableFact(Skip = "https://github.com/dotnet/orleans/issues/3993"), TestCategory("Functional")]
+        [SkippableFact, TestCategory("Functional")]
```
benjaminpetit: Shouldn't you remove the `SkippableFact` par entirely?

xiazen: No this test depends on AzureQueue to be set up, so it will still be skipped in jekins

xiazen:

xiazen:
@dotnet-bot test functional

Issue #4322, opened 2018-03-27
Intermittent failures of  SMS_AllSilosRestart_UnsubscribeConsumer test


sergeybykov:
https://orleans.visualstudio.com/Orleans/_TestManagement/Runs?_a=resultSummary&runId=4155&resultId=101189
>Error message
Orleans.Runtime.OrleansMessageRejectionException : No gateways available
Stack trace
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
at UnitTests.Streaming.Reliability.StreamReliabilityTests.<SMS_AllSilosRestart_UnsubscribeConsumer>d__40.MoveNext() in D:\build\agent\_work\15\s\test\TesterAzureUtils\Streaming\StreamReliabilityTests.cs:line 626
--- End of stack trace from previous location where exception was thrown ---
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

benjaminpetit:
Not a 2.0 blocker.

In the test we restart silos, and sometimes we try to make a call grom the client before it had a chance to refresh the list of gateway. Refreshing it more often + waiting a bit before making the final client call should fix this test.

ReubenBond:
Closing all test reliability issues. Test reliability has improved substantially. We can re-open for any test reliability issues which we still see.

PR #4327, opened 2018-03-27
Fix SMS_AllSilosRestart_UnsubscribeConsumer

benjaminpetit:
Related issue: #4322

### DIFF
--- a/test/TesterAzureUtils/Streaming/StreamReliabilityTests.cs
+++ b/test/TesterAzureUtils/Streaming/StreamReliabilityTests.cs
@@ -64,12 +64,13 @@ namespace UnitTests.Streaming.Reliability
                 {
                     gatewayOptions.ConnectionString = TestDefaultConfiguration.DataConnectionString;
                 })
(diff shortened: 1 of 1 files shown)


benjaminpetit:
@dotnet-bot test this please

benjaminpetit:
@dotnet-bot test functional

test/TesterAzureUtils/Streaming/StreamReliabilityTests.cs
```diff
@@ -621,7 +622,23 @@ public void Baseline_StreamRel_RestartSilos()
...
 
-            await consumerGrain.UnSubscribeFromAllStreams();
+            // Since we restart all silos, the client might not haave had enough
+            // time to reconnect to the new gateways. Let's retry the call if it 
+            // is the case
+            for (int i = 0; i < 3; i++)
+            {
+                try
+                {
+                    await consumerGrain.UnSubscribeFromAllStreams();
```
ReubenBond: The refresh time is 5s but we're only awaiting a total of 300ms here, is that fine?

benjaminpetit: Yes because when trying to send a request, if the client see that there are no gateway available, it will enqueue a refresh. The current call will fail, but the next should work

PR #4336, opened 2018-03-28
Adding migration path for azure worker role documentation

miker1423:


xiazen:
Thanks so much for the contribution ! @miker1423 really appreciate it .

PR #4290, opened 2018-03-23
Second pass at options overrides

jason-bragg:
- Added registration functions that use named options and register the options for logging
- Updated named components that use cluster options to support the overrides.

Unclear what to do about unnamed components that use cluster options.  At this time they are not being overridden.  Maybe we can use the component name, or abstraction name?  (IReminderTable?, for instance).  What to do with unnamed components is for discussion and planning possible additional pass, should not block this PR.

PR #4339, opened 2018-03-28
Update change log for final 2.0.0

sergeybykov:


xiazen:
is this the final one or just an intermedia step ?

sergeybykov:
Final.

changelog.md
```diff
@@ -7,6 +7,49 @@ All notable end-user facing changes are documented in this file.
 *Here are all the changes in `master` branch, and will be moved to the appropriate release once they are included in a published nuget package.
 The idea is to track end-user facing changes as they occur.*
 
+### [2.0.0] (changes since 2.0.0-rc2)
+
+- Major changes
+  - All included providers obtain ServiceId and ClusterId from the global ClusterOptions and do not have those properties on their own options classes (#4235, #4277, 4290)
+  - Use string for ServiceId instead of Guid (#4262)
```
xiazen: we should add a documentation page stating that now all partitionkey in persistence mechanism all use ServiceId now, instead of some use DeploymentId, some use ServiceId some use both. And link that doc here.

xiazen: I'm afraid the subtle change would mess up user's application in a way which is  hard to detect. If we don't state it here, it is hard for them to find out,

sergeybykov: It's too late to be afraid. :-)
I don't think this specific change is actually a big deal. A Guid is trivial to convert to a string.

The potential problem is if somebody used cluster ID instead of service ID for service state. And that's what #4235, #4277, and #4290 are more directly are related to. I don't think release notes is a good place to document that.

xiazen: I'm not saying this "Use string for ServiceId instead of Guid". The change I'm talking about is the unifying all partition key to use service Id , instead of deployment id, which I have no where to comment on. But I do brought up this concern many times and I think it deserve its own migration page..

xiazen: release note doesn't have space to document that, but I think it is important to call it out in the release note and link the doc page besides. Since release not is the first doc user is going to read when migrating.  Otherwise how do you suppose the user will discover this?

sergeybykov: I'm not sure about its own migration page, maybe, but we should explain in the configuration page for sure.

xiazen: offline discussion decided to add a paragraph to explain this nuance and how to migrate off it and link the doc in release note. But we can merge this one since we can add that link to release note directly. not necessarily in the change log.

PR #4344, opened 2018-03-28
[WIP] Improve activation & directory convergence

ReubenBond:
This is for consideration - I'm not expecting it to be merged as-is.

This changes the implementation of GrainDirectoryHandoffManager to use a queue for executing operations. Each operation must complete successfully before the next operation begins execution.

Using that system, we add resiliency to some of the grain directory balancing operations.
When cluster membership changes, duplicate activations on existing silos are reliably destroyed.


ReubenBond:
@dotnet-bot test functional

sergeybykov:
>Note that this will retry each operation forever, until it succeeds. We could add upper limits on the number of retries for a specific operation, either in terms of time or retry count.

I think we should never retry forever, and hence I think we do need a limit. IMO a configurable time-based limit is better than a retry count-based one.

ReubenBond:
Superseded by https://github.com/dotnet/orleans/pull/4344

PR #4365, opened 2018-03-30
Update Interceptors.md

ReubenBond:


src/Documentation/Advanced-Concepts/Interceptors.md
```diff
@@ -1,37 +1,55 @@
 ---
 layout: page
-title: Interceptors
+title: Grain Call Filters
 ---
 
-# Interceptors
+# Grain Call Filters
```
ReubenBond: Unsure if we should update the file name

PR #4376, opened 2018-03-30
Fix assembly version to 2.0.0 & increment build version

sergeybykov:
Fix assembly version to 2.0.0, so that assemblies stay compatible.
Bump up build default version to 2.0.3 (because we already published [2.0.2 version of Microsoft.Orleans.ServiceFabric](https://www.nuget.org/packages/Microsoft.Orleans.ServiceFabric/2.0.2).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -105,7 +105,8 @@
 
   <!-- Versioning properties -->
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


sergeybykov:
@dotnet-bot test functional

Issue #2574, opened 2017-01-07
Make timer callbacks respect reentrancy of grains


sergeybykov:
Today, timer callbacks interleave with grain methods (at `await` points) for both reentrant and non-reentrant grains. This is a source of confusion because developers, naturally, do not expect interleaving in non-reentrant grains. We need to make timer callbacks respect reentrancy of grain classes and not interleave for non-reentrant grains.

cata:
Heads-up: this is a breaking change to an actively used feature(/bug? :-) )

We are actively relying on timer callback reentrancy. This being said, I agree that the behavior is confusing and should be changed. I think a straightforward migration path would be to call a reentrant grain method (specified using `AlwaysInterleave` or `MayInterleave`) from the timer callback.

sergeybykov:
@cata, thanks for the heads-up. I didn't realize people rely on this behavior.

kogir:
This timer reentrancy initially surprised me as well, but now our code has also come to rely on it. The non-reentrant methods in our grains often control flags used by the timer callback.

Would be happy to migrate, but this "loophole" is currently the only way to have a partially reentrant grain - a useful feature that would be sad to lose.

sergeybykov:
Maybe we should make the behavior configurable then, to explicitly allow for both reentrant and non-reentrant execution of timer callbacks?

cata:
@sergeybykov - configurable behavior (an extra parameter on a `RegisterTimer` overload?) would work great for us. 
The *default* behavior could be changed to match the grain's default reentrancy policy, thus making it less surprising.

gabikliot:
I think that is exactly what we agreed upon and planned to do a long time ago. That is at least what I remember.

jamescarter-le:
@sergeybykov @gabikliot Please can I clarify if this behavior is applicable to Reminders too?

sergeybykov:
@jamescarter-le Reminders are different in behavior from timers because their ticks are delivered via true grain calls that follow reentrancy/interleaving rules.

shlomiw:
It's very important for timers callback not to interleave. Optional configuration (to allow re-entrant) is also welcomed.
When do you plan releasing it? 

Please consider also an option to register a timer as such it wont allow Orleans deactivating the Grain.
Thanks.

sergeybykov:
We don't have concrete plans for this. But we are open to contributions.

shlomiw:
[...]

PR #4382, opened 2018-03-31
Don't enforce reentrancy for one way requests

dVakulen:
The deadlock prevention mechanism shouldn't affect the one-way requests, as they can not cause the deadlocks. 

This PR fixes the following cases:
- A -> B -> A chain:
image 1
 - And A -> A chain: 
image

Thanks to @ReubenBond for this finding and images.


ReubenBond:
What about requests where the user calls `grain.SomeMethod().Ignore()`? It's not technically `[OneWay]`

dVakulen:
They will interleave. `Ignore` should not change semantics of the request anyway.

ReubenBond:
That could be an issue, though, since an ignored message can arrive after the initiating request has finished and can interleave with calls which were not a part of the initial call chain

dVakulen:
Ignored message will be only of type `Response`, and it's interleaving should be safe. This potential issue, while being too costly to track, can be documented.

dVakulen:
While being too costly to track this point should be documented.

src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -343,37 +343,33 @@ public bool CanInterleave(ActivationData targetActivation, Message incoming)
...
             {
-                return;
+                return false;
             }
 
-            if (sendingActivation?.RunningRequestsSenders.Contains(outgoing.TargetActivation) == true)
-            {
-                outgoing.IsAlwaysInterleave = true;
-            }
+            var isCallFromOriginatorActivation = incoming.SendingActivation.Equals(targetActivation.Running.SendingActivation);
```
ReubenBond: I don't quite understand this condition, could you explain the reasoning with a comment?

ReubenBond: Is the point here to avoid immediate reentrancy, eg in the case of `this.AsReference<TSelf>().SomeMethod().Ignore()`?

That behavior may be worthy of its own option, but I suppose 'allow call chain reentrancy' doesn't preclude some cases where reentrancy within a chain is disallowed.

ReubenBond:
I'm considering interactions between `[AlwaysInterleave]` or timers which make grain calls, & `AllowCallChainReentrancy` and I believe there can be an issue under some circumstances.

Eg, if we have a grain which has an `[AlwaysInterleave]` method, then we can get a deadlock on call-chain reentrancy. Consider the following between 2 activations, `A` and `B`:

1. `A` receives `RegularMethod` call from client & awaits `Task.Delay(long time)`
2. `A` receives `AlwaysInterleaveMethod` call from client and calls `B`
3. `B` receives `A`'s request and calls back into `A`

At 3, `A`'s `ActivationData.Running` value will correspond to the `RegularMethod` request and hence the reentrancy will be denied, since it only supports a single running request. The issue is apparent on [ActivationData.cs:488](https://github.com/dotnet/orleans/blob/a5ac788b37b9481caf3cbdfdba5c70f1436e715b/src/Orleans.Runtime/Catalog/ActivationData.cs#L488), where `Running` is not updated to account for the `AlwaysInterleaveMethod` request.

The same issue would be apparent for calls initiated from timers, but in that case there is the additional trouble of there being no `Message` for the timer whatsoever and so we never even record `Running.CallChainId`.

Perhaps we should switch `RunningRequestSenders` into `RunningRequests` and have it be a collection of `Message` - or is that too costly? Are there alternatives?

... or am I wrong about this issue? I may be.

Reentrancy is so perilous...

dVakulen:
The first mentioned case won't cause the deadlock if `AlwaysInterleaveMethod` call can be completed sooner than initial `RegularMethod`. 
 <details><summary> Behavior explanation</summary>
<p>
<br/>
While RegularMethod fills the Running property  - the outgoing messages will have the call chain id of this request. Deadlock can occur only if RegularMethod ends earlier than AlwaysInterleaveMethod thus nullifying Running property, support of this case will require efforts similar to supporting of ` Handle long request detection for reentrant activations` (https://github.com/dotnet/orleans/blob/master/src/Orleans.Runtime/Catalog/ActivationData.cs#L490), and could be handled in separate PR .  
</p>
</details>


</p>
</p>
The case with calls initiated from timers works due to not yet implemented https://github.com/dotnet/orleans/issues/2574, there's a possibility that it will be supported automatically  after making timer callbacks respect reentrancy of grains.

dVakulen:
@dotnet-bot test bvt please 
Reason: unrelated failure (https://ci.dot.net/job/dotnet_orleans/job/master/job/bvt_prtest/1838/)

src/Orleans.Runtime/Core/Dispatcher.cs
dVakulen: Sorry for late reply; the point here is to avoid reentrancy for multiple requests which are coming from single origin (fan out to single target ).

PR #4388, opened 2018-04-03
Add a chatroom app which uses SMS stream as a streaming sample

xiazen:
# Chat Room Sample OVerview

This is a simple sample using [Orleans Streaming feature](http://dotnet.github.io/orleans/Documentation/Orleans-Streams/index.html) to build a simple chatroom application. In this application, each client can 
- set its user name for ther chatroom application
- join a channel by channel name
- send messages to the channel
- query history messages in the channel
- query members in the channel
- leave the channel

Each client will
- receive messages sent to the channel by other clients and itself after joined a channel
- stop receive messages after leaved the channel

For the purpose to make this sample simple, one client can only join one channel. It cannot join multiple channels at the same time. So to join another channel, the client need to leave the current channel first.

## Running the sample
From Visual Studio, you can start the OrleansServer project first, wait for it to stablize, and then start multiple OrleansClient projects simultaneously (you can start the project by right click the project -> Debug -> Start new instance).

Alternatively, you can run from the command line:

To start the silo
```
OrleansServer\bin\Debug(Release)\net462\OrleansServer.exe
```

To start the client (you will have to use a different command window)
```
OrleansClient\bin\Debug(Release)\net462\OrleansClient.exe
```
If you build the sample app in debug mode, the exe binary will be in Debug folder. If you build it in release mode, then the exe binary will be in Release folder.

After the client started up, you will see instructions printed on the console, which tells you how to interact with a channel.

## Reference
This sample is based on [a sample writted by @centur](https://github.com/centur/altnet-streams-demo), but revised with changes.

### DIFF
--- a/Samples/2.0/ChatRoom/GrainImplementation/Channel.cs
+++ b/Samples/2.0/ChatRoom/GrainImplementation/Channel.cs
@@ -27,7 +27,6 @@ namespace GrainImplementation
 
 		public async Task<Guid> Join(string nickname)
 		{
-			if (onlineMembers.Contains(nickname)) return Guid.Empty;
 			onlineMembers.Add(nickname);
 
 			await stream.OnNextAsync(new ChatMsg("System", $"{nickname} joins the chat '{this.GetPrimaryKeyString()}' ..."));
@@ -37,8 +36,6 @@ namespace GrainImplementation
 
 		public async Task<Guid> Leave(string nickname)
 		{
-			if (!onlineMembers.Contains(nickname)) return Guid.Empty;
-
 			onlineMembers.Remove(nickname);
 			await stream.OnNextAsync(new ChatMsg("System", $"{nickname} leaves the chat..."));
 
--- a/Samples/2.0/ChatRoom/GrainImplementation/Channel.cs
+++ b/Samples/2.0/ChatRoom/GrainImplementation/Channel.cs
@@ -1,0 +1,72 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using GrainInterfaces;
+using GrainInterfaces.Model;
+using Orleans;
+using Orleans.Streams;
+using Utils;
+
+namespace GrainImplementation
+{
+	public class Channel : Grain, IChannel
+	{
+		private readonly List<ChatMsg> messages = new List<ChatMsg>(100);
+		private readonly List<string> onlineMembers = new List<string>(10);
+
+		private IAsyncStream<ChatMsg> stream;
+
+		public override Task OnActivateAsync()
+		{
+			var streamProvider = GetStreamProvider(Constants.ChatRoomStreamProvider);
+
+		    stream = streamProvider.GetStream<ChatMsg>(Guid.NewGuid(), Constants.CharRoomStreamNameSpace);
+            return base.OnActivateAsync();
+		}
+
+		public async Task<Guid> Join(string nickname)
+		{
+			if (onlineMembers.Contains(nickname)) return Guid.Empty;
+			onlineMembers.Add(nickname);
+
+			await stream.OnNextAsync(new ChatMsg("System", $"{nickname} joins the chat '{this.GetPrimaryKeyString()}' ..."));
+
+			return stream.Guid;
+		}
+
+		public async Task<Guid> Leave(string nickname)
+		{
+			if (!onlineMembers.Contains(nickname)) return Guid.Empty;
+
+			onlineMembers.Remove(nickname);
+			await stream.OnNextAsync(new ChatMsg("System", $"{nickname} leaves the chat..."));
+
+			return stream.Guid;
+		}
+
+		public async Task<bool> Message(ChatMsg msg)
+		{
+			messages.Add(msg);
+			await stream.OnNextAsync(msg);
+
+			return true;
+		}
+
+	    public Task<string[]> GetMembers()
+	    {
+	        return Task.FromResult(onlineMembers.ToArray());
+	    }
+
+	    public Task<ChatMsg[]> ReadHistory(int numberOfMessages)
+	    {
+	        var response = messages
+	            .OrderByDescending(x => x.Created)
+	            .Take(numberOfMessages)
+	            .OrderBy(x => x.Created)
+	            .ToArray();
+
+	        return Task.FromResult(response);
+	    }
+    }
+}
--- a/Samples/2.0/ChatRoom/OrleansClient/Program.cs
+++ b/Samples/2.0/ChatRoom/OrleansClient/Program.cs
@@ -53,6 +53,8 @@ namespace OrleansClient
                         .UseLocalhostClustering()
                         .ConfigureApplicationParts(parts => parts.AddApplicationPart(typeof(IChannel).Assembly).WithReferences())
 				        .ConfigureLogging(logging => logging.AddConsole())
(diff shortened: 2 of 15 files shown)


Samples/2.0/ChatRoom/GrainImplementation/Channel.cs
```diff
@@ -0,0 +1,72 @@
...
+		{
+			var streamProvider = GetStreamProvider(Constants.ChatRoomStreamProvider);
+
+		    stream = streamProvider.GetStream<ChatMsg>(Guid.NewGuid(), Constants.CharRoomStreamNameSpace);
+            return base.OnActivateAsync();
+		}
+
+		public async Task<Guid> Join(string nickname)
+		{
+			if (onlineMembers.Contains(nickname)) return Guid.Empty;
```
sergeybykov: It seems confusing to me to return a `Guid.Empty` here. I think it's fine for a second client for the same nickname to join the channel.

Samples/2.0/ChatRoom/OrleansClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Streams;
+using Utils;
+
+namespace OrleansClient
+{
+	class Program
+	{
+        //To make this sample simple
+        //In this sample, one client can only join one channel, hence we have a static variable of one channel name.
```
sergeybykov: We can as easily allow multiple clients to join the same channel.

Samples/2.0/ChatRoom/OrleansClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+			return null;
+		}
+
+		private static void PrintHints()
+		{
+			var menuColor = ConsoleColor.Magenta;
+			PrettyConsole.Line("Type '/j <channel>' to join specific channel", menuColor);
+		    PrettyConsole.Line("Type '/n <username>' to set your user name", menuColor);
```
sergeybykov: indentation issues.

Samples/2.0/ChatRoom/OrleansClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+			{
+				input = Console.ReadLine();
+
+				if (string.IsNullOrWhiteSpace(input)) continue;
+
+				if (input.StartsWith("/j"))
+				{
+					await JoinChannel(client, input.Replace("/j", "").Trim());
+				}
+                else if (input.StartsWith("/n"))
```
sergeybykov: Indentation.

Samples/2.0/ChatRoom/OrleansClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+				{
+				    userName = input.Replace("/n", "").Trim();
+				    PrettyConsole.Line($"Your user name is set to be {userName}", ConsoleColor.DarkGreen);
+                }
```
sergeybykov: Indentation.

Samples/2.0/ChatRoom/OrleansClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+				}
+				else if (input.StartsWith("/h"))
+				{
+				    await ShowCurrentChannelHistory(client);
+				}
+				else if (input.StartsWith("/m"))
+				{
+				    await ShowChannelMembers(client);
+				}
+                else if (!input.StartsWith("/exit"))
```
sergeybykov: Indentation.

Samples/2.0/ChatRoom/OrleansClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		private static async Task JoinChannel(IClusterClient client, string channelName)
+		{
+		    PrettyConsole.Line($"Joining to channel {channelName}");
+            joinedChannel = channelName;
+			var room = client.GetGrain<IChannel>(joinedChannel);
+		    var streamId = await room.Join(userName);
+            //empty stream id meaning client already joined this channel, so return 
+		    if (streamId == Guid.Empty) return;
```
sergeybykov: If we return the actual stream GUID, we won't have to exit here.

Samples/2.0/ChatRoom/OrleansClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+            var stream = client.GetStreamProvider(Constants.ChatRoomStreamProvider)
+		        .GetStream<ChatMsg>(streamId, Constants.CharRoomStreamNameSpace);
+            //sunscribe to the stream to receiver furthur messages sent to the chatroom
```
sergeybykov: Typo: "subscribe".

Samples/2.0/ChatRoom/Utils/Constants.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+
+namespace Utils
+{
+	public static class Constants
+	{
+		public const string ChatRoomStreamProvider = "ChatRoom";
```
sergeybykov: Indentation.

Samples/2.0/ChatRoom/OrleansServer/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+		            options.ClusterId = Constants.ClusterId;
+		            options.ServiceId = Constants.ServiceId;
+		        })
+		        .UseLocalhostClustering()
+		        .Configure<EndpointOptions>(options => options.AdvertisedIPAddress = IPAddress.Loopback)
+                .ConfigureApplicationParts(parts => parts.AddApplicationPart(typeof(Channel).Assembly).WithReferences())
+		        .ConfigureLogging(logging => logging.AddConsole())
+                //need to configure a grain storage called "PubSubStore" for using streaming with ExplicitSubscribe pubsub type
+                .AddMemoryGrainStorage("PubSubStore")
+		        .AddSimpleMessageStreamProvider(Constants.ChatRoomStreamProvider);
```
sergeybykov: Maybe we could also show here (with a commented out line) that one can easily replace `AddSimpleMessageStreamProvider` with `AddAzureQueueStreams` to get the messages go over Azure Queues instead?
What do you think?

Samples/2.0/ChatRoom/OrleansClient/Program.cs
xiazen: yes multiple client can join the same channel, which is the point of a chat room. it is just one client cannot join multiple channel at the same time. We can technically do that, but without proper UI, it would look confusing from the console. That's why I designed it this way.

Samples/2.0/ChatRoom/OrleansServer/Program.cs
xiazen: ok

xiazen: actually at a second thought, it might be confusing to do that, since it wouldn't be like "uncomment this, and the sample will just work", the project need to reference AzureQueueStreaming package, and add connection string, etc. I think I can just parse in the streaming doc link here and it will be much clearer to read the doc than doing some simple stuff here.

Samples/2.0/ChatRoom/OrleansClient/Program.cs
xiazen: yes sorry I forgtot to remove it

Issue #4391, opened 2018-04-03
Application Insights Telemetry failing to initialize


dsarfati:
I'm trying to add the Application Insights Telemtry to our 2.0 app and im getting the following exception 
```
System.InvalidOperationException: Unable to resolve service for type 'System.String' while attempting to activate 'Orleans.TelemetryConsumers.AI.AITelemetryConsumer'.
```
Looking at the code it looks like the only constructor on the `AITelemetryConsumer` takes a string, causing the exception. Simply adding a DI compatible constructor should fix the issue. I'll submit a PR if this makes sense.

xiazen:
yeah that seems like a bug in our side. `AITelemetryConsumer ` consumer should have its own options. So that user can configure the options to give a `instrumentationKey`, 

To work around this, there's two options 
-  inject a string singleton to DI as instrumentalKey, so when AITelemetryConsumer is constructed through `ActivatorUtilities.CreateInstance(serviceProvider, consumerType);`, ActivatorUtility would locate that string singleton and create AITelemetryConsumer. But this is risky since it assumes you can only have one string singleton in DI. so I'm not recommending this one. 
- make simple adjustment to AITelemetryConsumer to fix the issue, something like 
```
 public class AITelemetryConsumerOptions
    {
        public string InstrumentationKey { get; set; }
    }

public class AITelemetryConsumer : ITraceTelemetryConsumer, IEventTelemetryConsumer, IExceptionTelemetryConsumer,
        IDependencyTelemetryConsumer, IMetricTelemetryConsumer, IRequestTelemetryConsumer
    {
         public AITelemetryConsumer(IOptions<AITelemetryConsumerOptions> options)
        {
            _client = options.Value.InstrumentationKey != null
                ? new TelemetryClient(new Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration { InstrumentationKey = options.Value.InstrumentationKey })
                : new TelemetryClient();
        }
   }
```

And then you can configure it through IOptions. 
```
new SiloHostBuilder.Configure<AITelemetryConsumerOptions>(op => op.InstrumentalKey = "")
.Configure<TelemetryOptions>(op=>op.AddConsumer<AITelemetryConsumer>());
```
- write your own TelemetryConsumer for AI 

At the mean time, I will submit a pr to fix the bug, but the fix will be included in our next release. So if you don't want to wait, you can do either of the workaround.

xiazen:
submitted a PR #4392 to fix this

[...]

PR #4392, opened 2018-04-03
Fix telemetry consumer construction

xiazen:
- Delete NewRelicTelemetryConsumerOptions.cs , since NewRelicTelemetryConsumer had a zero arg constructor in 1.5, so not sure why this option is here.
- Wire AITelemetryConsumer with its options 

fix for #4391


ReubenBond:
Will users know to use `ApplicationInsightsTelemetryConsumerOptions`?

xiazen:
they will use AddApplicationInsightesTelemetryConsumer extension on silo and client builder.

PR #4395, opened 2018-04-04
Reduce allocs in Interner

ReubenBond:
This is a minor improvement, one in a string of several to come. This change reduces the allocs for each call by approximately 300 bytes and reduces RTT (in proc) by ~21us on my machine.

Allocations are like paper cuts: too many and you bleed out.

For this and the other allocation related PRs, the targets are found by profiling, looking at hot spots, and picking the low-hanging fruit.

Before:

 Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|-------:|----------:|
   Ping | 375.3 us | 166.1 us | 9.387 us | 3.4180 |  15.83 KB |

After "Reduce allocs in Interner" commit:

 Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|-------:|----------:|
   Ping | 360.5 us | 200.2 us | 11.31 us | 3.4180 |  15.66 KB |

After "Intern ActivationAddress" commit:

 Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|-------:|----------:|
   Ping | 354.4 us | 119.1 us | 6.731 us | 3.4180 |  15.54 KB |

### DIFF
--- a/src/Orleans.Core.Abstractions/IDs/ActivationAddress.cs
+++ b/src/Orleans.Core.Abstractions/IDs/ActivationAddress.cs
@@ -1,13 +1,16 @@
 using System;
+using Orleans.Core.Abstractions.Internal;
 
 namespace Orleans.Runtime
 {
     [Serializable]
     internal class ActivationAddress
     {
-        public GrainId Grain { get; private set; }
-        public ActivationId Activation { get; private set; }
-        public SiloAddress Silo { get; private set; }
+        private static readonly Interner<ValueTuple<SiloAddress, GrainId, ActivationId>, ActivationAddress> interner =
+            new Interner<(SiloAddress, GrainId, ActivationId), ActivationAddress>(128, TimeSpan.FromMinutes(1));
+        public GrainId Grain { get; }
+        public ActivationId Activation { get; }
+        public SiloAddress Silo { get; }
 
         public bool IsComplete
         {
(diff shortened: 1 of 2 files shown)


xiazen:
LGTM. but to be cautious, should we run load test to test how this play out in a distributed system?

src/Orleans.Core.Abstractions/IDs/ActivationAddress.cs
```diff
@@ -32,7 +35,7 @@ public static ActivationAddress GetAddress(SiloAddress silo, GrainId grain, Acti
             // Silo part is not mandatory
             if (grain == null) throw new ArgumentNullException("grain");
 
-            return new ActivationAddress(silo, grain, activation);
+            return interner.FindOrCreate((silo, grain, activation), key => new ActivationAddress(key.Item1, key.Item2, key.Item3));
```
xiazen: I would image this change would impact throughput more in applications which has a lot of new activations, since now `NewActivationAddress` requires a look up, instead of a allocation. Not sure whether we have a load test testing this scenario.

ReubenBond: I imagine this PR will have very little impact. We don't have any specific tests for this, but it might be most apparent in the transaction stress tests

PR #4396, opened 2018-04-04
Avoid costly Message.GetTargetHistory() on gateway reroute

ReubenBond:
Results in 1.5Kb less allocs per call and ~17us faster RTT

Before: 

 Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|-------:|----------:|
   Ping | 375.3 us | 166.1 us | 9.387 us | 3.4180 |  15.83 KB |

After:

 Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|-------:|----------:|
   Ping | 353.9 us | 6.983 us | 13.78 us | 3.4180 |  14.28 KB |

### DIFF
--- a/src/Orleans.Runtime/Silo/Silo.cs
+++ b/src/Orleans.Runtime/Silo/Silo.cs
@@ -187,7 +187,7 @@ namespace Orleans.Runtime
             // Initialize the message center
             messageCenter = Services.GetRequiredService<MessageCenter>();
             var dispatcher = this.Services.GetRequiredService<Dispatcher>();
-            messageCenter.RerouteHandler = message => dispatcher.SendMessage(message);
+            messageCenter.RerouteHandler = dispatcher.RerouteMessage;
             messageCenter.SniffIncomingMessage = runtimeClient.SniffIncomingMessage;
 
             // Now the router/directory service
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -187,7 +187,7 @@ public Silo(ILocalSiloDetails siloDetails, IServiceProvider services)
             // Initialize the message center
             messageCenter = Services.GetRequiredService<MessageCenter>();
             var dispatcher = this.Services.GetRequiredService<Dispatcher>();
-            messageCenter.RerouteHandler = dispatcher.RerouteMessage;
+            messageCenter.RerouteHandler = message => dispatcher.SendMessage(message);
```
sergeybykov: Why not `dispatcher.ResendMessageImpl(message)` as `RerouteMessage` used to do?

ReubenBond: This logic could do with a refactoring. RerouteHandler is called from RerouteMessage which is only called from GatewayAcceptor.HandleMessage when the message is not already addressed.

So they will have the same effect but this skips a bit of logic that would be skipped anyway.

I'll change it back since it's not the purpose of this PR

PR #4398, opened 2018-04-04
Use linked list for activation message queue

ReubenBond:
* Instead of a `List<T>`, use a `LinkedList<T>` for an activation's message queue.
* Responses are added to the front of the queue, Requests are added to the back of the queue.


AlgorithmsAreCool:
This looks interesting.

Is the activation message queue a hot structure?
If so, won't this will generate more garbage overtime?

sergeybykov:
>Is the activation message queue a hot structure?

Shouldn't be for most scenarios, as hot individual grains are usually a sign of an unbalanced design. But processing responses at a higher priority should help release memory held for outstanding requests sooner. I think that's what @ReubenBond is targeting here.

PR #4399, opened 2018-04-04
Remove per-call timer allocation

ReubenBond:
* Split `CallbackData` into `CallbackData` & `SharedCallbackData` to reduce the weight of each `CallbackData` and the cost associated with creating it.
* Avoid starting a .NET Timer for every grain call, instead, use a single timer in the runtime client and expire calls in succession

The timer period is set to 1 second in this PR - I figure that is a fine interval, but we could reduce it as we see fit, or make it configurable.

These callback timers were causing significant contention (due to how .NET's Timer class is implemented) and allocating ~400 bytes per call

cc @dVakulen - we talked about this back in January & Dmytro has an implementation of a timer wheel which may be suitable to use for Grain timers (which this PR doesn't address)

### DIFF
--- a/src/Orleans.Core/Runtime/CallbackData.cs
+++ b/src/Orleans.Core/Runtime/CallbackData.cs
@@ -1,111 +1,48 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
-using Orleans.Configuration;
 using Orleans.Transactions;
 
 namespace Orleans.Runtime
 {
-    /// <summary>
-    /// This interface is for use with the Orleans timers.
-    /// </summary>
-    internal interface ITimebound
+    internal class CallbackData
     {
-        /// <summary>
-        /// This method is called by the timer when the time out is reached.
-        /// </summary>
-        void OnTimeout();
-        TimeSpan RequestedTimeout();
-    }
-
-    internal class CallbackData : ITimebound, IDisposable
-    {
-        private readonly Action<Message, TaskCompletionSource<object>> callback;
-        private readonly Func<Message, bool> resendFunc;
-        private readonly Action<Message> unregister;
+        private readonly SharedCallbackData shared;
         private readonly TaskCompletionSource<object> context;
-        private readonly MessagingOptions messagingOptions;
+        private ValueStopwatch stopwatch;
 
-        private bool alreadyFired;
-        private TimeSpan timeout;
-        private SafeTimer timer;
-        private ITimeInterval timeSinceIssued;
-        private readonly ILogger logger;
-        private readonly ILogger timerLogger;
-        private readonly ApplicationRequestsStatisticsGroup requestStatistics;
+        public CallbackData(
+            SharedCallbackData shared,
+            TaskCompletionSource<object> ctx, 
+            Message msg)
+        {
+            this.shared = shared;
+            this.context = ctx;
+            this.Message = msg;
+            this.TransactionInfo = TransactionContext.GetTransactionInfo();
+            this.stopwatch = ValueStopwatch.StartNew();
+        }
+
         public ITransactionInfo TransactionInfo { get; set; }
 
         public Message Message { get; set; } // might hold metadata used by response pipeline
 
-        public CallbackData(
-            Action<Message, TaskCompletionSource<object>> callback, 
-            Func<Message, bool> resendFunc, 
-            TaskCompletionSource<object> ctx, 
-            Message msg, 
-            Action<Message> unregisterDelegate,
-            MessagingOptions messagingOptions,
-            ILogger logger,
-            ILogger timerLogger,
-            ApplicationRequestsStatisticsGroup requestStatistics)
+        public bool IsCompleted { get; private set; }
+
+        public bool IsExpired(long currentTimestamp)
         {
-            // We are never called without a callback func, but best to double check.
-            if (callback == null) throw new ArgumentNullException(nameof(callback));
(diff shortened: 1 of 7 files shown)


ReubenBond:
@dotnet-bot test this please

ReubenBond:
The failure is *kind-of* a legitimate failure: the test changes the response timeout after the silo has already started. We should remove that, in my opinion.

ReubenBond:
I didn't end up removing the existing ability to change the response timeout, but we can still do that in a later PR

ReubenBond:
Load tests look good

ReubenBond:
If anyone could review this, I'd much appreciate it - I've got another optimization queued up for when this goes in

src/Orleans.Core/Runtime/CallbackData.cs
```diff
@@ -183,37 +119,31 @@ private void OnFail(Message msg, Message error, string resendLogMessageFormat, b
...
+                    if (this.shared.Logger.IsEnabled(LogLevel.Debug)) this.shared.Logger.Debug(resendLogMessageFormat, msg.ResendCount, msg);
                     return;
                 }
 
                 alreadyFired = true;
-                DisposeTimer();
                 if (StatisticsCollector.CollectApplicationRequestsStats)
                 {
-                    timeSinceIssued.Stop();
+                    durationTimestamp = -(Stopwatch.GetTimestamp() - durationTimestamp);
```
dVakulen: Nit: Repetitive calculation (duplication of line 102) could be extracted into separate method.

src/Orleans.Core/Runtime/CallbackData.cs
```diff
@@ -1,108 +1,64 @@
...
+        {
+            return currentTimestamp - this.durationTimestamp > this.shared.ResponseTimeoutStopwatchTicks;
         }
 
-        private void TimeoutCallback(object obj)
+        public TimeSpan GetCallDuration()
         {
-            OnTimeout();
+            // A positive timestamp value indicates the start time of an ongoing call,
+            // a negative value indicates the negative total duration of a completed call.
```
dVakulen: Would it be more clear to introduce nullable `totalDuration` field and `requestStartTimestamp`, so that implicit meaning of negative values could be avoided?

src/Orleans.Core/Runtime/CallbackData.cs
ReubenBond: Yeah, that's true. I was wary of doing that because of how I'm abusing that single field to mean two things depending on the sign

ReubenBond:
Before: Timer.Change = 9% total time for my 110 second local ping pong test
image

After:
image

src/Orleans.Core/Runtime/CallbackData.cs
ReubenBond: Maybe we should abstract it into a struct instead, eg `ValueStopwatch`, what do you think?

dVakulen: I don't see another places where such struct could be used, and `ValueStopwatch` name does not fully reflect its purpose so it might be better not to add additional entity, when it can be avoided.

dVakulen: And public `startTimestamp` property. Yes, that makes sense.

ReubenBond:
I like this change with or without perf improvements, so I'm reopening it because it came up when I was trying to fix some other issue (statistics) which impact `CallbackData`.
Local benchmarks show an improvement and when I last ran load tests they showed a minor or not significant improvement.

At some point in the future we should consider extracting the CallbackData related functionality out from the runtime clients and into a more encapsulated class (similar to InvokableObjectManager in the local client PR)

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -132,6 +117,18 @@ internal void ConsumeServices(IServiceProvider services)
             this.ClientStatistics = this.ServiceProvider.GetRequiredService<ClientStatisticsManager>();
             this.serializationManager = this.ServiceProvider.GetRequiredService<SerializationManager>();
             this.messageFactory = this.ServiceProvider.GetService<MessageFactory>();
+            
+            this.sharedCallbackData = new SharedCallbackData(
+                this.TryResendMessage,
+                msg => this.UnRegisterCallback(msg.Id),
```
dVakulen: Nit: 'Unregister'. R could be lowercase.

dVakulen (approved):

src/Orleans.Core/Timers/ValueStopwatch.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public bool IsRunning => this.value > 0;
+        
+        public TimeSpan Elapsed => TimeSpan.FromTicks(this.ElapsedTicks);
+
+        public long ElapsedTicks
+        {
+            get
+            {
+                // A positive timestamp value indicates the start time of a running stopwatch,
```
dVakulen: This comment could be document comment, for it to be readable at usage site.

ReubenBond: Added comments to this and other members, thanks

ReubenBond:
Rebased. I believe this is ready for merge.

ReubenBond:
@dotnet-bot test bvt
@dotnet-bot test functional

ReubenBond:
Re-rebased 😊

Issue #4313, opened 2018-03-27
Redundant Dictionary allocation in RequestContext.Set ?


MelnikovIG:
There is a extra allocation of `Dictionary` with copying values from another `Dictionary` if it exists.
https://github.com/dotnet/orleans/blob/3f27bea31633e55dbf4cbdb96b1562021da2226b/src/Orleans.Core.Abstractions/Runtime/RequestContext.cs#L84-L89

Looks like comments outdated and article by link too. Does it make sense since  `CallContextData` based on `AsyncLocal<T>` not `CallContext` anymore? I tried remove this allocation and all still works fine. Could you investigate this question?

kutensky:
Hi @MelnikovIG. Approach proposed by you couldn't be done. Even if we use AsyncLocal, Value that is stored is a reference type. So when we enter a new thread and AsyncLocal copies Value of instance from previous thread, it copies a link. So when we edit dictionary, we edit the same dictionary that was used in previous thread. That's why need to reallocate dictionary each time we call Set.
It would be great if we could understand that this call of Set method is the first per current thread, but AsyncLocal class doesn't provide such feature. So allocating a dictionary each time we call Set is the only way to make sure that it doesn't depend on dictionaries from other threads.

ReubenBond:
Fixed by @kutensky in #4435

PR #4435, opened 2018-04-09
Improve Dictionary allocation

kutensky:
https://github.com/dotnet/orleans/issues/4313 
Approach proposed in the ticket couldn't be done. Even if we use AsyncLocal, Value that is stored is a reference type, so when we enter a new thread and AsyncLocal copies Value of instance from previous thread, it copies a link. So when we edit dictionary, we edit the same dictionary that was used in previous thread. That's why need to reallocate dictionary each time we call Set.
It would be great if we could understand that this call of Set method is the first per current thread, but AsyncLocal class doesn't provide such feature. So allocating a dictionary each time we call Set is the only way to make sure that it doesn't depend on dictionaries from other threads. 

But I found similar case in CoreClr repository. They also reallocate dictionary inside AsyncLocal too often. So they added minor improvement that should decrease size of memory allocation.
dotnet/coreclr@9bc06f7
I added the similar improvement to RequestContext.cs.

### DIFF
--- a/src/Orleans.Core.Abstractions/Runtime/RequestContext.cs
+++ b/src/Orleans.Core.Abstractions/Runtime/RequestContext.cs
@@ -76,18 +76,18 @@ namespace Orleans.Runtime
         {
             var values = CallContextData.Value;
 
-            if (values == null)
+            if (values != null)
             {
-                values = new Dictionary<string, object>();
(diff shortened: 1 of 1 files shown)


ReubenBond:
Thank you, @kutensky, and thank you for jumping in on this

PR #4438, opened 2018-04-10
Clarify UseDevelopmentClustering and UseLocalhostClustering

### LINKED ISSUE #4421: Local and Development clustering?
`SiloHostBuilder` has the following two methods:

image

I'm not sure if these are documented anywhere, but the documentation we get in VS doesn't tell much about what they actually do:

* `UseLocalhostClustering()` says: "Configures a localhost silo for development and testing".
* `UseDevelopmentClustering()` says: "Configures the silo to use development-only clustering".

What is the difference between localhost and development clustering? What is involved in the "development and testing" configuration? What does "development-only" clustering mean?

The method docs should be improved to explain what they actually do, and why there are two different methods for what seems to be the same thing. This should also be added to the online documentation if it is not already there.

ReubenBond:
Fixes #4421

### DIFF
--- a/src/Orleans.Core/Configuration/Options/ClusterOptions.cs
+++ b/src/Orleans.Core/Configuration/Options/ClusterOptions.cs
@@ -13,6 +13,11 @@ namespace Orleans.Configuration
         internal const string DevelopmentClusterId = "dev";
 
         /// <summary>
+        /// Default service id for development clusters.
+        /// </summary>
+        internal const string DevelopmentServiceId = "dev";
+
+        /// <summary>
         /// Gets or sets the cluster identity. This used to be called DeploymentId before Orleans 2.0 name.
         /// </summary>
         public string ClusterId { get; set; }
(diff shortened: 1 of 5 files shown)


ReubenBond:
I removed ClusterId & ServiceId parameters from the extension methods.

That means that there is a single way to specify ClusterId/ServiceId (ignoring legacy config). 
The values will still default to "dev" when using `UseLocalhostClustering`, but only if no value was provided.

In my opinion, this is cleaner. It means that users will have to manually specify the value when using `UseDevelopmentClustering`, but this seems alright to me.

`UseDevelopmentClustering` on the silo is paired with `UseStaticClustering` on the client. This is because there are other cases where the client might access the cluster using a well-known set of gateways. A static gateway list is fine for production use on the client. On the other hand, grain based membership is not fault tolerant and so it's usually not considered suitable for production.
`UseLocalhostClustering` on the silo is paired with `UseLocalhostClustering` on the client.

ReubenBond:
@benjaminpetit we discussed this for about 40 mins the other day. I thought we had decided to add clusterId + serviceId parameters to `UseLocalhostClustering` only, however when I went to implement it I found that it didn't feel so good. I think we should take this as-is and instead add a helper for configuring those two in a separate PR. wdyt?

benjaminpetit:
@ReubenBond let's discuss this live when you are available

ReubenBond:
Ok, how does this look, @benjaminpetit?

Issue #4445, opened 2018-04-10
Fine-grain control of package versions


sergeybykov:
Up until 2.0, we always published a new version of every package to nuget.org, even for a patch release. This was forcing users to upgrade all packages even what it wasn't actually necessary.

In 2.0 we introduced the abstraction packages that are meant to be much more, and change very infrequently, so that the majority of projects in an application solution wouldn't need to be updated when changes in a new version of Orleans are limited to the runtime packages or providers. Similarly, most of the provider packages don't need to be updated for fixes in the runtime.

The project structure is in place and the types have been placed to the right projects and packages. What we are missing is an ability to build those projects and packages independently, so that just one of a handful of packages can be built for a patch release or even a minor release if needed.

The challenge here is that for such a level of build flexibility an Orleans project needs to reference its other Orleans projects as either a project (via a relative path in the file system) or a NuGet package of a previously published version of it.

I created a proposal in a form of a PR that uses a set of environment variable to control what gets packaged and how dependencies are referenced.

`FullBuild` variable turn on and off this mode of selective packaging. When `FullBuild=true` (default), everything gets built and packaged as today. This is how we would continue to run functional and load tests, release major versions (3.0.0) and potentially minor ones (2.1.0), too.

When `FullBuild` is not set to `true`, a set of build variables defined in the top level `Directory.Build.props` file drive how invidual dependency are referenced and what packages get built.

galvesribeiro:
Finally! :)
Once that is merged we should have some not at the Announcements repo and/or the blog so people will know how the new releases are going to be moving forward.

sergeybykov:
Resolved via #4446, #4485, #4541, #4559, #4567.

PR #4446, opened 2018-04-10
Proposal for fine-grain control of package versions

sergeybykov:
This is a proposal for #4445.

https://github.com/dotnet/orleans/commit/4d9e8de210f8542a5b8533dd0f74b98be2825b78 defines the build variables.

https://github.com/dotnet/orleans/commit/0689e7ff27cb613eb257cae4325f84736795f7fd moves intra-solution project references from .csproj files into per-directory `Directory
.Build.props` files and conditions them to either point to a .csproj or to a NuGet package.

https://github.com/dotnet/orleans/commit/1c8c1574479b0b9a9a71a2142c87004dd0d97f42 sets the target version for `Core` and `Runtime` to 2.1.0 for the next minor release.

### DIFF
--- a/src/AWS/Orleans.Transactions.DynamoDB/Directory.Build.props
+++ b/src/AWS/Orleans.Transactions.DynamoDB/Directory.Build.props
@@ -1,0 +1,35 @@
+<Project>
+  <PropertyGroup>
+    <_ParentDirectoryBuildPropsPath Condition="'$(_DirectoryBuildPropsFile)' != ''">$([System.IO.Path]::Combine('..', '$(_DirectoryBuildPropsFile)'))</_ParentDirectoryBuildPropsPath>
+  </PropertyGroup>
+
+  <Import Project="$(_ParentDirectoryBuildPropsPath)" Condition="Exists('$(_ParentDirectoryBuildPropsPath)')"/>
+
+  <PropertyGroup Condition="$(OrleansTransactionsVersion)!=$(VersionPrefix)">
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <!-- Configure source link: https://github.com/ctaggart/SourceLink -->
+  <PropertyGroup>
+    <SourceLinkCreate>true</SourceLinkCreate>
+    <SourceLinkOriginUrl>https://github.com/dotnet/orleans</SourceLinkOriginUrl>
+  </PropertyGroup>
+
+  <Choose>
+    <When Condition="$(OrleansTransactionsVersion) == $(VersionPrefix)">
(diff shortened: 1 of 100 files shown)


xiazen:
what would happen say, Azure.Persistence depends on Orleans.Runtime 2.1 minimum, and AdoNet.Remider depends on Orleans.Runtime 2.0 minimum. One application depends on both Azure.Persistence, and AdoNet.Reminder, and 2.0 and 2.1 has some small breaking changes. What version of Orleans.Runtime will/should the app pick up?

sergeybykov:
>What version of Orleans.Runtime will/should the app pick up?

I think that's no different from any other dependencies. If X depends on 2.0, and Y depends on 2.1, a project that references both X and Y will automatically depend on 2.1.

sergeybykov:
@dotnet-bot test functional
Reason: seemingly unrelated failure of ServiceBus.Tests.MonitorTests.EHStatisticMonitorTests.EHStatistics_MonitorCalledAccordingly.

src/AWS/Orleans.Transactions.DynamoDB/Directory.Build.props
```diff
@@ -0,0 +1,35 @@
...
+  <!-- Configure source link: https://github.com/ctaggart/SourceLink -->
+  <PropertyGroup>
+    <SourceLinkCreate>true</SourceLinkCreate>
+    <SourceLinkOriginUrl>https://github.com/dotnet/orleans</SourceLinkOriginUrl>
+  </PropertyGroup>
+
+  <Choose>
+    <When Condition="$(OrleansTransactionsVersion) == $(VersionPrefix)">
+      <ItemGroup >
+        <ProjectReference Include="..\..\Orleans.Transactions\Orleans.Transactions.csproj" />
```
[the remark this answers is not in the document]
sergeybykov: Good point. I can do a pass later and replace all such relative paths that I inherited from VS.

src/Azure/Directory.Build.props
```diff
@@ -0,0 +1,48 @@
...
+  </PropertyGroup>
+
+  <Import Project="$(_ParentDirectoryBuildPropsPath)" Condition="Exists('$(_ParentDirectoryBuildPropsPath)')"/>
+
+  <PropertyGroup Condition="$(OrleansAzureVersion)!=$(VersionPrefix)">
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
ReubenBond: We could keep these in the src-level props file instead, since all projects under src need documentation. Same goes for the SoruceLink properties

sergeybykov: We already have them in the `src\Directory.Build.props` file. So we can simply remove them from all other .props files under src\.

ReubenBond:
The comments can be addressed in a later PR

Issue #4351, opened 2018-03-29
Ado.net storage provider try to insert null for non nullable column


AdamWyzgol:
After upgrade to 2.0 I got this error when using ado.net storage: 
` Exc level 0: System.Data.SqlClient.SqlException: Cannot insert the value NULL into column 'ServiceId', table 'Orleans.dbo.Storage'; column does not allow nulls. INSERT fails.`

AdamWyzgol:
probably before 2.0 in clusterOptions serviceId was set to default "00000000-0000-0000-0000-000000000000" now it's set to null, after configure serviceId to value that used to have everything is working. When serviceId is null ado.net storage is not working even when I set this column to nullable.

quaider:
serviceId's type may changes from guid to  string

veikkoeeva:
I'll page @xiazen as she did a lot of hard work on this lately and might know if there's a good place for PR to go and check the ID really is set. Very good you had yourself unblocked. 👍

sergeybykov:
Yes, this is because we changed `ServiceId` from `Guid` that implicitly defaulted to empty GUID to `string`.

Beware also of http://dotnet.github.io/orleans/Documentation/2.0/Migration1.5.html#provider-configuration.

veikkoeeva:
@sergeybykov To make a bit progress, is this behavior consistent across all providers and what would be the path of least surprise (e.g. provide a default if one isn't set or something else)?

sergeybykov:
One option is to add validation similar to #4160 to make sure service ID is configured.

ReubenBond:
We have a validator now, so this particular issue should no longer occur

PR #4450, opened 2018-04-11
Add validator for ClusterOptions

benjaminpetit:
Currently we check that `ClusterId` is set but not `ServiceId`. I added a validator to test this.


veikkoeeva:
Seems like relevant to #4351.

xiazen:

xiazen:
Orleans.Runtime.OrleansConfigurationException : Configuration for ClusterOptions is invalid. A non-empty value for ServiceId is required. See $https://aka.ms/orleans-troubleshooting for more information.

looks like a lot of our tests actually have invalid service id haha. this make the functional fail

jdom:
This is a big breaking change, as services already deployed will have to set this now, and effectively change their service's `ServiceId`. Do we have a way to opt-out of this validation and treat it like an explicitly empty service id? Also, what is the rationale for requiring a non-empty ServiceId? You might already have your cluster's storage isolated from anything else. Or you might even have no storage, reminders or stream providers that require a ServiceId... Maybe each of the providers that require ServiceId should add the validator (if it was not already added)? This way users don't have to fill in something they might not use

sergeybykov:
I don't consider this a breaking change. Service ID is required and was always required. It's implicitly used in multiple places. That requirement was masked by `Guid` defaulting to empty GUID prior to 2.0. I don't think we should provide an opt-out option or complicate things here. Service ID is easy to specify.

jdom:
Then perhaps explain 8n the exception message that if you are porting from 1.x and you were not explicitly setting the serviceid, then they might want to use `Guid.Empty.ToString()` (or whatever it is so that it matches the deployed service)

xiazen:
+1 on @jdom 's comment. if the user wasn't aware of setting serviceId before, and doesn't know the importance of keeping serviceId consistent in between deployments, it is highly possible they end up set a new serviceId, which is not `Guid.Empty.ToString()` (the old defaults) and end up losing state from previous deployments.

xiazen:
this kind of subtle breaking change should be called out clearly, though exception message and docs (we already called it out in docs). But exception message add double safty and more obvious

sergeybykov:
We called it out in http://dotnet.github.io/orleans/Documentation/Migration/Migration1.5.html#provider-configuration. Although that may not be easy to discover.

xiazen:
Orleans.Runtime.OrleansConfigurationException : Configuration for ClusterOptions is invalid. A non-empty value for ServiceId is required. See $https://aka.ms/orleans-troubleshooting for more information.

tests are failing because no valid serviceId was set for them.

benjaminpetit:
So the issue is that now `ServiceId` is mandatory, but there is no way to configure it via the legacy `ClientConfiguration`. And the static legacy `GrainClient` is initialized only with this configuration object.

Our options:

- Remove the static `GrainClient`
- Add a parameter `ServiceId` (and probably another one for `ClusterId`) to `GrainClient.Initialize`
- Add a static delegate that takes a `IClientBuilder` to allow customization of the static `GrainClient`.

The later solution is probably the best since we might have missed some other stuff in the transition to the builder pattern for the static `GrainClient`... But maybe it is time to remove this class.

jdom:
Why does the client need to know about the ServiceId? It should just care about which cluster it's connecting to, but seems like ServiceId should be an implementation detail of the the silos that clients should not be bothered with.

benjaminpetit:
Some providers that run on the client need the `ServiceId` (for example streaming)

jdom:
When it's needed on the client, must it always reflect what's configured in the service? If the answer is yes (which I assume it is), shouldn't this setting be transparently propagated from the Silo as soon as the client connects?
I would expect a similar thing should be done to the entire configuration, with the possibility for the client to override some values, but especially useful to avoid defining the EventHub connection strings on the client for example, just flow it from the Silo.
Also some knobs such as response timeout, etc could have defaults coming from the Silo instead of constant values.
I know that's a bigger change, but at least the ServiceId seems like it should have never belonged on the client, and it's very easy to have them mismatch and not know about it

ReubenBond:
@jdom should ClusterId be globally unique or scoped to ServiceId? I believe it *should* be the latter.
In that case, we ought to be setting ServiceId at configuration time, as we are.

ReubenBond:
`ServiceId + ClusterId` -> cluster membership
`ServiceId` -> anything service-wide (storage, streams)

One issue with silos flowing service id to clients is that ServiceId is a part of ClusterOptions, which means that it must be known ahead of time, otherwise having options flow like that (at first connection time) messes with a core assumption we have about options.

ReubenBond:
If people decide otherwise, I won't push back too hard, but that's my reasoning anyway

benjaminpetit:
So what I did:

- I added the ServiceId as a Guid in the legacy client configuration
- I added it too in the XML parsing logic.

Why a Guid? To replicated what is in the Silo legacy configuration. I didn't ant to have a string in one and a guid in another. If we want to change both, let's do it in another PR

Issue #4442, opened 2018-04-10
StructureMap and Orleans


misha-fm:
Hi! I try to integrate the StructureMap in Orleans (v 2.0.0).

I use the following code:

```cs
var builder = new SiloHostBuilder()
    .UseServiceProviderFactory(new StructureMapServiceProviderFactory(new MyRegistry());
```

`StructureMapServiceProviderFactory` implements a method:
```cs
public Registry CreateBuilder(IServiceCollection services)
{
    var registry = Registry ?? new Registry();
    registry.Populate(services);
    return registry;
}
```

But call to `registry.Populate(services)` throws an exception:
"Specified argument was out of the range of valid values.
Parameter name: Orleans.Runtime.StatelessWorkerPlacement must have at least one public constructor to be plugged in by StructureMap".

So, what is the correct way to inject StructureMap into Orleans DI framework?

xiazen:
which structure map nugget are you using?

I found this https://github.com/structuremap/StructureMap.Microsoft.DependencyInjection on the internet show me install `StructureMap.Microsoft.DependencyInjection` and then it has a sample showing how to plug StructureMap into ASP.NET DI abstraction, which is the abstraction Orleans integrated with. This seems to be the suggested way if user want to configure third party DI container  to use with ASP.NET DI abstractions. is this what you are using? your code doesn't look quite like the sample .

misha-fm:
Hi, @xiazen! Thank you for your reply!
Yes, I'm using that Nuget (StructureMap.Microsoft.DependencyInjection), and the code for method `CreateBuilder` I took from there. According to https://github.com/aspnet/DependencyInjection/issues/442, all implementations are populating Orleans services into the container: `registry.Populate(services);`. But as I wrote in my first post, it throws the exception saying that `Orleans.Runtime.StatelessWorkerPlacement` must have a public constructor in order to be plugged into the StructureMap container.

Tried to use Autofac in a similar manner: 
```cs
var autofacSpFactory = new AutofacServiceProviderFactory();
var builder = new SiloHostBuilder()
	.UseLocalhostClustering()
	.UseServiceProviderFactory(autofacSpFactory);
```

It throws the similar exception: "No accessible constructors were found for the type 'Orleans.Runtime.StatelessWorkerPlacement'."

StackTrace:
```cs
   at Autofac.Core.Activators.Reflection.DefaultConstructorFinder.GetDefaultPublicConstructors(Type type)
[...]

PR #4453, opened 2018-04-12
fix orleans integration with third party DI solution which requires public constructor

xiazen:
Third party DI solution, such as StructureMap, Autofac, requires services injected into DI have correct public constructor, which the runtime will be using to constructing them. We found a single instance of `StatelessWorkerPlacement`, which is injected into DI and doesn't have a public constructor. This prevent user to integrate Orleans with those third party DI solution.

Since there's no need for `StatelessWorkerPlacement` constructor to be internal, the class is already internal. So make the constructor public.

Also picked one of our test to use `StructureMap` as its DI solution, to test Orleans can integrate with third party DI solution, like `StructureMap`. Let me know if you want me to separate it into its own test. 

Fix for #4442

### DIFF
--- a/src/Orleans.Core.Abstractions/Placement/StatelessWorkerPlacement.cs
+++ b/src/Orleans.Core.Abstractions/Placement/StatelessWorkerPlacement.cs
@@ -9,7 +9,7 @@ namespace Orleans.Runtime
 
         public int MaxLocal { get; private set; }
 
-        internal StatelessWorkerPlacement(int maxLocal = -1)
+        public StatelessWorkerPlacement(int maxLocal = -1)
         {
             // If maxLocal was not specified on the StatelessWorkerAttribute, 
             // we will use the defaultMaxStatelessWorkers, which is System.Environment.ProcessorCount.
--- a/test/ServiceBus.Tests/ServiceBus.Tests.csproj
+++ b/test/ServiceBus.Tests/ServiceBus.Tests.csproj
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="$(FluentAssertionsVersion)" />
+    <PackageReference Include="StructureMap.Microsoft.DependencyInjection" Version="1.4.0" />
     <PackageReference Include="xunit" Version="$(xUnitVersion)" />
     <PackageReference Include="Xunit.SkippableFact" Version="$(XunitSkippableFactVersion)" />
     <DotNetCliToolReference Include="dotnet-xunit" Version="$(xUnitVersion)" />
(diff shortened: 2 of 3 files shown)


src/Orleans.Core.Abstractions/Placement/StatelessWorkerPlacement.cs
```diff
@@ -9,7 +9,7 @@ internal class StatelessWorkerPlacement : PlacementStrategy
 
         public int MaxLocal { get; private set; }
 
-        internal StatelessWorkerPlacement(int maxLocal = -1)
+        public StatelessWorkerPlacement(int maxLocal = -1)
```
ReubenBond: Maybe we should add a default constructor instead and make this argument non-optional on this ctor

test/ServiceBus.Tests/ServiceBus.Tests.csproj
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="$(FluentAssertionsVersion)" />
+    <PackageReference Include="StructureMap.Microsoft.DependencyInjection" Version="1.4.0" />
```
jdom: I think I would add an explicit test project that tests integration with several containers, using theories or base classes like we use to test different providers... Converting just one test in this project seems very random

xiazen: will do.

ReubenBond:
Need to also include the DI test project in the test command so that VSO and Jenkins run it

test/DependencyInjection.Tests/AspnetDiContainer/DependencyInjectionGrainTestsUsingASPNETDIContainer.cs
```diff
@@ -0,0 +1,34 @@
...
+                builder.Options.InitialSilosCount = 1;
+                builder.AddSiloBuilderConfigurator<TestSiloBuilderConfigurator>();
+                //Orleans would use ASP.NET DI container solution by default, so no need to configure ServiceProviderFactory here
+            }
+
+
+        }
+
+        public DependencyInjectionGrainTestsUsingASPNETDIContainer()
+            : base(new Fixture())
```
jdom: This will create a new cluster per test. Please follow the standard pattern to use class fixtures

test/DependencyInjection.Tests/AspnetDiContainer/DependencyInjectionGrainTestsUsingASPNETDIContainer.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Hosting;
+using Orleans.TestingHost;
+using TestExtensions;
+using UnitTests.Grains;
+
+namespace DependencyInjection.Tests.AspnetDIContainer
+{
+    [TestCategory("DI"), TestCategory("BVT")]
```
jdom: I'd make these Functional instead, as they'll slow down the BVTs... But feel free to push back.
Also, I believe there is already another category for DI tests... maybe it was DependencyInjection? Don't remember, and I might be mistaken

xiazen:
Hi someone merge this ~~ it is approved and checks passed

PR #4455, opened 2018-04-12
Azure blob storage provider: respect UseJson setting

### LINKED ISSUE #4439: Azure blob storage provider ignores UseJson option
Currently the provider does not access the UseJson property and will always use JSON for serialization.

ReubenBond:
Fixes #4439 

Submitted for testing

### DIFF
--- a/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureBlobStorage.cs
+++ b/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureBlobStorage.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics;
+using System.IO;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureBlobStorage.cs
```diff
@@ -63,11 +64,14 @@ public class AzureBlobGrainStorage : IGrainStorage, ILifecycleParticipant<ISiloL
...
-                string json;
-
+                byte[] contents;
                 try
                 {
-                    json = await blob.DownloadTextAsync().ConfigureAwait(false);
+                    using (var stream = new MemoryStream())
+                    {
+                        await blob.DownloadToStreamAsync(stream).ConfigureAwait(false);
+                        contents = stream.ToArray();
```
ReubenBond: This is not efficient, but it's the same as what's being done under the hood today, same goes for the JSON serialization

PR #4461, opened 2018-04-12
Move test grain interfaces and classes to Grains subdirectory

sergeybykov:
This is on top of #4460 to consolidate test grain interfaces and classes.


xiazen:
Tester.HeterogeneousSilosTests. are failing

Issue #4468, opened 2018-04-13
UseDevelopmentClustering and UseLocalhostClustering helper methods do not configure ServiceId


benjaminpetit:
UseDevelopmentClustering and UseLocalhostClustering helper methods configure a default ClusterId, but no ServiceId. This is an issue since ServiceIs is required too.

These methods should take an optional parameter (with a default value) for setting the ServiceId

sergeybykov:
Resolved via #4438.

PR #4469, opened 2018-04-13
Add serviceId parameters to ClientBuilder/ServiceHostBuilder helper methods

benjaminpetit:
Fix for #4468 

With this PR, test failures from PR #4450 should be gone.

I added check to not override ClusterId/ServiceId set from this methods if they were set explicitely by a call to `Configure<ClusterOptions>` before calling them.


ReubenBond:
Take a look at https://github.com/dotnet/orleans/pull/4438

PR #4476, opened 2018-04-16
changed framework to netcoreapp2.0

mxplusb:
The example uses .NET Framework 4.6.2. Changing it makes it easier for
others to get started; updated readme to reflect changes.


galvesribeiro:
Sorry but I'm confuse... I think the grain interface and the implementation projects still class libraries so it should be `netstandard2.0`  not `netcoreapp2.0`. The only ones to target `netcoreapp2.0` should be silo and client projects which are console apps... Am I missing something here?

benjaminpetit:
Yes you are right netstandard2.0 should be enough for grains class/interfaces. I may have merged this too fast.

PR #4481, opened 2018-04-17
Increase diagnostic logging in code generator

ReubenBond:
Adds log lines to explain why a serializer is generated for a particular type.
Only enabled when logging is set to `Trace`, eg via the csproj property `<OrleansCodeGenLogLevel>Trace</OrleansCodeGenLogLevel>`


ReubenBond:
@dotnet-bot test functional

Issue #4061, opened 2018-02-21
Update Samples and docs to show necessity of using ObserverSubscriptionManager.IsSubscribed()


finsterdexter:
Both the Orleans docs and several samples just call ObserverSubscriptionManager.Subscribe() without handling the OrleansException that is generated if the observer that is subbed/unsubbed is already subbed/unsubbed. That seems like an important failure state to handle, especially since the current docs actually suggest re-subbing as a means of improving reliability of this feature.

sergeybykov:
@finsterdexter Valid point. Thank you. Would you be willing to contribute an update to the docs and samples?

finsterdexter:
Yeah possibly. Gotta see if I can fit it into my spare time. I mainly wanted to float the issue before tackling it.

finsterdexter:
Yikes, I didn't realize all my commit --amends would end up on here... sorry about the mess!

![20161119151919 sorry_about_the_mess](https://user-images.githubusercontent.com/98887/38998653-a88416e8-43b5-11e8-9af1-eb9332f0e48d.gif)

xiazen:
thanks for pointing this out for us. Your PR looks good. I merged it.

xiazen:
close issue, since issue resolved through #4489

PR #4489, opened 2018-04-18
Update Observers doc page for handling case of already subbed/unsubbed (#4061)

finsterdexter:
This is to update the Observer doc to show the necessity of mitigating for the OrleansException that is thrown if you try to subscribe/unsubscribe an observer that is already added to/removed from the observer manager. #4061


xiazen:
Thank you so much !

PR #4490, opened 2018-04-18
Update samples to match docs for Observers (#4061)

finsterdexter:
This updates a couple of Samples to match the `IsSubscribed()` calls added to the docs in PR #4489. There wasn't a good Sample under 2.0, so maybe we still want to add a good Observer sample for 2.0, as I held off on updating too much of this 1.5 sample (such as updating all of the deprecated `TaskDone.Done` calls to `Task.CompletedTask`, which might make sense as part of a separate issue).


sergeybykov:
Thank you, @finsterdexter!

PR #4496, opened 2018-04-20
Golden path multi-resource transactional grain testing

xiazen:
Add golden path tests for transactional grains which consist of 2 and TransactionTestConstants.MaxCoordinatedTransactions transactional resources.


test/Transactions/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
```diff
@@ -14,30 +15,40 @@ protected GoldenPathTransactionTestRunner(IGrainFactory grainFactory, ITestOutpu
...
         public virtual async Task SingleGrainWriteTransaction(string transactionTestGrainClassName)
         {
             const int delta = 5;
             ITransactionTestGrain grain = RandomTestGrain(transactionTestGrainClassName);
-            int original = await grain.Get();
+            var original = await grain.Get();
             await grain.Add(delta);
-            int expected = original + delta;
-            int actual = await grain.Get();
+            var expected = original.Select(value => value += delta).ToArray();
```
ReubenBond: += is a little odd here, since changes to `value` will have no effect.

test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalGrains.cs
```diff
@@ -0,0 +1,144 @@
...
+        public static int Get(ITransactionalState<GrainData> data, ILogger logger)
+        {
+            logger.LogInformation($"Get {data.State.Value}.");
+            return data.State.Value;
+        }
+
+        public static int Add(ITransactionalState<GrainData> data, int numberToAdd, ILogger logger)
+        {
+            logger.LogInformation($"Adding {numberToAdd} to {data.State.Value}.");
+            data.State.Value += numberToAdd;
```
ReubenBond: In general, we should use a non-commutative operation (eg subtraction) for transaction tests in order to help ensure that the order of application is correct.

It doesn't matter so much here since this isn't a correctness test (or are they all correctness tests?)

test/Transactions/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
xiazen: yeah it should be just +

test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalGrains.cs
xiazen: why is this ? did I fail my highschool math test again?

ReubenBond: It's fine the way it is. It's just that we want to verify some of the properties of transactions (ACID, etc) and one way to verify some of those properties is to verify that the order of application is consistent. One way to verify order is to use a non-commutative operation on the value here.

LGTM, just want to give Jason a chance to look at it.

xiazen: I know it looks fine now since we don't have correctness tests. Was just asking for more info on non-commutative operation and how it verify the order of application. I can totally search it, but I guess it is faster to ask you,.

xiazen: and @jason-bragg can you take a look of this pr ?

test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalGrains.cs
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        public Task Deactivate()
+        {
+            DeactivateOnIdle();
+            return Task.CompletedTask;
+        }
+    }
+
+    public static class TransactionalGrainUtils
```
jason-bragg: Unclear why we need a utility class.  It doesn't seem to buy us anything and introduces ambiguity, as 'utility' is very unclear as to what it's supposed to do.
In general I discourage the use of 'utility' or 'helper' classes because their ambiguous nature allows them to become catchalls for random functions, harming clarity and maintainability.

test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalGrains.cs
jason-bragg: We'll likely need to adjust the operation logic when we add correctness testing, but accounting for this is, imo, out of scope for this PR.

Side note: non-commutative is insufficient, as it, to my recollection, concerns itself with the order of the datum within an operation, not the order of operations.  We need something that detects missing or out of order operations.   That is 100 - 1 - 2 == 100 - 2 - 1,  which doesn't help us.

test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalGrains.cs
```diff
@@ -0,0 +1,144 @@
...
+
+namespace Orleans.Transactions.Tests
+{
+    [Serializable]
+    public class GrainData
+    {
+        public int Value { get; set; }
+    }
+
+    public class EightStateTransactionalGrain : MultiStateTransactionalGrainBaseClass
```
jason-bragg: Goal was to support TransactionTestConstants.MaxCoordinatedTransactions, which is currently 8 but may be redefined in the future.
This should be possible by injecting the Transactional State factory into the constructor rather than the specific state objects, and creating as many as is needed.

test/Transactions/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
```diff
@@ -14,30 +15,40 @@ protected GoldenPathTransactionTestRunner(IGrainFactory grainFactory, ITestOutpu
...
+            var actual = await grain.Get();
             Assert.Equal(expected, actual);
         }
 
         [SkippableTheory]
         [InlineData(TransactionTestConstants.SingleStateTransactionalGrain)]
+        [InlineData(TransactionTestConstants.DoubleStateTransactionalGrain)]
+        [InlineData(TransactionTestConstants.EightStateTransactionalGrain)]
```
jason-bragg: It's probably ok, but it should be noted that the way these tests are written, transactions will now be tested against TransactionTestConstants.MaxCoordinatedTransactions * TransactionTestConstants.MaxCoordinatedTransactions transactional states.

test/Transactions/Orleans.Transactions.Tests/TransactionTestConstants.cs
```diff
@@ -14,5 +14,7 @@ public static class TransactionTestConstants
         // grain implementations
         public const string TransactionOrchestrationGrain = "Orleans.Transactions.Tests.TransactionOrchestrationGrain";
         public const string SingleStateTransactionalGrain = "Orleans.Transactions.Tests.SingleStateTransactionalGrain";
+        public const string DoubleStateTransactionalGrain = "Orleans.Transactions.Tests.DoubleStateTransactionalGrain";
```
jason-bragg: I understand this change is consistent with the previous pattern, but that pattern is error prone.  Suggest static readonly strings which get their value from the type, rather than constants.  The reason for this is that using the types provides compile time validation to the value.

test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalGrains.cs
xiazen: yes that was my confusion, since subtraction doesn't 100% verify the order either, so I was wondering what's other magic there I don't know of

test/Transactions/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
xiazen: I thought that's the goal?

xiazen: the task is add Golden path multi-resource transactional grain testing

test/Transactions/Orleans.Transactions.Tests/TransactionTestConstants.cs
xiazen: we cannot do it because these variable are used as `InlineData` which requires them to be const which I think might be why they are const in the first place

test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalGrains.cs
xiazen: it can be merged back to the transactionalGrainBaseClass.

test/Transactions/Orleans.Transactions.Tests/TransactionTestConstants.cs
jason-bragg: Ah, fair point.. sux, but don't see good way around.. nm

test/Transactions/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
jason-bragg: Understood, and I'm not necessarily pushing back on this approach, only pointing out that it's a bit beyond what we scoped, and that has implications.  The number of transactional resources is unbounded, but we've committed to testing the system up to 8 for now, which we've defined as TransactionTestConstants.MaxCoordinatedTransactions.  This value will likely change to 16 or 32 once the system is more stable.  The number of transactional resources is the sum of the resource across all the grain, but the original tests, built against grains with a single resource, treated the number of grains and the number of transaction resources as the same, which they are not.  Not correcting this in this pr means we're testing many more resources than we committed to.  If that works without incident, great, but if we encounter issues, we should consider scoping back these tests to the original resource goals.

PR #4502, opened 2018-04-23
Archive transaction logs

xiazen:
- Archive transaction logs
- Add minimum query API
- Add minimum testing on the feature

Since no enough information for me to know which queries will be most useful, so I didn't add any specific query API, just the general API which takes a `TableQuery`

### DIFF
--- a/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionArchiveLogOptions.cs
+++ b/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionArchiveLogOptions.cs
@@ -2,7 +2,7 @@
 using System.Collections.Generic;
 using System.Text;
 
-namespace Orleans.Transactions.AzureStorage.Storage.Development
+namespace Orleans.Configuration.Development
 {
     /// <summary>
     /// Option class to configure Azure transaction log archive behavior
--- a/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionArchiveLogOptions.cs
+++ b/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionArchiveLogOptions.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Orleans.Transactions.AzureStorage.Storage.Development
+{
+    /// <summary>
+    /// Option class to configure Azure transaction log archive behavior
+    /// </summary>
+    public class AzureTransactionArchiveLogOptions
+    {
+        public const bool DEFAULT_ARCHIVE_LOG = false;
+        //whether to archive commited transaction log or not. turned off by default
+        public bool ArchiveLog { get; set; } = DEFAULT_ARCHIVE_LOG;
+    }
+}
--- a/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogOptions.cs
+++ b/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogOptions.cs
@@ -10,6 +10,9 @@ namespace Orleans.Configuration
         public string ConnectionString { get; set; }
 
         public string TableName { get; set; } = "TransactionLog";
+
+        //whether to archive commited transaction log or not. turned off by default
+        public bool ArchiveLog = false;
     }
 
     public class AzureTransactionLogOptionsValidator : IConfigurationValidator
--- a/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogOptions.cs
+++ b/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogOptions.cs
@@ -10,9 +10,6 @@ namespace Orleans.Configuration
         public string ConnectionString { get; set; }
 
         public string TableName { get; set; } = "TransactionLog";
-
-        //whether to archive commited transaction log or not. turned off by default
-        public bool ArchiveLog = false;
     }
 
     public class AzureTransactionLogOptionsValidator : IConfigurationValidator
--- a/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
+++ b/src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
@@ -10,7 +10,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Orleans.Configuration;
 using Orleans.Serialization;
 using Orleans.Transactions.Abstractions;
-using DateTime = System.DateTime;
+using Orleans.Transactions.AzureStorage.Storage.Development;
 
 namespace Orleans.Transactions.AzureStorage
 {
@@ -24,7 +24,6 @@ namespace Orleans.Transactions.AzureStorage
 
         private const int BatchOperationLimit = 100;
         private const int CommitRecordsPerRow = 10;
-        private const string CommitRecordPartitionKey = "0";
 
         private const string StartRowPartitionKey = "1";
         private const string StartRowRowKey = "0";
@@ -38,6 +37,7 @@ namespace Orleans.Transactions.AzureStorage
 
         private long startRecordValue;
         private long nextLogSequenceNumber;
+        private readonly string CommitRecordPartitionKey;
 
         // Log iteration indexes, reused between operations
         private TableContinuationToken currentContinuationToken;
@@ -45,12 +45,16 @@ namespace Orleans.Transactions.AzureStorage
         private int currentQueryResultIndex;
         private List<CommitRecord> currentRowTransactions;
         private int currentRowTransactionsIndex;
-        private ClusterOptions clusterOptions;
-        public AzureTransactionLogStorage(SerializationManager serializationManager, IOptions<AzureTransactionLogOptions> configurationOptions, IOptions<ClusterOptions> clusterOptions)
+        private readonly ClusterOptions clusterOptions;
+        private readonly AzureTransactionArchiveLogOptions archiveLogOptions;
+        public AzureTransactionLogStorage(SerializationManager serializationManager, IOptions<AzureTransactionLogOptions> configurationOptions, 
+            IOptions<AzureTransactionArchiveLogOptions> archiveOptions, IOptions<ClusterOptions> clusterOptions)
         {
             this.serializationManager = serializationManager;
             this.options = configurationOptions.Value;
             this.clusterOptions = clusterOptions.Value;
+            this.archiveLogOptions = archiveOptions.Value;
+            this.CommitRecordPartitionKey = this.clusterOptions.ServiceId;
         }
 
         public async Task Initialize()
@@ -194,7 +198,7 @@ namespace Orleans.Transactions.AzureStorage
             {
                 var recordCount = Math.Min(transactionList.Count - nextRecord, CommitRecordsPerRow);
                 var transactionSegment = transactionList.GetRange(nextRecord, recordCount);
-                var commitRow = new CommitRow(nextLogSequenceNumber);
+                var commitRow = new CommitRow(nextLogSequenceNumber, this.clusterOptions.ServiceId);
 
                 foreach (var transaction in transactionSegment)
                 {
@@ -222,7 +226,7 @@ namespace Orleans.Transactions.AzureStorage
         public async Task<List<CommitRecord>> QueryArchivalRecords(TableQuery<ArchivalRow> query)
         {
             var continuationToken = default(TableContinuationToken);
-            var deseriazliedResults = new List<CommitRecord>();
+            var deserializedResults = new List<CommitRecord>();
             do
             {
                 var queryResult = await table.ExecuteQuerySegmentedAsync(query, continuationToken).ConfigureAwait(false);
@@ -233,23 +237,22 @@ namespace Orleans.Transactions.AzureStorage
                     foreach (var row in queryResult)
                     {
                         var transactions = DeserializeCommitRecords(row.Transactions);
-                        deseriazliedResults.AddRange(transactions);
+                        deserializedResults.AddRange(transactions);
                     }
                 }
-            } while (continuationToken != null);
+            } while (continuationToken != default(TableContinuationToken));
 
-            return deseriazliedResults;
+            return deserializedResults;
         }
 
         public async Task TruncateLog(long lsn)
         {
             var continuationToken = default(TableContinuationToken);
-            var truncatingTime = DateTime.UtcNow;
             var query = new TableQuery<CommitRow>().Where(TableQuery.CombineFilters(
                 TableQuery.GenerateFilterCondition(PartitionKey, QueryComparisons.Equal, CommitRecordPartitionKey),
                 TableOperators.And,
                 TableQuery.GenerateFilterCondition(RowKey, QueryComparisons.LessThanOrEqual, CommitRow.ToRowKey(lsn))));
-
+            var batchOperation = new TableBatchOperation();
             do
             {
                 var queryResult = await table.ExecuteQuerySegmentedAsync(query, continuationToken).ConfigureAwait(false);
@@ -258,8 +261,7 @@ namespace Orleans.Transactions.AzureStorage
 
                 if (queryResult.Results.Count > 0)
                 {
-                    var batchOperation = new TableBatchOperation();
-                    var archivalBatchOperation = new TableBatchOperation();
+
                     foreach (var row in queryResult)
                     {
                         var transactions = DeserializeCommitRecords(row.Transactions);
@@ -267,21 +269,15 @@ namespace Orleans.Transactions.AzureStorage
                         if (transactions.Count > 0 && transactions[transactions.Count - 1].LSN <= lsn)
                         {
                             batchOperation.Delete(row);
-                            if (this.options.ArchiveLog)
+                            if (this.archiveLogOptions.ArchiveLog)
                             {
-                                var archiveRow = new ArchivalRow(row.Transactions, transactions.Select(tx => tx.TransactionId).Min(), transactions.Select(tx => tx.LSN).Min(), this.clusterOptions.ClusterId, truncatingTime);
-                                archivalBatchOperation.Insert(archiveRow);
+                                var archiveRow = new ArchivalRow(this.clusterOptions, row.Transactions, transactions.Select(tx => tx.TransactionId).Min(), transactions.Select(tx => tx.LSN).Min());
+                                batchOperation.Insert(archiveRow);
                             }
 
                             if (batchOperation.Count == BatchOperationLimit)
                             {
                                 await table.ExecuteBatchAsync(batchOperation).ConfigureAwait(false);
-                                if (this.options.ArchiveLog)
-                                {
-                                    await table.ExecuteBatchAsync(archivalBatchOperation).ConfigureAwait(false);
-                                    archivalBatchOperation = new TableBatchOperation();
-                                }
-                                
                                 batchOperation = new TableBatchOperation();
                             }
                         }
@@ -292,7 +288,12 @@ namespace Orleans.Transactions.AzureStorage
                     }
                 }
 
-            } while (continuationToken != null);
+            } while (continuationToken != default(TableContinuationToken));
+
+            if (batchOperation.Count > 0)
+            {
+                await table.ExecuteBatchAsync(batchOperation).ConfigureAwait(false);
+            }
         }
 
         public static Factory<Task<ITransactionLogStorage>> Create(IServiceProvider serviceProvider)
@@ -359,23 +360,29 @@ namespace Orleans.Transactions.AzureStorage
 
         public class ArchivalRow : TableEntity
         {
+            public static string MinRowKey = MakeRowKey(long.MinValue);
+            public static string MaxRowKey = MakeRowKey(long.MaxValue);
+
+            public static string MakeRowKey(long firstTransactionId)
+            {
+                return $"arch_{CommitRow.ToRowKey(firstTransactionId)}";
+            }
+
             public ArchivalRow()
             {
             }
 
-            public ArchivalRow(byte[] transactions, long firstTransactionId, long firstLSN, string clusterId, DateTime archivalTime)
+            public ArchivalRow(ClusterOptions clusterOptions, byte[] transactions, long firstTransactionId, long firstLSN)
             {
                 this.Transactions = transactions;
-                this.ArchivalTime = archivalTime;
-                this.ClusterId = clusterId;
+                this.ClusterId = clusterOptions.ClusterId;
                 this.FirstLSN = firstLSN;
                 this.FirstTransactionId = firstTransactionId;
-                PartitionKey = $"{clusterId}${this.ArchivalTime.Ticks}";
-                RowKey = $"{this.ArchivalTime.Ticks}${firstTransactionId}";
+                PartitionKey = clusterOptions.ServiceId;
+                RowKey = MakeRowKey(firstTransactionId);
             }
 
             public string ClusterId { get; set; }
-            public DateTime ArchivalTime { get; set; }
             public long FirstLSN { get; set; }
             public long FirstTransactionId { get; set; }
             public byte[] Transactions { get; set; }
@@ -383,10 +390,10 @@ namespace Orleans.Transactions.AzureStorage
 
         private class CommitRow : TableEntity
         {
-            public CommitRow(long firstLSN)
+            public CommitRow(long firstLSN, string serviceId)
             {
                 // All entities are in the same partition for atomic read/writes.
-                PartitionKey = CommitRecordPartitionKey;
+                PartitionKey = serviceId;
                 RowKey = ToRowKey(firstLSN);
             }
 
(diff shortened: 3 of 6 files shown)


src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -234,19 +259,29 @@ public Task<long> GetStartRecord()
                 if (queryResult.Results.Count > 0)
                 {
                     var batchOperation = new TableBatchOperation();
-
+                    var archivalBatchOperation = new TableBatchOperation();
```
jason-bragg: The add and remove should be performed in the same batch operation, else they won't be atomic.
If archiving is on, we need reduce the size of the batch to half, as half of the operation will be delete, and half will be adds.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -322,6 +357,30 @@ private List<CommitRecord> DeserializeCommitRecords(byte[] serializerCommitRecor
...
+        public class ArchivalRow : TableEntity
+        {
+            public ArchivalRow()
+            {
+            }
+
+            public ArchivalRow(byte[] transactions, long firstTransactionId, long firstLSN, string clusterId, DateTime archivalTime)
+            {
+                this.Transactions = transactions;
+                this.ArchivalTime = archivalTime;
```
jason-bragg: we shouldn't need archive time as a property, as the entity should already have a timestamp on it.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -322,6 +357,30 @@ private List<CommitRecord> DeserializeCommitRecords(byte[] serializerCommitRecor
...
+                this.ClusterId = clusterId;
+                this.FirstLSN = firstLSN;
+                this.FirstTransactionId = firstTransactionId;
+                PartitionKey = $"{clusterId}${this.ArchivalTime.Ticks}";
```
jason-bragg: Probably should have delimiter between cluster id and archive time, for parsablity.
Ticks format should probably be normalized, suggest 16:x

jason-bragg: Suggest formal row key structure something like:
        private const string  RKBaseFormat "arch_{0}_";
        public static string RowKeyMin(string clusterId) => string.Format(RKBaseFormat,   clusterId);
        public static string RowKeyMax(string clusterId) => RowKeyMin(clusterId) + "~";
        public static string MakeRowKey(string clusterId, DateTime utcNow) => RowKeyMin(clusterId) +utcNow.Ticks.ToString("{0:x16}");

That way queries can be formed from the min to max row keys without knowledge or concern as to what those values are.

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            var logStorage = await StorageFactory(azureOptions);
+            var recordsNum = 2000;
+            var allTransactions = new List<CommitRecord>(recordsNum);
+            for (int i = 0; i< recordsNum; i++)
+            {
+                allTransactions.Add(new CommitRecord(){LSN=i, TransactionId = i});
+            }
+            await logStorage.Append(allTransactions);
+            //a subset of all transactions will be archived 
+            await logStorage.TruncateLog(recordsNum + 2);
```
jason-bragg: unclear why this would archive only a subset?

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            var query = new TableQuery<AzureTransactionLogStorage.ArchivalRow>().Where(
+                TableQuery.GenerateFilterCondition(nameof(AzureTransactionLogStorage.ArchivalRow.ClusterId), QueryComparisons.Equal, TestClusterId));
+            var archivalTransactions = await logStorage.QueryArchivalRecords(query);
+            //assert a subset of transaction has been archived and no duplicates
+            Assert.True(allTransactions.Select(tx=>tx.LSN).ToImmutableHashSet().IsSubsetOf(archivalTransactions.Select(tx=>tx.LSN)));
+            Assert.Equal(archivalTransactions.Select(tx => tx.LSN).Count(), archivalTransactions.Select(tx => tx.LSN).ToImmutableHashSet().Count);
```
jason-bragg: unclear what the expected outcome is.  Asserts usually have (expected, actual) with expected state being a clear an comprehensible state.  In this case it looks like we're comparing the actual count with the actual count..?

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogOptions.cs
```diff
@@ -10,6 +10,9 @@ public class AzureTransactionLogOptions
         public string ConnectionString { get; set; }
 
         public string TableName { get; set; } = "TransactionLog";
+
+        //whether to archive commited transaction log or not. turned off by default
+        public bool ArchiveLog = false;
```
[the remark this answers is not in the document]
xiazen: a property of what? mind be more specific?

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
xiazen: ok

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
xiazen: because it might now archive all . For example, if the batch operation didn't hit 100 operation count, the batch operation won't execute

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
xiazen: this line probably not needed. but we are comparing no duplicared LSN, which I commented above

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
xiazen: cannot do it since batch operation requires operation under the same partition key.

jdom: Be careful with this suggestion, because unless you rework the partition key from it's current value, then these can't go in the same batch operation

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -9,6 +10,7 @@
 using Orleans.Configuration;
 using Orleans.Serialization;
 using Orleans.Transactions.Abstractions;
+using DateTime = System.DateTime;
```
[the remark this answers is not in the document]
xiazen: I dont know what's happened there. will look into it. looks odd

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogOptions.cs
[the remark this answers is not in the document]
jason-bragg: public bool ArchiveLog {get; set;} = DEFAULT_ARCHIVE_LOG;
public const bool DEFAULT_ARCHIVE_LOG = true;

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
jason-bragg: I didn't notice the partition keys didn't match, my mistake.  We do need these to be atomic so suggest they be in the same partition.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
jason-bragg: I didn't read close enough.  Thought this was the row key.  Partition key should probably match log record partition key, and all of the above is suggested for the row key.  Also, data partition needs be by service id, rather than cluster Id.

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
jason-bragg: That sounds like a bug...

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
jason-bragg: archivalTransactions.Select(tx => tx.LSN).Count() returns number of archivalTransactions, correct?
archivalTransactions.Select(tx => tx.LSN).ToImmutableHashSet().Count()  returns number of archivalTransactions, correct?
So we're comparing the same things..?

xiazen: if a array is {1,1,2}, ToHashSet would make it {1, 2}. So their count would be different. 

it shouldn't happen with this case since LSN isn't the row key. and there won't be duplicated two row with the same row key. I can remove this line, but just they are not comparing the same thing. Or I missed what ToHashSet do ?

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
xiazen: yeah... you are right... it was a bug in the old code.. I wasn't thinking that is a bug because I assume the original code is bugless..

so log would only be truncated when the   if (batchOperation.Count == BatchOperationLimit)

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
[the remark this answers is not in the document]
xiazen: yeah you are right. will refine this. 

I was just put all the field I know which might be helpful there, and hoping the correctness tests would bring more information on what hot query we need . and then refine the partition key. 

but you are right using ArchivalTime as partition doesn't bring any query value

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
jdom: Can you explain the purpose of archiving? I thought it was so that the partition doesn't get extremely big and it affects the hot access to it, but if it's kept in the same one, maybe I missed the point.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -216,10 +219,32 @@ public Task<long> GetStartRecord()
...
+
+                if (queryResult.Results.Count > 0)
+                {
+                    foreach (var row in queryResult)
+                    {
+                        var transactions = DeserializeCommitRecords(row.Transactions);
+                        deseriazliedResults.AddRange(transactions);
+                    }
+                }
+            } while (continuationToken != null);
```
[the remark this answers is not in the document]
xiazen: will fix

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
xiazen: the purpose of archiving is for test purpose only. It is not designed for production use. I talked with Jason that hiding that `ArchiveLog` setting to another option in a test namespace to make this less discoverable. 

@jason-bragg was insisting on atomic operation I think it is because in that way, we won't lose transaction records which were being deleted originally. ofc there's ways to put them on another partition through two batch operations, and then handling operation failure in a way to ensure records not lost, for example use a queue to retry on failure. But we are not doing that to my understanding is because time is tight. But I will let @jason-bragg speak his reasoning to you, instead of me interpreting.

jason-bragg: Archiving, at this point, is mainly to assist in our recovery and correctness testing.  By deleting entries from the transaction log we loose history needed for correctness checks.  Instead of deleting the entries we (if configured) archive them.  For us to have confidence in the results, archive records need be created for all deleted transaction log entries.  There are many ways of doing this, but performing the operations atomically within the same partition is simple and fast.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -359,34 +360,40 @@ private List<CommitRecord> DeserializeCommitRecords(byte[] serializerCommitRecor
 
         public class ArchivalRow : TableEntity
         {
+            public static string MinRowKey = MakeRowKey(long.MinValue);
+            public static string MaxRowKey = MakeRowKey(long.MaxValue);
+
+            public static string MakeRowKey(long firstTransactionId)
+            {
+                return $"arch_{CommitRow.ToRowKey(firstTransactionId)}";
```
jason-bragg: With this pattern one could change the commit row, without knowing they are also changing the archive row unintentionally.  I suggest this entity be authoritative of it's rowkey, and not delegate any part of it to the commit row.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -359,34 +360,40 @@ private List<CommitRecord> DeserializeCommitRecords(byte[] serializerCommitRecor
...
             {
                 this.Transactions = transactions;
-                this.ArchivalTime = archivalTime;
-                this.ClusterId = clusterId;
+                this.ClusterId = clusterOptions.ClusterId;
                 this.FirstLSN = firstLSN;
                 this.FirstTransactionId = firstTransactionId;
-                PartitionKey = $"{clusterId}${this.ArchivalTime.Ticks}";
-                RowKey = $"{this.ArchivalTime.Ticks}${firstTransactionId}";
+                PartitionKey = clusterOptions.ServiceId;
```
jason-bragg: As in other cases, prefixing service Id, and either consistently using the CommitRecordPartitionKey, or building the pk.  Current pattern is inconsistent and subject to introducing bugs from future maintenance.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -359,34 +360,40 @@ private List<CommitRecord> DeserializeCommitRecords(byte[] serializerCommitRecor
...
         }
 
         private class CommitRow : TableEntity
         {
-            public CommitRow(long firstLSN)
+            public CommitRow(long firstLSN, string serviceId)
             {
                 // All entities are in the same partition for atomic read/writes.
-                PartitionKey = CommitRecordPartitionKey;
+                PartitionKey = serviceId;
```
jason-bragg: see other pk comments.

test/Benchmarks/Benchmarks/TransactionManager/TransactionManagerBentchmarks.cs
```diff
@@ -155,7 +156,8 @@ public Task<bool> Prepare(long transactionId, TransactionalResourceVersion? writ
                 TableName = $"TransactionLog{((uint)Guid.NewGuid().GetHashCode()) % 100000}",
```
jason-bragg: Now that we're partitioning the data by service Id, we don't need this hack.

test/Benchmarks/Benchmarks/TransactionManager/TransactionManagerBentchmarks.cs
```diff
@@ -155,7 +156,8 @@ public Task<bool> Prepare(long transactionId, TransactionalResourceVersion? writ
...
                 ConnectionString = "UseDevelopmentStorage=true"
             });
-            AzureTransactionLogStorage storage = new AzureTransactionLogStorage(client.ServiceProvider.GetRequiredService<SerializationManager>(), azureConfig, Options.Create(new ClusterOptions()));
+            AzureTransactionLogStorage storage = new AzureTransactionLogStorage(client.ServiceProvider.GetRequiredService<SerializationManager>(), azureConfig, 
+                Options.Create(new AzureTransactionArchiveLogOptions()), Options.Create(new ClusterOptions()));
```
jason-bragg: Suggest setting a serviceId in the cluster options.

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
```diff
@@ -30,35 +31,41 @@ public AzureTransactionLogTests()
...
             await logStorage.Append(allTransactions);
-            //a subset of all transactions will be archived 
-            await logStorage.TruncateLog(recordsNum + 2);
+            //all transactions will be archived 
+            var maxSLN = recordsNum + 2;
+            await logStorage.TruncateLog(maxSLN);
+            //since lsn and transaction id are the same in this test, so use it as the query on row key
             var query = new TableQuery<AzureTransactionLogStorage.ArchivalRow>().Where(
-                TableQuery.GenerateFilterCondition(nameof(AzureTransactionLogStorage.ArchivalRow.ClusterId), QueryComparisons.Equal, TestClusterId));
+                TableQuery.GenerateFilterCondition(nameof(AzureTransactionLogStorage.ArchivalRow.RowKey), QueryComparisons.LessThanOrEqual, AzureTransactionLogStorage.ArchivalRow.MakeRowKey(maxSLN)));
```
jason-bragg: Shouldn't query use minrowkey/maxrow key and limit search by partition?

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
```diff
@@ -30,35 +31,41 @@ public AzureTransactionLogTests()
...
 
-        private static async Task<AzureTransactionLogStorage> StorageFactory(AzureTransactionLogOptions azureOptions)
+        private static async Task<AzureTransactionLogStorage> StorageFactory(AzureTransactionLogOptions azureOptions, AzureTransactionArchiveLogOptions archiveOptions)
         {
             var config = new ClientConfiguration();
             var environment = SerializationTestEnvironment.InitializeWithDefaults(config);
             var azureConfig = Options.Create(azureOptions);
-            AzureTransactionLogStorage storage = new AzureTransactionLogStorage(environment.SerializationManager, azureConfig, Options.Create(new ClusterOptions(){ClusterId = TestClusterId}));
+            AzureTransactionLogStorage storage = new AzureTransactionLogStorage(environment.SerializationManager, azureConfig, 
+                Options.Create(archiveOptions), Options.Create(new ClusterOptions(){ClusterId = TestClusterId, ServiceId = "TestServiceID"}));
```
jason-bragg: Suggest different serviceId per tests.  Maybe guid.new.tostring?

test/Transactions/Orleans.Transactions.Azure.Test/GoldenPathTransactionManagerTest.cs
```diff
@@ -41,7 +42,8 @@ private static ITransactionManager MakeTransactionManager()
                 TableName = $"TransactionLog{((uint)Guid.NewGuid().GetHashCode()) % 100000}",
                 ConnectionString = TestDefaultConfiguration.DataConnectionString
             });
-            AzureTransactionLogStorage storage = new AzureTransactionLogStorage(environment.SerializationManager, azureConfig, Options.Create(new ClusterOptions()));
+            AzureTransactionLogStorage storage = new AzureTransactionLogStorage(environment.SerializationManager, azureConfig, 
+                Options.Create(new AzureTransactionArchiveLogOptions()), Options.Create(new ClusterOptions()));
```
jason-bragg: suggests setting service Id

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionLogTests.cs
jason-bragg: Using the entire query range ensures that archiving is not writing entities beyond where it should.

Limiting query to partition key prevents a table scan, which could be very expensive, especially for ci or other checks which may accumulate entities in the table over time.

Partition key also ensures that query is only querying entities from this deployment, in the case that multiple tests are being run at the same time using multiple serviceIds.

Probably should not delete table after run, this is not standard in our storage tests because it can cause failures in other test that may be running in parallel.  Also would hide bugs caused by improper partitioning of data.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionArchiveLogOptions.cs
```diff
@@ -2,7 +2,7 @@
 using System.Collections.Generic;
 using System.Text;
 
-namespace Orleans.Transactions.AzureStorage.Storage.Development
+namespace Orleans.Configuration.Development
```
ReubenBond: Is .Development the right namespace?

Is that because this is 'beta'/unstable or is it because it's not designed for production use?

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -388,13 +398,15 @@ public ArchivalRow(ClusterOptions clusterOptions, byte[] transactions, long firs
...
         {
-            public CommitRow(long firstLSN, string serviceId)
+            public const string MaxRowKey = "crrk_~";
+            public const string MinRowKey = "crrk_";
+            public CommitRow(ClusterOptions clusterOptions, long firstLSN)
             {
                 // All entities are in the same partition for atomic read/writes.
-                PartitionKey = serviceId;
-                RowKey = ToRowKey(firstLSN);
+                PartitionKey = ArchivalRow.MakePartitionKey(clusterOptions.ServiceId);
```
ReubenBond: Yeah? I would think we would give these different partition keys

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
```diff
@@ -388,13 +398,15 @@ public ArchivalRow(ClusterOptions clusterOptions, byte[] transactions, long firs
             public byte[] Transactions { get; set; }
         }
 
-        private class CommitRow : TableEntity
+        public class CommitRow : TableEntity
         {
-            public CommitRow(long firstLSN, string serviceId)
+            public const string MaxRowKey = "crrk_~";
```
ReubenBond: crrk = commit row row key?

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
ReubenBond: Ok, so the reasoning is that we want to 'move' the rows in a batch transaction. That seems fine, though it makes me wonder what the purpose of this feature is.

If it's to improve startup speed (allowing more records to be skipped during recovery) then that's easily accomplished using just the StartRow. If it's so that we can retain the entire tx log for archival purposes then I would assume that we want to be able to move that log into some colder storage, or at least a different storage account.

So I am missing the motivation for this feature.

jason-bragg: We need a history of what transactions have succeeded.  Truncation normally removes completed transactions, losing history.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
xiazen: but both partitionkey and rowkey are prefixed. if row key took cr_, what does partitionKey uses? 
come on it is not that ugly.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
ReubenBond: So why not use the StartRow for this? Then you just have a pointer to the first "non-archived" transaction and you don't need to modify any of the records. It's faster, too.

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
ReubenBond: It's ok, just curious. RowKey and PartitionKey can have the same prefix (or no prefix)

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionLogStorage.cs
ReubenBond: Given this is only indented for test cases, what's here is fine

src/Azure/Orleans.Transactions.AzureStorage/Storage/AzureTransactionArchiveLogOptions.cs
ReubenBond: Ok, got it

Issue #4507, opened 2018-04-24
Exception thrown in MembershipOracle.TryToSuspectOrKill when debug level set to Debug


benjaminpetit:
The following line is incorrect:

``` csharp
if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("-TryToSuspectOrKill {0}: The current status of {0} in the table is {1}, its entry is {2}", entry.SiloAddress.ToLongString(), entry.Status, entry.ToFullString());
````

The parameter `format` does not follow the behavior of `String.Format`, so it will throw an exception "System.FormatException : Index (zero based) must be greater than or equal to zero and less than the size of the argument list".

Format should respect this format: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.1&tabs=aspnetcore2x#log-message-template

We need to do a pass in Orleans codebase to see if we did the same mistake in other places.

andscho-msft:
Did this get past tests because the tests run with a higher minimum LogLevel? Perhaps we should use Trace for LogLevel in tests.

benjaminpetit:
Yes indeed. I am trying to run them with loglovel set to trace to see if it has a big impact or not in the test time. Some of our test relies on timing, so I wouldn't be surprised if we have more test failure using this log level

sergeybykov:
Resolved via #4508.

PR #4508, opened 2018-04-24
Fix Exception thrown in MembershipOracle.TryToSuspectOrKill

benjaminpetit:
Related issue: #4507 

This is the bare minimum fix to avoid any exception at runtime because of logging in Orleans. We will need to do several passes to remove entirely numbered format string.

### DIFF
--- a/src/Orleans.Runtime/MembershipService/MembershipOracle.cs
+++ b/src/Orleans.Runtime/MembershipService/MembershipOracle.cs
@@ -978,7 +978,11 @@ namespace Orleans.Runtime.MembershipService
             var tuple = table.Get(silo);
             var entry = tuple.Item1;
             string eTag = tuple.Item2;
-            if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("-TryToSuspectOrKill {0}: The current status of {0} in the table is {1}, its entry is {2}", entry.SiloAddress.ToLongString(), entry.Status, entry.ToFullString());
+            if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("-TryToSuspectOrKill {siloAddress}: The current status of {siloAddress} in the table is {status}, its entry is {sntry}",
+                entry.SiloAddress.ToLongString(), // First
+                entry.SiloAddress.ToLongString(), // Second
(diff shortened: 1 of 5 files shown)


test/TesterInternal/StreamingTests/SingleStreamTestRunner.cs
```diff
@@ -510,7 +510,7 @@ private Task ValidatePubSub(Guid streamId, string providerName)
                 activationCount = await consumer.GetNumActivations(this.client);
             }
             var expectActivationCount = 0;
-            logger.Info("Test {0} CheckGrainsDeactivated: {1}ActivationCount = {2}, Expected{1}ActivationCount = {3}", testNumber, str, activationCount, expectActivationCount);
+            logger.Info("Test {testNumber} CheckGrainsDeactivated: {type}ActivationCount = {activationCount}, Expected{type}ActivationCount = {expectActivationCount}", testNumber, str, activationCount, expectActivationCount);
```
ReubenBond: Is this right? The parameter count doesn't match? In the MembershipOracle change you duplicate the parameter which is used twice.

src/Orleans.Runtime/MembershipService/MembershipOracle.cs
```diff
@@ -978,7 +978,11 @@ private void LogFailedProbe(SiloAddress silo, int pingNumber, Exception failureR
             var tuple = table.Get(silo);
             var entry = tuple.Item1;
             string eTag = tuple.Item2;
-            if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("-TryToSuspectOrKill {0}: The current status of {0} in the table is {1}, its entry is {2}", entry.SiloAddress.ToLongString(), entry.Status, entry.ToFullString());
+            if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("-TryToSuspectOrKill {siloAddress}: The current status of {siloAddress} in the table is {status}, its entry is {sntry}",
```
ReubenBond: typo: 'sntry'

ReubenBond:
@dotnet-bot test this please

PR #4509, opened 2018-04-25
Consolidate SiloStatisticsOptions and ClientStatisticsOptions

ReubenBond:
The options system doesn't handle inheritance.
The divide between `SiloStatisticsOptions` and `ClientStatisticsOptions` complicates things unnecessarily.

Also split out `DeploymentLoadPublishingOptions`


src/Orleans.Core/Configuration/Options/StatisticsOptions.cs
```diff
@@ -1,20 +1,19 @@
...
-    public abstract class StatisticsOptions
+    public class StatisticsOptions
     {
         /// <summary>
         /// The PerfCounterWriteInterval property specifies the frequency of updating the windows performance counters.
         /// The default is 30 seconds.
         /// </summary>
         public TimeSpan PerfCountersWriteInterval { get; set; } = DEFAULT_PERF_COUNTERS_WRITE_PERIOD;
-        public static readonly TimeSpan DEFAULT_PERF_COUNTERS_WRITE_PERIOD = Constants.INFINITE_TIMESPAN;
+        public static readonly TimeSpan DEFAULT_PERF_COUNTERS_WRITE_PERIOD = TimeSpan.FromSeconds(30);
```
ReubenBond: This property is not used on the client, but the silo has a default of 30s

ReubenBond:
I think this PR should be straight forward and non-controversial.
It does change some user-facing API by removing those two classes and replacing them with their base classes, but that is a) the right move, and b) not a silent breakage

xiazen:
code LGTM. But this PR is built on the statement of `The options system doesn't handle inheritance.`, which definitely has mixed understanding in the team. Since I definitely heard of something like inheritance is fine with options . Can you send me articles which says it is not fine. And then we should do a scan over all the options we have and do a purge ?

ReubenBond:
It would be surprising if the options system did support inheritance (what would it mean?). Consider the following program:
```C#
using System;
using Microsoft.Extensions.DependencyInjection;

namespace Test
{
    class Program
    {
        static void Main(string[] args)
        {
            services.AddOptions();
            services.Configure<BaseOptions>(o => o.Num = 4);
            services.Configure<SubOptions>(o => o.Strang = "yo yo");
            var sp = services.BuildServiceProvider();

            var bas = sp.GetRequiredService<IOptions<BaseOptions>>();
            var sub = sp.GetRequiredService<IOptions<SubOptions>>();
            Console.WriteLine(bas.Value);
            Console.WriteLine(sub.Value);
        }
    }

    public class BaseOptions
    {
        public int Num { get; set; }

        {
            return $"{nameof(Num)}: {Num}";
        }
    }

    public class SubOptions : BaseOptions
    {
        public string Strang { get; set; }

        {
            return $"{base.ToString()}, {nameof(Strang)}: {Strang}";
        }
    }
}
```

The output is:

```
Num: 4
Num: 0, Strang: yo yo
```

xiazen:
oh oh , this is not good. I found other occurrence of option inheritance : 
 `MessagingOptions`, and `Silo/ClientMessagingOptions`

want to fix that in next pr? shouldn't be a breaking change

xiazen:
wait actually I took my words back. Looking at the program again, what it does is correct. right? 
SubOPtions.Num is 0 because you didn't configure it, so it should be 0

xiazen:
yeah you are not suppose to do that and you cannot do that because the base type is abstract. I guess we can merge this pr since SiloStatisticOptions and ClientStatisticOptions and the inheritance is more complex than needed. `Silo/ClientMessagingOptions` are more different with each other. It make more sense for them to have a base class. And we won't need to worry someone ask for the base type, since the base type is abstract , get it from options won't compile.

xiazen:
this is a small breaking change. I will @sergeybykov here for his opinion.

I'm not concerned by it since it is a easy to fix breaking change. But @sergeybykov may have other opinion since he is in charge of the semantic versioning .

PR #4510, opened 2018-04-25
ClientOptionsLogger and SiloOptionsLogger are not configuration options

MelnikovIG:
`ClientOptionsLogger` and `SiloOptionsLogger` are not configuration options and could not be used as ASP.NET `IOptions<T>` options.


ReubenBond:
Thanks :)

PR #4511, opened 2018-04-25
Fix logging level check on Grain exception

### LINKED ISSUE #4512: Fix logging level check on Grain exception
Currently Grain exception is not logged on SiloHost side if exception occurred and LogLevel is Warning

MelnikovIG:
Currently Grain exception is not logged on SiloHost side if exception occurred and LogLevel is Warning
Fixes #4512


src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -365,7 +365,7 @@ public void SniffIncomingMessage(Message message)
                 }
                 catch (Exception exc1)
                 {
-                    if (invokeExceptionLogger.IsEnabled(LogLevel.Debug) || message.Direction == Message.Directions.OneWay)
+                    if (invokeExceptionLogger.IsEnabled(LogLevel.Warning) || message.Direction == Message.Directions.OneWay)
                     {
                         invokeExceptionLogger.Warn(ErrorCode.GrainInvokeException,
```
ReubenBond: Perhaps this should be logged at a lower level instead. An exception from a grain call is not necessarily a 'warning' - it really depends on the developer. So in my opinion, these should be logged at Debug instead.

Are there other opinions on this?

MelnikovIG: Are we really want to suppress logging grain method exceptions in productions for silo host, where log level not set to Debug? There is no way to analyze what happened on silo host side, because this exception will not be logged nowhere. I think is is better warn if something goes wrong by default, and if developer want to ignore exception, he can catch it manually.

ReubenBond: Developers can log those exceptions by adding an `IIncomingGrainCallFilter` anyway, so it's not as though anything is being suppressed.

Should exceptions in a grain method show up as warnings in the developer's monitoring system?

jdom: Hmmm, I was going to say that the caller can handle and log these, but given this is on OneWay calls, I kind of agree of bumping up the severity. The reality is that the developers should fix the code so that one way grain methods do not throw, so having a Warning that annoys them makes sense.
We should of course not log a duplicate Warning for infrastructure-level exceptions if there are any, but for app level exceptions I think I agree with the change (don't feel that strongly either, but it's better to annoy them than them not knowing about it)

jdom: if for some reason they decide they *don't* want to fix their code (or they can't temporarily) and they are getting spammed, they can set the minimum log level to Error by setting it for the following category: `$"{typeof(Grain).FullName}.InvokeException"`.
So it's a better default IMO to get notified by these otherwise silent exceptions on one way methods.

ReubenBond: Okay, that makes sense to me: enabled by default, but they can opt-out.

@MelnikovIG would you be able to update the PR to remove the `OneWay` check, since it's not needed?

ReubenBond: In that case we should split the logic into an if/else-if. I don't mind either way, but given users can disable the warnings for these very easily, I kind-of like the idea of logging an exception at Warning by default

jdom: I'd prefer the split. My opinion is that on RPC calls, having a warning will almost always be spam, as it's normal to bubble up exceptions to the caller. But one way calls is a completely different thing, because it will be silently suppressed

MelnikovIG: So is it okay if it will be like that?
```                    
var exceptionMessage = "Exception during Grain method call of message: " + message;
if (message.Direction == Message.Directions.OneWay)
{
    invokeExceptionLogger.Warn(ErrorCode.GrainInvokeException, exceptionMessage, exc1);
}
else
{
    invokeExceptionLogger.Debug(ErrorCode.GrainInvokeException, exceptionMessage, exc1);
}

ReubenBond: change the else to `else if (invokeExceptionLogger.IsEnabled(LogLevel.Debug))` and inline the exception message allocation

MelnikovIG: Done

ReubenBond:
Thank you, @MelnikovIG!

Issue #4482, opened 2018-04-17
DynamoDb null reference on read


stephenlautier:
We are encountering a `NullReferenceException` on activation in special cases e.g. This seems to occur only when changing a type namespace.

This is the following error
```
Exc level 0: System.NullReferenceException: Object reference not set to an instance of an object.
   at Orleans.Storage.DynamoDBGrainStorage.ConvertFromStorageFormat(GrainStateRecord entity)
   at Orleans.Storage.DynamoDBGrainStorage.<ReadStateAsync>d__20.MoveNext()
```

### Additional info
We are using Orleans `2.0.0-rc2`

#### Dynamo Options
 - `config.UseJson = true;`
 ` config.TypeNameHandling` (unset)

Just reporting this maybe it gets handled better or more relevant exception

sergeybykov:
>This seems to occur only when changing a type namespace.

Do you mean the identity of the type of the persisted object changes between when it is saved to storage and when it is later read back?

stephenlautier:
@sergeybykov yes, for instance changing the namespace of the State class

e.g.

```cs
namespace Friend // rename to Enemy
{
    public class EntityState 
    {
          public string Name { get; set; }
    }
}
```

galvesribeiro:
I believe this is not a supported scenario on any provider as the full type name (namespace + type) is used to reference the state... Specifically on DynamoDB, it is used as Range index and RowKey.

I would appreciate if you have a full stacktrace for me :)

sergeybykov:
Yes, changing of the namespace changes the type identity, and without a special type forwarding logic of some sort, there is no way for any serializer to know that those type should be considered the same.

stephenlautier:
@galvesribeiro maybe you missed my last point, I'm not expecting it to fetch the old data and work. I reported this because its throwing a `NullReferenceException` on activation; So what I'm suggesting is either somehow just returns `null` or throw a better/nicer exception. In my example I would rather have a null than an exception because then data would be refetched and everything would have worked

#### Exception

#### message
```
[...]

PR #4513, opened 2018-04-25
Fix NRE on AWS DynamoDB storage provider

galvesribeiro:
Fix NRE on AWS DynamoDB storage provider. #4482


ReubenBond:
Thanks, Gutemberg 🙂

PR #4514, opened 2018-04-25
fix(dynamo): fix grain storage when an exception has occured

stephenlautier:
This was done to address https://github.com/dotnet/orleans/issues/4482#issuecomment-383768677

Fix for when using `UseJson` and an exception was thrown it was causing a null reference
Also included a minor refactor

### DIFF
--- a/src/AWS/Orleans.Persistence.DynamoDB/Provider/DynamoDBGrainStorage.cs
+++ b/src/AWS/Orleans.Persistence.DynamoDB/Provider/DynamoDBGrainStorage.cs
@@ -303,7 +303,7 @@ namespace Orleans.Storage
 
             try
             {
-                // Try to rehydrate
+                // rehydrate
                 if (binaryData?.Length > 0)
                 {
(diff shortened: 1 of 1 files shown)


src/AWS/Orleans.Persistence.DynamoDB/Provider/DynamoDBGrainStorage.cs
```diff
@@ -301,42 +301,37 @@ internal object ConvertFromStorageFormat(GrainStateRecord entity)
...
-                if (binaryData.Length > 0)
+                if (binaryData?.Length > 0)
                 {
                     sb.AppendFormat("Unable to convert from storage format GrainStateEntity.Data={0}", binaryData);
                 }
                 else if (!string.IsNullOrEmpty(stringData))
                 {
                     sb.AppendFormat("Unable to convert from storage format GrainStateEntity.StringData={0}", stringData);
                 }
-                if (dataValue != null)
```
stephenlautier: this was removed as it was never possible entering here

PR #4538, opened 2018-04-30
Azure table TransactionalStateStorage

jason-bragg:


src/Azure/Orleans.Transactions.AzureStorage/Hosting/AzureTableTransactionsSiloBuilderExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            return builder.ConfigureServices(services => services.AddAzureTableTransactionalStateStorage(name, configureOptions));
+        }
+
+        private static IServiceCollection AddAzureTableTransactionalStateStorage(this IServiceCollection services, string name,
+            Action<OptionsBuilder<AzureTableTransactionalStateOptions>> configureOptions = null)
+        {
+            configureOptions?.Invoke(services.AddOptions<AzureTableTransactionalStateOptions>(name));
+            services.ConfigureNamedOptionForLogging<AzureTableTransactionalStateOptions>(name);
+            services.TryAddSingleton<ITransactionalStateStorageFactory>(sp => sp.GetServiceByName<ITransactionalStateStorageFactory>(ProviderConstants.DEFAULT_STORAGE_PROVIDER_NAME));
```
ReubenBond: Should this be `name` here or am I misunderstanding?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+using Orleans.Transactions.Abstractions;
+
+namespace Orleans.Transactions.AzureStorage
+{
+    public class AzureTableTransactionalStateStorage<TState> : ITransactionalStateStorage<TState>
+        where TState : class, new()
+    {
+        private readonly CloudTable table;
+        private readonly string partition;
+        JsonSerializerSettings JsonSettings;
```
ReubenBond: no access modifiers?

jason-bragg: can you clarify?  I'm not sure what you mean.

src/Azure/Orleans.Transactions.AzureStorage/Hosting/AzureTableTransactionsSiloBuilderExtensions.cs
jason-bragg: This is just setting up a mapping for the default, so that if anyone resolves ITransactionalStateStorageFactory it will get the one named ProviderConstants.DEFAULT_STORAGE_PROVIDER_NAME

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorageFactory.cs
```diff
@@ -0,0 +1,78 @@
...
+            this.name = name;
+            this.options = options;
+            this.clusterOptions = clusterOptions.Value;
+            this.loggerFactory = loggerFactory;
+        }
+
+        public ITransactionalStateStorage<TState> Create<TState>(string stateName, IGrainActivationContext context) where TState : class, new()
+        {
+            string partitionKey = MakePartitionKey(context, stateName);
+            var settingsResolver = ActivatorUtilities.CreateInstance< JsonSerializerSettingsResolver>(context.ActivationServices);
```
ReubenBond: This feels a bit odd to me, since getting those settings creates a non-insignificant object graph. I feel that we should either:
* Use `IOptions<JsonSerializerSettings>` (which I realize might feel strange, but it works)
* Use `ActivatorUtilities.GetServiceOrCreateInstance` and optionally TryAdd `JsonSerializerSettings` to the container by default if this provider is being used
* Create the settings once in this class, therefore not creating them in the scope of the activation

I assume there's some intention that users can customize those settings on a per-activation basis, but I'm not sure that's worth the cost

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+                Task<KeyEntity> keyTask = ReadKey();
+                Task<List<StateEntity>> statesTask = ReadStates();
+                this.key = await keyTask;
+                this.states = await statesTask;
+                if (string.IsNullOrEmpty(this.key.ETag))
+                {
+                    return new TransactionalStorageLoadResponse<TState>();
+                }
+                TState commitedState = TryFindState(this.key.CommittedTransactionId, out TState found) ? found : new TState();
+                var pendingStates = states.Select(s => new PendingTransactionState<TState>(s.TransactionId, s.SequenceId, s.GetState<TState>(this.JsonSettings))).ToList();
```
ReubenBond: LINQ usage here is expressive but inefficient, particularly since we know the size of the result and therefore don't need a bunch of allocations for list resizing and for the Select closure. The alternative is to pre-allocate the pendingStates as `new List<PendingTransactionState>(states.Count)` and `foreach` over `states` adding them to the list.

Feel free to ignore, it's probably not a big deal

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+                if (!TryFindState(transactionIdToCommit, out TState state))
+                {
+                    this.logger.LogCritical("Transactional state non-recoverable error.  Attempting to confirm a transaction {TransactionId} for which no state exists.", transactionIdToCommit);
+                    throw new InvalidOperationException($"Transactional state non-recoverable error.  Attempting to confirm a transaction {transactionIdToCommit} for which no state exists.");
+                }
+
+                this.key.ETag = expectedETag;
+                this.key.Metadata = metadata;
+                this.key.CommittedTransactionId = transactionIdToCommit;
+                await WriteKey();
```
ReubenBond: ConfigureAwait?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+                this.key.Metadata = metadata;
+                if (string.IsNullOrEmpty(this.key.ETag))
+                    batchOperation.Insert(this.key);
+                else
+                    batchOperation.Replace(this.key);
+
+                // add new states
+                List<Tuple<string,long>> stored = this.states.Select(s => Tuple.Create(s.TransactionId, s.SequenceId)).ToList();
+                List<StateEntity> newStates = new List<StateEntity>();
+                foreach (PendingTransactionState<TState> pendingState in statesToPrepare.Where(p => !stored.Contains(Tuple.Create(p.TransactionId, p.SequenceId))))
```
ReubenBond: This is quite inefficient. I'll stop commenting on efficiency for the sake of expedience.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+                    this.logger.LogCritical("Transactional state non-recoverable error.  Attempting to confirm a transaction {TransactionId} for which no state exists.", transactionIdToCommit);
+                    throw new InvalidOperationException($"Transactional state non-recoverable error.  Attempting to confirm a transaction {transactionIdToCommit} for which no state exists.");
+                }
+
+                this.key.ETag = expectedETag;
+                this.key.Metadata = metadata;
+                this.key.CommittedTransactionId = transactionIdToCommit;
+                await WriteKey();
+                var dead = this.states.Where(p => string.Compare(p.TransactionId, transactionIdToCommit) <0).ToList();
+                this.states = this.states.Where(p => string.Compare(p.TransactionId, transactionIdToCommit) >= 0).ToList();
```
ReubenBond: We consider these APIs 'pubternal', I think - i.e, they're open but not really intended for users to implement until we're stable. Otherwise I wonder if it would be worth lifting some logic (like this cleanup) to a different layer.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+        private bool TryFindState(string transactionId, out TState state)
+        {
+            state = null;
+            if (string.IsNullOrEmpty(transactionId))
+                return false;
+
+            StateEntity entity = this.states.FirstOrDefault(s => s.TransactionId == transactionId);
+            if (entity == null)
+            {
+                throw new InvalidOperationException($"Transactional state non-recoverable error.  Commited state for transaction {transactionId} not found.");
```
ReubenBond: TryFind seems like it would return false if the transaction is not found. Maybe this exception should be pulled out to the call sites

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/StateEntity.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public static string MakeRowKey(string TransactionId, long sequenceId)
+        {
+            return AzureStorageUtils.SanitizeTableProperty($"{RKMin}{TransactionId}_{sequenceId.ToString("x16")}");
+        }
+
+        public T GetState<T>(JsonSerializerSettings JsonSettings)
+        {
+            return JsonConvert.DeserializeObject<T>(this.StateJson, JsonSettings);
+        }
```
ReubenBond: I'm still mulling over the question of storage format, sorry - coming up with a good solution is tricky given our existing legacy. For the time being, we can keep this the way it is and add another property for a different/flexible storage format.

src/Orleans.Core/Transactions/ITransactionalResource.cs
```diff
@@ -84,7 +84,7 @@ public int CompareTo(TransactionalResourceVersion other)
 
         public override string ToString()
         {
-            return $"{TransactionId}#{WriteNumber}";
+            return $"{TransactionId:x16}#{WriteNumber:x16}";
```
ReubenBond: WriteNumber:x8 since it's an int?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+            try
+            {
+                var batchOperation = new TableBatchOperation();
+
+                this.key.ETag = expectedETag;
+                this.key.Metadata = metadata;
+                if (string.IsNullOrEmpty(this.key.ETag))
+                    batchOperation.Insert(this.key);
+                else
+                    batchOperation.Replace(this.key);
```
xiazen: TableOperation.InsertOrReplace does this if for you

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+            TableQuerySegment<KeyEntity> queryResult = await table.ExecuteQuerySegmentedAsync(query, null).ConfigureAwait(false);
+            return queryResult.Results.Count == 0
+                ? new KeyEntity() { PartitionKey = this.partition }
+                : queryResult.Results[0];
+        }
+
+        private Task WriteKey()
+        {
+            return (string.IsNullOrEmpty(this.key.ETag))
+                ? this.table.ExecuteAsync(TableOperation.Insert(this.key))
```
xiazen: same here

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+            return true;
+        }
+
+        private async Task Cleanup(List<StateEntity> deadStates)
+        {
+            try
+            {
+                var batchOperation = new TableBatchOperation();
+                deadStates.ForEach(batchOperation.Delete);
+                await table.ExecuteBatchAsync(batchOperation).ConfigureAwait(false);
```
xiazen: batch op size limit check here  ?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorageFactory.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public void Participate(ISiloLifecycle lifecycle)
+        {
+            lifecycle.Subscribe(OptionFormattingUtilities.Name<AzureTableTransactionalStateStorageFactory>(this.name), this.options.InitStage, Init);
+        }
+
+        private string MakePartitionKey(IGrainActivationContext context, string stateName)
+        {
+            string grainKey = context.GrainInstance.GrainReference.ToKeyString();
+            var key = $"ts_{this.clusterOptions.ServiceId}_{grainKey}_{stateName}";
```
xiazen: wrap this string template into a static method? since I don't know what key it s creating, partition key or rowkey?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/StateEntity.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Orleans.Transactions.AzureStorage
+{
+    internal class StateEntity : TableEntity
+    {
+        public const string RKMin = "ts_";
+        public const string RKMax = "ts_~";
+
+        public string TransactionId { get; set; }
+        public long SequenceId { get; set; }
+        public string StateJson { get; set; }
```
xiazen: we only support json for now?

src/Azure/Shared/Storage/AzureStorageUtils.cs
```diff
@@ -25,10 +26,28 @@ namespace Orleans.Streaming.AzureStorage
...
+    /// Constants related to Azure table storage.
+    /// </summary>
+    internal static class AzureTableConstants
+    {
+        /// <summary>
+        /// ETag of value "*" to match any etag for conditional table operations (update, nerge, delete).
+        /// </summary>
+        public const string ANY_ETAG = "*";
+
+        public const string PKProperty = "PartitionKey";
```
xiazen: nameof(TableEntity.PartitionKey) ? for compile time safe

src/Azure/Shared/Storage/AzureStorageUtils.cs
```diff
@@ -25,10 +26,28 @@ namespace Orleans.Streaming.AzureStorage
...
+        public const string RKProperty = "RowKey";
+
+        public const int MaxBatchSize = 100;
```
xiazen: use `TableConstants.TableServiceBatchMaximumOperations`

src/Azure/Shared/Storage/AzureStorageUtils.cs
xiazen: or `TableConstants.PartitionKey`

https://docs.microsoft.com/en-us/dotnet/api/microsoft.windowsazure.storage.table.protocol.tableconstants?view=azure-dotnet

src/Azure/Shared/Storage/AzureStorageUtils.cs
```diff
@@ -25,10 +26,28 @@ namespace Orleans.Streaming.AzureStorage
...
+    /// </summary>
+    internal static class AzureTableConstants
+    {
+        /// <summary>
+        /// ETag of value "*" to match any etag for conditional table operations (update, nerge, delete).
+        /// </summary>
+        public const string ANY_ETAG = "*";
+
+        public const string PKProperty = "PartitionKey";
+        public const string RKProperty = "RowKey";
```
xiazen: similar here

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
xiazen: he means `private` or `internal` or `public` ?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
xiazen: so pending states include committed state? is this by design or?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+                    throw new InvalidOperationException($"Transactional state non-recoverable error.  Attempting to confirm a transaction {transactionIdToCommit} for which no state exists.");
+                }
+
+                this.key.ETag = expectedETag;
+                this.key.Metadata = metadata;
+                this.key.CommittedTransactionId = transactionIdToCommit;
+                await WriteKey();
+                var dead = this.states.Where(p => string.Compare(p.TransactionId, transactionIdToCommit) <0).ToList();
+                this.states = this.states.Where(p => string.Compare(p.TransactionId, transactionIdToCommit) >= 0).ToList();
+                Cleanup(dead).Ignore();
```
sebastianburckhardt: Why is the cleanup asynchronous, separate from the commit? Is that really necessary? I think doing everything in a single batch is cleaner and peforms better.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,184 @@
...
+                {
+                    var newState = StateEntity.Create(this.JsonSettings, this.partition, pendingState);
+                    newStates.Add(newState);
+                    batchOperation.Insert(newState);
+                }
+
+                if (batchOperation.Count > AzureTableConstants.MaxBatchSize)
+                {
+                    this.logger.LogError("Too many pending states. PendingStateCount {PendingStateCount}.", batchOperation.Count);
+                    throw new InvalidOperationException($"Too many pending states. PendingStateCount {batchOperation.Count}");
```
sebastianburckhardt: To make this robust, should probably prepare using multiple batches instead of throwing an exception. Not sure if that level of reliability is necessary, I believe this provider is not meant for production use?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/StateEntity.cs
```diff
@@ -0,0 +1,40 @@
...
+            string partitionKey, PendingTransactionState<T> pendingState)
+            where T : class, new()
+        {
+            return new StateEntity
+            {
+                PartitionKey = partitionKey,
+                RowKey = MakeRowKey(pendingState.TransactionId, pendingState.SequenceId),
+                TransactionId = pendingState.TransactionId,
+                SequenceId = pendingState.SequenceId,
+                StateJson = JsonConvert.SerializeObject(pendingState.State, JsonSettings)
```
sebastianburckhardt: Not robust if state is larger than what can be stored in Azure table row. Again, maybe o.k. if there is a warning somewhere against using this provider in production.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
sebastianburckhardt: Probably should break into multiple batches if above batch size limit.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorageFactory.cs
jason-bragg: TL;RD
I suggest the following:
*  Resolving the json settings once and using the same setting for all the storage instances, even though that technically violates the grain scoping.
* Add issue to support pluggable state serialization.
* Add issue to design and formalize handling of json serialization settings management.

--

This sux, imo.  The problem is deeper than this PR so I'll open issues to address in future.

At a high level this is related to how we handle serialization for persistence, an unresolved issue.  I've opted to use json rather than the serialization manager as an initial starting point.

To support this I need to  call GetDefaultSerializerSettings, which requires the type resolver and grain factory from the container.  Since this is in the context of a grain these 'should' be resolved using the grains container, as it's scoped, which is the case.  I could resolve this once for all storage instances, but that technically violates the scoping.  We've done this before, so it's not unprecedented so I'll change that, but overall the larger issue still remains.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: I agree with the concern. In general link statements are almost never efficient.  I'm unconvinced, at this time, that the performance implications justify the maintainabilty cost of not using link.  This seems like premature optimization, do you disagree?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: The ITransactionalStateStorage and ITransactionalStateStorageFactory interfaces are public, and we do expect custom implementations.  The patterns are still in flux, but the entities and cleanup logic here is specific to this implemetation.  Do you disagree?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: Will refactor.  There are 3 cases, transctionId requested is null, transactionId is valid and not found (error) transactionId is found and state is found.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
ReubenBond: The cleanup logic may be semi-specific, but I think most providers would need/want some kind of cleanup process.

Anyhow, users will know to copy this implementation.

src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
```diff
@@ -47,6 +45,8 @@ public PendingTransactionState(string transactionId, long sequenceId, TState sta
     public class TransactionalStorageLoadResponse<TState>
         where TState : class, new()
     {
+        public TransactionalStorageLoadResponse() : this(null, new TState(), null, new List<PendingTransactionState<TState>>()){}
```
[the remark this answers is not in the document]
jason-bragg: arg is IReadOnlyList.  Will look at, may find better way.

src/Orleans.Transactions/State/TransactionalStateStorageProviderWrapper.cs
```diff
@@ -17,56 +16,58 @@ internal class TransactionalStateStorageProviderWrapper<TState> : ITransactional
...
-            return stateStorage.Etag;
+            await this.StateStorage.WriteStateAsync();
+            return this.StateStorage.Etag;
         }
 
-        private IStorage<TransactionalStateRecord<TState>> GetStateStorage(string stateName)
+        private IStorage<TransactionalStateRecord<TState>> GetStateStorage()
         {
-            return this.stateStorages.GetOrAdd(stateName, name => new StateStorageBridge<TransactionalStateRecord<TState>>(name, this.context.GrainInstance.GrainReference, grainStorage, this.loggerFactory));
+            string fullStateName = $"{this.context.GrainInstance.GetType().FullName}-{this.stateName}";
```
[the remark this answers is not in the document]
jason-bragg: Will look into.
In general, I'm following the same pattern as storage bridge, with all of the same problems.  Figured we'd fix this when we address those issues, under the presumtion it is better to be wrong in a consistent manner then be wrong in many differnt ways.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: Insert or replace will not respect etag.  We want to respect etag if it's set.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: same answer :)

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/StateEntity.cs
jason-bragg: atm, yes.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: There is no difference between pending and committed states entities in storage.  The commited state is simply the state indicated by the key entity.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorageFactory.cs
```diff
@@ -0,0 +1,78 @@
...
+            this.options = options;
+            this.clusterOptions = clusterOptions.Value;
+            this.loggerFactory = loggerFactory;
+        }
+
+        public ITransactionalStateStorage<TState> Create<TState>(string stateName, IGrainActivationContext context) where TState : class, new()
+        {
+            string partitionKey = MakePartitionKey(context, stateName);
+            var settingsResolver = ActivatorUtilities.CreateInstance< JsonSerializerSettingsResolver>(context.ActivationServices);
+            return ActivatorUtilities.CreateInstance<AzureTableTransactionalStateStorage<TState>>(context.ActivationServices, this.table, partitionKey, settingsResolver.Settings);
```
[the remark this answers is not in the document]
jason-bragg: I could, but I'd need to resolve all of the dependencies from the container in this code.   This resolves them for me.  It also allows maintainers to add more dependencies without changing this code.  Basically this code only fills in the arguments that are not in the container.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: Doing it all in a single operation would be cleaner.  Agreed
Doing it all in a single operation would be faster.  Agreed

I elected to perform cleanup as a separate operation to reduce cost and risk of the critical operation.  The cleanup is optional in that it does not affect the core behavior other than making loads slower over time.  I didn't want to risk storage errors in the confirm path, which is critical.

There are other patterns that could be used:
* We could leave the entities in storage and scope the load query by starting it with the committed transaction id, but that means we couldn't perform the key and state queries in parallel.
* We could setup some sort of background operation that periodically cleaned up the table, but then we'd need to manage that operation.
* We could clean up the table once on load, but that would still effect load time.

I'm open to other solutions, I just don't want to introduce unnecessary risk to the commit operation.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: This provider is intended for production use, but system is still in development.  I'll add issues to revisit shortcomings.

Operation must be atomic, which is only provided by a single batch.  We could change api surface to support partial success, but given current api expectations and storage capabilities, best we can do is enforce this limitation.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/StateEntity.cs
jason-bragg: Understood.  The plan is to define expected behaviors (including state size and other limitations) as part of the API definition.  We've not done that yet.  Will add tracking issue.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: I agree with your suspicions, but don't think we've enough information to draw such conclusions at this time.  Some systems may not cleanup at all, but instead perform more targeted loads, others may archive the states for transaction history.  Some may keep all state in a single object (blob storage).  Even if cleanup is desired, the pattern I'm using is debatable. 

I advocate keeping the public surface as simple and direct as is practicable and geared towards the needs of the transaction orchestration.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorageFactory.cs
ReubenBond: It doesn't violate scoping. Violating scoping would be pulling something from a Scoped or Transient service from a Singleton or service. It's fine to pull a Singleton from a Scoped service.

If we move the JsonSettings up into the factory ctor then it's fine.

As for JSON vs SerializationManager, I'm not convinced that SerializationManager is inherently bad for persistence - it allows users to register serializers which are suitable for persistence and those will be used preferentially. The issue is that we want to somehow avoid using unsuitable serializers - almost as if having another SerializationManager which is configured without generated serializers or a fallback serializer.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
ReubenBond: Seems ok to me for now.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorageFactory.cs
jason-bragg: Regarding scoping.  Calling logic is not expected to know or care if the services are singleton, transient, scoped, or whatever, it only cares that it acquires a service from the container for its scope, which this would not, that's why I'm arguing it violates scoping.  Again, we've done this before, so I'm not necessarily opposed to it, but in general, I try to respect a grains scope as every violation risks unexpected behaviors in the future.

> As for JSON vs SerializationManager, I'm not convinced that SerializationManager is inherently bad for persistence.

Neither am I.  I claimed only that our persistence serialization problems are wider than this PR and unresolved.  Until we figure that out, serializing using json takes our serializers out of the picture.  Most users understand json, and understand the implications of it’s use, which is not the case for our serializers.

src/Azure/Shared/Storage/AzureStorageUtils.cs
jason-bragg: hmm, const no for azure :/

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: added it inside write key, where the external call is made.

src/Azure/Orleans.Transactions.AzureStorage/Hosting/AzureTableTransactionsSiloBuilderExtensions.cs
ReubenBond: It seems odd, but ok. Maybe it would be better to add a special case for default instead of unconditionally adding a registration to a service which may not exist.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
ReubenBond: If you have to pick one place to add the call, it should be in the top-level async method. The effect of `ConfigureAwait(false)` is local to the method which it is used in, it has no effect on the caller to that method.

In other words, this `await` here will continue on the captured context

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
sebastianburckhardt: > reduce cost and risk of the critical operation

I am not sure I see a difference in the "criticality" ... by the time this is called the transaction has already irreversably committed (i.e. commit record is present in TM log). So, everything that is happening here is really a "cleanup" operation.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
sebastianburckhardt: Ok, probably good enough for now. We may revisit this when looking at the storage interface for the distributed TM.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: bleh, I read up on this more.. ConfigureAwait sux.  Looks like it needs to be called at every level.  I thought it was only needed on the async call that actually blocked.   Will fix.. again.. bleh!??

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorage.cs
jason-bragg: > I am not sure I see a difference in the "criticality"
Deleting older states is optional cleanup.
Confirm success is crtical to operation.  This operation must eventually succeed and failures negativly effect the performacne.

Authoritatively you are correct, this is just notifying the grain that the transaction has been commited. but this operation *must* eventually succeed, and every failure keeps following transaction states in a pending state and dependent on the committed transaction.  This does not lead to errors, assuming this operation eventually succeeds, but does negatively affect the performance of the system.

This operation currently writes one single entity, which is as low risk as it gets, as opposed to writing a batch and dealing with the complexity (batch entity counts and size limits) and the risk in which that entails.  To risk this to perform a mostly optional cleanup operation in a cleaner and faster manner does not seem a good trade.

We can always cleanup and optimize the cleanup operation, but it is logically separable from the confirm operation which is not optional.

src/Azure/Orleans.Transactions.AzureStorage/Hosting/AzureTableTransactionsSiloBuilderExtensions.cs
jason-bragg: This is the pattern used to support legacy providers which indicated a provider was the default by naming it 'default', whether there was a provider named that or not.  I am not at all opposed to refining or altering this pattern, but we should be consistent.  I'll open an issue to review this pattern.

xiazen:
@dotnet-bot test functional

PR #4539, opened 2018-04-30
Add transaction recoverbility test runner

xiazen:
- Add transaction recoverability test runners. 
- Add transaction recoverability tests running against single TM. currently isn't included into functional or bvt, because single TM has recoverability problems , so the tests isn't passing. But it is higher priority to get the test runner merged in so we can test the new distributed TM. If you don't like it, I can remove the actual tests.

### DIFF
--- a/test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
+++ b/test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
@@ -1,0 +1,147 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.Extensions.Logging;
+using Orleans.Runtime;
+using Orleans.TestingHost;
+using Xunit;
+using Xunit.Abstractions;
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Providers;
+using Orleans.TestingHost.Utils;
+using Orleans.Hosting;
+
+namespace Orleans.Transactions.Tests
+{
+    public class TransactionRecoveryTestsRunner : TransactionTestRunnerBase
+    {
+        private readonly Random seed;
+        private readonly TestCluster testCluster;
+        private readonly ILogger logger;
+        private static TimeSpan RecoveryTimeout = TimeSpan.FromSeconds(60);
+        public TransactionRecoveryTestsRunner(TestCluster testCluster, ITestOutputHelper output)
+            : base(testCluster.GrainFactory, output)
+        {
+            this.testCluster = testCluster;
+            this.logger = this.testCluster.ServiceProvider.GetService<ILogger<TransactionRecoveryTestsRunner>>();
+            this.seed = new Random();
+        }
+
+        public class SiloBuilderConfigurator : ISiloBuilderConfigurator
+        {
+            public void Configure(ISiloHostBuilder hostBuilder)
+            {
+                hostBuilder.ConfigureServices(services => services.AddTransientNamedService<IControllable, TMGrainLocator>(typeof(TMGrainLocator).Name));
+            }
+        }
+
+        public virtual async Task TransactionWillRecoverAfterRandomSiloFailure(string transactionTestGrainClassName, bool killSiloWhichRunsTm)
+        {
+            const int grainCount = 100;
+            const int opsPerGrain = 2;
+            var txGrains = Enumerable.Range(0, grainCount)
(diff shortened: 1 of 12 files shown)


test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
```diff
@@ -0,0 +1,147 @@
...
+                .ToList();
+            var tpsBeforeInterruption = await RecordTPS(txGrains, opsPerGrain);
+            this.logger.LogInformation($"TPS before interruption is {tpsBeforeInterruption}");
+            if(killSiloWhichRunsTm)
+                await KillSilo(location => location.ActiveInCurrentSilo);
+            else
+            {
+                await KillSilo(location => !location.ActiveInCurrentSilo);
+            }
+            await TestingUtils.WaitUntilAsync(lastTry => CheckTPS(txGrains, tpsBeforeInterruption, opsPerGrain, lastTry), RecoveryTimeout);
```
jason-bragg: Why are we checking TPS? To ensure we recover, we need ensure all transaction operations begin to succeed, so checking for no failures seems more strait forward.  Load tests will vet this behavior under load.

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
```diff
@@ -0,0 +1,147 @@
...
+            var txGrains = Enumerable.Range(0, grainCount)
+                .Select(i => RandomTestGrain(transactionTestGrainClassName))
+                .ToList();
+            var tpsBeforeInterruption = await RecordTPS(txGrains, opsPerGrain);
+            this.logger.LogInformation($"TPS before interruption is {tpsBeforeInterruption}");
+            if(killSiloWhichRunsTm)
+                await KillSilo(location => location.ActiveInCurrentSilo);
+            else
+            {
+                await KillSilo(location => !location.ActiveInCurrentSilo);
```
jason-bragg: Won't this kill all silos that don't have the TM?  We only want to kill one..?

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            const int grainCount = 100;
+            const int opsPerGrain = 2;
+            var txGrains = Enumerable.Range(0, grainCount)
+                .Select(i => RandomTestGrain(transactionTestGrainClassName))
+                .ToList();
+            var tpsBeforeInterruption = await RecordTPS(txGrains, opsPerGrain);
+            this.logger.LogInformation($"TPS before interruption is {tpsBeforeInterruption}");
+            if(killSiloWhichRunsTm)
+                await KillSilo(location => location.ActiveInCurrentSilo);
```
jason-bragg: Should probably vet transactions are functioning correctly prior to killing silo.  For instance, 0 tps (due to all transactions failing) would allow this test to pass, when it should not.

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
xiazen: I disagree. No failures = recovered seems harsh, since aborts is considered normal behavior?

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
xiazen: no.. it only kills one which fits the predicate, which is picked randomly... go read the specific implementation

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
xiazen: see above. I think consider all transaction success as normal is harsh, since aborts should be viewed as normal

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
jason-bragg: Why should aborts be considered a normal behavior?  In cases were there are conflicts, it may (should not be but for now may) be considered normal, but we can structure the transaction so they do not conflict.

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionRecoveryTestsRunner.cs
jason-bragg: Yup, my oversight.

PR #4542, opened 2018-05-01
Fixed incorrect sample in DI docs

### LINKED ISSUE #4532: DI documentation incorrect client sample
This concerns the [Dependency Injection](https://dotnet.github.io/orleans/Documentation/Core-Features/Dependency-Injection.html) documentation page for Orleans 2.0.

There's this code sample that shows how the _client_ should be configured to use DI:

image

It seems to be a direct copy/paste from the sample above which is for the silo. The name of the builder variable was changed, but the rest was left the same.

dandago:
Fixes #4532


ReubenBond:
Thanks, @dandago

PR #4552, opened 2018-05-03
Add Client/SiloHost builder delegate to legacy GrainClient and SIlo/AzureSilo

benjaminpetit:
Since now providers need to be configured using the builder pattern, developer stuck with the legacy Silo and GrainClient can no longer configure thingks like storage, or stream providers.

This PR enable them to configure it so transition is more easy.


benjaminpetit:
@dotnet-bot test functional

Issue #2480, opened 2016-12-03
Does/will Orleans support the new ValueTask<T> awaitable? 


TsengSR:
I was just again going through the Orleans tutorials and when I looked at [this ](http://dotnet.github.io/orleans/Tutorials/A-Service-is-a-Collection-of-Communicating-Actors.html) one I just remembered that during the .NET Core development a new `ValueTask<T>` was introduced

`ValueTask<T>`' has the goal to reduce/avoid allocating of a Task class when returning an value which do not involve calling an async method. 

The example where I immediately thought this may be useful is this one (reduced to get to the point):

```
public class Employee : Grain, IEmployee
{
    public Task<int> GetLevel()
    {
        // allocates a Task
        return Task.FromResult(_level);
    }

    private int _level;
}
```

The line here `return Task.FromResult(_level);` will cause a task to be allocated even if there is no async operation involved. While in theory we could create a task from it and cache it we would have to that every time it changes. 

The interesting thing is, `ValueTask<T>` is a struct and stored on a stack rather than on the heap and when we frequently call code which only returns a value w/o involving async operation, it will lower the preasure on GC. 

The usage with it would be 

```
public class Employee : Grain, IEmployee
{
    public ValueTask<int> GetLevel()
    {
        return new ValueTask(_level);
    }

    private int _level;
}
```

We can even use it, when we have a caching layer and there is a chance async operation involved to reduce the allocation of Task. 

```
public class Employee : Grain, IEmployee
{
    public ValueTask<int> GetLevel()
    {
        int level;

        // if cached return synchronously, otherwise get it from asynchronously 
        if(!cache.TryGetValue("level", out level) 
        {
            // Task will only be allocated when the value is not in cache
            Task task = remoteGrain.GetLevel(...)
                .ContinueWith(_ => cache.AddValue("level", level));

            return new ValueTask(task);
        }

       return new ValueTask(level);
    }
}
```

There is also an excellent post by Bar Arnon on this topic [here](http://blog.i3arnon.com/2015/11/30/valuetask/). 

[...]

PR #4562, opened 2018-05-05
Support of ValueTask as a grain method return type

kutensky:
#2480
Added support of ValueTask as a return type of grain method.
It's a first step of ValueTask support. For now it's just allow users to use ValueTask as a return type to improve their grain performance. Later on we can try to improve our own performance by using ValueTask instead of Task.
Even so, if users start using ValueTask they will have more freedom of writing their code and when we implement full support of ValueTask, they will suddenly fill performance improvement without need of changing their code.


kutensky:
Hi @ReubenBond . I have some strange error in my pull request. There is 2 errors:
1.  DependencyInjection.Tests
It fails before the test starts. Error message is "The following constructor parameters did not have matching fixture data: BaseTestClusterFixture fixture".
But problem is that fails even on master branch. 
2. EHStatisticMonitorTests.
It pass on my location machine.

Could you give any advise?

kutensky:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test this please

kutensky:
@dotnet-bot test functional

xiazen:
@kutensky the error message you posted for DependencyInjection.Tests is not the actual error which caused the test fail to start. I believe there's more inner stack trace which can give us more insight.

xiazen:
and about the EHStatisticMontoring test, it fails intermittently because the test rely on timing a lot. So it failing might be benign. Rerun the test using the command Reuben is talking about, it will likely succeed.

kutensky:
@xiazen, @ReubenBond  Thanks for your help. After restarting tests, it says that all tests passed.

ReubenBond:
Great work @kutensky! Apologies for the delay in merging this

icanhasjonas:
Is there a reason for comparing strings instead of generic types?

```cs
  var same = method.ReturnType.GetGenericTypeDefinition().FullName == "System.Threading.Tasks.Task`1";
```

vs

```cs
  var same = method.ReturnType.GetGenericTypeDefinition() == typeof(Task<>);
```

kutensky:
@icanhasjonas I don't see any.

PR #4566, opened 2018-05-07
Azure table transactional storage for distributed TM

sebastianburckhardt:
New implementation for Azure table transactional storage for distributed TM.
Supports arbitrarily sized batches.

Passes existing transactions tests (NOTE: fails on devstorage, must use actual Azure table).
Testing is somewhat incomplete, we still need more tests to exercise all the recovery situations.

### DIFF
--- a/src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
+++ b/src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
@@ -1,0 +1,296 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage.Table;
+using Newtonsoft.Json;
+using Orleans.Transactions.Abstractions;
+using Orleans.Transactions.AzureStorage;
+
+namespace Orleans.Transactions.DistributedTM.AzureStorage
+{
+    public class AzureTableTransactionalStateStorage<TState> : ITransactionalStateStorage<TState>
+        where TState : class, new()
+    {
+        private readonly CloudTable table;
+        private readonly string partition;
+        private readonly JsonSerializerSettings jsonSettings;
+        private readonly ILogger logger;
+
+        private KeyEntity key;
+        private List<KeyValuePair<long, StateEntity>> states;
+
+        public AzureTableTransactionalStateStorage(CloudTable table, string partition, JsonSerializerSettings JsonSettings, ILogger<AzureTableTransactionalStateStorage<TState>> logger)
+        {
+            this.table = table;
+            this.partition = partition;
+            this.jsonSettings = JsonSettings;
+            this.logger = logger;
+        }
+
+        public async Task<TransactionalStorageLoadResponse<TState>> Load(string stateName)
+        {
+            try
+            {
+                var keyTask = ReadKey();
+                var statesTask = ReadStates();
+                key = await keyTask.ConfigureAwait(false);
+                states = await statesTask.ConfigureAwait(false);
+
+                if (string.IsNullOrEmpty(key.ETag))
+                {
+                    if (logger.IsEnabled(LogLevel.Trace))
+                        logger.LogTrace($"{partition} Loaded v0, fresh");
+
+                    // first time load
+                    return new TransactionalStorageLoadResponse<TState>();
+                }
+                else
+                {
+                    if (!FindState(this.key.CommittedSequenceId, out var pos))
+                    {
+                        var error = $"Storage state corrupted: no record for committed state";
+                        logger.LogCritical(error);
+                        throw new InvalidOperationException(error);
+                    }
+                    var committedState = states[pos].Value.GetState<TState>(this.jsonSettings);
+
+                    var PrepareRecordsToRecover = new List<PendingTransactionState<TState>>();
+                    for (int i = 0; i < states.Count; i++)
+                    {
+                        var kvp = states[i];
+
+                        // pending states for already committed transactions can be ignored
+                        if (kvp.Key <= key.CommittedSequenceId)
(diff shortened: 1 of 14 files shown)


src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,296 @@
...
+                    {
+                        if (FindState(s.SequenceId, out var pos))
+                        {
+                            // overwrite with new pending state
+                            var existing = states[pos].Value;
+                            existing.TransactionId = s.TransactionId;
+                            existing.TransactionTimestamp = s.TimeStamp;
+                            existing.TransactionManager = s.TransactionManager;
+                            existing.SetState(s.State, this.jsonSettings);
+                            batchOperation.Replace(existing);
```
jason-bragg: Under what conditions would a prepared state change?

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,296 @@
...
+                            batchOperation.Insert(entity);
+                            states.Insert(pos, new KeyValuePair<long, StateEntity>(s.SequenceId, entity));
+
+                            if (logger.IsEnabled(LogLevel.Trace))
+                                logger.LogTrace($"{partition}.{s.SequenceId:x16} Insert {entity.TransactionId}");
+                        }
+                    }
+
+            // third, persist metadata and commit position
+            key.Metadata = metadata;
```
jason-bragg: the key serves as a synchronizer, to prevent modification by multiple grains under edge conditions, like duplicate activations or deployments.  Every batch write needs to include the key, even if the key values don't change.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/StateEntity.cs
```diff
@@ -0,0 +1,53 @@
...
+using Orleans.Transactions.Abstractions;
+using System;
+
+namespace Orleans.Transactions.DistributedTM.AzureStorage
+{
+    internal class StateEntity : TableEntity
+    {
+        public static string MakeRowKey(long sequenceId)
+        {
+            return sequenceId.ToString("x16");
```
jason-bragg: Please prefix rowkey and use {RKPrefix} to KRPrefix~ range pattern for min max rowkey.  This is a tested pattern we're trying to standardize on for maintainability reasons.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        public AzureTableTransactionalStateStorage(CloudTable table, string partition, JsonSerializerSettings JsonSettings, ILogger<AzureTableTransactionalStateStorage<TState>> logger)
+        {
+            this.table = table;
+            this.partition = partition;
+            this.jsonSettings = JsonSettings;
+            this.logger = logger;
+        }
+
+        public async Task<TransactionalStorageLoadResponse<TState>> Load(string stateName)
```
jason-bragg: statename should probably be used from the interfaces calls (load, store)

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
```diff
@@ -0,0 +1,296 @@
...
+                {
+                    await table.ExecuteBatchAsync(batchOperation).ConfigureAwait(false);
+                }
+                else
+                {
+                    var portion = new TableBatchOperation();
+                    var enumerator = batchOperation.GetEnumerator();
+                    while (enumerator.MoveNext())
+                    {
+                        portion.Add(enumerator.Current);
```
jason-bragg: nifty.   I didn't know you could move operations from batch to batch! :)

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
sebastianburckhardt: the prepared states are indexed by sequence id (1,2,3,4...). These sequence ids serve as the row key in the table. When a transaction is aborted, the same sequence is reused by a different transaction.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
sebastianburckhardt: I don't prefer including statename... cleaner to specify it in the factory, which eliminates the ugly ConcurrentDictionary inside the provider.

I actually thought you already came to the same conclusion in #4538 - you removed statename from ITransactionalStateStorage, and removed the ConcurrentDictionary.

sebastianburckhardt: Looks like I did not consistently update my code to reflect the latest pattern. Will do that.

src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/DistributedTM/AzureTableTransactionalStateStorage.cs
sebastianburckhardt: Yeah. But I'll have to rewrite this part to make it easier to add the key to each batch.

PR #4567, opened 2018-05-08
Fix tests on partial builds

sergeybykov:
Bump up version of Azure extension packages and fix versions of some references.


sergeybykov:
@dotnet-bot test functional

PR #4568, opened 2018-05-08
Transactions simplified to use distributed transaction manager only.

jason-bragg:
Removed singleton transaction manager and abstractions related to that pattern.
Still work in progress, as it should be rebased on Sebastian's transaction storage work, once that goes in.

Removed dynamoDb transaction work because there was nothing left once log storage was removed.  Related assemblies and tests are still under source control and can be brought back if/when transactional storage impl is introduced.


src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -230,9 +230,8 @@ internal static void AddDefaultServices(HostBuilderContext context, IServiceColl
             services.AddFromExisting<IKeyedSerializer, ILBasedSerializer>();
             
             // Transactions
-            services.TryAddSingleton<ITransactionAgent, TransactionAgent>();
             services.TryAddSingleton<Factory<ITransactionAgent>>(sp => () => sp.GetRequiredService<ITransactionAgent>());
-            services.TryAddSingleton<ITransactionManagerService, DisabledTransactionManagerService>();
+            //services.TryAddSingleton<ITransactionAgent, DisabledTransactionAgent>();
```
sebastianburckhardt: is this comment still needed? Seems that the DisabledTransactionAgent is gone.

src/Orleans.Runtime/Transactions/ITransactionManagerService.cs
```diff
@@ -1,16 +1,8 @@
...
 {
-    public interface ITransactionManagerService
-    {
-        Task<StartTransactionsResponse> StartTransactions(List<TimeSpan> timeouts);
-        Task<CommitTransactionsResponse> CommitTransactions(List<TransactionInfo> transactions, HashSet<long> queries);
-        Task AbortTransaction(long transactionId, OrleansTransactionAbortedException reason);
-    }
-
     [Serializable]
     public struct CommitResult
```
sebastianburckhardt: Are the structs in this file still used somewhere? I can't find any occurrences.

jason-bragg:
@dotnet-bot test this please

jason-bragg:
@jsteinich At a high level, the distributed TM pattern picks a grain involved in the transaction to act as the TM for that transaction,  We don't have any design documentation for this (as far as I know) as it was developed in research.

The distributed TM logic was check in previously in "draft of distributed TM implementation #3820".  May be easyer to look at that PR when familiarizing yourself.  This PR mostly just pulls out the singleton TM and make the distributed tech the only one we support.

xiazen:
I see 369 bvt failing, while majority of them isn't tx related. WHat's going on? are they pass locally?

jason-bragg:
No, It's a bug.  Fixing.  Transaction agent is only added in transaction tests, but requested by all.  Adding default agent.

jason-bragg:

PR #4573, opened 2018-05-09
Fix Dependency Injection without changing Abstractions project

ReubenBond:


ReubenBond:
This works by replacing the constructorless registration of `StatelessWorkerPlacement` in the container with one which specifies the constructor (by using a lambda) and therefore StructureMap is happy with it. I would have provided a lambda to the original registration, but that requires changing a lot more code, since currently keyed services does not support that.

jason-bragg (approved):

PR #4581, opened 2018-05-11
Clock skew transaction testing

jason-bragg:
Ensure transaction timestamps work correctly even with clock skew and system time changes.


src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
```diff
@@ -13,40 +13,22 @@ namespace Orleans.Transactions
...
-        public TransactionAgent(ILoggerFactory loggerFactory, ILocalSiloDetails siloDetails)
-        {
-            logger = loggerFactory.CreateLogger(this.GetType().Name);
-        }
+        private readonly ILogger logger;
+        private readonly Stopwatch stopwatch = Stopwatch.StartNew();
+        private readonly TransactionClock clock;
 
-        public Task Start()
+        public TransactionAgent(IClock clock, ILoggerFactory loggerFactory, ILocalSiloDetails siloDetails)
```
xiazen: logger to be in the constructor?

src/Orleans.Transactions/DistributedTM/IClock.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace Orleans.Transactions
+{
+    /// <summary>
+    /// System clock abstraction
+    /// </summary>
+    public interface IClock
```
xiazen: if `IClock` is introduced just for testing purpose , suggesting make it internal, and make the test project do internal visible to. If not, I'd like to understand the intention of the new public interface

sebastianburckhardt:
I think we can get a bit more teeth out of these tests by using the injected "bad" clock not just in the TransactionAgent, but also in the TransactionalState. I will submit PR.

src/Orleans.Transactions/DistributedTM/TransactionClock.cs
```diff
@@ -0,0 +1,29 @@
+﻿
+using System;
+
+namespace Orleans.Transactions
+{
+    public class TransactionClock
```
sebastianburckhardt: I think "CausalClock" would be a better name since it describes what this thing provides, not just who uses it.

src/Orleans.Transactions/DistributedTM/IClock.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        /// <summary>
+        /// Current time in utc
+        /// </summary>
+        /// <returns></returns>
+        DateTime UtcNow();
```
sebastianburckhardt: To keep this closer to DateTime.UtcNow you may consider declaring it as a property `DateTime UtcNow { get; }`. Obviously, this is not important. Also perhaps that would be bad style (a getter with side effects).

test/Transactions/Orleans.Transactions.Tests/Runners/SkewedClock.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public class SkewedClock : IClock
+    {
+        private readonly Random rand = new Random(Guid.NewGuid().GetHashCode());
+        private readonly TimeSpan minSkew;
+        private readonly int skewRangeTicks;
+
+        public SkewedClock(TimeSpan minSkew, TimeSpan maxSkew)
+        {
+            this.minSkew = maxSkew;
```
sebastianburckhardt: shouldn't this be `this.minSkew = minSkew`?

src/Orleans.Transactions/DistributedTM/IClock.cs
jason-bragg: Testability is a design goal.  Having a clock abstraction as part of the public surface of the implementation violates no design principles I am aware of, and improves testability.  This is a common problem for components with dependencies on time.  There is not (to my knowledge) a common pattern for addressing this, but many systems do use clock abstractions for testability.

xiazen: hmm having abstraction is fine. Having a public abstraction just for testing isn't common to my understanding. If it is just for testing, can we keep it as a internal interface? this is what I meant by my previous comment.

src/Orleans.Transactions/DistributedTM/TransactionClock.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        private readonly Object lockable = new object();
```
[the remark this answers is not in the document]
jason-bragg: Usual rule (by my understanding) is for the owner of the vulnerable state to be responsible for ensuring synchronized access to it.  In this case, this is the TransactionClock.  We could assume that the clock has state from the TransactionAgent, and have it lock the entire clock, but that requires that system make assumptions about the TransactionClock impl, which is unnecessary and error prone.

src/Orleans.Transactions/DistributedTM/IClock.cs
jason-bragg: Will leave as is, for the reasons you stated.  We get very ittle from making this a property.

test/Transactions/Orleans.Transactions.Tests/Runners/SkewedClock.cs
jason-bragg: Doht!?

PR #4592, opened 2018-05-15
Fix tx recovery test intermittent failure

xiazen:
- removed unused testing method for old singleTM
- change stop silo method from ungracefully shutdown to gracefully shutdown. Recovery tests was failed before because silos are shutdown ungracefully, and alive silos somehow still keep sending those requests to the dead silo and get time out exception back. So I think it is failing because clustering isn't recovering from ungracefully shutdown, instead of tx isn't recovering. So to tests tx recovery in a clean box, I think shutdown silo gracefully is the right way. In this way, tx will be the only one which need to recovery, the test won't be testing clustering recovery.


xiazen:
another way could be using azure clustering which is more stable than grain based clustering. but still we will be also partially testing azure clustering recovery

PR #4595, opened 2018-05-15
Transaction concurrency tests

jason-bragg:
- failing against memory storage, unclear why.  Tests disabled, will open bug

### DIFF
--- a/test/Transactions/Orleans.Transactions.Azure.Test/TransactionConcurrencyTests.cs
+++ b/test/Transactions/Orleans.Transactions.Azure.Test/TransactionConcurrencyTests.cs
@@ -10,6 +10,7 @@ namespace Orleans.Transactions.AzureStorage.Tests
         public TransactionConcurrencyTests(TestFixture fixture, ITestOutputHelper output)
             : base(fixture.GrainFactory, output)
         {
+            fixture.EnsurePreconditionsMet();
         }
     }
 }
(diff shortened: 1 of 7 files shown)


test/Transactions/Orleans.Transactions.Azure.Test/TransactionConcurrencyTests.cs
```diff
@@ -0,0 +1,15 @@
...
+using Orleans.Transactions.Tests;
+
+namespace Orleans.Transactions.AzureStorage.Tests
+{
+    [TestCategory("Azure"), TestCategory("Transactions"), TestCategory("Functional")]
+    public class TransactionConcurrencyTests : TransactionConcurrencyTestRunner, IClassFixture<TestFixture>
+    {
+        public TransactionConcurrencyTests(TestFixture fixture, ITestOutputHelper output)
+            : base(fixture.GrainFactory, output)
+        {
```
xiazen: I think you aren't skiiping azure tests correctly. those are still run by Jenkins and failed

jason-bragg: This was by design.  I only intended to skip the memory test, because those were the only one's failing locally.  Azure tests passed for me.  This simplifies things, as I don't need to do this somewhat kludgy skip method, but rather just mark all the test as skip.  Will change that.

Bumbed these tests aren't passing.  Had hoped all was well and there was just some issue with the memory storage.

jason-bragg: Oh, I'm dork.. I see what you mean now.. NM

PR #4596, opened 2018-05-16
Remove unused tx metrics

xiazen:
TransactionsStatisticsGroup.cs  was used by single TM and is currently not used in the system. delete to clean the code base up


ReubenBond:
Hopefully we have some other way to collect useful tx metrics

xiazen:
guys don't just chat on my pr, your job here is to merge it.  :)

PR #4598, opened 2018-05-16
Fix bug where Issue is never assigned and retry delays never happen

jms69:
The class is RecordedConnectionIssue never assigns the Issue property which causes DelayBeforeRetry() to never delay.


ReubenBond:
@sebastianburckhardt this looks good to me. If you feel likewise then we can merge it.

PR #4603, opened 2018-05-17
Test MetaData can be (de)serialzied

xiazen:


xiazen:
test isn't passing because of serialization issues. so don't merge yet. need to fix that still

xiazen:
@sebastianburckhardt  this is the serialization issue I was talking about. MetaData is (de)serialized in StorageBatch, somehow the Deserialize doesn't work. This test reveals it. If you have an idea how to fix it, that would be great ! I tried suggested ways from the internet to serialize a class which contains a list of interface objects, but haven't found a way to fix this thing.

ReubenBond:
The issue with the `MetaData` (could we please fix the capitalization on this name?) deserialization is the custom converter in `TransactionParticipantExtensionExtensions.GetJsonSerializerSettings`

This is one way of 'fixing' it:
```C#
public static JsonSerializerSettings GetJsonSerializerSettings(ITypeResolver typeResolver, IGrainFactory grainFactory){
    var serializerSettings = OrleansJsonSerializer.GetDefaultSerializerSettings(typeResolver, grainFactory);
    serializerSettings.TypeNameHandling = TypeNameHandling.Auto;
    serializerSettings.PreserveReferencesHandling = PreserveReferencesHandling.None;
    serializerSettings.Converters.Add(new TransactionParticipantExtensionWrapper.CustomJsonConverter(grainFactory));
    serializerSettings.ContractResolver = serializerSettings.ContractResolver ?? new DefaultContractResolver();
    var contract = serializerSettings.ContractResolver.ResolveContract(typeof(ITransactionParticipant));
    contract.DefaultCreator = () => FormatterServices.GetUninitializedObject(typeof(TransactionParticipantExtensionWrapper));
    return serializerSettings;
}
```

Another way is to not use the custom json converter, but I'm not sure what's best

xiazen:
I think custom serializer are there for perf concerns ? but I will let @sebastianburckhardt speak for his reasons on the custom serializer for` Metadata `

sebastianburckhardt:
The custom Json converter was just a way to make it work... the algorithm needs to persist grain references in both the commit records and the prepare records, and I did not know how to do that in any other way. Feel free to substitute any other serialization that you find appropriate and that works.

ReubenBond:
Ok, it seems that we can remove the custom converter, then, since GrainReferences will be supported by virtue of the converter added by OrleansJsonSerializer. Worth testing.

PR #4608, opened 2018-05-18
Clean up and fix abort logic

sebastianburckhardt:
- remove unnecessary control flow complications and unnecessary calls to GetTransactionContext
- capture original exception in transaction context, avoid relying on grain response and serialization


sebastianburckhardt:
I got this to work as desired now.

src/Orleans.Transactions/Abstractions/TransactionInfo.cs
```diff
@@ -86,33 +85,57 @@ public void Join(ITransactionInfo x)
...
         }
 
+        public OrleansTransactionAbortedException MustAbort(SerializationManager sm)
+        {
+
+            if (OriginalException != null)
+            {
+                var reader = new BinaryTokenStreamReader(OriginalException);
+                return sm.Deserialize<OrleansTransactionAbortedException>(reader);
+            }
```
jason-bragg: Why serialize this yourself?  Doesn't codegen do this for you if you have a field "Exception OriginalException;"?

src/Orleans.Core/Transactions/ITransactionAgent.cs
```diff
@@ -22,16 +22,13 @@ public interface ITransactionAgent
...
-        /// <returns>None.</returns>
-        /// <remarks>
-        /// The method throws OrleansTransactionInDoubtException if the outcome of the Commit cannot be determined.
-        /// If any other exception is thrown then the transaction is aborted.
-        /// </remarks>
-        Task Commit(ITransactionInfo transactionInfo);
+        /// <returns>null if the transaction committed successfully, or an exception otherwise.
+        /// If the exception is OrleansTransactionInDoubtException, it means the outcome of the Commit cannot be determined; otherwise,
+        /// the transaction is guaranteed to not have taken effect.</returns>
+        Task<OrleansTransactionException> Commit(ITransactionInfo transactionInfo);
```
jason-bragg: Why have an exception as a return value, this seems very odd and unorthodox.  How is receiving an exception different from the call throwing an exception.

src/Orleans.Core/Transactions/ITransactionInfo.cs
```diff
@@ -13,9 +16,18 @@ public interface ITransactionInfo
         string Id { get; }
 
         /// <summary>
-        /// Indicates that the transaction has aborted.
+        /// Record an unhandled exception thrown by a grain call. This causes the transaction to abort.
         /// </summary>
-        bool IsAborted { get; set; }
+        /// <param name="e">The unhandled exception that was thrown by the grain call</param>
+        /// <param name="serializationManager">The serialization manager used for serializing the exception</param>
+        void RecordException(Exception e, SerializationManager serializationManager);
```
jason-bragg: Why do we need to know the reason for the abort in the TransactionInfo?  Wasn't IsAborted sufficient?

src/Orleans.Core/Transactions/ITransactionAgent.cs
ReubenBond: Exceptions only capture their stack when they are thrown, so returning an exception avoids that.

I would very much prefer we return a result type (some struct or class) which represents the commit rather than an exception

src/Orleans.Transactions/Abstractions/TransactionInfo.cs
sebastianburckhardt: You would think so... I tried that first. But got weird failure from Orleans serializer. What you see here represents a workaround.

ReubenBond: When issues like this pop up, please feel free to message in Gitter/email/IM - we can help

src/Orleans.Core/Transactions/ITransactionInfo.cs
sebastianburckhardt: The point is that the user needs to know why a transaction aborted - so it can take appropriate action. For example, if the exception is caused by a timed out lock, it makes sense to retry the transaction... but it the user code throws an unexpected exception (e.g. IndexOutOfBounds) or expected exception (e.g. InsufficientFundsException), retrying is not what you want.

The reason why I felt this needed to be improved is that while implementing my general-purpose randomized serializability testing, I noticed that the exceptions returned are (1) much too big and bulky because of frequent re-throwing, and (2) terribly nested, basically unusable, if the transactions themselves are nested.

The solution I am proposing here guarantees that (no matter how much nesting or re-throwing happens) the user will always see either (1) a OrleansTransactionAbortedException with InnerException=(exception thrown by user code that caused the abort), if user code was the culprit, or (2) one of the specific exception classes describing the problem, if some issue within the transaction runtime was the culprit.

src/Orleans.Core/Transactions/ITransactionAgent.cs
sebastianburckhardt: I will change this to return TransactionalStatus instead of an exception.

src/Orleans.Core/Transactions/ITransactionInfo.cs
jason-bragg: Understood.  I'm not convinced this is a good way to address the error handling, but for the general abort logic I think this PR goes in the right direction.

Reuben, Would you object to letting the PR go in with a work item to revisit the exception handling, in a future pass?

I'm suggesting this becuase, while I don't agree with the current pattern, we should be able to revisit it with minimal effect to the application layer, the solution is non-trivial, and cleaning up this abort logic can help us narrow on the remaining issues.

In short, I'd rather get this improvement in then clean it up as opposed to holding it on how some (mostly) internal details are handled.

sebastianburckhardt:
Ok, I revised this a  little.

1. the return type of Commit is now an enum, not an exception
2. I removed a few unneeded exception classes that were left over from the old algorithm
3. I added an exception class OrleansTransactionTransientFailureException that groups all the exceptions representing a failure caused by some hopefully transient effect (timeouts, cascading aborts, ...) that may go away when retried

src/Orleans.Core/Transactions/ITransactionInfo.cs
ReubenBond: Sounds fine to me

jason-bragg:
@sebastianburckhardt, I think this is ready to merge, once conflicts are resolved.

PR #4616, opened 2018-05-22
Tx fixes

sebastianburckhardt:
Several bug fixes for bugs that caused hangs in my randomized tests

- fix bug in commit logic that causes hangs in some situations
- fix 2 bugs in azure table transactional state storage (one in tracing, one causing hangs)
- make TA notify participants of aborts for read-only transactions

also, I made tracing in azure table transactional state storage more uniform.

### DIFF
--- a/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
+++ b/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
@@ -93,6 +93,9 @@ namespace Orleans.Transactions
                         if (logger.IsEnabled(LogLevel.Debug))
                             logger.Debug($"{stopwatch.Elapsed.TotalMilliseconds:f2} fail {transactionInfo.TransactionId} prepare response status={status}");
 
+                        foreach (var p in participants)
+                            p.Key.Abort(transactionInfo.TransactionId).Ignore();
+
                         throw status.ConvertToUserException(transactionInfo.TransactionId.ToString());
                     }
                 }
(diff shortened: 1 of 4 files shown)


src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
```diff
@@ -102,6 +105,9 @@ public Task<ITransactionInfo> StartTransaction(bool readOnly, TimeSpan timeout)
                 if (logger.IsEnabled(LogLevel.Debug))
                     logger.Debug($"{stopwatch.Elapsed.TotalMilliseconds:f2} timeout {transactionInfo.TransactionId} prepare responses");
 
+                foreach(var p in participants)
+                    p.Key.Abort(transactionInfo.TransactionId).Ignore();
```
jason-bragg: Can the agent make this call?  Isn't TM authoritative?  What if tm has committed transaction the, but this times out?

src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
```diff
@@ -93,6 +93,9 @@ public Task<ITransactionInfo> StartTransaction(bool readOnly, TimeSpan timeout)
                         if (logger.IsEnabled(LogLevel.Debug))
                             logger.Debug($"{stopwatch.Elapsed.TotalMilliseconds:f2} fail {transactionInfo.TransactionId} prepare response status={status}");
 
+                        foreach (var p in participants)
+                            p.Key.Abort(transactionInfo.TransactionId).Ignore();
```
jason-bragg: Shouldn't we wait for the aborts?  If not, won't this allow future calls to fail because they may depend on the aborted operation.

src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
sebastianburckhardt: This is the logic for read-only transaction. For those, there is no TM, instead the TA "commits" the transaction after it receives all the prepare responses. These responses are sent by the participants after they unlock, and after they persist the clock.

jason-bragg:
@dotnet-bot test Ubuntu please

jason-bragg:
@dotnet-bot test Ubuntu16.04 Release Build please

PR #4623, opened 2018-05-24
Convert counter values before calling ITelemetryProducer.TrackMetric

### LINKED ISSUE #4607: SerializationManager measures ticks, not milliseconds
The serialization manager measures the ticks: https://github.com/dotnet/orleans/blob/e87f0e458d36dccf145de06295475831331ff0bf/src/Orleans.Core/Serialization/SerializationManager.cs#L609

But the counter says: Serialization.Body.DeepCopy.Milliseconds.Current

ReubenBond:
Fixes #4607

cc @SebastianStehle are you able to verify this?


ReubenBond:
I locally verified this has the intended effect using the Orleans Dashboard as my `ITelemetryProducer`

PR #4628, opened 2018-05-25
Cleanup pass on Transaction abstractions and public surface.

jason-bragg:
- Removed leftover code from singleton TM.
- Moves agent abtraction from core to runtime abstractions
- Cleaned up agent initialization and usage pattern
- Reduced public surface of transactional state abstraction
- Moved transaction info out of abstractions
- Removed old unsused logging errorcodes.
- Moved serialization generation assembly statements to files that defined the serializable classes.


src/Orleans.Transactions/ErrorCodes.cs
```diff
@@ -1,61 +1,15 @@
...
 {
     /// <summary>
-    /// Orleans ServiceBus error codes
+    /// Orleans Transactions error codes
     /// </summary>
     internal enum OrleansTransactionsErrorCode
     {
         /// <summary>
-        /// Start of orlean servicebus errocodes
+        /// Start of orlean transactions errocodes
```
ReubenBond: Orleans
error codes

src/Orleans.Transactions/ErrorCodes.cs
```diff
@@ -1,61 +1,15 @@
...
-        internal static void Warn(this ILogger logger, OrleansTransactionsErrorCode errorCode, string message, Exception exception)
-        {
-            logger.Warn((int)errorCode,  message, new object[] { }, exception);
-        }
-
-        internal static void Error(this ILogger logger, OrleansTransactionsErrorCode errorCode, string message, Exception exception = null)
-        {
-            logger.Error((int)errorCode, message, exception);
-        }
+        // TODO - jbragg - add error codes for transaction errors - User Story 62
```
ReubenBond: IMO we should add a link to the story in VSO - it's easy to forget where User Story 62 lives

src/Orleans.Transactions/Abstractions/ITransactionalStateStorage.cs
```diff
@@ -1,7 +1,11 @@
-﻿using Orleans.Concurrency;
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Orleans.CodeGeneration;
+using Orleans.Concurrency;
+
+[assembly: GenerateSerializer(typeof(Orleans.Transactions.Abstractions.PendingTransactionState<>))]
+[assembly: GenerateSerializer(typeof(Orleans.Transactions.Abstractions.TransactionalStorageLoadResponse<>))]
```
ReubenBond: I prefer assembly-level attributes live in a file called AssemblyInfo.cs which only contains assembly-level metadata

src/Orleans.Transactions/ErrorCodes.cs
jason-bragg: Tasking is a mess right now, will remove reference to specific task as odds of link or task remaining the same by the time we get to fixing this is fairly weak.

PR #4638, opened 2018-05-31
Transaction load shedding based on txStartedPerSecond

xiazen:
Transaction load shedding based on txStartedPerSecond. 
Configurable through TransactionLoadShuddingOptions

### DIFF
--- a/src/Orleans.Core/Transactions/OrleansTransactionException.cs
+++ b/src/Orleans.Core/Transactions/OrleansTransactionException.cs
@@ -57,6 +57,18 @@ namespace Orleans.Transactions
     }
 
     /// <summary>
+    /// Signifies that transaction runtime is overloaded
+    /// </summary>
+    [Serializable]
+    public class OrleansTransactionOverloadException : OrleansTransactionException
+    {
+        public OrleansTransactionOverloadException()
+            : base("Transaction is overloaded on current silo, please try again later.")
+        {
+        }
+    }
+
+    /// <summary>
     /// Signifies that the runtime is unable to determine whether a transaction
     /// has committed.
     /// </summary>
--- a/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
+++ b/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
@@ -31,9 +31,9 @@ namespace Orleans.Transactions
             this.monitor = new PeriodicAction(options.Value.PerfCountersWriteInterval, this.ReportMetrics);
         }
 
-        public void TryReportMetrics()
+        public void TryReportMetrics(DateTime timestamp)
         {
-            this.monitor.TryAction(DateTime.UtcNow);
+            this.monitor.TryAction(timestamp);
         }
 
         private void ReportMetrics()
@@ -61,9 +61,9 @@ namespace Orleans.Transactions
         private readonly Stopwatch stopwatch = Stopwatch.StartNew();
         private readonly CausalClock clock;
         private readonly TransactionAgentStatistics statistics;
-        private readonly TransactionOverloadDetector overloadDetector;
+        private readonly ITransactionOverloadDetector overloadDetector;
 
-        public TransactionAgent(IClock clock, ILogger<TransactionAgent> logger, TransactionAgentStatistics statistics, TransactionOverloadDetector overloadDetector)
+        public TransactionAgent(IClock clock, ILogger<TransactionAgent> logger, TransactionAgentStatistics statistics, ITransactionOverloadDetector overloadDetector)
         {
             this.clock = new CausalClock(clock);
             this.logger = logger;
@@ -73,7 +73,7 @@ namespace Orleans.Transactions
 
         public Task<ITransactionInfo> StartTransaction(bool readOnly, TimeSpan timeout)
         {
-            this.statistics.TryReportMetrics();
+            this.statistics.TryReportMetrics(DateTime.UtcNow);
             if (overloadDetector.IsOverloaded())
                 throw new OrleansStartTransactionFailedException(new OrleansTransactionOverloadException());
 
--- a/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
+++ b/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
@@ -5,11 +5,52 @@ using System.Threading.Tasks;
 using Orleans.Runtime;
 using Orleans.Concurrency;
 using System.Linq;
+using System.Threading;
+using System.Transactions;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Orleans.Configuration;
 using Orleans.Transactions.Abstractions;
 
 namespace Orleans.Transactions
 {
+    public class TransactionAgentStatistics
+    {
+        public const string TransactionStartedPerSecondMetric = "TransactionAgent.TransactionStartedPerSecond";
+        //TransactionStartedPerSecond for last reporting period
+        public double TransactionStartedPerSecond { get; private set; }
+
+        public long TransactionStartedCounter { get; set; }
+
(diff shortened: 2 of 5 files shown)


src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
```diff
@@ -5,11 +5,52 @@
...
+        {
+            this.monitor.TryAction(DateTime.UtcNow);
+        }
+
+        private void ReportMetrics()
+        {
+            var now = DateTime.UtcNow;
+            var txStarted = TransactionStartedCounter;
+            var timelapseSinceLastReport = now - this.LastReportTime;
+            if (timelapseSinceLastReport.TotalSeconds <= 1)
```
jason-bragg: Suggest using (txStarted * 1000) / by time delta in ticks rather than seconds, in case sub-second report windows are configured.  It's also more accurate.

src/Orleans.Core/Transactions/OrleansTransactionException.cs
```diff
@@ -56,6 +56,18 @@ public OrleansStartTransactionFailedException(SerializationInfo info, StreamingC
         }
     }
 
+    /// <summary>
+    /// Signifies that transaction runtime is overloaded
+    /// </summary>
+    [Serializable]
+    public class OrleansTransactionOverloadException : OrleansTransactionException
```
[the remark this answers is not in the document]
xiazen: OrleansTransactionOverloadException is wrapped in a `OrleansStartTransactionFailedException`. So I'm not trying to make OrleansTransactionOverloadException a major exception which describe a behavior. I just make it a specialization under `OrleansStartTransactionFailedException`

src/Orleans.Transactions/DistributedTM/TransactionOverloadDetector.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            if (!this.options.Enabled)
+                return false;
+
+            this.monitor.TryAction(DateTime.UtcNow);
+            var txPerSecondInLastReportingPeriod = transactionStartedPerSecond;
+            var sinceLastReport = DateTime.UtcNow - lastCheckTime;
+
+            double txPerSecondCurrently;
+            if (sinceLastReport.TotalSeconds <= 1)
```
jason-bragg: Calculating transactions per second given transactions started at last check, transactions started now, and delta between now an last check is common logic that can/should be in shared private function.  Would be less code and consistent in case of change.

src/Orleans.Transactions/DistributedTM/TransactionOverloadDetector.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Orleans.Transactions
+{
+    public interface ITransactionOverloadDetector
```
jason-bragg: Not fond of OverloadDetector name, but consistency is more important than whether I like the name.

In this case the interface, implementation, and options are inconsistently named.  The TransactionOverloadDetector is a specialization of the interface, and name needs reflect the specialization's behavior (which it does not) and the options are associated with the specialization and should reflect it's naming, which it does not.

src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
```diff
@@ -18,21 +60,29 @@ internal class TransactionAgent : ITransactionAgent
         private readonly ILogger logger;
         private readonly Stopwatch stopwatch = Stopwatch.StartNew();
         private readonly CausalClock clock;
+        private readonly TransactionAgentStatistics statistics;
+        private readonly TransactionOverloadDetector overloadDetector;
 
-        public TransactionAgent(IClock clock, ILogger<TransactionAgent> logger)
+        public TransactionAgent(IClock clock, ILogger<TransactionAgent> logger, TransactionAgentStatistics statistics, TransactionOverloadDetector overloadDetector)
```
jason-bragg: TransactionAgent should only reference the abstraction, not the concrete class.

src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
```diff
@@ -5,11 +5,53 @@
...
+        private readonly ITelemetryProducer telemetryProducer;
+        private readonly PeriodicAction monitor;
+        public TransactionAgentStatistics(ITelemetryProducer telemetryProducer, IOptions<StatisticsOptions> options)
+        {
+            this.telemetryProducer = telemetryProducer;
+            this.lastReportTime = DateTime.UtcNow;
+            this.monitor = new PeriodicAction(options.Value.PerfCountersWriteInterval, this.ReportMetrics);
+        }
+
+        public void TryReportMetrics()
```
jason-bragg: should probably take a timestamp for testability.

src/Orleans.Transactions/DistributedTM/TransactionOverloadDetector.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    /// <summary>
+    /// Options for load shedding based on transaction rate 
+    /// </summary>
+    public class TransactionRateLoadSheddingOptions
+    {
+        /// <summary>
+        /// whether to turn on transaction load shedding. Default to false;
+        /// </summary>
+        public bool Enabled { get; set; } = true;
```
jason-bragg: y default to true?

src/Orleans.Transactions/DistributedTM/TransactionOverloadDetector.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    internal class TransactionOverloadDetector : ITransactionOverloadDetector
+    {
+        private readonly TransactionAgentStatistics statistics;
+        private readonly TransactionRateLoadSheddingOptions options;
+        private readonly PeriodicAction monitor;
+        private long transactionStartedAtLastCheck;
+        private double transactionStartedPerSecond;
+        private DateTime lastCheckTime;
+        private static readonly TimeSpan MetricsCheck = TimeSpan.FromSeconds(30);
```
jason-bragg: Should this be configurable?

src/Orleans.Core/Transactions/OrleansTransactionException.cs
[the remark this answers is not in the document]
xiazen: they would receive a OrleansStartTransactionFailedException(OrleansTransactionOverloadException) . so based on the inner exception, they know they are being throttled?

src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
xiazen: will fix this

src/Orleans.Transactions/DistributedTM/TransactionOverloadDetector.cs
xiazen: should be false

src/Orleans.Core/Transactions/OrleansTransactionException.cs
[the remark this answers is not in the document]
jason-bragg: Overload exception as an inner exception of a start failure exception is fine.  As long as there is a means of a client detecting this in a clear way.  Think this is fine.

src/Orleans.Transactions/DistributedTM/TransactionOverloadDetector.cs
xiazen: that depends on the design. I was intending to add more algorithm to TransactionOverloadDetector. So it will include more logic/algorithm in the future. Then in this case, calling it TransactionOverloadDetector  is consistent. But if you want to have multiple overload detector, every of which just focus on one type of load shedding , then maybe we should rename it... Not sure we want to go that route, we can discuss

Issue #4648, opened 2018-06-04
Needlessly controversial terminology


palpha:
https://github.com/dotnet/orleans/blob/cb7b101ee4847c865baefb31a78dcc6461dcf9f2/src/Orleans.Runtime/MembershipService/MembershipOracle.cs#L651
https://github.com/dotnet/orleans/blob/cb7b101ee4847c865baefb31a78dcc6461dcf9f2/src/Orleans.Runtime/MembershipService/MembershipOracle.cs#L682
https://github.com/dotnet/orleans/blob/cb7b101ee4847c865baefb31a78dcc6461dcf9f2/src/Orleans.Runtime/MembershipService/MembershipOracle.cs#L1010

There's no need to use the phrase "commit suicide". Self-destruct would be an okay euphemism, but "will shut down" feels like a perfectly good, neutral way to put it.

Feel free to avoid making any comments about "political correctness" here. ❤️

veikkoeeva:
@palpha As a related note, there has been discussion of using `EventId` and semantic logging in general. Maybe some of this come into consideration if/when there is such refactoring.

I'll pester @jdom a bit here since he was actively discussing logging ideas not long ago in Gitter. :)

sergeybykov:
Resolved via #4649.

PR #4649, opened 2018-06-05
Revised terminology. (#4648)

palpha:


sergeybykov:
@palpha Thank you for taking an initiative!

I suggest instead of "will self-destruct" we simply say "will stop".

sergeybykov:
@palpha Nothing to apologize for here. I totally agree - this king of details matter. Thank you for your contribution!

Issue #4666, opened 2018-06-07
#4479 broke Silo backward compatibility


benjaminpetit:
#4479 introduced a breaking change in the wire protocol between silos.

Not sure if it could be fixed easily, or if we could just revert this change.

@dVakulen?

PR #4479, opened 2018-04-16
 Convert IMembershipTableGrain into a SystemTarget

dVakulen:
Addresses https://github.com/dotnet/orleans/issues/3129.

PR #4668, opened 2018-06-08
Fix wire backward compatibility

dVakulen:
Addresses https://github.com/dotnet/orleans/issues/4666.

### DIFF
--- a/src/Orleans.Core.Abstractions/IDs/GrainId.cs
+++ b/src/Orleans.Core.Abstractions/IDs/GrainId.cs
@@ -48,6 +48,11 @@ namespace Orleans.Runtime
             return FindOrCreateGrainId(key);
         }
 
+        internal static GrainId GetSystemGrainId(Guid guid)
+        {
+            return FindOrCreateGrainId(UniqueKey.NewKey(guid, UniqueKey.Category.SystemGrain));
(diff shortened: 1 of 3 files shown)


src/Orleans.Core.Abstractions/IDs/GrainId.cs
```diff
@@ -48,6 +48,11 @@ internal static GrainId GetGrainId(UniqueKey key)
             return FindOrCreateGrainId(key);
         }
 
+        internal static GrainId GetSystemGrainId(Guid guid)
```
benjaminpetit: If I am not mistaken, it will introduce a breaking change in Abstaction, and we would like to avoid that... Could we move this helper method somewhere else?

dVakulen: That is just partial revert of changes done at https://github.com/dotnet/orleans/pull/4479/files#diff-6f8ed860e2abcd355cb00107fd1653b2L51.

benjaminpetit: Sorry I missed that!

sergeybykov:
Thank you, @dVakulen!

PR #4669, opened 2018-06-08
Tutorial on Custom Grain Storage for 2.0

Kimserey:
This is a tutorial on how to create a custom grain storage for 2.0.

It is taking as base the previous document on storage provider 1.5, but updated to follow the current grain storage implementation found in Azure persistence and AWS persistence

It is explaining implementation of `IGrainStorage` and `ILifecycleParticipant<ISiloLifecycle>`. It also explains how to create the factory and the extensions in order to make the grain storage registration available on the `ISiloHostBuilder`.

_I am not sure if this is the right place to merge, could you please advise and I will move the tutorial. Thanks!_

### DIFF
--- a/src/Tutorials/Custom-Grain-Storage.md
+++ b/src/Tutorials/Custom-Grain-Storage.md
@@ -91,6 +91,8 @@ The  create a constructor containing two fields, `storageName` to specify which 
 
 `IGrainFactory`, `ITypeResolver` and `JsonSerializerSettings` will be used in the next section where we will initilize the storage.
 
+We also take two options as argument, our own `FileGrainStorageOptions` and the `ClusterOptions`. Those will be needed for the implementation of the storage functionalities.
+
 ## Initializing the storage
 
 To initialize the storage, we register an `Init` function on the `ApplicationServices` lifecycle.
@@ -118,15 +120,24 @@ private Task Init(CancellationToken ct)
 }
 ```
 
+We also provide a common function to construct the filename ensuring uniqueness per service, grain Id and grain type.
+
+```csharp
+private string GetKeyString(string grainType, GrainReference grainReference)
+{
+    return $"{_clusterOptions.ServiceId}.{grainReference.ToKeyString()}.{grainType}";
+}
+```
+
 ## Reading State
 
-To read the state, to ensure uniqueness, we construct the filename using the grain key and the grain type.
+To read a grain state, we get the filename using the function we previously defined and combine it to the root directory coming from the options.
 
 ```csharp
 public async Task ReadStateAsync(string grainType, GrainReference grainReference, IGrainState grainState)
 {
-    var fName = grainReference.ToKeyString() + "." + grainType;
-    var path = Path.Combine(_rootDirectory, fName);
+    var fName = GetKeyString(grainType, grainReference);
+    var path = Path.Combine(_options.RootDirectory, fName);
 
     var fileInfo = new FileInfo(path);
     if (!fileInfo.Exists)
@@ -151,8 +162,8 @@ public async Task WriteStateAsync(string grainType, GrainReference grainReferenc
 {
     var storedData = JsonConvert.SerializeObject(grainState.State, _jsonSettings);
 
-    var fName = grainReference.ToKeyString() + "." + grainType;
-    var path = Path.Combine(_rootDirectory, fName);
+    var fName = GetKeyString(grainType, grainReference);
+    var path = Path.Combine(_options.RootDirectory, fName);
 
     var fileInfo = new FileInfo(path);
 
@@ -172,8 +183,8 @@ Clearing the state would be deleting the file if the file exists.
 ```csharp
 public Task ClearStateAsync(string grainType, GrainReference grainReference, IGrainState grainState)
 {
-    var fName = grainReference.ToKeyString() + "." + grainType;
-    var path = Path.Combine(_rootDirectory, fName);
+    var fName = GetKeyString(grainType, grainReference);
+    var path = Path.Combine(_options.RootDirectory, fName);
 
     var fileInfo = new FileInfo(path);
     if (fileInfo.Exists)
--- a/src/Tutorials/Custom-Grain-Storage.md
+++ b/src/Tutorials/Custom-Grain-Storage.md
@@ -155,8 +155,12 @@ public async Task ReadStateAsync(string grainType, GrainReference grainReference
         var storedData = await stream.ReadToEndAsync();
         grainState.State = JsonConvert.DeserializeObject(storedData, _jsonSettings);
     }
+
+    grainState.ETag = fileInfo.LastWriteTimeUtc.ToString();
 }
 ```
+
+We use the `fileInfo.LastWriteTimeUtc` as a ETag which will be used by other functions for inconsistency checks to prevent data loss.
 
 Note that for the deserialization, we use the `_jsonSettings` which was set on the `Init` function. This is important to be able to serialize/deserialize properly the state.
 
@@ -174,14 +178,23 @@ public async Task WriteStateAsync(string grainType, GrainReference grainReferenc
 
     var fileInfo = new FileInfo(path);
 
+    if (fileInfo.Exists && fileInfo.LastWriteTimeUtc.ToString() != grainState.ETag)
+    {
+        throw new InconsistentStateException($"Version conflict (WriteState): ServiceId={_clusterOptions.ServiceId} ProviderName={_storageName} GrainType={grainType} GrainReference={grainReference.ToKeyString()}.");
+    }
+
     using (var stream = new StreamWriter(fileInfo.Open(FileMode.Create, FileAccess.Write)))
     {
         await stream.WriteAsync(storedData);
     }
+
+    fileInfo.Refresh();
+    grainState.ETag = fileInfo.LastWriteTimeUtc.ToString();
 }
 ```
 
 Similarly as reading, we use `_jsonSettings` to write the state.
+The current ETag is used to check against the last updated time in UTC of the file. If the date is different, it means that another activation of the same grain changed the state concurrently. In this situation, we throw an `InconsistentStateException` which will result in the current activation being killed to prevent overwritting the state previously saved by the other activated grain.
 
 ## Clearing State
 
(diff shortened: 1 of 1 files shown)


src/Tutorials/Custom-Grain-Storage.md
```diff
@@ -0,0 +1,232 @@
...
+
+```csharp
+public async Task ReadStateAsync(string grainType, GrainReference grainReference, IGrainState grainState)
+{
+    var fName = grainReference.ToKeyString() + "." + grainType;
+    var path = Path.Combine(_rootDirectory, fName);
+
+    var fileInfo = new FileInfo(path);
+    if (!fileInfo.Exists)
+        return;
```
jason-bragg: If no state is found, probably want to set state to default.  Something like:

grainState.State = Activator.CreateInstance(grainState.State.GetType())

src/Tutorials/Custom-Grain-Storage.md
```diff
@@ -0,0 +1,232 @@
...
+
+    using (var stream = fileInfo.OpenText())
+    {
+        var storedData = await stream.ReadToEndAsync();
+        grainState.State = JsonConvert.DeserializeObject(storedData, _jsonSettings);
+    }
+}
+```
+
+Note that for the deserialization, we use the `_jsonSettings` which was set on the `Init` function. This is important to be able to serialize/deserialize properly the state.
```
jason-bragg: May want to note that serialization is an implementation detail.  Developers can use any serialization they choose.

src/Tutorials/Custom-Grain-Storage.md
```diff
@@ -0,0 +1,232 @@
...
+
+```csharp
+public Task ClearStateAsync(string grainType, GrainReference grainReference, IGrainState grainState)
+{
+    var fName = grainReference.ToKeyString() + "." + grainType;
+    var path = Path.Combine(_rootDirectory, fName);
+
+    var fileInfo = new FileInfo(path);
+    if (fileInfo.Exists)
+        fileInfo.Delete();
```
jason-bragg: Probably want to set state to default. Something like:
grainState.State = Activator.CreateInstance(grainState.State.GetType())

src/Tutorials/Custom-Grain-Storage.md
```diff
@@ -0,0 +1,232 @@
...
+After that we will create a factory which will allow us to scope the options setting to the provider name and at the same time create an instance of the `FileGrainStorage` to ease the registration to the service collection.
+
+```csharp
+public static class FileGrainStorageFactory
+{
+    internal static IGrainStorage Create(IServiceProvider services, string name)
+    {
+        IOptionsSnapshot<FileGrainStorageOptions> optionsSnapshot = services.GetRequiredService<IOptionsSnapshot<FileGrainStorageOptions>>();
+        var options = optionsSnapshot.Get(name);
+        return ActivatorUtilities.CreateInstance<FileGrainStorage>(services, name, options.RootDirectory);
```
jason-bragg: It's an implementation detail, but we tend to partition our data further using the service Id from ClusterOptions.  This allows the same code with the same configuration to be run on different services and not write to the same location.

src/Tutorials/Custom-Grain-Storage.md
```diff
@@ -0,0 +1,232 @@
...
+    var path = Path.Combine(_rootDirectory, fName);
+
+    var fileInfo = new FileInfo(path);
+    if (!fileInfo.Exists)
+        return;
+
+    using (var stream = fileInfo.OpenText())
+    {
+        var storedData = await stream.ReadToEndAsync();
+        grainState.State = JsonConvert.DeserializeObject(storedData, _jsonSettings);
```
jason-bragg: If the implementation does not set or check the grainState.Etag, data loss is possible.  In case of Etag violation an InconsistentStateException (or some specialization thereof) needs be thrown.

src/Tutorials/Custom-Grain-Storage.md
```diff
@@ -0,0 +1,232 @@
...
+{
+    public class FileGrainStorage : IGrainStorage, ILifecycleParticipant<ISiloLifecycle>
+    {
+        private readonly string _storageName;
+        private readonly string _rootDirectory;
+        private readonly IGrainFactory _grainFactory;
+        private readonly ITypeResolver _typeResolver;
+        private JsonSerializerSettings _jsonSettings;
+
+        public FileGrainStorage(string storageName, string rootDirectory, IGrainFactory grainFactory, ITypeResolver typeResolver)
```
jason-bragg: Suggest passing the entire FileGrainStorageOptions to the implementation rather than just the root directory, that way options are automatically piped through if more options are added.

src/Tutorials/Custom-Grain-Storage.md
```diff
@@ -0,0 +1,232 @@
...
+    internal static IGrainStorage Create(IServiceProvider services, string name)
+    {
+        IOptionsSnapshot<FileGrainStorageOptions> optionsSnapshot = services.GetRequiredService<IOptionsSnapshot<FileGrainStorageOptions>>();
+        var options = optionsSnapshot.Get(name);
+        return ActivatorUtilities.CreateInstance<FileGrainStorage>(services, name, options.RootDirectory);
+    }
+}
+```
+
+Lastly to register the grain storage, we create an extension on the `ISiloHostBuilder` which internally register the grain storage as a named service using `.AddSingletonNamedService(...)`, an extension provided by `Orleans.Core`.
```
jason-bragg: May want to be clear that two named services need be added, one for the named IGrainStorage service, and one for the named ILifecycleParticipant<ISiloLifecycle> service.

src/Tutorials/Custom-Grain-Storage.md
Kimserey: Nice, I changed the constructor to take the options as argument.

src/Tutorials/Custom-Grain-Storage.md
Kimserey: Oh, I didn't see that partition in the AWS DynamoDB persistence and the Azure blob persistence but I just saw it in the Azure Table impl.
I added the `IOptions<ClusterOptions>` as argument of the constructor and added the `.ServiceId` as part of a function constructing the filename. (`GetKeyString`).

src/Tutorials/Custom-Grain-Storage.md
Kimserey: Thanks for the comment, that's interesting!

> If the implementation does not set or check the grainState.Etag, data loss is possible
Would you have an example to share where data loss would occur due to Etag not being checked?

> In case of Etag violation an InconsistentStateException (or some specialization thereof) needs be thrown.
I didn't see what the ETag was used for in the current AWS Dynamo and Azure Blob grain storages. Would you be able to guide me on where to look at?

src/Tutorials/Custom-Grain-Storage.md
Kimserey: Nice, I added a note in the first section:

```
Json is an implementation detail, it is up to the developer to decide what serialization/deserialization protocol would fit the application. Another common format is binary format.
```

src/Tutorials/Custom-Grain-Storage.md
Kimserey: Nice, I added `grainState.State = Activator.CreateInstance(grainState.State.GetType());` to set the state back to a `default` instantiation.

src/Tutorials/Custom-Grain-Storage.md
Kimserey: Nice, if state not found, I added `grainState.State = Activator.CreateInstance(grainState.State.GetType());` to set the state to a `default` instantiation.

src/Tutorials/Custom-Grain-Storage.md
Kimserey: Thanks! I added an extra explanation on this.

src/Tutorials/Custom-Grain-Storage.md
veikkoeeva: @Kimserey It varies a bit by implementation, I suppose. In ADO.NET it's checked like https://github.com/dotnet/orleans/blob/master/src/AdoNet/Orleans.Persistence.AdoNet/Storage/Provider/AdoNetGrainStorage.cs#L193 . I believe all providers have not traditionally checked and thrown `InconsistentStateException` explicitly (I think ATS added check last winter).

veikkoeeva: Note also https://github.com/dotnet/orleans/blob/master/src/AdoNet/Orleans.Persistence.AdoNet/Storage/Provider/AdoNetGrainStorage.cs#L27 .

jason-bragg: >Would you have an example to share where data loss would occur due to Etag not being checked?

In the (rare) case of duplicate activations, one each grain would be able to load a modify and save the grain state and possibly overwrite the state changes of the other.

Assume A and B are the same grain, where B is a duplicate activation, caused by network partitioning, upgrade, or some other rare occurrence.

1) A loads state.
2) B loads state.
3) A modifies and saves state.
4) B modifies and saves state.

In the above, without etag validation the changes in 4) by B would overwrite the changes made by A in 3).  The expected behavior in the above would be for 4) to throw an InconsistentStateException, causing B to deactivate.

As you can probably see from looking through the grain storage implementations, they are inconsistent in many ways.  I won't go into the history of how we got to this point, but your document should help future implementations be more consistent in regards to the expected behaviors.

Kimserey:
Thanks. I'll get the ETag changes done later today.

src/Tutorials/Custom-Grain-Storage.md
Kimserey: Thanks @veikkoeeva and @jason-bragg, your comments helped be understand what was needed in term of ETag checks.

I added more information around ETag handling and also modified the code sections for `Read`, `Write` and `Clear`.

The behaviors are as followed:

- When grain reads state, ETag is set to last write date UTC
-  When grain writes state, current ETag is compared to last write date UTC of the file, if not equal, it means that a duplicate grain activation has written data in the state. So it throws exception making the activation kill itself
- When grain clears state, current ETag is compared to last write date UTC for the same reason as write

The previous behaviors also cater for non-existent files.

Thanks!

veikkoeeva: @Kimserey Maybe add an explanation that in real operation the time resolution isn't sufficient and there's issue with locking the file, but this is just an example. Otherwise someone probably takes and uses this as-is. :)

Kimserey: Thanks @veikkoeeva, I have added a stronger warning including the points you mentioned.

Issue #4417, opened 2018-04-07
Grain not found if in same project


dandago:
I'm following the [Hello World sample documentation](https://dotnet.github.io/orleans/Documentation/Samples-Overview/Hello-World.html). But rather than build a full project structure (as the [actual code sample](https://github.com/dotnet/orleans/tree/master/Samples/2.0/HelloWorld) does), I'm dumping everything (silo, client, grain and interface) in the same console app. This is because I want to keep it simple, and not complicate an introductory exercise with additional infrastructure. First I make it work, then I make it nice.

Unfortunately, the "make it work" part didn't play out very nicely. The grain is not being found, even though it sits literally in the same file as the `Main()` method.

![image](https://user-images.githubusercontent.com/13241564/38455497-286185e2-3a71-11e8-8cb8-1c3360663e51.png)

According to the [Server Configuration documentation](https://dotnet.github.io/orleans/Documentation/Deployment-and-Operations/Configuration-Guide/Server-Configuration.html), this should work without the need for additional configuration:

> "Although this step is not technically required (if not configured, Orleans will scan all assembly in the current folder), developers are encouraged to configure this. This step will help Orleans to load user assemblies and types."

However, I have also tried `.ConfigureApplicationParts(parts => parts.AddApplicationPart(typeof(HelloGrain).Assembly).WithReferences())` and `.ConfigureApplicationParts(parts => parts.AddFromAppDomain())` and I still get the same error.

I am attaching a repro project.

[Orleans2TestDrive.zip](https://github.com/dotnet/orleans/files/1886403/Orleans2TestDrive.zip)

ReubenBond:
Make sure you add the code generation package

dandago:
Yes, that solved it.

    Install-Package Microsoft.Orleans.OrleansCodeGenerator.Build

So it is required now, and there is no default if you don't install a codegen package?

ReubenBond:
That's correct, @dandago, there is no default.
For all intents and purposes, that package is required

ReubenBond:
Thank you, @dandago

PR #4671, opened 2018-06-08
Update NuGets.md

### LINKED ISSUE #4528: Orleans 2.0 documentation page inaccuracies
This thread relates to the [Orleans 2.0 documentation page](https://dotnet.github.io/orleans/Documentation/Orleans-2.0.html).

> Explicit programmatic specification of application assemblies that replaces automatic scanning of folders by the Orleans runtime upon silo or client initialization. Orleans will still automatically find relevant types, such as grain interfaces and classes, serializers, etc. in the specified assemblies, but **it will not anymore try to load every assembly it can find in the folder**. An optional helper method for loading all assemblies in the folder is provided for backward compatibility. See Configuration and Migration sections for more details.

I thought that automatic scanning for assemblies was the default behaviour - ref [Application Parts documentation](https://dotnet.github.io/orleans/Documentation/Deployment-and-Operations/Configuration-Guide/Server-Configuration.html#application-parts) and https://github.com/dotnet/orleans/issues/4420#issuecomment-379962127 .

> Overhaul of code generation. **While mostly invisible for developer**, code generation became much more robust in handling serialization of various possible types. Special handling is required for F# assemblies. See Code generation section for more details.

Code generation is not very invisible, given that Orleans won't work without it - ref https://github.com/dotnet/orleans/issues/4417 .

JillHeaden:
Move Code Generation package to Key Packages area, just below Built-time Code Generation package.
Rename "Additional Packages" section to "Event-Sourcing Package" 
Fixes #4528

### DIFF
--- a/src/Documentation/Getting-Started-With-Orleans/NuGets.md
+++ b/src/Documentation/Getting-Started-With-Orleans/NuGets.md
@@ -28,12 +28,18 @@ PM> Install-Package Microsoft.Orleans.OrleansCodeGenerator.Build
 Build time support for grain interfaces and implementation projects.
 Add it to your grain interfaces and implementation projects to enable code generation of grain references and serializers.
 
+### [Microsoft Orleans Code Generation](https://www.nuget.org/packages/Microsoft.Orleans.OrleansCodeGenerator/)
+
+```
+PM> Install-Package Microsoft.Orleans.OrleansCodeGenerator
+```
+Includes the run time code generator.
+
 ### [Microsoft Orleans Server Libraries](https://www.nuget.org/packages/Microsoft.Orleans.Server/)
 
 ```
 PM> Install-Package Microsoft.Orleans.Server
 ```
-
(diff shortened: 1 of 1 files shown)


sergeybykov:
Why do you think we should elevate the runtime code generation package? Compile time codegen is generally safer and more performant. So we encourage using it instead of the runtime codegen.

JillHeaden:
I got the impression that that package was in the wrong spot because it was required, as Reuben said in issue #4417.
#4417 was closed, but referenced by issue #4528 with the message:  "Code generation is not very invisible, given that Orleans won't work without it - ref #4417 ."

The other codegen package wasn't mentioned. Perhaps it was added after the ticket was opened?

sergeybykov:
I believe he meant `Microsoft.Orleans.OrleansCodeGenerator.Build`, which is build time codegen.

JillHeaden:
Closing

PR #4672, opened 2018-06-08
Remove unused settings MaxPendingWorkItemsHardLimit in SchedulingOptions

xiazen:
MaxPendingWorkItemsHardLimit is not used. No reason for it to be exposed through SchedulingOptions


xiazen:
potential breaking change to 2.0. So summon @sergeybykov to notice him

ReubenBond:
Is there a reason that this shouldn't be configurable?

ReubenBond:
@xiazen what I meant is that we ought to be able to globally configure this limit, i.e, to be able to configure a hard limit on the number of `Task`s which can be enqueued on a `WorkItemGroup`.

If you feel like that's not desirable, then this is the right course of action.

I'm supposing that there was some hallway discussion about what the right behavior is.

Either way is ok, I'm just saying we should be explicit about it.

xiazen:
oh I see. You are asking is it valuable to have the feature of load shedding on task limit. 

So a bit of context on this config, I track back on 1.5, there's no setting which is equivalent to this in 1.5. I don't know in which PR it introduced this setting but didn't use it in runtime. 

Whether it is valuable to add this , it is debatable. Let's bring it up to the team in sync meeting and decide. I feel like it can be valuable to play with it and test to see if it has perf gains when a cluster is overloaded. But in this pr, I assume it is a bug to expose this setting publicly since your configuration won't be effective in runtime. 

Does this make sense ?

ReubenBond:
Sounds fine to me :)

PR #4677, opened 2018-06-11
[WIP] Tx load shedding based on lock queue length

xiazen:
Here is my wip for load shedding based on read-write lock on transactional state. 
Rough algorithm:
- keep track of the `LockGroup` queue length in variable `lockQueueLength `
- reject new tx enqueue `LockGroup` if its length is larger than `QueueLengthHardLimit` 

Main problem for this algorithm is that it is hard to test it right now, the test actually hit the azure storage limit first than hit the queue length limit. Based on the statistics I record, the queue length is around 0.5 in average when hit storage throttling limit. I tried with memory storage too and it is similar issue.  So submit this WIP so more people can have an eye on this, in case I got the implementation wrong. 

Note:
- `QueueLengthHardLimit ` is currently hard coded, but should be configurable in the future
- I added `TransactionStateStatistics ` just to find out what's the proper value for `QueueLengthHardLimit`, by observing `LockGroup` queue length when tx runtime is overloaded with this load shedding turned off. I understand it may not be an optimal implementation for `TransactionStateStatistics`. We need to refactor it before check it in. One problem is reporting statistic on a per state level is too verbose. But if all state share the same `TransactionStateStatistics` singleton, it become a bottle neck. But refactoring it is not the main thing I want to talk about in this WIP. we can get to it in a later point

### DIFF
--- a/src/Orleans.Core/Lifecycle/LifecycleSubject.cs
+++ b/src/Orleans.Core/Lifecycle/LifecycleSubject.cs
@@ -47,7 +47,7 @@ namespace Orleans
                     Stopwatch stopWatch = Stopwatch.StartNew();
                     await Task.WhenAll(observerGroup.Select(orderedObserver => WrapExecution(ct, orderedObserver.Observer.OnStart)));
                     stopWatch.Stop();
-                    this.logger?.Info(ErrorCode.SiloStartPerfMeasure, $"Starting lifecycle stage {this.highStage} took {stopWatch.ElapsedMilliseconds} Milliseconds");
+                    //this.logger?.Info(ErrorCode.SiloStartPerfMeasure, $"Starting lifecycle stage {this.highStage} took {stopWatch.ElapsedMilliseconds} Milliseconds");
                 }
             }
             catch (Exception ex)
@@ -74,7 +74,7 @@ namespace Orleans
                     Stopwatch stopWatch = Stopwatch.StartNew();
                     await Task.WhenAll(observerGroup.Select(orderedObserver => WrapExecution(ct, orderedObserver.Observer.OnStop)));
                     stopWatch.Stop();
-                    this.logger?.Info(ErrorCode.SiloStartPerfMeasure, $"Stopping lifecycle stage {this.highStage} took {stopWatch.ElapsedMilliseconds} Milliseconds");
+                    //this.logger?.Info(ErrorCode.SiloStartPerfMeasure, $"Stopping lifecycle stage {this.highStage} took {stopWatch.ElapsedMilliseconds} Milliseconds");
                 }
                 catch (Exception ex)
                 {
--- a/src/Orleans.Transactions/State/ReaderWriterLock.cs
+++ b/src/Orleans.Transactions/State/ReaderWriterLock.cs
@@ -1,14 +1,57 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
+using System.Threading;
 using System.Threading.Tasks;
+using System.Transactions;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Orleans.Configuration;
 using Orleans.Runtime;
 using Orleans.Transactions;
 using Orleans.Transactions.Abstractions;
 
 namespace Orleans.Transactions
 {
+    public class TransactionStateStatistics
+    {
+        private const string AvgLockQueueLength = "TransactionalState.AvgLockQueueLength";
+        //Transaction started recorded at when metrics was reported last time
+        private double lockQueueLengthCounter;
+
+        private long lockQueueLengthReportCounter;
+        private readonly ITelemetryProducer telemetryProducer;
+        private readonly PeriodicAction monitor;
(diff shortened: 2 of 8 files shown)


src/Orleans.Core/Lifecycle/LifecycleSubject.cs
```diff
@@ -47,7 +47,7 @@ public LifecycleSubject(ILogger<LifecycleSubject> logger)
                     Stopwatch stopWatch = Stopwatch.StartNew();
                     await Task.WhenAll(observerGroup.Select(orderedObserver => WrapExecution(ct, orderedObserver.Observer.OnStart)));
                     stopWatch.Stop();
-                    this.logger?.Info(ErrorCode.SiloStartPerfMeasure, $"Starting lifecycle stage {this.highStage} took {stopWatch.ElapsedMilliseconds} Milliseconds");
+                    //this.logger?.Info(ErrorCode.SiloStartPerfMeasure, $"Starting lifecycle stage {this.highStage} took {stopWatch.ElapsedMilliseconds} Milliseconds");
```
benjaminpetit: Why commenting it? I think we should put it as verbose maybe, but definitely not remove it, a tleast not in this PR.

src/Orleans.Transactions/State/ReaderWriterLock.cs
```diff
@@ -1,14 +1,57 @@
...
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Orleans.Configuration;
 using Orleans.Runtime;
 using Orleans.Transactions;
 using Orleans.Transactions.Abstractions;
 
 namespace Orleans.Transactions
 {
+    public class TransactionStateStatistics
```
benjaminpetit: This file is already 500+ lines, could you put it in another file?

src/Orleans.Transactions/State/ReaderWriterLock.cs
```diff
@@ -250,13 +299,17 @@ private Task LockWork()
             }
 
             return task;
+
         }
 
+        private const int QueueLengthHardLimit = 10;
```
benjaminpetit: Do we plan to make this configurable?

xiazen: yes, I mentioned it in the description

src/Orleans.Core/Lifecycle/LifecycleSubject.cs
xiazen: I commented this out temporarily, and didn't uncomment it since this is a WIP pr. but this won't be commited if we merge this pr. And there's another pr out there to make the logging less verbose. My plan was to rebase on top of that if that pr got merged. 

sorry for the confusion

src/Orleans.Transactions/State/ReaderWriterLock.cs
xiazen: yes.  I'm actually not sure to check this statistic class in since it has problems and is not the goal for this pr, as I mentioend in the description.  I will move it to another file if I decided to check it in.

test/Benchmarks/Transactions/TransactionBenchmark.cs
```diff
@@ -99,7 +100,8 @@ public sealed class SiloTransactionConfigurator : ISiloBuilderConfigurator
         {
             public void Configure(ISiloHostBuilder hostBuilder)
             {
-                hostBuilder.UseDistributedTM();
+                hostBuilder.UseDistributedTM()
```
ReubenBond: Woah... we seriously need to change this extension method name ASAP

Opened #4688 about it.

src/Orleans.Transactions/State/ReaderWriterLock.cs
```diff
@@ -250,13 +299,17 @@ private Task LockWork()
             }
 
             return task;
+
```
ReubenBond: Please run some kind of style checking/fixing tool

ReubenBond:
Generally I'm not a fan of this approach, primarily because it assumes that the system will fail if the user attempts to process more than 10 transactions concurrently.

I believe a more suitable and adaptive approach would be based upon the expected duration of the queue rather than the number of items in it.

Ultimately I won't block this, but I think that a duration-based approach would be more suitable and *might* be able to be applied more generally. For example, could we implement a more accurate form of this check by supporting customization of `ActivationData.CheckOverloaded`?

xiazen:
I'm all for good ideas. Like in the description said, this is a WIP :) . Can you explain more on "implement a more accurate form of this check by supporting customization of ActivationData.CheckOverloaded" ? Do you mean have some sort of config in ActivationData realm which user can configure to be a certain way when they use tx, and configure to be another way when they don't?

ReubenBond:
Something like the ability to inject some `IActivationOverloadPredicate` or similar... maybe more general and inject a predicate which determines if an activation can enqueue/execute a message, where the overload check is one of those

jason-bragg:
I suggest we keep this prototype under source control in case the existing transaction load shedding is insufficient, but I don't think we should introduce this change until we are convinced simpler solutions are insufficient.

xiazen:
close this p, since there's no commitment scheduled for this feature in the short term. Would still keep it under source control in case we need to bring it up again.

PR #4673, opened 2018-06-08
Fix (or workaround?) for codegen using netcore/netstandard 2.1

benjaminpetit:
Fix for #4585

PR #4689, opened 2018-06-14
Use netcoreapp2.0 for msbuild target dll if using dotnet core msbuild but targeting full .net

benjaminpetit:
#4673 introduced a regression: when using the command `dotnet build` to build project that target the full .NET framework, the variable for `CoreAssembly` is not set, and the variable `TaskAssembly` depends on it.

The easiest workaround here is to always fallback to 2.0 dll

Maybe we could instead removing this task dependency, and rely on the fact that we should find the `dotnet` command in the path @ReubenBond ?


ReubenBond:
BTW, regarding the question: it would be nice if we could assume that `dotnet` is on the PATH and that it's the correct version, but we cannot. The user could be running the build using a different version of `dotnet` (maybe even a custom build in a folder somewhere).

The MSBuild task is the same as what ASP.NET use - unless they've changed that.

We could use inline C# in the MSBuild script instead of having a task dll 😈

Another option, which doesn't rely on `dotnet`, might be to package self-contained apps for every platform.

With Hagar, I had initially hoped to just run everything inside of an MSBuild task, but that's not tenable since there may be clashes with the loaded versions of various DLLs (eg, Roslyn, mscorlib). Still, using source-to-source for code generation eliminates most dll loading issues: no need to load application dlls for reflection, and Roslyn will happily load metadata from ref dlls.

That should be our goal and I *hope* we can somehow find time to do that.

PR #4699, opened 2018-06-19
remove internal

romanlutz:
Removing internal to allow for extensions


src/Azure/Orleans.Reminders.AzureStorage/Storage/AzureBasedReminderTable.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Orleans.Runtime.ReminderService
 {
-    internal class AzureBasedReminderTable : IReminderTable
+    class AzureBasedReminderTable : IReminderTable
```
sergeybykov: Did you mean to put `public` here?

romanlutz: Yes! Thanks.

PR #4707, opened 2018-06-23
Modify activation garbage collection code

aqa510415008:
Modify activation garbage collection code


ReubenBond:
It would be better for us to use `RuntimeTypeNameFormatter.Format` here

PR #4714, opened 2018-06-26
Fix TransactionParticipantExtensionWrapper serialization issues

xiazen:
- removed static `MetaData.SerializerSettings`. Now serializer settings are parsed through constructors
- fixed serialization settings. Should use `TypeNameHandling.All`, instead of `Auto`
- fixed inconsistent serialization settings used in `AzureTableTransactionStateStorage` and `TransactionalState`, `StorageBatch`
- add `[JsonProperty]` attr to `TransactionParticipantExtensionWrapper ` fields

The first bullet point is refactoring. The remaining is there to fix the issue of `TransactionParticipantExtensionWrapper` is serialized into a "{}" before persist into storage, which is a json string represents empty instance. This causes transaction recovery failing when restore transaction state from storage.

### DIFF
--- a/src/Orleans.Transactions/State/StorageBatch.cs
+++ b/src/Orleans.Transactions/State/StorageBatch.cs
@@ -74,7 +74,7 @@ namespace Orleans.Transactions
         private int confirm = 0;
         private int collect = 0;
         private int cancel = 0;
-        private JsonSerializerSettings serializerSettings;
+        public JsonSerializerSettings SerializerSettings { get; private set; }
         public MetaData MetaData { get; private set; }
 
         public string ETag { get; set; }
@@ -85,23 +85,19 @@ namespace Orleans.Transactions
             return $"batchsize={total} [{read}r {prepare}p {commit}c {confirm}cf {collect}cl {cancel}cc]";
         }
 
-        public StorageBatch(TransactionalStorageLoadResponse<TState> loadresponse, JsonSerializerSettings settings)
+        public StorageBatch(TransactionalStorageLoadResponse<TState> loadresponse, JsonSerializerSettings serializerSettings)
         {
-            if(settings == null)
-                throw new ArgumentNullException(nameof(settings));
-            this.serializerSettings = settings;
-            MetaData = ReadMetaData(loadresponse, this.serializerSettings);
+            this.SerializerSettings = serializerSettings??throw new ArgumentNullException(nameof(serializerSettings));
+            MetaData = ReadMetaData(loadresponse, this.SerializerSettings);
             ETag = loadresponse.ETag;
             confirmUpTo = loadresponse.CommittedSequenceId;
             cancelAbove = loadresponse.PendingStates.LastOrDefault()?.SequenceId ?? loadresponse.CommittedSequenceId;
             cancelAboveStart = cancelAbove;
         }
 
-        public StorageBatch(StorageBatch<TState> previous, JsonSerializerSettings settings)
+        public StorageBatch(StorageBatch<TState> previous)
         {
-            if (settings == null)
-                throw new ArgumentNullException(nameof(settings));
-            this.serializerSettings = settings;
+            this.SerializerSettings = previous.SerializerSettings;
             MetaData = previous.MetaData;
             confirmUpTo = previous.confirmUpTo;
             cancelAbove = previous.cancelAbove;
@@ -127,7 +123,7 @@ namespace Orleans.Transactions
 
         public Task<string> Store(ITransactionalStateStorage<TState> storage)
         {
-            var jsonMetaData = JsonConvert.SerializeObject(MetaData, this.serializerSettings);
+            var jsonMetaData = JsonConvert.SerializeObject(MetaData, this.SerializerSettings);
             var list = new List<PendingTransactionState<TState>>();
 
             if (prepares != null)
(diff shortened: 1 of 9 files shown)


src/Orleans.Transactions/State/StorageBatch.cs
```diff
@@ -74,38 +74,43 @@ internal class StorageBatch<TState> : ITransactionalStateStorageEvents<TState>
...
             cancelAbove = loadresponse.PendingStates.LastOrDefault()?.SequenceId ?? loadresponse.CommittedSequenceId;
             cancelAboveStart = cancelAbove;
         }
 
-        public StorageBatch(StorageBatch<TState> previous)
+        public StorageBatch(StorageBatch<TState> previous, JsonSerializerSettings settings)
         {
+            if (settings == null)
+                throw new ArgumentNullException(nameof(settings));
+            this.serializerSettings = settings;
```
jason-bragg: nit: - style comment, feel free to ignore
this.serializerSettings = settings ?? throw new ArgumentNullException(nameof(settings)); 
is less code and looks cleaner.

src/Orleans.Transactions/State/TransactionParticipantExtension.cs
```diff
@@ -3,11 +3,14 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Newtonsoft.Json;
 
 namespace Orleans.Transactions
 {
+    [Serializable]
     public class TransactionParticipantExtension : ITransactionParticipantExtension
```
jason-bragg: y does this need to be serializable?

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -120,11 +121,8 @@ IClock clock
             lockWorker = new BatchWorkerFromDelegate(LockWork);
             storageWorker = new BatchWorkerFromDelegate(StorageWork);
             confirmationWorker = new BatchWorkerFromDelegate(ConfirmationWork);
-
-            if (MetaData.SerializerSettings == null)
-            {
-                MetaData.SerializerSettings = TransactionParticipantExtensionExtensions.GetJsonSerializerSettings(typeResolver, grainFactory);
-            }
+            
+            this.serializerSettings = TransactionParticipantExtensionExtensions.GetJsonSerializerSettings(typeResolver, grainFactory);
```
jason-bragg: Suggest serializerSettings be passed in through constructor.  Push responsibility up a layer to the factory.

Advantages:
- Reduces dependencies, as typeResolver and grainFactory are only passed in to construct these settings.
- Improves testability, due to reduced dependencies, and responsivity.
- More flexible, as users can replace factory (not complicated) to customize xml settings.

src/Orleans.Transactions/State/StorageBatch.cs
```diff
@@ -74,38 +74,43 @@ internal class StorageBatch<TState> : ITransactionalStateStorageEvents<TState>
...
+            MetaData = ReadMetaData(loadresponse, this.serializerSettings);
             ETag = loadresponse.ETag;
             confirmUpTo = loadresponse.CommittedSequenceId;
             cancelAbove = loadresponse.PendingStates.LastOrDefault()?.SequenceId ?? loadresponse.CommittedSequenceId;
             cancelAboveStart = cancelAbove;
         }
 
-        public StorageBatch(StorageBatch<TState> previous)
+        public StorageBatch(StorageBatch<TState> previous, JsonSerializerSettings settings)
```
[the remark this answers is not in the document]
xiazen: ?? what do you mean by that? StorageBatch has a property for serializationSettings?

src/Orleans.Transactions/State/TransactionParticipantExtension.cs
xiazen: it is one of the fields of TransactionParticipantExtensionWrapper, which needs to be serialized right?

src/Orleans.Transactions/State/TransactionalState.cs
xiazen: will do

src/Orleans.Transactions/State/TransactionParticipantExtension.cs
jason-bragg: To my knowledge it shouldn't be.   The ITransactionParticipantExtension in the wrapper should be a grain cast to the extension with this concrete class being what handles the calls, not this class itself.

src/Orleans.Transactions/State/StorageBatch.cs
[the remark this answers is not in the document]
jason-bragg: StorageBatch has a JsonSerializerSettings field, so we should be able to get these settings from the passed in 'previous' StorageBatch, unless the batches are unrelated, but they seem part of the same state, so the settings should be consistent, right?  It should be ok to read the settings from the previous batch.

        public StorageBatch(StorageBatch<TState> previous)
        {
            this.serializerSettings = previous.serializerSettings; 
            this.MetaData = previous.MetaData;
                ...
        }

src/Orleans.Transactions/State/StorageBatch.cs
```diff
@@ -74,7 +74,7 @@ internal class StorageBatch<TState> : ITransactionalStateStorageEvents<TState>
         private int confirm = 0;
         private int collect = 0;
         private int cancel = 0;
-        private JsonSerializerSettings serializerSettings;
+        public JsonSerializerSettings SerializerSettings { get; private set; }
```
jason-bragg: this is fine, but not sure why this needs to be a public property.

PR #4736, opened 2018-07-03
Workaround for CoreFx/#30781 (Fixes #4505)

### LINKED ISSUE #4505: Orleans gets slower over time
Hi,

it is a little bit vague, but I have the impression that Orleans gets slower over time.

My specs are:

* .NET Core 2.0
* Running on Kubernetes in GCloud, 2 Nodes/Silos
* Co-Hosted with ASP.NET Core

I have a very simple grain, that just returns the snapshot (my own state system):

https://github.com/Squidex/squidex/blob/master/src/Squidex.Domain.Apps.Entities/Schemas/SchemaGrain.cs#L305

I use my own serializer with JSON.NET and my benchmarks show that the serialization takes 2-3 ms usually. I also tested it with 100x larger states that expected and the performance is great.

I added some profiling to the client side and I have experienced that the call takes 4-5 ms after I deploy it and several days later up to 200ms or even more (the max I have seen is 10seconds).

I also checked my MongoDB logs where I save all slow queries and there is nothing related.

I am a little bit lost.

ReubenBond:
Fixes #4505 

This implements a workaround for some divergent behavior between .NET Core and .NET Framework tracked by https://github.com/dotnet/corefx/issues/30781

Note that this is intended for a 2.0.4 release.

### DIFF
--- a/src/Orleans.Runtime/Scheduler/WorkQueue.cs
+++ b/src/Orleans.Runtime/Scheduler/WorkQueue.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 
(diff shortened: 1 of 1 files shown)


benjaminpetit:
> CI is not running against this branch. Is this just a cherrypick of something that is already on master? If not, shouldn't this be on master first?

I think it doesn't make sense to put it in master, since this fix is not applicable with the change to the scheduler, do you confirm Reuben?

ReubenBond:
Yes, this fix is only for the old scheduler. We can run tests in VSO instead.

PR #4739, opened 2018-07-03
Transaction attribution test

jason-bragg:
Added test cases to ensure transaction attribute options (NotSupported, Required, RequiredNew) properly maintain transaction context.

### DIFF
--- a/src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
+++ b/src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
@@ -281,14 +281,22 @@ namespace Orleans.CodeGenerator
                 options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.AlwaysInterleave.ToString()));
             }
 
-            if (GrainInterfaceUtils.IsNewTransactionRequired(method))
+            if (GrainInterfaceUtils.TryGetTransactionOption(method, out TransactionOption option))
             {
-                options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionRequiresNew.ToString()));
-            }
-
-            if (GrainInterfaceUtils.IsTransactionRequired(method))
-            {
-                options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionRequired.ToString()));
+                switch (option)
+                {
+                    case TransactionOption.NotSupported:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionNotSupported.ToString()));
+                        break;
+                    case TransactionOption.Required:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionRequired.ToString()));
+                        break;
+                    case TransactionOption.RequiresNew:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionRequiresNew.ToString()));
+                        break;
+                    default:
+                        throw new NotSupportedException($"Transaction option {options} is not supported.");
+                }
             }
 
             ExpressionSyntax allOptions;
--- a/src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
+++ b/src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
@@ -285,14 +285,22 @@ namespace Orleans.CodeGenerator
             {
                 switch (option)
                 {
-                    case TransactionOption.NotSupported:
-                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionNotSupported.ToString()));
+                    case TransactionOption.Suppress:
                         break;
-                    case TransactionOption.Required:
-                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionRequired.ToString()));
+                    case TransactionOption.CreateOrJoin:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionCreateOrJoin.ToString()));
                         break;
-                    case TransactionOption.RequiresNew:
-                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionRequiresNew.ToString()));
+                    case TransactionOption.Create:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionCreate.ToString()));
+                        break;
+                    case TransactionOption.Mandatory:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionMandatory.ToString()));
+                        break;
+                    case TransactionOption.Supported:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionSupported.ToString()));
+                        break;
+                    case TransactionOption.Never:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionNever.ToString()));
                         break;
                     default:
                         throw new NotSupportedException($"Transaction option {options} is not supported.");
--- a/src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
+++ b/src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
@@ -286,6 +286,7 @@ namespace Orleans.CodeGenerator
                 switch (option)
                 {
                     case TransactionOption.Suppress:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionSuppress.ToString()));
                         break;
                     case TransactionOption.CreateOrJoin:
                         options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionCreateOrJoin.ToString()));
@@ -293,14 +294,14 @@ namespace Orleans.CodeGenerator
                     case TransactionOption.Create:
                         options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionCreate.ToString()));
                         break;
-                    case TransactionOption.Mandatory:
-                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionMandatory.ToString()));
+                    case TransactionOption.Join:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionJoin.ToString()));
                         break;
                     case TransactionOption.Supported:
                         options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionSupported.ToString()));
                         break;
-                    case TransactionOption.Never:
-                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionNever.ToString()));
+                    case TransactionOption.NotAllowed:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionNotAllowed.ToString()));
                         break;
                     default:
                         throw new NotSupportedException($"Transaction option {options} is not supported.");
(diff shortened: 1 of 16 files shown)


src/Orleans.Core.Abstractions/CodeGeneration/InvokeMethodRequest.cs
```diff
@@ -67,7 +67,8 @@ public enum InvokeMethodOptions
 
         // Transactional method options. 
         // NOTE: keep in sync with TransactionOption enum.
-        TransactionRequired = 0x200,
-        TransactionRequiresNew = 0x400,
+        TransactionNotSupported = 0x200,
+        TransactionRequired = 0x400,
+        TransactionRequiresNew = 0x800,
```
ReubenBond: These are implemented like flags, but using them like flags could be contradictory.
Suggest `NotSupported` be the implied result of no value being set and leaving the existing values as-is.
Alternatively, if the default is "supported", then we can define `NotSupported = 0x600` so at least there's a meaning to that value and they're not flags (plus we save a precious bit 😉).

What do you think?

src/Orleans.Core.Abstractions/Transactions/TransactionAttribute.cs
```diff
@@ -20,8 +20,8 @@ public TransactionAttribute(TransactionOption requirement)
 
     public enum TransactionOption
     {
-        RequiresNew,
-        Required,
-        NotSupported
+        NotSupported, // Logic is not transactional.  If called within the context of a transaction, the context will not be passed ot the call.
```
ReubenBond: The comment makes it sound like it's fine to call a `NotSupported` method from within the context of a transaction (it just wont be a part of the transaction)... but I would have assumed from the name that this means that transactions are *explicitly not supported* and therefore calling such a method from within a transaction is an error because it will not behave transactionally (and therefore breaks ACID).

A method which explicitly does not support transactions silently taking part in a transaction worries me.

Maybe a more appropriate name for the described behavior would be `NotApplicable`, as in "the notion of a transaction simply does not apply to this method". Eg, perhaps the method simply returns the number 5 whenever called and there's nothing which the transactions subsystem needs to do to ensure that behaves as expected.

ReubenBond: We could rename this to `Suppress`, aligning it with .NET's `TransactionScopeOption` enum.

test/Transactions/Orleans.Transactions.Tests/Grains/TransactionAttributionGrain.cs
```diff
@@ -0,0 +1,71 @@
...
+            ITransactionInfo ti = TransactionContext.GetTransactionInfo();
+            Dictionary<int, List<string>> results = new Dictionary<int, List<string>>();
+            results[tier] = new List<string>(new[] { ti?.Id });
+
+            if (tiers.Count == 0)
+            {
+                return results;
+            }
+
+            KeyValuePair<int, List<ITransactionAttributionGrain>> nextTier = tiers.FirstOrDefault();
```
ReubenBond: Should this also be ordered?

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionAttributionTestRunner.cs
```diff
@@ -0,0 +1,166 @@
...
+            this.grainFactory = grainFactory;
+        }
+
+        [Fact]
+        public async Task FanoutToAllAttributesTest()
+        {
+            ITransactionAttributionGrain root = this.grainFactory.GetTransactionAttributionGrain(Guid.NewGuid());
+            Dictionary<int, List<ITransactionAttributionGrain>> tiers = new Dictionary<int, List<ITransactionAttributionGrain>>
+            {
+                [1] = new List<ITransactionAttributionGrain>(new[] {
```
ReubenBond: Could you add a comment describing why we start at 1 here? After reading the code it becomes clear, but it's not obvious at first. Maybe something like `// Tier 0 is the root grain. We pass a collection of grains in tier 1 so that tier 0 will fan out to those grains.`

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionAttributionTestRunner.cs
```diff
@@ -0,0 +1,166 @@
...
+            };
+            Dictionary<int, List<string>> results = await root.GetNestedTransactionIds(0, tiers);
+            foreach(KeyValuePair<int, List<string>> kvp in results)
+            {
+                this.output.WriteLine($"{kvp.Key} => {string.Join(",", kvp.Value)}");
+            }
+            // make sure there are 2 tiers
+            Assert.Equal(2, results.Count);
+
+            // make sure tier 0 has no transactionId
```
ReubenBond: `since the root grain has no TransactionOption specified`

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionAttributionTestRunner.cs
```diff
@@ -0,0 +1,166 @@
...
+        protected readonly ITestOutputHelper output;
+
+        protected TransactionAttributionTestRunner(IGrainFactory grainFactory, ITestOutputHelper output)
+        {
+            this.output = output;
+            this.grainFactory = grainFactory;
+        }
+
+        [Fact]
+        public async Task FanoutToAllAttributesTest()
```
ReubenBond: We should start describing out tests. Maybe something like:

```C#
/// <summary>
/// Tests the behavior of each <see cref="TransactionOption" /> value on the transaction context
/// when there is no ambient transaction.
/// </summary>
```

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionAttributionTestRunner.cs
```diff
@@ -0,0 +1,166 @@
...
+            Assert.Null(tier_1[0]);
+            Assert.Null(tier_1[1]);
+            Assert.NotNull(tier_1[2]);
+            Assert.NotNull(tier_1[3]);
+            // make sure the transaction id's differ
+            Assert.NotEqual(tier_1[2], tier_1[3]);
+        }
+
+        [Fact]
+        public async Task FanoutToAllAttributesFromTransactionTest()
```
ReubenBond:
```C#
/// <summary>
/// Tests the behavior of each <see cref="TransactionOption" /> value on the transaction context
/// when called in the context of an ongoing transaction.
/// </summary>
```

ReubenBond: maybe the wording is a little awkward

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionAttributionTestRunner.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            // make sure tier 0 has a transactionId
+            Assert.True(results.TryGetValue(0, out List<string> tier_0));
+            Assert.Single(tier_0);
+            Assert.NotNull(tier_0.First());
+
+            // check tier 1 transactionIds, should be null, null, guid1, guid2
+            Assert.True(results.TryGetValue(1, out List<string> tier_1));
+            Assert.Equal(2, tier_1.Count);
+            // make sure RequiresNew attributed transactionId does not matches parent
```
ReubenBond: nit: does not match the parent's

src/Orleans.Core.Abstractions/Transactions/TransactionAttribute.cs
xiazen: I think " If called within the context of a transaction, the context will not be passed ot the call. " is implementation details which may not mean much to the user. How about 

When you set this attribute value, orleans transactions runtime ensures that the method call never participates in a transaction, regardless of the transactional status of its caller. 


This is borrowed from COM+ transaction concept : https://docs.microsoft.com/en-us/windows/desktop/cossdk/configuring-transactions . They have a similar set of tx attr with us.

jason-bragg: Let's discuss the attributes meanings and names in the today's transaction sync.

test/Transactions/Orleans.Transactions.Tests/Grains/TransactionAttributionGrain.cs
jason-bragg: I will move to using flat array, will make it more clear.  Started that way and moved to dictionary for ease of coding, but it adds confusion.

test/Transactions/Orleans.Transactions.Tests/Runners/TransactionAttributionTestRunner.cs
jason-bragg: I will simplify to use array.  Should be clearer... i hope :/

src/Orleans.Core.Abstractions/Transactions/TransactionAttribute.cs
ReubenBond: > When you set this attribute value, orleans transactions runtime ensures that the method call never participates in a transaction, regardless of the transactional status of its caller.

`Never` is another possible value, used by EJB transactions for their equivalent to this enum and it means "do not start a transaction and throw if there is an existing transaction"

Yep, let's discuss in the meeting

src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
```diff
@@ -285,14 +285,22 @@ private static ArgumentSyntax GetInvokeOptions(MethodInfo method)
...
+                    case TransactionOption.Create:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionCreate.ToString()));
+                        break;
+                    case TransactionOption.Mandatory:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionMandatory.ToString()));
+                        break;
+                    case TransactionOption.Supported:
+                        options.Add(typeof(InvokeMethodOptions).GetNameSyntax().Member(InvokeMethodOptions.TransactionSupported.ToString()));
+                        break;
+                    case TransactionOption.Never:
```
ReubenBond: Perhaps `Never` should be `NotAllowed` or `Forbidden`. "Not allowed" is simple English, so I'm partial to that.

`Never` is odd, since it's a declaration that this method never has a transaction context when it's called, but it should instead indicate that it fails when there is a transaction context.

ReubenBond: (but I'm fine to leave this as-is)

Similarly, `Join` is consistent with the other names instead of `Mandatory`. Mandatory is ok, just not as consistent and not as obvious in terms of the resulting behavior.

src/Orleans.Core/Messaging/MessageFactory.cs
```diff
@@ -21,9 +21,32 @@ public MessageFactory(SerializationManager serializationManager, ILogger<Message
...
+                ((options & InvokeMethodOptions.TransactionCreateOrJoin) == 0 && 
+                 (options & InvokeMethodOptions.TransactionMandatory) == 0 &&
+                 (options & InvokeMethodOptions.TransactionSupported) == 0))
+            {
+                transactionInfo = null;
+            }
+
+            bool isTransactionRequired = (options & InvokeMethodOptions.TransactionCreate) != 0 ||
+                                         (options & InvokeMethodOptions.TransactionCreateOrJoin) != 0 ||
+                                         (options & InvokeMethodOptions.TransactionMandatory) != 0;
```
ReubenBond: We could probably arrange these so that it's just one bit that needs to be checked.

src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
jason-bragg: Meant to use Join instead of Mandatory, forgot to switch that over.  Will do.

src/Orleans.Core/Messaging/MessageFactory.cs
jason-bragg: Yeah, don't like the way this works.. Will look into.

src/Orleans.Core.Abstractions/CodeGeneration/InvokeMethodRequest.cs
```diff
@@ -68,10 +68,25 @@ public enum InvokeMethodOptions
         // Transactional method options. 
         // NOTE: keep in sync with TransactionOption enum.
         // Suppress is default, so no need for flag for it
-        TransactionCreateOrJoin = 0x200,
-        TransactionCreate = 0x400,
-        TransactionMandatory = 0x800,
-        TransactionSupported = 0x1000,
-        TransactionNever = 0x2000,
+        TransactionFlags = 0xE00,
```
ReubenBond: `TransactionOptionsMask` (since this isn't a valid value and this is a bit mask and they're no longer flags)

xiazen:
@dotnet-bot test this please

PR #4740, opened 2018-07-05
2.0.4 partial build

ReubenBond:
Please review each commit independently & in order, it will make it much easier.

Here's my reasoning for these changes:

The below snippet is from Orleans.Server’s `Directory.Build.props` from before this PR. Its job is to determine whether or not the Orleans.Server package will reference `Microsoft.Orleans.OrleansProviders` via a `ProjectReference` or a `PackageReference`.

image

In the present case I need to bump `Server` to 2.0.4, since it refs `Runtime` (and we’re trying to publish `Runtime` 2.0.4 to fix this CoreFx bug). Given the above code, that means `Server` will try to pull `Providers` 2.0.4 from NuGet (which doesn't exist yet). So that’s broken (unless I’m doing it wrong?)

Instead, all of these conditional `<PackageReference>` elements should use the *referenced* package's version from the root `Directory.Build.props` file. The dependency relationship implies that if `ServerVersion != ProvidersVersion`, then `ServerVersion > ProvidersVersion` and therefore `ProvidersVersion` must be published to NuGet already. Otherwise, (if they are equal) then we are currently building both packages and we should use a `<ProjectReference>` instead.

I had to split the Azure package group out into individually versioned packages because the published versions for those packages were inconsistent. Eg, 2.0.0 for most vs 2.0.3 for persistence. If preferred, I can undo this change and drag all Azure packages up to 2.0.4 even though only one package (cloud services) actually needs it.

In the case of the telemetry consumers, I pulled that forward to 2.0.4 since 2.0.3 (the currently specified version) is not published and as above, it would break the build now that we're moving on to 2.0.4.

In order to test locally, set the `FullBuild` environment variable to `false` and run `Build.cmd`. The result should be the following packages:
```
Microsoft.Orleans.CounterControl
Microsoft.Orleans.Hosting.AzureCloudServices
Microsoft.Orleans.Hosting.ServiceFabric
Microsoft.Orleans.OrleansAzureUtils (because of cloud services)
Microsoft.Orleans.OrleansManager
Microsoft.Orleans.OrleansRuntime
Microsoft.Orleans.OrleansRuntime
Microsoft.Orleans.OrleansTelemetryConsumers.AI
Microsoft.Orleans.OrleansTelemetryConsumers.Counters (in lieu of a 2.0.3 pkg)
Microsoft.Orleans.OrleansTelemetryConsumers.NewRelic
Microsoft.Orleans.Runtime.Legacy
Microsoft.Orleans.Server
Microsoft.Orleans.ServiceFabric
Microsoft.Orleans.TestingHost
Microsoft.Orleans.Transactions
Microsoft.Orleans.Transactions.AzureStorage
```


ReubenBond:
Build passes in VSO for full & partial build. Tests pass for partial build (didn't try full)

benjaminpetit:
Thanks!!!

> If preferred, I can undo this change and drag all Azure packages up to 2.0.4 even though only one package (cloud services) actually needs it.

I'm fine either way

Directory.Build.props
```diff
@@ -137,19 +137,19 @@
...
+    <OrleansAzureMetapackageVersion>2.0.4</OrleansAzureMetapackageVersion>
     <OrleansGoogleCloudProviderVersion>2.0.0</OrleansGoogleCloudProviderVersion>
-    <OrleansAzureCloudServicesVersion>2.0.3</OrleansAzureCloudServicesVersion>
-    <OrleansTestingHostVersion>2.0.3</OrleansTestingHostVersion>
-    <OrleansTransactionsVersion>2.0.3</OrleansTransactionsVersion>
-    <OrleansServiceFabricVersion>2.0.0</OrleansServiceFabricVersion>
+    <OrleansAzureCloudServicesVersion>2.0.4</OrleansAzureCloudServicesVersion>
+    <OrleansTestingHostVersion>2.0.4</OrleansTestingHostVersion>
+    <OrleansTransactionsVersion>2.0.4</OrleansTransactionsVersion>
+    <OrleansServiceFabricVersion>2.0.4</OrleansServiceFabricVersion>
```
jason-bragg: Why are going from 2.0.0 to 2.0.4 rather than 2.0.1?

PR #4741, opened 2018-07-05
Fix package versioning logic

ReubenBond:
This is very similar to #4740, but based on `master`, with all the changes there.

This should only need skimming compared with the original PR


ReubenBond:
Full build & partial build succeed in VSO. Running tests on the partial build.

ReubenBond:
Superseded by #4844

PR #4746, opened 2018-07-06
Update Configuring-AdoNet-Providers.md

janmarques:


ry8806:
Hey, i think instead of "splited" or "splitted"

it should just read "split" - https://simple.wiktionary.org/wiki/split

janmarques:
OK Ryan

PR #4749, opened 2018-07-06
Cleanup Reminders PartitionKey

### LINKED ISSUE #4719: AzureStorage Reminders does not replace invalide path chars in Row and Partition Keys
Hello,

We've experienced an issue recently when our Orleans Cluster has been migrated to Service Fabric.
We are using AzureStorage Table Service for clustering, GrainState persistence and Reminders.

As the sample shows us, we use the ServiceName of the fabric as the ServiceId of the cluster.
(https://github.com/dotnet/orleans/blob/master/Samples/2.0/ServiceFabric/Stateless/StatelessCalculatorService/StatelessCalculatorService.cs )
> // The service id is unique for the entire service over its lifetime. This is used to identify persistent state
> // such as reminders and grain state.
> options.ServiceId = fabricServiceContext.ServiceName.ToString();

This works well for GrainState PartitionKey, since invalid path chars like **/** or **\\** are replaced by underscores:
`fabric:_Silo.ServiceFabric_TheSilo_GrainReference=40080509dab600a3548839c6507585be030000003ab73acd`

But this seems to be not the case for GrainReminders. 
A lot of 400 bad request errors occured because of the PartitionKey name containing invalid path chars. 

> The following characters are not allowed in values for the PartitionKey and RowKey properties:
> 
> The forward slash (/) character
> The backslash (\\) character
> The number sign (#) character
> The question mark (?) character
> Control characters from U+0000 to U+001F, including:
> The horizontal tab (\t) character
> The linefeed (\n) character
> The carriage return (\r) character
> Control characters from U+007F to U+009F

The AzureTableReminderService should use the same sanitization mechanism for PartitionKeys and RowKeys of GrainReminders as GrainState. 

Thank you,

benjaminpetit:
Fix #4719

### DIFF
--- a/src/Azure/Orleans.Reminders.AzureStorage/Storage/RemindersTableManager.cs
+++ b/src/Azure/Orleans.Reminders.AzureStorage/Storage/RemindersTableManager.cs
@@ -44,8 +44,7 @@ namespace Orleans.Runtime.ReminderService
             // when comparisons will be done on strings, this will ensure that positive numbers are always greater than negative
             // string grainHash = number < 0 ? string.Format("0{0}", number.ToString("X")) : string.Format("1{0:d16}", number);
 
-            var grainHash = String.Format("{0:X8}", number);
-            return String.Format("{0}_{1}", serviceId, grainHash);
+            return AzureStorageUtils.SanitizeTableProperty($"{serviceId}_{number:X8}");
         }
 
 


jason-bragg:
@dotnet-bot test bvt please

jason-bragg:

jason-bragg:

PR #4753, opened 2018-07-10
[WIP] Support configuration of messaging options via MessagingOptions

### LINKED ISSUE #4751: Response timeout is not respected
We're using 2.0.3 and despite configured 10s timeout for both client (embedded in the same process) and silo, we sometimes see 30 second TimeoutException in logs:

```
System.TimeoutException: Response did not arrive on time in 00:00:30 for message: Request *cli/e89bccba@8e020588->S10.31.5.13:101:268855415*grn/6EF63849/00000000+133890 #289952: . Target History is: <S10.31.5.13:101:268855415:*grn/6EF63849/00000000+133890:>.
```

I don't understand how that could happen. We have that set in code:

```cs
client.Configure<MessagingOptions>(o => o.ResponseTimeout = TimeSpan.FromSeconds(10)); 
.....
cluster.Configure<MessagingOptions>(o => o.ResponseTimeout = TimeSpan.FromSeconds(10));
```

This usually happens after ungraceful shutdown of the silo ...

ReubenBond:
Fixes #4751

### DIFF
--- a/src/Orleans.Core/Core/DefaultClientServices.cs
+++ b/src/Orleans.Core/Core/DefaultClientServices.cs
@@ -1,5 +1,7 @@
+using System.Collections.Generic;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
 using Orleans.ApplicationParts;
 using Orleans.Configuration;
 using Orleans.Configuration.Validators;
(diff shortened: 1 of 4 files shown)


ReubenBond:
It would be ideal if we could remove the inheritance hierarchy here, possibly by splitting out the non-shared options into separate classes, or by hiding the abstract MessagingOptions class

ReubenBond:
It would be ideal if we could remove the inheritance hierarchy here and remove the confusion

jason-bragg:
I don't think we want to support configuring MessagingOptions because it's not intended for that use.  ClientMessagingOptions and SiloMessagingOptinos are what users should be using.  I think the goal of this change should be to ensure that users don't try to configure MessagingOptions.  Maybe by making it an interface?  Maybe just adding warning comments. Not sure, but supporting configuring it is, imo, not a good solution.

It should be noted that we have a similar issue with Statistics and possibly some other options classes, so whatever we settle on, we should be consistent.

xiazen:
I think the cleaner change here is to ensure users don't try to configure MessagingOptions, by removing the inheritance hierarchy. Either by make MessagingOptions internal or just duplicate shared properties in SiloMessagingOptions or ClientMessageingOptions. Or anything better than this.

ReubenBond:
Either way is fine. This issue has been brought up before.

I prefer we split non-common properties into their own types, hopefully with intuitive names

xiazen:
are we going to update this pr?

ReubenBond:
@jason-bragg 

> Suggest solution that prevents users from configuring MessagingOptions.

That's not really up to us, since the Microsoft.Extensions.Options package does not place a `new()` constraint on `TOptions`, so we can't stop the user from calling that unless we either delete the class or make it internal.

xiazen:
we can delete `MessagingOptions`, and separate SiloMessagingOptions and ClientMessaginOptions. This seem clean and simple to me.

sergeybykov:
>delete MessagingOptions

That would be a breaking change, that however will not break backward compatibility, as this is only used in configuration, correct?

xiazen:
@sergeybykov `MessagingOptions` is the base class for SiloMessageingOptions and ClientMessagingOptions, no one should be using it. If for configuration, users should use Silo/ClientMessagingOptions. So it is technically a breaking change, but I think the impact is low

jason-bragg:
What @xiazen said!  Technically, yes, this is a breaking change, but it shouldn't break anyone because no one should be using the MessagingOptions directly.

sergeybykov:
After some discussion, we concluded that there's no value in enabling the undesirable behavior of configuring via `MessagingOptions`. There might be a follow up PR to disable that entirely to prevent mistakes.

PR #4759, opened 2018-07-11
New Sample to go with new Tutorial

JillHeaden:
@jason-bragg

### DIFF
--- a/Samples/2.0/MathGrains/Client/Program.cs
+++ b/Samples/2.0/MathGrains/Client/Program.cs
@@ -1,0 +1,96 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Orleans;
+using Orleans.Runtime;
+using Orleans.Configuration;
+using GrainInterfaces;
+
+namespace OrleansClient
+{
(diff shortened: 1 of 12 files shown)


Samples/2.0/MathGrains/Client/Program.cs
```diff
@@ -0,0 +1,96 @@
...
+                        .UseLocalhostClustering()
+                        .Configure<ClusterOptions>(options =>
+                        {
+                            options.ClusterId = "dev";
+                            options.ServiceId = "HelloWorldApp";
+                        })
+                        .ConfigureLogging(logging => logging.AddConsole())
+                        .Build();
+
+                    await client.Connect();
```
ReubenBond: We should change this: some existing code does this, too, but there is a better way to make the client connection retry. We can use the `Connect(Func<Exception, Task<bool>>)` overload, like in #4711. The method will be much simpler as a result.

jason-bragg: made a similar cleanup to hello world.  When that get's pass review, you can model off of it, if you like.  Cleanup connect retry loop in hello world sample #4767

Issue #4706, opened 2018-06-23
PerformDeadlockDetection not working in 2.0


Xanatus:
Enabling PerformDeadlockDetection on Silo prevents Orleans Client from connecting with following TypeAccessException:

![image](https://user-images.githubusercontent.com/20849443/41804305-78c87750-76d7-11e8-8db0-67ce2cfcde6e.png)

![image](https://user-images.githubusercontent.com/20849443/41804275-e9835cea-76d6-11e8-96c2-4c27d2b5eaf6.png)

Not sure if I'm missing a certain package or if this is an Orleans bug.

jason-bragg:
@dVakulen, ideas?

dVakulen:
In order to use `PerformDeadlockDetection` client project needs to reference `Microsoft.Orleans.OrleansRuntime` package, 
and to instruct the code generator to generate code for `Runtime` assembly like following:
  ```csharp
new ClientBuilder()
   .ConfigureApplicationParts(parts => parts.AddApplicationPart(typeof(Orleans.Runtime.GrainService).Assembly)) 
```

<br>
 (RequestInvocationHistory has been recently moved from Core into Runtime, thus it's currenlty inaccessible in clients which usually does not reference the Runtime package)

jason-bragg:
> In order to use PerformDeadlockDetection client project needs to reference Microsoft.Orleans.OrleansRuntime package

I don't think configuring it on the client makes sense or that @Xanatus is trying to do that.  @Xanatus can you confirm?  The config line and logs are from the silo, correct?

Xanatus:
As I stated in my original post, I tried to enable it on the Silo of course.
I also tried setting it on the client because I thought maybe I had to enable it on both, but it made no difference.

Will try dVakulens suggestions on monday and let you know if it works.

jason-bragg:
@Xanatus, I'd hold off on trying @dVakulen's suggestions until he gets back to this thread.  I suspect his response was based off a misunderstanding of what you were doing, though I'm confident he'll clarify if I'm mistaken.

Looking at the PerformDeadlockDetection flag, I don't see any tests specifically setting it so it may be broken.  Likely the testing is using a legacy config, but I’ll try to track it down.  If there is no testing of this, that's something we need to address.

jason-bragg:
[...]

PR #4761, opened 2018-07-11
Moved request invocation history to core.

jason-bragg:
Moved request invocation history to orleans core so it's available on the client.
Should address #4706


ReubenBond:
@dotnet-bot test functional

Issue #4577, opened 2018-05-11
Sanitize "." for Azure Queue name and make Queue name configurable


KTazayan:
I'm getting exceptions when project name isn't so long or contains "." for example name is "SiloHost.NetCore".

1. Make Azure Queue name configurable or partially configurable. For example for case when silo host is SF service my queue name is "fabric--distrtestapplication-silohost.netcore-azurequeueprovider-0".  It contains 66 symbols instead of max allowed. I want to remove or configure "fabric--" prefix or something else.

2. Sanitize "." with dash "-" as other symbols in AzureQueueDataManager.SanitizeQueueName.

xiazen:
I agree

xiazen:
we can sanitize it without breaking anyone. And make azure queue configurable is already an agreed feature to add based on a similar issue before. just we haven't got down to do it.

xiazen:
submitted #4582 to sanitize . from azure queue name. Make azure queue name configurable needs more changes. But I don't have more bandwidth to do it now, so I suggest to put that feature in backlog. @KTazayan feel free to contribute to make it configurable before we get to it.

PR #4762, opened 2018-07-11
Make azure queue name configurable

xiazen:
-make azure queue name prefix configuration through `AzureQueueOptions`

-if not set, then default to its original value , which is serviceId-providerName

For issue #4577

### DIFF
--- a/src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamBuilder.cs
+++ b/src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamBuilder.cs
@@ -4,6 +4,7 @@ using Orleans.Providers.Streams.AzureQueue;
 using Orleans.Streams;
 using System;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 using Orleans;
 
 namespace Orleans.Streaming
@@ -31,15 +32,10 @@ namespace Orleans.Streaming
             this.Configure<AzureQueueOptions>(configureOptions);
             return this;
         }
+
         public SiloAzureQueueStreamConfigurator<TDataAdapter> ConfigureCache(int cacheSize = SimpleQueueCacheOptions.DEFAULT_CACHE_SIZE)
         {
             this.Configure<SimpleQueueCacheOptions>(ob => ob.Configure(options => options.CacheSize = cacheSize));
-            return this;
-        }
-
-        public SiloAzureQueueStreamConfigurator<TDataAdapter> ConfigurePartitioning(int numOfPartition = HashRingStreamQueueMapperOptions.DEFAULT_NUM_QUEUES)
-        {
-            this.Configure<HashRingStreamQueueMapperOptions>(ob => ob.Configure(options => options.TotalQueueCount = numOfPartition));
             return this;
         }
     }
--- a/src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamBuilder.cs
+++ b/src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamBuilder.cs
@@ -3,6 +3,7 @@ using Orleans.Hosting;
 using Orleans.Providers.Streams.AzureQueue;
 using Orleans.Streams;
 using System;
+using System.Linq;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Orleans;
@@ -25,6 +26,16 @@ namespace Orleans.Streaming
                         .AddTransient<IConfigurationValidator>(sp => new SimpleQueueCacheOptionsValidator(sp.GetOptionsByName<SimpleQueueCacheOptions>(name), name))
                         .ConfigureNamedOptionForLogging<HashRingStreamQueueMapperOptions>(name);
                 });
+            //configure default queue names
+            this.ConfigureAzureQueue(ob => ob.PostConfigure<IOptions<ClusterOptions>>((op, clusterOp) =>
+            {
+                if (op.QueueNames == null || op.QueueNames?.Count == 0)
+                {
+                    op.QueueNames =
+                        AzureQueueStreamProviderUtils.GenerateDefaultAzureQueueNames(clusterOp.Value.ServiceId,
+                            this.name);
+                }
+            }));
         }
 
         public SiloAzureQueueStreamConfigurator<TDataAdapter> ConfigureAzureQueue(Action<OptionsBuilder<AzureQueueOptions>> configureOptions)
@@ -39,7 +50,7 @@ namespace Orleans.Streaming
             return this;
         }
     }
-
+    
     public class ClusterClientAzureQueueStreamConfigurator<TDataAdapter> : ClusterClientPersistentStreamConfigurator
           where TDataAdapter : IAzureQueueDataAdapter
     {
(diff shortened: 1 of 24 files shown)


ReubenBond:
> AzureQueueDataManagerTests.cs(41,49): error CS1729: 'AzureQueueDataManager' does not contain a constructor that takes 5 arguments [D:\j\workspace\functional_pr---83d31c72\test\Extensions\TesterAzureUtils\Tester.AzureUtils.csproj

src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamBuilder.cs
```diff
@@ -29,6 +30,15 @@ public SiloAzureQueueStreamConfigurator(string name, ISiloHostBuilder builder)
         public SiloAzureQueueStreamConfigurator<TDataAdapter> ConfigureAzureQueue(Action<OptionsBuilder<AzureQueueOptions>> configureOptions)
         {
             this.Configure<AzureQueueOptions>(configureOptions);
+            this.Configure<AzureQueueOptions>(b => b.PostConfigure<IOptions<ClusterOptions>>((op, clusterOp) =>
+            {
+                if (op.QueueNamePrefix == null)
+                {
+                    var serviceId = clusterOp.Value?.ServiceId;
+                    var defaultQueueNamePrefix = serviceId + "-" + this.name;
+                    op.QueueNamePrefix = defaultQueueNamePrefix;
```
jason-bragg: I don't think we should handle this override in the configuration stage.  It makes it difficult to override the behavior.

I'd think what we'd want is for the AzureQueueAdapterFactory to take a concrete list of the queue names that we build in the static Create function using the QueueNamePrefix, serviceId, provider name, and HashRingStreamQueueMapperOptions.TotalQueueCount.  If the patterns we support aren't what a developer needs, they'd need only write their own static Create function and construct the queue name list any way they want (for instance, providing a list of pre-existing queues they want to process events from).

src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamProviderUtils.cs
```diff
@@ -17,52 +17,44 @@ public class AzureQueueStreamProviderUtils
...
-                }
+            // TODO: Do not assume defaults !? - jbragg
+            var queueMapper = new HashRingBasedStreamQueueMapper(new HashRingStreamQueueMapperOptions(), queueNamePrefix);
+            List<QueueId> allQueues = queueMapper.GetAllQueues().ToList();
 
-                await Task.WhenAll(deleteTasks);
+            var deleteTasks = new List<Task>();
+            foreach (var queueId in allQueues)
+            {
+                var manager = new AzureQueueDataManager(loggerFactory, queueId.ToString(), storageConnectionString);
```
jason-bragg: The queue name needs to map to a queueId but does not need to be the name of the queueId.  QueueId is unnecessarily constrained.  A team with an existing set of queues "A", "B", "C", which they wanted to process using Orleans streams would not be supported by queue names constrained by QueueId naming.

jason-bragg:
Was there an issue for this?  If so, can we link it to this PR plz.

xiazen:
for issue #4577

src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamBuilder.cs
xiazen: I don't think we should handle this override in the configuration stage. It makes it difficult to override the behavior.

What behavior you want to override here? And why exactly `AzureQueueAdapterFactory` is the better place to configure the azure queue name, instead of the AzureQueueOptions. I mean I can argue the same for `AzureTableStorageOptions.TableName`, why is the table name in the AzureTableStorageOptions, instead of the `AzureTableGrainStorageFactory `

src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamProviderUtils.cs
xiazen: I'm not sure what feature you are suggesting here. If a team want to configure azure queue name completely, then they need to give a list of their queue name to AzureQueueStreamProvider which has the same queue amount of the queues configured with AzureQueueStreamProvider. This requires user to have certain level of understanding for AzureQueueStreamProvider. 

So let's say there's a new user just want to configure a AzureQueueStreamProvider with connection string and play with it after, how do we do in this situation? 

Just btw you might missed the context of this PR. The goal for this PR is to fix #4577 , which is to not use unsupported symbol in Azure queue name , while the old queue name is serviceId-providerName by default, so configuring azure queue name is hard. So serviceId and providerName should contain unsupported symbol. This is hard to know from the user's side since it is implementation details. So I think this PR solve that problem

xiazen:
update with the ability to configure azure queue name entirely.

xiazen:

src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamBuilder.cs
```diff
@@ -31,17 +32,12 @@ public SiloAzureQueueStreamConfigurator<TDataAdapter> ConfigureAzureQueue(Action
...
             return this;
         }
+
         public SiloAzureQueueStreamConfigurator<TDataAdapter> ConfigureCache(int cacheSize = SimpleQueueCacheOptions.DEFAULT_CACHE_SIZE)
         {
             this.Configure<SimpleQueueCacheOptions>(ob => ob.Configure(options => options.CacheSize = cacheSize));
             return this;
         }
-
-        public SiloAzureQueueStreamConfigurator<TDataAdapter> ConfigurePartitioning(int numOfPartition = HashRingStreamQueueMapperOptions.DEFAULT_NUM_QUEUES)
```
jason-bragg: Suggest leaving this or adding something like it GenerateQueueNames(..)? to help maintain the original names.  Basically a helper functions that uses AzureQueueUtilities.GenerateQueueNames to build the names for users azure queues.

jason-bragg: Or maybe add a null check on the names to generate the names if none are provided by the user.

xiazen:
we can add default behavior back. Just think this way is cleaner. But open to adding default queue names back option

src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/AzureQueueStreamBuilder.cs
xiazen: I was planning to do the later, if queue names are null , then configure default ones using PostConfigure, just as what I used in my first pass of this PR.

xiazen: I think if we want to retain the original behavior , which is generate a set of queue names by default, we should use PostConfigure, just as how our other default options values are configured. Configure through a GenerateQueueName() method seems clumsy and not really retaining original behavior, since user still need to call it.

xiazen:
I fixed the feedback two days ago but pushed the commit to the wrong branch. But they are all here now. Please take a look

jason-bragg:

xiazen:
test all passed !

Issue #4612, opened 2018-05-20
WriteStateAsync() works a lot slower since upgrading to Orleans2


Gabrn:
Since upgrading to Orleans2 we have had general slowness in writing grain state. We tried to pin down the issue and we gauged the time it took for us to write jsons of about 1mb each to azure blob storage and the mean went up from about 100ms to 400ms in peak times. 
We looked in the blob storage to see if the data got any larger but it was about the same size (1mb).
We suspected the upgrade to orleans2 to be the cause and when we downgraded back to orleans1 the write time got back to normal (this was the only change we made) so if the cause was something else the downgrade should not have effected this.

This is the graph of the `writeStateAsync()` time from one of our grains "expert-grain":
![screen shot 2018-05-20 at 15 11 21](https://user-images.githubusercontent.com/10612685/40278772-1790bb16-5c40-11e8-902e-57e1159c0be7.png)

We noticed that after restarting the nodes the write time goes down for a few hours before going up again (we used a script to restart the nodes every 4 hours for that reason).

Any suggestions to how we can investigate this further?

sergeybykov:
Sorry about delayed response.

Are you running on .NET Core or Full Framework?

In theory, the Orleans runtime shouldn't get in the way of writing state to storage. That's almost a straight pass-through.

There was an upgrade in `WindowsAzure.Storage` from 7.0 to 8.2.1. Can you try running the Orleans 1.x version (BTW, which one are you running there) with `WindowsAzure.Storage` 8.2.1 to eliminate the possibility that this upgrade caused something?

Are you seeing any memory leak in those few hours?

talarari:
Hi @sergeybykov, i work with @Gabrn, Looks like were using WindowsAzure.Storage 8.6.0.
latest version is 9.2.0.
We're using orleans 1.5.2

sergeybykov:
So with WAS 8.6.0 1.5.2 runs fine but 2.0 with the same 8.6.0 on Full .NET Framework runs slower?

talarari:
We're experiencing some slowdown for writeStateAsync on 1.5.2 as well. It might be the same problem but I'm not sure.

As we get higher traffic we make a lot of writeStateAsync calls, we're using azure blob storage.
We're seeing write requests to the blob storage average at 150 ms but getting slower 3+ seconds every now and then.
[...]

PR #4763, opened 2018-07-11
Added grain storage to benchmarks

jason-bragg:
Added a grain storage benchmark to help track down #4612.

Unfortunately, not able to repro the issue (yet).

### DIFF
--- a/test/Benchmarks/GrainStorage/GrainStorageBenchmark.cs
+++ b/test/Benchmarks/GrainStorage/GrainStorageBenchmark.cs
@@ -1,0 +1,114 @@
+﻿using System;
+using System.Threading.Tasks;
+using System.Diagnostics;
+using Orleans.Hosting;
+using Orleans.TestingHost;
+using TestExtensions;
+using BenchmarkGrainInterfaces.GrainStorage;
+
+namespace Benchmarks.GrainStorage
+{
+    public class GrainStorageBenchmark
+    {
+        private TestCluster host;
+
+        public void MemorySetup()
+        {
+            var builder = new TestClusterBuilder();
+            builder.AddSiloBuilderConfigurator<SiloMemoryStorageConfigurator>();
+            this.host = builder.Build();
+            this.host.Deploy();
+        }
+
+        public void AzureTableSetup()
+        {
+            var builder = new TestClusterBuilder();
+            builder.AddSiloBuilderConfigurator<SiloAzureTableStorageConfigurator>();
+            this.host = builder.Build();
+            this.host.Deploy();
+        }
+
+        public void AzureBlobSetup()
+        {
+            var builder = new TestClusterBuilder();
+            builder.AddSiloBuilderConfigurator<SiloAzureBlobStorageConfigurator>();
+            this.host = builder.Build();
+            this.host.Deploy();
+        }
+
+        public class SiloMemoryStorageConfigurator : ISiloBuilderConfigurator
+        {
+            public void Configure(ISiloHostBuilder hostBuilder)
+            {
+                hostBuilder.AddMemoryGrainStorageAsDefault();
+            }
+        }
+
+        public class SiloAzureTableStorageConfigurator : ISiloBuilderConfigurator
+        {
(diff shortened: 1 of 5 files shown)


test/Benchmarks/GrainStorage/GrainStorageBenchmark.cs
```diff
@@ -0,0 +1,114 @@
...
+            IPersistentGrain persistentGrain = this.host.Client.GetGrain<IPersistentGrain>(Guid.NewGuid());
+            // activate grain
+            await persistentGrain.Set(0);
+            int stored = 0;
+            int failed = 0;
+            TimeSpan max = TimeSpan.MinValue;
+            Stopwatch sw = Stopwatch.StartNew();
+            while (running())
+            {
+                Stopwatch time = Stopwatch.StartNew();
```
ReubenBond: Maybe re-use the stopwatch. It shouldn't matter, but generally we should keep benchmarking code as light as possible.

test/Benchmarks/Program.cs
```diff
@@ -86,6 +87,45 @@ class Program
...
+                    benchmark.AzureTableSetup();
+                    return benchmark;
+                },
+                benchmark => benchmark.RunAsync().GetAwaiter().GetResult(),
+                benchmark => benchmark.Teardown());
+            },
+            ["GrainStorage.AzureBlob"] = () =>
+            {
+                RunBenchmark(
+                "Running Transactions benchmark",
```
ReubenBond: Names here need fixing

test/Grains/BenchmarkGrains/GrainStorage/PersistentGrain.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public async Task<int> Set(int value)
+        {
+            Stopwatch sw = Stopwatch.StartNew();
+            try
+            {
+                this.State = value;
+                await this.WriteStateAsync();
+                sw.Stop();
+                object[] args = { this.GetPrimaryKey(), sw.ElapsedMilliseconds };
```
ReubenBond: I don't quite understand the reasoning behind an explicit array, but that's fine.

ReubenBond:
Mostly just the text for the benchmarks

test/Grains/BenchmarkGrains/GrainStorage/PersistentGrain.cs
jason-bragg: Stupid heap analyzer was complaining about everything.. I turned it off :/  This code was written while it was still on and being neurotic.

test/Benchmarks/GrainStorage/GrainStorageBenchmark.cs
```diff
@@ -83,27 +83,28 @@ public void Configure(ISiloHostBuilder hostBuilder)
...
             int stored = 0;
             int failed = 0;
-            TimeSpan max = TimeSpan.MinValue;
+            TimeSpan maxCalltime = TimeSpan.MinValue;
             Stopwatch sw = Stopwatch.StartNew();
+            Stopwatch calltime = new Stopwatch();
             while (running())
             {
-                Stopwatch time = Stopwatch.StartNew();
+                calltime.Start();
```
ReubenBond: Restart?

test/Grains/BenchmarkGrains/GrainStorage/PersistentGrain.cs
ReubenBond: hahaha, yeah, my bad. It is definitely too picky. At most it should just give you friendly hints, never warnings.

ReubenBond: Instead, maybe we can add it to the solution with a rules file which reduces the severity of all of its diagnostics to the lowest visible level (which doesn't show up in build output - if that exists). Obviously you need to allocate some times and sometimes you simply don't care (since it's not performance critical code).

For the test subtree, we should have it disabled.

PR #4764, opened 2018-07-12
Remove all instances of #region

ReubenBond:
`#region` is yuck. There are better ways to organize code and if a particular file is so massive as to warrant them, then developers should feel that pain and fix the core issue rather than hiding it behind `#region` directives.


ReubenBond:
@dotnet-bot test functional

jason-bragg (approved):
Ok with this change, but this feels a bit like format-wars, which never ends well..

Note: In the comments for this change, it was argued

> “if a particular file is so massive as to warrant them, then developers should feel that pain and fix the core issue rather than hiding it behind #region directives”.

This change doesn’t cleanup any of the complexity that may have led developers to use regions, it just removes the regions.  If code was so massive as to warrant them for clarity, this PR will harm clarity.

ReubenBond:
@dotnet-bot test functional

ReubenBond:
I find that regions harm clarity, since they hide the code.

The vocal parts of the .NET community are fairly unanimous that they don't want regions any more, and I'm in agreement.

PR #4766, opened 2018-07-12
Changed link to documentation 

JillHeaden:
Changed link to go directly to Documentation Intro page, skipping "secondary" page

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Then either restart VS, or go to the task manager and kill the processes that st
 Documentation
 =============
 
-Documentation is located [here](http://dotnet.github.io/orleans/)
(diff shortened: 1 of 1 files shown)


JillHeaden:
@jason-bragg

jason-bragg:
Is all the information on the secondary page covered on the main page or somewhere easy to find?
I agree we've a lot of duplicated information and should get rid of the second page, just want to make sure all the info is covered elsewhere before moving the page.

xiazen:
I kinda like the main page as the first page people see, since it highlighted the three important points about orleans: scalability, low latency and simplified concurrency. And also it touched the simple grain interface. I like the main page better since it serves better as a marking material. It is simple, straightforward to highlight all the points user will be interested in orleans.

PR #4768, opened 2018-07-13
Transaction participant wrapper json serialization fix

jason-bragg:
Transaction wrapper was not serializing the participant when using json.


xiazen:
@dotnet-bot test this please

xiazen:
do you need to serialize `localParticipants ` in this case? mark it as json property? Is this the transactional state you are referring? 
```
  public class TransactionParticipantExtension : ITransactionParticipantExtension
    {
        private readonly Dictionary<string, ITransactionParticipant> localParticipants = new Dictionary<string, ITransactionParticipant>();
}
```

PR #4776, opened 2018-07-17
Fix partial build

xiazen:
#4739  introduced breaking change in Orleans.Core.Abstraction, Orleans.Core, Orleans.Transaction without set the new package version correctly, which broke partial build.


xiazen:
local full build succeed

PR #4780, opened 2018-07-18
Bump orleans package version to 2.1

xiazen:
We recently introduced a breaking change in Core.Abstraction, which means we need to bump every package version to 2.1 as well. More info in #4776


xiazen:
@dotnet-bot test this please

ReubenBond:
Everyone fine with this?

If not, we can revert it later 😉

PR #4781, opened 2018-07-18
More Transaction Metrics

jason-bragg:
Added more transaction metrics, and benchmark tests for transaction throttling.

### DIFF
--- a/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
+++ b/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
@@ -68,16 +68,22 @@ namespace Orleans.Transactions
                 }
             }
 
-            TransactionalStatus status = (writeParticipants == null)
-                ? await CommitReadOnlyTransaction(transactionInfo)
-                : await CommitReadWriteTransaction(transactionInfo, writeParticipants);
-
-            if (status == TransactionalStatus.Ok)
-                this.statistics.TrackTransactionSucceeded();
-            else
+            try
+            {
+                TransactionalStatus status = (writeParticipants == null)
+                    ? await CommitReadOnlyTransaction(transactionInfo)
+                    : await CommitReadWriteTransaction(transactionInfo, writeParticipants);
+                if (status == TransactionalStatus.Ok)
+                    this.statistics.TrackTransactionSucceeded();
+                else
+                    this.statistics.TrackTransactionFailed();
+                return status;
+            }
+            catch (Exception)
+            {
                 this.statistics.TrackTransactionFailed();
-
-            return status;
+                throw;
+            }
         }
 
         private async Task<TransactionalStatus> CommitReadOnlyTransaction(TransactionInfo transactionInfo)
--- a/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
+++ b/src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
@@ -2,56 +2,14 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading.Tasks;
+using System.Linq;
+using Microsoft.Extensions.Logging;
 using Orleans.Runtime;
 using Orleans.Concurrency;
-using System.Linq;
-using System.Threading;
-using System.Transactions;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
-using Orleans.Configuration;
 using Orleans.Transactions.Abstractions;
 
 namespace Orleans.Transactions
 {
-    public class TransactionAgentStatistics
-    {
-        public long TransactionStartedCounter { get; set; }
-
-        private const string TransactionStartedPerSecondMetric = "TransactionAgent.TransactionStartedPerSecond";
-        //Transaction started recorded at when metrics was reported last time
-        private long transactionStartedAtLastReported;
-        private DateTime lastReportTime;
-        private readonly ITelemetryProducer telemetryProducer;
-        private readonly PeriodicAction monitor;
-        public TransactionAgentStatistics(ITelemetryProducer telemetryProducer, IOptions<StatisticsOptions> options)
-        {
-            this.telemetryProducer = telemetryProducer;
-            this.lastReportTime = DateTime.UtcNow;
-            this.monitor = new PeriodicAction(options.Value.PerfCountersWriteInterval, this.ReportMetrics);
-        }
-
-        public void TryReportMetrics(DateTime timestamp)
-        {
-            this.monitor.TryAction(timestamp);
-        }
-
-        private void ReportMetrics()
-        {
-            var now = DateTime.UtcNow;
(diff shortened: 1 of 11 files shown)


src/Orleans.Transactions/DistributedTM/TransactionOverloadDetector.cs
```diff
@@ -34,28 +31,29 @@ public class TransactionRateLoadSheddingOptions
 
     public class TransactionOverloadDetector : ITransactionOverloadDetector
     {
-        private readonly TransactionAgentStatistics statistics;
+        private readonly ITransactionAgentStatistics statistics;
         private readonly TransactionRateLoadSheddingOptions options;
         private readonly PeriodicAction monitor;
-        private long transactionStartedAtLastCheck;
+        private TransactionAgentStatistics lastStatistics;
```
xiazen: this should be ITransactionAgentStatistic too?

src/Orleans.Transactions/DistributedTM/TransactionAgentStatistics.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            Interlocked.Increment(ref this.transactionsFailed);
+        }
+
+        public void TrackTransactionThrottled()
+        {
+            Interlocked.Increment(ref this.transactionsThrottled);
+        }
+
+        public static TransactionAgentStatistics Create(ITransactionAgentStatistics initialStatistics)
```
xiazen: why isn't this return a `ITransactionAgentStatistics` as well?

src/Orleans.Transactions/DistributedTM/TransactionOverloadDetector.cs
```diff
@@ -66,8 +64,7 @@ public bool IsOverloaded()
 
             DateTime now = DateTime.UtcNow;
             this.monitor.TryAction(now);
-            long startCounter = this.statistics.TransactionStartedCounter;
-            double txPerSecondCurrently = CalculateTps(this.transactionStartedAtLastCheck, this.lastCheckTime, startCounter, now);
+            double txPerSecondCurrently = CalculateTps(this.lastStatistics.TransactionsStarted, this.lastCheckTime, this.statistics.TransactionsStarted, now);
```
xiazen: where do you track ITransactionAgentStatistic.Throttled ? did I miss anything. I thought it will be tracked here

jason-bragg: It's tracked in the transaction agent, where the IsOverloaded is called.  The agent is responsible for updating it's statistics.

src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
```diff
@@ -108,14 +68,16 @@ public Task<TransactionalStatus> Commit(ITransactionInfo info)
                 }
             }
 
-            if (writeParticipants == null)
-            {
-                return CommitReadOnlyTransaction(transactionInfo);
-            }
+            TransactionalStatus status = (writeParticipants == null)
```
ReubenBond: Does there need a try/catch to ensure we track failed txns correctly, or are these async methods 'guaranteed' to not throw?

src/Orleans.Transactions/DistributedTM/TransactionAgentStatistics.cs
ReubenBond: This method seems more like a copy method than a create method

src/Orleans.Transactions/DistributedTM/TransactionAgentStatisticsReporter.cs
```diff
@@ -0,0 +1,87 @@
...
+        private const string TransactionsStartedPerSecondMetric = "TransactionAgent.TransactionsStarted_PerSecond";
+
+        private const string SuccessfulTransactionsTotalMetric = "TransactionAgent.SuccessfulTransactions_Total";
+        private const string SuccessfulTransactionsPerSecondMetric = "TransactionAgent.SuccessfulTransactions_PerSecond";
+
+        private const string FailedTransactionsTotalMetric = "TransactionAgent.FailedTransactions_Total";
+        private const string FailedTransactionsPerSecondMetric = "TransactionAgent.FailedTransactions_PerSecond";
+
+        private const string ThrottledTransactionsTotalMetric = "TransactionAgent.ThrottledTransactions_Total";
+        private const string ThrottledTransactionsPerSecondMetric = "TransactionAgent.ThrottledTransactions_PerSecond";
```
ReubenBond: Convention is to use only `.` as separators - see [StatisticNames.cs](https://github.com/dotnet/orleans/blob/3d2bf21e1f43fb24ec8bcb81aae9e6165c797a3a/src/Orleans.Core/Statistics/StatisticNames.cs#L107)

src/Orleans.Transactions/DistributedTM/TransactionAgentStatisticsReporter.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+
+        private void ReportMetrics(object ignore)
+        {
+            var now = DateTime.UtcNow;
+            var currentReported = TransactionAgentStatistics.Create(statistics);
+
+            this.telemetryProducer.TrackMetric(TransactionsStartedTotalMetric, currentReported.TransactionsStarted);
+            this.telemetryProducer.TrackMetric(TransactionsStartedPerSecondMetric, PerSecond(this.lastReported.TransactionsStarted, currentReported.TransactionsStarted, now - lastReportTime));
```
ReubenBond: Could extract `var reportPeriod = now - lastReportTime`

src/Orleans.Transactions/Hosting/SiloBuilderExtensions.cs
```diff
@@ -13,25 +13,24 @@ public static class SiloBuilderExtensions
         /// <summary>
         /// Configure cluster to use the distributed TM algorithm
         /// </summary>
-        public static ISiloHostBuilder UseDistributedTM(this ISiloHostBuilder builder)
+        public static ISiloHostBuilder UseDistributedTM(this ISiloHostBuilder builder, bool withReporter = true)
```
ReubenBond: This new parameter is not obvious and hence could benefit from a doc comment.

Slightly prefer naming it `reportStatistics`.

I'm not entirely convinced that this needs to be exposed - or could we implement this using the options pattern somehow?

jason-bragg: Issue here is that users can provide their own statistics reporting.  So they can configure tm with reporting, or without, and provide their own.  I'll add a comment.
As for being part of the options, the Tm has no options now and this is really about what we're setting up in the container, not how the TM is configured. :/

jason-bragg:

veikkoeeva:
I'll link as tangential to https://github.com/dotnet/orleans/issues/4805#issuecomment-408969770 since it appears I just linked an example doing a bit of the same.

jason-bragg:

Issue #2106, opened 2016-08-31
Automate building docs with DocFX


jdom:
Now that DocFX was integrated by #1970, it requires a few manual steps to build the HTML documentation from the markdown files. We should automate this process to reduce friction.
We can split the documentation source (markdown files) from the generated html into 2 different branches (and potentially even put both of them in a new orleans-docs repo, but we'd need to sort out the domain change issue in that case, so it's not a pre-req). Let's call them `docs` and `gh-pages` respectively for now.
The script can be triggered by Jenkins every time the `docs` branch changes, and should:
1. Pull the `gh-pages` branch into a clone in a working folder
2. Delete all the generated HTML from it
3. Run docfx so that it builds and puts the generated HTML on that `gh-pages` clone
4. Add all files and commit
5. Push to `gh-pages`

Sounds reasonable?

After we have this basic automation in-place, we can take a look at doing something with #88 to generate API docs from the XML comments too.

jason-bragg:
All for auto-dox, though I'm concerned about updating the main doc repository to point to auto-generated dox from the latest code.  I think we'll need main docs to point to the generated docs of the latest released version (starting with 1.3).

I suggest keeping auto generated docs per release (again starting with 1.3) and a link to the unreleased latest docs that can be updated as described above.

jdom:
Agreed, I envision this in a similar way that the script can fetch the changes from either `master` or some other branch or tag, so we can build docs whatever version of the API. Still, that's something we should discuss in #88 after we have this particular issue implemented, which is not for generating the API from code, but just the docs from the markdown files.

erictuvesson:
@sergeybykov + @jason-bragg, what is the status on this and https://github.com/dotnet/orleans/issues/88?

Considering that there is a lot of documentation on the `docs` branch using Docfx.

sergeybykov:
No progress on API docs. Any help would be much appreciated. 🙂

ReubenBond:
Closing as duplicate (close enough) of #88. By the way, I took a stab at this and couldn't prevent DocFx from crashing. Here's the issue: https://github.com/dotnet/docfx/issues/4341

PR #4791, opened 2018-07-23
Created Orleans.Documentation folder with new doc structure

JillHeaden:
I have created a new folder to contain the reorganized Orleans documentation. 
When compiled with DocFX, the new site is contained inside a folder next to the source. 

I will be delighted to demonstrate all of this at the next meeting.

@jason-bragg 
@ReubenBond


JillHeaden:
Removed compiled site folder containing HTML files, etc.

1.5/Documentation/Advanced-Concepts/Activation-Garbage-Collection.html
```diff
@@ -0,0 +1,231 @@
+﻿<!DOCTYPE html>
+<!--[if IE]><![endif]-->
```
xiazen: still html file here

xiazen: or is this supposed to be html files since it is 1.5 docs? If yes, suggest separate 1.5 docs and 2.0 docs into 2 pr to help review. This PR is too big I even have trouble loading.

jdom:
Is the plan to have the docs in the master branch? I strongly suggest we have them in a different repo (or at least in a different branch as it currently is), and then have the generated HTML in another branch altogether. So that the source of the documentation is updated in one branch, and then have automation to update the other branch.

But having either source of docs or the html itself in the master branch of the code seems like a bad idea because it's unlikely that they will ever be in sync with what's currently checked in. Also the docs might contain documentation about multiple versions...
Also, I believe we would be in a better shape to host the docs in docs.microsoft.com (although that's pure speculation on my part, from what I've seen in the organization structure of other MS projects).

BTW, these thoughts are related to https://github.com/dotnet/orleans/issues/2106

Just my 2 cents.

jdom:
Also, be extra careful with binary files (such as pptx or high res images), as having them in the main repo (and worse in the main code branch) increases the repo size considerably, and the updates to them will keep each full version of it in the git history forever.

jason-bragg:
@jdom, The plan, atm, is to keep the doc source in the same repo and branch as the code source and keep them updated together as much as is practicable.  A build step will be added to generate and publish the html.

"It's unlikely that they will ever be in sync with what's currently checked in"
The doc source 'should' be kept up to date with what's checked in, but the published documentation will be updated with nuget releases.

"Also the docs might contain documentation about multiple versions..."
The plan is to publish docs with each released nuget version, rather than have a single doc that is expected to cover the nuances of every release.

Keeping the doc source with the code source also makes it simpler to generate api references from the source and allows us to require updates to documentation as part of any change to the public surface in the code review process.

jdom:
> The doc source 'should' be kept up to date with what's checked in, but the published documentation will be updated with nuget releases.

Would that mean that there will be no way to update the docs after a package gets released? (for example, add some explanation to how to migrate something from an older version, or even documentation that applies to a previous release?)

> "Also the docs might contain documentation about multiple versions..."
The plan is to publish docs with each released nuget version, rather than have a single doc that is expected to cover the nuances of every release.

So the docs are only browsable locally or will there be a published version online? Will that contain only the latest? What will be the experience to browse and search for something of a previous release?
BTW, I assume this means publishing to docs.microsoft.com is no longer a goal?

> Keeping the doc source with the code source also makes it simpler to generate api references from the source and allows us to require updates to documentation as part of any change to the public surface in the code review process.

Agree with the latter, although the former is extremely easy to do with the automation that is already planned to update the generated HTML branch.

One note though, I wouldn't make this decision lightly thinking that later on you can move it to a different repo when there's bandwidth (since you mention that's the plan *ATM*), because it will mean that this git history will remain in the code branch forever, and it will be especially harmful since I see a lot of binaries in the PR.

jason-bragg:
>Would that mean that there will be no way to update the docs after a package gets released?

I don't see why that would be the case.  We can update the release branch like we can for any source, then rebuil the published docs.

>So the docs are only browsable locally or will there be a published version online? 

The documents for the most recently released version will definitely be published.  Older versions are debatable, but I don't see why we'd not publish online docs for every version of orleans we're supporting.

>What will be the experience to browse and search for something of a previous release?

I suspect we'd just provide links to the docs for all supported versions, organized by version.  Not sure.  Still working most of this out.  Having docs for a specific version is definitely lighter for the maintainers and probably more clear for the community.

>I see a lot of binaries in the PR.

I wasn't really concerned about binaries, but yea, we should give that some thought.  We may want to keep binaries and other assets in another branch.  They are content, not source anyway.

PR #4796, opened 2018-07-24
Verify exception handling logic when multiple particiant throw

xiazen:
Verify exception handling logic when multiple particiant throw. The expected logic is only one exception is picked as the inner exception of OrleansTransactionAbortException


xiazen:
can we merge it ?!

PR #4799, opened 2018-07-26
fix/improve transactional state behavior after InconsistentStateException

sebastianburckhardt:
As mentioned by @xiazen a couple weeks ago, the transactional state does not quite do the right thing when catching `InconsistentStateException`. 

This PR adds the following behavior to a transactional state facet that catches an `InconsistentStateException`:

1. calls `DeactivateOnIdle`
2. aborts all transactions in the wait queue (i.e. transactions that have not accessed the state yet but are queued up waiting to enter the lock)

### DIFF
--- a/src/Orleans.Transactions/State/ReaderWriterLock.cs
+++ b/src/Orleans.Transactions/State/ReaderWriterLock.cs
@@ -471,6 +471,27 @@ namespace Orleans.Transactions
             }
         }
 
+        // aborts all queued transactions that are not in the lock yet
+        private void AbortQueuedTransactions()
+        {
+            var pos = currentGroup?.Next;
+            while (pos != null)
+            {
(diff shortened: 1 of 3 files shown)


src/Orleans.Transactions/State/ReaderWriterLock.cs
```diff
@@ -471,6 +471,27 @@ private void AbortExecutingTransactions(string indication)
...
+        {
+            var pos = currentGroup?.Next;
+            while (pos != null)
+            {
+                if (pos.Tasks != null)
+                {
+                    foreach (var t in pos.Tasks)
+                    {
+                        // running the task will abort the transaction because it is not in currentGroup
+                        t.RunSynchronously();
```
xiazen: is there a better to abort the transaction ? this seems hacky

PR #4810, opened 2018-08-01
Limited separation of transaction orchestration and participation.

jason-bragg:
First pass at separating transaction orchestration (transaction manager) from participation (transactional resource).

Goal is to eventually be able to support a variety of transactional resources (state, queues, event sourcing, ...) as well as advanced transactional patterns like Transfer of Commit (TOC).

This PR should have no behavioral changes, only code reorganization.
- Added transaction manager and transactional resource interfaces
- Removed partial classes, separating logic into a read/write lock, and transaction queue, transaction manager and transactional resource.

More work needs be done, this is just a first pass.


src/Orleans.Transactions/Abstractions/ITransactionParticipant.cs
```diff
@@ -8,28 +8,58 @@ namespace Orleans.Transactions.Abstractions
...
-        /// Participants respond after committing or aborting the read.
+        /// One-way message sent by TA to all participants except TM.  
         /// </summary>
         /// <param name="transactionId">the id of the transaction to prepare</param>
         /// <param name="accessCount">number of reads/writes performed on this participant by this transaction</param>
         /// <param name="timeStamp">the commit timestamp for this transaction</param>
+        /// <param name="transactionManager">the transaction manager for this transaction</param>
         /// <returns></returns>
-        Task<TransactionalStatus> CommitReadOnly(Guid transactionId, AccessCounter accessCount, DateTime timeStamp);
+        Task Prepare(Guid transactionId, AccessCounter accessCount,
```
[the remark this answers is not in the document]
jason-bragg: These are not grain interfaces or extensions, so attribution doesn't really make sense here.
Should probably not have the one-way noted in the comment, or if we do, we should return void.
Will clean up call patterns in future pass.

jason-bragg:

src/Orleans.Transactions/State/TransactionManager.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace Orleans.Transactions.State
+{
+    internal class TransactionManager<TState> : ITransactionManager
+               where TState : class, new()
+    {
+        private readonly TransactionQueue<TState> queue;
+
+        public TransactionManager(TransactionQueue<TState> queue)
+        {
+            this.queue = queue;
```
ReubenBond: nit: consider adding a null check here

src/Orleans.Transactions/State/ReaderWriterLock.cs
```diff
@@ -252,6 +160,201 @@ private Task LockWork()
...
+                {
+                    TransactionId = transactionId
+                };
+                status = TransactionalStatus.BrokenLock;
+                return false;
+            }
+            else if (record.NumberReads != accessCount.Reads
+                   || record.NumberWrites != accessCount.Writes)
+            {
+                Rollback(transactionId, true);
```
ReubenBond: We dropped a message here - maybe at some point in future we should add Trace-level logging here instead.

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -4,338 +4,215 @@
...
+        private readonly TransactionalStateOptions options;
+        private readonly ITransactionParticipant resource;
+        private readonly Action deactivate;
+        private readonly ITransactionalStateStorage<TState> storage;
+        private readonly JsonSerializerSettings serializerSettings;
+        private readonly BatchWorker storageWorker;
+        private readonly BatchWorker confirmationWorker;
+        private readonly ILogger logger;
+        private readonly Dictionary<Guid, TransactionRecord<TState>> confirmationTasks;
+        private CommitQueue<TState> commitQueue;
```
ReubenBond: This line irks me so much. The field is never assigned. This only works because `CommitQueue<TState>` is a struct. It's unintuitive that this could be correct.

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -365,9 +242,9 @@ private void NotifyOfAbort(TransactionRecord<TState> entry, TransactionalStatus
 
                         // tell remote participants
                         foreach (var p in entry.WriteParticipants)
-                            if (p != thisParticipant)
+                            if (p != resource)
```
ReubenBond: This should be `!p.Equals(resource)`

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -867,7 +739,7 @@ private async Task ConfirmationTask(TransactionRecord<TState> record)
 
                 foreach (var p in record.WriteParticipants)
                 {
-                    if (p != thisParticipant)
+                    if (p != resource)
```
ReubenBond: `.Equals`, since `ITransactionParticipant` is `IEquatable`

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -1,236 +1,296 @@
...
         private ITransactionParticipant thisParticipant;
-
-        // storage
-        private ITransactionalStateStorage<TState> storage;
+        private TransactionQueue<TState> queue;
+        private TransactionalResource<TState> resource;
+        private TransactionManager<TState> transactionManager;
 
         private string stateName;
         private string StateName => stateName ?? (stateName = StoredName());
```
ReubenBond: This is never used

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -1,236 +1,296 @@
...
-                    var tm = (pr.TransactionManager == null) ? null :
-                        (ITransactionParticipant) JsonConvert.DeserializeObject<ITransactionParticipant>(pr.TransactionManager, this.serializerSettings);
+                        logger.Debug($"update-lock write v{record.SequenceNumber} {record.TransactionId} {record.Timestamp:o}");
+
+                    // record this write in the transaction info data structure
+                    info.RecordWrite(this.thisParticipant, record.Timestamp);
 
-                    commitQueue.Add(new TransactionRecord<TState>()
+                    // record this participant as a TM candidate
+                    if (info.TMCandidate != resource)
```
ReubenBond: .Equals

src/Orleans.Transactions/State/TransactionalState.cs
```diff
@@ -1,236 +1,296 @@
...
+                        }
+                    }
+
+                    // perform the write
+                    TResult result = default(TResult);
+                    try
+                    {
+                        detectReentrancy = true;
+
+                        if (updateAction != null)
```
ReubenBond: This is always true, since it's guarded with an exception at the top of the method

ReubenBond:
LGTM. I mostly verified that this is isomorphic to the existing code.
There are some comments which you may want to address before merging but if you prefer to tackle them later then that's fine, too.

ReubenBond:
I pushed suggested changes - same as with the other PR: if you feel that they are appropriate then I'm happy to merge. Otherwise feel free to revert.

PR #4815, opened 2018-08-03
Auto-installing grain extensions

jason-bragg:
Grains can now load grain extension from the container upon request rather than requiring the grain to register them during activation.

### DIFF
--- a/src/Orleans.Runtime.Abstractions/Hosting/HostingGrainExtensions.cs
+++ b/src/Orleans.Runtime.Abstractions/Hosting/HostingGrainExtensions.cs
@@ -1,0 +1,16 @@
+﻿using Orleans.CodeGeneration;
+using Orleans.Runtime;
+
+namespace Orleans.Hosting
+{
+    public static class HostingGrainExtensions
+    {
+        public static ISiloHostBuilder AddGrainExtension<TExtensionInterface,TExtension>(this ISiloHostBuilder builder)
+            where TExtensionInterface : class, IGrainExtension
+            where TExtension : class, TExtensionInterface
+        {
+            int interfaceId = GrainInterfaceUtils.GetGrainInterfaceId(typeof(TExtensionInterface));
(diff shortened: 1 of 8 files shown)


ReubenBond:
You're the best, thank you for implementing this! I'll try to take a look ASAP and also review the txn manager separation code.

ReubenBond:
@jason-bragg I made some changes rather than leave comments. If you feel that the changes are appropriate then we can merge.

src/Orleans.Runtime.Abstractions/Hosting/HostingGrainExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// Methods for configuring <see cref="IGrainExtension"/>s on a silo.
+    /// </summary>
+    public static class HostingGrainExtensions
+    {
+        /// <summary>
+        /// Registers a grain extension implementation for the specified interface.
+        /// </summary>
+        /// <typeparam name="TExtensionInterface">The <see cref="IGrainExtension"/> interface being registered.</typeparam>
+        /// <typeparam name="TExtension">The implementation of <typeparamref name="TExtensionInterface"/>.</typeparam>
+        public static ISiloHostBuilder AddGrainExtension<TExtensionInterface, TExtension>(this ISiloHostBuilder builder)
```
jdom: Should we provide a way to filter in which grain types (and/or instances thinking about Orleankka) the extension is installed?

jason-bragg: I suspect there will be cases where users will want more control over which grains extensions are installed on.  We could probably do this via silo configuration, attribution, or some sort of requirements or validation check..  I suggest we expand this feature as needed, rather than speculate what users will need.   In any case, this core capability is a prerequisite to any more intelligent installation of extensions.

PR #4816, opened 2018-08-03
[WIP] Add memory threshold

COCPORN:
A naive implementation that allows for configurable suppression of grain collection based on memory pressure.

### DIFF
--- a/src/Orleans.Runtime/Catalog/Catalog.cs
+++ b/src/Orleans.Runtime/Catalog/Catalog.cs
@@ -34,7 +34,7 @@ namespace Orleans.Runtime
 
             public NonExistentActivationException() : base("NonExistentActivationException") { }
             public NonExistentActivationException(string msg) : base(msg) { }
-            public NonExistentActivationException(string message, Exception innerException) 
+            public NonExistentActivationException(string message, Exception innerException)
                 : base(message, innerException) { }
 
             public NonExistentActivationException(string msg, ActivationAddress nonExistentActivation, bool isStatelessWorker)
@@ -96,6 +96,7 @@ namespace Orleans.Runtime
         private readonly ILoggerFactory loggerFactory;
         private readonly IOptions<GrainCollectionOptions> collectionOptions;
         private readonly IOptions<SiloMessagingOptions> messagingOptions;
+        private readonly long memoryThresholdInMegabytes;
         public Catalog(
             ILocalSiloDetails localSiloDetails,
             ILocalGrainDirectory grainDirectory,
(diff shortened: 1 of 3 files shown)


src/Orleans.Runtime/Configuration/Options/GrainCollectionOptions.cs
```diff
@@ -25,5 +25,8 @@ public class GrainCollectionOptions
         /// Period of inactivity necessary for a grain to be available for collection and deactivation by grain type.
         /// </summary>
         public Dictionary<string, TimeSpan> ClassSpecificCollectionAge { get; set; } = new Dictionary<string, TimeSpan>();
+
+        public long CollectionMemoryThresholdInMegabytes { get; set; } = DEFAULT_COLLECTION_MEMORY_THRESHOLD_IN_MEGABYTES;
```
ReubenBond: This needs a doc comment to explain the behavior.

When this value is non-zero, collection will be suspended until memory usage passes the specified threshold.

Additionally, does this force early collection if the memory usage is above the threshold? I believe that's the "least surprises" behavior.

ReubenBond:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test functional

src/Orleans.Runtime/Configuration/Options/GrainCollectionOptions.cs
COCPORN: The implementation is very simple at this point. I'll add some documentation, that was a stupid oversight. Right now it basically uses the instrumentation that is already in place to decide whether or not to do a collection of grains at all, and if it decides to do so it does so using the existing logic where it follows the normal rules of Orleans grain collection. That is, given that the current memory usage as returned by GC.GetTotalMemory() is greater than the configured threshold, it will run collection as usual, if not it will simply suppress collection.

src/Orleans.Runtime/Catalog/Catalog.cs
```diff
@@ -244,21 +246,31 @@ private async Task CollectActivationsImpl(bool scanStale, TimeSpan ageLimit = de
...
+                watch.Stop();
+                logger.Info(ErrorCode.Catalog_CollectionSuppressed,
+                    $"Collection suppressed#{number}: memory={memBefore}MB, #activations={activations.Count}, memoryPressureThreshold={memoryThresholdInMegabytes}, suppression time={watch.Elapsed}.");
+            }
+            else
             {
-                count = list.Count;
-                if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("CollectActivations{0}", list.ToStrings(d => d.Grain.ToString() + d.ActivationId));
-                await DeactivateActivationsFromCollector(list);
+                logger.Info(ErrorCode.Catalog_BeforeCollection, "Before collection#{0}: memory={1}MB, #activations={2}, collector={3}.",
```
benjaminpetit: Be careful, the new logging system used in 2.0 and forward does not support index based string for logging (yes we are still using this in some part of orleans…)

More infos: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.1#log-message-template

benjaminpetit:
Sorry for the delay., and thanks for your proposal!

After further consideration, we think that this behavior can be potentially very harmful, since it has some "cutoff" value.

Grains implementation can currently ask for their deactivation, and the idle time before automatic collection is configurable per type.

Feel free to discuss here if you disagree with this decision.

PR #4820, opened 2018-08-07
Separation of Transaction Manager and Transactional Resource

jason-bragg:
Separation of Transaction Manager and Transactional Resource
- Remove transaction participant
- Full separation of tm and resource grain extensions.
- Grain references are used everwhere and cast to extensions where needed.

Next step: Add TOC support


src/Orleans.Transactions/Abstractions/ITransactionManager.cs
```diff
@@ -0,0 +1,72 @@
...
+        Task Prepared(Guid transactionId, DateTime timeStamp, ParticipantId resource, TransactionalStatus status);
+
+        /// <summary>
+        /// One-way message sent by participants to TM, to let TM know they are still waiting to hear about
+        /// the fate of a transaction.
+        /// </summary>
+        /// <param name="transactionId">The id of the transaction</param>
+        /// <param name="timeStamp">The commit timestamp of the transaction</param>
+        /// <param name="resource">The participant sending the message</param>
+        Task Ping(Guid transactionId, DateTime timeStamp, ParticipantId resource);
```
ReubenBond: Should this `Prepared` above return void?

src/Orleans.Transactions/DistributedTM/TransactionInfo.cs
```diff
@@ -201,9 +194,40 @@ public override string ToString()
...
+                TMCandidate.Reference != null ? $" TM={TMCandidate.GetHashCode()}({TMBatchSize})" : ""
             );
         }
     }
+
+    public struct ParticipantId
+    {
+        public static IEqualityComparer<ParticipantId> Comparer = new IdComparer();
+
+        public GrainReference Reference;
```
ReubenBond: I think this struct is immutable, so we should express that in the definition.

With `Comparer`, we should also mark it `readonly`

src/Orleans.Transactions/DistributedTM/TransactionInfo.cs
```diff
@@ -201,9 +194,40 @@ public override string ToString()
...
+        public override string ToString()
+        {
+            return $"ParticipantId.{Reference}.{Name}";
+        }
+
+        private class IdComparer : IEqualityComparer<ParticipantId>
+        {
+            public bool Equals(ParticipantId x, ParticipantId y)
+            {
+                return x.Name == y.Name && Equals(x.Reference, y.Reference);
```
ReubenBond: This is ok, but preferable to explicitly use `StringComparer.Ordinal`.

src/Orleans.Transactions/State/ContextResourceFactoryExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Orleans.Transactions.State
+{
+    internal class ResourceFactoryRegistry<T> : Dictionary<string, Func<T>> { };
+
+    internal static class ContextResourceFactoryExtensions
+    {
+        public static void RegisterResourceFactory<T>(this IGrainActivationContext context, string name, Func<T> factory)
+        {
+            ResourceFactoryRegistry<T> registery = context.GetResourceFactoryRegistry<T>(true);
+            registery[name] = factory;
```
ReubenBond: typo: registry

src/Orleans.Transactions/State/ContextResourceFactoryExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+                ? (ResourceFactoryRegistry<T>)resourceFactoryRegistryObj
+                : default(ResourceFactoryRegistry<T>);
+            if(createIfNotExists && result == null)
+            {
+                context.Items[resourceFactoryRegistryName] = result ?? (result = new ResourceFactoryRegistry<T>());
+            }
+            return result;
+        }
+
+        private static string ResourceFactoryRegistryName<T>() => $"{typeof(T).FullName}+ResourceFactoryRegistry";
```
ReubenBond: I suppose this is just for uniqueness, but instead of .FullName, it's better to use `RuntimeTypeNameFormatter.Format(typeof(T))` because the results are cleaner (no asm versions/keys for generics).

src/Orleans.Transactions/State/TransactionManagerExtension.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (!this.managers.TryGetValue(resourceId, out ITransactionManager manager))
+            {
+                this.managers[resourceId] = manager = this.factories[resourceId].Invoke();
+            }
+            return manager.PrepareAndCommit(transactionId, accessCount, timeStamp, writeResources, totalResources);
+        }
+
+        public Task Prepared(string resourceId, Guid transactionId, DateTime timestamp, ParticipantId resource, TransactionalStatus status)
+        {
+            if (!this.managers.TryGetValue(resourceId, out ITransactionManager manager))
```
ReubenBond: It looks like we could extract these 3 lines out into a 'GetManager' method.

src/Orleans.Transactions/Abstractions/ITransactionManager.cs
jason-bragg: Intentionally not changing these call patterns until necessary.  Still more work to do for TOC which will likely necessitate interface changes, so little value in changing this at this time.

Issue #1909, opened 2016-07-07
ImplicitStreamSubscription Attribute inherit = false


dsarfati:
I was trying to implement a base grain class that has an implicit subscription, but it looks like Orleans doesn't support the `ImplicitStreamSubscription` on any base classes. Looking at the code it looks like this was intentionally excluded in `ImplicitStreamSubscriberTable.cs`
 `object[] attribs = grainClass.GetCustomAttributes(typeof(ImplicitStreamSubscriptionAttribute), inherit: false);`

Is there a reason for excluding the inheritance?

sergeybykov:
IIRC the thinking was that it is unlikely that one would want to implicitly subscribe multiple grain subclasses  to the same set of streams, and that inheriting the attribute would make easy to miss the fact of implicit subscription.

dsarfati:
I would normally agree with you that the normal use cases for implicit subscriptions probably don't require inheritance. We are building up a robust set of data processors that share some behavior (a single implicit subscription & several explicit subscriptions), so we were trying to build a base processor that would handle that shared logic for us. Any opposition to adding a flag to the `ImplicitStreamSubscription` that allows inheritance? I can open up the PR today if that seems like a reasonable solution.

sergeybykov:
>so we were trying to build a base processor that would handle that shared logic for us.

As an abstract base grain class?

dsarfati:
Correct. ` public abstract class WorkflowProcessGrain<T> : Grain<T>`

jason-bragg:
addressed in - Allow implicit subscription attribute to be inheritable #4824

krlm:
@sergeybykov  How this is supposed to work? Currently I'm getting following error when my subclass inherits from abstract base class marked with `ImplicitStreamSubscription` attribute. 
```
InvalidOperationException: A suitable constructor for type 'XXX.Infrastructure.EventSourcing.ProjectionBase`1[TEvent]' could not be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.
```

PR #4824, opened 2018-08-07
Allow implicit subscription attribute to be inheritable

jason-bragg:
ImplicitStreamSubscription Attribute inherit = false #1909


xiazen:
is there an issue asking for this ?

jason-bragg:
ImplicitStreamSubscription Attribute inherit = false #1909

Issue #4692, opened 2018-06-15
Docs: Possible mistake in "Unreliable Deployment on a Cluster of Dedicated Servers"


andersjonsson:
I tried to follow the guide for DevelopmentClustering
https://github.com/dotnet/orleans/blob/gh-pages/src/Documentation/Deployment-and-Operations/Configuration-Guide/Typical-Configurations.md/#unreliable-deployment-on-a-cluster-of-dedicated-servers

When I ran the code the client couldn't connect, so I tried changing the client port, in the .UseStaticClustering()-call to the gateway port (30000). That seemed to do the trick.

I guess that it's a typo in the Docs but I'm extremely new to Orleans, so I don't want to assume to much.

If it is a typo I can fix it and submit a PR

Also, the ClusterId and and ServiceId mismatch between the silo and the client config, but I caught that before I even tried to run the code.

JillHeaden:
HELP WANTED : Can someone please confirm that the suggested change is correct? If so, I will be glad to fix the document. (Also, can someone please add the "Help Wanted" label on this issue? Thanks!) 
@jason-bragg

ReubenBond:
This is correct - the port should be the gateway port, 30000.

JillHeaden:
Thank you, @ReubenBond !

JillHeaden:
To confirm, line 129:
`  .UseStaticClustering(gateways)`

... is being changed to...
`    .UseStaticClustering(gatewayPort: 30000)`

ReubenBond:
Apologies for not commenting earlier. The change should be that 11111 is changed to 30000 on lines 123, 124, & 126.

``` diff
var gateways = new IPEndPoint[]
{
-    new IPEndPoint(PRIMARY_SILO_IP_ADDRESS, 11111),
+    new IPEndPoint(PRIMARY_SILO_IP_ADDRESS, 30000),
-    new IPEndPoint(OTHER_SILO__IP_ADDRESS_1, 11111),
+    new IPEndPoint(OTHER_SILO__IP_ADDRESS_1, 30000),
    [...]
-    new IPEndPoint(OTHER_SILO__IP_ADDRESS_N, 11111),
+    new IPEndPoint(OTHER_SILO__IP_ADDRESS_N, 30000),
};
```

JillHeaden:
No problem. I will undo the other change (as noted in a comment on the PR) and make this change instead. 

Thanks!

ReubenBond:
Fixed by #4948

PR #4829, opened 2018-08-08
4692 gateway

JillHeaden:
Fixed issue #4692


src/Documentation/Deployment-and-Operations/Configuration-Guide/Typical-Configurations.md
```diff
@@ -126,7 +126,7 @@ var gateways = new IPEndPoint[]
     new IPEndPoint(OTHER_SILO__IP_ADDRESS_N, 11111),
 };
 var client = new ClientBuilder()
-    .UseStaticClustering(gateways)
+    .UseStaticClustering(gatewayPort: 30000)
```
xiazen: I don't think this compiles. UseStaticClusterting take a `IPEndPoint[] endpoints` as param, not just port. I think what @ReubenBond means in the issue is the port in `gateways` are supposed to be 30000, instead of 11111

PR #4831, opened 2018-08-09
Added support for priority transaction manager. 

jason-bragg:
Only one priority transaction manager is allowed per transaction.  Priority TM be used to detect TOC in future changes.

Also moved ReadonlyCommit to resource interface from tm interface.  Call was placed in wrong interface when split.

### DIFF
--- a/src/Orleans.Transactions/DistributedTM/ParticipantId.cs
+++ b/src/Orleans.Transactions/DistributedTM/ParticipantId.cs
@@ -1,0 +1,96 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Orleans.Concurrency;
+using Orleans.Runtime;
+using Orleans.Transactions.Abstractions;
+
+namespace Orleans.Transactions
+{
+    [Serializable]
+    [Immutable]
+    public readonly struct ParticipantId
+    {
+        public static readonly IEqualityComparer<ParticipantId> Comparer = new IdComparer();
+
+        [Flags]
+        public enum Role
+        {
+            Resource = 1 << 0,
+            Manager = 1 << 1,
+            PriorityManager = 1 << 2
+        }
+
+        public string Name { get; }
+        public GrainReference Reference { get; }
+        public Role SupportedRoles { get; }
+
+        public ParticipantId(string name, GrainReference reference, Role supportedRoles)
+        {
+            this.Name = name;
+            this.Reference = reference;
+            this.SupportedRoles = supportedRoles;
+        }
+
+        public override string ToString()
+        {
+            return $"ParticipantId.{Name}.{Reference}";
+        }
+
+        private class IdComparer : IEqualityComparer<ParticipantId>
+        {
+            public bool Equals(ParticipantId x, ParticipantId y)
+            {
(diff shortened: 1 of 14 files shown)


src/Orleans.Transactions/DistributedTM/TransactionAgent.cs
```diff
@@ -221,5 +213,25 @@ public void Abort(ITransactionInfo info, OrleansTransactionAbortedException reas
                  .Ignore();
             }
         }
+
+        // TODO: make overridable - jbragg
+        private KeyValuePair<ParticipantId,AccessCounter> SelectManager(TransactionInfo transactionInfo)
+        {
+            List<KeyValuePair<ParticipantId, AccessCounter>> priorityManagers = transactionInfo.Participants.SelectPriorityManagers().ToList();
```
xiazen: if there can only have 1 priority manager, then why `SelectPriorityManagers` returns multiple?

src/Orleans.Transactions/DistributedTM/TransactionInfo.cs
```diff
@@ -39,8 +39,6 @@ public TransactionInfo(TransactionInfo other)
             this.IsReadOnly = other.IsReadOnly;
             this.TimeStamp = other.TimeStamp;
             this.Priority = other.Priority;
-            this.TMCandidate = other.TMCandidate;
-            this.TMBatchSize = other.TMBatchSize;
```
xiazen: I understand TMCandidate has been moved to TransactionInfo? But where are we setting TMBatchSize now? 
I guess my question is , why is this removed?

jason-bragg: Batch size was an optimization for TM selection.  It was not strictly necessary, but may influence performance.  I suspect it's a premature optimization, but we'll see.  If we get perf drop from this, we can revisit.

src/Orleans.Transactions/DistributedTM/ParticipantId.cs
```diff
@@ -0,0 +1,96 @@
...
+        public enum Role
+        {
+            Resource = 1 << 0,
+            Manager = 1 << 1,
+            PriorityManager = 1 << 2
+        }
+
+        public string Name { get; }
+        public GrainReference Reference { get; }
+        public Role SupportedRoles { get; }
```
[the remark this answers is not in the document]
sebastianburckhardt: It may be nice for application to be able to specify priority TM even if not using TOC. I certainly have encountered this need in my applications (e.g. for my TPC-C workload I wanted to control which grain is the TM).

src/Orleans.Transactions/DistributedTM/TransactionInfo.cs
sebastianburckhardt: yep, we don't have super strong evidence that this optimization is important in general. I have seen it make a big difference in some cases but in the long run we may be better off just giving the application some control over TM selection than trying to do it automatically.

src/Orleans.Transactions/DistributedTM/ParticipantId.cs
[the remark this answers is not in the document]
sebastianburckhardt: I would keep it simple and just add a configuration option "PriorityTM" to the configuration object for the TransactionalState facet.

PR #4849, opened 2018-08-14
Changed link to documentation to use https instead of http

Expecho:
When I access the documentation using http://dotnet.github.io/orleans/Documentation/Introduction.html I am missing the navigation bar. The https version works fine.


sergeybykov:
Thank you, @Expecho!

PR #4851, opened 2018-08-15
transaction consistency tests and error injection

sebastianburckhardt:
the consistency tests I have been using for testing the transaction implementation.

- includes a test runner that runs randomized transaction tests using various scale and randomization settings.  Running all of those takes a long time (because of the large number of tests, and the volume of the tests) but this is sort of intentional for this type of test (random testing is quite good at discovering subtle race conditions if run with enough volume).

- each consistency test records a lot of data while executing the random transactions (state version numbers read & written by each transaction for each grain)  that is checked for serializability in the end. Also, I check if any exceptions encountered are expected for the type of test being run. For example, there should not be any transient exceptions under low congestion and when avoiding deadlocks, and there should not be any unknown exceptions when storage injection is off.

- These tests can run with or without fault injection. 

- The fault injector randomly injects three kinds of exceptions:
    - an uninterpreted exception before or after writing to storage
    - an InconsistentStateException before writing to storage

### DIFF
--- a/src/Orleans.Transactions/Abstractions/ITransactionalFaultInjector.cs
+++ b/src/Orleans.Transactions/Abstractions/ITransactionalFaultInjector.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Orleans.Transactions
+{
+    public interface ITransactionalFaultInjector
+    {
+        void BeforeStore();
+        void AfterStore();
+
+    }
+
+
+}
--- a/test/Transactions/Orleans.Transactions.Azure.Test/FaultInjectionConsistencyTests.cs
+++ b/test/Transactions/Orleans.Transactions.Azure.Test/FaultInjectionConsistencyTests.cs
@@ -1,0 +1,134 @@
+﻿using Orleans.TestingHost;
+using Orleans.Transactions.Tests;
+using Orleans.Transactions.Tests.Consistency;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Orleans.Transactions.AzureStorage.Tests
+{
+    [TestCategory("BVT"), TestCategory("Transactions")]
+    public class FaultInjectionConsistencyTests: TransactionTestRunnerBase, IClassFixture<FaultInjectedTestFixture>
+    {
+        public FaultInjectionConsistencyTests(FaultInjectedTestFixture fixture, ITestOutputHelper output)
+            : base(fixture.GrainFactory, output)
+        { }
+
+        [SkippableTheory]
+        // high congestion
+        [InlineData(2, 2, true, true, ReadWriteDetermination.PerGrain)]
+        [InlineData(2, 3, true, true, ReadWriteDetermination.PerGrain)]
+        [InlineData(2, 4, true, true, ReadWriteDetermination.PerGrain)]
+        [InlineData(2, 5, true, true, ReadWriteDetermination.PerGrain)]
+        [InlineData(2, 2, true, true, ReadWriteDetermination.PerTransaction)]
+        [InlineData(2, 3, true, true, ReadWriteDetermination.PerTransaction)]
+        [InlineData(2, 4, true, true, ReadWriteDetermination.PerTransaction)]
+        [InlineData(2, 5, true, true, ReadWriteDetermination.PerTransaction)]
+        [InlineData(2, 2, true, true, ReadWriteDetermination.PerAccess)]
+        [InlineData(2, 3, true, true, ReadWriteDetermination.PerAccess)]
+        [InlineData(2, 4, true, true, ReadWriteDetermination.PerAccess)]
+        [InlineData(2, 5, true, true, ReadWriteDetermination.PerAccess)]
+        [InlineData(2, 2, false, true, ReadWriteDetermination.PerAccess)]
+        [InlineData(2, 3, false, true, ReadWriteDetermination.PerAccess)]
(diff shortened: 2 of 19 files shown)


src/Orleans.Transactions/Abstractions/ITransactionalFaultInjector.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Orleans.Transactions
+{
+    public interface ITransactionalFaultInjector
```
xiazen: it is a bit odd to include the testing interface in TransactionAbstraction package. I run into the same problem in my PR #4874 . I'm thinking moving this interface to Transaction.Tests package, and inject it into fault injection `ITransactionalStateStorage`, by decorating existing `ITransactionalStateStorage`

sebastianburckhardt: I was just copying the way the skewed clock injection works - if that pattern is not o.k. then the skewed clock should be moved as well.

xiazen: hmm @jason-bragg  here since he added the IClock interface. I think IClock interface is not just testing specific , it provides ability to inject other type of clock ? @jason-bragg please clarity your intention

sebastianburckhardt:
I am removing the work-in-progress flag since the tests are now all passing (after fixes in #4887).

test/Transactions/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
```diff
@@ -161,5 +170,119 @@ public virtual async Task MultiWriteToSingleGrainTransaction(string grainStates)
...
+        [InlineData(1000, 2, false, false, ReadWriteDetermination.PerTransaction)]
+        [InlineData(1000, 3, false, false, ReadWriteDetermination.PerTransaction)]
+        [InlineData(1000, 4, false, false, ReadWriteDetermination.PerTransaction)]
+        [InlineData(1000, 5, false, false, ReadWriteDetermination.PerTransaction)]
+        [InlineData(1000, 2, false, false, ReadWriteDetermination.PerAccess)]
+        [InlineData(1000, 3, false, false, ReadWriteDetermination.PerAccess)]
+        [InlineData(1000, 4, false, false, ReadWriteDetermination.PerAccess)]
+        [InlineData(1000, 5, false, false, ReadWriteDetermination.PerAccess)]
+
+        public virtual async Task RandomizedConsistency(int numGrains, int scale, bool avoidDeadlocks, bool avoidTimeouts, ReadWriteDetermination readwrite)
```
jason-bragg: Not sure if this should be in the golden path test runner.  If it's expected to succeed consistently, then this is fine.  Just questioning to make sure it's in the right place.

test/Transactions/Orleans.Transactions.Tests/Memory/FaultInjectionConsistencyTests.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Orleans.Transactions.Tests.Memory
+{
+    [TestCategory("BVT"), TestCategory("Transactions")]
```
jason-bragg: I was under the impression that these tests were expected to be used as developer tests initially, with nightly runs to be added later.  Marking them as BVT means they'll be run with every code submission.  If these tests are not expected to succeed consistently, we shouldn't mark them BVT.

test/Transactions/Orleans.Transactions.Azure.Test/FaultInjectionConsistencyTests.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Orleans.Transactions.AzureStorage.Tests
+{
+    [TestCategory("BVT"), TestCategory("Transactions")]
```
jason-bragg: Azure tests should be marked azure, and not be marked bvt unless we're confident that they will pass consistently.

src/Orleans.Transactions/Abstractions/ITransactionalFaultInjector.cs
jason-bragg: This should probably be implemented as an ITransactionalStateStorage decorator, but this is ok for now, we'll need to restructure in future pass.

test/Transactions/Orleans.Transactions.Tests/Memory/FaultInjectionConsistencyTests.cs
sebastianburckhardt: yep, that was not the intention. I will mark a small subset as BVT.

test/Transactions/Orleans.Transactions.Azure.Test/FaultInjectionConsistencyTests.cs
sebastianburckhardt: ok

test/Transactions/Orleans.Transactions.Tests/Runners/GoldenPathTransactionTestRunner.cs
jason-bragg: Understood, but the test classes are mostly simple cruft code.  We can start with 2, just memory and azure, and add clocked versions later If you'd prefer.

sebastianburckhardt:
I will clean this up a bit to improve the categorization and remove some code duplication. 
Looks like we can use a single runner for all tests, and six specialized tests classes.

Issue #4852, opened 2018-08-16
Revisit handling of StatelessWorker grains during shutdown


sergeybykov:
Base on the feedback in #4757, we need to reconsider how StatelessWorker grains are handled during silo shutdown.

benjaminpetit:
Fixed with #4853

Issue #4757, opened 2018-07-11
Incorrect silo shutdown behavior


yevhen:
We're experiencing serious problems with a rolling update after migration to 2.0. Let me explain our architecture: we have a single heterogenous Azure queue processed by multiple workers (Competing Consumers pattern). Each message in this queue may be destined to a differrent actor. The setup is quite similar to how built-in Azure Queue Streaming Provider works. See diagram below.

![2018-07-11 12_06_43-untitled diagram xml - draw io](https://user-images.githubusercontent.com/107726/42562033-5785be38-8503-11e8-9e9e-055862a2d9c4.png)

Each node runs pulling agent (queue poller)  and the corresponding TPL DataFlow pipeline outside of the silo but within the same process with the silo (ie embedded). It could happen that recipent actor may reside on a different silo and so the client may forward the message to another silo, which is normal of course. Also there number of other actors that may exchange messages across silos, also normal, of course, since that's the whole point of having Orleans.

The problem arises when we try to shutdown the silo during rolling update. My expectation of Orleans' silo shutdown behavior when `StopAsync()` is called:
1. Don't accept any *new* requests into the silo, neither from clients nor from other silos (basically, close silo gateway)
2. Wait until all outstanding requests are completed (or failed or timed out) and return results
3. Deactivate all actors
4. Done

Instead, what we see from logs is that upon receiving `StopAsync()` Orleans almost immedietely deactivates all actors and then for 1 minute it bounces requests from other clients/silos by trying to forward them to non-existent activations (and they are not reactivated on the current silo of course since it's shutting down), and after 1 minute it kills itself after silo stop timeout is expired (which is 1 minute as I can see from the code). During this time it (Orleans.Runtime.Dispatcher) spills tens of thousands of log messages like:

```
Trying to forward after Non-existent activation, ForwardCount = 0. Message Request S10.31.5.14:100:268955504*cli/586efeb6@0c49119e->S10.31.5.14:100:268955504*grn/2FB327A8/00000000+124531@dd334eba #17019
```

And

```
Forwarding failed: tried to forward message Request S10.31.5.14:100:268955504*cli/586efeb6@0c49119e->S10.31.5.14:100:268955504*grn/2FB327A8/00000000+136155@897e8889 #17002[ForwardCount=1]:  for 1 times after Non-existent activation to invalid activation. Rejecting now.
```

And
 
```
[...]

PR #4853, opened 2018-08-16
Do not place stateless worker locally if the silo is stopping

benjaminpetit:
Fix for #4852 and #4757


ReubenBond:
@dotnet-bot test functional

ReubenBond:
Should this be applied to all placement providers?

PR #4857, opened 2018-08-17
TransactionRecovery test fixes

ReubenBond:
Improve diagnostics & fix JSON serialization issue with `BitArrayState`


ReubenBond:
(note that this does not fix the tests entirely)

PR #4864, opened 2018-08-21
Doc reorg

JillHeaden:


Documentation/grains/transactions.html
```diff
@@ -5,9 +5,9 @@
   <head>
     <meta charset="utf-8">
```
xiazen: I can understand that for 1.5, we only kept the html files, so we are making updates to html files. But isn't for 2.0 docs, we should be reorg .md file, instead of html files? why is this reshuffle of html files ?

xiazen: sorry I missed the md file changes. ;P , which caused the shuffle of those html files. we are safe here

PR #4874, opened 2018-08-23
Add storage exception injection logic into fault injection test runner

xiazen:
Adding fault injection test case
- inject storage exception before store, during Prepare
- inject storage exception after store, during Prepare 
- inject storage exception before store, during Confirm
- inject storage exception after store, during Confirm
- inject storage exception before store, during PrepareAndCommit
- inject storage exception after store, during PrepareAndCommit

injecting exception before store simulates situation that state update failed to persist
injecting exception after store simulates situation that state update succeed to persist, but runtime think it failed

currently all tests failed... I don't think I have a test logic issue. But please give me a second eye.


src/Orleans.Transactions/Abstractions/ITransactionFaultInjector.cs
```diff
@@ -0,0 +1,12 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Orleans.Transactions.Abstractions
+{
+    public interface ITransactionFaultInjector
+    {
+        void BeforeStore();
+        void AfterStore();
+    }
```
ReubenBond: Seems odd that a fault injector would be added into the public abstractions package... is there a better way?

Eg, could we generalize this to `ITransactionStorageObserver` or a name with a similarly general meaning?

ReubenBond: Even better, could this be implemented using something like a decorator pattern instead? It seems that we could decorate `ITransactionalStateStorage<TState>` as-needed and avoid special-casing the code. As a bonus, doing so would remove the class of errors whereby we forget to call the Before/After methods around the Store call.

xiazen: the storage exception injection logic is too deep into the implementation, if using decorator pattern, I might need to decorate every class.. I will see what I can do there. may decorate ITransactionalStateStorage is a good alternative. But then we need to decorate different storage provider implementation if we have multiple

xiazen:
@sebastianburckhardt  this is the failed storage  exception injection tests I mentioned in the meeting. You can see more on what I described in the meeting here

test/Transactions/Orleans.Transactions.Azure.Test/FaultInjection/FaultInjectionAzureTableTransactionStateStorage.cs
```diff
@@ -0,0 +1,85 @@
...
+            string metadata,
+
+            // a list of transactions to prepare.
+            List<PendingTransactionState<TState>> statesToPrepare,
+
+            // if non-null, commit all pending transaction up to and including this sequence number.
+            long? commitUpTo,
+
+            // if non-null, abort all pending transactions with sequence numbers strictly larger than this one.
+            long? abortAfter
```
ReubenBond: Usually doc comments like these would go above the method, like:

```xml
/// <summary>
/// Register a timer to send regular callbacks to this grain.
/// This timer will keep the current grain from being deactivated.
/// </summary>
/// <param name="asyncCallback"></param>
/// <param name="state"></param>
/// <param name="dueTime"></param>
/// <param name="period"></param>
/// <param name="name"></param>
/// <returns></returns>
```

You could include this info there.

test/Transactions/Orleans.Transactions.Tests/Runners/FaultInjectionTransactionTestRunner.cs
```diff
@@ -0,0 +1,90 @@
...
+    public class FaultInjectionTransactionTestRunner : TransactionTestRunnerBase
+    {
+        public FaultInjectionTransactionTestRunner(IGrainFactory grainFactory, ITestOutputHelper output)
+         : base(grainFactory, output)
+        { }
+
+        [SkippableFact]
+        public async Task SingleGrainReadTransaction()
+        {
+            const int expected = 5;
```
veikkoeeva: This is a nit, but I believe .NET naming conventions call for capitalizing constants.

test/Transactions/Orleans.Transactions.Azure.Test/FaultInjection/FaultInjectionAzureTableTransactionStateStorage.cs
```diff
@@ -0,0 +1,85 @@
...
+            var result = await this.stateStorage.Store(expectedETag, metadata, statesToPrepare, commitUpTo, abortAfter);
+            faultInjector.AfterStore();
+            return result;
+        }
+    }
+
+    public class FaultInjectionAzureTableTransactionStateStorageFactory : ITransactionalStateStorageFactory,
+        ILifecycleParticipant<ISiloLifecycle>
+    {
+        private readonly AzureTableTransactionalStateStorageFactory factory;
```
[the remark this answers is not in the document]
jason-bragg: We could probably make this more generic, but the fault tolerance of the system should, in theory, be storage independent, so we should really only need to run the storage fault injection tests against a single storage to identify the issues.  If we do have a more generic need in the future, we can generalize this at the time.

xiazen:

test/Transactions/Orleans.Transactions.Azure.Test/FaultInjection/FaultInjectionAzureTableTransactionStateStorage.cs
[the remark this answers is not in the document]
xiazen: It probably can be made more generic, but currently there's only one type of fault injection tx state used. So more generic version of this doesn't have much usage now. We can make it more generic when we have more than 1 testTxState used in tests.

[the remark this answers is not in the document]
ReubenBond: Sounds good to me

ReubenBond:
Needs rebase

xiazen:
rebased and squashed for easy rabse. waiting for tests.

PR #4875, opened 2018-08-23
Revisit silo stop/shutdown timeout

benjaminpetit:

Issue #4774, opened 2018-07-17
Race condition with interleaved messages using TCS


yevhen:
We're extensively using the async batcher pattern described by @hataytuna in this book
https://github.com/hataytuna/Distributed/raw/master/Applied%20Actor%20Model%20with%20Orleans%20v1.1.pdf It is crucial to performance boosts we got from batching in our system.

We figured out that problem we're experiencing with graceful shutdown described in #4757 is due to Orleans not being able to deactivate some of the grains. This only happens when TaskCompletionSource is used (along with the buffering and timer-based completion).

I have a failing test (#4773) for this scenario you can pull and check. On the surface, it looks like after DeactivateOnIdle is called Orleans disposes the timer while there interleaved requests, which leaves tasks in the uncompleted state, thus preventing Orleans from deactivating the grain.

Please, advice what can be done wrt to this issue.

yevhen:
Found bug in my test. Seem that it works. Now I'm completely confused why exactly the same code (and test) doesn't work properly in my app.

yevhen:
Well, we found that problem is with one of out actor middlewares. After we have removed the problematic actor filter it started to work. We'll be investigating the reasons. Sorry for false alarm.

hataytuna:
Glad to hear it.

Cheers,
Hatay

________________________________

yevhen:
Unfortunately, it's still an issue. I've added more repeats to the test so now it fails more reliably. Please confirm if the test is wrong or there is an issue. None of the things helped and we still have some actors stuck in production environment during rolling restart :(

yevhen:
Ok, guys, now I can definitely confirm that it's the bug in Orleans code. 

I've changed [this](https://github.com/dotnet/orleans/blob/master/src/Orleans.Runtime/Catalog/ActivationData.cs#L394) method to:

```cs
        public void PrepareForDeactivation()
        {
            SetState(ActivationState.Deactivating);
            deactivationStartTime = DateTime.UtcNow;
            if (!IsCurrentlyExecuting)
                StopAllTimers();
        }
```
(did custom private build) and now my system shuts down in production smoothly without any issues.

I don't know how to write the test which will fail reliably (to show the presence of the bug) but it can be easily figured out analytically. Take look at this method:

https://github.com/dotnet/orleans/blob/839e95303f897c0bcc6b86222ff03d43b1420e21/src/Orleans.Runtime/Catalog/Catalog.cs#L867-L886

[...]

PR #4883, opened 2018-08-23
Add timeout mecanism for grain deactivation

benjaminpetit:
It's not the fix I intended to write, but I thinks used in conjunction with #4875, it should fix the issues encountered in #4774 and #4757

### DIFF
--- a/src/Orleans.Core/Async/MultiTaskCompletionSource.cs
+++ b/src/Orleans.Core/Async/MultiTaskCompletionSource.cs
@@ -26,11 +26,6 @@ namespace Orleans
             get { return tcs.Task; }
         }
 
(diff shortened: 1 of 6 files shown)


src/Orleans.Core/Async/MultiTaskCompletionSource.cs
```diff
@@ -25,6 +26,10 @@ public Task Task
             get { return tcs.Task; }
         }
 
+        public bool TrySetCancelled(CancellationToken ct)
```
ReubenBond: Spelling of Cancelled is inconsistent with TPL

ReubenBond: (and afaict, this method is never called)

PR #4888, opened 2018-08-24
Fixes 'clusters and clients' similar to 'grains'

JillHeaden:


sergeybykov:
Looks like this one was superseded by #4902?

PR #4890, opened 2018-08-25
Feature per grain collection attribute

aRajeshKumar:


src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Orleans.Configuration
```
ReubenBond: I think this can live in the root `Orleans` namespace

src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            get
+            {
+                return minutes;
+            }
+            set
+            {
+                if (value <= 0d)
+                {
+                    throw new ArgumentOutOfRangeException("Collection Age Limit must be a positive number.");
```
ReubenBond: We should remove this, or move it to the getter. The reason is that if the setter fails then the user will see a TypeLoadException at runtime and they are hard to debug.

An alternative approach is to create an implementation of `IConfigurationValidator` which receives `IOptions<GrainCollectionOptions>` in its constructor and then checks every one of the values in its `ValidateConfiguration` method. The benefit of doing it that way is that it will catch all cases where the user enters an invalid value (including when they configure `GrainCollectionOptions` in code) and the user will see the exception on startup and it's consistent with how the rest of the configuration is validated.

src/Orleans.Runtime/Catalog/Catalog.cs
```diff
@@ -472,14 +473,14 @@ public void GetGrainTypeInfo(int typeCode, out string grainClass, out PlacementS
                 {
                     TimeSpan ageLimit = this.collectionOptions.Value.ClassSpecificCollectionAge.TryGetValue(grainType, out TimeSpan limit)
                         ? limit
-                        : collectionOptions.Value.CollectionAge;
+                        : TimeSpan.FromMinutes(CollectionAgeLimitConstants.DefaultCollectionAgeLimitInMinutes);
```
ReubenBond: I believe this should not be changed.

sergeybykov:
Functional test failures appear related to the change:
>UnitTests.StuckGrainTests.StuckGrainTests.StuckGrainTest_Basic
UnitTests.ActivationsLifeCycleTests.ActivationCollectorTests.ActivationCollectorShouldCollectIdleActivations
UnitTests.ActivationsLifeCycleTests.ActivationCollectorTests.ManualCollectionShouldNotCollectBusyActivations
UnitTests.ActivationsLifeCycleTests.ActivationCollectorTests.ActivationCollectorShouldNotCollectBusyActivations

src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
aRajeshKumar: This approach looks very nice. Let me see if I can add this capability.

src/Orleans.Core/Configuration/CollectionAgeLimitConstants.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Orleans.Configuration
+{
+    internal class CollectionAgeLimitConstants
+    {
+        public static readonly double DefaultCollectionAgeLimitInMinutes = 120;
```
ReubenBond: I prefer a default of zero (i.e, remove this and make `Minutes` an auto-property) - we can let the config validator throw when no value is set.

ReubenBond:
@dotnet-bot test bvt

sergeybykov:
Thank you, @aRajeshKumar!

PR #4891, opened 2018-08-27
Move actor identity content from 1.5 tutorials to grains section

JillHeaden:
The 1.5 Tutorial page "Actor Identity" is informational, not procedural - it should be included among other informational pages.

### DIFF
--- a/src/Documentation/grains/actor_identity.md
+++ b/src/Documentation/grains/actor_identity.md
@@ -1,0 +1,126 @@
+---
+layout: page
+title: Actor Identity
+---
+
+# Actor Identity
+
+In object-oriented environments, the identity of an object is hard to distinguish from a reference to it.
+Thus, when an object is created using new, the reference you get back represents all aspects of its identity except those that map the object to some external entity that it represents.
+
+In distributed systems, object references cannot represent instance identity, since references are typically limited to a single address space.
+That is certainly the case for .NET references.
+Furthermore, a virtual actor must have an identity regardless of whether it is active, so that we can activate it on demand.
+Therefore grains have a primary key.
+The primary key can be either a GUID (A Globally Unique Identifier), a long integer, or a string.
+
+The primary key is scoped to the grain type.
+Therefore, the complete identity of a grain is formed from the actor type and its key.
+
+The caller of the grain decides a long, a GUID, or a string scheme should be used.
+In fact the underlying data is the same, so the schemes can be used interchangeably.
+When a long is used, a GUID is actually created, and padded with zeros.
+
+Situations that require a singleton grain instance, such as a dictionary or registry, benefit from using `0` (a valid GUID) as its key.
+This is merely a convention, but by adhering, it becomes clear at the call site that it is what is going on, as we saw in the first tutorial:
+
+## Using GUIDs
+
+GUIDs are useful when there are several processes that could request a grain, such as a number of web servers in a web farm.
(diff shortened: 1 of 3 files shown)


src/Documentation/grains/actor_identity.md
```diff
@@ -0,0 +1,123 @@
+---
+layout: page
+title: Actor Identity
```
ReubenBond: I think we should change the name to Grain identity and generally replace 'actor' with 'grain' in most/all docs (except in places where we're specifically talking about the Actor Model, if there are any)

ReubenBond: Maybe @sergeybykov has an opinion on this.

"virtual actor" would become "grain"

src/Documentation/grains/actor_identity.md
```diff
@@ -0,0 +1,123 @@
...
+The primary key can be either a GUID (A Globally Unique Identifier), a long integer, or a string.
+
+The primary key is scoped to the grain type.
+Therefore, the complete identity of a grain is formed from the actor type and its key.
+
+The caller of the grain decides a long, a GUID, or a string scheme should be used.
+In fact the underlying data is the same, so the schemes can be used interchangeably.
+When a long is used, a GUID is actually created, and padded with zeros.
+
+Situations that require a singleton grain instance, such as a dictionary or registry, benefit from using `0` (a valid GUID) as its key.
```
ReubenBond: This line seems dated - maybe it should say `Guid.Empty` instead of `0`.

src/Documentation/grains/actor_identity.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+## Using Longs
+
+A long integer is also available, which would make sense if the grain is persisted to a relational database, where numerical indexes are preferred over GUIDs.
+
+Referencing a grain by long integer in client code:
+
+``` csharp
+var grain = GrainClient.GrainFactory.GetGrain<IExample>(1);
```
ReubenBond: We should change `GrainClient.GrainFactory` to just `grainFactory` here and in other example code.

For history, there used to be a static class called `GrainClient` which users would use to interact with Orleans, but that static class was removed and replaced with the current approach (build the client using ClientBuilder, etc)

JillHeaden:
@ReubenBond @sergeybykov 
Should these changes to the document be in a separate issue ticket?

sergeybykov:
@JillHeaden I don't think we need a separate issue. Just need to clean up the language.

src/Documentation/grains/actor_identity.md
JillHeaden: The line now reads:

Situations that require a singleton grain instance, such as a dictionary or registry, benefit from using `Guid.Empty` as its key.

I took out the '(a valid GUID)' clause because I figured that it was explaining that '0' is a valid GUID - and now '0' is gone.

src/Documentation/grains/actor_identity.md
JillHeaden: I have renamed the file in one PR and this change will be made throughout this document. 
I will seek and change this in other files, too. 

Thanks, @ReubenBond !

src/Documentation/grains/actor_identity.md
JillHeaden: Thank you for the change suggestion and the history. It's neat to know how Orleans has evolved. I found and replaced 3 instances in this document and will be looking through the rest of the docs later.

JillHeaden:
The changes requested above are already in the document: http://dotnet.github.io/orleans/Documentation/grains/grain_identity.html

This PR can be closed.

PR #4895, opened 2018-08-27
Use netcoreapp2.0 for msbuild target dll if using dotnet core msbuild but targeting full .net

benjaminpetit:
(Cherry-pick of #4689)

#4673 introduced a regression: when using the command `dotnet build` to build project that target the full .NET framework, the variable for `CoreAssembly` is not set, and the variable `TaskAssembly` depends on it.

The easiest workaround here is to always fallback to 2.0 dll

Maybe we could instead removing this task dependency, and rely on the fact that we should find the `dotnet` command in the path @ReubenBond ?


sergeybykov:
@dotnet-bot test this please

ReubenBond:
@benjaminpetit I'm fine with your suggestion.

The code used to find `dotnet` can be resurrected when we move to a framework-agnostic code generator (i.e, one binary for all frameworks)

PR #4897, opened 2018-08-27
Bump codegen version to 2.0.5

benjaminpetit:


sergeybykov:
`Fix pkg ref in some tests csproj` commit is needed in `master` as well.

PR #4907, opened 2018-08-29
Propagate message serialization exceptions to callers

### LINKED ISSUE #4793: Serialization errors on exceptions thrown from grain methods cause client timeouts
Take grain interface:

```
interface IFoo
{
    Task DoSomething();
}
```

If, for whatever reason, DoSomething() encounters an unhandled exception that reaches the runtime, and the runtime cannot serialize it back to the client, the runtime sends nothing back to the client causing an eventual client timeout.

The expected behavior is the client would receive some generic serialization failed exception.  Without this these types of issues are very difficult to debug.

ReubenBond:
Fixes #4793 

There are 6 scenarios being tested, which can be summed up as `{Serialization, Deserialization} exception sending to a {client, grain} from a {client, grain}` except I didn't add tests for the client-to-client cases (they go via silo anyway).

In the end, there's similar logic for handling these exceptions in:
* `GatewayConnection.OnMessageSerializationFailure` - client to gateway (request)
* `InvokableObjectManager.LocalObjectMessagePumpAsync` - client to gateway (response)
* `Gateway.GatewaySender.OnMessageSerializationFailure` - gateway to client
* `SiloMessageSender.OnMessageSerializationFailure` - silo to silo

The client-to-gateway response scenarios are a little contrived, since client objects are not supposed to be able to return `Task` or `Task<T>`, only `void`. The code generator blocks `Task`/`Task<T>` for external users. Nevertheless, it's worth adding tests for the scenario just in case, and it helped to uncover some issues and places we could improve logging.

### DIFF
--- a/src/Orleans.Core/Messaging/GatewayConnection.cs
+++ b/src/Orleans.Core/Messaging/GatewayConnection.cs
@@ -250,28 +250,42 @@ namespace Orleans.Messaging
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
         protected override void OnMessageSerializationFailure(Message msg, Exception exc)
         {
-            // we only get here if we failed to serialise the msg (or any other catastrophic failure).
-            // Request msg fails to serialise on the sending silo, so we just enqueue a rejection msg.
-            Log.Warn(ErrorCode.ProxyClient_SerializationError, $"Unexpected error serializing message to gateway {Address}.", exc);
-            FailMessage(msg, $"Unexpected error serializing message to gateway {Address}. {exc}");
-            if (msg.Direction == Message.Directions.Request || msg.Direction == Message.Directions.OneWay)
-            {
-                return;
-            }
-
+            // we only get here if we failed to serialize the msg (or any other catastrophic failure).
+            // Request msg fails to serialize on the sender, so we just enqueue a rejection msg.
             // Response msg fails to serialize on the responding silo, so we try to send an error response back.
-            // if we failed sending an original response, turn the response body into an error and reply with it.
-            msg.Result = Message.ResponseTypes.Error;
-            msg.BodyObject = Response.ExceptionResponse(exc);
-            try
+            this.Log.LogWarning(
+                (int) ErrorCode.ProxyClient_SerializationError,
+                "Unexpected error serializing message {Message}: {Exception}",
+                msg,
+                exc);
+
+            msg.ReleaseBodyAndHeaderBuffers();
+            MessagingStatisticsGroup.OnFailedSentMessage(msg);
+
+            var retryCount = msg.RetryCount ?? 0;
+
+            if (msg.Direction == Message.Directions.Request)
             {
-                MsgCenter.SendMessage(msg);
+                this.MsgCenter.RejectMessage(msg, $"Unable to serialize message. Encountered exception: {exc?.GetType()}: {exc?.Message}", exc);
             }
-            catch (Exception ex)
+            else if (msg.Direction == Message.Directions.Response && retryCount < 1)
             {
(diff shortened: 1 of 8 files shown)


src/Orleans.Core/Messaging/GatewayConnection.cs
```diff
@@ -250,28 +250,48 @@ protected override void OnGetSendingSocketFailure(Message msg, string error)
...
         {
-            // we only get here if we failed to serialise the msg (or any other catastrophic failure).
-            // Request msg fails to serialise on the sending silo, so we just enqueue a rejection msg.
-            Log.Warn(ErrorCode.ProxyClient_SerializationError, $"Unexpected error serializing message to gateway {Address}.", exc);
-            FailMessage(msg, $"Unexpected error serializing message to gateway {Address}. {exc}");
-            if (msg.Direction == Message.Directions.Request || msg.Direction == Message.Directions.OneWay)
+            // we only get here if we failed to serialize the msg (or any other catastrophic failure).
+            // Request msg fails to serialize on the sender, so we just enqueue a rejection msg.
+            // Response msg fails to serialize on the responding silo, so we try to send an error response back.
+            if (this.Log.IsEnabled(LogLevel.Warning))
```
xiazen: nit: I don't think we need to check everywhere if logger has Warning turned on. Most people set logging equal or above info. this check turns true in most cases. 
we usually check it for Debug and Trace cases

src/Orleans.Core/Messaging/GatewayConnection.cs
```diff
@@ -250,28 +250,48 @@ protected override void OnGetSendingSocketFailure(Message msg, string error)
...
             {
-                // If we still can't serialize, drop the message on the floor
-                Log.Warn(ErrorCode.ProxyClient_DroppingMsg, "Unable to serialize message - DROPPING " + msg, ex);
-                msg.ReleaseBodyAndHeaderBuffers();
+                // if we failed sending an original response, turn the response body into an error and reply with it.
+                // unless we have already tried sending the response multiple times.
+                msg.Result = Message.ResponseTypes.Error;
+                msg.BodyObject = Response.ExceptionResponse(exc);
+                msg.RetryCount = retryCount + 1;
+                this.MsgCenter.SendMessage(msg);
```
xiazen: what if serialization on this msg failed as well? do we want to add a try catch here, if this exception failed to serialize, we flatten it to a string, or a minimal Exception to return to the caller, stating serialization issue happened?

src/Orleans.Core/Runtime/InvokableObjectManager.cs
```diff
@@ -170,9 +189,17 @@ private async Task LocalObjectMessagePumpAsync(LocalObjectData objectData)
                     else if (message.Direction != Message.Directions.OneWay)
                         this.SendResponseAsync(message, resultObject);
```
xiazen: why is this method name SendResponseAsyc, it is not async method though.

src/Orleans.Core/Messaging/GatewayConnection.cs
ReubenBond: SendMessage just enqueues the message, so no serialization happens at that stage. If this also fails to be serialized then something is very wrong.
In the case that we're failing while sending a request message, the rejection will be handled locally and therefore won't be serialized. For response messages, the entire message body is being replaced by a different exception, so if that doesn't work then we have big issues (catastrophe).

If the message still fails then it could be a RequestContext issue, but that means that we won't even be able to deserialize message headers (so we have no information to go by)

Longer term, I'd like to move RequestContext out of the headers for this very reason.

src/Orleans.Core/Runtime/InvokableObjectManager.cs
ReubenBond: Maybe because it's non-blocking, but I'm not sure. It could be renamed at some point.

Issue #4855, opened 2018-08-17
[Question] about GrainStorage and the internal VersionStoreGrain


alirezajm:
I'm writing a storage provider using efcore. 
The provider works against a configured dbcontext, which means any type being used as storage should have a corresponding DbSet<> on the context. So arbitrary types are not welcomed here, as the backing storage is not happy with them (SQL Server for example).

Now, the problem raises If the provider is used as the default grain storage. This causes the internal ```VersionStoreGrain``` to malfunction because of the not configured ```VersionStoreGrainState```. 
Since the ```IVersionStoreGrain``` is an internal interface I can't add another implementation for it (Haven't really thought about if it's a good idea or not).

So to summarize my options:
- Provide another implementation for IVersionStoreGrain, (which is not doable cause everything in there is internal)
- Handle ```VersionStoreGrainState``` differently. This probably suggest an internal dbcontext to handle Orleans related stuff. Which also causes the provider to break if I'm missing another internal type.
- Handle arbitrary types using a fallback provider or a key-value table.
- Don't allow the provider to be used as the default storage.

I know using entity framework inherently suggest that arbitrary types can't be used but for this to work it should be able to be used as the only grain storage provider in the silo.

I'll probably go with option 3 with my current knowledge of Orleans. Any advice is very appreciated. I'm gonna open source this as soon as I finish this and write more tests.

benjaminpetit:
The public interface to implement if you want to implement your own storage for versioning rules is `IVersionStore`.

I think this is the only built-in grain that requires storage in Orleans.

alirezajm:
Ok so there's little documentation about this.
I assume one wouldn't really want to persist versioning strategies as the ```IVersionStore``` is about code compatibility. Am I right?

alirezajm:
Another thing that is a little bit puzzling is I couldn't find any other storage providers (like Orleans.Persistence.AdoNet) handling this issue.

alirezajm:
Can you tell me if my understanding of this is correct?

The built-in ```VersionStore``` uses the internal ```VersionStoreGrain``` and it'll be activated only once in the cluster which makes sense as the whole cluster should follow the same strategy. But doesn't any silo joining the cluster can change the strategy? This causes any deployment to be able to cause undesired effects. Again I have no idea yet who calls for example ```VersionStore.SetSelectorStrategy```, which is declared on ```IVersionManager```.

[...]

PR #4911, opened 2018-08-29
Do not try to register GrainVersionStore if an implementation of IVersionStore is already registered

benjaminpetit:
Fix for #4855

### DIFF
--- a/src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
+++ b/src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
@@ -38,6 +38,7 @@ using Microsoft.Extensions.Options;
 
 using Orleans.Configuration.Validators;
 using Orleans.Runtime.Configuration;
+using System.Linq;
 
 namespace Orleans.Hosting
 {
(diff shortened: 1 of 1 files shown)


benjaminpetit:
@dotnet-bot test functional

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -171,8 +172,11 @@ internal static void AddDefaultServices(HostBuilderContext context, IServiceColl
             services.TryAddSingleton<VersionSelectorManager>();
             services.TryAddSingleton<CachedVersionSelectorManager>();
             // Version selector strategy
-            services.TryAddSingleton<GrainVersionStore>();
-            services.AddFromExisting<IVersionStore, GrainVersionStore>();
```
ReubenBond: This can be `TryAddFromExisting` instead

ReubenBond: (but it doesn't matter)

PR #4913, opened 2018-08-30
Add back .editorconfig

pentp:
This adds back the .editorconfig file from #3872 that was removed by #3963.
I fixed some bugs in the file (removed severity from formatting rules that don't support it) and changed all warnings to suggestions or even to "none" for some rules.
Changed some of the rules to better match the existing codebase, but it's not always consistent, so suggestions for style changes are welcome.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,202 @@
+# http://EditorConfig.org
+
+# This file is the top-most EditorConfig file
+root = true
+
+# All Files
+[*]
+charset                  = utf-8
+end_of_line              = crlf
+indent_style             = space
+indent_size              = 4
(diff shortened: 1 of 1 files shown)


ReubenBond:
Thanks for bringing this up again, @pentp. I forked your branch and made some local modifications and tested whether or not they would have a negative impact on the developer experience (poor dev experience was the reason for reverting this). I believe that the dev experience is better now, so I'll push to your branch and we can merge if others are happy.

ReubenBond:
@xiazen if you have time, could you please pull this branch and see if the build experience on your dev machine is good in VS/console? It works well for me, but your setup may be different.

If it's good then we can merge it

Issue #4835, opened 2018-08-09
Change Orleans.TelemetryConsumers.Counters project target platform to netstandard or netcoreapp


MelnikovIG:
Currently `Orleans.TelemetryConsumers.Counters` targets `net461` framework. I am unable to use it with .net core app. I have an exception building `SiloHost` with `AddPerfCountersTelemetryConsumer` extension method
`System.IO.FileNotFoundException: 'Could not load file or assembly 'System.Diagnostics.PerformanceCounter, Version=0.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.' `

Looks like it uses full framework version of `System.Diagnostics.PerformanceCounter`. I think we can use netstandart version of `System.Diagnostics.PerformanceCounter` since  it is already available for netstandart https://www.nuget.org/packages/System.Diagnostics.PerformanceCounter/

ReubenBond:
Fixed by #4914

PR #4914, opened 2018-08-30
Migrate Orleans.TelemetryConsumers.Counters to netstandard

benjaminpetit:
See #4835


src/TelemetryConsumers/Orleans.TelemetryConsumers.Counters/OrleansPerformanceCounterInstaller.cs
```diff
@@ -1,94 +0,0 @@
...
-using Orleans.Metadata;
-using Orleans.Runtime.Configuration;
-
-namespace OrleansTelemetryConsumers.Counters
-{
-    /// <summary>
-    /// Providers installer hooks for registering Orleans custom performance counters.
-    /// </summary>
-    [RunInstaller(true)]
-    public class OrleansPerformanceCounterInstaller : Installer
```
xiazen: why is this removed? This installer will be invoked if the assembly is installed in Visual Studio Custom Action Installer or the Installutil.exe, to my understanding.

benjaminpetit: Because there is no official nuget for `System.Configuration.Install`, as far as I know.

User can still use `OrleansCounterControl` to install them using a .exe

PR #4916, opened 2018-08-31
[1.5.5] Propagate message serialization errors to callers

ReubenBond:
This is a port of #4907 to the 1.5.x branch

### DIFF
--- a/src/Orleans/Messaging/GatewayConnection.cs
+++ b/src/Orleans/Messaging/GatewayConnection.cs
@@ -246,28 +246,48 @@ namespace Orleans.Messaging
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
         protected override void OnMessageSerializationFailure(Message msg, Exception exc)
         {
-            // we only get here if we failed to serialise the msg (or any other catastrophic failure).
-            // Request msg fails to serialise on the sending silo, so we just enqueue a rejection msg.
-            Log.Warn(ErrorCode.ProxyClient_SerializationError, String.Format("Unexpected error serializing message to gateway {0}.", Address), exc);
-            FailMessage(msg, String.Format("Unexpected error serializing message to gateway {0}. {1}", Address, exc));
-            if (msg.Direction == Message.Directions.Request || msg.Direction == Message.Directions.OneWay)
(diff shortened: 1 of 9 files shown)


src/Orleans/Messaging/GatewayConnection.cs
```diff
@@ -246,28 +246,48 @@ protected override void OnGetSendingSocketFailure(Message msg, string error)
...
+                this.MsgCenter.RejectMessage(msg, $"Unable to serialize message. Encountered exception: {exc?.GetType()}: {exc?.Message}", exc);
+            }
+            else if (msg.Direction == Message.Directions.Response && retryCount < 1)
+            {
+                // if we failed sending an original response, turn the response body into an error and reply with it.
+                // unless we have already tried sending the response multiple times.
+                msg.Result = Message.ResponseTypes.Error;
+                msg.BodyObject = Response.ExceptionResponse(exc);
+                msg.RetryCount = retryCount + 1;
+                this.MsgCenter.SendMessage(msg);
```
xiazen: what if serialization on this msg failed as well? do we want to add a try catch here, if this exception failed to serialize, we flatten it to a string, or a minimal Exception to return to the caller, stating serialization issue happened?

src/OrleansRuntime/Messaging/Gateway.cs
```diff
@@ -538,10 +538,47 @@ private void OnMessageSerializationFailure(Message msg, Exception exc)
                 // we only get here if we failed to serialize the msg (or any other catastrophic failure).
                 // Request msg fails to serialize on the sending silo, so we just enqueue a rejection msg.
                 // Response msg fails to serialize on the responding silo, so we try to send an error response back.
-                Log.Warn(ErrorCode.Messaging_Gateway_SerializationError, String.Format("Unexpected error serializing message {0} on the gateway", msg.ToString()), exc);
+                if (this.Log.IsWarning)
```
xiazen: nit: I don't think we need to check everywhere if logger has **Warning** turned on. Most people set logging equal or above info. this check turns true in most cases.

xiazen: I think we usually check it for Debug or Trace level logs

sergeybykov:
Is this still an up to date port of #4907?

ReubenBond:
Is it, @sergeybykov

PR #4918, opened 2018-08-31
Rename actor identity to grain identity

JillHeaden:
File rename only. Next step is editing content.


src/Documentation/grains/grain_identity.md
```diff
@@ -0,0 +1,123 @@
+---
+layout: page
+title: Actor Identity
```
sergeybykov: It still say "actor identity" in multiple places.

PR #4921, opened 2018-08-31
Change typical configurationsmd gateway30000 issue4692

JillHeaden:
fixes issue #4692 

I submitted a PR about this previously, but I may not have done it correctly.


sergeybykov:
This is mixed with the actor identity change. So it cannot be merged separately. It is better to prepare these independently.

JillHeaden:
These issues have been re-done separately and I believe this PR can be closed.

PR #4922, opened 2018-08-31
Benefitsmd change actor to grain

JillHeaden:
Changes most instances of "actor" to "grain" 
Makes formatting adhere to documentation style guidelines
Makes minor grammatical changes


sergeybykov:
Needs to be rebased on top of the current docs branch.

JillHeaden:
The content of this file has already been updated via a different PR, but the file name needs to be changed to have a lower-case 'b' 

Current name: dotnet.github.io/orleans/Documentation/Benefits.html

sergeybykov:
`docs` branch is the source of truth, and it already has the file named as lower-case - https://github.com/dotnet/orleans/blob/docs/src/Documentation/benefits.md.

`gh-pages` branch gets rebuilt periodically, and at that point it picks up the changes and makes them visible at dotnet.github.io. You can always build the `docs` branch locally to look at the latest state.

sergeybykov:
I see that it still says "Actor Identity" instead of "Grain Identity."

PR #4923, opened 2018-08-31
Transction recovery bug fixes.

jason-bragg:


xiazen:

jason-bragg:
@dotnet-bot test this please

jason-bragg:
No really.. 
@dotnet-bot test this please

jason-bragg:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot test this please

src/Orleans.TestingHost/Utils/TestingUtils.cs
```diff
@@ -74,19 +74,21 @@ public static ILoggerFactory CreateDefaultLoggerFactory(string filePath, LoggerF
...
                 async () =>
                 {
                     bool passed;
                     do
                     {
-                        // need to wait a bit to before re-checking the condition.
-                        await Task.Delay(TimeSpan.FromSeconds(1));
                         passed = await predicate(false);
+                        // need to wait a bit to before re-checking the condition.
+                        if (!passed) await Task.Delay(delayOnFail.Value);
```
ReubenBond: I feel that we add `keepGoing[0]` to the guard here.

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -37,6 +37,10 @@ internal class TransactionQueue<TState>
         private Dictionary<DateTime, PMessages> unprocessedPreparedMessages;
         private class PMessages
         {
+            public PMessages(TransactionalStatus status)
```
ReubenBond: We should rename this to PrepareMessages or something more descriptive.

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -286,20 +286,22 @@ public void NotifyOfPing(Guid transactionId, DateTime timeStamp, ParticipantId r
...
-                            .Cancel(resource.Name, transactionId, timeStamp, TransactionalStatus.PresumedAbort);
+                            .Cancel(resource.Name, transactionId, timeStamp, TransactionalStatus.PresumedAbort).Ignore();
                 }
             }
         }
 
         public async Task NotifyOfConfirm(Guid transactionId, DateTime timeStamp)
         {
+            if (logger.IsEnabled(LogLevel.Trace))
+                logger.Trace($"NotifyOfConfirm: {transactionId} {timeStamp}");
```
ReubenBond: We should probably start using the "semantic logging" approach for new messages. That means removing the `$` and adding `transactionId` etc as parameters.

Not blocking.

test/Transactions/Orleans.Transactions.Azure.Test/TransactionRecoveryTests.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Orleans.Transactions.AzureStorage.Tests
 {
-    [TestCategory("Azure"), TestCategory("Transactions"), TestCategory("Functional")]
+    [TestCategory("Azure"), TestCategory("Transactions")]
```
ReubenBond: Without the Functional category, these tests will never run

test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalBitArrayGrain.cs
```diff
@@ -153,8 +153,8 @@ public static BitArrayState Apply(BitArrayState left, BitArrayState right, Func<
             // Continue with the non-overlapping portion.
             for (; i < result.value.Length; i++)
             {
-                var leftVal = left.value.Length < i ? left.value[i] : 0;
-                var rightVal = right.value.Length < i ? right.value[i] : 0;
+                var leftVal = left.value.Length > i ? left.value[i] : 0;
+                var rightVal = right.value.Length > i ? right.value[i] : 0;
                 result.value[i] = op(leftVal, rightVal);
```
ReubenBond: 😅 oops, sorry

ReubenBond:
Only the comment about `keepGoing[0]` needs to be addressed

src/Orleans.TestingHost/Utils/TestingUtils.cs
jason-bragg: Definitely debatable.  I'd considered including it.

I didn't because the behavior in that case would be for the final predicate(true) call to be made immediately after the last failed call without any delay, which was the original behavior, but probably not what we want.

What is the behavior we want?  I think we want the call to be made repeatedly until it succeeds, with a delay before each retry.  If the timeout fires, I'd think we'd still want to have a delay before the retry.

Do you disagree?

test/Transactions/Orleans.Transactions.Azure.Test/TransactionRecoveryTests.cs
jason-bragg: This is buy design, at this point.  They are mostly stable now, but still have some odd failures i'm working out, and can run longer than we usually like.  Still working out kinks in the system :/

src/Orleans.Transactions/State/TransactionQueue.cs
jason-bragg: I'll do this in a separate targeted PR.  Don't want to clutter or delay this one, as it's blocking other changes.

jason-bragg:

PR #4926, opened 2018-09-03
Fix fault injection tests and enable them 

xiazen:
I didn't know confirm would be retried on storage error, so I didn't reset fault injection control after confirmed once, which cause some of the tests failing.


xiazen:
closed this and reopened #4927 , failed at tooling.

PR #4929, opened 2018-09-04
Fix README.md documentation link

jlorich:
The link to documentation in the readme went to a page which does not exists. It's been updated to link to the root path for the documentation site.


sergeybykov:
Thank you, @jlorich!

PR #4931, opened 2018-09-04
Rename file from 'Benefits.md' to 'benefits.md', change 'actor' to ...

JillHeaden:
…'grain' where needed, and change grammar and wording.

This combines and replaces two other PRs that had problems.


ReubenBond:
Renaming a file to change the casing is annoying on Windows because Windows is case-insensitive.

I followed the answer here https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git/40307511 and pushed an update to your branch. I opened `src\Documentation` using a command prompt
and ran `git mv -f Benefits.md benefits.md`.

PR #4933, opened 2018-09-04
Remove 'Types of Grains.md" placeholder page, insert content from...

JillHeaden:
…"Actor Identity.md", and replace "actor" with "grain" throughout document.
Also, reformat  to Documentation Guidelines and make grammatical changes.

### DIFF
--- a/src/Documentation/grains/grain_identity.md
+++ b/src/Documentation/grains/grain_identity.md
@@ -1,0 +1,123 @@
+---
+layout: page
+title: Grain Identity
+---
+
+# Grain Identity
+
+In object-oriented environments, the identity of an object is hard to distinguish from a reference to it.
+Thus, when an object is created using new, the reference you get back represents all aspects of its identity except those that map the object to some external entity that it represents.
+
+In distributed systems, object references cannot represent instance identity, since references are typically limited to a single address space.
+That is certainly the case for .NET references.
+Furthermore, a virtual actor must have an identity regardless of whether it is active, so that we can activate it on demand.
+Therefore grains have a primary key.
+The primary key can be a Globally Unique Identifier (GUID), a long integer, or a string.
+
+The primary key is scoped to the grain type.
+Therefore, the complete identity of a grain is formed from the actor type and its key.
+
+The caller of the grain decides a long, a GUID, or a string scheme should be used.
+In fact the underlying data is the same, so the schemes can be used interchangeably.
+When a long integer is used, a GUID is actually created and padded with zeros.
+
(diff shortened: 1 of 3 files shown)


src/Documentation/grains/grain_identity.md
```diff
@@ -0,0 +1,123 @@
...
+---
+
+# Grain Identity
+
+In object-oriented environments, the identity of an object is hard to distinguish from a reference to it.
+Thus, when an object is created using new, the reference you get back represents all aspects of its identity except those that map the object to some external entity that it represents.
+
+In distributed systems, object references cannot represent instance identity, since references are typically limited to a single address space.
+That is certainly the case for .NET references.
+Furthermore, a virtual actor must have an identity regardless of whether it is active, so that we can activate it on demand.
```
ReubenBond: IMO we should write "grain" instead of "virtual actor" here (and replace actor with grain on line 18 below, too). @sergeybykov, what do you think?

sergeybykov: I agree - we should be talking about grains, and not confuse people by using actors or virtual actors interchangeably.

src/Documentation/grains/grain_identity.md
```diff
@@ -0,0 +1,123 @@
...
+
+To access the compound key in the grain, we can call an overload on the `GetPrimaryKey` method:
+
+``` csharp
+public class ExampleGrain : Orleans.Grain, IExampleGrain
+{
+    public Task Hello()
+    {
+	string keyExtension;
+        long primaryKey = this.GetPrimaryKey(out keyExtension);
```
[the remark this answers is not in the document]
JillHeaden: Fixed!

src/Documentation/grains/grain_identity.md
JillHeaden: Fixed - thank you!

src/Documentation/grains/grain_identity.md
```diff
@@ -111,10 +118,10 @@ public class ExampleGrain : Orleans.Grain, IExampleGrain
 {
     public Task Hello()
     {
-	string keyExtension;
-        long primaryKey = this.GetPrimaryKey(out keyExtension);
+		string keyExtension;
+        long primaryKey = this.GetPrimaryKeyLong(out keyExtension);
```
ReubenBond: I believe the spacing here is still off - but it probably looks correct on your screen. The issue appears to be that you're indenting using tabs but our convention is to use spaces.

If you're using Visual Studio Code to edit the documents (recommended), then you can switch to using spaces for indentation:
image

Issue #4935, opened 2018-09-04
Race condition with SimpleQueueCache allows for broken cursor


berdon:
See https://gitter.im/dotnet/orleans?at=5b8ed56958a3797aa301e47e:

Essentially, a cursor being spun up with an empty sequence token was set to the first available token but not tracked. This allows for periodic purging to purge the item with the token causing the cursor to report a cache miss forever.

The fix is pretty simple and supplied by @jason-bragg. I'll throw a PR up shortly.

jason-bragg:
Fix merged.
Thanks Berdon

PR #4936, opened 2018-09-04
#4935 Fixing race condition with simple queue cache

berdon:
Related to #4935 
- Cursor was allowed to be in a state dependent on a token
  while allowing the containing bucket to be purged


jason-bragg:
Thanks Berdon!

PR #4945, opened 2018-09-06
IsOrleansShallowCopyable fixes

pentp:
- `Guid` is shallow copyable and primitive.
- `Nullable<T>` is shallow copyable if the wrapped type is shallow copyable.
- `Tuple<T,...>` is shallow copyable if all the element types are shallow copyable.
- Value types are shallow copyable if all public and private instance fields are shallow copyable.
Previously private fields were ignored and public static fields were checked instead and generic types were not allowed.
- Changed the static constructor to a field initializer for better performance (static constructors enforce precise initialization timing which can be expensive).

### DIFF
--- a/src/Orleans.Core/Serialization/TypeUtilities.cs
+++ b/src/Orleans.Core/Serialization/TypeUtilities.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Concurrent;
-using System.Linq;
 using System.Net;
 using System.Reflection;
 using System.Text;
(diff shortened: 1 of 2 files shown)


src/Orleans.Core/Serialization/TypeUtilities.cs
```diff
@@ -21,86 +20,76 @@ internal static bool IsOrleansPrimitive(this Type t)
...
+                if (def == typeof(Nullable<>)
+                    || def == typeof(Tuple<>)
+                    || def == typeof(Tuple<,>)
+                    || def == typeof(Tuple<,,>)
+                    || def == typeof(Tuple<,,,>)
+                    || def == typeof(Tuple<,,,,>)
+                    || def == typeof(Tuple<,,,,,>)
+                    || def == typeof(Tuple<,,,,,,>)
+                    || def == typeof(Tuple<,,,,,,,>))
+                    return Array.TrueForAll(t.GenericTypeArguments, a => IsOrleansShallowCopyable(a));
```
ReubenBond: Could also add ValueTuple here

ReubenBond:
Thank you, @pentp!

src/Orleans.Core/Serialization/TypeUtilities.cs
pentp: ValueTuple should be covered by the general value-type code below already.

PR #4949, opened 2018-09-06
TOC aborts under service failure

jason-bragg:
Fixed TOC aborts under service failure
Added minimal failure testing for TOC.

### DIFF
--- a/src/Orleans.Transactions/State/TransactionQueue.cs
+++ b/src/Orleans.Transactions/State/TransactionQueue.cs
@@ -22,13 +22,13 @@ namespace Orleans.Transactions.State
         private readonly JsonSerializerSettings serializerSettings;
         private readonly BatchWorker storageWorker;
         private readonly BatchWorker confirmationWorker;
-        private readonly ILogger logger;
+        protected readonly ILogger logger;
         private readonly Dictionary<Guid, TransactionRecord<TState>> confirmationTasks;
         private CommitQueue<TState> commitQueue = new CommitQueue<TState>();
 
         private StorageBatch<TState> storageBatch;
 
-        private TransactionalStatus problemFlag;
+        protected TransactionalStatus problemFlag;
         // the queues handling the various stages
 
         private int failCounter;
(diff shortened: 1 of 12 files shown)


src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -482,6 +482,10 @@ private async Task StorageWork()
                     {
                         // process all committable entries, adding storage events to the storage batch
                         CollectEventsForBatch(committableEntries);
+                        if (problemFlag != TransactionalStatus.Ok)
```
ReubenBond: `problemFlag` reminds me of [errno](http://man7.org/linux/man-pages/man3/errno.3.html) and [GetLastError](https://msdn.microsoft.com/en-us/library/windows/desktop/ms679360%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396), which makes me uncomfortable.

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -482,6 +482,10 @@ private async Task StorageWork()
...
+                        {
+                            return;
+                        }
```
ReubenBond: Should this be calling `NotifyOfRestore()`?

ReubenBond:
LGTM as long as the Notify comment is addressed

src/Orleans.Transactions/State/TransactionQueue.cs
jason-bragg: Agree.  I openly acknowledge that this solution is kludgy, almost to the point of being a hack.  All three loops (lock worker, storage worker, and confirmation worker) need refactored.  I'm trying to fix things within the current code until I've time to do refactoring passes.

src/Orleans.Transactions/State/TransactionQueue.cs
jason-bragg: No.  The finally block will check the problemFlag and handle any issues.  If Notify needs be called again, the finally block will do so.

PR #4950, opened 2018-09-07
Fixed TransactionFaultInjectionTests

jason-bragg:
- Addressed transaction recovery issues.
- Fixed minor test issues.
- Still not added to functional, because they take too long to run.  Need to fix the abort issue, then we can remove delays and to functional

### DIFF
--- a/src/Orleans.Transactions/State/TransactionManager.cs
+++ b/src/Orleans.Transactions/State/TransactionManager.cs
@@ -15,8 +15,9 @@ namespace Orleans.Transactions.State
             this.queue = queue ?? throw new ArgumentNullException(nameof(queue));
         }
 
-        public Task<TransactionalStatus> PrepareAndCommit(Guid transactionId, AccessCounter accessCount, DateTime timeStamp, List<ParticipantId> writeResources, int totalResources)
+        public async Task<TransactionalStatus> PrepareAndCommit(Guid transactionId, AccessCounter accessCount, DateTime timeStamp, List<ParticipantId> writeResources, int totalResources)
         {
+            await this.queue.Ready();
             // validate the lock
             var valid = this.queue.RWLock.ValidateLock(transactionId, accessCount, out var status, out var record);
 
@@ -37,19 +38,19 @@ namespace Orleans.Transactions.State
             }
 
             this.queue.RWLock.Notify();
-            return record.PromiseForTA.Task;
+            return await record.PromiseForTA.Task;
         }
 
-        public Task Prepared(Guid transactionId, DateTime timeStamp, ParticipantId resource, TransactionalStatus status)
+        public async Task Prepared(Guid transactionId, DateTime timeStamp, ParticipantId resource, TransactionalStatus status)
         {
+            await this.queue.Ready();
             this.queue.NotifyOfPrepared(transactionId, timeStamp, status);
-            return Task.CompletedTask;
         }
 
-        public Task Ping(Guid transactionId, DateTime timeStamp, ParticipantId resource)
+        public async Task Ping(Guid transactionId, DateTime timeStamp, ParticipantId resource)
         {
+            await this.queue.Ready();
             this.queue.NotifyOfPing(transactionId, timeStamp, resource);
-            return Task.CompletedTask;
         }
     }
 }
--- a/src/Orleans.Transactions/State/TransactionManager.cs
+++ b/src/Orleans.Transactions/State/TransactionManager.cs
@@ -15,9 +15,8 @@ namespace Orleans.Transactions.State
             this.queue = queue ?? throw new ArgumentNullException(nameof(queue));
         }
 
-        public async Task<TransactionalStatus> PrepareAndCommit(Guid transactionId, AccessCounter accessCount, DateTime timeStamp, List<ParticipantId> writeResources, int totalResources)
+        public Task<TransactionalStatus> PrepareAndCommit(Guid transactionId, AccessCounter accessCount, DateTime timeStamp, List<ParticipantId> writeResources, int totalResources)
         {
-            await this.queue.Ready();
             // validate the lock
             var valid = this.queue.RWLock.ValidateLock(transactionId, accessCount, out var status, out var record);
 
@@ -38,13 +37,13 @@ namespace Orleans.Transactions.State
             }
 
             this.queue.RWLock.Notify();
-            return await record.PromiseForTA.Task;
+            return record.PromiseForTA.Task;
         }
 
-        public async Task Prepared(Guid transactionId, DateTime timeStamp, ParticipantId resource, TransactionalStatus status)
+        public Task Prepared(Guid transactionId, DateTime timeStamp, ParticipantId resource, TransactionalStatus status)
         {
-            await this.queue.Ready();
             this.queue.NotifyOfPrepared(transactionId, timeStamp, status);
+            return Task.CompletedTask;
         }
 
         public async Task Ping(Guid transactionId, DateTime timeStamp, ParticipantId resource)
--- a/src/Orleans.Transactions/State/TransactionQueue.cs
+++ b/src/Orleans.Transactions/State/TransactionQueue.cs
@@ -24,7 +24,8 @@ namespace Orleans.Transactions.State
         private readonly BatchWorker confirmationWorker;
         protected readonly ILogger logger;
         private readonly Dictionary<Guid, TransactionRecord<TState>> confirmationTasks;
-        private CommitQueue<TState> commitQueue = new CommitQueue<TState>();
+        private CommitQueue<TState> commitQueue;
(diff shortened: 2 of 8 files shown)


src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -347,6 +354,24 @@ public void NotifyOfCancel(Guid transactionId, DateTime timeStamp, Transactional
...
+        {
+            await Ready();
+            this.restoreTask = Restore();
+            await this.restoreTask;
+        }
+
+        /// <summary>
+        /// called on activation, and when recovering from storage conflicts or other exceptions.
+        /// </summary>
+        public async Task Ready()
```
ReubenBond: The doc comment is maybe a little inaccurate/non-descriptive

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -225,6 +228,7 @@ public void NotifyOfAbort(TransactionRecord<TState> entry, TransactionalStatus s
                         if (logger.IsEnabled(LogLevel.Trace))
                             logger.Trace($"aborting RemoteCommitEntry {entry.Timestamp:o} status={status}");
 
+                        entry.ConfirmationResponsePromise?.TrySetException(new OrleansException("Confirm failed: Status {status}"));
```
ReubenBond: The exception message string needs `$` so that status is included

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -347,6 +354,24 @@ public void NotifyOfCancel(Guid transactionId, DateTime timeStamp, Transactional
...
+            this.restoreTask = Restore();
+            await this.restoreTask;
+        }
+
+        /// <summary>
+        /// called on activation, and when recovering from storage conflicts or other exceptions.
+        /// </summary>
+        public async Task Ready()
+        {
+            if (!this.restoreTask.IsCompleted)
```
ReubenBond: If `Restore` fails (eg `storage.Load()` exception) then I imagine we'd need to restart recovery on the next call.

This logic will instead ignore failures on subsequent calls.
That seems incorrect/dangerous to me.

For example, this could manifest after a write conflict.

jason-bragg: First, this solution sucks, but addresses the problem until we can refactor the call patterns.

The issue encountered was that in recovery scenarios, we clear the locks and commit logs, then reload and recover.  The reload is an async operation, surrendering the grains turn.  This allows new reentrant calls to be processed while the state is empty and still being reloaded. 

By holding the restore task and awaiting it before any incoming reentrant call we block all such calls while the queue is restoring.  This call does not need to trigger anything, just delay new operations until any pending restore is complete.

In the case where there is no restore error, these awaits should be fast as they will mostly occur on completed tasks.  In the case where there IS a restore error, all the pending calls will fail, which is good.  The internal recovery logic is responsible for retrying the restore, not the Ready calls, they just make sure it’s complete before allowing new requests.

sebastianburckhardt: I have an idea for a simpler solution - we can use the reader-writer lock to make callers wait for the load to complete. I will try that out.

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -225,6 +228,8 @@ public void NotifyOfAbort(TransactionRecord<TState> entry, TransactionalStatus s
                         if (logger.IsEnabled(LogLevel.Trace))
                             logger.Trace($"aborting RemoteCommitEntry {entry.Timestamp:o} status={status}");
 
+                        entry.ConfirmationResponsePromise?.TrySetException(new OrleansException($"Confirm failed: Status {status}"));
```
sebastianburckhardt: Code is unreachable, should never happen: cannot receive an abort if the transaction has committed on the TM.

src/Orleans.Transactions/State/TransactionManager.cs
```diff
@@ -15,8 +15,9 @@ public TransactionManager(TransactionQueue<TState> queue)
             this.queue = queue ?? throw new ArgumentNullException(nameof(queue));
         }
 
-        public Task<TransactionalStatus> PrepareAndCommit(Guid transactionId, AccessCounter accessCount, DateTime timeStamp, List<ParticipantId> writeResources, int totalResources)
+        public async Task<TransactionalStatus> PrepareAndCommit(Guid transactionId, AccessCounter accessCount, DateTime timeStamp, List<ParticipantId> writeResources, int totalResources)
         {
+            await this.queue.Ready();
```
sebastianburckhardt: no need to wait here. (If grain is undergoing restore, lock validation is guaranteed to fail.)

src/Orleans.Transactions/State/TransactionManager.cs
```diff
@@ -37,19 +38,19 @@ public Task<TransactionalStatus> PrepareAndCommit(Guid transactionId, AccessCoun
...
 
             this.queue.RWLock.Notify();
-            return record.PromiseForTA.Task;
+            return await record.PromiseForTA.Task;
         }
 
-        public Task Prepared(Guid transactionId, DateTime timeStamp, ParticipantId resource, TransactionalStatus status)
+        public async Task Prepared(Guid transactionId, DateTime timeStamp, ParticipantId resource, TransactionalStatus status)
         {
+            await this.queue.Ready();
```
sebastianburckhardt: No need to wait here.

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -328,6 +333,9 @@ public async Task NotifyOfConfirm(Guid transactionId, DateTime timeStamp)
 
         public void NotifyOfCancel(Guid transactionId, DateTime timeStamp, TransactionalStatus status)
         {
+            if (logger.IsEnabled(LogLevel.Trace))
+                logger.Trace("NotifyOfConfirm. TransactionId: {TransactionId}, TimeStamp: {TimeStamp} Status: {TransactionalStatus}", transactionId, timeStamp, status);
```
sebastianburckhardt: method name in trace string does not match... you probably meant "NotifyOfCancel". Perhaps use {nameof(NotifyOfConfirm)} to ensure trace message stays in sync with method name?

src/Orleans.Transactions/State/TransactionalResource.cs
```diff
@@ -14,8 +14,9 @@ public TransactionalResource(TransactionQueue<TState> queue)
             this.queue = queue;
         }
 
-        public Task<TransactionalStatus> CommitReadOnly(Guid transactionId, AccessCounter accessCount, DateTime timeStamp)
+        public async Task<TransactionalStatus> CommitReadOnly(Guid transactionId, AccessCounter accessCount, DateTime timeStamp)
         {
+            await this.queue.Ready();
```
sebastianburckhardt: no need to wait here.

src/Orleans.Transactions/State/TransactionalResource.cs
```diff
@@ -33,32 +34,33 @@ public Task<TransactionalStatus> CommitReadOnly(Guid transactionId, AccessCounte
...
         public async Task Confirm(Guid transactionId, DateTime timeStamp)
         {
+            await this.queue.Ready();
             await this.queue.NotifyOfConfirm(transactionId, timeStamp);
         }
 
-        public Task Prepare(Guid transactionId, AccessCounter accessCount, DateTime timeStamp, ParticipantId transactionManager)
+        public async Task Prepare(Guid transactionId, AccessCounter accessCount, DateTime timeStamp, ParticipantId transactionManager)
         {
+            await this.queue.Ready();
```
sebastianburckhardt: no need to wait here.

src/Orleans.Transactions/State/TransactionQueue.cs
jason-bragg: NotifyOfAbort is called during any storage error.  If a storage error occurs when a confirmation is being processed, the promise will never be resolved, and the TM will timeout.  Sending this exception allows the TM to recover faster.  It's not strictly necessary, but recovery is delayed greatly when we don't resolve pending operations and rely on orleans timeouts.

sebastianburckhardt: ok, you are right - I mistook the "NotifyOfAbort" method for the code that handles the "Abort" message sent by the TA. But that one is called Abort. 

Is there some ambiguity with the "NotifyOf" prefix? Here it seems to mean "notify other participants of", but in other cases it means "notify this participant of".

ReubenBond:
Test failure unrelated

ReubenBond:
@dotnet-bot test functional

PR #4953, opened 2018-09-07
Minor revisions to transaction tracing

sebastianburckhardt:
This is a minor refactoring to give us better information when reading traces of failed tests.


src/Orleans.Transactions/State/ReaderWriterLock.cs
```diff
@@ -99,6 +99,9 @@ private class LockGroup : Dictionary<Guid, TransactionRecord<TState>>
                 if (group == currentGroup)
                 {
                     group.Deadline = DateTime.UtcNow + this.options.LockTimeout;
+
+                    if (logger.IsEnabled(LogLevel.Trace))
+                        logger.Trace($"set lock expiration at {group.Deadline:o}");
```
jason-bragg: For all new logging we're using "semantic logging".  We've not updated exiting logging, but that's planned.

https://softwareengineering.stackexchange.com/questions/312197/benefits-of-structured-logging-vs-basic-logging

For semantic logging this would look something like.
logger.Trace("set lock expiration at {Deadline}", group.Deadline.Tostring("o"));

xiazen:
needs rebase

PR #4961, opened 2018-09-11
Fix azure streaming tests

xiazen:
some azure streaming tests failed due to diff introduced to WaitUntilAsync in pr #4923 . It is not necessarily the changes in WaitUtilAcync is wrong, is how those tests uses it is slightly wrong IMO. For a quick fix, let's revert WaitUtilAsync's behavior to its original when `delayOnFail` is not set. @jason-bragg let me know if this affects the tx recovery fix you put in #4923 . 

The azure steaming tests failed due to that timing become hasher after WaitUtilAsync change. Before the pr, WaitUtilAsync would be waiting 1 sec regardless, but after the pr, it won't if predicate succeed. For tests such as `StreamingTests_Consumer_Producer_UnSubscribe`, the test use WaitUtilAsync to ensure producer produced. After the pr, producer would produce far less, leaving little time for consumer to react before being checked. The time of a message sent from producer to consumer is short, but apparently not that short ;)

### DIFF
--- a/src/Orleans.TestingHost/Utils/TestingUtils.cs
+++ b/src/Orleans.TestingHost/Utils/TestingUtils.cs
@@ -86,9 +86,9 @@ namespace Orleans.TestingHost.Utils
                     bool passed;
                     do
                     {
-                        passed = await predicate(false);
                         // need to wait a bit to before re-checking the condition.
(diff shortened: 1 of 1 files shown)


jason-bragg:

PR #4968, opened 2018-09-12
More fixed to TOC

jason-bragg:
Added commit operation to pre store actions.
This still doesn't work correctly with batching in all cases, but does not violate non-toc transaction semantics.  TOC is still not production ready.

### DIFF
--- a/src/Orleans.Transactions/State/StorageBatch.cs
+++ b/src/Orleans.Transactions/State/StorageBatch.cs
@@ -63,7 +63,8 @@ namespace Orleans.Transactions
 
         // follow-up actions, to be executed after storing this batch
         private List<Action> followUpActions;
-
+        private List<Func<Task<bool>>> storeConditions;
+        
         // counters for each type of event
         private int total = 0;
         private int prepare = 0;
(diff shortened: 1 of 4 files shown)


ReubenBond:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test functional

src/Orleans.Transactions/State/StorageBatch.cs
```diff
@@ -256,5 +257,24 @@ public void FollowUpAction(Action action)
             }
             followUpActions.Add(action);
         }
+
+        public void AddStorePreCondition(Func<Task<bool>> action)
+        {
+            if (this.storeConditions == null)
```
xiazen: why do we lazy init this one? can we init this in the constructor?

jason-bragg: We can.  I used lazy init cause:
- Only toc uses it, so it won't be used most of the time.
- Was following the pattern used by the after actions (the 'when in rome' rule).

I don't mind changing this if you feel strongly that it should be initialized in the constructor.

xiazen: It is a small fix, I'd like for it to be fixed here. the more cleaner code is, the better for maintenance

jason-bragg: Will do, sorry I didn't get back to the feedback in time.. :/

PR #4974, opened 2018-09-12
Remove Ideas for Contributions page.

JillHeaden:
This PR removes the page and the listing from the TOC.yml


sergeybykov:
Need to update src\Documentation\resources\Contributing.md that has a broken link now.

PR #4983, opened 2018-09-13
Added support for TableName on AWS legacy configurator

galvesribeiro:


ReubenBond:
@dotnet-bot test this please

src/AWS/Orleans.Clustering.DynamoDB/Membership/LegacyDynamoDBMembershipConfigurator.cs
```diff
@@ -44,6 +45,14 @@ internal static void ParseDataConnectionString(string dataConnectionString, Dyna
                     options.Service = value[1];
             }
 
+            var tableNameConfig = parameters.Where(p => p.Contains(TableNamePropertyName)).FirstOrDefault();
```
ReubenBond: nit: This where condition can be moved into the FirstOrDefault call.
nit2: For correctness, the where condition should be checking that TableNamePropertyName occurs before any `=` symbol, otherwise that could be the value instead of the key and result in incorrect behavior. It's very unlikely, though.

PR #4995, opened 2018-09-17
Revert Microsoft.ServiceFabric.Services version

rmja:
Upgrading even the patch version of the `Microsoft.ServiceFabric.Services` is a breaking change, and requires the user to upgrade their SF cluster. Please consider to revert this back to 3.0.456.


rmja:
@ReubenBond you can ignore this. I found out that the reason for me getting errors with the new library version was that i had disabled "Just My Code" in the debugger to inspect Orleans source, and that just incidently cought a lot of exceptions in `Microsoft.ServiceFabric.Services`. Re-enabling "Just My Code" shows that the new library version works with my cluster. Sorry...

PR #4996, opened 2018-09-17
Resolve transaction on abort.

jason-bragg:
Abort and Cancel calls are no longer one way.  When transactions return abort exceptions they can usually be retried immediately.  In some rare cases this will cause a single cascade abort.

Correctness testing and fault testing now enabled.


jason-bragg:

Issue #1496, opened 2016-02-24
Azure Blob storage provider does not sanitize blob names


sergeybykov:
@hyha600 pointed to this issue in https://github.com/dotnet/orleans/commit/c2506ea16b0bbe59fdb67e457e276e8463396f86#commitcomment-16287142.

Need to sanitize illegal special characters the same way the Azure Table provider does.

hyha:
Also please consider if we have a grain defined with generic class:

**public class B<T> : Grain<CState>, ICGrain<T> where T : class**

Using table storage, we can see an entry in GrainStates table with PK and RowKey as:

**A.BGrain`1[[CGrain, CGrain, Version=1.1.0.0, Culture=neutral, PublicKeyToken=null]]**
(The real class name is much longer than this, this is class FullName)

All "space", "comma", "equal", '`' chars seems working in Table Storage, but definitely will not work in Blob name:
https://msdn.microsoft.com/en-us/library/azure/dd135715.aspx

veikkoeeva:
Cross-linked more directly: https://github.com/dotnet/orleans/issues/1497. :)

richorama:
I believe blob names to be more permissive than table keys: https://msdn.microsoft.com/en-us/library/azure/dd135715.aspx

richorama:
Could I suggest we create a unit test which demonstrates a problem?

aRajeshKumar:
As per Azure Docs: https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata, the following are rules:

- A blob name must conforming to the following naming rules: 
- A blob name can contain any combination of characters. 
- A blob name must be at least one character long and cannot be more than 1,024 characters long, for blobs in Azure Storage. 
- The Azure Storage emulator supports blob names up to 256 characters long. For more information, see Use the Azure storage emulator for development and testing.
- Blob names are case-sensitive. 
- Reserved URL characters must be properly escaped. 
- The number of path segments comprising the blob name cannot exceed 254. A path segment is the string between consecutive delimiter characters (e.g., the forward slash '/') that corresponds to the name of a virtual directory. 

Azure Storage dll is exposing a method to validate the blob name. Src location: https://github.com/Azure/azure-storage-net/blob/master/Lib/Common/NameValidator.cs

)Method Name: **ValidateblobName**

Is this enough or we expecting to add more validations?

xiazen:
fixed by @aRajeshKumar in #5020

PR #5020, opened 2018-09-24
Added Validations for Blob Names and refactored the ContianerName & TableName validations

aRajeshKumar:
PR for: Azure Blob storage provider does not sanitize blob names (https://github.com/dotnet/orleans/issues/1496)


xiazen:
good job, I didn't know azure cloud wrote those NameValidator already.  Thank you !

PR #5033, opened 2018-09-27
Make MaxSockets in SocketManager configurable

sergeybykov:
Make MaxSockets in SocketManager configurable.
Raise the default from 200 to 500.

With clusters larger than 200 silos the hardcoded limit on the number of sockets causes massive churn of sockets that get constantly closed and reopened.


jason-bragg:

Issue #5021, opened 2018-09-24
The new code generator adds everything to feature.AddKnownType


pentp:
This was a problem with the old code generator already as documented in #4437, but with the new code generator it's a lot worse.

For example I have a functional tests project that has a single grain interface with a few simple calls, it references various other projects. Both code generators write about 200 lines of code for the grain method invoker, but then they generate endless `feature.AddKnownType` lines - the old one about 700 lines, the new one about **26000** lines! The generated assembly grows from 300KB to 6MB and is really slow to start up.

What's the purpose of `feature.AddKnownType`? It looks like it adds every type that is visible to Roslyn during codegen or something similar. And the added information doesn't look useful at all - it's just a list of type names.

ReubenBond:
Thanks for reporting, @pentp. I've opened #5026 with a fix. Hope it works well for you. There will still be some (maybe many) `AddKnownType` lines for assemblies which are referenced by your project but not by other projects (eg not referenced by Orleans), but the number should be reduced so that there are less duplicates.

This is a side effect of how type serialization works which I hope to remove in a future release, maybe 3.0.

ReubenBond:
@pentp this is a side effect of how type serialization works - it needs to know the list of all possible types before it encounters them during deserialization. If there's a *possibility* that a type might be serialized, it's included. That means that very few types are excluded.

We could consider a switch which tells the code generator to modify its behavior and try to be more conservative about which types it includes.

Ultimately, it would be good to entirely avoid this process by using a different method for type serialization.

Scanning all possible types at runtime would force all assemblies and transitive dependencies to be loaded during startup.

Alternatively, what do you suggest?

ReubenBond:
@pentp this is a side effect of how type serialization works - it needs to know the list of all possible types before it encounters them during deserialization. If there's a *possibility* that a type might be serialized, it's included. That means that very few types are excluded.

We could consider a switch which tells the code generator to modify its behavior and try to be more conservative about which types it includes.

Ultimately, it would be good to entirely avoid this process by using a different method for type serialization.

Scanning all possible types at runtime would force all assemblies and transitive dependencies to be loaded during startup (which is slow).

I'm open to suggestions.

pentp:
The old code generator included a lot less types, so it was tolerable (and still worked) - maybe the list of possible types can be reasonably limited?  I have some stats for known types of old vs new:

Project | Old | New
------------ | ------------ | -------------
Interfaces | 5053 | 7876
[...]

PR #5036, opened 2018-09-28
Filter static types from list of types known to serializer

ReubenBond:
See #5021

### DIFF
--- a/src/Orleans.CodeGenerator/CodeGenerator.cs
+++ b/src/Orleans.CodeGenerator/CodeGenerator.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


src/Orleans.CodeGenerator/CodeGenerator.cs
```diff
@@ -488,11 +488,16 @@ private bool ValidForKnownTypes(INamedTypeSymbol type)
                 case TypeKind.TypeParameter:
                 case TypeKind.Submission:
                 {
-                    if (this.log.IsEnabled(LogLevel.Trace)) this.log.LogTrace($"{nameof(ValidForKnownTypes)} skipping type {type} of kind {type.TypeKind}");
+                    // if (this.log.IsEnabled(LogLevel.Trace)) this.log.LogTrace($"{nameof(ValidForKnownTypes)} skipping type {type} of kind {type.TypeKind}");
```
sergeybykov: Why comment this line out?

ReubenBond: To match the other commented out lines, no real reason. I could delete them all or guard them using a different condition

sergeybykov: I think we either leave them uncommented and guarded by 
```csharp
if (this.log.IsEnabled(LogLevel.Trace))
```
or delete them if they aren't needed. We try not to leave commented out code in the codebase as a general rule.

ReubenBond: deleted

PR #5045, opened 2018-10-01
Cleaned up StreamId code

bboyle1234:
The ValueTuple type allowed a lovely cleanup of the StreamId class. 
When the Record types becomes a c# feature, another cleanup can be made.

### DIFF
--- a/src/Orleans.Core/IDs/StreamId.cs
+++ b/src/Orleans.Core/IDs/StreamId.cs
@@ -35,7 +35,7 @@ namespace Orleans.Streams
         private StreamId((Guid Guid, string ProviderName, string Namespace) key)
         {
             Key = key;
-            UniformHashCode = new Lazy<uint>(() => CalculateUniformHashCode(key));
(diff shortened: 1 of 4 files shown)


sergeybykov:
@benjaminpetit Do you think this will break backward compatibility?

bboyle1234:
I just realised last night that lazy initialization of `UniformHashCode` might not work on a deserialized `StreamId` with my changes. I need to test it, can we pause this PR until I've run some tests to confirm?

MelnikovIG:
> I just realised last night that lazy initialization of `UniformHashCode` might not work on a deserialized `StreamId` with my changes. I need to test it, can we pause this PR until I've run some tests to confirm?

I think you can use WIP label in PR title to hold review https://github.com/apps/wip

ReubenBond:
This change breaks wire format compatibility. Closing for now, but we can reopen it if there's a way to implement it while retaining compatibility.

PR #5061, opened 2018-10-02
Start using Span and new language features for increasing Orleans perfomance

MelnikovIG:
Initial PR to inject Span and C# 7.3 optimizations in Orleans project with simple example of using it, where we avoid unnecessary GC array allocations.


ReubenBond:
We are certainly keen to start making use of span, @MelnikovIG - so thank you for kicking this off!

`UseLocalhostClustering` is only called during configuration time, so I wouldn't consider it a candidate for `Span<T>` - I think the most appropriate place to start would be `ByteArrayBuilder`. Maybe refactoring that and `BinaryTokenStreamReader` and `BinaryTokenStreamWriter` together would be fitting, but I'm not sure.

MelnikovIG:
@ReubenBond it is hard to optimize `BinaryTokenStreamReader` and `BinaryTokenStreamWriter` with `Span` without platform support of them. NetCore 2.1 has support of using `Span` in primitive type constructors, but netstandard2.0 does not. But i found another place where we can avoid a lot of string allocations with `Span` (GrainReference.cs file), i have updated this PR, can you take a look?

src/Orleans.Core/Core/ClientBuilderExtensions.cs
```diff
@@ -188,7 +189,28 @@ public static IClientBuilder ConfigureLogging(this IClientBuilder builder, Actio
...
+        }
+
+        /// <summary>
+        /// Configures the client to connect to a silo on the localhost.
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="gatewayPorts">The local silo gateway ports.</param>
+        /// <param name="serviceId">The service id.</param>
+        /// <param name="clusterId">The cluster id.</param>
+        public static IClientBuilder UseLocalhostClustering(this IClientBuilder builder,
```
ReubenBond: I feel this change is unnecessary, but the rest look good to me. Generally speaking, APIs which are only called once are not candidates for spanification

MelnikovIG: Done

MelnikovIG:
@ReubenBond Updated with code review changes

csyszf:
The current `BinaryTokenStreamReader` have been optimized to avoid allocating, what the reading operation does mostly is : a) convert bytes slices to primitive type, b) copy payload bytes and deliver them to serializers. The former almost has no extra allocations because we can use the source byte array with offset to avoid allocating, and for b), deliver a non-allocated buffer slice maybe useful, but for now, all serializers need a copy of bytes(or string for json serializer) to serialize...
Perhaps `BinaryTokenStreamReader` can expose an extra api like `ReadOnlySequence[byte] ReadBytes(int count)` to support future version serializers.
BTW, @MelnikovIG , `netstandard2.0` can use `MemoryMarshal.Read<T>(ReadOnlySpan<byte> span)` to convert primitive types, reference: [BufferReader_binary.cs](https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs)

And for `BinaryTokenStreamWriter`/`ByteArrayBuilder`, similarly, no slicing and no extra allocation.

MelnikovIG:
@csyszf  Thanks, ```MemoryMarshal.Read<T>(ReadOnlySpan<byte> span)``` is good point for future optimizations. I think this PR can be initial point of using Spans in Orleans.

ReubenBond:
I would accept rewriting BinaryTokenStreamReader/ByteArrayBuilder/etc to improve their performance and to replace the BufferPool with `ArrayPool<byte>.Shared`. I consider those temporary or "stepping stone" solutions, though, and ultimately I would like us to move towards the approach adopted in Hagar in [Reader.cs](https://github.com/ReubenBond/Hagar/blob/master/src/Hagar/Buffers/Reader.cs) and [Writer.cs](https://github.com/ReubenBond/Hagar/blob/master/src/Hagar/Buffers/Writer.cs). In other words, we don't directly manage buffers, but instead are provided buffers by the consumer (which will also be Orleans, but this allows us to more easily use System.IO.Pipelines or perhaps Bedrock and related projects)

csyszf:
@ReubenBond `Hagar.Buffers.Reader/Writer`/`Hagar.Buffers.Writer` LGTM.
So there will be two pairs of `BuffReader/BuffWriter`: one for decoding/encoding message, one for serializing/deserializing message body, and they both read from the same buffers and write to the same `IBufferWriter<byte>`(Pipe.Writer).
I'll give it a try.

sergeybykov:
1 functional test failed - `UnitTests.ActivationsLifeCycleTests.DeactivateOnIdleTests.DeactivateOnIdle_NonExistentActivation_1`. Possibly noise.

sergeybykov:
@dotnet-bot test functionsl

sergeybykov:
@dotnet-bot test functional

ReubenBond:
Thank you, Igor!

sillyousu:
hi @MelnikovIG  , building in VS(15.8.7)  failed in my local machine.  
Do I miss something?

Issue #5069, opened 2018-10-03
Service Fabric Mesh support


mholo65:
I started thinking about adding support for [Service Fabric Mesh](https://docs.microsoft.com/en-us/azure/service-fabric-mesh/service-fabric-mesh-overview).

So far the docker-aspnet-core example works like a charm, example [here](https://github.com/mholo65/mesh-orleans-sample/tree/414d0d73c3a1e977de1667a977fc2e153645b992). But I want to take this further and use Service Fabric Mesh for clustering, and (when Service Fabric 6.4 lands) use Standalone Reliable Collections for persisting grain state.

For clustering, I started looking at current the current implementation in `Microsoft.Orleans.Clustering.ServiceFabric` and _"I think"_ that a similar approach would be doable, but instead of taking dependency on `System.Fabric.FabricClient`, one should try using the [Service Fabric Client REST Api](https://docs.microsoft.com/en-us/rest/api/servicefabric/sfclient-index#applicationresource-apissfclient-index-applicationresourcemd) instead.

Now to the questions:
- Is adding support for Service Fabric Mesh on your roadmap? AFAIK, Mesh will not bring the Actor programming model currently available in Service Fabric Native. People, (including me) are looking for alternatives if moving to Service Fabric Mesh. I would love to help adding Mesh support for Orleans.
- As I'm new to Orleans (and specifically extending Orleans), I might need some help on getting started. To add support for new clustering providers, am I right in saying that I need to implement the following services?:
  - `IMembershipOracle` or `IMembershipTable`
  - `IGatewayListProvider`.

sergeybykov:
@ReubenBond was about to publish a Mesh sample, but ran into an issue running it locally. Let's see what's the overlap here.

ReubenBond:
Thanks for opening this, @mholo65 :)

I have been working on various aspects of Service Fabric support lately.

For now, Mesh support is just in the form of a sample. When Standalone Reliable Collections are released, it would be *fantastic* if you wanted to add support for grain persistence. If you're working on clustering for Mesh, then implementing `IMembershipTable` backed by Standalone Reliable Collections would be the best approach, IMHO.

`IMembershipOracle` is used by the current SF clustering provider, but that isn't being shipped because of issues that users experienced in production which essentially come down to the provider not guaranteeing what `IMembershipTable` does. I'm looking into relaxing the requirements.

[...]

PR #5071, opened 2018-10-04
Service Fabric Mesh sample project

ReubenBond:
This is largely a port of the Service Fabric Mesh "Voting App" sample to use Orleans instead of an in-memory dictionary


Samples/2.0/ServiceFabricMesh/VotingApp/VotingData/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+        public static async Task Main(string[] args)
+        {
+            var silo = new SiloHostBuilder()
+                .ConfigureEndpoints(siloPort: 30001, gatewayPort: 30002, listenOnAnyHostAddress: true)
+                .Configure<ClusterOptions>(options =>
+                {
+                    options.ClusterId = "dev";
+                    options.ServiceId = "votingapp";
+                })
+                .UseAzureStorageClustering(options => options.ConnectionString = Environment.GetEnvironmentVariable("CLUSTERING_CONNECTION_STRING"))
```
mholo65: Regarding https://github.com/dotnet/orleans/issues/5069, do you know what it would take to support clustering out-of-the-box inside Mesh without the need for external storage? Once standalone reliable collections land, that’s one option, but maybe not ideal? It somehow feels like it could be tighter integrated with Setvice Fabric.

ReubenBond: Standalone reliable collections will make it possible. I don't believe that Mesh has any suitable facilities for clustering yet - clustering currently has to guarantee some things around consistency which DNS isn't suited for.

I'll reply on the issue you opened (thanks for that)

PR #5073, opened 2018-10-05
[WIP] Support Service Fabric Stateful Service hosting

### LINKED ISSUE #1059: Support deep Service Fabric integration

ReubenBond:
Fixes #1059. I consider this satisfying enough of #1059 that we can close that issue and add any extra "good to have" features in separate PRs with individual issues if required.

This PR will be split up into multiple parts:

1. Enabling packaging of `Microsoft.Orleans.Clustering.ServiceFabric` package, which was not included in 2.0. `IMembershipOracle` ..
2. `StatefulServicePlacement` - so that activations are placed deterministically by mapping from grain id to SF partition id (SF partition ids / counts are fixed at deployment time and the service must be deleted and redeployed to 'change' the partition count/scheme).
3. Changes to support placement directors which do not use the grain directory. In short, adds an `IsUsingGrainDirectory` prop to `PlacementStrategy` and removes the existing special-casing around `StatelessWorkerPlacement` being the only placement strategy which can bypass the directory. (PR #5074)
4. Changes to support activations which have a deterministic `ActivationId` (related to the above) so that a grain which is placed using `StatefulServicePlacement` has a fixed `ActivationId` which can be computed from its `GrainId` (i.e, 1:1 relationship). It's possible that this will be contentious, since previously an `ActivationId` was ephemeral. This change is important, though, because it lets us route calls without needing to learn the `ActivationId` from another source (the grain directory). This is implemented by adding an `IsDeterministicActivationId` property to `PlacementStrategy`, which is then used in `PlacementDirectorsManager.AddActivation`. (PR #5082)
5. A storage provider backed by Service Fabric Reliable Collections.
6. Hosting utilities for wiring everything together, to make hosting Orleans on SF as simple as possible.


NickDarvey:
RE: `StatefulServicePlacement` 
Is there any chance the application developer can influence the placement of grains?

## Value
The scenario I was thinking was where you might have an AccountGrain and DeviceGrain which have an affinity (the account owns the device) and are very chatty (the account aggregates stats from its devices), so you might want them placed locally to each other.

## Implementation
This might be very misguided and stem from not understanding how Orleans does placement and partitioning... but...

If I were to try and do this in something like a consumer of Azure Event Hubs or Service Fabric Stateful Services service I would pick the account ID as the _partition key_ for both accounts and devices so they end up on the same partition (and I'd ensure run those partitions on the same node).

I think introducing a partition key might be a little far-fetched here (because it'd need to be supplied everywhere you fetch a proxy for a grain), but could developers supply their own function for picking a partition instead of `t.GrainIdentity.GetUniformHashCode();` etcetera (in [StatefulServicePlacementDirector.OnAddActivation](https://github.com/dotnet/orleans/pull/5073/files#diff-a3a271edc419109459a5a14589e1f2aaR64))?

In my accounts-device example, perhaps you could encode the account ID in the grain ID of the device and the supplied function would just be plucking the account ID, running it through whichever hash and picking the partition.

aL3891:
Why did you close this one? It sounded so cool :/

ReubenBond:
Apologies for the slow reply, @aL3891 

We were cleaning up dormant PRs & closed this along with the others.

I have been focusing on other areas (eg, we rewrote networking on top of the Bedrock abstractions) and haven't had the time to complete this. We have internal teams running on SF via Stateless Services (their services are still stateful, but that state isn't managed by SF).

If you think this would be useful to you then please feel free to pick it up - I can advise.

aL3891:
No worries, using bedrock abstractions sound very nice as well :) its just that with the SF actors not seeing much in the ways of new features, i've been looking at alternatives and this would be a great fit.

I dont know enough about or orleans to pick this one up in a timley manner i think, but i'd love to dig more into it.

PR #5074, opened 2018-10-05
Allow placement strategies to skip directory registration

ReubenBond:
Changes to support placement directors which do not use the grain directory. In short, adds an `IsUsingGrainDirectory` prop to `PlacementStrategy` and removes some existing special-casing around `StatelessWorkerPlacement` being the only placement strategy which can bypass the directory.

This is a supporting PR for Service Fabric Stateful Service hosting #5073

### DIFF
--- a/src/Orleans.Core.Abstractions/Placement/PlacementStrategy.cs
+++ b/src/Orleans.Core.Abstractions/Placement/PlacementStrategy.cs
@@ -5,5 +5,10 @@ namespace Orleans.Runtime
     [Serializable]
     public abstract class PlacementStrategy
     {
+        /// <summary>
+        /// Returns a value indicating whether or not this placement strategy requires activations to be registered in
+        /// the grain directory.
+        /// </summary>
+        public virtual bool IsUsingGrainDirectory => true;
     }
 }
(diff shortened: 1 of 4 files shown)


src/Orleans.Core.Abstractions/Placement/PlacementStrategy.cs
```diff
@@ -5,5 +5,10 @@ namespace Orleans.Runtime
     [Serializable]
     public abstract class PlacementStrategy
     {
+        /// <summary>
+        /// Returns a value indicating whether or not this placement strategy requires activations to be registered in
+        /// the grain directory.
+        /// </summary>
+        public virtual bool IsUsingGrainDirectory => true;
```
sergeybykov: Will this break backward compatibility with 2.0 and 2/1?

benjaminpetit: It will because `PlacementStrategy` is sent from silo-to-silo when we merge the list of supported grains.

ReubenBond: This isn't serialized, though: it's not backed by a field.

If it was defined as follows, then that would break serialization:

```C#
public virtual bool IsUsingGrainDirectory { get; } = true;
```

sergeybykov: That's good. Will it work though for cluster in a (temporary) hybrid state with some silos and clients running a version (2.1) without this change and some a version (2.2) with it?

ReubenBond: I believe so - there's no serialization or behavioral change in this PR until we start adding new placement strategies.

PR #5078, opened 2018-10-07
(docs) fix links to presentations

aprooks:
GH-pages does not support links with spaces so I've escaped them.


sergeybykov:
Thank you, @aprooks!

PR #5082, opened 2018-10-09
Allow placement strategies to specify deterministic activation ids

ReubenBond:
When grains are placed using the (upcoming) Service Fabric Stateful Service strategy, there is no grain directory involvement and hence we need a mechanism to determine which activation id to use when routing messages.

This PR enables that functionality but keeps it *`internal`* so that it will only be exposed to the SF library for now. The reasoning here is that it's unlikely anyone else will need it and if someone stumbles upon this they may opt-in to it for their placement strategy without understanding the consequences.

If `PlacementStrategy.IsDeterministicActivationId` is `true`, then all activations for a particular grain will have an `ActivationId` whose `Key` is equal to the grain's `GrainId.Key`. If `false`, then no behavior changes and `ActivationId` is randomly generated.

Without this PR, initial calls to already-activated grains placed using `StatefulServicePlacement` will round-trip to the target silo using the incorrect generated `ActivationId` and the silo will need to forward those calls to the correct activation. The response message will invalidate the caller's routing cache, but will not specify the activation's new address, so the caller will continue to call using incorrect `ActivationId`s. This is *benign* but unpleasant and inefficient. Users would almost certainly complain about warnings and other garbage in their logs. It is also slower. This PR enables the first and subsequent call to succeed without any round-trips for routing info.

See #5073 under part 4:
> Changes to support activations which have a deterministic `ActivationId` (related to the above) so that a grain which is placed using `StatefulServicePlacement` has a fixed `ActivationId` which can be computed from its `GrainId` (i.e, 1:1 relationship). It's possible that this will be contentious, since previously an `ActivationId` was ephemeral. This change is important, though, because it lets us route calls without needing to learn the `ActivationId` from another source (the grain directory). This is implemented by adding an `IsDeterministicActivationId` property to `PlacementStrategy`, which is then used in `PlacementDirectorsManager.AddActivation`. (PR #5082)

### DIFF
--- a/src/Orleans.Core.Abstractions/Placement/PlacementStrategy.cs
+++ b/src/Orleans.Core.Abstractions/Placement/PlacementStrategy.cs
@@ -10,5 +10,11 @@ namespace Orleans.Runtime
         /// the grain directory.
         /// </summary>
         public virtual bool IsUsingGrainDirectory => true;
+
+        /// <summary>
+        /// Returns a value indicating whether or not activations using this strategy must have deterministic activation ids.
+        /// If true then activations have activation ids equal to their grain id, otherwise activations are given unique ids.
+        /// </summary>
+        internal virtual bool IsDeterministicActivationId => false;
     }
 }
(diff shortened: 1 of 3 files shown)


ReubenBond:
@dotnet-bot test functional

PR #5095, opened 2018-10-12
Rewrite BufferPool with ArrayPool<byte>.Shared

csyszf:
Let BufferPool wrap `ArrayPool<byte>.Shared` instead manage byte arrays directly, similar to what MemoryPool does.

Since [`ArrayPool<byte>.Shared` doesn't accept arrays which not from pool](https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs#L181), the workaround for this brokes the optimization in long bytes writing(like bytes from serializer).


src/Orleans.Core/Messaging/BufferPool.cs
```diff
@@ -1,80 +1,45 @@
...
+using System.Buffers;
 using System.Collections.Generic;
-using System.Threading;
+using System.Runtime.CompilerServices;
 using Orleans.Configuration;
 namespace Orleans.Runtime
 {
     internal class BufferPool
     {
+        private const int MaximumBufferSize = int.MaxValue;
```
ReubenBond: Is this unused?

csyszf: Yes, forgot to delete it. This is for `MemoryPool<T>.MaxBufferSize` which I'm planning to implement.

src/Orleans.Core/Messaging/BufferPool.cs
```diff
@@ -1,80 +1,45 @@
...
-            buffers = new ConcurrentBag<byte[]>();
 
-            var globalPoolSizeStat = IntValueStatistic.FindOrCreate(StatisticNames.SERIALIZATION_BUFFERPOOL_BUFFERS_INPOOL,
-                                                                    () => Count);
-            allocatedBufferCounter = CounterStatistic.FindOrCreate(StatisticNames.SERIALIZATION_BUFFERPOOL_ALLOCATED_BUFFERS);
+            // Some of statistics can not be recorded,
+            // because we aren't able to find out whether ArrayPool<byte>.Shared drops returned byte array or not.
             checkedOutBufferCounter = CounterStatistic.FindOrCreate(StatisticNames.SERIALIZATION_BUFFERPOOL_CHECKED_OUT_BUFFERS);
             checkedInBufferCounter = CounterStatistic.FindOrCreate(StatisticNames.SERIALIZATION_BUFFERPOOL_CHECKED_IN_BUFFERS);
-            droppedBufferCounter = CounterStatistic.FindOrCreate(StatisticNames.SERIALIZATION_BUFFERPOOL_DROPPED_BUFFERS);
```
ReubenBond: If these StatisticsNames are no longer applicable, we should delete them

ReubenBond:
I need to run distributed tests to determine whether or not there's a noticeable performance impact - I'll try to run it over night.

src/Orleans.Core/Messaging/BufferPool.cs
```diff
@@ -1,80 +1,45 @@
...
-        public string Name
-        {
-            get;
-            private set;
-        }
-
         internal static void InitGlobalBufferPool(MessagingOptions messagingOptions)
         {
-            GlobalPool = new BufferPool(messagingOptions.BufferPoolBufferSize, messagingOptions.BufferPoolMaxSize, messagingOptions.BufferPoolPreallocationSize, "Global");
+            GlobalPool = new BufferPool(messagingOptions.BufferPoolBufferSize, messagingOptions.BufferPoolPreallocationSize);
```
ReubenBond: Should this be renamed to MinimumBufferSize or something similar? It's no longer the exact size

sergeybykov:
@dotnet-bot test this please

ReubenBond:
Because this adds a new dependency, we shouldn't include it in 2.2.0, but I'll see if we can include it in 2.3.0. I need to test it in our load test cluster to see how it behaves.

src/Orleans.Core/Messaging/ByteArrayBuilder.cs
```diff
@@ -125,32 +125,21 @@ private void EnsureRoomFor(int n)
         /// <returns></returns>
         public ByteArrayBuilder Append(byte[] array)
         {
+            // We shouldn't introduce external buffer Since we need return byte array to ArrayPool<byte>.Shared.
```
ReubenBond: @sergeybykov @jason-bragg I have no issue with removing the "large foreign buffer" optimization, at least for now. What do you think?

An alternative approach may be to explicitly mark which buffers are pooled and which aren't.

src/Orleans.Core/Orleans.Core.csproj
```diff
@@ -30,12 +30,14 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjectionVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingVersion)" />
     <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)" />
+    <PackageReference Include="System.Buffers" Version="4.5.0" />
```
sergeybykov: Don't we need a variable to this version? And another one for `System.Memory`?

ReubenBond: yep, I have a change locally to fix that

src/Orleans.Core/Messaging/ByteArrayBuilder.cs
ReubenBond: I spoke with @jason-bragg and he is fine with this particular change

ReubenBond:
@csyszf do you mind if I remove the `MemoryPool<T>` implementation from `BufferPool` and the related classes (including `BufferSegment`)? After that, and after I perform more testing, I am ready to merge this.

Those other changes can come later, when there's a PR which uses them

ReubenBond:
I made those changes and rebased upon master, @csyszf. If the changes look good to you, please remove the [WIP] tag and we can merge this after load tests are complete

ReubenBond:
@dotnet-bot test functional

ReubenBond:
FYI, @csyszf and I discussed this PR after running performance tests. The implementation of ArrayPool<T>.Shared may not be suitable for our uses in its current form since it's designed for cases where the same thread rents & returns the buffer (and has degraded performance in other cases).

In our case, the Gateway does *not* follow this pattern because it attempts to deserialize the message body as late as possible (eg, on the receiving grain's thread, just before invocation). The primary reason for this is that there is no use deserializing a message which will just be reserialized and sent to a foreign silo (high probability when using gateway).

We need to determine what is the best approach. It may be that we use a different pool in the case of Gateway and Silo-to-Silo serialization. I'm not sure yet.

As Sun points out, the non-shared ArrayPool implementation does not have the same behavior (making it more suitable for sharing between threads). So as an intermediate step, this PR could be changed to use the non-shared implementation and we can investigate further later on.

Sun, what do you think?

csyszf:
@ReubenBond , I've created a branch for ConfigurableArrayPool, but I could not get a stable benchmark result.
I'll cherry pick that commit to this PR, Would you please test this and check out if it will fix the performance lossing?

And I think use dedicated pools for each case is a good idea, but I think it should be easier to do that after we implement Pipeline-based I/O and processing, like What Kestrel did.

ReubenBond:
I'll run tests on the latest commit. I was experimenting with changing the top-level serialization interface to support Pipelines here: https://github.com/ReubenBond/orleans/tree/perf/serializer-interface

ReubenBond:
Moving to 2.4.0 for now, since we're still working on it

ReubenBond:
This PR is no longer needed since we now have this functionality in the new networking layer which has been merged with #5436

Issue #5058, opened 2018-10-02
Log compact/aggregation not working after 2.x migration


galvesribeiro:
Hello folks!

According to the [docs](http://dotnet.github.io/orleans/Documentation/clusters_and_clients/monitoring/silo_error_code_monitoring.html): 

> If multiple logs of the same log code occur within a designated time interval (the default is >5 within 1 minute) then additional log messages with that log code are suppressed and output as a "bulk" entry with log code equal to the original log code + 500000. So for example, multiple 100157 entries will show in the logs as 5 x 100157 + 1 x 600157 log entry per minute.

After migrating to 2.x and using the new logging infrastructure based on MEL, the logs aren't being aggregated anymore.

Can someone take a look why?

galvesribeiro:
Ping

ReubenBond:
FYI I believe this is not an Orleans issue per-se.

Documentation should be updated to remove mention of de-duplication.

This is necessarily a concern of logging or log consumers & outside the scope of Orleans.

sergeybykov:
@JillHeaden Please update the doc page to remove the info about log bulking (deduplication).

galvesribeiro:
Ok cool. Now at least we will know that should not happen anymore. Thanks!

galvesribeiro:
Okey, looks like @JillHeaden fixed the docs. Closing the issue, thanks!

PR #5096, opened 2018-10-12
galvesribeiro reported an issue with the description of 'Bulk log mes…

JillHeaden:
…sage compaction' - Reuben replied that "Documentation should be updated to remove mention of de-duplication.

This is necessarily a concern of logging or log consumers & outside the scope of Orleans."

I am assuming that this means the entire entry for "Bulk log message compaction" should be removed from the table.


JillHeaden:
Fixes #5058

PR #5098, opened 2018-10-13
Fixed bug in transaction confirmation logic

jason-bragg:
- Refactored confirmation management.
- Confirmation task only resoved after write to storage.

### DIFF
--- a/src/Orleans.Transactions/State/ConfirmationWorker.cs
+++ b/src/Orleans.Transactions/State/ConfirmationWorker.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


xiazen:

src/Orleans.Transactions/State/ConfirmationWorker.cs
```diff
@@ -0,0 +1,151 @@
...
+                        p,
+                        transactionId,
+                        timestamp,
+                        () => p.Reference.AsReference<ITransactionalResourceExtension>()
+                            .Confirm(p.Name, transactionId, timestamp),
+                        this.logger))
+                    .ToList();
+
+            // attempts to confirm all, will retry every ConfirmationRetryDelay until all succeed
+            while ((await Task.WhenAll(confirmations.Select(c => c.Confirmed()))).Any(b => !b))
```
xiazen: is it ok to just keep retry without a limit? what if one of the participant is stuck and never return ? what's the expected behavior in this case?

jason-bragg: Unfortunately algorithm requires confirmations eventually succeed.  This logic only retries those that failed, but it will retry forever (or until grain/silo go away).

jason-bragg: Side note, original logic did this as well, just in a different manner.

PR #5101, opened 2018-10-16
Reenabled transaction recovery tests

jason-bragg:


jason-bragg:

PR #5105, opened 2018-10-17
The link to the last successful development build in the README is broken.

jthelin:


dzmitry-lahoda:
successful != stable . here is better pull #5111

jthelin:
Yes, successful != stable, per Jenkins terminology.
https://wiki.jenkins.io/display/JENKINS/Terminology

It's a subtle distinction, and I am not sure whether the feedback loop even exists for orleans to mark "successful" builds as not "stable". 

It is also not clear to me whether a link to the "latest clean development branch build from CI" even should try to make that distinction (after all, a "development" build is by definition not making any claims to be a "stable" release build - only that "all the tests passed") - but I am not going to make this into a religious issue.

@dzmitry-lahoda  - Am i correct in thinking that you are suggesting to replace this PR with PR #5111 ?

I would be fine with withdrawing this PR in favor of #5111 once all the functional tests pass for that PR.

dzmitry-lahoda:
Yes. I suggest to replace #5105 with #5111.
Unfortunately I cannot make all tests pass. I changed only `readme.md`, but some `unstable` irrelevant tests failed.  
Need to think how to reapply. I did changes from browser, so may create new pull and close this one. I do not see retry button in Jenkins for me. Will do first I think and create new pull.

jthelin:
Replaced by #5111

PR #5111, opened 2018-10-19
fixed build links and C# compilation

dzmitry-lahoda:


sergeybykov:
Thank you, @dzmitry-lahoda!

PR #5113, opened 2018-10-20
Update AccountTransfer sample 

andhesky:
* Update sample to the newer version of the transaction interface.
* Add ReadKey in sample client after an exception is thrown to keep the window open long enough to read the exception
* Fix an exception to remove reference to deprecated functionality.


sergeybykov:
Thank you, @andhesky!

PR #5133, opened 2018-10-25
Add event on gateway count changed

mikkoairaksinen:
Based on gitter discussion with @ReubenBond 

Adds an event that the user can listen for changes in the gateway counts. Contains the current count of gateways as well as the previous count (to differentiate between gainin and losing gateways).

It's hard to test the reconnect situation. The only way I found so far to make a test that passes is included, but it's not clean. I also had to implement a Reconnect functionality in the ClientMessageCenter for the test, but it seems to not actually connect to any silos. The gateways reconnect and send the proper events, but any grain requests after Reconnect time out. This makes me wonder if this is only a test cluster scenario (possibly due to faulty Reconnect implementation) or can it also happen in a regular cluster (which would make the entire event useless) 

TODO:
- [ ] Proper tests (need to figure out how to either fix the Reconnect in ClientMessageCenter or some other way of properly testing the reconnect situation)

- [x]  Whitespace cleanup


ReubenBond:
@dotnet-bot test this please

ReubenBond:
Hi @mikkoairaksinen, apologies for not getting to this until now.

Generally looks good to me. I will make a few minor edits before merge.

If anyone else has comments on this new event, please chime in here.

src/Orleans.Core/Messaging/ClientMessageCenter.cs
```diff
@@ -403,7 +403,14 @@ public void Disconnect()
...
         public void Reconnect()
         {
-            throw new NotImplementedException("Reconnect");
+            //TODO: This is probably not be the proper way to reconnect since the connection to the test cluster is sometimes not regained
+            foreach (var connection in gatewayConnections.Values.ToArray())
+            {
+                if (!connection.IsLive)
+                {
+                    connection.Start();
+                }
```
ReubenBond: This doesn't seem right to me. @benjaminpetit do you have any thoughts on how this should be implemented or if it should be implemented at all?

ReubenBond: On the other hand, given it's explicitly for testing only, I think it's fine

ReubenBond: Currently this wont work because `IsLive` is never set back to `true` among other potential issues (`ThreadState`) inside of `GatewayConnection`.

mikkoairaksinen: Apologies for the slow reply. I currently don't have access to my Windows machine and it seems not possible to use my Mac to do any proper development on the project, so until I get my Windows machine fixed I can't do any additional work on this pull request. (Sidenote: even on the Windows machine doing a clean checkout would not work properly as a lot of tests kept failing randomly, not sure if there was something wrong with the setup. Would still be nice to have some FAQ on how to setup the project as a new contributor as the steps in the README did not work out of the box and there was a bunch of stuff I had to do to even run a single test)

As for the Reconnect, I had strong suspicions this was not the proper way to do it. It was not easy to figure out what the proper way is (frankly, I still don't know). I tried various approaches to make a new Gateway fire a GatewayCountChanged event with an increased gateway count, but none of them seemed to work. If anyone can point out the proper way to do this then I can fix the code and make the tests pass properly.

ReubenBond:
If we remove the test and undo the `Reconnect()` change then I'm happy to merge this as-is

ReubenBond:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot  test this please

ReubenBond:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot test bvt

PR #5139, opened 2018-10-26
Update Who-Is-Using-Orleans.html

jkonecki:
Added MaxADR to the list of companies using Orleans


JillHeaden:
@sergeybykov - in light of this, should I put this page back in the 2.0 documents?

sergeybykov:
@jkonecki Thank you for this! There are a couple of problems though.

1. We deleted the page because it became rather misleading due to the fact that a very small subset of users were willing to add them to the list. Need to check why we still have the html page.

2. We moved doc sources to the docs branch, and that's what doc PRs need to target.

jkonecki:
No worries. If you don't maintain this page anymore you can abandon the PR.

PR #5148, opened 2018-10-30
Log warning when ClusterMembershipOptions.ValidateInitialConnectivity=true

benjaminpetit:
While it can be fine to run with this setting set to true in test environment, it's generally a very bad idea to use it in production.

This PR just add a warning when a silo tries to join a cluster with this setting enabled.

### DIFF
--- a/src/Orleans.Runtime/MembershipService/MembershipOracle.cs
+++ b/src/Orleans.Runtime/MembershipService/MembershipOracle.cs
@@ -473,7 +473,7 @@ namespace Orleans.Runtime.MembershipService
                 else
                     logger.Warn(
                         ErrorCode.MembershipSendingPreJoinPing,
-                        $"${nameof(ClusterMembershipOptions.ValidateInitialConnectivity)} is set to false. This is NOT a supported value for production environment.");
+                        $"${nameof(ClusterMembershipOptions.ValidateInitialConnectivity)} is set to false. This is NOT recommended for a production environment.");
             }
             
             TableVersion next = table.Version.Next();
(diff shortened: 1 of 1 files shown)


src/Orleans.Runtime/MembershipService/MembershipOracle.cs
```diff
@@ -466,8 +466,15 @@ private async Task<bool> TryUpdateMyStatusGlobalOnce(SiloStatus newStatus)
...
-            if (newStatus == SiloStatus.Active && this.clusterMembershipOptions.ValidateInitialConnectivity)
-                await GetJoiningPreconditionPromise(table);
+            if (newStatus == SiloStatus.Active)
+            {
+                if (this.clusterMembershipOptions.ValidateInitialConnectivity)
+                    await GetJoiningPreconditionPromise(table);
+                else
+                    logger.Warn(
+                        ErrorCode.MembershipSendingPreJoinPing,
+                        $"${nameof(ClusterMembershipOptions.ValidateInitialConnectivity)} is set to false. This is NOT a supported value for production environment.");
```
sergeybykov: I'd say it's supported but not recommended.

ReubenBond:
@gabikliot as far as I understand it, the issues come when silos don't have stable identities, which is mostly the case in container hosts or Service Fabric where the port is not fixed. Generally these environments have large churn (rolling upgrades, for ex, or maybe cluster balancing or maintenance). If a silo terminates without updating membership then its successor (with likely a different id) will see it, fail to connect to it, and hence fail to join the cluster until the 5 minute lee-way period is over.

@benjaminpetit correct me if I'm wrong

Issue #5166, opened 2018-11-07
Orleans TestCluster Deploy causes deadlock with xUnit


MelnikovIG:
In case of trying run TestCluster Deploy parallel with xUnit deadlock occured.

Steps to reproduce:
1) Create as many simple classes as you have logical processors or more (i.e i have 8 logical processors, i need to create 8 or more same classes)
```   
 public class Class1
    {
        [Fact, TestCategory("Functional")]
        public async Task Test()
        {
            var builder = new TestClusterBuilder(2);
            builder.Options.ServiceId = Guid.NewGuid().ToString();
            builder.ConfigureHostConfiguration(TestDefaultConfiguration.ConfigureHostConfiguration);
            var testCluster = builder.Build();

            await testCluster.DeployAsync();
        }
    }
```
2) Run all tests from created classes with xUnit

Result: deadlock at https://github.com/dotnet/orleans/blob/master/src/Orleans.TestingHost/InProcessSiloHandle.cs#L34

I think xUnit custom SyncronizationContext not allows such way of executing async metod result. I have solution with full async flow, but it can break public testing API.

PR #5167, opened 2018-11-07
Fix test cluster deploy deadlock

MelnikovIG:
Fixes https://github.com/dotnet/orleans/issues/5166 , but has some breacking changes public test APIs

### DIFF
--- a/src/Azure/Orleans.Hosting.AzureCloudServices/Hosting/AzureSilo.cs
+++ b/src/Azure/Orleans.Hosting.AzureCloudServices/Hosting/AzureSilo.cs
@@ -208,7 +208,7 @@ namespace Orleans.Runtime.Host
             if (config == null)
             {
                 host = new SiloHost(instanceName);
-                host.LoadOrleansConfig(); // Load config from file + Initializes logger configurations
+                host.LoadConfig(); // Load config from file + Initializes logger configurations
             }
             else
             {
@@ -285,7 +285,7 @@ namespace Orleans.Runtime.Host
 
             host.ConfigureSiloHostDelegate = ConfigureSiloHostDelegate;
 
-            host.InitializeOrleansSilo();
+            host.InitializeSilo();
             return StartSilo();
         }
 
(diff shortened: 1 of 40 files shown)


benjaminpetit:
Thanks @MelnikovIG !

I am all for removing these `GetAwaiter().GetResult()` calls. I don't think it is an issue if we have breaking changes here, since it's only public test APIs.

However, here the changes are only on the return type of the method, so I can see that, if someone isn't careful when upgrading, will not see the change immediately, and will not await some of the calls.

We could rename the method with an "Async" suffix to them, to make the breaking change more obvious :). I know some people don't like the "async word everywhere", but I think it's not a bad thing when you mix async and synchronous calls on the same class.

I also see that you didn't changed the Stop method, any strong reason on why we shouldn't make this call also async (and maybe provide with some "legacy" synchronous stop method, for the dispose method for example) ?

MelnikovIG:
@benjaminpetit marked breacking changes methods with async keyword, but i didnt changed stop methods yet, because they have a lot of dependencies.

@sergeybykov removed Orleans word from method names.

MelnikovIG:
@benjaminpetit hm, some tests started fail with
```System.Runtime.Serialization.SerializationException: 'Type 'System.Threading.Tasks.Task`1' in assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' is not marked as serializable.'```
I think i need you help to handle this exception, its strange.

MelnikovIG:
@dotnet-bot test functional

MelnikovIG:
@dotnet-bot test functional

MelnikovIG:
@dotnet-bot test bvt

MelnikovIG:
@dotnet-bot test functional

sergeybykov:
@MelnikovIG 👍 I kicked off a full functional run in VSO. Just to make sure because this touches TestCluster.

benjaminpetit:
Thanks!

PR #5174, opened 2018-11-12
Corrected link to sample grain for streams

Kritner:
Looks like some code was moved around at some point, correcting link.


sergeybykov:
Thank you, @Kritner!

PR #5191, opened 2018-11-22
trying to hide something which leads to errors

dzmitry-lahoda:
I did edit in github editor and it seems added new line. so I am trying to enforce documentation by some ide tricks.


sergeybykov:
Thank you, @dzmitry-lahoda!

PR #5198, opened 2018-11-28
Simple way to delay invoke an async call

bluexo:
Register an timer to delay invoke an async call , the timer will be destroyed automatically when first timer tick

```C#
 public abstract class Grain : IAddressable, ILifecycleParticipant<IGrainLifecycle>
 {
   ...
        protected void DelayInvoke(Func<Task> asyncCallback, TimeSpan dueTime)
        {
            this.RegisterTimer(async obj => await asyncCallback(), null, dueTime, TimeSpan.FromMilliseconds(-1));
        }
    ...
 }

```

### DIFF
--- a/src/Orleans.Core.Abstractions/Core/Grain.cs
+++ b/src/Orleans.Core.Abstractions/Core/Grain.cs
@@ -120,6 +120,18 @@ namespace Orleans
         }
 
         /// <summary>
+        /// Simple way to delay invoke an async callback
+        /// the timer will be destroyed automatically when first timer tick
+        /// </summary>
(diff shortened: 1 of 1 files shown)


bluexo:
@sergeybykov 
Some one want to do delay call , the general approach is ```Task.Delay(xxx)``` , but that will be block threads .

The `RegisterTimer`  like an periodic calls , must manual destroy , nobody knows that the timer period equals -1 millisecond will be destroyed automatically when first timer tick . 

In my game project , often use delay invoke , such as cast spell , I want to do that create a ```waitable , cancelable and continuable```  DelayInvoke method like unity3d ```Coroutine``` ,  use ```RegisterTimer```  is the first step

PR #5201, opened 2018-11-28
Add TimerManager as Task.Delay replacement

ReubenBond:
This is an alternative timer implementation for use in Transactions to reduce the lock contention on the `Timer` queue caused by `Task.Delay` inside transaction management code.

How it works:

* Each thread which registers a timer does so by adding a `DelayTimer` node to a linked list of `DelayTimer` instances. That list is stored in a thread local structure, `ThreadLocalQueue`.
* `TimerManager` maintains an array of all threads' `ThreadLocalQueue` instances in `allQueues`.
* Every 50 milliseconds, a .NET `Timer` (`QueueChecker`) fires to service all queues via `CheckQueues()`. The timer enumerates all thread-local queues in `allQueues`. For each queue:
  * Try to obtain a lock on the queue (using `queue.Lock`)
  * If the lock fails, increment `queue.Starvation` and if the queue is starved then spin until a lock can be obtained.
  * Enumerate all timers in the queue and move the expired timers into a new list (`ExpiredTimers`).
* Once all queues have been checked, enumerate the `ExpiredTimers` list and fire each one
* Discard the list of fired timers (all timers are one-shot)

This code is based on earlier work by @dVakulen in #2060

### DIFF
--- a/src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
+++ b/src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Runtime.CompilerServices;
 using System.Threading;
 
@@ -8,15 +9,25 @@ namespace Orleans.Threading
     /// </summary>
     internal sealed class RecursiveInterlockedExchangeLock
     {
-        private const int Unlocked = -1;
-        private int lockState = Unlocked;
+        private const int UNLOCKED = -1;
+
+        [ThreadStatic]
+        private static int threadId;
+        private int lockState = UNLOCKED;
+        private readonly Func<bool> spinCondition;
+
+        public RecursiveInterlockedExchangeLock()
+        {
+            this.spinCondition = this.TryGet;
+        }
+
+        private static int ThreadId => threadId != 0 ? threadId : threadId = Thread.CurrentThread.ManagedThreadId;
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public bool TryGet()
         {
-            var threadId = Thread.CurrentThread.ManagedThreadId;
-            var previousValue = Interlocked.CompareExchange(ref this.lockState, threadId, Unlocked);
-            return previousValue == Unlocked || previousValue == threadId;
+            var previousValue = Interlocked.CompareExchange(ref this.lockState, ThreadId, UNLOCKED);
+            return previousValue == UNLOCKED || previousValue == ThreadId;
         }
         
         /// <summary>
@@ -30,25 +41,16 @@ namespace Orleans.Threading
                 return;
             }
 
-            Spin();
-
-            void Spin()
-            {
-                var spinWait = new SpinWait();
-                while (!this.TryGet())
-                {
-                    spinWait.SpinOnce();
-                }
-            }
+            SpinWait.SpinUntil(this.spinCondition);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void Release() => Interlocked.Exchange(ref this.lockState, Unlocked);
+        public void Release() => Interlocked.Exchange(ref this.lockState, UNLOCKED);
 
         public override string ToString()
         {
             var state = Volatile.Read(ref this.lockState);
-            return state == Unlocked ? "Unlocked" : $"Locked by Thread {state}";
+            return state == UNLOCKED ? "Unlocked" : $"Locked by Thread {state}";
         }
     }
 }
--- a/src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
+++ b/src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
@@ -12,7 +12,7 @@ namespace Orleans.Threading
         private const int UNLOCKED = -1;
 
         [ThreadStatic]
-        private static int threadId;
+        private static int localThreadId;
         private int lockState = UNLOCKED;
         private readonly Func<bool> spinCondition;
 
@@ -21,7 +21,7 @@ namespace Orleans.Threading
             this.spinCondition = this.TryGet;
         }
 
-        private static int ThreadId => threadId != 0 ? threadId : threadId = Thread.CurrentThread.ManagedThreadId;
+        private static int ThreadId => localThreadId != 0 ? localThreadId : localThreadId = Thread.CurrentThread.ManagedThreadId;
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public bool TryGet()
@@ -45,7 +45,12 @@ namespace Orleans.Threading
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void Release() => Interlocked.Exchange(ref this.lockState, UNLOCKED);
+        public bool TryRelease()
+        {
+            var threadId = ThreadId;
+            var previousValue = Interlocked.CompareExchange(ref this.lockState, UNLOCKED, threadId);
+            return previousValue == UNLOCKED || previousValue == threadId;
+        }
 
         public override string ToString()
         {
--- a/src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
+++ b/src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
@@ -1,0 +1,54 @@
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Orleans.Threading
+{
+    /// <summary>
+    /// Lightweight recursive lock.
+    /// </summary>
+    internal sealed class RecursiveInterlockedExchangeLock
+    {
+        private const int Unlocked = -1;
+        private int lockState = Unlocked;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool TryGet()
+        {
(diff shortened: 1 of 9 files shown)


src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading;
+
+namespace Orleans.Threading
+{
+    /// <summary>
+    /// Lightweight recursive lock.
+    /// </summary>
+    internal sealed class RecursiveInterlockedExchangeLock
+    {
+        private const int Unlocked = -1;
```
sergeybykov: All-caps for the constant - `UNLOCKED`?

src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
```diff
@@ -0,0 +1,54 @@
...
+        private int lockState = Unlocked;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool TryGet()
+        {
+            var threadId = Thread.CurrentThread.ManagedThreadId;
```
sergeybykov: Wouldn't it be better to get thread ID once and keep in a `ThreadLocal` variable instead?

src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                return;
+            }
+
+            Spin();
+
+            void Spin()
+            {
+                var spinWait = new SpinWait();
+                while (!this.TryGet())
```
sergeybykov: Why not `SpinUntil(this.TryGet())`?

ReubenBond: Sounds good to me. It involves adding a new field, `Func<bool> spinCondition` to avoid allocations.

src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
ReubenBond: Fixed

src/Orleans.Core/Threading/RecursiveInterlockedExchangeLock.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Orleans.Threading
+{
+    /// <summary>
+    /// Lightweight recursive lock.
+    /// </summary>
+    internal sealed class RecursiveInterlockedExchangeLock
+    {
+        private const int Unlocked = -1;
+        private int lockState = Unlocked;
```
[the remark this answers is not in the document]
ReubenBond: We only interact with it using `Interlocked` methods, except for in the `ToString` which directly performs a `Volatile.Read`

If it was marked `volatile` then VS would complain that volatile takes no effect when you're using `ref`:

image

src/Orleans.Core/Timers/TimerManager.cs
```diff
@@ -0,0 +1,363 @@
...
+        }
+
+        /// <summary>
+        /// Registers a timer.
+        /// </summary>
+        public static void Register(T timer)
+        {
+            ExpiredTimers expired = null;
+            var queue = EnsureCurrentThreadHasQueue();
+            queue.Lock.Get();
```
sergeybykov: Shouldn't this line be inside the `try` to ensure `queue.Lock.Release()` will always execute?

src/Orleans.Core/Timers/TimerManager.cs
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    internal interface ITimerCallback
+    {
+        /// <summary>
+        /// The UTC time when this timer is due.
+        /// </summary>
+        DateTime DueTime { get; }
+
+        bool IsCanceled { get; }
```
sergeybykov: `IsCanceled` doesn't appear to be used.

src/Orleans.Core/Timers/TimerManager.cs
```diff
@@ -0,0 +1,363 @@
...
+    /// <summary>
+    /// Manages timers of a specified type, firing them after they expire.
+    /// </summary>
+    /// <typeparam name="T">The timer type.</typeparam>
+    internal static class TimerManager<T> where T : class, ITimerCallback, ILinkedListElement<T>
+    {
+        /// <summary>
+        /// The maximum number of times a queue can be denied servicing before servicing is mandatory.
+        /// </summary>
+        private const int MaxStarvation = 2;
```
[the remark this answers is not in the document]
ReubenBond: `TimerManager<T>` is static at the moment, we'd have to change that before making it configurable

src/Orleans.Core/Timers/TimerManager.cs
ReubenBond: In effect, it's not. It's from Dmitry's code. The reason for it was for things such as request timeouts, where usually we cancel the timer before it fires. It could also serve to support `CancellationToken`.

I can remove it for now, though, if that's preferred.

src/Orleans.Core/Timers/TimerManager.cs
ReubenBond: Perhaps I should add a `TryRelease` which only releases if the current thread actually holds the lock?

src/Orleans.Core/Timers/TimerManager.cs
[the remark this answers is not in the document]
sergeybykov: That's fine for now then.

src/Orleans.Core/Timers/TimerManager.cs
sergeybykov: My concern is the opposite - if the `try` doesn't execute (for whatever reason) the lock would never be released. I think putting the `Lock.Get()` line inside the try *should* guarantee that the lock will always get released.

src/Orleans.Core/Timers/TimerManager.cs
sergeybykov: That's fine with me.

src/Orleans.Core/Timers/TimerManager.cs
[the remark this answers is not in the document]
ReubenBond: It would be reasonable to want to configure `TimerTickMilliseconds`. `TimerTickMilliseconds` could also be made dynamic, but I figured that was too much of a change for now.

The NuGet package used to originally test the theory used a dedicated thread instead of a `Timer`.

src/Orleans.Core/Timers/TimerManager.cs
```diff
@@ -0,0 +1,354 @@
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Orleans.Threading;
+
+namespace Orleans.Timers
+{
+    public static class TimerManager
```
jason-bragg: Shouldn't public surface be interface, so components using it can use mocks during testing?

jason-bragg: Why static?  I thought we were actively moving away from the use of statics?

src/Orleans.Core/Timers/TimerManager.cs
```diff
@@ -0,0 +1,354 @@
...
+    {
+        public static Task Delay(TimeSpan timeout) => DelayUntil(DateTime.UtcNow + timeout);
```
jason-bragg: If we do make an interface for this, suggest Delay be an extension function.

src/Orleans.Core/Timers/TimerManager.cs
ReubenBond: That depends on where the assumed exception comes from. If it came from `SpinWait` then putting `Get` into the lock and *not* having try-release semantics could be incorrect.

ReubenBond: So I did both anyhow. The return value of `TryRelease` is never used and it will always succeed if the lock was acquired.

src/Orleans.Core/Timers/TimerManager.cs
ReubenBond: `Task.Delay` isn't an interface

ReubenBond: Similar to Task.Delay design. Timer management is naturally static and we'd need to rework a huge amount of code to inject some kind of `ITimerManager`

src/Orleans.Core/Timers/TimerManager.cs
ReubenBond: makes sense

src/Orleans.Core/Timers/TimerManager.cs
ReubenBond: I removed it. We will have the PR branch to consult if we ever want to bring it back

src/Orleans.Core/Timers/TimerManager.cs
jason-bragg: I don't think we use task.delay a lot outside of test.

How about a compromise, we keep implementations static and internal but wrap it in an interface.  That way we can address this later without changes to the public surface.  Once we expose this new feature as part of our public surface as a static, our options become more limited.

ReubenBond:
I made the changes requested in our meeting and also included the Transactions changes to use `ITimerManager` in this PR.

PR #5207, opened 2018-11-30
[WIP] Simplified BatchWorker

ReubenBond:
This implements a simpler/cheaper alternative to `BatchWorker` used in Transactions, Event Sourcing, and Multi-clustering.

The implementation sits alongside the existing implementation and only Transactions code is modified to use it (other consumers will still use the original `BatchWorker`).

The purpose of the batch worker is to deduplicate requests to process some backlog of work. It does this through its `Notify()` and `Notify(DateTime dueTime)` interface.

* `Notify()` schedules work immediately and if the worker is currently processing requests then it ensures that the worker will loop at least one more time to process the most recent request.
* `Notify(DateTime dueTime)` calls `Notify()` at the specified time.

### DIFF
--- a/src/Orleans.Core/Async/BatchWorker.cs
+++ b/src/Orleans.Core/Async/BatchWorker.cs
@@ -1,9 +1,95 @@
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Orleans.Timers.Internal;
 
 namespace Orleans
 {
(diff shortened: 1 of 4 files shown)


ReubenBond:
@dotnet-bot test functional

ReubenBond:
Closing this because it isn't necessary

PR #5210, opened 2018-12-01
Optimize memory allocation with custom EqualityComparer

ar4ebald:
The `BinaryTokenStreamWriter` uses dictionaries with `RuntimeTypeHandle` keys to lookup appropriate `SerializationTokenType` and serialization action for every field in an object graph. 
Since `RuntimeTypeHandle` does not implement IEquatable interface (even though it has `Equals(RuntimeTypeHandle)` method) dictionaries are using `ObjectEqualityComparer` which in turn uses `Equals(object)` and that causes a boxing allocation to occur at every lookup of a dictionary.
In this PR I'm proposing a usage of a custom `RuntimeTypeHandlerEqualityComparer` to eliminate allocation.

### DIFF
--- a/src/Orleans.Core/Serialization/BinaryTokenStreamWriter.cs
+++ b/src/Orleans.Core/Serialization/BinaryTokenStreamWriter.cs
@@ -91,7 +91,7 @@ namespace Orleans.Serialization
 
         static BinaryTokenStreamWriter()
         {
-            typeTokens = new Dictionary<RuntimeTypeHandle, SerializationTokenType>();
+            typeTokens = new Dictionary<RuntimeTypeHandle, SerializationTokenType>(RuntimeTypeHandlerEqualityComparer.Instance);
             typeTokens[typeof(bool).TypeHandle] = SerializationTokenType.Boolean;
             typeTokens[typeof(int).TypeHandle] = SerializationTokenType.Int;
             typeTokens[typeof(uint).TypeHandle] = SerializationTokenType.Uint;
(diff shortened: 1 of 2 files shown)


MelnikovIG:
I think some benchmarks will be good to prove improvements?

ar4ebald:
> I think some benchmarks will be good to prove improvements?

I have made a repo with a sample application for that benchmark -> https://github.com/ar4ebald/orleans-memory-allocation-test
I used dotTrace to measure memory traffic. Without optimization it allocated over 3GB in 50 secs just for Dictionary.TryGetValue. You can see it in dotTrace window:
image
Then I switched to an optimized version (it actually have another optimization of boxing, but that'll be anpther PR) and this traffix is gone. Memory traffic reduced drastically. 
dotTrace with new version:
image
P.S.
The other 3.8GB of memory from first snapshot is being allocated due to boxing in serialization of each element of an input list

MelnikovIG:
I am not sure it is correct benchmark since you are benchmark whole Orleans infrastructure. You can try  benchmark with https://github.com/dotnet/BenchmarkDotNet just for BinaryTokenStreamWriter class without other dependencies (try benchmark on different runtimes). So we can see how you are using this class with optimizations and perfomance results.

ReubenBond:
There are also some [serialization benchmarks](https://github.com/dotnet/orleans/blob/8d8cd7041370c6a02403e33fe4bb934a1fd589dc/test/Benchmarks/Serialization/SerializationBenchmarks.cs) in the Benchmarks project, run via `dotnet Benchmarks.dll Serialization`

Those benchmarks themselves might be allocation-heavy and likely could be tweaked.
I think the version in Hagar is likely a more suitable base: https://github.com/ReubenBond/Hagar/blob/5274ace1688ce4ebf2e254f810347eae07f1afd6/test/Benchmarks/ComplexTypeBenchmarks.cs#L156

ReubenBond:
It's also worth investigating whether or not just switching to `Type` instead of `RuntimeTypeHandle` would be better - what do you think?

ReubenBond:
Thank you for this PR, by the way! It's great to see more people diving into the internals!

If you're not already familiar, we've got a Gitter room: https://gitter.im/dotnet/orleans

PR #5213, opened 2018-12-04
Fix AzureSilo startup

ReubenBond:
We saw an issue with `AzureSilo` which amounts to discrepancies between the membership table entry published manually by `AzureSilo` during startup and the entry published by the `IMembershipTable` implementation.

Two main changes: 

* Do not manually publish an entry into the table
* Do not manually configure a silo generation (which will be overwritten later anyway).

We can include this post 2.2.0-rc1


ReubenBond:
@dotnet-bot test functional

Opened #5216 for the failure

PR #5214, opened 2018-12-04
Support ProxyGatewayEndpoint from legacy configuration

ReubenBond:
Propagates the `ProxyGatewayEndpoint` setting from legacy `NodeConfiguration` to `EndpointOptions`. The silo-to-silo endpoint options are being propagated correctly and don't need to change.

~~`GatewayPort` is redundant and should not be used if `GatewayEndpoint` is set, but it shouldn't hurt to explicitly set it anyway.~~ EDIT: setting gateway port here `GatewayPort` wasn't redundant: it was being used to determine whether or not to start a gateway, etc. I added changes to fix that.


ReubenBond:
@dotnet-bot test functional

Opened #5215 for the initial failure

PR #5223, opened 2018-12-05
Add warm up to transaction recovery tests and reduce grain logging

xiazen:
TransactionWillRecoverAfterRandomSiloFailure test failed on the first assert, where we send transaction request to grains for the first time, before we shutdown or kill silo. Failure are prepareTimeout and break lock. So I suspect it is because grains are taking too long to activate or grains logging are too verbose. So in order to make TransactionWillRecoverAfterRandomSiloFailure passing more consistent,  added a ping method to the grain interface and call Ping() to activate grains before transaction calls, and reduce grain logging level.

### DIFF
--- a/test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalBitArrayGrain.cs
+++ b/test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalBitArrayGrain.cs
@@ -211,13 +211,18 @@ namespace Orleans.Transactions.Tests.Correctness
             this.dataArray = dataArray;
             this.loggerFactory = loggerFactory;
         }
-
+        
         public override Task OnActivateAsync()
         {
             this.logger = this.loggerFactory.CreateLogger(this.GetGrainIdentity().ToString());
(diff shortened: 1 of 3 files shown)


test/Transactions/Orleans.Transactions.Tests/Grains/MultiStateTransactionalBitArrayGrain.cs
```diff
@@ -211,23 +211,28 @@ public class MultiStateTransactionalBitArrayGrain : Grain, ITransactionalBitArra
             this.dataArray = dataArray;
             this.loggerFactory = loggerFactory;
         }
-
+        
         public override Task OnActivateAsync()
         {
             this.logger = this.loggerFactory.CreateLogger(this.GetGrainIdentity().ToString());
-            this.logger.LogInformation($"GrainId : {this.GetPrimaryKey()}.");
+            this.logger.LogTrace($"GrainId : {this.GetPrimaryKey()}.");
```
ReubenBond: Perhaps this should have an if statement to guard it

ReubenBond:
This seems fine, but it shouldn't be necessary

PR #5232, opened 2018-12-07
Move transaction test runners to Transaction.Testkit projects

xiazen:
- Move transaction test runners to Transaction.Testkit projects
- Make transaction test projects use Transaction.Testkit.xUnit


src/Orleans.Transactions.Testkit.Xunit/ConsistencyTransactionTestRunner.cs
```diff
@@ -1,25 +1,23 @@
...
+using Orleans.TestingHost;
+using Orleans.Transactions.TestKit.Base.Consistency;
 using Xunit;
 using Xunit.Abstractions;
 
-namespace Orleans.Transactions.Tests
+namespace Orleans.Transactions.TestKit.xUnit
 {
-    public abstract class ConsistencyTransactionTestRunner : TransactionTestRunnerBase
+    public class ConsistencyTransactionTestRunnerxUnit : Base.ConsistencyTransactionTestRunner
```
jason-bragg: I don't think we need the Base. qualifier in the namespace.  Seems like the naming convention for Foo test of FooTestsRunner, FooTestRunnerXUnit, AzureFooTests Is suffice.  Base seems to imply some inheritance model, which we use, but don't have to.  Some runners in the system don't use inheritance.

xiazen: valid concern. Will remove.

jdom:
yay, I'm glad test kits are finally happening :)
BTW, what's the purpose of having a base project which is testing-framework agnostic? I'd just commit to using xUnit and that's it... Anyone writing a new provider will have to use xUnit to run the suite, which doesn't sound like a terrible thing. On the other hand, re-writing all the theories in a different test framework seems like it's too much work and potentially be done in a way that is inconsistent, loosing all the value of having the test kit for correctness validation in the first place.

xiazen:
the purpose is that if someone isn't using xUnit, we can later introduce another package which depends on other test framework which uses the test runners in the base project. All the actual test runner is in the base project, xUnit one just a collection of test runners which inherit the base ones

PR #5248, opened 2018-12-12
Added 'first' and 'last' to grain lifecycle stages.

jason-bragg:
Added first and last stages to grain lifecycle stages, to align with service lifecycle stages


ReubenBond:
If someone tries to register `GrainLifecycleStage.Last + 1` then they will be in for a shock.

We could set first to 0 instead and throw if anything is registered which is less than zero.

PR #5249, opened 2018-12-12
Revert: Fix call chain reentrancy (#5145, #5225)

sergeybykov:
#5145 apparently caused a noticeable performance regression (12%-15%). This is to revert it along with #5225 that accompanied it. We can revisit the change and look at optimizing it post 2.2.0.


ReubenBond:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test functional

PR #5251, opened 2018-12-12
Use typed loggers so it is easier to enable debug transaction logging

jsteinich:
It took longer than expected to enable debug transaction logging without enabling all debug logging since most of it isn't using typed loggers. This change should allow just using `Orleans.Transactions` as a source filter.


jsteinich:
Wow did I fail on this one. Sorry about that. I'll re-open as an issue for discussion.

PR #5253, opened 2018-12-12
Replaces previous Transaction page.

JillHeaden:
Updates content with new information

### DIFF
--- a/src/Documentation/grains/transactions.md
+++ b/src/Documentation/grains/transactions.md
@@ -13,7 +13,11 @@ Orleans transactions are opt-in.
 A silo must be configured to use transactions.
 If it is not, any calls to transactional methods on grains will receive an `OrleansTransactionsDisabledException`.
 To enable transactions on a silo, call `UseTransactions()` on the silo host builder.
-`var builder = new SiloHostBuilder().UseTransactions();`
+
+```csharp
+var builder = new SiloHostBuilder().UseTransactions();
+
+```
 
 ### Transactional State Storage
 
@@ -24,13 +28,15 @@ To use transaction-specific storage, the user can configure their silo using any
 
 Example:
 
-``` csharp
+```csharp
+
 var builder = new SiloHostBuilder()
-    .AddAzureTableTransactionalStateStorage(“TransactionStore”, options =>
+    .AddAzureTableTransactionalStateStorage("TransactionStore, options =>
     {
         options.ConnectionString = ”YOUR_STORAGE_CONNECTION_STRING”);
     })
     .UseTransactions();
+
 ```
 
 For development purposes, if transaction specific storage is not available for the data store you need, an `IGrainStorage` implementation may be used instead.
@@ -46,27 +52,30 @@ The attribute needs indicate how the grain call behaves in a transactional envir
 It indicates this via the transaction options below:
 
 - `TransactionOption.Create` - Call is transactional and will always create a new transaction context, even if called within an existing transaction context.
-- `TransactionOption.Join Call` - Logic is transactional but can only be called within the context of an existing transaction.
+- `TransactionOption.Join` - Call logic is transactional but can only be called within the context of an existing transaction.
 - `TransactionOption.CreateOrJoin` - Call is transactional. If called within the context of a transaction, it will use that context, else it will create a new context.
 - `TransactionOption.Suppress` - Call is not transactional but can be called from within a transaction. If called within the context of a transaction, the context will not be passed to the call.
 - `TransactionOption.Supported` - Call is not transactional but supports transactions. If called within the context of a transaction, the context will be passed to the call.
-- `TransactionOption.NotAllowed` - Call is not transactional and cannot be called from within a transaction. If called within the context of a transaction, it will throw a not supported exception.
+- `TransactionOption.NotAllowed` - Call is not transactional and cannot be called from within a transaction. If called within the context of a transaction, it will throw a `NotSupportedException`.
 
 Calls can be marked as “Create”, meaning the call will always start its own transaction.
 
-``` csharp
+```csharp
+
 public interface IATMGrain : IGrainWithIntegerKey
 {
     [Transaction(TransactionOption.Create)]
     Task Transfer(Guid fromAccount, Guid toAccount, uint amountToTransfer);
 }
+
 ```
 
-The Transfer operation in the above atm grain will always start a new transaction which involves the two referenced accounts.
+The Transfer operation in the above ATM grain will always start a new transaction which involves the two referenced accounts.
 The transactional operations Withdraw and Deposit on the account grain are marked “Join”, indicating that they can only be called within the context of an existing transaction, which would be the case if they were called during `IATMGrain.Transfer(…)`.
 The `GetBalance` call is marked `CreateOrJoin` so it can be called from within an existing transaction, like via `IATMGrain.Transfer(…)`, or on its own.
 
-``` csharp
+```csharp
+
 public interface IAccountGrain : IGrainWithGuidKey
 {
     [Transaction(TransactionOption.Join)]
@@ -78,6 +87,7 @@ public interface IAccountGrain : IGrainWithGuidKey
     [Transaction(TransactionOption.CreateOrJoin)]
     Task<uint> GetBalance();
 }
+
 ```
 
 ### Grain Implementations
@@ -85,32 +95,37 @@ public interface IAccountGrain : IGrainWithGuidKey
 A grain implementation needs to use an `ITransactionalState` facet (see Facet System) to manage grain state via ACID transactions.
 
 ```csharp
+
     public interface ITransactionalState<TState>  
         where TState : class, new()
     {
         Task<TResult> PerformRead<TResult>(Func<TState, TResult> readFunction);
         Task<TResult> PerformUpdate<TResult>(Func<TState, TResult> updateFunction);
     }
+
 ```
 
 All read or write access to the persisted state must be performed via synchronous functions passed to the transactional state facet.
 This allows the transaction system to perform or cancel these operations transactionally.
 To use a transactional state within a grain, one needs only define a serializable state class to be persisted and declare the transactional state in the grain’s constructor with a `TransactionalState` attribute which declares the state name, and (optionally) which transactional state storage to use (see Setup).
 
-``` csharp
-    [AttributeUsage(AttributeTargets.Parameter)]
-    public class TransactionalStateAttribute : Attribute
+```csharp
+
+[AttributeUsage(AttributeTargets.Parameter)]
+public class TransactionalStateAttribute : Attribute
+{
+    public TransactionalStateAttribute(string stateName, string storageName = null)
     {
-        public TransactionalStateAttribute(string stateName, string storageName = null)
-        {
-            …
-        }
+      …
     }
+}
+
 ```
 
 Example:
 
-``` csharp
+```csharp
+
 public class AccountGrain : Grain, IAccountGrain
 {
     private readonly ITransactionalState<Balance> balance;
@@ -122,12 +137,12 @@ public class AccountGrain : Grain, IAccountGrain
         this.balance = balance ?? throw new ArgumentNullException(nameof(balance));
     }
 
-    Task IAccountGrain.Deposit(uint ammount)
+    Task IAccountGrain.Deposit(uint amount)
     {
         return this.balance.PerformUpdate(x => x.Value += amount);
     }
 
-    Task IAccountGrain.Withdrawal(uint ammount)
+    Task IAccountGrain.Withdrawal(uint amount)
     {
         return this.balance.PerformUpdate(x => x.Value -= amount);
     }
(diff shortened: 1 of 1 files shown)


src/Documentation/grains/transactions.md
```diff
@@ -3,82 +3,110 @@ layout: page
...
-
-The TransactionsConfiguration contains the following configurable values:
-* TransactionIdAllocationBatchSize - To avoid a storage call on every transaction start, transaction Ids are allocated in batches. This is the number of new transaction Ids allocated per transaction Id generation request (storage call).
-* AvailableTransactionIdThreshold - A new batch of transaction Ids will be automatically allocated if the available ids drop below this threshold.
-* TransactionRecordPreservationDuration - How long to preserve a transaction record in the TM memory after the transaction has completed.  This is used to answer queries about the outcome of the transaction.
+Orleans transactions are opt-in.
+A silo must be configured to use transactions.
+If it is not, any calls to transactional methods on grains will receive an `OrleansTransactionsDisabledException`.
+To enable transactions on a silo, call `UseTransactions()` on the silo host builder.
+`var builder = new SiloHostBuilder().UseTransactions();`
```
ReubenBond: The user also needs to configure storage

C# code should be placed inside a block comment with syntax highlighting (triple back-tick)

For DocFx we have to specify C# like this (as far as I know:
    ``` csharp

src/Documentation/grains/transactions.md
```diff
@@ -3,82 +3,110 @@ layout: page
...
 var builder = new SiloHostBuilder()
-    [...]
-    .AddAzureTableGrainStorageAsDefault(options => options.ConnectionString = ”YOUR_STORAGE_CONNECTION_STRING”)
-    .UseInClusterTransactionManager()
-    .UseAzureTransactionLog(options => options.ConnectionString = ”YOUR_STORAGE_CONNECTION_STRING”) 
-    .UseTransactionalState();
-
-var host = builder.Build();
-await host.StartAsync();
+    .AddAzureTableTransactionalStateStorage(“TransactionStore”, options =>
```
ReubenBond: The quotes around TransactionStore here are smart quotes. We need to change them to regular quotes, `"`

src/Documentation/grains/transactions.md
```diff
@@ -3,82 +3,110 @@ layout: page
...
-For a grain to support transactions, transactional operations on a grain must be marked as being part of a transaction using the “Transaction” attribute.  Calls can be marked as “Required”, meaning the call can take part in a transaction among multiple other grains or as “RequiresNew” indicating that it will already start its own transaction.
+### Grain Interfaces
 
-Example:
+For a grain to support transactions, transactional operations on a grain interface must be marked as being part of a transaction using the “Transaction” attribute.
+The attribute needs indicate how the grain call behaves in a transactional environment.
+It indicates this via the transaction options below:
+
+- `TransactionOption.Create` - Call is transactional and will always create a new transaction context, even if called within an existing transaction context.
+- `TransactionOption.Join Call` - Logic is transactional but can only be called within the context of an existing transaction.
```
ReubenBond: `Join Call` -> `Join`

src/Documentation/grains/transactions.md
```diff
@@ -3,82 +3,110 @@ layout: page
...
+- `TransactionOption.CreateOrJoin` - Call is transactional. If called within the context of a transaction, it will use that context, else it will create a new context.
+- `TransactionOption.Suppress` - Call is not transactional but can be called from within a transaction. If called within the context of a transaction, the context will not be passed to the call.
+- `TransactionOption.Supported` - Call is not transactional but supports transactions. If called within the context of a transaction, the context will be passed to the call.
+- `TransactionOption.NotAllowed` - Call is not transactional and cannot be called from within a transaction. If called within the context of a transaction, it will throw a not supported exception.
```
ReubenBond: Perhaps specify the precise exception here, `System.NotSupportedException` or just `NotSupportedException`

src/Documentation/grains/transactions.md
```diff
@@ -3,82 +3,110 @@ layout: page
...
+
+``` csharp
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public class TransactionalStateAttribute : Attribute
+    {
+        public TransactionalStateAttribute(string stateName, string storageName = null)
+        {
+            …
+        }
+    }
```
ReubenBond: I believe we should remove the initial indenting here and in any other code blocks

src/Documentation/grains/transactions.md
```diff
@@ -88,31 +116,54 @@ public class AccountGrain : Grain, IAccountGrain
...
     {
-        return Task.FromResult(this.balance.State.Value);
+        return this.balance.PerformRead(x => x.Value);
     }
-}
+}`
+```
+
+In the above example, the attribute `TransactionalState` is used to declare that the ‘balance’ constructor argument should be associated with a transactional state named “balance”.
+With this declaration, Orleans will wire up an `ITransactionalState` instance with a state loaded from the transactional state storage named "TransactionStore" (see Setup).
```
ReubenBond: Maybe "inject" instead of "wire up", but I don't hold a strong opinion

src/Documentation/grains/transactions.md
```diff
@@ -88,31 +116,54 @@ public class AccountGrain : Grain, IAccountGrain
...
+The state can be modified via `PerformUpdate` or read via `PerformRead`.
+The transaction infrastructure will ensure that any such changes performed as part of a transaction, even among multiple grains distributed over an Orleans cluster, will either all be committed or reverted upon completion of the grain call which created the transaction (`IATMGrain.Transfer` in the above examples).
+
+### Calling Transactions
+
+Transactional methods on a grain interface are called the same as any other grain call.
```
ReubenBond: "the same" -> "the same way"? I'm not sure what sounds more natural

src/Documentation/grains/transactions.md
```diff
@@ -88,31 +116,54 @@ public class AccountGrain : Grain, IAccountGrain
...
-        return Task.CompletedTask;
+        return this.balance.PerformUpdate(x => x.Value += amount);
     }
 
     Task IAccountGrain.Withdrawal(uint ammount)
     {
-        this.balance.State.Value -= ammount;
-        this.balance.Save();
-        return Task.CompletedTask;
+        return this.balance.PerformUpdate(x => x.Value -= amount);
```
ReubenBond: I know this is an example, but there should be logic here to check for integer underflow. Otherwise, subtracting 1 from a balance of 0 makes the user a billionaire and we are relying on logic specified in `ATMGrain`, which I cannot see defined here.

src/Documentation/grains/transactions.md
```diff
@@ -88,31 +116,54 @@ public class AccountGrain : Grain, IAccountGrain
...
+### Calling Transactions
+
+Transactional methods on a grain interface are called the same as any other grain call.
+
+``` csharp
+    IATMGrain atm = client.GetGrain<IATMGrain>(0);
+    Guid from = Guid.NewGuid();
+    Guid to = Guid.NewGuid();
+    await atm.Transfer(from, to, 100);
+    uint fromBalance = await client.GetGrain<IAccountGrain>(from).GetBalance();
```
ReubenBond: nit: it's awkward to me to have a `uint` for balance, I would expect a `decimal` or a signed `int` (eg, if the currency is non-divisible tokens of some sort). Recommend `int`.

Overdrawing an account (having a negative balance) is normal.

This assumes we don't have validation logic (see above comment)

src/Documentation/grains/transactions.md
```diff
@@ -88,31 +116,54 @@ public class AccountGrain : Grain, IAccountGrain
...
+    uint toBalance = await client.GetGrain<IAccountGrain>(to).GetBalance();
 ```
 
-In the above example the attribute TransactionalState is used to declare that the ‘balance’ construction argument should be associated with a transactional state named “balance”.  With this declaration Orleans will wire up an ITransactionalState<Balance> instance with a state loaded from the default storage provider for the grain to use.  The state can be accessed and modified via the State property.  Any changes to the state need be recorded by calling ‘Save()’.  The transaction infrastructure will ensure that any such changes performed as part of a transaction, even among multiple grains distributed over an Orleans cluster, will either all be committed or reverted together.
+In the above calls, an atm grain is used to transfer 100 units of currency from one account to another.
```
ReubenBond: "atm" -> ATM

src/Documentation/grains/transactions.md
```diff
@@ -88,31 +116,54 @@ public class AccountGrain : Grain, IAccountGrain
...
+After the transfer is complete, both accounts are queried to get their current balance.
+The currency transfer as well as both account queries are performed as ACID transactions.
 
+As seen in the above example, transactions can return values within a task, like other grain calls, but upon call failure they will not throw application exceptions, but rather an `OrleansTransactionException` or `TimeoutException`.
```
ReubenBond: task -> `Task`?

ReubenBond: So we always wrap any user exceptions in transactional methods?

I feel uneasy about that, I'm not convinced it's the right thing to do from a user's perspective.

src/Documentation/grains/transactions.md
```diff
@@ -88,31 +116,54 @@ public class AccountGrain : Grain, IAccountGrain
...
+If the application throws an exception during the transaction and that exception causes the transaction to fail (as opposed to other system failures) the application exception will be the inner exception of the `OrleansTransactionException`.
+If a transaction exception is thrown of type `OrleansTransactionAbortedException`, the transaction failed and can be retried.
+Any other exception thrown indicates that the transaction is an unknown state.
```
ReubenBond: "the transaction terminated with an unknown state." perhaps?

ReubenBond:
Added comments. /cc @jason-bragg

src/Documentation/grains/transactions.md
JillHeaden: For the storage configuration addition, is mentioning the need to do so enough? Or should there be another subsection about it - @jason-bragg?


I fixed the markdown code formatting.

src/Documentation/grains/transactions.md
JillHeaden: Done.

>    .AddAzureTableTransactionalStateStorage("TransactionStore", options =>

src/Documentation/grains/transactions.md
JillHeaden: - `TransactionOption.Join` - Call logic is transactional but can only be called within the context of an existing transaction.

src/Documentation/grains/transactions.md
JillHeaden: I went with the less verbose option, to keep it like the other verbiage.

- `TransactionOption.NotAllowed` - Call is not transactional and cannot be called from within a transaction. If called within the context of a transaction, it will throw a `NotSupportedException`.

src/Documentation/grains/transactions.md
JillHeaden: I believe this, too.

```csharp

[AttributeUsage(AttributeTargets.Parameter)]
public class TransactionalStateAttribute : Attribute
{
    public TransactionalStateAttribute(string stateName, string storageName = null)
    {
      …
    }
}

```

src/Documentation/grains/transactions.md
JillHeaden: I made the change. When @jason-bragg sends out the related "Setup" document (I'm assuming?) I will take note of any more suitable phrase

> With this declaration, Orleans will inject an `ITransactionalState` instance with a state loaded from the transactional state storage named "TransactionStore" (see Setup).

src/Documentation/grains/transactions.md
JillHeaden: I agree. How about this:

>Transactional methods on a grain interface are called like any other grain call.

src/Documentation/grains/transactions.md
JillHeaden: This is beyond my current skill level. Can you send an example, please?

Also, I removed the extra 'm' from amount. That IS within my skill level.

src/Documentation/grains/transactions.md
JillHeaden: I'm happy to change this if it's okay with @jason-bragg

src/Documentation/grains/transactions.md
JillHeaden: So it is an Automatic Teller Machine (ATM) grain.
I fixed both non-code instances.

src/Documentation/grains/transactions.md
JillHeaden: The user exception question is another one for @jason-bragg 

And did you mean this...

> As seen in the above example, transactions can return values within a Task<T>, like other grain calls...

src/Documentation/grains/transactions.md
JillHeaden: I changed that. I could also see "... the transaction terminated **in** an unknown state."

src/Documentation/grains/transactions.md
jason-bragg: Storage has it's onw section under setup.

src/Documentation/grains/transactions.md
jason-bragg: This is just meant to show the pattern.  We can improve the logic but I don't think that will make the doc any clearer.  same with changing uint to int.

src/Documentation/grains/transactions.md
jason-bragg: Agreed, Task<T> is more clear.

Issue #4744, opened 2018-07-05
Support adding cluster clients to an app's IServiceCollection


jdom:
Since most modern frontends user either ASP.NET (Mvc or WebApi) and with the new generic Host that was shipped in Microsoft.Extensions.Hosting 2.1, we can expect most Orleans clients will be using the `IServiceCollection` abstraction for building their frontend hosts.
It would be good to provide a simple way to configure an `IClusterClientFactory` that can build 1 or more cluster clients. This can benefit from passing the cross-cutting services configured in the app's container into each of the `IClusterClient`s that get built.
In particular for now, pass in the logger factory and the host's configuration (whatever config providers that were configured using Microsoft.Extensions.Configuration), and potentially then use Polly for connection retries, etc.

Note that I'm explicitly not suggesting we should share the app's container with the IClusterClient(s), but instead share it with the factory and keep the cluster client's container as an internal implementation that is isolated from the app's container or lifecycle.

We can use [HttpClientFactory as an inspiration](https://github.com/aspnet/HttpClientFactory/blob/master/samples/HttpClientFactorySample/Program.cs).

Then, configuring an Orleans cluster client factory would be a matter of calling an extension method on an `IServiceCollection` and configuring the client (or named clients) normally, except without explicitly configuring Logging or Configuration.

```
// ... configure MVC, logging, etc in a generic Host

services.AddClusterClient("some client identifier", c =>
{
    c.UseAzureStorageClustering(...)
    c.Configure<ClusterOptions>(options =>
    {
        options.ClusterId = "dev";
        options.ServiceId = "HelloWorldApp";
    });
});
```
Of course we can also support a default (non-named) client.

Then you can just inject the `IClusterClientFactory` in your app code to get a hold of a particular client (or just inject the `IClusterClient` if it was registered as default.
The factory can also act as a hosted service (if using the generic Host) to gracefully shutdown all the clients when the host is shutting down.

For reasons why I don't advocate sharing the container, 1 is obvious: it would make the assumption that a client library must be configured in the same way as the app host, and it wouldn't even support multiple clients. For others, there's [this longer chat thread](https://gitter.im/dotnet/orleans?at=5b3e746370efc60660b06f0f) with @galvesribeiro

galvesribeiro:
Good! That factory pattern from HTTP is really nice. However one point:

> it would make the assumption that a client library must be configured in the same way as the app host

[...]

PR #5261, opened 2018-12-14
Microsoft.Extensions.Hosting support

### LINKED ISSUE #4702: Support Microsoft.Extensions.Hosting.Abstractions
In Orleans 2.0, we added `SiloHostBuilder`, which was designed to align with `IHostBuilder` from the `Microsoft.Extensions.Hosting.Abstractions` package (which wasn't yet released).

The intention was that `SiloHostBuilder` would be temporary and we would add support for the [generic host](https://github.com/aspnet/Hosting/issues/1163) once it became available.

It's available now, so we can consider adding support for it.

ASP.NET folks have a similar issue opened for `WebHostBuilder`: https://github.com/aspnet/Hosting/issues/1421. They've scheduled it for 3.0.0, so I imagine it will be a year or more before they move over.

I have a PoC for this on a branch here: https://github.com/dotnet/orleans/compare/master...ReubenBond:feature-generic-host

cc @galvesribeiro & @attilah who asked for this today

galvesribeiro:
This PR adds [Microsoft.Extensions.Hosting](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-2.1) support for Orleans (aka _Generic Host_).

This PR is being worked out aligned with @ReubenBond and the strategy described [here](https://github.com/dotnet/orleans/issues/4702#issuecomment-446684609).

Fixes #4702


src/Orleans.Runtime/Hosting/GenericHostExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.Extensions.Hosting;
+
+namespace Orleans.Hosting
+{
+    public static class GenericHostExtensions
+    {
+        public static IHostBuilder AddOrleans(this IHostBuilder hostBuilder, Action<ISiloBuilder> configureDelegate)
+        {
+            if (configureDelegate == null) throw new ArgumentNullException(nameof(configureDelegate));
+            var siloBuilder = new SiloBuilder(hostBuilder);
```
ReubenBond: This should probably use either the `Properties` dictionary or `IServiceCollection` - see how `ApplicationParts` does it via `ApplicationPartManagerExtensions.GetApplicationPartManager(builder.Properties)`.

AspNetCore uses `IServiceCollection` for the same thing: https://github.com/aspnet/AspNetCore/blob/321327f84b2b22dcff2e9beb06a9a64236c5cced/src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L74 << maybe that approach offers some benefits - unclear.

Happy for you to take either route. It ultimately shouldn't matter because the `SiloBuilder` itself is stateless. Maybe in future that will change.

pashchuk:
@galvesribeiro I see that progress has stopped 23 days ago, probably you are busy enough.  
Sorry for mentioning you but do you have any plans to finish this in the nearest future.  
I'm the one who would like to see this feature in the Orleans, so please let me know if I can help you somehow ;)

galvesribeiro:
@palpha thanks for the interest.

This PR is on @ReubenBond's pipeline. He had few other high priority issues but rest assured he will get to it asap.

Samples/2.2/GenericHost/src/OrleansClient/Program.cs
```diff
@@ -0,0 +1,88 @@
...
+                Console.ReadKey();
+                return 1;
+            }
+        }
+
+        private static async Task<IClusterClient> StartClientWithRetries()
+        {
+            attempt = 0;
+            IClusterClient client;
+            client = new ClientBuilder()
```
SebastianStehle: Why is the client not generated with the generic host interface? The client could live within an mvc app

galvesribeiro: Basically this PR is to make the silo compatible with the new hosting abstractions.

Yes, I do have the client already working but I prefer to put it outside in another PR.

jdom: BTW, I see the client emulating the HttpClientFactory, since you don't "host" a client.... it wouldn't be uncommmon to have more than 1 client, each client with it's own DI container (it should be almost an implementation detail of the client), and just sharing it's instance (the entry point) with the parent host (ie: ASP.NET Core) container. So basically the client builder wouldn't be replaced with the generic Host (as it should continue to work standalone), but instead there would be some extension methods to have the generic host configure the client factory, similar to how you could use HttpClient without an ASP.NET Core app.

SebastianStehle: Sure, but the client builder should use the same DI container, so that it can share configuration and logging and other services from the host.

jdom: Configuration and logging are some cross-cutting concerns that could be flowed into the builder from the parent host, without the need for them to share the container. I explained it in length here: #4744

ReubenBond:
@dotnet-bot test functional

Samples/2.2/GenericHost/src/SiloHost/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+                                options.ServiceId = "HelloWorldApp";
+                            })
+                            .Configure<EndpointOptions>(options => options.AdvertisedIPAddress = IPAddress.Loopback)
+                            .ConfigureApplicationParts(parts => parts.AddApplicationPart(typeof(HelloGrain).Assembly).WithReferences())
+                            .ConfigureLogging(logging => logging.AddConsole());
+                    })
+                    .Build();
+                await host.RunAsync();
+
+                await host.StopAsync();
```
ReubenBond: Perhaps this should wait for the user to hit Ctrl+C before terminating?

SebastianStehle: Just remove the `StopAsync()`. The host handles it for you.

src/Orleans.Runtime/Hosting/GenericHostExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// Configures the host builder to host an Orleans silo.
+        /// </summary>
+        /// <param name="hostBuilder">The host builder.</param>
+        /// <param name="configureDelegate">The delegate used to configure the silo.</param>
+        /// <returns>The host builder.</returns>
+        /// <remarks>
+        /// Calling this method multiple times on the same <see cref="IHostBuilder"/> instance will result in one silo being configured.
+        /// However, the effects of <paramref name="configureDelegate"/> will be applied once for each call.
+        /// </remarks>
+        public static IHostBuilder AddOrleans(this IHostBuilder hostBuilder, Action<ISiloBuilder> configureDelegate)
```
SebastianStehle: There should be an overload where you get a reference to the IConfiguration object in the delegate.

src/Orleans.Runtime/Hosting/SiloHostedService.cs
```diff
@@ -0,0 +1,48 @@
...
+            this.serviceProvider = serviceProvider;
+            this.ValidateSystemConfiguration();
+
+            this.siloHost = serviceProvider.GetRequiredService<ISiloHost>();
+            this.logger = serviceProvider.GetRequiredService<ILoggerFactory>().CreateLogger<SiloHostedService>();
+        }
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            this.logger.LogInformation("Starting Orleans Silo.");
```
SebastianStehle: If a hosted service does not start the Host will not fail to start. I think you should catch the exception here and stop the process. See: https://github.com/aspnet/AspNetCore/issues/5900

src/Orleans.Runtime/Hosting/GenericHostExtensions.cs
ReubenBond: Not sure if that's possible under this sub-builder pattern. The configuration can be pulled from the container or from the `hostBuilderContext`, so it can be used in `ConfigureServices(...)` calls, but it doesn't exist until `Build` is called on the `IHostBuilder`.

If we had separate containers for host & silo then it would be possible, since we could presumably access the host config while constructing the silo builder.

Samples/2.2/GenericHost/src/SiloHost/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace OrleansSiloHost
+{
+    public class Program
+    {
+        public static async Task<int> Main(string[] args)
+        {
+            try
+            {
+                var host = new HostBuilder()
+                    .AddOrleans(orleansBuilder =>
```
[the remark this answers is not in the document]
ReubenBond: Will fix

src/Orleans.Runtime/Hosting/SiloHostedService.cs
ReubenBond: @davidfowl do you have an opinion here? I see that the behavior is changing for 3.0

Samples/2.2/GenericHost/src/SiloHost/Program.cs
ReubenBond: Fixed

ReubenBond:
Approved for 2.3.0 during team sync

ReubenBond:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test functional

davidfowl:
cc @Tratcher @glennc @rynowak

src/Orleans.Runtime/Hosting/GenericHostExtensions.cs
SebastianStehle: You can just call the delegate inside the ConfigureServices of the Host, then it will be available. I think it should be possible to make decisions based on the environment and the configuration, e.g. which storage you want to use.

src/Orleans.Runtime/Hosting/SiloHostedService.cs
davidfowl: Stop the process? What would that look like?

SebastianStehle: Using IApplicationLifetime and calling StopApplication() ;)

Samples/2.2/GenericHost/src/OrleansClient/Program.cs
SebastianStehle: Of course, but that is about connection pools for clustering, just as an example

src/Orleans.Runtime/Hosting/SiloHostedService.cs
davidfowl: I’m not 100% sure that would work well. We don’t exactly plan for people to call Stop inside of StartAsync. I wouldn’t be surprised if it broke in weird ways

ReubenBond: I'd be ok with either:

* Holding this PR until Hosting propagates `IHostedService.StartAsync` exceptions out through `IHost.StartAsync`
* Providing some way for users to customize the behavior here.

I'd really rather not hard-code behavior to terminate the process abruptly when `StartAsync` fails. That would be painful UX, IMHO.

Or is there a suggested alternative?

sergeybykov: I don't think we should be stopping the process we don't own. Bubbling an exception up seems like the right and expected behavior here to me.

sergeybykov: To me, the question is who own the process. If we (the framework) didn't start it, we shouldn't kill it. I think we should signal to the app (with an exception) that we failed to start, and the app can kill the process if it desires so.

Tratcher: Note aspnet/AspNetCore#5900 only applies to webhost. The generic host already propagates StartAsync exceptions. Since we're moving AspNetCore to use generic host in 3.0 anyways I don't see any reason to block on a bug in webhost.

src/Orleans.Runtime/Hosting/GenericHostExtensions.cs
ReubenBond: @SebastianStehle are we allowed to call ConfigureServices from within ConfigureServices? I imagine it would throw from [here](https://github.com/aspnet/Extensions/blob/0de62ae930da31048ba7e54c8cd0c6c0bcbd9095/src/Hosting/Hosting/src/HostBuilder.cs#L197) since the collection would change during enumeration.

It could be made to work, though, by changing HostBuilder.

src/Orleans.Runtime/Hosting/SiloHostedService.cs
ReubenBond: Thanks, @Tratcher, I didn't realize that. Given that this, I suppose we should consider this a non-issue (or a known issue with webhost).

SebastianStehle: But please notify the host then, I would like to use this and need to stop the process. I had an issue with that because right now I host the silo in an hosted service and first it looked like the deployment went fine and then everything blew up later.

src/Orleans.Runtime/Hosting/GenericHostExtensions.cs
Tratcher: Don't call ConfigureServices from within ConfigureServices. Invoke the second ConfigureServices action inline instead.

src/Orleans.Runtime/Hosting/SiloHostedService.cs
ReubenBond: @SebastianStehle can you use `HostBuilder` instead of `WebHostBuilder`?

`WebHostBuilder` (really [HostedServiceExecutor](https://github.com/aspnet/AspNetCore/blob/1aa50faa290ecda304507981cd01ed92651d5e34/src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs#L30)) has this issue but `HostBuilder` does not. `WebHostBuilder` will be fixed in 3.0.

You could work around `WebHostBuilder`'s implementation by manually removing `SiloHostedService`, eg:

```C#
var defaultSiloHostedService = services.FirstOrDefault(svc => svc.ImplementationType.Name == "SiloHostedService");
                    if (defaultSiloHostedService != null) services.Remove(defaultSiloHostedService);
```

Then add your own `IHostedService` based on the existing `SiloHostedService` (nothing internal in there).

Would that work?

The alternative is that we introduce some mechanism for you to customize the behavior but I feel that's adding API/code where there shouldn't be any.

src/Orleans.Runtime/Hosting/GenericHostExtensions.cs
ReubenBond: See this commit: https://github.com/dotnet/orleans/pull/5261/commits/e36231d52b088eaefeba3cf9aa8ca3e9a923dd61

src/Orleans.Runtime.Abstractions/Hosting/ISiloBuilder.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.Hosting
+{
+    /// <summary>
+    /// Functionality for building <see cref="ISiloHost"/> instances.
+    /// </summary>
+    public interface ISiloBuilder
```
jdom: >ISiloBuilder [](start = 21, length = 12)

What if we just made the SiloHostBuilder implement this? I suppose it's done this way to avoid a breaking change and allow chaining calls that return ISiloHostBuilder, but is it worth it? or perhaps mark ISiloHostBuilder as obsolete and soon get rid of it. Adding duplicate extension methods for both ISiloBuilder and iSiloHostBuilder (triplicate if you count IServiceCollection, and then all the combinations that pass the context too) is a tax not just for the core team, but for every extension developer out there, and easy to get wrong.

src/Orleans.Runtime/Hosting/SiloHostedService.cs
SebastianStehle: Sure, that should work, I was not aware that HostBuilder does not have this problem.

src/Orleans.Runtime.Abstractions/Hosting/ISiloBuilder.cs
ReubenBond: Avoiding breakages is precisely the reason for doing it this way. We could also have implemented the extensions using generic parameters to return the same type as passed in and avoided duplication but that is more complicated and degrades the experience for methods which have non-inferred generic parameters (since the usr would also have to specify the silo builder type parameter).

We should remove the IServiceCollection parameters and consider obsoleting the ISHB bits soon

jdom:
Not from me. Couldn't finish a thorough review, but I liked what I saw

davidfowl:
Will this be available on a myget feed once merged? I'd like to play with it.

ReubenBond:
@davidfowl yep, it will be show up here: https://dotnet.myget.org/F/orleans-ci/api/v3/index.json

sergeybykov:
Thank you so much, @galvesribeiro! This is a major step toward making integration of Orleans with ASP.NET and other parts of the .NET Core stack easier and more natural.

galvesribeiro:
Thanks for accepting it. Looking forward for the release!

sergeybykov:
@davidfowl This is now available on https://dotnet.myget.org/gallery/orleans-ci as 2.2.4-ci.20190207.3.

davidfowl:
On it!

davidfowl:
UseOrleans should be in the Microsoft.Extensions.Hosting namespace.

ReubenBond:
Good point, thanks. Opened #5355 to rectify.

sergeybykov:
Merged and is included in 2.2.4-ci.20190208.3 on https://dotnet.myget.org/gallery/orleans-ci.

Issue #5250, opened 2018-12-12
A transaction with only 1 write participant incorrectly assumes that the write participant is local


jsteinich:
Setup:
Orleans 2.2.0-rc1
Have a transaction with 2 or more participants where only 1 of them writes.
*It may or may not be necessary for the 1st participant to be read only.
Wait for `RemoteTransactionPingFrequency` (60) seconds.
Observe the transaction get reverted.

Cause:
https://github.com/dotnet/orleans/blob/fb7699a779883a1cec27775037575012b156269e/src/Orleans.Transactions/State/TransactionQueue.cs#L777
The local commit assumes that it doesn't need to notify other participants if there is only one write participant. This is definitely incorrect if the TM isn't the write participant and I believe also incorrect if it is (though arguably other code should handle that).

Possible Solution:
I'm thinking changing it to 
```c#
if (entry.WriteParticipants.Count > 1 || (entry.WriteParticipants.Count > 0 && !entry.WriteParticipants[0].Equals(this.resource)))
```
would fix this particular problem, but I have some doubts about how read only participants are handled in read + write transactions since at least on the surface it looks like they are still expecting to get the confirmations.

I'm pretty sure this is a regression from 2.1.

jason-bragg:
In the current pattern, the TM must also be a write participant due to an optimization to perform prepare and commit in the same operation on the TM (see ITransactionManager.PrepareAndCommit), so if there is only one write participant it must be the TM.

Is this different from what you're seeing?

For clarity, I'm not saying there isn't a bug here, just trying to make sure I understand the behavior you're seeing an how that deviates from the design.

jsteinich:
That makes sense.
Yes, I'm getting a read only participant as the TM.

If no priority manager is present, `TransactionAgent` assigns the first "eligible" participant as the manager. Eligibility is defined at creation time and doesn't take into account the current transaction.

PR #5267, opened 2018-12-17
Transaction Manager must be a write participant

jsteinich:
Addresses https://github.com/dotnet/orleans/issues/5250

I didn't add any checks for priority managers since they are kind of special, but perhaps we should.


sergeybykov:
Thank you, @jsteinich!

sergeybykov:
[2.2.1](https://github.com/dotnet/orleans/releases/tag/v2.2.1) has been released with this and another fix to transactions. Thanks again, @jsteinich!

PR #5268, opened 2018-12-18
Add configurable timeout to wait for queued messages being forwarded

xiazen:
On my local test for "forwarding messages during silo graceful shutdown", one of the reasons for message forwarding failure is 

- OutboundMessageQueue stopped before queued message being forwarded to other silo. 

The causes can be
- application messages weren't queued into OutboundMessageQueue before OutboundMessageQueue  stopped
- messages queued into OutboundMessageQueue didn't finish forwarding before it stopped

This PR tries to address the two causeses by 
- adding a 2 sec delay after `catalog.DeactivateAllActivations()`, this hopefully give enough time for the messages (queued messages for deactivated activations) to be queued into OutboundMessageQueue 
- adding a RerouteQueuedMessagesOnSiloShutdownTimeout as the max wait time for OutboundMessageQueue to finish forwarding before stopping.

### DIFF
--- a/src/Orleans.Runtime/Configuration/Options/SiloMessagingOptions.cs
+++ b/src/Orleans.Runtime/Configuration/Options/SiloMessagingOptions.cs
@@ -76,5 +76,13 @@ namespace Orleans.Configuration
         /// For test only - Do not use in production
         /// </summary>
         public bool AssumeHomogenousSilosForTesting { get; set; } = false;
+
+        public static TimeSpan DEFAULT_REROUTE_QUEUED_MESSAFES_On_SILO_SHUTDOWN_TIMEOUT { get; set; } = TimeSpan.FromSeconds(10);
+
+        /// <summary>
+        /// How long the silo will wait for rerouting queued mesages, before it continues shutting down. 
+        /// </summary>
+        public TimeSpan RerouteQueuedMessagesOnSiloShutdownTimeout { get; set; } =
+            DEFAULT_REROUTE_QUEUED_MESSAFES_On_SILO_SHUTDOWN_TIMEOUT;
     }
 }
--- a/src/Orleans.Runtime/Configuration/Options/SiloMessagingOptions.cs
+++ b/src/Orleans.Runtime/Configuration/Options/SiloMessagingOptions.cs
@@ -77,12 +77,12 @@ namespace Orleans.Configuration
         /// </summary>
         public bool AssumeHomogenousSilosForTesting { get; set; } = false;
 
-        public static TimeSpan DEFAULT_REROUTE_QUEUED_MESSAFES_On_SILO_SHUTDOWN_TIMEOUT { get; set; } = TimeSpan.FromSeconds(10);
+        public static TimeSpan DEFAULT_SHUTDOWN_REROUTE_TIMEOUT { get; set; } = TimeSpan.FromSeconds(10);
 
         /// <summary>
         /// How long the silo will wait for rerouting queued mesages, before it continues shutting down. 
         /// </summary>
-        public TimeSpan RerouteQueuedMessagesOnSiloShutdownTimeout { get; set; } =
-            DEFAULT_REROUTE_QUEUED_MESSAFES_On_SILO_SHUTDOWN_TIMEOUT;
+        public TimeSpan ShutdownRerouteTimeout { get; set; } =
+            DEFAULT_SHUTDOWN_REROUTE_TIMEOUT;
     }
 }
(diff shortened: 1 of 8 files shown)


sergeybykov:
@dotnet-bot test this please

src/Orleans.Runtime/Configuration/Options/SiloMessagingOptions.cs
```diff
@@ -76,5 +76,13 @@ public class SiloMessagingOptions : MessagingOptions
         /// For test only - Do not use in production
         /// </summary>
         public bool AssumeHomogenousSilosForTesting { get; set; } = false;
+
+        public static TimeSpan DEFAULT_REROUTE_QUEUED_MESSAFES_On_SILO_SHUTDOWN_TIMEOUT { get; set; } = TimeSpan.FromSeconds(10);
```
sergeybykov: Typo: "MESSAFES", Also the name seems too long. Suggest "DEFAULT_SHUTDOWN_REROUTE_TIMOUT" instead.

src/Orleans.Runtime/Configuration/Options/SiloMessagingOptions.cs
```diff
@@ -76,5 +76,13 @@ public class SiloMessagingOptions : MessagingOptions
...
+
+        /// <summary>
+        /// How long the silo will wait for rerouting queued mesages, before it continues shutting down. 
+        /// </summary>
+        public TimeSpan RerouteQueuedMessagesOnSiloShutdownTimeout { get; set; } =
```
sergeybykov: Name seems too long. Suggest "ShutdownRerouteTimeout" instead.

src/Orleans.Runtime/Messaging/MessageCenter.cs
```diff
@@ -117,6 +119,19 @@ public void PrepareToStop()
...
 
+        private void WaitToRerouteAllQueuedMessages()
+        {
+            DateTime maxWaitTime = DateTime.UtcNow + this.messagingOptions.RerouteQueuedMessagesOnSiloShutdownTimeout;
+            while (DateTime.UtcNow < maxWaitTime)
+            {
+                var applicationMessageQueueLength = this.OutboundQueue.ApplicationMessageCount;
+                if (applicationMessageQueueLength == 0)
+                    break;
+                Task.Delay(TimeSpan.FromSeconds(1)).Wait();
```
sergeybykov: If we are blocking the thread anyway, there's no need to use Task.Delay(). Can do Thread.Sleep() instead.

Also, we might want to reduce the iteration wait time here to lower than a second because, hopefully, in most cases forwarding will be very fast, especially in test scenarios. Maybe 100 milliseconds instead?

src/Orleans.Runtime/Messaging/OutboundMessageQueue.cs
```diff
@@ -28,6 +28,15 @@ public int Count
             }
         }
 
+        public int ApplicationMessageCount
+        {
+            get
+            {
+                int n = senders.Where(sender => sender.IsValueCreated).Sum(sender => sender.Value.Count);
```
sergeybykov: Avoid code duplication on line 25 by changing it to:
```csharp
int n = ApplicationMessageCount;
```

Also, it might be cleaner to make it a method instead - `GetApplicationMessageCount()` because it performs some computation, and doesn't simply return a stored value. Similarly, I think `Count` should be refactored to `GetCount()`.

src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -761,6 +761,8 @@ await scheduler.QueueTask(this.membershipOracle.ShutDown, this.membershipOracleC
                         .WithCancellation(ct, "MembershipOracle Shutting down failed because the task was cancelled");
                     // Deactivate all grains
                     SafeExecute(() => catalog.DeactivateAllActivations().Wait(ct));
+                    //wait for all queued message sent to OutboundMessageQueue before MessageCenter stop and OutboundMessageQueue stop. 
+                    await Task.Delay(TimeSpan.FromSeconds(2));
```
sergeybykov: Why hardcoded 2 seconds? Don't we want to wait here for forwarding of queued messages?

xiazen: this is just to wait for messages to be enqueued into OutboundMessageQueue, not actually forwarding yet. So 2 sec should be enough. What should we do in your opinion>?

sergeybykov: Doesn't `catalog.DeactivateAllActivations().Wait(ct)` return only after all activations got deactivated and their messages got enqueued for forwarding?

sergeybykov:
>error CS1061: 'IOutboundMessageQueue' does not contain a definition for 'ApplicationMessageCount' and no accessible extension method 'ApplicationMessageCount' accepting a first argument of type 'IOutboundMessageQueue' could be found (are you missing a using directive or an assembly reference?)

sergeybykov:
@dotnet-bot test functional

PR #5276, opened 2018-12-22
Sort out LocalGrainDirectory shutdown sequence

xiazen:
- move LocalGrainDirectory stop to `silo.OnBecomeActiveStop` after `MembershipOracle.shutdown`, instead being trigged by MembershipOracleShutdown -> Sent notification to its SiloStatatusListeners about its shuttingdown -> LocalGrainDirectory.Stop(). and not awaiting the call. 

- sort out its shutdown sequence to be : 

1. exclude itself from GrainDirectoryPartition calculation, since it is shutting down and cannot serve as as GrainDirectoryPartition anymore
2. Stop Mantainers
3. GrainPartition handoff and we are awaiting this call 
4. Clear local activation cache
5. MarkStopPreparationCompleted

### DIFF
--- a/src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
+++ b/src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
@@ -184,7 +184,6 @@ namespace Orleans.Runtime.GrainDirectory
             Dictionary<GrainId, IGrainInfo> batchUpdate = localDirectory.DirectoryPartition.GetItems();
 
             await HandoffMyPartitionUponStop(batchUpdate, silosHoldingMyPartitionCopy, true);
-            localDirectory.DirectoryPartition.Clear();
         }
 
         internal void ProcessSiloAddEvent(SiloAddress addedSilo)
--- a/src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
+++ b/src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Orleans.Runtime.Configuration;
@@ -157,12 +158,12 @@ namespace Orleans.Runtime.GrainDirectory
             }
         }
 
-        internal void ProcessSiloStoppingEvent()
+        internal Task ProcessSiloStoppingEvent()
         {
-            ProcessSiloStoppingEvent_Impl();
+            return ProcessSiloStoppingEvent_Impl();
         }
 
-        private async void ProcessSiloStoppingEvent_Impl()
+        private async Task ProcessSiloStoppingEvent_Impl()
         {
             if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("Processing silo stopping event");
 
@@ -181,15 +182,9 @@ namespace Orleans.Runtime.GrainDirectory
             }
             // take a copy of the current directory partition
             Dictionary<GrainId, IGrainInfo> batchUpdate = localDirectory.DirectoryPartition.GetItems();
-            try
-            {
-                await HandoffMyPartitionUponStop(batchUpdate, silosHoldingMyPartitionCopy, true);
-                localDirectory.MarkStopPreparationCompleted();
-            }
-            catch (Exception exc)
-            {
-                localDirectory.MarkStopPreparationFailed(exc);
-            }
+
+            await HandoffMyPartitionUponStop(batchUpdate, silosHoldingMyPartitionCopy, true);
+            localDirectory.DirectoryPartition.Clear();
         }
 
         internal void ProcessSiloAddEvent(SiloAddress addedSilo)
@@ -215,7 +210,7 @@ namespace Orleans.Runtime.GrainDirectory
                     GrainDirectoryPartition splitPart = localDirectory.DirectoryPartition.Split(
                         grain =>
                         {
-                            var s = localDirectory.CalculateTargetSilo(grain);
+                            var s = localDirectory.CalculateGrainDirectoryPartition(grain);
                             return (s != null) && !localDirectory.MyAddress.Equals(s);
                         }, false);
                     List<ActivationAddress> splitPartListSingle = splitPart.ToListOfActivations(true);
@@ -261,7 +256,7 @@ namespace Orleans.Runtime.GrainDirectory
                             grain =>
                             {
                                 // Need to review the 2nd line condition.
-                                var s = localDirectory.CalculateTargetSilo(grain);
+                                var s = localDirectory.CalculateGrainDirectoryPartition(grain);
                                 return (s != null) && !predecessorOfNewSilo.Equals(s);
                             }, true);
                         directoryPartitionsMap[addedSilo] = splitPart;
--- a/src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
+++ b/src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
@@ -52,9 +52,6 @@ namespace Orleans.Runtime.GrainDirectory
         public RemoteGrainDirectory CacheValidator { get; private set; }
         public ClusterGrainDirectory RemoteClusterGrainDirectory { get; private set; }
 
-        private readonly TaskCompletionSource<bool> stopPreparationResolver;
-        public Task StopPreparationCompletion { get { return stopPreparationResolver.Task; } }
-
         internal OrleansTaskScheduler Scheduler { get; private set; }
 
         internal GrainDirectoryHandoffManager HandoffManager { get; private set; }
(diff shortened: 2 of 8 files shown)


src/Orleans.Runtime/MembershipService/MembershipOracle.cs
```diff
@@ -712,9 +722,9 @@ private void GossipToOthers(SiloAddress updatedSilo, SiloStatus updatedStatus)
                             throw exc;
                         }
                         return true;
-                    })
-                    .Ignore();
```
sergeybykov: Is this intentional that this `.Ignore()` is removed?

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -181,15 +182,9 @@ private async void ProcessSiloStoppingEvent_Impl()
...
-            {
-                await HandoffMyPartitionUponStop(batchUpdate, silosHoldingMyPartitionCopy, true);
-                localDirectory.MarkStopPreparationCompleted();
-            }
-            catch (Exception exc)
-            {
-                localDirectory.MarkStopPreparationFailed(exc);
-            }
+
+            await HandoffMyPartitionUponStop(batchUpdate, silosHoldingMyPartitionCopy, true);
```
sergeybykov: What is this throws?

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -972,7 +975,7 @@ public void InvalidateCacheEntry(ActivationAddress activationAddress, bool inval
             // for multi-cluster registration, the local directory may cache remote activations
             // and we need to remove them here, on the fast path, to avoid forwarding the message
             // to the wrong destination again
-            if (invalidateDirectoryAlso && CalculateTargetSilo(grainId).Equals(MyAddress))
+            if (invalidateDirectoryAlso && CalculateGrainDirectoryPartition(grainId).Equals(MyAddress))
```
sergeybykov: If `CalculateGrainDirectoryPartition` returns `null`, this will cause a NRE. Wouldn't it be safer to flip the check to
```csharp
MyAddress.Equals(CalculateGrainDirectoryPartition(grainId))
``` 
instead.

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -988,7 +991,7 @@ public void InvalidateCacheEntry(ActivationAddress activationAddress, bool inval
         /// <returns></returns>
         public SiloAddress GetPrimaryForGrain(GrainId grain)
```
sergeybykov: `GetPrimaryForGrain` can now also return `null`, which needs to be accounted for by its callers.

src/Orleans.Runtime/MembershipService/MembershipOracle.cs
```diff
@@ -395,7 +397,15 @@ private async Task UpdateMyStatusGlobal(SiloStatus status)
                 {
                     if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("-Silo {0} Successfully updated my Status in the Membership table to {1}", MyAddress.ToLongString(), status);
                     membershipOracleData.UpdateMyStatusLocal(status);
-                    GossipMyStatus();
+                    if (status == SiloStatus.Stopping || status == SiloStatus.ShuttingDown || status == SiloStatus.Dead)
+                    {
+                        GossipMyStatus().Wait(shutdownGossipTimeout);
```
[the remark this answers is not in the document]
xiazen: you are right, not sure why I used Wait here. If it throws, I think logging for now is sufficient?

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
xiazen: its caller LocalGrainDirectory catches it and `MarkStopPreparationFailed()`.

sergeybykov: But `localDirectory.DirectoryPartition.Clear()` will not get called as a result. Won't that cause a problem with forwarding of enqueued requests?

xiazen: yeah you are right. we need to catch it, log it  and continue

xiazen: I moved  DirectoryPartition.Clear(); to LocalGrainDirectory.Stop() after HandoffMyPartitionUponStop

sergeybykov:
The failure in functional tests looks potentially related:

https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/2483/testReport/junit/Tester.Forwarding/ShutdownSiloTests/SiloGracefulShutdown_ForwardPendingRequest/
>Orleans.Runtime.OrleansMessageRejectionException : Forwarding failed: tried to forward message NewPlacement Request S127.0.0.1:26291:284172193*cli/b20f44b7@4d95a69f->S127.0.0.1:26291:284172193*grn/CE1C219C/4020809b@b70215aa #5427[ForwardCount=2]: for 2 times after Non-existent activation to invalid activation. Rejecting now.

sergeybykov:
@dotnet-bot test functional

xiazen:
@sergeybykov I include that test into functional in this PR, hoping the fix would fix all issues in request forwarding. at least in that test's scenario. But I guess it didn't? Should I exclude it from functionals? It wasn't part of functionals because it failes frequently.

sergeybykov:
Let's keep the test out for now, until we improve the shutdown process to be more deterministic.

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -501,14 +480,16 @@ public SiloAddress CalculateTargetSilo(GrainId grainId, bool excludeThisSiloIfSt
             int hash = unchecked((int)grainId.GetUniformHashCode());
 
             // excludeMySelf from being a TargetSilo if we're not running and the excludeThisSIloIfStopping flag is true. see the comment in the Stop method.
-            bool excludeMySelf = !Running && excludeThisSiloIfStopping;
+            // excludeThisSIloIfStopping flag was removed because we believe that flag complicates things unnecessarily. We can add it back if it turns out that flag 
+            // is doing something valuable.
```
ReubenBond: Not sure we need a comment about dead code - the comment can be added to the PR comments instead

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -792,10 +773,18 @@ public bool LocalLookup(GrainId grain, out AddressesAndTag result)
...
+
 
-            if (log.IsEnabled(LogLevel.Debug)) log.Debug("Silo {0} tries to lookup for {1}-->{2} ({3}-->{4})", MyAddress, grain, silo, grain.GetUniformHashCode(), silo.GetConsistentHashCode());
+            if (log.IsEnabled(LogLevel.Debug)) log.Debug("Silo {0} tries to lookup for {1}-->{2} ({3}-->{4})", MyAddress, grain, silo, grain.GetUniformHashCode(), silo?.GetConsistentHashCode());
+
+            //this will only happen if I'm the only silo in the cluster and I'm shutting down
+            if (silo == null)
+            {
+                if (log.IsEnabled(LogLevel.Trace)) log.Trace("LocalLookup mine {0}=null", grain);
+                result = new AddressesAndTag();
```
ReubenBond: Should this be null instead?

src/Orleans.Runtime/MembershipService/MembershipOracle.cs
```diff
@@ -142,7 +144,7 @@ public async Task BecomeActive()
                 MembershipTableData table = await membershipTableProvider.ReadAll();
                 await ProcessTableUpdate(table, "BecomeActive", true);
                     
-                GossipMyStatus(); // only now read and stored the table locally.
+                GossipMyStatus().Ignore(); // only now read and stored the table locally.
```
ReubenBond: Unsure of the purpose of this. Are we gossiping our status twice here, once in UpdateMyStatusGlobal and again immediately after?

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
xiazen: I think it is useful to add something here in case this flag is useful somehow. Add in the pr would juts got lost after it merged

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
xiazen: its caller expects it not be null. And also see line 808 and line 811. it is doing the same thing. I was copying logic from there.

src/Orleans.Runtime/MembershipService/MembershipOracle.cs
xiazen: not sure why. But it was there before. If we conclude this line is useless and want to remove it, we can do it in another PR IMO. But the comment besides it indicates it might be here for a reason.

ReubenBond:
@dotnet-bot test functional

Looks like a docker index issue?

ReubenBond:
@dotnet-bot test Ubuntu16.04 Release Build

PR #5296, opened 2019-01-05
Use AzureClustering to make SiloGracefulShutdown_ForwardPendingRequest pass more consistently

xiazen:


ReubenBond:
Looks like there's an issue with the connection string

PR #5313, opened 2019-01-11
Enable EventHub streaming tests

xiazen:
Enable EventHub streaming tests 
- will be skipped on check in 
- will be in run in vso

vso event hub connection string  updated to the new one.


xiazen:
poke

PR #5319, opened 2019-01-18
Fix defensive check in LogConsistentGrain

emiaj:
Found this low hanging fruit around the Orleans.Core project.

The original intent of this code is to have two possible error messages:

1.  `$"Cannot find consistency provider with Name={attr.ProviderName} for grain type {this.GetType().FullName}"`
2.  `$"No consistency provider manager found loading grain type {this.GetType().FullName}";`

However, the wrapping `if`statement above that says:

`if (attr != null && defaultFactory == null)`

...makes the expression below to always fall into the first error message because it evaluates always to `true`, thus, making impossible for the error message to be the second one.

Furthermore, there is another safety check below that ultimately ensures `defaultFactory` is not null and throws an exception if so.

### DIFF
--- a/src/Orleans.Core/LogConsistency/LogConsistentGrain.cs
+++ b/src/Orleans.Core/LogConsistency/LogConsistentGrain.cs
@@ -107,7 +107,7 @@ namespace Orleans.LogConsistency
                 : this.ServiceProvider.GetService<ILogViewAdaptorFactory>();
             if (attr != null && defaultFactory == null)
             {
-                var errMsg = $"Cannot find consistency provider with Name={attr.ProviderName} for grain type {this.GetType().FullName}"
+                var errMsg = $"Cannot find consistency provider with Name={attr.ProviderName} for grain type {this.GetType().FullName}";
                 throw new BadGrainStorageConfigException(errMsg);
             }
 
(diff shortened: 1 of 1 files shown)


sergeybykov:
It seems to me the right fix here is to change the condition to 
```csharp
 if (defaultFactory == null)
```
and leave both of the error messages for the two possible cases.

emiaj:
@sergeybykov I thought initially the same.
But this part specifically deals with the case `A provider name has been specified explicitly using an attribute but it couldn't be found`.

Because if you haven't decorated the inheritor with this attribute (`LogConsistencyProviderAttribute`), then there's more code below that attempts to initialize this field using a default provider:

```csharp
            // use default if none found
            defaultFactory = defaultFactory ?? this.DefaultAdaptorFactory;
            if (defaultFactory == null)
            {
                var errMsg = $"No log consistency provider found loading grain type {this.GetType().FullName}";
                throw new BadGrainStorageConfigException(errMsg);
            };
```

sergeybykov:
Yes, I think you are right.

sergeybykov:
@dotnet-bot test functional

sergeybykov:
Thank you, @emiaj!

emiaj:
Wow thanks @sergeybykov 
Such a minor thing but very proud of myself at the moment LOL

PR #5336, opened 2019-01-25
Batching batch containers pulling agent retrieves from cache

cunqueiromarco:
- This change attempts to add a layer of batching in the PersistentStreamPullingAgent. If configured, the 
   PersistentStreamPullingAgent will accumulate several BatchContainers received from a cache and place 
   them in an instance of BatchContainerBatch before delivering.

I'll add tests and fix the white spaces changes that were somehow included in the pull request. Mostly creating this pull request for visibility.


src/Orleans.Core/Streams/QueueAdapters/BatchContainerBatch.cs
```diff
@@ -0,0 +1,67 @@
...
+        public string StreamNamespace { get; }
+
+        /// <summary>
+        /// Batch containers comprising this batch
+        /// </summary>
+        public IEnumerable<IBatchContainer> BatchContainers { get; }
+
+        public BatchContainerBatch(IEnumerable<IBatchContainer> batchContainers)
+        {
+            if ((batchContainers == null) || !batchContainers.Any())
```
jason-bragg: Using IEnumerable<T> in this way can allow for multiple enumerations.  Suggest either converting it to a List<T> or using List<T> as input arg

src/Orleans.Core/Streams/QueueAdapters/BatchContainerBatch.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public IEnumerable<Tuple<T, StreamSequenceToken>> GetEvents<T>()
+        {
+            return this.BatchContainers.SelectMany(batchContainer => batchContainer.GetEvents<T>());
+        }
+
+        public bool ImportRequestContext()
+        {
+            return true;
```
jason-bragg: Need to set call import context on each batch prior to returning it's events in GetEvents.  Suggest returning events via yield loop in GetEvents to support this.

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -604,6 +604,48 @@ private async Task RunConsumerCursor(StreamConsumerData consumerData, IStreamFil
             }
         }
 
+        private IBatchContainer GetBatchForConsumer(IQueueCacheCursor cursor)
+        {
+            if (this.options.BatchContainerBatchSize == 1)
```
jason-bragg: <= 1, in case of misconfiguration?

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -604,6 +604,48 @@ private async Task RunConsumerCursor(StreamConsumerData consumerData, IStreamFil
...
+                if (i == 0)
+                {
+                    return null;
+                }
+
+                return new BatchContainerBatch(batchContainers);
+            }
+            else
+            {
+                throw new OrleansConfigurationException("StreamPullingAgentOptions.BatchContainerBatchSize is negative");
```
jason-bragg: Suggest validation code a configuration time over failure at runtime.

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -604,6 +604,48 @@ private async Task RunConsumerCursor(StreamConsumerData consumerData, IStreamFil
...
+                var batchContainers = new List<IBatchContainer>();
+
+                for (i = 0; i < this.options.BatchContainerBatchSize; i++)
+                {
+                    if (!cursor.MoveNext())
+                    {
+                        break;
+                    }
+
+                    batchContainers.Add(cursor.GetCurrent(out ignore));
```
jason-bragg: Where is back filtering logic applied?

src/Orleans.Core/Streams/QueueAdapters/IBatchContainerBatch.cs
```diff
@@ -0,0 +1,15 @@
+using System.Collections.Generic;
+
+namespace Orleans.Streams
+{
+    /// <summary>
+    /// A batch of queue messages (see IBatchContainer for description of batch contents)
+    /// </summary>
+    public interface IBatchContainerBatch : IBatchContainer
```
jason-bragg: Who uses this interface?

cunqueiromarco: BatchContainerBatch implements the interface.

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
cunqueiromarco: Is this supposed to read "Suggest validation code at configuration time over failure at runtime"?

src/Orleans.Core/Streams/QueueAdapters/BatchContainerBatch.cs
cunqueiromarco: Are you saying that within GetEvents itself I should call ImportRequestContext on each batch and then yield that batch?

jason-bragg: Yes, or something of the like.  The expected behavior is that when events from a batch are passed to the application, they are done so within the calling context of the batch they are in.  This may not be possible for batched observers.  I'll speak to the team about that behavior.

src/Orleans.Core/Streams/QueueAdapters/IBatchContainerBatch.cs
jason-bragg: Yes, but what calls the interface.  If nothing calls it, then it is unnecessary as an abstraction, correct?

cunqueiromarco: It was previously used for logic in code that was going to send batches to the application layer. But since I'm reverting that change, I'll remove this interface.

src/Orleans.Core/Streams/QueueAdapters/BatchContainerBatch.cs
cunqueiromarco: Done. Added import context in StreamSubscriptionHandleImpl when iterating over the IBatchContainers which compose the IBatchContainerBatch.

src/Orleans.Core/Streams/Internal/StreamSubscriptionHandleImpl.cs
```diff
@@ -112,6 +120,23 @@ public async Task<StreamHandshakeToken> DeliverItem(object item, StreamSequenceT
...
 
+        public async Task NextBatch(IBatchContainerBatch batchContainerBatch)
+        {
+            bool isRequestContextSet;
+            foreach (var batchContainer in batchContainerBatch.BatchContainers)
+            {
+                isRequestContextSet = batchContainer.ImportRequestContext();
+                foreach (var itemTuple in batchContainer.GetEvents<T>())
+                {
+                    await NextItem(itemTuple.Item1, itemTuple.Item2);
```
xiazen: I don't quite understand this. Why don't we deliver the whole batch by calling `OnNextBatchAsync`, but still deliver the batch item by item?

cunqueiromarco: Correct me if I'm wrong, but `OnNextBatchAsync` is actually not supported. If it was supported, then yes, I would be calling `OnNextBatchAsync` for each `IBatchContainer`.

src/Orleans.Core/Streams/QueueAdapters/BatchContainerBatch.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Orleans.Streams
+{
+    /// <summary>
+    /// A batch of batch containers, that if configured (see StreamPullingAgentOptions), will be the data pulled by the
+    /// PersistenStreamPullingAgent from it's underlying cache
+    /// </summary>
+    class BatchContainerBatch : IBatchContainerBatch
```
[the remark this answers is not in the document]
xiazen: nvm, ignore me, I think I understand the use case now.

src/Orleans.Core/Streams/Internal/StreamSubscriptionHandleImpl.cs
xiazen: I see. I was thinking it might be because of that. But I was confused because I heard you guys are interested in implementing `OnNextBatchAsync` as well, and I'm not sure whether that is implemented in this pr or the next one. 

thanks for explaining.

cunqueiromarco:
@dotnet-bot test functional

cunqueiromarco:
@dotnet-bot test functional

cunqueiromarco:
@dotnet-bot test this please

cunqueiromarco:
All tests now pass.

PR #5344, opened 2019-01-31
Always Interleave modified to also be interleavable.

jason-bragg:


src/Orleans.Runtime/Catalog/ActivationData.cs
```diff
@@ -456,7 +456,7 @@ public void SetCollectionTicket(DateTime ticket)
         /// </summary>
         internal bool IsUsingGrainDirectory => this.PlacedUsing.IsUsingGrainDirectory;
 
-        public Message Running { get; private set; }
+        public Message Blocking { get; private set; }
```
ReubenBond: I describe this to myself as "the message which is currently blocking non-interleaving messages from execution"

ReubenBond: This probably warrants a comment

PR #5350, opened 2019-02-05
Migrate blog content from shut down MSDN blogs to docs

sergeybykov:
We can later mirror the content on some blogging engine if we want. For now, at least it'll be available as part of our docs.

You can see a preview on http://sergeybykov.github.io/orleans/Blog/index.html.


ReubenBond:
I would change Blog to blog in the URL slug, but otherwise LGTM. I think we can do that afterwards with no ill effect.

PR #5353, opened 2019-02-08
Skip intermitent failing test

xiazen:
There's couple streaming test failing intermittently in vso. Since I'm still investigating them and don't know why it is happening, so skip them for now.

### DIFF
--- a/test/Extensions/ServiceBus.Tests/Streaming/EHStreamProviderCheckpointTests.cs
+++ b/test/Extensions/ServiceBus.Tests/Streaming/EHStreamProviderCheckpointTests.cs
@@ -84,14 +84,14 @@ namespace ServiceBus.Tests.StreamingTests
             }
         }
 
-        [SkippableFact(Skip="Investigating intermitent failures")]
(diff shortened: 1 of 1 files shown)


test/Extensions/ServiceBus.Tests/Streaming/EHStreamProviderCheckpointTests.cs
```diff
@@ -84,14 +84,14 @@ public void Configure(IConfiguration configuration, IClientBuilder clientBuilder
             }
         }
 
-        [SkippableFact]
+        [SkippableFact(Skip="Investigating intermitent failures")]
```
sergeybykov: I think a better pattern is to open an issue with names of the intermittently failing tests listed and put the URL to that issue as the description for `Skip`. For example, like in https://github.com/dotnet/orleans/blob/master/test/TesterInternal/GeoClusterTests/MultiClusterNetworkTests.cs#L149

xiazen: will do

PR #5354, opened 2019-02-08
Add handling when pulling agent fails RegisterAsProducer to pubsub

xiazen:


src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -453,7 +453,17 @@ private async Task<bool> ReadFromQueue(QueueId myQueueId, IQueueAdapterReceiver
...
                     streamData.RefreshActivity(now);
-                    StartInactiveCursors(streamData, startToken); // if this is an existing stream, start any inactive cursors
+                    if (streamData.StreamRegistered)
+                    {
+                        StartInactiveCursors(streamData,
+                            startToken); // if this is an existing stream, start any inactive cursors
+                    }
+                    else
+                    {
+                        this.logger.LogInformation($"Pulled new messages in stream {streamId} from the queue, but pulling agent haven't succeeded in" +
```
sergeybykov: Should this rather be a warning?

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -762,7 +773,24 @@ private async Task RegisterAsStreamProducer(StreamId streamId, StreamSequenceTok
                 if (pubSub == null) throw new NullReferenceException("Found pubSub reference not set up correctly in RetreaveNewStream");
 
                 IStreamProducerExtension meAsStreamProducer = this.AsReference<IStreamProducerExtension>();
-                ISet<PubSubSubscriptionState> streamData = await pubSub.RegisterProducer(streamId, streamProviderName, meAsStreamProducer);
+                ISet<PubSubSubscriptionState> streamData = null;
+                var callPubsubSucceed = false;
+                while (!callPubsubSucceed)
+                {
+                    var retryInterval = TimeSpan.FromSeconds(1);
```
sergeybykov: Should this be configurable?

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
xiazen: yeah it could be. if code hit this, then there might be message delay, but not message loss. So warning should be proper.

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
xiazen: I'm OK with both. we can add a setting to PullingAgentOptions.

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -453,7 +453,17 @@ private async Task<bool> ReadFromQueue(QueueId myQueueId, IQueueAdapterReceiver
...
-                    StartInactiveCursors(streamData, startToken); // if this is an existing stream, start any inactive cursors
+                    if (streamData.StreamRegistered)
+                    {
+                        StartInactiveCursors(streamData,
+                            startToken); // if this is an existing stream, start any inactive cursors
+                    }
+                    else
+                    {
+                        this.logger.LogWarning($"Pulled new messages in stream {streamId} from the queue, but pulling agent haven't succeeded in" +
+                                                   $"RegisterStream yet, will start deliver on this stream after RegisterStream succeeded");
```
jason-bragg: This warning could spam a lot..
Maybe instead add a check in CleanupPubSubCache to log any stream that is still unregistered?
Other options?

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -762,7 +773,23 @@ private async Task RegisterAsStreamProducer(StreamId streamId, StreamSequenceTok
...
+                    try
+                    {
+                        streamData = await pubSub.RegisterProducer(streamId, streamProviderName, meAsStreamProducer);
+                        callPubsubSucceed = true;
+                        break;
+                    }
+                    catch (Exception e)
+                    {
+                        logger.Error(ErrorCode.PersistentStreamPullingAgent_17, $"RegisterAsStreamProducer failed due to {e}, retrying after {retryInterval}", e);
+                        await Task.Delay(this.options.RetryPubsubInterval);
```
jason-bragg: Suggest using retry policy, with exponential backoff, as we do with other actions we need to retry.

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
```diff
@@ -762,7 +773,23 @@ private async Task RegisterAsStreamProducer(StreamId streamId, StreamSequenceTok
                 if (pubSub == null) throw new NullReferenceException("Found pubSub reference not set up correctly in RetreaveNewStream");
 
                 IStreamProducerExtension meAsStreamProducer = this.AsReference<IStreamProducerExtension>();
-                ISet<PubSubSubscriptionState> streamData = await pubSub.RegisterProducer(streamId, streamProviderName, meAsStreamProducer);
+                ISet<PubSubSubscriptionState> streamData = null;
+                var callPubsubSucceed = false;
+                while (!callPubsubSucceed)
```
jason-bragg: need to break out if IsShutdown

jason-bragg:
While reviewing this, I think I've found another issue in the pulling agent (not introduced by this change).

The registration with the pubsub does two things; it gets a set of subscriptions on a stream, and notifies the pubsup to tell us about any subscription changes on that stream.  If a stream is inactive long enough, it's data will be removed from the agent's cache, but the agent will still be registered with the pubsub and get notified of subscription changes.  If the pubsub notifies the agent of a new subscription, we put the stream data back into the cache, but we don't get the full set of subscription data from the pubsub again, so if events do show up, they'll only be sent to the added subscriptions, not the one's we removed from the cache.

I raise this issue here as we may want to address it as part of this fix.  We can fix it in a separate pass if preferred, but wanted to bring it up for consideration.

xiazen:
@jason-bragg not sure if that is an issue. Yes the pulling agent will have a partial list of subscribers when the stream is inactive. But the moment it gets message from the queue, it will call pubsub again (since this stream isn't in pubsub cache), and get the full list of subscribers. Is this right?

jason-bragg:
@xiazen,
> the moment it gets message from the queue, it will call pubsub again (since this stream isn't in pubsub cache), and get the full list of subscribers. Is this right?

In most cases, yes, this is correct.  The problem is related to a specific edge case.  Consider this scenario:

1) Events for a stream are read from the queue and the agent registers as a producer of that stream with pubsub.
2) Stream becomes inactive long enough for subscription information to be removed from the subscription cache.
3) A new consumer subscribes to the stream, causing the pubsub system to notify the agent of the news subscription.  In this case the agent will add the stream data for the new subscription to the subscription cache.
4) Stream becomes active again, but agent doesn't call to pubsub to get full list of subscriptions because it already has subscription data for that stream, added when the pubsub system notified the agent of the new subscription.

The result is that the events will only be delivered to the newly added subscription, not the full list of subscriptions in the pubsub, because the agent never refreshed its list of subscriptions cleared due to inactivity.

xiazen:
@jason-bragg yeah..... I think you are right.... this is bad. 

I want to suggest doing some special handing when stream is inactive, and a new consumer just subscribed, such as calling pubsub to get the full list of consumers. But I feel like maybe a better solution is to not clear the cache. It is unclear to me how much perf gain does the cache give us. And if we don't clear it, then it is less error prone.

src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingAgent.cs
xiazen: hmm maybe we log it as Debug? @sergeybykov what's your opinion? 

> Maybe instead add a check in CleanupPubSubCache to log any stream that is still unregistered?

not sure why this is related. Can you explain more?

jason-bragg: Yeah, probably doesn't belong in CleanupPubSubCache.  The reason I suggested that is because that check is done less often, so we'd not span the logs.. but conceptually, it makes no sense :/

PR #5373, opened 2019-02-16
Persistent state facet

jason-bragg:
This change Introduces a storage facet IPersistentState<T> which allows grains to define state which can be stored in using the existing IGrainStorage implementations.

### DIFF
--- a/src/Orleans.Runtime.Abstractions/Facet/Persistent/IPersistentState.cs
+++ b/src/Orleans.Runtime.Abstractions/Facet/Persistent/IPersistentState.cs
@@ -1,0 +1,9 @@
+using Orleans.Core;
+
+namespace Orleans.Runtime
+{
+    public interface IPersistentState<TState> : IStorage<TState>
+        where TState : new()
+    {
+    }
+}
--- a/src/Orleans.Runtime/Facet/Persistent/PersistentStateStorageFactory.cs
+++ b/src/Orleans.Runtime/Facet/Persistent/PersistentStateStorageFactory.cs
@@ -1,10 +1,10 @@
-using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Orleans.Core;
 using Orleans.Storage;
+using Orleans.Utilities;
 
 namespace Orleans.Runtime
 {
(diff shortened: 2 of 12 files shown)


src/Orleans.Runtime/Facet/Persistent/PersistentStateStorageFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+            string fullStateName = GetFullStateName(context, cfg);
+            var bridge = new PersistentStateBridge<TState>(fullStateName, context, storageProvider);
+            bridge.Participate(context.ObservableLifecycle);
+            return bridge;
+        }
+
+        protected virtual string GetFullStateName(IGrainActivationContext context, IPersistentStateConfiguration cfg)
+        {
+            return $"{context.GrainType.FullName}.{cfg.StateName}";
```
ReubenBond: We should use `RuntimeTypeNameFormatter.Format(x)` instead of `FullName`

src/Orleans.Runtime.Abstractions/Facet/Persistent/IPersistentState.cs
```diff
@@ -0,0 +1,9 @@
+using Orleans.Core;
+
+namespace Orleans.Runtime
+{
+    public interface IPersistentState<TState> : IStorage<TState>
+        where TState : new()
```
ReubenBond: Minor: we could drop the `new()` constraint on `TState` by using `ActivatorUtilities`. I don't believe there are any major downsides. Maybe there are reasons that a user would want to inject services into their state, maybe not.

ReubenBond:
Love it :) I added one comment but it's not worth holding the train for it.

ReubenBond:
One more thing: it would be nice if users had a way to migrate from `Grain<T>` to facets. It's not a must, though: this can be used for new cases.

ReubenBond:
I didn't notice earlier, but the attribute arguments seem backwards to me, particularly since the state name feels more optional than the provider name.

Generally I feel arguments should be ordered in terms of most significant to least significant (most discriminatory to least). Storage provider seems more significant than state name to me. Do other's have opinions? cc @xiazen @benjaminpetit @sergeybykov 

Current example looks like this:
``` C#
public GrainStorageTestGrain(
    [PersistentState("state", "GrainStorageForTest")]
    IPersistentState<PersistenceTestGrainState> persistentState)
```

jason-bragg:
@ReubenBond, 
At the moment the state name is required, while the storage provider name is not.  This is why the state name is first.  If no storage provider is provided, the system will use the default.

The state name is required because using this facet allows for multiple states to be associated with the grain, unlike grain<T>, and each needs be uniquely identified.

In previous incarnations, the state name was also optional, and if not specified the argument name ("persistentState" in this case) would be used, but I moved away from this because I felt it was better to be explicit.

ReubenBond:
In this PR, state name is set to `parameter.Name` if not otherwise specified.

veikkoeeva:
@jason-bragg Do you have link to documentation by any chance? :)

Issue #5377, opened 2019-02-19
Sample GPSTracker is broken


kimsama:
Hey, 

The sample, GPSTracker is broken. Running it got the following error in a browser:

```
(index):58 Uncaught TypeError: Microsoft.Maps.Location is not a constructor
    at HTMLDocument.<anonymous> ((index):58)
    at c (jquery-1.10.2.min.js:4)
    at Object.fireWith [as resolveWith] (jquery-1.10.2.min.js:4)
    at Function.ready (jquery-1.10.2.min.js:4)
    at HTMLDocument.q (jquery-1.10.2.min.js:4)
(anonymous) @ (index):58
c @ jquery-1.10.2.min.js:4
fireWith @ jquery-1.10.2.min.js:4
ready @ jquery-1.10.2.min.js:4
q @ jquery-1.10.2.min.js:4
:48777/favicon.ico:1 Failed to load resource: the server responded with a status of 404 (Not Found)
(index):53 listening for messages
```
Error is:
> Uncaught TypeError: Microsoft.Maps.Location is not a constructor.

The reason for that error is the code doesn't access the Microsoft.Maps namespace outside of a function and that the function isn't triggered before the Microsoft.Maps namespace does, otherwise, it won't be available. It is recommended using the callback parameter of the map script URL to call a map load function. This will ensure that the map script is fully loaded before the sample's code tries to access it.

Please, see  [_Customize Map Options on Load_](https://docs.microsoft.com/en-us/bingmaps/v8-web-control/map-control-concepts/map/customize-map-options-on-load
) page in bing map site :

ReubenBond:
Fixed in #5378

Thanks for reporting & fixing @kimsama! Please know that you can always find us in the Gitter channel here: https://gitter.im/dotnet/orleans

PR #5378, opened 2019-02-19
resolved the issue #5377

kimsama:
- Changed to wait until the map script is loaded.  Microsoft.Maps namespace should be used after the map control script is loaded. 
- Changed the code to make V8 map script loads asynchronously.


ReubenBond:
Thank you, @kimsama & welcome!

PR #5379, opened 2019-02-19
ConvertJsonToFromStorageFormatWithCustomJsonProperties test fix

jason-bragg:
- made is skippable

### DIFF
--- a/test/Extensions/TesterAzureUtils/Persistence/PersistenceProviderTests.cs
+++ b/test/Extensions/TesterAzureUtils/Persistence/PersistenceProviderTests.cs
@@ -202,7 +202,7 @@ namespace Tester.AzureUtils.Persistence
             Assert.Equal(initialState.C, convertedState.C);
         }
 
-        [Fact, TestCategory("Functional"), TestCategory("Azure")]
+        [SkippableFact, TestCategory("Functional"), TestCategory("Azure")]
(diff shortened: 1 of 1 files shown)


ReubenBond:
It's possible we need to move this test to `test\Extensions\TesterAzureUtils\Persistence\PersistenceGrainTests_AzureBlobStore_Json.cs` - I don't see anything which would call `CheckPreconditionsOrThrow` in the current class. Sorry

test/Extensions/TesterAzureUtils/Persistence/PersistenceProviderTests.cs
```diff
@@ -202,7 +202,7 @@ public async Task AzureTableStorage_ConvertToFromStorageFormat(int? stringLength
             Assert.Equal(initialState.C, convertedState.C);
         }
 
-        [Fact, TestCategory("Functional"), TestCategory("Azure")]
+        [SkippableFact, TestCategory("Functional"), TestCategory("Azure")]
```
xiazen: do you need to do `TestUtils.CheckForAzureStorage();` to throw the SkipException which would make this test skipped when condition not applied?

xiazen: what's the skipping condition?  Mark it as `SkippableFact` wouldn't necessary skip it.

PR #5381, opened 2019-02-20
Remove new() requirement from persistent state facet and Grain<T>

jason-bragg:
Note: This change introduced interface changes in core.abstractions.  We may not want to do this at this time.


ReubenBond:
We agreed that we can close this and consider the change in 3.0. It's a breaking change.

PR #5384, opened 2019-02-21
Upgade to eventhub 2.2.1

jason-bragg:


src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubConstants.cs
```diff
@@ -2,7 +2,6 @@ namespace Orleans.ServiceBus.Providers
 {
     internal class EventHubConstants
     {
-        public readonly static string StartOfStream =
-            Microsoft.Azure.EventHubs.PartitionReceiver.StartOfStream;
```
xiazen: why this change? there isn't something like `Microsoft.Azure.EventHubs.PartitionReceiver.StartOfStream` in new pkg anymore?

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
```diff
@@ -96,7 +96,7 @@ public class EventHubAdapterFactory : IQueueAdapterFactory, IQueueAdapter, IQueu
         /// <summary>
         /// Factory to create a IEventHubReceiver
         /// </summary>
-        protected Func<EventHubPartitionSettings, string, ILogger, ITelemetryProducer, Task<IEventHubReceiver>> EventHubReceiverFactory;
```
xiazen: yeah I can see using API in the new evenhub pkg, a async Task isn't needed anymore. But this is a breaking change indeed. I don't have strong opinion on this, since most people won't be affected. Just want to point that out.

src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubConstants.cs
jason-bragg: No, they don't use '-1' as a signal offset for the start of the partition, but we still need it to not break backwards compatibility with checkpoint entities that may have been written with it.

PR #5386, opened 2019-02-22
Prereqs information for `LoadSheddingOptions`

Kritner:
I'm not sure if this really belongs here, but didn't see much otherwise regarding `LoadShedding` in the docs.  `LoadShedding` appears to only be functional once a `IHostEnvironmentStatistics` implementation has been registered - calling that out in the "Used for" column.


sergeybykov:
Thank you, @Kritner!

PR #5389, opened 2019-02-22
Enable "cleaning" of all dead entries in the membership table

benjaminpetit:
Right now, we don't delete old entries in the membership table, and that can be an issue since it will grow forever. 

This PR enable deleting old silo entries, and implement it in the azure membership provider. Other providers still need to implement this logic.

### DIFF
--- a/src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
+++ b/src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
@@ -196,7 +196,7 @@ namespace Orleans.AzureUtils
             return entriesList.Count();
         }
 
-        public async Task DeleteDeadMembershipTableEntries(DateTime beforeDate)
+        public async Task DeleteDeadMembershipTableEntries(DateTimeOffset beforeDate)
         {
             var entriesList = (await FindAllSiloEntries())
                 .Where(entry => entry.Item1.Status == INSTANCE_STATUS_DEAD && entry.Item1.Timestamp < beforeDate)
--- a/src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
+++ b/src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
@@ -190,19 +190,36 @@ namespace Orleans.AzureUtils
 
             var entries = await storage.ReadAllTableEntriesForPartitionAsync(clusterId);
             var entriesList = new List<Tuple<SiloInstanceTableEntry, string>>(entries);
+
+            await DeleteEntriesBatch(entriesList);
+
+            return entriesList.Count();
+        }
+
+        public async Task DeleteDeadMembershipTableEntries(DateTime beforeDate)
+        {
+            var entriesList = (await FindAllSiloEntries())
+                .Where(entry => entry.Item1.Status == INSTANCE_STATUS_DEAD && entry.Item1.Timestamp < beforeDate)
+                .ToList();
+
+            await DeleteEntriesBatch(entriesList);
+        }
+
+        private async Task DeleteEntriesBatch(List<Tuple<SiloInstanceTableEntry, string>> entriesList)
+        {
             if (entriesList.Count <= AzureTableDefaultPolicies.MAX_BULK_UPDATE_ROWS)
             {
                 await storage.DeleteTableEntriesAsync(entriesList);
-            }else
+            }
+            else
             {
-                List<Task> tasks = new List<Task>();
+                var tasks = new List<Task>();
                 foreach (var batch in entriesList.BatchIEnumerable(AzureTableDefaultPolicies.MAX_BULK_UPDATE_ROWS))
                 {
                     tasks.Add(storage.DeleteTableEntriesAsync(batch));
                 }
                 await Task.WhenAll(tasks);
             }
-            return entriesList.Count();
         }
 
         internal async Task<List<Tuple<SiloInstanceTableEntry, string>>> FindSiloEntryAndTableVersionRow(SiloAddress siloAddress)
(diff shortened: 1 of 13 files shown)


src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
```diff
@@ -190,19 +190,36 @@ internal async Task<int> DeleteTableEntries(string clusterId)
 
             var entries = await storage.ReadAllTableEntriesForPartitionAsync(clusterId);
             var entriesList = new List<Tuple<SiloInstanceTableEntry, string>>(entries);
+
+            await DeleteEntriesBatch(entriesList);
+
+            return entriesList.Count();
+        }
+
+        public async Task DeleteDeadMembershipTableEntries(DateTime beforeDate)
```
ReubenBond: This should be a `DateTimeOffset`

src/Orleans.Runtime/MembershipService/MembershipOracle.cs
```diff
@@ -892,6 +924,30 @@ private void OnIAmAliveUpdateInTableTimer(object data)
...
+                this.membershipTableProvider
+                        .DeleteDeadMembershipTableEntries(dateLimit)
+                        .ContinueWith(task =>
+                        {
+                            if (task.IsFaulted)
+                                this.logger.Error(ErrorCode.MembershipCleanDeadEntriesFailure, "CleanupEntries failed", task.Exception);
+                            return true;
+                        }).Ignore();
+            }
+            catch (MissingMethodException)
```
ReubenBond: Do we need to also catch `NotImplementedException`?

src/Orleans.Core/Configuration/Options/ClusterMembershipOptions.cs
```diff
@@ -91,6 +91,19 @@ public class ClusterMembershipOptions
...
+        /// </summary>
+        public TimeSpan DeleteEntriesOlderThan { get; set; } = DEFAULT_DELETE_ENTRIES_OLDER_THAN;
+        public static readonly TimeSpan DEFAULT_DELETE_ENTRIES_OLDER_THAN = TimeSpan.FromDays(7);
+
+        /// <summary>
+        /// If set, will check if there are older entries than <see cref="DeleteEntriesOlderThan"/> in the membership table,
+        /// and will delete them
+        /// </summary>
+        public TimeSpan? CleanupDeadEntriesTimeout { get; set; } = DEFAULT_DELETE_ENTRIES_OLDER_THAN;
+        public static readonly TimeSpan? DEFAULT_CLEANUP_DEAD_ENTRIES_TIMEOUT = null;
```
ReubenBond: This was confusing to me before reading the code. I think we should rename it to something like `DefunctSiloCleanupPeriod` and change the comment to:

```xml
/// <summary>
/// The duration between membership table cleanup operations. When this period elapses, all defunct silo
// entries older than <see cref="DefunctSiloExpiration" /> are removed. This value is per-silo.
/// </summary>
```

Thoughts? I'm not set on those names, but 'timeout' to me says that the process will be cancelled after that amount of time.

benjaminpetit: Yes I like your names better

src/Orleans.Runtime/MembershipService/MembershipOracle.cs
benjaminpetit: I thought that they were caught in the `Continue` but it seems you are right

ReubenBond: Maybe we should use an async method instead of `.ContinueWith().Ignore()`, so the error handling is flat

ReubenBond:
many test failures, unsure why

sergeybykov:
@dotnet-bot test bvt
@dotnet-bot test functional

ReubenBond:
Made some minor changes. LGTM & will squash merge once tests complete again.

ReubenBond:
@dotnet-bot test bvt

Doesn't seem to be running

ReubenBond:
@dotnet-bot test functional

ReubenBond:
@dotnet-bot test functional

jinhong-:
I noticed this is listed as a non-breaking change. If i have a custom membership provider that did not upgrade, wouldn't that break when i upgrade to Orleans 2.3? What would be then considered as breaking change?

benjaminpetit:
@jinhong- If you are compiling your custom membership provider at the same time as you are compiling your silos then you will have to add this method to your interface to compile properly. You can throw `NotImplementedException` if you don't want to implement the logic.

If you are referencing via NuGet for example, then it should be fine since it will be catched by the runtime.

jinhong-:
@benjaminpetit  Okay. So in theory i could use a Membership Provider installed via NuGet package compiled with Orleans 2.2 in the new Orleans 2.3?

On another note, i notice this line seemed a little strange to me.
https://github.com/benjaminpetit/orleans/blob/5b97f1c0d53e3de10dda7530f214033009acb155/src/Orleans.Runtime/MembershipService/MembershipOracle.cs#L942
`catch (Exception ex) when (ex is NotImplementedException || ex is NotImplementedException)`
Why would you need a or condition that checks for the same exception twice?

benjaminpetit:
@jinhong- you are right, the condition should be

`catch (Exception ex) when (ex is NotImplementedException || ex is MissingMethodException)`

benjaminpetit:
> Okay. So in theory i could use a Membership Provider installed via NuGet package compiled with Orleans 2.2 in the new Orleans 2.3?

Yes it should work

RainingNight:
Why Consul NotImplementedException?

```csharp
// src/Orleans.Clustering.Consul/ConsulBasedMembershipTable.cs
public Task CleanupDefunctSiloEntries(DateTimeOffset beforeDate)
{
     throw new NotImplementedException();
}
```

benjaminpetit:
Because we didn't implemented the logic for Consul yet. But feel free to open a PR with the implementation if you can, I would be happy to review it and merge it!

PR #5395, opened 2019-02-25
Enable HostedClient by default

### LINKED ISSUE #5337: Enable hosted client by default

ReubenBond:
Fixes #5337


ReubenBond:
@dotnet-bot test this please

I think the CI servers rebooted?

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -274,8 +274,21 @@ internal static void AddDefaultServices(IApplicationPartManager applicationPartM
             services.AddTransient<IConfigurationValidator, ClusterOptionsValidator>();
             services.AddTransient<IConfigurationValidator, SiloClusteringValidator>();
 
-            // Disable hosted client by default.
-            services.TryAddSingleton<IHostedClient, DisabledHostedClient>();
+            // Enable hosted client.
+            services.AddSingleton<HostedClient>();
+            services.AddFromExisting<IHostedClient, HostedClient>();
```
jason-bragg: So users can no longer provider their own impl of IHostedClient as default?  I'm cool with that, just want to make sure we're aware of this change.  Users can still register IHostedClient implementaions by name as a named service if they need to use a custom impl or connect to multiple clusters (I think..?)

ReubenBond: `IHostedClient` is intentionally internal. It doesn't even need an interface anymore since it's always enabled and we only have one valid implementation of it. There's no pressing reason to remove the interface yet, though.

To connect to a remote cluster, the external client would currently have to be used.

ReubenBond:
@dotnet-bot test this

ReubenBond:
I think the CI queue is stuck

ReubenBond:
@dotnet-bot test this please

PR #5400, opened 2019-02-26
Adds a LoadSheddingValidator class

### LINKED ISSUE #4389: No implementation of IHostEnvironmentStatistics was found
Warning	100708	Orleans.Runtime.NoOpHostEnvironmentStatistics]	No implementation of IHostEnvironmentStatistics was found. Load shedding will not work yet.

What should I write in StartSilo

Kritner:
- Adds as a default validator within `DefaultSiloServices`
- Should only validate `LoadSheddingOptions` when not null
- Checks LoadSheddingLimit not more than 100% CPU
- Ensures a non default and non null implementation of `IHostEnvironmentStatistics` has been registered.
- Fixes #4389

I did not see (at a glance) any test classes for validator classes... this seems like it should work... maybe?


ReubenBond:
@dotnet-bot test functional

benjaminpetit:
Did you at least tested locally? I remember that there is some issues when trying to get something from the service provider... But you put it at then end so I think it should work

Kritner:
I updated the `LoadSheddingValidator` to take in its actual dependencies rather than relying on the `IServiceProvider`, I saw a combination of both used in the other validators, and I guess it makes more sense for the class to depend on its actual dependencies rather than the IOC container.

Also added a few tests to the NonSilo tests class - wasn't really sure where these should go as there didn't seem to be separate classes that were for testing the individual validators. Additionally I wanted to make sure the SiloHostBuilder was injecting the things the way I thought it was, so I *guess* they work there? I was worried about having its own test class where I'm not relying on the builder to do the injection would be too "unit"y rather than "integration"y - hopefully that makes sense!

ReubenBond:
@dotnet-bot test functional

sergeybykov:
@dotnet-bot test functional

benjaminpetit:
Let's try again

@dotnet-bot test functional

Kritner:
Not having much luck with that flaky test I guess eh? And now there's new merge conflicts 😂

sergeybykov:
Thank you, @Kritner!

PR #5405, opened 2019-02-27
hack to allow physical file provider to be serialized using json

jason-bragg:


test/TestInfrastructure/TestExtensions/TestDefaultConfiguration.cs
```diff
@@ -48,6 +52,39 @@ public static void ConfigureHostConfiguration(IConfigurationBuilder builder)
...
         }
 
+        /// <summary>
+        /// Hack, allowing PhysicalFileProvider to be serialized using json
+        /// </summary>
+        [Serializable]
+        [JsonObject(MemberSerialization.OptIn)]
+        private class MyPhysicalFileProvider : IFileProvider
+        {
+            private PhysicalFileProvider fileProvider;
```
ReubenBond: Alternatively/additionally, this could have `[NonSerialized]` on it.

xiazen:
just curious, why is this change?

jason-bragg:
It seems the upgrade of Microsoft.Configuration included changes to PhysicalFileProvider which made it no longer serializable via json.  It's not clear that it was ever supposed to be serializable.  It's likely we're using the configuration sources incorrectly.  That's why this is a bit of a hack.  It just works around the issue to unblock our testing, but we may want to reconsider serializing configuration sources.

jason-bragg:
oops.. hit close instead of comment.. bleh...

test/TestInfrastructure/TestExtensions/TestDefaultConfiguration.cs
```diff
@@ -48,6 +52,39 @@ public static void ConfigureHostConfiguration(IConfigurationBuilder builder)
             builder.AddEnvironmentVariables("Orleans");
         }
 
+        /// <summary>
+        /// Hack, allowing PhysicalFileProvider to be serialized using json
+        /// </summary>
+        [Serializable]
+        [JsonObject(MemberSerialization.OptIn)]
+        private class MyPhysicalFileProvider : IFileProvider
```
jdom: nit: I'd call it something more explicit, such as `SerializablePhysicalFileProvider`

xiazen:
@jason-bragg opps seems my change did this....

PR #5406, opened 2019-02-27
Remove response callback using a single operation

ReubenBond:
Removes response callbacks using a single operation instead of TryGet + TryRemove. I don't expect the impact to be significant unless perhaps the system is under very high concurrent load.


ReubenBond:
@dotnet-bot test functional

PR #5409, opened 2019-02-28
Minor serialization perf & cleanliness improvements

### LINKED ISSUE #4443: High memory consumption by SerializationContext in SerializationManager
I noticed memory slowly being exhausted on client side servers.  A process dump from one server shows over 5GB of memory for SerializationContext records.  This averages 11MB / instance.

Count    TotalSize        Class Name
480        5629484224  System.Collections.Generic.Dictionary`2+Entry[[System.Object, mscorlib],[Orleans.Serialization.SerializationContext+Record, Orleans]][]

Running !gcroot on a couple of these instances shows they are coming from the serializationContext field in the SerializationManager class.

00000007652bdd50 00007fff7c103534 Orleans.Messaging.ProxiedMessageCenter.WaitMessage(Categories, System.Threading.CancellationToken)
        rbp+10: 00000007652bddb0
            ->  0000010b15165fd8 Orleans.Messaging.ProxiedMessageCenter
            ->  0000010794ebad60 Orleans.Serialization.SerializationManager
            ->  0000010794ebaee0 System.Threading.ThreadLocal`1[[Orleans.Serialization.SerializationContext, Orleans]]
            ->  0000010794ebaf08 System.Threading.ThreadLocal`1+LinkedSlot[[Orleans.Serialization.SerializationContext, Orleans]]
            ->  0000010b1cde8998 System.Threading.ThreadLocal`1+LinkedSlot[[Orleans.Serialization.SerializationContext, Orleans]]
.....

This process dump showed 591 background threads and based on the count above a lot of these seem to have an Orleans serialization context in thread local storage.  Would it be more memory efficient to keep a pool of SerializationContexts inside the SerializationManager instead of using ThreadLocal storage?  

Our application does have objects with large serialization subgraphs but even if we reduced this from an average of 11MB to say 2MB if there are 500 threads with a SerializationContext in local storage we're still looking at 1GB of consumed memory.  We don't create threads and most of the threads in the process dump are marked (Threadpool Worker) so I'm not sure we have a good way to reduce thread count.

ReubenBond:
Minor serializer perf fixes. Can hold until after 2.3.0 is tagged.

Fixes #4443

Also see comment: https://github.com/dotnet/orleans/pull/5210#issuecomment-443859708

### DIFF
--- a/src/Orleans.Core/Serialization/BinaryTokenStreamReader.cs
+++ b/src/Orleans.Core/Serialization/BinaryTokenStreamReader.cs
@@ -19,7 +19,7 @@ namespace Orleans.Serialization
     {
         /// <summary> Read an <c>GrainId</c> value from the stream. </summary>
         /// <returns>Data from current position in stream, converted to the appropriate output type.</returns>
-        internal static GrainId ReadGrainId(this IBinaryTokenStreamReader @this)
+        internal static GrainId ReadGrainId<TReader>(this TReader @this) where TReader : IBinaryTokenStreamReader
         {
             UniqueKey key = @this.ReadUniqueKey();
             return GrainId.GetGrainId(key);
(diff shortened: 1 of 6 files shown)


src/Orleans.Core/Serialization/BinaryTokenStreamReader.cs
```diff
@@ -867,6 +865,14 @@ public byte ReadByte()
             return buff[offset];
         }
 
+        public byte PeekByte()
```
xiazen: should `PeekToken` reuse this ? I see the two methods are oddly alike.

xiazen: LGTM except some minor comments

ReubenBond: I think `PeekToken` does use this - see line 77 above

ReubenBond: Maybe I should delete the `BinaryTokenStreamReader.PeekToken` impl. We can do that in another PR, though

PR #5413, opened 2019-02-28
Throw during startup if no grain classes/interfaces in app parts

### LINKED ISSUE #5288: SiloHostBuilder.Build() should throw if AddApplicationPart() added no app grain classes
#4785 exposed this gap - if a grain class projects lacks the codegen NuGet package, then calling `AddApplicationPart()` on the grain assembly ends up being an no-op due to missing metadata. However, `SiloHostBuilder.Build()` executes fine without warning the users about likely missing grain classes. This is an easy to make mistake when learning Orleans.

The proposal is to have `SiloHostBuilder.Build()` throw if no application grain classes were added.

ReubenBond:
Fixes #5288

### DIFF
--- a/src/Orleans.Core/Configuration/Validators/ApplicationPartValidator.cs
+++ b/src/Orleans.Core/Configuration/Validators/ApplicationPartValidator.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using System.Reflection;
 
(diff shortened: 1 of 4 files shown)


ReubenBond:
@dotnet-bot test functional

Seems like a transient error

PR #5416, opened 2019-03-02
Correcting arrangement in a few LoadSheddingValidator tests

Kritner:
- Had not set up the `LoadSheddingValidator` in `SiloHostBuilder_LoadSheddingValidatorPassesWhenLoadSheddingDisabled`
- `SiloHostBuilder_LoadSheddingValidatorFailsWithNoRegisteredHostEnvironmentStatistics` was failing expectedly, but for the wrong reason. Corrected test to pass under `LoadSheddingOptions` configuration, but to fail under the lack of a registered `IHostEnvironmentStatistics` - which was the intention of the test.


benjaminpetit:
Thanks you for following up :)

PR #5423, opened 2019-03-05
Linux version of IHostEnvironmentStatistics

martinothamar:
Started work on an implementation for `IHostEnvironmentStatistics` for Linux. Currently only one implementation for perf counters. 

Would like to have unit tests for these, would like some suggestions on how to accomplish that. There are libraries for mocking filesystem but not sure if that dependency is wanted, maybe there's another way? Doesn't look like there's any tests for the perfcounters one.

Haven't implemented anything for the `FloatValueStatistic` for CPU, GC etc. Should I? I see from .NET Core 2.2 we can get some events from CoreCLR ETW provider.

Is it wanted/possible to consolidate this with the perfcounter one so that Orleans can inject the right instance based on `System.Environment.OSVersion`?


martinothamar:
I think I should be implementing all the `IntValueStatistic`/`FloatValueStatistic` related stuff that exists in the perfcounter statistic implementation. As things are now I have to copy paste the relevant code into the linux implementation. I could abstract that into a common module if we could consolidate IHostEnvironmentStatistics into one project. Maybe with some csproj magic we could conditionally reference performance counter library? If we don't want to consolidate we could perhaps create a `TelementryConsumers.Abstractions` package where the statistics stuff is defined. Need some input on this.

src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
```diff
@@ -2,10 +2,30 @@ namespace Orleans.Statistics
...
+        /// Total memory available for allocation on the host in bytes
+        /// i.e. 14426476000L for 14 gb
+        /// </summary>
+        /// <value>14426476000</value>
         long? AvailableMemory { get; }
+
+        /// <summary>
+        /// Total memory usage in bytes as reported from <code>GC.GetTotalMemory(false)</code>
+        /// </summary>
+        long MemoryUsage { get; }
```
martinothamar: So I added this to the interface, but we could probably make naming more obvious here since the other three properties refer to the  host, while this property refers only to the process

src/Orleans.Core/Orleans.Core.csproj
```diff
@@ -19,10 +19,6 @@
     </Content>
   </ItemGroup>
 
-  <ItemGroup>
-    <Compile Remove="Statistics\RuntimeStatisticsGroup.cs" />
```
martinothamar: This file looked like a duplication of the perf-counter implementation so deleted it

src/Orleans.Core.Abstractions/Statistics/IHostEnvironmentStatistics.cs
benjaminpetit: We already have this value in `IAppEnvironmentStatistics.MemoryUsage` (implemented in `AppEnvironmentStatistics`). Since this logic in cross plat, I don't think we should add it here.

Yes, there is a `MemoryUsage` field in `PerfCounterEnvironmentStatistics` but it's not used anywhere in Orleans (as far as I know), I guess we kept it to avoid a breaking change.

src/TelemetryConsumers/Orleans.TelemetryConsumers.Linux/LinuxEnvironmentStatistics.cs
```diff
@@ -0,0 +1,276 @@
...
+                    _logger.LogWarning(StartErrMsg_Unknown);
+                return;
+            }
+
+            _logger.LogTrace($"Starting {nameof(LinuxEnvironmentStatistics)}");
+
+            _cts = new CancellationTokenSource();
+            ct.Register(() => _cts.Cancel());
+
+            _monitorTask = await Task.Factory.StartNew(
```
benjaminpetit: Not 100% sure, but woudln't it be better to simply use a `Timer` for that task instead of a look with a `Wait`?

src/TelemetryConsumers/Orleans.TelemetryConsumers.Linux/LinuxEnvironmentStatistics.cs
```diff
@@ -0,0 +1,276 @@
...
+        public void Participate(IClusterClientLifecycle lifecycle)
+        {
+            lifecycle.Subscribe(ServiceLifecycleStage.RuntimeInitialize, this);
+        }
+
+        public async Task OnStart(CancellationToken ct)
+        {
+            var isLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
+
+            if (!isLinux)
```
benjaminpetit: I think this check should be in an option validator, this way it will throw when calling `Build()` on the silo builder instead of just displaying a warning at runtime.

You could also check that you have access to the files needed

src/TelemetryConsumers/Orleans.TelemetryConsumers.Linux/LinuxEnvironmentStatistics.cs
```diff
@@ -0,0 +1,276 @@
...
+                ThreadPool.GetMaxThreads(out var maxWorkerThreads, out var maxCompletionPortThreads);
+
+                ThreadPool.GetAvailableThreads(out var workerThreads, out var completionPortThreads);
+                return maxCompletionPortThreads - completionPortThreads;
+            });
+        }
+
+        private async Task Monitor(CancellationToken ct)
+        {
+            for (int i = 0; ; i++)
```
benjaminpetit: Not a big fan of letting a int getting incremented "forever". Can you reset this counter in this loop at least?

src/TelemetryConsumers/Orleans.TelemetryConsumers.Linux/LinuxEnvironmentStatistics.cs
```diff
@@ -0,0 +1,276 @@
...
+            {
+                _logger.LogWarning($"Failed to parse '/proc/stat' output correctly. Line: {cpuUsageLine}");
+                return;
+            }
+
+            var cpuNumbers = cpuNumberStrings.Select(long.Parse).ToArray();
+            var idleTime = cpuNumbers[3];
+            var totalTime = cpuNumbers.Sum();
+
+            if (i > 0)
```
benjaminpetit: If for some reason it fails at the first call, `i` will be superior to `0` and you might report some wirted data. I don't think it is a huge issue.

benjaminpetit:
> Would like to have unit tests for these, would like some suggestions on how to accomplish that. There are libraries for mocking filesystem but not sure if that dependency is wanted, maybe there's another way? Doesn't look like there's any tests for the perfcounters one.

That's a good question. We use the perfcounter one a lot in our load tests, so when they are broken we should be able to catch it that way :). You could add another indirection layer to be able to mock these fiels read without a mocking framework, but it might a bit overkill TBH. I'm fine not having unit tests for that since it will not change too much.

> Haven't implemented anything for the FloatValueStatistic for CPU, GC etc. Should I? I see from .NET Core 2.2 we can get some events from CoreCLR ETW provider.

Not sure to understand what do you want to implement here? Go you mean writting values for stuff like `StatisticNames.RUNTIME_GC_GENCOLLECTIONCOUNT`, `StatisticNames.RUNTIME_GC_PERCENTOFTIMEINGC`, ect? If you know a way to get them from Linux, then go for it! If not, this implementation is still a start :)

> Is it wanted/possible to consolidate this with the perfcounter one so that Orleans can inject the right instance based on System.Environment.OSVersion?

I think having a separate project is fine, and probably safer.

src/TelemetryConsumers/Orleans.TelemetryConsumers.Linux/LinuxEnvironmentStatistics.cs
martinothamar: `Wait` uses `Timer` underneath AFAIK, and personally think just the loop looks simpler. Will change if you want

src/TelemetryConsumers/Orleans.TelemetryConsumers.Linux/LinuxEnvironmentStatistics.cs
martinothamar: Good point

benjaminpetit:
Thanks you! I will merge it as soon as tests are passing

benjaminpetit:
Unrelated test failures

PR #5424, opened 2019-03-06
Improve activation & directory convergence

benjaminpetit:
Rebasing and fixing comments from #4344


src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -271,6 +263,137 @@ internal void ProcessSiloAddEvent(SiloAddress addedSilo)
...
+                {
+                    if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("Sending " + splitPartListMulti.Count + " entries to " + addedSilo);
+                }
+
+                await localDirectory.GetDirectoryReference(addedSilo).AcceptSplitPartition(splitPartListSingle, splitPartListMulti);
+            }
+            else
+            {
+                if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("Silo " + addedSilo + " is no longer active and therefore cannot receive this partition split");
+            }
```
ReubenBond: So if we take long enough to successfully process the added silo that the silo is no longer active then we remove the grains from our directory without deactivating them.

I think it's an unsafe change of behavior. Should we return at this point instead?

ReubenBond: If so, should be a warning log?

src/Orleans.Core/SystemTargetInterfaces/IRemoteGrainDirectory.cs
```diff
@@ -76,5 +76,13 @@ internal interface IRemoteGrainDirectory : ISystemTarget, IGrainDirectory
...
         /// <returns></returns>
         Task RemoveHandoffPartition(SiloAddress source);
+
+        /// <summary>
+        /// Registers activations from a split partition with this directory.
+        /// </summary>
+        /// <param name="singleActivations">The single-activation registrations from the split partition.</param>
+        /// <param name="multiActivations">The multiple-activation registrations from the split partition.</param>
+        /// <returns></returns>
+        Task AcceptSplitPartition(List<ActivationAddress> singleActivations, List<ActivationAddress> multiActivations);
```
ReubenBond: At some point we should remove multi-activation support from the directory, but not in this PR

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
benjaminpetit: Yes you are right we should return instead. But the change of status can also happen after this check...

benjaminpetit:
Wasn't able to repro the tes failures locally... Let's see if it pass this time

benjaminpetit:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot test this please

ReubenBond:
@dotnet-bot test this please

ReubenBond:
The functional test failure here is nothing to be concerned about - it's a known issue and I'll submit a PR to change the test.

benjaminpetit:
@dotnet-bot test functional

benjaminpetit:
Argh closed by mistake. All tests were passing (except functional on devops...)

PR #5425, opened 2019-03-06
Refactored stream batch behaviors to support batch consumption.

jason-bragg:
Only supported by persistent streams.

### DIFF
--- a/src/Orleans.Core.Abstractions/Streams/Core/IAsyncBatchObserver.cs
+++ b/src/Orleans.Core.Abstractions/Streams/Core/IAsyncBatchObserver.cs
@@ -4,9 +4,9 @@ using System.Threading.Tasks;
 
 namespace Orleans.Streams
 {
-    public class OrderedItem<T>
+    public class SequentialItem<T>
     {
-        public OrderedItem(T item, StreamSequenceToken token)
+        public SequentialItem(T item, StreamSequenceToken token)
         {
             this.Item = item;
             this.Token = token;
@@ -47,7 +47,7 @@ namespace Orleans.Streams
         /// </summary>
         /// <param name="items">The item to be passed.</param>
         /// <returns>A Task that is completed when the item has been accepted.</returns>
-        Task OnNextAsync(IList<OrderedItem<T>> items);
+        Task OnNextAsync(IList<SequentialItem<T>> items);
 
         /// <summary>
         /// Notifies the consumer that the stream was completed.
--- a/src/Orleans.Core.Abstractions/Streams/Core/IAsyncBatchObserver.cs
+++ b/src/Orleans.Core.Abstractions/Streams/Core/IAsyncBatchObserver.cs
@@ -1,41 +1,53 @@
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace Orleans.Streams
 {
+    public class OrderedItem<T>
+    {
+        public OrderedItem(T item, StreamSequenceToken token)
+        {
+            this.Item = item;
+            this.Token = token;
+        }
+
+        public T Item { get; }
+        public StreamSequenceToken Token { get; }
+    }
+
     /// <summary>
-    /// This interface generalizes the standard .NET IObserver interface to allow asynchronous production of items.
+    /// This interface generalizes the IAsyncObserver interface to allow production and consumption of batches of items.
     /// <para>
     /// Note that this interface is implemented by item consumers and invoked (used) by item producers.
     /// This means that the consumer endpoint of a stream implements this interface.
     /// </para>
     /// </summary>
     /// <typeparam name="T">The type of object consumed by the observer.</typeparam>
-    public interface IAsyncObserver<in T>
+    public interface IAsyncBatchObserver<T>
     {
         /// <summary>
-        /// Passes the next item to the consumer.
+        /// Passes the next batch of items to the consumer.
         /// <para>
-        /// The Task returned from this method should be completed when the item's processing has been
+        /// The Task returned from this method should be completed when the items' processing has been
         /// sufficiently processed by the consumer to meet any behavioral guarantees.
         /// </para>
         /// <para>
         /// When the consumer is the (producer endpoint of) a stream, the Task is completed when the stream implementation
-        /// has accepted responsibility for the item and is assured of meeting its delivery guarantees.
-        /// For instance, a stream based on a durable queue would complete the Task when the item has been durably saved.
+        /// has accepted responsibility for the items and is assured of meeting its delivery guarantees.
+        /// For instance, a stream based on a durable queue would complete the Task when the items have been durably saved.
         /// A stream that provides best-effort at most once delivery would return a Task that is already complete.
         /// </para>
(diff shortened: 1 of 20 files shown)


src/Orleans.Core/Streams/Internal/StreamImpl.cs
```diff
@@ -113,6 +118,14 @@ public Task OnErrorAsync(Exception ex)
             return GetConsumerInterface().ResumeAsync(handle, observer, token);
         }
 
+        internal Task<StreamSubscriptionHandle<T>> ResumeAsync(
```
xiazen: what would happen if a user used both ResumeAsync method.

jason-bragg: Last resume call should win.  This is because it creates a new handle and the new handle will only have the most recent observable.  If both are called on the same handle one should fail because the handle should be invalidated after new handle is created.

src/Orleans.Core/Streams/Internal/IInternalAsyncObservable.cs
```diff
@@ -18,7 +23,7 @@ internal interface IInternalAsyncObservable<T> : IAsyncObservable<T>
     }
 
         
-    internal interface IInternalAsyncBatchObserver<in T> : IAsyncBatchObserver<T>
+    internal interface IInternalAsyncBatchObserver<T> : IAsyncObserver<T>
```
kylemc: I don't understand what this interface is used for anymore...

src/Orleans.Core/Streams/Internal/StreamSubscriptionHandleImpl.cs
```diff
@@ -84,58 +94,68 @@ public async Task<StreamHandshakeToken> DeliverBatch(IBatchContainer batch, Stre
...
-            foreach (var batchContainer in batchContainerBatch.BatchContainers)
+            if (this.observer != null)
             {
-                isRequestContextSet = batchContainer.ImportRequestContext();
-                foreach (var itemTuple in batchContainer.GetEvents<T>())
-                {
-                    await NextItem(itemTuple.Item1, itemTuple.Item2);
-                }
 
-                if (isRequestContextSet)
```
kylemc: Was it intentional to remove the "RequestContext" support?

src/Orleans.Core.Abstractions/Streams/Core/IAsyncBatchObserver.cs
```diff
@@ -0,0 +1,71 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Orleans.Streams
+{
+    public class OrderedItem<T>
```
xiazen: Ordered is a bit weird naming for me here, since this is the only place has "Ordered" keyword , among IAsyncObserver/Observerble or BatchObserver/Observable interfaces

I kinda like the name Kyle suggested "StreamItem"

src/Orleans.Core/Streams/Internal/IInternalAsyncObservable.cs
xiazen: should be this `IAsyncBatchObserver` instead of `IAsyncObserver` ?

src/Orleans.Core.Abstractions/Streams/Core/IAsyncBatchObserver.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// <summary>
+    /// This interface generalizes the IAsyncObserver interface to allow production and consumption of batches of items.
+    /// <para>
+    /// Note that this interface is implemented by item consumers and invoked (used) by item producers.
+    /// This means that the consumer endpoint of a stream implements this interface.
+    /// </para>
+    /// </summary>
+    /// <typeparam name="T">The type of object consumed by the observer.</typeparam>
+    public interface IAsyncBatchObserver<T>
```
xiazen: Can we do something like following, which is in the code base now. 

```
public interface IAsyncBatchObserver<in T> : IAsyncObserver<T>{
Task OnNextAsync(IList<OrderedItem<T>> items)
}
```

So we can avoid having two ResumeAsync methods, and also two observers side by side in our StreamConsumer class, and other stream classes. 

And when delivering, if the observer implemeted IAsyncBatchObserver, then we call its OnBatchNext method, otherwise, we call the OnNext method. 

I'm not saying we should do that, since I haven't coded that direction myself so I don't know the possible naunce.  But this direction seems to be able to eliminate the need to have two observer side by side in our stream classes, which isn't ideal for readablity and maintenance.

src/Orleans.Core/Streams/Internal/IInternalAsyncObservable.cs
jason-bragg: yeah, it's legacy junk we've not clear out yet. IMO. :/

src/Orleans.Core.Abstractions/Streams/Core/IAsyncBatchObserver.cs
jason-bragg: The concern is that the introduction of the sequence token added ordering information.  The fact that the term isn't used anywhere doesn't negate this.  I'll think more on this and try and find a better way of expressing it, but I do think it needs be clear to users.

src/Orleans.Core/Streams/Internal/IInternalAsyncObservable.cs
jason-bragg: I opted to inherit from only IAsyncObserver since the feature request was only for batched reads, not batched writes, but that was due to the below issues, which I've not yet found a 'good' solution.

IAsyncBatchObserver  and IAsyncObserver share functions, so inheriting from both interfaces cause conflicts.  We can work around them, but kinda sucks.
Original batch observer inherited from observer to avoid this, but that requires implementors to implement both batch and not batch calls and makes it unclear which will be called.
We could take the common functions and include them in a shared interfaces so their' not common, but that changes existing IAsyncObserver interface, which I was trying to avoid.

src/Orleans.Core.Abstractions/Streams/Core/IAsyncBatchObserver.cs
jason-bragg: We can do this, but it's, IMO, a really bad public surface.

Implementors of IAsyncBatchObserver would need implement the non-batch onNextAsync, when its unnecessary.
It would be unclear to users under which conditions each onNextAsync was called.
It would not be clear that one should get a different behavior by subscribing using a batched observer, for either users or implementors of the observable.

Over all, I think we'd be adopting a bad public surface for the sake of avoiding making our already debt-ridden implementation mildly more debt-ridden.

src/Orleans.Core/Streams/Internal/StreamSubscriptionHandleImpl.cs
jason-bragg: Yes, but only in the case where one enabled batch delivery with a batched observer, so not in this case.  Will fix.

PR #5430, opened 2019-03-09
Fix inconsistencies between docs and examples

vanderstack:
In the source version of this document under the header Orleans Clustering Information the documentation described the ClusterId as being set to "my-first-cluster", but the example code above this documentation was setting ClusterId to "orleans-docker". The example code has been updated to set the ClusterId to "my-first-cluster" in line with the documentation. 

In the same section ServiceId was being described as set to "AspNetSampleApp", but in the Server Configuration section the example code was setting ServiceId to "MyAwesomeOrleansService". This example code has been updated to set the ServiceId to "AspNetSampleApp" in line with the documentation.


sergeybykov:
Thank you, @vanderstack!

PR #5433, opened 2019-03-11
Implementing Chirper Sample on v2.2

JorgeCandeias:
This PR implements the v1.x Chirper sample on Orleans v2.2.

- .NET Core 2.1.
- The newer host builder components are used but not the generic host integration yet.
- Command line client rebuilt for simplicity.
- ChirperAccount grain key changed to string (for username) for demo simplicity.
- Legacy ObserverManager replaced with an internal collection.
- Network load generators *not* migrated as they appear to depend on windows-only performance counters.

Legacy files from 1.x have not been touched. Only new files for v2.2 are added.

There's more improvements that could be done, e.g. removing aged-out observers, but not sure it is warranted on a sample. That said, happy to apply fixes or improvements as needed.


sergeybykov:
Thank you, @JorgeCandeias!

PR #5445, opened 2019-03-14
Grain extensions are now available on system targets and Grain services

jason-bragg:


src/Orleans.Runtime/Catalog/ActivationData.cs
```diff
@@ -24,139 +24,6 @@ namespace Orleans.Runtime
     /// </summary>
     internal class ActivationData : IGrainActivationContext, IActivationData, IInvokable, IDisposable
     {
-        // This class is used for activations that have extension invokers. It keeps a dictionary of 
-        // invoker objects to use with the activation, and extend the default invoker
-        // defined for the grain class.
-        // Note that in all cases we never have more than one copy of an actual invoker;
-        // we may have a ExtensionInvoker per activation, in the worst case.
-        private class ExtensionInvoker : IGrainMethodInvoker, IGrainExtensionMap
```
ReubenBond: This inner class always annoyed me 🙂 thanks for moving it

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -812,15 +812,26 @@ public void RemoveExtension(IGrainExtension handler)
             return false;
         }
 
+        private ExtensionInvoker GetCurrentExtensionInvoker()
+        {
+            return (RuntimeContext.CurrentActivationContext.ContextType == SchedulingContextType.SystemTarget)
+                ? (RuntimeContext.CurrentActivationContext as SchedulingContext)?.SystemTarget.ExtensionInvoker
+                : GetCurrentActivationData().ExtensionInvoker;
+        }
```
ReubenBond: These methods should access `RuntimeContext.CurrentActivationContext` only once. It's a thread local, not as cheap as a regular field access.

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -812,15 +812,26 @@ public void RemoveExtension(IGrainExtension handler)
...
+
         private ActivationData GetCurrentActivationData()
         {
-            var activationData = (IActivationData) (RuntimeContext.CurrentActivationContext as SchedulingContext)?.Activation;
-            if (activationData == null)
-                throw new InvalidOperationException("Attempting to GetCurrentActivationData when not in an activation scope");
-            return (ActivationData)activationData;
+            if (TryGetCurrentActivationData(out ActivationData activationData)) return activationData;
+            throw new InvalidOperationException("Attempting to GetCurrentActivationData when not in an activation scope");
```
ReubenBond: Prefer we refactor this throw into a separate local function inside of `GetCurrentActivationData` so that the method can be inlined

xiazen:
what is the magic code for running AzureDevOps tests? @benjaminpetit

ReubenBond:
You can also go to 'Details' and use one of the re-run options
image

ReubenBond:
Ben fixed one of the failing tests in #5449, which I just merged.

Issue #4621, opened 2018-05-23
Stream data adapter


jason-bragg:
Custom handling of the conversion of application data to and from the underlying queue is a common concern for users of  Orleans persistent streams.  There are many reasons for this:

* The default conversion logic does not handle schema changes for event data or changes to the Orleans data structures, so if an application alters the contents of the persisted events, or the Orleans team changes how the events are wrapped, the old events will become unreadable.
* Without changes to the conversion, application developers cannot read/process events using streaming for pre-existing queues or data written to queues by other processes.
* Without changes to the conversion, application developers cannot write events in formats suitable for pre-existing systems or external consumers.

Some stream providers, like azure queue, memory, and eventhub, provide some support for custom data conversion, but there is no common pattern and some patterns are quite painful.

Given that **_every_** internal service using Orleans persistent streams has had to customize their handling of queue data for one of the above reasons, and in every case it’s made the use of Orleans persistent stream significantly more complicated, I am of the opinion that addressing this should be a fairly high priority.

jason-bragg:
This is an example of what one would need to do to customize the how data is written to and read from eventhub using eventhub streams.

The areas marked REPLACE would need be replaced with application specific code.

    public class CustomDataStreamAdapterFactory : EventHubAdapterFactory
    {
        private readonly EventHubOptions ehOptions;
        private readonly StreamCacheEvictionOptions evictionOptions;
        private readonly StreamStatisticOptions statisticOptions;
        private readonly SerializationManager serializationManager;
        private EventHubClient client;
        
        public CustomDataStreamAdapterFactory(string name, IOptionsSnapshot<EventHubOptions> ehOptions, IOptionsSnapshot<EventHubReceiverOptions> receiverOptions,
            IOptionsSnapshot<EventHubStreamCachePressureOptions> cacheOptions, IOptionsSnapshot<StreamCacheEvictionOptions> evictionOptions, IOptionsSnapshot<StreamStatisticOptions> statisticOptions, 
            IServiceProvider serviceProvider, SerializationManager serializationManager, ITelemetryProducer telemetryProducer, ILoggerFactory loggerFactory)
            : base(name, ehOptions.Get(name), receiverOptions.Get(name), cacheOptions.Get(name), evictionOptions.Get(name), statisticOptions.Get(name), serviceProvider, serializationManager, telemetryProducer, loggerFactory)
        {
            this.ehOptions = ehOptions.Get(name);
            this.evictionOptions = evictionOptions.Get(name);
[...]

PR #5450, opened 2019-03-15
Added IQueueData adapter for persistent streams.

jason-bragg:
- Updated AzureQueueStream to use IQueueDataAdapter, replacing IAzureQueueDataAdapter.
- Cleaned up azure queue stream provider configuration to support IQueueDataAdapter as a configurable component.

### DIFF
--- a/src/Azure/Orleans.Streaming.AzureStorage/Hosting/ClientBuilderExtensions.cs
+++ b/src/Azure/Orleans.Streaming.AzureStorage/Hosting/ClientBuilderExtensions.cs
@@ -17,7 +17,6 @@ namespace Orleans.Hosting
             //the constructor wires up DI with AzureQueueStream, so has to be called regardless configure is null or not
             var configurator = new ClusterClientAzureQueueStreamConfigurator(name, builder);
             configure?.Invoke(configurator);
-            configurator.ConfigureDefaults();
             return builder;
         }
 
--- a/src/Azure/Orleans.Streaming.AzureStorage/Hosting/ClientBuilderExtensions.cs
+++ b/src/Azure/Orleans.Streaming.AzureStorage/Hosting/ClientBuilderExtensions.cs
@@ -1,7 +1,6 @@
-﻿using System;
+using System;
 using Microsoft.Extensions.Options;
 using Orleans.Configuration;
-using Orleans.Providers.Streams.AzureQueue;
 using Orleans.Streaming;
 
 namespace Orleans.Hosting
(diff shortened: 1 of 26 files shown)


src/Azure/Orleans.Streaming.AzureStorage/Hosting/ClientBuilderExtensions.cs
```diff
@@ -11,25 +10,24 @@ public static class ClientBuilderExtensions
         /// <summary>
         /// Configure cluster client to use azure queue persistent streams. 
         /// </summary>
-        public static IClientBuilder AddAzureQueueStreams<TDataAdapter>(this IClientBuilder builder,
+        public static IClientBuilder AddAzureQueueStreams(this IClientBuilder builder,
```
ReubenBond: I greatly prefer this API 👍

jason-bragg:
This is the first part of https://github.com/dotnet/orleans/issues/4621.
Next step will be to apply this pattern to EventHubStreams.
If the pattern holds well for AzureQueueStreams and EventHubStreams, we can roll out the pattern to the remaining persistent streams.

src/Azure/Orleans.Streaming.AzureStorage/Hosting/ClientBuilderExtensions.cs
```diff
@@ -11,25 +10,24 @@ public static class ClientBuilderExtensions
...
             string name,
-            Action<ClusterClientAzureQueueStreamConfigurator<TDataAdapter>> configure)
-            where TDataAdapter : IAzureQueueDataAdapter
+            Action<IClusterClientAzureQueueStreamConfigurator> configure)
         {
             //the constructor wires up DI with AzureQueueStream, so has to be called regardless configure is null or not
-            var configurator = new ClusterClientAzureQueueStreamConfigurator<TDataAdapter>(name, builder);
+            var configurator = new ClusterClientAzureQueueStreamConfigurator(name, builder);
             configure?.Invoke(configurator);
+            configurator.ConfigureDefaults();
```
JanEggers: should the defaults be before configure.invoke?

jason-bragg: Not the way I'm configuring them.  Basically the defaults configure required components with their defaults -if- they've not already been configured.

I'm not terribly happy with this solution, but not come up with a better approach, as this requires creators of a configurator to understand the expected order, and if they forget or get it wrong the configurator may not function correctly.  I'd prefer a pattern that's less prone to user error.

Open to suggestions.

jason-bragg: Dropped ConfigureDefaults in favor of a default service (rather than a named service).  If a named service is not found for a provider, it will fail over to the global default.  Application developers can override this default by registering it in the container before or after the configuring azure streams.
Will show up in next push.

jason-bragg:

jason-bragg:

PR #5453, opened 2019-03-17
Migrate v1.x Presence Service sample to v2.2

JorgeCandeias:
This PR migrates the presence service sample from v1.x to v2.2.

* Targets .NET Core 2.1
* Models are now immutable
* Replaced BinaryFormatter with MessagePack
* Legacy v1.x sample left as-is

Question - This sample still makes use of observers. Is it worth replacing these with the streaming API?


JorgeCandeias:
Looks like the build failed on something unrelated.

ReubenBond:
LGTM, thanks very much, @JorgeCandeias!

PR #5457, opened 2019-03-18
Made stream fault handler a plugable component.

jason-bragg:


src/Orleans.Runtime/Streams/PersistentStream/PersistentStreamPullingManager.cs
```diff
@@ -42,35 +43,25 @@ internal class PersistentStreamPullingManager : SystemTarget, IPersistentStreamP
...
-                throw new ArgumentNullException("streamQueueBalancer", "IStreamQueueBalancer streamQueueBalancer reference should not be null");
-            }
 
             queuesToAgentsMap = new Dictionary<QueueId, PersistentStreamPullingAgent>();
             streamProviderName = strProviderName;
-            providerRuntime = runtime;
-            pubSub = streamPubSub;
+            providerRuntime = runtime ?? throw new ArgumentNullException("runtime", "IStreamProviderRuntime runtime reference should not be null");
+            pubSub = streamPubSub ?? throw new ArgumentNullException("streamPubSub", "StreamPubSub reference should not be null");
+            this.failureHandler = failureHandler ?? throw new ArgumentNullException("failureHandler", "Stream failure handler should not be null");
```
ReubenBond: nit: use `nameof(failureHandler)` - no need to change here

src/AWS/Orleans.Streaming.SQS/Hosting/ClientBuilderExtensions.cs
```diff
@@ -23,6 +23,7 @@ public static IClientBuilder AddSqsStreams(this IClientBuilder builder, string n
         {
             var configurator = new ClusterClientSqsStreamConfigurator(name, builder);
             configure?.Invoke(configurator);
+            configurator.ConfigureDefaults();
```
xiazen: the same comment with your other pr

ReubenBond:
Looks like this needs a rebase

src/Orleans.Runtime.Abstractions/Streams/SiloPersistentStreamConfigurator.cs
```diff
@@ -49,6 +50,7 @@ public SiloPersistentStreamConfigurator(string name, Action<Action<IServiceColle
             ConfigureComponent<ILifecycleParticipant<ISiloLifecycle>>(PersistentStreamProvider.ParticipateIn<ISiloLifecycle>);
             ConfigureComponent<IQueueAdapterFactory>(adapterFactory);
             ConfigureComponent<IConfigurationValidator>(PersistentStreamStorageConfigurationValidator.Create);
+            base.configureDelegate(services => services.TryAddSingleton<IStreamFailureHandler>(NoOpStreamDeliveryFailureHandler.Create));
```
xiazen: should this be Singleton or transient? Before this pr,  the behavior is more like a transient, meaning stream provider cache the handler factory, and create a new handler each time the pulling agent asks for one.  And also , before this pr, failure handler can be initialized with an asynchronously, see `Func<string, Task<IStreamFailureHandler>> StreamFailureHandlerFactory`  

I bring this up because this is more than a compile time breaking change, it is also a functionality/behavior  breaking change,

jason-bragg: For the specific impl "NoOpStreamDeliveryFailureHandler", a singleton is safe.

In the case of a user configured handler, all configured components have been singletons so far, and I think that's fine for this one as well.  I understand that this is a deviation from the previous pattern, but I suspect most implementations will not contain per queue state.  If they do need per queue state, the application developer can register their impl as a transient named service.

Regarding the loss of asynchronous construction, if an impl needs asynchronous initialization, the pattern we've used in other components (like grain storage) is for them to subscribe to the silo lifecycle and initialize themselves at startup, rather than waiting until they're resolved from the container.  I don't think maintaining the previous asynchronous construction (for compatibility reasons) is sufficient to justify deviating from the common pattern we've settled on in other areas of the system for the following reasons.
- I don't think most users have even implemented a failure handler due to the difficulty of doing so prior to this change, so this will be a breaking change for very few.
- I don't think asynchronous initialization will be necessary very often.
- If asynchronous initialization is necessary, it's better to perform it at silo startup than when the component is resolved from the container, as we have better testing and recovery patterns around silo initialization problems.

xiazen:
needs rebase.

src/Orleans.Core/Streams/QueueAdapters/IQueueAdapterFactory.cs
```diff
@@ -26,12 +24,5 @@ public interface IQueueAdapterFactory
         /// </summary>
         /// <returns></returns>
         IStreamQueueMapper GetStreamQueueMapper();
-
-        /// <summary>
-        /// Acquire delivery failure handler for a queue
-        /// </summary>
-        /// <param name="queueId"></param>
-        /// <returns></returns>
-        Task<IStreamFailureHandler> GetDeliveryFailureHandler(QueueId queueId);
```
JanEggers: any chance this pr is extended and the whole IQueueAdapterFactory is replaced by DI?

jason-bragg: That's roughly the plan, long term.

The queue adapter was just supposed to be a thin abstraction layer over the underlying queuing system to allow the persistent stream infrastructure to run on a range of queuing solutions.  Over time and under delivery pressures it took on more features and complexity, in part because there was no uniform factory pattern in Orleans at the time.

Streams was one of the systems that prompted the adoption of DI, as it necessitated either the development of our own factory patterns or leveraging existing solutions.  DI was selected, and we've been (slowly) refactoring streams and other systems over time to leverage it.

There is a fair amount of technical debt in our streaming infrastructure, but the systems work sufficiently well to rarely become the priority.  We welcome help cleaning it up! :)

PR #5458, opened 2019-03-18
Component configurator function order independant

jason-bragg:
Modified component configurator extension functions so order of configuration no longer matters.

With component specialization via interfaces, as IAAA : IAA: IA, extension functions on the more specialized (IAAA) interfaces had to be used first, as the extension functions on the less specific interface (IA) would return the less specific interface, limiting all future extension functions to that limited surface.

Example from streams, the following would work:

    .AddEventHubStreams(StreamProviderName, b=>b
        .ConfigureEventHub(ob=>ob.Configure(
            options =>
            {
                options.ConnectionString = TestDefaultConfiguration.EventHubConnectionString;
                options.ConsumerGroup = EHConsumerGroup;
                options.Path = EHPath;
            }))
        .ConfigureStreamPubSub(StreamPubSubType.ImplicitOnly));

but this wouldn't

    .AddEventHubStreams(StreamProviderName, b=>b
        .ConfigureStreamPubSub(StreamPubSubType.ImplicitOnly)
        .ConfigureEventHub(ob=>ob.Configure(
            options =>
            {
                options.ConnectionString = TestDefaultConfiguration.EventHubConnectionString;
                options.ConsumerGroup = EHConsumerGroup;
                options.Path = EHPath;
            })));

Because the ConfigureEventHub extension function was hung on a more specialized interface than ConfigureEventHub, so it needed to be used first.

With this change, they both work.


src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubStreamBuilder.cs
```diff
@@ -1,14 +1,57 @@
...
+
+    public static class SiloEventHubStreamConfiguratorExtensions
+    {
+        public static TConfigurator ConfigureCheckpointer<TConfigurator,TOptions>(this TConfigurator configurator, Func<IServiceProvider, string, IStreamQueueCheckpointerFactory> checkpointerFactoryBuilder, Action<OptionsBuilder<TOptions>> configureOptions)
+            where TConfigurator : ISiloEventHubStreamConfigurator
+            where TOptions : class, new()
+        {
+            configurator.ConfigureComponent<TOptions, IStreamQueueCheckpointerFactory>(checkpointerFactoryBuilder, configureOptions);
+            return configurator;
+        }
```
ReubenBond: I take it users aren't meant to call this method directly? The type parameter inference might make it a little ugly. This is the reason we didn't go with this pattern (builder methods returning constrained generics) for @galvesribeiro's Generic Host PR: because if the user has to specify one generic parameter then they must specify all generic parameters.

I'm not pushing back

galvesribeiro: Yeah that was what I originally did for that PR, but then I saw @ReubenBond's point and makes total sense...

jason-bragg: Agreed that this can get clumsy in some cases.
I'd not yet come up with a better pattern.
Alternatives?

jason-bragg: I'd not that I don't think most usages will require users to specify the generics.  See the usage of ConfigureCheckpointer later in this file.  The generics are listed for readability reasons, but are not required.

Though I sued:

    configurator.ConfigureCheckpointer<TConfigurator,AzureTableStreamCheckpointerOptions>(
        EventHubCheckpointerFactory.CreateFactory, configureOptions);

The usage:

    configurator.ConfigureCheckpointer(
        EventHubCheckpointerFactory.CreateFactory, configureOptions);

resolves fine, without need for users to specify configurator type.

jason-bragg:

jason-bragg:
@dotnet-bot test bvt please

PR #5460, opened 2019-03-19
Fix upper-case typo in warning log

eduherminio:
I couldn't help  ¯\_(ツ)_/¯


ReubenBond:
:) Thanks for finding & fixing that, @eduherminio!

PR #5471, opened 2019-03-23
Ad-hoc Reactive Caching / Computation Sample

JorgeCandeias:
This PR adds a new ad-hoc reactive caching and computation algorithm sample.

It's based on the [Reactive Caching for Composed Services](https://www.microsoft.com/en-us/research/publication/reactive-caching-for-composed-services/) documentation and own experience hacking Orleans at client site.

I would appreciate a second pair of eyes on this one! Detailed workings in the readme file.

### DIFF
--- a/Samples/2.3/AdHocReactiveCaching/src/Grains/AggregatorCacheGrain.cs
+++ b/Samples/2.3/AdHocReactiveCaching/src/Grains/AggregatorCacheGrain.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Threading;
 using System.Threading.Tasks;
 using Grains.Models;
 using Microsoft.Extensions.Logging;
@@ -13,9 +12,8 @@ namespace Grains
     public class AggregatorCacheGrain : Grain, IAggregatorCacheGrain
     {
         private readonly ILogger<AggregatorCacheGrain> _logger;
-        private readonly CancellationTokenSource _token = new CancellationTokenSource();
         private VersionedValue<int> _cache;
-        private Task _pollTask;
+        private IDisposable _pollTimer;
 
         public AggregatorCacheGrain(ILogger<AggregatorCacheGrain> logger)
         {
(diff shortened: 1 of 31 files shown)


Samples/2.3/AdHocReactiveCaching/src/Grains/AggregatorCacheGrain.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        public override Task OnDeactivateAsync()
+        {
+            _token.Cancel();
+
+            return base.OnDeactivateAsync();
+        }
+
+        private async Task PollAsync()
```
sergeybykov: Asynchronous loops like this is something that we generally don't recommend as a pattern because it is easy to get into troubles with them. The preferred way is to schedule a timer instead. Each time tick would execute an equivalent of an iteration of the loop, but with a cleaner separation of individual tasks.

Samples/2.3/AdHocReactiveCaching/src/Grains/AggregatorGrain.cs
```diff
@@ -0,0 +1,105 @@
...
+        public Task<VersionedValue<int>> GetAsync() => Task.FromResult(_sumValue);
+
+        public Task<VersionedValue<int>> LongPollAsync(int knownVersion) =>
+            knownVersion == _sumValue.Version
+            ? _wait.Task
+            : Task.FromResult(_sumValue);
+
+        private async Task LongPollLeftAsync()
+        {
+            while (!_cancellation.IsCancellationRequested)
```
sergeybykov: Same issue with an infinite loop here.

Samples/2.3/AdHocReactiveCaching/src/Grains/AggregatorGrain.cs
```diff
@@ -0,0 +1,105 @@
...
+                }
+                catch (TimeoutException)
+                {
+                }
+            }
+        }
+
+        private async Task LongPollRightAsync()
+        {
+            while (!_cancellation.IsCancellationRequested)
```
sergeybykov: And here.

Samples/2.3/AdHocReactiveCaching/src/Grains/ProducerCacheGrain.cs
```diff
@@ -0,0 +1,63 @@
...
+        public override Task OnDeactivateAsync()
+        {
+            _token.Cancel();
+
+            return base.OnDeactivateAsync();
+        }
+
+        private async Task PollAsync()
+        {
+            while (!_token.IsCancellationRequested)
```
sergeybykov: And here.

Samples/2.3/AdHocReactiveCaching/src/Grains/ProducerCacheGrain.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                try
+                {
+                    _cache = await GrainFactory.GetGrain<IProducerGrain>(GrainKey).LongPollAsync(_cache.Version);
```
sergeybykov: Also, is there a way to avoid the code duplication by maybe reusing a helper method that takes a lambda to invoke?

sergeybykov:
Thank you, @JorgeCandeias!

JorgeCandeias:
Wait, this wasn't resolved yet! My bad, shoud've WIP it! I guess I'll raise another PR with the rest.

sergeybykov:
No worries. We can improve the sample with another PR.

PR #5475, opened 2019-03-26
Observers have been Replaced with Streams

RehanSaeed:
I'm new to Orleans. This page threw me off until I read somewhere that Streams had replaced this functionality. If there is docs for older versions, then this document should go there.


martinothamar:
It still exists in a legacy package I think, in that case maybe the doc should just be updated to reflect that along with the reasoning for why.

ReubenBond:
Why did the page throw you off? Streams aren't a 1:1 replacement for Observers, which are a lower-level mechanism for silo->client calls. They are no plans to remove observers

ReubenBond:
I see, the references to `ObserverSubscriptionManager<T>` are out of date, since that's legacy now. I prefer to use a simple helper class for that: https://gist.github.com/ReubenBond/0bb430a52eedce50e6976f2f89154f5a

JorgeCandeias:
The `IGrainObserver` class still exists and is still useful for pushing light-weight throw-away notifications (e.g. user interface toasts). The `ObserverSubscriptionManager` however was moved to the legacy package as it's often best to write your own. See the [Chirper Sample](https://github.com/dotnet/orleans/tree/master/Samples/2.2/Chirper) for a simple example of observers in use.

RehanSaeed:
> Streams aren't a 1:1 replacement for Observers, which are a lower-level mechanism for silo->client calls. They are no plans to remove observers

Seeing this sentence at the top of the documentation, would be super useful.

RehanSaeed:
I just discovered that [this page](https://dotnet.github.io/orleans/Documentation/clusters_and_clients/index.html#receiving-notifications) in the docs talks about this but it's buried in the "What is a Grain Client?" page.

sergeybykov:
@RehanSaeed We should close this PR, right? Observers are still a feature and will continue to exist for a foreseeable future. Improving the documentation is a separate matter.

PR #5477, opened 2019-03-27
Fix for the correct way to add App Insights

RehanSaeed:


sergeybykov:
Thank you, @RehanSaeed!

Issue #5485, opened 2019-03-29
Question: How can one access the IReminderRegistry from a static Grain extension method.


JorgeCandeias:
I'm attempting to write a static extension method for the `Grain` type, which, in addition to other things, also registers a timer.

The easiest way would be to call `RegisterTimer()` from the grain, however that method is protected and hence not available to a static extension method.

I'd like to avoid weird stuff like creating an extension `.UseSomething()` method on `IHostBuilder` to expose the timer registry to some static variable.

Is there some nice way to reach the `ITimerRegistry` in this context?

ReubenBond:
All of the important methods/properties of `Grain` are either `protected` or `internal`. If `Grain.Runtime` was exposed then you would be able to access `grain.Runtime.TimerRegistry`.

To the best of my knowledge, `Grain` is never exposed to user code directly except perhaps via `IIncomingGrainCallContext` (and in that case it's exposed as `IAddressable`).

Perhaps we should expose `IGrainRuntime` via `Grain`'s `Runtime` property. What do you think?

There are protections in the `IGrainRuntime` impl to prevent access from outside of the grain's execution context (eg, you can't access it from `Task.Run`). I assume that's fine for you.

ReubenBond:
If you only need this functionality for your own grains then you can create a grain base class or interface which exposes `IGrainRuntime` property in some way (inject it if needed).

JorgeCandeias:
This is to improve the ad-hoc reactive caching sample PR (and the improvement will spill into out own solutions here). I'm fine with deriving the class for my own solution. However I'd prefer having something less intrusive for the benefit of the sample users.

---

> Perhaps we should expose IGrainRuntime via Grain's Runtime property. What do you think?

I'm always weary of exposing API surface *by default* without a good reason, as it's hard to go back on that decision. If it's just for this use case, I don't think that API change is warranted.

> There are protections in the IGrainRuntime impl to prevent access from outside of the grain's execution context (eg, you can't access it from Task.Run). I assume that's fine for you.

That's fine - the extension is meant to be called from within the grain. It's a comfort method, nothing else.

---

Right now I'm considering the interface way, e.g.:

* Create an interface that allows implementations to expose a timer registry, for example:
``` csharp
public interface ISomeExposingInterface
{
    ITimerRegistry TimerRegistry { get; }
}
```
* Implement the interface on the given grain, thereby *explicitly* exposing the TimerRegistry.
* Extend the interface with the static method instead of the grain class.

[...]

PR #5487, opened 2019-03-29
Reactive Caching Sample Leftovers

JorgeCandeias:
Here are the rest of the changes for now. If #5485 surfaces a better approach, I'll come back to this.

Not super happy with [this commit](https://github.com/dotnet/orleans/commit/54ab4c1abae8e9b2a8d0f3266bd8777e9108bead) though. Is there any way we can disable the timeout warnings only for specific calls?

Also, very grateful for the tips on the previous PR! 👍


sergeybykov:
>Not super happy with this commit though. Is there any way we can disable the timeout warnings only for specific calls?

I don't think masking the timeouts is a good thing to do. That would point people to a wrong direction. Maybe the code should simply add a timer for a period shorter than the timeout (e.g. 8 seconds in this sample) to make sure the `Task` gets resolved before the call times out?

I tried adding the following lines to ProducerGrain.cs while removing the log filtering, and it seems to work without timeouts:
```csharp
private TimeSpan timeoutTimerDelay = TimeSpan.FromSeconds(8);
private IDisposable _timeoutTimer;
…

_wait = new TaskCompletionSource<VersionedValue<int>>();

_timeoutTimer = RegisterTimer(_ =>  
{
          if (!_wait.Task.IsCompleted)
         _wait.TrySetResult(_state);
         _wait = new TaskCompletionSource<VersionedValue<int>>();
          return Task.CompletedTask;
},
null,
timeoutTimerDelay,
timeoutTimerDelay);
```

JorgeCandeias:
I agree, that's a better approach, thank you for the feedback. I'm wondering if this could be even cleaner by returning a `Task.WhenAny()` to the caller, where one of the tasks is a delay with `timeoutTimerDelay` (a different one per call and resolving to an empty `VersionedValue<T>` by design) and the other one is the task from the `TaskCompletionSource`. Maybe it would let each call make better use of its timeout allowance, instead of having more recent calls break early because of the timer. Not yet sure if this will work though. I'll push some changes either way.

sergeybykov:
>I'm wondering if this could be even cleaner by returning a Task.WhenAny() to the caller, where one of the tasks is a delay with timeoutTimerDelay

We have an extension method for that - `OrleansTaskExtentions.WithTimeout()`. I didn't employ it here just to avoid throwing timeout exceptions as it would.

JorgeCandeias:
A few changes:

* Promoted the concept of version to its own `VersionToken` that makes explicit the special case of "No Version" (aka "move along, nothing to see here yet").
* `PublisherGrain` now fulfills a long poll with any new value it generates or... an explicit default of "none" just before each calling promise breaks.
* `ReactiveGrain` now has a `RegisterReactivePollAsync()` method that starts to look a bit like an API method. This sets up the long polling behaviour in one go.

The logs still show timeouts, but these are now due to the sample itself logging them, not to Orleans breaking promises.

JorgeCandeias:
Forgot to remove the logging filters. Will do that later, please don’t accept this yet.

JorgeCandeias:
Done.

sergeybykov:
Thank you, @JorgeCandeias!

PR #5493, opened 2019-04-01
Add Kubernetes support to Docker sample

dgkanatsios:
This is a WIP PR to add support for Kubernetes on the Docker sample. I played with the sample myself, wrote a simple Makefile to test and I thought of PR'ing it to get some feedback whether you want something like that. E.g. want the Kubernetes sample combined with the Docker one or maybe we should create a separate Kubernetes sample? I'm open to suggestions :)

Btw, whereas deployment works (tested locally with Docker for Windows and remotely using Azure Kubernetes Service), I'm getting an Exception after some time and the silo dies. Any hints?

```
orleans-silo-6988656b9-jh5wz orleans-silo Silo started
orleans-silo-6988656b9-jh5wz orleans-silo warn: Orleans.Runtime.MembershipService.MembershipOracleData[100618]
orleans-silo-6988656b9-jh5wz orleans-silo       I should be Dead according to membership table (in CleanupTableEntries): entry = SiloAddress=S10.244.0.62:11111:291805778 SiloName=Silo_98c6d Status=Dead.
orleans-silo-6988656b9-jh5wz orleans-silo fail: Orleans.Runtime.MembershipService.MembershipOracleData[100627]
orleans-silo-6988656b9-jh5wz orleans-silo       I have been told I am dead, so this silo will stop! I should be Dead according to membership table (in CleanupTableEntries): entry = SiloAddress=S10.244.0.62:11111:291805778 SiloName=Silo_98c6d Status=Dead.
orleans-silo-6988656b9-jh5wz orleans-silo fail: Orleans.Runtime.MembershipService.MembershipOracleData[100627]
orleans-silo-6988656b9-jh5wz orleans-silo       INTERNAL FAILURE! About to crash! Fail message is: I have been told I am dead, so this silo will stop! I should be Dead according to membership table (in CleanupTableEntries): entry = SiloAddress=S10.244.0.62:11111:291805778 SiloName=Silo_98c6d Status=Dead.
```

Btw, I'm pretty new to Orleans so apologies in advance if I missed something.

### DIFF
--- a/Samples/2.0/docker-aspnet-core/API/API.csproj
+++ b/Samples/2.0/docker-aspnet-core/API/API.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <DockerComposeProjectPath>..\docker-compose.dcproj</DockerComposeProjectPath>
+    <ServerGarbageCollection>false</ServerGarbageCollection> <!-- https://blog.markvincze.com/troubleshooting-high-memory-usage-with-asp-net-core-on-kubernetes/ -->
   </PropertyGroup>
 
   <ItemGroup>
--- a/Samples/2.0/docker-aspnet-core/Makefile
+++ b/Samples/2.0/docker-aspnet-core/Makefile
@@ -1,0 +1,47 @@
+# a Makefile to test this sample on Kubernetes
+# you can test it locally (e.g. using Docker for Windows/Mac) or remotely, on your cloud provider of choice
+# do not forget to set Azure Storage connection string on Silo/Program.cs and API/Startup.cs
+
+# your Docker registry (we suppose you have logged in and have push access)
+REGISTRY ?= docker.io/dgkanatsios
+# the context of your local Kubernetes cluster
+LOCAL_K8S_CLUSTER=docker-for-desktop
+# the contect of your remote Kubernetes cluster
+REMOTE_K8S_CLUSTER=aks
+# version/tag of the images that will be pushed to Docker Hub
+VERSION=0.0.8
+
+API_PROJECT_NAME=orleans-api
+SILO_PROJECT_NAME=orleans-silo
+
+# tag of the images that will be pushed for local development
+TAG?=$(shell git rev-list HEAD --max-count=1 --abbrev-commit)
+
+buildlocal: 
+		docker build -f ./Silo/Dockerfile -t $(SILO_PROJECT_NAME):$(TAG) .
+		docker build -f ./API/Dockerfile -t $(API_PROJECT_NAME):$(TAG) .
+		docker system prune -f
+deploylocal: uselocalcontext
+		kubectl config use-context $(LOCAL_K8S_CLUSTER)
+		sed -e 's/orleans-silo-image/$(SILO_PROJECT_NAME):$(TAG)/' -e 's/orleans-api-image/$(API_PROJECT_NAME):$(TAG)/' deploy.yaml | kubectl apply -f -
+cleandeploylocal: uselocalcontext clean
+uselocalcontext:
+		kubectl config use-context $(LOCAL_K8S_CLUSTER)
+
+buildremote: 
+		docker build -f ./Silo/Dockerfile -t $(REGISTRY)/$(SILO_PROJECT_NAME):$(VERSION) .
+		docker build -f ./API/Dockerfile -t $(REGISTRY)/$(API_PROJECT_NAME):$(VERSION) .
+		docker system prune -f
+pushremote:
+		docker push $(REGISTRY)/$(SILO_PROJECT_NAME):$(VERSION)
+		docker push $(REGISTRY)/$(API_PROJECT_NAME):$(VERSION)
+deployremote: useremotecontext
+		sed -e 's,orleans-silo-image,$(REGISTRY)/$(SILO_PROJECT_NAME):$(VERSION),' -e 's,orleans-api-image,$(REGISTRY)/$(API_PROJECT_NAME):$(VERSION),' deploy.yaml | kubectl apply -f -
+		# after a few minutes, you can do a `kubectl get service` to get the Public IP endpoint
+cleandeployremote: useremotecontext clean
+useremotecontext:	
+		kubectl config use-context $(REMOTE_K8S_CLUSTER)
+
+clean:
+		kubectl delete deployment $(SILO_PROJECT_NAME) $(API_PROJECT_NAME)
+		kubectl delete service $(API_PROJECT_NAME)	
--- a/Samples/2.0/docker-aspnet-core/Makefile
+++ b/Samples/2.0/docker-aspnet-core/Makefile
@@ -7,9 +7,9 @@ REGISTRY ?= docker.io/dgkanatsios
 # the context of your local Kubernetes cluster
 LOCAL_K8S_CLUSTER=docker-for-desktop
 # the contect of your remote Kubernetes cluster
-REMOTE_K8S_CLUSTER=aks
+REMOTE_K8S_CLUSTER=aksorleans
 # version/tag of the images that will be pushed to Docker Hub
-VERSION=0.0.8
+VERSION=0.0.14
 
 API_PROJECT_NAME=orleans-api
 SILO_PROJECT_NAME=orleans-silo
--- a/Samples/2.0/docker-aspnet-core/README.md
+++ b/Samples/2.0/docker-aspnet-core/README.md
@@ -76,9 +76,17 @@ If you want to remove remote resources, use:
 make cleandeployremote
 ```
 
+### Using StatefulSets
+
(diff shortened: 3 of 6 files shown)


andhesky:
Based on those logs, I am not sure of the problem. It seems that the silo get voted dead between the 1st and 2nd line of the logs. The cluster membership table is in Azure Table by default with the table name OrleansSiloInstances.  If you are deploying with multiple silos, it could be that they are unable to ping each other and therefore [vote each other dead](http://dotnet.github.io/orleans/Documentation/implementation/cluster_management.html). If there is a single silo then it works slightly differently and the silo must be marking itself dead for some reason. I would add additional logging and look for a message where a silo is being marked as dead.

As for the approach. @sergeybykov or @ReubenBond would be better to comment.

dgkanatsios:
Thanks @andhesky. I did some modifications (e.g. randomizing the Port(s) that Silo uses), Silo still crashing with the same message. Any hints on where I should log to see what's going on?
However, if I increase the number of the Silo replicas to more than 1 (I used 5), sample works flawlessly.

sergeybykov:
@dgkanatsios Can you share your silo configuration code? Those TCP ports aren't blocked by the firewall?
@andhesky 
>The cluster membership table is in Azure Table by default

Not unless the silo is explicitly configured to use Azure Table for cluster membership.

dgkanatsios:
so, I'm using the sample here -> https://github.com/dotnet/orleans/tree/master/Samples/2.0/docker-aspnet-core

The only changes I've made are
- add the Makefile (which is in this PR)
- created an Azure Storage account and added its connection string [here](https://github.com/dotnet/orleans/blob/master/Samples/2.0/docker-aspnet-core/API/Startup.cs#L49) and [here](https://github.com/dotnet/orleans/blob/master/Samples/2.0/docker-aspnet-core/Silo/Program.cs#L21)
- modified code [here](https://github.com/dotnet/orleans/blob/master/Samples/2.0/docker-aspnet-core/Silo/Program.cs#L29) so that ports are random

Kubernetes Pods generally can talk to each other freely, so no firewall between them. I'm using hostNetworking to create the Pods so all should probably be good. Btw, do Silos talk to each other via IP or DNS names? Can you think of a reason why a *single* Silo would mark itself as dead? As I mentioned, 5 did work, 2 did work as well, but once I scaled down to 1, it went boom :(

Thanks! I'd also like some feedback on how to document it after we fix it, since Kubernetes is pretty hot at the moment.

dgkanatsios:
[silotrace.txt](https://github.com/dotnet/orleans/files/3033858/silotrace.txt)

activated Trace on the Silo, checking

andhesky:
Silos use IP to talk to each other. Specifically, when using AzureStorage clustering as you are doing, they will write their IP to the table and then other Silos and clients pick it up from there to use it.

Since it only happens with one silo, it does not seem to be a networking issue.  I took a look at the trace, but it doesn't seem to contain the issue.  I don't see the "I should be Dead" anywhere in that log.

dgkanatsios:
thanks @andhesky 
Another trace
[silotrace2.txt](https://github.com/dotnet/orleans/files/3035024/silotrace2.txt)
and an export from Azure Table (bear in mind that I stopped capturing silo logs a bit before the third silo was created)
[OrleansSiloInstances.typed.txt](https://github.com/dotnet/orleans/files/3035028/OrleansSiloInstances.typed.txt)

JoshCollinsMSFT:
From the original log this is a known issue with reusing IPs and a single silo in a cluster.  If the membership table says that instance (IP + a couple other things) is dead, and there are no other silos to communicate with to come to a consensus, it won't mark itself as alive.  This is a common problem when debugging locally, which we usually work around by dropping the membership table (using the ATS emulator).  When deploying, even for development scenarios, it is always recommended to have 2+ silos to avoid these problems.  Assuming you don't put resource constraints on your deployment definitions, running 2+ pods with silos doesn't really mean you need more hardware.

dgkanatsios:
thanks @JoshCollinsMSFT, this makes sense, I thought the issue would have to do with the fact of the Silos running as Pods. 
Some questions then, on top of that:
- so, if we do change the Membership provider, the bug would disappear (talking for a single Silo)?
- do you recommend another Membership provider, in this case?
- how would you deal with Silo scaling? Is Kubernetes HPA (with CPU/RAM metrics) enough?
- not sure what you mean in your last sentence -> `running 2+ pods with silos doesn't really mean you need more hardware`. Technically, on a production environment you should have requests/limits set on your Pods, so you should need some scaling, right?

JoshCollinsMSFT:
I wouldn't really call this a bug per se, clusters are not really designed to be ran with a single silo, so that really breaks the membership oracle's algorithm.

Re: your questions
- Perhaps, I don't have a whole lot of experience with membership providers outside of ATS
- A better recommendation isn't to deploy clusters with single silos
- We're still investigating the best strategies here for our production readiness.  HPA is one piece of the puzzle for sure, but node autoscale is still a concern and is something that the internal AKS customers are working with the AKS team on for coming up with a better solution than the current script they have available.
- For production environments, sure, but for non-prod environments (i.e. dev/test envs that don't mirror prod) your pod density can be arbitrary.  But for prod envs you definitely will need to do some scale/loading testing to determine your per silo resource constraints.  Note that Orleans is designed with fewer, larger silos in mind, but a lot of that is based on the default algorithms and strategies used for grain placement and directories.

Note on autoscaling - Orleans itself doesn't have a built-in rebalancing algorithm for grain placement.  This is something we're going to be looking at as well, with K8s autoscaling we'll need to replace/augment the default grain placement component with one that will give higher priority to under utilized silos (i.e. less grains than per silo average, CPU/memory, etc).

dgkanatsios:
ping @andhesky @JoshCollinsMSFT @sergeybykov for a review

dgkanatsios:
Added another commit that uses CPU/RAM limits for the API and the Silo Pods, describes the potential use of StatefulSets (thanks to [James Sukhabut](https://github.com/jsukhabut)) as well as sets ServerGarbageCollection to false based on [this](https://blog.markvincze.com/troubleshooting-high-memory-usage-with-asp-net-core-on-kubernetes/).

Samples/2.0/docker-aspnet-core/API/API.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <DockerComposeProjectPath>..\docker-compose.dcproj</DockerComposeProjectPath>
+    <ServerGarbageCollection>false</ServerGarbageCollection> <!-- https://blog.markvincze.com/troubleshooting-high-memory-usage-with-asp-net-core-on-kubernetes/ -->
```
benjaminpetit: This is interesting... But we know that Orleans will run slower without that, so I don't know if we should set this value here. 

See https://dotnet.github.io/orleans/Documentation/clusters_and_clients/configuration_guide/configuring_.NET_garbage_collection.htm

Samples/2.0/docker-aspnet-core/Makefile
```diff
@@ -0,0 +1,47 @@
...
+# tag of the images that will be pushed for local development
+TAG?=$(shell git rev-list HEAD --max-count=1 --abbrev-commit)
+
+buildlocal: 
+		docker build -f ./Silo/Dockerfile -t $(SILO_PROJECT_NAME):$(TAG) .
+		docker build -f ./API/Dockerfile -t $(API_PROJECT_NAME):$(TAG) .
+		docker system prune -f
+deploylocal: uselocalcontext
+		kubectl config use-context $(LOCAL_K8S_CLUSTER)
+		sed -e 's/orleans-silo-image/$(SILO_PROJECT_NAME):$(TAG)/' -e 's/orleans-api-image/$(API_PROJECT_NAME):$(TAG)/' deploy.yaml | kubectl apply -f -
```
benjaminpetit: Does `sed` will work on a Windows machine with just docker and kubectl installed?

Samples/2.0/docker-aspnet-core/API/API.csproj
dgkanatsios: Thanks for the review!
Pods tended to crash overnight without this setting (due to Out Of Memory errors) which confused both clusters (Kubernetes and Orleans). This led to many Pods creating/killed etc.
I *think* this will be fixed in .NET Core 3 ([source](https://devblogs.microsoft.com/dotnet/announcing-net-core-3-preview-3/)). Maybe we can leave this now and remove it when .NET Core 3 shipped.

Samples/2.0/docker-aspnet-core/Makefile
dgkanatsios: Nope, but `make` will not work either. Maybe add `use make if on Linux/WSL`?

galvesribeiro: I only saw that now... I would avoid use make and sed for those examples... better use msbuild...

PR #5509, opened 2019-04-04
Remove redundant white-space from grain_lifecycle.md

7rakir:
Removed redundant white-space and fixed case of a method's name.


sergeybykov:
Thank you, @7rakir!

PR #5514, opened 2019-04-06
Enable TransactionalStateStorageTestRunner to test with custom type

xiazen:
TransactionalStateStorageTestRunner was hard coded to test manipulating integer state only. This PR enable it to test with any type, as long as the type is a reference type and implements `IEquatable`. To use this, users must do followings
- The custom type must implement `IEquatable`, and user should inject the custom type's validation logic through `IEquatable`. 
- the `Func<TState> stateFactory` feeded to `TransactionalStateStorageTestRunner ` constructor should return a different value each time, to ensure test coverage.


test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionalStateStorageTests.cs
```diff
@@ -19,17 +19,23 @@
...
+        }
     }
 
     public class AzureTransactionalStateStorageTests : TransactionalStateStorageTestRunnerxUnit<TestState>, IClassFixture<TestFixture>
     {
         private const string tableName = "StateStorageTests";
         private const string partition = "testpartition";
         public AzureTransactionalStateStorageTests(TestFixture fixture, ITestOutputHelper testOutput)
-            :base(()=>StateStorageFactory(fixture), ()=>new TestState(), fixture.GrainFactory, testOutput)
+            :base(()=>StateStorageFactory(fixture), ()=>new TestState(){State = TestState.Seed.Next()}, fixture.GrainFactory, testOutput)
```
jason-bragg: Should state factory have some hint for making different states deterministically, rather than relying on randomness?  Like maybe take in optional integer?  As is, either all expected states are the same as each other and the default state, allowing for a wide range of bugs to go undetected, or the factory must introduce randomness, which may make test failures intermittent and hard to reproduce.

src/Orleans.Transactions.Testkit.Xunit/TransactionalStateStorageTestRunner.cs
```diff
@@ -7,7 +7,7 @@
 namespace Orleans.Transactions.TestKit.xUnit
 {
     public abstract class TransactionalStateStorageTestRunnerxUnit<TState> : TransactionalStateStorageTestRunner<TState>
-        where TState: class, ITestState, new()
+        where TState: class, IEquatable<TState>, new()
```
JoshCollinsMSFT: >IEquatable<TState> [](start = 29, length = 18)

Given the usage of ShouldBeEquivalentTo() IEquatable isn't necessary.  ShouldBeEquivalentTo() will do an object graph walk to do comparison. Implementers can use either IEquatable<> or EquivalencyAssertionOptions<TState> (look at the second parameter of ShouldBeEquivalentTo()) to control object comparison further if necessary.

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionalStateStorageTests.cs
JoshCollinsMSFT: I agree with this.  My implementation used a int seed parameter to make the output deterministic.

---
In reply to: [273162605](https://github.com/dotnet/orleans/pull/5514#discussion_r273162605) [](ancestors = 273162605)

JoshCollinsMSFT:
>         protected TransactionalStateStorageTestRunner(Func<Task<ITransactionalStateStorage<TState>>> stateStorageFactory, Func<TState> stateFactory, 

State factory should take a seed component to make its output deterministic.

---
Refers to: src/Orleans.Transactions.TestkitBase/TestRunners/TransactionalStateStorageTestRunner.cs:24 in 05b2268. [](commit_id = 05b2268d8fa7f4592d5fea50a97e5bbb96ccb1fc, deletion_comment = False)

src/Orleans.Transactions.TestkitBase/TestRunners/TransactionalStateStorageTestRunner.cs
```diff
@@ -8,7 +8,7 @@
 namespace Orleans.Transactions.TestKit
 {
     public abstract class TransactionalStateStorageTestRunner<TState> : TransactionTestRunnerBase
-        where TState : class, ITestState, new()
+        where TState : class, IEquatable<TState>, new()
```
JoshCollinsMSFT: >IEquatable<TState> [](start = 30, length = 18)

Again, not needed since using ShouldBeEquivalentTo()

src/Orleans.Transactions.TestkitBase/TestRunners/TransactionalStateStorageTestRunner.cs
```diff
@@ -133,7 +133,7 @@ public virtual async Task WrongEtags()
             loadresponse.PendingStates.Should().BeEmpty();
         }
 
-        private PendingTransactionState<TState> MakePendingState(long seqno, int val, bool tm)
+        private PendingTransactionState<TState> MakePendingState(long seqno, TState val, bool tm)
```
JoshCollinsMSFT: >TState val [](start = 77, length = 10)

You can make this value the seed parameter and just call the state factory within MakePendingState()

test/Transactions/Orleans.Transactions.Azure.Test/AzureTransactionalStateStorageTests.cs
xiazen: will do

src/Orleans.Transactions.Testkit.Xunit/TransactionalStateStorageTestRunner.cs
xiazen: will do

PR #5526, opened 2019-04-11
[WIP] Remove catalog activations locking

ReubenBond:


ReubenBond:
@dotnet-bot test bvt

known flaky test

src/Orleans.Runtime/Catalog/Catalog.cs
```diff
@@ -443,20 +439,19 @@ public void GetGrainTypeInfo(int typeCode, out string grainClass, out PlacementS
             activatedPromise = Task.CompletedTask;
             PlacementStrategy placement;
 
-            lock (activations)
+            while (true)
```
xiazen: this smells bad. why are we adding infinite retry here while the original logic didn't?

jdom:
>                     // We want to do this (RegisterMessageTarget) under the same lock that we tested TryGetActivationData. They both access ActivationDirectory.

no longer true

---
Refers to: src/Orleans.Runtime/Catalog/Catalog.cs:474 in cfd0c4b. [](commit_id = cfd0c4bf8d2f74abfb450b3db1c0af5d39fa6257, deletion_comment = False)

jdom:
>         private void OnSiloStatusChange(SiloAddress updatedSilo, SiloStatus status)

can this be called concurrently or is this single threaded? Because if it's the first option, you'll have to use some kind of locking primitive, not necessarily on `activations`.

---
Refers to: src/Orleans.Runtime/Catalog/Catalog.cs:1383 in cfd0c4b. [](commit_id = cfd0c4bf8d2f74abfb450b3db1c0af5d39fa6257, deletion_comment = False)

src/Orleans.Runtime/Catalog/Catalog.cs
jdom: this is a standard lock-free approach when dealing with optimistic concurrency conflicts

---
In reply to: [274565380](https://github.com/dotnet/orleans/pull/5526#discussion_r274565380) [](ancestors = 274565380)

xiazen: I see. Thanks for explaining !

PR #5527, opened 2019-04-12
Add RabbitMQ as a Stream Provider

mxplusb:
This is my first significant PR, and I think I got some stuff wrong, but I wanted to submit it as a draft and work through the feedback while it's still a WIP.

## RabbitMQ Stream Provider

This implements all of the Stream Provider interfaces for formal RabbitMQ support. My team and I (@zednis, et al) have a use case for RabbitMQ, and since I were going to implement it for our use cases, it seemed better to donate the effort for others.

### TODO

I need to address these items before it's ready for a more formal review. Also, please let me know if there's anything major I'm missing that's not covered under the list below.

- [ ] Write tests

I used the GCP PubSub provider as the reference item, and I saw the tests in the original PR, but they've moved locations and I cannot find them. I will also need a way to require RabbitMQ to be available so the test suite can be run. I would love some advice in this area because I'd like for folks to be able to do local testing. RabbitMQ server is licensed with the Mozilla Public License 1.1, so there should be no distribution concerns when working with the RabbitMQ server binaries.

- [ ] Pull design pattern

The C# RabbitMQ client is designed to be very simple with a totally hidden implementation. As part of that it, when I was working through the implementation, it felt really clunky and didn't seem to mesh well with the current provider interface designs for Stream Providers. I would love to hear some feedback on the design pattern that I used to make sure it feels _right_ because it feels a bit wrong to me. Some of this is likely because of the RabbitMQ client implementation, but I'd love some more constructive feedback on the topic.

- [ ] Pushing API?

Are there any plans for a more asynchronous queue layer? RabbitMQ is a provider that would benefit from that as there is first class support for pushing messages to the client.

- [ ] AMQP Interfaces?

Are there any plans for an interface that's a bit closer to the core AMQP design patterns? It would've been super useful to be able to have backend primitives/interfaces that were much more closely aligned with Exchanges, Queues, Dead Lettering, etc.

- [ ] Nuget Package

I need to make a Nuget package out of this, but I'm not sure how that works.

- [ ] Error Codes?

The GCP provider has error codes. I didn't implement any specific error codes, is that something I need to do? If so, is there guidance on how to go about that?

- [ ] Namespace

I didn't know what to put for the namespace so I put `Orleans.Providers.RabbitMQ.Streams.RabbitMQ`. There's no real reason (from my perspective) you can't use RabbitMQ for cluster membership and health checks, so it could be used in other ways. RabbitMQ is owned by my former employer, Pivotal, but I do not represent them and I didn't feel comfortable with `Orleans.Providers.Pivotal.Streams.RabbitMQ` at the time. I'm happy to fix the namespacing to whatever the group things.

Edit: namespace comments.


src/Orleans.Streaming.RabbitMQ/Providers/Streams/RabbitMQ/RabbitMQAdapter.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        /// <summary>
+        /// Sends messages to RabbitMQ.
+        /// </summary>
+        public Task QueueMessageBatchAsync<T>(Guid streamGuid, string streamNamespace, IEnumerable<T> events, StreamSequenceToken token, Dictionary<string, object> requestContext)
+        {
+            var queueId = _queueMapper.GetQueueForStream(streamGuid, streamNamespace);
+            var logger = _loggerFactory.CreateLogger($"{this.GetType().FullName}.{queueId}");
+            var manager = new RabbitMQManager(_options, logger);
```
justmara: Manager is created from scratch (with all the connection open, exchange/queue setup logics) for every message batch?

src/Orleans.Streaming.RabbitMQ/Providers/Streams/RabbitMQ/RabbitMQAdapterReceiver.cs
```diff
@@ -0,0 +1,101 @@
...
+            var colonyRef = _manager;
+            if (colonyRef == null) return Task.FromResult<IList<IBatchContainer>>(null);
+
+            var upstream = new List<IBatchContainer>();
+
+            // rabbitmq does not have a pullN interface, so we just loop through how many messages we want to pull.
+            for (var i = 0; i < maxCount; i++)
+            {
+                var msg = colonyRef.ReceiveMessage();
+                var container = _dataAdapter.FromPullResponseMessage(msg, _lastReadMessage++);
```
justmara: msg will be `null` if there are no messages in queue (or when `maxCount` is greater than rabbitmq's Prefetch Limit).

justmara:
Did you look at existing RabbitMQ provider's realizations? There are at least two already (both looks abandoned and half-done though, so we decided to not use'em in production):

1. https://github.com/OrleansContrib/Orleans.Providers.RabbitMQ
2. https://github.com/zitmen/Orleans.Streams.RabbitMqStreamProvider

src/Orleans.Streaming.RabbitMQ/Providers/Streams/RabbitMQ/RabbitMQAdapter.cs
mxplusb: This is where I was a bit confused on how the Stream provider worked internally. I wasn't sure if it was a one-time setup and then it polled for messages or it was instantiated on every stream. If that's the case, I can likely slim some of this down a bit, I'll just need to figure out a better configuration story so it's a configure once situation. Any advice on how to do that?

mxplusb: Ah, I see how it was [done](https://github.com/zitmen/Orleans.Streams.RabbitMqStreamProvider/blob/master/Orleans.Streams.RabbitMqStreamProvider/SiloBuilderExtensions.cs).

src/Orleans.Streaming.RabbitMQ/Providers/Streams/RabbitMQ/RabbitMQAdapterReceiver.cs
mxplusb: I currently have the logic set to acknowledge every message received, so there shouldn't be too much concern on hitting the prefetch limit (at least based on my understanding of prefetching in Rabbit) - I can definitely document that when I figure out where to put it. It's conservative, but it's safer for clients, and the `basic.qos` tag can be set on the RabbitMQ side to adjust as needed.

mxplusb:
@justmara I didn't know those existed, but I reviewed them now and I do agree they look abandoned and half-finished, so ideally this can solve that problem. My team's use cases plus yours and the other two contrib attempts prove there is a need for a RabbitMQ provider, so hopefully this is the one that can start working the problem in a consistently supported matter.

I figure if this PR gets accepted I've signed myself up as the maintainer of this bit of code, which is fine, so at least folks would have a more formally supported implementation that keeps up with changes and issues.

jason-bragg:
Some thoughts.

The two stream providers I'd suggest modeling off of are either azure queue (AQ) stream provider or eventhub (EH) stream provider.  The main difference here is in error handling and ordering guarantees.  If you want recoverable reliably ordered streams you'd need an underlying queue that supports that and should model off of EH.  Otherwise best effort delivery (like AQ) is probably best.  I don't know if RabbitMQ supports ordered reliable data as EH, or is more like AQ.

With 2.0 we've moved to a more componentize architecture for stream providers and have entirely dropped the original provider model.  This transition isn't complete but if you're building a new provider it's probably good to know the direction we're going.  Stream provider adapters have several sub-components like receivers, message caches, data adapters, and the like.  The model we're moving towards is for all these sub-components to be obtained from the container by name, using the same name as the provider.  In this way each can be setup as a named service in the container, and application developers have the ability to provide their own behavior and configuration for each component.  A good case for this in your provider is the data adapter.

Additionally, initialization of stream providers now use the silo lifecycle rather than the legacy provider model.  This probably doesn't affect your development much, but it may be good to understand if there are one-time initialization steps needed for your provider.

While I always encourage community contributions, we've a small team and have had difficulties maintaining contributions that we don't use and can't test properly (like GCP, and SQS).  So my preference for this change would be for it to be added to contrib rather than our main repo.  That's just my preference though, rest of team may feel otherwise. :)  The biggest issue with this is testing.  We've several 'test runners' for streams that can be run against any stream provider, but these test runners aren't packaged (nuget?) in a way that would allow them to be used from contrib.

mxplusb:
Thanks for the feedback, I will go ahead and take that action!

andersjonsson:
Is this effort still ongoing?

We would like to have persistent Orleans streams, on-premise, and RabbitMQ sounds like a good candidate. 
I guess Kafka is also an option, but Kafka seems to have a steeper learning curve and I'm not sure if it's the right choice for us. We're "just" looking for a reliable work/event queue.

I haven't worked with RabbitMQ before, but if I can be of any assistance please let me know

mxplusb:
@andersjonsson I ended up dropping this work stream because I wasn't able to sanely (re: sustainably) map RabbitMQ AMQP constructs to Orleans stream provider constructs. The Orleans internal documentation wasn't very good and/or consistent, so it made implementing this very difficult. I think my biggest pain point was around sharding, since RabbitMQ doesn't have a consistent or native sharding construct. It does support sharding through plugins, but it's only through exchanges, and I wasn't able to map exchanges to Orleans internal constructs well enough.

I think the ultimate problem (as I voiced in #5603) was that Orleans Streams seemed very purpose built for a single provider, and everything else feels bolted on. Personally, I would've much rathered Orleans Streams have an AMQP-driven interface, where you implement a layered CRUD interface (for messages, queues, and exchanges) on top of a generic client interface to control the interface layers.

Due to these difficulties I decided I will build the abstraction layer as a message router, which does exactly what I described as my preferred solution, and then calls a Grain the normal way.

PR #5530, opened 2019-04-12
Add codegen error for non-awaitable grain interface methods

ReubenBond:


ReubenBond:
This could be a Roslyn analyzer instead/also

ReubenBond:
@benjaminpetit pls

PR #5539, opened 2019-04-16
Dropped fluent support for Named Service Configurator

jason-bragg:
Note: this will likely break our load and bc tests.  Working on fixes.


ReubenBond:
@jason-bragg I'll merge when you're ready

Issue #5543, opened 2019-04-23
Cannot resolve scoped service 'Microsoft.Extensions.Options.IOptionsSnapshot`1[Orleans.Configuration.AzureBlobStorageOptions]' from root provider.


Alxandr:
I enabled validation of services on the service provider, and this gives me an `InvalidOperationException` with the message `Cannot resolve scoped service 'Microsoft.Extensions.Options.IOptionsSnapshot`1[Orleans.Configuration.AzureBlobStorageOptions]' from root provider.`.

The issue stems from the `GetService` call here: https://github.com/dotnet/orleans/blob/master/src/Azure/Orleans.Persistence.AzureStorage/Hosting/AzureBlobSiloBuilderExtensions.cs#L110.

It seems that calling `GetService(MyScopedServiceType)` on a root service provider is disallowed. Any reason this is using a factory function, and not just using normal constructor injection?

Alxandr:
Replacing with `IOptionsMonitor` solves the problem.

wanton7:
I'm getting similar error ```Microsoft.Extensions.Options.IOptionsSnapshot`1[Orleans.Configuration.AdoNetGrainStorageOptions]``` when trying to use AddAdoNetGrainStorageAsDefault with ASP.NET Core 3.0 preview4.

Alxandr:
@wanton7 yeah, there is the same error ~20 places in the code.

PR #5545, opened 2019-04-23
Fix DI scope issue in azure blob

Alxandr:
Closes: #5543

### DIFF
--- a/src/AWS/Orleans.Persistence.DynamoDB/Hosting/DynamoDBSiloBuilderExtensions.cs
+++ b/src/AWS/Orleans.Persistence.DynamoDB/Hosting/DynamoDBSiloBuilderExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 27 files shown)


Alxandr:
Turns out this happens more places. Will get it working and update PR.

Alxandr:
Are there flacky timings in the functional tests? Cause I can't see how I broke any of the tests that have failed.

xiazen:
Yes sometimes. I think your PR is fine, since both bvt and functional passed. The AzureOps one are new, and run identical tests as bvt and functional. They are more flacky.

xiazen:
Thanks for the contribution! although I think the right fix here is to use constructor injection, instead of resolving it directly from DI. We probably just missed those cases when we move our configurator to use Options. If you can help fix that, it would be great!!

Thanks again :)

Alxandr:
Some functional tests should probably be made/updated to use the strict DI scope checks to see if there are more of these errors btw. I just caught the `IOptionsSnapshot` one.

xiazen:
Yes you are right. I will put out a PR for it.

Issue #5560, opened 2019-04-26
Client does not close connection to a dead Silo


benjaminpetit:
If a Silo stop responding to other silos for some reason, it will be voted `Dead`. The silo in question will then shutdown itself, closing all sockets.

Clients does not check for this status, and instead just get the list of active gateways on a regular basis. If a silo has been voted `Dead`, but for some reason its socket are still open, clients will continue to send message to this faulty silo.

sergeybykov:
Resolved via #5561.

PR #5561, opened 2019-04-26
On the client, close gateway connection to dead silos

benjaminpetit:
Fix for #5560

### DIFF
--- a/src/Orleans.Core/Messaging/GatewayManager.cs
+++ b/src/Orleans.Core/Messaging/GatewayManager.cs
@@ -280,7 +280,7 @@ namespace Orleans.Messaging
                 }
             }
 
-            this.messageCenter.CleanupGatewayConnections(cachedLiveGateways);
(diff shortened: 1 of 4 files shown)


benjaminpetit:
Jenkins build are failing because of disk space issues

src/Orleans.Core/Messaging/GatewayManager.cs
```diff
@@ -274,6 +279,8 @@ private void UpdateLiveGatewaysSnapshot(IEnumerable<Uri> currentKnownGateways, T
                             prevRefresh);
                 }
             }
+
+            this.messageCenter.CleanupGatewayConnections(cachedLiveGateways);
```
ReubenBond: Given that `null` is being passed in for `messageCenter` in the `GatewaySelectionTest.cs` file, should this be a `?.` call instead?

ReubenBond:
Made 1 trivial change. It's a little bit disconcerting that we never clean up the `ClientMessageCenter.gatewayConnections` dictionary while the client is alive.

PR #5567, opened 2019-05-01
Fix ordered list with incorrect indentation.

st0le:


st0le:
In markdown, 4 spaces is code.

PR #5570, opened 2019-05-02
Prevent NullReferenceException with some storage providers when state is Nullable<T>

### LINKED ISSUE #5569: Possible NullReferenceException in AzureTableStorage.ReadStateAsync for Grain<Nullable<T>>
If `Nullable<T>` is used for grain state type, the following line will throw a NullReference Exception: https://github.com/dotnet/orleans/blob/v2.3.1/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureTableStorage.cs#L83

This would happen as `StateStorageBridge<TState>` instantiates it as `new GrainState<TState>(new TState())`, and when `TState` is e.g. `int?`, then `(new int?()) == null)`.

I'm going to submit a PR that fixes that shortly.

ilyalukyanov:
Fixes #5569 

The change has been tested in a real cluster.


sergeybykov:
Thank you, @ilyalukyanov!

PR #5574, opened 2019-05-03
[WIP] Remove base class from [Client,Silo]MessagingOptions

ReubenBond:
This separates the `MessagingOptions` base class from `ClientMessagingOptions` and `SiloMessagingOptions`.

This is a breaking change: users may have to fix some code so that their application continues to compile.

Effectively subsumes #4753.

### DIFF
--- a/src/Orleans.Core.Legacy/Configuration/LegacyConfigurationExtensions.cs
+++ b/src/Orleans.Core.Legacy/Configuration/LegacyConfigurationExtensions.cs
@@ -49,11 +49,15 @@ namespace Orleans.Configuration
             // Translate legacy configuration to new Options
             services.Configure<ClientMessagingOptions>(options =>
             {
-                CopyCommonMessagingOptions(configuration, options);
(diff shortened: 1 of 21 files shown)


xiazen:
This will require users to configure two options class(MessagingOptions, and Client/SiloMessagingOptions), instead of one. 

This is better than before, But I still want to bring up the options to have just Client/SiloMessagingOptions, and copy fields in MessagingOptions over to them. I'd like to hear your thought on this.

ReubenBond:
Having two entirely separate classes which happen to have the same properties means that we cannot have common classes which use those properties. 

Eg, for `ResponseTimeout` we would need to find some uniform way to access that.

We could potentially have an internal "IMessagingOptions" which both classes implement and then register IMessagingOptions in the container based upon `IOptions<ClientMessagingOptions>`, etc.

We could alternatively register it as `IOptions<IMessagingOptions>` and still implement it using the concrete classes.

PR #5580, opened 2019-05-07
Refactor EventHubDataAdapter to be plugable

jason-bragg:
Attempting to address "Stream data adapter #4621" for EventHub streams.
EventHubDataAdapter is now pluggable.  It can now be configured without customizing queue adapter or cache.

Other changes:
- Simplified pooled cache to only use CachedMessage
- Simplified eviction strategy to only use CachedMessage
- Removed the need for custom comparer logic.
- Added interned stream namespace to CachedMessage to prevent the need to encode it into pooled memory segment.
- Refactored IQueueDataAdapter to support common send behaviors.

Overall, pooled cache pattern is much cleaner and maintainable, at the cost of some customizability.


src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/IAzureQueueDataAdapter.cs
```diff
@@ -73,7 +73,7 @@ public AzureQueueDataAdapterV2(SerializationManager serializationManager)
         /// <summary>
         /// Creates a cloud queue message from stream event data.
         /// </summary>
-        public CloudQueueMessage ToQueueMessage<T>(Guid streamGuid, string streamNamespace, IEnumerable<T> events, Dictionary<string, object> requestContext)
+        public CloudQueueMessage ToQueueMessage<T>(Guid streamGuid, string streamNamespace, IEnumerable<T> events, StreamSequenceToken token, Dictionary<string, object> requestContext)
```
xiazen: Is this AzureQueueDataAdaptor related change to this pr?

src/Orleans.Core.Abstractions/Streams/Core/StreamSequenceToken.cs
```diff
@@ -10,6 +10,16 @@ namespace Orleans.Streams
     [Serializable]
     public abstract class StreamSequenceToken : IEquatable<StreamSequenceToken>, IComparable<StreamSequenceToken>
     {
+        /// <summary>
+        /// Number of event batches in stream prior to this event batch
+        /// </summary>
+        public long SequenceNumber { get; protected set; }
```
xiazen: what does this mean to other persistent stream providers?

xiazen: what does this mean to other stream providers which doesn't support rewindability

src/Azure/Orleans.Streaming.AzureStorage/Providers/Streams/AzureQueue/IAzureQueueDataAdapter.cs
jason-bragg: Yes, unfortunately.  The interface surface changed, thus so need the implementations.

src/Orleans.Core.Abstractions/Streams/Core/StreamSequenceToken.cs
jason-bragg: It means that ordering information needs be expressed in a long value.

This could be a problem if systems have more complex or convoluted ordering information.

It could also be an issue if provider implementations have persisted tokens that don't fit this.  They may need to perform some translation between the old token and a new compatible one.  This should not be an issue for any we maintain, but custom tokens may encounter some issues.

jason-bragg: It shouldn't matter for those, as they shouldn't rely on tokens.  In those cases, this information is only used to order data in the cache for delivery.

jason-bragg: The cache ordering requirement already existed even for non-rewindable stream providers, thus this is not a new constraint, it was just not expressed in the token.

xiazen:
@jason-bragg LGTM. do you want to run it by back-compact test, and if it went ok, we merge it?

PR #5582, opened 2019-05-07
Fix #5565 - NullReferenceException caused by call filter

### LINKED ISSUE #5565: NullReferenceException when you catch exceptions
Hi,
when you just catch execptions with a filter like:

```
    public sealed class LoggingFilter : IIncomingGrainCallFilter
    {
        private readonly ILogger<LoggingFilter> log;

        public LoggingFilter(ILogger<LoggingFilter> log)
        {
            this.log = log;
        }

        public async Task Invoke(IIncomingGrainCallContext context)
        {
            try
            {
                await context.Invoke();
            }
            catch (Exception ex)
            {
                log.LogError(ex, "foo");
            }
        }
    }
```

you get a NullReferenceException. I think such a filter should be allowed for grains that void result (Task).

ReubenBond:
Fixes #5565

When a `Task` returns `null`, that value cannot be cast to `T` using a hard-cast because that will cause a `NullReferenceException`.

Also adds `ConfigureAwait(false)` inside some private Task helpers for cleanliness.

I am hoping this builds on the build servers, but the `is` syntax I'm using may not be supported there.

cc @SebastianStehle

### DIFF
--- a/src/Orleans.Core/Async/TaskExtensions.cs
+++ b/src/Orleans.Core/Async/TaskExtensions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


ReubenBond:
@dotnet-bot test functional

PR #5436, opened 2019-03-12
Networking system rewrite

ReubenBond:
Replaces the existing networking system with a system based upon work originally done for Kestrel, dubbed [Bedrock](https://github.com/aspnet/AspNetCore/issues/4772).

The core purpose of this is to modernize the networking stack and eventually support implementing TLS (#828). In addition, Orleans gains improved support for non-TCP/non-Socket networking, including the ability to host on Kestrel.

Fixes #1372
Closes #307

PR #5585, opened 2019-05-07
Groundwork for span-friendly serialization

ReubenBond:
This is a pre-requisite for #5436 and lays some groundwork for a more span-friendly serializer (networking replat is based on pipes which use Span<byte>/Memory<byte>)

### DIFF
--- a/src/Orleans.Core.Abstractions/Serialization/IBinaryTokenStreamReader.cs
+++ b/src/Orleans.Core.Abstractions/Serialization/IBinaryTokenStreamReader.cs
@@ -7,19 +7,8 @@ namespace Orleans.Serialization
 {
     public interface IBinaryTokenStreamReader
     {
-        /// <summary>
-        /// Resets this instance with the provided data.
-        /// </summary>
-        /// <param name="buffs">The underlying buffers.</param>
-        void Reset(IList<ArraySegment<byte>> buffs);
-
         /// <summary> Current read position in the stream. </summary>
         int CurrentPosition { get; }
-
-        /// <summary>
-        /// Gets the total length.
-        /// </summary>
-        int Length { get; }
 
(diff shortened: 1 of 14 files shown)


SebastianStehle:
If you use Pipes I would definitely wait for this task: https://github.com/dotnet/corefx/issues/32588

Otherwise you have a lot of fun to deal with ReadOnlySequence<byte> by yourself.

ReubenBond:
I'm basing this off the API & samples in the issue you linked - it's all geared around delimiters.
Either way, we already handle non-contiguous buffers in the serializer, so extending that for `ReadOnlySequence<byte>` was quite simple (see linked PR), eg:

``` C#
[MethodImpl(MethodImplOptions.AggressiveInlining)]
public uint ReadUInt32()
{
    const int width = 4;
    if (this.bufferPos + width > this.bufferSize) return ReadSlower();

    var result = BinaryPrimitives.ReadUInt32LittleEndian(this.currentSpan.Span.Slice(this.bufferPos, width));
    this.bufferPos += width;
    return result;

    uint ReadSlower()
    {
        uint b1 = ReadByte();
        uint b2 = ReadByte();
        uint b3 = ReadByte();
        uint b4 = ReadByte();

        return b1 | (b2 << 8) | (b3 << 16) | (b4 << 24);
    }
}
```

SebastianStehle:
I don't think so. If you have messages that do not fit into one buffer or are just not aligned to the buffer then you can get multiple byte arrays for one message. So you get a message that starts at position 100 at the first buffer and ends at position 50 of buffer 3 as an example. And then it becomes annoying to write efficient deserializers because you have to deal with the offset calculation and you can even not assume anymore that an 4 byte integer is located in a single array. 

The reader should make it much easier.

PR #5586, opened 2019-05-07
Implement IApplicationLifetime for ClientBuilder/SiloHostBuilder

ReubenBond:
This is a pre-requisite for #5436, which uses `IApplicationLifetime` (from shared Hosting package) to ensure all connections are closed during shutdown.

This aligns `ClusterClient` and `SiloHostBuilder` more closely with the generic `HostBuilder` from `Microsoft.Extensions.Hosting`.

### DIFF
--- a/src/Orleans.Core/Core/ClusterClient.cs
+++ b/src/Orleans.Core/Core/ClusterClient.cs
@@ -9,6 +9,8 @@ using Microsoft.Extensions.Options;
 using Orleans.Configuration;
 using Orleans.Runtime;
 using Orleans.Streams;
+using Microsoft.Extensions.Hosting;
+using Orleans.Hosting;
 
 namespace Orleans
 {
@@ -20,6 +22,7 @@ namespace Orleans
         private readonly IRuntimeClient runtimeClient;
         private readonly ClusterClientLifecycle clusterClientLifecycle;
         private readonly AsyncLock initLock = new AsyncLock();
+        private readonly ClientApplicationLifetime applicationLifetime;
         private LifecycleState state = LifecycleState.Created;
 
         private enum LifecycleState
@@ -61,6 +64,9 @@ namespace Orleans
             {
                 participant?.Participate(clusterClientLifecycle);
             }
+
+            // It is fine for this field to be null in the case that the silo is not the host.
+            this.applicationLifetime = runtimeClient.ServiceProvider.GetService<IApplicationLifetime>() as ClientApplicationLifetime;
         }
 
         /// <inheritdoc />
@@ -120,6 +126,8 @@ namespace Orleans
                 await this.clusterClientLifecycle.OnStart().ConfigureAwait(false);
                 this.state = LifecycleState.Started;
             }
+
+            this.applicationLifetime?.NotifyStarted();
         }
 
         /// <inheritdoc />
@@ -131,6 +139,7 @@ namespace Orleans
         private async Task Stop(bool gracefully)
         {
             if (this.IsDisposing) return;
+            this.applicationLifetime?.StopApplication();
             using (await this.initLock.LockAsync().ConfigureAwait(false))
             {
                 if (this.state == LifecycleState.Disposed) return;
@@ -161,6 +170,8 @@ namespace Orleans
                     if (this.state == LifecycleState.Disposing) this.state = LifecycleState.Invalid;
                 }
             }
+
+            this.applicationLifetime?.NotifyStopped();
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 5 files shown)


src/Orleans.Core/Core/ClusterClient.cs
```diff
@@ -20,6 +22,7 @@ internal class ClusterClient : IInternalClusterClient
         private readonly IRuntimeClient runtimeClient;
         private readonly ClusterClientLifecycle clusterClientLifecycle;
         private readonly AsyncLock initLock = new AsyncLock();
+        private readonly ClientApplicationLifetime applicationLifetime;
```
jason-bragg: Why use concrete ClientApplicationLifetime  class when only ApplicationLifetime is necessary?

src/Orleans.Core/Core/ClusterClient.cs
```diff
@@ -61,6 +64,9 @@ public ClusterClient(IRuntimeClient runtimeClient, ILoggerFactory loggerFactory,
             {
                 participant?.Participate(clusterClientLifecycle);
             }
+
+            // It is fine for this field to be null in the case that the silo is not the host.
+            this.applicationLifetime = runtimeClient.ServiceProvider.GetService<IApplicationLifetime>() as ClientApplicationLifetime;
```
jason-bragg: This cast seems error prone.  If we need cast is ApplicationLifetime shouldn't we resolve that, and just register the same instance in container for ApplicationLifetime and IApplicationLifetime

src/Orleans.Runtime/Hosting/Generic/SiloWrapper.cs
```diff
@@ -1,20 +1,26 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 using Orleans.Runtime;
 
 namespace Orleans.Hosting
 {
     internal class SiloWrapper : ISiloHost
     {
         private readonly Silo silo;
+        private readonly SiloApplicationLifetime applicationLifetime;
```
jason-bragg: Same questions as for ClientApplicationLifetime

src/Orleans.Core/Core/ClusterClient.cs
ReubenBond: Just to make it very clear that this will only be set when the host is the client

src/Orleans.Core/Core/ClusterClient.cs
ReubenBond: This is very intentional: we want the `IApplicationLifetime` implementation registered in the container but only if it's implemented by `ClientApplicationLifetime`. i.e, the lifetime is optional. We do not want to fire lifetime events if the host wasn't created by `ClientBuilder`. See above comment.

src/Orleans.Core/Hosting/ApplicationLifetime.cs
```diff
@@ -0,0 +1,119 @@
...
+            // which will no-op since the first call already requested cancellation, get a chance to execute.
+            lock (this.stoppingSource)
+            {
+                try
+                {
+                    this.ExecuteHandlers(this.stoppingSource);
+                }
+                catch (Exception ex)
+                {
+                    this.logger.LogError("An error occurred stopping the application", ex);
```
ifle: Instead of
```c#
this.logger.LogError("An error occurred stopping the application", ex);
```
Should be
```c#
this.logger.LogError(ex, "An error occurred stopping the application");
```

src/Orleans.Core/Hosting/ApplicationLifetime.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// </summary>
+        public void NotifyStarted()
+        {
+            try
+            {
+                this.ExecuteHandlers(this.startedSource);
+            }
+            catch (Exception ex)
+            {
+                this.logger.LogError("An error occurred starting the application", ex);
```
ifle: Instead of
```c#
this.logger.LogError("An error occurred starting the application", ex);
```
Should be
```c#
this.logger.LogError(ex, "An error occurred starting the application");
```

src/Orleans.Core/Hosting/ApplicationLifetime.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// </summary>
+        public void NotifyStopped()
+        {
+            try
+            {
+                this.ExecuteHandlers(this.stoppedSource);
+            }
+            catch (Exception ex)
+            {
+                this.logger.LogError("An error occurred stopping the application", ex);
```
ifle: Instead of
```c#
this.logger.LogError("An error occurred stopping the application", ex);
```
Should be
```c#
this.logger.LogError(ex, "An error occurred stopping the application");
```

PR #5587, opened 2019-05-08
Allow default(ImmutableArray<T>) to be serialized

Liversage:
`ImmutableArray<T>` is a value type and the default value will throw `NullReferenceException` when the `Length` property is read. The serializer will read this property to determine how many elements in the array should be written to the stream making it impossible to serialize `default(ImmutableArray<T>)`.

This change ensures that `default(ImmutableArray<T>)` can be serialized without throwing a `NullReferenceException`. The value `default(ImmutableArray<T>)` is indicated by writing `-1` as the length of the array to the stream. Note that `ImmutableArray<T>.Empty` has a length of `0` so the code is able to distinguish between the two values.

The deserializer will make a call to `context.RecordObject`. I do not quite understand the purpose of this call and if it is meaningful to do with a value type but I included it in the added code because it was already in the existing code.

I have added a test that verifies that `default(ImmutableArray<T>)` can round-trip when serialized.

### DIFF
--- a/src/Orleans.Core/Serialization/BuiltInTypes.cs
+++ b/src/Orleans.Core/Serialization/BuiltInTypes.cs
@@ -1105,28 +1105,41 @@ namespace Orleans.Serialization
 
         internal static void SerializeImmutableArray<K>(object untypedInput, ISerializationContext context, Type typeExpected)
         {
-            var dict = (ImmutableArray<K>)untypedInput;
+            var array = (ImmutableArray<K>)untypedInput;
 
-            context.StreamWriter.Write(dict.Length);
-            foreach (var pair in dict)
+            if (array == default)
             {
-                SerializationManager.SerializeInner(pair, context, typeof(K));
+                context.StreamWriter.Write(-1);
+                return;
+            }
+
+            context.StreamWriter.Write(array.Length);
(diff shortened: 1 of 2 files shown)


Liversage:
The **functional** check had a failed test:

[`UnitTests.ActivationsLifeCycleTests.ActivationCollectorTests.ActivationCollectorShouldNotCollectBusyActivationsSpecifiedInPerTypeConfiguration`](https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/2816/testReport/junit/UnitTests.ActivationsLifeCycleTests/ActivationCollectorTests/ActivationCollectorShouldNotCollectBusyActivationsSpecifiedInPerTypeConfiguration/)

with the following exception message:

> System.Net.Sockets.SocketException : An attempt was made to access a socket in a way forbidden by its access permissions

It seems unlikely to me that my change caused a regression in this test but any advice on how to resolve it will be appreciated.

ReubenBond:
@Liversage that exception message is prevalent on those old Jenkins machines - I think something kills the job/sockets sometimes. Nothing to worry about.

ReubenBond:
Thank you, @Liversage!

PR #5588, opened 2019-05-08
Execute tasks scheduled against defunct activations

ReubenBond:
Currently, `Task`s which are scheduled against an activation's `TaskScheduler` will be ignored if that activation has become defunct. This is generally not a problem and ignoring these tasks is fine (eg, it could be a non-awaited task from a grain method call). However, there are some cases where not executing these tasks can lead to larger issues. For example, we have encountered a library which captures an activation's `TaskScheduler` and use it for servicing a work queue.

This PR changes the behavior of `WorkItemGroup` so that `Task`s scheduled after an activation has terminated are still executed.

A warning will still be emitted and that warning will be upgraded to an error after 1 minute (configurable via `SchedulingOptions.StoppedWorkItemGroupWarningInterval`). The error will then be logged at most once per minute per activation.

This allows systems to continue functioning when a library misbehaves and gives the application developer more tools to find the misbehaving code.

Since this allows grain code to continue executing after being deactivated (eg, due to an end-user programming error), this PR also adds additional runtime checks so that invalid activations cannot access runtime services as easily.

### DIFF
--- a/src/Orleans.Runtime/Core/GrainRuntime.cs
+++ b/src/Orleans.Runtime/Core/GrainRuntime.cs
@@ -107,6 +107,7 @@ namespace Orleans.Runtime
 
             if (context == null)
             {
+                // Move exceptions into local functions to help inlining this method.
                 ThrowMissingContext();
                 void ThrowMissingContext() => throw new InvalidOperationException("Activation access violation. A non-activation thread attempted to access activation services.");
             }
@@ -115,6 +116,7 @@ namespace Orleans.Runtime
                 && schedulingContext.Activation is ActivationData activation
                 && activation.State == ActivationState.Invalid)
             {
+                // Move exceptions into local functions to help inlining this method.
                 ThrowInvalidActivation(activation);
                 void ThrowInvalidActivation(ActivationData activationData) => throw new InvalidOperationException($"Attempt to access an invalid activation: {activationData}");
             }
--- a/src/Orleans.Runtime/Core/GrainRuntime.cs
+++ b/src/Orleans.Runtime/Core/GrainRuntime.cs
@@ -5,6 +5,8 @@ using Orleans.Configuration;
 using Orleans.Core;
 using Orleans.Timers;
 using Orleans.Storage;
+using Orleans.Runtime.Scheduler;
+using System.Runtime.CompilerServices;
 
 namespace Orleans.Runtime
 {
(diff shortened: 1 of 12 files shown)


src/Orleans.Runtime/Core/GrainRuntime.cs
```diff
@@ -99,11 +101,22 @@ public IStorage<TGrainState> GetStorage<TGrainState>(Grain grain) where TGrainSt
...
 
-        private static void CheckRuntimeContext()
+        public static void CheckRuntimeContext()
         {
-            if (RuntimeContext.Current == null)
+            var context = RuntimeContext.Current;
+
+            if (context == null)
+            {
+                ThrowMissingContext();
```
benjaminpetit: Minor but I think we should add a comment on why you are declaring a function here

ReubenBond:
Let's merge it

src/Orleans.Runtime/Core/GrainRuntime.cs
ReubenBond: It's for the JIT only, to allow inlining. I think once people know this pattern then the comment at every usage is redundant. For those that don't know the pattern, the usage is harmless if a little verbose.

It's mentioned in this post: https://reubenbond.github.io/posts/dotnet-perf-tuning

veikkoeeva: I wonder if some sort of a `ThrowHelper` should be added and extensively used. I also wonder if there's something to grab (otherwise than inspiration) from CoreFx. Just throwing this ideas here.

veikkoeeva: https://github.com/dotnet/corefx/pull/35213 via https://github.com/dotnet/corefx/pull/35213

> Across .NET Core, we’re also paying more attention to assembly size, in particular as it can impact ahead-of-time (AOT) compilation. PRs like dotnet/corefx#35213, which employs “ThrowHelpers” in the heavily-generic LINQ, help to reduce generated code size, which has benefits in and of itself but can also help with other areas of performance.

pentp:
Does this fix https://github.com/Azure/azure-cosmos-dotnet-v2/issues/687?

ReubenBond:
In a way, @pentp. The behavior of capturing `TaskScheduler.Current` by CosmosDB still needs to be fixed to use `TaskScheduler.Default`, but at least this allows it to continue operating.

PR #5589, opened 2019-05-08
Add Analyzers to Orleans

mholo65:
- Setup basic Analyzer project
- Setup basic Tests project
- Create analyzer for AlwaysInterleaveAttribute

### DIFF
--- a/Orleans.sln
+++ b/Orleans.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2036
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28803.202
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{4CD3AA9E-D937-48CA-BB6C-158E12257D23}"
 EndProject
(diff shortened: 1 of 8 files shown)


mholo65:
Basics for Analyzers.

Are there any docs on how to create new projects? It seems that if I try to copy samples from the other projects (e.g. removing `<TargetFramworks>` from .csproj), then Visual Studio will convert my project to a legacy project. Also, where should a new test project be hooked up in order to be run by CI? Is it sufficient to add a new line to `Test.cmd`?

ping @ReubenBond

ReubenBond:
I pushed some changes, @mholo65 & I rebased on top of master

ReubenBond:
I'll merge when tests complete - your tests are running in CI now.

ReubenBond:
Thank you, @mholo65! 🎉

mholo65:
Thanks @ReubenBond! In order to make use of Analyzers by default, it might be wise to have `Orleans.Core.Abstractions` depend on `Orleans.Analyzers`. This is how EF does it. https://github.com/aspnet/EntityFrameworkCore/blob/99fb2d4e3d52f8a3be67fe6815a4fc288e63fe50/src/EFCore/EFCore.csproj#L43

jdom:
Notice there that it's being added to EFCore, not EFCore.Abstractions. We shouldn't be adding a dependency to the abstractions project that will make it release with a faster cadence, as we've been there already, and it's not fun

ReubenBond:
I'm not sure where to add it just yet. Grain/grain interface projects only require a reference to the .Abstractions project. We could also add it to the CodeGenerator package, but that feels awkward

PR #5592, opened 2019-05-09
Fix NRE in TestCluster.cs

MelnikovIG:
```await this.InternalClient?.Close();``` and similar causes NRE when ```InternalClient``` is null, because we are trying to await null Task.


ReubenBond:
Thank you, @MelnikovIG

PR #5594, opened 2019-05-09
Rebuild when files are removed from a project

ReubenBond:
This PR targets the reflection-based code generator.

When a file is removed from a project, we should re-run code generation.
We should also ignore the generated `$(ProjectName).orleans.g.cs` file when determining whether or not to run code generation.

The focus of the PR is the added text, but I also updated the naming conventions to be inline with the Roslyn-based code generator's targets file.


ReubenBond:
@dotnet-bot test functional

PR #5608, opened 2019-05-13
Fix ReadLineStartingWithAsync for LinuxEnvironmentStatistics

martinothamar:
Don't know how I missed this (been using a different implementation internally).
Has coincidentally worked for `MemTotal` since it's at the top of the file but doesnt work for `MemAvailable`.

Sorry about this


sergeybykov:
Thank you, @martinothamar!

PR #5618, opened 2019-05-19
[WIP] [DRAFT] Microsoft Faster Integration Sample

JorgeCandeias:
This new sample demonstrates rudimentary integration with [Microsoft Faster](https://github.com/Microsoft/FASTER) within the context of a grain, for the purpose of automatic partitioning of massive lookups. The idea is to use Orleans grains to provide state sharding across a cluster, while using Faster to provide the disk spilling lookup.

This sample is a work-in-progress. Integration is proving challenging due to the conflicting designs between Orleans' and Faster's thread concurrency models, which work well on their own but will compete for resources when brought together.

The sample contains two approaches at the moment:

* A grain that starts and stops ad-hoc Faster sessions on the Thread Pool as needed.
* A grain that creates a fixed number of affinitized threads and uses a rudimentary command model to interact with them.

I'm yet to try a shared GrainService or to take the Faster logic out of the silo altogether as that defeats the objective of automatic partitioning.

I've asked @badrishc from the Faster project to help review this and see if there are opportunities for improving things.


badrishc:
I will leave others to comment on the Orleans side of the integration for now, but from the point of view of how FASTER is being used, I think this is a good starting point.

As background, the basic FASTER model is based on a set of threads, each of which can start a session with FASTER, perform a sequence of operations, and later stop the session. Operations consist of bursts of activity, with periodic invocations of a Refresh() call to FASTER in between. It would _not_ be okay for a thread/session to go to sleep (i.e., stop calling Refresh) without stopping its session.

This pattern matches well with the fixed-threads version of your samples. However, this does raise questions on resource sharing with Orleans grains threads. For scale out, I assume the idea is to have a FASTER grain for each (coarse-grained) hash bucket range of the hash space. Then, Orleans may schedule these "state" grains across the cluster as it wishes. Key requests would then need to be routed by Orleans to the appropriate grains in order to be served.

The ad-hoc session example you have created is interesting as well because it interacts better with the async/await thread pool model, but I am unsure of the performance hit with short lived sessions. Perhaps this will not be a problem if the bottlenecks are elsewhere. Related, I am starting an effort to add async/await support in FASTER, at https://github.com/microsoft/FASTER/pull/130.

FASTER supports a periodic checkpointing/commit mechanism, that would be interesting to parameterize and expose. In other words, users may ask us to persist every e.g., 1 or 10 seconds, and this would correspond to a Checkpoint operation (across all threads) every 10 seconds, giving users (incremental) state persistence until that point. Read more about the incremental commit mechanism used in FASTER [here](https://www.microsoft.com/en-us/research/uploads/prod/2019/01/cpr-sigmod19.pdf).

Finally, would it make sense to use the [storage provider](https://dotnet.github.io/orleans/Documentation/grains/grain_persistence/index.html) notion of Orleans, i.e., create a FASTER storage provider, as an alternative to these approaches? It's not clear if this makes sense, based on the documentation. For example, with the storage provider model, who would be responsible for moving keys to the new silo when a grain migrates?

Samples/2.3/FasterSample/src/Grains/FasterThreadPoolGrain.cs
```diff
@@ -0,0 +1,408 @@
...
+                    LookupItem input = null;
+                    LookupItem output = null;
+                    switch (lookup.Read(ref _key, ref input, ref output, Empty.Default, 0))
+                    {
+                        case Status.OK:
+                            result.Add(output);
+                            break;
+
+                        case Status.PENDING:
+                            if (lookup.CompletePending(true))
```
badrishc: What you have is correct. However, you don't have to call CompletePending immediately after every operation in the batch. This is because CompletePending(true) spin-waits until all pending requests until then are completed. Thus all operations become synchronous.

Instead, you can simply call CompletePending(true) once after foreach, in order to complete all pending operations until that point.

Note that the pending operation results are reported via callback code (such as Functions.ReadCompletionCallback) during CompletePending(). You would therefore need to store the results array and the correct offset in the results array inside Context, so that ReadCompletionCallback can populate the result at the correct position in the result array.

Samples/2.3/FasterSample/src/Grains/FasterThreadPoolGrain.cs
```diff
@@ -0,0 +1,408 @@
...
+        public Task<LookupItem> TryGetAsync(int key) => Task.Run(async () =>
+        {
+            await semaphore.WaitAsync();
+
+            var session = Guid.Empty;
+            try
+            {
+                session = lookup.StartSession();
+                LookupItem result = null;
+                if (lookup.Read(ref key, ref result, ref result, Empty.Default, 0) == Status.ERROR)
```
badrishc: A Read operation may return Status.PENDING, in which case we need to call CompletePending(true) in the same session/thread, either now or later on. If you call CompletePending(true) immediately after the Read, it will result in the operation forced to complete synchronously (the call will spin-wait for completion).

A different technique is to simulate the async/await model. To do this, you can associate the Read operation with a `TContext` of a `TaskCompletionSource<TOutput>` instead of `Empty`. Further, let the ReadCompletionCallback be responsible for setting this task as completed. You return this task to the user in the lookup path so they can await on it. We would still need to call CompletePending(false) on the session periodically to ensure that the I/O completions are processed and callbacks issued.

badrishc: ### More on I/O and CompletePending

When a FASTER operation performs I/O, it returns immediately to the calling thread with a Status.PENDING. The asynchronous I/O completion callback (on the thread pool) eventually enqueues the operation result into a pending queue of completed operations, that is _local to the original session/thread_. The user is responsible for later calling `CompletePending` in order to drain this queue and issue the actual callbacks such as `ReadCompletionCallback`. This model allows threads to perform callback work locally on their own thread, and lets users control when the callback work occurs, e.g., after a batch of requests are issued.

Also, `CompletePending(false)` simply dequeues and processes callbacks for whatever operations have completed until that point - it does not wait. OTOH, `CompletePending(true)` spin-waits until all pending operations until that point are completed and dequeued by the thread.

Samples/2.3/FasterSample/src/Grains/FasterThreadPoolGrain.cs
JorgeCandeias: This grain now uses a `TaskCompletionSource<LookupItem>` as the operation context. Either the grain or the callback will resolve that completion, depending on whether the operation completed synchronously or asynchronously. The trade-off here is more allocations. I could over-engineer this with object pools and value tasks but I think there's more merit in changing the underlying operations so they are async themselves as per https://github.com/microsoft/FASTER/pull/130, so I'll look into that instead.

Samples/2.3/FasterSample/src/Grains/FasterThreadPoolGrain.cs
JorgeCandeias: That makes sense, many thanks for the detailed explanation.

sergeybykov:
@JorgeCandeias Are you still working on this?

JorgeCandeias:
@sergeybykov It's still on my list, just swamped to my eyeballs at the moment. I see @badrishc has been busy committing to https://github.com/microsoft/FASTER/pull/130 but I haven't had a chance to review those changes.

badrishc:
@JorgeCandeias @ReubenBond We now have full async support in FASTER KV, merged to master (https://github.com/microsoft/FASTER/pull/130). The need for periodically calling Refresh has also been eliminated. Would love to hear your comments on using the same.

JorgeCandeias:
Many thanks for this @badrishc and apologies for the late reply. We're going to field test this work of art and I'll come back to this issue to update the sample.

sergeybykov:
@JorgeCandeias Were you able to get back to this?

JorgeCandeias:
@sergeybykov We ended up not using Faster at the time, in favour of a more applicable approach for the client's use case. Faster has its merits regardless in Orleans. Let me come back to this on the weekend.

badrishc:
FASTER has evolved a lot since this first effort. At this point, it behaves exactly like any other standard C# library with respect to threading, making the integration sample much simpler than what we had to do in here. For example, Refresh is no longer needed. Also full async support has been added as well.

We think an Orleans sample at this point will be much easier and behave naturally as well. We support both local SSD and Azure Storage as data backends. Look forward to hearing what u think!

Get it from the new Nuget feed at https://www.nuget.org/packages/Microsoft.FASTER.Core/ or sources from https://github.com/Microsoft/FASTER.

JorgeCandeias:
@badrishc I've now gone through the Faster repository. Indeed it has come a long way, in fact this old sample format can no longer do it justice. I'll do one from scratch to highlight the different scenarios, and will likely bin the current code in the process. @sergeybykov please leave this PR open for a while longer.

oising:
Any more thoughts here, @JorgeCandeias ? I know life can intervene sometimes but I'm eager to try some things in the area myself and what you're doing looks great.

ReubenBond:
Much has changed in FASTER and Orleans. Integrating them should be easier than before, especially if you use Azure Page Blobs as the IDevice implementation. I'll close this since it's been idle for a long time and is still in draft. Many thanks for your contributions, @JorgeCandeias!

JorgeCandeias:
I'm sad I haven't come around on this but it has fallen below other priorities on my list. I do believe I'll come back to this in the context of training material but I think I'll end up covering Trill before covering Faster.

PR #5619, opened 2019-05-20
Create GrainReferenceKeyInfo

### LINKED ISSUE #3810: Possible to make GrainReference and GrainId publicly accessible?
Everything important about GrainReference and GrainId is internal. This is fine for normal use, but really hurts when trying to implement custom storage providers or serializers.

One specific case for which I really need to access the internals of GrainId is to implement grain reference serialization/deserialization in my custom serializer, which is optimized for storage space and having only a ~100 character ToKeyString() result to store is not ideal when I could, under the best circumstances, have between maybe 4 and 20 bytes to store (not counting the possibility of long string keys).

I can just take the source and change the members to be public, but I'm wondering if there is a specific reason they were made internal (other than to simplify the API), and if they can't be made public altogether.

Alxandr:
Create new type GrainReferenceKeyInfo and methods for getting key info from a GrainReference, and getting a GrainReference form the key info. This enables serializers not part of orleans to serialize/deserialize GrainReferences into more space compact representations than the GetKeyString() format.

Resolves #3810.

### DIFF
--- a/src/Orleans.Core.Abstractions/IDs/GrainId.cs
+++ b/src/Orleans.Core.Abstractions/IDs/GrainId.cs
@@ -293,7 +293,7 @@ namespace Orleans.Runtime
         }
 
         /// <summary>
-        /// Return this GrainId ina  standard components form, suitable for later use with the <see cref="FromKeyInfo"/> method.
+        /// Return this GrainId in a standard components form, suitable for later use with the <see cref="FromKeyInfo"/> method.
         /// </summary>
         /// <returns>GrainId in a standard components form.</returns>
         internal (ulong, ulong, ulong, string) ToKeyInfo()
--- a/src/Orleans.Core.Abstractions/IDs/GrainId.cs
+++ b/src/Orleans.Core.Abstractions/IDs/GrainId.cs
@@ -293,6 +293,15 @@ namespace Orleans.Runtime
         }
 
         /// <summary>
+        /// Return this GrainId ina  standard components form, suitable for later use with the <see cref="FromKeyInfo"/> method.
+        /// </summary>
+        /// <returns>GrainId in a standard components form.</returns>
+        internal (ulong, ulong, ulong, string) ToKeyInfo()
+        {
+            return (Key.N0, Key.N1, Key.TypeCodeData, Key.KeyExt);
+        }
+
+        /// <summary>
         /// Create a new GrainId object by parsing string in a standard form returned from <c>ToParsableString</c> method.
         /// </summary>
         /// <param name="grainId">String containing the GrainId info to be parsed.</param>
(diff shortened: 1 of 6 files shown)


src/Orleans.Core.Abstractions/IDs/GrainId.cs
```diff
@@ -292,6 +292,15 @@ internal string ToParsableString()
             return Key.ToHexString();
         }
 
+        /// <summary>
+        /// Return this GrainId ina  standard components form, suitable for later use with the <see cref="FromKeyInfo"/> method.
```
veikkoeeva: Nit: a typo: "ina".

src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
```diff
@@ -388,6 +388,23 @@ public string ToKeyString()
             }
             return String.Format("{0}={1}", GRAIN_REFERENCE_STR, GrainId.ToParsableString());
         }
+
+        public GrainReferenceKeyInfo ToKeyInfo()
```
veikkoeeva: While self-explaining,  I wonder if public methods should have comments on them. Someone else to decide. :)

src/Orleans.Core.Abstractions/Serialization/GrainReferenceKeyInfo.cs
```diff
@@ -0,0 +1,212 @@
+using System;
+using System.Net;
+using Orleans.Runtime;
+
+namespace Orleans.Serialization
+{
+    /// <summary>
+    /// type is a low level representation of graui reference keys to enable
```
veikkoeeva: Nits: Should start with a capital 't', "graui".

src/Orleans.Core.Abstractions/Serialization/GrainReferenceKeyInfo.cs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            get
+            {
+                var category = UniqueKey.GetCategory(Key.Item3);
+                return category == UniqueKey.Category.KeyExtGrain
+                    || category == UniqueKey.Category.GeoClient; // geo clients use the KeyExt string to specify the cluster id
+            }
+        }
+
+        public (Guid key, ulong typeCode) KeyAsGuid()
```
veikkoeeva: Just a note like before that while looking self explanatory, I wonder if should be commented too. Probably not, but for someone else to decide.

src/Orleans.Core.Abstractions/IDs/GrainId.cs
Alxandr: Nice. There is also a double space after `ina` 😛. Thanks for finding these 🙂.

sergeybykov:
@Alxandr Thanks for submitting this PR! 

To clarify, we want to hold it for a while until we sort a few urgent things. Also, for 4.0 (yeah, I know 😊) we are thinking about revamping the grain identity scheme altogether. That's another reason we want to wait a little bit, to see how those things could align. Thanks again!

src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
ReubenBond: They should, yes. We may have build warnings for public, undocumented members

ReubenBond:
LGTM, keeping in mind the points Sergey mentioned above

Alxandr:
@sergeybykov I think I might have misinterpreted parts of what you said originally, so just to clarify:

1. We're waiting a bit with this PR due to more urgent things.
2. For 4.0 (which is probably in a long time?) we plan to revamp grain identity, so this might become obsolete.

Cause I first read that this PR would have to wait for 4.0 :P

sergeybykov:
@Alxandr Correct. The point I was trying to make is that in the 4.0 time frame this will likely go away. But that's okay.

sergeybykov:
Thank you, @Alxandr!

PR #5656, opened 2019-05-31
Make transaction log group max size configurable

jsteinich:
When trying to determine cause of broken lock exceptions this was useful.
I can also see it being used to adapt to different storage latencies.


xiazen:
LGTM. Thanks for the contribution!

PR #5662, opened 2019-06-03
Fixes #5661 by allowing configuration to pass in value of MetadataPro…

### LINKED ISSUE #5661: Unable to set MetadataPropertyHandling on serializer in AdoNet persistence
WHen  using `TypeNameHandling` settings which write the type info to the serialized JSON, deserialization only works when the `$id` and `$type` property are the first two properties in the serialized JSON string.

For example, it is possible to get this serialized JSON:

```
{
  "$id": "1",
  "$type": "HelloWorld.Grains.GreetingArchive, HelloWorld.Grains",
  "Greetings": {
    "$type": "System.Collections.Generic.List`1[[HelloWorld.Interfaces.IInterface, HelloWorld.Interfaces]], System.Private.CoreLib",
    "$values": [
      {
        "$id": "2",
        "Foo": "Bar",
        "$type": "HelloWorld.Interfaces.Item, HelloWorld.Interfaces",
        "Message": "Good morning, my friend!"
      }
    ]
  }
}
```

Which will not deserialize

>Could not create an instance of type HelloWorld.Interfaces.IInterface. Type is an interface or abstract class and cannot be instantiated. Path 'Greetings.$values[0].Foo', line 1, position 252

However, it is possible for JSON.NET to deserialize this JSON by setting `MetadataPropertyHandling = MetadataPropertyHandling.ReadAhead` - however this option is not exposed by `AdoNetGrainStorageOptions` (as an aside, nor are many other options on the `JsonSerializerSettings`).

This issue causes many problems on deserialization when grain state contains an array/collection of concrete interface implementations.

jam13c:
…pertyHandling to serializer settings

### DIFF
--- a/src/Orleans.Core/Serialization/OrleansJsonSerializer.cs
+++ b/src/Orleans.Core/Serialization/OrleansJsonSerializer.cs
@@ -72,10 +72,10 @@ namespace Orleans.Serialization
             bool useFullAssemblyNames = config.GetBoolProperty(UseFullAssemblyNamesProperty, false);
             bool indentJson = config.GetBoolProperty(IndentJsonProperty, false);
             TypeNameHandling typeNameHandling = config.GetEnumProperty(TypeNameHandlingProperty, settings.TypeNameHandling);
-            return UpdateSerializerSettings(settings, useFullAssemblyNames, indentJson, typeNameHandling, null);
+            return UpdateSerializerSettings(settings, useFullAssemblyNames, indentJson, typeNameHandling);
         }
 
-        public static JsonSerializerSettings UpdateSerializerSettings(JsonSerializerSettings settings, bool useFullAssemblyNames, bool indentJson, TypeNameHandling? typeNameHandling, Action<JsonSerializerSettings> configureSettings)
+        public static JsonSerializerSettings UpdateSerializerSettings(JsonSerializerSettings settings, bool useFullAssemblyNames, bool indentJson, TypeNameHandling? typeNameHandling)
         {
             if (useFullAssemblyNames)
             {
@@ -91,10 +91,7 @@ namespace Orleans.Serialization
             {
                 settings.TypeNameHandling = typeNameHandling.Value;
             }
-            if(configureSettings != null)
-            {
-                configureSettings(settings);
-            }
+           
             return settings;
         }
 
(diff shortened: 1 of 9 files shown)


src/Orleans.Core/Serialization/OrleansJsonSerializer.cs
```diff
@@ -71,10 +72,10 @@ public static JsonSerializerSettings UpdateSerializerSettings(JsonSerializerSett
             bool useFullAssemblyNames = config.GetBoolProperty(UseFullAssemblyNamesProperty, false);
             bool indentJson = config.GetBoolProperty(IndentJsonProperty, false);
             TypeNameHandling typeNameHandling = config.GetEnumProperty(TypeNameHandlingProperty, settings.TypeNameHandling);
-            return UpdateSerializerSettings(settings, useFullAssemblyNames, indentJson, typeNameHandling);
+            return UpdateSerializerSettings(settings, useFullAssemblyNames, indentJson, typeNameHandling, null);
         }
 
-        public static JsonSerializerSettings UpdateSerializerSettings(JsonSerializerSettings settings, bool useFullAssemblyNames, bool indentJson, TypeNameHandling? typeNameHandling)
+        public static JsonSerializerSettings UpdateSerializerSettings(JsonSerializerSettings settings, bool useFullAssemblyNames, bool indentJson, TypeNameHandling? typeNameHandling, Action<JsonSerializerSettings> configureSettings)
```
ReubenBond: Since this is a public method, it's best to not break back-compat.

The configureSettings parameter does not need to be passed in to this method: callers can each call it.
eg:
```C#
this.JsonSettings = OrleansJsonSerializer.UpdateSerializerSettings(OrleansJsonSerializer.GetDefaultSerializerSettings(this.typeResolver, this.grainFactory), this.options.UseFullAssemblyNames, this.options.IndentJson, this.options.TypeNameHandling);

this.options.ConfigureJsonSerializerSettings?.Invoke(this.JsonSettings);
```

jam13c: Ok, never thought of that. Will make the change.

xiazen:
LGTM! @ReubenBond

PR #5677, opened 2019-06-10
Reduce delay localdirectory when cluster membership is not stable

benjaminpetit:
When the `LocalGrainDirectory` receives a register/unregister request that needs to be forwarded (it can happen when all silos doesn't have the same view of the cluster membership), it currently wait 5 seconds before forwarding, to allow some time for the membership changes to propagate.

200ms seems to be a better value since with gossip the membership change should propagate quickly.

### DIFF
--- a/src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
+++ b/src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
@@ -34,8 +34,8 @@ namespace Orleans.Runtime.GrainDirectory
         private Action<SiloAddress, SiloStatus> catalogOnSiloRemoved;
 
         // Consider: move these constants into an apropriate place
-        internal const int HOP_LIMIT = 3; // forward a remote request no more than two times
-        public static readonly TimeSpan RETRY_DELAY = TimeSpan.FromSeconds(5); // Pause 5 seconds between forwards to let the membership directory settle down
+        internal const int HOP_LIMIT = 6; // forward a remote request no more than 4 times
+        public static readonly TimeSpan RETRY_DELAY = TimeSpan.FromMilliseconds(200); // Pause 5 seconds between forwards to let the membership directory settle down
 
         protected SiloAddress Seed { get { return seed; } }
 
@@ -586,7 +586,8 @@ namespace Orleans.Runtime.GrainDirectory
             if (hopCount > 0 && forwardAddress != null)
             {
                 await Task.Delay(RETRY_DELAY);
-                forwardAddress = this.CheckIfShouldForward(address.Grain, hopCount, "RegisterAsync(recheck)");
+                forwardAddress = this.CheckIfShouldForward(address.Grain, hopCount, "RegisterAsync");
+                this.log.LogWarning($"RegisterAsync - It seems we are not the owner of activation {address}, trying to forward it to {forwardAddress} (hopCount={hopCount})");
             }
 
             if (forwardAddress == null)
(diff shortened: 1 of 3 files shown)


src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -34,8 +34,8 @@ internal class LocalGrainDirectory :
         private Action<SiloAddress, SiloStatus> catalogOnSiloRemoved;
 
         // Consider: move these constants into an apropriate place
-        internal const int HOP_LIMIT = 3; // forward a remote request no more than two times
-        public static readonly TimeSpan RETRY_DELAY = TimeSpan.FromSeconds(5); // Pause 5 seconds between forwards to let the membership directory settle down
+        internal const int HOP_LIMIT = 6; // forward a remote request no more than 4 times
+        public static readonly TimeSpan RETRY_DELAY = TimeSpan.FromMilliseconds(200); // Pause 5 seconds between forwards to let the membership directory settle down
```
sergeybykov: The comment still says "5 seconds".

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -586,7 +586,8 @@ public async Task<AddressAndTag> RegisterAsync(ActivationAddress address, bool s
             if (hopCount > 0 && forwardAddress != null)
             {
                 await Task.Delay(RETRY_DELAY);
-                forwardAddress = this.CheckIfShouldForward(address.Grain, hopCount, "RegisterAsync(recheck)");
+                forwardAddress = this.CheckIfShouldForward(address.Grain, hopCount, "RegisterAsync");
+                this.log.LogWarning($"RegisterAsync - It seems we are not the owner of activation {address}, trying to forward it to {forwardAddress} (hopCount={hopCount})");
```
sergeybykov: These semi-duplicate message strings bother me a bit. Can we collapse them?

src/Orleans.Runtime/MembershipService/MembershipOracle.cs
```diff
@@ -12,7 +12,7 @@ namespace Orleans.Runtime.MembershipService
 {
     internal class MembershipOracle : SystemTarget, IMembershipOracle, IMembershipService
     {
-        private readonly static TimeSpan shutdownGossipTimeout = TimeSpan.FromMilliseconds(30);
+        private readonly static TimeSpan shutdownGossipTimeout = TimeSpan.FromSeconds(10);
```
sergeybykov: This seems too drastic of a change.

benjaminpetit: I disagree, it's just that we need time to send one request per silo (in parallel). 30ms was way too short, so we always had a timeout, even though the task was not cancelled.

sergeybykov: The timeout is benign noise, right? The gossip call goes in parallel to all silos telling them to reread the table. While 30ms is clearly too aggressive for that, wouldn't 500ms or 1s be more than enough of time for silos to read the table?

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
benjaminpetit: We could, but we would end up creating multiple string anyway, since there is the operation, the grain identity, ect. And we will propably lose the variable name in the log string if we use a string.Format

ReubenBond: These should use structured logging rather than string templating, imo

src/Orleans.Runtime/MembershipService/MembershipOracle.cs
ReubenBond: Up to 5s sounds reasonable to me

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
benjaminpetit: This code will be replaced soon anyway, let's keep it simple for now

Issue #5671, opened 2019-06-05
not support ISiloBuilder.AddIncomingGrainCallFilter 


lfzm:
Cannot use the `AddIncomingGrainCallFilter` function when using `Microsoft.Extensions.Hosting`

idea：
1. support `ISiloBuilder.AddIncomingGrainCallFilter `
2. `IServiceCollection.AddIncomingGrainCallFilter`  open public

lfzm:
Temporary solution：
```
ISiloBuilder.ConfigureServices(services => services.AddSingleton<IIncomingGrainCallFilter, OrleansExceptionsFiltered>());
```

ReubenBond:
@sergeybykov we merged a fix in #5466, but haven't picked it up for a patch release yet (my mistake)

sergeybykov:
Oh, then this is already resolved.

stephenlautier:
Was this released on version 2.3.4?

ReubenBond:
@stephenlautier unfortunately this was not included in 2.3.4. We plan to make another release shortly and will include it. 

The extension method seems to resolve for me: 
![image](https://user-images.githubusercontent.com/203839/59391545-6e95d700-8d29-11e9-8921-4e880de730d5.png)

stephenlautier:
@ReubenBond ahh oki that explains it, I was testing on 2.3.4, not on master.

I got confused because i was seeing the source and it was added 22 March (quite back), and somehow on VS i was getting 2 classes but couldnt access the `Orleans.Runtime` one, just the other

stephenlautier:
Just checked on an orleans sample project, and im just not seeing the `ISiloBuilder` extensions, so if they are resolving correctly on master then it should work after release. Sorry about this

ReubenBond:
It's my mistake, @stephenlautier

lfzm:
V2.3.5 completed  [#5466](https://github.com/dotnet/orleans/pull/5466)

PR #5684, opened 2019-06-12
Cherry-picked fixes for 2.3.5

### LINKED ISSUE #5661: Unable to set MetadataPropertyHandling on serializer in AdoNet persistence

sergeybykov:
Fixes #5661 by allowing configuration to pass in value of MetadataPro… (#5662)
Make transaction log group max size configurable (#5656)
Reduce delay local directory when cluster membership is not stable (#5677)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -111,7 +111,7 @@
   <!-- Versioning properties -->
   <PropertyGroup>
     <AssemblyVersion>2.0.0.0</AssemblyVersion>
-    <VersionPrefix Condition=" '$(VersionPrefix)'=='' ">2.3.0</VersionPrefix>
+    <VersionPrefix Condition=" '$(VersionPrefix)'=='' ">2.4.0</VersionPrefix>
   </PropertyGroup>
 
   <!-- For Debug builds generated a date/time dependent version suffix -->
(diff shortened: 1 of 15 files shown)


ReubenBond:
We should also include https://github.com/dotnet/orleans/issues/5671

PR #5688, opened 2019-06-13
Fix #5473 - StackOverflowException in codegen for recursively defined types

### LINKED ISSUE #5473: Microsoft.Orleans.CodeGenerator.MSBuild 2.2.4 infinite loop
The following code is probably causing an infinite loop in `Microsoft.Orleans.CodeGenerator.MSBuild`. 
When I build the following code, it hangs after generating `..orleans.g.args.txt`

this is the exit error message:
```
Orleans.CodeGenerator - command-line = SourceToSource /code/GrainInterfaces/obj/Debug/netstandard2.0/GrainInterfaces.orleans.g.args.txt
<...hangs for a long time...>
         Process is terminating due to StackOverflowException.
```

This code compiles without problem with `Microsoft.Orleans.OrleansCodeGenerator.Build`, `2.2.4`

GrainInterface.cs

```
    public interface IFoo : IGrainWithGuidKey
    {
        Task Bar(Complex request);
    }
```

Complex.cs
```
    public class Complex : Simple<Complex>
    {    
    }
    
    public abstract class Simple<T> where T : Simple<T>
    {
        public T DoSomething(T input)
        {
            // do something to input
            return input;
        }
    }
```

It is probably this line `public abstract class Simple<T> where T : Simple<T>` that is causing the infinite loop.

ReubenBond:
Fixes #5473

### DIFF
--- a/src/Orleans.CodeGenerator/Analysis/CompilationAnalyzer.cs
+++ b/src/Orleans.CodeGenerator/Analysis/CompilationAnalyzer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


ReubenBond:
I think we can merge this if someone is ready to review it

PR #5691, opened 2019-06-14
Add core tracing events

xiazen:
Adding core tracing events for application layer grain requests.

### DIFF
--- a/src/Orleans.Core.Abstractions/Concurrency/GrainAttributeConcurrency.cs
+++ b/src/Orleans.Core.Abstractions/Concurrency/GrainAttributeConcurrency.cs
@@ -1,4 +1,4 @@
-﻿using Orleans.Placement;
+using Orleans.Placement;
 using Orleans.Runtime;
 using System;
 using System.Collections.Generic;
--- a/src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
+++ b/src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Diagnostics;
+using System.Diagnostics.Tracing;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Orleans.CodeGeneration;
--- a/src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
+++ b/src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
@@ -1,6 +1,4 @@
 using System;
-using System.Diagnostics;
-using System.Diagnostics.Tracing;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Orleans.CodeGeneration;
--- a/src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
+++ b/src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
@@ -9,7 +9,10 @@ namespace Orleans.Runtime
     [EventSource(Name = "Microsoft-Orleans-CallBackData")]
     internal sealed class OrleansCallBackDataEvent : EventSource
     {
-        public static readonly OrleansCallBackDataEvent Log = new OrleansCallBackDataEvent();
+        private static readonly OrleansCallBackDataEvent Log = new OrleansCallBackDataEvent();
+        public static readonly Action OnTimeoutAction = Log.OnTimeout;
+        public static readonly Action OnTargetSiloFailAction = Log.OnTargetSiloFail;
+        public static readonly Action DoCallbackAction = Log.DoCallback;
         public void OnTimeout()
         {
             WriteEvent(1);
@@ -29,7 +32,10 @@ namespace Orleans.Runtime
     [EventSource(Name = "Microsoft-Orleans-OutsideRuntimeClient")]
     internal sealed class OrleansOutsideRuntimeClientEvent : EventSource
     {
-        public static readonly OrleansOutsideRuntimeClientEvent Log = new OrleansOutsideRuntimeClientEvent();
+        private static readonly OrleansOutsideRuntimeClientEvent Log = new OrleansOutsideRuntimeClientEvent();
+        public static readonly Action SendRequestAction = Log.SendRequest;
+        public static readonly Action ReceiveResponseAction = Log.ReceiveResponse;
+        public static readonly Action SendResponseAction = Log.SendResponse;
         public void SendRequest()
         {
             WriteEvent(1);
--- a/src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
+++ b/src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
@@ -1,71 +1,57 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
 using System.Text;
 
-namespace Orleans.DistributedTracing.EventSourceEvents
+namespace Orleans.Runtime
 {
-    [EventSource(Name = "Microsoft-Orleans-CallBackDataEvent")]
-    public class OrleansCallBackDataEvent : EventSource
+    [EventSource(Name = "Microsoft-Orleans-CallBackData")]
+    internal sealed class OrleansCallBackDataEvent : EventSource
     {
         public static readonly OrleansCallBackDataEvent Log = new OrleansCallBackDataEvent();
-        public void OnTimeoutStart()
+        public void OnTimeout()
         {
             WriteEvent(1);
         }
-        public void OnTimeoutStop()
+
+        public void OnTargetSiloFail()
         {
             WriteEvent(2);
         }
 
-        public void OnTargetSiloFailStart()
+        public void DoCallback()
         {
             WriteEvent(3);
-        }
-        public void OnTargetSiloFailStop()
-        {
-            WriteEvent(4);
-        }
-
-        public void DoCallbackStart()
-        {
-            WriteEvent(5);
-        }
-        public void DoCallbackStop()
-        {
-            WriteEvent(6);
         }
     }
 
-    [EventSource(Name = "Microsoft-Orleans-OutsideRuntimeClientEvent")]
-    public class OrleansOutsideRuntimeClientEvent : EventSource
+    [EventSource(Name = "Microsoft-Orleans-OutsideRuntimeClient")]
+    internal sealed class OrleansOutsideRuntimeClientEvent : EventSource
     {
         public static readonly OrleansOutsideRuntimeClientEvent Log = new OrleansOutsideRuntimeClientEvent();
-        public void SendRequestStart()
+        public void SendRequest()
         {
             WriteEvent(1);
         }
-
-        public void SendRequestStop()
+        public void ReceiveResponse()
         {
             WriteEvent(2);
         }
-        public void ReceiveResponseStart()
+
+        public void SendResponse()
         {
             WriteEvent(3);
         }
-        public void ReceiveResponseStop()
-        {
-            WriteEvent(4);
-        }
+    }
 
-        public void SendResponseStart()
+    internal static class EventSourceUtils
+    {
+        public static void EmitEvent(Message message, Action emitAction)
         {
-            WriteEvent(5);
-        }
-        public void SendResponseStop()
-        {
-            WriteEvent(6);
+            EventSource.SetCurrentThreadActivityId(message.TraceContext?.ActivityId??Guid.Empty, out var previousId);
+            emitAction();
+            EventSource.SetCurrentThreadActivityId(previousId);
         }
     }
 }
--- a/src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
+++ b/src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
@@ -1,0 +1,71 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Text;
+
+namespace Orleans.DistributedTracing.EventSourceEvents
+{
+    [EventSource(Name = "Microsoft-Orleans-CallBackDataEvent")]
+    public class OrleansCallBackDataEvent : EventSource
+    {
+        public static readonly OrleansCallBackDataEvent Log = new OrleansCallBackDataEvent();
+        public void OnTimeoutStart()
+        {
+            WriteEvent(1);
+        }
+        public void OnTimeoutStop()
+        {
+            WriteEvent(2);
+        }
+
+        public void OnTargetSiloFailStart()
+        {
+            WriteEvent(3);
+        }
+        public void OnTargetSiloFailStop()
+        {
+            WriteEvent(4);
+        }
+
+        public void DoCallbackStart()
+        {
+            WriteEvent(5);
+        }
+        public void DoCallbackStop()
+        {
+            WriteEvent(6);
+        }
+    }
+
+    [EventSource(Name = "Microsoft-Orleans-OutsideRuntimeClientEvent")]
+    public class OrleansOutsideRuntimeClientEvent : EventSource
+    {
+        public static readonly OrleansOutsideRuntimeClientEvent Log = new OrleansOutsideRuntimeClientEvent();
+        public void SendRequestStart()
+        {
+            WriteEvent(1);
+        }
+
+        public void SendRequestStop()
+        {
+            WriteEvent(2);
+        }
+        public void ReceiveResponseStart()
+        {
+            WriteEvent(3);
+        }
+        public void ReceiveResponseStop()
+        {
+            WriteEvent(4);
+        }
+
+        public void SendResponseStart()
+        {
+            WriteEvent(5);
+        }
+        public void SendResponseStop()
+        {
+            WriteEvent(6);
+        }
+    }
+}
--- a/src/Orleans.Core/Messaging/Message.cs
+++ b/src/Orleans.Core/Messaging/Message.cs
@@ -252,9 +252,9 @@ namespace Orleans.Runtime
             set { Headers.CallChainId = value; }
         }
         
-        public Guid ActivityId {
-            get { return Headers.ActivityId; }
-            set { Headers.ActivityId = value; }
+        public TraceContext TraceContext {
+            get { return Headers.TraceContext; }
+            set { Headers.TraceContext = value; }
         }
 
         public ActivationAddress SendingAddress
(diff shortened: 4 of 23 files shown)


src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
```diff
@@ -1,11 +1,27 @@
...
+using System.Diagnostics;
+using System.Diagnostics.Tracing;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Orleans.CodeGeneration;
 using Orleans.Serialization;
 
 namespace Orleans.Runtime
 {
+    public class OrleansGrainReferenceEvent : EventSource
```
ReubenBond: We should specify the name explicitly & attribute the event-emitting methods, too

src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
```diff
@@ -330,7 +346,15 @@ protected void InvokeOneWayMethod(int methodId, object[] arguments, InvokeMethod
...
-            return this.Runtime.InvokeMethodAsync<T>(this, methodId, arguments, options | invokeMethodOptions, silo);
+            try
+            {
+                OrleansGrainReferenceEvent.Log.InvokeMethodAsyncStart();
+                return this.Runtime.InvokeMethodAsync<T>(this, methodId, arguments, options | invokeMethodOptions, silo);
+            }
+            finally
+            {
+                OrleansGrainReferenceEvent.Log.InvokeMethodAsyncStop();
+            }
```
ReubenBond: The *Stop event needs to be emitted once the task completes, otherwise I think it breaks activity support

src/Orleans.Core/Messaging/Message.cs
```diff
@@ -9,6 +10,20 @@
 
 namespace Orleans.Runtime
 {
+    internal static class EventSourceUtils
+    {
+        public static void SetEventSourceActivityIdIfNotNull(this Message msg)
+        {
+            if(msg.ActivityId != null)
+                EventSource.SetCurrentThreadActivityId(msg.ActivityId.Guid);
```
ReubenBond: Unsure, but do we need to track the existing thread activity id so we can restore it on completion?

Also: formatting

src/Orleans.Core/Messaging/Message.cs
```diff
@@ -1253,6 +1284,11 @@ public static void Serializer(object untypedInput, ISerializationContext context
                     WriteObj(context, typeof(GuidId), input.TargetObserverId);
                 }
 
+                if ((headers & Headers.ACTIVITY_ID) != Headers.NONE)
+                {
+                    WriteObj(context, typeof(GuidId), input.ActivityId);
+                }
```
ReubenBond: Should this be the last field written since we just added it with the last flag (1 << 30)? I think that may be required for back compat

ReubenBond: Should this be the last field written since we just added it with the last flag (1 << 30)? I think that may be required for back compat

src/Orleans.Core/Messaging/MessageFactory.cs
```diff
@@ -20,11 +22,13 @@ public MessageFactory(SerializationManager serializationManager, ILogger<Message
 
         public Message CreateMessage(InvokeMethodRequest request, InvokeMethodOptions options)
         {
+           
             var message = new Message
             {
                 Category = Message.Categories.Application,
                 Direction = (options & InvokeMethodOptions.OneWay) != 0 ? Message.Directions.OneWay : Message.Directions.Request,
                 Id = CorrelationId.GetNext(),
+                ActivityId = GuidId.GetGuidId(EventSource.CurrentThreadActivityId),
```
ReubenBond: I think this should be a regular Guid. `GuidId` would introduce an allocation (it's essentially boxing a Guid) + indirection which I'm not sure is necessary.

src/Orleans.Core/Orleans.Core.csproj
```diff
@@ -22,6 +22,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="$(MicrosoftExtensionsConfigurationVersion)" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelVersion)" />
+    <PackageReference Include="Microsoft.Extensions.DiagnosticAdapter" Version="2.2.0" />
```
ReubenBond: If this package is needed, we should add the version to the props file & use that

src/Orleans.Core/Runtime/CallbackData.cs
```diff
@@ -1,10 +1,24 @@
 using System;
+using System.Diagnostics.Tracing;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Orleans.Transactions;
 
 namespace Orleans.Runtime
 {
+    public class OrleansCallBackDataEvent : EventSource
```
ReubenBond: These classes should perhaps go in a separate file, maybe on that's specific for event sources

src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
xiazen: what do you mean? The name will be OrleansGrainReferenceEvent  by default, is that not enough?

src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
xiazen: I see , let me do that

src/Orleans.Core/Messaging/Message.cs
xiazen: I see, will do

src/Orleans.Core/Messaging/MessageFactory.cs
xiazen: but using a Guid will make message header always has this field, not sure if we want to parse an empty ActivityId if users didn't set one.

src/Orleans.Core/Orleans.Core.csproj
xiazen: will do

src/Orleans.Core/Runtime/CallbackData.cs
xiazen: go to the same file in core abstraction? Or each project holds its own file.

src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
ReubenBond: I think the convention for event source names is more like "Microsott-Orleans-GrainReference" or something like that

src/Orleans.Core/Messaging/MessageFactory.cs
xiazen: Oh right, I meant to not set it if the CurrentThreadActivityId is empty, hence the using GuidId instead of Guid, since some headers may not have this field. But failed to do it yet.

src/Orleans.Core/Messaging/Message.cs
```diff
@@ -1127,6 +1143,7 @@ internal Headers GetHeadersMask()
                 headers = string.IsNullOrEmpty(_rejectionInfo) ? headers & ~Headers.REJECTION_INFO : headers | Headers.REJECTION_INFO;
                 headers = _requestContextData == null || _requestContextData.Count == 0 ? headers & ~Headers.REQUEST_CONTEXT : headers | Headers.REQUEST_CONTEXT;
                 headers = _callChainId == null ? headers & ~Headers.CALL_CHAIN_ID : headers | Headers.CALL_CHAIN_ID;
+                headers = activityId == null? headers & ~Headers.ACTIVITY_ID : headers | Headers.ACTIVITY_ID;
```
ReubenBond: Compare to Guid.Empty?

src/Orleans.Core/Runtime/CallbackData.cs
```diff
@@ -38,61 +77,100 @@ public bool IsExpired(long currentTimestamp)
...
                 return;
-            var msg = this.Message; // Local working copy
-
-            string messageHistory = msg.GetTargetHistory();
-            string errorMsg = $"Response did not arrive on time in {timeout} for message: {msg}. Target History is: {messageHistory}.";
-            this.shared.Logger.Warn(ErrorCode.Runtime_Error_100157, "{0} About to break its promise.", errorMsg);
+            var previousActivityId = EventSource.CurrentThreadActivityId;
+            try
+            {
+                EventSource.SetCurrentThreadActivityId(this.Message.ActivityId);
```
ReubenBond: Note that you can set + get the activity id in one call: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource.setcurrentthreadactivityid?view=netframework-4.8#System_Diagnostics_Tracing_EventSource_SetCurrentThreadActivityId_System_Guid_System_Guid__

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -19,6 +20,44 @@
 
 namespace Orleans
 {
+    [EventSource(Name="Microsoft-Orleans-OutsideRuntimeClientEvent")]
+    public class OrleansOutsideRuntimeClientEvent : EventSource
+    {
+        public static OrleansOutsideRuntimeClientEvent Log = new OrleansOutsideRuntimeClientEvent();
+        public void SendRequestStart()
+        {
+            if(EventSource.CurrentThreadActivityId!=Guid.Empty)
```
ReubenBond: I wonder if we should check for `this.Enabled()` instead?

xiazen: this.enabled will be checked by event source itself, https://referencesource.microsoft.com/#mscorlib/system/diagnostics/eventing/eventsource.cs,ae4fe1a0959bdb62

based on their source file

src/Orleans.Core/Runtime/CallbackData.cs
xiazen: yeah I know. but it won't work with try-finally.

src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -1,14 +1,33 @@
...
 using Orleans.Messaging;
 using Orleans.Serialization;
-using System.Threading.Tasks;
 
 namespace Orleans.Runtime.Messaging
 {
+    [EventSource(Name = "Microsoft-Orleans-IncomingMessageAcceptorEvent")]
+    public class OrleansIncomingMessageAcceptorEvent : EventSource
+    {
+        public static OrleansIncomingMessageAcceptorEvent Log = new OrleansIncomingMessageAcceptorEvent();
```
ReubenBond: These should be readonly

src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
```diff
@@ -1,4 +1,6 @@
 using System;
+using System.Diagnostics;
+using System.Diagnostics.Tracing;
```
ReubenBond: Revert unchanged files / InvokeMethodRequest.cs

src/Orleans.Core/Messaging/Message.cs
```diff
@@ -250,6 +251,11 @@ public CorrelationId CallChainId
             get { return Headers.CallChainId; }
             set { Headers.CallChainId = value; }
         }
+        
+        public Guid ActivityId {
```
ReubenBond: Formatting

src/Orleans.Core/Runtime/CallbackData.cs
```diff
@@ -38,61 +40,97 @@ public bool IsExpired(long currentTimestamp)
 
         public void OnTimeout(TimeSpan timeout)
         {
+            // set/clear activityid
```
ReubenBond: remove

src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
```diff
@@ -0,0 +1,71 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Text;
+
+namespace Orleans.DistributedTracing.EventSourceEvents
```
ReubenBond: IMO the namespace should be `Orleans` or `Orleans.Runtime` & the classes should all be marked `internal sealed`

src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+    [EventSource(Name = "Microsoft-Orleans-CallBackDataEvent")]
```
ReubenBond: Remove `Event` from the end of the name and consider a name like `"Microsoft-Orleans-Client"` & merge these two classes

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -304,46 +306,56 @@ private void RunClientMessagePump(CancellationToken ct)
 
             while (listenForMessages)
             {
+                // set/clear activityid
```
ReubenBond: remove

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -304,46 +306,56 @@ private void RunClientMessagePump(CancellationToken ct)
...
+                            logger.Error(ErrorCode.Runtime_Error_100327, $"Message not supported: {message}.");
+                            break;
+                    }
                 }
+                finally
+                {
+                    EventSource.SetCurrentThreadActivityId(previousId);
+                }
+
             }
```
ReubenBond: remove superfluous newlines

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -378,7 +399,17 @@ public void Reconnect()
         public void SendRequest(GrainReference target, InvokeMethodRequest request, TaskCompletionSource<object> context, string debugContext = null, InvokeMethodOptions options = InvokeMethodOptions.None, string genericArguments = null)
         {
             var message = this.messageFactory.CreateMessage(request, options);
-            SendRequestMessage(target, message, context, debugContext, options, genericArguments);
+            EventSource.SetCurrentThreadActivityId(message.ActivityId, out var previousId);
+            try
+            {
+                OrleansOutsideRuntimeClientEvent.Log.SendRequestStart() ;
```
ReubenBond: nit: fix spacing before `;`

src/Orleans.Core/Runtime/SharedCallbackData.cs
```diff
@@ -80,7 +80,7 @@ public void ResponseCallback(Message message, TaskCompletionSource<object> conte
                             {
                                 message.RejectionInfo = "Unable to send request - no rejection info available";
                             }
-                            rejection = new OrleansMessageRejectionException(message.RejectionInfo);
+                            rejection = new OrleansMessageRejectionException(message.RejectionInfo);   
                         }
```
ReubenBond: revert whitespace change

src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -72,130 +74,144 @@ internal class Dispatcher
...
-            }
-
-            // check if its targeted at a new activation
-            if (message.TargetGrain.IsSystemTarget)
-            {
-                MessagingProcessingStatisticsGroup.OnDispatcherMessageProcessedError(message, "ReceiveMessage on system target.");
-                throw new InvalidOperationException("Dispatcher was called ReceiveMessage on system target for " + message);
-            }
-
+            // set/clear activityid
```
ReubenBond: remove

src/Orleans.Core/Messaging/Message.cs
```diff
@@ -1383,6 +1405,9 @@ public static object Deserializer(Type expected, IDeserializationContext context
                 if ((headers & Headers.TRANSACTION_INFO) != Headers.NONE)
                     result.TransactionInfo = SerializationManager.DeserializeInner<ITransactionInfo>(context);
 
+                if ((headers & Headers.ACTIVITY_ID) != Headers.NONE)
+                    result.ActivityId = reader.ReadGuid();
```
ReubenBond: Given that we're going to implement proper distributed tracing support, should we make this some type other than a guid? eg, something which can also hold the various fields needed by the w3c trace context?

xiazen: hmm. Taken how orthogonal EventSource and Activity are, I think we should have a separate thing for w3c trace context, instead of glob them together. The link between the two is quite weak.

ReubenBond: I generally agree. It's just that:
* Each header added to `Message` has a cost and the next header we add will need to be some kind of *ExtendedHeaders` header or something else which retains the flexibility of adding more headers.
* In our meeting with Liudmila & Noah, we discussed deriving `ActivityId` from the distributed tracing context - so a separate field might not be required.

src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
```diff
@@ -1,71 +1,57 @@
...
         }
-        public void OnTimeoutStop()
+
+        public void OnTargetSiloFail()
         {
             WriteEvent(2);
         }
 
-        public void OnTargetSiloFailStart()
+        public void DoCallback()
```
ReubenBond: Is it right to call this "OnSuccess" or not really?

(we can rename later)

src/Orleans.Core.Abstractions/Concurrency/GrainAttributeConcurrency.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Orleans.Placement;
+using Orleans.Placement;
```
xiazen: I tried to clean this, but my git doesn't think this is a diff...

src/Orleans.Core/DistributedTracing/EventSourceEvents.cs
xiazen: I think we should make this as simple as possible, this is a 1 to 1 map to the method name it is used in. I'm of the opinion that event source should be like logging, let's not complicate this

src/Orleans.Core/Runtime/GrainReferenceRuntime.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging;
```
xiazen: tried to clean this too.. but my git doesn't think this is a diff...

src/Orleans.Core/Runtime/CallbackData.cs
```diff
@@ -54,7 +57,7 @@ public void OnTargetSiloFail()
         {
             if (this.IsCompleted)
                 return;
-
+            EventSourceUtils.EmitEvent(this.Message, OrleansCallBackDataEvent.Log.OnTargetSiloFail);
```
ReubenBond: This pattern is very expensive. Each time a call to `EmitEvent` is made, it will allocate a new `Action` for this second parameter.

If we cache the `Action` in a field on the EventSource class then the allocation goes away and it should be much cheaper.

src/Orleans.Core/Runtime/GrainReferenceRuntime.cs
ReubenBond: No problems. You can revert the change regardless of what git thinks. Next time

src/Orleans.Core/Runtime/CallbackData.cs
ReubenBond: I think if we just fix this then it can be merged

PR #5694, opened 2019-06-17
ClusterClient: only call IRuntimeClient.Reset for OutsideRuntimeClient

ReubenBond:
Fixes some benign exceptions during shutdown now that ClusterClient is also used on silos


ReubenBond:
@jason-bragg please take a look

PR #5695, opened 2019-06-18
Expose versioning from membership

ReubenBond:
This is a major refactoring of the core membership system.

The impetus is that we want to expose monotonic membership versions from membership for use in other parts of the system (eg, directory). The membership system already maintains a global version, primarily for the purpose of concurrency control. In order for these versions to be consistent with membership locally (under concurrency & mutation) this PR breaks membership into smaller components with an immutable core. The core is `MembershipTableManager` and it is responsible for all direct interactions with the membership table (provided by `IMembershipTable`). Other components within membership consume membership as a series of immutable, consistent table snapshots.

The main components are:
* `MembershipTableManager` - responsible for all direct interactions with the shared membership table. Performs refreshes periodically or on demand. Exposes membership information in two ways:
  * The most recent snapshot
  * The sequence of observed snapshots
* `MembershipAgent` - responsible for representing the local silo in the membership table: reflecting the current lifecycle stage in membership & periodically updating the *IAmAlive* entry for this silo.
* `ClusterHealthMonitor` - determines which silos this silo is responsible for monitoring, spawns individual `SiloHealthMonitor` instances, & periodically asks them to probe the remote silos. When a silo is determine to be unhealthy locally, it calls `MembershipTableManager.TrySuspectOrKill`.
* `SiloStatusOracle` - implements `ISiloStatusOracle` for compatibility purposes, obsoleting some methods.
* `MembershipTableCleanupAgent` - responsible for periodically removing old, defunct membership table entries.

This PR does not change the public surface significantly, but the plan is to eventually expose a simplified view of membership for other components to consume.


src/Orleans.Runtime/Utilities/AsyncEnumerable.cs
```diff
@@ -0,0 +1,241 @@
...
+            {
+                this.value = value;
+                this.next = new TaskCompletionSource<Element>(TaskCreationOptions.RunContinuationsAsynchronously);
+            }
+
+            public static Element CreateInitial() => new Element(
+                AsyncEnumerable.InitialValue,
+                new TaskCompletionSource<Element>(TaskCreationOptions.RunContinuationsAsynchronously));
+
+            public static Element CreateDisposed()
```
jason-bragg: nit - Can't dispose element be an immutable static single instance.

ReubenBond: I think it's cleaner to create a new one each time. In theory there should be no difference

ReubenBond:
BVT failed because lifecycle took too long to start... because ExpectedClusterSize = 20, which means it waits up to 5 seconds arbitrarily during startup...

src/Orleans.Runtime/Utilities/AsyncEnumerable.cs
ReubenBond: Although I don't hold a strong opinion here

PR #5702, opened 2019-06-20
Remove ExpectedClusterSize & add MaxOperationBackoffTime

ReubenBond:


src/Orleans.Core/Configuration/Options/ClusterMembershipOptions.cs
```diff
@@ -57,10 +57,10 @@ public class ClusterMembershipOptions
         public static readonly TimeSpan DEFAULT_LIVENESS_MAX_JOIN_ATTEMPT_TIME = TimeSpan.FromMinutes(5); // 5 min
 
         /// <summary>
-        /// The expected size of a cluster. Need not be very accurate, can be an overestimate.
+        /// The maximum amount of time to wait between operation retries.
         /// </summary>
-        public int ExpectedClusterSize { get; set; } = DEFAULT_LIVENESS_EXPECTED_CLUSTER_SIZE;
-        public static readonly int DEFAULT_LIVENESS_EXPECTED_CLUSTER_SIZE = 20;
+        public TimeSpan MaxOperationBackoffTime { get; set; } = DEFAULT_MAX_OPERATION_BACKOFF_TIME;
```
jason-bragg: Do we need this?  We use an exponential backoff.  I suspect we should be able to set a reasonable default.

ReubenBond: I'm ok with not having it, but I was being conservative by having a config value.

PR #5708, opened 2019-06-26
Fix the test trace file name on Unix systems

srollinet:
Use the correct path separator on Unix systems, otherwise it creates files with names like 'logs\testcluster-2019-06-25t15-34-58-179_MainClient.log'


ReubenBond:
Thank you, @srollinet!

PR #5710, opened 2019-06-26
Fix CategoryDiscoverer first-chance exception while debugging

ReubenBond:


test/TestInfrastructure/TestExtensions/TestCategory.cs
```diff
@@ -25,6 +25,10 @@ public class TestCategoryAttribute : Attribute, ITraitAttribute
 
 public class CategoryDiscoverer : ITraitDiscoverer
 {
+    public CategoryDiscoverer(IMessageSink diagnosticMessageSink)
```
xiazen: can you explain more on what the issue is? 

And also, this class doesn't seem to be used anywhere. Can we just remove it?

ReubenBond: This class essentially translates our `[Category("BVT")]` attributes into `[Trait("Category", "BVT")]` attributes. It's important for how Xunit finds our tests.

I notice during debugging that Xunit throws an exception (which is later caught) because we are missing this constructor. It is arguably an Xunit issue (they should check first), but this fixes the exception regardless.

PR #5711, opened 2019-06-26
Improve lifecycle logging

ReubenBond:
Log lifecycle levels using stage names rather than numeric values.
Use structured logging


ReubenBond:
My hope is that this improves the friendliness of the lifecycle log messages

PR #5715, opened 2019-06-26
Dispose TestCluster after tests

ReubenBond:


ReubenBond:
The impetus for this PR is that *not* disposing the `TestCluster` means some aspects of it may continue running after the test should have completed. Eg, potentially some timers.

That makes it harder to debug stuck tests, etc, and may cause other noise.

RehanSaeed:
Do I need to call `StopAllSilos` before `Dispose`? I'm using an xUnit `IClassFixture` to avoid creating the cluster every time for each test as it does take some time.

```c#
public class ClusterFixture : IDisposable
{
    public ClusterFixture()
    {
        var builder = new TestClusterBuilder();
        builder.AddClientBuilderConfigurator<TestClientBuilderConfigurator>();
        builder.AddSiloBuilderConfigurator<TestSiloBuilderConfigurator>();
        this.Cluster = builder.Build();
        this.Cluster.Deploy();
    }

    public TestCluster Cluster { get; }

    public void Dispose()
    {
        this.Cluster.StopAllSilos();
        this.Cluster.Dispose();
    }
}

public class HelloGrainTest : IClassFixture<ClusterFixture>
{
    private readonly TestCluster cluster;

    public HelloGrainTest(ClusterFixture fixture) => this.cluster = fixture.Cluster;

    [Fact]
    public async Task SayHello_PassName_ReturnsGreeting()
    {
        var grain = this.cluster.GrainFactory.GetGrain<IHelloGrain>(Guid.NewGuid());

        var greeting = await grain.SayHello("Rehan");

        Assert.Equal("Hello Rehan!", greeting);
    }

    [Fact]
    public async Task SayHello_PassName_CountIncremented()
    {
        var helloGrain = this.cluster.GrainFactory.GetGrain<IHelloGrain>(Guid.NewGuid());
        var counterGrain = this.cluster.GrainFactory.GetGrain<ICounterGrain>(Guid.Empty);

        await helloGrain.SayHello("Rehan");

        await Task.Delay(TimeSpan.FromSeconds(2));
        var count = await counterGrain.GetCount();

        Assert.Equal(1L, count);
    }

    [Fact]
    public async Task SayHello_PassName_SaidHelloPublished()
    {
        string hello = null;
        var helloGrain = this.cluster.GrainFactory.GetGrain<IHelloGrain>(Guid.NewGuid());
        var streamProvider = this.cluster.Client.GetStreamProvider(StreamProviderName.Default);
        var stream = streamProvider.GetStream<string>(Guid.Empty, StreamName.SaidHello);
        var subscription = await stream.SubscribeAsync(
            (x, token) =>
            {
                hello = x;
                return Task.CompletedTask;
            });

        await helloGrain.SayHello("Rehan");

        Assert.Equal("Rehan", hello);
    }
}
```

ReubenBond:
See how it's done in [BaseTestClusterFixture](https://github.com/dotnet/orleans/pull/5715/files#diff-e9d703ed9913adc664ec938784372551)

You can implement `Xunit.IAsyncLifetime` & call StopAllSilos in `DisposeAsync`. In BaseTestClusterFixture we are doing both (though the synchronous `Dispose` is likely not needed given `DisposeAsync`)

RehanSaeed:
So let me get this straight. The recommended approach is `StopAllSilosAsync` and `Dispose` is just there as a synchronous backup. Finally `DisposeAsync` is coming, so use that in .NET Core 3.0.

ReubenBond:
Yes, pretty much. Either way is fine as long as the test cluster gets shutdown when the fixture is no longer needed.

`DisposeAsync` is on `Xunit.IAsyncLifetime`, so you can use it today.

`System.IAsyncDisposable` is coming with .NET Core 3.0, but that's a separate matter (Xunit will probably support it, but I'm not sure)

PR #5717, opened 2019-06-26
Fix AgentTests

ReubenBond:
I'm not sure how this test was passing, but `TryAddSingleton` for a type which is supposed to have many registrations like `ILifecycleParticipant<T>` isn't right.


ReubenBond:
Failed due to unrelated flaky test

PR #5724, opened 2019-06-26
DeploymentLoadPublisher: use silo lifecycle

ReubenBond:
Moving more components out of Silo.cs

### DIFF
--- a/src/Orleans.Runtime/Placement/DeploymentLoadPublisher.cs
+++ b/src/Orleans.Runtime/Placement/DeploymentLoadPublisher.cs
@@ -249,15 +249,15 @@ namespace Orleans.Runtime
 
             async Task OnActiveStop(CancellationToken ct)
             {
-                this.publishTimer.Dispose();
                 this.siloStatusOracle.UnSubscribeFromSiloStatusEvents(this);
+                this.publishTimer.Cancel();
                 await Task.WhenAny(ct.WhenCancelled(), Task.WhenAll(tasks));
             }
         }
 
         public void Dispose()
         {
-            this.publishTimer?.Dispose();
+            this.publishTimer.Dispose();
         }
     }
 }
--- a/src/Orleans.Runtime/Placement/DeploymentLoadPublisher.cs
+++ b/src/Orleans.Runtime/Placement/DeploymentLoadPublisher.cs
@@ -2,11 +2,14 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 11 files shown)


src/Orleans.Runtime/Placement/DeploymentLoadPublisher.cs
```diff
@@ -224,8 +231,33 @@ private void OnSiloStatusChange(SiloAddress updatedSilo, SiloStatus status)
...
+            Task OnActiveStart(CancellationToken ct)
+            {
+                tasks.Add(this.ScheduleTask(this.PublishStatistics));
+                this.siloStatusOracle.SubscribeToSiloStatusEvents(this);
+                return Task.CompletedTask;
+            }
+
+            async Task OnActiveStop(CancellationToken ct)
+            {
+                this.publishTimer.Dispose();
```
jason-bragg: Suggest calling this.Dispose() to avoid duplicate code and setting publishTimer = null to avoid double dispose.

ReubenBond: It's a readonly, so how about I add a `IAsyncTimer.Cancel()` method and call that here instead? I think that has a clearer intention.

Double-dispose is 'guaranteed' (by [IDisposable docs](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?view=netframework-4.8)) to be permissible:

> If an object's Dispose method is called more than once, the object must ignore all calls after the first one. The object must not throw an exception if its Dispose method is called multiple times. Instance methods other than Dispose can throw an ObjectDisposedException when resources are already disposed.

jason-bragg: Agreed.  Just need to ensure our impl follows that rule.

ReubenBond:
Addressed review feedback

src/Orleans.Runtime/Timers/AsyncTimer.cs
```diff
@@ -84,6 +86,16 @@ public bool CheckHealth(DateTime lastCheckTime)
...
+        public void Cancel() => this.cancellation.Cancel(throwOnFirstException: false);
+
+        public void Dispose()
+        {
+            if (!this.disposed)
+            {
+                this.disposed = true;
+                this.Cancel();
+                this.cancellation.Dispose();
+            }
```
ReubenBond: We have a 'disposed' check here because calling `CancellationTokenSource.Cancel()` might throw if the CTS is already disposed

Issue #5568, opened 2019-05-02
"Failed to get a local IP address" on Linux with localhost clustering


bojidar-bg:
**OS**: Arch Linux, updated 30.04.2019
**Linux kernel**: 5.0.10-arch1-1-ARCH
**Orleans version**: 2.2.0

**What happened**:
I am creating a cluster with UseLocalhostClustering, but I cannot connect from a client running on the same machine without having internet connection (or when on a tethered connection).
I suspect this is due to the loopback address having UNKNOWN state.

Related to #5097.

Stacktrace of the error:
```
Unhandled Exception: System.AggregateException: One or more errors occurred. (Failed to get a local IP address.) ---> 
Orleans.Runtime.OrleansException: Failed to get a local IP address.
    at Orleans.Runtime.Configuration.ConfigUtilities.GetLocalIPAddress(AddressFamily family, String interfaceName)
    at Orleans.OutsideRuntimeClient.ConsumeServices(IServiceProvider services)
    at Orleans.ClientBuilder.Build()
    at … (non-orleans code)
```

`ip addr list` output (when on a tethered connection from an andorid phone):
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: wlp3s0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc mq state DOWN group default qlen 1000
    link/ether 56:2f:10:d5:d9:f7 brd ff:ff:ff:ff:ff:ff
3: enp0s20u3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UNKNOWN group default qlen 1000
    link/ether 8a:10:9e:df:40:22 brd ff:ff:ff:ff:ff:ff
    inet 192.168.42.135/24 brd 192.168.42.255 scope global dynamic noprefixroute enp0s20u3
       valid_lft 3511sec preferred_lft 3511sec
    inet6 …/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
```

`ip addr list` when not connected to the internet:
```
[...]

PR #5727, opened 2019-06-27
Change Client IP Address selection logic to use routing tables

bojidar-bg:
Addresses #5568.

Haven't tested the pull request, but at least it builds. If someone knows of an easy way to test a PR to Orleans under Linux, I would be glad to hear.

Based on code from [this StackOverflow question](https://stackoverflow.com/questions/15086863/c-sharp-know-with-which-ip-we-use-to-communicate-with-another-ip).

This PR is mostly a proof-of-concept for now.

### DIFF
--- a/src/Orleans.Core/Configuration/ConfigUtilities.cs
+++ b/src/Orleans.Core/Configuration/ConfigUtilities.cs
@@ -424,45 +424,58 @@ namespace Orleans.Runtime.Configuration
         /// If there are multiple addresses in the correct family in the server's DNS record, the first will be returned.
         /// </summary>
         /// <returns>The server's IPv4 address.</returns>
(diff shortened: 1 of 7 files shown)


sergeybykov:
@bojidar-bg This is still a draft. Curious if you are interested in continuing it. Thanks.

bojidar-bg:
@sergeybykov Well, I'm currently not working with Orleans, so my interest in working on the PR has declined. I would happily complete it if I am sure that the feature is wanted by the maintainers though.

sergeybykov:
We believe that functionality like this makes more sense outside of Orleans, maybe as a helper class to figure out what IP endpoints to use, so that those discovered endpoints are passed as part of Orleans configuration. That would avoid having this corner case complexity in the Orleans codebase and would clearly separate the concerns.

PR #5735, opened 2019-07-03
Use nameof instead of magic string

Youssef1313:


ReubenBond:
Thank you, @Youssef1313!

PR #5737, opened 2019-07-05
Directory: perform early handoff & forward to successor

ReubenBond:
This PR tweaks the behavior of the grain directory during membership change (startup, shutdown):

* During startup, wait until the local silo has received a "split partition" before allowing the local directory to serve requests
* During shutdown, perform handoff before any membership change has occurred
* During shutdown, after handoff has completed, forward all requests which should be served by the local silo (according to membership) to the would-be successor of the local silo (i.e, the silo which the local silo handed-off its directory partition to).
* Allow serving requests which should be served by a different silo (according to membership) if that silo has handed off its partition into this silo (i.e, indicating that it is in the process of shutting down).

### DIFF
--- a/src/Orleans.Runtime/GrainDirectory/DirectoryMembershipSnapshot.cs
+++ b/src/Orleans.Runtime/GrainDirectory/DirectoryMembershipSnapshot.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics.Contracts;
 using System.Text;
@@ -33,6 +32,7 @@ namespace Orleans.Runtime.GrainDirectory
             {
                 if (member.Value.Status == SiloStatus.Active)
                 {
+                    ++this.ActiveMemberCount;
                     var silo = member.Value.SiloAddress;
                     activeMembers.Add(silo);
                 }
@@ -61,6 +61,11 @@ namespace Orleans.Runtime.GrainDirectory
         /// The monotonically increasing membership version associated with this snapshot.
         /// </summary>
         public ClusterMembershipSnapshot ClusterMembership { get; }
+
+        /// <summary>
+        /// The number of active silos.
+        /// </summary>
+        public int ActiveMemberCount { get; }
 
         /// <summary>
         /// Returns the <see cref="SiloAddress"/> which owns the directory partition of the provided grain.
@@ -100,7 +105,7 @@ namespace Orleans.Runtime.GrainDirectory
                 }
             }
 
-            if (siloAddress == null)
+            if (siloAddress is null)
             {
                 // If not found in the traversal, last silo will do (we are on a ring).
                 // We checked above to make sure that the list isn't empty, so this should always be safe.
--- a/src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
+++ b/src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Orleans.Runtime.Scheduler;
@@ -21,13 +20,21 @@ namespace Orleans.Runtime.GrainDirectory
         private readonly LocalGrainDirectory localDirectory;
         private readonly IClusterMembershipService clusterMembership;
         private readonly IInternalGrainFactory grainFactory;
-        private readonly Dictionary<SiloAddress, GrainDirectoryPartition> directoryPartitionsMap;
-        private readonly List<SiloAddress> silosHoldingMyPartition;
-        private readonly Dictionary<SiloAddress, Task> lastPromise;
+        private readonly Dictionary<SiloAddress, GrainDirectoryPartition> ongoingHandoffs;
         private readonly ILogger logger;
         private readonly Factory<GrainDirectoryPartition> createPartion;
         private readonly Queue<(string name, Func<Task> action)> pendingOperations = new Queue<(string name, Func<Task> action)>();
         private readonly AsyncLock executorLock = new AsyncLock();
+
+        /// <summary>
+        /// Whether or not this silo has handed off its directory to another silo.
+        /// </summary>
+        public bool HasPerformedHandoff { get; private set; }
+
+        /// <summary>
+        /// Whether or not this silo has received a directory split from another silo.
+        /// </summary>
+        public bool HasReceivedSplit { get; private set; }
 
         internal GrainDirectoryHandoffManager(
             ILocalSiloDetails localSiloDetails,
@@ -43,37 +50,19 @@ namespace Orleans.Runtime.GrainDirectory
             this.clusterMembership = clusterMembership;
             this.grainFactory = grainFactory;
             this.createPartion = createPartion;
-            directoryPartitionsMap = new Dictionary<SiloAddress, GrainDirectoryPartition>();
-            silosHoldingMyPartition = new List<SiloAddress>();
-            lastPromise = new Dictionary<SiloAddress, Task>();
+            ongoingHandoffs = new Dictionary<SiloAddress, GrainDirectoryPartition>();
         }
 
-        internal List<ActivationAddress> GetHandedOffInfo(GrainId grain)
+        private async Task HandoffMyPartitionUponStop(Dictionary<GrainId, IGrainInfo> batchUpdate, SiloAddress destination, bool isFullCopy)
         {
-            lock (this)
-            {
-                foreach (var partition in directoryPartitionsMap.Values)
-                {
-                    var result = partition.LookUpActivations(grain);
-                    if (result.Addresses != null)
-                        return result.Addresses;
-                }
-            }
-            return null;
-        }
-
-        private async Task HandoffMyPartitionUponStop(Dictionary<GrainId, IGrainInfo> batchUpdate, List<SiloAddress> silosHoldingMyPartitionCopy, bool isFullCopy)
-        {
-            if (batchUpdate.Count == 0 || silosHoldingMyPartitionCopy.Count == 0)
+            if (batchUpdate.Count == 0 || destination is null)
             {
                 if (logger.IsEnabled(LogLevel.Debug)) logger.Debug((isFullCopy ? "FULL" : "DELTA") + " handoff finished with empty delta (nothing to send)");
                 return;
             }
 
             if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("Sending {0} items to my {1}: (ring status is {2})", 
-                batchUpdate.Count, silosHoldingMyPartitionCopy.ToStrings(), localDirectory.DirectoryMembershipSnapshot.ToDetailedString());
-
-            var tasks = new List<Task>();
+                batchUpdate.Count, destination, localDirectory.DirectoryMembershipSnapshot.ToDetailedString());
 
             var n = 0;
             var chunk = new Dictionary<GrainId, IGrainInfo>();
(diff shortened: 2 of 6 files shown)


src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -569,26 +613,17 @@ public Task UnregisterAfterNonexistingActivation(ActivationAddress addr, SiloAdd
 
         public async Task UnregisterAsync(ActivationAddress address, UnregistrationCause cause, int hopCount)
         {
+            var callerMembershipVersion = default(MembershipVersion);
```
ReubenBond: The idea is that eventually we will send the caller's membership version along with requests, but for now we declare it to be 'default', which indicates to `GetPartitionOwner` that if the local silo does not believe it's the correct silo for a request then it will refresh to the latest membership before continuing.

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -569,26 +613,17 @@ public Task UnregisterAfterNonexistingActivation(ActivationAddress addr, SiloAdd
...
-            {
-                await this.RefreshMembership();
-                forwardAddress = this.CheckIfShouldForward(address.Grain, hopCount, "UnregisterAsync");
-                if (forwardAddress != null)
-                {
-                    this.log.LogWarning($"UnregisterAsync - It seems we are not the owner of activation {address}, trying to forward it to {forwardAddress} (hopCount={hopCount})");
-                }
-            }
-
-            if (forwardAddress == null)
```
ReubenBond: This behavior is now consolidated into `GetPartitionOwner`, which replaces `CheckIfShouldForward`

src/Orleans.Runtime/GrainDirectory/DirectoryMembershipSnapshot.cs
```diff
@@ -100,7 +105,7 @@ internal SiloAddress CalculateGrainDirectoryPartition(GrainId grainId)
                 }
             }
 
-            if (siloAddress == null)
+            if (siloAddress is null)
```
veikkoeeva: I see! :) I wonder if it would be educational to include a comment for https://stackoverflow.com/questions/40676426/what-is-the-difference-between-x-is-null-and-x-null/40676671 or maybe even something like https://twitter.com/fransbouma/status/1115201371182071808 ? Personally I have enjoyed seeing gems like this explained, also first drawn attention to if otherwise missing while scanning.

Some comments were the reason I initially became intereste of the first public release of Orleans. :)

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -196,6 +186,12 @@ internal async Task ProcessSiloStoppingEvent(DirectoryMembershipSnapshot members
             Dictionary<GrainId, IGrainInfo> batchUpdate = localDirectory.DirectoryPartition.GetItems();
 
             await HandoffMyPartitionUponStop(batchUpdate, silosHoldingMyPartitionCopy, true);
+
+            // This will cause remote write requests to be forwarded to the silo that will become the new owner.
+            // Requests might bounce back and forth for a while as membership stabilizes, but they will either be served by the
+            // new owner of the grain, or will wind up failing. In either case, we avoid requests succeeding at this silo after we've
+            // completed handoff, which could cause them to not get handed off to the new owner.
+            this.HasPerformedHandoff = true;
```
veikkoeeva: Should this comment be added to https://github.com/dotnet/orleans/pull/5737/files#diff-eaa3e506dd98e5155c01b308f0dd4962R33 too?

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -212,37 +208,79 @@ internal void ProcessSiloAddEvent(DirectoryMembershipSnapshot membershipSnapshot
...
-                    grain =>
+                if (successor.Equals(addedSilo))
+                {
+                    // split my local directory and send to my new immediate successor his share
+                    if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("Splitting my partition between me and " + addedSilo);
+                    GrainDirectoryPartition splitPart = localDirectory.DirectoryPartition.Split(
+                        grain =>
+                        {
+                            var s = membershipSnapshot.CalculateGrainDirectoryPartition(grain);
+                            return (s != null) && !this.localSiloDetails.SiloAddress.Equals(s);
```
veikkoeeva: As a side note to interested readers, it's something to think how much work it would be to move from `SiloAddress` as in in `IPEndpoint` within to other ways to identify siloes.

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -212,37 +208,79 @@ internal void ProcessSiloAddEvent(DirectoryMembershipSnapshot membershipSnapshot
...
-            
+                        if (logger.IsEnabled(LogLevel.Debug))
+                        {
+                            logger.LogDebug("Splitting partition of {Predecessor} and creating a copy for {AddedSilo}", predecessorOfNewSilo, addedSilo);
+                        }
+
+                        GrainDirectoryPartition splitPart = partitionToSplit.Split(
+                            grain =>
+                            {
+                                // Need to review the 2nd line condition.
```
veikkoeeva: Why it needs to be reviewed? Can the reason be added to comments succintly?

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -212,37 +208,79 @@ internal void ProcessSiloAddEvent(DirectoryMembershipSnapshot membershipSnapshot
...
+
                 // remove partition of one of the old successors that we do not need to now
                 SiloAddress oldSuccessor = directoryPartitionsMap.FirstOrDefault(pair => !successor.Equals(pair.Key)).Key;
                 if (oldSuccessor == null) return;
 
-                if (logger.IsEnabled(LogLevel.Debug)) logger.Debug("Removing copy of the directory partition of silo " + oldSuccessor + " (holding copy of " + addedSilo + " instead)");
+                if (logger.IsEnabled(LogLevel.Debug))
+                {
+                    logger.LogDebug(
+                        "Removing copy of the directory partition of silo {OldSuccessor} (holding copy of {AddedSilo} instead)",
```
veikkoeeva: A tangential note, I wonder at which point and which way, if at all, it would make sense to move into using https://docs.microsoft.com/fi-fi/dotnet/api/microsoft.extensions.logging.eventid ? As one big refactoring or a plugin at a time? I could imagine this could be tied to the refactoring @xiazen is doing about loggin and defining a set of IDs (as `const` with appropriate messages) and then mandating/testing any one is only used once. Would make add to choices to build analytics on running clusters (SIEM). An example I have in mind (just composing something here as an example):

```
/// <summary>
    /// Ties together the event ID and corresponding format string.
    /// </summary>
    [DebuggerDisplay("EventStruct(Id = Id, FormatString = {FormatString})")]
    public readonly struct EventStruct
    {
        /// <summary>
        /// The ID of the logging event. This should be unique per purpose.
        /// </summary>
        public EventId Id { get; }

        /// <summary>
        /// The string used to format the message related to the <see cref="Id"/>.
        /// </summary>
        public string FormatString { get; }


        /// <summary>
        /// Constructs an event together with its formatting string.
        /// </summary>
        /// <param name="id">The ID of the logging event. This should be unique per purpose.</param>
        /// <param name="formatString">The string used to format the message related to the <paramref name="id"/>.</param>
        public EventStruct(in EventId id, string formatString)
        {
            Id = id;
            FormatString = formatString ?? throw new ArgumentNullException(nameof(formatString));
        }
    }

 /// <summary>
    /// Events used to log application activities.
    /// </summary>
    public static class Events
    {
        /// <summary>
        /// Something, something...
        /// </summary>
       /// <remarks>Check called only once in the framework.</remarks>
        public static EventStruct Something { get; } = new EventStruct(new EventId(1, nameof(Something)), "Something in {something}.");
    }
```

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -212,37 +208,79 @@ internal void ProcessSiloAddEvent(DirectoryMembershipSnapshot membershipSnapshot
...
+                }
+
                 directoryPartitionsMap.Remove(oldSuccessor);
             }
         }
 
-        private async Task ProcessAddedSiloAsync(DirectoryMembershipSnapshot membershipSnapshot, SiloAddress addedSilo, List<ActivationAddress> splitPartListSingle, List<ActivationAddress> splitPartListMulti)
+        private async Task ProcessAddedSiloAsync(
+            DirectoryMembershipSnapshot membershipSnapshot,
+            SiloAddress addedSilo, List<ActivationAddress> splitPartListSingle,
```
veikkoeeva: Nit: should these parameters be on a separate line?

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -437,35 +436,89 @@ private bool IsValidSilo(SiloAddress silo)
...
+                if (this.MyAddress.Equals(result) && !this.initialStabilizationComplete)
+                {
+                    var remainingWaitCycles = 80;
+
+                    // Wait for handoff only if there is more than one member in the cluster.
+                    while (this.DirectoryMembershipSnapshot.ActiveMemberCount > 1
+                        && !this.HandoffManager.HasReceivedSplit
+                        && remainingWaitCycles-- > 0)
+                    {
+                        await TimerManager.Delay(TimeSpan.FromMilliseconds(50));
```
veikkoeeva: I wonder if there is a heuristic for this `50 ms` and the `80` cycles? If there is, maybe name constants. Otherwise probably OK as in "shaked something out of stetson".

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
veikkoeeva: Should this be turned to a comment in code? I mean it's easier to read comments in code rather than Git history.

src/Orleans.Runtime/GrainDirectory/DirectoryMembershipSnapshot.cs
ReubenBond: Such a comment wouldn't be specific to the code, but rather a discussion on general-purpose C# language constructs. I think the meaning is too obvious to include it in every place that a coding style is used. A curious user can find that stack overflow answer.

If we add such comments & links everywhere then the code will drown & intention will be lost in discussion of language semantics.

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
ReubenBond: Not sure, this code just moved from one place to another (i.e, not my code). Could probably remove it.

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
ReubenBond: Leveraging `EventId` and the [`LoggerMessage.Define` pattern](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?view=aspnetcore-2.2) would be beneficial, but I want to save it for another time

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
ReubenBond: It's fairly arbitrary

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
ReubenBond: Possibly the first part (~"eventually pass this as a parameter") should, but the second part is an implementation detail of GetPartitonOwner

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -437,35 +436,89 @@ private bool IsValidSilo(SiloAddress silo)
...
+                    }
+
+                    // Even if a split partition has not been received yet, continue as though it has
```
jason-bragg: To get a sense of how well this loop is behaving as intended, can we log how long this loop takes to resolve, and the resulting this.HandoffManager.HasReceivedSplit?

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
ReubenBond: That would be great :)

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -212,37 +208,79 @@ internal void ProcessSiloAddEvent(DirectoryMembershipSnapshot membershipSnapshot
...
+                    List<ActivationAddress> splitPartListSingle = splitPart.ToListOfActivations(true);
+                    List<ActivationAddress> splitPartListMulti = splitPart.ToListOfActivations(false);
+
+                    EnqueueOperation(
+                        $"{nameof(ProcessSiloAddEvent)}({addedSilo})",
+                        () => ProcessAddedSiloAsync(membershipSnapshot, addedSilo, splitPartListSingle, splitPartListMulti));
+                }
+                else
+                {
+                    // adjust partitions by splitting them accordingly between new and old silos
```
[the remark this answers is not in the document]
ReubenBond: (discussed in-person) This is just restoring how the code was in 2.3.5

src/Orleans.Runtime/GrainDirectory/GrainDirectoryHandoffManager.cs
```diff
@@ -368,6 +406,8 @@ private async Task AcceptExistingRegistrationsAsync(List<ActivationAddress> sing
                     }
                 }
             }
+
+            this.HasReceivedSplit = true;
```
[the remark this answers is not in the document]
ReubenBond: (discussed in-person) This is never cleared - this is only used during startup to signal that the local directory can start processing requests

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
ReubenBond: sure thing

src/Orleans.Runtime/GrainDirectory/DirectoryMembershipSnapshot.cs
```diff
@@ -62,6 +62,11 @@ private static int CompareSiloAddressesForRing(SiloAddress left, SiloAddress rig
         /// </summary>
         public ClusterMembershipSnapshot ClusterMembership { get; }
 
+        /// <summary>
+        /// The number of active silos.
+        /// </summary>
+        public int ActiveMemberCount { get; }
```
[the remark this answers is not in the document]
ReubenBond: I made the distinction here just because we may implement what we were discussing with each silo owning multiple points on the ring to improve balance. If we did that then `ring.Count` may no longer be consistent with what this property represents.

The info cannot get out of sync: this is an immutable class

PR #5739, opened 2019-07-08
Fixed reminder issue.

jonas-johansson:
**Symptom**

Not all registered reminders were received when having a large set of registered reminders in a DynamoDBReminderTable.

**Reason**

DynamoDBReminderTable.ReadRows calls DynamobDBStorage.ScanAsync which in turn calls AmazonDynamoDBClient.ScanAsync once. Calling AmazonDynamoDBClient.ScanAsync only once is the problem, because in certain cases it bails out and needs to be called again to continue the scan from where it left off. Those cases are listed here: https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/DynamoDBv2/MDynamoDBScanAsyncScanRequestCancellationToken.html

**Solution**

Call AmazonDynamoDBClient.ScanAsync until the scan has completed.

### DIFF
--- a/src/AWS/Shared/Storage/DynamoDBStorage.cs
+++ b/src/AWS/Shared/Storage/DynamoDBStorage.cs
@@ -1,4 +1,4 @@
-﻿using Amazon.DynamoDBv2;
+using Amazon.DynamoDBv2;
 using Amazon.DynamoDBv2.Model;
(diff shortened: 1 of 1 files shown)


xiazen:
LGTM. Thanks for spotting this issue!

PR #5740, opened 2019-07-09
WIP - Silo Service Placement

jason-bragg:
Add placement strategy to place services on specific silos using grains.  Should be able to replace system targets.

### DIFF
--- a/src/Orleans.Core.Abstractions/Placement/PlacementAttribute.cs
+++ b/src/Orleans.Core.Abstractions/Placement/PlacementAttribute.cs
@@ -88,7 +88,7 @@ namespace Orleans.Placement
 
     public static class SiloServicePlacementKeyFormat
     {
-        private const char Separator = '.';
+        private const char Separator = '|';
         private static readonly char[] Separators = { Separator };
 
         public static bool TryParsePrimaryKey(string primaryKey, out string key)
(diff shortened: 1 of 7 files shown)


src/Orleans.Core.Abstractions/Placement/PlacementAttribute.cs
```diff
@@ -69,4 +70,75 @@ public sealed class ActivationCountBasedPlacementAttribute : PlacementAttribute
...
+
+        internal static string BuildPrimaryKey(SiloAddress silo, string key)
+        {
+            return $"{BuildSiloTag(silo)}{key}";
+        }
+
+
+        private static string BuildSiloTag(SiloAddress silo)
+        {
+            return $"{silo.GetConsistentHashCode()}{Separator}";
```
ReubenBond: Why not use a parseable SiloAddress? I think the intent would be more obvious (debugging, too)

src/Orleans.Runtime/Placement/SiloServicePlacementDirector.cs
```diff
@@ -0,0 +1,46 @@
...
+            return TrySelectActivationSynchronously(strategy, target, runtime, out PlacementResult placementResult)
+                ? Task.FromResult(placementResult)
+                : Task.FromResult(default(PlacementResult));
+        }
+
+        public bool TrySelectActivationSynchronously(
+            PlacementStrategy strategy, GrainId target, IPlacementRuntime runtime, out PlacementResult placementResult)
+        {
+            AddressesAndTag addressesAndTag;
+            if (runtime.FastLookup(target, out addressesAndTag) && addressesAndTag.Addresses.Count > 0)
```
ReubenBond: We can avoid the directory entirely

src/Orleans.Core.Abstractions/Placement/PlacementAttribute.cs
jason-bragg: I wanted to use something that I was confident wouldn't have the separator, else I'd need to normalize.  I'll switch to normalized parseable string.

src/Orleans.Runtime/Placement/SiloServicePlacementDirector.cs
ReubenBond: (resolved: this is correct behavior)

jason-bragg: Can, but it doesn't hurt to call it in case addr is in local cache.  Sticking to normal flow as much as possible.

jason-bragg:
prototype work, but we should probable do this in a less kludgy way.  Closing

PR #5745, opened 2019-07-10
HostedClient - use a slim IClusterClient implementation

ReubenBond:
Instead of re-using `ClusterClient`, use a purpose-built `IClusterClient` implementation on silos.

This simplifies behavior by avoiding `ClusterClient`, which was intended for use with external clients & performs a bunch of work which is not needed on the silo (eg, connect/disconnect from the cluster).


ReubenBond:
Ignore test failure, it's a flaky test (which is now disabled)

PR #5751, opened 2019-07-13
Fix incorrectly configured listening ports in tests

ReubenBond:
Legacy configuration support messes with how port configuration works

### DIFF
--- a/src/Orleans.TestingHost/TestClusterBuilder.cs
+++ b/src/Orleans.TestingHost/TestClusterBuilder.cs
@@ -154,7 +154,7 @@ namespace Orleans.TestingHost
                 basePort = basePort - (basePort % consecutivePortsToCheck);
                 int endPort = basePort + consecutivePortsToCheck;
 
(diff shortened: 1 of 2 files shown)


ReubenBond:
Some of our tests only pass because we are passing `SO_REUSEADDR` to the sockets (which is [not advisable](https://docs.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse#using-so_reuseaddr)) - those tests are lucky that they don't check behavior of separate silo/gw endpoints (eg, like some forwarding tests do).

The rest pass because our default test cluster uses legacy config which happens to configure the socket options correctly.

If I comment out this line then we see that a bunch of BVTs which already do not use legacy configuration fail:
image

image

The error is `System.Net.Sockets.SocketException : Only one usage of each socket address (protocol/network address/port) is normally permitted`

ReubenBond:
This also caused me a little grief in the Bedrock PR, since that doesn't use SO_REUSEADDR

PR #5759, opened 2019-07-16
Replace Environment.FailFast with Environment.Exit

benjaminpetit:
On some machine,  `Environment.FailFast` triggers a popup "xxx has stopped working". We believe it can stalls some deployments when they encounter this error.

### DIFF
--- a/src/Orleans.Core/Logging/ILoggerExtensions.cs
+++ b/src/Orleans.Core/Logging/ILoggerExtensions.cs
@@ -275,7 +275,9 @@ namespace Orleans.Runtime
             {
                 logger.Error(ErrorCode.Logger_ProcessCrashing, "INTERNAL FAILURE! Process crashing!");
 
-                Environment.FailFast("Unrecoverable failure: " + message);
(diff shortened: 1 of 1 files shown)


src/Orleans.Core/Logging/ILoggerExtensions.cs
```diff
@@ -275,7 +275,9 @@ internal static void Fail(this ILogger logger, ErrorCode errorCode, string messa
             {
                 logger.Error(ErrorCode.Logger_ProcessCrashing, "INTERNAL FAILURE! Process crashing!");
 
-                Environment.FailFast("Unrecoverable failure: " + message);
+                // Environment.FailFast triggers a popup on some machine "xxx has stopped working"
+                // Environment.FailFast("Unrecoverable failure: " + message);
+                Environment.Exit(128);
```
ReubenBond: Delete the commented code?

benjaminpetit: I wanted to keep it if we want to switch again, but yes I can delete it

PR #5763, opened 2019-07-18
Fix #5686 - Json serialization with Postgres

### LINKED ISSUE #5686: Json serialization, references, wrong ordering
Using Json serialization when you have multiple references to the same object usually creates a graph such that the first time an object is encountered it has the properties and an `$id` which identifies that instance. Thereafter, in the JSON a `$ref` property is used to identify that pre-defined object and a reference is inserted.

I have a situation where the `$ref` comes _before_ the definition of the object, which means that is being deserialized to `null`. 

This is a real headscratcher, because I cant reproduce the same problem in isolation using a simpler object graph and JSON.NET. Butmy fairly complex grain state exhibits the problem _every time_.

Has anyone ever experienced this? 

Below is a (snipped) copy of my JSON demonstrating the problem. You'll notice 2 `"$ref" : "5"` that come before the definition of `"$id": "5"`:

```
{
  "$id": "1",
  "$type": "Redacted.GrainSTateTypeName, Redacted.GrainAssembly",
  "Block": {
    "$id": "2",
    "$type": "Redacted.TypeName, Redacted.Assembly",
    "Terms": {
      "$type": "Redacted.TypeName.SubTypeName[], Redacted.Assembly",
      "$values": [
        {
          "$ref": "5"
        }
      ]
    },
    "Action": "Transition",
    "Driver": {
      "$id": "27",
      "$type": "Redacted.OtherTypeName, Redacted.Assembly",
      "Terms": {
        "$type": "Redacted.TypeName.SubTypeName[], Redacted.Assembly",
        "$values": [
          {
            "$ref": "5"
          }
        ]
      },
     [....snip...]
    "currentStateRecord": {
      "$id": "4",
      "$type": "Redected.TypeName.StateTypeName, Redacted.Assembly",
      "State": 1,
      "Terms": {
        "$type": "Redacted.TypeName.SubTypeName[], Redacted.Assembly",
        "$values": [
          {
            "$id": "5",
            "Name": "qty",
            "$type": "Redacted.TypeName.SubTypeName, Redacted.Assembly",
            "value": 10
          }
        ]
      }
   [...snip...]
   }
}
```

srollinet:
fixes #5686
replace jsonb columns by text, to ensure the properties ordering is not modified.

### DIFF
--- a/src/AdoNet/Orleans.Persistence.AdoNet/PostgreSQL-Persistence.sql
+++ b/src/AdoNet/Orleans.Persistence.AdoNet/PostgreSQL-Persistence.sql
@@ -1,4 +1,4 @@
-﻿CREATE TABLE Storage
+CREATE TABLE Storage
 (
     grainidhash integer NOT NULL,
     grainidn0 bigint NOT NULL,
@@ -9,7 +9,7 @@
     serviceid character varying(150)  NOT NULL,
     payloadbinary bytea,
     payloadxml xml,
-    payloadjson jsonb,
+    payloadjson text,
     modifiedon timestamp without time zone NOT NULL,
     version integer
 );
@@ -28,7 +28,7 @@ CREATE OR REPLACE FUNCTION writetostorage(
     _serviceid character varying,
     _grainstateversion integer,
     _payloadbinary bytea,
-    _payloadjson jsonb,
+    _payloadjson text,
     _payloadxml xml)
     RETURNS TABLE(newgrainstateversion integer)
     LANGUAGE 'plpgsql'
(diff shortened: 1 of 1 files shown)


srollinet:
I don't know if it's worth doing a unit test that cover advanced serialization scenarios?

veikkoeeva:
@srollinet There are tests already, but they didn't catch this. :) They start at https://github.com/dotnet/orleans/blob/dad62cc16d4d6de671563f6b1e54196f617926a5/test/Extensions/TesterAdoNet/StorageTests/Relational/PostgreSqlStorageTests.cs .

This case can probably easily checked just so that someone alters a deployed table and checks if the problem goes away. Then also runs the tests just to be sure.

ashtonkj:
Question: Why replace it with text instead of just replacing it with `json` type? I've been using Orleans with Postgres for a while and I just updated the tables to json rather than jsonb as jsonb is the one that reorders the data.

veikkoeeva:
@ashtonkj The related discussion starts at https://github.com/dotnet/orleans/issues/5686#issuecomment-511956976 . It would be valuable if we could have your opinion. The short story is that ADO.NET handles it as .NET `string` type when it reads and writes and that if the database handles other than text types, it apparently consumes some more resources if also potentially storing the data more compactly and also more efficiently for subqueries. However Orleans doesn't do subqueries but reads all at once and anyone who needs to do "off-Orleans processing" can change the type. Then there might be potential problems, at least with `jsonb` there was.

ashtonkj:
@veikkoeeva - Happy with that rationale.

sergeybykov:
Thank you, @srollinet!

PR #5770, opened 2019-07-20
Remove options and logic for automatic resending of messages

### LINKED ISSUE #5667: Remove message retry functionality
It is a bad idea to blindly retry messages that fail to process/deliver. Because of that the default retry count is set to zero. We need to remove this code altogether to simplify the codebase and prevent the potential confusion. We should have done that as part of 2.0 effort.

sergeybykov:
Resolves #5667.

Automatic message resend was a bad idea. We forgot to remove the logic in 2.0.


ReubenBond:
@sergeybykov I force-pushed an update on top of master

PR #5775, opened 2019-07-22
docs: small spelling fixes and added links

ry8806:
spelling fixes

and added proper links to articles mentioned (previously just plain text)


sergeybykov:
Thank you, @ry8806!

PR #5776, opened 2019-07-22
Implement full-duplex silo-to-silo connections

ReubenBond:
This PR upgrades the networking protocol in a backwards-compatible manner to add support for full-duplex silo-to-silo connections.

Currently silos establish outgoing connections to each other in order to send messages. This results in each silo pair having two connections between them, where each connection is used for traffic in a single direction.

By sending a `SiloAddress` along with the connection preamble, we can support full-duplex connections & save on connections.


ReubenBond:
🚀 :shipit:  🚢

PR #5785, opened 2019-07-24
Networking fixes

ReubenBond:
* Fix wire protocol backwards compatibility with older versions
* Implement most feedback from #5436
* Add protocol version option for upgradeability of network protocol
* General cleanup in networking code

### DIFF
--- a/src/Orleans.Runtime/Networking/SiloConnection.cs
+++ b/src/Orleans.Runtime/Networking/SiloConnection.cs
@@ -8,16 +8,16 @@ using Orleans.Networking.Shared;
 
 namespace Orleans.Runtime.Messaging
 {
-    internal sealed class SiloConnection : Connection
+    internal sealed class SiloConnection : Connection, ISiloConnection
     {
         private readonly MessageCenter messageCenter;
         private readonly MessageFactory messageFactory;
         private readonly ISiloStatusOracle siloStatusOracle;
         private readonly ConnectionManager connectionManager;
-        private readonly SiloAddress myAddress;
-        private SiloAddress remoteSiloAddress;
+        private readonly ConnectionOptions connectionOptions;
 
         public SiloConnection(
+            SiloAddress remoteSiloAddress,
             ConnectionContext connection,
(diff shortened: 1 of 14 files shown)


src/Orleans.Runtime/Networking/SiloConnection.cs
```diff
@@ -158,20 +165,46 @@ protected override async Task RunInternal()
...
             finally
             {
-                if (!(this.remoteSiloAddress is null)) this.connectionManager.OnConnectionTerminated(this.remoteSiloAddress, this);
+                if (!(this.RemoteSiloAddress is null)) this.connectionManager.OnConnectionTerminated(this.RemoteSiloAddress, this);
             }
 
-            Task WritePreamble() => ConnectionPreamble.Write(this.Context, Constants.SiloDirectConnectionId, this.myAddress);
+            async Task WritePreamble()
+            {
+                await ConnectionPreamble.Write(
```
JanEggers: this does not need to be async just return the task

src/Orleans.Runtime/Networking/SiloConnectionFactory.cs
```diff
@@ -47,7 +49,24 @@ protected override Connection CreateConnection(ConnectionContext context)
...
+
+        private void EnsureInitialized()
+        {
+            if (!isInitialized)
+            {
+                lock (this.initializationLock)
+                {
+                    if (!isInitialized)
+                    {
+                        this.messageCenter = this.serviceProvider.GetRequiredService<MessageCenter>();
```
JanEggers: why cant you use  ctor injection for these services?

ReubenBond: Circular dependency

src/Orleans.Runtime/Networking/SiloConnection.cs
ReubenBond: It's a ValueTask

ReubenBond:
I'm splitting this up into smaller PRs

PR #5796, opened 2019-07-29
Send a snapshot of the membership table on gossip

benjaminpetit:
Right now, when a silo change a silo entry in the membership table, it will "gossip" the new silo/status to other silos in the cluster. Since silos can miss notifications, it's not safe for them to infer anything based on this information alone, so they have to go to storage and refresh the all table.

In this PR, now silo that has modified the membership table will send a full snapshot of it, with the version number. This way, other silos in the cluster can either ignore the gossip if they are already up-to-date (or ahead), or can simply use it directly, without reading in the storage.

This should reduce the time of membership change propagation and reduce contention on the membership table.


src/Orleans.Runtime/MembershipService/MembershipTableManager.cs
```diff
@@ -93,6 +93,19 @@ public async Task Refresh()
...
+        {
+            // Check if a refresh is underway
+            var pending = this.pendingRefresh;
+            if (pending != null && !pending.IsCompleted)
+            {
+                await pending;
+            }
+
+            // If we are behind, let's take directly the snapshot in param
+            this.updates.Publish(snapshot);
```
ReubenBond: I think this should be TryPublish instead, since the publish may fail if an equal/later version was already published

ReubenBond:
LGTM. Only other question is whether or not dead silos should be filtered from the snapshot. For now I think we're safe to include them, but we can keep that in mind for future.

PR #5799, opened 2019-07-30
Improve codegen's .NET Core 3 compatibility

### LINKED ISSUE #5793: MSBuild codegen failing when only .NET Core 3 preview SDK is available
It would appear something is causing `Microsoft.Orleans.CodeGenerator.MSBuild` to fall back to trying to use .NET Core 2.0.0 SDK when only .NET Core 3 Preview7 is available.

When using only 3.0 I'm getting the following message when attempting to build a .net standard 2.0 library containing a grain:

```
  It was not possible to find any compatible framework version
  The specified framework 'Microsoft.NETCore.App', version '2.0.0' was not found.
    - The following frameworks were found:
        3.0.0-preview7-27912-14 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]

  You can resolve the problem by installing the specified framework and/or SDK.

  The .NET Core frameworks can be found at:
    - https://aka.ms/dotnet-download
```

I've setup a reproduction case at https://github.com/Vhab/OrleansCodegenNetCore3

It's using docker as it's the easiest way to get a clean environment to test in.

It contains 3 images, one with .NET Core 2,2, one with 3.0p7 and one with both.
The first and last one build fine, but the one with only 3.0 fails.

The repo includes a readme with how to run the reproduction cases.

Vhab:
Fixes #5793 using @ReubenBond 's instructions.

Successfully compiles a minimal netstandard2.0 project on Linux .NET Core 2.2, 3.0 Preview 7 and a combination of the 2.
Previously this would fail on the 3.0 variant.

### DIFF
--- a/src/Orleans.CodeGenerator.MSBuild/runtimeconfig.template.json
+++ b/src/Orleans.CodeGenerator.MSBuild/runtimeconfig.template.json
@@ -1,0 +1,3 @@
+{
+    "rollForwardOnNoCandidateFx": 2
+}


ReubenBond:
It looks like we can use a `runtimeconfig.template.json` file instead of explicitly passing the command line option: https://natemcmaster.com/blog/2019/01/09/netcore-primitives-3/#runtimeconfigtemplatejson

I think that would be preferable. If we could use a csproj setting instead, that would be even better (but I don't believe we can)

Vhab:
That sounds quite sensible. I'll try it out a bit later as I'm wrapping up for today.

ReubenBond:
Thank you, @Vhab - much appreciated!

Vhab:
Yep, that did the trick as well, and is a much nicer solution.

ReubenBond:
Thank you, @Vhab 
@sergeybykov we should include this in the next patch release.

PR #5800, opened 2019-07-30
Remove LocalSilo from MembershipTableSnapshot. Detect death in gossip

ReubenBond:
Follow-up to #5796

### DIFF
--- a/src/Orleans.Runtime/MembershipService/MembershipTableManager.cs
+++ b/src/Orleans.Runtime/MembershipService/MembershipTableManager.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


src/Orleans.Runtime/MembershipService/MembershipTableManager.cs
```diff
@@ -102,6 +103,24 @@ public async Task RefreshFromSnapshot(MembershipTableSnapshot snapshot)
...
+                    this.KillMyselfLocally(msg);
+                }
+
+                snapshot = MembershipTableSnapshot.Create(localSiloEntry.WithStatus(this.CurrentStatus), snapshot);
+            }
+            else
+            {
+                snapshot = MembershipTableSnapshot.Create(this.CreateLocalSiloEntry(this.CurrentStatus), snapshot);
+            }
+
```
jason-bragg: Why modify the membership info?  Shouldn't we act on the authoritative version rather than modifying it?

ReubenBond: Will attempt to address in follow up PR

PR #5807, opened 2019-07-31
Introduce support for network protocol versioning

ReubenBond:
This PR modifies the connection preamble which is sent whenever a connection is established in order to support network protocol versioning in a forward- and backward-compatible manner.

For end-users, the protocol versions can only be incremented or decremented by one per deployment (i.e, no skipping ahead until all nodes are updated to at least version 1 on Orleans 3.0).

The current Orleans protocol does not send a preamble in both directions. Instead, the client/silo establishing the connection sends a preamble and the listening side receives it. This PR expands that behavior so that both sides send a preamble as long as the connecting node's protocol version is at least `Version2`.

This is a 3.0-only change and does not need to be back-ported to 2.x

### DIFF
--- a/src/Orleans.Runtime/Networking/SiloConnection.cs
+++ b/src/Orleans.Runtime/Networking/SiloConnection.cs
@@ -174,7 +174,7 @@ namespace Orleans.Runtime.Messaging
                         // Inbound connection
                         var protocolVersion = await ReadPreamble();
 
-                        // To support graceful transition to higher protocol versions, send a preamble the remote endpoint supports it.
+                        // To support graceful transition to higher protocol versions, send a preamble if the remote endpoint supports it.
                         if (protocolVersion >= NetworkProtocolVersion.Version2)
                         {
                             await WritePreamble();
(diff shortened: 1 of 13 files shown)


src/Orleans.Runtime/Networking/SiloConnection.cs
```diff
@@ -158,23 +165,67 @@ protected override async Task RunInternal()
...
+                if (this.connectionOptions.ProtocolVersion == NetworkProtocolVersion.Version1)
+                {
+                    // This version of the protocol does not support symmetric preamble, so either send or receive preamble depending on
+                    // Whether or not this is an inbound or outbound connection.
+                    if (this.RemoteSiloAddress is null)
+                    {
+                        // Inbound connection
+                        var protocolVersion = await ReadPreamble();
+
+                        // To support graceful transition to higher protocol versions, send a preamble the remote endpoint supports it.
```
sergeybykov: "send a preamble **if** the remote endpoint supports it" ?

PR #5811, opened 2019-07-31
Check that a silo is not known to be dead before attempting a connection

ReubenBond:
This also removes checks for dead silos from other locations, since those are not needed & can cause issues during shutdown in test scenarios (where SystemTargetBasedMembership is used)


ReubenBond:
@benjaminpetit what do you think?

PR #5812, opened 2019-07-31
Fixed bug preventing OnError from being called on batch consumers.

jason-bragg:
On complete as well.

### DIFF
--- a/src/Orleans.Core/Streams/Internal/StreamSubscriptionHandleImpl.cs
+++ b/src/Orleans.Core/Streams/Internal/StreamSubscriptionHandleImpl.cs
@@ -203,12 +203,20 @@ namespace Orleans.Streams
 
         public Task CompleteStream()
         {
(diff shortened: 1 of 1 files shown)


src/Orleans.Core/Streams/Internal/StreamSubscriptionHandleImpl.cs
```diff
@@ -203,12 +203,20 @@ private Task NextItems(IEnumerable<Tuple<T, StreamSequenceToken>> items)
 
         public Task CompleteStream()
         {
-            return this.observer == null ? Task.CompletedTask : this.observer.OnCompletedAsync();
+            return this.observer is null
```
ReubenBond: I think this line would be cleaner using null coalescence:

```C#
return this.observer?.OnCompletedAsync() ?? this.batchObserver?.OnCompletedAsync() ?? Task.CompletedTask;
```

PR #5814, opened 2019-07-31
Log a warning when blocking application messages in MessageCenter

ReubenBond:
The intent here it to make it more obvious when a message is being dropped without a response


SebastianStehle:
Whats the advantage of the IsEnabled check? Is the formatting and allocations not deferred anyway?

ReubenBond:
There's no significant advantage. Warning is almost always enabled anyway. It's largely just habit

PR #5815, opened 2019-07-31
Cleanup transaction confirmation worker logging.

jason-bragg:


src/Orleans.Transactions/State/ConfirmationWorker.cs
```diff
@@ -145,10 +145,10 @@ public async Task<bool> Confirmed()
                 {
                     await this.pending;
                     this.complete = true;
-                } catch(Exception)
+                } catch(Exception ex)
                 {
                     this.pending = null;
-                    logger.LogWarning($"Confirmation of transactation {transactionId} with timestamp {timestamp} to participant {paricipant} failed.  Retrying");
+                    logger.LogWarning(ex, "Confirmation of transaction {TransactionId} with timestamp {Timestamp} to participant {Participant} failed.  Retrying", this.transactionId, this.timestamp, this.participant);
```
ReubenBond: I believe the exception will often not be included in the log message unless you explicitly include it

jason-bragg: hmm.  My understanding is that this is the recommended way for logging exceptions.  We can put it in the message as well, but that will duplicate the information, and for a long exception, that could nearly double the size.
Thinks it's worth it?

ReubenBond: It just depends whether you want to see the exception in the output or not

jason-bragg: Let's start with this and add more if it's needed.

PR #5825, opened 2019-08-05
Reject failed activations and fix possible race condition

benjaminpetit:
- Add another terminal state for activation to reject messages when `OnActivateAsync`failed
- Unregister failed activation from the directory before marking it as `Invalid` to avoid forwarding messages to the same activation
- Do not forward messages if the activation is in the `Deactivating` state. All pending messages will be forwarded after the activation has been unregistered from the directory

### DIFF
--- a/src/Orleans.Runtime/Core/Dispatcher.cs
+++ b/src/Orleans.Runtime/Core/Dispatcher.cs
@@ -262,15 +262,7 @@ namespace Orleans.Runtime
         {
             lock (targetActivation)
             {
-                if (targetActivation.State == ActivationState.Invalid)
-                {
-                    ProcessRequestToInvalidActivation(
-                        message,
-                        targetActivation.Address,
-                        targetActivation.ForwardingAddress,
-                        "ReceiveRequest");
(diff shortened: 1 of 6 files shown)


src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -238,10 +238,10 @@ private void ReceiveResponse(Message message, ActivationData targetActivation)
         {
             lock (targetActivation)
             {
-                if (targetActivation.State == ActivationState.Invalid)
+                if (targetActivation.State == ActivationState.Invalid || targetActivation.State == ActivationState.FailedToActivate)
                 {
                     logger.Warn(ErrorCode.Dispatcher_Receive_InvalidActivation,
-                        "Response received for invalid activation {0}", message);
+                        "Response received for {state} activation {message}", targetActivation.State, message);
```
ReubenBond: I think we need to change this to `LogWarning((int)ErrorCode.`. Also, guidelines say to use PascalCase for the template parameter names

PR #5832, opened 2019-08-07
Stop background transaction processing when a grain deactivates

ReubenBond:
Several parts of the transaction processing system rely on background workers for transaction servicing. For example, there is a storage worker, a lock worker, and a confirmation worker.

When an activation terminates, these workers must be made aware so that they can also gracefully terminate.

This PR is a follow-up to #5821, extending the termination to the storage worker & lock worker

### DIFF
--- a/src/Orleans.Transactions/State/ReaderWriterLock.cs
+++ b/src/Orleans.Transactions/State/ReaderWriterLock.cs
@@ -21,6 +21,7 @@ namespace Orleans.Transactions.State
         private BatchWorker lockWorker;
         private BatchWorker storageWorker;
         private readonly ILogger logger;
+        private readonly IActivationLifetime activationLifetime;
 
         // the linked list of lock groups
         // the head is the group that is currently holding the lock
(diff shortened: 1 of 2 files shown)


src/Orleans.Transactions/State/ReaderWriterLock.cs
```diff
@@ -274,108 +277,114 @@ public async Task Rollback(Guid guid, bool notify)
 
         private async Task LockWork()
         {
-            var now = DateTime.UtcNow;
-
-            if (currentGroup != null)
+            // Stop pumping lock work if this activation is stopping/stopped.
+            if (this.activationLifetime.OnDeactivating.IsCancellationRequested) return;
```
jason-bragg: Would it make sense to update the BatchWorker to take a cancelation token and stop scheduling work, rather than having each Work call need to implement this?

src/Orleans.Transactions/State/TransactionQueue.cs
```diff
@@ -496,94 +498,100 @@ public int BatchableOperationsCount()
 
         private async Task StorageWork()
         {
-            try
+            // Stop if this activation is stopping/stopped.
+            if (this.activationLifetime.OnDeactivating.IsCancellationRequested) return;
+
+            using (this.activationLifetime.BlockDeactivation())
```
JoshCollinsMSFT: Not much of a concern about this going to deactivated between?  Will BlockDeactivation() throw if the deactivation has started?

ReubenBond: It won't, it's just a hint to not deactivate while some block of code is executing. It's currently internal to tx asm

ReubenBond:
I considered updating BatchWorker (also used in event sourcing), but I'm not sure it's worthwhile yet

ReubenBond:
🚢 choo choo

PR #5844, opened 2019-08-09
Fix potential deadlock between Catalog and LocalGrainDirectory

ReubenBond:
We've identified a potential silo deadlock:
* `LocalGrainDirectory.RemoveServer` (called on membership change) **locks** `membershipCache` and calls into the catalog which deactivates grains (taking a **lock** on each activation it's deactivating)
* `Catalog.RecoverFailedInitActivation` (called when a new activation cannot be added to the directory, eg because that silo just shut down) is executed under the activation's **lock** and calls into `LocalGrainDirectory.CalculateGrainDirectoryPartition` which tries to **lock** `membershipCache`

This PR is an attempt to fix that.

### DIFF
--- a/src/Orleans.Runtime/Utilities/ImmutableCollectionExtensions.cs
+++ b/src/Orleans.Runtime/Utilities/ImmutableCollectionExtensions.cs
@@ -1,0 +1,29 @@
+using System;
+using System.Collections.Immutable;
+
+namespace Orleans.Runtime.Utilities
+{
+    internal static class ImmutableCollectionExtensions
+    {
+        public static int FindIndex<T>(this ImmutableArray<T> collection, Func<T, bool> predicate)
+        {
+            var index = 0;
+            foreach (var item in collection)
+            {
+                if (predicate(item)) return index;
+            }
+
+            return -1;
+        }
+
+        public static int FindLastIndex<T>(this ImmutableArray<T> collection, Func<T, bool> predicate)
+        {
+            for (var index = collection.Length - 1; index >= 0; --index)
+            {
+                if (predicate(collection[index])) return index;
+            }
+
+            return -1;
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Orleans.Runtime/Utilities/ImmutableCollectionExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Orleans.Runtime.Utilities
+{
+    internal static class ImmutableCollectionExtensions
+    {
+        public static int FindIndex<T>(this ImmutableArray<T> collection, Func<T, bool> predicate)
+        {
+            var index = 0;
+            foreach (var item in collection)
+            {
+                if (predicate(item)) return index;
```
sergeybykov: This will always return 0. Was the intent here to return index of the item that satisfies the predicate?

ReubenBond: Oh that's terrible, can you submit a fix?

sergeybykov: Sure

ReubenBond: How embarassing

sergeybykov: Need more sleep? 😉

ReubenBond: Let's chalk it up to that 😅

ReubenBond: I wonder how many issues this caused in previous tests (this code is all based on the previous bits which were reverted) 🤔🤔🤔

PR #5860, opened 2019-08-13
[WIP] Support ConnectionOptions.OpenConnectionTimeout

ReubenBond:


src/Orleans.Core/Messaging/ClientMessageCenter.cs
```diff
@@ -212,9 +213,23 @@ async Task SendAsync(ValueTask<Connection> task, Message message)
...
+                                state => this.SendMessage((Message)state),
+                                message,
+                                CancellationToken.None,
+                                TaskCreationOptions.DenyChildAttach,
+                                TaskScheduler.Default);
+                        }
+                        else
+                        {
+                            this.RejectMessage(message, $"Unable to send message due to exception {exception}", exception);
+                        }
```
ReubenBond: The purpose here is to circuit break cases where no gateways are available. If a transient exception occurs, we will retry up to `DEFAULT_MAX_MESSAGE_SEND_RETRIES` times, but promptly reject if the count has been exceeded.

src/Orleans.Core/Messaging/ClientMessageCenter.cs
```diff
@@ -293,32 +308,28 @@ private ValueTask<Connection> GetGatewayConnection(Message msg)
             var gatewayConnection = this.connectionManager.GetConnection(addr);
             if (gatewayConnection.IsCompletedSuccessfully)
             {
-                var result = this.TryUpdateConnection(index, gatewayConnection.Result, weakRef);
-                if (result == null) return this.GetGatewayConnection(msg);
-                return new ValueTask<Connection>(result);
+                this.UpdateBucket(index, gatewayConnection.Result);
+                return gatewayConnection;
             }
```
ReubenBond: The other changes in this file are about simplifying the logic. We have a single lock for modifying the buckets & WeakRefs and we re-use the WeakRefs, updating them instead of replacing them. Hopefully it's easier to understand than before.

src/Orleans.Core/Networking/ConnectionManager.cs
```diff
@@ -17,11 +18,12 @@ internal sealed class ConnectionManager
         private static int nextConnection;
 
         private readonly ConcurrentDictionary<SiloAddress, ConnectionEntry> connections = new ConcurrentDictionary<SiloAddress, ConnectionEntry>();
+
         private readonly ConnectionOptions connectionOptions;
         private readonly ConnectionFactory connectionFactory;
         private readonly INetworkingTrace trace;
         private readonly CancellationTokenSource cancellation = new CancellationTokenSource();
-        private readonly object collectionLock = new object();
+        private readonly object lockObj = new object();
```
ReubenBond: There was poor performance/results in recovery tests and I wanted to simplify this class & remove the ~'async spin locks'. So this just uses one regular C# lock now, covering all modifications to `connections` (which is a `ConcurrentDictionary` because we don't lock on reads) and all modifications to the constituent `ConnectionEntry` objects, which are now mutable classes instead of immutable structs, since it simplified the logic (IMO)

src/Orleans.Core/Networking/ConnectionManager.cs
```diff
@@ -53,158 +55,162 @@ public int ConnectionCount
...
-
-                    if (entry.Connections.Length >= this.connectionOptions.ConnectionsPerEndpoint)
+                    if (entry.PendingConnection != null
+                        && entry.Connections.Length < this.connectionOptions.ConnectionsPerEndpoint)
                     {
-                        result = entry.Connections;
-                        break;
+                        // Start a new connection attempt in the background.
+                        this.GetConnectionAsync(endpoint).Ignore();
                     }
```
ReubenBond: Previously, when a `GetConnection(SiloAddress)` call was received, it would always try to create a new connection if we had less than `ConnectionsPerEndpoint` connections. That's not optimal since it stalls more messages than are needed. 

Instead we will now fire off an async attempt to get a connection and return an established connection synchronously.

src/Orleans.Core/Networking/ConnectionManager.cs
```diff
@@ -53,158 +55,162 @@ public int ConnectionCount
...
+                    return connection;
+                }
+
+                // Wait for the pending attempt to complete
+                var pendingTask = entry.PendingConnection;
+                if (pendingTask != null && !pendingTask.IsCompleted)
                 {
-                    entry = this.GetOrCreateEntry(address, ref acquiredConnectionLock);
+                    await pendingTask.ConfigureAwait(false);
+                }
```
ReubenBond: Previously there was a kind of 'async spinlock' whereby only one connection attempt for a given endpoint could execute at a time and the others would await a `Task.Delay(...)` before re-checking. Instead we now share a `Task` and other connection attempts await that.

src/Orleans.Core/Networking/ConnectionManager.cs
```diff
@@ -53,158 +55,162 @@ public int ConnectionCount
...
+                {
+                    pendingTask = entry.PendingConnection;
+                    if (pendingTask == null || pendingTask.IsCompleted)
                     {
-                        var ticks = Math.Max(lastFailure.Ticks, entry.LastFailure.Value.Ticks);
-                        lastFailure = new DateTime(ticks);
+                        pendingTask = this.ConnectAsync(endpoint);
+                        entry.PendingConnection = pendingTask;
                     }
+                }
```
ReubenBond: Note that we *start* the async connection attempt under the lock and await it outside of the lock. This is to ensure that we correctly set `PendingConnection` and don't have a race for multiple pending connections. Another alternative would be to use a `TaskCompletionSource` which could be assigned from within the lock but completed in the background by `ConnectAsync`

src/Orleans.Core/Networking/ConnectionManager.cs
```diff
@@ -53,158 +55,162 @@ public int ConnectionCount
...
                 {
-                    if (acquiredConnectionLock) entry.ReleaseLock();
+                    // Clean up the pending connection task.
+                    lock (this.lockObj)
+                    {
+                        if (ReferenceEquals(entry.PendingConnection, pendingTask))
+                        {
+                            entry.PendingConnection = null;
+                        }
+                    }
```
ReubenBond: We only update `PendingConnection` if we didn't lose the race.

src/Orleans.Core/Networking/ConnectionManager.cs
```diff
@@ -53,158 +55,162 @@ public int ConnectionCount
...
-                }
 
                 return entry;
             }
         }
 
-        private async ValueTask<Connection> ConnectAsync(SiloAddress address)
+        private async Task<Connection> ConnectAsync(SiloAddress address)
         {
+            await Task.Yield();
```
ReubenBond: Yield immediately - see the note about starting `ConnectAsync` under a lock

src/Orleans.Core/Networking/ConnectionManager.cs
```diff
@@ -383,42 +366,55 @@ private IDisposable BeginConnectionScope(Connection connection)
...
 
-        private struct ConnectionEntry
+        private class ConnectionEntry
         {
-            public readonly DateTime? LastFailure;
-            public readonly ImmutableArray<Connection> Connections;
-            private readonly int[] lockObj;
+            public Task PendingConnection { get; set; }
+            public DateTime? LastFailure { get; set; }
+            public ImmutableArray<Connection> Connections { get; set; }
```
ReubenBond: Connections can be accessed outside of a lock, so we use an immutable collection here.

src/Orleans.Runtime/GrainDirectory/LocalGrainDirectory.cs
```diff
@@ -486,9 +486,10 @@ public SiloAddress CalculateGrainDirectoryPartition(GrainId grainId)
                     if (Seed == null)
                     {
                         var errorMsg =
-                            $"MembershipTable cannot run without Seed node. Please check your silo configuration make sure it specifies a SeedNode element. " +
-                            $"This is in either the configuration file or the {nameof(NetworkingOptions)} configuration. " +
-                            " Alternatively, you may want to use reliable membership, such as Azure Table.";
+                            $"Development clustering cannot run without a primary silo. " +
+                            $"Please configure {nameof(DevelopmentClusterMembershipOptions)}.{nameof(DevelopmentClusterMembershipOptions.PrimarySiloEndpoint)} " +
+                            "or provide a primary silo address to the UseDevelopmentClustering extension. " +
+                            "Alternatively, you may want to use reliable membership, such as Azure Table.";
```
ReubenBond: This comment was already wrong, but I updated it because it referenced the now `[Obsolete]` `NetworkingOptions` type

src/Orleans.Runtime/Networking/SiloConnectionFactory.cs
```diff
@@ -32,6 +35,15 @@ internal sealed class SiloConnectionFactory : ConnectionFactory
             this.localSiloDetails = localSiloDetails;
         }
 
+        public override ValueTask<Connection> ConnectAsync(SiloAddress address, CancellationToken cancellationToken)
+        {
+            EnsureInitialized();
+
+            if (this.siloStatusOracle.IsDeadSilo(address)) throw new ConnectionAbortedException($"Denying connection to known-dead silo {address}");
```
ReubenBond: This check is performed here just in case there are multiple messages already waiting for a connection. This lets us largely avoid a race, eg when a silo is declared dead while several tasks are already waiting to get a `Connection` to send it a message.

There is still a case where the silo might be declared dead while a particular connection attempt is pending. We could theoretically fix but I don't think the added complexity would be worthwhile: the connection attempt will timeout fairly quickly and fail and retries will fail-fast.

ReubenBond:
Setting to WIP since I want to investigate something first

PR #5872, opened 2019-08-16
Remove unused "SetupSqlScriptFileNames" , It will cause the test to fail

bluexo:


veikkoeeva:
@bluexo You are right. I tested this myself too. The setup fails and as a conclusion the system marks the tests as inconclusive since it thinks there just wasn't MySQL or PostgreSQL installed without default settings to be connected. When I applied this fix, the tests succeeded.

I add a tangential issue here, the setup could be simpler and make better use of XUnit external configuration (which `TryLoadAndMergeWithCustomSettings` mimicks a bit).

PR #5874, opened 2019-08-16
Add release notes for 3.0.0-beta1

sergeybykov:


shersh:
What's with performance? Increased\Decreased or stagnation?
Any benchmarks?

ReubenBond:
@shersh The performance increase for the networking change was between ~8% and ~140% in our distributed load tests (depending on the test): https://github.com/dotnet/orleans/pull/5436#issuecomment-514424902

PR #5887, opened 2019-08-22
Improve graceful deactivation of grains performing transaction work

ReubenBond:
This PR is mostly intended to reduce log noise since this behavior is benign. Nevertheless, I believe it's worthwhile to clean up the behavior here.

If a grain fails to fully activate and has some transaction processing work which need to be performed then it's possible for the grain to get into a state where the background transaction workers will never terminate.

Additionally, when a grain deactivates while it's performing background transaction work, we may emit a warning to the effect of `"Enqueuing task {Task} to a work item group which should have terminated. Likely reasons are that the task is not being 'awaited' properly or a TaskScheduler was captured and is being used to schedule tasks after a grain has been deactivated."`
This PR attempts to prevent this situation by blocking deactivation until the background workers are aware of it and can gracefully terminate.

### DIFF
--- a/src/Orleans.Core/Timers/TimerManager.cs
+++ b/src/Orleans.Core/Timers/TimerManager.cs
@@ -8,40 +8,45 @@ namespace Orleans.Timers.Internal
 {
     public interface ITimerManager
     {
-        Task Delay(TimeSpan timeSpan);
+        Task<bool> Delay(TimeSpan timeSpan, CancellationToken cancellationToken = default(CancellationToken));
     }
 
     internal class TimerManagerImpl : ITimerManager
     {
-        public Task Delay(TimeSpan timeSpan) => TimerManager.Delay(timeSpan);
+        public Task<bool> Delay(TimeSpan timeSpan, CancellationToken cancellationToken = default(CancellationToken)) => TimerManager.Delay(timeSpan, cancellationToken);
     }
 
     internal static class TimerManager
     {
-        public static Task Delay(TimeSpan timeSpan) => DelayUntil(DateTime.UtcNow + timeSpan);
+        public static Task<bool> Delay(TimeSpan timeSpan, CancellationToken cancellationToken = default(CancellationToken)) => DelayUntil(DateTime.UtcNow + timeSpan, cancellationToken);
 
-        public static Task DelayUntil(DateTime dueTime)
+        public static Task<bool> DelayUntil(DateTime dueTime, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var result = new DelayTimer(dueTime);
+            var result = new DelayTimer(dueTime, cancellationToken);
             TimerManager<DelayTimer>.Register(result);
             return result.Completion;
         }
 
         private sealed class DelayTimer : ITimerCallback, ILinkedListElement<DelayTimer>
         {
-            private readonly TaskCompletionSource<int> completion =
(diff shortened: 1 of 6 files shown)


src/Orleans.Core/Timers/TimerManager.cs
```diff
@@ -8,40 +8,45 @@ namespace Orleans.Timers.Internal
...
 
-            public DelayTimer(DateTime dueTime)
+            public DelayTimer(DateTime dueTime, CancellationToken cancellationToken)
             {
                 this.DueTime = dueTime;
+                this.CancellationToken = cancellationToken;
             }
 
-            public Task Completion => this.completion.Task;
+            public Task<bool> Completion => this.completion.Task;
```
jason-bragg: Why return bool?

src/Orleans.Transactions/State/ConfirmationWorker.cs
```diff
@@ -80,17 +80,21 @@ private async Task NotifyAll(Guid transactionId, DateTime timestamp, List<Partic
...
-                    var results = await Task.WhenAll(confirmations.Select(c => c.Confirmed()));
-                    return results.Any(b => !b);
+                    var confirmationResults = await Task.WhenAll(confirmations.Select(c => c.Confirmed()));
+                    hasPendingConfirmations = false;
+                    foreach (var confirmed in confirmationResults)
+                    {
+                        hasPendingConfirmations |= !confirmed;
+                    }
+
+                    if (!hasPendingConfirmations) break;
```
jason-bragg: if we're breaking out here, and hasPendingConfirmations can't change before loop starts again, why have hasPendingConfirmations as a loop condition?

ReubenBond: Good point

src/Orleans.Core/Timers/TimerManager.cs
ReubenBond: I could revert this. The reasoning was for the caller to be able to distinguish between a canceled and completed delay call without throwing a cancellation exception (cancellation is not exceptional)

src/Orleans.Transactions/State/ConfirmationWorker.cs
```diff
@@ -80,17 +80,21 @@ private async Task NotifyAll(Guid transactionId, DateTime timestamp, List<Partic
...
-            }
-
-            async Task<bool> HasPendingConfirmations(List<Confirmation> values)
+            var hasPendingConfirmations = true;
+            while (!ct.IsCancellationRequested && hasPendingConfirmations)
             {
                 using (this.activationLifetime.BlockDeactivation())
                 {
-                    var results = await Task.WhenAll(confirmations.Select(c => c.Confirmed()));
-                    return results.Any(b => !b);
```
jason-bragg: not convinced perf benefit of avoiding link here justifies cost of simplicity/maintainability of any()

ReubenBond: I don't disagree. I simplified the logic in the latest commit

src/Orleans.Core/Timers/TimerManager.cs
jason-bragg: I'm not opposed to it, just thought since these were intended to replace task.Delay, they should probably follow the same pattern.  I'll leave it to you're discretion.

PR #5901, opened 2019-08-23
[2.4.2] Fix header deserialization error handling

benjaminpetit:
In `IncomingMessageBuffer.TryDecodeMessage`, do not set the `out` parameter `msg` if headers deserialization fails.

Otherwise, since `Message.Headers` is never `null`, we do not clear the buffer and close the socket. It might provoke some `NullReferenceException` later too.

### DIFF
--- a/src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs
+++ b/src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs
@@ -666,22 +666,31 @@ namespace Orleans.Runtime.Messaging
                         }
                         catch (Exception exception)
                         {
-                            // If deserialization completely failed or the message was one-way, rethrow the exception
-                            // so that it can be handled at another level.
-                            if (msg == null || msg.Direction != Message.Directions.Request)
+                            // If deserialization completely failed rethrow the exception so that it can be handled at another level.
+                            if (msg == null)
                             {
+                                // Logging will be done in the other catch statement
                                 throw;
                             }
 
-                            // The message body was not successfully decoded, but the headers were.
-                            // Send a fast fail to the caller.
-                            MessagingStatisticsGroup.OnRejectedMessage(msg);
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -668,7 +668,7 @@ public void ProcessReceived(SocketAsyncEventArgs e)
                         {
                             // If deserialization completely failed or the message was one-way, rethrow the exception
                             // so that it can be handled at another level.
-                            if (msg?.Headers == null || msg.Direction != Message.Directions.Request)
+                            if (msg == null || msg.Direction != Message.Directions.Request)
```
jason-bragg: I think the behavior we want here is:

if we have a header and the message is a 'request', generate a response, log and move on.
if we have a header and the message is not a request, we log and move on.
if we have no header, we throw, which closes socket and resets buffer.

Does that seem like what we want?

src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs
```diff
@@ -666,22 +666,31 @@ public void ProcessReceived(SocketAsyncEventArgs e)
...
-                            var response = this.messageFactory.CreateResponseMessage(msg);
-                            response.Result = Message.ResponseTypes.Error;
-                            response.BodyObject = Response.ExceptionResponse(exception);
-                            
-                            // Send the error response and continue processing the next message.
-                            this.IMA.MessageCenter.SendMessage(response);
+                            IMA.Log.Error(
+                                ErrorCode.MessagingProcessReceiveBufferException,
+                                $"Exception when deserializing message {msg}",
+                                exception);
```
ReubenBond: Will this actually print the exception to logs? I think we found that for the console logger it won't include the exception in the message

Issue #5905, opened 2019-08-25
Managing grains without subclassing: Getting the IGrainActivationContext?


bddckr:
# Summary

I need a `Grain` instance to pass to `IGrainRuntime.DeactivateOnIdle` and other public, accessible API - how do I get that without forcing the target grain type to have a reference to a `GrainServiceClient` that then itself can get to the `IGrainActivationContext`?

Should `Grain.Data` be `public` instead of `internal`? Is it a bad idea?

# Issue

How do you handle such things like using the API that is (seemingly) only available to grains themselves? E.g. `RegisterTimer`, `DelayDeactivation` and `DeactivateOnIdle`. I want to call such API from some managing grains "from the outside".

## Example

I have some patterns like a grain that should be kept alive. I want to re-use the implementation of that pattern but I don't like having this as a grain superclass to inherit from - composition is really needed here instead, as I have various differences in the kinds of grain implementations and we can't have multiple superclasses of course :smile:

What I ended up with are three things basically:
* An interface with a single method `StayActivated` that literally does nothing and just returns `Task.Completed`. (Thanks to C# 8's default interface implementations this is basically just a marker interface and needs no copy-pasta in each implementing grain!)
* A `GrainServiceClient` which any implementing grain gets via constructor injection, then keeps a reference and calls `KeepActivated(this)` or `AllowDeactivation(this)` whenever necessary. This client is added as a **scoped** service to ensure client instance gets the correct `IGrainActivationContext`.
* A `GrainService` to keep a list of all grains to repeatedly call `StayActivated` on.

The only reason this works is because I can get to calling the `Grain` instance in the client via `IGrainActivationContext.GrainInstance`. After the client delays the activation it also calls into the `GrainService` as I don't want the scoped client to do the keep-alive logic - it dies with the grain getting deactivated after all AFAIK.

While this allows me to re-use this pattern implementation now by just

* Tagging the grain types with the interface.
* Adding a constructor argument + reference to the `GrainServiceClient`.
* Calling the client's methods.

**I am still having a need to allow "outsiders" to tell the service to keep a different grain activated.**

SebastianStehle:
I think this is in general a good approach. Grains should be more composable. Right now you have to use inheritance to accomplish a lot of things and the different interface like IGrainWithStringKey and IGrainWithGuidKey make it even more harder to implement generic extensions.

SebastianStehle:
[...]

Issue #4519, opened 2018-04-27
Feature request: IActivationFilter


SebastianStehle:
In obit you have this extension interface: https://github.com/orbit/orbit/blob/master/actors/core/src/main/java/cloud/orbit/actors/extensions/LifetimeExtension.java 

I thing it would be great to introduce IActionFilter with a similar signature to IIncomingGrainCallFilter.

Use cases:

1. Custom storage solutions
2. Logging
3. Exception handling
4. Integration to Orleans Dashboard

ReubenBond:
How should the API look, @SebastianStehle/others?

SebastianStehle:
The IIncomingGrainCallFilter and IIncomingGrainCallContext is great. Would be good to have something similar with a Invoke delegate to control when your code is running.

ravarnamsft:
+1 on this request. We have a need for the same. It would be great to have something like IIncomingGrainCallFilter for activations and de-activations

mkowalskigps:
Is this something that is in scope for the 4.0 release?

ReubenBond:
@mkowalskigps if there's an API proposal, then we could. A PR would be welcome

hankovich:
Is there progress on this issue?

tomachristian:
Needed by us as well.

mikekistler:
As @ReubenBond indicated above, the next step is to create an API proposal. This does not need to be elaborate -- just describe the APIs to be added and what they would do / how they could be used. We would welcome a API proposal from the community and Reuben has offered to guide anyone willing to make the effort.

PR #5907, opened 2019-08-26
Memory usage for activation data improved.

SebastianStehle:
This PR creates a memory optimized version of the Grain lifecycle subject. The ConcurrentDictionary creates a lot of sparse arrays and consumes a lot of memory. In the context of the grain I think that concurrency is not a topic.

With concurrent dictionary (100k Grains)
image

With simple list (100k Grains)
image


SebastianStehle:
I was expecting this feedback. For me it seems that there are 3 lifecycle subjects:

* Clients
* Silos
* Grains

For my understanding thread safety is not needed for all three of them, but I am not sure.

SebastianStehle:
I just changed the base implementation to use a list instead.

SebastianStehle:
Before making more improvements it would be great to also consider other feature requests like:

* https://github.com/dotnet/orleans/issues/4519
* https://github.com/dotnet/orleans/issues/5905

I think the grain structure in general is hard to use, because it its not easy to add functionalities and it could also be easier to test grains.

sergeybykov:
I like this. If we don't have concurrency in the use cases today, no need to pay the perf price. If we might see a concurrent use, should we simply add an optional lock that would be used only if `LifecycleSubject` is instantiated with a concurrency flag?

Allocations caused by LINQ are a concern. We see some workloads where grains are activated to handle a single request. The impact of such allocations is even worse in those cases.

SebastianStehle:
The numbers are from a test where 50.000k grains are created, each for one request, but I have not seen the impact of the linq statements in the analysis.

jason-bragg:
The concurrency issue, imo, is not as much a question of what we do today, because the lifecycle is an extensibility point, but more a question of what do we support.  If modification of the lifecycle needs be performed in a single thread, then we need have some means of ensuring that users are aware of this and that patterns we support ensure this. 

Grains:
The patterns described in the grain lifecycle docs look to be thread safe.  Thoughts?

https://dotnet.github.io/orleans/Documentation/grains/grain_lifecycle.html

Silo
The patterns described in the silo lifecycle docs look to be thread safe.  Thoughts?

https://dotnet.github.io/orleans/Documentation/clusters_and_clients/silo_lifecycle.html

The patterns described in the lifecycle core system would require thread safe implementations, but thats mainly describing the abstractions, so we can deal with thread safety when we get there.

https://dotnet.github.io/orleans/Documentation/implementation/orleans_lifecycle.html

From the above, I think it's safe to make this change, as the proscribed patterns are safe with this change.  Someone please sanity check me here.

sergeybykov:
Thank you, @SebastianStehle!

PR #5909, opened 2019-08-26
Adds a sample that runs and tests locally in reliable configuration

### LINKED ISSUE #411: One-Box Deployment Scenarios
Transferring the discussion about One-Box deployment scenario from #408 

@veikkoeeva Wrote:

The scenario I have in mind is to run "cluster" locally. I'll get to the point later, but allow me to take a bit of your time and paint scene.

An example project could be two servers with Windows Services and whatnot, two web frontends and then an Orleans system of say, three silos. Having this working on a single developer machine by simply pressing F5 will allow even more novice developers form a picture how the overall system is supposed to interact with its constituent pieces and mocked integration points. It's easy to draw a A4 sized poster having nice pictures and then mapping the projects to appropriate nodes (e.g. a deployment diagram). Arrangement such as this allows also [exploratory testing](https://en.wikipedia.org/wiki/Exploratory_testing), real testers, do testing "real time" (not only hitting the system on sprint late or something like that). For instance, every time feature branch hits master, a tester can either continue testing with the local version or pull the newest from origin master.

Maybe less rational idea is that it's nice to _feel the system_. That is, it's not just tests running occasionally or dumping something to servers (in enterprise setting, access to servers isn't granted), but when developing the front-end, one could have the silos and others systems up (on one's leisure) and debug the web front-end against "a live system". The "feel the system" here, while not real servers, relays some feel of how the topology fits together and how the grains and other parts work together and that knowledge is to some extent transferable to reasoning the real system too.

One could run something like [chaos monkey](http://techblog.netflix.com/2010/12/5-lessons-weve-learned-using-aws.html) (heh, this is a bit like the colocation mentioned in the blog post) during a coffee break with coworkes just for the fun of it and observe the system and see it should be doing what planned. Part of the "feel the system" is feeding into human curiosity and pattern recognizer mind to spot what has not been thought about and make a good use of debugger that is readily available if something unexpected happens (as it will).

I noticed [OrleansHostWrapper](https://github.com/dotnet/orleans/blob/5e964005d0cb88f38f30123ba079ddb46513adbf/src/OrleansVSTools/VSProjectSiloHost/OrleansHostWrapper.cs) while developing my [gist](https://gist.github.com/veikkoeeva/4414609e828375c898f3), but I don't wish the wrapper to bundle an arguments parser (I'd prefer something like [UnionArgParser](https://nessos.github.io/UnionArgParser/tutorial.html) in any event) and whatnot together. I'd like to be a bare-bones cluster manager I can run both in my testing project (with the framework of my choice) as easily as in my gist. Moreover, I'd like to be able to shut down, restart and add silos at my leisure. Easy configurability and the possibility to decide if running in the same process or even in separate processes (this might have production usage too). My gist has some problems in that I'll change the naming stuff, redo a bit about the [configuration part](https://gist.github.com/veikkoeeva/4414609e828375c898f3#file-gistfile1-fs-L93) and many other things. One issue is that I may want to run an Orleans cluster in a console application (or more likely with [TopShelf](http://topshelf-project.com/)) in a faul-tolerant style when I get the SQL server part done. My hunch is that it should be nearly as fast as in-memory system.

Arranging a projects like this has worked rather well, at least in the projects I've been involved in. :) The last one, which is on the finishing lines (makes me too busy), has several servers running IIS, Windows Services, reverse proxies (with things like TLS termination), integration points and so forth and this has worked well. We stub out integration points and don't run reverse proxies (we could configure IIS Express instances if it it made sense), but otherwise the system locally mimicks that of production. Common problems are that, for instance, front-end developers may find that, for some reason, the data loading/inserting feels sluggish. The reason could be as simple as overly broad query to database, missing index (or bad index as loading data to system from an integration point soiled them), too much data put on the line or whatever, but it can be examined quickly, on the spot, in debugger, when it occurs sometimes as simply as checking the call stack. Which makes the more web oriented guys on occassion step through stored procedures too in order to trace something down (but that's another story, though great in educational sense).

So, in essence, I'd like to have an easy way to have the silos running like the APIs and front-ends on IIS Express on my example project and it'd be super-nice if I could put the same arrangment to test projects and run a more formal battery through it (as in the example project). In the particular project example I mentioned, we have the code and we share the test data with outside testers. It takes mere seconds to bulk-load about 300 MiB to LocalDb and if so needed, if the testers have found something, we can just backup their database, load it in on local system and see how it might have happened.

Assuming in future Orleans relational database there's less destruction on data (i.e. less UPDATE and DELETE clauses), it could be sensible to just load data in for troubleshooting locally and observe the system state or run analytical queries to dig out certain aspects. This scenario, and the above configurability in local debugging and testing (and production) may also explain my enthusiasm on having DI and interceptors and whanot so as to have the possibility to have a good visualization on what happens in the system should there be a need.

I'm not sure if this answers any of your questions or gives thoughts. This is more of a "systems thinking" point for me. In any event, what I'm after isn't doable in this PR (probably not even in several ones) and my vision is not something that is necessarily a shared one. Though if you have questions, I'm glad to answer them.

veikkoeeva:
This adds a sample that:
- Can test multiple siloes in reliable configuration locally.
- Can start an API and a cluster locally in reliable configuration and test them.
- Shows how local storage, SQL Server in this case, can be used and even bulk-loaded
  with data in tests and how snapshots can be used.
- Shows how to do cross-cutting tests:
  - Capture log content from the system and test them.
  - Checks server sent headers.
- Shows some real deployment issues such as security.
- Allows to extend with same technique to Storage Emulator in tests and locally.
- Allows to extend identity testing, e.g. IdentityServer with the same technique.

Closes #411


veikkoeeva:
This isn't likely good to go yet, but wanted to finally open this in case it would be useful to others.

There is more into the sample here than that description, but I'm a bit pressed on time so I write later a better one. My thinking is to later just add features on top of this, such as IdentityServer, Trill and others and test them likewise in a cross-cutting manner. Even add chaos engineering aspects. One is also a PWA front-end that would use https://github.com/kblok/puppeteer-sharp or something similar then to run the tests through the whole system.

The gist is to run Orleans locally in relieable mode, both in tests and also when debugging. I have privately a demo where I have IdentityServer also plugged in so that the system can test potentially even complex policy handling while testing. Though it needs to be cleaned up and I took a slight shortcut so that in tests it also open resource owner flow (password and username) while I would like to in tests too use just PKCE (with the web), we'll see how this goes.

sergeybykov:
This is epic!

I just wonder if there's a more descriptive name for it than "BigSample." 🙂

veikkoeeva:
@sergeybykov Perhaps `OneBoxDeploymentSample`? :)

This is probably difficult to review in GH and I don't know how keen anyone is to download the branch to try it, but I think I will write a better documentation of how it works, the intended development flow and what things I intend to incorporate over the time and iron out some code kinks and pitfals (e.g. don't blurp out stack traces while in production and an internal error occurs). This code has been sitting on my computer almost a year. While the age shows in some parts (e.g. no IoC, yet), I thought I could try to push it over the edge already.

veikkoeeva:
I also think I should add a few general resources such as

- [How Complex Systems Fail](https://www.youtube.com/watch?v=2S0k12uZR14) by **Richard Cook**
- [Antics, drift, and chaos](https://www.youtube.com/watch?v=SM2uXpmyJmA) by **Lorin Hochstein**

Maybe articles such as [Simple testing can prevent most critical failures](https://blog.acolyer.org/2016/10/06/simple-testing-can-prevent-most-critical-failures/).

Goes to show a bit clearer how one might want to think complex systems and testing them. The tooling seem to be there too and isn't anymore invasive than the logging example, but probably have to put add in subsequent PRs

sergeybykov:
>@sergeybykov Perhaps OneBoxDeploymentSample? :)

I like "OneBoxDeployment". I think "Sample" is superfluous though. 🙂

veikkoeeva:
@sergeybykov, @ReubenBond can you see something here? Could be good to go as far as I'm concerned, me or someone can then refine things in separate PRs (e.g. update packages) later if there isn't anything fatal. I also wrote `Readme.md` that could be useful, see at https://github.com/dotnet/orleans/pull/5909/files#diff-bfa748e7941a1faa44354b3856662e5e. A bit of tangential OK? OK to list ideas for further additions?

sergeybykov:
Many thanks, @veikkoeeva! This is the most comprehensive sample we have by far!

PR #5914, opened 2019-08-28
Lease based queue balancer refactor.

jason-bragg:
No longer uses deployment configuration.
Serialize changes to avoid interleaving.
More configuration options.
Conformed configuration patterns to Orleans common patterns, including replacing the azure blob lease settings with options.

### DIFF
--- a/src/Azure/Orleans.Streaming.AzureStorage/Options/AzureBlobLeaseProviderOptions.cs
+++ b/src/Azure/Orleans.Streaming.AzureStorage/Options/AzureBlobLeaseProviderOptions.cs
@@ -1,0 +1,9 @@
+namespace Orleans.Configuration
+{
+    public class AzureBlobLeaseProviderOptions
+    {
+        [RedactConnectionString]
+        public string DataConnectionString { get; set; }
+        public string BlobContainerName { get; set; }
+    }
+}
--- a/src/Azure/Orleans.Streaming.AzureStorage/Options/AzureBlobLeaseProviderOptions.cs
+++ b/src/Azure/Orleans.Streaming.AzureStorage/Options/AzureBlobLeaseProviderOptions.cs
@@ -1,3 +1,9 @@
+using System;
+using Microsoft.Extensions.Options;
+using Microsoft.WindowsAzure.Storage;
+using Orleans.Runtime;
+using Orleans.Streaming.AzureStorage;
+
 namespace Orleans.Configuration
 {
     public class AzureBlobLeaseProviderOptions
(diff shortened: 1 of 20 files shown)


src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -9,84 +9,25 @@
...
     public class LeaseBasedQueueBalancer : QueueBalancerBase, IStreamQueueBalancer
     {
-        /// <summary>
-        /// Lease category for LeaseBasedQueueBalancer
-        /// </summary>
-        public const string LeaseCategory = "QueueBalancer";
-
         private class AcquiredQueue 
         {
+            public int LeaseId { get; set; }
```
ReubenBond: What does this id represent, is it an index? I prefer a string if there's no specific meaning to it (and it isn't used as a lease itself)

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -180,75 +120,58 @@ private async Task MaintainLeases()
...
-            else if (this.myQueues.Count > this.maximumResponsibility)
-            {
-                await this.ReleaseLeasesToMeetResponsibility();
+                bool allLeasesRenewed = await this.RenewLeases();
+                // if we lost some leases during renew after leaseAquisitionTimer stopped, restart it
+                if (!allLeasesRenewed &&
+                    this.leaseAquisitionTimer == null &&
+                    !base.Cancellation.IsCancellationRequested)
+                {
+                    this.leaseAquisitionTimer = this.timerRegistry.RegisterTimer(null, this.AcquireLeasesToMeetResponsibility, null, TimeSpan.Zero, this.options.LeaseAquisitionPeriod);
```
ReubenBond: not blocking: split over multiple lines

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -419,43 +379,52 @@ private async Task ScheduleUpdateResponsibilities(HashSet<SiloAddress> activeSil
             }
         }
 
+        /// <summary>
+        /// Greedy silos
+        /// </summary>
+        /// <param name="overflow"></param>
+        /// <param name="activeSilos"></param>
+        /// <returns></returns>
+        private bool AmGreedy(int overflow, HashSet<SiloAddress> activeSilos)
```
ReubenBond: Not understanding the summary comment.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -419,43 +379,52 @@ private async Task ScheduleUpdateResponsibilities(HashSet<SiloAddress> activeSil
...
+        {
+            // If using multiple stream providers, this will select the same silos to be greedy for
+            //   all providers, aggravating inbalance as stream provider count increases.
```
ReubenBond: In practice I don't think this will be an issue, but I'm glad you highlighted it

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
ReubenBond: This could probably be a local function instead, I think

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
```diff
@@ -257,40 +180,53 @@ private async Task ReleaseLeasesToMeetResponsibility()
...
 
             // tracks how many remaining possible leases there are.
-            int possibleLeaseCount = this.queueSelector.Count - this.myQueues.Count;
+            var possibleLeaseCount = this.queueSelector.Count - this.myQueues.Count;
             if (this.Logger.IsEnabled(LogLevel.Debug))
             {
                 this.Logger.LogDebug("Holding leased for {QueueCount} queues.  Trying to acquire {AquireQueueCount} queues to reach {TargetQueueCount} of a possible {PossibleLeaseCount}", this.myQueues.Count, leasesToAquire, expectedTotalLeaseCount, possibleLeaseCount);
             }
 
+            Stopwatch sw = Stopwatch.StartNew();
```
ReubenBond: nit: favor ValueStopwatch

src/Azure/Orleans.Streaming.AzureStorage/Options/AzureBlobLeaseProviderOptions.cs
```diff
@@ -7,4 +13,73 @@ public class AzureBlobLeaseProviderOptions
...
+        private AzureBlobLeaseProviderOptionsValidator(AzureBlobLeaseProviderOptions options, string name)
+        {
+            this.options = options;
+            this.name = name ?? string.Empty;
+        }
+
+        public void ValidateConfiguration()
+        {
+            // name can be null, but not empty or white space.
+            if(this.name != null && string.IsNullOrWhiteSpace(this.name))
```
ReubenBond: This double check is redundant

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: Understood.  A little history:  This is an issue in an existing service which has one or more stream providers per tenant.  Addressing it was a specific ask by the team when working out the requirements of the lease base queue balancer.  This note is primarily to help track this shortcoming, should future maintainers need address it.

src/Orleans.Runtime/Streams/QueueBalancer/LeaseBasedQueueBalancer.cs
jason-bragg: Yeah, poor name.  It's used to track the order in which the leases were obtained in, so that this information is available to algorithms that select which leases to release should the provider need to shed leases.
Will try to make it more clear.

jason-bragg: Found bug in lease ordering.. bleh.. fixing..

PR #5917, opened 2019-08-28
Stateless worker local compatibility check

JohnMorman:
This changes the placement of stateless workers to not just blindly choose the local silo as the target. Instead we check if the local silo is compatible first, if it is then we choose it, otherwise we back off to the random logic which we previously used when the silo was shutting down.


benjaminpetit:
Thanks!

Could you add a test case for stateless worker? In the code we have sometimes some specifics code path for this type.

Look into the test class `HeterogeneousTests`. Ping me if you need any help

JohnMorman:
Added two tests based off of the HeterogeneousTests, let me know if that is the correct place to add or if you have any other suggestions

ReubenBond:
Thanks, @JohnMorman!

Issue #3883, opened 2018-01-18
SerializationException under high load


ReubenBond:
This issue tracks a non-root-caused issue whereby a SerializationException is thrown in the Message deserialization path, usually during header deserialization, when the system is under high load.

I don't know how to reproduce the issue.

**Example 1**

`Unexpected token '185' introducing type specifier`

**Example 2**

```
   Exception trying to process 43634 bytes from endpoint xx.xx.xxx.xx:58094  
Exc level 0: System.Runtime.Serialization.SerializationException: Attempt to read past the end of the input stream: CurrentPosition=20, n=1851879244, totalLength=523283
   at Orleans.Serialization.BinaryTokenStreamReader.CheckLength(Int32 n, Int32& offset, Boolean& safeToUse)
   at Orleans.Serialization.BinaryTokenStreamReader.CheckLength(Int32 n, Int32& offset)
   at Orleans.Serialization.BinaryTokenStreamReader.ReadString()
   at Orleans.Runtime.Message.HeadersContainer.Deserializer(Type expected, IDeserializationContext context)
   at Orleans.Serialization.SerializationManager.DeserializeMessageHeaders(IDeserializationContext context)
   at Orleans.Runtime.IncomingMessageBuffer.TryDecodeMessage(Message& msg)
   at Orleans.Runtime.Messaging.IncomingMessageAcceptor.ReceiveCallbackContext.ProcessReceived(SocketAsyncEventArgs e)
```

**Example 3**

```
!!!!!!!!!! Exception deserializing message body Exc level 0: System.ArgumentException: Extended key field is not null in non-extended UniqueIdentifier.
at Orleans.Runtime.UniqueKey.ValidateKeyExt(String keyExt, UInt64 typeCodeData)
at Orleans.Runtime.UniqueKey.NewKey(UInt64 n0, UInt64 n1, UInt64 typeCodeData, String keyExt)
at Orleans.Serialization.BinaryTokenStreamReader.ReadUniqueKey()
at Orleans.Serialization.BinaryTokenStreamReader.TryReadSimpleType(Object& result, SerializationTokenType& token)
at Orleans.Serialization.SerializationManager.DeserializeInner(Type expected, IDeserializationContext context)
[...]

PR #5920, opened 2019-08-30
Do not call release header/body on a message in the dispatcher

benjaminpetit:
This can cause corruption in messaging. 

The component that actually send the data via the socket the one responsible for this body & header release.


ReubenBond:
Fixes #3883

PR #5923, opened 2019-08-31
Stream configuration namespace cleanup.

jason-bragg:
Inconsistent name spaces harm configuration discoverability.
- Move all configurators in Orleans.Streams namespace to Orleans.Hosting
- Move all configurators in Orleans.Configuration namespace to Orleans.Hosting
- Added developer extension functions for EventDataGeneratorStream
- Cleaned up naming issues with Memory Stream Configurator.

### DIFF
--- a/src/Azure/Orleans.Streaming.EventHubs/Hosting/DeveloperExtensions.cs
+++ b/src/Azure/Orleans.Streaming.EventHubs/Hosting/DeveloperExtensions.cs
@@ -1,0 +1,37 @@
+using System;
+
+namespace Orleans.Hosting.Developer
+{
+    public static class SiloBuilderExtensions
+    {
+        /// <summary>
+        /// Configure silo to use event data generator streams.
+        /// </summary>
+        public static ISiloHostBuilder AddEventDataGeneratorStreams(
+            this ISiloHostBuilder builder,
+            string name,
+            Action<IEventDataGeneratorStreamConfigurator> configure)
+        {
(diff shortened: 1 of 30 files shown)


src/Azure/Orleans.Streaming.EventHubs/Hosting/DeveloperExtensions.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+
+namespace Orleans.Hosting.Developer
```
ReubenBond: I feel these should still be in `Orleans.Hosting` - the `.Developer` ns will make them very difficult to discover

jason-bragg: Understood.  These are not surfaces intended for production use.  They exist for developers and test.  Being less discoverable was the intention.  So I suspect the question we need resolve here is whether they should be less discoverable?

ReubenBond: I'm not sure. It's a hard question. A part of me wonders: should we be checking some config flag? Eg, something like `IHostEnvironment.IsDevelopment`, or maybe something more Orleans-specific?

That's not an immediate solution... but in the short term maybe we could fix this with regular naming. My feeling is that we should make the functionality we offer discoverable but I don't know the use-case for generators in order to decide what value they have: if they are just for dev/test then I'm happy for no one to know about them (and even happier for us to delete them).

PR #5929, opened 2019-09-04
Fix wrong link in stream documentation

Tino92:
Seems like SampleStreamingGrain.cs was moved, so the current link in the documentation is returning 404.


ReubenBond:
Thank you, @Tino92!

PR #5935, opened 2019-09-06
Migrate to ASP.NET "Bedrock" abstractions

ReubenBond:
Deletes our copies of networking abstractions in favor of the shared Bedrock abstractions, in preparation for 3.0.

Ideally we could also remove SocketConnection, etc, but there are no publicly accessible versions of those yet so we still have a significant amount of duplicated code.

cc @davidfowl


ReubenBond:
image

This is a little concerning, maybe I have an assembly mismatch locally (running tests on .NET Framework)

davidfowl:
Any repro?

ReubenBond:
It reproduces on my home machine on both .NET Fx & .NET Core 2.0.
Running any test will repro it. eg, JSON_EchoJson.

This is quirky, VS is showing the exception fields under a field of type `ReadOnlySequence<byte>` :
image

davidfowl:
.NET Core 2.0? What about 2.1 (2.0 is out of support).

ReubenBond:
Repros on 2.1 & 2.2

davidfowl:
Pull together as minimal a repro as you can an file an issue

ReubenBond:
@davidfowl the bug was caused by an error in my code - in one place I was advancing the PipeReader before actually consuming the data. The fix is simple:

``` diff
diff --git a/src/Orleans.Core/Networking/ConnectionPreamble.cs b/src/Orleans.Core/Networking/ConnectionPreamble.cs
index f3f552595..7635a69ad 100644
--- a/src/Orleans.Core/Networking/ConnectionPreamble.cs
+++ b/src/Orleans.Core/Networking/ConnectionPreamble.cs
@@ -96,8 +96,9 @@ int ReadLength(ref ReadOnlySequence<byte> b)
             }
 
             var payloadBuffer = buffer.Slice(0, length);
-            input.AdvanceTo(payloadBuffer.End);
 
+            try
+            {
                 var reader = new BinaryTokenStreamReader2(payloadBuffer);
                 var grainId = reader.ReadGrainId();
 
@@ -123,6 +124,11 @@ int ReadLength(ref ReadOnlySequence<byte> b)
                 }
 
                 return (grainId, protocolVersion, siloAddress);
+            }
+            finally
+            {
+                input.AdvanceTo(payloadBuffer.End);
+            }
 
             void CheckForCompletion(ref ReadResult r)
             {

```

PR #5945, opened 2019-09-11
Upgrade to .NET Core 3.0 SDK

ReubenBond:
Several changes are required in order to support building on the .NET Core 3.0 SDK.
This PR makes those changes and upgrades our build to the .NET Core 3.0 preview9 SDK


ReubenBond:
Note that this is based on top of #5946 because that PR removes some unnecessary metapackages which would break the build without modification (see Orleans.Client.csproj).

ReubenBond:
Seems that our Tests.cmd doesn't pick up the .NET version like Build.cmd does

Issue #5879, opened 2019-08-21
Kinesis Streams Sequence Number much bigger than attributes in StreamSequenceToken introduced with #5580


Frank-Krick:
We have an implementation of a persistent stream adapter to consume Amazon Kinesis streams which we are using in production for several month now.

Amazon Kinesis streams have a sequence number consisting of 128 digits (186 Bits) which is much bigger than the attributes SequenceNumber and EventIndex defined in StreamSequenceToken. The changes in #5580 therefore break our stream provider implementation.

StreamSequenceToken seems to be very much specific to Azure streams and it is not apparent how an implementation of StreamSequenceToken should be mapped to other persistent stream providers and it seems to be not a very good choice to tie the StreamSequenceToken to closely with stream implementations from a given vendor.

jason-bragg:
Hmm.. Interesting problem.

I’ll start with some clarifications regarding the changes made in 3.0 which have brought us here.  For Orleans persistent streams to integrate with other systems, including legacy systems and data feeds not generated by streams themselves, there was a strong need to be able to specialize how data was converted to and from the underlying queuing system.  This necessitated a customizable translation between the data in the queue and the data in the cache.  Layers already existed for these translations, but were quite cumbersome to use.

To expose this layer as a pluggable component, we needed both the cache and the queue sides of the translation to be formalized.  The queue side is defined by the underlying persistent queue so we need conform to that, but the cache side could be standardized.

Previously the cache was quite generic, too generic to easily support a uniform surface.  This was by design, as we wanted to support the caching of a wide variety of data, as not to unnecessarily limit the utility of these systems.  To support the pluggable data adapter, we elected to simplify the cache by formalizing it around what we felt was a reasonable set of information which we believed most queuing systems could be conformed to.  This was an understood tradeoff, simplicity at the cost of flexibility, made on the suspicion that most queues could be conformed to the pattern, even if it wasn’t an exact match.  This pattern was vetted against all of the current persistent stream providers we support (not all from azure) as well as a couple other queues out there we were aware of (like kafka), and the pattern seemed to hold.  We unfortunately didn’t vet that it would work for Kinesis.  This was an oversight; Kinesis is a major queuing service we’d prefer work well with our features.

[...]

PR #5948, opened 2019-09-13
WIP - Use opaque sequence token in message cache

jason-bragg:
We now use opaque streamId, sequence token, and offsets in stream message cache.

This is intended to address https://github.com/dotnet/orleans/issues/5879
Will also help us move to more robust streamId.

PooledQueueCache is mostly ok, but cruft translating opaque cached data to pre-existing types is a bit ick.

Also, eviction strategy still has logic that shouldn't be there.

Getting this out for review, but want to let the patterns bake for a bit before committing to this path.

@Frank-Krick


ReubenBond:
Closing due to idleness. It will remain accessible for the motivated.

PR #5953, opened 2019-09-15
Fix useful queries database name

veikkoeeva:
Before this sample was committed to the repository, it was
renamed from BigSample to OneBoxDeployment. However, these queries
had the previous name. This commit fixes that.


sergeybykov:
Thank you, @veikkoeeva!

PR #5954, opened 2019-09-15
Improves queries by adding lock hinting to membership protocol

veikkoeeva:
In OneBoxDeployment sample if there are approximately 8-16 siloes starting
simultaneously, one or a few can end up in a primary key constraint
violation since the IF NOT EXISTS check followed by INSERT is not atomic.
This commit remedies that situation according to OneBoxDeployment
(it already has this fix).

Notable that it's not optimal to enforce the query engine with hints like
this and there is a better way, but it would require a larger change.


veikkoeeva:
The tests pass locally.

PR #5961, opened 2019-09-18
Make CollectionAgeLimitAttribute easy to use!

bluexo:
This PR make CollectionAgeLimitAttribute user friendly . 

```C#
// This grain has 10 hours age limit
[CollectionAgeLimit(Hours = 10)]
class SampleGrainClass1 : Grain ...
{
}

// This grain has 1 day +10 hours + 30 minutes age limit
[CollectionAgeLimit(Days = 1, Hours = 10 , Minutes = 30)]
class SampleGrainClass2 : Grain ...
{
}

//This grains has 32767 days age limit
[CollectionAgeLimit(AlwaysActive = true)]
class SampleGrainClass3 : Grain ...
{
}
```

### DIFF
--- a/src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
+++ b/src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
@@ -8,6 +8,8 @@ namespace Orleans.Configuration
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
     public class CollectionAgeLimitAttribute : Attribute
     {
+        public static readonly TimeSpan DEFAULT_COLLECTION_AGE_LIMIT = TimeSpan.FromHours(2);
+
         public readonly TimeSpan MinAgeLimit = TimeSpan.FromMinutes(1);
 
         public double Days { get; set; } 
--- a/src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
+++ b/src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
@@ -10,9 +10,9 @@ namespace Orleans.Configuration
     {
         public readonly TimeSpan MinAgeLimit = TimeSpan.FromMinutes(1);
 
-        public int Days { get; set; } 
-        public int Hours { get; set; } 
-        public int Minutes { get; set; } 
+        public double Days { get; set; } 
+        public double Hours { get; set; } 
+        public double Minutes { get; set; } 
 
         public bool AlwaysActive { get; set; }
 
--- a/src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
+++ b/src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 namespace Orleans.Configuration
 {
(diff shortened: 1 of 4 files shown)


src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
```diff
@@ -8,6 +8,25 @@ namespace Orleans.Configuration
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
     public class CollectionAgeLimitAttribute : Attribute
     {
-        public double Minutes { get; set; }
+        public const int DefaultAgeMinutes = 10;
```
sergeybykov: Why 10 minutes? The default today is 2 hours.

src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
```diff
@@ -8,6 +8,25 @@ namespace Orleans.Configuration
...
+        public bool AlwaysActive { get; set; }
+
+        public TimeSpan Amount
+        {
+            get
+            {
+                var span = AlwaysActive
+                ? TimeSpan.FromDays(short.MaxValue)
+                : TimeSpan.FromDays(Days) + TimeSpan.FromHours(Hours) + TimeSpan.FromMinutes(Minutes);
+                return span < TimeSpan.FromMinutes(DefaultAgeMinutes)
```
sergeybykov: Why this comparison? Would we want to return the default unless any of the values - `Days`, `Hours`, `Minutes` - is non-zero?

src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
bluexo: Can move the ```GrainCollectionOptions``` from ```Orleans.Runtime``` project to ```Orleans.Core``` project ?
 if so I can use the ```GrainCollectionOptions.DEFAULT_COLLECTION_AGE_LIMIT```

> Why 10 minutes? The default today is 2 hours.

src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
bluexo: I think the ```DefaultAgeMinutes``` should recommend value , should I change this line to ```span < TimeSpan.Zero``` ?

> Why this comparison? Would we want to return the default unless any of the values - Days, Hours, Minutes - is non-zero?

src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
sergeybykov: I don't think moving `GrainCollectionOptions` to `Orleans.Core` makes a lot of sense. But we can move `DEFAULT_COLLECTION_AGE_LIMIT` from `GrainCollectionOptions` to `CollectionAgeLimitAttribute`. I submitted https://github.com/bluexo/orleans/pull/1 with the change.

src/Orleans.Core/Configuration/CollectionAgeLimitAttribute.cs
sergeybykov: The logic I think is right is in https://github.com/bluexo/orleans/pull/1.

sergeybykov:
Thank you, @bluexo!

Issue #5158, opened 2018-11-02
Client connection using AdoNetClustering  fails on Docker container


fduman:
- Used Orleans 2.1.0. 
- Run on .NET Core 2.1, Docker linux container.

[orleans.log](https://github.com/dotnet/orleans/files/2542565/orleans.log)
(frontend is Orleans Client app. backend is silo. Db is Postgresql)

Using StaticClustering configuration runs perfectly.

I saw an issue similar with this at https://github.com/Azure/service-fabric-issues/issues/1182

sergeybykov:
Does the frontend have network access to the silo gateway endpoint `192.168.32.3:30000`? It complains that it couldn't connect to the endpoint:
>Connection to 192.168.32.3:30000 could not be established in 00:00:05

fduman:
The frontend has access to the backend. They are on the same docker network. It is interesting that backend silo says “Remote socket closed while receiving connection preamble data from endpoint 192.168.32.4:34992” for the frontend connection.

benjaminpetit:
Can you post your client and silos configurations?

fduman:
@benjaminpetit I prepared a sample for the issue. You can execute "docker-compose up" at sample's root folder. Then you should hit http://[Your docker machine IP]:8081/api/values

The web application can't connect to the Silo for the first time.

My Docker Version is 18.06.1-ce, build e68fc7a

I use the default Docker machine ISO on my Mac.

[OrleansDockerTest.zip](https://github.com/dotnet/orleans/files/2603186/OrleansDockerTest.zip)

fduman:
I said there was no problem with static clustering configuration, but I was wrong.

fduman:
@benjaminpetit Is there anything I can do for the issue?

sergeybykov:
@fduman @benjaminpetit went on a leave until early February. I'm not sure how much we can help you with running on Mac. Asking in https://gitter.im/dotnet/orleans might be a more expedient way to get help with that.

Gandii:
@fduman had the same issue with almost identical setup playing around with the orleans docker sample, realized that the silo was setting its address to 127.0.0.1 when checking the Membeor ship table instead of its Docker network IP.

`
         .Configure<EndpointOptions>(options => options.AdvertisedIPAddress = ip)
`
then use any method to get the the silos/container IP or even set a fixed IP for the container.

`
         IPAddress.TryParse("172.20.0.4",out var ip);

            //or

[...]

PR #5963, opened 2019-09-19
Fix Connect blocked when ConnectAsync completed synchronously

tcz717:
Fix for #5158

Acccodring to [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.connectasync?view=netcore-2.2), `ConnectAsync(SocketAsyncEventArgs)` will return bool 

> true if the I/O operation is pending. The Completed event on the e parameter will be raised upon completion of the operation.
> 
> false if the I/O operation completed synchronously. In this case, The Completed event on the e parameter will not be raised and the e object passed as a parameter may be examined immediately after the method call returns to retrieve the result of the operation.

In windows, looks like `ConnectAsync` will alwasy return true (pending) even if operation is done immediately https://github.com/dotnet/corefx/blob/d391103/src/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEventArgs.Windows.cs#L163-L164 . 

But in Linux host, the [corefx](https://github.com/dotnet/corefx/blob/d391103/src/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncContext.Unix.cs#L14-L33) says:
```csharp
    // The asynchronous socket operations here generally do the following:
    // (1) If the operation queue is Ready (queue is empty), try to perform the operation immediately, non-blocking.
    // If this completes (i.e. does not return EWOULDBLOCK), then we return the results immediately
    // for both success (SocketError.Success) or failure.
    // No callback will happen; callers are expected to handle these synchronous completions themselves.
```

That means if operation completed synchronously, the Completed callback will be never invoked, which will make `AutoResetEvent.WaitOne` keeping waiting until timeout. That's why the server did not receive `Preamble` and timeout [here](https://github.com/dotnet/orleans/blob/v2.4.2/src/Orleans.Runtime/Messaging/IncomingMessageAcceptor.cs#L207)

I see you plan to rewrite network stack in 3.0.0. I don't if this bug will be solved. But as for 2.x version, my possible solution is to change Connect method https://github.com/dotnet/orleans/blob/2.4.3/src/Orleans.Core/Messaging/SocketManager.cs#L196-L209 like 
```csharp
        internal static void Connect(Socket s, IPEndPoint endPoint, TimeSpan connectionTimeout)
        {
            var signal = new AutoResetEvent(false);
            var e = new SocketAsyncEventArgs();
            e.RemoteEndPoint = endPoint;
            e.Completed += (sender, eventArgs) => signal.Set();
            bool pending = s.ConnectAsync(e);

            if (pending && !signal.WaitOne(connectionTimeout))
                throw new TimeoutException($"Connection to {endPoint} could not be established in {connectionTimeout}");

            if (e.SocketError != SocketError.Success || !s.Connected)
                throw new OrleansException($"Could not connect to {endPoint}: {e.SocketError}");
        }
```

### DIFF
--- a/src/Orleans.Core/Messaging/SocketManager.cs
+++ b/src/Orleans.Core/Messaging/SocketManager.cs
@@ -199,9 +199,9 @@ namespace Orleans.Runtime
             var e = new SocketAsyncEventArgs();
             e.RemoteEndPoint = endPoint;
             e.Completed += (sender, eventArgs) => signal.Set();
-            s.ConnectAsync(e);
+            bool pending = s.ConnectAsync(e);
 
-            if (!signal.WaitOne(connectionTimeout))
+            if (pending && !signal.WaitOne(connectionTimeout))
                 throw new TimeoutException($"Connection to {endPoint} could not be established in {connectionTimeout}");
 
             if (e.SocketError != SocketError.Success || !s.Connected)


ReubenBond:
Thank you, @tcz717! Very interesting. I'm not sure how ConnectAsync could complete with success synchronously, but this looks like the right change regardless. In the new revision of the code we perform the correct check: https://github.com/dotnet/orleans/blob/cc4749c9be837f3f6dd82ece9cb4ed7a9c5dc798/src/Orleans.Core/Networking/Shared/SocketConnectionFactory.cs#L43

tcz717:
Nice! I just tested my client with this change. And now my client can connect to the server normally. Before we upgrade to 3.0, would you mind merging my change into 2.4.3, so that we can use Orleans in Linux as soon as possible?

PR #5974, opened 2019-09-25
Clear RequestContext when spawning connections

ReubenBond:
This was causing reliability issues because new connections could occasionally be started due to a Ping message, which has a `"Ping"` `RequestContext` header set. The connection would then capture that `RequestContext` and would import it into messages created by the `SiloConnection` itself. These messages would then be treated as though they were Ping messages by the remote silo.

Examples of such messages are:
* Rejections for messages destined to an old silo generation
* Directory lookups - this is trickier, it relies on an optimization whereby `IncomingMessageAgent.ReceiveMessage` is called synchronously from the connection 

This would result in timeouts (which can prevent a silo from starting), active silos being declared dead, obscure exceptions such as `Unrecoverable Rejection (info: System.InvalidOperationException: Expected result of type Orleans.GrainDirectory.AddressesAndTag but encountered a null value.)`, and `NullReferenceExceptions` from `ActivationDirectory.FindTarget`

### DIFF
--- a/src/Orleans.Core/Networking/Connection.cs
+++ b/src/Orleans.Core/Networking/Connection.cs
@@ -71,6 +71,8 @@ namespace Orleans.Runtime.Messaging
         /// <returns>A <see cref="Task"/> which completes when the connection terminates and has completed processing.</returns>
         public async Task Run()
         {
+            RequestContext.Clear();
+
             Exception error = default;
             try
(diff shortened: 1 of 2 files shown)


ReubenBond:
Merging myself after approval - we've already tested this in VSO

PR #5983, opened 2019-09-28
Upgrade packages & target netcoreapp3.0

ReubenBond:
~net461->net472~ This update caused tests to hang so I reverted this change & will investigate
netcoreapp*->netcoreapp3.0
Upgrade 'core' packages to their latest versions
Various updates to make sure everything builds cleanly


ReubenBond:
Need .NET 4.7.2 or 4.8.0 on the PR Build machines

Issue #5978, opened 2019-09-26
Exception deserialization failure


u-less:
When the client does not refer to the server's Exception type, Client will fail and timeout error will be raised.I think there should be a general approach.

HermesNew:
This may be a bug, because the Type can't be found, causing a serialization error, causing the exception to handle the infinite loop, and finally thow out of timeoutexception.
I also encountered the same problem. PgSQL throws an exception because Orleans can't find a reference to Npgsql.dll, so Orleans finally throws a timeoutexception.

ReubenBond:
@luohuazhiyu the explanation (other than the infinite loop) which @HermesNew provided is also my strongest suspicion. I agree that there should be a general approach. Do you have any suggestions for how this should be done generally?

luohuazhiyu:
@ReubenBond Grain is similar to the microservice API,Detailed exception information should be handled by the server,The client should have a generic Exception type to throw simple exception information Or handle only unknown types of exceptions.

veikkoeeva:
Can we ensure there is some kind of an identifier the very least that flows from silo to client in order to match exceptions? On one hand one might want to enforce a [trust boundary](https://en.wikipedia.org/wiki/Trust_boundary) on exception propagation here, on another hand it may not always be the case. Also intriguing is [RFC 7807](https://tools.ietf.org/html/rfc7807) or `Either` type of solution that would be easy to assemble, like installing a global exception handler as in ASP.NET Core.

HermesNew:
@ReubenBond IIncomingGrainCallFilter can get specific exceptions, and the client can only wait for a timeout. An infinite loop is my guess because the server won't have any return until it times out.It is easy to reappear, I think it is a serious problem.

ReubenBond:
Good feedback, thank you! I am considering various improvements to the experience here.
One such improvement is #5998

ReubenBond:
The main issue here was fixed in #5998, so at least *some* exception will come back to the original caller very quickly. That exception may be a serialization/deserialization exception but at least it's something which you can identify and use to fix issues quickly.

ReubenBond:
I'm going to shift this to 3.1.0 for now so that we have time to properly design the feature. As mentioned above, the experience will be much better in 3.0 regardless

ReubenBond:
[...]

PR #5998, opened 2019-10-01
Propagate message [de]serialization exceptions to callers

### LINKED ISSUE #4748: Exception thrown during OnActivateAsync not posted to client?
I try this: a grain implemented Grain(T), while the entity T can not be found in storage during OnActivateAsync, a EntityNotFoundException will be thrown. And the exception will not be posted to the client. How can I handle this error?

### LINKED ISSUE #5397: Test Reliability: ExceptionPropagation_ClientToGrain_SerializationFailure
UnitTests.General.ExceptionPropagationTests.ExceptionPropagation_ClientToGrain_SerializationFailure is flaky. Example test run: https://ci.dot.net/job/dotnet_orleans/job/master/job/functional_prtest/2578/testReport/junit/UnitTests.General/ExceptionPropagationTests/ExceptionPropagation_ClientToGrain_SerializationFailure/

```
Stacktrace

MESSAGE:
Assert.Throws() Failure\r\nExpected: typeof(System.NotSupportedException)\r\nActual: 
typeof(System.TimeoutException): Response did not arrive on time in 00:00:30 for message: Request 
*cli/1770639c@44a482a6->S127.0.0.1:42555:0*grn/415DDB74/41aaf750 #5478: . Target History is: 
<S127.0.0.1:42555:0:*grn/415DDB74/41aaf750:>.

+++++++++++++++++++
STACK TRACE:
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() at 
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

ReubenBond:
Currently when deserialization fails for a response message we do not propagate that deserialization exception to the caller. This PR adds behavior to propate such exceptions back to the original caller.

Fixes #4748

Fixes #5397

xref #5978

### DIFF
--- a/src/Orleans.Core/Networking/Connection.cs
+++ b/src/Orleans.Core/Networking/Connection.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Orleans.Configuration;
 using Orleans.Messaging;
 
 namespace Orleans.Runtime.Messaging
@@ -33,11 +34,13 @@ namespace Orleans.Runtime.Messaging
         protected Connection(
             ConnectionContext connection,
             ConnectionDelegate middleware,
+            MessageFactory messageFactory,
             IServiceProvider serviceProvider,
             INetworkingTrace trace)
         {
             this.Context = connection ?? throw new ArgumentNullException(nameof(connection));
             this.middleware = middleware ?? throw new ArgumentNullException(nameof(middleware));
+            this.MessageFactory = messageFactory ?? throw new ArgumentNullException(nameof(messageFactory));
             this.serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
             this.Log = trace ?? throw new ArgumentNullException(nameof(trace));
             this.outgoingMessages = Channel.CreateUnbounded<Message>(OutgoingMessageChannelOptions);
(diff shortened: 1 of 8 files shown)


src/Orleans.Core/Networking/Connection.cs
```diff
@@ -408,5 +400,50 @@ private static EndPoint NormalizeEndpoint(EndPoint endpoint)
...
+                response.Result = Message.ResponseTypes.Error;
+                response.BodyObject = Response.ExceptionResponse(exception);
+
+                // Send the error response and continue processing the next message.
+                this.MessageCenter.SendMessage(response);
+            }
+            else if (message.Direction == Message.Directions.Response)
+            {
+                // If the message was a response, propagate the exception to the intended recipient.
+                message.Result = Message.ResponseTypes.Error;
```
sergeybykov: Wouldn't it be better to create a new message here instead of mutating the failed one?

ReubenBond: We can. In the original version of this I had a `Message.ShallowCopy()` method and used that, however this approach of mutating the message is the common pattern from the existing code so I reverted to doing the same.

I can go with the copy approach instead if you prefer it

ReubenBond:
Marked as WIP because I am adding additional test coverage

src/Orleans.Core/Networking/Connection.cs
sergeybykov: I'm fine with leaving it as is.

ReubenBond:
Test failures caused by client object being garbage collected during the test 😅

PR #5999, opened 2019-10-01
Fixed wrong condition for getting logContext

DamirAinullin:
The condition type.IsSerializable already exists. Need to replace it with TypeHasKnownBase.


ReubenBond:
Thank you, @DamirAinullin!

PR #6000, opened 2019-10-01
Remove InternalsVisibleTo set for extensions by making necessary internal types public

### LINKED ISSUE #4477: Orleans Core public surface insufficient for extension development
External and application engineers cannot build extension libraries on Orleans without access to internals.  This is evidenced by our internally maintained extension libraries, many of which require access to Orleans Core internals.

[assembly: InternalsVisibleTo("Orleans.Clustering.AzureStorage")]
[assembly: InternalsVisibleTo("Orleans.Clustering.DynamoDB")]
[assembly: InternalsVisibleTo("Orleans.Hosting.AzureCloudServices")]
[assembly: InternalsVisibleTo("Orleans.Persistence.AzureStorage")]
[assembly: InternalsVisibleTo("Orleans.Persistence.DynamoDB")]
[assembly: InternalsVisibleTo("Orleans.Reminders.AzureStorage")]
[assembly: InternalsVisibleTo("Orleans.Reminders.DynamoDB")]
[assembly: InternalsVisibleTo("Orleans.Streaming.AzureStorage")]
[assembly: InternalsVisibleTo("Orleans.Streaming.EventHubs")]
[assembly: InternalsVisibleTo("Orleans.Streaming.SQS")]
[assembly: InternalsVisibleTo("Orleans.TelemetryConsumers.Counters")]
[assembly: InternalsVisibleTo("Orleans.Transactions.DynamoDB")]

Internal access needs be removed for extension libraries.  Where extension libraries currently require internal access, either alternate solutions need be found or the accessed capabilities need be exposed in maintainable public apis.

sergeybykov:
Fixes #4477.


src/Orleans.Core/Async/AsyncExecutorWithRetries.cs
```diff
@@ -9,7 +9,7 @@ namespace Orleans
     /// This class is a convenient utility class to execute a certain asynchronous function with retries,
     /// allowing to specify custom retry filters and policies.
     /// </summary>
-    internal static class AsyncExecutorWithRetries
+    public static class AsyncExecutorWithRetries
```
ReubenBond: I'm not huge on exposing these since they are general utility classes. I'd be fine to copy them to the other places that they're needed or even include them via shared source.

Ultimately this is also ok if we don't like the alternatives.

src/Orleans.Core/Async/TaskExtensions.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Orleans
 {
-    internal static class OrleansTaskExtentions
+    public static class OrleansTaskExtentions
```
ReubenBond: Likewise here. I'd prefer we namespace these to something internal at least, since otherwise we can potentially introduce some clashes

src/Orleans.Core/Logging/ILoggerExtensions.cs
```diff
@@ -252,7 +252,7 @@ internal static void Assert(this ILogger logger, ErrorCode errorCode, bool condi
             logger.Fail(errorCode, "Assert failed with message = " + message);
         }
 
-        internal static void Fail(this ILogger logger, ErrorCode errorCode, string message)
+        public static void Fail(this ILogger logger, ErrorCode errorCode, string message)
```
ReubenBond: In time we should attempt to remove our dependence on these

sergeybykov: I'm fine with leaving it as is.

ReubenBond: The main impetus is for semantic logging, since that lets our users perform richer log searches

src/Orleans.Core/Logging/ErrorCodes.cs
```diff
@@ -4,7 +4,7 @@ namespace Orleans
     /// <summary>
     /// The set of error types used by the Orleans runtime libraries for logging errors.
     /// </summary>
-    internal enum ErrorCode
```
jason-bragg: Don't think our error codes should be public.
Makes it more likely app developers will use them, which can create logging confusion.
Suggest well defined ranges of error codes for core, extensions, and services, and move those used by extension libraries into those libraries.
Since existing codes may not conform to the defined ranges, we shouldn't enforce the ranges, but rather make it an 'orleans best practices' problem, wherein communicating the practice is probably sufficient.
See OrleansServiceBusErrorCode for example.

src/Orleans.Core/Logging/ILoggerExtensions.cs
```diff
@@ -95,7 +95,7 @@ public static void Debug(this ILogger logger, int logCode, string format, params
             logger.LogDebug(logCode, format, args);
         }
 
-        internal static void Debug(this ILogger logger, ErrorCode logCode, string format, params object[] args)
```
jason-bragg: If we organize error codes as described in other comment, we can keep these internal.

src/Orleans.Core/Statistics/FloatValueStatistic.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Orleans.Runtime
 {
-    internal class FloatValueStatistic : ICounter<float>
```
jason-bragg: Should these be in Orleans.Runtime.Internal?  I don't think we want public using these surfaces as, imo, they are subject to change/removal.  If we at least hide them behind an "internal" namespace, then that should mostly hide them and indicate that we're not supporting them as part of our public surface.

jason-bragg: Above comment applies to all ICounters

PR #6005, opened 2019-10-03
Use MemFree when MemAvailable is not present

howard0su:
On Windows WSL1, there is no MemAvailable line in meminfo file.
Use MemFree which gives similiar information.


sergeybykov:
Thank you, @howard0su!

PR #6014, opened 2019-10-05
Simplify ConnectionListener.RunAsync

ReubenBond:
Fixing in response to feedback from @davidfowl

### DIFF
--- a/src/Orleans.Runtime/Networking/ConnectionListener.cs
+++ b/src/Orleans.Runtime/Networking/ConnectionListener.cs
@@ -8,7 +8,6 @@ using Microsoft.AspNetCore.Connections;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Orleans.Configuration;
-using Orleans.Internal;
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Networking/ConnectionListener.cs
```diff
@@ -81,21 +81,11 @@ public void Start()
         private async Task RunAsync()
         {
             var runCancellation = this.cancellation.Token;
-            var runCancellationTask = runCancellation.WhenCancelled();
             while (!runCancellation.IsCancellationRequested)
```
davidfowl: while (true), let the AcceptAsync call throw.

sergeybykov:
Functional test consistently get stuck with this change.

davidfowl:
My guess would be that unbind isn’t being used to yield the accept loop. The token isn’t respected by the socket transport

ReubenBond:
I've fixed it in the latest commit and aligned the style with Kestrel's

PR #6035, opened 2019-10-10
Add TLS middleware with sample

### LINKED ISSUE #828: Adding optional TLS/encryption to transport channel

ReubenBond:
This PR adds support for securing client-to-silo and silo-to-silo connections using mutual-TLS.
Adds several `UseTls` methods to `ISiloBuilder`, `IClientBuilder`, and `ISiloHostBuilder`.

TLS can be configured on the silo like so:

``` C#
var host = new HostBuilder()
    .UseOrleans((context, siloBuilder) =>
    {
        var isDevelopment = context.HostingEnvironment.IsDevelopment();
        siloBuilder
            .UseLocalhostClustering(serviceId: "HelloWorldApp", clusterId: "dev")
            .UseTls(StoreName.My, "fakedomain.faketld", allowInvalid: isDevelopment, StoreLocation.LocalMachine, options =>
            {
                if (isDevelopment)
                {
                    options.AllowAnyRemoteCertificate();
                }
            });
    })
    .ConfigureLogging(logging => logging.AddConsole())
    .Build();
```

By default, when the `UseTls` extension method is used, both sides of all connections are required to present a certificate.

This adds a new package: `Microsoft.Orleans.Connections.Security` which targets `netcoreapp3.0` ***only*** and therefore .NET Core 3.0 is required to use this functionality.

A sample application is included under Samples\3.0\TransportLayerSecurity.

There are some rough edges here which I would especially like feedback on.  For example, a `TargetHost` must be set whenever a connection is made. I would like to know how users would expect to set this value. In the provided sample the value is set to a fixed value:

``` C#
.UseTls(StoreName.My, "fakedomain.faketld", allowInvalid: true, StoreLocation.LocalMachine, options =>
{
    options.OnAuthenticateAsClient = (connection, sslOptions) =>
    {
        sslOptions.TargetHost = "fakedomain.faketld";
    };
    // NOTE: Do not do this in a production environment
    options.AllowAnyRemoteCertificate();
})
```

I understand that this will not suffice for all cases. The `connection` parameter has a `Features` property which allows the users to determine which endpoint is being connected to (among other things). This could be used to determine the correct `TargetHost` value.

Fixes #828

xref https://github.com/aspnet/AspNetCore/issues/12809

### DIFF
--- a/Orleans.sln
+++ b/Orleans.sln
@@ -202,6 +202,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Orleans.Analyzers", "src\Or
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Analyzers.Tests", "test\Analyzers.Tests\Analyzers.Tests.csproj", "{B94232D8-B8C2-45D8-AE09-7D8A0B0D88ED}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Orleans.Connections.Security", "src\Orleans.Connections.Security\Orleans.Connections.Security.csproj", "{8E01A6EB-DE96-4DFD-AA7D-B07078F12372}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Orleans.Connections.Security.Tests", "test\Orleans.Connections.Security.Tests\Orleans.Connections.Security.Tests.csproj", "{D53D80CC-3E14-4499-B03F-610A5D3F6359}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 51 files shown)


src/Orleans.Connections.Security/Security/TlsConnectionAdapterOptions.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using Microsoft.AspNetCore.Connections;
+
+namespace Orleans.Connections.Security
+{
+    /// <summary>
+    /// Settings for how TLS connections are handled.
+    /// </summary>
+    public class TlsConnectionAdapterOptions
```
davidfowl: Don't replicate the Adapter naming mistake we made 😄

ReubenBond: s/Https/Tls 😅 
Perhaps `TlsOptions` is a nicer name for this, or `TlsConnectionOptions`

JanEggers:
@ReubenBond can you please let me know when there is a preview that includes this. I would like to use the TLS middleware for this: https://github.com/chkr1011/MQTTnet/tree/master/Source/MQTTnet.AspnetCore

ReubenBond:
@JanEggers you can take the current commit if you like - that MQTT project doesn't reference Orleans & the important bits are internal so I don't think a build of this would be useful

JanEggers:
@ReubenBond thx for letting me know, I didnt look at the code in detail yet. I was hoping I could just use the extension method on IConnectionBuilderand be done.

ReubenBond:
@JanEggers eventually yes, but those things are marked as internal in this PR because the user isn't supposed to interact with them. The intention it to upstream this to Bedrock eventually

RehanSaeed:
Is it possible to configure TLS using `IConfiguration` similar to the way [HTTPS can be configured for Kestrel](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-3.0#configureiconfiguration)?

```json
{
  "Kestrel": {
    "Endpoints": {
      "Http": {
        "Url": "http://localhost:5000"
      },

      "HttpsInlineCertFile": {
        "Url": "https://localhost:5001",
        "Certificate": {
          "Path": "<path to .pfx file>",
          "Password": "<certificate password>"
        }
      },

      "HttpsInlineCertStore": {
        "Url": "https://localhost:5002",
        "Certificate": {
          "Subject": "<subject; required>",
          "Store": "<certificate store; required>",
          "Location": "<location; defaults to CurrentUser>",
          "AllowInvalid": "<true or false; defaults to false>"
        }
      },

      "HttpsDefaultCert": {
        "Url": "https://localhost:5003"
      },

      "Https": {
        "Url": "https://*:5004",
        "Certificate": {
          "Path": "<path to .pfx file>",
          "Password": "<certificate password>"
        }
      }
    },
    "Certificates": {
      "Default": {
        "Path": "<path to .pfx file>",
        "Password": "<certificate password>"
      }
    }
  }
}
```

ReubenBond:
The current implementation does not explicitly add support for that, but it doesn't stop you from using IConfiguration yourself. This could be a good area for contribution

PR #6039, opened 2019-10-11
Add section on use of Task.Factory.StartNew()

### LINKED ISSUE #5977: Add information about Task.Factory.StartNew gotchas in the docs
The current documentation about running tasks inside a grain (https://dotnet.github.io/orleans/Documentation/grains/external_tasks_and_grains.html) suggests to use `Task.Factory.StartNew()`, but there are some issues regarding this, and it would be nice if those were pointed out in the doc.

Specifically, it should be stated that `Task.Factory.StartNew()` does not natively support async delegates, and so it should be used as `Task.Factory.StartNew(MyMethodAsync).Unwrap()` in this case.

rmja:
... with async delegates. Fixes #5977

### DIFF
--- a/src/Documentation/grains/external_tasks_and_grains.md
+++ b/src/Documentation/grains/external_tasks_and_grains.md
@@ -11,13 +11,21 @@ In some cases grain code might need to “break out” of the Orleans task sched
 
 ### Task based APIs:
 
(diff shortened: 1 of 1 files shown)


ReubenBond:
Many thanks, Rasmus!

Issue #5873, opened 2019-08-16
Add a warning about runtime running with trace level higher than Info


sergeybykov:
This was suggested by @EranOfer in https://github.com/dotnet/orleans/issues/5851#issuecomment-520709746

zHaytam:
Hello, I would like to take this as my first issue if possible.  
Where should I begin?

ReubenBond:
Fixed by #6041

PR #6041, opened 2019-10-11
Add warning message at startup

RozzaysRed:
Adding warning message at startup if debug level logging is set and recommends setting log level to info for production.

Addresses #5873

### DIFF
--- a/src/Orleans.Runtime/Silo/Silo.cs
+++ b/src/Orleans.Runtime/Silo/Silo.cs
@@ -162,7 +162,7 @@ namespace Orleans.Runtime
             {
                 var highestLogLevel = logger.IsEnabled(LogLevel.Trace) ? nameof(LogLevel.Trace) : nameof(LogLevel.Debug);
                 logger.LogWarning(
-                    ErrorCode.SiloGcWarning,
+                    new EventId((int)ErrorCode.SiloGcWarning),
                     $"A verbose logging level ({highestLogLevel}) is configured. This will impact performance. The recommended log level is {nameof(LogLevel.Information)}.");
             }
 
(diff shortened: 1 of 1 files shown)


src/Orleans.Runtime/Silo/Silo.cs
```diff
@@ -158,6 +158,8 @@ public Silo(ILocalSiloDetails siloDetails, IServiceProvider services)
                 logger.Warn(ErrorCode.SiloGcWarning, "Note: ServerGC only kicks in on multi-core systems (settings enabling ServerGC have no effect on single-core machines).");
             }
 
+            if (logger.IsEnabled(LogLevel.Debug)) logger.Warn(ErrorCode.SiloGcWarning, "Debug level logging set : This will impact performance of ServerGC.  Recommended to set log level to Info for production");
```
ReubenBond: We should also check for `Trace` (I've seen this happen)

veikkoeeva: Also I wonder if it would be better to use `nameof` here for 'Info' in case of refactoring; also the same for 'ServerGC' so that it points to the actual type to set it or to a MS docs link.

I mean this is "clear to us", but I see currently people moving from Java and other ecosystems and it's not that clear to where to look at.

RozzaysRed: I'll add that.  Also, as I was going over it again, I was thinking I should add an else to `if (!GCSettings.IsServerGC)` which is right above where the log is sitting since the issue is with the server gc.  If the ServerGC isn't on, then there's no reason to log the message right?

ReubenBond: Performance will be severely impacted under high load with or without ServerGC

ReubenBond: @RozzaysRed @veikkoeeva [I pushed a commit](https://github.com/dotnet/orleans/pull/6041/commits/e6fb652d9c513668623dad635f5a9dc41d530745), how does that look to you?

veikkoeeva: @ReubenBond Strikes a well balanced tone for the log message and how to to do/refactor logging related Orleans code. 👍

ReubenBond:
Thank you, @RozzaysRed!

PR #6042, opened 2019-10-12
 Prevent Orleans + Kestrel from interfering with each other's networking services

### LINKED ISSUE #6040: Orleans + ASP.NET co-hosting can cause NullReferenceException in ASP.NET
This is specific to Orleans 3.0-rc1.

Currently both Orleans & Kestrel register an implementation for `IConnectionListenerFactory`. Both have different requirements of the created ConnectionContexts, however and so when they are both configured on the same `IHostBuilder`/container we can observe `NullReferenceException`s from ASP.NET. Registering them in the reverse order would likely cause similar issues in Orleans.

Example:

``` C#
Host.CreateDefaultBuilder(args)
    .UseOrleans((context, builder) =>
    {
        builder
            .Configure<SiloMessagingOptions>(opts =>
            {
                opts.PropagateActivityId = true;
            })
            .UseLocalhostClustering()
            .AddMemoryGrainStorageAsDefault();
    })
    .ConfigureWebHostDefaults(webBuilder =>
    {
        webBuilder.UseStartup<Startup>();
    });
```

Kestrel registers an implementation here: https://github.com/aspnet/AspNetCore/blob/00e0b97466a3aa7e535b1e34fb04c11f5970fb28/src/Servers/Kestrel/Kestrel/src/WebHostBuilderKestrelExtensions.cs#L32

Orleans registers an implementation here: https://github.com/dotnet/orleans/blob/e816f02f12df34e9e554ae2b898528c42d9d87a4/src/Orleans.Runtime/Hosting/DefaultSiloServices.cs#L355

We can change our registration to disambiguate it from Kestrel's, but issues may occur with other users of these abstractions. I imagine the problem would also exist with `IConnectionFactory` (eg, maybe Orleans + SignalR).

cc @davidfowl

ReubenBond:
This is a quick fix for #6040. Ideally we will implement something more robust and configurable but this will suffice for 3.0.0

Fixes #6040

Note that this is based on top of #6035, so that comes first after which I will rebase this.

### DIFF
--- a/src/Orleans.Core/Core/DefaultClientServices.cs
+++ b/src/Orleans.Core/Core/DefaultClientServices.cs
@@ -99,7 +99,11 @@ namespace Orleans
             // Networking
             services.TryAddSingleton<ConnectionManager>();
             services.AddSingleton<ILifecycleParticipant<IClusterClientLifecycle>, ConnectionManagerLifecycleAdapter<IClusterClientLifecycle>>();
-            services.TryAddSingleton<IConnectionFactory, SocketConnectionFactory>();
+
+            services.AddSingletonKeyedService<object, IConnectionFactory>(
+                ClientOutboundConnectionFactory.ServicesKey,
+                (sp, key) => ActivatorUtilities.CreateInstance<SocketConnectionFactory>(sp));
+
             services.TryAddTransient<IMessageSerializer>(sp => ActivatorUtilities.CreateInstance<MessageSerializer>(sp,
                 sp.GetRequiredService<IOptions<ClientMessagingOptions>>().Value.MaxMessageHeaderSize,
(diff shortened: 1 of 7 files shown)


ReubenBond:
cc @davidfowl, this is the workaround I'm going with for now: https://github.com/dotnet/orleans/pull/6042/commits/bc8cbe9c5aba7347b87abaadaf64c0dd7ec1ace8

galvesribeiro:
Looking forward to see it merged! That bites me hard 😄

galvesribeiro:
rc2! :shipit: 

😄

SebastianStehle:
rtm!

sergeybykov:
rc2 is out - https://github.com/dotnet/orleans/releases/tag/v3.0.0-rc2. No plans for further changes before RTM.

PR #6044, opened 2019-10-14
[WIP] Move sql scripts into a separate NuGet package

### LINKED ISSUE #4867: Avoiding the OrleansAdoNetContent folder
The solution to #4243 causes repositories and projects to be polluted by the OrleansAdoNetContent folder. One has to employ various hacks to ensure that the folder is removed after package restore, and in the case of `dotnet restore` there doesn't even seem to exist any target to use with AfterTargets to get rid of the files.

I see how the files could be useful for someone just starting out with Orleans, but not for an established project. In my case, the files are possibly harmful, as I'm running on a modified schema and these files would only confuse my colleagues.

A possible solution would be to move the copying of these files into the OrleansSQLUtils project. One can then reference the individual packages to bypass the copying of the SQL files.

Has anyone else spent way too much time fighting these files? :) Any other ideas as to what can be done about it?

sergeybykov:
Fixes #4867.


sergeybykov:
Closing this in favor of #6049.

Issue #5865, opened 2019-08-14
Log parameter names improvement for structured loggers


srollinet:
A lot of log messages use the "old" String.Format syntax, with `{0}`  style parameters 

    Log.Info("Starting AsyncAgent {0} on managed thread {1}", Name, Thread.CurrentThread.ManagedThreadId)

When you use a structured logger, it results in 2 parameters named `0` and `1`, which is not a really useful information.

It can also cause indexation errors if you index the log messages. In my case I use Graylog to manage my logs and I see a lot of indexing errors because the same parameter with name `0` is not always of the same type. (int, string, ...)

This could be replaced by something like

    Log.Info("Starting AsyncAgent {AgentName} on managed thread {ThreadId}", Name, Thread.CurrentThread.ManagedThreadId)

The goal is maybe not to change it everywhere, but at least to improve the most common log messages in a first time and to improve the other logs over time.

ReubenBond:
We've been using structured logging for new log messages but haven't gone back to convert most older messages yet. I was hoping for an analyzer to be released which would help to ensure that no errors were introduced during the conversion. Please feel free to PR individual fixes

bddckr:
I've been looking for the same kind of analyzer. There's `Microsoft.Extensions.Logging.Analyzers` but it's [not shipping](https://github.com/aspnet/Extensions/blob/39d5064a04546c9acaa2871c4e07ece93d332dd7/src/Logging/Logging.Analyzers/src/Microsoft.Extensions.Logging.Analyzers.csproj#L11-L12) - https://github.com/aspnet/Extensions/issues/658.

peter-perot:
Yeah, logging looks really weird:

![image](https://user-images.githubusercontent.com/15671569/68952130-9fd66f80-07bf-11ea-8fbc-0fba15b89c70.png)

Especially in those case where you use _log parameters_ (not `string.Format` or the `$""` syntax) to output bars:

https://github.com/dotnet/orleans/blob/c683fa8f2fddf310c2dfc3c9cf43db4fb80be1ad/src/Orleans.Core/Runtime/OutsideRuntimeClient.cs#L199

This results in some really weird output like this when using a structured logger like Serilog:

```text
 "----------" Started OutsideRuntimeClient with Global Client ID:
```

MaximTkachenko:
Hi, is this issue still actual? If so I can fix it in particular project(s).

ReubenBond:
[...]

PR #6045, opened 2019-10-14
Added structured logging

Expecho:
Implemented structured logging (see #5865) for the project OrleansProviders. Doing it project by project makes it a bit more manageable than to do the whole lot in one changeset.

### DIFF
--- a/src/OrleansProviders/Storage/MemoryStorage.cs
+++ b/src/OrleansProviders/Storage/MemoryStorage.cs
@@ -56,7 +56,7 @@ namespace Orleans.Storage
             this.grainFactory = grainFactory;
 
             //Init
-            logger.Info("Init: Name={0} NumStorageGrains={1}", name, this.options.NumStorageGrains);
+            logger.Info("Init: Name={Name} NumStorageGrains={NumStorageGrains}", name, this.options.NumStorageGrains);
 
             storageGrains = new Lazy<IMemoryStorageGrain>[this.options.NumStorageGrains];
             for (int i = 0; i < this.options.NumStorageGrains; i++)
@@ -72,7 +72,7 @@ namespace Orleans.Storage
         {
             IList<Tuple<string, string>> keys = MakeKeys(grainType, grainReference).ToList();
 
-            if (logger.IsEnabled(LogLevel.Trace)) logger.Trace("Read Keys={0}", StorageProviderUtils.PrintKeys(keys));
+            if (logger.IsEnabled(LogLevel.Trace)) logger.Trace("Read Keys={Keys}", StorageProviderUtils.PrintKeys(keys));
             
             string id = HierarchicalKeyStore.MakeStoreKey(keys);
             IMemoryStorageGrain storageGrain = GetStorageGrain(id);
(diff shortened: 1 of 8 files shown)


src/OrleansProviders/Storage/MemoryStorage.cs
```diff
@@ -56,7 +56,7 @@ public MemoryGrainStorage(string name, MemoryGrainStorageOptions options, ILogge
             this.grainFactory = grainFactory;
 
             //Init
-            logger.Info("Init: Name={0} NumStorageGrains={1}", name, this.options.NumStorageGrains);
+            logger.Info("Init: Name={Name} NumStorageGrains={NumStorageGrains}", name, this.options.NumStorageGrains);
```
ReubenBond: Use `LogInformation` instead of `Info` here and similarly elsewhere

Expecho: > Use `LogInformation` instead of `Info` here and similarly elsewhere

Done!

ReubenBond:
Thank you, Peter, much appreciated!

Expecho:
> Thank you, Peter, much appreciated!

You're welcome. Now I dipped  my toes in it (and noticed the additional changes to look for) what would be the next best project in the solution to tackle?

JanEggers:
@ReubenBond , @Expecho  did you consider this to squeeze out some more performance?

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?view=aspnetcore-3.0

Expecho:
> @ReubenBond , @Expecho did you consider this to squeeze out some more performance?
> 
> https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?view=aspnetcore-3.0

I get the point, but it might requires some upfront design regarding the eventids to make sure there is some consistency.

Work could be split in two seperate issues. First move to structured logging and after that see where we could benefit from performance.

srollinet:
@JanEggers interesting link!

I only wonder it it will make this kind of pattern less maintainable
```
if (logger.IsEnabled(LogLevel.Trace)) logger.LogTrace("Read Keys={Keys}", StorageProviderUtils.PrintKeys(keys));
```
Because using `LoggerMessage.Define`, the LogLevel is defined statically, so you must check at 2 different places to ensure the LogLevels are consistent.

But this is not a big issue

JanEggers:
@srollinet you actually can remove the if in usercode because its in the box:

https://github.com/aspnet/Extensions/blob/66007df6e5b68683de0cab0dc6a8552265318f20/src/Logging/Logging.Abstractions/src/LoggerMessage.cs#L100-L116

JanEggers:
on the other hand StorageProviderUtils.PrintKeys should not be used in a structured logging approach as it removes the structure of the argument so it would be better to just pass keys without formatting them.

srollinet:
@JanEggers yes, looking the Orleans code, this pattern is used a lot and it's probably to avoid parameters evaluation.

And yes, with structured logging there are maybe other ways to format the parameters :)

PR #6047, opened 2019-10-15
Transactions: support larger state sizes in Azure Table Storage

ReubenBond:
This implementation retains backwards compatibility with existing stored data while adding support for states which are too large to fit into a single Azure Table Storage column (which holds a maximum of 32 * 1024 chars)

### DIFF
--- a/src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/StateEntity.cs
+++ b/src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/StateEntity.cs
@@ -1,54 +1,180 @@
 using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
 using Microsoft.Azure.Cosmos.Table;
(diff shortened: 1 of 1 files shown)


src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/StateEntity.cs
```diff
@@ -1,54 +1,180 @@
...
+        }
+
+        public string TransactionManager
+        {
+            get => this.GetPropertyOrDefault(nameof(this.TransactionManager))?.StringValue;
+            set => this.properties[nameof(this.TransactionManager)] = new EntityProperty(value);
+        }
 
-        public DateTime TransactionTimestamp { get; set; }
+        public string StateJson { get => this.GetStateInternal(); set => this.SetStateInternal(value); }
```
jason-bragg: Don't think we need the StateJson property any longer

PR #6053, opened 2019-10-17
Added consistent logging for all messages dropped due to expiry

jason-bragg:


src/Orleans.Core/Messaging/Message.cs
```diff
@@ -548,8 +549,9 @@ public static Message CreatePromptExceptionResponse(Message request, Exception e
             };
         }
 
-        internal void DropExpiredMessage(MessagingStatisticsGroup.Phase phase)
+        internal void DropExpiredMessage(ILogger logger, MessagingStatisticsGroup.Phase phase)
         {
+            logger.LogWarning((int)ErrorCode.Messaging_DroppingExpiredMessage, "Dropped expired message during {Phase} phase.  Message: {Message}", Enum.GetName(typeof(MessagingStatisticsGroup.Phase), phase), logger.IsEnabled(LogLevel.Trace) ? this.ToLongString() : this.ToString());
```
ReubenBond: Can't we just use `phase` here without the (expensive) `Enum.GetName` call?
image

Other than that, there's a double-space in the message.

PR #6057, opened 2019-10-19
Improved activation queue monitoring

jason-bragg:
- Added load stages to prevent spamming logs under load.
- Added max lifetime and spike high load metrics

### DIFF
--- a/src/Orleans.Runtime/Catalog/ActivationData.cs
+++ b/src/Orleans.Runtime/Catalog/ActivationData.cs
@@ -387,6 +387,17 @@ namespace Orleans.Runtime
         /// </summary>
         private List<Message> waiting;
 
+        private enum LoadState
+        {
+            Low,
+            Hot,
+            Overload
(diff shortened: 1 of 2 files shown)


src/Orleans.Runtime/Catalog/ActivationData.cs
```diff
@@ -387,6 +387,17 @@ public void ResetRunning(Message message)
...
+        private enum LoadState
+        {
+            Low,
+            Hot,
+            Overload
+        }
+
+        private LoadState loadState = LoadState.Low;
+        int lifetimeHighLoad = 0;
+        int spikeHighestLoad = 0;
```
ReubenBond: I like the idea of more monitoring, but this adds 3 new fields per activation. Can we do it in a cheaper way?

jason-bragg:
Probably better to leave as is and add a counter for queue size.  Closing

PR #6071, opened 2019-10-28
Azure table grain storage inconsistent state on not found

jason-bragg:
Azure table grain storage returns inconsistent state if modifying state that does not exist.


ReubenBond:
Should we also add this behavior to blob storage?

jason-bragg:
Possibly.  It could be an issue in other providers.  I've not investigated.

PR #6092, opened 2019-11-05
Update OneBoxDeployment to .NET Core 3

veikkoeeva:
Updates OneBoxDeployment to .NET Core 3 while also cleaning code where appropriate.


veikkoeeva:
This is not ready, but I thought that could open a PR nevertheless.

ReubenBond:
Thanks, Veikko ☺

veikkoeeva:
@ReubenBond, all right. I submit the rest of changes in separate PR(s). :) For instance, in problem case the problem detail result doesn't have the headers etc. set like I would and didnt't have time to find out. Consequently also the test that checks the error resolts has set these correctly fails. Maybe not too fatal, maybe better to have something in and not waiting the next three weeks I get things resolved. :D

PR #6097, opened 2019-11-07
Fix for MySql clustering script

oleggolovkovuss:
The script is failing because the `OrleansQuery` table is not created before trying to insert something there. I think it was lost during "Split something" refactoring which this fix is copied from. I think all other clustering ADO.Net scripts have the same problem

### DIFF
--- a/src/AdoNet/Orleans.Clustering.AdoNet/MySQL-Clustering.sql
+++ b/src/AdoNet/Orleans.Clustering.AdoNet/MySQL-Clustering.sql
@@ -1,3 +1,11 @@
+CREATE TABLE OrleansQuery
+(
+    QueryKey VARCHAR(64) NOT NULL,
+    QueryText VARCHAR(8000) NOT NULL,
+
+    CONSTRAINT OrleansQuery_Key PRIMARY KEY(QueryKey)
+);
+
 -- For each deployment, there will be only one (active) membership version table version column which will be updated periodically.
 CREATE TABLE OrleansMembershipVersionTable
 (


oleggolovkovuss:
Looks like the same problem exists for the persistence setup scripts. Also all of the MySQL scripts aren't working from the C# code (using `MySql.Data`) because of the `DELIMITER` command

sergeybykov:
@veikkoeeva Do you approve?

veikkoeeva:
The required part is at https://github.com/dotnet/orleans/tree/master/src/AdoNet/Shared and needs to be deployed first.

The rationale for splitting was that not all tables and queries are needed in all cases so split like to only those that are needed. Since this is the case, maybe this isn't the right fix but instruct people to deploy that one first.

Though it may be worth considering if the scripts should be merged into one. This would require a bit changes to testing too. Though if one does this, it's tempting to consider removing the `#ifdef` constructs, have a dll for the core piece that also has the scripts and then separate packages for reminders, membership etc. that has the small pieces of code.

oleggolovkovuss:
@veikkoeeva oh, thanks, I see. I can't find any mention of `Shared` either at 
https://dotnet.github.io/orleans/Documentation/grains/grain_persistence/relational_storage.html
or at
https://dotnet.github.io/orleans/Documentation/clusters_and_clients/configuration_guide/adonet_configuration.html
though so I thought it was lost during refactoring and I had to recreate it myself. I guess a complete setup sample from C# would be helpful

veikkoeeva:
@oleggolovkovuss The docs seem to be perpetually in process to be improved. :) You are right of course that this should be told better.

There is one bigger sample at https://github.com/dotnet/orleans/tree/master/Samples/OneBoxDeployment if you are interested. It uses [dacpac](https://docs.microsoft.com/en-us/sql/relational-databases/data-tier-applications/data-tier-applications) and so is not directly applicble to your case. You could though switch the connection strings to point to MySQL and switch the ADO.NET connector library.

Related to the previous single script consideration, if there were a single script and it were available in known location, it could deploy it to database during setup. The tests do this, for instance. It has occurred to me to enhance the ADO.NET configuration surface so that there'd be an extension method like `Deploy` that would take the database name (and ADO.NET vendor information) and create the database and deploy the scripts there. It likely would be helpful to many people while still maintaining flexibility to run any tool one might use in one's enterprise, dealing with security or whatever other concerns there may be.

sergeybykov:
I submitted #6118 to add information about Shared/Main SQL scripts.

sergeybykov:
Now that the docs were updated, should we close this?

sergeybykov:
>Besides a small typo Persiatence I guess so, thanks for the feedback!

Thanks for pointing it out. I submitted #6124 to fix it.

PR #6101, opened 2019-11-07
Add note about Visual Studio Community Edition

fdbeirao:
The `ArtifactReference` and `HintPath` in `OneBoxDeployment.Database.sqlproj` assume that Visual Studio 2019 **Enterprise** edition are installed.

If the user is using Visual Studio 2019 Community edition, she will observe ~115 build errors, where the first is:

`SQL72027: File "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\Extensions\Microsoft\SQLDB\Extensions\SqlServer\150\SqlSchemas\master.dacpac" does not exist.`

I understand if this PR is not merged *as-is*, but I just spent ~1h and went through adding a bunch of options to my Visual Studio 2019 Community Edition, before actually figuring out what the issue was.

It is my hope that this single line can help at least my future self.

### DIFF
--- a/Samples/OneBoxDeployment/Readme.md
+++ b/Samples/OneBoxDeployment/Readme.md
@@ -20,6 +20,8 @@ At any time the database can be reset or snapshots deleted. There shouldn't be a
 
 ## Setting up the SQL Server Database
 
+0. (_optional_) If you are using Visual Studio 2019 Community Edition, you need to edit the `OneBoxDeployment.Database.sqlproj` file, and set the `ArtifactReference` and `HintPath` to point to `Community`, instead of `Enterprise`.
+
 1. This needs to be done only the first time
 Open the _OneBoxDeployment.Database_ project and open _Debug_ tab. If the target connection string isn't set, put _Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=OneBoxDeployment.Database;Integrated Security=True;Persist Security Info=False;Pooling=True;MultipleActiveResultSets=True;Connect Timeout=60;Encrypt=False;TrustServerCertificate=True_ to it. This deploys the database to LocalDb with the current user rights.
 


sergeybykov:
Thank you, @fdbeirao!

veikkoeeva:
This definitely should be automated somehow. Looks like the source of this problem is the few additional system views to query some health data from the database. They could be removed from this sample, though better fix would be make it just more robust. Have to think this...

fdbeirao:
A hackish workaround that would "just work", but not be the most brilliant thing ever, would be in the `OneBoxDeployment.Database.sqlproj` to have something like:

```xml
  <ItemGroup>
    <ArtifactReference Condition="Exists('C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\Extensions\Microsoft\SQLDB\Extensions\SqlServer\150\SqlSchemas\master.dacpac')" Include="C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\Extensions\Microsoft\SQLDB\Extensions\SqlServer\150\SqlSchemas\master.dacpac">
      <HintPath>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\Extensions\Microsoft\SQLDB\Extensions\SqlServer\150\SqlSchemas\master.dacpac</HintPath>
      <SuppressMissingDependenciesErrors>False</SuppressMissingDependenciesErrors>
      <DatabaseVariableLiteralValue>master</DatabaseVariableLiteralValue>
    </ArtifactReference>
    <ArtifactReference Condition="Exists('C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\Extensions\Microsoft\SQLDB\Extensions\SqlServer\150\SqlSchemas\master.dacpac')" Include="C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\Extensions\Microsoft\SQLDB\Extensions\SqlServer\150\SqlSchemas\master.dacpac">
      <HintPath>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\Extensions\Microsoft\SQLDB\Extensions\SqlServer\150\SqlSchemas\master.dacpac</HintPath>
      <SuppressMissingDependenciesErrors>False</SuppressMissingDependenciesErrors>
      <DatabaseVariableLiteralValue>master</DatabaseVariableLiteralValue>
    </ArtifactReference>
  </ItemGroup>
```

I think this could be refactored, but I am not 100% familiar with this XML programming syntax. This would work for both Enterprise and Community editions, without the user having to modify anything locally.

veikkoeeva:
@fdbeirao Do you have energy to send a PR for that? :) It looks like it would improve the situation.

I document a few more thoughts since it is likely someone ends up with this same trouble.

Beyond these and labouring a bit, there seem to be `$(DacPacRootPath)` variable, which could afford a robust alternative to listing default install locations. Build Tools is one that probably has a different path depending on environment. A factor that does weigh in when considering solutions is that `.sqlproj` still requires .NET Full, which is a small bump in the road when plugging this into a CI/CD solution. 

I made a few quick searches and list some cases about finding the packages, previous bugs, blog posts and then a few Nuget packages. The larger case of using solutions like dacpac is to do controlled deployments and it could be that some of the tools can also solve this problem by injectig search logic or perhaps including the master database schema.

- https://github.com/microsoft/azure-pipelines-tasks/issues/10499
- https://feedback.azure.com/forums/908035-sql-server/suggestions/32897047-visual-studio-2017-ssdt-adds-hardcoded-master-dacp
- https://abelsquidhead.com/index.php/2019/09/02/github-actions-2-0-lets-do-something-a-little-more-involved/
- https://www.gatevnotes.com/continuous-integration-continuous-deployment-of-ssdt-projects-creating-azure-devops-pipelines/
- https://www.nuget.org/packages/Dacpac.Tool/

Additional note: One could remove the troublesome views and consequently remove the need to reference the master database schema. Though solving this for the community and saving collective time would be better.

PR #6102, opened 2019-11-08
Fix routing of gateway count changed events to registered servi…

### LINKED ISSUE #6098: AddGatewayCountChangedHandler does not work
`AddGatewayCountChangedHandler` doesn't work because the implementation of `OutsideRuntimeClient` doesn't forward on the event like he does for connection lost - https://github.com/dotnet/orleans/blob/c683fa8f2fddf310c2dfc3c9cf43db4fb80be1ad/src/Orleans.Core/Runtime/OutsideRuntimeClient.cs#L121

This means there doesn't seem to be a simple way to know when a connection is re-established? Something that is necessary for client subscription to be resumed.

mindlink:
Update the `OutsideRuntimeClient` to add registered `GatewayCountChanged` services as handlers of the internal event. This fixes #6098.

Additionally, when disposed remove the registered handlers in cases where the cluster client is disposed and registered handlers live longer.

### DIFF
--- a/src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
+++ b/src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
@@ -124,6 +124,12 @@ namespace Orleans
                     this.ClusterConnectionLost += handler;
                 }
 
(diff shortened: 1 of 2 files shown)


ReubenBond:
Big thank you, @mindlink!

PR #6109, opened 2019-11-11
Removing Extra Angle Bracket

tecfield:
Removes extra angle bracket in `<JournaledGrain<StateType,EventType>`


sergeybykov:
Thank you, @tecfield!

Issue #6096, opened 2019-11-07
Logger names for PubSubRendezvousGrain too fine grained


eisendle:
I'm investigating a memory issue in our orleans cluster. While investigating a process dump, I discovered there is a large amount of memory retained inside a dictionary in Microsoft.Extension.Logging,LoggerFactory. This dictionary keeps a mapping between LoggerName and logger. It seems that foreach stream event the PubSubRendezvousGrain is handling, a new logger name and logger is created. 

In the dump, the this dictionary contains over 40000 loggers which results in 26 MB of memory retention and I suspect it just keeps growing. For reference we use Serilog as logging provider.

Here some examples of created logger names:

`Orleans.Streams.PubSubRendezvousGrain.S172.31.201.57:50000:310208942.*grn/716E8E94/40a234f781217745a9800a6c5a9bb98806000000716e8e94+XXXStreamProvider_XXX.XXXChanged-0x5F8B8B46`

`Orleans.Streams.PubSubRendezvousGrain.S172.31.201.57:50000:310208942.*grn/716E8E94/4161ca03098d9df78600e17e3aa7dda206000000716e8e94+XXXStreamProvider_XXX.XXXChanged-0x18AADBD0`

`Orleans.Streams.PubSubRendezvousGrain.S172.31.201.57:50000:310208942.*grn/716E8E94/47ab2e70ab630b13a05933bcab52978e06000000716e8e94+XXXStreamProvider_XXX.XXXChanged-0xB5C9A4D0`

I think it would be better when the logger names in case would more coarse grained and the grain identity and event information should be moved to the logged message.

So the proposed logger name should just be:

`Orleans.Streams.PubSubRendezvousGrain`

eisendle:
Resolved via #6122

PR #6122, opened 2019-11-14
Do not include grain identifier in the ILogger category name

sehra:
In the constructor of PubSubRendezvousGrain, it requests an ILogger using category name based on the grain identifier. This might not be an issue depending on the ILoggerFactory implementation, but the one in Microsoft.Extensions.Logging caches them all in a dictionary based on that category name.


ReubenBond:
Thank you, @sehra!

sergeybykov:
Does this also resolve #6096?
/cc @eisendle

eisendle:
@sergeybykov Yep, this should resolve the issue.

PR #6126, opened 2019-11-17
DynamoDB: support extended membership protocol

### LINKED ISSUE #6099: DynamoDB clustering: support extended membership protocol
The current implementation of the DynamoDB clustering provider only supports the initial membership protocol, not the later [extension to the protocol](https://dotnet.github.io/orleans/Documentation/implementation/cluster_management.html#extension-to-totally-order-membership-views) to support totally ordered views of cluster membership.

It's important to support this so that the provider will function correctly.

See this PR which adds such support to Consul: https://github.com/dotnet/orleans/pull/6095

jsteinich:
Implements the extended membership protocol for DynamoDB.

Fixes #6099 

Still need to go through testing.

I noticed a few things while implementing this:

1. `ReadAll` isn't guaranteed to return all silo entries at the same version. It seemed like other implementations can also be inconsistent, but I'm not familiar enough to say for certain.
2. The Azure implementation seems to be more aggressive at throwing exceptions when rows don't exist than other implementations. I'm not sure what the preference is.
3. `DynamoDBMembershipTable` is getting a bit hard to read with all the low level dynamo db. It could be broken up, but it may be better to use a higher level api. That type of refactoring seems outside the scope of this PR though.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -96,7 +96,7 @@
     <FSharpCoreVersion>4.7.0</FSharpCoreVersion>
 
     <!-- 3rd party packages -->
-    <AWSSDKDynamoDBv2Version>3.3.4.17</AWSSDKDynamoDBv2Version>
+    <AWSSDKDynamoDBv2Version>3.3.102.2</AWSSDKDynamoDBv2Version>
     <AWSSDKSQSVersion>3.3.2.7</AWSSDKSQSVersion>
     <BondCoreCSharpVersion>5.3.1</BondCoreCSharpVersion>
     <ConsulVersion>0.7.2.3</ConsulVersion>
--- a/src/AWS/Orleans.Clustering.DynamoDB/Membership/DynamoDBMembershipTable.cs
+++ b/src/AWS/Orleans.Clustering.DynamoDB/Membership/DynamoDBMembershipTable.cs
@@ -7,6 +7,8 @@ using Orleans.Runtime;
 using Orleans.Runtime.MembershipService;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
 using System.Net;
 using System.Text;
 using System.Threading.Tasks;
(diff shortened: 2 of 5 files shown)


jsteinich:
@sergeybykov I wasn't able to schedule time at work, so my time to work on this has been limited. I'm hoping to spend some more time this weekend, but realistically, it will probably be another week before I feel confident in the changes since I'd like to run in a test cluster for a few days after finishing initial tests.

@ReubenBond Any comments on my observations?

ReubenBond:
> `ReadAll` isn't guaranteed to return all silo entries at the same version. It seemed like other implementations can also be inconsistent, but I'm not familiar enough to say for certain.

The Consul implementation performs the reads in a transaction, so it's ok. There may be an issue with the Azure Table Storage implementation - I am not able to confirm yet.

To rectify issues like that could store `TableVersion` along with each modified row and check if any entry has a `TableVersion` higher than the version 'row'. If so, retry the read (potentially logging a warning). That relies on the `TableVersion` row being read first (an algo which doesn't rely on that is possible, too, but more complex: store the modifed row keys with the version row and check their versions match). Just an idea. If there's a way to ensure consistent reads at the database level, then that's preferable.

> The Azure implementation seems to be more aggressive at throwing exceptions when rows don't exist than other implementations. I'm not sure what the preference is.

I'm not sure either way. Generally if an expected row doesn't exist, that's an error, but if it doesn't affect semantics then it can be ignored. Both the Azure & Consul providers will fail indefinitely if the user manually deletes some data or drops the tables.

> `DynamoDBMembershipTable` is getting a bit hard to read with all the low level dynamo db. It could be broken up, but it may be better to use a higher level api. That type of refactoring seems outside the scope of this PR though.

I agree that refactoring should be done in a separate PR.

jsteinich:
I've updated `ReadAll` with the simple method of consistency. I've also switched the existing membership tests to run with extended membership and got them passing.

I was hoping to put into a test cluster and run for awhile, but it is still on 2.3.x so it will take me a bit longer to try that out (still hoping to start that this weekend).

jsteinich:
I now have a test cluster running with these changes. If everything seems stable after a few days, I'll drop the wip tag.

ReubenBond:
Huge thank you, @jsteinich!

PR #6127, opened 2019-11-17
Allow negative values in TypeCodeAttribute. Fixes #6114

### LINKED ISSUE #6114: Codegen ignores negative integer type codes for TypeCodeOverride attribute
The grain interface and class codes could be a negative integer as returned by [Utils.CalculateIdHash](https://github.com/dotnet/orleans/blob/master/src/Orleans.Core/Utils/Utils.cs#L200) But codegen skips such codes and instead is falling back to re-calculation.
https://github.com/dotnet/orleans/blob/bc41bf386095a335b7ccaeda431b5ebf2fd9ac60/src/Orleans.Core/CodeGeneration/GrainInterfaceUtils.cs#L424

Having `TypeCodeOverride` is a nice workaround for system evolution scenarios where names/assemblies of the grain interfaces and classes may change, and the ability to leave the old type code is crucial. It would be great if this could be patch fixed for 2.4.

yevhen:
That need to be cherry-picked into 2.4 branch. BTW, what is the workflow for merging PRs into multiple branches? Should I create another PR or is that done only by maintainers?

### DIFF
--- a/src/Orleans.Core/CodeGeneration/GrainInterfaceUtils.cs
+++ b/src/Orleans.Core/CodeGeneration/GrainInterfaceUtils.cs
@@ -421,10 +421,7 @@ namespace Orleans.CodeGeneration
         private static int GetTypeCode(Type grainInterfaceOrClass)
         {
             var attr = grainInterfaceOrClass.GetCustomAttributes<TypeCodeOverrideAttribute>(false).FirstOrDefault();
-            if (attr != null && attr.TypeCode > 0)
-            {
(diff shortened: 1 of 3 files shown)


ReubenBond:
Thanks, @yevhen. The typical workflow for us is that we will create a commit to rebase this onto 2.4.x

yevhen:
> Thanks, @yevhen. The typical workflow for us is that we will create a commit to rebase this onto 2.4.x

Super! Thanks!

PR #6137, opened 2019-11-23
Update tutorial_1.md

jeramyRR:
Changed pass to across.  I think it is more often phrased as "across machine boundaries" than "past machine boundaries".
Added public to SayHello method as it's an interface implementation and must be public.

### DIFF
--- a/src/Documentation/tutorials_and_samples/tutorial_1.md
+++ b/src/Documentation/tutorials_and_samples/tutorial_1.md
@@ -62,7 +62,7 @@ You will also probably need to add `using` statements.
 
 `Microsoft.Orleans.Core.Abstractions` is needed everywhere. It included in both `Microsoft.Orleans.Server` and `Microsoft.Orleans.Client`.
 
-`Microsoft.Orleans.CodeGenerator.MSBuild` automatically generates code that is needed to make calls to grains pass machine boundaries. So it needed in both `GrainInterfaces` and `Grains` projects.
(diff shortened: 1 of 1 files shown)


src/Documentation/tutorials_and_samples/tutorial_1.md
```diff
@@ -99,7 +99,7 @@ namespace OrleansBasics
             this.logger = logger;
         }
 
-        Task<string> IHello.SayHello(string greeting)
+        public Task<string> IHello.SayHello(string greeting)
```
ReubenBond: Because this is an explicit interface implementation (maybe it was to show that that's where the method is defined for the purpose of the tutorial), it cannot have an accessibility modifier

jeramyRR: This isn't in the interface.  It's the method implementation.  Without the public, it won't compile.

jeramyRR: image

ReubenBond: I'm referring to the `IHello.` text before `SayHello` in the original code:

``` C#
Task<string> IHello.SayHello(string greeting)
```

If you add `public` then you also must remove `IHello.`, since explicit interface implementations cannot have access modifiers.

Valid:
``` C#
Task<string> IHello.SayHello(string greeting)
```

Valid:
``` C#
public Task<string> SayHello(string greeting)
```

Invalid:
``` C#
public Task<string> IHello.SayHello(string greeting)
```

jeramyRR: Ah, I learned something new today.  I didn't even notice the IHello prefix to SayHello.

PR #6138, opened 2019-11-24
Update local_development_configuration.md

jeramyRR:
Adds link to Orleans 3.0 Hello World sample, while directing readers to older versions if wanted.
Fixes loopback wrapper ending character.  Was a ' instead of a \`.
Adds space between `ISiloBuilder.Configure` and "method".


ReubenBond:
Thanks again!

PR #6154, opened 2019-12-03
Implement mTLS support for .NET Standard 2.0

juyuz:
Target Orleans.Connections.Security to .net standard 2.0


ReubenBond:
Thanks for submitting this, @juyuz - I'll take a closer look soon. There are some sync-over-async paths which I need to double-check before we can merge.

ReubenBond:
@juyuz do you mind if I take over this PR? I'll likely multi-target the project. I just need to clean a few things up (sync blocking in places, etc)

juyuz:
Go ahead! Thanks for the help.

src/Orleans.Connections.Security/Security/TlsServerConnectionMiddleware.cs
```diff
@@ -151,35 +152,49 @@ private async Task InnerOnConnectionAsync(ConnectionContext context)
...
                     };
 
                     _options.OnAuthenticateAsServer?.Invoke(context, sslOptions);
 
-                    await sslStream.AuthenticateAsServerAsync(sslOptions, CancellationToken.None);
+                    await sslStream.AuthenticateAsServerAsync(
+                        sslOptions.ServerCertificate,
+                        sslOptions.ClientCertificateRequired,
+                        sslOptions.EnabledSslProtocols,
+                        sslOptions.CertificateRevocationCheckMode == X509RevocationMode.Online);
```
ReubenBond: I have some reservations around this. Specifically, there are other options on `SslServerAuthenticationOptions` which we aren't reflecting in this call, so the user misses out on the opportunity to access those options.

ReubenBond: The new approach allows users to modify the underlying options type if they want (on .NET Core only, since .NET Standard 2.0 does not support it)

ReubenBond:
@juyuz could you please take a look and let me know if this fits your needs?

juyuz:
> @juyuz could you please take a look and let me know if this fits your needs?

Yes, this change works for us. Thanks.

src/Orleans.Connections.Security/Security/TlsClientConnectionMiddleware.cs
```diff
@@ -65,15 +66,15 @@ private async Task InnerOnConnectionAsync(ConnectionContext context)
                 leaveOpen: true
             );
 
-            SslDuplexPipe sslDuplexPipe = null;
+            TlsDuplexPipe sslDuplexPipe = null;
```
benjaminpetit: minor but it should be renamed to `tlsDuplexPipe` now

src/Orleans.Connections.Security/Security/TlsServerAuthenticationOptions.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public object SslServerAuthenticationOptions => this.Value;
+#else
+        public X509Certificate ServerCertificate { get; set; }
+
+        public ServerCertificateSelectionCallback ServerCertificateSelectionCallback { get; set; }
+
+        public bool ClientCertificateRequired { get; set; }
+
+        public SslProtocols EnabledSslProtocols { get; set; }
```
benjaminpetit: Why not set the default value as in `TlsOptions.SslProtocols`? I think the default value for `SslProtocols` is to use obsolete protocols

ReubenBond: There's `SslProtocols.Default`, which uses obsolete protocols, but the default value of `SslProtocols` is `None` (`0`), which means no protocols are specified.

We can either choose `SslProtocols.None` and make the user specify, or pick some value which we believe is safe today (but may not be safe in future).

src/Orleans.Connections.Security/Security/TlsClientConnectionMiddleware.cs
ReubenBond: Thanks, fixed

ReubenBond:
I believe this is ready

PR #6190, opened 2019-12-19
Multi-target Orleans sln and tests

ReubenBond:
This won't pass on CI just yet - and there are probably a bunch of other changes needed.

Multi-targets test projects to netcoreapp3.1 & net461 (we should bump that to net48 soon, or at least net472)

Multi-targets Orleans.Core & Orleans.Runtime to netcoreapp3.1 & netstandard2.0.

I had to bump some package versions to get things aligning nicely on netcoreapp3.1


ReubenBond:
Functionals won't run correctly here until I update the pipeline, but they pass in an updated version of the pipeline.

src/Orleans.CodeGenerator/Analysis/CompilationAnalyzer.cs
```diff
@@ -422,7 +422,7 @@ public void AddDependencyType(ITypeSymbol type)
             if (named.IsGenericType && !named.IsUnboundGenericType)
             {
                 var unbound = named.ConstructUnboundGenericType();
-                if (unbound.Equals(this.wellKnownTypes.Task_1))
+                if (SymbolEqualityComparer.Default.Equals(unbound, this.wellKnownTypes.Task_1))
```
ReubenBond: An analyzer started complaining about how we were comparing symbols.

Issue #6189, opened 2019-12-18
MayInterleave attribute throws when in base class


cjenp:
Hi, 

i have an abstract base class that has the MayInterleave attribute specified. The public
static IsReentrant method is also defined in this class. When i create a class that derives from the base class i get the following error on silo startup

> System.InvalidOperationException: 'Class SampleGrain.MyGrain doesn't declare public static method with name IsReentrant specified in MayInterleave attribute'

Im currently solving this problem with adding the IsReentrant method to each derived class and using the new keyword.

From what i can see this could be solved by adding the BindingFlags.FlattenHierarchy flag to the GetMethod() call in GrainTypeData.cs, or am i missing something?
```C#
var callbackMethodName = grainType.GetCustomAttribute<MayInterleaveAttribute>().CallbackMethodName;
var method = grainType.GetMethod(callbackMethodName, BindingFlags.Public | BindingFlags.Static);
if (method == null)
      throw new InvalidOperationException($"Class {grainType.FullName} doesn't declare public static method " + $"with name {callbackMethodName} specified in MayInterleave attribute");
```

ReubenBond:
Good find, are you able to submit a PR? I think the search should start from the class which the attribute is defined on and traverse upwards from there

cjenp:
Sure, i will make the changes and submit a PR.

ReubenBond:
Fixed by #6192 - thanks, @cjenp!

PR #6192, opened 2019-12-19
Allow MayInterleaveAttribute on base grains. Fix for issue #6189

cjenp:
This fixes an exception that occures when the MayInterleaveAttirbute is defined for a base grain.


ReubenBond:
Thank you for this, @cjenp!

PR #6199, opened 2019-12-24
Fixes for 2.4.5 cherry-picked from master

sergeybykov:


ReubenBond:
We can either ignore CI or add a global.json

sergeybykov:
Let's try to add global.json. I don't understand what needs to be there though.

ReubenBond:
It's because I edited the CI pipeline to make builds on .NET Core 3.1 work for the multi-target work. Master had a global.json already, but the 2.x branch doesn't.

sergeybykov:
Looks like it still breaks on Linux and MacOS. I can't set it to 3.1.100 without fixing building of F# libraries. Maybe it's not worth fixing in 2.4.* anyway.

PR #6208, opened 2020-01-02
refactor(blazor sample): use asp.net core hosting like the webapi template

ElderJames:


ReubenBond:
Thanks for the PR! Could you change it to avoid using the `Startup` class and instead configure the Web builder directly? You can still have separate methods for ConfigureServices/etc.

ElderJames:
I think using the `Startup` class is more like the .net sdk template.

ElderJames:
I fix it now.

ReubenBond:
Thank you, @ElderJames!

PR #6216, opened 2020-01-07
Minor perf tweak for RequestContext when removing last item

ReubenBond:
When removing the last entry from a RequestContext, set the value to null instead of creating a new dictionary with a copy of the current context and removing the last item.


veikkoeeva:
More out of curiosity than other things, how does `null` improve performance here? I have not looked into this, likely just lazy question. I think if it's serialization, but if not, thinking static, empty Dictionary as an alternative.

ReubenBond:
This avoids allocating a dictionary (couple allocs) needlessly. It was a drive-by PR while I was resurrecting my distributed tracing branch (which is where I pulled this from)

PR #6221, opened 2020-01-08
Add OrleansDebuggerHelper.GetGrainInstance to aid in local debugging

### LINKED ISSUE #6220: Debugger support for interrogating grains
When I'm debugging, I am often working in a single silo.  The methods I am using will typically have some grain references to other grains that I will be calling or have called.  It would be really nice to have a debugger helper that allowed me to go from a grain id to the C# object underlying the grain.  This helper would need to be something that could be evaluated in the immediate or watch windows of the debugger.   If the grain id refers to a non-local grain then clearly it needs to return null or throw an exception.

ReubenBond:
Fixes #6220 
cc @OracPrime

Example usage (from Immediate Window in VS debugger):

```C#
Orleans.Runtime.Utilities.OrleansDebuggerHelper.GetGrainInstance(myGrainReference)
```

### DIFF
--- a/src/Orleans.Runtime/Utilities/OrleansDebuggerHelper.cs
+++ b/src/Orleans.Runtime/Utilities/OrleansDebuggerHelper.cs
@@ -1,0 +1,40 @@
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Orleans.Runtime.Utilities
+{
+    /// <summary>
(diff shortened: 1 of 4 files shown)


OracPrime:
Should it be GetLocalGrainInstance ?  Or is it going to instantiate a grain which is currently remote?

ReubenBond:
It will try to get a grain instance and return null if it fails

PR #6232, opened 2020-01-14
Update tutorial_1

Mrchazaaa:
The Grains project also depends on `Microsoft.Orleans.CodeGenerator.MSBuild`


ReubenBond:
Thank you for the contribution - I think you may have misread point 3 in the list.
The way it is in this PR sounds odd to me: ~"add the logging abstractions and codegen package for logging".

Mrchazaaa:
Hey, thanks for catching that. I think it looks good now.

ReubenBond:
Thanks, @Mrchazaaa!

PR #6234, opened 2020-01-14
Avoid capturing ExecutionContext in GrainTimer and other timers

### LINKED ISSUE #5563: AsyncLocal values flow into timer callbacks and stream observers
The values of AsyncLocals will flow into timer callbacks and stream observers from their registration point. RequestContext uses an AsyncLocal to manage properties, which is probably why they are like this, but it shouldn't affect all AsyncLocals.

This is quite unintuitive for me because I would only expect these to flow whenever I call an async method.

In my project `Trace.Begin` will initialize an AsyncLocal and disposing it will set it back to null.  `$1` is the grain instance and `$2` is the value that was stored in the AsyncLocal (which has been cleared to null).

When the timer callback is called I see the AsyncLocal value from the point of registration.

Wrapping the timer registration with the following prevents the values from flowing through:
```csharp
using (ExecutionContext.SuppressFlow())
{
    // register things here ...
}
```

ReubenBond:
Fixes #5563

This means that `RequestContext` will not automatically flow into timers, etc. My belief is that this is the correct behavior and that the current behavior is a likely source of bugs.

### DIFF
--- a/src/Orleans.Core/Timers/AsyncTaskSafeTimer.cs
+++ b/src/Orleans.Core/Timers/AsyncTaskSafeTimer.cs
@@ -4,44 +4,14 @@ using Microsoft.Extensions.Logging;
 
 namespace Orleans.Runtime
 {
-    internal class AsyncTaskSafeTimer : IDisposable
+    internal class AsyncTaskSafeTimer : SafeTimerBase
     {
-        private readonly SafeTimerBase safeTimerBase;
-
-        public AsyncTaskSafeTimer(ILogger logger, Func<object, Task> asynTaskCallback, object state)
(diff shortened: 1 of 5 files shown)


ReubenBond:
(flaky test - unrelated)

ReubenBond:
This is a breaking change. It is *possible* that a user accidentally relies on RequestContext/AsyncLocals/ExecutionContext flowing into timers.

I think we should merge this as a fix and call it out in the release notes as a minor breaking change.
@sergeybykov - wdyt?

sergeybykov:
IMO it would be very odd to rely on capturing context and seeing it in timer callbacks.

PR #6239, opened 2020-01-15
use possible instead of possibly

patrick-m-kendall:


ReubenBond:
Thank you, @patrick-m-kendall!

PR #6241, opened 2020-01-16
Enable remaining test suites

ReubenBond:


Test.cmd
```diff
@@ -39,13 +39,10 @@ set TESTS=^
 %CMDHOME%\test\Transactions\Orleans.Transactions.Tests,^
 %CMDHOME%\test\Transactions\Orleans.Transactions.Azure.Test,^
 %CMDHOME%\test\TestInfrastructure\Orleans.TestingHost.Tests,^
-%CMDHOME%\test\DependencyInjection.Tests
-
-:: Add to TESTS once dotnet-xunit supports .NET Core 3.0 (post dotnet-xunit v2.4.1)
```
ReubenBond: This note here is essentially what this PR fixes. We stopped using `dotnet-xunit` since it doesn't get updated and doesn't support the newer .NET Core releases. Instead, we use `dotnet test` (still XUnit, but via VSTest).

PR #6251, opened 2020-01-21
Adds MySqlConnector library using invariant MySql.Data.MySqlConnector

### LINKED ISSUE #6230: Add MySqlConnector AdoNet Provider
Would it be possible for the next release to add the [MySQLConnector](https://github.com/mysql-net/MySqlConnector) ADO.NET Driver support to Orleans? 

MySQLConnector Provider is a drop in replacement for Oracle's MySQL Provider, but provides true async support (rather than the "sync over async" that Oracle's MySQL Provider implements). 

While the namespaces are unchanged to `MySql.Data.MySqlClient`, using `InvariantNameMySql` is not working, because the provider is [hard-coded](https://github.com/dotnet/orleans/blob/dad62cc16d4d6de671563f6b1e54196f617926a5/src/AdoNet/Shared/Storage/DbConnectionFactory.cs#L30) to look for `MySql.Data` assembly, but MySqlConnector assembly is called `MySqlConnector` and hence can't be accessed when using the `AdoNetInvariants.InvariantNameMySql`.

Due to true lack of async support, Oracle's MySQL Provider is 6 times slower than similarly setup MSSQL Provider (a load test on MSSQL Server does run in 22 Seconds where Oracle's MySQL Provider takes 5 minutes and 30 seconds)

veikkoeeva:
This adds support for MySqlConnector library using MySql.Data.MySqlConnector
invariant string.

Note: MySql.Data.MySqlConnector library uses the same namespace as the
official Oracle library 'MySql.Data.MySql' which has traditionally
interpreted as the invariant name also. Hence in Orleans the invariant
string is different than in the library used. Likely support for
DbProviderFactory changes this as using that one can use load library
separate from the invariant.

Fixes #6230.

### DIFF
--- a/src/AdoNet/Shared/Storage/AdoNetInvariants.cs
+++ b/src/AdoNet/Shared/Storage/AdoNetInvariants.cs
@@ -21,13 +21,18 @@ namespace Orleans.Tests.SqlUtils
         /// <summary>
         /// A list of the supported invariants.
         /// </summary>
+        /// <remarks>The invariant names here do not match the namespaces as is often the convention.
+        /// Current exception is MySQL Connector library that uses the same invariant as MySQL compared
+        /// to the official Oracle distribution.</remarks>
         public static ICollection<string> Invariants { get; } = new Collection<string>(new List<string>(new[]
         {
             InvariantNameMySql,
             InvariantNameOracleDatabase,
             InvariantNamePostgreSql,
             InvariantNameSqlLite,
-            InvariantNameSqlServer
+            InvariantNameSqlServer,
+            InvariantNameSqlServerDotnetCore,
+            InvariantNameMySqlConnector
         }));
 
         /// <summary>
@@ -59,5 +64,10 @@ namespace Orleans.Tests.SqlUtils
         /// Dotnet core Microsoft SQL Server invariant name string.
         /// </summary>
         public const string InvariantNameSqlServerDotnetCore = "Microsoft.Data.SqlClient";
+
+        /// <summary>
+        /// An open source implementation of the MySQL connector library.
+        /// </summary>
+        public const string InvariantNameMySqlConnector = "MySql.Data.MySqlConnector";
     }
 }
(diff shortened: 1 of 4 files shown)


src/AdoNet/Shared/Storage/AdoNetInvariants.cs
```diff
@@ -21,13 +21,18 @@ internal static class AdoNetInvariants
...
+        /// to the official Oracle distribution.</remarks>
         public static ICollection<string> Invariants { get; } = new Collection<string>(new List<string>(new[]
         {
             InvariantNameMySql,
             InvariantNameOracleDatabase,
             InvariantNamePostgreSql,
             InvariantNameSqlLite,
-            InvariantNameSqlServer
+            InvariantNameSqlServer,
+            InvariantNameSqlServerDotnetCore,
```
veikkoeeva: Added here also the SQL Server .NET Core library in order which they appear later. It wasn't catched in the review last time (this is for information purposes if one wants to print all invariants).

test/Extensions/TesterAdoNet/Tester.AdoNet.csproj
```diff
@@ -7,6 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="1.1.0" />
```
veikkoeeva: I added these so one could quickly test also using these libraries by switching the hardcoded invariants. Not strictly necessary, although they're a dependency. I can remove if that's not OK.

veikkoeeva:
As a side note, I would like to work on the test infrastructure especially on ADO.NET side to make it use DI & async in XUnit setup and dispose now that I know it's possible. :)

ReubenBond:
Something F# quotations related. You may need to perform a clean rebuild - it looks like some types changed after that code was generated.

veikkoeeva:
It helped, thanks @ReubenBond. I updated some invariants in tests and checked the MySQL tests pass with this another connector. However, I did not add facilities to permanently run MySQL tests with bot of these providiers, only the official Oracle one is in the tests. Have to think in the future what could be a good way to test alternative providers if that would be needed.

Issue #6219, opened 2020-01-08
Adding Transactional Deadlock Detection


jjmason:
# The Problem
When using transactions in production, @jsteinich  and I have encountered deadlocks that are difficult to reproduce and predict.  While the ultimate solution is to order accesses and structure data to prevent these situations altogether,  we would like to be able to do two things to keep our systems stable and push toward that goal:

1. Quickly (~1 sec) abort and retry the request, preferably knowing that it's a deadlock and being able to tune our retry logic appropriately.  Most deadlocks are resolved by retry - if we can do these retries relatively quickly the impact on player experience is minimized.
2. Diagnose the deadlock by examining Silo logs.  Currently we have a transaction id and a request message, which leaves a lot of guesswork in determining the cause of the deadlock.  It would be nice to see which grains are being locked for all transactions involved in the deadlock.

While reducing the transaction break-lock timeout would enable (1), in production we see a small number of transactions take a very long time (> 5 sec).  Aborting and retrying these transactions after a short time interval won't help, and in many cases will make things worse. 

# Our Proposal
We'd like to add an (optional) runtime deadlock detection system to orleans transactions.  This would work something like the following:
* Each lock group has two deadlines - one to check for deadlocks, and one to abort the transaction normally.  When the first deadline is hit, we run the deadlock detector, and abort with a special "deadlock" status if one is found.  The deadlock detection would also log the transactions and  grains that are part of the cycle.
* If deadlock detection is enabled, each ReaderWriterLock sends information about waits and locks to a per silo service.   
* If the deadlock check deadline is reached for a particular lock group, we request a deadlock check for the transactions that are part of that group.  
   * As an optimization, we could first check for deadlocks in the per silo state.
   * If no local deadlocks are found, we send a message containing the suspected transaction(s) to a single grain.  This grain requests relevant lock/wait information for each silo (essentially, relevant means 'reachable from the suspected transaction id, via either locks or waits') from the local tracking services.
  * With this information, we use cycle detection to identify deadlocks and trigger aborts.
 * If deadlock detection is disabled, we never send info to the local tracking service or request deadlock checks, so the overhead in this case is minimal.

Before we start building this, we'd like the Orleans team's take on the proposal in general and advice on a few specific problems.
* The current prototype we're working with uses simple locking to handle concurrency in the per silo service - it might be easier/better to use message passing by working with the orleans runtime.  
[...]

PR #6253, opened 2020-01-21
Deadlock Detection

jjmason:
This PR is intended as a starting point for discussion around adding support for transactional deadlock detection.  It **is not intended to be merged**, or even be a starting point for something that we can merge.   

See [Issue 6219](https://github.com/dotnet/orleans/issues/6219) for discussion.  

The general outline of this proposal is as follows:

 * `ReaderWriterLock` has an optional `ITransactionalResourceObserver`, which is a singleton service that tracks the silo's current locks and waits.  
* `LockGroup` has a `DeadlockDeadline`, similar to `Deadline`.  When this is exceeded while holding or waiting to enter the lock, we send a one-way message to the `DeadlockDetector` grain.  Although it's not really well represented in this PR, we can try to run the deadlock detection locally as well to avoid the overhead of remote calls.
* `DeadlockDetector` uses `IManagementGrain.SendControlCommandToProvider` to request locks related to the offending transactions and grains, which the `SimpleTransactionalLockObserver` finds by performing a dfs over its local state.  Oftentimes we can detect a local cycle as part of this dfs, which saves some work on the part of the `DeadlockDetector`. 
* Once the `DeadlockDetector` has a `WaitForGraph` for all transactions and resources, it uses the normal cycle finding algorithm to locate a cycle that is causing a deadlock.  If one exists, the transactions involved are aborted.

The "tests" included in this proposal don't really test anything, but by examining the logs we can observe that deadlocks are detected and reported correctly.  

No transactions are actually aborted by this version!  I had an implementation that did break locks, but it involved adding a new method to `ITransactionalResource`, which might not be necessary.


KSemenenko:
Randomly I think I found solution https://github.com/managedcode/Orleans.Graph

so we can track calls and check if there is cycles and deadlocks

ReubenBond:
This PR is currently conflicted with main, and maintainers cannot update the head branch because `maintainer_can_modify` is disabled and the authenticated maintainer has no push access to `PerBlue/orleans`.

The refreshed rebased/ported branch is available at `ReubenBond/orleans:pr-6253-deadlock-detection` (`f6a23ef2c7a4be156ac3cfa5a5d53af530375eb8`), based on current `dotnet/orleans` main `24b02591e27c3d5855cef43db24e5d4a3f8c7a6f`.

The six self-review findings were addressed: contention-driven deadline rearming and earliest-deadline arbitration, immutable versioned local snapshots with per-silo replacement rounds, deduplicated transaction-correlated break delivery, isolated event collection, and corrected wait-for-graph coverage. Follow-up reviews found and fixed timeout definiteness and late-worker deadline ordering; the final review reports no significant issues.

`dotnet build Orleans.slnx -bl` succeeds with 0 errors (3 existing `Benchmarks.AdoNet` MSB3277 warnings). Focused deadlock tests pass 15/15 on both `net8.0` and `net10.0`.

The public option XML documentation now states the contention and deadline-order guarantees. No sample or external documentation update is needed because this change refines internal runtime behavior and existing opt-in configuration rather than adding another usage workflow.

Please enable maintainer edits or update the PR head from that branch.

PR #6260, opened 2020-01-24
Expose exception when initializing PerfCounterEnvironmentStatistics

JorgeCandeias:
This exposes the exception thrown when failing to connect to performance counters. This is to help troubleshoot the underlying issue when repairing the counter config does not work. Note that I couldn't find any test class to update. Do you want one? And is there a cleaner solution than this?


ReubenBond:
Thank you, @JorgeCandeias!

PR #6267, opened 2020-01-28
Membership: allow a minimum grace period during ungraceful shutdown

ReubenBond:
Also use `await foreach` instead of our manually coded solution

### DIFF
--- a/src/Orleans.Runtime/MembershipService/ClusterHealthMonitor.cs
+++ b/src/Orleans.Runtime/MembershipService/ClusterHealthMonitor.cs
@@ -109,15 +109,12 @@ namespace Orleans.Runtime.MembershipService
 
         private async Task ProcessMembershipUpdates()
         {
-            IAsyncEnumerator<MembershipTableSnapshot> enumerator = default;
             try
             {
                 if (this.log.IsEnabled(LogLevel.Debug)) this.log.LogDebug("Starting to process membership updates");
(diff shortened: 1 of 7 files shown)


src/Orleans.Runtime/MembershipService/ClusterHealthMonitor.cs
```diff
@@ -346,8 +342,9 @@ Task OnActiveStop(CancellationToken ct)
 
                 this.monitoredSilos = this.monitoredSilos.Clear();
 
-                // Stop waiting for graceful shutdown when the provided cancellation token is cancelled
-                return Task.WhenAny(ct.WhenCancelled(), Task.WhenAll(tasks));
+                // Allow some minimum time for graceful shutdown.
+                var gracePeriod = Task.WhenAll(Task.Delay(TimeSpan.FromSeconds(5)), ct.WhenCancelled());
```
sergeybykov: Should we make the grace period configurable? Or at the very minimum make it a constant as it is used multiple times below?

ReubenBond: Making it configurable may be too much ceremony for too little benefit. I'll make it a constant

ReubenBond: Fixed

PR #6268, opened 2020-01-29
Add support for DynamoDB Pay Per Request Billing Mode

icanhasjonas:
Adds support for `PAY_PER_REQUEST` `BillingMode` for DynamoDB if you set `ReadCapacityUnits` and `WriteCapacityUnits` to `0`


jsteinich:
I wonder if the config should be more explicit (i.e. another property (did that in a custom provider implementation)). At a minimum, it would be good to document the behavior.

icanhasjonas:
I 100% agree. I just didn't want to break the API for now and still get the functionality. I'm not perfectly happy with it tbh and I agree with your reasoning.

sergeybykov:
Thank you, @icanhasjonas!

PR #6269, opened 2020-01-29
Use RegionEndpoint.GetBySystemName() to resolve AWS region

icanhasjonas:
- [x] remove bad `ServiceURL` setter
 - [x] use AWS `RegionEndpoint.GetBySystemName` to resolve regions, but still default to `us-west-2`

Why is `us-west-2` the default btw?

### DIFF
--- a/src/AWS/Shared/AWSUtils.cs
+++ b/src/AWS/Shared/AWSUtils.cs
@@ -1,4 +1,4 @@
-﻿using Amazon;
+using Amazon;
 using System;
 
 #if CLUSTERING_DYNAMODB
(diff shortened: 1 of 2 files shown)


jsteinich:
I believe `ServiceURL` is there to support using something like [LocalStack](https://github.com/localstack/localstack), so we may not want to remove it.

icanhasjonas:
So, yes, but setting Region clears the `ServiceURL` property on the configuration object. You can set one, but not both, and it happens to be that Region is set after ServiceURL and that's the only reason it works. I didn't remove the ServiceURL for the local testing case and I believe this change better aligns the configuration with for example the SQS component, that already removed it.

sergeybykov:
Thank you, @icanhasjonas!

PR #6273, opened 2020-01-30
Best Practices

Zeroshi:
Converted and rewrote some of the Best Practices from the presentation slide. The tutorials will need to be recreated ass well sincxe the dont exist any longer and would likely be outdated.


SebastianStehle:
Cool stuff...What is still unclear for me is the best practices around the number of grains? If you are talking about 100 million IOT devices, is this a number that is acceptable for grains or do we need a grain per device "group"?

Zeroshi:
How would you set up a device group? Are you saying something like thermostat model x group versus thermostat grain?

src/Documentation/resources/BestPractices
```diff
@@ -0,0 +1,422 @@
+# Best Practices
```
ReubenBond: I think this file should be called best_practices.md, or at least end in .md

Zeroshi:
Oh good call. I'll kill it and recommit

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/ghei36>
________________________________

src/Documentation/resources/BestPractices
Zeroshi: Good call!

Zeroshi: done.

Zeroshi:
> 
> 
> Cool stuff...What is still unclear for me is the best practices around the number of grains? If you are talking about 100 million IOT devices, is this a number that is acceptable for grains or do we need a grain per device "group"?

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+    (TAP))
+
+-   [await](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/)
+    is the best syntax to use when composing async operations
+
+-   Common Scenarios:
+
+    -   Return a concrete value:
+
+        -   return Task.FromResult(value);
```
erictuvesson: I think that code should always be kept in code blocks, it makes it easier to read and follow along. 😃

SebastianStehle:
> How would you set up a device group? Are you saying something like thermostat model x group versus thermostat grain?

Either by something from the domain model, e.g. grain per customer, room, region, building...or just a simple module operation to create partitions.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+
+-   Failures can happen at any time and are a normal occurrence
+
+    -   Lost grains are automatically reactivated
+
+    -   Grains that are making in-process calls can fail or timeout
+
+    -   Orleans provides it best effort for grain delivery
+
+    -   Any network message can be lost and should be retired by the application
```
JanEggers: retired  => retried

src/Documentation/resources/Best_Practices.md
Zeroshi: Ah yes! done! thanks!

src/Documentation/resources/Best_Practices.md
Zeroshi: thank you! corrected!

Zeroshi:
> 
> 
> > How would you set up a device group? Are you saying something like thermostat model x group versus thermostat grain?
> 
> Either by something from the domain model, e.g. grain per customer, room, region, building...or just a simple modulo operation to create partitions.

Ah got it. This is based on the grain size in the document. 
"Grain Size
•	Optimal throughput is achieved by using multiple smaller grains rather than a few larger grains. However, the best practice of choosing a grain size and type is base on the application domain model.
–	Example: Users, Orders, etc."

If this doesn't answer your question, please let me know.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
+# Best Practices
+
+Orleans was designed in 2014 with the intention of removing low-level
```
sergeybykov: Orleans has been designed and built much earlier. It was running in production in late 2011. Also, the intent had nothing to do with improving the actor model per se. I think it's better to rephrase the into to something like the following.

"Orleans was built with the goal to greatly simplify building of distributed scalable applications, especially for the cloud. Orleans invented the [Virtual Actor Model](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/Orleans-MSR-TR-2014-41.pdf) as an evolution of the Actor Model optimized for the cloud scenarios.

Grains (virtual actors) are the base building blocks of an Orleans-based application. They encapsulate state and behavior of application entities and maintain their lifecycle, The programming model of Orleans and the characteristics of its runtime fit some types of applications better than others. This document is intended to capture some of the tried and proven application patterns that work well in Orleans."

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+abstractions from actor model frameworks. Grains (actors stored in memory) are
+used to maintain the lifecycle among other features. Orleans is based on the
+actor model and fits some projects better that others.
+
+## Orleans should be considered when:
+
+-   Significant number (hundreds to millions) of loosely coupled entities
```
sergeybykov: I'm not sure how best to reflect the fact that there are two number here really - total number of entities, which can easily be in the billions and trillions, and the subset of that total number that is activated at any point in time.

User profiles are a good example. It is easy to have a grain for every person on Earth within a small cluster, so long as only a subset of users are active at any point in time.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+
+    -   Example: Internet of Things (IoT) – individual devices turning on/off
```
sergeybykov: I would add some other examples: user profiles, purchase orders, application/game sessions, stocks.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+
+-   Entities are small enough to be single-threaded
+
+    -   Example: Determine if stock should be purchased based on current price
+
+-   Workload is interactive
+
+    -   Example: request-response, start/monitor/complete
+
+-   More than one server is expected or may be required\\
```
sergeybykov: Typo - "\\"?

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+
+    -   Orleans runs on a cluster which is expanded by adding servers to expand
+        the cluster
+
+-   Global coordination is not needed or on a smaller scale between a few
+    entities at a time
+
+    -   Orleans maintains its clusters and talks to other clusters. Each cluster
```
sergeybykov: I don't see how this illustrates the topic. Maybe something along the lines of:
"Scalability and performance of execution is achieved by parallelizing and distributed a large number of mostly independent tasks with no single point of synchronization." instead?

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+        is able to communicate through tcp/ip
+
+-   *Various entities are used at different times*
```
sergeybykov: This is debatable. it is entirely possible to keep all entities of certain type in memory, for example, all users or all devices. And for some applications that's the best. Of course, the number of entities must be reasonable in those case. But in general, there's no requirement for the set of active entities to differ over time, even though that's a popular case.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+    -   Each grain maintains its own states and should not be shared.
+
+-   A small number of large entities that may be multithreaded
+
+    -   A microservice may be a better option when supporting complex logic in a
+        single service
+
+-   Global coordination and/or consistency is needed
+
+    -   Orleans maintains the silos within its own framework
```
sergeybykov: Here again the example doesn't seem to fit the topic. Maybe "Such global coordination would severely limit performance of an Orleans-based application."?

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+
+    -   Batch jobs, Single Instruction Multiple Data (SIMD) tasks
+
+    -   This depends on the need of the application and may be a fit for Orleans
+
+## Grains
+
+**Overview**:
+
+-   Actors are not objects, however they are similar
```
sergeybykov: As a general rule, we try to refrain from talking about actors. I think it is better to talk about grains instead.

I would also start with a positive statement. For example, "Grains resemble objects. However, they are distributed, virtual, and asynchronous."

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,422 @@
...
+
+        -   `var tasks = new List\<Task\>();  
+            foreach(var grain in grains)  
+            { tasks.Add(grain.Foo()) }  
+            await Task.WhenAll(tasks);  
+            DoMoreWork();`
+
+**Implementation of Grains**:
+
+-   When to use a [StatelessWorker]
```
sergeybykov: I would start with the first cardinal rule or grain implementation - never perform a thread-blocking operation within grain code. Examples: synchronously waiting for an IO operation or a web service call, taking a lock, running a tight loop waiting for a condition, etc. All operations other than local computations must be explicitly asynchronous.

src/Documentation/resources/Best_Practices.md
Zeroshi: Corrected.

src/Documentation/resources/Best_Practices.md
Zeroshi: Corrected

src/Documentation/resources/Best_Practices.md
Zeroshi: corrected

src/Documentation/resources/Best_Practices.md
Zeroshi: corrected

src/Documentation/resources/Best_Practices.md
Zeroshi: corrected

src/Documentation/resources/Best_Practices.md
Zeroshi: corrected

src/Documentation/resources/Best_Practices.md
Zeroshi: corrected

src/Documentation/resources/Best_Practices.md
Zeroshi: That's a much better way to put it!

Zeroshi: corrected

src/Documentation/resources/Best_Practices.md
Zeroshi: corrected

SebastianStehle:
> > > How would you set up a device group? Are you saying something like thermostat model x group versus thermostat grain?
> > 
> > 
> > Either by something from the domain model, e.g. grain per customer, room, region, building...or just a simple modulo operation to create partitions.
> 
> Ah got it. This is based on the grain size in the document.
> "Grain Size
> • Optimal throughput is achieved by using multiple smaller grains rather than a few larger grains. However, the best practice of choosing a grain size and type is base on the application domain model.
> – Example: Users, Orders, etc."
> 
> If this doesn't answer your question, please let me know.

This is a typical consultancy answer: It depends.. :D

Zeroshi:
> 
> 
> > > > How would you set up a device group? Are you saying something like thermostat model x group versus thermostat grain?
> > > 
> > > 
> > > Either by something from the domain model, e.g. grain per customer, room, region, building...or just a simple modulo operation to create partitions.
> > 
> > 
> > Ah got it. This is based on the grain size in the document.
> > "Grain Size
> > • Optimal throughput is achieved by using multiple smaller grains rather than a few larger grains. However, the best practice of choosing a grain size and type is base on the application domain model.
> > – Example: Users, Orders, etc."
> > If this doesn't answer your question, please let me know.
> 
> This is a typical consultancy answer: It depends.. :D

Ha, well it does depend. If you want to highest throughput then super small grains would be desired. That can create the issue of design, implementation, maintenance, etc. The way I go about it is based on its domain role. For instance, i would make a profile grain that talks to a payer grain. Depending on the complexity, i may break the payer grain into specifics such as debit, mastercard, etc.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   A **timer** is an alternative method to write updates periodically.
+
+        -   The timer allows the application to determine the amount of
+            “eventual consistency”/statelessness allowed.
+
+        -   Timing (immediate/none/minutes) can also be controlled as to when to
+            update.
+
+    -   Each grain class can only be associated with one storage provider.
```
sergeybykov: This part wouldn't be needed for `IPersistentState<TState>` or can be rephrased for it instead.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+-   MemoryStorage (Data stored in memory without durable persistence) is used
+    *only* for debugging and unit testing.
+
+-   AzureTableStorage
+
+    -   Configure the Azure storage account information with an optional
+        DeleteStateOnClear (hard or soft deletions)
+
+    -   Orleans serializer efficiently stores binary data in one Azure table
```
sergeybykov: Binary serializer is not recommended. JSON is a more flexible option.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+        cell
+
+    -   Data size limit == max size of the Azure column which is 64kb of binary
+        data
+
+    -   Community contributed code that extends the use of multiple table
+        columns which increases the overall maximum size to 1mb.
+
+-   SharedStorageProvider will write data to multiple underlying storage
```
sergeybykov: We removed sharded storage provided a while back.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+How to deal with Storage Operation Failures
+
+-   Grains and storage providers can await storage operations and *retry*
+    failures as needed
+
+-   Unhandled failures will propagate back to the caller and will be seen by the
+    client as a broken promise
+
+-   Other than the initial read, there is not a concept that automatically
```
sergeybykov: If grain lets an `InconsistentStateException` escape, it gets automatically deactivated.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+    Clear, ReadStateAsync())
+
+-   The storage behavior is typically configurable (Batch writing, Hard or Soft
+    Deletions, etc.) and defined by the storage provider
+
+## Cluster Management
+
+-   Orleans automatically manages clusters
+
+    -   Worker roles (silo) can fail and join at any time which is handled
```
sergeybykov: I think talking about worker roles (a.k.a Classic Cloud Services) is confusing here. I think "node" is a more general term.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+        automatically
+
+    -   A silo instance table exists for diagnostics
```
sergeybykov: The tables exist first and foremost for the clustering protocol to work. It is also very helpful for diagnostics because it keeps history of all silos in the cluster.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   There are also configuration options of an aggressive or a more lenient
+        failure detection
+
+-   Failures can happen at any time and are a normal occurrence
+
+    -   Lost grains are automatically reactivated
```
sergeybykov: I think this may create a misunderstanding that grain can simply disappear. It is probably better to be more explicit. For example, "In case of a silo failure, grains there were activated on that silo get automatically reactivated later on other silos in the cluster."

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   Grains that are making in-process calls can fail or timeout
```
sergeybykov: I'm not sure I understand what this is referring to.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   Orleans provides it best effort for grain delivery
```
sergeybykov: I would rephrase this as "Message delivery guarantee in Orleans is at-most-once", if that's what you were referring to.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   All of the network messages can be lost and should be retried by the application
```
sergeybykov: I would rephrase this as "It is a responsibility of the caller to retry any failed calls if needed."

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+        if it is crucial
+
+        -   Common practice is to retry from end-to-end from the client/front
+            end
+
+-   Currently there is not a graceful shutdown
```
sergeybykov: Graceful shutdown is supported.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   A Azure upgrade or reboot is treated as a node failure
```
sergeybykov: I would take this one out as it is confusing.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Deployment and Production Management
+
+Service Monitoring
+
+-   Orleans provided information
```
sergeybykov: This is an outdated list.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+-   Create a custom performance counters
+
+Scaling out and in
+
+-   Monitor the Service-Level Agreement (SLA)
+
+-   Add or Remove instances
+
+-   Orleans automatically rebalances and takes advantage of the new hardware
```
sergeybykov: This is not 100% accurate. We don't rebalance activated grains when a new silo is added to the cluster.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Logging and Testing
+
+-   Logging, Tracing, and Monitoring
+
+    -   GrainBase.GetLogger() exposes Info(), Warn(), Error(), Verbose()
+
+    -   .NET tracing and Runtimes trace to a local file by default
```
sergeybykov: This is not true anymore.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+-   Add or Remove instances
+
+-   Orleans automatically rebalances and takes advantage of the new hardware
+
+## Logging and Testing
+
+-   Logging, Tracing, and Monitoring
+
+    -   GrainBase.GetLogger() exposes Info(), Warn(), Error(), Verbose()
```
sergeybykov: The recommended way is to inject logger via DI. For example, like here - https://github.com/dotnet/orleans/blob/master/Samples/3.0/HelloWorld/src/HelloWorld.Grains/HelloGrain.cs#L14.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   .NET tracing and Runtimes trace to a local file by default
+
+    -   Azure Diagnostics is also easily consumed
```
sergeybykov: I would take this out.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   Logger.LogConsumer.Add() allows the developer to choose their own
```
sergeybykov: This is outdated since we switched to leveraging `Microsoft.Extensions.Logging` in 2.0.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+        logging consumer
+
+    -   Default tracing levels can be overridden for grains and runtime
+        components
+
+        -   Example: `<TraceLevelOverrideLogPrefix="Application"
```
sergeybykov: This is all gone since 2.0.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+            TraceLevel=“Verbose">`
+
+        -   Example: `<TraceLevelOverrideLogPrefix=“Runtime"
+            TraceLevel=“Warning">`
+
+Testing
+
+-   UnitTestingBase class is used for unit testing
```
sergeybykov: `UnitTestingBase` is gone.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+-   Starting two or more silos in app domains and a client in the main app
```
sergeybykov: This is outdated since 2.0.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+    domain
+
+-   Simplifies the versions of what is used internally
+
+Troubleshooting
+
+-   Logs (Silo and Frontend logging)
+
+    -   WAD may not pick up the logging in the event of a startup failure
```
sergeybykov: This is Azure specific and outdated. I would take it out.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+    -   Remote Desktop Protocol (RDP) to the machines for verification
```
sergeybykov: This is not the recommended way anymore.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+-   Silo isn’t starting
+
+    -   Check OrleansSiloInstances to determine if the silo registered there.
+
+    -   Make sure that firewall is open for TCP ports: 11111 and 30000
+
+    -   Check the logs and don’t forget that there is an extra log for startup
+        errors
+
+    -   RDP into the server or worker role instance and attempt to start it
```
sergeybykov: I would take this out.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
+# Best Practices
+
+Orleans was built with the goal to greatly simplify building of distributed scalable applications, especially for the cloud. Orleans invented the Virtual Actor Model as an evolution of the Actor Model optimized for the cloud scenarios.
+
+Grains (virtual actors) are the base building blocks of an Orleans-based application. They encapsulate state and behavior of application entities and maintain their lifecycle, The programming model of Orleans and the characteristics of its runtime fit some types of applications better than others. This document is intended to capture some of the tried and proven application patterns that work well in Orleans.
```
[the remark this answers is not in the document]
Zeroshi: Thank you. That does make it much easier to maintain.

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Orleans should be considered when:
+
+-   Significant number (hundreds, millions, billions, and even trillions) of loosely coupled entities. To put the number in perspective, Orleans can easily create a grain for every person on Earth in a small cluster, so long as a subset of that total number is active at any point in time. 
+
+    -   Example: Additonal examples are: user profiles, purchase orders, application/game sessions, stocks
```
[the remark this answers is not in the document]
Zeroshi: that would be due to a lack of sleep. thank you for looking over the document

src/Documentation/resources/Best_Practices.md
Zeroshi: This was based on the best practices presentation. I was assuming that it meant grains may fail or timeout when making calls. I re-wrote it as, "Grains have an ability to timeout. A retry solution such as Polly can assist with retries."

src/Documentation/resources/Best_Practices.md
Zeroshi: Thank you. I asked that one in Gitter because I was pretty sure if was. Thank you for validating.

src/Documentation/resources/Best_Practices.md
Zeroshi: removed for now

src/Documentation/resources/Best_Practices.md
Zeroshi: That's very insightful!

src/Documentation/resources/Best_Practices.md
Zeroshi: Does Dynatrace work?

Zeroshi: removed the .net tracing for the moment

src/Documentation/resources/Best_Practices.md
```diff
@@ -0,0 +1,421 @@
...
+extensible storage functionality.
+
+-   Tutorial: *Needs to be created*
+
+**Overview**:
+
+-   Orleans.IGrainState is extended by a .NET interface which contains fields
+    that should be included in the grain’s persisted state.
+
+-   GrainBase\<T\> is extended by the grain class that adds a strongly typed
```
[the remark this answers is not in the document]
Zeroshi: Thanks for the catch!

src/Documentation/resources/Best_Practices.md
Zeroshi: Updated to, "PersistetState classes, like other grain classes, can only be associated with one storage provider." Thanks again!

sergeybykov:
Thank you, @Zeroshi!

PR #6275, opened 2020-02-02
Support Grain Persistency TTL On dynamo DB

AmirSasson:
Support a TimeToLive option for Grains Persistency on DynamoDB.
Value is provides as a TimeSpan to the the DynamoDBStorageOptions class.
and each "StateWrite" will update the record Attribute.

Table creation process was also updated to enable the TTL feature when needed.


sergeybykov:
Thank you, @AmirSasson!

pipermatt:
Sweet... I was literally coming here searching for this very option. Thanks @AmirSasson !

PR #6285, opened 2020-02-04
Fixing code gen for ValueTask

jsteinich:
Currently `ValueTask<T>` is supported for grain interfaces, but `ValueTask` isn't because it generates a return. This PR just removes the return in that case.


ReubenBond:
Thanks, looks good to me.

PR #6287, opened 2020-02-05
Change TTL Dynamo DB Persistenecy Attribute name, as it is reserved attribute name of DynamoDB

AmirSasson:
Deserializing failed when Grains are saved with TTL attribute which is apparently a reserved dynamo DB word.
(although according to [doc](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-how-to.html) it was seemed fine and best practice.)


sergeybykov:
Thank you, @AmirSasson!

PR #6290, opened 2020-02-08
[WIP] Cherry-picked fixes and improvements for 3.0.3

sergeybykov:


ReubenBond:
Should we include https://github.com/dotnet/orleans/pull/6154

sergeybykov:
It seems to me it would be difficult to include #6154 without enabling multi-targeting for the whole solution. At least, I'm struggling with that. Should we include all of the multi-targeting changes as well as .NET Core perf optimization, and release that as a 3.1.0 instead?

ReubenBond:
Sounds like a fine plan to me

sergeybykov:
I pushed current `master` to 3.1.0 branch. We can build Beta and RC builds from it. I'm not even sure we need a Beta for that. Maybe build it as 3.1.0-rc1 and save the ceremony?

I'll close this PR for now.

PR #6303, opened 2020-02-13
WIP: Logging categories

### LINKED ISSUE #6289: Logging category names are inconsistent
Under Microsoft.Extensions.Logging, it is normal to use a type name complete with namespace, for the logging category.

Most of Orleans does indeed do this, sometimes with fictitious subtype names where it makes sense. That is all good. But there are a surprising number of locations where this is not followed, or has other anomalies.

This includes:
* `RuntimeCodeGen` (No namespace),
* `MembershipOracle` (No Namespace),
* `LeaseBasedQueueBalancer-{name}` (No namespace, use of `-` as separator)
* `AsyncTimer-{name}` (No namespace, use of `-` as separator)
* `SiloHealthMonitor/<SiloAddress>` (no namespace, use of `/` as separator)
* `Orleans.Runtime.ReminderService.GrainBasedReminderTable_<Data.Address>` (use of `_` as seperator)
* `Orleans.Streams.PersistentStreamPullingManager-<streamProviderName>` (use of `-` as separator)
* `Microsoft.Orleans.Networking` (weirdly begins Microsoft.Orleans),
* `Microsoft.Orleans.Messaging.<Subcategories>` (Weirdly begins Microsoft.Orleans).
* `Orleans.RuntimeSiloLogStatistics`/`Orleans.RuntimeClientLogStatistics` (It seems really odd that there is no '.' after Runtime).
* `Orleans.Runtime.Scheduler <SchedulingContext.Name>.WorkItemGroup`, (use of space of seperator. [1])
* Last, but certainly not least, `AsynchAgent` and `TaskSchedulerAgent`, both strip leading "Orleans." off of the type name which ends up getting used by the logger, resulting in categories that begin with `Runtime.`

------

The ones that lack namespaces (or lack the leading `Orleans.`) are obnoxious, as it makes it more difficult to configure the logging level for Orleans internals Separately from the logging level for other parts of the application. I was expected to just be able to configure for `Orleans`, but instead would need `Orleans`, `Runtime`, `Microsoft.Orleans`, and all the class names missing a namespace.

I don't have a problem with separating dynamic data with something other than a simple dot, but currently there is a mixture of `_`, `/`, `-`, space, and sometimes even `.` for that. (Dot is being used to seperate dynamic data more often than any of the others are.)

That said, dynamic values should in logging category names should be used with caution, since using them with an unbounded set of values can effectively be a slow memory leak. See #6096 for details. So dynamic values that have a small fixed set of values, like stream providers, is not a big deal, but including things like grain identity information can be somewhat suspect. Especially concerning is `TransactionalState`, and `TransactionComitter`. 

-----

So what I would like to see happen:
1. Add the namespace to the 5 or so cases that are missing it.
2. Change `AsynchAgent`, and `TaskSchedulerAgent` to use the full namespace as a basis for the logging category.
3. Add the missing dot in `SiloLogStatistics`, and `ClientLogStatistics`, unless that was deliberate.
4. Evaluate usages of dynamic values in category names, and remove them in cases where they are likely to effectively be memory leaks.
5. Consider standardizing the separators used for dynamically determined data in logging category names. (This is not particularly important.)
6. Consider changing the few `Microsoft.Orleans.` categories to just `Orleans.`. (This is definitely not important, and leaving them as-is may make sense, since some people may have these configured specially.)

Footnotes:
[1] Also this actually uses the namespace and type names of the subclass, which is rather odd when the class in question is internal with no subclasses, especially since for non system priority contexts, the hypothetical subclass's name is not used.

KevinCathcart:
Fixes #6289.

The first commit is proposed changes. The second one flags all the dynamic data cases with comments, as some decision still needs to be made for each one.

I've not done anything with the `Microsoft.Orleans.` categories, as I'm not sure if changing them are worth any possible breakage of existing deployment's log filtering schemes.

### DIFF
--- a/src/AWS/Orleans.Persistence.DynamoDB/Provider/DynamoDBGrainStorage.cs
+++ b/src/AWS/Orleans.Persistence.DynamoDB/Provider/DynamoDBGrainStorage.cs
@@ -52,7 +52,7 @@ namespace Orleans.Storage
         {
             this.name = name;
             this.loggerFactory = loggerFactory;
-            var loggerName = $"{typeof(DynamoDBGrainStorage).FullName}.{name}";
+            var loggerName = $"{typeof(DynamoDBGrainStorage).FullName}.{name}"; // dynamic data
             this.logger = loggerFactory.CreateLogger(loggerName);
             this.options = options;
             this.serializationManager = serializationManager;
--- a/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureBlobStorage.cs
+++ b/src/Azure/Orleans.Persistence.AzureStorage/Providers/Storage/AzureBlobStorage.cs
@@ -50,7 +50,7 @@ namespace Orleans.Storage
             this.grainFactory = grainFactory;
             this.typeResolver = typeResolver;
             this.loggerFactory = loggerFactory;
-            this.logger = this.loggerFactory.CreateLogger($"{typeof(AzureTableGrainStorageFactory).FullName}.{name}");
+            this.logger = this.loggerFactory.CreateLogger($"{typeof(AzureTableGrainStorageFactory).FullName}.{name}"); // dynamic data
         }
 
         /// <summary> Read state data function for this storage provider. </summary>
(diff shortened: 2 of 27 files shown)


KevinCathcart:
Arg... Somehow this includes the changes from my other pull request. Yuck. Will rebase it to not include those.

sergeybykov:
@KevinCathcart Is this still WIP?

ReubenBond:
@KevinCathcart your commits are included in #6430, so we can close this. We can bring it back if some changes were missed. Thank you for doing this

PR #6305, opened 2020-02-14
Remove current multicluster implementation

benjaminpetit:
This is a preparation step for the future pluggable directory


ReubenBond:
Should we also remove the UniqueKey.Category GeoClient member?

src/Orleans.Core/Runtime/OutsideRuntimeClient.cs
```diff
@@ -321,24 +319,6 @@ private async Task RunClientMessagePump()
...
-                    clientId = message.TargetGrain;
-                    transport.UpdateClientId(clientId);
-                    CurrentActivationAddress = ActivationAddress.GetAddress(transport.MyAddress, clientId, CurrentActivationAddress.Activation);
-                }
-                else
-                {
-                    clientId = handshakeClientId;
-                }
-            }
-
```
ReubenBond: 🙌 This hack was so ugly

PR #6309, opened 2020-02-17
nuget list formatting and send more messages

aktxyz:
2 simple things in here 

1) reformat the list of nuget packages to make a little easier to read

2) added a loop to the client to send more than one message and have a reconnect option ... I think this makes it a little more fun on the first run :)

### DIFF
--- a/src/Documentation/tutorials_and_samples/tutorial_1.md
+++ b/src/Documentation/tutorials_and_samples/tutorial_1.md
@@ -52,10 +52,19 @@ You will also probably need to add `using` statements.
 
 ## Add Orleans NuGet Packages
 
-1. To the Silo project, add the `Microsoft.Orleans.Server` and `Microsoft.Extensions.Logging.Console` NuGet packages.
-2. To the Client project, add the `Microsoft.Orleans.Client` and `Microsoft.Extensions.Logging.Console` NuGet packages.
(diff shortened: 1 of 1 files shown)


src/Documentation/tutorials_and_samples/tutorial_1.md
```diff
@@ -198,10 +207,20 @@ namespace OrleansBasics
...
-                using (var client = await ConnectClient())
+                var exitRequested = false;
+                while (!exitRequested)
                 {
-                    await DoClientWork(client);
-                    Console.ReadKey();
+                    using (var client = await ConnectClient())
+                    {
+                        var reconnectRequested = false;
+                        while (!reconnectRequested)
```
ReubenBond: This disconnect/reconnect functionality is rather unconventional. I would never expect that kind of behavior in a real-world application, or any behavior which involves disposing and recreating the client.

aktxyz: agreed ... but then a sample application is not real-world either ... I just thought it made more fun than send one message and done

sergeybykov: I think this is likely to show an anti-pattern - of disposing a cluster client, even if that's not the intent of the sample. The simple and straightforward approach we recommend is to initialize a cluster client once, and then hold on to it for the life of the app.

Issue #6310, opened 2020-02-18
Ability to dump statistics on demand


OracPrime:
I'd quite like to be able to dump statistics at particular times, rather than just wait for a timer to go off.
I'm happy to submit a PR, but wanted to float my design here and see if it meets approval.

The statistics are done by ClientStatisticsManager and SiloStatisticsManager, both of which are internal classes with no public interface.  They are each registered as singletons inthe service container.

I propose defining an IStatisticsManager interface with a single Dump method on it.  Because the context for registering the singletons is different, I believe I can make both ClientStatisticsManager and SiloStatisticsManager implement IStatisticsManager.

Users wishing to use the manual dump would then use code like this

      ISiloHostBuilder builder = new SiloHostBuilder()
                                                         .NormalConfigCalls();
      var host = builder.Build();
      var stats = host.Services.GetService<IStatisticsManager>();

and then stats.Dump() when required.

Does that sound reasonable/correct?

PR #6311, opened 2020-02-18
Manual stats dump #6310

OracPrime:
As per #6310, allow users to dump statistics on demand, rather than just on a timer.


ReubenBond:
I'd prefer a name more specific to your use case - eg, IStatisticsDumper or IDumpLogStatistics. The interface doesn't currently have any methods related to management.

Be aware that we may remove these statistics entirely in future and replace them with .NET Core `EventCounters` - likely with purpose-built interfaces for you to intercept them and do with them what you wish

OracPrime:
I can push another commit with IStatisticsDumper if you like? I'm not attached to the name; just thought that it was the public face of the managers.  The fact that the only public management is controlling when things are dumped is just the current state.

ReubenBond:
This is fine, I'll merge when tests complete

sergeybykov:
Thank you, @OracPrime!

Issue #6276, opened 2020-02-02
Autofac compatibilty issue


alfeg:
Hello, I have a issue regarding compatibility with `Autofac` and `GatewayPort = 0`

I cannot figure out how to configure it to play nice with Orleans.
Here is a minimal repro on sample project https://github.com/dotnet/orleans/compare/master...alfeg:autofac-compatilibity?expand=1
exception is  
```
DependencyResolutionException: An exception was thrown while activating λ:Orleans.ILifecycleParticipant`1[[Orleans.Runtime.ISiloLifecycle, Orleans.Runtime.Abstractions, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null]][] -> Orleans.Runtime.Messaging.GatewayConnectionListener -> λ:Orleans.Runtime.Messaging.Gateway.
...
DependencyResolutionException: A delegate registered to create instances of 'Orleans.Runtime.Messaging.Gateway' returned null.
```
at https://github.com/dotnet/orleans/blob/master/src/Orleans.Runtime/Silo/Silo.cs#L243

Issues is seems to be in registration that return `null`
https://github.com/dotnet/orleans/blob/master/src/Orleans.Runtime/Hosting/DefaultSiloServices.cs#L128 Autofac doesn't like it at all

Is there a sane way to disable Gateway with Autofac?

alfeg:
Found a temporary workaround. Remove registration of `GatewayConnectionListener`

```csharp
.UseOrleans(...)
.ConfigureServices(services =>
{
    var gateWayListener = services.FirstOrDefault(s => s.ImplementationType?.Name == "GatewayConnectionListener");
    services.Remove(gateWayListener);
})
```

benjaminpetit:
You are right, Autofac doesn't like `null`. The line causing the issue, as you pointed:

``csharp
services.AddSingleton<Gateway>(sp => sp.GetRequiredService<MessageCenter>().Gateway);
``

I think we can remove this and use directly `MessageCenter.Gateway` inside the constructor of `GatewayConnectionListener`. It's not ideal, but our registration of the `Gateway` is not really great either.

Trying to validate this solution and will open a PR with the workaround. Thanks for raising this issue!

alfeg:
@benjaminpetit Issues seems to be fixed in latest release and no longer reproducible.

PR #6312, opened 2020-02-19
Avoid registering Gateway in DI since it can be null

benjaminpetit:
Workaround for #6276


benjaminpetit:
(please don't merge until I managed to add a unit test)

PR #6320, opened 2020-02-20
Fix SQL Server connection string

veikkoeeva:
The new .NET Core SQL Server connector library has made
has removed the obsolote "asynchronous processing = true"
attribute and the tests crash to exception that tells this
keyword is unsupported. Removing it from the connection string
allows the tests run green again.


sergeybykov:
Thank you, @veikkoeeva!

PR #6327, opened 2020-02-22
[AWS:DynamoDB] Don't set `ServiceUrl` when region is provided.

icanhasjonas:
For reference, setting `RegionEndpoint` overwrites the `ServiceUrl` setting anyway,
    and the only reason it works now is because of the order the
    properties are set.


sergeybykov:
Thank you, @icanhasjonas!

PR #6328, opened 2020-02-22
[AWS:DynamoDB] Explicit setting for `UseProvisionedThroughput`

icanhasjonas:
This fixes a previous issue where a Read/Write capacity of `0` would imply a non-provisioned strategy. Setting `UseProvisionedThroughput` to `false` now changes the `BILLING_MODE` to `PAY_PER_REQUEST` as expected.

Defaults are kept (`UseProvisionedThroughput = true`) to not change or break existing behavior.


sergeybykov:
Thank you, @icanhasjonas!

PR #6331, opened 2020-02-22
Fix deadlocks with MySQL Storage

### LINKED ISSUE #6316: Deadlocks when using MySql with Persistence.AdoNet
I had to migrate one project from Postgres to MySql (MariaDB more precisely) and I had a lot of deadlock problems when massively persisting grain states

The issue is caused by Gap Locks (https://dev.mysql.com/doc/refman/8.0/en/innodb-locking.html#innodb-gap-locks)

Changing the transaction's isolation level from `REPEATABLE READ` (default in MySql) to `READ COMMITTED` in `WriteToStorage` procedure fixes the problem.

I don't see any advantage to use `REPEATABLE READ` in this context @veikkoeeva any thoughts?

If this solution is ok for you, I'd be happy to make a PR

srollinet:
Set isolation level to `READ COMMITED` to avoid Gap Lock issues
Fixes #6316

### DIFF
--- a/src/AdoNet/Orleans.Persistence.AdoNet/MySQL-Persistence.sql
+++ b/src/AdoNet/Orleans.Persistence.AdoNet/MySQL-Persistence.sql
@@ -111,6 +111,8 @@ BEGIN
 
     SET _newGrainStateVersion = _GrainStateVersion;
 
+    -- Default level is REPEATABLE READ and may cause Gap Lock issues
+    SET TRANSACTION ISOLATION LEVEL READ COMMITTED;
     START TRANSACTION;
     UPDATE Storage
     SET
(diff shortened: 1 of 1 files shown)


srollinet:
@veikkoeeva let me know if you prefer to define the isolation level globally, rather to change it only in the procedures

veikkoeeva:
@srollinet This seems to be the better option.

To be explicit, my reasoning is as follows: I believe in MySQL the global options requires super priviledges, a restart and would apply to all schemas. It's a least an inconvenience and in enterprise setting likely not possible to pull off. I'm thinking especially like deparment X considering Orleans and then needing to abandon Orleans due to restrictions like this.

sergeybykov:
Thank you, @srollinet!

PR #6336, opened 2020-02-23
AdoNet: Rename Storage table to OrleansStorage for consistency with other tables.

alexrp:
This table is the odd one out:

```
test=# \d
                        List of relations
 Schema |             Name              | Type  | Owner
--------+-------------------------------+-------+-------
 public | orleansmembershiptable        | table | test
 public | orleansmembershipversiontable | table | test
 public | orleansquery                  | table | test
 public | orleansreminderstable         | table | test
 public | storage                       | table | test

```

This shouldn't affect existing deployments, I think, since the queries that access the table are all written into the `OrleansQuery` table in these scripts in the first place.


alexrp:
(Note: This might create a merge conflict with #6331.)

sergeybykov:
Thank you, @alexrp!

PR #6340, opened 2020-02-25
Basic support for pluggable grain directory

benjaminpetit:
Out of scope for this PR;

- Per grain type directory (will come in another PR)
- CosmosDB implementation

### DIFF
--- a/src/Orleans.Core/GrainDirectory/IGrainLocator.cs
+++ b/src/Orleans.Core/GrainDirectory/IGrainLocator.cs
@@ -1,0 +1,24 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Orleans.Runtime;
+
+namespace Orleans.GrainDirectory
+{
+    /// <summary>
+    /// Used to locate Grain activation in the cluster
+    /// </summary>
+    interface IGrainLocator
+    {
+        Task<ActivationAddress> Register(ActivationAddress address);
+
+        Task Unregister(ActivationAddress address, UnregistrationCause cause);
+
+        Task UnregisterMany(List<ActivationAddress> addresses, UnregistrationCause cause);
+
+        Task<List<ActivationAddress>> Lookup(GrainId grainId);
+
+        bool TryLocalLookup(GrainId grainId, out List<ActivationAddress> addresses);
+    }
+}
--- a/src/Orleans.Runtime/Core/Dispatcher.cs
+++ b/src/Orleans.Runtime/Core/Dispatcher.cs
@@ -168,7 +168,9 @@ namespace Orleans.Runtime
                             {
                                 try
                                 {
-                                    logger.Trace("UnregisterAfterNonexistingActivation addr={0} origin={1}", nonExistentActivation, origin);
+                                    if (this.logger.IsEnabled(LogLevel.Trace))
+                                        logger.Trace("UnregisterAfterNonexistingActivation addr={ActivationAddress} origin={SiloAddress}", nonExistentActivation, origin);
+
                                     await this.grainLocator.Unregister(nonExistentActivation, UnregistrationCause.NonexistentActivation);
                                 }
                                 catch (Exception exc)
--- a/src/Orleans.Runtime/Core/Dispatcher.cs
+++ b/src/Orleans.Runtime/Core/Dispatcher.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Orleans.CodeGeneration;
 using Orleans.Configuration;
+using Orleans.GrainDirectory;
 using Orleans.Runtime.GrainDirectory;
 using Orleans.Runtime.Messaging;
 using Orleans.Runtime.Placement;
@@ -26,6 +27,7 @@ namespace Orleans.Runtime
         private readonly SiloMessagingOptions messagingOptions;
         private readonly PlacementDirectorsManager placementDirectorsManager;
         private readonly ILocalGrainDirectory localGrainDirectory;
+        private readonly IGrainLocator grainLocator;
         private readonly ActivationCollector activationCollector;
         private readonly MessageFactory messageFactory;
         private readonly CompatibilityDirectorManager compatibilityDirectorManager;
@@ -40,6 +42,7 @@ namespace Orleans.Runtime
             IOptions<SiloMessagingOptions> messagingOptions,
             PlacementDirectorsManager placementDirectorsManager,
             ILocalGrainDirectory localGrainDirectory,
+            IGrainLocator grainLocator,
             ActivationCollector activationCollector,
             MessageFactory messageFactory,
             CompatibilityDirectorManager compatibilityDirectorManager,
@@ -54,6 +57,7 @@ namespace Orleans.Runtime
             this.invokeWorkItemLogger = loggerFactory.CreateLogger<InvokeWorkItem>();
             this.placementDirectorsManager = placementDirectorsManager;
             this.localGrainDirectory = localGrainDirectory;
+            this.grainLocator = grainLocator;
             this.activationCollector = activationCollector;
             this.messageFactory = messageFactory;
             this.compatibilityDirectorManager = compatibilityDirectorManager;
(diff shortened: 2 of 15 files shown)


src/Orleans.Core/GrainDirectory/IGrainLocator.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        Task<ActivationAddress> Register(ActivationAddress address);
+
+        Task Unregister(ActivationAddress address, UnregistrationCause cause);
+
+        Task UnregisterMany(List<ActivationAddress> addresses, UnregistrationCause cause);
+
+        Task<List<ActivationAddress>> Lookup(GrainId grainId);
+
+        bool TryLocalLookup(GrainId grainId, out List<ActivationAddress> addresses);
```
ReubenBond: Can this and Lookup not be merged into one `ValueTask<List<ActivationAddress>>` method?

src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -164,8 +168,8 @@ public void ReceiveMessage(Message message)
                             {
                                 try
                                 {
-                                    await this.localGrainDirectory.UnregisterAfterNonexistingActivation(
-                                        nonExistentActivation, origin);
+                                    logger.Trace("UnregisterAfterNonexistingActivation addr={0} origin={1}", nonExistentActivation, origin);
```
ReubenBond: This trace message needs a guard

ReubenBond: Also, prefer to use LogTrace instead of our string.format-based shim

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Orleans.Configuration;
+using Orleans.GrainDirectory;
+
+namespace Orleans.Runtime.GrainDirectory
+{
+    /// <summary>
+    /// Implementation of <see cref="IGrainLocator"/> that uses an <see cref="IGrainDirectory"/> store.
+    /// </summary>
+    internal class GrainLocator : IGrainLocator
```
ReubenBond: nit: naming? Maybe DefaultGrainLocator or DirectoryGrainLocator?

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        public async Task<ActivationAddress> Register(ActivationAddress address)
+        {
+            if (address.Grain.IsClient)
+                return await this.inClusterGrainLocator.Register(address);
+
+            var result = await this.grainDirectory.Register(ConvertToGrainAddress(address));
+            var activationAddress = ConvertToActivationAddress(result);
+            this.cache.AddOrUpdate(
```
ReubenBond: Do client addresses need to be cached, too?

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -0,0 +1,120 @@
...
+                activationAddress.Grain,
+                new List<Tuple<SiloAddress,ActivationId>>() { Tuple.Create(activationAddress.Silo, activationAddress.Activation) },
```
ReubenBond: nit: in a future PR, we should replace this Tuple with either a ValueTuple or a purpose-built struct

ReubenBond: Benefit of a ValueTuple is automatic GetHashCode/Equality, etc, if that matters.

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -0,0 +1,120 @@
...
+        public bool TryLocalLookup(GrainId grainId, out List<ActivationAddress> addresses)
+        {
+            if (grainId.IsClient)
+                return this.inClusterGrainLocator.TryLocalLookup(grainId, out addresses);
+
+            if (this.cache.LookUp(grainId, out var results))
+            {
+                addresses = results
+                    .Select(tuple => ActivationAddress.GetAddress(tuple.Item1, grainId, tuple.Item2))
+                    .ToList();
```
ReubenBond: This is clean but not cheap

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            addresses = null;
+            return false;
+        }
+
+        public async Task Unregister(ActivationAddress address, UnregistrationCause cause)
+        {
+            if (address.Grain.IsClient)
+                await this.inClusterGrainLocator.Unregister(address, cause);
```
ReubenBond: We aren't returning after this, so we're going to try unregistering the client from two directories, even though we only registered it in one.

Might be worthwhile enforcing braces around conditional/loop bodies. It's one of the things which Go enforces even though they are light on syntax in other places, because omitting them can make errors more subtle.

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -0,0 +1,120 @@
...
+            await this.grainDirectory.Unregister(ConvertToGrainAddress(address));
+            this.cache.Remove(address.Grain);
+        }
+
+        public async Task UnregisterMany(List<ActivationAddress> addresses, UnregistrationCause cause)
+        {
+            var tasks = addresses.Select(addr => Unregister(addr, cause)).ToList();
+            await Task.WhenAll(tasks);
+            foreach (var addr in addresses)
+                this.cache.Remove(addr.Grain);
```
ReubenBond: Clients are currently never cached, but they will be removed here. Also, this seems to remove grains from cache twice, since Unregister already calls cache.Remove

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                return ActivationAddress.GetAddress(
+                        SiloAddress.FromParsableString(addr.SiloAddress),
+                        GrainId.FromParsableString(addr.GrainId),
+                        ActivationId.GetActivationId(UniqueKey.Parse(addr.ActivationId.AsSpan())));
+            }
+            catch (Exception)
+            {
+
+                throw;
```
ReubenBond: remove try/catch? I realize we intend to remove much of this code over time anyhow.

src/Orleans.Runtime/Core/Dispatcher.cs
benjaminpetit: I just restored a log line that was wrongly deleted in the multi-cluster PR. I will update it

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
benjaminpetit: Bad copy-paste on my part while trying to split this into multiple PR. Other method just return from this if statement

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
benjaminpetit: They are cached in the `ILocalGrainDirectory`  implementation

src/Orleans.Core/GrainDirectory/IGrainLocator.cs
benjaminpetit: Not "easily" at least, since in `Catalog` we have two explicit different method: `Catalog.FastLookup` and `Catalog.FullLookup`. The synchronous method is used by the `Dispatcher` when trying to address a message synchronously.

We could definitely change that also to simplify the flow.

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
benjaminpetit: Forgot a return here. Client are cached in the `LocalGrainDirectory` implementation

sergeybykov:
Ah, I wanted to suggest some minor changes. I guess I can submit a separate PR for that.

ReubenBond:
I think subsequent PRs are the fastest way to evolve this

RehanSaeed:
Any docs for this?

ReubenBond:
@RehanSaeed see #6572

PR #6344, opened 2020-02-25
Fix CleanupDefunctSiloMembership & MembershipTableTests

pipermatt:
So I got DynamoDB running locally in a Docker container so that I could run the AWSUtils.Tests and verify my change.

It seems that for DynamoDB-related things, these tests are mostly broken. The constructors for DynamoDBGrainStorage and SiloLifecycleSubject have changed, so the attempts to instantiate these fail in DynamoDBStorageProviderTests.cs. I fixed this, but then also found that the maximum size DynamoDB allows for an item is 400KB and some of those tests are trying to write items considerably larger than that. 

Other AWSUtils.Tests are failing too. I may take a stab at fixing these later in the week, but this is what I had time for today. The Membership ones pass, and that's the code I was changing.


sergeybykov:
Thank you, @pipermatt!

PR #6347, opened 2020-02-25
CodeGenerator fixes

pentp:
We're in the process of upgrading to Orleans 3.x and trying out the new code generator (again).
I discovered several bugs and regressions compared to the legacy code-gen:
* Fix `ValueTask` type loading to be optional (for netstandard 2.0).
* Fix `ImmutableAttribute` to actually work.
* Don't generate (invalid) serializer types for enums.
* Don't record invalid (empty) copies of value types using `RecordCopy`.
Theoretically this might result in observable changes in behavior, but the current logic is invalid anyway.
Probably should remove `RecordCopy` calls from `BuiltInTypes` also for value types (but the copies are at least valid objects there).
`ILSerializerGenerator.EmitCopier` has the same bug - it boxes and records invalid (empty) objects for value types.
* For shallow copyable types don't create a new object in DeepCopier. In case of value types a new box is still created, but no field-by-field copying is done.
* Mark `enum, Type, nullable, Tuple, ValueTuple` types shallow copyable when suitable.

### DIFF
--- a/src/Orleans.CodeGenerator/Generators/SerializerGenerator.cs
+++ b/src/Orleans.CodeGenerator/Generators/SerializerGenerator.cs
@@ -5,13 +5,12 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
-using System.Threading;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.Extensions.Logging;
 using Orleans.CodeGenerator.Model;
 using Orleans.CodeGenerator.Utilities;
-using Microsoft.Extensions.Logging;
 using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
 using ITypeSymbol = Microsoft.CodeAnalysis.ITypeSymbol;
 
@@ -98,7 +97,7 @@ namespace Orleans.CodeGenerator.Generators
                         break;
                 }
             }
-            
+
             return CodeGenerator.ToolName + b;
         }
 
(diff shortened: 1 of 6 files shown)


src/Orleans.CodeGenerator/Generators/SerializerGenerator.cs
```diff
@@ -346,11 +345,10 @@ private MemberDeclarationSyntax GenerateSerializerMethod(INamedTypeSymbol type,
         private MemberDeclarationSyntax GenerateDeepCopierMethod(INamedTypeSymbol type, List<FieldInfoMember> fields, SemanticModel model)
         {
             var originalVariable = IdentifierName("original");
-            var inputVariable = IdentifierName("input");
-            var resultVariable = IdentifierName("result");
 
             var body = new List<StatementSyntax>();
-            if (type.HasInterface(wellKnownTypes.ImmutableAttribute))
+            if (type.HasAttribute(wellKnownTypes.ImmutableAttribute)
+                || SymbolEqualityComparer.Default.Equals(wellKnownTypes.Immutable_1, type))
```
ReubenBond: Does SymbolEqualityComparer.Default.Equals return true for comparisons between `Immutable<>` and `Immutable<MyClass>`?

pentp: This is just a special case for the `Immutable<T>` type itself that gets generated for `Orleans.Core`. Added it just in case that ends up actually used for deep copying (though it shouldn't).

pentp: Maybe it would be better to actually skip emitting that `Immutable<T>` serializer altogether as it should always us the built-in serializer?

ReubenBond: We should consider adding a `[SkipCodeGeneration]` attribute, or something more specific and add it to `Immutable<T>`

ReubenBond:
Thank you, @pentp - it is very much appreciated!

PR #6352, opened 2020-02-26
Improve TLS troubleshooting experience

### LINKED ISSUE #6317: Improve troubleshooting experience for TLS
Currently, authentication errors which occur during TLS negotiation are not clearly visible in logs.

We should improve the experience to assist users in troubleshooting configuration issues.

Eg, recently an internal user hit a permissions issue getting the private key for a certificate. The logs didn't expose the underlying exception, so debugging it wasn't as simple as it should have been.

ReubenBond:
Fixes #6317

### DIFF
--- a/src/Orleans.Connections.Security/Hosting/HostingExtensions.IClientBuilder.cs
+++ b/src/Orleans.Connections.Security/Hosting/HostingExtensions.IClientBuilder.cs
@@ -60,7 +60,7 @@ namespace Orleans
 
             if (!certificate.HasPrivateKey)
             {
-                throw new ArgumentException($"Certificate {certificate.ToString(verbose: true)} does not contain a private key", nameof(certificate));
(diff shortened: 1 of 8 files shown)


src/Orleans.Connections.Security/Hosting/HostingExtensions.IClientBuilder.cs
```diff
@@ -48,16 +48,21 @@ public static class OrleansConnectionSecurityHostingExtensions
...
                 throw new ArgumentNullException(nameof(certificate));
             }
 
-            if (configureOptions == null)
+            if (configureOptions is null)
             {
                 throw new ArgumentNullException(nameof(configureOptions));
             }
 
+            if (!certificate.HasPrivateKey)
```
sergeybykov: It's better to put the check and throw into a helper method to avoid code duplication.

ReubenBond: Thanks, done

PR #6354, opened 2020-02-26
Add UnregisterMany to IGrainDirectory to support batch unregistrations 

sergeybykov:
Add UnregisterMany to IGrainDirectory to support batch unregistrations.
Rename InCluster to Dht.

This is a followup to #6340.

### DIFF
--- a/src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
+++ b/src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
@@ -16,10 +16,10 @@ namespace Orleans.Runtime.GrainDirectory
     internal class GrainLocator : IGrainLocator
     {
         private readonly IGrainDirectory grainDirectory;
-        private readonly InClusterGrainLocator inClusterGrainLocator;
+        private readonly DhtGrainLocator inClusterGrainLocator;
         private readonly IGrainDirectoryCache cache;
 
-        public GrainLocator(IGrainDirectory grainDirectory, InClusterGrainLocator inClusterGrainLocator)
+        public GrainLocator(IGrainDirectory grainDirectory, DhtGrainLocator inClusterGrainLocator)
         {
             this.grainDirectory = grainDirectory;
             this.inClusterGrainLocator = inClusterGrainLocator;
--- a/src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
+++ b/src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
@@ -76,10 +76,13 @@ namespace Orleans.Runtime.GrainDirectory
             return false;
         }
 
-        public Task UnregisterMany(List<ActivationAddress> addresses, UnregistrationCause cause)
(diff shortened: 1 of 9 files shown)


ReubenBond:
Will there be a subsequent PR to consume the functionality? (I can't see any usage here)

Another option could be to have an optional interface, like `IBulkGrainDirectory` which can have these extra *Many methods, but this LGTM, too.

sergeybykov:
`UnregisterMany` is used by `Catalog` - https://github.com/dotnet/orleans/blob/master/src/Orleans.Runtime/Catalog/Catalog.cs#L1007. This change only makes that code path more efficient in https://github.com/dotnet/orleans/pull/6354/files#diff-2ce13868a2e1d9bd6db970b4c8a618f3L93.

I don't think we need `IBulkGrainDirectory` because I see the real usage asymmetric: grains are registered one at a time, but might get unregistered in batches upon a silo failure, etc. SO, I believe `UnregisterMany` is the only batch operation we need.

ReubenBond:
Ah, I see how I misread the code. Got it.

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -75,23 +76,16 @@ public bool TryLocalLookup(GrainId grainId, out List<ActivationAddress> addresse
...
-            {
-                await this.inClusterGrainLocator.Unregister(address, cause);
-            }
-            else
-            {
-                await this.grainDirectory.Unregister(ConvertToGrainAddress(address));
-                this.cache.Remove(address.Grain);
-            }
+            var grainAddresses = addresses.Select(addr => ConvertToGrainAddress(addr)).ToList();
+            return this.grainDirectory.UnregisterMany(grainAddresses);
```
benjaminpetit: We should also remove these addresses from the cache

src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -76,10 +76,13 @@ public bool TryLocalLookup(GrainId grainId, out List<ActivationAddress> addresse
...
 
-        public Task UnregisterMany(List<ActivationAddress> addresses, UnregistrationCause cause)
+        public async Task UnregisterMany(List<ActivationAddress> addresses, UnregistrationCause cause)
         {
             var grainAddresses = addresses.Select(addr => ConvertToGrainAddress(addr)).ToList();
-            return this.grainDirectory.UnregisterMany(grainAddresses);
+            await this.grainDirectory.UnregisterMany(grainAddresses);
+
+            foreach (var address in addresses)
+                this.cache.Remove(address.Grain);
```
ReubenBond: If UnregisterMany throws, the cache wont be updated - but some or all of the registrations may have been unregistered.

We could choose to ignore this because cache will be cleaned up eventually by other means, or we could put the foreach into a finally block

sergeybykov: Thanks. Did that and also updates the single version of `Unregister` the same way.

PR #6360, opened 2020-02-27
Expand documentation on client connectivity and dependency injection.

### LINKED ISSUE #6359: Is a ClusterClient expected to recover from intermittent connection issues?
I've seen cohosting and dependency injection examples where people are injecting a singleton `IClusterClient` at startup (which is also what I'm currently doing), but looking through the documentation, I couldn't actually find out whether the client is expected to eventually recover from intermittent connection issues.

This is of course an important guarantee to have since, if the client is *not* guaranteed to recover and one is using a singleton instance, the application would need to be restarted to obtain a working client instance again.

This seems like an obvious enough issue that I would imagine this guarantee is actually there, but I would rest easier knowing for sure whether that's the case.

alexrp:
Closes #6359.

### DIFF
--- a/src/Documentation/clusters_and_clients/index.md
+++ b/src/Documentation/clusters_and_clients/index.md
@@ -171,6 +171,7 @@ public class Program
                 services.AddSingleton<ClusterClientHostedService>();
                 services.AddSingleton<IHostedService>(sp => sp.GetService<ClusterClientHostedService>());
                 services.AddSingleton<IClusterClient>(sp => sp.GetService<ClusterClientHostedService>().Client);
+                services.AddSingleton<IGrainFactory>(sp => sp.GetService<ClusterClientHostedService>().Client);
             })
             .ConfigureLogging(builder => builder.AddConsole())
             .RunConsoleAsync();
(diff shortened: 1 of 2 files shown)


alexrp:
cc @ReubenBond

src/Documentation/clusters_and_clients/index.md
```diff
@@ -86,6 +86,117 @@ So it is a responsibility of the application code to build a higher level reliab
...
+public class Program
+{
+    static Task Main()
+    {
+        return new HostBuilder()
+            .ConfigureServices(services =>
+            {
+                services.AddSingleton<ClusterClientHostedService>();
+                services.AddSingleton<IHostedService>(sp => sp.GetService<ClusterClientHostedService>());
+                services.AddSingleton<IClusterClient>(sp => sp.GetService<ClusterClientHostedService>().Client);
```
ReubenBond: We should add IGrainFactory here, too

```suggestion
                services.AddSingleton<IClusterClient>(sp => sp.GetService<ClusterClientHostedService>().Client);
                services.AddSingleton<IGrainFactory>(sp => sp.GetService<ClusterClientHostedService>().Client);
```

alexrp: This does make me wonder when you would actually want to use `IGrainFactory` specifically. `IClusterClient` exposes more functionality and seems to always be available when an `IGrainFactory` is, so is there actually a reason one might want to get an `IGrainFactory` specifically?

No objection to the suggested change either way, though.

ReubenBond: Yes, `IGrainFactory` is usually the preferred choice. The main difference, as far as application developers are concerned, is that `IClusterClient` has a convenience method for getting stream providers.

PR #6362, opened 2020-02-27
Remove unnecessary provider runtime members

ReubenBond:
Minor cleanup


ReubenBond:
One unrelated test failure (socket collision)

PR #6365, opened 2020-02-28
Reorient RuntimeContext around IGrainContext

ReubenBond:
`IGrainContext` is a new interface, based on `IActivationData` and is intended to be a unified representation of both Grains and SystemTargets from the perspective of the runtime. Therefore, the runtime will interact with this type to interact with the grain (activate, send messages, deactivate, etc)

This PR preparation for aligning SystemTargets & Grains (goal being less internal special-casing) and for allowing SystemTargets/Grains to make their own scheduling decisions, by moving scheduling into SystemTarget/ActivationData instead of outside.

Also removes SchedulingContext & ISchedulingContext


src/Orleans.Runtime/Core/InsideRuntimeClient.cs
```diff
@@ -669,16 +650,7 @@ public void ReceiveResponse(Message message)
...
-        {
-            get
-            {
-                if (RuntimeContext.Current == null) return this.HostedClient.ToString();
-
-                var currentActivation = GetCurrentGrainContext();
-                return currentActivation.Address.ToString();
-            }
-        }
+        public string CurrentActivationIdentity => RuntimeContext.CurrentGrainContext?.Address.ToString() ?? this.HostedClient.ToString();
```
sergeybykov: Shouldn't the value be caches?

src/Orleans.Runtime/Core/SystemTarget.cs
```diff
@@ -51,6 +48,18 @@ internal ExtensionInvoker ExtensionInvoker
 
         IGrainReferenceRuntime ISystemTargetBase.GrainReferenceRuntime => this.RuntimeClient.GrainReferenceRuntime;
 
+        GrainReference IGrainContext.GrainReference => GrainReference.FromGrainId(this.grainId, this.RuntimeClient.GrainReferenceRuntime, systemTargetSilo: this.Silo);
```
sergeybykov: Shouldn't we cache the value instead of allocating a new one for every call?

src/Orleans.Runtime/Core/SystemTarget.cs
```diff
@@ -51,6 +48,18 @@ internal ExtensionInvoker ExtensionInvoker
...
+
+        GrainId IGrainContext.GrainId => this.grainId;
+
+        IAddressable IGrainContext.GrainInstance => this;
+
+        ActivationId IGrainContext.ActivationId => this.ActivationId;
+
+        ActivationAddress IGrainContext.Address => ActivationAddress.GetAddress(this.Silo, this.grainId, this.ActivationId);
```
sergeybykov: Same here - allocate on every call?

src/Orleans.Runtime/Utilities/SiloUnobservedExceptionsHandler.cs
```diff
@@ -51,19 +51,17 @@ private void InternalUnobservedTaskExceptionHandler(object sender, UnobservedTas
...
             var baseException = aggrException.GetBaseException();
-            var tplTask = (Task)sender;
-            var contextObj = tplTask.AsyncState;
-            var context = contextObj as ISchedulingContext;
+            var context = (sender as Task)?.AsyncState;
 
             try
             {
-                UnobservedExceptionHandler(context, baseException);
+                this.logger.LogError((int)ErrorCode.Runtime_Error_100104, baseException, "Silo caught an unobserved exception thrown from context {Context}: {Exception}", context, baseException);
```
sergeybykov: Missing $ for string interpolation?

ReubenBond: This is a logger template string, not interpolated - the variables Context & Exception will appear with their positional counterparts in logging systems (eg, columns in geneva or seq)

src/Orleans.Runtime/Core/InsideRuntimeClient.cs
ReubenBond: I think it's not worth it - this is only used in rare logs.

ReubenBond:
Unrelated failure

Arshia001:
I have no idea how, but since updating from 3.0.0 to 3.1.1, a piece of code I had running in production for three years doesn't run anymore, complaining that `System.InvalidOperationException: 'Cannot create a local object reference from a grain.'
`, even though I'm perfectly sure the class isn't a grain. Having read through the source and viewed the history, I think this change is to blame somehow?

I'm trying to create an observer in the `StartAsync` of a hosted service. I'm 250% sure it's not a grain.

src/Orleans.Runtime/Utilities/SiloUnobservedExceptionsHandler.cs
SebastianStehle: Please do not put the exceptions into  the log message. The ILogger interface has a dedicated parameter for exceptions, there is no need to log them twice. I have issues with GCE with that because Google Stackdriver analyzes the logs and sometimes I get exceptions twice then.

ReubenBond: @SebastianStehle I'm happy to remove them in a follow up PR. The reason for doing this in the first place is that not all loggers emit the exception (eg, the Console logger does not - afaik)

sergeybykov:
@Arshia001 Sorry about that. We published a wrong build for 3.1.1. It has been unpublished since.

src/Orleans.Runtime/Utilities/SiloUnobservedExceptionsHandler.cs
SebastianStehle: I think it does: https://github.com/dotnet/extensions/blob/master/src/Logging/Logging.Console/src/ConsoleLogger.cs#L135

Arshia001:
@sergeybykov so it happens even to the best, huh? XD

sergeybykov:
>@sergeybykov so it happens even to the best, huh? XD

It happens when somebody dumb like myself is involved. 🙁

PR #6366, opened 2020-02-28
Basic implementation of IGrainDirectory using the Azure Table API

benjaminpetit:


src/Azure/Orleans.GrainDirectory.AzureStorage/AzureTableGrainDirectory.cs
```diff
@@ -0,0 +1,148 @@
...
+            var result = await this.tableDataManager.ReadSingleTableEntryAsync(this.clusterId, AzureTableUtils.SanitizeTableProperty(address.GrainId));
+
+            // No entry found
+            if (result == null)
+                return;
+
+            // Check if the entry in storage match the one we were asked to delete
+            var entity = result.Item1;
+            if (entity.ActivationId == address.ActivationId)
+                await this.tableDataManager.DeleteTableEntryAsync(GrainDirectoryEntity.FromGrainAddress(this.clusterId, address), entity.ETag);
```
ReubenBond: Maybe the ActivationId should be the ETag :P 

... eventually

src/Orleans.Core.Abstractions/GrainDirectory/IGrainDirectory.cs
```diff
@@ -58,5 +58,22 @@ public class GrainAddress
...
+                   this.ActivationId == address.ActivationId;
+        }
+
+        public override int GetHashCode()
+        {
+            var hashCode = 1043893337;
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(this.SiloAddress);
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(this.GrainId);
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(this.ActivationId);
+            return hashCode;
```
ReubenBond: We can pull in the Microsoft.Bcl.HashCode package and use `HashCode.Combine` here instead (but this is fine, too)

src/Azure/Shared/Storage/AzureTableDataManager.cs
```diff
@@ -216,6 +218,45 @@ public async Task<string> UpsertTableEntryAsync(T data)
...
+        {
+            const string operation = "InsertTableEntry";
+            var startTime = DateTime.UtcNow;
+            if (Logger.IsEnabled(LogLevel.Trace)) Logger.Trace("{0} entry {1} into table {2}", operation, data, TableName);
+
+            try
+            {
+                try
+                {
+                    // WAS:
```
sergeybykov: Don't need this commented out code?

src/Azure/Shared/Storage/AzureTableDataManager.cs
```diff
@@ -216,6 +218,45 @@ public async Task<string> UpsertTableEntryAsync(T data)
...
+                    var opResult = await tableReference.ExecuteAsync(TableOperation.Insert(data));
+                    return (true, opResult.Etag);
+                }
+                catch (StorageException storageException) when (storageException?.RequestInformation?.HttpStatusCode == (int)HttpStatusCode.Conflict)
+                {
+                    return (false, null);
+                }
+                catch (Exception exc)
+                {
+                    Logger.Warn((int)Utilities.ErrorCode.AzureTable_06,
```
sergeybykov: This is copy-pasted from above. Can we avoid code duplication like that with a helper method instead? There's a number of similar cases in this code file.

sergeybykov: BTW, I don't understand why we same "Intermediate" here.

src/Orleans.Core.Abstractions/GrainDirectory/IGrainDirectory.cs
benjaminpetit: I just want to avoid pulling too many dependencies on `Abstractions`...

@sergeybykov not sure what you are suggesting here. It's a `GetHashCode` generated from VS2019

sergeybykov: I'm fine with that.

ReubenBond:
I think we're waiting for at least one fix from Ben, to correctly sanitize the RowId column

PR #6370, opened 2020-02-28
Simplify OrleansTaskScheduler

ReubenBond:
This PR removes some unnecessary code from OrleansTaskScheduler, including the `TaskScheduler` base class itself (so is the name still fitting? I don't know).

This involves hoisting a few lines of code into `WorkItemGroup` (the core 'schedule this' code), but mostly removing unneeded code.

In addition, the PR removes some unnecessary wrapping/allocations where a closure is wrapped in a ClosureWorkItem, which is then wrapped in a Task and scheduled.

Instead, the closure can often be passed directly as an Action and only wrapped in a Task if the a context is present - otherwise it is scheduled directly on the ThreadPool.

There's more which could be done, but I think this is a good start to unraveling and simplifying some of the remaining scheduling code.


src/Orleans.Runtime/GrainDirectory/ClientObserverRegistrar.cs
```diff
@@ -54,7 +54,7 @@ internal void SetHostedClient(IHostedClient client)
             this.hostedClient = client;
             if (client != null)
             {
-                this.scheduler.QueueAction(Start, this).Ignore();
+                this.scheduler.QueueAction(Start, this);
```
ReubenBond: Note that this method is now implemented on OrleansTaskScheduler itself rather than as an extension. It is a void method now.
The extension method which this line previously called (which returns a Task) was renamed to `QueueActionAsync`.

sergeybykov:
Build failure?

ReubenBond:
Oops, pushed a fix.

PR #6374, opened 2020-03-01
Fix errors reported by GitHub Semmle code analysis tools.

jthelin:
https://lgtm.com/projects/g/dotnet/orleans/?mode=list&severity=error

Fix errors reported by GitHub Semmle code analysis tools:

- Missing X-Frame-Options HTTP header - https://lgtm.com/rules/1506102236348/
- Using a package with a known vulnerability - https://lgtm.com/rules/1506697777412/
- Possible loss of precision - https://lgtm.com/rules/1506096756023/

Ensure Test.cmd clears the `Platform` environment variable (same as in Build.cmd) in case it is run on separate cmd session (eg Appveyor builds)

Remove unused $outDir param from Parallel-Tests.ps1 script.

### DIFF
--- a/Parallel-Tests.ps1
+++ b/Parallel-Tests.ps1
@@ -3,7 +3,7 @@ param(
     [string] $testFilter,
     [string] $dotnet)
 
-$maxDegreeOfParallelism = ($env:NUMBER_OF_PROCESSORS, 4).Minimum
+$maxDegreeOfParallelism = [math]::min($env:NUMBER_OF_PROCESSORS, 4)
 Write-Host "Max Job Parallelism = $maxDegreeOfParallelism"
 
 $failed = $false
--- a/Parallel-Tests.ps1
+++ b/Parallel-Tests.ps1
@@ -3,7 +3,7 @@ param(
     [string] $testFilter,
     [string] $dotnet)
 
-$maxDegreeOfParallelism = $env:NUMBER_OF_PROCESSORS
+$maxDegreeOfParallelism = ($env:NUMBER_OF_PROCESSORS, 4).Minimum
 Write-Host "Max Job Parallelism = $maxDegreeOfParallelism"
 
 $failed = $false
--- a/Parallel-Tests.ps1
+++ b/Parallel-Tests.ps1
@@ -1,10 +1,11 @@
 param(
     [string[]] $directories,
     [string] $testFilter,
-    [string] $outDir,
     [string] $dotnet)
 
-$maxDegreeOfParallelism = 4
+$maxDegreeOfParallelism = $env:NUMBER_OF_PROCESSORS
+Write-Host "Max Job Parallelism = $maxDegreeOfParallelism"
+
 $failed = $false
 
 if( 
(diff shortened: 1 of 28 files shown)


jthelin:
The first run of the **Functional** test suite failed with what looks like a system clock resolution edge case glitch in test case Tester.AzureUtils.Persistence.PersistenceProviderTests_Local.PersistenceProvider_Memory_FixedLatency_WriteRead

```
Write: Expected minimum latency = 00:00:00.2000000 Actual = 00:00:00.1975619

   at Tester.AzureUtils.Persistence.PersistenceProviderTests_Local.PersistenceProvider_Memory_FixedLatency_WriteRead() in C:\agent\_work\2\s\test\Extensions\TesterAzureUtils\Persistence\PersistenceProviderTests.cs:line 237
```

Hopefully rerunning the test will get a better result next time!

jthelin:
There are also another 222 Warnings and 34 Recommendations reported by Semmle code analysis, if anyone wants to lend a hand with some code cleanup work.

https://lgtm.com/projects/g/dotnet/orleans/?mode=tree

jthelin:
PR #6378 addresses the unrelated test glitch found when running the Functional test suite for this PR.

Parallel-Tests.ps1
```diff
@@ -1,10 +1,11 @@
 param(
     [string[]] $directories,
     [string] $testFilter,
-    [string] $outDir,
     [string] $dotnet)
 
-$maxDegreeOfParallelism = 4
```
sergeybykov: This was intentionally constrained because we saw issues with too many tests trying to execute in parallel. There's a chance that issue is gone now, but I doubt it.

Parallel-Tests.ps1
```diff
@@ -1,10 +1,11 @@
 param(
     [string[]] $directories,
     [string] $testFilter,
-    [string] $outDir,
```
sergeybykov: This might be used in the ADO build definition. I'm not 100% positive though.

Test.cmd
```diff
@@ -15,12 +19,6 @@ call Ensure-DotNetSdk.cmd
 pushd "%CMDHOME%"
 @cd
 
-SET TestResultDir=%CMDHOME%\Binaries\%BuildConfiguration%\TestResults
```
sergeybykov: Why delete this? I believe it is used by the ADO build.

Parallel-Tests.ps1
jthelin: I was actually trying to accommodate environments with less that 4 (v-)cores available, which were causing the same problems that you probably originally encountered.
I can change this to Max(num_procs, 4) if preferred.

Parallel-Tests.ps1
jthelin: This param is not used anywhere else in this script, which was why it seems like a candidate for removal. 
All the ADO tests linked to GitHub PR validation seem to pass OK, because they used the `Test.cmd` script which calls this `Parallel-Tests.ps1` script. 
Do you have other (internal) builds that call `Parallel-Tests.ps1` directly?

Test.cmd
jthelin: This was the value passed as the $outDir param in `Parallel-Tests.ps1`, which isn't used in that script. 
Sounds like you are saying this is read as an environment variable by something in one of the builds?
Would you prefer I can leave the EV even though the param is not used?

jthelin: I un-removed these EV's, although I can't find where they might be used.

Parallel-Tests.ps1
jthelin: Fixed: Limit parallel test jobs to max 4, or NUMBER_OF_PROCESSORS if that is less.

Parallel-Tests.ps1
sergeybykov: I checked - we do simply call Test.cmd. You are right.

sergeybykov:
Thank you, Jorgen!

Issue #6330, opened 2020-02-22
Reminders period overflow issue in ADO.NET Reminders Table


xontab:
In the ADO.NET reminders table implementation, the period is stored in milliseconds as Integer.  Thus it is not possible to set it more than 24.8 days, as it will overflow and thus the wrong value is stored in the DB.  At the moment there is no validation so it will not throw any exceptions or errors when this happens.

https://github.com/dotnet/orleans/blob/38e1e86f5af92994bcc83b37c5228df95be6b892/src/AdoNet/Shared/Storage/DbStoredQueries.cs#L399-L402

We are ready to create a PR with a fix, however there are a number of options:
* Option 1: Change it to `Long` - This will ensure backwards compatibility with the existing DB Schema
* Option 2: Change it to `String` - Similar to other reminder storage providers (example Azure and DynamoDB)
* Option 3: Change it to `Double` - Same type used for `TotalMilliseconds` to support both whole and fractional milliseconds

What do you think?

sergeybykov:
Tagging @veikkoeeva for his opinion.

veikkoeeva:
I think Orleans cannot support fractional milliseconds and I'm not aware of such plans. Also looking at https://github.com/dotnet/orleans/blob/master/src/Orleans.Runtime/ReminderService/ReminderRegistry.cs#L30 the maximum is `4294967294` ms of type `uint`, or 49.7 days while calculations use `long` if I got my mental pseudocalculations correct.

In practice see:

1. Needs to be changed to all database scripts to `long` to match code.
2. Needs to be changed in the code.
3. Tests run. I can run the MySQL and PostgreSQL and/or SQL Server depending what is needed if this is needed.

So, the change to `long` a necessary fix unless @sergeybykov or @xontab can come up with other arguments.

As a side note, I remember this topic came up in the very early days while checking other bugs during refactoring the codebase. In hindsight this ought to have refactored already back then, but the plan was also to make a better implementation of the reminder table (narrow name indexes, good range index for queries taking cues from `Storage`). In enterprise space there are solutions and 3rd party services doing this in scaleable with good enough accuracy and resolution, so this is an important subservice. Here we are again, glad to see this receiving attention. Go for it! :)

xontab:
[...]

PR #6390, opened 2020-03-11
Reminders period overflow issue in ADO.NET Reminders Table

xontab:
Issue https://github.com/dotnet/orleans/issues/6330

Updates:
* Changed data types from `int` to `long` in reminders ADO.NET when period is greater than max int (due to compatibility reasons)


veikkoeeva:
I have not tested this, but apparently @xontab has changed all the places I could immediately think of. Did you give a run for this code, @xontab ? This looks OK to me.

xontab:
I ran a number of manual tests to make sure that the reminders still work even with the old schema to make sure that the fix is backwards compatible.  I ran the following scenarios on SQL Server, Postgres, MySQL, Oracle (all containerised) using both the old and new schema changes. 

Scenario 1: Within integer range - 60,000ms (to make sure that it works on both schemas)
Scenario 2: Max supported period value - 4,294,967,294ms (this is the fix, thus it only works on the new schema)
Scenario 3: Period greater than 4,294,967,294ms (this will always throw a validation error)

To make sure that the reminders are still working properly, I ran the following steps:
* Used `RegisterOrUpdateReminder` to run the above scenarios
* Restarted the Silo (to make sure that the correct values are read from the DB)
* Verified that the correct values are loaded in the Orleans Dashboard

* Verified that the reminders are still firing for the first 3 periods (by changing the computer date and time settings)

If you want me to run more test scenarios, please let me know.

veikkoeeva:
@xontab This looks reasonable. It is a lot more work to introduce new tests to the system to check these. Should we record an issue to add more tests to the system (and where to add them as a pointer to others), @ReubenBond or @sergeybykov ?

sergeybykov:
Thank you, @xontab and @veikkoeeva!

Feel free to open a tracking issue for additional tests.

xontab:
@veikkoeeva I didn't add any integration tests as all the above tests depend on a different DB engines which might be a bit tricky to integrate with the CI/CD pipeline, however I guess we can start by adding tests which use an `InMemoryReminderTable`.  I am ready to add more tests related to Reminders to help out and contribute to the project. :)

veikkoeeva:
@xontab That is true about test pipeline. Currently at least the ADO.NET persistence tests are not run at least on these GH tests.

Currently persistence tests are done one way to accomondate for the different DB engines and make them show nicely on the test filter. There are better ways these days besides having a next iteration and just not getting it done in the middle of night. :)

I would like to refactor persistence tests to use dependency injection and asynchronous setup similar to https://github.com/dotnet/orleans/blob/master/Samples/OneBoxDeployment/test/OneBoxDeployment.IntegrationTests/IntegrationTestFixture.cs#L58 with default DB values but so that they're more easily configurable with one's custom values. Indeed, as picked via the normal .NET Core configuration mechanism.

There would be plenty of work in that regard there. Since starting a startup and and working on more or less global projects I've had too little time lately to accomplish bigger tasts other than pushing iteration one of OneBoxDeployment over finish line. :)

PR #6399, opened 2020-03-13
Reduce port clashes in TestCluster

ReubenBond:
This will not help as much on Linux, since named mutexes are only process-wide on Linux.
On Windows, this ought to work fairly well - we'll see 😅


ReubenBond:
I need to address some issues with this and perform some thread synchronization (yuck)

slonopotamus:
Why don't you just bind to port=0 and let OS _guarantee_ you a free port? Present solution is full of race conditions and doesn't work reliably.

ReubenBond:
The current solution uses in-memory transports and doesn't bind to ports during tests, by default.

There is a reason why we also don't do what you're suggesting, though: the port is a part of the silos identity and it needs to know that before the bind occurs. Something different could work, but it would require significant changes.

slonopotamus:
If tests are not binding to TCP ports, what's the point in TestClusterPortAllocator _at all_?

ReubenBond:
Tests don't bind to TCP ports by default. We have some tests which use the TestCluster which spawn hosts in separate processes or test socket logic. Those tests do use TCP ports.

PR #6409, opened 2020-03-18
Use the overload of ConcurrentDictionary.GetOrAdd that takes a method

abe545:
I was just browsing the code, and noticed this randomly.
It seems like it wasn't actually useful previously, as the dictionary seems to be an attempt to not have to call `GenerateDeterministicGuid` every time it tries to get the queue grain.


ReubenBond:
Thanks, @abe545 :)

PR #6413, opened 2020-03-19
Avoid prematurely releasing mutex in TestClusterPortAllocator

ReubenBond:


src/Orleans.TestingHost/TestClusterPortAllocator.cs
```diff
@@ -167,6 +167,7 @@ public bool Acquire(string name)
                                 // Acquired
                                 _mutexes[name] = mutex;
                                 Interlocked.Increment(ref result[0]);
+                                return;
```
ReubenBond: *whoops*

PR #6414, opened 2020-03-20
Allow grain call filters to retry calls

### LINKED ISSUE #6391: Grain calls retry policy
Is there a standard and centralized way to implement retry strategies when calling grain methods from a client or from another grain?

I want to use Polly to retry calls that throw specific exceptions, like `ConnectionFailedException`. But I don't know where to put this code, without doing it manually everywhere I call a grain method.

I tried with an `OutgoingGrainCallFilter`, but it seems `context.Invoke()` cannot be called multiple times.

Thanks in advance for your help!

ReubenBond:
Fixes #6391


benjaminpetit:
/azp run Azure DevOps - BVT

LGTM, will merge as soon as BVT passes, just to be sure

PR #6417, opened 2020-03-20
Additional information

Zeroshi:
Added nuget package that should be used with the client along with the using statement in the code.


sergeybykov:
Thank you, @Zeroshi!

PR #6421, opened 2020-03-20
Fix flaky locator tests

benjaminpetit:


src/Orleans.Runtime/GrainDirectory/GrainLocator.cs
```diff
@@ -160,26 +170,27 @@ public void Participate(ISiloLifecycle lifecycle)
...
-                ListenToClusterChange().Ignore();
+                this.listenToClusterChangeTask = ListenToClusterChange();
                 return Task.CompletedTask;
             };
-            Task onStop(CancellationToken ct)
+            async Task onStop(CancellationToken ct)
             {
                 this.shutdownToken.Cancel();
-                return Task.CompletedTask;
+                await listenToClusterChangeTask;
```
ReubenBond: Check that it's not null first?

May also want to skip awaiting if `ct.IsCancellationRequested` - could this ever timeout?

BTW, I think all of our other local functions use upper case names.

benjaminpetit: It could timeout yes, since we do a `await this.grainDirectory.UnregisterSilos(deadSilos)` in `ListenToClusterChange`. I will propagate the `shutdownToken` where necessary, I think that will be enough

PR #6428, opened 2020-03-24
WIP: AWS S3 Grain Storage Provider

icanhasjonas:
S3 is a viable alternative for slow state access grains that have a small or large state.

It's very inexpensive and supports a huge read/write capacity, although much slower than `DynamoDB`

Typical costs:
```
  $5   / million writes
  $0.4 / million reads
 (+ data transfer costs)
```

It's very efficient for heavy reads but can get pretty expensive on heavy writes compared to DynamoDB.

Heads up; This is *NOT* a slower 1:1 alternative to `DynamoDB` as `S3` offers fewer data guarantees compared `DynamoDB`. For example, putting an existing object on S3 has eventual consistency while creating an object has instant consistency. There is no way that I know of that allows for optimistic locking or conditional writes.

### DIFF
--- a/src/AWS/Orleans.Persistence.S3/Provider/S3GrainStorage.cs
+++ b/src/AWS/Orleans.Persistence.S3/Provider/S3GrainStorage.cs
@@ -1,0 +1,128 @@
+using System;
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Amazon.S3;
+using Amazon.S3.Model;
+using Microsoft.Extensions.Logging;
+using Orleans.Persistence.S3.Options;
+using Orleans.Runtime;
+using Orleans.Serialization;
+using Orleans.Storage;
+
+namespace Orleans.Persistence.S3.Provider
+{
+    public class S3GrainStorage : IGrainStorage, ILifecycleParticipant<ISiloLifecycle>, ILifecycleObserver
+    {
+        private readonly string name;
+        private readonly S3StorageOptions options;
+        private readonly SerializationManager serializationManager;
+        private readonly IAmazonS3 s3;
+        private readonly IS3GrainStorageKeyFormatter keyFormatter;
+        private readonly ILogger logger;
+
+        public S3GrainStorage(
+            string name,
+            S3StorageOptions options,
+            SerializationManager serializationManager,
+            IAmazonS3 s3,
+            IS3GrainStorageKeyFormatter keyFormatter,
+            ILoggerFactory loggerFactory)
+        {
+            this.name = name;
+            this.options = options;
+            this.serializationManager = serializationManager;
+            this.s3 = s3;
+            this.keyFormatter = keyFormatter;
+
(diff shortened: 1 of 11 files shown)


src/AWS/Orleans.Persistence.S3/Provider/S3GrainStorage.cs
```diff
@@ -0,0 +1,128 @@
...
+                    Key = this.keyFormatter.FormatKey(this.name, grainType, grainReference)
+                }).ConfigureAwait(false);
+            }
+            catch (AmazonS3Exception e) when (e.StatusCode == HttpStatusCode.NotFound)
+            {
+                // Ignore not found errors
+            }
+        }
+
+        public void Participate(ISiloLifecycle lifecycle) => lifecycle.Subscribe<S3GrainStorage>(this.options.InitStage, this);
```
icanhasjonas: Will probably remove `ILifecycleParticipant` support as it's not really needed any more

src/AWS/Orleans.Persistence.S3/StreamExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+                }
+
+                if (bytesRead < bufferSize)
+                {
+                    break;
+                }
+            }
+
+            return result;
+        }
```
icanhasjonas: Not used anymore

src/AWS/Orleans.Persistence.S3/StreamExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+                await stream
+                    .CopyToAsync(mem, 8192, cancellationToken)
+                    .ConfigureAwait(false);
+            }
+
+            return mem.AsArraySegment();
+        }
+
+        public static ArraySegment<byte> MergeToSingleSegment(this IReadOnlyList<ArraySegment<byte>> segments)
+        {
```
icanhasjonas: Add check for `Count == 0`

jsteinich:
One concern I have is that the consistency controls around duplicate activations works differently than other storage providers, i.e. others check on write to make sure they have the most up to date state. Perhaps this won't be a big concern since it seems unlikely to be used in a write heavy application.

icanhasjonas:
That's a good point. There are no guarantees of order on S3 and you are right, this provider is not a good choice for applications requiring any one of those guarantees.

sergeybykov:
Without support for etags, the provider would not conform to the protocol. Blind writes are dangerous regardless of how write heavy an application is. Doesn't S3 support some form of etags?

icanhasjonas:
> Without support for etags, the provider would not conform to the protocol. Blind writes are dangerous regardless of how write heavy an application is. Doesn't S3 support some form of etags?

`ETag` on read, only. There are, as I can see, no support for conditional writes.

sergeybykov:
>ETag on read, only. There are, as I can see, no support for conditional writes.

Then I think we shouldn't do it as the behavior is inconsistent with all other persistence providers and can lead to occasional data corruption/loss.

Issue #6383, opened 2020-03-04
Expose configuration for Azure Storage timeouts


ReubenBond:
See `AzureTableDefaultPolicies`. These values should be configurable on a per-provider basis.

Currently, users do not have any mechanism for customizing these timeouts or other policies.

* [ ] Expose for all providers
* [ ] Ensure all storage methods in (eg) AzureTableDataManager and AzureBlobGrainStorage respect policies

veikkoeeva:
@ReubenBond Does that "expose for all providers" mean all that can be used in Orleans and so a general mechanism that storage providers can rely on configuring user customizable retries? I take that "(eg) AzureTableDataManager" means as for an example but not excluding others.

If so, such a mechanism would be welcome and I think it should be dynamic since a real world scenario would be SLA per customer per time of day or some other event (that then extends to dynamic billing).

ReubenBond:
@veikkoeeva I'm not sure about a general mechanism. Different policies are relevant to different instances of a provider. I think as a first pass, we should just try to make sure that the relevant knobs are accessible and consistently put to use

veikkoeeva:
@ReubenBond I think some kind of a layered approach would be needed. So that the way they are exposed looks similar across all providers even if the provider specific knobs may vary. Then maybe some convenience functions that can be provider specific. I think lambdas, naturally.

An example way of thinking, for ADO.NET some improvements could already be like `.UseMysql` whereas it would then apply the right constants, use lambda to load the right connector (or by default like now try to load either of the most used ones) and so on. To be explicit about one thing, I don't suggest creating a Nuget package per ADO provider to have these functions.

ReubenBond:
I believe this was fixed by #6432

Issue #4557, opened 2018-05-04
Azure retry policy in Orleans


jason-bragg:
Pretty much the same issue as "Json serialization settings in Orleans" #4555, but with azure table retry policies.

PR #6432, opened 2020-03-25
Expose timeouts for Azure Table Storage

govenkat-ms:
https://github.com/dotnet/orleans/issues/6383.


src/Azure/Shared/Storage/AzureTableDataManager.cs
```diff
@@ -708,7 +719,7 @@ private CloudTableClient GetCloudTableOperationsClient()
                 CloudStorageAccount storageAccount = AzureTableUtils.GetCloudStorageAccount(ConnectionString);
                 CloudTableClient operationsClient = storageAccount.CreateCloudTableClient();
                 operationsClient.DefaultRequestOptions.RetryPolicy = AzureTableDefaultPolicies.TableOperationRetryPolicy;
-                operationsClient.DefaultRequestOptions.ServerTimeout = AzureTableDefaultPolicies.TableOperationTimeout;
+                operationsClient.DefaultRequestOptions.ServerTimeout = this.tableOperationTimeout;
```
keriehm: It would also be useful to control the MaximumExecutionTime property here (the client-side timeout).

src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
```diff
@@ -25,38 +25,44 @@ internal class OrleansSiloInstanceManager
         private readonly AzureTableDataManager<SiloInstanceTableEntry> storage;
         private readonly ILogger logger;
 
-        internal static TimeSpan initTimeout = AzureTableDefaultPolicies.TableCreationTimeout;
```
jason-bragg: Shouldn't we get rid of AzureTableDefaultPolicies.TableCreationTimeout, and AzureTableDefaultPolicies.TableOperationTimeout.  It would be best to be consistent, and not have duplicate settings.

src/Azure/Shared/Storage/AzureStorageOptions.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace Orleans.Configuration
+{
+    public abstract class AzureStorageOptions
+    {
+        public TimeSpan CreationTimeout { get; set; }
```
jason-bragg: Shouldn't we set these to the defaults from AzureTableDefaultPolicies?

src/Azure/Orleans.Clustering.AzureStorage/AzureBasedMembershipTable.cs
```diff
@@ -40,8 +39,13 @@ public async Task InitializeMembershipTable(bool tryInitTableVersion)
...
             LogFormatter.SetExceptionDecoder(typeof(StorageException), AzureTableUtils.PrintStorageException);
 
-            tableManager = await OrleansSiloInstanceManager.GetManager(
-                this.clusterId, options.ConnectionString, options.TableName, this.loggerFactory);
+            this.tableManager = await OrleansSiloInstanceManager.GetManager(
+                this.clusterId,
+                this.options.ConnectionString,
+                this.options.TableName,
+                this.loggerFactory,
+                this.options.CreationTimeout,
```
jason-bragg: Could pass in options class rather than each specific option.  Class didn't originally because it was written before we started using options.

src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
govenkat-ms: It's not really duplicate, since users have the option to not specify a timeout, in which case Orleans would need to default to a value.

src/Azure/Shared/Storage/AzureTableDataManager.cs
govenkat-ms: Good point. Based on the retry guidance: https://docs.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific, it's better to configure the MaxExecutionTime instead.

src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
jason-bragg: My main concern is that if we've duplicate settings for the same behavior it will be more difficult for us and other users to figure out which they should be setting/using.  As it is, we've a single consistent place for these values, but no good pattern for users to specify them, especially on a per provider basis.  If we're going to address this, my preference would be to do so consistently.  Even if we don't change all of the azure settings in this PR, the pattern should be consistent and extendable to all azure storage points.

As you probably only need this for specific areas, I don't think it's reasonable to block this change due to incompleteness, when we can create a work item to migrate the changes across the other storage points, but it is important to get the pattern consistent.

I noted this issue a while back, but it's been in the backlog for a bit.  https://github.com/dotnet/orleans/issues/4557  @ReubenBond, @sergeybykov, @benjaminpetit, thoughts?  If I'm the only one concerned, then I'm probably being too nit-picky (it happens :])

ReubenBond: Yes, I think we should remove `AzureTableDefaultPolicies` altogether and adopt a consistent pattern.

Regarding this inheritance-based design, we used a similar pattern for {Client,Silo}MessagingOptions, and regretted it, because users made the mistake of configuring the base class (`.Configure<AzureStorageOperationOptions>(...)`) instead of the correct class, resulting in misconfiguration (no configuration). See #5574, for example.

I would prefer we use composition instead of inheritance. For example, by adding a field called `StoragePolicy`, of type `AzureStoragePolicyOptions` to the existing provider options (`AzureStorageClusteringOptions`, etc). That policy class would contain the properties currently in `AzureTableDefaultPolicies`.

How does that sound?

jason-bragg: Regarding inheritance vs composition, the inheritance pattern used in this PR 'shouldn't' have the same issue we ran into with MessagingOptions because the base class is abstract, so it's comparable to an interface, hence I was less concerned with that.

For consistency, I think we should settle on what behaviors we want to support.  I'm confident we minimally want to allow users to provide their own policy settings per provider.  Do we want to also support a configurable global default?

If we don't want a configurable global default, then an inheritance model or composition model with default values both 'should' get us what we want.  If we want to support a configurable global default, things get more complicated, because we'd need a way to fallback to a configured global default if settings were not configured in the provider (as we do with ClusterOptions).  For this reason, I suspect supporting a configurable global default is probably not warranted.

If we agree that a configurable global default isn't needed, then we need only move the defaults from AzureTableDefaultPolicies (and remove that class) to the new configuration class (whether it's introduced via inherited or composed).

govenkat-ms: But the thing is that this "StoragePolicy" property has to be applied to all AzureStorage<>Options anyway which requires some form of inheritance to avoid duplication.

govenkat-ms: Spoke with Reuben offline, tableName is now abstract, and the policyOptions is now injectable into every Orleans Provider. Removed the Global config.

govenkat-ms:
PubSub_Store_ReadError test passes on my local machine. Is this a known flaky test?

src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
```diff
@@ -78,24 +78,21 @@ public void RegisterSiloInstance(SiloInstanceTableEntry entry)
...
+            Task.WaitAll(new Task[] { storage.UpsertTableEntryAsync(entry) });
         }
 
         public void UnregisterSiloInstance(SiloInstanceTableEntry entry)
         {
             entry.Status = INSTANCE_STATUS_DEAD;
             logger.Info(ErrorCode.Runtime_Error_100271, "Unregistering silo instance: {0}", entry.ToString());
-            storage.UpsertTableEntryAsync(entry)
-                .WaitWithThrow(AzureTableDefaultPolicies.TableOperationTimeout);
+            Task.WaitAll(new Task[] { storage.UpsertTableEntryAsync(entry) });
```
jason-bragg: Why use wait all then there is one task?
Aren't we loosing timeout here?

src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
```diff
@@ -78,24 +78,21 @@ public void RegisterSiloInstance(SiloInstanceTableEntry entry)
...
         }
 
         public void ActivateSiloInstance(SiloInstanceTableEntry entry)
         {
             logger.Info(ErrorCode.Runtime_Error_100272, "Activating silo instance: {0}", entry.ToString());
             entry.Status = INSTANCE_STATUS_ACTIVE;
-            storage.UpsertTableEntryAsync(entry)
-                .WaitWithThrow(AzureTableDefaultPolicies.TableOperationTimeout);
+            Task.WaitAll(new Task[] { storage.UpsertTableEntryAsync(entry) });
```
jason-bragg: Same comment as from line  88, and for all similar changes.

src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
govenkat-ms: Addressed them in the new iteration

src/Azure/Orleans.Clustering.AzureStorage/OrleansSiloInstanceManager.cs
govenkat-ms: Addressed them in the new iteration

sergeybykov:
Thank you, @govenkat-ms!

src/Azure/Shared/Storage/AzureStoragePolicyOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace Orleans.Transactions.AzureStorage
+#elif ORLEANS_DIRECTORY
+namespace Orleans.GrainDirectory.AzureStorage
+#else
+// No default namespace intentionally to cause compile errors if something is not defined
+#endif
+{
+    public class AzureStoragePolicyOptions
+    {
+        public int MAX_BULK_UPDATE_ROWS { get; set; }
```
ReubenBond, line 29: We should use a consistent naming scheme here.

src/Azure/Shared/Storage/AzureStoragePolicyOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+
+			this.PauseBetweenBusyRetries = (!Debugger.IsAttached)
+					? TimeSpan.FromMilliseconds(500)
+					: TimeSpan.FromSeconds(5);
+
+			this.CreationTimeout = TimeSpan.FromMilliseconds(this.PauseBetweenCreationRetries.TotalMilliseconds * this.MaxCreationRetries * 3);
+			this.OperationTimeout = TimeSpan.FromMilliseconds(this.PauseBetweenOperationRetries.TotalMilliseconds * this.MaxOperationRetries * 6);
+			this.BusyRetriesTimeout = TimeSpan.FromMilliseconds(this.PauseBetweenBusyRetries.TotalMilliseconds * this.MaxBusyRetries);
+			this.CreationRetryPolicy = new LinearRetry(this.PauseBetweenCreationRetries, this.MaxCreationRetries);
+			this.OperationRetryPolicy = new LinearRetry(this.PauseBetweenOperationRetries, this.MaxOperationRetries);
```
ReubenBond, line 70: indentation is broken in this method

src/Azure/Shared/Storage/AzureTableDataManager.cs
```diff
@@ -48,23 +48,25 @@ internal class AzureTableDataManager<T> where T : class, ITableEntity, new()
         /// <summary> Connection string for the Azure storage account used to host this table. </summary>
         protected string ConnectionString { get; set; }
 
-        private CloudTable tableReference;
+        public CloudTable Table { get; private set; }
 
-        public CloudTable Table => tableReference;
+        public AzureStoragePolicyOptions StoragePolicyOptions;
```
ReubenBond, line 53: This should not be a public, mutable _field_. At the minimum, it should be a property. It should also be immutable.

src/Azure/Shared/Storage/AzureTableDataManager.cs
```diff
@@ -48,23 +48,25 @@ internal class AzureTableDataManager<T> where T : class, ITableEntity, new()
...
 
         /// <summary>
         /// Constructor
         /// </summary>
         /// <param name="tableName">Name of the table to be connected to.</param>
         /// <param name="storageConnectionString">Connection string for the Azure storage account used to host this table.</param>
         /// <param name="logger">Logger to use.</param>
-        public AzureTableDataManager(string tableName, string storageConnectionString, ILogger logger)
+        /// <param name="storagePolicyOptions">Optional Storage Policy Configuration.</param>
+        public AzureTableDataManager(string tableName, string storageConnectionString, ILogger logger, AzureStoragePolicyOptions storagePolicyOptions = default)
```
ReubenBond, line 62: This really should not allow a `null` value for the policy object. It should throw if it's null.

The same goes for every time a default parameter value is allowed.

src/Azure/Shared/Storage/AzureTableDataManager.cs
```diff
@@ -48,23 +48,25 @@ internal class AzureTableDataManager<T> where T : class, ITableEntity, new()
...
         {
             Logger = logger;
             TableName = tableName;
             ConnectionString = storageConnectionString;
 
             AzureTableUtils.ValidateTableName(tableName);
+            this.StoragePolicyOptions ??= storagePolicyOptions;
```
ReubenBond, line 69: As per the above, this should be a regular =, not a ??=.

src/Azure/Shared/Storage/AzureStorageOperationOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+#else
+// No default namespace intentionally to cause compile errors if something is not defined
+#endif
+{
+    public abstract class AzureStorageOperationOptions
+    {
+        /// <summary>
+        /// Azure Storage Policy Options
+        /// </summary>
+        public AzureStoragePolicyOptions StoragePolicyOptions { get; set; }
```
ReubenBond, line 31: Make this get-only and set it to `new AzureStoragePolicyOptions()`

src/Azure/Orleans.Reminders.AzureStorage/Storage/RemindersTableManager.cs
```diff
@@ -98,9 +90,15 @@ await singleton.InitTableAsync()
...
-        private RemindersTableManager(string serviceId, string clusterId, string storageConnectionString, string tableName, ILoggerFactory loggerFactory)
-            : base(tableName, storageConnectionString, loggerFactory.CreateLogger<RemindersTableManager>())
+        
+        private RemindersTableManager(
+            string serviceId,
+            string clusterId,
+            string storageConnectionString,
+            string tableName,
+            ILoggerFactory loggerFactory,
+            AzureStoragePolicyOptions storagePolicyOptions = default)
```
ReubenBond, line 100: Throw if null, don't allow default value.

PR #6439, opened 2020-03-27
Streamline Dictionary use and remove some dead code

pentp:
* Streamline `Dictionary` use and remove some dead code
* Use `IsDefined` instead of `GetCustomAttributes().Any()`


pentp:
I rebased and cleaned up some of the changes and removed the race condition fix (will create a separate PR for that).

ReubenBond:
Thanks, @pentp, it's a nice cleanup

Issue #6345, opened 2020-02-25
Error doing DeleteMessage for Azure storage queue + Receiving events multiple times


rallets:
Hi, it looks like I've encountered the same behaviour as in [Receiving events from Azure Queue Stream Provider multiple times #890](https://github.com/dotnet/orleans/issues/890) and I was wondering if it still the same situation.
I've a project using Orleans 3.0.2 (with Storage emulator) and I've a grain that generate 1000 msg (from a integration/load test) and are processed pretty slowly. It looks like that exactly after 1 minute the queue is not purged, all messages are back to the queue and then some are re-processed. Others messages are not processed at all and they keep to stay in the queue (800/1000). That means it's not just a transient error, it's not recoverable. 
So I get a ton of:

```
fail: Orleans.AzureUtils.AzureQueueDataManager[103211]
      Error doing DeleteMessage for Azure storage queue yourprefix-azurequeueprovider-0
      Exception = Microsoft.Azure.Storage.StorageException: The specified message does not exist.
```

I tried to increase ConfigurePullingAgent.MaxEventDeliveryTime to 60 minutes, but after a silo startup (webapi w/direct client-cohosting) it process some hundred messages and then is back to the same error. The same if I clean the queue and restart the process.

I was wondering which are the possible workaround. Should I just reduce the CacheSize? or it's possible to increase a timeout?
Thank you

sergeybykov:
Have you tried running the code against the actual Azure Queue, not the emulator? Sometimes the emulator behaves differently from the actual service and might have its own bugs.

rallets:
Hi @sergeybykov, sorry for the delay. I just tried with a real Azure Queue, and the behaviour is the same.

jason-bragg:
What is the visibility timeout on the queue?  I believe the default is 30 seconds.
How long does it take to process a message in your service?

Events are read, put in the cache, then delivered for processing by the consumers.  Once successfully delivered, they are deleted from the queue.  If this sequence takes longer than the 'visibility time' the queue will regenerate the message.

rallets:
I was using the default (null), and by the [docs ](https://docs.microsoft.com/en-us/rest/api/storageservices/get-messages) it looks like it's 30 seconds.
I tried increasing `AzureQueueOptions.MessageVisibilityTimeout` and all went fine, but then I got a ton of
`warn: Orleans.AzureUtils.AzureQueueDataManager[103213]
[...]

PR #6441, opened 2020-03-28
Tuning Azure Queue Streams

rallets:
Collecting input from @jason-bragg and issue [6345](https://github.com/dotnet/orleans/issues/6345#issuecomment-595419498) about tuning azure queue streams. I found it very interesting and helpful to better understand how the system works. I thought was smart to share it in the docs.

### DIFF
--- a/src/Documentation/implementation/streams_implementation/azure_queue_streams.md
+++ b/src/Documentation/implementation/streams_implementation/azure_queue_streams.md
@@ -64,7 +64,7 @@ In a production system can emerge the need of tuning over the default configurat
 
 ### Example
 
-We assume a system with this characteristics:
+Assuming a system with this characteristics:
 
 * 100 streams,
 * 10 queues,
@@ -80,7 +80,7 @@ Hence **20 streams/queue** (10 streams/queue x 2 as safety factor) is probably r
 
 * Messages/minute: This means each queue will be expected to process up to **1200 messages/minute** (60 messages x 20 streams).
 
-The we can determine the visibility time to use:
+Then we can determine the visibility time to use:
 
 * Visibility time: The cache time (1 minute) is configured to hold 1 minute of messages (so 1200 messages, as we calculated messages/minute above).
 We assumed that each message takes 30 ms to process, then we can expect messages to spend up to 36 seconds in the cache (0.030 sec x 1200 msg = 36 sec), so the visibility time - doubled for safety - would need be over 72 seconds (36 sec of time in cache x 2).
--- a/src/Documentation/implementation/streams_implementation/azure_queue_streams.md
+++ b/src/Documentation/implementation/streams_implementation/azure_queue_streams.md
@@ -53,3 +53,62 @@ hostBuilder
     options.QueueNames = new List<string> { "yourprefix-azurequeueprovider-0" };
   }))
 ```
+
+## Tuning
+
+In a production system can emerge the need of tuning over the default configuration. When tuning there are some factors that should be considered, and it's service specific.
+
+1. First, most of the settings are per queue, so for a large number of streams, the load on each queue can be reduced by configuring more queues.
+2. Since streams process messages in order per stream, the gating factor will be the number of events being sent on a single stream.
+3. A reasonable balance of **cache time** (`StreamPullingAgentOptions.GetQueueMsgsTimerPeriod`) vs **visibility time** (`AzureQueueOptions.MessageVisibilityTimeout`) is that the visibility should be configured **to double** the time messages **are expected** to be in the cache.
+
+### Example
+
+We assume a system with this characteristics:
+
+* 100 streams,
+* 10 queues,
+* Each stream processing 60 messages per minute,
+* Each message takes around 30ms to process,
+* 1 minute worth of messages in cache (cache time).
+
+So we can calculate some parameters of the system:
+
+* Streams/queue: Even balancing of streams across queues would be an ideal 10 streams/queue (100 streams / 10 queues).
+But since streams won't always be evenly balanced over the queues, doubling (or even tripling) the ideal is safer than expecting ideal distribution.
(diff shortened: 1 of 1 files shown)


src/Documentation/implementation/streams_implementation/azure_queue_streams.md
```diff
@@ -53,3 +53,62 @@ hostBuilder
...
+
+In a production system can emerge the need of tuning over the default configuration. When tuning there are some factors that should be considered, and it's service specific.
+
+1. First, most of the settings are per queue, so for a large number of streams, the load on each queue can be reduced by configuring more queues.
+2. Since streams process messages in order per stream, the gating factor will be the number of events being sent on a single stream.
+3. A reasonable balance of **cache time** (`StreamPullingAgentOptions.GetQueueMsgsTimerPeriod`) vs **visibility time** (`AzureQueueOptions.MessageVisibilityTimeout`) is that the visibility should be configured **to double** the time messages **are expected** to be in the cache.
+
+### Example
+
+We assume a system with this characteristics:
```
jason-bragg: Minor nit pick about wording here.  Can we start with 'Assuming" rather then "We assume".  I make this suggestion because a reader not being attentive to the fact that this is an assumption within an example may interpret this as being the assumptions we (the Orleans team) are making about their service, rather than the assumptions for a specific example.  "Assuming", IMO, makes it more clear that the analysis below is based on the assumption.

src/Documentation/implementation/streams_implementation/azure_queue_streams.md
```diff
@@ -53,3 +53,62 @@ hostBuilder
...
+
+So we can calculate some parameters of the system:
+
+* Streams/queue: Even balancing of streams across queues would be an ideal 10 streams/queue (100 streams / 10 queues).
+But since streams won't always be evenly balanced over the queues, doubling (or even tripling) the ideal is safer than expecting ideal distribution.
+Hence **20 streams/queue** (10 streams/queue x 2 as safety factor) is probably reasonable.
+
+* Messages/minute: This means each queue will be expected to process up to **1200 messages/minute** (60 messages x 20 streams).
+
+The we can determine the visibility time to use:
```
jason-bragg: "The we can"  => "We can"?

src/Documentation/implementation/streams_implementation/azure_queue_streams.md
```diff
@@ -53,3 +53,62 @@ hostBuilder
...
+        options.ConnectionString = "xxx";
+        options.QueueNames = new List<string> {
+          "yourprefix-azurequeueprovider-1",
+          [...]
+          "yourprefix-azurequeueprovider-10",
+        };
+        options.MessageVisibilityTimeout = TimeSpan.FromSeconds(72);
+      }));
+    configurator.ConfigurePullingAgent(ob => ob.Configure(options => {
+      options.GetQueueMsgsTimerPeriod = TimeSpan.FromMinutes(1);
```
jason-bragg: Since nothing in this post explains the GetQueueMsgsTimerPeriod, should we remove this setting from the example?  Most of the others should be understood (like ConnectionString) or are relent to the example (like QueueNames and MessageVisibilityTimeout).  Only GetQueueMsgsTimerPeriod seems out of place for this explanation.

src/Documentation/implementation/streams_implementation/azure_queue_streams.md
jason-bragg: or "Then we can"?  Not sure.. your call :)

src/Documentation/implementation/streams_implementation/azure_queue_streams.md
rallets: I agree, much better with your suggestion!

src/Documentation/implementation/streams_implementation/azure_queue_streams.md
rallets: Fixed, thanks

src/Documentation/implementation/streams_implementation/azure_queue_streams.md
rallets: Then I misunderstood something... It looks like I wrongly supposed that "1 minute worth of messages in cache (cache time)" could be set - looking to the source code - by GetQueueMsgsTimerPeriod. If this is not the case, which setting can we use to configure it?

jason-bragg: Ah, I see.  Yeah I missed that too.  Good catch.  AzureQueueStreams (or any persistent stream provider using the SimpleQueueCache) configures cache by size, not time, using ConfigureCacheSize.  This can be a bit confusing because persistent stream providers which use the PooledQueueCache are configured by time rather than size, so it's easy to get confused as it's not always clear which cache a persistent stream provider is using.  Originally both were configured by size, but we changed the PooledQueueCache to use time.  Eventually I'd like to get rid of the SimpleQueueCache to avoid the confusion, but I digress.

From the example the cache size for 1 minute of data would be 1200 (default is 4k I think).  So I think we'd need something like:

    configurator.ConfigureCacheSize(1200);

Regarding GetQueueMsgsTimerPeriod: GetQueueMsgsTimerPeriod configures how often we query the queue when no data is returned or the cache is full.  The pulling agent will read from the queue as long as the cache is not full and the queue is returning data.  If the queue is empty, or the cache is full we don't want to free spin checking as that will burn up a lot of cpu, so this setting controls how often we retry.  By default it's 100ms.

jason-bragg:
Much thanks @rallets!

rallets:
Sorry @jason-bragg for the long delay and thanks for the very detailed explanation... hence I made the correction to remove the confusion about time and size in a new [PR 6542](https://github.com/dotnet/orleans/pull/6542)

PR #6442, opened 2020-03-29
Added projects and associated nuget packes to a table.

Zeroshi:
It seemed like it would be easier to find and read in a table format rather than text.


sergeybykov:
Thank you, @Zeroshi!

PR #6444, opened 2020-03-29
Dont use iowait in cpu calcs on linux

mchandler-manticore:
as it causes the silo to think its using 100% cpu when its not

iowait (since Linux 2.5.41)
                            (5) Time waiting for I/O to complete.  This
                            value is not reliable, for the following rea‐
                            sons:

                            1. The CPU will not wait for I/O to complete;
                               iowait is the time that a task is waiting for
                               I/O to complete.  When a CPU goes into idle
                               state for outstanding task I/O, another task
                               will be scheduled on this CPU.

                            2. On a multi-core CPU, the task waiting for I/O
                               to complete is not running on any CPU, so the
                               iowait of each CPU is difficult to calculate.

                            3. The value in this field may decrease in cer‐
                               tain conditions.

http://man7.org/linux/man-pages/man5/proc.5.html


sergeybykov:
Thank you, @mchandler-manticore!

Issue #3917, opened 2018-01-27
IStreamProviderFactory would be nice


JanEggers:
hi there,

it would be nice if there was a IStreamProviderFactory like IGrainFactory so you could write extensions against it that work on the Client and in the Grain.

```csharp
public interface IStreamProviderFactory 
{
        IStreamProvider GetStreamProvider(string name);
}

public interface IClusterClient : IStreamProviderFactory 
{
     ...
}

public abstract class Grain : IAddressable, ILifecycleParticipant<IGrainLifecycle>, IStreamProviderFactory 
    {
     ...
    }

```

yevhen:
Another interesting api option is to introduce StreamReference similar to GrainReference which encapsulates stream provider name and stream id and get rid of IStreamProviderFactory->IStreamProvider->GetStream ceremony. That will also make mocking more streamlined.

galvesribeiro:
Both very interesting suggestions... Indeed there is a big boilerplate at `OnActivateAsync()` in order to use streams...

xiazen:
I want to understand where the interface design came from. What problem it tries to solve? Does it come from the pain points of using streams? What are those pain points?  So I basically want to understand the problem, then visit the solution :) 

I think you mentioned one pain point of  "get rid of IStreamProviderFactory->IStreamProvider->GetStream ceremony". Is this method basically too long and annoy, or there's other paint points attached to it ?

xiazen:
Is it for making test mocking easier?

galvesribeiro:
> get rid of IStreamProviderFactory->IStreamProvider->GetStream ceremony

and

> making test mocking easier

Both are big annoying problems for the development experience. 

In other words, make it easier to mock/test and remove the boilerplate to get a stream reference.

galvesribeiro:
@xiazen the point is, why does the grain implementation need to know what provider it needs to use?

The grain code should be aware only about the stream id. The `GetStreamProvider()` call is totally not idiomatic. It could be registered at startup type a factory which gets injected to the grain and allow it to just `providerFactory.GetStream(streamId)`. 

Or even better. Don't need a factory to be injected at all. The stream reference could be directly injected as facet. 

The way @yevhen suggested, to be similar to GrainReference, will bring us more one dependency on the inheritance of `Grain`, which is what we are trying to avoid (so in a dream day, we would have POCO grains!).

[...]

PR #6446, opened 2020-03-30
Proof of concept: Remove Grain Base Requirement

KevinCathcart:
I present here a proof-of-concept for allowing grains to not inherit from the `Grain` base class. However, **this pull request is intended to get a conversation started** about this topic as much as it is a code suggestion. Even if none of this code gets used, I'm hopefull that the table of how to access the base class's functionality without without using the base class might be useful, as might the the other analysis in the message, or the following ones. 

This result allows for these classes to use basically all Orleans functionality with the exception of log-consistency/event-sourcing, since we do not have a facet version of those yet. This arguably is not yet fully POCO, since we still require the grain interface, and some functionality requires implementing specific interfaces.

## Philosophy

The goal here was implementing the simplest thing that works, and documenting how it is possible to inject all the needed functionality (and analyze where the current options appear suboptimal). I know in the meetup. Ruben talked about one of the benefits of POCO classes being reduced magic, and that certainly happens to an extent here, but there is still magic remaining, and arguably some additional magic added by this pull request.

Rather than trying to reduce that remaining magic further, this pull request simply does what is needed to keep that working. Instead, I will outline one way to a decent chunk  of the remaining magic, in a later post in this pull request.

## Approach
The approach used here is that every object that implements `IGrain` (except for the proxies derived from `GrainReference`) is a grain by definition. If manually constructed, it is a half-baked grain, that the runtime knows nothing about, but that is not really different from manually constructing a classic grain. 

This was surprising easy to accomplish, with the main complication being that we needed a way to get from a grain instance to its activation data. Beyond that, it was mostly changing `Grain` to `IGrain` in quite a few API, adding some checks that we were passed a grain and not a generated `GrainReference` proxy, and changing some code that used properties on the grain base class to get information via the existing extension methods instead. 

Handling activation data lookup is especially tricky, because we want to be able to access that data in static contexts, like the previously mentioned extension methods. For classic grains derived from `Grain`, I continue to use the existing `Data` field as a performance optimization. For other grains, I use a static `ConditionalWeakTable` for associating this data. (If anybody is not familiar, that class is designed to enable associating additional data to an arbitrary CLR reference type object, with the garbage collector effectively treating the data as being stored in a field of the object, rather than in the table itself.)

## Mapping of `Grain` base class functionality
**Note:** Some of these are not ideal, and are discussed more in the footnotes:

| Functionality | How To Use |
|---------------|------------|
| `this.GrainFactory` | Receive `IGrainFactory` as constructor parameter |
| State | use the `IPersistentState` facet |
| `this.IdentityString` | `this.GetGrainIdentity().IdentityString` |
| `this.RegisterTimer(...)` |  Receive `ITimerRegistry` as constructor parameter |
| `this.RegisterOrUpdateReminder(...)`<br>`this.UnregisterReminder(...)`<br>`this.GetReminder(...)`<br>`this.GetReminders()` |  Receive `IReminderRegistry` as constructor parameter, and use its methods [0] |
| `this.GetStreamProvider(name)` | See Footnote [1] |
| `this.DeactivateOnIdle()`<br>`this.DelayDeactivation(timeSpan)` | Receive `IGrainRuntime` as constructor parameter and use its methods [2] 
| overriding `OnActivateAsync`<br>overriding `OnDeactivateAsync` | Implement `ILifecycleParticipant<IGrainLifecycle>` [3] |
| `this.RuntimeIdentity` | Receive `IGrainRuntime` as constructor parameter and use `SiloIdentity` property [2] |
| `this.ServiceProvider` | Receive `IServiceProvider` as constructor parameter |
| `this.GrainReference` | `this.AsWeaklyTypedReference` [4] |

**Footnotes**:

**[0]:** Note that some checks from the `Grain` base class may be missing, like the check that grain implements `IRemindable` for registering a reminder. We might want to migrate the check?
**[1]:** Currently this requires either injecting `IClusterClient` (which would be overkill), or injecting `IServiceProvider` and calling `serviceProvider.GetRequiredServiceByName<IStreamProvider>(name)`. For the latter, users should probably not need to know about the named registrations. A nicer option might be something like an `IStreamProviderFactory` (#3917).
**[2]:** `IGrainRuntime` feels too low level to recommend users inject, so introducing some new user friendly injectable or facet for this would make better sense.
**[3]:** There are a bunch of alternative ways we can implement this which may be nicer for users, like an interface with those two methods that the grain can optionally implement, and have `GrainCreator` wire them to the lifecycle if it does implement the interface. 
**[4]:** I had to make that public, as a few places in the existing code base need to get the untyped Grain Reference for a grain, but did not have access to internals.

### DIFF
--- a/src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorageFactory.cs
+++ b/src/Azure/Orleans.Transactions.AzureStorage/TransactionalState/AzureTableTransactionalStateStorageFactory.cs
@@ -51,7 +51,7 @@ namespace Orleans.Transactions.AzureStorage
 
         private string MakePartitionKey(IGrainActivationContext context, string stateName)
         {
-            string grainKey = context.GrainInstance.GrainReference.ToShortKeyString();
+            string grainKey = context.GrainInstance.AsWeaklyTypedReference().ToShortKeyString();
             var key = $"{grainKey}_{this.clusterOptions.ServiceId}_{stateName}";
             return AzureTableUtils.SanitizeTableProperty(key);
         }
--- a/src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
+++ b/src/Orleans.CodeGeneration/GrainReferenceGenerator.cs
@@ -336,7 +336,11 @@ namespace Orleans.CodeGenerator
             {
                 return
                     SF.ConditionalExpression(
-                        SF.BinaryExpression(SyntaxKind.IsExpression, argIdentifier, typeof(Grain).GetTypeSyntax()),
+                        SF.BinaryExpression(SyntaxKind.LogicalAndExpression,
+                            SF.BinaryExpression(SyntaxKind.IsExpression, argIdentifier, typeof(IGrain).GetTypeSyntax()),
+                            SF.PrefixUnaryExpression(SyntaxKind.LogicalNotExpression,
+                                SF.ParenthesizedExpression(
+                                    SF.BinaryExpression(SyntaxKind.IsExpression, argIdentifier, typeof(GrainReference).GetTypeSyntax())))),
                         SF.InvocationExpression(argIdentifier.Member("AsReference", arg.ParameterType)),
                         argIdentifier);
             }
--- a/src/Orleans.CodeGenerator/Compatibility/OrleansLegacyCompat.cs
+++ b/src/Orleans.CodeGenerator/Compatibility/OrleansLegacyCompat.cs
@@ -167,11 +167,16 @@ namespace Orleans.CodeGenerator.Compatibility
             if (type.TypeKind != TypeKind.Class) return false;
 
             var orig = type.OriginalDefinition;
-            return HasBase(orig, types.Grain) && !IsMarkerType(types, orig);
+            return Implements(orig, types.IGrain) && !IsMarkerType(types, orig) && !HasBase(orig, types.GrainReference);
 
             bool IsMarkerType(WellKnownTypes l, INamedTypeSymbol t)
             {
                 return SymbolEqualityComparer.Default.Equals(t, l.Grain) || SymbolEqualityComparer.Default.Equals(t, l.GrainOfT);
+            }
+
+            bool Implements(INamedTypeSymbol symbol, ITypeSymbol type)
+            {
+                return symbol.AllInterfaces.Any(i => SymbolEqualityComparer.Default.Equals(type,i));
             }
 
             bool HasBase(INamedTypeSymbol t, INamedTypeSymbol baseType)
--- a/src/Orleans.CodeGenerator/Generators/GrainReferenceGenerator.cs
+++ b/src/Orleans.CodeGenerator/Generators/GrainReferenceGenerator.cs
@@ -269,7 +269,11 @@ namespace Orleans.CodeGenerator.Generators
                 {
                     return
                         ConditionalExpression(
-                            BinaryExpression(SyntaxKind.IsExpression, identifier, wellKnownTypes.Grain.ToTypeSyntax()),
+                            BinaryExpression(SyntaxKind.LogicalAndExpression,
+                                BinaryExpression(SyntaxKind.IsExpression, identifier, wellKnownTypes.Grain.ToTypeSyntax()),
+                                PrefixUnaryExpression(SyntaxKind.LogicalNotExpression,
+                                    ParenthesizedExpression(
+                                        BinaryExpression(SyntaxKind.IsExpression, identifier, wellKnownTypes.GrainReference.ToTypeSyntax())))),
                             InvocationExpression(identifier.Member("AsReference".ToGenericName().AddTypeArgumentListArguments(arg.Type.ToTypeSyntax()))),
                             identifier);
                 }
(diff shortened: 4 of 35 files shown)


KevinCathcart:
Some additional notes/thoughts:

1. I have included an additional commit that changes a few of the grains in the test suite to not derive from `Grain` as proof that this functionality works.

2. There are some new test failures for the `ClientAddressableTests`. These tests used "pseudo-grains" (classes implementing a grain interface, without deriving from `Grain`) rather than observers. Under the new definition these are half-baked grains, rather than pseudo-grains. On the other hand, `OutsideRuntimeClient` includes code to disallow creating local object references to grains. Combined this caused test failure.

   Since I'm not sure we really want to support client addressables derived from actual grain interfaces, I've changed the psuedo-grains to derive from `IAddressable`. I'd be happy to change this in some other way, if for some reason we really do want client addressables that claim to be grains.

3. This does not seem to meaningfully conflict with Reuben's WIP Grain Id changes, which is a good sign.

4. One important limitation is that this Pull Request currently does not support mocking the Grain Identity for unit testing for these new POCO-ish grains. If getting the grain identity information occured via DI rather than extension methods, then this would be trivial.

KevinCathcart:
# More radical proposal for further reducing magic

I almost hesitate to propose this, given how disruptive it could be to user code, but if reducing magic is a goal, this seems like one reasonable way to approach it:

## The major change

Get rid of the magical extension methods from `GrainExtensions`. As a reminder, the functionality provided includes the following groups that will be discussed seperately:
* `Cast`/`AsReference`/`AsWeaklyTypeReference`/`BindGrainReference`
* Grain Identity and Key related extensions
* InvokeOneWay

### GrainReference manipulation methods.
The `Cast`, `AsReference` and `AsWeaklyTypedGrainReference` functionality would be moved to the GrainFactory interface, and be viewed as constructing a new GrainReference that points to the passed in addressible. For efficientcy reasons, it might not always construct a new one. Some of this functionality would likely get exposed
on some internal interfaces, for the benefit of internal users of the functionality. (Note: Interestingly `Cast<>()` already eventually delegates to the Grain factory anyway.)

Cast functionality would also be provided directly on the `GrainReference` base class, so if you know for sure that an `IAddressable` is a strongly typed grain reference, you can cast/pattern match it to `GrainReference`, and invoke it's .Cast<>` method to avoid needing to have a GrainFactory handy.

As for `BindGrainReference`: that could remain an extension, but I'm not sure there is much point, as it simply delegates to the same method on the passed in GrainFactory.

### Grain Identity and Key extensions

For API compat for by far the most common use case of these, they could be implemented on the `Grain` base class. 

For POCO style grains, a new scoped injectable interface would provide access to the grain ID (and perhaps other relavant IDs as well, like ActivationID, and the SiloIdentity string?). This fixes the lack of mocking support for Grain Identity in the proof of concept, since mocking of this new interface would support that.

For cases where you are passed an arbitrary IGrain/IAddressable, and want the grain information, a new extension on `IAddressable` would be be available, that also takes an IGrainFactory, and basically does `return grainFactory.GetWeaklyTypedReference(addressable).GrainId;`. Nothing magical about that extension method. If you know for sure that an `IAddressable` is a strongly typed grain reference, you can cast/pattern match it to `GrainReference`, and access its GrainIdentity property directly to avoid needing to have a GrainFactory handy.

### InvokeOneWay
Honestly, this one could remain as an extension method, but that throws if the IAddressible passed is not already a GrainReference. Invoking this directly on a real instance of a grain is unlikely to be a particularly common scenario. 

## Impact on this pull request

I believe with the above changes, the mapping from a grain object reference to its Activation data would no longer need to be static, and could live inside some runtime component. If tied in properly it could even be possible to use a normal `ConcurrentDictionary` with a `ReferenceEquals based `IEqualityComparer`, rather than a ConditonalWeakTable.

Other than that, this is mostly just additional changes on top of what this pull request contains.

KevinCathcart:
Looking closely at what I needed to change to completely eliminate `Grain` from the the public API surface (except as a base class), shows that it is not very much. (According to the diffstat only 64 existing lines were deleted or changed in the main commit, which also includes changes in the non-public APIs. Obviously that is not counting any newly added lines.)

Very few public API's actually take the `Grain` type as input, and even fewer supply it as output. The public API where it is visible as an output output case is `IGrainActivationContext`'s GrainInstance property. And yeah, most of those that do take it as input could be reasonably changed to take a public `IGrainContext` instead.

The interesting cases are the API's that currently accept an IAddressable, and assume the instance provided is one of: GrainReference/subclass, ISystemTarget, or the Grain base class. In the latter scenario, the APIs are usually trying to get either the `GrainId`, or the `GrainReference`.

But obviously APIs that take `IAddressable` generally cannot be changed to take an `IGrainContext` specifically because they also want to accept GrainReference proxies. It is those APIs that necessitate the need to lookup IActivationData.

ReubenBond:
I'll close this PR for now. It's a good proof of concept (thank you, Kevin), but I don't think we can merge it in this shape and the main branch has changed significantly enough that it would likely only be used as a reference for building the final implementation

KevinCathcart:
Yeah. Obviously at this point it would need to be done differently anyway.
I probably should have closed it myself a fair while back, but pretty much
just forgot about it.

ReubenBond:
Thanks, Kevin :) I think a change like this can be implemented for real now, too: the new activation system is in a place where I feel happy with it - especially once the open PRs are merged

tomachristian:
@ReubenBond what should we be using instead of `InvokeOneWay`? Sorry if this has been answered somewhere else.

PR #6460, opened 2020-04-06
Consistently sanitize properties for Azure Table Storage reminders implementation

### LINKED ISSUE #6449: Inconsistent property sanitization in Azure Table Storage reminders
ServiceId is not always santized. For example, see: https://github.com/dotnet/orleans/blob/205ebc3b2c175ebe6b5d7e887697b85e7f184766/src/Azure/Orleans.Reminders.AzureStorage/Storage/RemindersTableManager.cs#L112

ReubenBond:
Fixes #6449

### DIFF
--- a/src/Azure/Orleans.Reminders.AzureStorage/Storage/RemindersTableManager.cs
+++ b/src/Azure/Orleans.Reminders.AzureStorage/Storage/RemindersTableManager.cs
@@ -25,8 +25,14 @@ namespace Orleans.Runtime.ReminderService
 
         public static string ConstructRowKey(GrainReference grainRef, string reminderName)
         {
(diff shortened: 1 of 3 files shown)


src/Azure/Orleans.Reminders.AzureStorage/Storage/RemindersTableManager.cs
```diff
@@ -219,31 +233,24 @@ internal async Task<bool> DeleteReminderEntryConditionally(ReminderTableEntry re
 
         internal async Task DeleteTableEntries()
         {
-            if (ServiceId.Equals(Guid.Empty) && ClusterId == null)
```
ReubenBond, line 222: Comparing Guid & string using .Equals... ClusterId is not allowed to be null

PR #6467, opened 2020-04-07
Setting eventIndex

adyada:
Setting EventIndex in FromQueueMessage method. In our use case we get a message from eventhub and then duplicate that message to go to multiple grains, while duplicating we set different EventIndex for each message. 

So, setting EventIndex in FromQueueMessage method will cause us to copy and modify one less function from orleans code to support our use case, if this doesn't cause any other issue in orleans

### DIFF
--- a/src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubDataAdapter.cs
+++ b/src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubDataAdapter.cs
@@ -67,6 +67,7 @@ namespace Orleans.ServiceBus.Providers
                 StreamGuid = streamPosition.StreamIdentity.Guid,
                 StreamNamespace = streamPosition.StreamIdentity.Namespace != null ? string.Intern(streamPosition.StreamIdentity.Namespace) : null,
                 SequenceNumber = queueMessage.SystemProperties.SequenceNumber,
+                EventIndex = streamPosition.SequenceToken.EventIndex,
                 EnqueueTimeUtc = queueMessage.SystemProperties.EnqueuedTimeUtc,
                 DequeueTimeUtc = dequeueTime,
                 Segment = EncodeMessageIntoSegment(queueMessage, getSegment)


jason-bragg:
Can you plz run the functional eventhub tests on this PR.  CI doesn't run those.
To do this you'll need to add a secrets file to the folder the orleans tree is in.

For example if the orleans code is in C:/user/bob/code/orleans, the secrets file would need be in C:/user/bob/code/

The secrets file is a json file as below:
{
  "EventHubConnectionString": "Endpoint=<your eventhub secret>"
} 

The hub setup for the testing needs have 4 partitions, a path of  "ehorleanstest", and a consumer group named "orleansnightly"

Thanks

---

As for the described scenario, orleans streaming does not use the sequence token (or EventIndex) in event routing, only the streamId, so what you're describing "shouldn't" work unless you're doing some custom routing somehow..?  I suggest building custom orleans nugets with this change and testing them in your environment to ensure this will do what you intended.

sergeybykov:
I kicked off a functional run in ADO - https://orleans.visualstudio.com/Orleans/_build/results?buildId=33370&view=results

adyada:
@jason-bragg  we also do some other customizations/hacks to make that scenario work.  We map event indexes to different stream Ids.  In case you are interested, you might have access to this link internally. 

https://msdata.visualstudio.com/Vienna/_git/vienna?path=%2Fsrc%2Fazureml-api%2Fsrc%2FCommon%2FOrleans%2FStreaming

This has worked for us so far without much problems.  To make this scenario work, I had to copy and modify some code from orleans, so this PR reduces one such function that I would need to copy. 

We don't need to necessarily merge this PR. I am happy copying and modifying that function :) . But this was just one line change, so I thought to merge it in the orleans repo if it doesn't do any other harm.

sergeybykov:
Failed | 0
-- | --
Passed | 2543
Others | 528
Total tests | 3071

jason-bragg:
Thanks for contributing  @adyada

adyada:
Thanks for merging this PR

PR #6473, opened 2020-04-09
Do not try to group grain directory unregistration calls.

benjaminpetit:


src/Orleans.Runtime/Catalog/ActivationData.cs
```diff
@@ -586,8 +586,13 @@ public void ResetKeepAliveRequest()
                     OnInactive = new List<Action>();
                 }
                 OnInactive.Add(action);
+                if (!IsCurrentlyExecuting)
+                {
+                    RunOnInactive();
```
ReubenBond, line 591: Locks are reentrant, so it's ok, but the lock will be re-acquired by this call. Unsure what the cost of that is, if at all significant. An alternative would be an "*UnderLock" or "Unsafe*" alternative without a lock, to signify that the lock has already been taken.

src/Orleans.Runtime/Catalog/Catalog.cs
```diff
@@ -746,18 +744,24 @@ public bool TryGetActivationData(ActivationId activationId, out ActivationData d
             return data != null;
         }
 
-        private Task DeactivateActivationsFromCollector(List<ActivationData> list)
+        private async Task DeactivateActivationsFromCollector(List<ActivationData> list)
         {
+            var cts = new CancellationTokenSource(this.collectionOptions.Value.DeactivationTimeout);
+            var taskCompletionSourceList = new TaskCompletionSource<object>[list.Count];
```
ReubenBond, line 750: Does it make sense to use the `MultiTaskCompletionSource` here instead? It ought to be cheaper/faster, but I don't know if there are semantics from having individual TCS' which would help

src/Orleans.Runtime/Catalog/Catalog.cs
```diff
@@ -1363,20 +1202,27 @@ public SiloStatus LocalSiloStatus
...
+            foreach (var activationData in TryGetActivationDatas(addresses))
             {
-                ActivationData data;
-                if (TryGetActivationData(activationAddress.Activation, out data))
-                    datas.Add(data);
+                DestroyActivationAsync(activationData, taskCompletionSourceList[counter], timeoutTokenSource.Token);
+                counter++;
             }
-            return datas;
+            return Task.WhenAll(taskCompletionSourceList.Select(tcs => tcs.Task));
```
ReubenBond, line 1225: Why not make DestroyActivationAsync true async (`async Task`, not `async void`) and eliminate the TCS, waiting on the true task instance?

Also, wont `taskCompletionSourceList[counter]` throw an index out of bounds exception?

PR #6474, opened 2020-04-10
Send rejections for messages enqueued on stopped outbound queue

ReubenBond:
We shouldn't drop messages, unless they are expired or we have no reasonable action to take on them.

In this case, I think we can reject messages sent to a stopped OutboundMessagingQueue (which happens during shutdown), to make some problems easier to debug.

### DIFF
--- a/src/Orleans.Runtime/Messaging/OutboundMessageQueue.cs
+++ b/src/Orleans.Runtime/Messaging/OutboundMessageQueue.cs
@@ -44,11 +44,12 @@ namespace Orleans.Runtime.Messaging
 
         public void SendMessage(Message msg)
         {
-            if (msg == null) throw new ArgumentNullException("msg", "Can't send a null message.");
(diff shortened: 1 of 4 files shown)


src/Orleans.Runtime/Messaging/OutboundMessageQueue.cs
```diff
@@ -82,34 +84,23 @@ public void SendMessage(Message msg)
             {
                 if (stopped)
                 {
-                    logger.Info(ErrorCode.Runtime_Error_100115, "Message was queued for sending after outbound queue was stopped: {0}", msg);
+                    logger.LogInformation((int)ErrorCode.Runtime_Error_100115, "Message was queued for sending after outbound queue was stopped: {Message}", msg);
+                    messageCenter.SendRejection(msg, Message.RejectionTypes.Unrecoverable, "Message was queued for sending after outbound queue was stopped");
```
ReubenBond, line 88: This is the important line (and 52, which is identical). The other changes are cleanliness changes.

PR #6479, opened 2020-04-10
Clean up some parts of IGrainTypeResolver and related

ReubenBond:


src/Orleans.Core/Runtime/Legacy/GrainTypeResolver.cs
```diff
@@ -1,36 +1,37 @@
...
         private readonly Dictionary<string, GrainInterfaceData> typeToInterfaceData;
+
+#pragma warning disable IDE0051 // Remove unused private members
+#pragma warning disable CS0169 // Remove unused private members
+        // Unused. Retained for serializer compatibility.
         private readonly Dictionary<int, GrainInterfaceData> table;
+
+        // Unused. Retained for serializer compatibility.
         private readonly HashSet<string> loadedGrainAsemblies;
+#pragma warning restore CS0169 // Remove unused private members
```
jdom, line 25: are we still using the non-back compat serializer for the internal protocol?

src/Orleans.Runtime/GrainTypeManager/GenericGrainTypeData.cs
```diff
@@ -1,27 +0,0 @@
-using System;
-using System.Reflection;
-
-namespace Orleans.Runtime
-{
-    [Serializable]
-    internal class GenericGrainTypeData : GrainTypeData
```
jdom, line 7: why was this serializable? I assume it was an oversight and wasn't really being used as that. But just a heads up in case it was really being sent over the wire. In that case the change wouldn't be back compat.

jdom (approved):

src/Orleans.Core/Runtime/Legacy/GrainTypeResolver.cs
ReubenBond: lol

ReubenBond: My hope is that we can change this in 4.0, since we now support network protocol negotiation and can hence add a new core serialization format.

src/Orleans.Runtime/GrainTypeManager/GenericGrainTypeData.cs
ReubenBond: I think marking these as serializable was an oversight. I removed the attribute from the base class.

My plan is to move to a more stringly/open-ended format for exchanging type & other metadata for 4.0.

PR #6482, opened 2020-04-11
Streamline LINQ/Enumerable use

pentp:


pentp:
Why is `string.Equals` better than just a normal comparison? And why add `StringComparison.Ordinal` when it's the default?

ReubenBond:
Sometimes a refactoring will mean that `==` no longer resolves to `string`'s implementation, so we should be explicit. One such case was fixed last week.

As for `StringComparison.Ordinal`, the primary reason is just to make sure the first point remains true (because `string.Equals(a, b)` is defined for `object` inputs). See best practices here: https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

PR #6483, opened 2020-04-12
Stopped WorkItemGroup logging enhanchement

pentp:
This tries to add some more context to the warning about incorrect use of a stopped scheduler.
This will turn `Task.AsyncState: System.Action` into something like `Task.AsyncState: ContinuationWrapper.Invoke: MoveNextRunner.Run: DummyTest+<RunSomethingAsync>d__1`.

Not sure if this is the reasonable way to get this data, maybe @stephentoub knows how to do this more cleanly? Using `Task.GetDelegatesFromContinuationObject(task.AsyncState)` maybe?


ReubenBond:
Great, thanks!

PR #6491, opened 2020-04-16
Remove ObserverId from GrainReference, and add ClientGrainId & ObserverGraindId

ReubenBond:


src/Orleans.Core.Abstractions/IDs/ClientGrainId.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        /// <summary>
+        /// Creates a new <see cref="ClientGrainId"/> instance.
+        /// </summary>
+        public static ClientGrainId Create() => Create(Guid.NewGuid().ToString("N"));
+
+        /// <summary>
+        /// Creates a new <see cref="ClientGrainId"/> instance.
+        /// </summary>
+        public static ClientGrainId Create(string id) => Create(IdSpan.Create(id));
```
benjaminpetit, line 29: Do we plan to enable the client name to be configurable? It might be dangerous.

ReubenBond: I'm not sure. It can be useful for the hosted client case.

PR #6493, opened 2020-04-18
Less allocations on serialization manager

ThiagoT1:
The point is to avoid allocating an array just for the lengths of the array being (de)serialized, unless we have to.


sergeybykov:
Thank you, @ThiagoT1!

PR #6496, opened 2020-04-20
Fix typo in Adventure sample

atifaziz:
This PR fixes a small typo I noticed while studying the source of the Adventure sample.


sergeybykov:
Thank you, @atifaziz!

PR #6503, opened 2020-04-22
New type metadata system

ReubenBond:
This PR introduces a new type metadata system for Orleans and demonstrates its usage by integrating it with the Placement & Directory systems.

In order for clients and silos to communicate and coordinate effectively, they need to know some information about each other. This information includes:

* Which grains are available on each silo
* Which grain interfaces are available on each silo

For each of those grains, Orleans needs to know (among other things):

* Which placement strategy it uses, which may be parameterized (eg, StatelessWorker has a MaxLocal property)
* Which grain locator/directory it uses
* Which interfaces it implements

For each grain interfaces, Orleans needs to know:

* The version of the interface loaded by the silo
* The primary implementation of the interface

# Motivation

The current implementation of the Type Metadata system is strongly-typed.
This strong typing adds some rigidity and extensibility problems which are burdensome for a system which needs to support smooth rolling upgrades. For example, adding a new `PlacementStrategy` would require a two-stage rollout today (otherwise, there could be serialization errors which would prevent smooth operation).

Today, grain and grain interfaces are identified using integer type codes (essentially, a hash of the type, but the value can be overridden). These type codes are not human readable and do not align with the new GrainId system, which uses stringly identifiers.

# Design

This Type Metadata system introduced in this PR intends to solve some of the problems and provide a much more extensible system for building features reliant on that type metadata.

At the core, the system is comprised of `string` to `string` dictionaries which describe the properties of grains and grain interfaces. Each silo compiles a manifest of the grains and grain interfaces it has locally, and shares that manifest with other silos to create a cluster manifest. This cluster manifest is also shared with the client.

The cluster manifest is then used to direct individual systems. For example, a `PlacementStrategyResolver`, which is responsible for identifying the `PlacementStrategy` which a given grain uses, reads the `placement-strategy` property from the cluster manifest entry corresponding to the grain type.

In the case of placement, the system is architecturally organized like so:

* `PlacementService` is the central point for placement decisions
* It uses `PlacementStrategyResolver` to get the `PlacementStrategy` for a grain
* It passes that strategy to `PlacemetnDirectorResolver` to find the correct `IPlacementDirector`
* It calls the `IPlacementDirector` using that `PlacementStrategy` to make placement decisions.

A similar layout is used for the grain locator/directory.

# Detail

There are three major stages to the Type Metadata system:

* Compiling local type metadata (`SiloManifest`)
* Discovering type metadata for other silos (`ClusterManifest`)
* Consuming type metadata in subsystems (`GrainPropertiesResolver`)

To aid in understanding, we will look at those stages one-by-one

## Compiling Type Metadata

`SiloManifestProvider` is the central type here. It is constructed through dependency injection and combines application parts with configured providers.

Application parts consist primarily of assemblies and pre-configured helpers which analyze those assemblies (relying on generated code) to provide a list of grain classes & grain interfaces.

The configured providers, of type `IGrainPropertiesProvider` and `IGrainInterfacePropertiesProvider`, and invoked using each of the grain or interface types (respectively) and are responsible for updating a `Dictionary<string, string>` for grains and grain interfaces.

Implementations of these providers are configured which examine attributes. Any attribute which adorns a grain/interface type and which implements the `IGrainPropertiesProviderAttribute`/`IGrainInterfacePropertiesProviderAttribute`  interface will have the chance to populate metadata for that type.

For example, here is the full implementation of `GrainDirectoryAttribute`:

``` C#
[AttributeUsage(AttributeTargets.Class)]
public sealed class GrainDirectoryAttribute : Attribute, IGrainPropertiesProviderAttribute
{
    public const string DEFAULT_GRAIN_DIRECTORY = "default";

    public string GrainDirectoryName { get; set; }

    public GrainDirectoryAttribute()
    {
        this.GrainDirectoryName = DEFAULT_GRAIN_DIRECTORY;
    }

    public void Populate(IServiceProvider services, Type grainClass, GrainType grainType, Dictionary<string, string> properties)
    {
        properties[WellKnownGrainTypeProperties.GrainDirectory] = this.GrainDirectoryName ?? DEFAULT_GRAIN_DIRECTORY;
    }
}
```

This attribute will propagate the value of its `GrainDirectoryName` property to the `grain-directory` key on the `properties` dictionary.

A similar implementation exists for `PlacementStrategyAttribute` with some additional indirection caused by the presence of the `PlacementStrategy` type.

Once constructed, each `SiloManifest` is immutable and a silo never updates its manifest.

## Discovering Type Metadata for the cluster

Silo manifests created in the previous stage are synchronized around the cluster to form a `ClusterManifest` on each silo and client. The mechanism by which this happens is different for silos and clients.

In `ClusterManifestProvider`, silos listen for membership updates from `ClusterMembershipService`. Each time a silo joins or leaves the cluster, the provider attempts to update its local `ClusterManifest`, adding or removing an entry. When a silo joins, the provider sends a request to a `SystemTarget` on that silo to get its `SiloManifest`. If this request fails, it will be retried indefinitely (with a delay) until it succeeds or that silo leaves the cluster.

The updated `ClusterManifest` is then made available to consumers.

## Consuming Type Metadata

The primary means for consuming type metadata is via the `GrainPropertiesResolver`, which maps from `GrainType` to the `GrainProperties` instance which contains the `string`-to-`string` dictionary of metadata.

Systems which rely on these properties (for example, `PlacementStrategyResolver`) will therefore call into this service as a part of their normal resolution process, and cache their final result (eg, the `PlacementStrategy`, in this case).


src/Orleans.Core.Abstractions/Manifest/GrainInterfaceId.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Diagnostics;
+
+namespace Orleans.Metadata
+{
+    /// <summary>
+    /// Uniquely identifies a grain interface.
+    /// </summary>
+    [Serializable]
+    [DebuggerDisplay("{Value}")]
+    public readonly struct GrainInterfaceId : IEquatable<GrainInterfaceId>
```
benjaminpetit, line 11: IMO naming is too similar to `GrainId`. What about `GrainInterfaceTypeCode` or something like that?

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
```diff
@@ -283,6 +285,23 @@ internal static void AddDefaultServices(IApplicationPartManager applicationPartM
...
+            services.AddSingleton<SiloManifest>(sp => sp.GetRequiredService<SiloManifestProvider>().SiloManifest);
+            services.AddSingleton<Metadata.GrainTypeResolver>();
+            services.AddSingleton<Metadata.IGrainTypeProvider, LegacyGrainTypeResolver>();
+            services.AddSingleton<GrainPropertiesResolver>();
+            services.AddSingleton<GrainInterfaceIdResolver>();
+            services.AddSingleton<Metadata.IGrainTypeProvider, AttributeGrainTypeProvider>();
+            services.AddSingleton<IGrainInterfacePropertiesProvider, AttributeGrainInterfacePropertiesProvider>();
+            services.AddSingleton<IGrainPropertiesProvider, AttributeGrainPropertiesProvider>();
+            services.AddSingleton<IGrainInterfacePropertiesProvider, DiagnosticInfoGrainPropertiesProvider>();
+            services.AddSingleton<IGrainPropertiesProvider, DiagnosticInfoGrainPropertiesProvider>();
```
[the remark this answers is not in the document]
ReubenBond: It won't make much difference, but we could

src/Orleans.Core.Abstractions/Runtime/GrainReference.cs
```diff
@@ -14,7 +14,6 @@ namespace Orleans.Runtime
     public class GrainReference : IAddressable, IEquatable<GrainReference>, ISerializable
     {
         private readonly string genericArguments;
-        private readonly GuidId observerId;
```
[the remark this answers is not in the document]
ReubenBond: Yep

src/Orleans.Runtime/Hosting/DefaultSiloServices.cs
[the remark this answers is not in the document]
KevinCathcart: yeah, feels rather 6 and one-half-dozen to me. Makes no real difference which way you say it.

src/Orleans.Core.Abstractions/IDs/GrainId.cs
```diff
@@ -15,6 +15,8 @@ namespace Orleans.Runtime
     [StructLayout(LayoutKind.Auto)]
     public readonly struct GrainId : IEquatable<GrainId>, IComparable<GrainId>, ISerializable
     {
+        private static readonly char[] SegmentSeparator = new[] { '/' };
```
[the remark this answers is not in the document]
ReubenBond: Everything will clash. I don't see a reason not to use this. Anything will need to be escaped if it's going to be used in a URL, etc, especially considering the components themselves currently don't have any restriction on what they can hold

src/Orleans.Core.Abstractions/Manifest/GrainInterfaceId.cs
ReubenBond: I'm not a fan of the TypeCode name, especially the "code" part, but I'm open to suggestions. It could be `GrainInterfaceDescriptor` if you prefer.

src/Orleans.Core.Abstractions/IDs/GrainId.cs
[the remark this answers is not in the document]
KevinCathcart: I think you understood what i was trying to say backwards.

Currently, this code when parsing will stop at the first slash, whether it was the one ToString inserted or not.

If a user included the slash in a GrainType, then the provided code will incorrectly parse the result. To avoid that would require escaping the GrainType in the ToString method, and unescaping in Parse, or (alternatively) forbidding forward slash from being part of GrainType.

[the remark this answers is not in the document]
ReubenBond: Agreed, I'm thinking of adding input sanitation soon, but not just yet. Feel free to PR them if you'd like

ReubenBond:
Open questions:

* Should we ensure that joining silos have sync'd ClusterManifest?

ReubenBond:
@benjaminpetit I removed the WIP tag. Please take a look and merge when ready

src/Orleans.Core.Abstractions/Placement/PlacementStrategy.cs
```diff
@@ -16,5 +22,14 @@ public abstract class PlacementStrategy
...
         /// </summary>
         internal virtual bool IsDeterministicActivationId => false;
+
+        public virtual void Initialize(GrainProperties properties)
+        {
+        }
+
+        public virtual void PopulateGrainProperties(IServiceProvider services, Type grainClass, GrainType grainType, Dictionary<string, string> properties)
+        {
+            properties[WellKnownGrainTypeProperties.PlacementStrategy] = this.GetType().Name;
```
benjaminpetit, line 32: Do you see a use case where we would never set this property? Not a lot of people will implement their own `PlacementStrategy`, but I assume it will be easy to overlook calling the base method.

Not very important anyway.

src/Orleans.Core/Manifest/DiagnosticInfoGrainPropertiesProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+        public void Populate(Type grainClass, GrainType grainType, Dictionary<string, string> properties)
+        {
+            properties["diag.type"] = RuntimeTypeNameFormatter.Format(grainClass);
+            properties["diag.asm"] = grainClass.Assembly.GetName().Name;
+        }
+
+        public void Populate(Type interfaceType, GrainInterfaceId interfaceId, Dictionary<string, string> properties)
+        {
+            properties["diag.type"] = RuntimeTypeNameFormatter.Format(interfaceType);
+            properties["diag.asm"] = interfaceType.Assembly.GetName().Name;
```
benjaminpetit, line 19: really NIT: duplicate strings

src/Orleans.Runtime/GrainDirectory/GrainLocatorResolver.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                result = this.clientGrainLocator;
+            }
+            else if (this.grainDirectoryResolver.HasNonDefaultDirectory(grainType))
+            {
+                result = this.cachedGrainLocator;
+            }
+            else
+            {
+                result = this.dhtGrainLocator;
```
benjaminpetit, line 44: What if the default is not set to the `dhtGrainLocator`?

benjaminpetit (commented):
Only question I have really is the one in the `GrainLocatorResolver`

src/Orleans.Core.Abstractions/Placement/PlacementStrategy.cs
ReubenBond: Yeah, it's definitely possible. If a developer wants to override `PopulateGrainProperties` then that's fine, but caveat emptor, I suppose.

The stateless worker overrides this and adds another property

src/Orleans.Core/Manifest/DiagnosticInfoGrainPropertiesProvider.cs
ReubenBond: I figured these diagnostic properties didn't warrant an entry in the WellKnown* classes

ReubenBond: (because I don't want anyone to rely on them)

src/Orleans.Runtime/Hosting/Generic/SiloWrapper.cs
```diff
@@ -51,12 +51,21 @@ public async Task StopAsync(CancellationToken cancellationToken)
...
+
+        public async ValueTask DisposeAsync()
         {
             if (!isDisposing)
             {
                 this.isDisposing = true;
-                (this.Services as IDisposable)?.Dispose();
+                if (this.Services is IAsyncDisposable asyncDisposable)
+                {
+                    await asyncDisposable.DisposeAsync();
```
jdom, line 63: is this the typical behavior (to preserve ordering)? or should you parallelize `DisposeAsync()` calls?

jdom (commented):
I didn't do a full review, just eyeballed it a little bit. I assume this is using a wire serialization format that allows evolution?

src/Orleans.Core/Manifest/DiagnosticInfoGrainPropertiesProvider.cs
benjaminpetit: meh it was mostly to have some private string const in the class to avoid duplicate string in memory, but the compiler is probably clever enough

src/Orleans.Core/Manifest/GrainTypeResolver.cs
```diff
@@ -0,0 +1,76 @@
...
+                {
+                    return grainType;
+                }
+            }
+
+            // Conventions are used as a fallback
+            return GetGrainTypeByConvention(type);
+        }
+
+        private static GrainType GetGrainTypeByConvention(Type type)
```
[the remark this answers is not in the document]
ReubenBond: Nah, users can already provide a resolver if they want, so I think we don't need to go too crazy

src/Orleans.Runtime/GrainDirectory/GrainLocatorResolver.cs
ReubenBond: I think this is the same behavior as it is today. Either way, we can merge now & follow up

src/Orleans.Core/Manifest/DiagnosticInfoGrainPropertiesProvider.cs
ReubenBond: Yeah, it is - it will intern them

src/Orleans.Core.Abstractions/IDs/GrainId.cs
```diff
@@ -123,7 +160,9 @@ public int CompareTo(GrainId other)
         public static bool operator <(GrainId a, GrainId b) => a.CompareTo(b) < 0;
 
         /// <inheritdoc/>
-        public override string ToString() => $"{Type.ToStringUtf8()}/{Key.ToStringUtf8()}";
+        public override string ToString() => $"{Type.ToStringUtf8()}{SegmentSeparator[0]}{Key.ToStringUtf8()}";
```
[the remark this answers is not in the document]
ReubenBond: Fixed in branch, will include in next set of PRs

src/Orleans.Core.Abstractions/Manifest/ClusterManifest.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Immutable;
+using Orleans.Runtime;
+
+namespace Orleans.Metadata
+{
+    /// <summary>
+    /// Information about types which are available in the cluster.
+    /// </summary>
+    [Serializable]
+    public class ClusterManifest
```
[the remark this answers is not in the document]
ReubenBond: C# 9 features would come in very handy

src/Orleans.Runtime/Hosting/Generic/SiloWrapper.cs
ReubenBond: There's only one call here - the important thing is to call DisposeAsync preferentially and only fall back to Dispose if async isn't supported

PR #6505, opened 2020-04-23
Fix name of data service

jacob-l:
Previous name is LightstreamerDataService, but we register DataService in Silo. Looks like the correct name is DataService


sergeybykov:
Thank you, @jacob-l!

PR #6524, opened 2020-05-05
Simple accumulator to batch grain unregistration in DhtGrainLocator 

benjaminpetit:


src/Orleans.Runtime/GrainDirectory/DhtGrainLocator.cs
```diff
@@ -34,9 +40,64 @@ public async Task<ActivationAddress> Register(ActivationAddress address)
             => (await this.localGrainDirectory.RegisterAsync(address, singleActivation: true)).Address;
 
         public  Task Unregister(ActivationAddress address, UnregistrationCause cause)
-            => this.localGrainDirectory.UnregisterAsync(address, cause);
+        {
+            var tcs = new TaskCompletionSource<object>();
```
ReubenBond, line 44: We should add `TaskCreationOptions.RunContinuationsAsynchronously` - see https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

PR #6525, opened 2020-05-07
Do not reject rejection messages locally. Drop them instead

benjaminpetit:


src/Orleans.Runtime/Messaging/MessageCenter.cs
```diff
@@ -202,10 +202,20 @@ public bool TrySendLocal(Message message)
         internal void SendRejection(Message msg, Message.RejectionTypes rejectionType, string reason)
         {
             MessagingStatisticsGroup.OnRejectedMessage(msg);
-            if (string.IsNullOrEmpty(reason)) reason = string.Format("Rejection from silo {0} - Unknown reason.", MyAddress);
-            Message error = this.messageFactory.CreateRejectionResponse(msg, rejectionType, reason);
-            // rejection msgs are always originated in the local silo, they are never remote.
-            this.OnReceivedMessage(error);
+
+            if (msg.Result == Message.ResponseTypes.Rejection)
```
sergeybykov, line 206: Should we make it more explicit:  `if( msg.Direction == Message.Directions.Response && msg.Result == Message.ResponseTypes.Rejection)` even though the end result will be the same?

PR #6528, opened 2020-05-09
Grammar in Reentrancy docs.

bchavez:
>Grain activations are single-threaded and, by default, process each request from beginning to completion before the next request can begin ***being*** processing. In some circumstances, it may be desirable for ....

Removed ***being*** above.


sergeybykov:
Thank you, @bchavez!

PR #6543, opened 2020-05-21
Grain Directory Redis implementation

benjaminpetit:


src/Orleans.GrainDirectory.Redis/Options/RedisGrainDirectoryOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+        private readonly RedisGrainDirectoryOptions options;
+
+        public RedisGrainDirectoryOptionsValidator(RedisGrainDirectoryOptions options)
+        {
+            this.options = options;
+        }
+
+        public void ValidateConfiguration()
+        {
+        }
```
ReubenBond, line 24: Is this missing an implementation?

src/Orleans.GrainDirectory.Redis/Options/RedisGrainDirectoryOptions.cs
```diff
@@ -21,6 +39,10 @@ public RedisGrainDirectoryOptionsValidator(RedisGrainDirectoryOptions options)
 
         public void ValidateConfiguration()
         {
+            if (this.options == null)
```
ReubenBond, line 42: Can this ever be null? We should be using IOptions<T>, not just T, right?

Then, we can check the `.ConfigurationOptions` isn't null, instead

src/Orleans.GrainDirectory.Redis/RedisGrainDirectory.cs
```diff
@@ -43,7 +51,12 @@ public async Task<GrainAddress> Register(GrainAddress address)
 
         public async Task Unregister(GrainAddress address)
         {
-            await this.database.ScriptEvaluateAsync(this.deleteScript, new { key = address.GrainId, val = JsonConvert.SerializeObject(address) });
+            var key = GetKey(address.GrainId);
+
+            var tx = this.database.CreateTransaction();
+            tx.AddCondition(Condition.StringEqual(key, JsonConvert.SerializeObject(address)));
+            tx.KeyDeleteAsync(key).Ignore();
+            await tx.ExecuteAsync();
```
ReubenBond, line 59: I believe we need to check the return value and potentially the condition: https://github.com/OrleansContrib/Orleans.Redis/blob/15b63ca48f11b715c6d05da1a60e85c9f8812247/src/Orleans.Clustering.Redis/RedisMembershipTable.cs#L75

ReubenBond: Although, I suppose in this case we don't care if it actually was deleted

benjaminpetit: Maybe we can log it if we want? The current implementation has no log at all, it might be useful in the future to have some

src/Orleans.GrainDirectory.Redis/Hosting/RedisGrainDirectoryExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Orleans.Configuration;
+using Orleans.GrainDirectory;
+using Orleans.GrainDirectory.Redis;
+using Orleans.Runtime;
+
+namespace Orleans.Hosting
+{
+    public static class RedisGrainDirectoryExtensions
```
ReubenBond, line 11: We probably want to add some doc comments on these methods before final

src/Orleans.GrainDirectory.Redis/Options/RedisGrainDirectoryOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class RedactRedisConfigurationOptions : RedactAttribute
+    {
+        public override string Redact(object value) => value is ConfigurationOptions cfg ? cfg.ToString(includePassword: false) : base.Redact(value);
+    }
+
+    public class RedisGrainDirectoryOptionsValidator : IConfigurationValidator
+    {
+        private readonly RedisGrainDirectoryOptions options;
+
+        public RedisGrainDirectoryOptionsValidator(RedisGrainDirectoryOptions options)
```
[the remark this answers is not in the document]
benjaminpetit: ` IOptionsMonitor<T>.Get` doesn't return an `IOptions<T>` but a `T` directly. We do the same for other named providers.

src/Orleans.GrainDirectory.Redis/Options/RedisGrainDirectoryOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// Configure the Redis client
+        /// </summary>
+        [RedactRedisConfigurationOptions]
+        public ConfigurationOptions ConfigurationOptions { get; set; }
+
+        /// <summary>
+        /// Entry expiry, null by default. A value should be set ONLY for ephemeral environments (like in tests).
+        /// Setting a value different from null will cause duplicate activations in the cluster.
+        /// </summary>
+        public TimeSpan? EntryExpiry { get; set; } = null;
```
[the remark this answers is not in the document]
benjaminpetit: If you are sure that your activations will not be active for more than the value you set here, maybe? But I think we should not advocate for that. This setting is useful for test environment, for the moment it's why it's there

PR #6545, opened 2020-05-21
Remove reflection-based code generator

ReubenBond:


ReubenBond:
This passed tests, but gh missed the notification

PR #6547, opened 2020-05-24
Make Deployment and Multi-Cluster Support documentation visible in sidebar

logiclabs:
toc.yml appears to be invalid (has duplicate items nodes), which seems to prevent Deployment and Multi-cluster support documentation links from being rendered.

### DIFF
--- a/src/Documentation/toc.yml
+++ b/src/Documentation/toc.yml
@@ -14,7 +14,6 @@
     href: index.md
   - name: Deployment
     href: deployment/toc.yml
-  items:
   - name: Multi-Cluster Support
     href: index.md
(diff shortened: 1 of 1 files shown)


sergeybykov:
Thank you, @logiclabs, for submitting this PR. 

I don't believe it does what you intended it to though. 

The other problem, unrelated to your PR, is that the `deployment` folder contains a bunch of outdated doc files that either need to be updated (e.g. for the current configuration API) or probably retired.

We removed support for multi-cluster in 3.2.0 and 4.0.0. So adding back those doc pages to the menu would also be confusing.

PR #6521, opened 2020-05-03
Support ValueTask as [OneWay] Methods Return Type

ThiagoT1:

PR #6549, opened 2020-05-24
Fix netstandard2.0 compatibility

pentp:
This restores `netstandard2.0` compatibility broken in #6521.

Also updated .editorconfig with two new rules and modified `this.` qualification rules to better match the reality of the project style.


ReubenBond:
Ah, thank you, @pentp

PR #6557, opened 2020-05-28
More graceful termination of network connections

ReubenBond:


jdom:
>                         this.Context.Abort();

won't this Abort cause non-graceful terminations? Since in the code above you just ask for everything to finish (asynchrnonously), dropping any enqueued work, but completing (asynchronously)their current work

---
Refers to: src/Orleans.Core/Networking/Connection.cs:204 in 78b590d. [](commit_id = 78b590d1e49726c41d64bd133d2a110896bb2659, deletion_comment = False)

ReubenBond:
@jdom do you know of another way?

jdom:
@ReubenBond I actually haven't checked the implementation of `Abort`, nor know how `CancelPendingRead` nor `TryComplete` work. It was mainly just a heads up, as naming is similar to TPL DataFlow for example, and in there, there's ways to initiate completion, and then a different Task you can call Wait on (with a timeout) to effectively initiate a non-graceful abort. But this implementation may be OK as is

ReubenBond:
No worries. This impl is intended to generally try to send all enqueued outgoing messages before closing the incoming channel and aborting the connection

PR #6560, opened 2020-05-28
Made ShouldPurge protected and virtual in eviction strategy.

adyada:
We are seeing some cases where some items from cache are purged before they get delivered to a grain, as they satisfy the time purge criteria. It could also be because of some customizations we have done to Orleans streaming. 

So, just making ShouldPurge as public method so that we can customize it, like only purge cache messages whose sequence token is less than min read sequence token in last 15 mins or something.

### DIFF
--- a/src/OrleansProviders/Streams/Common/PooledCache/ChronologicalEvictionStrategy.cs
+++ b/src/OrleansProviders/Streams/Common/PooledCache/ChronologicalEvictionStrategy.cs
@@ -84,8 +84,10 @@ namespace Orleans.Providers.Streams.Common
             this.periodicMonitoring?.TryAction(nowUtc);
         }
 
-        /// <inheritdoc />
-        public virtual bool ShouldPurge(ref CachedMessage cachedMessage, ref CachedMessage newestCachedMessage, DateTime nowUtc)
+        /// <summary>
+        /// Given a purge cached message, indicates whether it should be purged from the cache.
+        /// </summary>
+        protected virtual bool ShouldPurge(ref CachedMessage cachedMessage, ref CachedMessage newestCachedMessage, DateTime nowUtc)
         {
             TimeSpan timeInCache = nowUtc - cachedMessage.DequeueTimeUtc;
             // age of message relative to the most recent event in the cache.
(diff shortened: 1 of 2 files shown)


src/OrleansProviders/Streams/Common/PooledCache/IEvictionStrategy.cs
```diff
@@ -28,6 +28,15 @@ public interface IEvictionStrategy
...
         void OnBlockAllocated(FixedSizeBuffer newBlock);
+
+        /// <summary>
+        /// Given a purge cached message, indicates whether it should be purged from the cache.
+        /// </summary>
+        /// <param name="cachedMessage"></param>
+        /// <param name="newestCachedMessage"></param>
+        /// <param name="nowUtc"></param>
+        /// <returns></returns>
+        bool ShouldPurge(ref CachedMessage cachedMessage, ref CachedMessage newestCachedMessage, DateTime nowUtc);
```
jason-bragg, line 39: Why add this to the abstraction.  Is called by other?  Shouldn't currently be, as it was private.

src/OrleansProviders/Streams/Common/PooledCache/ChronologicalEvictionStrategy.cs
```diff
@@ -84,6 +84,16 @@ public void PerformPurge(DateTime nowUtc)
             this.periodicMonitoring?.TryAction(nowUtc);
         }
 
+        /// <inheritdoc />
+        public virtual bool ShouldPurge(ref CachedMessage cachedMessage, ref CachedMessage newestCachedMessage, DateTime nowUtc)
```
jason-bragg, line 88: Unless it needs to be public, suggest making it a protected virtual.  Specializations can override, but still an implementation detail of this strategy.

adyada: Okay, will make it protected virtual, somehow that skipped my mind that a "protected virtual" would also work.

jason-bragg:
Thank you @adyada !

adyada:
Thanks @jason-bragg for quick review and merging it.

PR #6563, opened 2020-05-29
Remove unused AppDomainTestHooks type

ReubenBond:


ReubenBond:
This is needed for some later PRs, since it removes some code that is no longer going to be supported.

PR #6566, opened 2020-05-29
Log the ManagedThreadId instead of the Type name in WorkItemGroup.

youfoundkris:
Our logs were showing the Threads property with the Type name.
This tiny PR removes the calls to ToString on the ManagedThreadId.
This should make sure the logs will contain the ManagedThreadId again.

Example:
`"Thread": "System.Threading.Thread"`


ReubenBond:
Thanks for the PR :) The change to log the actual id instead of the type name was included in #6408, so there's a solid chance you're running on a version without this change

sergeybykov:
I see that thread ID is logged properly. So indeed, this change is not needed anymore.

Thank you, @youfoundkris, for your effort!

Issue #5704, opened 2019-06-21
Add Analyzer/CodeGen error for invalid grain interface signatures


ReubenBond:
Grain interfaces do not support parameter modifiers such as `out`/`ref`/`in`.

We should add clear error messages via the Orleans.Analyzers package and codegen to notify users when they attempt to use an unsupported modifier.

See #5670 & #5530

An analyzer was added in https://github.com/dotnet/orleans/pull/5589 - it could be extended

(note: support for `in` could be added to the code generator, too)

miker1423:
If no one has take it, I'll like to take it :)

ReubenBond:
That would be fantastic, @miker1423 !

erictuvesson:
What is the status here @ReubenBond + @miker1423 ?

ReubenBond:
I haven't had cycles for it, and it would be a welcome contribution

miker1423:
Hello @erictuvesson, I had one half finished, it just needs testing, I pushed it to my fork, if you want to take over this issue, go ahead, but I could finish it by the weekend.

erictuvesson:
Hi @miker1423, if you feel like you have time that would be great; otherwise I could take a look.

sergeybykov:
Resolved via #6567. Thank you, @miker1423!

PR #6567, opened 2020-05-31
Adding reference parameter analyzer

miker1423:
This analyzer checks method declarations on interfaces or classes, where the IGrain class is implemented, to check for reference parameters and marking them as errors.

This closes issue #5704.


src/Orleans.Analyzers/NoRefParamsDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Orleans.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class NoRefParamsDiagnosticAnalyzer : DiagnosticAnalyzer
+    {
+        private const string BaseInterfaceName = "IGrain";
```
ReubenBond, line 13:
```suggestion
        private const string BaseInterfaceName = "IAddressable";
```

src/Orleans.Analyzers/NoRefParamsDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,55 @@
...
+            context.RegisterSyntaxNodeAction(AnalyzeSyntax, SyntaxKind.MethodDeclaration);
+        }
+
+        private static void AnalyzeSyntax(SyntaxNodeAnalysisContext context)
+        {
+            if (!(context.Node is MethodDeclarationSyntax syntax)) return;
+
+            var symbol = context.SemanticModel.GetDeclaredSymbol(syntax);
+            var isClassOrInterface = symbol.ContainingType.TypeKind == TypeKind.Interface ||
+                                     symbol.ContainingType.TypeKind == TypeKind.Class;
```
ReubenBond, line 36: We should only do this for interface, I think - any objections?

src/Orleans.Analyzers/NoRefParamsDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (!isClassOrInterface) return;
+
+            var implementedInterfaces = symbol.ContainingType
+                                              .AllInterfaces
+                                              .Select(interfaceDef => interfaceDef.Name);
+            if (!implementedInterfaces.Contains(BaseInterfaceName)) return;
+
+            foreach(var param in symbol.Parameters)
+            {
+                if (param.RefKind == RefKind.None) continue;
```
ReubenBond, line 47: Do we want to allow `in` params?

src/Orleans.Analyzers/NoRefParamsDiagnosticAnalyzer.cs
miker1423: Checking only for interfaces seems better, because to write a Grain you would always need the interface, and the only methods that should be checked would be the ones delcared in the interface, is that right?

src/Orleans.Analyzers/NoRefParamsDiagnosticAnalyzer.cs
miker1423: I don't know if ```in``` is now supported, if so, they should be allowed.

src/Orleans.Analyzers/NoRefParamsDiagnosticAnalyzer.cs
ReubenBond: That's right

src/Orleans.Analyzers/NoRefParamsDiagnosticAnalyzer.cs
ReubenBond: It wouldn't have the same semantics as what people would likely expect, since you can't reference memory in a remote machine, but there's no reason for us to block it.

Let's keep this code in your PR as-is for now and we can always broaden it up later if we decide we want to support it.

PR #6570, opened 2020-06-01
Catch Exception instead of RedisException

benjaminpetit:


src/Orleans.GrainDirectory.Redis/RedisGrainDirectory.cs
```diff
@@ -44,7 +44,7 @@ public async Task<GrainAddress> Lookup(string grainId)
 
                 return JsonConvert.DeserializeObject<GrainAddress>(result);
             }
-            catch (RedisException ex)
+            catch (Exception ex)
```
sergeybykov, line 47: This will catch even serializable exceptions, and lose their call stacks and other details. Is there no good way to only catch a narrower set of exceptions?

benjaminpetit: You are right, but we still want to log here too. I made the change to rethrow the original exception if it's not an exception from the Redis library

PR #6573, opened 2020-06-02
Use TaskCompletionSource.RunContinuationsAsynchronously

ReubenBond:
See https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

### DIFF
--- a/src/AdoNet/Shared/Storage/OrleansRelationalDownloadStream.cs
+++ b/src/AdoNet/Shared/Storage/OrleansRelationalDownloadStream.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Data.Common;
 using System.IO;
 using System.Threading;
@@ -186,7 +186,7 @@ namespace Orleans.Tests.SqlUtils
 
             if(cancellationToken.IsCancellationRequested)
             {
-                var tcs = new TaskCompletionSource<int>();
+                var tcs = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
                 tcs.SetCanceled();
                 return tcs.Task;
             }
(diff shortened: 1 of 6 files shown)


src/AdoNet/Shared/Storage/OrleansRelationalDownloadStream.cs
```diff
@@ -203,7 +203,7 @@ public override Task<int> ReadAsync(byte[] buffer, int offset, int count, Cancel
             catch(Exception e)
             {
                 //Due to call to Read, this is for sure a IOException and can be thrown out.
-                var tcs = new TaskCompletionSource<int>();
+                var tcs = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
                 tcs.SetException(e);
```
icanhasjonas, line 207: This actually makes no sense in this context. You're not in an async context, and the task you are returning is already completed. You should opt for using `Task.FromException()` and `Task. FromCanceled`

ReubenBond: Well spotted. The PR was mostly find and replace. When these methods were originally written, FromException and FromCanceled didn't exist (neither did Task.CompletedTask), hence the TCS.

Would you be willing to PR your suggestion?

icanhasjonas: Yeah, I figured as much :-)
I can do

ReubenBond: Thank you, @icanhasjonas, I appreciate the effort

PR #6577, opened 2020-06-04
Observe discarded ping task results

pentp:
Always observe ping task results, even if they are discarded (canceled).

### DIFF
--- a/src/Orleans.Core.Abstractions/Async/PublicOrleansTaskExtensions.cs
+++ b/src/Orleans.Core.Abstractions/Async/PublicOrleansTaskExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Orleans.Core.Abstractions/Async/PublicOrleansTaskExtensions.cs
```diff
@@ -31,7 +31,7 @@ public static void Ignore(this Task task)
                 task.ContinueWith(
                     IgnoreTaskContinuation,
                     CancellationToken.None,
-                    TaskContinuationOptions.NotOnRanToCompletion | TaskContinuationOptions.ExecuteSynchronously,
+                    TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.ExecuteSynchronously,
```
ReubenBond, line 34: Why this change?

pentp: To avoid running the continuation if the task is canceled, just more optimal.

ReubenBond: I ran some tests and the behavior seems fine

ReubenBond:
Thank you, Pent!

Issue #6517, opened 2020-04-30
Feature Request - A Way to Know if the State has Been Persisted Before


zeus82:
I would be helpful to know if an `IPersistentState<T>` has been written before.  The scenario I'm thinking of is when your creating a new entity:
1. I want to know if that entity has been created before
2. There are actions that need to be performed when its being created for the first time
We could do something like this:
```
public interface IPersistentState<TState> where TState : new()
{
  TState State { get; set; }
  string Etag { get; }
  bool WasLoadedFromPersistence { get; }
  Task ClearStateAsync();
  Task WriteStateAsync();
  Task ReadStateAsync();
}
```

`WasLoadedFromPersistence` would be `false` the first time the state is loaded, and it would be forever `true` after you call `WriteStateAsync`.  Ideally, it would return to `false` when you call `ClearStateAsync`.

I understand this would be a breaking change because all the persistence providers would need to set it appropriately.

ReubenBond:
Would checking that the ETag is not null/empty be sufficient, or perhaps adding a field (eg, CreationDate) to the TState?

zeus82:
CreateDate could would.  The problem with using the ETag is that according to the [documentation](https://dotnet.github.io/orleans/Documentation/grains/grain_persistence/index.html#storage-provider-semantics), storage providers don't need to set it.

zeus82:
I've implemented this feature locally, and I would like to put up a PR to have it reviewed.  What branch should the PR merge into?

This would be a breaking change for existing storage providers that aren't in this repo.

EdeMeijer:
Just my 2 cents, I have a lot of 'entity' grains and I follow a pattern where each one has an `Exists` state field. Each entity needs to be created explicitly once with a `Create()` function, which does the initial setup and sets the flag to true. In every other function I assert that the `Exists` flag is true, and in the create function I assert that it is false. I have some base class for entity grains that handles most of this logic in a generic way.

sergeybykov:
@zeus82 Does the approach with an explicit boolean flag would work for you? I'm personally hesitant about adding a flag like that as a standard feature because it would impose a tax of persisting that field on everyone.

zeus82:
I was thinking that field could be computed by the provided when it goes to load the data from the backing store - all the providers in this repo are explicit on when they load the data from the store vs creating a new instance.  That was the only tax would really be the bool property added to the `IPersistentState` interface.

sergeybykov:
[...]

PR #6580, opened 2020-06-05
Added 'RecordExists' flag to perisistent store so that grains can det…

zeus82:
…ect if a record exists already

### DIFF
--- a/src/AWS/Orleans.Persistence.DynamoDB/Provider/DynamoDBGrainStorage.cs
+++ b/src/AWS/Orleans.Persistence.DynamoDB/Provider/DynamoDBGrainStorage.cs
@@ -149,6 +149,7 @@ namespace Orleans.Storage
             if (record != null)
             {
                 var loadedState = ConvertFromStorageFormat(record, grainState.Type);
+                grainState.RecordExists = loadedState != null;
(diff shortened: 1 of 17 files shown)


zeus82:
This PR is for i[ssue 6517](https://github.com/dotnet/orleans/issues/6517).  I just want to make sure the approach I took is acceptable.  If it is I will add some unit tests.

EdeMeijer:
Good idea, but how will this affect my custom written storage providers? Just adding a method to an existing interface like that is a backwards incompatible change I'm afraid.

zeus82:
I didn't change `IGrainStorage`, so this will not break custom storage providers.  Until the custom provider is upgraded, that value will always be false.

benjaminpetit:
Thanks for your PR (and sorry for the delay)! I think your approach is totally fine, it would be great if you could add some unit tests. If you don't have time, I can add them myself to your PR.

Let me know what you prefer.

zeus82:
Thanks @benjaminpetit .  I added some asserts to some of the tests, I think that covers it, but if I missed anything, please let me know!

benjaminpetit:
/azp run Azure DevOps - Functional 

LGTM, I will merge it as soon as functional pass. Failure isn't related, but you never know...

Many thanks

PR #6585, opened 2020-06-06
GrainId: Activations & References

ReubenBond:
This is a rather large PR, which:

* Merges `IGrainContext` and `IGrainActivationContext` into one interface
* Adds `IGrainContextAccessor` for accessing the current context, since `IGrainActivationContext` will no longer be available as a scoped service (it involved some awkward hackery to make that work originally, and I wanted to remove it)
* Implements grain components, for extending the behavior of grains
* Implements grain bindings, for declaratively specifying which resources a grain interacts with (currently only used for implicit stream subs)
* Adds better support for generics via messaging/RPC changes
* Implements a more flexible system for creating grain instances as well as grain contexts
* Removes the now-obsoleted type metadata system

## Grain components

To make grains more openly extensible, this PR introduces the concept of grain components. Components are very closely analogous to ASP.NET's "Features" and we could call them features if we want, but this naming more closely aligns with the game development world and I feel it is more intuitive.

Components are exposed via two methods on `IGrainContext`: `void SetComponent<T>(T value)` and `T GetComponent<T>()`. They are backed by a `Dictionary<Type, object>`, though some components can be implemented directly by the `IGrainContext` implementation.

The idea is that components also form some hierarchy: when looking for an implementation of a component, the grain context will check if it has an implementation of it, then whether some shared, per-GrainType `GrainSharedComponents` object does, and (not implemented) whether some default components object can provide it.

Components are currently used for: `IGrainActivator` (to activate grain instances, see below), `IGrainExtensionBinder` and grain extensions themselves, `GrainCanInterleave` (`IsReentrant`, `MayInterleavePredicate`), and `ResourceFactoryRegistry` (used by transactions) and **components replace the `IDictionary<object, object> Items` property from `IGrainActivationContext` which this PR removes.**

In future, we can consider adding other pieces to components, such as `IGrainLocator`.

Components can be configured via `IConfigureGrain` implementations for each grain (see Grains below), and via `IConfigureGrainSharedComponents` instances for the `GrainSharedComponents` object.

## Grain bindings

Grain bindings are per-GrainType collections of string dictionaries that are intended to act as a way to share declarative information about resources which a grain interacts with. Azure Functions has a similar concept. These bindings are propagated throughout the cluster via the existing type metadata system (i.e, `ClusterManifest` / `GrainProperties`). In this PR, bindings are used for implicit stream subscription predicates, but they are intended to be used openly.

* Represented by `GrainBindings` (array of string dictionaries)
* Resolved by `GrainBindingsResolver`, from `GrainProperties` in `ClusterManifest`
* Added to `GrainProperties` by `IGrainBindingsProviderAttribute` impls (can be extended to other kinds of providers later, if we want).

## Messaging

Since its inception, Orleans has used integer-based hash codes to identify grain types, interface types, and interface methods. This has worked very well, but complicates debugging (which grain/interface/method is being called?), portability across runtimes, and support for generics (introducing some cases and assumptions which cannot be solved by the existing system).

This PR brings a few changes for the core message format:

* All messages have an `InterfaceId` header of type `GrainInterfaceId`, which is a stringly identifier for a grain interface.
* All messages have an `InterfaceVersion` header, lifting this field from `InvokeMethodRequest` (in the body of the message), since it is needed in identifying compatibility (along with interface id and grain type).
* `NewGrainType` and `GenericGrainType` headers are removed, since a grain's implementation type can be fully resolved from its `GrainType`

## Grains

Currently, the responsibility for creating an instance of a grain largely resides in the `Catalog` and grain instances are very tightly coupled to `ActivationData` instances. This PR adds a new class, `GrainContextActivator`, with one method `IGrainContext CreateInstance(ActivationAddress address)`. This class is responsible for creating an `IGrainContext` (today, `ActivationData` is the only implementation), including the associated Grain instance itself. To do this, it relies on a series of extensibility points:

* `IGrainContextActivatorProvider` implementations from DI are called in registration order to return an `IGrainContextActivator` which is then cached per `GrainType`.
* `IGrainContextActivator.CreateContext(ActivationAddress)` is called to construct the grain context, including the grain instance
* `IConfigureGrainContext` instances which apply to a given `GrainType` are called on the constructed `IGrainContext`. These instances are also cached, similarly to above, and provided by `IConfigureGrainContextProvider` implementations from DI

The only implementation of `IGrainContextActivatorProvider` included with this PR is `ActivationDataActivatorProvider`. This provider uses `GrainClassMap` (driven by `GrainClassFeature` and hence application parts / codegen) to find the implementation class for a `GrainType` and will construct any class which derives from `Grain`, returning an instance of `ActivationDataActivator` to do the actual construction.

`ActivationDataActivator` uses the `GrainSharedComponents` to get an `IGrainActivator` to create the actual grain object. The default activator is configured in `ConfigureDefaultGrainActivator : IConfigureGrainSharedComponents`.

## Grain References

This PR reduces the `GrainReference` class down to 2 fields: an `IdSpan` that is the `Key` part of `GrainId`, and `GrainReferenceShared` which is a shared object for all references of a given `GrainInterfaceId` + `GrainType` pair.

Grain references need to be instantiated so that they can be used. The existing system for doing this is centered around `GrainInterfaceData`, which is shared around the cluster and accessed from `GrainTypeResolver`. Once a grain reference type is resolved from there, `GrainReferenceFactory` and `GrainCasterFactory` perform the actual construction into a typed (generated) `GrainReference` implementation using a little bit of IL.

Starting with this PR, it is the responsibility of `GrainRefrenceActivator` to create grain references. This uses `IGrainreferenceActivatorProvider` instances from DI to construct an `IGrainReferenceActivator` capable of constructing references for a given `GrainType` and `GrainInterfaceId`. The resulting `IGrainReferenceActivator` has a single method: `GrainReference CreateReference(GrainId grainId)`, which is used to finally construct the `GrainReference`.

The advantage of this approach is that the resulting references can easily share common properties (any information which is common to the grain type and/or the interface id), and only a single lookup is required (after initialization) in order to find a suitable activator to use.


src/Orleans.Runtime/Facet/Persistent/PersistentStateStorageFactory.cs
```diff
@@ -25,22 +25,21 @@ public IPersistentState<TState> Create<TState>(IGrainActivationContext context,
             return bridge;
         }
 
-        protected virtual string GetFullStateName(IGrainActivationContext context, IPersistentStateConfiguration cfg)
+        protected virtual string GetFullStateName(IGrainContext context, IPersistentStateConfiguration cfg)
         {
-            return $"{RuntimeTypeNameFormatter.Format(context.GrainType)}.{cfg.StateName}";
+            return $"{context.GrainId}.{cfg.StateName}";
```
EdeMeijer, line 30: @ReubenBond are you sure this is correct? I'm working on a custom storage provider and was trying to figure out how the 'grain type' value was being constructed and noticed this changed since the latest release. Currently, grain types are fully formatted types that can be parsed by `Type.GetType()` + the state name, but here it seems you're changing the type to a grain ID, which would at the least be backwards incompatible. Maybe I'm missing something.

EdeMeijer (commented):
Could you check this one inline comment?

src/Orleans.Runtime/Facet/Persistent/PersistentStateStorageFactory.cs
ReubenBond: You're right. I don't have a final solution for that yet, but I was considering we would want to either fork providers or provide an adapter in order to support migration from grains using legacy ids to the simplified id system.

I would also be ok with a PR to revert this particular change, if that helps your case.

We will need to audit the changes before final release to make sure that a migration path is viable.

Note that this code has not been included in any releases - all releases are made from 3.x.x tags and do not include these commits.

PR #6586, opened 2020-06-07
Constrain work done under a lock in BatchWorker

pentp:
Currently `BatchWorker.Start` runs arbitrary code under the lock by invoking `Work()`.
Also, if `Work()` throws an exception, then  `currentWorkCycle` can be null and Start can fail with a NRE.
And because all of this is done under a lock, `startingCurrentWorkCycle` can never be observed to be true and there is no need for volatile as a full memory barrier is implied by lock anyway.

### DIFF
--- a/src/Orleans.Core/Async/BatchWorker.cs
+++ b/src/Orleans.Core/Async/BatchWorker.cs
@@ -3,6 +3,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Orleans.Timers.Internal;
 
+#nullable enable
 namespace Orleans
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


src/Orleans.Core/Async/BatchWorker.cs
```diff
@@ -157,88 +147,54 @@ private void CheckForMoreWork()
...
-                await await waitfortasktask;
-            }
-            else if (waitfortask != null)
-            {
-                await waitfortask;
-            }
+        private Task CreateNextWorkCyclePromise()
+        {
+            // it's OK to run any continuations synchrnously because this promise only gets signaled at the very end of CheckForMoreWork
+            nextWorkCyclePromise = new TaskCompletionSource<Task>();
```
ReubenBond, line 177: Prefer if we pass `TaskCreationOptions.RunContinuationsAsynchronously` for the reasons mentioned in the guidance here: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

ReubenBond: Looks good to me other than this, and possibly the same for the Task.Factory call (though with `TaskCreationOptions` instead). Hard to be sure that user code attaching continuations to these tasks will not end up causing a deadlock.

This structure of this class is not ideal in my eyes. I feel that a processor loop with a semaphore could be a cleaner approach. Still, this does look cleaner to me.

pentp: I understand the guidance, but in this specific case it's safe and more optimal to potentially run continuations inline because it's signaled when CheckForMoreWork has finished and CheckForMoreWork is always scheduled asynchronously.

pentp: `Task.Factory` does not have the same issues with continuations that TaskCompletionSource has because there is no explicit signaling of work completion.

ReubenBond: Sounds fine to me

ReubenBond: One last thing, though: was this PR in response to an issue you hit, or is it more speculative?

ReubenBond:
Thanks again, Pent

src/Orleans.Core/Async/BatchWorker.cs
pentp: No specific issue, I was just looking for `Task<Task>` uses and then noticed the problems in BatchWorker.Start method.

PR #6589, opened 2020-06-09
Stop sending messages to a gateway that is shutting down

benjaminpetit:
When a silo shutdown, notify clients connected to the local gateway to stop sending messages.

While the current implementation is a bit hacky, we got very good results, and it shouldn't break backward compatibility, so we should be able to backport this fix to 3.2.x and potentially 3.1.x branches.


benjaminpetit:
Replaced by #6643

PR #6592, opened 2020-06-12
Support deterministic builds with CodeGenerator

### LINKED ISSUE #6495: Orleans generating non-deterministic class names
When building code with /p:[Deterministic](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/deterministic-compiler-option)=true, we expect that code generated by Orleans would also produce the same outputs for the same inputs. While Orleans does support incremental builds (avoid running code gen when inputs have not changed), still invoking generator for the same inputs (e.g. in two clean builds) will produce slightly different assemblies. This is mostly due to the use of `Guid.NewGuid()` in:

* https://github.com/dotnet/orleans/blob/ffabe550a7891e7378f9f076e88a8e3c184d967f/src/Orleans.CodeGeneration/FeaturePopulatorGenerator.cs#L23
* https://github.com/dotnet/orleans/blob/ffabe550a7891e7378f9f076e88a8e3c184d967f/src/Orleans.CodeGenerator/Generators/FeaturePopulatorGenerator.cs#L23

Can the generated code adhere to the [Deterministic](https://blog.paranoidcoding.com/2016/04/05/deterministic-builds-in-roslyn.html) mode?

rikbosch:
This PR contains the changes required to fix #6495 

The `FeaturePopulator` classname will be generated by the sanitized assembly-name (if known).

This will enable the support of deterministic builds of projects with Orleans codegeneration enabled.


ReubenBond:
Thank you, @rikbosch - much appreciated!

PR #6598, opened 2020-06-18
Docs

Abramalin:
Updating DoClientWork method in Client program using interpolation, optional, for readability. Interpolation is used in the HelloGrain.cs class.


sergeybykov:
Thank you, @Abramalin!

PR #6621, opened 2020-06-29
Initial message scheduling reorganization

ReubenBond:


src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -215,238 +68,77 @@ public void ReceiveMessage(Message message)
...
+            ActivationAddress forwardingAddress,
             string failedOperation,
             Exception exc = null,
             bool rejectMessages = false)
         {
             // Just use this opportunity to invalidate local Cache Entry as well. 
             if (oldAddress != null)
             {
-                this.localGrainDirectory.InvalidateCacheEntry(oldAddress);
+                var localGrainDirectory = this.RuntimeClient.ServiceProvider.GetRequiredService<ILocalGrainDirectory>();
```
benjaminpetit, line 120: Any reason to not have it in the constructor anymore? To refactor in the future more easily?

src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -652,21 +294,9 @@ void OnAddressingFailure(Message m, IGrainContext activation, Exception ex)
...
-
         /// <summary>
         /// Resolve target address for a message
-        /// - use transaction info
-        /// - check ordering info in message and sending activation
-        /// - use sender's placement strategy
         /// </summary>
-        /// <param name="message"></param>
         /// <returns>Resolve when message is addressed (modifies message fields)</returns>
         private Task AddressMessage(Message message)
```
benjaminpetit, line 301: Shoould we change this method to return a `ValueTask` instead?

src/Orleans.Runtime/Core/Dispatcher.cs
```diff
@@ -719,99 +349,107 @@ internal void SendResponse(Message request, Response response)
...
+                try
+                {
+                    var targetActivation = catalog.GetOrCreateActivation(
+                        msg.TargetAddress,
+                        msg.IsNewPlacement,
+                        msg.RequestContextData);
+
+                    if (targetActivation is null)
+                    {
+                        logger.LogInformation(
```
benjaminpetit, line 413: Should we move this log line inside the `else` to avoid logging two times for the response case?

src/Orleans.Runtime/Core/Dispatcher.cs
ReubenBond: No good reason, at least not anymore. It probably made sense to me at the time (maybe to break a cyclic loop which no longer exists).

Thanks

src/Orleans.Runtime/Core/Dispatcher.cs
ReubenBond: I think we can leave it for now because it's an untyped task (no return value) so the synchronous path can return `Task.CompletedTask`

ReubenBond: We *should* convert `IActivationSelector`. I can do that in this PR if you're ok with it

src/Orleans.Runtime/Core/Dispatcher.cs
ReubenBond: Yes, well spotted, thanks

src/Orleans.Runtime/Core/Dispatcher.cs
ReubenBond: Added, it's not too much change

ReubenBond:
Addressed feedback and fixed conflicts

ReubenBond:
Distributed tests pass - please merge when ready

Issue #6028, opened 2019-10-09
Could not load file or assembly 'Microsoft.Bcl.AsyncInterfaces, Dependency missing in Orleans 3.0 RC1?


emilekberg:
I might be a bit quick to jump the gun, but i wanted to try out orelans-3.0.0-RC1 after it got released. going from 3.0.0-Beta1 broke to me, seems like Microsoft.Bcl.AsyncInterfaces is missing. 
got this error: 
```
Exception has occurred: CLR/System.IO.FileNotFoundException
An exception of type 'System.IO.FileNotFoundException' occurred in Orleans.Runtime.dll but was not handled in user code: 'Could not load file or assembly 'Microsoft.Bcl.AsyncInterfaces, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. Det går inte att hitta filen.'
   at Orleans.Hosting.DefaultSiloServices.AddDefaultServices(IApplicationPartManager applicationPartManager, IServiceCollection services)
   at Orleans.Hosting.CoreHostingExtensions.<>c.<ConfigureDefaults>b__6_0(HostBuilderContext context, IServiceCollection services)
   at Orleans.Hosting.SiloBuilder.Build(HostBuilderContext context, IServiceCollection serviceCollection)
   at Microsoft.Extensions.Hosting.GenericHostExtensions.<>c__DisplayClass0_0.<UseOrleans>b__0(HostBuilderContext context, IServiceCollection services)
   at Microsoft.Extensions.Hosting.HostBuilder.CreateServiceProvider()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
   at TestOrleans.ServiceHost.Program.<Main>d__0.MoveNext() in E:\dev\orleans\TestOrleans.ServiceHost\Program.cs:line 18
```
installing it made it work however. Wanted to drop the info somewhere in case it got missing somewhere. Saw no info in the release notes about this

ReubenBond:
@terrajobst did we make a silly packaging mistake here concerning IAsyncEnumerable? I don't believe we're referencing `Microsoft.Bcl.AsyncInterfaces` explicitly anywhere.

terrajobst:
What platform are you on (.NET Core 2.x or .NET Core 3.0)?

Someone in your transitive closure probably does. Looks like `Microsoft.AspNetCore.Connections.Abstractions` for example does, but only for .NET Standard 2.0.

ReubenBond:
[...]

PR #6638, opened 2020-07-07
Remove redundant NuGet package references, particularly for .NET Core.

teo-tsirpanis:


sergeybykov:
@teo-tsirpanis Thank you for the PR!

TBH I'm a bit nervous about removing those seemingly redundant dependencies. Some of them were added as fixes and workarounds for external bugs and dependency version ambiguities. For example, `Microsoft.Bcl.AsyncInterfaces` was added in #6031 as a workaround for #6028 

Some of those issues/bugs have probably been fixed by now. However, it's not clear to me how to validate that.

We also tend to avoid conditional references, such as
>Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">

to keep things simpler.

teo-tsirpanis:
@sergeybykov #6028 was about Orleans 3.0.0-RC1 when it was targeting only .NET Standard 2.0. `Microsoft.Bcl.AsyncInterfaces` is still needed on .NET Standard but not on .NET Core 3.0+ (notice that it does not target directly `netcoreapp3.0` and CI still succeeds after the explicit references are removed.

Furthermore, packages like [`System.Net.Http`](https://www.nuget.org/packages/System.Net.Http/) which stuck in version 4.3.* have not been updated since a couple of years and were only needed by libraries targeting .NET Standard 1.*. An extreme example is [`System.Data.Common`](https://www.nuget.org/packages/System.Data.Common/) whose latest version was released in 2016.

> We also tend to avoid conditional references [...] to keep things simpler.

I admit the complexity of the project files slightly raises but not much in my opinion; at least grouping the dependencies in an `ItemGroup` is better than a `<PackageReference Condition="..." Include="..."/>`. The benefits of this miniscule complexity however are tangible: the dependency tree becomes smaller, unmaintained packages are stopped being used, and the good practice of keeping dependencies at the minimum necessary is honored.

> [I]t's not clear to me how to validate that

First of all the CI succeeded. Second, we can take a look at the issues that led to the inclusion of these packages and try to reproduce them. I will further investigate.

teo-tsirpanis:
Orleans targets .NET Standard 2.0 since 87690de which occured in September 2017. The `System.*` NuGet package references already existed for the times where Orleans was targeting .NET Standard 1.5, but were never removed until now.

---

As for the conditional references, I just thought of `'$(TargetFramework)' == 'netstandard2.0'` but the approach in the PR is more robust and will automatically work if Orleans hypothetically targets .NET Standard _2.1_ in the future.

sergeybykov:
Since `master` branch is effectively 4.0, we don't actually need to be too cautious about changes like this. Worst case, we'll sort any issues once we have a preview or beta build out.

Thank you, @teo-tsirpanis!

PR #6640, opened 2020-07-08
Fix unobserved ArgumentOutOfRangeException from Task.Delay

pentp:
Reminders are allowed to have dueTime/period > int.MaxValue milliseconds (about 24 days) which causes `AsyncTimer.NextTick` to call `Task.Delay` with invalid arguments, the result isn't observed and the next reminder tick is immediately run...

### DIFF
--- a/src/Orleans.Runtime/Timers/AsyncTimer.cs
+++ b/src/Orleans.Runtime/Timers/AsyncTimer.cs
@@ -64,13 +64,14 @@ namespace Orleans.Runtime
                     if (delay > maxDelay)
                     {
                         delay -= maxDelay;
-                        await Task.Delay(maxDelay, cancellation.Token);
+                        await Task.Delay(maxDelay, cancellation.Token).ConfigureAwait(false);
                     }
 
-                    await Task.Delay(delay, cancellation.Token);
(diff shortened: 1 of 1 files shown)


ReubenBond:
Thanks, looks good to me. In the future, perhaps we should adopt something more similar to SignalR's TimerAwaitable instead: https://github.com/dotnet/aspnetcore/blob/master/src/SignalR/common/Shared/TimerAwaitable.cs

ReubenBond:
By the way, @pentp, I am considering partially reverting this to use the cancellationTask again, instead of passing the CancellationToken to Task.Delay, since the CT version throws first-chance exceptions during shutdown and I prefer to avoid them wherever possible

pentp:
Sounds reasonable - it felt like an improvement for the common case, but I wasn't really sure about the trade-offs also.

ReubenBond:
I suppose if we were to change it again, it would be better to use TimerAwaitable from SignalR source, or a variant of it, modified to fit our needs

pentp:
Or you could wrap the Task.Delay calls with Task.WhenAny (with a single task) - this would also avoid cancellation exceptions.

PR #6672, opened 2020-08-03
Analyze running activations and report back status to callers for long-running and blocked requests

ReubenBond:
Adds a new `ActivationWorkloadMonitor` type that periodically scans active activations to analyze running and blocked requests. Requests which have been running or blocked for a configurable period of time are responded to with a status update message that is printed as a warning. Eventual timeout exceptions also include the latest status messages.

The hope is that this will improve debuggability for some users and make persistent issues more apparent.

Also adds the ability to tier timeouts based upon the most recent status update. Eg, for messages with no response, we can timeout early. For messages which are running, we can allow them to keep running past the usual 30s timeout (although we would need to also consider this when performing message expiry).


ReubenBond:
In future, we can use this mechanism at the center of deadlock detection

PR #6673, opened 2020-08-03
Reduce package dependencies

pentp:
#6638 removed many redundant package references, but there are some entirely unused references also:
* `System.Management` is only used to query total system memory from WMI - this can be replaced with a single API call.
* `Orleans.Hosting.AzureCloudServices` and `Orleans.Hosting.ServiceFabric` referenced unrelated `AzureStorage`/`OrleansProviders` projects.
* Many projects referenced `Microsoft.Extensions.Logging.Console`, but it's actually only used in `CodeGenerator.MSBuild`.
* `CodeGenerator.MSBuild` referenced `Newtonsoft.Json` (unused).
* `Orleans.Connections.Security` had redundant references (duplicated from `Orleans.Runtime.Abstractions`).


ReubenBond:
Thanks, @pentp! I appreciate it

PR #6681, opened 2020-08-11
Implement == and != operators in StreamId and InternalStreamId

benjaminpetit:


src/Orleans.Core.Abstractions/IDs/StreamId.cs
```diff
@@ -74,6 +74,12 @@ public static StreamId Create(byte[] ns, byte[] key)
 
         public bool Equals(StreamId other) => FullKey.Span.SequenceEqual(other.FullKey.Span);
 
+        public override bool Equals(object obj) => obj is StreamId ? Equals((StreamId)obj) : false;
```
ReubenBond, line 77:
```suggestion
        public override bool Equals(object obj) => obj is StreamId other && this.Equals(other);
```

src/Orleans.Core.Abstractions/IDs/StreamId.cs
```diff
@@ -123,7 +129,13 @@ private InternalStreamId(SerializationInfo info, StreamingContext context)
 
         public bool Equals(InternalStreamId other) => StreamId.Equals(other) && ProviderName.Equals(other.ProviderName);
 
-        public int CompareTo(InternalStreamId other) => throw new NotImplementedException();
+        public override bool Equals(object obj) => obj is InternalStreamId ? Equals((InternalStreamId)obj) : false;
```
ReubenBond, line 132:
```suggestion
        public override bool Equals(object obj) => obj is InternalStreamId other && this.Equals(other);
```

PR #6682, opened 2020-08-11
Pluggable StreamId to GrainId mapper

benjaminpetit:


src/Orleans.Core.Abstractions/Manifest/GrainProperties.cs
```diff
@@ -121,6 +121,11 @@ public static class WellKnownGrainTypeProperties
         /// </summary>
         public const string StreamBindingIncludeNamespaceKey = "include-namespace";
 
+        /// <summary>
+        /// Key type of the grain, if it implement a legacy interface. Valid values are nameof(String), nameof(Int32) and nameof(Guid)
```
ReubenBond, line 125: Should this be Int64?

src/Orleans.Core/Streams/Core/IStreamIdMapper.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Common interface for component that map a StreamId to a GrainId
+    /// </summary>
+    public interface IStreamIdMapper
+    {
+        /// <summary>
+        /// Get the corresponding GrainId from the StreamId
+        /// </summary>
+        GrainId GetGrainId(GrainBindings grainBindings, StreamId streamId);
```
ReubenBond, line 14: I think ultimately a model which is compositional would be better. Eg, TryGetGrainId, so you can run them in sequence before falling back to a default

src/Orleans.Core/Streams/PubSub/ImplicitStreamSubscriberTable.cs
```diff
@@ -79,14 +82,15 @@ private Cache BuildCache(MajorMinorVersion version, ImmutableDictionary<GrainTyp
...
-                    bool includeNamespaceInGrainId = false;
-                    if (grainBinding.TryGetValue(WellKnownGrainTypeProperties.StreamBindingIncludeNamespaceKey, out var value)
-                        && string.Equals(value, "true", StringComparison.OrdinalIgnoreCase))
+                    if (!grainBinding.TryGetValue(WellKnownGrainTypeProperties.StreamIdMapperKey, out var mapperName))
                     {
-                        includeNamespaceInGrainId = true;
+                        throw new KeyNotFoundException(
+                           $"Stream binding for grain type {binding.GrainType} is missing a \"{WellKnownGrainTypeProperties.StreamIdMapperKey}\" value");
                     }
+                    var streamIdMapper = _serviceProvider.GetServiceByName<IStreamIdMapper>(string.IsNullOrWhiteSpace(mapperName) ? DefaultStreamIdMapper.Name : mapperName);
```
ReubenBond, line 90: I think we should pass the default name through the grain bindings, rather than pass null/empty and map it here

PR #6648, opened 2020-07-13
AAD authentication for Azure providers (blob, queue & table)

aelij:
Note this PR changes a few public APIs (`AzureQueueDataManager` and `AzureQueueAdapterFactory`) due to the new Azure Queue SDK, which uses `string` instead of the `CloudQueueMessage` class for sending messages.

PR #6686, opened 2020-08-14
Cherry-pick commits for 3.3.0

sergeybykov:
Improve ClusterClient disposal (#6583)
Added 'RecordExists' flag to perisistent store so that grains can detect if a record exists already (#6580)
AAD authentication for Azure providers (blob, queue & table) (#6648)


pentp:
The `RecordExists` flag should be mentioned as a potentially breaking change - for example OrleansTestKit doesn't implement these new properties yet.

sergeybykov:
Good point, @pentp. Will update the release notes.

PR #6693, opened 2020-08-17
Fix ordering in ASP.NET Co-hosting sample

ReubenBond:


clement128:
hi @ReubenBond may I ask why change the ordering? thanks

ReubenBond:
The silo needs to be started before accepting requests

clement128:
@ReubenBond hmm that's interested, I used same ordering on production since few months ago, never encountered any error. it happen randomly?

ReubenBond:
It can happen only during startup, for a brief period. If you don't receive traffic immediately, then it ought to be fine

clement128:
thanks

PR #6710, opened 2020-08-24
Update .NET SDK version

ReubenBond:


ReubenBond:
This is failing due to an issue with the Use .NET SDK Azure Pipelines task

PR #5408, opened 2019-02-28
Removed filtering from streaming infrastructure

jason-bragg:
Breaking changes to core.abstractions, so this is, at earliest, a 3.0 change

Issue #5939, opened 2019-09-07
Retain producer-side stream filtering


cata:
**Background**

Producer-side stream filtering is to be removed in an upcoming revision of Orleans (per ~#5408~ #6711).
See relevant Gitter excerpts below:
* [August 13, 2019 3:04 PM](https://gitter.im/dotnet/orleans?at=5d535e35d03a7c63e62256d5)
* [August 14, 2019 7:23 AM](https://gitter.im/dotnet/orleans?at=5d5443ad9507182477d9d6be)
* [August 14, 2019 8:04 AM](https://gitter.im/dotnet/orleans?at=5d544d494e17537f522fbd4e)

In short, the intent is to remove the feature due to maintainability and design-related concerns in the face of the pub-sub system becoming a public surface.
My understanding is that the primary design concern stems from the way the filter predicate is specified.  In the current implementation, the identity of the filter predicate (represented by a class name and method name tuple) is supplied by the subscribing party. In addition to constraining  eligible predicates to  static methods, the mechanism used to extract this identity employs delegates and leads to an undesirable dependency whereby the predicate function must be available to both producer and consumer. While this may appear fairly benign in the case in which both the producer and the consumer are grains, this requirement becomes harder to justify when consuming streams from an Orleans client. 

**Advocating for retaining producer-side filtering** 

> Please note that the scenarios described below have been significantly simplified for the sake of brevity

We use per-grain streams to publish incremental grain state changes, in a variation on event sourcing. Clients subscribe to these streams to render changes to grain state in real time. Upon subscribing, a new client receives an optimized version of the change log, allowing it to "catch-up" with the current grain state - we are currently using the feature in production to selectively deliver this class of messages to only the newly subscribed client. 

In addition to simple state(s), grains can hold collections - changes to these collections are similarly propagated to clients in real time. Clients can construct queries over these collections, the results of which are also updated in real time - we are currently using the feature in production to selectively deliver initial query results to only the client that has made the query. 

Continuing to be able to filter such messages on the producer side is very important to us.

**Possible approaches to retaining producer-side filtering**
 
It appears that allowing arbitrary client-supplied predicates while addressing the coupling concern would require serializing the predicate closure provided by the consumer at subscription time and then deserializing and executing it on the producer side.
[...]

PR #6711, opened 2020-08-24
Removed filtering from streaming infrastructure

benjaminpetit:
Replace #5408


cata:
Please see [this comment](https://github.com/dotnet/orleans/pull/5408#issuecomment-528568705) and #5939, which is still open.
Can you please share more information as to what the plan is regarding the above?

benjaminpetit:
For now: no defined plan, except cleaning.

We want first to have clean, reliable, and consistent API across all providers. This current implementation of stream filtering does not satisfy these requirements.

But we are open to suggestion and PR! I've read your issue; we can chat a bit and see what we can do.

ReubenBond:
One conflict

benjaminpetit:
Rebased

PR #6716, opened 2020-08-26
Improve error message for OpenConnectionTimeout

pentp:
This improves the error message for connection opening timeouts.
Also updated `SocketConnectionFactory.ConnectAsync` to properly support cancellation (mostly copied from dotnet/runtime repo).

Message diff for OrleansMessageRejectionException looks similar to this:
```diff
+Exception while sending message: Orleans.Runtime.Messaging.ConnectionFailedException: Connection attempt to endpoint S10.0.0.4:11111:332519906 timed out after 00:00:03
-Exception while sending message: Orleans.Runtime.Messaging.ConnectionFailedException: Unable to connect to endpoint S10.0.0.4:11111:332519906. See InnerException ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
-   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
-   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
-   at Orleans.Internal.OrleansTaskExtentions.<MakeCancellable>d__25`1.MoveNext() in D:\build\agent\_work\25\s\src\Orleans.Core\Async\TaskExtensions.cs:line 409
---- End of stack trace from previous location where exception was thrown ---
-   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
-   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
-   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
-   at Orleans.Runtime.Messaging.ConnectionManager.<ConnectAsync>d__20.MoveNext() in D:\build\agent\_work\25\s\src\Orleans.Core\Networking\ConnectionManager.cs:line 239
---- End of inner exception stack trace ---
   at Orleans.Runtime.Messaging.ConnectionManager.<ConnectAsync>d__20.MoveNext() in D:\build\agent\_work\25\s\src\Orleans.Core\Networking\ConnectionManager.cs:line 263
 --- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
 --- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Orleans.Runtime.Messaging.ConnectionManager.<GetConnectionAsync>d__15.MoveNext() in D:\build\agent\_work\25\s\src\Orleans.Core\Networking\ConnectionManager.cs:line 114
   at Orleans.Runtime.Messaging.ConnectionManager.<GetConnectionAsync>d__15.MoveNext() in D:\build\agent\_work\25\s\src\Orleans.Core\Networking\ConnectionManager.cs:line 114
 --- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Orleans.Runtime.Messaging.OutboundMessageQueue.<<SendMessage>g__SendAsync|10_0>d.MoveNext() in D:\build\agent\_work\25\s\src\Orleans.Runtime\Messaging\OutboundMessageQueue.cs:line 121
 --- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Orleans.Runtime.Messaging.OutboundMessageQueue.<<SendMessage>g__SendAsync|10_0>d.MoveNext() in D:\build\agent\_work\25\s\src\Orleans.Runtime\Messaging\OutboundMessageQueue.cs:line 121

### DIFF
--- a/src/Orleans.Core/Networking/Connection.cs
+++ b/src/Orleans.Core/Networking/Connection.cs
@@ -37,7 +37,6 @@ namespace Orleans.Runtime.Messaging
         private readonly ConnectionDelegate middleware;
         private readonly Channel<Message> outgoingMessages;
         private readonly ChannelWriter<Message> outgoingMessageWriter;
-        private readonly object _closeLock = new object();
         private readonly List<Message> inflight = new List<Message>(4);
         private readonly TaskCompletionSource<int> _transportConnectionClosed = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
         private IDuplexPipe _transport;
@@ -113,12 +112,12 @@ namespace Orleans.Runtime.Messaging
             return connection.RunInternal();
         }
 
-        protected virtual async Task RunInternal()
+        protected virtual Task RunInternal()
         {
             _transport = this.Context.Transport;
             _processIncomingTask = this.ProcessIncoming();
             _processOutgoingTask = this.ProcessOutgoing();
-            await Task.WhenAll(_processIncomingTask, _processOutgoingTask);
+            return Task.WhenAll(_processIncomingTask, _processOutgoingTask);
         }
 
         /// <summary>
@@ -130,13 +129,10 @@ namespace Orleans.Runtime.Messaging
 
         protected abstract void RetryMessage(Message msg, Exception ex = null);
 
-        public async Task CloseAsync(Exception exception)
+        public Task CloseAsync(Exception exception)
         {
             StartClosing(exception);
-            if (_closeTask is Task task && !task.IsCompleted)
-            {
-                await _closeTask;
-            }
+            return _closeTask;
         }
 
         private void OnTransportConnectionClosed()
@@ -152,18 +148,11 @@ namespace Orleans.Runtime.Messaging
                 return;
             }
 
-            TaskCompletionSource<int> completion;
-            lock (_closeLock)
+            var task = new Task<Task>(FinishClosing);
+            if (Interlocked.CompareExchange(ref _closeTask, task.Unwrap(), null) is object)
             {
-                if (_closeTask is object)
-                {
-                    return;
-                }
-
-                completion = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
-                _closeTask = completion.Task;
+                return;
             }
-
 
             if (this.Log.IsEnabled(LogLevel.Information))
             {
@@ -173,22 +162,7 @@ namespace Orleans.Runtime.Messaging
                     this);
             }
 
-            _ = WrapCloseAsync(this, completion);
-
-            // Propagate the result of the close method to the task completion source.
-            static async Task WrapCloseAsync(Connection self, TaskCompletionSource<int> completion)
-            {
-                try
-                {
-                    await Task.Yield();
-                    await self.FinishClosing().ConfigureAwait(false);
-                    completion.SetResult(0);
-                }
-                catch (Exception closeException)
-                {
-                    completion.SetException(closeException);
-                }
-            }
+            task.Start();
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/Orleans.Core/Networking/ConnectionManager.cs [hunk 1]
```diff
@@ -325,8 +323,9 @@ public async Task CloseAsync(SiloAddress endpoint)
                     }
                 }
 
-                await Task.WhenAll(closeTasks);
+                return Task.WhenAll(closeTasks);
```
ReubenBond, line 326: await is used here for debuggability - to retain stacks. I prefer that over performance for cold paths like this.
Was there a reason other than performance for this change?

ReubenBond (commented):
Looks good to me except the comments about intentional use of async/await

src/Orleans.Core/Networking/Connection.cs
```diff
@@ -113,12 +112,12 @@ private static Task OnConnectedAsync(ConnectionContext context)
...
         }
 
-        protected virtual async Task RunInternal()
+        protected virtual Task RunInternal()
         {
             _transport = this.Context.Transport;
             _processIncomingTask = this.ProcessIncoming();
             _processOutgoingTask = this.ProcessOutgoing();
-            await Task.WhenAll(_processIncomingTask, _processOutgoingTask);
+            return Task.WhenAll(_processIncomingTask, _processOutgoingTask);
```
ReubenBond, line 120: Await here is intentional - see the other comments

src/Orleans.Core/Networking/Connection.cs
```diff
@@ -130,13 +129,10 @@ protected virtual async Task RunInternal()
...
 
-        public async Task CloseAsync(Exception exception)
+        public Task CloseAsync(Exception exception)
         {
             StartClosing(exception);
-            if (_closeTask is Task task && !task.IsCompleted)
-            {
-                await _closeTask;
-            }
+            return _closeTask;
```
ReubenBond, line 135: Await here is intentional - see the other comments

ReubenBond (approved):

veikkoeeva:
I drop
https://twitter.com/Scooletz/status/1362328009160679424
https://github.com/dotnet/orleans/search?q=Interlocked.Exchange
regarding `Interlocked.Exchange`.

Just in case here also as this showed up. Unfortunately I do not have the bandwith to right now to see further (maybe over the weekend if needed).

pentp:
I'm using `Interlocked.CompareExchange` here to conditionally set/start a task, not sure what I should do with that tweet or search?

veikkoeeva:
@pentp Likely nothing. I had once bug with `Interlocked.CompareExchange` and volatility in tests that prompoted to check if there is such a situation elsewhere. Maybe not, that tweet reminded me about it, so I ran a search over it.

ReubenBond:
One of the most nefarious things to look out for with cmpxchg is the ABA problem, but our uses in this codebase are hopefully all straightforward enough that we do not encounter it https://en.wikipedia.org/wiki/ABA_problem

pentp:
The new logic turned out to be so efficient that it exposed a race condition that was previously masked by scheduling delays during ConnectionListener shutdown.

ReubenBond:
Thanks, @pentp, looks good to me. Where was the race?

pentp:
The loop that called `CloseAsync` for all connections and then awaited for their completion would go into a near-infinite loop if `Connection.Run` didn't get a chance to run its continuation first (which would remove itself from the dictionary).

But this retrying loop isn't needed anymore since #6955 made `RunAcceptLoop` add the connections to the dictionary itself and after it has finished there can be no new connections.

PR #6731, opened 2020-09-04
Fixes #6675: Update IEventHubDataAdapter to support StreamId to partition mapping (#6676)

### LINKED ISSUE #6675: Hot partitions due to PartitionKey used in Event Hub Stream Provider

benjaminpetit:
* Fixes #6675: Include stream Namespace in PartitionKey for Event Hub messages
* Update IEventHubDataAdapter to support partition to StreamId mapping

@alexmg

### DIFF
--- a/src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
+++ b/src/Azure/Orleans.Streaming.EventHubs/Providers/Streams/EventHub/EventHubAdapterFactory.cs
@@ -225,7 +225,8 @@ namespace Orleans.ServiceBus.Providers
             Dictionary<string, object> requestContext)
         {
             EventData eventData = this.dataAdapter.ToQueueMessage(streamGuid, streamNamespace, events, token, requestContext);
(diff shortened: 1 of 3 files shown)


alexmg:
Awesome! Thank you @benjaminpetit and @sergeybykov.

